PR #52378, opened 2026-08-07
[DRAFT] Gemma3 model optimizations

tvardhineniTT:
## Summary
- Ports Gemma3 decode/prefill performance optimizations to Blackhole (single- and multi-chip):
  tuned matmul program configs/grids, fused create-heads (`generic_op` grid-fill), concat-heads
  sharding, sharded prefill RMSNorm, and a traced path.
- Adds env-gated **on-device force-argmax sampling** (`TT_GEMMA3_ODS=1`) for multi-device greedy
  decode. Gemma3's ~262K vocab exceeds the 64K-per-device on-device-sampling limit; force-argmax
  argmaxes the full logits on-device after the split-logit all-gather, removing the per-token host
  round-trip. Minimal, opt-in change in `tt_transformers` (default off → no impact on other models).
- Keeps **27B (and any non-12B) on stock matmul configs** — the 12B-tuned LM-head/FF2/QKV grids
  deadlock/overflow on 27B's tile counts, so they're gated to `dim==3840`.
- Adds `PERF_COMMANDS.md` with exact repro commands, cache notes, and measured numbers.

## Results (ISL 128 / OSL 128, batch-1, text-only; decode = 1000/mean_tpot)

| Model | HW | Decode tok/s | TTFT | ms/token |
|---|---|---|---|---|
| Gemma3-12B | 1× P150 | 31.42 | 67.8 ms | 31.8 |
| Gemma3-12B | 2× P150 | 44.37 | 111.3 ms | 22.5 |
| Gemma3-27B | 2× P150 | 18.25 | 128.2 ms | 54.8 |
| Gemma3-27B | 4× P150 | 27.49 | 140.0 ms | 36.4 |

- 12B optimized from 21 → 31 (1×) and 29 → 45 (2×). 

## Accuracy
- Top-1/Top-5 token accuracy A/B-verified per optimization seam on 12B: Top-1 ~90–91.5%,
  Top-5 97–98%, all passing the reference check (shipped lm_head = BFP8).

## Test plan
- [ ] `performance-batch-1` decode re-verified for all 4 variants (commands in PERF_COMMANDS.md)
- [ ] `accuracy-batch-1` top1/top5 on 12B
- [ ] CI

### DIFF
--- a/models/demos/multimodal/gemma3/PERF_COMMANDS.md
+++ b/models/demos/multimodal/gemma3/PERF_COMMANDS.md
@@ -1,0 +1,103 @@
+# Gemma3 Performance Benchmark Commands (Blackhole)
+
+Repro commands for benchmarking Gemma3-12B / 27B decode throughput on Blackhole.
+All numbers below are **ISL 128 / OSL 128, batch-1**, traced, with on-device sampling (ODS)
+on multi-chip. Decode throughput is reported the apples-to-apples way we align with server CI:
+`decode tps = 1000 / mean_tpot`.
+
+## Common preamble
+
+```bash
+cd "$TT_METAL_HOME"        # tt-metal repo root
+source python_env/bin/activate
+export TT_METAL_HOME=$PWD PYTHONPATH=$PWD ARCH_NAME=blackhole
+MGD=$PWD/tt_metal/fabric/mesh_graph_descriptors      # mesh graph descriptor dir
+
+# Reset before each run. NOTE: reset is per-BOARD. A 2x P300 host has 2 boards
+# (indices 0 and 1), each with 2 chips -> `tt-smi -r 0 1` resets all 4 chips.
+tt-smi -r 0 1
+```
+
+## ISL / OSL
+
+- All `performance-batch-1` runs use `input_data_questions_prefill_128.json`.
+- Prefill is bucketed to the **128** sequence-length bucket, so **ISL = 128**.
+- **OSL** is `max_generated_tokens` in the `performance-batch-1` parametrize entry of
+  `text_demo.py` (stock default **200**). For an exact **ISL 128 / OSL 128** measurement, set
+  that entry to `max_generated_tokens=128` and `stop_at_eos=False` (fixed 128 decode steps).
+- Decode t/s/u is steady-state (per-token) and effectively independent of OSL; TTFT reflects
+  the ISL-128 prefill.
+
+---
+
+## TEXT-ONLY  (`text_demo.py`, `-k performance-batch-1`)
+
+### Gemma3-12B — 1 chip (P150)
+```bash
+TT_VISIBLE_DEVICES=0 MESH_DEVICE=P150 HF_MODEL=google/gemma-3-12b-it \
+TT_MESH_GRAPH_DESC_PATH=$MGD/p150_mesh_graph_descriptor.textproto \
+pytest models/demos/multimodal/gemma3/demo/text_demo.py -k "performance-batch-1" -s -q
+```
+
+### Gemma3-12B — 2 chips (1× P300) + ODS
+```bash
+TT_VISIBLE_DEVICES=0,1 MESH_DEVICE=P300 HF_MODEL=google/gemma-3-12b-it TT_GEMMA3_ODS=1 \
+TT_MESH_GRAPH_DESC_PATH=$MGD/p300_mesh_graph_descriptor.textproto \
+pytest models/demos/multimodal/gemma3/demo/text_demo.py -k "performance-batch-1" -s -q
+```
+
+### Gemma3-27B — 2 chips (1× P300) + ODS
+```bash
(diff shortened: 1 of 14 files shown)


mtairum:
Is this still a draft like the name implies, or ready for review?

yieldthought:
Codex review:

- [P1] [models/demos/multimodal/gemma3/tt/model_config.py:525](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/model_config.py#L525) caches each derived interleaved QKV weight in a module-global dictionary keyed only by `buffer_address()`. That numeric address is not a stable tensor identity: it can be reused after a model is destroyed, and the same address can exist on a different device. Because the dictionary is never cleared and retains only the converted tensor, a later model construction can see an existing key at [models/demos/multimodal/gemma3/tt/model_config.py:604](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/model_config.py#L604) and silently run with the previous model/device's QKV weights. Suggestion: make the converted weight instance-owned (for example on the `Attention` object) or key it by an owning tensor/device identity with an explicit lifetime.

- [P1] [models/demos/multimodal/gemma3/tt/pipeline.py:176](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/pipeline.py#L176) declares the prefill stage self-traced, which tells the perf adapter not to capture it, but Gemma returns an empty prefill trace allow-list at [models/demos/multimodal/gemma3/tt/model_config.py:2100](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/model_config.py#L2100). Consequently the `enable_trace=True` call still takes the eager branch selected at [models/demos/multimodal/gemma3/tt/gemma_multimodal_generator.py:315](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/gemma_multimodal_generator.py#L315). The reported `TRACE_STAGE_MS[prefill]` therefore measures an eager stage while labeling it trace+1CQ. Suggestion: either enable and verify a real supported prefill trace before declaring `prefill_self_traced`, or leave that flag false and let the harness treat the stage as eager.

- [P1] [models/tt_transformers/tt/model_config.py:1129](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/tt_transformers/tt/model_config.py#L1129) applies `TT_GEMMA3_ODS=1` in the shared base `ModelArgs` without checking the model family. Any non-Galaxy model constructed in that environment gets `allow_force_argmax=True`, and [models/tt_transformers/tt/model.py:164](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/tt_transformers/tt/model.py#L164) then treats force-argmax capability as general on-device-sampling support even when the normal per-device top-k width is unsupported. A later non-greedy request disables force argmax inside `TTSampling` and enters the path the 64K gate was meant to exclude; unrelated models in the same process are affected too. Suggestion: scope the option to Gemma and keep unsupported non-greedy requests on the host path.

- [P1] [models/tt_transformers/tt/model.py:157](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/tt_transformers/tt/model.py#L157) now conflicts with current `main` exactly at the sampling-support decision. Main uses padded vocab width and `TTSampling.num_single_device_vocab_splits()` to support large single-device vocabularies, while this branch replaces that logic with the older `vocab_size // sampling_splits` heuristic. Resolving the conflict by taking the PR side would regress main's large-vocab split support and reintroduce floor/unpadded width decisions. Suggestion: rebase and layer the Gemma force-argmax exception on top of main's current support calculation.

- [P1] [models/demos/multimodal/gemma3/tt/cpp_create_heads.py:93](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/models/demos/multimodal/gemma3/tt/cpp_create_heads.py#L93) lets the custom tile-addressing kernel run based only on rank, dtype, and element divisibility. The stock operation rejects non-device/non-TILE inputs at [ttnn/cpp/ttnn/operations/experimental/transformer/nlp_create_qkv_heads/device/nlp_create_qkv_heads_device_operation.cpp:21](https://github.com/tenstorrent/tt-metal/blob/bad06ce8611b1b870ca0991386789adb98620c93/ttnn/cpp/ttnn/operations/experimental/transformer/nlp_create_qkv_heads/device/nlp_create_qkv_heads_device_operation.cpp#L21), but the import-time wrapper intercepts before that validator. A row-major tensor or a TILE tensor with non-32x32 geometry can therefore be interpreted using hard-coded 32x32 page math and silently produce wrong heads. Suggestion: enforce device storage, allocation, TILE layout, standard tile geometry, and the supported memory-layout contract in `can_run`, falling through to the stock op otherwise.

PR #52380, opened 2026-08-07
Generalize multi_scale_deformable_attn to support D values that are multiples of 16

### LINKED ISSUE #52328: [Bounty $1,500] Generalize multi_scale_deformable_attn to support D values that are multiples of 16
## Summary

`ttnn.experimental.multi_scale_deformable_attn` hard-requires the value tensor's last dimension to
be exactly `D = 32`. Any other width is rejected at validation. The source carries a `TODO`
describing the intended generalization, and the restriction is a kernel-layout assumption rather
than a hardware or algorithmic limit.

This is a forward-looking generality improvement, not an active model blocker. See
[Scope and motivation](#scope-and-motivation-stated-plainly) for an honest statement of impact.

## Current behavior

`ttnn/cpp/ttnn/operations/experimental/multi_scale_deformable_attn/device/multi_scale_deformable_attn_device_operation.cpp:68-74`

```cpp
// TODO: generalize the kernel to support arbitrary D (multiple of 16). The
// reader/writer currently assume D=32: a single tile row is split into
// exactly two 32-byte halves placed in TL+TR (or BL+BR) faces, and
// HALF_STICK_NBYTES is hardcoded to 32 in the kernels. Supporting D=64
// or D=16 means deriving the per-row layout from element_size + D and
// looping over multiple (half-)faces per row.
TT_FATAL(vs[-1] == 32, "value's last dim (D) must be 32, got {}", static_cast<uint32_t>(vs[-1]));
```

`TT_FATAL` is enforced in all build configurations, so this is a hard failure rather than a
debug-only check.

Test coverage reflects the same constraint: the correctness test parametrizes `D` with `[32]` only,
so there is no coverage for other valid widths.

## What already works

The restriction is narrower than it first appears. The following are already in place, and a
contributor should not spend time rebuilding them:

- **The program factory is already D-parametric.** `D` is passed to the kernels as a compile-time
  argument.
- **Value and output stick sizes are already derived from `D`**, computed as `D * element_size`
  rather than as a fixed constant. The scratch circular buffers size off those values.
- **The tile-layout helper (`msda_tile_layout.hpp`) is D-agnostic.** It is pure face geometry and
  needs no changes.
- **The NoC read path already exposes the offset lever needed.** The existing `noc_async_read` call
  takes an explicit byte offset, currently passed as `0`.

One clarification that has caused confusion before: the constant `TILE_MAX_ROWS = 32` in this op
refers to the **query** dimension, not `D`. It is unrelated to this restriction.

## Scope of work

The remaining `D = 32` assumption lives in the two dataflow kernels:

- `device/kernels/dataflow/reader_msda.cpp`
- `device/kernels/dataflow/writer_msda.cpp`

Specifically:

1. **`HALF_STICK_NBYTES` is hardcoded to 32** in both kernels. It should be derived from
   `D * element_size` instead.
2. **Two copy loops in each kernel assume a row splits into exactly two 32-byte halves** placed in
   the TL+TR (or BL+BR) faces. These need to loop over the correct number of (half-)faces per row,
   derived from `D`.
3. **One loop-nesting change** is needed so a row can span more than one face pair, which is what
   `D = 64` requires.
4. **Relax the validation** at `multi_scale_deformable_attn_device_operation.cpp:74` from
   `vs[-1] == 32` to `vs[-1] % 16 == 0`, with a clear error message naming the supported rule.

`D = 16` is the smallest useful increment of work. It fits within a single face pair and requires
only items 1 and 4 plus a narrowed copy length. `D = 64` exercises the loop-nesting change in
item 3.

## Acceptance criteria

- [ ] `D` values of 16, 32 and 64 produce numerically correct output, validated against the existing
      PyTorch reference at the same PCC threshold used for `D = 32` today.
- [ ] The existing `D = 32` path is unchanged in behavior and performance.
- [ ] Validation rejects non-multiples of 16 with an actionable error message.
- [ ] The correctness test's `D` parametrization is extended from `[32]` to at least `[16, 32, 64]`.
- [ ] No regression in the existing multi_scale_deformable_attn test suite.

## Testing notes

The test already parametrizes `D`, so extending coverage is close to free. Adding values to the
existing parametrize list is sufficient, and no new test scaffolding is required.

Correctness is the bar for v1. Performance parity with the existing `D = 32` path is expected only
in the sense that the `D = 32` case must not regress.

## Scope and motivation, stated plainly

This is **not** currently blocking a model bring-up. The model referenced in #46067,
DINO-5scale-Swin-L, uses 256 embedding dims across 8 heads, giving `D = 32`, so it is already served
by the current path.

The value here is forward-looking. Deformable-attention variants with different head counts produce
different per-head widths, and today each such configuration hits a hard failure with no fallback.
Generalizing `D` removes a shape-specific cliff from an operator that is otherwise general.

Effort is estimated at a few days rather than weeks, given how much of the parametrization already
exists.

## Related issues

- **#46067**: multi_scale_deformable_attn tracking issue (open). Does not mention the `D` restriction.
- **#35746**: [BEVFormer] Implement Multi-Scale Deformable Attention (open).
- The operator landed via PR **#45411**.

AJ0070:
### Ticket
Closes #52328

### Problem description

`ttnn.experimental.multi_scale_deformable_attn` hard-required the value tensor's last dimension to be exactly `D = 32`, enforced by a `TT_FATAL` in validation. The reader and writer kernels assumed one value stick fills exactly the lo+hi face halves of a single tile row, with `HALF_STICK_NBYTES` hardcoded to 32.

### What's changed

The per-row face layout is now derived from `D` instead of being hardcoded. A value stick spans `ceil(D / 32)` tiles laid side by side, with the trailing tile half-filled when `D % 32 == 16`.

- **reader_msda.cpp**: the stick scatter loops over `N_D_TILES` d-tiles per (p, corner), computing per-tile lo/hi word counts from `D`. Word counts are derived from the `D` compile-time arg rather than the alignment-padded stick byte size, so DRAM alignment padding never enters the face math.
- **writer_msda.cpp**: the row gather mirrors the scatter across d-tiles; `D` added as a compile-time arg.
- **msda_compute.cpp**: per reduction group, waits for `n_d_tiles` input tiles plus the single shared scalar tile (the combined attn x bilinear weight is per query row and independent of D), multiply-accumulates each d-tile into its own output slot via `pack_tile<true>(0, cb, k)` with the existing L1-accumulate scheme.
- **program factory**: computes `n_d_tiles`, sizes the input/output tile CBs at `2 * n_d_tiles` pages (double-buffered blocks), and passes the new compile-time args.
- **validation**: relaxed from `D == 32` to `D % 16 == 0` with an actionable message.

The D=32 path is unchanged in behavior: for D=32 the derived constants collapse to exactly the previous values (`LO_WORDS = HI_WORDS = 8`, `N_D_TILES = 1`).

Unused lanes of a trailing half-filled d-tile follow the op's existing stale-bytes contract (the writer never reads those lanes back), documented in the reader comments.

### Checklist (results from N300, Wormhole)

- [x] D=32 regression gate on this branch: **32/32 passed**
- [x] Full correctness matrix, D in {16, 32, 48, 64} x (N, h/w, Q, P, align_corners): **128/128 passed**
- [x] Negative validation tests (D = 8, 24, 40 rejected with "multiple of 16" error): **3/3 passed**
- [x] Baseline sanity: prebuilt wheel on the same board, D=32: 32/32 passed

Test environment: single N300 (both chips visible), Ubuntu 22.04, source build of this branch, `build_metal.sh --release`.

### Notes for reviewers

- The scalar tile is shared across all d-tiles of a reduction group; only the input tile count scales with D. This keeps the reader's NoC read count per (p, corner) unchanged (the extra scatter passes reuse the already-staged sticks in L1 scratch).
- `TILE_MAX_ROWS = 32` in this op refers to the query dimension and is unrelated to D.
- Any D that is a positive multiple of 16 is accepted; L1 usage grows with `2 * ceil(D/32)` tiles per CB, which stays trivial for realistic D.


AJ0070:
@jberkowitzTT Can you please re-trigger CI?

AJ0070:
@jberkowitzTT CI is green!

AJ0070:
@jberkowitzTT Can you please ping someone to review the code?

jberkowitzTT:
Getting engineering review soon

bbradelTT (approved):

PR #52384, opened 2026-08-07
[Feature] Tensor prefetcher: mcast-in0 on K-row-major weights, and stop requiring a page-multiple GCB size

jbaumanTT:
### PR Category
Feature, Cleanup

### Summary
Initial mcast-in0 support for the Tensor prefetcher carried three restrictions the mechanism underneath it does not have.

**Legacy K-row-major weights were rejected.** The mcast geometry check delegated to the receiver-contiguous validator, which requires an `NdShardSpec`, and the GCB factory refused any mcast config on a `WIDTH_SHARDED` weight outright. Nothing below that needed the restriction: the sender divides a bank's shard into whatever `block_count` the request names (`compute_tensor_layout_krow_major`) and slices each block across that bank's receivers, so the per-receiver page stream is the same one mcast already reads off a receiver-contiguous weight. Production models carry K-row-major weights, so feeding one to an mcast-in0 matmul previously meant reallocating it.

`validate_krow_major_weight_for_matmul_1d` is the K-row-major counterpart to the recv-contig validator, so each layout's weight↔matmul contract lives in exactly one place, and `tensor_prefetcher_block_count_for_matmul_1d` dispatches to it on the detected layout. `build_matmul_1d_gcb_krow_major` delegates its per-weight checks there rather than inlining them, accepts mcast configs (grid capacity ≥ receiver count, as the recv-contig builder already allowed), and derives the GCB page from the consumer's block count instead of assuming a ring.

Streaming (`stream_in1`) stays receiver-contiguous-only — it needs a per-receiver rotation, which only that layout supports.

**A GCB size that wasn't a whole number of in1 K-block pages was fatal.** The mcast path asserted `gcb.size() % in1_block_size == 0` in both the device op and the program factory, forcing callers to know the op's page size to allocate a buffer. The gather path has always floored instead (`(size / page) * page`), so the mcast CB now floors the same way and the surviving check is the one that matters: the window must hold at least two whole pages.

**The GCB size floor was computed across configs instead of per config.** It took the largest page over all configs and multiplied it by the largest block count over all configs. Those two maxima can come from different configs, so a GCB shared by a batched gather (small page, whole layer resident) and an mcast (large page, two-page window) was held to a size neither consumer asked for — and the product can exceed the 2 MB remote-CB cap, making a mix unconstructible that each consumer fits individually. Each `(config, weight)` pair is now checked against its own window as the builders walk them (`validate_gcb_size_for_config`); the cap is size-independent and stays a single check after the loop. The floor only ever drops, so nothing that passed before starts failing.

One guard goes the other way. The K-row-major sender slices its bank's shard by a single receivers-per-bank number, so `tensor_prefetcher_block_count_for_matmul_1d` now requires every GCB sender to drive exactly `num_global_cb_receivers` receivers. The manager rejects split and partial-bank sender topologies and requires the receiver total to divide evenly across banks, but an uneven fan-out that still averages out — `(1, 3, 2, 2)` over four banks — clears both of those and then gets sliced at the average, silently delivering the wrong pages. `build_matmul_1d_gcb_krow_major` already enforces uniformity where it builds the mapping; the exposure was a hand-built GCB reaching a matmul through this entry point.

Docs-wise this also corrects `prefetcher_matmul_design.md`, which stated the layout restriction as contract.

### Notes for reviewers
- **The layout restriction came down to which validator the mcast path happened to call** — worth confirming that reading. The claim is that `compute_tensor_layout_krow_major` already accepts an arbitrary per-tensor `block_count` (it ceil-divides K by it), so nothing under the factory rejected legacy weights. The mcast test pins this: `block_count` is `K_tiles / in0_block_w`, deliberately *different* from the receiver count, so a ring-derived block count would deliver the wrong pages rather than silently working.
- `build_matmul_1d_gcb_krow_major` is the biggest diff, but most of it is deletion — ~75 lines of inlined per-weight checks replaced by the shared validator call. The gather path through it should be behavior-identical, size floor included: a batched gather's `block_count` is `ring_size`, so its own floor is the same `max_page * ring_size` the cross-config rule produced for a gather-only GCB.
- The gather-vs-mcast grid rule is checked per config rather than once from `program_configs[0]`, matching the recv-contig builder, so a GCB shared by a gather and an mcast matmul gets the right rule for each.
- `in1_remote_cb_size` in the mcast factory is hoisted out of the `if` only so the existing `log_debug` reports the floored size rather than the raw GCB size.
- The per-sender fan-out check is a no-op for any GCB built through `create_global_circular_buffer_for_matmul_1d`. It only reaches callers who hand-build one through the generic `create_global_circular_buffer_for_tensor_prefetcher`.

### Test coverage
- `test_tensor_prefetcher_mcast_in0[weight_layout × gcb_size_misalign_bytes]` — mcast PCC on all three weight layouts (`krow_major`, `recv_contig_strided`, `recv_contig_contiguous`), each at an exact page multiple and at a ragged size (`4 * page + 64`), asserting the derived block count on the way through.
- `test_create_global_circular_buffer_for_matmul_1d_sizes_per_config` — a batched gather sharing a GCB with an mcast whose page is 16× larger. Asserts that the cross-config product exceeds what either consumer needs, that the real floor is accepted, and that one byte under it fails naming the config that needs the larger window.
- `test_tensor_prefetcher_block_count_rejects_uneven_krow_major_fanout` — a hand-built GCB with `(1, 3, 2, 2, …)` fan-out, same receiver total as a uniform mapping, is rejected.

On p150 (7 DRAM banks, harvested):
- `test_prefetcher_BH_tensor_large.py`: 56 passed, 3 skipped (pre-existing "expects 8 unharvested DRAM banks" gate)
- `test_prefetcher_BH_validator.py`: 27 passed, 3 skipped (same gate)

One existing test needed a change: `test_..._rejects_undersized` matches the assert message on the literal `num_blocks`, and the reworded floor message keeps that token deliberately so the test still covers the case it was written for.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jbauman/mcast-in0-krow-major-gcb)

### DIFF
--- a/ttnn/core/global_circular_buffer.cpp
+++ b/ttnn/core/global_circular_buffer.cpp
@@ -381,7 +381,8 @@ uint32_t validate_krow_major_weight_for_matmul_1d(
         // has always required K to divide by it, so keep rejecting exactly the configs it always did.
         TT_FATAL(
             weight_K_tiles % program_config.in0_block_w == 0,
-            "weight K ({} tiles) must be divisible by in0_block_w ({})",
+            "legacy K-row-major gather still requires weight K ({} tiles) divisible by config in0_block_w ({}); "
+            "the matmul derives K-block width from K/ring_size, not this field",
             weight_K_tiles,
             program_config.in0_block_w);
     }
--- a/ttnn/core/global_circular_buffer.cpp
+++ b/ttnn/core/global_circular_buffer.cpp
@@ -4,7 +4,6 @@
 
 #include "ttnn/global_circular_buffer.hpp"
 
-#include <algorithm>
 #include <memory>
 #include <tt_stl/assert.hpp>
 #include <tt-metalium/buffer.hpp>
@@ -149,14 +148,42 @@ void validate_grid_for_consumer(
     }
 }
 
-// Shared GCB size validation: the caller-supplied size must clear the floor its consumers need — a
-// full layer for a batched gather matmul (it does wait_front(num_blocks)), or a double-buffered
-// window when every consumer sharing the GCB reads K-blocks FIFO — and fit the remote-CB cap.
+// One consumer's own floor on the GCB window: a batched gather matmul does wait_front(block_count)
+// and needs a whole layer resident, while a FIFO consumer takes K-blocks as they land and needs only
+// a double-buffered window.
+//
+// Checked per (config, weight) as the builders walk them, rather than once against the largest page
+// and the largest block count over all configs — those two maxima can come from different configs,
+// and their product demands a buffer neither consumer asked for. A GCB shared by a batched gather
+// (small page, whole layer) and an mcast (large page, two blocks) is the case that bites: the
+// product can exceed kMaxCbPagesBytes and make the mix unconstructible even though each consumer
+// fits on its own.
 //
 // No L1 budget check here — receivers may have very different L1 usage on top of the GCB (matmul
 // in0/in1/out/interm CBs etc.) and we don't have enough context at the factory to compute a real cap.
 // Callers must size the GCB to fit their own L1.
-//
+void validate_gcb_size_for_config(
+    uint32_t size,
+    const ttnn::operations::matmul::MatmulMultiCoreReuseMultiCast1DProgramConfig& config,
+    size_t config_index,
+    uint32_t page_bytes,
+    uint32_t block_count) {
+    const bool fifo = is_fifo_consumer(config);
+    const uint32_t num_blocks = fifo ? kFifoMinWindowBlocks : block_count;
+    const uint32_t min_size = page_bytes * num_blocks;
+    TT_FATAL(
+        size >= min_size,
+        "GCB size ({} B) must be at least num_blocks * page ({} * {} = {} B) for program_configs[{}]. {}",
+        size,
+        num_blocks,
+        page_bytes,
+        min_size,
+        config_index,
+        fifo ? "A FIFO consumer takes K-blocks as they arrive but still needs a double-buffered window."
+             : "A batched gather matmul does wait_front(num_blocks), so it needs that many pages "
+               "buffered before it consumes.");
+}
+
 // kMaxCbPagesBytes is a cap on fifo_aligned_num_pages = fifo_size /
 // REMOTE_CIRCULAR_BUFFER_ALIGNED_PAGE_SIZE. Two reasons it exists:
 //   1. The NoC stream overlay's STREAM_REMOTE_DEST_BUF_SIZE register holds the buffer size in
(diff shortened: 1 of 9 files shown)


bbradelTT (approved):

abhullar-tt (approved):

ttnn/core/global_circular_buffer.cpp
```diff
@@ -383,6 +289,136 @@ uint32_t validate_recv_contig_weight_for_matmul_1d(
...
+    const uint32_t block_count = block_count_for_consumer(program_config, weight_K_tiles, receiver_count);
+    if (program_config.gather_in0) {
+        // Legacy-only extra guard, predating this layout's mcast consumer: gather ignores
+        // cfg.in0_block_w (it derives the K-block width from K/ring_size), but the K-row-major builder
+        // has always required K to divide by it, so keep rejecting exactly the configs it always did.
+        TT_FATAL(
+            weight_K_tiles % program_config.in0_block_w == 0,
+            "weight K ({} tiles) must be divisible by in0_block_w ({})",
+            weight_K_tiles,
+            program_config.in0_block_w);
```
vtsilytskyiTT, line 387:
```suggestion
TT_FATAL(
    weight_K_tiles % program_config.in0_block_w == 0,
    "legacy K-row-major gather still requires weight K ({} tiles) divisible by config in0_block_w ({}); "
    "the matmul derives K-block width from K/ring_size, not this field",
    weight_K_tiles,
    program_config.in0_block_w);
```

vtsilytskyiTT (approved):
Looks good. GCB became more structured.

ttnn/core/global_circular_buffer.cpp
jbaumanTT: Done.

PR #50947, opened 2026-07-23
Feat: Query Class For FSD Hierarchy Fields

agupta-TT:
### PR Category
Feature

### Summary

Closes #50788 

### Notes for reviewers
Continues instance_path work of #50787 to make instance paths queryable. Use Longest Common Path to quantify relation between 2 hosts, and thus be used in  determining hierarchy level of a connection between 2 hosts.

Also some filtering fixes/adjustments to warn on edge/rare cases

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:agupta/fsd-query-instance-path)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:agupta/fsd-query-instance-path)

PR #52385, opened 2026-08-07
tools/scaleout: split source-file lists into sources.cmake

afuller-TT:
## Why

Raised by @afuller-TT: `tools/scaleout` hasn't yet received the `sources.cmake` treatment that `tt_metal` and `ttnn` already have, so routine PRs touching scaleout tooling end up pinging metal infra for no reason. Concrete example: #50947 only needed to register a new `factory_system_descriptor/query.{cpp,hpp}` pair, but had to touch `tools/scaleout/CMakeLists.txt` to do it.

`tools/**/scaleout/**/CMakeLists.txt` is CODEOWNERS-gated to the scaleout owners **plus** `@tenstorrent/metalium-developers-infra`. `sources.cmake` is a different filename, so it isn't matched by that rule at all — it falls through to the broader `tools/**/scaleout/` entry, which is already scaleout-owned with no infra co-owner. No CODEOWNERS changes needed; the ownership fix is entirely a side effect of the file split (same reasoning already used for `tt_metal/tools/watcher_dump` and friends).

## What changed

- **`tools/scaleout/CMakeLists.txt`**: moved every literal source list (the `scaleout_tools` object library's public/private files, the proto schema list, and every executable's/library's source file(s): `run_cluster_validation`, `run_fabric_manager`, `2d_big_mesh_cabling_gen`, `run_cabling_generator`, `run_regen_descriptors`, `generate_cluster_descriptor`, `generate_mgd_lib`, `generate_mgd`, `generate_rank_bindings`) into a new `tools/scaleout/sources.cmake`. `CMakeLists.txt` now only does `add_executable(name)` / `add_library(name TYPE)` + `target_sources(name PRIVATE ${VAR})`, matching the pattern already used in `ttnn/CMakeLists.txt` and `tt_metal/tools/watcher_dump/CMakeLists.txt`.
- **`tools/tests/scaleout/CMakeLists.txt`**: same treatment for its six test executables, including `test_link_retraining`'s multi-file source list (which reaches into `tools/scaleout/validation/utils/`).

No behavior change — every `set(...)` list is a verbatim move of the file paths that were previously inline, just renamed to `ALL_CAPS` variables (e.g. `SCALEOUT_TOOLS_SRCS`, `RUN_CLUSTER_VALIDATION_SRCS`). Each new variable is defined exactly once in `sources.cmake` and referenced exactly once in `CMakeLists.txt` (checked mechanically, not just by eye).

## Verification

No build environment available in my sandbox (the checkout here doesn't have submodules initialized, so `cmake` configure fails before it ever reaches these files) — please confirm CI is green before merging. I did verify structurally: diffed the new `sources.cmake` file lists against the exact strings removed from `CMakeLists.txt` (no paths added/dropped/reordered-in-a-meaningful-way), and confirmed every new variable name is unique and used exactly once.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:afuller/split-tools-cmakes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:afuller/split-tools-cmakes)


afuller-TT:
Cleaned up per feedback: dropped the redundant `CMAKE_CURRENT_SOURCE_DIR` prefix from `sources.cmake` entries so it no longer drowns out an actually-suspicious cross-directory reference.

One nuance worth flagging: `SCALEOUT_PROTO_SCHEMAS` still needs an absolute path, since those entries feed a custom build command (`GENERATE_PROTO_FILES`/`add_custom_command`) whose relative-path arguments resolve against the *build* dir, not the source dir — confirmed with an isolated repro before touching it. Kept that list itself as plain relative paths in `sources.cmake` and moved the `list(TRANSFORM ... PREPEND /)` into `CMakeLists.txt` right where it's consumed, with a comment explaining why.

Also added a FIXME on the `tools/tests/scaleout` `PROJECT_SOURCE_DIR` reach into `tools/scaleout/validation/utils/` — left the reach itself in place since removing it would mean duplicating those source files, but flagged it as the bad practice it is.

nsextonTT (approved):

Riddy21 (approved):

PR #52386, opened 2026-08-07
Fix build-wrapper.yaml fetch-depth type error on workflow_dispatch

afuller-TT:
## Summary
- `build-wrapper.yaml` passes `fetch-depth: ${{ inputs.fetch-depth }}` into `build-artifact.yaml`'s number-typed `fetch-depth` input, but `fetch-depth` is only declared under the `workflow_call` trigger, not `workflow_dispatch`.
- On manual dispatch, `inputs.fetch-depth` is undefined and GHA coerces it to `''`, which the reusable-workflow input evaluator rejects against `type: number` ("Unexpected value ''" at line 135, col 20), killing the whole run before any build job starts.
- Fixed both occurrences (`build-dispatch` job and the matrix `build` job) with `${{ inputs.fetch-depth || 500 }}`, matching the existing `||` fallback pattern already used in this file for `harbor-prefix`/`build-type`/`enable-lto`.

## Test plan
- [x] Dispatched `build-wrapper.yaml` manually on this branch — run completed `success` (previously every manual dispatch failed in ~1m45s before any build started): https://github.com/tenstorrent/tt-metal/actions/runs/31143905298

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:afuller/fix-build-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:afuller/fix-build-workflow)

### DIFF
--- a/.github/workflows/build-wrapper.yaml
+++ b/.github/workflows/build-wrapper.yaml
@@ -132,7 +132,7 @@ jobs:
       tracy: true
       runs-on: ${{ inputs['runs-on'] }}
       ref: ${{ inputs.ref }}
-      fetch-depth: ${{ inputs.fetch-depth }}
+      fetch-depth: ${{ inputs.fetch-depth || 500 }}
 
   # build job runs the matrix - either when called from merge-gate with pre-built tags,
   # or when called via workflow_call without the platform dropdown
@@ -177,7 +177,7 @@ jobs:
       tracy: true
       runs-on: ${{ inputs['runs-on'] }}
       ref: ${{ inputs.ref }}
-      fetch-depth: ${{ inputs.fetch-depth }}
+      fetch-depth: ${{ inputs.fetch-depth || 500 }}
       # Pass pre-built docker image tags based on platform
       ci-build-docker-image: ${{
         (matrix.config.platform == 'Ubuntu 24.04' && inputs.ubuntu-2404-ci-build-tag) ||
--- a/.github/workflows/build-wrapper.yaml
+++ b/.github/workflows/build-wrapper.yaml
@@ -95,6 +95,11 @@ on:
         type: boolean
         default: false
         description: "Enable Link Time Optimization (LTO)"
+      fetch-depth:
+        required: false
+        type: string
+        default: "500"
+        description: "Git fetch depth passed through to build-artifact (0 fetches full history)"
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


afuller-TT:
Good catch — fixed in 34b8e62. Declared `fetch-depth` as a proper `workflow_dispatch` input (string, default `"500"`) instead of relying on a `||` fallback, so `inputs.fetch-depth` is always defined and an explicit `fetch-depth: 0` from a `workflow_call` caller now passes through unchanged instead of being coerced to 500.

nsextonTT (approved):

afuller-TT:
Update: the `fetch-depth` string-input approach in my last comment turned out to hit a known GitHub Actions runner bug ([actions/runner#2848](https://github.com/actions/runner/issues/2848)) — forwarding a `workflow_dispatch`-sourced value into a reusable workflow's `type: number` input fails with `Unexpected value` regardless of how it's typed on the caller side.

Final fix (3162266): branch on `github.event_name` instead of on the input value at all — `workflow_dispatch` runs get the literal number `500`, everything else passes `inputs.fetch-depth` through unchanged (already a proper number via `workflow_call`'s typed default, so an explicit `fetch-depth: 0` caller is preserved). This never forwards a `workflow_dispatch`-sourced value into the reusable call's number input, so the runner bug path is never hit.

Re-verified with a live dispatch: https://github.com/tenstorrent/tt-metal/actions/runs/31145742168 — `build-dispatch` compiled successfully.

PR #52413, opened 2026-08-07
DFlash prefill: multi-turn + multi-user drafter context-KV

### LINKED ISSUE #50725: Dflash prefill: Padding aware and block cyclic KV cache layout
### Is your feature request related to a problem? Please describe

Currently, Dflash is implemented to run as 5k one shot prefill.

### Describe the solution you'd like

Implement Dflash that is aware of the padding.
Decide on the KV cache layout (align with the decode) - it should be block-cyclic previously seen in chunked prefill (ND shared layout).

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

nbabinTT:
### PR Category
Feature

### Summary
The DFlash drafter's context-KV path landed in #52802 and runs during Kimi prefill, but only for one shape: a single request, a single chunk, starting at position 0. That is enough to validate the drafter in isolation, not to serve with it. This PR brings the drafter cache up to the same footing as the verifier's KVPE cache — deep enough for a full sequence, wide enough for many users, and writable at an arbitrary offset — which is exactly what a second conversation turn (resume at 5000, not 0) and concurrent users need. It all stays behind the same `PREFILL_DFLASH` flag; `num_users=1` reproduces the old single-slot shape and `PREFILL_PRODUCER_MULTI_TURN_PROB=0` the old producer behaviour, so nothing changes when the new paths are unused.

The work is mostly reuse rather than new machinery. Offset and per-user writes go through the same `update_padded_kv_cache` op and user-major slot layout (`slot = user * num_layers + layer`) the MLA chunked-prefill path already uses; the cache is allocated deep (`max_seq_len × num_users`) and — importantly — ND-DRAM-sharded like KVPE rather than interleaved. The migration address table emits one address per (bank, 32-token block), which only describes the buffer when each block is contiguous in a single bank, so this is a correctness requirement, not tuning (it also clears the standing `ND_DRAM_SHARDED` TODO). The drafter keeps its native Qwen3 half-split RoPE, with the per-chunk cos/sin gathered on the host from the writer's block-cyclic staircase, so the stored-K format decode consumes is unchanged.

The drafter's K/V then join the verifier's merged KV chunk-address table (one config per K/V per kv-head), so the same migration hand-off can describe them, and two golden-trace PCC gates check the result — one reading the device cache directly, one reading it back through the published table over UMD, the exact path migration runs on.

Out of scope here: decode-side pad masking (the drafter's context attention is unmasked, so that is a decode-time correctness concern — prefill's part is handing decode the valid length) and the `tt-llm-engine` consumer itself, which is not vendored in this repo, so migration is validated up to the published table, not the downstream pull. Builds on #52802. Closes #50725.

### What changed
- **Drafter** (`tt_dflash_drafter.py`) — `forward` takes `kv_actual_global` + `slot_idx` and writes through `update_padded_kv_cache` (mesh-aware, per-chip offset derived on device) in place of `fill_cache_for_user_`; the per-chunk RoPE table is gathered from the writer's block-cyclic staircase (`rotated_chip_positions`), keeping the half-split `rotary_embedding_hf` convention.
- **Cache** (`utils/kv_cache_utils.py`) — `allocate_dflash_kv_cache` gains `num_users`, allocates to full `max_seq_len` depth in user-major slots, and moves to the ND-DRAM-sharded spec KVPE uses (resolves the `ND_DRAM_SHARDED` TODO).
- **Migration** (`runners/kv_chunk_table.py`) — new `populate_kv_chunk_address_table_dflash` + `dflash_config_name`; the drafter K/V join the merged table as `2 · num_kv_heads` further configs, with single-member device groups (heads are TP-sharded, not replicated) and an explicit strided bank walk instead of the contiguous counter.
- **Validation** (new `dflash_prefill/dflash_kv_validation.py`) — `dflash_kv_cache_pcc_check` (device readback) and `dflash_kv_table_pcc_check` (through the published table over UMD); V is checked before K so a V-pass/K-fail localizes any fault to the rope table.
- **Runtime** (`tt_prefill_runtime.py`) — allocates the cache at `max_seq_len × num_users`, drives the drafter with the verifier's `(offset, slot)`, registers the drafter caches into the merged table on the last rank / single-stage path, and adds an optional bring-up PCC hook.
- **Producer** (`common/prefill/runners/prefill_producer.py`, `prefill_runner.py`) — per-slot `prefix_len` + `turn_idx`, the `PREFILL_PRODUCER_MULTI_TURN_PROB` knob, and an absolute `real_len` (a resumed turn starts mid-cache, so chunk counts no longer describe it); the runner records per-slot `real_start`/`real_end`.
- **Config** (`dflash_drafter_config.py`) — `num_target_layers` is now `int | None = None` (fail-closed) so a checkpoint that omits it cannot silently pass the verifier cross-check.
- **Tests + CI** — `test_dflash.py` gains a multi-turn case and offset/straddle params; new `test_kv_cache_table.py` round-trips the merged table; the trace-table test covers the drafter configs.

### Testing (Blackhole)
- `test_dflash.py` on an 8×4 Galaxy — `test_dflash_multiturn_pcc` streams several turns and PCCs the drafter's K/V against the real HF `DFlashDraftModel`; added `offset` / `midchip_straddle` params exercise non-zero and tile-straddling write offsets (e.g. `offset=32`).
- `test_kv_cache_table.py` (new) — builds, serializes, and reads back the merged chunk-address table including the drafter's per-(K/V, kv-head) configs.
- Runner golden-PCC gates (`dflash_kv_validation`) — device-cache readback and through-the-table (UMD) readback, the migration path, wired into the runner and run when a golden trace is present.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nbabin/dflash_multi_turn)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nbabin/dflash_multi_turn)


models/demos/common/prefill/docs/PREFILL_MIGRATION_TESTING.md
```diff
@@ -234,6 +234,23 @@ crossed slot→prompt wiring or a slot-index bug in the address table still pass
...
 prompt; set `max_requests == num_users` for one resident request per slot.
 
+### Multi-turn conversations
+
+`PREFILL_PRODUCER_MULTI_TURN_PROB` (manifest: `workload.multi_turn_prob`, default `0.0`) is the probability
+that a recycled slot **continues** its conversation instead of starting a fresh one. A continued turn
+resumes writing at the previous turn's length **aligned down to 32** and replays the ≤31 dropped tokens as
+part of its first chunk; aligning up instead would leave a permanent unwritten hole mid-sequence, and a
+sub-tile write offset is rejected outright (`update_padded_kv_cache` asserts `kv_actual_global % 32 == 0`,
+and the kernel's staircase disagrees with the host mirror off-tile). The replay is PCC-idempotent because
```
pavlepopovic, line 241: we can use this for regular (not just dflash) runner testing?

models/demos/common/prefill/runners/prefill_producer.py
```diff
@@ -941,33 +1005,74 @@ def _read_slot_kv_and_check_pcc_mla(table, device_map: dict, slot_id: int, real_
...
+        if check_dflash:
+            # None when no golden is configured (the drafter golden is prompt-specific, so opt-in).
+            dflash_pcc = dflash_kv_table_pcc_check(
+                kv_table, slot_id, real_len, read_config_slice=read_dflash_slice, threshold=dflash_threshold
+            )
+            if dflash_pcc is not None:
+                min_dflash_overall = dflash_pcc if min_dflash_overall is None else min(min_dflash_overall, dflash_pcc)
+                if dflash_pcc < dflash_threshold:
+                    dflash_failures.append((slot_id, real_len, dflash_pcc))
+
```
pavlepopovic, line 1071: with mtp and potentially dflash for other models coming (it might be different to this one?), this could get messy, could this be abstracted away a little ? So we call something like model.check_kv_cache() and it internally decides how many kv caches are there and how to compare them ? doesn't have to be part of this PR but could be related to https://github.com/tenstorrent/tt-metal/issues/53336

models/demos/deepseek_v3_d_p/tests/test_kv_cache_table.py
```diff
@@ -441,6 +442,166 @@ def test_kimi_kv_cache_mock(
...
+                num_kv_heads=num_kv_heads,
+                head_idx=head_idx,
+                num_users=num_users,
+                config_id=base + head_idx,
+            )
+
+    def bf8_round(reference):
+        return ttnn.to_torch(ttnn.from_torch(reference, dtype=ttnn.bfloat8_b, layout=ttnn.TILE_LAYOUT)).to(
+            torch.bfloat16
+        )
```
pavlepopovic: pls add a comment what this does and why :)

models/demos/deepseek_v3_d_p/tt/runners/kv_chunk_table.py
```diff
@@ -111,21 +131,25 @@ def build_and_serialize_kv_chunk_table(
 
     primary_cache = kvpe_cache.storage
     all_caches = (primary_cache,) + ((index_kv_cache,) if index_kv_cache is not None else ())
```
pavlepopovic: can all_caches also contain drafter_cache?

models/demos/deepseek_v3_d_p/tt/runners/kv_chunk_table.py
```diff
@@ -156,54 +180,102 @@ def _builder(*, config, chunk_size_bytes, num_users):
...
+                seq_len=seq_len,
+                sp_axis=sp_axis,
+                tt_kvpe_cache=cache,
+                chunk_size_bytes=cfg.chunk_size_bytes,
+                num_users=num_users,
+                config_id=config_id,
+                stage_layout=stage_layout,
+            )
+        else:  # one global kv-head of the drafter's K or V cache
+            populate_kv_chunk_address_table_dflash(
```
pavlepopovic, line 280: again can this also be delegated to model adapter to decide what to create?

models/demos/common/prefill/docs/PREFILL_MIGRATION_TESTING.md
nbabinTT: Yes, we can use it for the runner test

models/demos/common/prefill/runners/prefill_producer.py
nbabinTT: We will have to address it in https://github.com/tenstorrent/tt-metal/issues/53336

models/demos/deepseek_v3_d_p/tt/runners/kv_chunk_table.py
nbabinTT: should be addressed as the part of https://github.com/tenstorrent/tt-metal/issues/53336

models/demos/deepseek_v3_d_p/tests/test_kv_cache_table.py
nbabinTT: added

models/demos/deepseek_v3_d_p/tt/runners/kv_chunk_table.py
nbabinTT: refactored

jjovicicTT (approved):

mtairum (approved):

mbezuljTT (approved):

PR #52437, opened 2026-08-07
[Metal 2.0] Port moreh/moreh_layer_norm_backward

VirdhatchaniKN:
# Port `moreh_layer_norm_backward` to Metal 2.0

Converts both program factories and all eight kernels from the legacy `ProgramDescriptor` path to
the Metal 2.0 host API. No behavioral change intended — same kernels, same work, expressed through
`ProgramSpec` / `ProgramRunArgs` with named bindings instead of raw circular-buffer indices.

**Scope:** 13 files, all inside the op directory — 2 factories (rewritten), 2 device-op headers
(2 lines each), 8 kernels, 1 new shared helper header.

## Notable details

- **`unpack_modes`** are hand-listed, one `UnpackToSrc` per Float32 buffer a compute kernel
  consumes (6 in `gamma_beta_grad`, 8 in `input_grad`), each next to the binding it belongs to.
  Metal 2.0 has no implicit default once 32-bit Dest is on, and a loop over bindings would
  silently cover a buffer nobody chose a mode for.
- **`opt_level` set explicitly to `O3`** on every compute spec. Legacy `ComputeConfig` defaulted
  to O3, Metal 2.0 `CompilerOptions` defaults to O2 — omitting it would be a quiet regression.
- **Per-core-group multiplicity preserved:** one compute `KernelSpec` per core group in its own
  `WorkUnitSpec`, keeping the per-group tile count a compile-time arg.
- **`input_grad`'s small/large choice** selects the reader source, the compute source *and* the
  buffer set (small needs one extra scratch buffer). Both branches are built explicitly.
- **New helper header** — five helpers shared across ~100 call sites, in a header rather than
  anonymous namespaces because `ttnn_op_moreh` is a unity build, where those merge and collide.
  It duplicates the peer `layernorm_distributed` helpers instead of including them, to avoid
  introducing cross-op coupling.
- `dst_full_sync_en` maps to `double_buffer_dest` **inverted**. Neither factory needs semaphores.

## Testing

`pytest tests/ttnn/nightly/unit_tests/operations/moreh/test_moreh_layer_norm.py` → **50 passed,
49 skipped** (28 exercise the backward op), zero failures. Build clean.

Kernel equivalence was checked by comparing the ordered LLK/helper call sequence against `main`:
143 / 219 / 300 calls in the three compute kernels, identical position for position.

The large `input_grad` algorithm is unreachable from any in-tree shape, so it was verified with a
temporary local scaffold (`[2, 16384]`, `normalized_dims=1`): 2 passed, large branch selected both
times, small branch never taken. That scaffold is not part of this PR.

## Follow-ups, not addressed here

None block review; each predates this PR or is out of its scope.

1. **Suspected pre-existing bug** — the small `input_grad` kernel does `wait_front(2)` where the
   large one does `pop_front(2)`. Looks like a leak rather than a deliberate difference, but
   changing it would be a behavior change, so it is carried across faithfully.
2. **No in-tree coverage of the large algorithm** — worth adding a shape that reaches it.
3. **`is_groupnorm` is never true**, making the `mask_w` path and eleven sibling branches dead.
   Preserved behind an `#ifdef DO_MASK_W` that is never defined rather than deleted; removing it
   means deciding the fate of the whole groupnorm scaffold, which is an owner call.
4. **`gen1_compute_config` is boilerplate** every Gen1-only op repeats — probably belongs in a
   shared Metal 2.0 header.
5. **`packer_l1_acc` handling and `log_info` severity** left exactly as found.

## Note for reviewers

The Metal 2.0 audit, plan and port report were committed early so the vetting work was reviewable,
then removed in the final commit — process artifacts, not code this op needs. They remain in this
branch's history, and no source file ever cited them, so removal leaves no dangling reference.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/moreh_layer_norm_b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/moreh_layer_norm_b)


ttnn/cpp/ttnn/operations/moreh/moreh_layer_norm_backward/device/kernels/moreh_layer_norm_backward_gamma_beta_grad_kernel.cpp
```diff
@@ -5,53 +5,48 @@
...
+    DataflowBuffer dfb_dy_obj(dfb::dy);          // output_grad(==dy)
+    DataflowBuffer dfb_x_obj(dfb::x);            // input(==x)
+    DataflowBuffer dfb_mean_obj(dfb::mean);      // mean
+    DataflowBuffer dfb_rstd_obj(dfb::rstd);      // rstd
+    DataflowBuffer dfb_scaler_obj(dfb::scaler);  // scaler
+    // mask_h is allocated only when the last row tile of the normalized region is partial, so its
+    // handle exists only on that path — the gate has to run at the preprocessor, before name lookup.
+#ifdef DO_MASK_H
+    DataflowBuffer dfb_mask_h_obj(dfb::mask_h);  // mask_h
+#endif
```
mcw-anasuya, line 30: These blocks are dead as stated in the comments. Is it okay to remove them?

mcw-anasuya (approved):

ttnn/cpp/ttnn/operations/moreh/moreh_layer_norm_backward/device/kernels/moreh_layer_norm_backward_gamma_beta_grad_kernel.cpp
VirdhatchaniKN: Good question, and thanks for pushing on it — but I'd rather not remove them in this PR.

They are genuinely unreachable, so the deletion would be safe. The catch is *why* they're
unreachable: `mask_w` is dead only because the factory hardwires `is_groupnorm = false`
(`moreh_layer_norm_backward_gamma_beta_grad_program_factory.cpp:102`, and `:115` in the InputGrad
factory). That single fact also kills every other `is_groupnorm` branch in these kernels — 6 in
`..._gamma_beta_grad_kernel.cpp`, 2 in `..._input_grad_small_kernel.cpp`, 3 in
`..._input_grad_large_kernel.cpp` — plus the `is_groupnorm` compile-time argument itself and the
`do_mask_w` derivation that feeds these blocks.

So removing `mask_w` alone would leave the groupnorm scaffolding half torn down: the mask path gone,
the rest still reading as live code. Whether to wire groupnorm up or retire it is a behavioral call for
the op owner, and either way it's one change covering all of that scaffolding at once — not something a
Metal 2.0 port should decide, since the port is meant to be a host-API swap with no behavior change.

Two more reasons to keep it here specifically:

- It's why the `#ifdef` exists at all. Under the legacy API this was harmless dead code; under Metal 2.0
  `dfb::mask_w` doesn't exist unless the host binds it, and a `constexpr bool do_mask_w = false` does
  **not** stop name lookup — so without the gate the kernel wouldn't compile. The gate is the port work;
  the dead path is pre-existing.
- Keeping the port diff scope-tight means that if a regression bisects to this commit, it's a Metal 2.0
  issue and nothing else.

What I've done instead is expand the comment so the next reader doesn't have to re-derive this
(20e8d3f851b) — it now says the path is kept deliberately and why. The full extent of the dead
groupnorm scaffolding is also written up as an open item in `METAL2_PORT_REPORT.md` for whoever picks up
that decision.

Happy to file the groupnorm cleanup as a separate issue/PR if that's useful — just say the word.

PR #52072, opened 2026-08-04
Add Gemma-2 TTNN bring-up scaffolding

sudhanshu112233shukla:
**Title**
#51398
Gemma-2 TTNN bring-up scaffolding in tt_transformers
**Description**
This change adds the first shared-framework wiring for Gemma-2 text models in models/tt_transformers:
Adds Gemma-2 HF/model aliases and cache path mappings.
Adds Gemma-2 prefill chunk-size entries for the supported device matrix.
Enables Gemma-2 defaults that the architecture expects:RMSNorm unit offset
embedding scale by sqrt(hidden_size)
alternating sliding-window / global attention layer pattern

Adds final logit soft-capping in the shared LMHead.
Adds targeted unit coverage for:Gemma-2 base model name normalization
final-logit softcap formula

**What this does not yet fully solve:**
Attention-logit soft-capping is still not implemented in the shared fused TTNN SDPA path.
I left that as an explicit NotImplementedError rather than silently producing incorrect outputs.

PR #40397, opened 2026-03-20
[TT-Transformers] Add padding to vocab-size to improve top-k performance on multichip systems

ifabijanic-tt:
### Problem description
Llama-3.3-70B, Qwen-3-32B and other models had a performance drop in throughput. For Llama-3.3-70B it went from 15.2 tok/s/user to 14 tok/s/user on T3K devices on the `performance-ci-32 test`. 

### What's changed
The perf drop happened with this commit: https://github.com/tenstorrent/tt-metal/commit/927a5409f43b370e00053805fe2d5206ed87f4c0

This commit included changes to top-k calculations and fixes on this op. One of the changes was the inclusion of bionic sort which required inputs to be in power of 2 sizes to work correctly. Because of that additional check was added which checked for that. If the check failed decode sampling of Top is not done on multiple chips but failed back to single core calculation. This would decrease performance significantly on multi chip devices. 

Llama-3.3-70B has vocabulary size of 128256 and in the current code `compute_padded_vocab_size()` only pads to a tile-aligned multiple of `32 * num_devices`, not a power of two, in `model_config.py`. On a Loudbox -> TP=8, the per-device width is `128256 / 8 = 16032`, which passes the old multi-core checks (`>=8192, <65536, k<=64`). However this number is not a power of 2 so it fails the new power-of-two check. Because of that the decode sampling of TopK falls back to single core and this kill throughput.

The fix in this PR checks enforces a check on TT-Transformer models to see if the vocab_size // num_devices is a power of two, and pads it if needed.
For Llama-3.3-70B, this fix increases performance from previous 14 tok/s/user to 16.2 tok/s/user.

#### Model tests
- [ ] T3K Demo - https://github.com/tenstorrent/tt-metal/actions/runs/23487549333
- [ ] Single card demo - https://github.com/tenstorrent/tt-metal/actions/runs/23487554651

PR #52444, opened 2026-08-07
[skip CI] [Models] Raise ValueError for invalid sampling_splits

mtairum:
### Problem

`should_pad_sampling_logits_to_power_of_2` and its test have contradicted each other since both were added in #40397.

The implementation warns and returns `False`:

```python
if sampling_splits < 1:
    logger.warning(f"Sampling_splits must be >= 1, got {sampling_splits}")
    return False
```

The test asserts it raises:

```python
def test_should_pad_sampling_logits_to_power_of_2_rejects_invalid_sampling_splits():
    with pytest.raises(ValueError, match="sampling_splits must be >= 1"):
        should_pad_sampling_logits_to_power_of_2("Llama-3.1-70B", 128256, 0)
```

This has gone unnoticed because `models/tt_transformers/tests/test_model_config_utils.py` is not referenced by any pipeline registry, so the test never runs in CI.

### Fix

Raise instead of warning. This matches the sibling function directly above it, `compute_padded_vocab_size`, which already raises `ValueError` for the same class of invalid input:

```python
if num_devices < 1:
    raise ValueError(f"num_devices must be >= 1, got {num_devices}")
```

### Impact

None in practice. The only production caller is `model_config.py:2731`, where `sampling_splits` is either `self.num_devices` (guarded `> 0`) or the constant `2`, so the branch is unreachable. The change matters for correctness rather than behaviour: an invalid argument now fails loudly instead of silently returning `False`, which would quietly disable the sampling-padding optimization rather than surfacing the bug.

Split out of #52072, where this one-line change was bundled with unrelated Gemma-2 work.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/fix-sampling-splits-valueerror)

### DIFF
--- a/models/tt_transformers/tt/model_config.py
+++ b/models/tt_transformers/tt/model_config.py
@@ -108,8 +108,7 @@ def should_pad_sampling_logits_to_power_of_2(
 ) -> bool:
     # Enable optional sampling padding for models that regress to single-core TopK. More info at issue #40399
     if sampling_splits < 1:
-        logger.warning(f"Sampling_splits must be >= 1, got {sampling_splits}")
-        return False
+        raise ValueError(f"sampling_splits must be >= 1, got {sampling_splits}")
 
     per_device_vocab = padded_vocab_size // sampling_splits
     return per_device_vocab > 0 and (per_device_vocab & (per_device_vocab - 1)) != 0


yieldthought:
Codex review:

[P2] Fix the still-failing sampling-padding test case

This PR fixes the invalid `sampling_splits` expectation, but [models/tt_transformers/tests/test_model_config_utils.py:39](https://github.com/tenstorrent/tt-metal/blob/b39cf5096d9cf1159b71b85524838e763e3260ee/models/tt_transformers/tests/test_model_config_utils.py#L39) still expects `should_pad_sampling_logits_to_power_of_2("Llama-3.1-8B", 128256, 4)` to return `False`. The implementation at [models/tt_transformers/tt/model_config.py:114](https://github.com/tenstorrent/tt-metal/blob/b39cf5096d9cf1159b71b85524838e763e3260ee/models/tt_transformers/tt/model_config.py#L114) ignores `base_model_name`; for that case `128256 // 4` is `32064`, which is non-zero and not a power of two, so the helper returns `True`. If the goal is to make this unregistered test file valid before adding it to CI, this PR still leaves it failing. Suggestion: either update the 8B expectation to `True`/remove the unused model dimension, or add the missing model gate if 8B is intentionally unsupported.

yieldthought (approved):
Codex: Approved per Mark's explicit Miguel vacation instruction after posting coreview results. Miguel should use his judgment on whether to address any findings before submission.

PR #52445, opened 2026-08-07
profiler: fail when zones overlap across threads

mvlahovicTT:
## What

`_assert_zones_valid` only ever checked that zone starts and ends pair up. Within a single thread that is guaranteed anyway, since entries are written in program order, so it could not catch the failure that actually happened: one thread racing into TILE_LOOP while another was still in INIT.

That went unnoticed through the whole fast tilize and untilize investigation, because both builds were wrong identically, so the with-counters versus no-counters difference stayed at zero while both numbers were wrong.

The new check asserts that no thread opens a phase before every thread has closed the previous one, grouped by run so it still holds when `raw()` is handed several runs at once. Only reported markers are checked, so a kernel can still keep an unreported zone such as UNINIT on a thread that some run type skips.

## Nesting is a supported shape

A zone inside a zone is not an overlap, and the check is built to allow it. `trisc.cpp` wraps every thread's kernel in `ZONE_SCOPED("KERNEL")`, so on hardware INIT and TILE_LOOP always sit inside a longer zone on their own thread. A zone that contains another zone on the same thread is treated as a wrapper and excluded from the phase sequence, leaving INIT and TILE_LOOP as the compared pair.

That inference is sound rather than a guess: within one thread the entries are written in program order, so two zones there are either nested or fully disjoint, never partially overlapping. Same-thread containment therefore only ever means real nesting.

Three tests pin this down, and they fail if the exemption is removed:

- `test_zones_nested_in_a_wrapper_are_not_an_overlap` — KERNEL around INIT and TILE_LOOP passes both views.
- `test_a_wrapper_does_not_mask_an_overlap_inside_it` — the wrapper does not hide a real cross-thread overlap.
- `test_wrapper_on_threads_without_phases_is_not_an_overlap` — ISOLATE run types, where every thread emits KERNEL but only the measured one emits phases.

Two limits follow from deriving wrappers this way, both recorded in the docstring. A phase that gains child zones becomes a wrapper itself and drops out of the comparison, so nesting a zone inside TILE_LOOP narrows what is checked to INIT against that child. And threads that name the same phase differently compare as separate phases, so a pipelined pair such as UNPACK_WORK on unpack and PACK_WORK on pack would read as an overlap. Neither affects any current kernel; both would need addressing before instrumentation grows in those directions.

## Quasar is skipped

Quasar's perf sources still use bare `ZONE_SCOPED` and have no rendezvous available to them: `llk_barrier` needs a semaphore nothing else uses, and Quasar's three are all owned by the ops. The invariant is not something they can satisfy yet, and asserting it there would fail every Quasar perf test rather than report a regression.

## Verified in both directions

Negative: fires at 176 cycles on a deliberately reverted kernel, and on an untagged frame whose `run_index` is all-NA (the case that silently skipped before review).
Positive: silent across 1620 `perf_fast_untilize` variants once the kernels use `START_PERF_MEASURE`.

An assertion only ever verified in the passing direction is not worth much, so the new nesting tests were checked the same way: deleting the wrapper exemption fails all three of them plus the pre-existing `l1_to_l1` fixture.

## Testing

Blackhole, speed of light, producer and consumer, with counters, passing. 39 host-only tests in `test_perf_report_hw_free.py` pass; `pre-commit` clean including `black (llk)`.
### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mvlahovic/perf-zone-invariants)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mvlahovic/perf-zone-invariants)

### DIFF
--- a/tt_metal/tt-llk/tests/python_tests/helpers/perf/schema.py
+++ b/tt_metal/tt-llk/tests/python_tests/helpers/perf/schema.py
@@ -12,6 +12,9 @@ class PerfSchemaError(AssertionError):
 
 MARKER = "marker"
 
+# Zones that use bare ZONE_SCOPED (no entry rendezvous), exempt from the cross-thread ordering invariant.
+NON_RENDEZVOUS_MARKERS = frozenset({"UNINIT"})
+
 FORMAT_HEADERS = (
     "formats.input_A",
     "formats.input_B",
--- a/tt_metal/tt-llk/tests/python_tests/helpers/profiler.py
+++ b/tt_metal/tt-llk/tests/python_tests/helpers/profiler.py
@@ -9,9 +9,17 @@ from typing import ClassVar
 
 import pandas as pd
 
+from .chip_architecture import ChipArchitecture, get_chip_architecture
 from .device_io import read_words_from_device
 from .llk_params import PerfRunType
-from .perf.schema import MARKER, MEAN, STD, stat_column, stat_prefix
+from .perf.schema import (
+    MARKER,
+    MEAN,
+    NON_RENDEZVOUS_MARKERS,
+    STD,
+    stat_column,
+    stat_prefix,
+)
 from .test_config import TestConfig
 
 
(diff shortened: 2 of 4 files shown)


tt_metal/tt-llk/tests/python_tests/helpers/profiler.py
```diff
@@ -79,6 +88,53 @@ def _assert_zones_valid(
...
+        is not something Quasar can satisfy yet, and asserting it there would fail every perf test
+        rather than report a regression.
+        """
+        if get_chip_architecture() == ChipArchitecture.QUASAR:
+            return
+        if MARKER not in starts.columns or "timestamp" not in starts.columns:
+            return
+        # raw() is also called on several runs concatenated together, so compare within a run.
+        group_cols = ["run_index"] if "run_index" in starts.columns else []
+        keys = starts.groupby(group_cols).groups.keys() if group_cols else [None]
```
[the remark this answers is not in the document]
mvlahovicTT: Confirmed and fixed, and this is the failure mode I care most about since an assertion that passes without checking is worse than none. Verified the mechanism: run_index is declared in the schema so the column test is always true, get_data() leaves it NA, and groupby drops NA keys, so an untagged frame yields zero groups and the loop never runs.

It now buckets NA rows together rather than dropping them. Verified with an untagged frame carrying a deliberate 176 cycle overlap: silent before, raises now. Used explicit masks rather than dropna=False because groupby(dropna=False).groups raises on an all-NA Int32 key on this pandas version.

Worth noting the main perf path was never vacuous, since perf.py tags before concat, so the 1620-variant result this PR reports still stands.

tt_metal/tt-llk/tests/python_tests/helpers/perf.py
```diff
@@ -40,8 +42,6 @@
 # Zone/marker names emitted by MEASURE_PERF_COUNTERS, in ID order. These must
 # match the marker values the kernels record; a mismatch silently empties the
 # TILE_LOOP mask in _postprocess_tile_loop (no KeyError raised).
```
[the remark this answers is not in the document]
mvlahovicTT: Right, it dangled. Moved to the constants in perf_schema.py and kept the _postprocess_tile_loop caveat, which is still true.

tt_metal/tt-llk/tests/python_tests/helpers/perf_schema.py
```diff
@@ -13,6 +13,10 @@ class PerfSchemaError(AssertionError):
 
 MARKER = "marker"
 
+# Zone markers that the report carries a column for. profiler.py enforces their ordering.
```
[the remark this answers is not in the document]
mvlahovicTT: Correct on both counts. They are values in the marker column, one report row per zone, not columns. Rewritten to say that, and to state what the check actually enforces: INIT closes on every thread before any thread opens TILE_LOOP, non-Quasar only.

tt_metal/tt-llk/tests/python_tests/helpers/perf/schema.py
```diff
@@ -12,6 +12,18 @@ class PerfSchemaError(AssertionError):
 
 MARKER = "marker"
 
+# Zone/marker names emitted by MEASURE_PERF_COUNTERS, in ID order. These must
+# match the marker values the kernels record; a mismatch silently empties the
```
skrsmanovicTT: Looks like double definition of same thing

tt_metal/tt-llk/tests/python_tests/helpers/metrics.py
skrsmanovicTT, line 1: is this by mistake part of this PR since it is chaning some metrics definition ?

tt_metal/tt-llk/tests/python_tests/test_perf_report_hw_free.py
```diff
@@ -241,8 +241,10 @@ def _one_run_events(seed: int) -> pd.DataFrame:
     """
     rows = []
     ts = 100
-    for thread in _THREADS:
-        for marker, mid in _MARKERS:
+    # Markers outer, threads inner: every thread closes INIT before any thread
```
skrsmanovicTT: Should we have also tests that is "failing on purpose" to confirm that infra is catching this error behaviour?

tt_metal/tt-llk/tests/python_tests/helpers/metrics.py
mvlahovicTT: you are right, it did not belong here, it is tied to another PR

tt_metal/tt-llk/tests/python_tests/test_perf_report_hw_free.py
mvlahovicTT: added

tt_metal/tt-llk/tests/python_tests/helpers/profiler.py
```diff
@@ -89,6 +168,7 @@ def raw(self) -> pd.DataFrame:
         end_entries = self.df[self.df["type"] == "ZONE_END"]
 
         self._assert_zones_valid(start_entries, end_entries)
+        self._assert_zones_dont_overlap(start_entries, end_entries)
```
sstanisicTT: IIRC the Profiler DF is ordered so that `ZONE_END` is the next row following `ZONE_START` so you don't need the filters above if you change the implementation a bit

sstanisicTT: This should probably also be taken into account for `_assert_zones_valid` but we can fix that later

sstanisicTT (commented):
The assertion check should be in perf.py, no profiler.py because it's specific to how we write performance microbenchmarks, not general profiler use-cases

tt_metal/tt-llk/tests/python_tests/helpers/profiler.py
mvlahovicTT: done

sstanisicTT (approved):

Issue #51350, opened 2026-07-28
Support SiTU activation function


pmilojevicTT:
Kimi K3 uses Sigmoid Tanh Unit (SiTU) activation function, while our current compute kernel support only SiLU and SwiGLU-OAI activation functions.

This is current PyTorch imeplemntation

```
# Register Moonshot-specific activation functions
class SituAndMul(nn.Module):
    """
    SituAndMul activation: beta * tanh(gate / beta) * sigmoid(gate) * up
    When linear_beta is set, up is also transformed by linear_beta * tanh(up / linear_beta).
    """

    def __init__(self, beta: float = 1.0, linear_beta: float | None = None):
        super().__init__()
        self.beta = beta
        self.linear_beta = linear_beta

    def forward(self, x: torch.Tensor) -> torch.Tensor:
        d = x.shape[-1] // 2
        gate = x[..., :d].to(torch.float32)
        up = x[..., d:].to(torch.float32)
        situ_a = self.beta * torch.tanh(gate / self.beta) * torch.sigmoid(gate)
        if self.linear_beta is not None:
            up = self.linear_beta * torch.tanh(up / self.linear_beta)
        return (situ_a * up).to(x.dtype)
```

Both self.beta and self.linear_beta are used in Kimi K3 so both activation functions for GATE and UP are needed.

pmilojevicTT:
Current changes on branch https://github.com/tenstorrent/tt-metal/compare/pmilojevic/51350-situ

kgrujcic-tt:
Implemented in PR #52448 (branch `kgrujcic/51350-situ`): adds `ttnn.softcap` (Blackhole SFPU unary, the up half) and `ttnn.situ_glu(gate, up, beta1, beta2=None)` (binary composite over pre-split gate/up; `beta2=None` skips the second tanh). `situ_gate` is not exposed standalone.

kgrujcic-tt:
Asked by Filip Vranic to wait for his `tanh` changes before merge, to isolate any potential. Regression. Found 0 issues. Proceeding further with this PR.

pmilojevicTT:
Initial PR for this was merged https://github.com/tenstorrent/tt-metal/pull/52448 currently working on potential improvement for SiTU https://github.com/tenstorrent/tt-metal/issues/52753

PR #52448, opened 2026-08-07
[Feature] Add softcap unary op and situ_glu binary activation (Kimi K3 SiTU-GLU)

kgrujcic-tt:
Adds Moonshot Kimi K3's SiTU-GLU activation to ttnn, closing #51350.

## What's added

- **`ttnn.softcap(x, beta)`** - `beta * tanh(x / beta)`, a Blackhole SFPU unary op. Bounds the input smoothly to +/-beta (the up half of SiTU; also general soft-capping, e.g. Gemma logit softcapping).
- **`ttnn.situ_glu(gate, up, beta1, beta2)`** - SiTU-GLU over pre-split `gate`/`up`:
  `(beta1 * tanh(gate/beta1) * sigmoid(gate)) * (beta2 * tanh(up/beta2))`. Implemented as a composite of `softcap` / `sigmoid` / `multiply`; the caller decides how to extract `gate` and `up`.
  - Composite intermediates are kept in **L1** for the routed-expert activation width (`hidden <= 3072`, K3 `moe_intermediate_size`) to skip the DRAM round-trip between the composed ops, and fall back to **DRAM** for wider activations (shared expert 6144, dense FFN 33792). The final output always uses the caller's memory config.
- **`situ_glu_tile`** (`api/compute/situ_glu.h`, SFPU math in `ckernels/blackhole/.../llk_sfpu/ckernel_sfpu_situ_glu.h`) - a dedicated single-pass **fused binary SFPU op** (`gate` and `up` pinned in DST, no intermediate materialized), modeled on the gpt-oss `swiglu_sfpu.h`. It is validated standalone and is **not yet wired** into an expert compute kernel - that integration is a follow-up PR. Blackhole only: it lives in the blackhole ckernel tree and its compute API is declared under `ARCH_BLACKHOLE`.
- **`_sfpu_softcap_`** - softcap gains a `vFloat` helper alongside its dst-loop op, matching the shape `tanh` and `sigmoid` already have, so the fused op calls it for both tanh halves instead of re-inlining the math. No pre-existing LLK file is modified: this PR adds `ckernel_sfpu_softcap.h` and registers the new op, nothing more. The fused op keeps its own reciprocal for the sigmoid half, since reusing the stock one would require changing existing LLK signatures.

Both ops run **bf16 / bfp8_b** activations and are Blackhole-only.

## Perf

All numbers below re-measured after rebasing onto main `30023337d7f`.

Blackhole p150, device kernel time per call in us (tracy `-p -r`, `DEVICE KERNEL DURATION` summed over the composite's ops, mean of 3 iterations after 2 warm-ups). `hidden` is the FFN activation width (Kimi K3: 3072 = routed-expert `moe_intermediate_size`, 33792 = dense FFN `intermediate_size`).

`softcap` (unary):

| hidden | #tokens | bf16 | bfp8_b |
|--:|--:|--:|--:|
| 3072 | 32 | 2.2 | 1.9 |
| 3072 | 256 | 8.1 | 6.0 |
| 3072 | 1024 | 41.5 | 21.6 |
| 3072 | 5120 | 199.3 | 97.8 |
| 33792 | 32 | 12.6 | 8.2 |
| 33792 | 256 | 107.6 | 54.4 |
| 33792 | 1024 | 445.8 | 212.5 |
| 33792 | 5120 | 2410.7 | 1056.9 |

`situ_glu` (composite; L1 intermediates at hidden<=3072, DRAM at hidden>3072; 5 device ops per call):

| hidden | #tokens | bf16 | bfp8_b | branch |
|--:|--:|--:|--:|:--|
| 3072 | 32 | 10.9 | 8.6 | L1 |
| 3072 | 256 | 33.7 | 27.4 | L1 |
| 3072 | 1024 | 122.7 | 95.6 | L1 |
| 3072 | 5120 | 559.0 | 440.9 | L1 |
| 33792 | 32 | 71.4 | 47.2 | DRAM |
| 33792 | 256 | 567.0 | 321.4 | DRAM |
| 33792 | 1024 | 2345.4 | 1260.7 | DRAM |
| 33792 | 5120 | 12801.6 | 6264.6 | DRAM |

(5120 = tile-aligned 5k tokens. hidden=3072 is the routed expert, which stays in L1; 33792 is the dense FFN.)

These are memory-bandwidth-bound elementwise ops (time scales with `tokens * hidden`). At 5120 tokens `situ_glu` costs 2.8x a single `softcap` on the L1 branch (bf16; 4.5x for bfp8_b, whose single `softcap` is itself 2x faster) and 5.3x on the DRAM branch, where every intermediate round-trips. bfp8_b is ~2x faster wherever the op is DRAM-bound; the L1 branch gains less, since its intermediates are already on-chip.

### Utilization

Fused `situ_glu_tile`, math thread (`TRISC_1`), Blackhole at 1350 MHz. `cyc/tile` is the `TRISC-KERNEL` zone divided by tiles (median of 12 dispatches, uninstrumented kernel). `SFPU cyc/tile` is a device zone around the SFPU call itself (per-tile median; the profiler records at most 125 zones per dispatch, so the 160-tile row uses that median scaled to 160), and `util` is `SFPU cyc/tile * tiles / TRISC-KERNEL`. tiles = tokens / 32.

| #tokens | tiles | bf16 cyc/tile | bf16 SFPU cyc/tile | bf16 util | bfp8_b cyc/tile | bfp8_b SFPU cyc/tile | bfp8_b util |
|--:|--:|--:|--:|--:|--:|--:|--:|
| 32 | 1 | 3822 | 3640 | 94.8% | 3784 | 3586 | 94.0% |
| 256 | 8 | 3100 | 2928 | 94.7% | 3089 | 2925 | 94.6% |
| 1024 | 32 | 3006 | 2926 | 97.2% | 2998 | 2927 | 97.3% |
| 5120 | 160 | 2980 | 2926 | 97.9% | 2974 | 2927 | 98.0% |

The fused op is SFPU-compute-bound: ~95% at 1 tile (the one-time init and warm-up are unamortized) rising to ~98% steady-state. Steady-state SFPU cost is a constant ~2927 cyc/tile in both dtypes; the first tile of a dispatch costs ~3600-3640 (cold instruction-cache fill plus SFPU pipeline warm-up), which is what keeps the 1-tile row high. bf16 and bfp8_b match within ~0.1% since the SFPU computes on unpacked values in DST regardless of input format. The composite is instead DRAM-bound, which is the headroom the fused kernel recovers when wired into an expert.

### softcap reuse is cycle-neutral

The fused kernel calls `_sfpu_softcap_` for both tanh halves instead of re-inlining the expression. Math-thread (`TRISC_1`) `TRISC-KERNEL` zone cycles, median of 12 dispatches per config, re-measured on the rebased branch:

| tiles | bf16 before | bf16 after | delta | bfp8_b before | bfp8_b after | delta |
|--:|--:|--:|--:|--:|--:|--:|
| 1 | 3812 | 3822 | +0.25% | 3770 | 3784 | +0.37% |
| 8 | 24788 | 24802 | +0.06% | 24718 | 24712 | -0.02% |
| 32 | 96206 | 96178 | -0.03% | 95940 | 95948 | +0.01% |
| 160 | 476842 | 476856 | +0.00% | 475860 | 475880 | +0.00% |

Every delta sits inside the per-dispatch warm-up spread. PCC is unchanged, with the max abs error against the torch golden identical to five significant figures (bf16 `5.0461e-01`, bfp8_b `4.3735e+00`).

## Tests

- `softcap`: shapes `[100]` and `[4,128,32]` x {bf16, bfp8_b}, plus a full bf16-domain sweep (saturation / NaN / Inf edges) and a zero-beta guard test.
- `situ_glu`: one case per memory branch (`hidden` 3072 routed -> L1, 6144 shared -> DRAM), covering both dtypes; PCC vs torch golden, plus a zero-beta guard test.
- `situ_glu_tile`: standalone `generic_op` harness (binary test kernel + reader), {bf16, bfp8_b} x {bf16, fp32} dst, PCC vs torch golden.

Cross-linked with #51350.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kgrujcic/51350-situ)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kgrujcic/51350-situ)


tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,136 @@
...
+    } else {
+        exp_neg_x = _sfpu_exp_21f_bf16_<true>(-x);
+    }
+    return _situ_glu_reciprocal_<is_fp32_dest_acc_en ? 2 : 1>(1.0f + exp_neg_x);
+}
+
+template <bool is_fp32_dest_acc_en, int ITERATIONS = 8, class Config = SituGluConfigKimi>
+inline void calculate_situ_glu(const uint gate_tile_idx, const uint up_tile_idx, const uint out_tile_idx) {
+    constexpr float beta_gate = Config::beta_gate;
+    constexpr float inv_beta_gate = 1.0f / Config::beta_gate;
```
pmilojevicTT, line 74: Can we also precalculate these inverse constants?

kgrujcic-tt: We could, but that wouldn't be a functional blocker. All potential perf improvements should be kept for follow up PRs that aim to improve perf.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,136 @@
...
+struct SituGluConfigKimi {
+    static constexpr float beta_gate = 4.0f;
+    static constexpr float beta_up = 25.0f;
+};
+
+// Newton reciprocal with 2.0 as a literal. The stock sfpu_reciprocal_iter reads that
+// constant from vConstFloatPrgm0, which tanh_init has loaded with a tanh coefficient.
+// Otherwise identical to sfpu_reciprocal_iter; keep in sync with recip.h.
+template <int MAX_ITER>
+sfpi_inline sfpi::vFloat _situ_glu_reciprocal_(const sfpi::vFloat x) {
```
fvranicTT, line 39: Can we move this code do ckernel_sfpu_situ_glu.h?

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,136 @@
...
+    sfpi::vFloat y = sfpi::approx_recip(x);
```
fvranicTT, line 40: `approx_recip` exists only on BH.

kgrujcic-tt: This is BH-only implementation. You think it's not properly placed then?

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
kgrujcic-tt: We could, but I'm missing why?

kgrujcic-tt: Moved. The SFPU math now lives in `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h` and the kernel-facing entry point is `tt_metal/hw/inc/api/compute/situ_glu.h` (`situ_glu_tile` / `situ_glu_tile_init`), same split softcap uses in this PR. The hand-rolled llk wrappers are gone; the op goes through `SFPU_BINARY_CALL`. Commit 70e106b8337.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
kgrujcic-tt: Right, and that is now enforced by placement rather than by a hand-written `#error`: the code sits under `ckernels/blackhole/` and the compute API in `api/compute/situ_glu.h` is declared inside `#if defined(ARCH_BLACKHOLE)`. Same guard added to `api/compute/eltwise_unary/softcap.h`. Commit 70e106b8337.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
kgrujcic-tt: Correction to my earlier reply: these already are compile-time constants. `beta_gate`/`beta_up` come from the config struct as `static constexpr float`, so `1.0f / Config::beta_gate` is folded by the compiler and no division or LReg-held beta reaches the kernel. Nothing left to precompute here.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_softcap.h
```diff
@@ -0,0 +1,47 @@
...
+
+    for (int d = 0; d < ITERATIONS; d++) {
+        // Always the Sollya polynomial, never _sfpu_tanh_fp32_accurate_, in both dst
+        // modes: the accurate expm1 tanh together with runtime beta exhausts the SFPU
+        // LReg file, which is a hard compile abort ("cannot store sfpu register")
+        // rather than a slowdown. Callers needing fp32-grade tanh should use tanh_tile.
+        //
+        // Round once, after the beta rescale -- rounding tanh first would discard bits
+        // that the multiply by beta then amplifies.
+        sfpi::vFloat result = _sfpu_tanh_polynomial_(sfpi::dst_reg[0] * inv_beta) * beta;
```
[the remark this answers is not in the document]
kgrujcic-tt: Not fixing: this matches the stock BH tanh. `calculate_tanh` / `tanh_tile` map NaN to 1.0 through the same `min(., 1.0)` clamp, so making softcap alone propagate NaN would make it inconsistent with the op it is built on and cost an extra setcc/mov per element in a hot path. The behaviour is documented in the header ("+/-Inf clamps to +/-beta; NaN does not propagate"), in the compute API doc, and the full-domain test masks NaN explicitly rather than silently passing.

tt_metal/hw/inc/api/compute/eltwise_unary/softcap.h
```diff
@@ -0,0 +1,56 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "api/compute/common_globals.h"
+#ifdef TRISC_MATH
+#include "ckernel_sfpu_softcap.h"
+#include "llk_math_eltwise_unary_sfpu_macros.h"
```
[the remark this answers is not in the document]
kgrujcic-tt: Addressed in 70e106b8337: `api/compute/eltwise_unary/softcap.h` now declares the API inside `#if defined(ARCH_BLACKHOLE)`, so including it on Wormhole is harmless and calling `softcap_tile` there is a clear "not declared" error instead of a missing-header failure deep in the LLK include chain. The host-side `TT_FATAL` on arch stays as the real guard. Per-arch LLK implementations are out of scope for this PR - there is no Wormhole `_sfpu_softcap_`.

ttnn/cpp/ttnn/operations/eltwise/binary/binary_nanobind.cpp
```diff
@@ -2037,6 +2088,12 @@ void py_module(nb::module_& mod) {
         detail::kFloatOnlyDtypes,
         detail::kSameDtypeRequiredFootnote);
 
+    detail::bind_situ_glu<"situ_glu">(
+        mod,
+        R"doc(Computes Moonshot's SiTU-GLU activation over the pre-split :attr:`gate` and :attr:`up` tensors. When :attr:`beta2` is `None`, the up half is left untransformed.)doc",
+        R"doc(\mathrm{output\_tensor}_i = \left(\verb|beta1| \cdot \tanh(\mathrm{gate}_i / \verb|beta1|) \cdot \sigma(\mathrm{gate}_i)\right) \cdot \verb|up|_i)doc",
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed earlier in this PR: `beta2` is a required argument, and the rendered equation is `(beta1 * tanh(gate/beta1) * sigmoid(gate)) * (beta2 * tanh(up/beta2))` - the up factor is no longer raw `up`.

ttnn/ttnn/operations/unary.py
```diff
@@ -621,6 +621,15 @@ def _golden_function_celu(input_tensor_a, *args, alpha=1.0, **kwargs):
 ttnn.attach_golden_function(ttnn.celu, golden_function=_golden_function_celu)
 
 
+def _golden_function_softcap(input_tensor_a, *args, beta, **kwargs):
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in 5c4b00c4: `_golden_function_softcap(input_tensor_a, beta, *args, **kwargs)` takes beta positionally, so the golden runs in comparison/fallback mode.

ttnn/ttnn/operations/binary.py
```diff
@@ -246,6 +246,20 @@ def _golden_function_hypot(input_tensor_a, input_tensor_b, *args, **kwargs):
 ttnn.attach_golden_function(ttnn.hypot, golden_function=_golden_function_hypot)
 
 
+def _golden_function_situ_glu(gate, up, *args, beta1, beta2=None, **kwargs):
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in 5c4b00c4: the golden is now `_golden_function_situ_glu(gate, up, beta1, beta2, *args, **kwargs)`, matching the positional nanobind signature.

ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
```diff
@@ -59,6 +59,19 @@ void UnaryDeviceOperation::validate_on_program_cache_miss(
...
+        if (op.type() == operations::unary::UnaryOpType::SOFTCAP) {
+            // ckernel_sfpu_softcap.h and the SfpuType registration it needs exist only
+            // under hw/ckernels/blackhole. Without this the kernel reaches JIT and
+            // dies on a missing header, which points nowhere useful.
+            TT_FATAL(
+                input_tensor.device()->arch() == tt::ARCH::BLACKHOLE,
+                "Unary: SOFTCAP is implemented for Blackhole only, got arch {}",
+                input_tensor.device()->arch());
+            break;
+        }
```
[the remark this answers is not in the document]
kgrujcic-tt: Now fully addressed. 5c4b00c4 added the arch check; 20955c225fa adds the non-zero beta guard next to it, so standalone softcap and fused unary chains share it, with `test_softcap_zero_beta_guard` covering it. situ_glu keeps its own guard on beta1/beta2 in binary_composite_op.cpp (it never reaches the unary op with a zero beta), covered by `test_situ_glu_zero_beta_guard`.

tests/ttnn/unit_tests/operations/eltwise/test_unary.py
```diff
@@ -2784,3 +2784,88 @@ def test_unary_mish(torch_dtype, ttnn_dtype, fast_and_approximate_mode, device):
...
+    # No range limiting: x/beta may overflow the polynomial's Horner chain to inf,
+    # but the min(., 1.0) clamp that bounds tanh turns that back into exactly beta.
+    input_tensor = ttnn.from_torch(in_data, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    tt_res = ttnn.to_torch(ttnn.softcap(input_tensor, SOFTCAP_BETA))
+    golden = ttnn.get_golden_function(ttnn.softcap)(in_data, beta=SOFTCAP_BETA, device=device)
+
+    # tanh is bounded by 1, so beta is a hard bound; an overshoot means the polynomial's
+    # saturation clamp is not holding.
+    assert tt_res.to(torch.float32).abs().max().item() <= SOFTCAP_BETA * (1.0 + 1e-3)
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in 5c4b00c4: both assertions now report the measured max and the bound, e.g. `softcap overshoot: max |out| ... > bound ...`.

tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
```diff
@@ -894,3 +895,50 @@ def test_unary_right_shift(input_shapes, device):
...
+    gate_tt = ttnn.from_torch(gate, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    up_tt = ttnn.from_torch(up, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    tt_res = ttnn.to_torch(ttnn.situ_glu(gate_tt, up_tt, SITU_GLU_BETA1, beta2=beta2))
+    golden = ttnn.get_golden_function(ttnn.situ_glu)(gate, up, beta1=SITU_GLU_BETA1, beta2=beta2)
+
+    if beta2 is not None:
+        # Both halves are bounded: |situ_a| <= beta1, |up_half| <= beta2.
+        bound = SITU_GLU_BETA1 * SITU_GLU_BETA2 * (1.0 + 1e-3)
+        assert tt_res.to(torch.float32).abs().max().item() <= bound
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in 5c4b00c4: the assertion reports the measured max magnitude and the bound.

fvranicTT (approved):

tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
```diff
@@ -894,3 +895,53 @@ def test_unary_right_shift(input_shapes, device):
...
+    tt_res = ttnn.to_torch(out)
+    golden = ttnn.get_golden_function(ttnn.situ_glu)(gate, up, beta1=SITU_GLU_BETA1, beta2=SITU_GLU_BETA2)
+
+    is_bfp8 = ttnn_dtype == ttnn.bfloat8_b
+    # Both halves are bounded: |situ_a| <= beta1, |up_half| <= beta2.
+    bound = SITU_GLU_BETA1 * SITU_GLU_BETA2 * (1.0 + (5e-2 if is_bfp8 else 1e-3))
+    max_abs = tt_res.to(torch.float32).abs().max().item()
+    assert max_abs <= bound, f"situ_glu overshoot: max |out| {max_abs:.4f} > bound {bound:.4f}"
+
+    assert_with_pcc(golden, tt_res, pcc=0.99 if is_bfp8 else 0.999)
```
vtsilytskyiTT: Nit: `assert_with_ulp` is a better option. PCC shows correlation, ULP shows deviation from expected result.

tests/ttnn/unit_tests/operations/eltwise/test_situ_glu_sfpu.py
```diff
@@ -0,0 +1,152 @@
...
+    # |situ_a| <= beta_gate, |up_half| <= beta_up -> |result| <= their product.
+    bound = BETA_GATE * BETA_UP * (1.0 + (5e-2 if in_name == "bfp8_b" else 1e-3))
+    assert actual.to(torch.float32).abs().max().item() <= bound
+
+    g = golden.to(torch.float32)
+    a = actual.to(torch.float32)
+    logger.info(f"{in_name} fp32_dst={fp32_dest}: max abs err {(a - g).abs().max().item():.4e}")
+
+    pcc = 0.99 if in_name == "bfp8_b" else 0.999
+    assert_with_pcc(g, a, pcc=pcc)
```
vtsilytskyiTT: Same, check `assert_with_ulp` as well, please.

tests/ttnn/unit_tests/operations/eltwise/test_situ_glu_sfpu.py
```diff
@@ -0,0 +1,152 @@
...
+    golden = situ_glu_reference(gate_t, up_t)
+    actual = _run(device, gate_t, up_t, in_dtype, page_bytes, fp32_dest)
+
+    # |situ_a| <= beta_gate, |up_half| <= beta_up -> |result| <= their product.
+    bound = BETA_GATE * BETA_UP * (1.0 + (5e-2 if in_name == "bfp8_b" else 1e-3))
+    assert actual.to(torch.float32).abs().max().item() <= bound
+
+    g = golden.to(torch.float32)
+    a = actual.to(torch.float32)
+    logger.info(f"{in_name} fp32_dst={fp32_dest}: max abs err {(a - g).abs().max().item():.4e}")
```
vtsilytskyiTT:
```suggestion
    logger.debug(f"{in_name} fp32_dst={fp32_dest}: max abs err {(a - g).abs().max().item():.4e}")
```
This test will run often, let's not waste milliseconds on output, that nobody normally reads.

tests/ttnn/unit_tests/operations/eltwise/test_unary.py
```diff
@@ -2784,3 +2784,105 @@ def test_unary_mish(torch_dtype, ttnn_dtype, fast_and_approximate_mode, device):
...
+        (torch.Size([4, 128, 32])),
+    ),
+)
+@pytest.mark.parametrize("ttnn_dtype", [ttnn.bfloat16, ttnn.bfloat8_b])
+def test_unary_softcap(input_shapes, ttnn_dtype, device):
+    if ttnn_dtype == ttnn.bfloat8_b:
+        # bfp8_b shares one exponent per 16-element block, so the full-range stress input
+        # (1e-5 .. 3e38 in a single tile) is neither representable nor a realistic activation.
+        # Use a bounded range that still spans the near-linear and saturated (|x| >> beta) regions.
+        torch.manual_seed(0)
```
vtsilytskyiTT: `torch.manual_seed(0)` is present in both arms. Let's extract it.

tests/ttnn/unit_tests/operations/eltwise/test_unary.py
```diff
@@ -2784,3 +2784,105 @@ def test_unary_mish(torch_dtype, ttnn_dtype, fast_and_approximate_mode, device):
...
+    max_abs = result.to(torch.float32).abs().max().item()
+    fd_bound = SOFTCAP_BETA * (1.0 + 1e-3)
+    assert max_abs <= fd_bound, f"softcap overshoot: max |out| {max_abs:.4f} > bound {fd_bound:.4f}"
+    assert not torch.isnan(result).any(), "finite input produced NaN"
+
+    g = golden.to(torch.float32)
+    r = result.to(torch.float32)
+    mask = g.abs() > SOFTCAP_FLUSH_FLOOR
+    rel_err = ((r[mask] - g[mask]).abs() / g[mask].abs()).max().item()
+    assert rel_err < 1.5e-2, f"full bf16 domain max rel err {rel_err:.4e}"
```
vtsilytskyiTT: You might want to use `comp_allclose` or even `comp_allclose_and_pcc` from `models/common/utility_functions.py`. And yes, `comp_ulp` is located in the same file :)

vtsilytskyiTT (commented):
Generally OK. My main suggestion is to add allclose/ulp alongside PCC. PCC alone is not reliable.

tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
kgrujcic-tt: Added. The bf16 arm now asserts `assert_with_ulp(..., ulp_threshold=6)` alongside PCC; worst measured distance is 3.0 ULP across the three composed ops (softcap, sigmoid, multiply). The bfp8_b arm stays PCC-only on purpose: with one exponent shared per 16-element block, a small element next to a large one is hundreds of bf16 ULP off before the op even runs (measured 780), so a ULP gate there would be measuring the format, not the kernel. Noted in a comment next to the thresholds.

tests/ttnn/unit_tests/operations/eltwise/test_situ_glu_sfpu.py
kgrujcic-tt: Done: `ulp_threshold=4` for the bf16 arm (worst measured 1.4 in both dst modes), PCC kept alongside. bfp8_b stays PCC-only for the same reason as the composite test - its inputs are quantized before the SFPU sees them, so the output sits ~343 bf16 ULP from the torch golden regardless of kernel accuracy.

tests/ttnn/unit_tests/operations/eltwise/test_situ_glu_sfpu.py
kgrujcic-tt: Applied.

tests/ttnn/unit_tests/operations/eltwise/test_unary.py
kgrujcic-tt: Extracted - `torch.manual_seed(0)` now runs once at the top of the test.

tests/ttnn/unit_tests/operations/eltwise/test_unary.py
kgrujcic-tt: Switched to `assert_with_ulp` (tests/ttnn/utils_for_testing.py), which wraps `comp_ulp` from models/common/utility_functions.py and reports the worst element on failure. It replaces the hand-rolled relative-error bound: over the full bf16 domain the polynomial lands within 0.72 ULP, so the threshold is 2. The subnormal-flush region is still excluded and checked separately, since there the golden is denormal-tiny and ULP-vs-golden is not meaningful. Kept PCC alongside per your summary comment; I stayed with the assert_with_* wrappers rather than comp_allclose so the file keeps one assertion style.

vtsilytskyiTT (approved):
LGTM

nvelickovicTT (approved):

ttnn/cpp/ttnn/operations/eltwise/binary/device/binary_composite_op.cpp
```diff
@@ -886,4 +886,35 @@ Tensor bias_gelu(
...
+    float beta1,
+    float beta2,
+    const std::optional<MemoryConfig>& output_mem_config) {
+    using namespace operations::unary;
+
+    // softcap precomputes 1/beta, so zero would emit inf.
+    TT_FATAL(beta1 != 0.0f && beta2 != 0.0f, "situ_glu: beta1 and beta2 must be non-zero");
+
+    const std::optional<MemoryConfig> interm_mem = gate.logical_shape()[-1] <= SITU_GLU_L1_MAX_HIDDEN
+                                                       ? std::optional<MemoryConfig>(ttnn::L1_MEMORY_CONFIG)
```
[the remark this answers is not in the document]
kgrujcic-tt: Real, and worse than the numbers here suggest. Verified on p150: 8192 tokens x 3072 bf16 (144 MB peak) still fits, 16384 hard-fails in bank_manager with Not enough space to allocate 100663296 B L1 buffer. 16k is a plausible prefill chunk, so this was a live crash. Fixed in 149cb1d7e68: the L1 branch now also requires 3 * intermediate bytes to fit a 3/4-of-L1 budget (three are live at the peak, as you noted), with a regression test at 8192 tokens. 5120 tokens x 3072 stays on L1, so the perf table still describes the code. Sharded inputs now skip the L1 branch entirely rather than taking the untested unshard/reshard path.

ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
```diff
@@ -59,6 +59,24 @@ void UnaryDeviceOperation::validate_on_program_cache_miss(
...
+            TT_FATAL(
+                input_tensor.device()->arch() == tt::ARCH::BLACKHOLE,
+                "Unary: SOFTCAP is implemented for Blackhole only, got arch {}",
+                input_tensor.device()->arch());
+            // beta reaches the kernel as (beta, 1/beta), so zero would hand the SFPU inf and
+            // return something that is not beta * tanh(x / beta). Guards the fused chains too.
+            const auto beta = op.get_param_if<float>(0);
+            TT_FATAL(beta.has_value(), "Unary: SOFTCAP requires a float beta parameter");
+            TT_FATAL(*beta != 0.0f, "Unary: SOFTCAP requires a non-zero beta");
+            break;
```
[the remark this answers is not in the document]
kgrujcic-tt: Both fixed in 149cb1d7e68. The `break` is gone - beta is per-op, so `unary_chain(t, [softcap(2.0), softcap(0.0)])` did pass validation and now raises. Added the dtype guard too: fp32 in, polynomial-grade accuracy out is a silent trap, and the binding advertises BFLOAT16/BFLOAT8_B. Both verified on device.

ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_op_utils.cpp
```diff
@@ -614,6 +615,15 @@ std::pair<std::string, std::string> get_op_init_and_func_parameterized(
...
                     std::bit_cast<uint32_t>(1.0f / param0))};
+        // softcap takes beta with 1/beta precomputed here, so the kernel never divides.
+        case UnaryOpType::SOFTCAP:
+            return {
+                "softcap_tile_init();",
+                fmt::format(
+                    "softcap_tile({}, {:#x}u, {:#x}u);",
+                    idst,
+                    std::bit_cast<uint32_t>(param0),
+                    std::bit_cast<uint32_t>(1.0f / param0))};
```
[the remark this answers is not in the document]
kgrujcic-tt: Half of this is real and fixed. The zero-beta check now sits in unary_op_utils.cpp where `1.0f / param0` is formed, so the fused activation path is covered rather than relying on UnaryDeviceOperation::validate. The arch check I am leaving where it is: `get_op_init_and_func` has no device handle, and there is no generic arch-capability mechanism for fused activations to hook into - inventing one for a single op is out of scope for this PR. On Wormhole the fused path fails at JIT with an undeclared `softcap_tile`, which names the op and the arch-guarded header.

ttnn/ttnn/operations/binary.py
```diff
@@ -246,6 +246,19 @@ def _golden_function_hypot(input_tensor_a, input_tensor_b, *args, **kwargs):
...
 
 
+def _golden_function_situ_glu(gate, up, beta1, beta2, *args, **kwargs):
+    import torch
+
+    g = gate.to(torch.float32)
+    u = up.to(torch.float32)
+    situ_a = beta1 * torch.tanh(g / beta1) * torch.sigmoid(g)
+    up_half = beta2 * torch.tanh(u / beta2)
+    return (situ_a * up_half).to(gate.dtype)
```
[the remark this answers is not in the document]
kgrujcic-tt: Good catch - fixed in 149cb1d7e68. The golden now stays fp32 like the others in that file. Re-measured worst case is 3.02 ULP (vs 3.0 with the downcast), so the threshold of 6 stands, but the check now measures what it claims to.

tests/tt_metal/tt_metal/test_kernels/compute/situ_glu.cpp
```diff
@@ -0,0 +1,52 @@
...
+
+    for (uint32_t t = 0; t < num_tiles; ++t) {
+        gate_cb.wait_front(1);
+        up_cb.wait_front(1);
+        out_cb.reserve_back(1);
+
+        tile_regs_acquire();
+        copy_tile(cb_gate, 0, 0);  // gate -> dst[0]
+        copy_tile(cb_up, 0, 1);    // up   -> dst[1]
+        situ_glu_tile(0, 1, 0);
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in 149cb1d7e68: `dst_out` is a compile-time arg and the test is parametrized over 0 (aliasing the gate operand, which is what an expert kernel does) and 2 (separate slot), so a kernel ignoring `out_tile_idx` now fails. 8 cases, all passing. Also named the dst indices.

tests/tt_metal/tt_metal/test_kernels/compute/situ_glu.cpp
```diff
@@ -0,0 +1,52 @@
...
+    situ_glu_tile_init();
+
+    for (uint32_t t = 0; t < num_tiles; ++t) {
+        gate_cb.wait_front(1);
+        up_cb.wait_front(1);
+        out_cb.reserve_back(1);
+
+        tile_regs_acquire();
+        copy_tile(cb_gate, 0, 0);  // gate -> dst[0]
+        copy_tile(cb_up, 0, 1);    // up   -> dst[1]
```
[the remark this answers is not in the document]
kgrujcic-tt: Right. The two CBs always share a format in this harness, so `copy_tile_to_dst_init_short_with_dt` would be a no-op here; rather than add a call that is never exercised, I stated the requirement at `init_sfpu` - both input CBs must share a data format, since init_sfpu configures the unpacker from cb_gate and copy_tile only updates the L1 address. A mixed-format consumer needs the reconfig, and that belongs with the kernel that does it.

tt_metal/hw/inc/api/compute/eltwise_unary/softcap.h
```diff
@@ -0,0 +1,62 @@
...
+namespace ckernel {
+
+// clang-format off
+/**
+ * Performs element-wise computation of softcap(x) = beta * tanh(x / beta) on
+ * each element of a tile in DST register at index tile_index. The DST register
+ * buffer must be in acquired state via *acquire_dst* call. This call is blocking
+ * and is only available on the compute engine.
+ *
+ * The output is bounded by beta: near-linear for |x| << beta, saturating smoothly at
```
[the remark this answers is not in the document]
kgrujcic-tt: Confirmed and fixed in 149cb1d7e68. Both doc comments now say the bound is exact in fp32 and the packed result can land up to half an output-dtype ULP above beta, and `SOFTCAP_BOUND_TOL[bfloat16]` is `2**-8` so it no longer depends on 25.0 being bf16-exact.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_softcap.h
```diff
@@ -0,0 +1,48 @@
...
+// divides (same contract as celu's alpha / alpha_recip).
+//
+// Always the Sollya polynomial, never _sfpu_tanh_fp32_accurate_: the accurate tanh plus a
+// runtime beta exhausts the SFPU LReg file, a hard compile abort rather than a slowdown.
+// Use tanh_tile if fp32-grade tanh is needed.
+//
+// +/-Inf clamps to +/-beta; NaN does not propagate, matching stock calculate_tanh. Leaves
+// rounding to the caller, so fused callers can round once at the end.
+sfpi_inline sfpi::vFloat _sfpu_softcap_(sfpi::vFloat x, sfpi::vFloat beta, sfpi::vFloat inv_beta) {
+    return _sfpu_tanh_polynomial_(x * inv_beta) * beta;
```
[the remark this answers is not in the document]
kgrujcic-tt: Measured on p150: `softcap(-0.0)` returns +0.0, so the mechanism you describe holds. Not adding the `copysgn` though - `ttnn.tanh(-0.0)` returns +0.0 on the same run, so softcap matches the op it is built on, and a fix here would make one member of the family stricter than the rest for an input no activation carries. Documented instead, next to the existing NaN note in ckernel_sfpu_softcap.h and softcap.h.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,101 @@
...
+inline void calculate_situ_glu(const uint gate_tile_idx, const uint up_tile_idx, const uint out_tile_idx) {
+    constexpr float beta_gate = Config::beta_gate;
+    constexpr float inv_beta_gate = 1.0f / Config::beta_gate;
+    constexpr float beta_up = Config::beta_up;
+    constexpr float inv_beta_up = 1.0f / Config::beta_up;
+    constexpr uint dst_tile_size = 32;  // 32 rows per tile in SFPU addressing
+
+    for (int d = 0; d < ITERATIONS; d++) {
+        sfpi::vFloat gate = sfpi::dst_reg[gate_tile_idx * dst_tile_size];
+        sfpi::vFloat up = sfpi::dst_reg[up_tile_idx * dst_tile_size];
```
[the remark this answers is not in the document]
kgrujcic-tt: Leaving as is. There is no evidence of a spill: the fused op measures a constant 2926-2927 SFPU cyc/tile from 8 tiles up, at 97-98% SFPU utilisation of the math-thread kernel, and both dst modes pass. sfpi is free to schedule the `dst_reg` read regardless of where it appears in the source, so moving the line is not a live-range guarantee - it would be churn justified by a suspicion. Worth revisiting with a measurement if the fp32-dest path ever shows up hot in the expert kernel.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,101 @@
...
+
+// Betas are compile-time so the kernel never divides and no LReg holds them.
+struct SituGluConfigKimi {
+    static constexpr float beta_gate = 4.0f;
+    static constexpr float beta_up = 25.0f;
+};
+
+// Newton reciprocal with 2.0 as a literal. The stock sfpu_reciprocal_iter reads that
+// constant from vConstFloatPrgm0, which tanh_init has loaded with a tanh coefficient.
+// Otherwise identical to sfpu_reciprocal_iter; keep in sync with recip.h.
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed the comment, declined the refactor. The claim was inaccurate for `MAX_ITER == 0` and now reads: identical for MAX_ITER >= 1, the only values instantiated here, dropping the stock MAX_ITER == 0 case. Adding a template parameter to `sfpu_reciprocal_iter` means editing a shared LLK header that every reciprocal user compiles against, and this PR is deliberately a pure addition - no pre-existing LLK file is modified. That is a reasonable follow-up on its own, not something to fold into a new-op PR.

tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
```diff
@@ -894,3 +895,64 @@ def test_unary_right_shift(input_shapes, device):
...
+    # Span the saturating and near-linear regions of both halves.
+    gate = torch.empty(input_shape, dtype=torch.bfloat16).uniform_(-30.0, 30.0)
+    up = torch.empty(input_shape, dtype=torch.bfloat16).uniform_(-30.0, 30.0)
+
+    gate_tt = ttnn.from_torch(gate, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    up_tt = ttnn.from_torch(up, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    out = ttnn.situ_glu(gate_tt, up_tt, SITU_GLU_BETA1, SITU_GLU_BETA2)
+    # Output placement follows the input, not the possibly-L1 intermediates.
+    assert out.memory_config().buffer_type == gate_tt.memory_config().buffer_type
```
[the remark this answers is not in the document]
kgrujcic-tt: Agreed that the comment overstated it - reworded to say the cases cover numerics on both sides of the split, and that the assertion checks placement rather than which branch ran. Keeping the assertion: it is a regression guard for a real bug fixed earlier in this PR (the output used to inherit the possibly-L1 intermediate config), so it is cheap and it has caught something before. Which branch ran is now covered separately by test_situ_glu_l1_intermediates_fall_back.

tt_metal/hw/inc/api/compute/situ_glu.h
```diff
@@ -0,0 +1,66 @@
...
+    MATH((SFPU_BINARY_CALL(
+        DST_SYNC_MODE,
+        DST_ACCUM_MODE,
+        calculate_situ_glu,
+        (DST_ACCUM_MODE, 8 /* ITERATIONS */, sfpu::SituGluConfigKimi),
+        idst0,
+        idst1,
+        odst,
+        vector_mode)));
+}
```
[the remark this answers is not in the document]
kgrujcic-tt: Tried it; it does not compile. `ckernel_sfpu_situ_glu.h` is included under `#ifdef TRISC_MATH` (it pulls in sfpi and cmath_common, like every other compute API header does), but a default template argument is parsed on all three TRISCs: TRISC_0 gives error: 'sfpu' does not name a type. `softcap_tile` avoids this by taking beta at runtime, which for this op would put four live values in LRegs on a kernel where tanh_init has already claimed all three vConstFloatPrgm. So the config stays compile-time and the doc points second callers at `calculate_situ_glu`; a runtime-beta entry point can come with the model that needs it.

ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_op_utils.cpp
```diff
@@ -614,6 +615,18 @@ std::pair<std::string, std::string> get_op_init_and_func_parameterized(
                     idst,
                     std::bit_cast<uint32_t>(param0),
                     std::bit_cast<uint32_t>(1.0f / param0))};
+        // softcap takes beta with 1/beta precomputed here, so the kernel never divides. The
+        // check is repeated from UnaryDeviceOperation::validate because fused activation chains
+        // (ttnn::multiply's lhs/rhs/post_activations) reach this emitter without going through it.
+        case UnaryOpType::SOFTCAP:
+            TT_FATAL(param0 != 0.0f, "SOFTCAP requires a non-zero beta");
```
Aswinmcw, line 622: The zero-beta check was moved here precisely because fused chains bypass `UnaryDeviceOperation::validate`, but the FLOAT32 guard added in the same commit stayed behind on the bypassed side — so `ttnn.multiply(a, b, post_activations=[(UnaryOpType.SOFTCAP, 2.0)])` on fp32 still reaches the kernel. Unlike the arch check, that one fails silently (fp32 out with polynomial-grade error), which is the trap the guard was added for. `input_dtype` is already a parameter of this function and other cases `TT_FATAL` on it, so it can move here too.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_situ_glu.h
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// sigmoid(x) = 1 / (1 + exp(-x)); both exp variants are free of vConstFloatPrgm.
+template <bool is_fp32_dest_acc_en>
+sfpi_inline sfpi::vFloat _situ_glu_sigmoid_(sfpi::vFloat x) {
+    sfpi::vFloat exp_neg_x;
+    if constexpr (is_fp32_dest_acc_en) {
+        exp_neg_x = _sfpu_exp_accurate_<true>(-x);
+    } else {
+        exp_neg_x = _sfpu_exp_21f_bf16_<true>(-x);
```
Aswinmcw, line 66: `<true>` in the `is_fp32_dest_acc_en == false` branch is correct (it suppresses the intermediate bf16 round so `calculate_situ_glu` rounds once at the store) but reads like an inverted flag next to `_sfpu_exp_accurate_<true>` above. Worth a one-line comment so a later cleanup doesn't "fix" it into a double-round precision regression.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_softcap.h
```diff
@@ -0,0 +1,49 @@
...
+// both matching stock calculate_tanh / tanh_tile rather than torch. Leaves rounding to the
+// caller, so fused callers can round once at the end.
+sfpi_inline sfpi::vFloat _sfpu_softcap_(sfpi::vFloat x, sfpi::vFloat beta, sfpi::vFloat inv_beta) {
+    return _sfpu_tanh_polynomial_(x * inv_beta) * beta;
+}
+
+inline void softcap_init() { tanh_init</*APPROXIMATION_MODE=*/false, /*is_fp32_dest_acc_en=*/false>(); }
+
+// beta and its reciprocal arrive as fp32 bit patterns.
+template <bool APPROXIMATION_MODE, bool is_fp32_dest_acc_en = false, int ITERATIONS = 8>
```
Aswinmcw, line 32: `APPROXIMATION_MODE` is never used in the body, yet `softcap_tile` passes the kernel-wide `APPROX` — a kernel built with `math_approx_mode=True` silently gets the exact polynomial. Either drop the parameter or `static_assert(!APPROXIMATION_MODE, "softcap has no approximate variant")` so it's a build error instead of a surprise.

Aswinmcw (approved):
Approving — one gap worth a follow-up commit (first comment), two are nits.

PR #52452, opened 2026-08-07
Make ttnn.prod accuracy observable in test_prod_all

iwroszTT:
### Problem

Nothing in `test_prod_all.py` can reliably detect a numerical error in `ttnn.prod`.

- `test_prod` gets its stimulus from `torch.randint(1, 5, shape)`. Every shape and dtype overflows the golden to `+inf`. Once the golden is non-finite the Frobenius check is switched off, and PCC is skipped for scalar outputs because `ttnn.prod` with no `dim` returns a single value (rank-0) which leaves `torch.allclose(inf, inf)` as the only assertion still doing anything.
- `test_prod_all_block_float` uses ones and `2.0`s: powers of two are exact at any mantissa width.
- `test_prod_all_block_float_nontrivial` tests the numerics, but with thresholds several times looser than the error it would need to catch.

### What this changes

**`test_prod` gets a stimulus whose product is bounded.** Dyadic factors (`k/2**n`, each exact in a few mantissa bits), renormalised by an exact power of two so the result lands in `[2**-0.5, 2**0.5]`. That avoids `+inf` and the denormals.

**The golden is computed in float64.** `torch.prod` on a bfloat16 tensor runs its own serial bfloat16 reduction, so comparing against it compares two approximations with different accumulation orders.

**Three new characterisation tests.** Every operand entering the FPU is truncated to the source-register mantissa, and `prod` multiplies every element of its input, so the relative error grows with the number of elements whose mantissa does not fit — not with tile count. Measured on Blackhole p100a:

| stimulus | relative error |
|---|---|
| 16 tiles, one full-mantissa factor per tile | −1.64e−2 |
| 1024 non-unit elements in a single tile | −3.06e−1 |
| 1024 dyadic elements in a single tile (control) | +6.28e−7 |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:iwrosz/prod-accuracy-tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:iwrosz/prod-accuracy-tests)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/reduction/test_prod_all.py
+++ b/tests/ttnn/nightly/unit_tests/operations/reduction/test_prod_all.py
@@ -11,13 +11,49 @@ from tests.ttnn.utils_for_testing import assert_numeric_metrics
 from tests.ttnn.nightly.unit_tests.operations.reduction.utility_functions import ttnn_prod
 
 
+# Dyadic factors (k / 2**n): each needs at most two stored mantissa bits, so every *operand*
+# survives the FPU source registers bit-for-bit, where those registers keep only 10 stored
+# mantissa bits and truncate the rest (see test_prod_all_fp32_truncation_accuracy below).
+# The running product is not exact - products of dyadic values are dyadic but their significands
+# grow, e.g. 1.5**15 = 3**15 / 2**15 needs 24 bits - so a few float32 ulps still accumulate in the
+# host-side reduction; measured worst case over these shapes is 3.9e-7, hence the 1e-5 tolerance.
+# The point is that no error comes from the input datapath, which keeps test_prod a check on
+# *what* prod computes rather than on how accurately it computes it.
+_DYADIC_FACTORS = (1.5, 0.75, 1.25, 0.625, 2.0, 0.5)
+
+
+def _exact_product_input(shape, cpu_dtype, seed=2023):
+    """Random dyadic factors, renormalised so the full product lands in [2**-0.5, 2**0.5].
+
+    A product of thousands of factors drifts exponentially. The previous `randint(1, 5)` stimulus
+    overflowed the golden to +inf for every shape in this test, which left `allclose(inf, inf)` as
+    the only assertion that still did anything; drawing factors around 1.0 instead pushes it the
+    other way, to ~1e-35 for the 4096-element shapes. That is only three orders of magnitude above
+    the smallest normal float32 (~1.2e-38, and bfloat16 has the same exponent range), so a
+    different seed or a larger shape would underflow into denormals where relative error stops
+    being meaningful.
+
+    Rounding log2 of the product to the nearest integer and dividing element 0 by that power of
+    two lands the product in [2**-0.5, 2**0.5]. Scaling by a power of two only changes an exponent
+    field, so no element becomes inexact.
+    """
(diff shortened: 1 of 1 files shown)


tests/ttnn/nightly/unit_tests/operations/reduction/test_prod_all.py
```diff
@@ -11,13 +11,49 @@
...
+# host-side reduction; measured worst case over these shapes is 3.9e-7, hence the 1e-5 tolerance.
+# The point is that no error comes from the input datapath, which keeps test_prod a check on
+# *what* prod computes rather than on how accurately it computes it.
+_DYADIC_FACTORS = (1.5, 0.75, 1.25, 0.625, 2.0, 0.5)
+
+
+def _exact_product_input(shape, cpu_dtype, seed=2023):
+    """Random dyadic factors, renormalised so the full product lands in [2**-0.5, 2**0.5].
+
+    A product of thousands of factors drifts exponentially. The previous `randint(1, 5)` stimulus
```
fplavecTT: "the previous" narrates what used to be here, which may be confusing to future reader. In most cases, comments that describe current state of code are more useful (of course it's OK to have a warning against doing something wrong, but it should be framed as such, not as a retrospective on past state of the code).

tests/ttnn/nightly/unit_tests/operations/reduction/test_prod_all.py
```diff
@@ -42,7 +78,12 @@ def test_prod(shapes, npu_dtype, device):
 
     (tt_input, tt_output, torch_input) = get_tensors(shapes, shapes, device, npu_dtype)
 
-    torch_output = torch.prod(torch_input)
+    # Reduce in float64 and cast back. torch.prod on a bfloat16 tensor runs its own serial
+    # bfloat16 reduction, which accumulates a different error than the device's and in a
```
fplavecTT: Similar, but different issue here. "bfloat16 tensor runs", "which accumulates",... make it sound as if the code is doing prod with bfloat16, which doesn't seem to be the case. Changeing the tense would make this correspond to code and easier to understand: "torch.prod on a bfloat16 tensor *would* run its own serial bfloat16 reduction, which *would* accumulate"...

fplavecTT (approved):
Great improvement to our test coverage

bbradelTT (approved):

Issue #47602, opened 2026-06-20
ttnn.slice can sometimes crash in a way similar to #45089


Hex27:
### Component / Area

ttnn

### Issue Type (optional)

Runtime Crash

### Observed

This error occurred when changing some tensor input sizes from 32 to 128

```
RuntimeError: TT_FATAL @ /home/user/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/tt_metal/impl/kernels/kernel.cpp:577: logical_core.x < this->core_to_runtime_args_.size() && logical_core.y < this->core_to_runtime_args_[logical_core.x].size()
info:
Cannot get runtime args for kernel slice_reader_unary_unpad_dims_rm_interleaved_start_id that is not placed on core 0-1
backtrace:
 --- /home/user/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/build_Release/lib/_ttnncpp.so(+0x1426638) [0x7706e2226638]
 --- tt::tt_metal::Kernel::runtime_args_data(tt::xy_pair const&)
 --- tt::tt_metal::GetRuntimeArgs(tt::tt_metal::Program const&, unsigned int, tt::xy_pair const&)
 --- ttnn::prim::SliceRmProgramFactory::override_runtime_arguments(tt::tt_metal::program_cache::detail::CachedProgram<ttnn::prim::SliceRmProgramFactory::shared_variables_t>&, ttnn::prim::SliceParams const&, ttnn::prim::SliceInputs const&, tt::tt_metal::Tensor&)
 --- /home/user/tt-xla/third_party/tt-mlir/src/tt-mlir/third_party/tt-metal/src/tt-metal/build_Release/lib/_ttnncpp.so(+0xaf523c) [0x7706e18f523c]
[...]

PR #46385, opened 2026-06-08
#45821: fix program-cache hash collisions (strong mixer + exact-key collision resolution)

dgomezTT:
Closes #45821.

This PR fixes program-cache hash collisions at **two levels**: it makes them astronomically improbable (a strong mixer), and then makes them impossible to observe as a wrong result (exact key comparison). The first is the minimal fix; the second is the structural guarantee.

---

## 1. Strong hash mixer (the acute fix)

`ttsl::hash::hash_objects` (the combiner every program-cache key flows through) used the classic **32-bit `boost::hash_combine`**:

```cpp
seed ^= hash_object(arg) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```

Its avalanche is poor for the small, structured integers that dominate our keys (tensor shapes, dtypes), so distinct sequences collide in 64 bits — e.g. shapes `[3, 17, 1, 1]` and `[1, 152, 1, 1]` (both reachable through `batch_norm`'s internal `permute`) hash **identically**. The cache is `unordered_map<uint64_t, …>` with no equality fallback, so `PermuteDeviceOperation` (no custom `compute_program_hash`) reused a program built for the wrong shape → wrong results. `disable_and_clear_program_cache()` between the two tests makes it pass, confirming cache reuse is the cause. Measured: **94,477 collisions** across the 2.56M shapes with dims in `[0, 40)⁴`.

There is no `std::hash_combine`, and `tt_stl` deliberately avoids Boost (a foundational, PCH'd header), so we inline the **splitmix64 finalizer (Stafford "variant 13")** — the same family of multiply-xorshift mixer Boost (≥1.81) and Abseil use internally — with only `<cstdint>`, **no new dependency**:

```cpp
hash_t x = seed + 0x9e3779b97f4a7c15ULL + hash_object(arg);
x = (x ^ (x >> 30)) * 0xbf58476d1ce4e5b9ULL;
x = (x ^ (x >> 27)) * 0x94d049bb133111ebULL;
seed = x ^ (x >> 31);
```

Collision-free across the full 2.56M-shape sweep. Residual collision probability at realistic cache sizes (thousands of entries) is ~`N²/2⁶⁵` ≈ **10⁻¹³**.

## 2. Exact key comparison (the structural guarantee)

10⁻¹³ is a *probability*, not a guarantee — the cache had opted out of the textbook hash-map contract by keying on the bare `uint64` and discarding the key. This restores it:

[...]

PR #52459, opened 2026-08-07
[ttnn] slice: drop the stale custom compute_program_hash (ON HOLD - costs one cache entry)

dgomezTT:
### Problem

`SliceDeviceOperation::compute_program_hash` exists to work around "weak distribution for
small-integer shape sequences (boost::hash_combine style)". That rationale no longer holds:

- the default combiner became splitmix64 in #46385;
- the framework additionally builds a collision-free **canonical key**, compared with `operator==`
  on every hash hit;
- declaring a custom hash **opts the op out of that canonical key**
  (`mesh_device_operation_adapter.hpp:871-883`).

So the hash that was added for safety leaves slice strictly less protected than an op on the default.

### Why the default is equivalent

`SliceParams` has no `attribute_names`, so the default reflection hash keys all 8 members — exactly
what the custom hash hand-listed. Everything else in the body is derived:

| custom hash term | under the default |
| --- | --- |
| the 8 attributes | reflected in full |
| `factory.index()` | pure function of (attributes, tensor_args), both keyed |
| `compute_output_specs(...)` | same |
| input / start_tensor `rank`, `logical`, `padded`, `layout`, `dtype`, `memory_config` | all derive from `TensorSpec = (logical_shape, tensor_layout)`, keyed whole |
| — | default additionally keys `end_tensor` and `preallocated_output`, which the custom hash ignored — this is **not** free, see below |

`override_runtime_arguments` is untouched — it re-applies per-dispatch state and is unrelated to keying.

### Measured cost: one extra entry on the preallocated-output axis

`preallocated_output` rides `tensor_args`, so the default keys it and a slice writing into a
caller-provided output no longer shares the plain slice's program. Measured, plain slice then one
`output_tensor=` slice:

| | old hash | new hash |
| --- | --- | --- |
| plain slice | 1 entry | 1 entry |
| + one preallocated-output slice | **2** | **3** |

The old reuse was legitimate — the output buffer rides a binding, so one program serves both — so this
is a genuine cache-utilization cost of one entry on this axis, not a false-hit fix. Pinned by
`test_slice_cache_preallocated_output_adds_entry` so it stays visible. Everywhere else measured, counts
are unchanged.

### Verification (wormhole, single device)

Baseline measured on the unmodified tree, then re-measured after removal:

| suite | before | after |
| --- | --- | --- |
| `test_slice_program_cache.py` (10 cache-entry assertions, added here) | 8 passed (2 new tests pin post-change behaviour) | 10 passed |
| `test_slice.py::test_slice_rm_sharded_with_program_cache` (`total == 3`) | passed | passed |
| `test_slice.py::test_slice_rm_program_cache_collison` (#47602 false-hit net) | passed | passed |
| `test_slice.py` full | 439 passed, 38 skipped | 439 passed, 38 skipped |

`test_slice_cache_hit_reapplies_buffers` is the freeze test: three dispatches of different random data
in different buffers on one cached program, each PCC-checked against its own input. A frozen buffer
address or per-dispatch scalar fails it.

`nm` on the rebuilt `_ttnncpp.so` confirms `ttnn::prim::SliceDeviceOperation::compute_program_hash`
is gone, so the runs exercised the default hash and not a stale binary.

### Not covered

- Built with `--disable-profiler`; correctness and cache-entry counts only, no perf claim.
- Single-device only. The default hash folds in mesh coordinates where the custom one did not, so a
  multi-device run is worth having before merge.

### DIFF
--- a/tests/ttnn/unit_tests/operations/data_movement/test_slice_program_cache.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_slice_program_cache.py
@@ -1,0 +1,168 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Program-cache behaviour for ttnn.slice.
+
+SliceDeviceOperation carried a custom compute_program_hash that hand-listed all 8
+SliceParams members plus derived state (factory index, output spec, per-tensor
+shape/layout/dtype). SliceParams has no attribute_names, so the framework default
+hash keys the same 8 members, derives the rest, and additionally keys end_tensor
+and preallocated_output -- which the custom hash ignored.
+
+These tests pin the cache-entry counts on every axis so the removal is proven,
+not asserted: they were written against the custom hash and must give identical
+counts without it.
+"""
+
+import pytest
+import torch
+
+import ttnn
+from tests.ttnn.utils_for_testing import assert_with_pcc
+
+
+@pytest.fixture
+def isolate_program_cache(device):
+    """Ensure each test starts with an empty program cache and cleans up after."""
+    device.disable_and_clear_program_cache()
+    device.enable_program_cache()
+    yield
+    device.disable_and_clear_program_cache()
+
+
+def run_slice(device, shape, begins, ends, layout=ttnn.TILE_LAYOUT, memory_config=ttnn.DRAM_MEMORY_CONFIG):
+    torch_input = torch.randn(shape, dtype=torch.bfloat16)
+    slices = tuple(slice(begins[i], ends[i]) for i in range(len(begins)))
+    torch_output = torch_input[slices]
+
+    tt_input = ttnn.from_torch(
+        torch_input, layout=layout, dtype=ttnn.bfloat16, device=device, memory_config=memory_config
+    )
+    with device.cache_entries_counter.measure():
+        tt_output = ttnn.slice(tt_input, begins, ends, memory_config=memory_config)
+
+    assert_with_pcc(torch_output, ttnn.to_torch(tt_output), 0.999)
+
+
+def run_slice_tensor_args(device, shape, begins, ends, dim, layout=ttnn.TILE_LAYOUT):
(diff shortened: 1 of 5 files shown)


dgomezTT:
HOLD - do not merge as-is. Measured behaviour change, not a neutral cleanup.

The removal costs **one extra program-cache entry** when an app slices both with and without a
preallocated output. Measured on two binaries differing only in this change:

| sequence | old hash | new hash |
| --- | --- | --- |
| plain slice alone | 1 | 1 |
| preallocated-output slice alone | 2 | 2 |
| the same preallocated slice twice | 2 | 2 |
| plain slice, then preallocated slice | **2** | **3** |

Isolated, the two hashes are identical. The delta is only in the mixed sequence: under the custom hash
the preallocated dispatch reused the plain dispatch program, and under the default it does not, because
`preallocated_output` is a `std::optional<Tensor>` in `SliceInputs` and the default hash keys engaged
vs disengaged.

**That reuse was safe, so the extra entry buys nothing.** Verified directly: on the old binary, running
plain-then-preallocated with *different* input data returns the correct preallocated result (PCC 0.999).
The output buffer rides a binding, so one program legitimately serves both. This is over-keying, not a
false-hit fix - correcting what an earlier revision of this description claimed.

`slice` is the only op in this batch with an optional tensor in `tensor_args`; `Pool2D::tensor_args_t`
and `ReshapeViewInputs` each hold a single required tensor, so #52460 and #52461 have no equivalent
exposure.

dgomezTT:
Closing: this changes behaviour. The removal costs one extra program-cache entry when an app slices
both with and without a preallocated output (plain-then-preallocated: 2 entries before, 3 after), and
the reuse it breaks was verified safe, so the extra entry buys no correctness. A cleanup has to be
cache-entry-neutral; this one is not, so it does not land.

The underlying observation stands for whenever this is revisited: the hash justifies itself with a
weak-distribution rationale that #46385 made obsolete, and declaring it opts slice out of the framework
canonical key. Reviving it needs a way to keep the optional preallocated_output out of the key without
a freeform hash.

PR #52391, opened 2026-08-07
[feature] Implement vararg CTA for metal 2.0

riverwuTT:
### PR Category

Feature

### Summary

vararg CTA allows an op writer to specify a dynamic number of positional CTA for their kernel.

This PR adds Metal 2.0 support for this feature, this is needed to unblock porting of concat op.

Closes #45388

### Notes for reviewers
- This PR reuses the old positional CTA injection system, where vararg CTA are placed at the 0-th index of the CTA array, while other systems using the positional CTA (mainly tensor binding) follows the vararg CTA.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/vararg-cta)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/vararg-cta)

PR #52490, opened 2026-08-07
[Cleanup] Tighten vararg CTA comments

### LINKED ISSUE #45388: Add vararg CTAs as a temporary shim for variadic ops
With Metal 2.0, we want to get rid of positional arguments in favor of named arguments.

The trouble is that some kernels are written to accept a variable number of arguments. Metal 2.0 currently supports runtime varargs for this reason. We intend to deprecate this once typed argument support is available.

Variadic kernels (e.g. `concat`, `halo`) that accept a variable number of tensor arguments actually _need_ to take a variable number of positional CTAs. This wasn't anticipated. 

This should also go away with `std::array` arguments, but we need it in the short term.

riverwuTT:
Stacked on `riverwu/vararg-cta` — targets that branch, not `main`. Addresses the "Cleanup slop comments" TODO on #52391.

Comment-only. No functional change; no line of code moved.

### The main thing

The sentence *"Metal 2.0 CTA varargs do not use the legacy/positional CTA infrastructure"* appeared in **9 places** across the 8 files. It is now stated once, canonically, in `compile_time_args.h` — the header a confused kernel author actually opens — with short references elsewhere.

Two of those 9 were a **verbatim** duplicate: the same 3-line block above `kernel_compile_time_args` and again above `get_ct_arg` five lines later.

### Everything else

| | |
|---|---|
| Grammar / typos | `"vararg CTAs does not"` → *do not* (x2), `"Metal 2.0 user should"` (x2), a comma splice in `program_spec.cpp`, `Legacy` → `legacy`, en-dash in `non–Metal 2.0` |
| Wrong-line comment | `// Binding CTA_OFFSET stays 0` sat above `static_assert(...is_dram)`, which does not check `cta_offset` — that check is on the host side, 20 lines down. Retargeted to what the assert actually proves. (`CTA_OFFSET` also is not an identifier; the field is `cta_offset`.) |
| Thin public API doc | `KernelAdvancedOptions::compile_time_varargs` had *"Values of the varargs for the kernel."* — the least-documented field in the struct, despite being the feature. Now says how to read them in a kernel and that they are part of the cache key. |
| Unexplained non-obvious code | `compute_hash` hashing the vararg *length* had a comment that did not mention length. Now explains it, reusing the file's existing `[a, b] vs [ab] collision` phrasing. |
| Pure restatement | `// Metal 2.0 path for injecting compile time varargs.` above `kernel->set_compile_time_varargs(...)`, and the numbered list of three accessors directly above the code emitting exactly those three accessors. |

### Notes for reviewers

- **Design question left open, not answered.** Your PR body flags the `get_ct_arg` signpost as "open for debate". I kept the signpost and only deduplicated it. If the outcome is a hard error rather than a comment, this all changes.
- `get_compile_time_arg_val` (the documented public macro) has a doxygen block I did **not** touch. It already scopes itself to "provided during kernel creation using `CreateKernel` calls", which arguably excludes varargs on its own — but it is the one place a kernel author is most likely to read. Say the word and I will add a line there.
- The two hash tests sit under the new banner, but the pre-existing `Kernel hash sensitivity to TensorParameter spec` banner starts ~15 lines later. They may belong there instead. Left alone — that is a move, not a comment fix.

### Out of scope, flagged not fixed

- `get_compile_time_vararg(uint32_t idx)` (the runtime-index overload) has **no bounds check**. With zero varargs it indexes a zero-length array. The template overload `static_assert`s; this one does not.
- Nothing in the tree references **#45388** any more — the `TODO` that cited it was the thing this PR removed. Might be worth a `Closes #45388` in the #52391 description.

### Verification

Comment-only, so no rebuild. Confirmed clang-format-clean: every changed line is within the 120-col limit and `clang-format` (repo `.clang-format`, `ReflowComments: true`) reflows none of the new comments. The violations my local clang-format 20 reports in these files are all pre-existing and outside the changed lines — a version skew against the repo pin, not something this PR introduces.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:brain/vararg-cta-comments)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:brain/vararg-cta-comments)

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -4134,12 +4134,15 @@ TEST_F(ProgramSpecTestGen1, TtKernelComputeShimCompiles) {
 }
 
 // ============================================================================
-// Compile-time varargs — mock-device bake & JIT smoke
+// Compile-time varargs (Metal 2.0)
 // ============================================================================
 //
-// Host: KernelAdvancedOptions::compile_time_varargs (values fixed at ProgramSpec time).
-// Kernel: get_compile_time_vararg* — literals baked into kernel_args_generated.h (Metal 2.0).
-// Separate from positional KERNEL_COMPILE_TIME_ARGS (TensorBinding DSpec / legacy CreateKernel).
+// KernelAdvancedOptions::compile_time_varargs values are fixed at ProgramSpec time and baked into
+// kernel_args_generated.h as literals, which is what makes a static_assert in the kernel source a
+// usable assertion here: if a value did not reach the kernel, the JIT compile fails.
+//
+// These tests compile on a MOCK Wormhole device (no silicon, no dispatch) via
+// detail::CompileProgram, so they cover the host -> genfiles -> RISC-V compile path only.
 
 TEST_F(ProgramSpecTestGen1, CompileTimeVarargsReadableFromKernel) {
     NodeCoord node{0, 0};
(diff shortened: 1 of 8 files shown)


riverwuTT:
what??

PR #52496, opened 2026-08-07
tt-train: add AttentionMaskType::None for bidirectional attention in fused SDPA

antonibertel:
## Summary

Adds bidirectional (unmasked) attention support to `ttml::ops::scaled_dot_product_attention()`. MAE-style models need every position attending to every other position; today that requires passing an all-ones "Arbitrary" mask, which computes and adds a zero bias for no benefit.

New `mask_type` parameter (`std::optional<AttentionMaskType>`) lets a caller request `None`, `Causal`, or `Arbitrary` explicitly, reusing the enum tt-train already has internally.

**Backward compatible**: unset reproduces the prior behavior (Arbitrary if a mask is given, Causal otherwise). Passing `Arbitrary` without a mask, or a mask with a non-`Arbitrary` type, throws.

## Changes

- `ops/scaled_dot_product_attention.{cpp,hpp}`: `mask_type` param, resolved to `AttentionMaskType::None` when requested
- `nanobind/nb_ops.cpp`: exposes `mask_type` and the `AttentionMaskType` enum (`NONE`/`CAUSAL`/`ARBITRARY`, registered by hand because `None` is a Python keyword)
- `sdpa_fw` compute kernel: `AttentionMaskType::None` skips `apply_mask_on_reg` and the mask CB pop (previously the non-causal path unconditionally touched a mask CB that isn't created for `None`)
- `sdpa_bw` Q kernel: `compute_u_scalar_row` row-sum now uses `sfpu_reduce<SUM, Float32, REDUCE_ROW>` on DST instead of a matmul with a ones column; the ones-column CB (`c_7`) and its writer generation are removed from the Q kernel since nothing reads it anymore
- `tests/ops/sdpa_bw_op_test.cpp`: float and composite references run unmasked for `None`; nightly test at a ViT-L MAE encoder shape (B=24, H=16, S=544, D=64)

## Testing

Validated against a ViT-L-scale MAE pretraining workload on Blackhole hardware.

https://claude.ai/code/session_01LiUNkrLQK3YEbTbkecAMC1


antonibertel:
Built and verified on hardware after merging upstream `main` (c3ce01c9) — **Blackhole p300c**, 1 chip used, tt-metal built in the `tt-metalium/ubuntu-22.04-dev` container (`build_metal.sh --build-tt-train`, clang-20). Host build, `ttml_tests`, and the `_ttml` nanobind module compiled without errors; device kernels compiled on first run without errors (only the pre-existing `mask_addr_generator` unused-variable warning in `sdpa_bw_q_reader_kernel.cpp`, which this PR does not touch).

**C++ tests run (`./build/tt-train/tests/ttml_tests`)**

| Test | Runs | Result |
|---|---|---|
| `SDPABackwardTest.NIGHTLY_NoMask_ProductionEncoderShape` (B=24,S=544,H=16,D=64, mask=None) | 4 | PASS (~28–35 s each) |
| `SDPABackwardTest.NIGHTLY_NanoGPTConfig` (Arbitrary mask) | 3 | PASS |
| `SDPABackwardTest.NIGHTLY_LargerSequence` (Arbitrary mask) | 3 | PASS |
| `SDPABackwardTest.NIGHTLY_CausalMask_NanoGPTConfig` | 4 | PASS |
| `SDPABackwardTest.NIGHTLY_CausalMask_LargerSequence` | 5 | PASS (see note) |
| `SDPABackwardTest.NIGHTLY_RingAttentionMergeSimulation` (uses `AttentionMaskType::None` in the KV path; run with its hard-coded `GTEST_SKIP` temporarily removed, not committed) | 3 | PASS |
| `SDPABackwardTest.NoMask_SmallBatch` (new, B=2,S=128,H=4,D=64, mask=None) | 8 | PASS, deterministic |
| `SDPABackwardTest.DiffVDim_SingleTile`, `ShapeMismatch_*`, `Validation_*` | 3 | PASS |
| `SDPAForwardTest.*` (18 tests incl. both `NIGHTLY_*`, NoMask/Causal/Arbitrary/DiffVDim/GQA) | 1 | 18/18 PASS |

The `SDPABackwardTest.*` suite was also run as one process three times: 13/13 pass each time. Note: the SDPA `NIGHTLY_*` tests are gated only by CI's gtest filter, not by `ENABLE_NIGHTLY_TT_TRAIN_TESTS`, so they were run by name.

**Note on one hang:** the very first full-suite run hung inside `NIGHTLY_CausalMask_LargerSequence` (host spinning in `completion_queue_wait_front` during a tensor readback, after `NanoGPTConfig`, `LargerSequence`, `CausalMask_NanoGPTConfig` had passed in the same process). After `tt-smi -r` it never reproduced: that test passed 5/5 afterwards, alone, paired with the preceding test, and inside three more full-suite runs. I could not tell whether it is related to this PR or a pre-existing intermittent issue; flagging it rather than hiding it.

**Python (`_ttml` nanobind)**
- `ttml.ops.attention.AttentionMaskType` exposes `NONE`, `CAUSAL`, `ARBITRARY`.
- `scaled_dot_product_attention(q, k, v, mask_type=AttentionMaskType.NONE)` vs `torch.nn.functional.scaled_dot_product_attention` with no mask (B=2,H=4,S=128,D=64, bf16 inputs): out/dQ/dK/dV all within atol=rtol=3e-2 (max abs err 0.0035 / 0.0058 / 0.0069 / 0.0077). `mask_type=CAUSAL` and the no-argument default both match `is_causal=True` reference. NONE output differs from the causal reference (i.e. the mask really is skipped). Passing a mask tensor together with `mask_type=NONE` raises `ValueError` as intended.
- `pytest tt-train/tests/python/test_deepseek_mla_sdpa.py`: 1 passed.

**Formatting:** `git clang-format --diff upstream/main` (clang-format 19.1.4, same as the pre-commit pin) reports no changes for the PR diff; the remaining pre-commit hooks pass on the changed files.

Not run: Wormhole. Commit `f457b738` adds the small non-nightly test; the merge commit `c3ce01c9` brings in upstream `main`.

antonibertel:
Hey hey @minaliuTT, you marked that PR as high complexity and high risk, even though it only changes a few files. The risk may indeed be high, but if you have good test coverage on your end, maybe the complexity isn't actually that high?

PR #52500, opened 2026-08-07
Add fused distributed GroupNorm

### LINKED ISSUE #50143: Add support fused distributed group norm
Add support fused distributed group norm typically used in dit VAES. The plan is to align it with the currently supported distributed rms norm and layer norm.

sadesoyeTT:
### PR Category
Feature

### Summary
This PR adds fused spatiality distributed group-norm so tensors stay sharded without the need to gather and re_partition 
     To link an issue: Closes #50143 

### Notes for reviewers
Only integrated in DiT models (specifically incoming Flux2). This implementation had been separated from existing groupnorm op to reduce the chance of breaking any model that uses the current groupnorm. It basically applies the chan's parallel algoritm over the current wellford's  They can be merged in the future.

### Relevant tests
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t1-unit-tests.yaml?query=branch:sadesoye/fused_silu_distributed_GN) — `model=dit-common` runs `models/tt_dit/tests/unit/`, which covers the fused
  distributed GroupNorm on N150 (local) and BH QuietBox (4-device, exercises the fabric all-gather).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:sadesoye/fused_silu_distributed_GN)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:sadesoye/fused_silu_distributed_GN)

### DIFF
--- a/models/tt_dit/layers/normalization.py
+++ b/models/tt_dit/layers/normalization.py
@@ -11,7 +11,12 @@ import torch
 
 import ttnn
 
+from .linear import _apply_activation_fn
 from .module import Module, Parameter
+
+# Activations the fused distributed GroupNorm op can absorb into its output stage.
+# Anything else is applied as a separate op, exactly like Linear's unfused activations.
+_FUSABLE_GROUP_NORM_ACTIVATIONS = frozenset({"silu"})
 
 
 class RMSNorm(Module):
@@ -444,6 +449,7 @@ class GroupNorm(Module):
         mesh_device: ttnn.MeshDevice,
         mesh_axis: int | None = None,
         core_grid: ttnn.CoreGrid | None = None,
+        activation_fn: str | None = None,
     ) -> None:
         """
         Args:
@@ -454,9 +460,12 @@ class GroupNorm(Module):
             mesh_axis: The mesh axis to use for sharding.
             core_grid: The core grid to use.
             num_out_blocks: The number of output blocks to use.
+            activation_fn: Applied after the norm. ``ttnn.group_norm`` cannot fuse one, so it
+                always runs as a separate op here.
         """
         super().__init__()
 
+        self.activation_fn = activation_fn
         self.eps = eps
         self.mesh_device = mesh_device
         self.mesh_axis = mesh_axis
@@ -569,7 +578,231 @@ class GroupNorm(Module):
         x = ttnn.group_norm(x, **kwargs)
         x = x.reshape([batch_size, height, width, channels])
 
-        return x
+        return _apply_activation_fn(x, self.activation_fn)
+
+
+class DistributedGroupNorm(Module):
+    """GroupNorm with optional spatial stats all-gather on ``cluster_axis``.
+
+    ``cluster_axis=None``: local ``GroupNorm`` (including channel ``mesh_axis`` TP).
+    ``cluster_axis=int``: fused ``dit_fused_distributed_groupnorm``
+    (PRE → fabric AG on that axis → POST). γ/β and the input mask use the same
+    DRAM-packed row-major GroupNorm layout as ``ttnn.group_norm``, with
+    ``num_virtual_cols == 1``. Optional ``mesh_axis`` still shards channels
+    (whole groups per device); orthogonal to ``cluster_axis``. Per-device
+    channel count must be divisible by 32.
+
+    ``activation_fn`` (e.g. ``"silu"``) is applied after the norm. On the distributed path it
+    is folded into the op's output stage, removing the separate activation op's full-tensor
+    DRAM read/write; otherwise it runs as a separate op. Either way the caller gets
+    ``activation(group_norm(x))`` and does not need to know which happened.
+    """
+
+    def __init__(
+        self,
+        num_channels: int,
+        num_groups: int,
+        *,
+        eps: float = 1e-5,
+        mesh_device: ttnn.MeshDevice,
+        cluster_axis: int | None = None,
+        mesh_axis: int | None = None,
+        ccl_manager=None,
+        core_grid: ttnn.CoreGrid | None = None,
+        activation_fn: str | None = None,
+    ) -> None:
+        super().__init__()
+
+        assert num_channels % num_groups == 0, "num_channels must be divisible by num_groups"
+        assert num_channels % 32 == 0, "num_channels must be divisible by tile size (fused v1)"
+
+        # Consume the activation into the op when the fused path can absorb it; whatever is
+        # left in self.activation_fn is applied separately by forward().
+        self.activation_fn = activation_fn
+        self.fused_activation = None
+        if cluster_axis is not None and activation_fn in _FUSABLE_GROUP_NORM_ACTIVATIONS:
+            self.fused_activation = activation_fn
+            self.activation_fn = None
+
+        self.num_channels = num_channels
+        self.num_groups = num_groups
+        self.eps = eps
+        self.mesh_device = mesh_device
+        self.cluster_axis = cluster_axis
+        self.mesh_axis = mesh_axis
+        self.ccl_manager = ccl_manager
+        self.core_grid = core_grid or ttnn.CoreGrid(x=8, y=8)
+
+        self.channel_devices = tuple(mesh_device.shape)[mesh_axis] if mesh_axis is not None else 1
+        self.cluster_size = tuple(mesh_device.shape)[cluster_axis] if cluster_axis is not None else 1
+        assert num_groups % self.channel_devices == 0, "num_groups must be divisible by channel mesh width"
+
+        if cluster_axis is not None and self.cluster_size > 1:
+            assert ccl_manager is not None, "ccl_manager is required when cluster_axis width > 1"
+
+        self.num_local_channels = num_channels // self.channel_devices
+        self.num_local_groups = num_groups // self.channel_devices
+
+        if cluster_axis is not None:
+            assert (
+                self.num_local_channels % 32 == 0
+            ), f"fused path requires per-device C % 32 == 0, got {self.num_local_channels}"
+            assert core_grid is None, (
+                "core_grid is not supported on the fused path (cluster_axis is not None): "
+                "dit_fused_distributed_groupnorm always uses the full compute grid"
+            )
+
+        if cluster_axis is None:
+            # Existing GroupNorm path (packed γ/β/mask, channel TP).
+            self.inner = GroupNorm(
+                num_channels=num_channels,
+                num_groups=num_groups,
+                eps=eps,
+                mesh_device=mesh_device,
+                mesh_axis=mesh_axis,
+                core_grid=self.core_grid,
+            )
+            self.weight = None
+            self.bias = None
+        else:
+            # Fused path: reuses the welford GroupNorm kernels, so γ/β/mask are prepped exactly
+            # like ttnn.group_norm (DRAM-packed RM γ/β + input_mask), but with a single worker
+            # core per device (num_virtual_cols == 1).
+            self.inner = None
+            num_padded_channels = math.ceil(self.num_local_channels / 32) * 32
+            assert num_padded_channels % self.num_local_channels == 0, "padded channels must divide channels"
+            num_padded_groups = self.num_local_groups * (num_padded_channels // self.num_local_channels)
+            self.num_virtual_cols = 1
+            self.num_padded_channels = num_padded_channels
+            self.num_padded_groups = num_padded_groups
+
+            weight_shape = [
+                self.channel_devices,
+                1,
+                math.ceil(num_padded_channels // self.num_virtual_cols / 32) * self.num_virtual_cols,
+                32,
+            ]
+            block_wt = ttnn.operations.normalization.find_max_tile_span(
+                num_padded_channels, num_padded_channels // num_padded_groups, 32
+            )
+            mask_shape = [1, num_padded_groups, 32, 32 * block_wt]
+
+            self.weight = Parameter(
+                total_shape=weight_shape,
+                layout=ttnn.ROW_MAJOR_LAYOUT,
+                mesh_axes=[mesh_axis, None, None, None],
+                device=mesh_device,
+            )
+            self.bias = Parameter(
+                total_shape=weight_shape,
+                layout=ttnn.ROW_MAJOR_LAYOUT,
+                mesh_axes=[mesh_axis, None, None, None],
+                device=mesh_device,
+            )
+            self.mask = Parameter(total_shape=mask_shape, device=mesh_device)
+
+    @classmethod
+    def from_torch(
+        cls,
+        torch_ref: torch.nn.GroupNorm,
+        *,
+        mesh_device: ttnn.MeshDevice,
+        cluster_axis: int | None = None,
+        mesh_axis: int | None = None,
+        ccl_manager=None,
+        core_grid: ttnn.CoreGrid | None = None,
+        activation_fn: str | None = None,
+    ) -> DistributedGroupNorm:
+        module = cls(
+            num_channels=torch_ref.num_channels,
+            num_groups=torch_ref.num_groups,
+            eps=torch_ref.eps,
+            mesh_device=mesh_device,
+            cluster_axis=cluster_axis,
+            mesh_axis=mesh_axis,
+            ccl_manager=ccl_manager,
+            core_grid=core_grid,
+            activation_fn=activation_fn,
+        )
+        module.load_torch_state_dict(torch_ref.state_dict())
+        return module
+
+    def _prepare_torch_state(self, state: dict[str, torch.Tensor]) -> None:
+        if self.inner is not None:
+            renamed = {f"inner.{k}": v for k, v in list(state.items())}
+            state.clear()
+            state.update(renamed)
+            return
+        if "weight" in state:
+            state["weight"] = self._prepare_param(state["weight"])
+        if "bias" in state:
+            state["bias"] = self._prepare_param(state["bias"])
+        input_mask = ttnn.create_group_norm_input_mask(
+            self.num_padded_channels, self.num_padded_groups, self.num_virtual_cols
+        )
+        state["mask"] = ttnn.to_torch(input_mask)
+
+    def _prepare_param(self, param: torch.Tensor) -> torch.Tensor:
+        expected_shape = (self.num_local_channels * self.channel_devices,)
+        assert param.shape == expected_shape, f"expected shape {expected_shape}, got {param.shape}"
+        padding = self.num_padded_channels - self.num_local_channels
+        params = [torch.nn.functional.pad(t, (0, padding)) for t in param.chunk(self.channel_devices)]
+        torch_sharded_lst = [
+            ttnn.create_group_norm_weight_bias_rm(t, self.num_padded_channels, self.num_virtual_cols) for t in params
+        ]
+        return torch.cat(torch_sharded_lst, dim=0)
+
+    def forward(self, x: ttnn.Tensor, num_out_blocks=-1, compute_kernel_config=None) -> ttnn.Tensor:
+        if self.inner is not None:
+            out = self.inner.forward(x, num_out_blocks=num_out_blocks, compute_kernel_config=compute_kernel_config)
+            return _apply_activation_fn(out, self.activation_fn)
+
+        assert num_out_blocks == -1, (
+            "num_out_blocks is not supported on the fused path (cluster_axis is not None): "
+            "dit_fused_distributed_groupnorm derives its own block split"
+        )
+
+        batch_size, height, width, channels = x.shape
+        assert (
+            channels == self.num_local_channels
+        ), f"last dim {channels} != num_local_channels {self.num_local_channels}"
+        x4 = x.reshape([batch_size, 1, width * height, channels])
+
+        semaphores = self.ccl_manager.get_ag_ping_pong_semaphore(self.cluster_axis) if self.cluster_size > 1 else []
+        persistent = None
+        if self.cluster_size > 1:
+            # The fused activation does not change the stats geometry (it is applied after the
+            # AG, on the output tile), so it is deliberately NOT part of the cache key.
+            persistent = self.ccl_manager.get_fused_norm_stats_buffer(
+                ("gn", tuple(x4.shape), self.num_padded_groups, self.cluster_axis),
+                lambda: ttnn.experimental.dit_fused_distributed_groupnorm_create_stats_buffer(
+                    x4,
+                    self.num_padded_groups,
+                    self.cluster_axis,
+                    self.mesh_device,
+                ),
+            )
+
+        # Mirror GroupNorm.forward kwargs; cluster_axis / mesh / semaphore / topology
+        # are the only distributed additions.
+        out = ttnn.experimental.dit_fused_distributed_groupnorm(
+            x4,
+            num_groups=self.num_padded_groups,
+            epsilon=self.eps,
+            cluster_axis=self.cluster_axis,
+            mesh_device=self.mesh_device,
+            multi_device_global_semaphore=semaphores,
+            topology=self.ccl_manager.topology if self.ccl_manager is not None else ttnn.Topology.Linear,
+            input_mask=self.mask.data,
+            weight=self.weight.data,
+            bias=self.bias.data,
+            persistent_output_buffer=persistent,
+            compute_kernel_config=compute_kernel_config,
+            activation=self.fused_activation,
+        )
+        out = out.reshape([batch_size, height, width, channels])
+        # No-op when the activation was consumed into the op above.
+        return _apply_activation_fn(out, self.activation_fn)
 
 
 class GroupNorm3D(Module):
--- a/models/tt_dit/models/vae/vae.py
+++ b/models/tt_dit/models/vae/vae.py
@@ -12,9 +12,9 @@ import ttnn
 from models.common.utility_functions import is_blackhole
 
 from ...layers.conv2d import Conv2d
-from ...layers.linear import ColParallelLinear, RowParallelLinear
+from ...layers.linear import ColParallelLinear, RowParallelLinear, _apply_activation_fn
 from ...layers.module import Module, ModuleList, Parameter
-from ...layers.normalization import DistributedRMSNorm, GroupNorm, RMSNorm
+from ...layers.normalization import DistributedGroupNorm, DistributedRMSNorm, GroupNorm, RMSNorm
 from ...parallel.manager import CCLManager
 from ...utils import tensor
 from ...utils.conv3d import get_conv3d_config
@@ -109,8 +109,11 @@ def _get_neighbor_pad_num_links(ccl_manager: CCLManager, input_tensor: ttnn.Tens
 
 
 class VaeRmsNorm(Module):
-    def __init__(self, num_channels: int, *, eps: float, ctx: VaeContext) -> None:
+    def __init__(self, num_channels: int, *, eps: float, ctx: VaeContext, activation_fn: str | None = None) -> None:
         super().__init__()
+
+        # Neither RMSNorm path here fuses an activation, so it always runs as a separate op.
+        self.activation_fn = activation_fn
 
         tp_axis_size = ctx.device.shape[ctx.tp_axis] if ctx.tp_axis is not None else 1
 
@@ -156,7 +159,7 @@ class VaeRmsNorm(Module):
             bs, h, w, c = x.shape
             x = ttnn.reshape(x, [1, 1, bs * h * w, c])
             x = self.inner.forward(x)
-            return ttnn.reshape(x, [bs, h, w, c])
+            return _apply_activation_fn(ttnn.reshape(x, [bs, h, w, c]), self.activation_fn)
 
         norm = ttnn.mean(ttnn.pow(x, 2), dim=-1, keepdim=True)
 
@@ -171,7 +174,7 @@ class VaeRmsNorm(Module):
             norm = norm * (1 / n)
 
         norm = ttnn.rsqrt(norm + self._eps)
-        return x * (norm * self.gamma.data)
+        return _apply_activation_fn(x * (norm * self.gamma.data), self.activation_fn)
 
 
 class _VaeConv2dConv3d(Module):
@@ -391,9 +394,9 @@ class VaeResnetBlock(Module):
     def __init__(self, *, in_channels: int, out_channels: int, norm: VaeNormDesc, ctx: VaeContext) -> None:
         super().__init__()
 
-        self.norm1 = _norm(norm, num_channels=in_channels, ctx=ctx)
+        self.norm1 = _norm(norm, num_channels=in_channels, ctx=ctx, activation_fn="silu")
         self.conv1 = VaeConv2d(in_channels, out_channels, kernel_size=3, padding=1, ctx=ctx)
-        self.norm2 = _norm(norm, num_channels=out_channels, ctx=ctx)
+        self.norm2 = _norm(norm, num_channels=out_channels, ctx=ctx, activation_fn="silu")
         self.conv2 = VaeConv2d(out_channels, out_channels, kernel_size=3, padding=1, ctx=ctx)
 
         self.conv_shortcut = (
@@ -416,11 +419,9 @@ class VaeResnetBlock(Module):
         h = x
 
         h = self.norm1.forward(h)
-        h = ttnn.silu(h, output_tensor=h)
         h = self.conv1.forward(h)
 
         h = self.norm2.forward(h)
-        h = ttnn.silu(h, output_tensor=h)
         h = self.conv2.forward(h)
 
         if self.conv_shortcut is not None:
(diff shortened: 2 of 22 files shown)


vtsilytskyiTT (commented):
This PR adds new kernels with an old API. Did you try to utilize Metal 2.0 API? It is object-oriented and error-proof. See docs/source/tt-metalium/tt_metal/apis/kernel_apis/data_movement/device_api_migration_guide.md and already migrated kernels.

sadesoyeTT:
> This PR adds new kernels with an old API. Did you try to utilize Metal 2.0 API? It is object-oriented and error-proof. See docs/source/tt-metalium/tt_metal/apis/kernel_apis/data_movement/device_api_migration_guide.md and already migrated kernels.

Thanks for the comment. Initial implementation pretty much copied over the groupnorm code. That has been removed now, and the current implementation now mostly reuses exiting kernels. PTAL again.

jonathansuTT (approved):

vsureshTT (approved):
Nice pr, looks solid

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
```diff
@@ -365,13 +443,140 @@ void kernel_main() {
...
+        noc_async_atomic_barrier();
+        cb_pop_front(stats_local_cb_id, 1);
+
+        // 2. Wait for the ring gather to land in DRAM (forwarder increments our go-sem).
+        volatile tt_l1_ptr uint32_t* go_sem_ptr = reinterpret_cast<volatile tt_l1_ptr uint32_t*>(go_sem_addr);
+        noc_semaphore_wait_min(go_sem_ptr, 1);
+        noc_semaphore_set(go_sem_ptr, 0);
+
+        // 3. Read the ring_size sub-sticks (our slot) from DRAM into the gathered CB.
+        cb_reserve_back(stats_gathered_cb_id, ring_size);
```
bbradelTT: Please use device 1.1 / 2.0 APIs

sadesoyeTT: Agreed! It can certainly be done in a future PR for the entire GN code. However, I didn't want to touch code path that other production models currently rely on, and I'm not sure how much additional code bloat mixing both APIs will cause.

models/tt_dit/models/vae/vae.py
```diff
@@ -694,47 +714,58 @@ def _prepare_torch_state(self, state: dict[str, torch.Tensor]) -> None:
...
-        # zero-padding rows that mesh_partition may have added for tile alignment.
-        logical_h_full = x.shape[1] * self._ctx.h_factor
-
-        x = _all_gather_hw(self._ctx, x)  # → [N, H_gathered, W, C], TILE_LAYOUT
-
-        padded_h = x.shape[1]
-        needs_trim = padded_h > logical_h_full
-        if needs_trim:
-            x = x[:, :logical_h_full, :, :]
+        if self._gather_w:
```
ayerofieiev-tt, line 720: **Bug: the pad-trim and pad-restore steps are deleted.**

The old `forward` removed the tile-padding rows before the statistics (`logical_h_full` / `needs_trim`). It restored the zero rows after the norm. The new path does not do this.

The fused op only checks `H*W % 32 == 0`. Tile-padded shards satisfy this check. The Welford count then includes the zero rows. The mean and the variance become wrong. There is no error. After the norm, the pad rows hold `gn(0)*gamma+beta` (or `silu` of that). The next conv reads these rows.

Recommendation: keep the trim and restore steps, or reject padded shards with an assert.

models/tt_dit/models/vae/vae.py
```diff
@@ -694,47 +714,58 @@ def _prepare_torch_state(self, state: dict[str, torch.Tensor]) -> None:
...
+            assert self._ctx.ccl_manager is not None
+            if x.layout != ttnn.TILE_LAYOUT:
+                x = ttnn.to_layout(x, ttnn.TILE_LAYOUT)
+            x = self._ctx.ccl_manager.all_gather(x, dim=2, mesh_axis=self._ctx.w_mesh_axis, use_hyperparams=True)
 
         x = self.inner.forward(x, compute_kernel_config=self._compute_kernel_config)
```
ayerofieiev-tt, line 724: **Bug: the 1-D SP path lost the TILE-layout conversion.**

The old path called `_all_gather_hw`. That helper always converted the input to TILE layout. The new path sends `x` directly to the fused op when only one spatial dim is sharded. The op requires TILE layout (`TT_FATAL`: "Input layout must be TILE").

A ROW_MAJOR input crashes on a path that worked before this PR. Only the `_gather_w` branch has the `to_layout` guard.

Recommendation: move the layout guard out of the `_gather_w` branch, so it applies to all inputs.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
```diff
@@ -0,0 +1,278 @@
...
+    const std::optional<Tensor>& weight,
+    const std::optional<Tensor>& bias,
+    const std::optional<MemoryConfig>& memory_config,
+    std::optional<DeviceComputeKernelConfig> compute_kernel_config,
+    const std::optional<Tensor>& persistent_output_buffer,
+    std::optional<tt::tt_metal::SubDeviceId> subdevice_id,
+    const std::optional<ttnn::operations::unary::UnaryWithParam>& fused_activation) {
+    using OperationType = ttnn::experimental::prim::DitFusedDistributedGroupnormDeviceOperation;
+
+    auto arch = is_device_tensor(input_tensor) ? input_tensor.device()->arch() : ttnn::GetDefaultDevice()->arch();
```
ayerofieiev-tt: **Bug: possible null-pointer crash for a host tensor.**

This line calls `ttnn::GetDefaultDevice()->arch()` when the input is not a device tensor. `GetDefaultDevice()` can return `nullptr` (it is set only by `SetDefaultDevice`). The process then crashes with a segfault. The validate error "Input must be on device" never fires.

The `mesh_device` parameter is in scope.

Recommendation: use `mesh_device.arch()`.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+void DitFusedDistributedGroupnormDeviceOperation::validate_on_program_cache_miss(
+    const operation_attributes_t& args, const tensor_args_t& tensor_args) {
+    const auto& input = tensor_args.input;
+    const auto& gamma = tensor_args.gamma;
+    const auto& beta = tensor_args.beta;
+
+    TT_FATAL(input.storage_type() == StorageType::DEVICE, "Input must be on device");
+    TT_FATAL(input.buffer() != nullptr, "Input must be allocated");
+    TT_FATAL(input.layout() == Layout::TILE, "Input layout must be TILE, got {}", input.layout());
```
ayerofieiev-tt, line 30: **Bug: validate does not reject sharded memory layouts.**

`validate_on_program_cache_miss` has no memory-layout check on the input or on `output_mem_config`. Stock groupnorm validates sharded inputs and routes them to a sharded factory. This op accepts a sharded tensor and runs the interleaved mcast factory. The block and CB geometry ignores the shard spec.

The result is garbage output or a watcher error, not a clean message.

Recommendation: add a `TT_FATAL` that requires INTERLEAVED input and output.

models/tt_dit/layers/normalization.py
```diff
@@ -569,7 +578,222 @@ def forward(self, x: ttnn.Tensor, num_out_blocks=-1, compute_kernel_config=None)
...
+        semaphores = self.ccl_manager.get_ag_ping_pong_semaphore(self.cluster_axis) if self.cluster_size > 1 else []
+        persistent = None
+        if self.cluster_size > 1:
+            # The fused activation does not change the stats geometry (it is applied after the
+            # AG, on the output tile), so it is deliberately NOT part of the cache key.
+            persistent = self.ccl_manager.get_fused_norm_stats_buffer(
+                ("gn", tuple(x4.shape), self.num_local_groups, self.cluster_axis),
+                lambda: ttnn.experimental.dit_fused_distributed_groupnorm_create_stats_buffer(
+                    x4,
+                    self.num_local_groups,
```
ayerofieiev-tt: **Latent bug: two different group counts.**

`forward` creates the stats buffer with `num_local_groups`. It calls the op with `num_groups=self.num_padded_groups`. The two values are equal today only because of the `%32` assert. The padding math in `__init__` (`pad`, `ceil`, `num_padded_groups`) is dead code today.

If padding becomes real, the buffer geometry and the op group count desync. The best case is a shape `TT_FATAL`. The worst case is silently wrong cross-device stats.

Recommendation: pass the same group count to both calls, or remove the dead padding path.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
```diff
@@ -163,6 +192,26 @@ void kernel_main() {
 
     const auto src_a = TensorAccessor(src0_args, src_addr);
 
+#ifdef GN_DISTRIBUTED_AG
+    // ---- Fabric all-gather setup ----
+    // The host appends the fabric args AFTER the variable-length noc_coord_x / noc_coord_y lists
+    // (each num_mcast_cores long). Those lists start at 12, 17, or 22 depending on which first/last
+    // mcast sub-groups are present, so the fabric block begins at that base + 2*num_mcast_cores.
+    const uint32_t noc_arg_base = (has_mcast_first_group && has_mcast_last_group)   ? 22u
```
ayerofieiev-tt: **Fragile: hand-coded runtime-arg offsets with two owners.**

The kernel computes the fabric arg base from fixed offsets (12/17/22) plus `2*num_mcast_cores`. Two host factories own this positional layout: the stock groupnorm mcast factory and the new fused factory. No assert ties the kernel offsets to either factory. The host side has the same risk: `override_runtime_arguments` applies one captured index to every sender core.

One inserted or reordered runtime arg silently shifts the fabric block. The kernel then reads a wrong stats-DRAM address or forwarder coordinate. The result is a hang or corruption, on multi-device only.

Recommendation: pass the fabric base as a compile-time arg, or put the fabric args at a fixed front position.

models/tt_dit/layers/normalization.py
```diff
@@ -569,7 +578,222 @@ def forward(self, x: ttnn.Tensor, num_out_blocks=-1, compute_kernel_config=None)
...
+        expected_shape = (self.num_local_channels * self.channel_devices,)
+        assert param.shape == expected_shape, f"expected shape {expected_shape}, got {param.shape}"
+        padding = self.num_padded_channels - self.num_local_channels
+        params = [torch.nn.functional.pad(t, (0, padding)) for t in param.chunk(self.channel_devices)]
+        torch_sharded_lst = [
+            ttnn.create_group_norm_weight_bias_rm(t, self.num_padded_channels, self.num_virtual_cols) for t in params
+        ]
+        return torch.cat(torch_sharded_lst, dim=0)
+
+    def forward(self, x: ttnn.Tensor, num_out_blocks=-1, compute_kernel_config=None) -> ttnn.Tensor:
```
ayerofieiev-tt, line 755: **API bug: `num_out_blocks` and `core_grid` are silently ignored.**

`forward` accepts `num_out_blocks`. The constructor accepts `core_grid`. The fused path forwards neither. The C++ factory uses its own heuristic and claims the full device grid plus one forwarder core.

A caller that tunes L1 usage, or reserves cores for another workload, gets no effect and no warning.

Recommendation: forward the parameters, or assert that they are unset on the fused path.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.cpp
```diff
@@ -0,0 +1,1000 @@
...
+            all_reduction_cores,
+            tt::CBIndex::c_6,
+            per_core_Nt * gamma_beta_single_tile_size,
+            gamma_beta_cb_data_format,
+            gamma_beta_single_tile_size);
+    }
+    create_cb(
+        all_reduction_cores,
+        tt::CBIndex::c_28,
+        input_mask->physical_volume() * input_mask->element_size(),
```
ayerofieiev-tt: **L1 waste: the mask CB holds the full mask on every core.**

The c_28 CB size is `input_mask->physical_volume() * element_size` — the full mask across all virtual columns. Each core uses only `num_groups_per_core * block_wt` tiles. The stock welford factory sizes this CB per core (groupnorm_mcast_program_factory.cpp:328).

The over-allocation is about `num_virtual_cols` times (typically 8x) on every reduction core. Large configs can fail L1 allocation here while stock `ttnn.group_norm` fits.

Recommendation: size the CB as `input_mask_num_tiles_per_core * in_mask_single_tile_size`.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/dit_fused_distributed_groupnorm.cpp
```diff
@@ -0,0 +1,139 @@
...
+        stats_shape,
+        tt::tt_metal::TensorLayout(DataType::FLOAT32, tt::tt_metal::PageConfig(Layout::ROW_MAJOR), stats_mem));
+    return ttnn::create_device_tensor(spec, &const_cast<MeshDevice&>(mesh_device));
+}
+
+}  // namespace ttnn::experimental
+
+namespace ttnn::experimental::prim {
+
+uint32_t gn_num_virtual_cols(uint32_t channels, uint32_t num_groups, uint32_t grid_x) {
```
ayerofieiev-tt: **Reuse: four helpers are re-implemented in this PR.**

`gn_num_virtual_cols` is a copy of `ttnn::operations::normalization::compute_num_virtual_cols` (groupnorm_grid_utils.hpp:37). The Python layer packs gamma/beta/mask against the same column rule. If the shared rule changes, this private copy drifts. The host-prepared mask geometry then disagrees with the device program, silently.

Three more copies exist in this PR: the ring-distance block (`ttnn::ccl::get_forward_backward_line_mcast_distance`, factory ~line 233), the `num_out_blocks` heuristic (`ttnn::prim::groupnorm_heuristic_num_out_blocks`, factory ~line 153), and the local `create_cb` lambda (`tt::tt_metal::create_cb` in the already-included cb_utils.hpp).

Recommendation: call the existing helpers.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
```diff
@@ -365,13 +443,140 @@ void kernel_main() {
...
             }
+#endif
 
             local_means_ptr += local_stride_per_group;
             local_vars_ptr += local_stride_per_group;
             global_means_ptr += 2 * single_tile_size_bytes;
             global_vars_ptr += 2 * single_tile_size_bytes;
         }
 
+#ifdef GN_DISTRIBUTED_AG
```
ayerofieiev-tt, line 446: **Altitude: the isolation claim does not hold.**

The PR keeps the op separate "to reduce the chance of breaking" stock groupnorm. But it adds ~180 lines behind `GN_DISTRIBUTED_AG` to the shared reader kernels, with a different handshake sequence per build. Only the experimental op compiles this branch. Stock groupnorm CI never builds it. The op also creates another op's private kernel (`dit_fused_distributed_rmsnorm/.../dit_rmsnorm_fused_forwarder.cpp`) with only a comment as the contract.

A groupnorm kernel refactor can silently break the fused op. A change to the rmsnorm forwarder can do the same.

Recommendation: extract the shared reduction into an included helper and give the fused op its own thin readers, or promote the forwarder kernel to a shared CCL location with both ops as named consumers.

pavlepopovic (approved):

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
bbradelTT: GN is already on the device 2.0 API.
E.g. if you look at ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp you'll see code like
```
        noc.async_read_barrier();
        dfb_repack_out.pop_front(per_core_N);
```
You are mixing both APIs by re-introducing device 1.0 API calls.

Please make sure everything is using device 2.0 APIs in this PR.

bbradelTT (changes requested):
Needs to not re-introduce device 1.0 API calls.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+    ttnn::Shape output_shape({logical[0], logical[1], logical[2], logical[3]});
+    specs.emplace_back(output_shape, TensorLayout(input.dtype(), PageConfig(Layout::TILE), args.output_mem_config));
+
+    const auto sizing = compute_sizing(args, input);
+    if (!sizing.is_local) {
+        const uint32_t floats_per_page = sizing.page_size_bytes / sizeof(float);
+        ttnn::Shape stats_shape({1u, 1u, sizing.total_pages, floats_per_page});
+        MemoryConfig stats_mem{tt::tt_metal::TensorMemoryLayout::INTERLEAVED, tt::tt_metal::BufferType::DRAM};
+        specs.emplace_back(stats_shape, TensorLayout(DataType::FLOAT32, PageConfig(Layout::ROW_MAJOR), stats_mem));
```
scardozaTT: the stats spec creation is duplicated in 3 places: here + `dit_fused_distributed_groupnorm.cpp:86` + checked dim-by-dim at line 150. Perhaps create a shared helper? (similar to `make_stats_tensor_spec(sizing)` in `dit_fused_distributed_rmsnorm_device_operation_types.hpp:158`)

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.hpp
```diff
@@ -0,0 +1,60 @@
...
+    std::vector<tt::tt_metal::KernelHandle> writer_kernel_ids;
+    std::vector<tt::tt_metal::KernelHandle> compute_kernel_ids;
+    std::vector<tt::tt_metal::KernelHandle> forwarder_kernel_ids;
+    std::vector<tt::tt_metal::CoreCoord> forwarder_cores;
+    std::vector<tt::tt_metal::CoreCoord> cores;           // all reduction cores (writer/compute)
+    std::vector<tt::tt_metal::CoreCoord> sender_cores;    // mcast-group masters (sender reader)
+    std::vector<tt::tt_metal::CoreCoord> receiver_cores;  // non-master cores (receiver reader)
+    // Index of the stats DRAM scratch address inside the master reader's runtime-args vector.
+    // Empty on the is_local (ring_size==1) path.
+    std::optional<size_t> stats_dram_addr_writer_arg_idx;
```
scardozaTT: Name says writer, but this is assigned from `stats_dram_addr_reader_arg_idx` and consumed against `reader_kernel_ids[0]`. Rename to `stats_dram_addr_reader_arg_idx`?

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.cpp
```diff
@@ -0,0 +1,1000 @@
...
+    // Kernels are the stock welford GroupNorm kernels plus the rmsnorm CCL forwarder. The only
+    // fused-specific behaviour lives behind GN_DISTRIBUTED_AG in the two reader kernels; the
+    // compute/writer kernels are used verbatim. At ring_size == 1 the define is not set, so the
+    // fused op compiles the exact stock kernel text.
+    const std::string gn_kernel_base = "ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/";
+    const std::string ccl_kernel_base =
+        "ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_rmsnorm/device/kernels/";
+
+    // Stats CBs are bf16 (cb_data_format above); the reader's cross-device stick layout requires it.
+    const bool stats_is_fp32 = cb_data_format == tt::DataFormat::Float32;
```
scardozaTT, line 455: This isn't correctly querying the stats CB dtype.
Separately, the payload is bf16 pairs (`num_groups_per_core * 4`) while these CBs and the DRAM tensor both declare fp32. Is this intentional?

scardozaTT (approved):
Approving to unblock

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
sadesoyeTT: Thanks for pointing it out. I will certainly get it updated before merge.

mbezuljTT (approved):
for sources.cmake

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/kernels/dataflow/dit_gn_welford_receiver.cpp
```diff
@@ -0,0 +1,252 @@
...
+
+            local_means_ptr += local_stride_per_group;
+            local_vars_ptr += local_stride_per_group;
+            global_means_ptr += 2 * single_tile_size_bytes;
+            global_vars_ptr += 2 * single_tile_size_bytes;
+        }
+
+        // Batched handshake: signal the master ONCE that all groups' partials are ready, then wait
+        // ONCE for its single batched mcast-back of the GLOBAL (mean, var) for every group.
+        reduce_receiver_sem.up(noc, mcast_sender_noc_x, mcast_sender_noc_y, 1);
```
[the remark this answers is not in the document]
sadesoyeTT: Stale — this file no longer exists. The first revision copied the groupnorm kernels wholesale; the current implementation reuses the stock welford readers instead, and `dit_gn_welford_receiver.cpp` was deleted in that rewrite.

For what replaced it: the sender's fabric block addresses the forwarder through a `UnicastEndpoint`, and the receiver signals with the Device 2.0 `Semaphore::up(noc, x, y, 1)` — both handle coordinate mapping internally.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
sadesoyeTT: Done — the `GN_DISTRIBUTED_AG` block is now fully Device 2.0: `DataflowBuffer` for the CB ops, `noc.async_write` / `async_read` with `UnicastEndpoint` and `TensorAccessor`, `Semaphore::up` / `wait_min` / `set`, and `noc.async_read_barrier()` / `async_write_barrier()` / `async_atomic_barrier()`. `safe_get_noc_addr` is gone, so the `noc_addr.h` include went with it and the file's includes are back to the stock set.

One thing I'd like your read on. `fwd_packet_buf_addr` is the packet CB's address on the *forwarder* core, used as a remote NoC destination. `DataflowBuffer::get_write_ptr()` adds `L1_UNCACHED_OFFSET`, and the NoC-safe `get_noc_write_addr()` is private — its traits `static_assert` on `AddressType::LOCAL_L1`, so a DFB can't serve as a remote unicast destination either. I've used `get_write_ptr()`, which is what the stock code two lines up already does with `global_base_ptr` for the peer read and the mcast-back. If there's an intended 2.0 idiom for a remote CB address that I've missed, happy to switch.

models/tt_dit/models/vae/vae.py
sadesoyeTT: This case is unreachable

`mesh_partition` hard-requires exact divisibility and never pads

models/tt_dit/models/vae/vae.py
sadesoyeTT: Fixed, though placed differently than suggested.

The `to_layout` in `_all_gather_hw` was the *gather kernel's* guard — its docstring says "Returns TILE_LAYOUT (the gather kernel requires it)" — and the new `_gather_w` branch keeps it correctly positioned for its own `all_gather`. Moving that one out would relocate a guard away from the thing that needs it.

Instead I hoisted a single guard to the top of `_VaeSpatialGroupNorm.forward`, so it covers both consumers — the gather on the 2-D path and the fused op on both paths — and removed the copy inside the branch.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
sadesoyeTT: Fixed — `mesh_device.arch()`, which is in scope and `const`. A host tensor now reaches `validate_on_program_cache_miss` and gets the "Input must be on device" `TT_FATAL` instead of segfaulting.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
sadesoyeTT: Added `TT_FATAL`s requiring INTERLEAVED on both the input and `output_mem_config`.

The mechanism is a little different from the description, worth noting for the record: addressing would actually have been correct, since every input/output/gamma/beta/mask access goes through `TensorAccessor` and `TensorAccessorArgs` picks the `Sharded` config straight off the buffer. The reasons to reject it are elsewhere — stock gates its own welford path to non-sharded (`use_welford && !a.is_sharded()`), the reduction grid is derived from the logical shape and ignores the shard grid entirely, and an L1-sharded input plus this factory's CBs can overflow L1. So it's an unsupported-configuration guard rather than a garbage-output fix.

models/tt_dit/layers/normalization.py
sadesoyeTT: Fixed — the stats buffer now uses `num_padded_groups`, matching the `num_groups` passed to the op, and the cache key uses the same value.

I kept the padding path rather than deleting it. The reused welford kernels handle padded channels natively (that's what the input mask is for) and the factory already reads `padded_shape()` throughout, so it's a real capability gated only by two validate lines that read `logical_shape()`. Deleting the arithmetic would have made it look unsupported when it's closer than that.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/dataflow/welford_reader_mcast_sender_unary_gn.cpp
sadesoyeTT: Fixed — and it was worse than fragile on the host side.

The index was captured from the *first* sender core only (`if (!stats_dram_addr_reader_arg_idx.has_value())`) and then applied to every sender core in `override_runtime_arguments`. Since `is_rectangle_grid` depends on where each group's run of cores falls relative to grid row wraps, `has_mcast_first_group` / `has_mcast_last_group` genuinely differ between groups — so on any config with non-uniform groups that wrote the stats DRAM address over an unrelated runtime arg on a cache hit.

Now:
- The kernel's if/else stores the base it was already computing into `noc_arg_base`, and the fabric block reuses it. The 12/17/22 literals exist in one place instead of two, and the file got 6 lines shorter.
- The host records the index per core in a `std::vector<size_t>` parallel to `sender_cores`, and the override indexes by core.
- Two `TT_FATAL`s tie the layouts: the host's computed fabric base must equal `rt.size()` where the block is pushed, and the emitted group size must equal the compiled `num_cores_per_mcast_group`.

models/tt_dit/layers/normalization.py
sadesoyeTT: Asserted unset on the fused path rather than forwarded — the op has no parameter for either (the factory derives its own block split and claims `compute_with_storage_grid_size()`), so honoring them would need new plumbing through the op. Both remain honored on the local `cluster_axis is None` path, which forwards them to the inner `GroupNorm`. Also dropped `core_grid` from the two tests that were passing it into the fused path.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.cpp
sadesoyeTT: Fixed — sized `input_mask_num_tiles_per_core * in_mask_single_tile_size`, matching stock's welford branch at `groupnorm_mcast_program_factory.cpp:324-329`. Also hoisted that value so the CB size and the per-column `input_mask_tile_start_id` wrap derive from one expression.

The consumer never holds more than that: the writer reserves exactly `block_w * num_groups_per_core` and reads its slice starting from the per-core `input_mask_tile_start_id`.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/dit_fused_distributed_groupnorm.cpp
sadesoyeTT: All four now call the shared versions:

- `compute_num_virtual_cols` — the private copy is deleted. Arg order differs (`grid_x, num_groups, num_channels`), so both call sites were rewritten rather than renamed. This was the one that mattered most, since the Python layer packs gamma/beta/mask against the same rule.
- `ttnn::prim::groupnorm_heuristic_num_out_blocks`
- `tt::tt_metal::create_cb` from the already-included `cb_utils.hpp` — 19 call sites converted
- `get_forward_backward_line_mcast_distance(..., static_alternate=true)` — this one was line-for-line identical to the inline block

One deliberate behavior change: the old `num_out_blocks` loop had a third condition `num_out_blocks < block_ht` that didn't actually clamp — with `heuristic=8, block_ht=3` it produced 4 and then tripped its own `TT_FATAL` two lines later. Replaced with `std::min(..., block_ht)`, which only changes outcomes in cases that previously aborted.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_device_operation.cpp
sadesoyeTT: Added `make_stats_tensor_spec(sizing)` to the types header, mirroring `make_stats_tensor_spec` in the rmsnorm op. All three sites derive from it now, and validate's dim-by-dim block collapses to a single `buf.tensor_spec() == expected_spec`.

That does make validate slightly stricter — full `TensorSpec` equality also covers page config and alignment, not just the five fields being checked individually — which seems right, since the expected spec is exactly what `create_stats_buffer` produces. The device/allocation checks stay separate, as those aren't part of a `TensorSpec`.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.hpp
sadesoyeTT: Renamed to `stats_dram_addr_reader_arg_idx`. It also changed type to `std::vector<size_t>` — see the runtime-arg-offsets thread; the index genuinely varies per sender core, and applying one captured value to all of them was a live bug.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.cpp
sadesoyeTT: Both parts are intentional, but the code didn't say so — so I've fixed the comments rather than the logic.

**On the query:** `stats_is_fp32` refers to the *welford* stats CBs (ex_partial / ex_global / ex2pe), which are created with `cb_data_format`. The fp32 ones you spotted are the separate fabric staging CBs. The expression is verbatim stock — `groupnorm_mcast_program_factory.cpp:82` and `groupnorm_sharded_program_factory.cpp:70` both say `cb_data_format == tt::DataFormat::Float32` — so I kept it. It's tautological here only because the fused op pins `cb_data_format`, and it tracks correctly if that ever becomes derived the way stock's `im_data_format` is. Added a comment naming which CB group it refers to.

**On bf16 payload in an fp32 container:** intentional. fp32 is a byte container — it gives 4-byte element granularity so `page_size_bytes / sizeof(float)` works and every sub-stick offset stays 64 B-aligned, which the Blackhole DRAM read path requires. Commented at the staging-CB declaration.

Also documented why `cb_data_format` is pinned to bf16 in the first place, since that's the constraint that makes the whole thing rigid: the cross-device gather packs each group's `[mean, var]` as a bf16 pair (`stick_bytes = num_groups_per_core * 4`), so fp32 stats would need a different stick layout and stride. That's what the kernel's `static_assert(stats_is_fp32 == 0)` is guarding.

bbradelTT (approved):

PR #52505, opened 2026-08-07
[Feature] Freeze a Metal 1.0 slice tile factory for the Python fusion framework

dgomezTT:
### Problem

`models/experimental/ops/descriptors/` (the experimental fusion framework) is the only Python consumer of the descriptor pybinds. It calls `<Factory>.create_descriptor(...)` to get a `ProgramDescriptor` and runs it through `generic_op`.

`generic_op` accepts only `ProgramDescriptor`/`MeshProgramDescriptor`. A Metal 2.0 factory produces `ProgramArtifacts` via `create_program_artifacts`, and no `ProgramArtifacts -> ProgramDescriptor` lowering exists. So porting a factory silently removes it from the fusion framework.

This already happened: `experimental/quasar/slice` is fully MetalV2 and its `bind_slice_descriptor` no longer binds any factory; the three ported `quasar/matmul` factories carry comments saying their `create_descriptor` entry point was dropped.

### Change

Adds `ttnn::for_python::SliceTileProgramFactory` — a Metal 1.0 copy of `ttnn::prim::SliceTileProgramFactory::create_descriptor`, frozen. It is not in `SliceDeviceOperation::program_factory_t` and is not reachable from any dispatch path; only nanobind refers to it.

`ttnn::prim::SliceTileProgramFactory` is now free to be ported to Metal 2.0 without breaking Python. The new type lives outside `device/` and carries a "do not port" comment.

Python-visible name becomes `SliceTileProgramFactoryForPython` so a caller can tell they are on the frozen path.

### Scope notes

- `SliceParams` / `SliceInputs` / `create_output_tensors` / `compute_output_specs` stay bound to production — they survive the port.
- The copy still calls the pure shape helper `get_tiled_start_offset`. That does not encode the execution model, so forking it would buy no migration safety and only create drift.

### Testing

Built with `build_metal.sh`. 258 fusion tests pass locally on hardware (`tests/ttnn/unit_tests/operations/fused/parallel_sequential/`), including the slice chains.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/for-python-slice)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/for-python-slice)

### DIFF
--- a/ttnn/cpp/ttnn/operations/data_movement/slice/slice_program_factory_tile_for_python.hpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/slice/slice_program_factory_tile_for_python.hpp
@@ -1,0 +1,18 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <tt-metalium/program_descriptors.hpp>
+#include "ttnn/operations/data_movement/slice/device/slice_device_operation_types.hpp"
+
+namespace ttnn::for_python {
+
+// Metal 1.0 copy of ttnn::prim::SliceTileProgramFactory, frozen for the Python fusion framework.
+// Not part of SliceDeviceOperation and not on any dispatch path — do not port it to Metal 2.0.
+struct SliceTileProgramFactory {
+    static tt::tt_metal::ProgramDescriptor create_descriptor(
+        const ttnn::prim::SliceParams& args, const ttnn::prim::SliceInputs& tensor_args, Tensor& output);
+};
+
+}  // namespace ttnn::for_python
--- a/ttnn/ttnn/__init__.py
+++ b/ttnn/ttnn/__init__.py
@@ -608,7 +608,7 @@ from ttnn._ttnn.operations.data_movement import (
     SliceParams,
     SliceInputs,
     SliceDeviceOperation,
-    SliceTileProgramFactory,
+    SliceTileProgramFactoryForPython,
 )
 
 import pathlib
--- a/models/experimental/ops/descriptors/data_movement/slice.py
+++ b/models/experimental/ops/descriptors/data_movement/slice.py
@@ -51,7 +51,7 @@ def slice(
 
     output_tensor = ttnn.SliceDeviceOperation.create_output_tensors(params, tensor_args)
 
-    descriptor = ttnn.SliceTileProgramFactory.create_descriptor(params, tensor_args, output_tensor)
+    descriptor = ttnn.SliceTileProgramFactoryForPython.create_descriptor(params, tensor_args, output_tensor)
 
     return OpDescriptor(
         descriptor=descriptor,
--- a/ttnn/cpp/ttnn/operations/data_movement/CMakeLists.txt
+++ b/ttnn/cpp/ttnn/operations/data_movement/CMakeLists.txt
@@ -244,6 +244,7 @@ target_sources(
             sharded_partial/sharded_to_interleaved_partial/device/sharded_to_interleaved_partial_program_factory.hpp
             # slice/
             slice/slice_nanobind.hpp
+            slice/slice_program_factory_tile_for_python.hpp
             slice/device/slice_device_operation.hpp
             slice/device/slice_device_operation_types.hpp
             slice/device/slice_program_factory_rm.hpp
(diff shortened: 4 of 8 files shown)


ttnn/cpp/ttnn/operations/data_movement/slice/slice_program_factory_tile_for_python.hpp
```diff
@@ -0,0 +1,18 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+#pragma once
+
+#include <tt-metalium/program_descriptors.hpp>
+#include "ttnn/operations/data_movement/slice/device/slice_device_operation_types.hpp"
+
+namespace ttnn::for_python {
```
vtsilytskyiTT, line 9: I don't like this naming. for_python doesn't reflect purpose. And, as I understand, it is temporary, until migration complete. Why not `ttnn::deprecated` then? Or something similar?

vtsilytskyiTT (approved):
Approve to unblock. Would prefer another naming.

ttnn/ttnn/__init__.py
```diff
@@ -608,7 +608,7 @@ def get_arch_name():
     SliceParams,
     SliceInputs,
     SliceDeviceOperation,
-    SliceTileProgramFactory,
+    SliceTileProgramFactoryForPython,
```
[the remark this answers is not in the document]
dgomezTT: Intentional, and there are no callers left to break — I checked the whole tree on this branch.

The only remaining references to `SliceTileProgramFactory` are three comment/docstring lines in
`tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_slice.py` (lines 8, 9,
212). No Python code reads the name.

Keeping it as a compatibility alias would defeat the point of the PR. This binding exists solely so the
experimental fusion framework can reach a program factory; freezing a Metal 1.0 copy for that consumer
is what lets the *live* slice factory stop exporting its internals and migrate to Metal 2.0. An alias
pointing back at the live factory would keep exactly the coupling being removed.

ttnn/cpp/ttnn/operations/data_movement/slice/slice_program_factory_tile_for_python.hpp
dgomezTT: Fair point, and you're right that it is temporary.

What these are, precisely: byte-frozen copies of the Metal 1.0 factory, kept alive only so the
experimental Python fusion framework (`models/experimental/ops/descriptors/`) keeps working while the
live factory migrates to Metal 2.0. They get deleted once that framework stops needing a
`ProgramDescriptor`.

I'd push back slightly on `deprecated`: nothing here was ever a supported API being phased out, and the
name would suggest callers should migrate off it, when in fact it is the *only* thing its one caller
should use for now. What the name should convey is "frozen snapshot, do not extend, has an expiry".

So `ttnn::frozen` or `ttnn::fusion` both read better to me than `for_python` — `frozen` says why you
must not touch it, `fusion` says who it is for. I have no strong preference between those two.

The rename spans this PR plus the layernorm (#52506) and matmul (#52507) ones, so I'd rather do it once
across all three than piecemeal. Tell me which name you want and I'll apply it everywhere; leaving this
thread open until then.

dgomezTT:
Closing this — we're taking a different direction.

Rather than freeze a Metal 1.0 copy of the factory so the Python binding keeps working, the binding
itself gets deleted. The op-porting recipe already removes it as part of the port, so freezing a copy
just adds a second surface that has to be kept in sync with production for a binding that is going
away regardless.

This PR's own review made the cost concrete: the frozen sharded copies had drifted from their source
(a pre-#52277 snapshot with fragments grafted back by hand), which dropped the idle-core kernels for
non-rectangular shard grids while keeping that change's validation. And the freeze only ever covered
host code — the device kernels stayed shared and moving, so the copies could fall out of sync again
at any time with nothing in the build to catch it.

Not merging any of the three freeze PRs (#52505 slice, #52506 layernorm, #52507 matmul).

PR #52513, opened 2026-08-07
[Feature]  Fused CCL + matmul: fabric-bound minimal-matmul, strided AGMM and MMRS

nwoodallTT:
# [Feature] Fused CCL + matmul: fabric-bound minimal-matmul, strided AGMM and MMRS

### PR Category

Feature

---

> **Stacked PR 1 of 3.** Base: `main`. This PR is self-contained — it adds new
> `ttnn.experimental.*` ops and their unit coverage, with no model-side changes.
> PR 2 (neighbor-pad halo + conv3d) and PR 3 (ring SDPA + LTX pipeline wiring)
> stack on top of it.

## Summary

New CCL and matmul machinery that lets a collective overlap with the math that
consumes it, rather than completing into DRAM and being read back.
`60 files changed, 6129 insertions(+), 505 deletions(-)` across 14 commits.

**No changes to existing public APIs.** New ops land under
`ttnn.experimental.*`; existing signatures gain optional parameters only
(`fused_ternary_input_a/b`, `mm_progress_counters`, `num_ag_workers`), each
defaulting to prior behavior.

Each subsection ends with the commits it covers, so a reviewer can read the
rationale and then go straight to the diff. Subjects are given alongside the
short SHAs because the SHAs move on every rebase.

### Fused CCL + matmul: overlapping the collective with the math

The motivating pipeline's cost is dominated by TP matmuls whose operands must be
gathered or whose results must be scattered. Previously the collective completed
into DRAM and the matmul then read it back — the collective and the math never
overlapped, and the round-trip itself cost DRAM/NoC bandwidth.

Three pieces remove that serialization:

**Per-worker, per-core signaling.** `AllGatherFusedOpSignaler` generalizes from
a fixed 3-semaphore layout to `2N+1` —
`[backward_0..N-1, forward_0..N-1, self]` where `N = num_ag_workers`. Each
all-gather worker increments its own semaphore, so a consumer can wait on all N
portions of a k-block rather than a single aggregated per-direction signal.
`N=1` reproduces the legacy layout byte-for-byte, so existing callers are
untouched. A second mechanism, `signal_op_per_core`, drops the worker barrier
entirely: each worker bumps its own slot in a per-core counter array on every
receiver core and continues, so per-block skew flows into the consumer's
pipeline instead of stalling the whole grid. Producer and consumer must derive
the same row-major core index for this to be correct.

**Banded all-gather delivery.** The strided all-gather now splits each chunk
into M-row bands, letting the matmul start on a partial k-block. An
interleaved-middle schedule runs two independent direction cursors so the middle
forward/backward region is consumed one-backward-one-forward rather than drained
as whole-device groups, with each cursor preserving its within-direction order so
the monotonic per-direction semaphores stay valid. Interleaving is disabled when
any k-block straddles a device boundary (`expected == 2`), because two devices
co-complete such a block on different direction passes and interleaving would
reorder those pair-completions.

**A matmul-signal aggregator.** One aggregator core per all-gather direction on
the receiving device. Each of the N writer workers fabric-increments its own
semaphore there as its portion lands (ordered after the data on that worker's
connection); the aggregator waits for all N, then increments the matmul's single
direction semaphore once. The matmul is decoupled from the AG reader's forwarding
pace while its cores still see only 3 semaphores. The aggregator's iteration
cadence deliberately mirrors the reader's remote-receive loop so the number and
order of matmul signals match the legacy reader-signaled path exactly.

Transport-level additions: scatter-write packs up to 4 tiles with distinct
destination NoC addresses per fabric packet, capped by real per-packet page
capacity so 4 degrades gracefully; the worker-to-fabric path moves to Mux V2
(dual-RISC forwarder + manager); and on an even ring the diametric slice is
split-forwarded across both links, with direction 1 receiving and relaying one
extra half. Split-forwarding is gated on
`Ring && ring_size % 2 == 0 && ring_size > 2`.

*Commits:* `aa9436aa8b9` ccl: per-core op signaling · `858d76fe5a0` M-row bands
with per-band signaling · `76671d02614` matmul-signal aggregator core ·
`8a116cce56b` split-forward the diametric slice · `e7ae8ba9e5c` SCATTER4, mux v2
sender, aggregator placement

### Fabric-bound minimal-matmul

A matmul variant that consumes `in0` directly from fabric-delivered bands
instead of a completed DRAM tensor — 3 new kernels plus a shared dataflow header
and a 1,179-line program factory. The dataflow primitives that make it work:
block reads that can skip the completion barrier so band *s*'s reads stay in
flight while band *s+1* is awaited (a single barrier after the last band),
CB-base byte offsets so bands land in their own slots rather than the base, and a
row-ownership predicate for an interleaved two-NoC output write that overlaps row
writes instead of running them back-to-back.

Band-interleave processes a forward remote k-block and the following backward one
a band at a time (`fwd.b0, bwd.b0, …`) rather than draining each whole; this
requires two k-blocks resident in the in1 scratch simultaneously, which is what
sizes that CB. The in1 scratch also lets compute run per band without re-reading
DRAM, and is created unconditionally (`nb == 1` when not banding) so the in1
dataflow keeps a single code path.

Hooking this into the default factory also fixed two latent issues: compile-time
arg offsets were advanced by a fixed `NUM_TENSOR_ACCESSOR_ARGS()` multiple even
though accessor arg counts are tensor-config dependent (now
`get_tensor_accessor_args_cta_offset<N, offset>()` /
`next_compile_time_args_offset()`), and a deferred-write branch fired on the
fused RS path where `defer_write` is false.

*Commits:* `31ef7a5130f` fabric-bound kernel variants · `35506e36bae`
fabric-bound program factory · `2d244c65f56` hook the fabric-bound path into the
default factory (also carries the two fixes above)

### sAGMM and MMRS

**`strided_all_gather_minimal_matmul_async`** wires the fabric-bound factory in
and adds the core placement it needs: fabric senders sit beside the mux at the
in0 chain's far (high-coordinate) end, with the sender index on the axis
perpendicular to the chain so senders up to `full_grid_size.y - 1` hops from
their mux funnel their writes together. Forward and backward senders each bind to
one direction's mux. It also gains a fused addcmul epilogue,
`out = ternary_a + scalar * matmul_out * ternary_b`; supplying both ternary
inputs enables `FUSE_TERNARY`, with the scalar riding in `matmul_struct`.

**`minimal_matmul_strided_reduce_scatter_async`** fuses matmul with strided
reduce-scatter and hands the matmul output off in L1: the output is block-sharded
across the matmul core grid so the RS reader's `TensorAccessor` reads it straight
from the matmul cores' L1, removing the ~M*N DRAM write plus read that was
driving DRAM/NoC contention. RS intermediate and output specs keep their DRAM
config.

Its progress signaling is per-core rather than aggregate: a flat L1 array of
`mm_cores_x * mm_cores_y` uint32 counters, at the same L1 address on every RS
worker core. Each MM core increments its own slot per completed output block and
the RS reader waits **per tile** on the slot of the core that produced that tile,
so a fast MM core never stalls behind a slow peer. Counters are monotonic across
batches; the reader zeroes them at kernel start, which is safe because both
kernels launch together while the MM's first signal is a full output block
(~100 us) away.

The op accepts that counter array as an optional caller-allocated tensor
(`mm_progress_counters`, defaulting to `std::nullopt` — per-program allocation,
the prior behavior). Sharing one array across many cached MMRS programs is what
the caller in PR 3 does, and the reason is L1 fragmentation: per-program arrays
are retained for each program's cached life, and because L1 is handed out
top-down each small permanent block pins the freed space above it. After enough
cached MMRS programs a later large-CB op has no contiguous L1 left. The array's
address is baked into reader and MM runtime args at build time and never
re-pointed, so it must stay at that address for the life of every cached program
that uses it; a `TT_FATAL` enforces this.

*Commits:* `e7a11ee8d08` sagmm: fabric-bound all-gather-matmul op · `f63dedc6288`
mmrs: strided reduce-scatter with L1 hand-off · `5876dede27d` mmrs: accept a
caller-allocated shared L1 progress counter

---

## Test results

Blackhole galaxy, 32 chips, 4x8 mesh, 12x10 compute grid, `FABRIC_1D_RING`,
`num_links=2`.

**MMRS unit test**, `ltx_ff2 4864x4096x4096`, b457 blocking, fabric_ring, fused,
axis_1 — 1 passed, PCC 0.99998 across all devices.

New coverage for strided all-gather (including variable-chunk), strided AGMM, and
MMRS across t3000, TG, and Blackhole-galaxy nightly suites.

*Commits:* `d8bf975c5c7` test: coverage for strided AGMM and MMRS ops ·
`35645eaaab4` scripts: make safe-pytest dispatch timeout and reset configurable

---

## Notes for reviewers

**Where to focus.** The fabric-bound kernels and their program factory are ~3,800
lines with no top-level design comment; the band pipeline's architecture is the
hardest thing here to review and would benefit most from a design note.

**Shared MMRS counter lifetime.** The shared L1 progress-counter address is baked
into runtime args and never rebound, so correctness depends on the allocation
outliving every cached program that references it. This is enforced by a
`TT_FATAL` rather than by rebinding on `override_runtime_arguments`; rebinding
would be the more robust design.

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nwoodall/ltx-ccl-matmul-fabric)


jonathansuTT (approved):
best PR ever

sjameelTT (approved):

Issue #42146, opened 2026-04-15
Conv3d auto-blocking overflows L1 for large kernels (Qwen 2.5 VL)


ctr-lelanchelianTT:
## Description

`ttnn::experimental::conv3d` overflows L1 when called **without an explicit `Conv3dConfig`** (auto-blocking path) for large kernels such as Qwen 2.5 VL's patch embedding convolution.

### Parameters that trigger the overflow

```python
nn.Conv3d(in_channels=3, out_channels=1280, kernel_size=(2, 14, 14), stride=(2, 14, 14), bias=False)
# input shape: (1024, 3, 2, 14, 14)
```

### Error

```
TT_THROW: Statically allocated circular buffers on core range
[(x=0,y=0) - (x=7,y=7)] grow to 1738528 B which is beyond max L1 size of 1499136 B
```

Source: `tt_metal/impl/program/program.cpp`

## Root Cause

The auto-blocking logic in `conv3d.cpp` uses a **fixed** `C_in_block = TILE_WIDTH (32)` regardless of kernel size. For large spatial kernels like `(2, 14, 14)` with padded `C_in = 32`, this produces:

- `patch_size = 2 × 14 × 14 × 32 = 12,544`
- `matmul_K_t = ceil(12544 / 32) = 392` tiles

The `vol2col_tiled` and `weight_tiled` circular buffers each need `392 × tile_size` bytes, totaling ~1.6 MB — exceeding the 1.5 MB L1 limit.

### Working workaround (requires explicit config)

Using `C_in_block=16, C_out_block=32` fits in L1 (as shown in `test_conv3d_qwen_shapes`), but callers that rely on auto-blocking (e.g., tt-xla) hit the overflow.

## Related

- tt-xla issue: https://github.com/tenstorrent/tt-xla/issues/3983

PR #52514, opened 2026-08-07
[Feature] Neighbor-pad halo exchange and halo-mode conv3d

nwoodallTT:
# [Feature] Neighbor-pad halo exchange and halo-mode conv3d

### PR Category

Feature

---

> **Stacked PR 2 of 3.** Base: PR 1 (`nwoodall/ltx-ccl-matmul-fabric`). The
> changes here are file-disjoint from PR 1 — the stacking is only so PR 3 can
> have a single base containing both. This PR adds new `ttnn.experimental.*` ops
> plus halo support on the existing experimental conv3d, with no model-side
> changes.

## Summary

Spatially sharded conv3d needs its neighbors' rows and columns. This PR adds the
fabric halo exchange that supplies them and teaches conv3d to read from it
directly. `40 files changed, 5720 insertions(+), 63 deletions(-)` across 4
commits, including 6 new device kernels.

**No changes to existing public APIs.** The new ops land under
`ttnn.experimental.*`; conv3d gains optional parameters only (`halo_buffer`,
`logical_h_mask`, `logical_w_mask`, `pad_offset_tensor`, `output_pad_h`,
`output_pad_w`), each defaulting to prior behavior.

### neighbor_pad_halo

Exchanges halo rows/columns between neighbors and writes **only** the compact
halo buffer `[H-top | H-bot | W-left | W-right]` — no interior copy, so the cost
is the transport rather than a full padded-activation rewrite.

H and W exchanges are separated so the W pass is H-independent: H rows first
(`nf * input_H_dev`), then corner rows (`nf * 2 * padding_h`). The halo-buffer
addressing helper is shared between the W reader and the W writer specifically so
the two cannot drift — an addressing mismatch between them is silent corruption,
not a crash.

Zero-padding covers both cases the callers need: in-place border-only fill on an
already-padded buffer, and a repack that grows H and W.

### halo_scatter

Writes the exchanged halo into a persistent padded activation. This is the path
for a conv3d that wants a fully materialized padded input — it gets one without
re-copying the interior on every step.

### Standalone conv3d: halo-mode reads

conv3d gains the ability to read its boundary taps straight from the compact halo
buffer instead of from a materialized padded tensor, which is what makes the
compact exchange above useful end to end. Also added:

- a padded-output mode (allocate a spatially padded output; the writer fills only
  the interior)
- opt-in logical-pad masking on the halo path
- conservative default blocking under two stated constraints:
  `kernel_vol * C_in_block ≡ 0 (mod TILE_WIDTH)` for weight tile alignment, and
  `C_in_block % l1_alignment == 0`

*Commits:* `f200e705a41` neighbor-pad: add halo exchange op for spatially-sharded
conv3d (adds the op and registers it) · `30f9ccc9507` conv3d: halo-mode reads and
per-shape blocking

---

## Test results

Blackhole galaxy, 32 chips, 4x8 mesh, `FABRIC_1D_RING`, `num_links=2`.

New correctness coverage for the halo exchange across mesh geometries, padding
widths, and both H and W axes.

*Commits:* `790b273d9ff` test: coverage for the neighbor-pad halo op

---

## Notes for reviewers

**Blackhole-specific DRAM bank count.** The halo kernels
(`np_h_reader.cpp:28`, `np_h_mux_writer.cpp:34`, `np_phase2_w_reader.cpp:35`)
hardcode `constexpr uint32_t NP_NUM_DRAM_BANKS = 8` to drive their bank-strided
coalescing loops. That matches Blackhole, which is the only platform this has
been brought up on, but it should become a compile-time arg derived from the
device before the op is used elsewhere. Flagging it explicitly rather than
leaving it to be discovered.

**W-reader / W-writer addressing.** The shared halo-buffer addressing helper is
load-bearing: if the reader and writer ever compute offsets independently they
will disagree silently. Worth a look to confirm the sharing is airtight.

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nwoodall/ltx-neighbor-pad-halo-conv3d)


jonathansuTT (approved):

sjameelTT (approved):

ttnn/cpp/ttnn/operations/experimental/conv3d/device/conv3d_program_factory.cpp
```diff
@@ -294,6 +294,11 @@ tt::tt_metal::ProgramDescriptor Conv3dProgramFactory::create_descriptor(
     log_debug(tt::LogOp, "CB matmul_result_rm: page_size={} bytes, num_pages={}", tile_size, matmul_M_t * matmul_N_t);
 
     bool is_padding_zeros = operation_attributes.padding_mode == "zeros";
+    // Halo-aware mode: spatial (H/W) boundary conv-window positions read from the compact halo buffer instead
+    const bool halo_mode = tensor_args.halo_buffer.has_value() && is_padding_zeros;
```
mstaletovicTT: We should instead TT_FATAL when padding mode is not "zeros"

mstaletovicTT:
conv3d_device_operation.cpp:39 (and the validate function)

Nothing validates the dimensions of the compact halo tensor. Dimensions for them should be derivable from the mesh setup etc. You might want to add some checks

ttnn/cpp/ttnn/operations/experimental/conv3d/conv3d.hpp
```diff
@@ -29,6 +29,12 @@ ttnn::Tensor conv3d(
     const std::string& padding_mode_ = "zeros",
     uint32_t groups_ = 1,
     const std::optional<MemoryConfig>& memory_config = std::nullopt,
-    std::optional<DeviceComputeKernelConfig> compute_kernel_config = std::nullopt);
+    std::optional<DeviceComputeKernelConfig> compute_kernel_config = std::nullopt,
+    const std::optional<ttnn::Tensor>& halo_buffer = std::nullopt,
+    uint32_t logical_h_mask = 0,
+    uint32_t logical_w_mask = 0,
+    const std::optional<ttnn::Tensor>& pad_offset_tensor = std::nullopt,
```
mstaletovicTT, line 36: I am not sure if I fully understand why we need to pass the information about mesh position through a tensor this way. Couldnt we get this kind of information from the MeshCoordinate of the device (like most CCL's do, including the one implemented in this PR)

ttnn/cpp/ttnn/operations/experimental/conv3d/device/kernels/reader_vol2col.cpp
```diff
@@ -393,10 +412,64 @@ void gather_rows_to_shard(
                     }
                 }
                 if constexpr (check_padding) {
+                    // Universal logical-pad mask (halo mode): any position whose GLOBAL coord is beyond the logical
+                    bool logical_masked = false;
+                    if (g_mask_logical_h != 0 &&
```
mstaletovicTT: `g_mask_logical_h` is a runtime variable. It is set to 0 in line 875 (and 13 aswell), and it being changed is guarded by `mask_mode`, which is `constexpr`, so you would think that the compiler can figure out whether this check should happen, or if it can be skipped. However, since `gather_rows_to_shard` is not inlined (and it shouldn't be) the compiler still might not be able to figure out that `g_mask_logical_h` can be known to be zero at compile time.



This, in turn, makes it so that this runtime check happens on all pre-existing paths that took the `gather_rows_to_shard` approach, and adds cycles.

I actually got Codex to compile some old path (`mask_mode` = off) and it did contain the runtime if.

So we should thread constexpr `mask_mode` and if constexpr direcly on it, or these if's will exist on the hot path for pre-existing paths

mstaletovicTT (approved):
Approving to unblock, but please address the final comments about the introduced runtime ifs.

Not so picky on the other comments, especially if you have built a lot on the current "pad_tensor" system, but I would prefer if we did not have to use it and if we could instead just directly get the mesh coordinate

Also please run L2 nightly tests with conv enabled

mstaletovicTT:
▎ The comment-collapse commit cut several of these mid-sentence:
  ▎
  ▎ | Line                                  | Ends at                            |
  ▎ |---------------------------------------|------------------------------------|
  ▎ | conv3d_device_operation_types.hpp:89  | …global spatial index is >=        |
  ▎ | conv3d_device_operation_types.hpp:92  | …into the interior of a spatially  |
  ▎ | conv3d_device_operation_types.hpp:101 | …positions from this compact       |
  ▎ | reader_vol2col.cpp:11                 | …zeros sticks whose GLOBAL spatial |
  ▎ | reader_vol2col.cpp:415                | …coord is beyond the logical       |
  ▎ | writer.cpp:66                         | …interior of a spatially padded    |
  ▎ | conv3d_program_factory.cpp:297        | …halo buffer instead               |
  ▎
  ▎ It also removed a pre-existing, unrelated comment in conv3d_nanobind.cpp explaining C_in_block = 0 semantics and its link to #42146/#47316 — that one's worth restoring regardless.
  ▎
  ▎ Separately, conv3d_device_operation_types.hpp:89 says masking is "halo mode only", but mask_mode (conv3d_program_factory.cpp:300) is computed independently and works standalone. Either the comment is stale or the combination should be enforced.

Some comments got truncated in the process

Issue #51127, opened 2026-07-26
[tt_metal/impl] Ten defects: allocator OOB write, stale dispatch command re-read, MX encoder rounding/mask, profiler control-vector overflow


amahmudTT:
Ten defects in `tt_metal/impl`, each causing silent wrong results, memory corruption, or a hang. Line references are against `02246470f99`; the code quoted in each entry is the reliable anchor if lines have moved. Found by code inspection.

They span seven subsystems with different owners, so they are grouped below by subsystem and can be fixed and closed independently.

| # | Location | Defect |
|---|---|---|
| 1 | `tt_metal/impl/allocator/algorithms/free_list_opt.cpp:572` | FreeListOpt::shrink_size() indexes block_prev_block_ with block_next_block_[block_to_shrink] without checking for the -1 (no next block) sentinel, producing an ou... |
| 2 | `tt_metal/impl/dispatch/kernels/cq_dispatch.cpp:1159` | process_go_signal_mcast_cmd re-reads cmd->mcast.num_unicast_txns after the kernel has intentionally overwritten that word of the command with the go-signal value,... |
| 3 | `tt_metal/impl/data_format/mx_common.hpp:221` | The subnormal branch of the MX float encoder rounds the shifted significand with input_width=24 while that value still has 23 fraction bits, so every subnormal MX... |
| 4 | `tt_metal/impl/data_format/mx_common.hpp:239` | MX float encoder masks the output mantissa with `params.elem_man_max` (a maximum VALUE, not a bit mask); for MXFP8 E4M3 (MxFp8P) elem_man_max == 0x6, so every odd... |
| 5 | `tt_metal/impl/debug/dprint_server.cpp:590` | poll_print_buffer() never validates the rpos it reads from L1 against DEVICE_PRINT_RESET_BUFFER_MAGIC (0xF0E1D2C3); when rpos holds that sentinel, `print_buffer_s... |
| 6 | `tt_metal/impl/kernels/kernel.cpp:507` | ComputeKernel::config_hash() omits config_.bfp8_pack_precise, so two compute kernels that differ only in bfp8_pack_precise get the same kernel hash, the same kern... |
| 7 | `tt_metal/impl/metal2_host_api/program_spec.cpp:2330` | The Metal 2.0 TensorParameter CTA encoder pushes the raw bank count into the num_banks CTA word instead of tensor_accessor::pack_num_banks(n_banks, is_shard_conti... |
[...]

PR #49417, opened 2026-07-08
[Feature] Bringup basic profiler on quasar

shengxiangjiTT:
### PR Category
Feature

### Summary
Relates to #48927 
Basic bringup of the profiler on quasar, wall clock counter only. The profiler is L1-only on quasar for now (no DRAM drain). Running kernels with profiler enabled will report the per-processor FW/kernel latencies into `profile_log_device.csv` similar to WH/BH.
Testing: The `test_custom_cycle_count` and `test_custom_cycle_count_slow_dispatch` programming examples are uplifted to metal 2.0 to enable running on quasar. The measured kernel latencies are between 104100 to 104600 cycles (with the kernel itself being an unrolled 104000 nops)

### Notes for reviewers
- WH/BH mailbox sizes increased by 128B as a result of the control vector size change

### Testing

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-profiler.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/perf-device-models.yaml?query=branch:shengxiangji/profiler-48927)

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:shengxiangji/profiler-48927)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:shengxiangji/profiler-48927)

PR #52549, opened 2026-08-07
[Diagnostic] Restore pre-#49417 WH/BH L1 map to confirm #51415 root c…

sanjaysundaramTT:
### PR Category
Test Only

### Summary
Not a merge candidate — this is the A/B half of a root-cause experiment for #51415
(`test_fabric_mux_v2_throughput` payload_sweep_1s_2048B_buf8_r0: 17.523 B/c vs golden
18.536, -5.5%).

Bisected to #49417, which widened `PROFILER_L1_CONTROL_VECTOR_SIZE` 32 → 64 words for
Quasar. That vector lives in `profiler_msg_t` inside `dev_msgs_t`, so the WH/BH mailbox
grew 128B, moving `MEM_MAP_END` 0x8840 → 0x88C0 and the L1 allocator base
0x19C40 → 0x19CC0. The mux-v2 ubench derives its entire buffer layout from that base.

This splits the slot reservation out of `PROFILER_MAX_RISC_COUNT` into
`PROFILER_CONTROL_RISC_SLOTS` and sets it back to 5, restoring the pre-#49417 L1 map
exactly. Relates to #51415, #51127.

### Notes for reviewers
Quasar genuinely needs 24 slots, so the real fix is to make this arch-dependent —
deliberately out of scope here. Also adds the bounds `static_assert` the control vector
never had; note it does **not** hold on main today (last `ControlBuffer` entry lands at
index 64 in a 64-word vector), which is the overflow in #51127.

### Testing Conducted --> Failed
Local T3K runs of `test_fabric_mux_v2_throughput` after #49417. The failing case stayed
at ~17.52 B/c (0.945x vs golden 18.536) unless the pre-#49417 mailbox / `MEM_MAP_END`
was restored. Sliding only the ubench buffers did not recover it:

- `align_up(base, 128/256/512/1024/4096)` — still 17.52. Power-of-two align-up from
  `0x19CC0` lands on phase `0x00`, not golden's `0x40`, and never moves `KERNEL_CONFIG`.
- `base + 1024` — still 17.52 (`+1024` is a multiple of the 256B L1 bank period, so
  the bank phase is unchanged).
- `base + 128` (rebuild confirmed: 42/105 JIT hits) — data phase restored to `0x40`,
  still 17.526 B/c. Mux/drainer/sender start address is not the cause.
- Forwarder+manager `-O3` — attempted as a kernel-codegen check; does not undo the
  128B `MEM_MAP_END` / BRISC XIP shift.

The only config that returned 18.535 B/c was restoring `PROFILER_CONTROL_RISC_SLOTS=5`
(CI, including with an extra `+1024` on the ubench base). That is diagnostic, not a
fix: Quasar still needs 24 slots. Ubench address hacks cannot replace an
arch-dependent control-vector size.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:sanjaysundaram/repro-51415-profiler-mailbox)

### DIFF
--- a/tests/tt_metal/tt_metal/perf_microbenchmark/routing/fabric_mux_v2_benchmark_program.cpp
+++ b/tests/tt_metal/tt_metal/perf_microbenchmark/routing/fabric_mux_v2_benchmark_program.cpp
@@ -18,6 +18,7 @@
 #include <tt-metalium/experimental/fabric/fabric.hpp>
 #include <tt-metalium/kernel_types.hpp>
 #include <tt-metalium/program.hpp>
+#include <tt-metalium/tt_align.hpp>
 #include <tt-metalium/tt_metal.hpp>
 
 #include "context/metal_context.hpp"
@@ -616,8 +617,9 @@ StandaloneMuxV2BenchmarkRunResult run_standalone_mux_v2_benchmark_once(
     const uint64_t expected_sender_bytes = static_cast<uint64_t>(resolved_payload_size_bytes) * num_packets;
     const uint64_t expected_aggregate_bytes = expected_sender_bytes * benchmark_case.num_senders;
 
-    const auto l1_unreserved_base_address =
-        static_cast<uint32_t>(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1)) + 1024;
+    constexpr uint32_t kL1PageSizeBytes = 1024;
+    const auto l1_unreserved_base_address = static_cast<uint32_t>(
+        tt::align(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1), kL1PageSizeBytes));
     const auto packet_header_size_bytes =
         static_cast<uint32_t>(tt::tt_fabric::get_tt_fabric_packet_header_size_bytes());
     const auto channel_buffer_size_bytes = packet_header_size_bytes + resolved_payload_size_bytes;
--- a/tests/tt_metal/tt_metal/perf_microbenchmark/routing/fabric_mux_v2_benchmark_program.cpp
+++ b/tests/tt_metal/tt_metal/perf_microbenchmark/routing/fabric_mux_v2_benchmark_program.cpp
@@ -617,7 +617,7 @@ StandaloneMuxV2BenchmarkRunResult run_standalone_mux_v2_benchmark_once(
     const uint64_t expected_aggregate_bytes = expected_sender_bytes * benchmark_case.num_senders;
 
     const auto l1_unreserved_base_address =
-        static_cast<uint32_t>(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1));
+        static_cast<uint32_t>(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1)) + 1024;
     const auto packet_header_size_bytes =
         static_cast<uint32_t>(tt::tt_fabric::get_tt_fabric_packet_header_size_bytes());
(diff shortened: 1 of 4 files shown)


shengxiangjiTT:
Synced with Sanjay over slack. I don't think changing the profiler is the right fix here. Sounds like this is a fabric perf degradation caused by a mailbox size increase, and is possibly caused by the change in the expected layout of the L1 addresses. If the perf is dependent on that, the kernel doing the allocation should make sure that the base address being used is expected (e.g. is a multiple of 1024B).

tests/tt_metal/tt_metal/perf_microbenchmark/routing/fabric_mux_v2_benchmark_program.cpp
```diff
@@ -617,7 +617,7 @@ StandaloneMuxV2BenchmarkRunResult run_standalone_mux_v2_benchmark_once(
     const uint64_t expected_aggregate_bytes = expected_sender_bytes * benchmark_case.num_senders;
 
     const auto l1_unreserved_base_address =
-        static_cast<uint32_t>(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1));
+        static_cast<uint32_t>(device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1)) + 1024;
```
aagarwalTT: can we use alignment APIs to align this address to 1K page instead of just adding 1K?

tt_metal/hostdevcommon/api/hostdevcommon/profiler_common.h
```diff
@@ -44,6 +44,9 @@ enum BufferIndex {
 // (8 DM + 4 x 4 TRISC).
 static constexpr std::uint32_t PROFILER_MAX_RISC_COUNT = 24;
 
+// TODO: make this architecture (wh/bh) specific --> needs to be moved out of this common profiler file.
+static constexpr std::uint32_t PROFILER_CONTROL_RISC_SLOTS = 5;
```
aagarwalTT: do we still need these men map/profiler changes?

sanjaysundaramTT: No, I will remove them.

PR #52551, opened 2026-08-07
[Feature] Add support for per core allocator on socket buffers

alingTT:
### PR Category
Feature

### Summary
Currently the data buffer of mesh sockets are lock step allocated, this means that the data buffer needs to be allocated even on all worker cores, which could be an issue for L1 space critical use cases. This PR adds support for per core allocation on socket buffers.

### CI Status
APC/Sanity tests: https://github.com/tenstorrent/tt-metal/actions/runs/31519721116

### DIFF
--- a/tt_metal/distributed/mesh_socket_utils.cpp
+++ b/tt_metal/distributed/mesh_socket_utils.cpp
@@ -275,14 +275,15 @@ std::shared_ptr<MeshBuffer> create_socket_data_buffer(
             socket_mem_config.socket_storage_type == BufferType::L1,
             "Per-core socket allocation is only supported for L1 storage (got {}).",
             socket_mem_config.socket_storage_type);
-        std::set<CoreCoord> receiver_cores;
+        std::unordered_set<MeshCoreCoord> receiver_cores;
         for (const auto& connection : config.socket_connection_config) {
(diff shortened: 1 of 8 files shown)


tt_metal/distributed/mesh_socket_utils.cpp
```diff
@@ -261,7 +269,30 @@ std::shared_ptr<MeshBuffer> create_socket_data_buffer(
 
     uint32_t num_data_cores = 0;
     CoreRangeSet shard_grid;
-    if (socket_mem_config.socket_storage_type == BufferType::DRAM) {
+    bool per_core = socket_mem_config.per_core_allocation;
+    if (per_core) {
```
alingTT, line 273: at least in blaze there is not a scenario where we have more than 1 connection per `SocketConfig`  so i think we can assert for this behaviour for now? unless there is a usecase for where we need more than 1 receiver core per socket config, if there is then i dont think the current `SocketPeerDescriptor` is sufficient to represent the data buffer addresses per receiver core.

tt-asaigal (approved):

riverwuTT (approved):

ntarafdar (approved):

jbaumanTT (approved):

TT-BrianLiu (approved):

Issue #48928, opened 2026-07-03
[test_perf_e2e_resnet50.py] | (T3K) T3000 perf tests | t3000-perf-tests / t3k_CNN_resnet50_model_perf_tests [wh_llmbox_perf]


ebanerjeeTT:
## CI Failure Report

**Summary:** [CI] ResNet50 non-trace inference time ~19x above target in t3k CNN perf tests (wh_llmbox_perf)

**Workflow:** (T3K) T3000 perf tests
**Job:** t3000-perf-tests / t3k_CNN_resnet50_model_perf_tests [wh_llmbox_perf]

### Regression timeline

- **Last passing commit:** N/A (not provided)
- **First failing commit:** `a3f0fca15cf1` — 2026-07-03T00:47:34Z — [job](https://github.com/tenstorrent/tt-metal/actions/runs/28630967894/job/84908724182)
- **Most recent failing commit:** `939b6db3b0f4` — 2026-07-03T12:43:53Z — [job](https://github.com/tenstorrent/tt-metal/actions/runs/28661347422/job/85004869006)

### Failing job URLs (last 2 runs)

- https://github.com/tenstorrent/tt-metal/actions/runs/30452986433/job/90584852888
- https://github.com/tenstorrent/tt-metal/actions/runs/30360088106/job/90280229980

### Failing test path(s)

- `models/demos/vision/classification/resnet50/ttnn_resnet/tests/test_perf_e2e_resnet50.py`
- Post-test performance check: `models/perf/merge_perf_results.py` (via `run_t3000_resnet50_tests`)

### Error excerpt (terminal failure from logs)

**Run 1** (commit `939b6db3b0f4`):
```
[FAIL] The following models failed 'Inference Time (sec)' checks:
  - ttnn_resnet50_2cqs_batch_size128_224x224_batchsize128: measured=0.3444, expected=0.0170, tolerance=[0.0136, 0.0170] (1925.9% regression)
  - ttnn_resnet50_batch_size128_224x224_batchsize128: measured=0.3434, expected=0.0175, tolerance=[0.0140, 0.0175] (1862.3% regression)
  - ttnn_resnet50_trace_batch_size128_224x224_batchsize128: measured=0.0039, expected=0.0038, tolerance=[0.0030, 0.0038] (2.6% regression)
Exception: Performance checks failed - see above for models with regressions or significant improvements requiring baseline updates
##[error]Process completed with exit code 1.
```

[...]

PR #52556, opened 2026-08-08
[ttnn] pad: replace get_dynamic_runtime_args with a factory override_runtime_arguments

dgomezTT:
### What changed

`ttnn.pad` no longer declares `get_dynamic_runtime_args`. The height-sharded row-major factory
(`PadRmShardedHeightOnlyProgramFactory`) now owns its cache-hit re-derivation through
`override_runtime_arguments` on the **program factory** — its natural home, next to
`create_descriptor` — so the other six pad factories are untouched and keep the behaviour they have
today.

The old hook re-applied no real value: it re-wrote writer arg 0 (the output shard height, which is
pinned by the hash) on one core purely to *trip* the framework's descriptor fast path, so that
`apply_resolved_bindings` would re-patch the sharded CB base addresses (#48928). The new hook does
that job directly.

### Why

Part of the `get_dynamic_runtime_args` elimination. `override_runtime_arguments` supersedes it and
the adapter's `static_assert` forbids declaring both, so the legacy hook is deleted in the same
change.

### What varies per dispatch

Pad has no custom `compute_program_hash`, so every per-core reader/writer arg is already pinned by
the hashed shapes, shard specs and attributes. The only per-dispatch state is the two sharded CB
base addresses — the input and output buffers ride on `cb.buffer`, not on a runtime arg. The
override re-points exactly those two through a minimal CB-only `ProgramDescriptor`: O(1), no
per-core `GetRuntimeArgs`, no `create_descriptor` rebuild on the hit path
(`scripts/detect_override_rebuild.py` passes).

### The cache-hit test

`test_pad_rm_sharded_height_only_override_addr_change` runs the same pad twice, keeping the first
input and output alive so the second pair is forced to different L1 addresses. It asserts both
address changes (so the test cannot silently go vacuous), asserts the second dispatch is a
program-cache hit, and requires bit-exact output for the second input.

Being explicit about what it does and does not prove:

- It **does** guard the new mechanism. With the CB re-pointing removed from the override body, the
  second dispatch reads the first input's shards and writes the first output's shards, and the test
  fails on the second data comparison (verified on a build with the two `cbs.push_back` lines
  removed).
- It is **not** a regression test against `main`: on unmodified `main` it passes, because the old
  `get_dynamic_runtime_args` had the framework patch the same two CB addresses. This migration is
  behaviour-preserving, so no test can fail before and pass after. The test exists to keep the new,
  explicit hook honest.

### Also in this diff

`tests/.../test_pad.py` had four pre-existing `pytest.raises` sites that the `prefer-expect-error`
pre-commit hook rejects for any commit touching the file; converted them to the `expect_error`
fixture. No behavioural change — same counts before and after.

### Testing

- `test_pad.py`: 507 passed, 63 skipped, 8 xfailed
- `test_pad_subcoregrids.py`: 58 passed
- `scripts/detect_override_rebuild.py` on the changed files: clean

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/gd-pad-override)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/gd-pad-override)

### DIFF
--- a/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
@@ -380,6 +380,61 @@ def test_pad_rm_sharded(device, n, c, h, w, padding, torch_padding, value, shard
         device.set_program_cache_misses_allowed(False)
 
 
+def test_pad_rm_sharded_height_only_override_addr_change(device):
+    """Cache-hit hook (override_runtime_arguments) for the CB-bound height-sharded RM factory: buffer
+    addresses are excluded from the pad hash, so a second identical pad at DIFFERENT input AND output
+    addresses must hit the cache and still re-point both sharded CB bases instead of reading/writing
+    the first dispatch's L1 shards."""
+    torch.manual_seed(48928)
+
+    n, c, h, w = 1, 3, 224, 256
+    padded_shape = [n, c, 256, w]
+    input_tensor_start = [0, 0, 0, 0]
+    pad_value = 8.0
+
+    num_cores_x, num_cores_y = 8, 7
+    ncores = num_cores_x * num_cores_y
+    shard_grid = ttnn.CoreRangeSet(
+        {ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(num_cores_x - 1, num_cores_y - 1))}
+    )
+    in_cfg = ttnn.MemoryConfig(
+        ttnn.TensorMemoryLayout.HEIGHT_SHARDED,
+        ttnn.BufferType.L1,
+        ttnn.ShardSpec(shard_grid, (ttnn.core.divup(n * c * h, ncores), w), ttnn.ShardOrientation.ROW_MAJOR),
+    )
+    out_cfg = ttnn.MemoryConfig(
+        ttnn.TensorMemoryLayout.HEIGHT_SHARDED,
+        ttnn.BufferType.L1,
+        ttnn.ShardSpec(
+            shard_grid,
+            (ttnn.core.divup(n * c * padded_shape[2], ncores), w),
+            ttnn.ShardOrientation.ROW_MAJOR,
+        ),
+    )
+
+    def _make():
+        t = torch.rand((n, c, h, w), dtype=torch.bfloat16)
+        return t, ttnn.from_torch(
+            t, dtype=ttnn.bfloat16, layout=ttnn.ROW_MAJOR_LAYOUT, device=device, memory_config=in_cfg
+        )
+
+    def _run(tt_in):
+        return ttnn.pad(tt_in, padded_shape, input_tensor_start, pad_value, memory_config=out_cfg)
+
+    torch_a, tt_a = _make()
+    tt_out_a = _run(tt_a)
+    assert_equal(pad_reference(torch_a, padded_shape, input_tensor_start, pad_value), ttnn.to_torch(tt_out_a))
+    entries_after_first = device.num_program_cache_entries()
+
+    # Keep A alive so B's input and output both land at different L1 addresses.
+    torch_b, tt_b = _make()
+    assert tt_b.buffer_address() != tt_a.buffer_address(), "second input must land at a different address"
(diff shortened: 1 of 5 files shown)


dgomezTT:
Heads up before anyone reviews this: the regression test added here does not actually
validate the new override. I ran it on hardware and then ran a negative control.

- Test passes with the override intact.
- Test still passes with the override body replaced by a bare `ret` (verified in the
  loaded `_ttnncpp.so`: symbol size `0x2f8` -> `0x1`).
- The override *is* invoked, once, on the cache-hit dispatch (checked with an fprintf),
  so the hook is wired correctly.

Reason: this factory is CB-bound, so it registers no `Buffer*` runtime args and
`resolved_bindings.rt_args` is empty. The adapter's cache-hit fast path is therefore not
eligible and it falls through to a full `invoke_per_coord()` +
`apply_descriptor_runtime_args()` rebuild, which re-derives every address. Cache-hit
correctness holds no matter what the hook does.

That matches what the deleted `get_dynamic_runtime_args` said about itself: it re-applied
one unchanged value (writer arg0 = output shard height) purely to "trip the fast-path
instead of rebuilding create_descriptor". It was a performance opt-in, not a correctness
fix. The property worth guarding is "no rebuild per dispatch", and nothing at runtime
currently reports which path was taken, so a pytest can't assert it.

Also worth knowing: this test never ran in CI. The `ttnn data movement group` in
`tests/pipeline_reorg/ttnn_sanity_tests.yaml` is an allowlist of individual node ids
(`test_pad.py::test_pad` only), and the whole-directory run in `ops_unit_tests.yaml` is
cron-only on manual dispatch. Green sanity here says nothing about the new test.

Holding this until we decide the direction rather than land a test that can't fail.

dgomezTT:
Retracting my previous comment — it was wrong. The test is fine and the override is
load-bearing. Please disregard that analysis.

What went wrong: my "negative control" ran against an incoherent build. I neutered the
override and rebuilt only the `ttnncpp` target, and the library pytest actually loaded
still contained the working code, so the test passed and I read that as "the test cannot
fail". Classic stale-`.so`.

Redone from a clean tree, rebuilding and refreshing the loaded
`build_Release/lib/_ttnncpp.so` each time, verifying the override's symbol size in the
loaded library (`0x2f8` real vs `0x1` neutered):

- real override -> test PASSES
- override replaced by a no-op -> test FAILS with `Data mismatch`

So it discriminates in both directions, and dropping the CB re-pointing genuinely corrupts
the second dispatch. My claim that the adapter's rebuild fallback made correctness
unconditional was also wrong: this factory declares `override_runtime_arguments`, so the
adapter takes the override branch, which by design does no `resolve_bindings` and never
rebuilds. The override is the only thing re-pointing those two CB bases on a cache hit.

The one part of the earlier comment that still holds: this test does not run in PR CI. The
`ttnn data movement group` in `tests/pipeline_reorg/ttnn_sanity_tests.yaml` is an allowlist
of individual node ids (`test_pad.py::test_pad` only), and the whole-directory run lives in
`ops_unit_tests.yaml`, which is cron-only on manual dispatch. I ran it on a wormhole_b0
card locally instead: passes with the override, fails without it.

Nothing about the diff changed as a result of this; it is ready for review.

ntarafdar (approved):

Issue #52593, opened 2026-08-08
[Numeric] ttnn.pow silently turns a NaN base/exponent into a finite value (no is_nan guard in the power fast path)


peter941221:
### Component / Area

ttnn numerics / SFPU kernels (`ttnn.pow`, `ttnn.rpow`)

### Observed

`ttnn.pow(x, y)` does not propagate NaN operands. If either `x` or `y` is NaN, the
SFPU fast path returns a finite value instead of NaN.

Root cause is the log2-based power fast path:

- `ckernel_sfpu_binary_pow.h` and `ckernel_sfpu_unary_power.h`
  (`tt_metal/hw/ckernels/<arch>/metal/llk_api/llk_sfpu/`)
- it normalizes the base with:
  ```cpp
  sfpi::vFloat abs_base = sfpi::abs(base);
  sfpi::vFloat m = sfpi::setexp(abs_base, 127);
  sfpi::vInt exp = sfpi::exexp(abs_base);
  ```
- `sfpi::setexp(NaN, 127)` rewrites a NaN operand into a finite value (it keeps the
  non-zero mantissa and forces the exponent to 127), so the NaN is silently destroyed.
- there is no `sfpi::is_nan` assertion anywhere in these two power kernels
  (`grep` finds zero uses), so nothing recovers the NaN on a subsequent branch.

Note: the existing `base == 0 && pow < 0` and negative-base handlers are not NaN
classifiers and do not cover this path.

### Expected vs observed

According to IEEE/C99/torch golden (`ttnn.pow` is registered against `torch.pow`):

| case | golden | observed (host block model) |
|---|---|---|
| `pow(NaN, 2.0)`   | NaN      | finite garbage |
| `pow(NaN, ±0.0)`  | 1        | 1 (via existing path) |
| `pow(1.0, NaN)`   | 1        | finite garbage |
| `pow(x, NaN)`     | NaN      | finite garbage |

The NaN-base idempotencies `pow(NaN, ±0) = 1` and `pow(base, 0) = 1` must be preserved,
so a fix must classify NaN with the sign-independent `sfpi::is_nan` and only force NaN
where the exponent/base is not one of the idempotent cases.

### Steps to reproduce

No device needed for the mechanism; the source-level claim is that `setexp` on a NaN
operand produces a finite value and no `is_nan` branch recovers it. A focused test
would construct NaN operands natively (e.g. `torch.full(..., float('nan'), dtype=torch.bfloat16)`)
[...]

PR #52599, opened 2026-08-08
ttnn.pow: propagate NaN base/exponent instead of returning a finite value

peter941221:
**PR Category**
Bug fix

**Summary**
`ttnn.pow` silently turns a NaN base or exponent into a finite number.

Root cause: the SFPU power fast path normalizes the base with
`setexp(abs_base, 127)` before the log2 polynomial. `setexp(NaN, 127)`
rewrites the exponent field while keeping the non-zero mantissa, so the NaN
bit pattern is destroyed before the result is produced, and there is no
`is_nan` guard anywhere in the power path to recover it.

Change: add a representation-neutral NaN guard in the binary power dispatcher
so both the bf16 (21f) and fp32 (f32) paths, on Wormhole and Blackhole, follow
IEEE-754 `pow()` semantics:
- `pow(NaN, y) = NaN` unless `y == +-0` (-> 1)
- `pow(x, NaN) = NaN` unless `x == 1.0` (-> 1)
The guard uses the existing `sfpi::is_nan()` primitive. The existing zero-base,
negative-base, and overflow/underflow handling is unchanged.

**Notes for reviewers**
- The idempotency exceptions are preserved on purpose: a blanket "any NaN ->
  NaN" would wrongly turn `pow(NaN, 0)` and `pow(1, NaN)` into NaN.
- Validation: host IEEE/numpy model (the 11-case matrix in #52593) matches
  numpy golden after the fix; a PyTorch regression in
  `tests/ttnn/unit_tests/operations/eltwise/test_pow.py` asserts the NaN cases
  against `torch.pow`. Device runtime accuracy of finite lanes is out of scope;
  this change is limited to NaN propagation semantics.

**Fixes** #52593

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
@@ -2,8 +2,8 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
-import torch
 import pytest
+import torch
 import ttnn
 from tests.ttnn.nightly.unit_tests.operations.eltwise.backward.utility_funcs import data_gen_with_range, compare_pcc
 from tests.ttnn.utils_for_testing import (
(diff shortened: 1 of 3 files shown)


mateusznowakTT:
@peter941221 : Thanks for the PR. Unfortunately, this change drops pow performance by 25% - unless there is a model/application that requires this exact behavior we won't merge it only for sake of IEEE compliance. Even then, it would have to be opt-in flag to enable this.

PR #52608, opened 2026-08-08
Fix xielu exponent underflow producing Inf/NaN for large negative inputs

### LINKED ISSUE #52607: [Bug] xielu returns Inf/NaN and values near 1e38 for large negative inputs
### Component / Area

ML operations (SFPU / LLK)

### Issue Type (optional)

Bad Outputs

### Observed

`ttnn.xielu` returns non-finite values and values around 1e38, including sign flips, for
large negative inputs where the true result is a small positive number around 26.

Measured on an N300 (Wormhole B0) sweeping x from -95 to -80 in 0.25 steps (60 inputs),
compared against the op's own golden function (`ttnn/ttnn/operations/unary.py`,
`alpha_p = alpha_n = 0.8`, `beta = 0.5`, `eps = -1e-6`):

| dtype | non-finite outputs | outputs > 1e30 | sign flips | worst input | device | true value |
| --- | --- | --- | --- | --- | --- | --- |
| bfloat16 | 1 | 9 | 1 | x = -88.5 | **Inf** | 25.75 |
| float32 | 3 | 8 | 2 | x = -90.75 | **NaN** | 26.425 |

**Root cause.** The large-negative branch calls the private helper `_sfpu_neg_exp_f32_` in
`tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_xielu.h`. That helper
clamps the scaled exponent at `UNDERFLOW_THRESHOLD = -126.5`, which stops the Cody-Waite
range reduction short for very negative inputs. The polynomial result `p` therefore keeps a
small exponent while `k_int` stays pinned near the clamp, so

```cpp
sfpi::vInt new_exp = p_exp + k_int;   // goes <= 0
result = sfpi::setexp(p, new_exp);    // unguarded
```

`setexp` writes only the low bits of `new_exp` into the exponent field, so a non-positive
`new_exp` wraps to a large positive exponent. The helper returns a huge value or Inf where
the true result should underflow toward zero, and the surrounding
`alpha_n * exp_term + beta * x` then produces the wrong final value.

**This is not a flush-to-zero artifact.** Across the entire failing window the true output
never falls below **|y| = 23.275**, far above the |y| >= 1e-3 threshold used in #45520 to
classify FTZ behaviour as expected. The device returns a wrong large number, not a flushed
small one.

**Related but distinct.** #51543 and #51544 cover the same defect class (dropped fp32
overflow/underflow guards around `setexp`) in the Quasar accurate-exp path. This is a
different file and a different helper on Wormhole and Blackhole, and it is not fixed by
#51544. #34594 fixed the analogous exp overflow near x = -88.

**No existing test covers this region.** The only xielu test,
`tests/ttnn/unit_tests/operations/eltwise/test_activation.py::test_xielu`, is fed
`torch.randn([32, 32])`, so |x| never approaches 89.

### Expected

`xielu` should return a finite value close to the golden reference for all finite inputs.
At x = -88.5 the expected result is about 25.75, and at x = -90.75 about 26.425.

### 1. Steps (exact commands)

```python
import torch
import ttnn

def golden(x, alpha_p=0.8, alpha_n=0.8):
    x = x.to(torch.float32)
    beta, eps = 0.5, -1e-6
    pos = alpha_p * x * x + beta * x
    xc = torch.minimum(x, torch.full_like(x, eps))
    neg = alpha_n * torch.expm1(xc) - alpha_n * x + beta * x
    return torch.where(x > 0, pos, neg)

xs = torch.arange(-95.0, -80.0, 0.25, dtype=torch.float32)
buf = torch.full((1, 1, 32, 64), -1.0, dtype=torch.float32)
buf[0, 0, 0, : xs.numel()] = xs

t = ttnn.from_torch(buf, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=dev)
got = ttnn.to_torch(ttnn.xielu(t))[0, 0, 0, : xs.numel()].float()

print("non-finite outputs:", int((~torch.isfinite(got)).sum()))
print("outputs > 1e30:", int((got.abs() > 1e30).sum()))
i = int((got - golden(xs)).abs().argmax())
print("worst x", float(xs[i]), "device", float(got[i]), "true", float(golden(xs)[i]))
ttnn.close_device(dev)
```

No input data needed. Any finite input in roughly [-90.5, -87.7] reproduces it; the sweep
above covers the region.

### 3. Frequency

Consistent and deterministic. Reproduces on every run, in both bfloat16 and float32.

### 1. Software Versions

tt-metal main at 565641307c.

### 2. Hardware Details

Wormhole B0, N300 (2 chips). The affected code is present in both the `wormhole_b0` and
`blackhole` copies of `ckernel_sfpu_xielu.h`.

### Is this a regression?

No. The unguarded `setexp` has been present since xielu was added in #38637. Every commit
touching `ckernel_sfpu_xielu.h` since then has been a mechanical API migration.

### Regression Details

_No response_

### Logs & Diagnostics

After adding the exponent guard, measured on the same N300 over the same sweep:

| dtype | non-finite outputs | outputs > 1e30 | max absolute error |
| --- | --- | --- | --- |
| bfloat16 | 0 | 0 | 0.125 (1 ULP at this magnitude) |
| float32 | 0 | 0 | 1e-05 |

Results over the normal operating range (x in [-10, 10]) are unchanged: bfloat16 max
absolute error 0.553009 and float32 8e-06 both before and after.

A fix is proposed in the linked PR, which guards the exponent the same way
`_sfpu_exp_fp32_accurate_` does in `ckernel_sfpu_exp.h`, patches both architecture copies,
and adds a regression test that fails without the fix.

### Priority

Medium

### Impact

Silent wrong results from an activation function. A model using xielu with any activation
reaching roughly -88 will see Inf, NaN or values near 1e38 propagate instead of a value
around 26, which will show up as a training or inference divergence rather than an obvious
error. #34594 documents this failure mode for the analogous exp bug.

AJ0070:
### Ticket

Fixes #52607

### Problem description

`ttnn.xielu` returns non-finite values and values around 1e38, including sign flips, for
large negative inputs where the true result is a small positive number around 26.

The large-negative branch scales by `2**k` through `setexp`. The private helper
`_sfpu_neg_exp_f32_` clamps the scaled exponent at `UNDERFLOW_THRESHOLD = -126.5`, which
stops the Cody-Waite range reduction short for very negative inputs. The polynomial result
`p` therefore keeps a small exponent while `k_int` stays pinned near the clamp, so
`new_exp = p_exp + k_int` goes non-positive. `setexp` writes only the low bits of `new_exp`
into the exponent field, so a non-positive `new_exp` wraps to a large positive exponent.

Measured on an N300 (Wormhole B0), sweeping x from -95 to -80 in 0.25 steps against the op's
own golden function with `alpha_p = alpha_n = 0.8`:

| dtype | non-finite outputs | outputs > 1e30 | worst input | device | true value |
| --- | --- | --- | --- | --- | --- |
| bfloat16 | 1 | 9 | x = -88.5 | Inf | 25.75 |
| float32 | 3 | 8 | x = -90.75 | NaN | 26.425 |

This is not a flush-to-zero artifact: across the entire failing window the true output never
falls below |y| = 23.275.

### What's changed

Guard the exponent before `setexp`, mirroring the guard `_sfpu_exp_fp32_accurate_` already
uses in `ckernel_sfpu_exp.h`. Only the underflow half is needed, because this helper is
called exclusively with negative inputs, so `exp(val) < 1` and the exponent cannot overflow.
`result` is already initialised to `0.0f`, which is the correct underflow answer.

Applied to both the `wormhole_b0` and `blackhole` copies of `ckernel_sfpu_xielu.h`.

Adds `test_xielu_large_negative`, covering x in [-95, -80] for both dtypes and two alpha
combinations. The existing `test_xielu` is fed `torch.randn([32, 32])`, so |x| never
approaches 89 and the region was untested.

### Verification on N300 (Wormhole B0)

After the fix, over the same sweep:

| dtype | non-finite outputs | outputs > 1e30 | max absolute error |
| --- | --- | --- | --- |
| bfloat16 | 0 | 0 | 0.125 (1 ULP at this magnitude) |
| float32 | 0 | 0 | 1e-05 |

No regression over the normal operating range (x in [-10, 10]): bfloat16 max absolute error
0.553009 and float32 8e-06, both identical before and after the change.

Test results on hardware:

- `pytest tests/ttnn/unit_tests/operations/eltwise/test_activation.py -k xielu`: **12 passed**
  (8 existing plus 4 new).
- With the kernel guard reverted and the new test kept, the new test **fails 4 of 4** with
  "xielu returned a non-finite value for a finite input", confirming it is a genuine
  regression test.

### Checklist

- [x] Tested on Wormhole B0 (N300) hardware
- [x] Applied to both wormhole_b0 and blackhole kernel copies


AJ0070:
This one is ready whenever a reviewer has a cycle. Context for picking it up cold:

- Fixes #52607: `ttnn.xielu` returns Inf/NaN and values around 1e38 with sign flips for large negative inputs, where the true result is a small positive number. The large-negative branch scales by `2**k` through `setexp` and overflows before the guard; the fix bounds the exponent so the branch stays finite.
- Ships with a regression test covering the previously non-finite range.
- CI is clean; the only red checks are pre-existing infrastructure jobs unrelated to this change.

Per the earlier discussion this is landing as an ordinary contribution rather than a bounty. It is currently marked BEHIND but is conflict-free; happy to rebase onto current main if that makes review easier. @jberkowitzTT no rush, just flagging it since it has been quiet for a week.

dchenTT (approved):

AJ0070:
Merged latest `main` (322d58254c) to clear the merge conflict — this is now conflict-free and ready to merge pending CodeOwners sign-off.

The only conflict was in `test_activation.py`: #53700 migrated the bfloat16 coverage tests out of this file into `test_unary_category1_bfloat16.py`. I took that migration as-is (dropping the moved `test_lgamma_bfloat16` rather than reintroducing it) and kept only this PR's own addition, `test_xielu_large_negative`. The branch now diffs against `main` as exactly the three intended files, and the two kernel changes were unaffected by the merge.

Thanks @dchenTT for the approval.

AJ0070:
Approved and conflict-free since the 21st, still waiting on CodeOwners. @jberkowitzTT could you re-ping?

fvranicTT (approved):

PR #52611, opened 2026-08-08
fix(eltwise): invert the divisor instead of its square in four backward composites (same op count)

### LINKED ISSUE #52376: [ttnn]: backward composites invert the divisor's square, losing the gradient for |x| < 1.08e-19 and again for |x| > 2^63
### Issue Type (optional)

Bad Outputs

### Observed

Four backward composites evaluate a term of the form `k / x²` by forming `x²` and inverting it.
That intermediate has a much narrower usable range than the result does, so the gradient is lost at
**both** ends of the divisor's range while the exact answer is an ordinary number:

```
rdiv_bw(grad=1, input=1e-19, scalar=1e-30)  ->  -inf   exact: -1e8
rdiv_bw(grad=1, input=1e19,  scalar=1e19)   ->   0     exact: -1e-19
```

The information is destroyed one step early, and no amount of accuracy in the reciprocal can
recover it.

### Three thresholds, all from the same intermediate

| condition | what happens | result |
|---|---|---|
| `\|x\| < 1.0842e-19` | `x²` falls below the smallest normal and flushes to zero | `inf` |
| `\|x\| > 2⁶³` | `1/x²` falls below the smallest normal — **the reciprocal** flushes | `0` |
| `\|x\| > 2⁶⁴` | `x²` exceeds the largest normal, so the reciprocal receives `inf` | `0` |

`2⁶³ = 9.2233720e18` is exact rather than approximate: `(2⁶³)² = 2¹²⁶ = 8.507059e37`, and
`1/2¹²⁶ = 1.1754943e-38` is the smallest normal float32, so the reciprocal underflows for anything
past that point.

The middle row is the one worth naming. Between `2⁶³` and `2⁶⁴` **the square is perfectly
representable** — `(9.22e18)² = 8.5e37`, well inside `FLT_MAX = 3.4028235e38` — and the loss
happens in the reciprocal instead. A check written only against the square overflowing misses that
half of the affected range. The `input=1e19` row above is exactly this case.

### Root cause

`ttnn/cpp/ttnn/operations/eltwise/unary_backward/unary_backward.cpp`, `rdiv_bw`:

```cpp
ttnn::multiply(
    ttnn::neg(grad, output_mem_config),
    (ttnn::multiply(
        ttnn::reciprocal(ttnn::square(input, output_mem_config)), scalar, std::nullopt, output_mem_config)),
    std::nullopt,
    output_mem_config),
```

The quantity being computed is `-grad * scalar / x²`. Written that way, both `x²` **and** `1/x²`
must be representable — but the *result* need not be anywhere near either edge, and in the rows
above it is not.

### Where it appears

The eltwise backward directories were swept for the same `square` → `reciprocal` pattern rather
than reporting one instance:

| site | file | line at the SHA in Software Versions |
|---|---|---|
| `rdiv_bw` | `eltwise/unary_backward/unary_backward.cpp` | 193 |
| `reciprocal_bw` | `eltwise/unary_backward/unary_backward.cpp` | 1352 |
| `div_bw` (tensor/tensor, `other_grad`) | `eltwise/binary_backward/binary_backward.cpp` | 716 |
| `addcdiv_bw` (`tensor2` gradient) | `eltwise/ternary_backward/ternary_backward.cpp` | 61 |
| `atan2_bw` | `eltwise/binary_backward/binary_backward.cpp` | 103 |

**`atan2_bw` is named but treated separately.** It has the same `SQUARE` → `RECIP` chain, on
`hypot(a, b)`, and the same failure — 592 of 2078 representable cases — but its reciprocal is
shared between both outputs, so the reformulation in Expected costs one extra multiply per element
there.
That is a tradeoff rather than a free correction, and it seemed better to name it than to fold it
in quietly.

**The scalar overload of `div_bw` is not affected.** It computes `float inv_scalar = 1.0f / scalar`
on the host and never forms a tensor square. `div_bw` has two overloads in that file; only the
tensor/tensor one — the one returning `std::vector<std::optional<Tensor>>` — has this.

### Measured

Every input `float32`, with the device's flush-to-zero behaviour on subnormals modelled, compared
against the same expression evaluated in `float64` and rounded to `float32`. Cases whose exact
answer is not representable are excluded — infinity is the right answer there and is not being
judged. Cases whose exact answer is subnormal are excluded and counted separately: those flush to
zero identically before and after, which is architectural rather than a defect.

| site | representable cases | wrong today | wrong after the reformulation | regressions |
|---|---|---|---|---|
| `rdiv_bw` | 1440 | 182 | 0 | 0 |
| `div_bw` (`other_grad`) | 1670 | 282 | 0 | 0 |
| `reciprocal_bw` | 952 | 112 | 0 | 0 |
| `addcdiv_bw` | 750 | 90 | 0 | 0 |
| `atan2_bw` (not proposed for change) | 2078 | 592 | 0 | 0 |

Of `rdiv_bw`'s 182, **72 are the small end and 110 the large end**, so a fix aimed at one end alone
would leave most of it in place.

The control on those counts: three deliberately wrong variants of the replacement — numerator not
interleaved, sign flipped, one reciprocal instead of two — were pushed through the same comparison
and reported 174, 1188 and 1328 failures rather than zero, so the checker is not passing everything
it is handed.

### Why the existing tests miss it

Backward coverage draws divisors from moderate ranges, where both the square and its reciprocal are
harmless:

- `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_rdiv.py` builds its
  input with `data_gen_pt_tt`, which is `torch.randn(...)` — standard normal, so `|x|` stays around
  1.
- `test_backward_reciprocal.py` and `test_backward_addcdiv.py` use `data_gen_with_range` with
  bounds of `-1..1` and `-100..100`.
- `tests/sweep_framework/sweeps/eltwise/unary_backward/rdiv_bw/rdiv_bw.py` and
  `tests/sweep_framework/sweeps/eltwise/binary_backward/div_bw/` draw from `low=-100, high=100`
  (`low=0.1, high=100` for the divisor).

Nothing samples `|x|` below `1e-19` or above `1e18`, and no invariant is asserted that would catch
this independently of the sampled values.

### Not the same as the open issues on these ops

`#48240` ("Eltwise host code: 28 bugs") does reach `div_bw`, on different defects: item 1 is that
`div_bw` discards the results of two `where()` edge cases, and item 18 is an unchecked array index
in the `RDIV` rounding-mode lookup. Neither concerns the squared intermediate.

`#52094` reports `ttnn.div` returning `0` for large denominators. That is the SFPU division kernel —
`calculate_sfpu_binary_div` and `sfpu_reciprocal_iter` — a different code path from these host-side
composites. **It is not being claimed here**, and the reformulation in Expected does not touch it: the
loss described above happens in the composite's own `square`/`reciprocal` pair, before any division
kernel is reached.

`#51831` (bfloat8_b sub-tile) is a different mechanism, and states that bfloat16 and float32 are
correct for the ops it covers.

### Expected

The gradient should match the autograd reference whenever the exact gradient is representable in
the output dtype — including when the divisor's square, or the reciprocal of that square, is not.
For the two rows in Observed:

| call | today | expected |
|---|---|---|
| `rdiv_bw(grad=1, input=1e-19, scalar=1e-30)` | `-inf` | `-1e8` |
| `rdiv_bw(grad=1, input=1e19, scalar=1e19)` | `0` | `-1e-19` |

That expectation is not a preference. PyTorch evaluates the same derivative without ever forming
the square: `div_tensor_other_backward` computes `-grad * ((self / other) / other)`, dividing twice.
That matters beyond being prior art — **those torch backward passes are the registered goldens for
these ops**, so the reference the existing tests already compare against returns the finite value
while the device path returns `inf` or `0`. Measured rather than assumed, `torch 2.13.0+cpu`,
`float32`:

```
d/dx (1e-30 / x) at x = 1e-19  ->  -1e8
d/dx (1e-30 / x) at x = 1e-20  ->  -1e10
d/dx (1e19  / x) at x = 1e19   ->  -1e-19
```

### The reformulation that gets there, at no op cost

Inverting `x` first and folding the numerator in between the two multiplies keeps every
intermediate in range:

```
   reciprocal(square(x)) * k       ->       (reciprocal(x) * k) * reciprocal(x)
   1 multiply + 1 recip + 1 multiply         1 recip + 1 multiply + 1 multiply
```

Same op count, same dependency depth, no extra tensor. `addcdiv_bw` comes out one op cheaper,
because the reciprocal of `tensor2` is already being computed for the other gradient and can be
shared instead of squaring and inverting a second time.

Folding the numerator *between* the two multiplies is load-bearing rather than cosmetic: computing
`reciprocal(x) * reciprocal(x)` first overflows, and fails 174 of the same cases the current code
fails.

### Success criteria

1. `rdiv_bw`, `reciprocal_bw`, the `other_grad` output of `div_bw` and the `tensor2` gradient of
   `addcdiv_bw` match the autograd reference to the ops' existing tolerance whenever the exact
   gradient is representable.
2. Specifically for `|x| < 1.0842e-19` and `|x| > 2⁶³`, including the band between `2⁶³` and `2⁶⁴`
   where the square is still representable and only the reciprocal underflows.
3. `float32` and `bfloat16` — they share an exponent field and therefore share all three
   thresholds.
4. Values in the currently-working band are unchanged, and a test samples divisors across the
   exponent range rather than relying on a random draw landing there.
5. No added element-wise work on the supported band.

### 1. Read the four expressions

```bash
git clone https://github.com/tenstorrent/tt-metal.git && cd tt-metal
git checkout 68b985e7ed5164faf2ac95ba3a0530c8a1488482
grep -rn "reciprocal(ttnn::square" ttnn/cpp/ttnn/operations/eltwise/
```

Four hits, one per affected site (`unary_backward.cpp:193` and `:1352`,
`binary_backward.cpp:716`, `ternary_backward.cpp:61`). `atan2_bw` reaches the same chain through
`binary_backward.cpp:103`, where `SQUARE` and `RECIP` are composed as `EltwiseUnaryWithParam`.

### 2. Reproduce the arithmetic on the host — no device needed

The failure is in the order of the host-side composite, so it reproduces from the expression alone.
Save as `repro_divbw.py` and run with `python repro_divbw.py`:

```python
import torch

TINY = torch.finfo(torch.float32).tiny  # 1.1754944e-38

def ftz(t):  # model the device flushing subnormals to zero
    return torch.where(t.abs() < TINY, torch.zeros_like(t), t)

def today(grad, x, k):  # -grad * (reciprocal(square(x)) * k)
    return ftz(-grad * ftz(ftz(torch.reciprocal(ftz(x * x))) * k))

def proposed(grad, x, k):  # -grad * ((reciprocal(x) * k) * reciprocal(x))
    r = ftz(torch.reciprocal(x))
    return ftz(-grad * ftz(ftz(r * k) * r))

def exact(grad, x, k):  # float64, rounded back to float32
    return (-grad.double() * k / (x.double() ** 2)).float()

one = torch.ones(1)
print(f"{'x':>9} {'scalar':>9} {'today':>13} {'proposed':>13} {'exact':>13}")
for xv, kv in [(1e-18, 1e-30), (1e-19, 1e-30), (1e-20, 1e-30),
               (1e18, 1e18), (1e19, 1e19), (1e20, 1e20), (3.0, 6.0)]:
    x = torch.tensor([xv])
    print(f"{xv:9.0e} {kv:9.0e} {today(one, x, kv).item():13.6g} "
          f"{proposed(one, x, kv).item():13.6g} {exact(one, x, kv).item():13.6g}")

print()
print(f"smallest normal float32 = {TINY:.6e}   sqrt = {TINY ** 0.5:.6e}")
print(f"2**63 = {2.0 ** 63:.6e}   (2**63)**2 = {(2.0 ** 63) ** 2:.6e}   1/that = {1 / (2.0 ** 63) ** 2:.6e}")
print(f"2**64 = {2.0 ** 64:.6e}   FLT_MAX    = {torch.finfo(torch.float32).max:.6e}")

print()
for xv in (1e-19, 1e-20, 1e19):
    x = torch.tensor([xv], requires_grad=True)
    torch.div(torch.tensor([1e-30 if xv < 1 else 1e19]), x).backward(torch.ones(1))
    print(f"torch reference: d/dx (k/x) at x={xv:.0e} -> {x.grad.item():.6g}")
```

`1e-18` and `1e18` sit just inside the working band at either end and are there as controls that
the band is untouched. Output is in the Logs box.

### 3. On a machine with a card

```python
import torch, ttnn

for xv, k in [(1e-18, 1e-30), (1e-19, 1e-30), (1e-20, 1e-30),
              (1e18, 1e18), (1e19, 1e19), (1e20, 1e20)]:
    x = torch.full((1, 1, 32, 32), xv, dtype=torch.float32)
    g = torch.ones((1, 1, 32, 32), dtype=torch.float32)
    xt = ttnn.from_torch(x, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
    gt = ttnn.from_torch(g, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
    out = ttnn.to_torch(ttnn.rdiv_bw(gt, xt, k)[0])[0, 0, 0, 0].item()

    xr = torch.tensor([xv], requires_grad=True)
    torch.div(torch.tensor([k]), xr).backward(torch.ones(1))
    print(f"x={xv:9.0e} scalar={k:9.0e}  ttnn={out:13.6g}  torch={xr.grad.item():13.6g}")
ttnn.close_device(device)
```

The same rows should disagree there, with `-inf` at the small end and `0` at the large end against
a finite torch value. This step has not been run here — see the Hardware Details box.

No external data, model or checkpoint. The repro builds its own tensors: single-element (or
32x32-filled) `float32` tensors holding a constant, plus a gradient tensor of ones.

The divisors are chosen by exponent, not sampled:

| `input` | `scalar` | why it is in the list |
|---|---|---|
| `1e-18` | `1e-30` | control — just inside the working band at the small end |
| `1e-19` | `1e-30` | below `sqrt(FLT_MIN) = 1.0842e-19`; the square flushes |
| `1e-20` | `1e-30` | further below the same threshold |
| `1e18` | `1e18` | control — just inside the working band at the large end |
| `1e19` | `1e19` | above `2⁶³`; the square is fine, the reciprocal underflows |
| `1e20` | `1e20` | above `2⁶⁴`; the square overflows |
| `3` | `6` | ordinary value, unchanged either way |

Grad is `1` everywhere so the printed number is the derivative itself.

The same shape of input applies to the other three sites: `div_bw` and `addcdiv_bw` take the
extreme value in the divisor position (`other` and `tensor2` respectively), `reciprocal_bw` in its
only input.

### 3. Frequency

Consistent - deterministic, every run.

The three thresholds are properties of the float32 exponent range, not of a sampled value: the
square of anything below `1.0842e-19` is subnormal, and the reciprocal of the square of anything
above `2⁶³` is subnormal. There is no timing, no layout and no random draw involved, so a given
input either falls in the affected band or it does not, and it does so identically on every run.

`bfloat16` shares the float32 exponent field, so it shares all three thresholds exactly.

### 1. Software Versions

| | |
|---|---|
| tt-metal | `main` at `68b985e7ed5164faf2ac95ba3a0530c8a1488482` (2026-08-04) |
| PyTorch | `2.13.0+cpu` |
| Python | `3.14.5` |
| OS | Windows 11 (host-side analysis only) |

All line numbers quoted in this report are at that SHA.

The host repro needs only PyTorch; it does not require a tt-metal build, because the expressions it
reproduces are the composite's own host-side op sequence.

### 2. Hardware Details

**No Tenstorrent card was used.** Saying so up front, because it bounds what the numbers in this
report are worth.

What this report actually rests on:

- **Source reading** of the four composites at the SHA in the previous box. The op sequence
  `square` → `reciprocal` → `multiply` is read directly from
  `unary_backward.cpp`, `binary_backward.cpp` and `ternary_backward.cpp`.
- **Reimplementing that op sequence in `float32`** on the host, with flush-to-zero on subnormals
  modelled explicitly, and comparing it against the same expression evaluated in `float64` and
  rounded back to `float32`, and against `torch`'s own backward pass. That is where every count in
  the Measured table comes from. The script is in the Steps box and its output is in the Logs box.
- **One related measurement under ttsim** (the functional simulator, Wormhole target, `bfloat16`,
  clean container): `ttnn.reciprocal` returns `0` once its result would be subnormal —
  `reciprocal(8e37) = 1.24896e-38` but `reciprocal(1e38) = 0`, against `1e-38` from torch. A
  round-trip control in the same run (`from_torch(5e-39)` came back as `4.95912e-39`) shows the
  simulator does represent subnormals, so that `0` is specific to the reciprocal rather than a
  general flush in the data path. That is the mechanism behind the second row of the thresholds
  table. **`float32` eltwise could not be exercised under ttsim**, so the `float32` figures in this
  report are the host model, not a simulator run.

So: the thresholds and the op sequence are read from the source and verified arithmetically; the
per-op failure counts are from the host model; the reciprocal's subnormal behaviour has one
`bfloat16` simulator data point behind it. Nothing here was observed on silicon, and the on-device
step in the Steps box is written out precisely so someone with a card can confirm or refute it
quickly.

### Is this a regression?

Unknown

### Regression Details

Not bisected, and no claim that these expressions ever behaved differently. They are present in the
form quoted at `68b985e7ed5164faf2ac95ba3a0530c8a1488482`; the history behind them was not walked,
so "Unknown" is the honest answer rather than "No".

### Logs & Diagnostics

Output of the host repro in the Steps box, `torch 2.13.0+cpu` / Python 3.14.5. `today` is the
composite's current op order with subnormal flush-to-zero modelled; `proposed` is the reformulation
described in Expected; `exact` is `float64` rounded back to `float32`.

```
        x    scalar         today      proposed         exact
    1e-18     1e-30        -1e+06        -1e+06        -1e+06
    1e-19     1e-30          -inf        -1e+08        -1e+08
    1e-20     1e-30          -inf        -1e+10        -1e+10
    1e+18     1e+18        -1e-18        -1e-18        -1e-18
    1e+19     1e+19             0        -1e-19        -1e-19
    1e+20     1e+20             0        -1e-20        -1e-20
    3e+00     6e+00     -0.666667     -0.666667     -0.666667

smallest normal float32 = 1.175494e-38   sqrt = 1.084202e-19
2**63 = 9.223372e+18   (2**63)**2 = 8.507059e+37   1/that = 1.175494e-38
2**64 = 1.844674e+19   FLT_MAX    = 3.402823e+38

torch reference: d/dx (k/x) at x=1e-19 -> -1e+08
torch reference: d/dx (k/x) at x=1e-20 -> -1e+10
torch reference: d/dx (k/x) at x=1e+19 -> -1e-19
```

The `1e+18` and `3` rows are the controls: the working band is identical under both orderings.
The `1e+19` row is the middle threshold — `(1e19)² = 1e38` is comfortably inside `FLT_MAX`, and the
value is still lost, in the reciprocal.

No device log to attach; the on-device snippet in Steps prints the same table with `ttnn.rdiv_bw`
in the `today` column, and has not been run here.

### Priority

P2

### Impact

These are backward ops, so the wrong value does not surface where it is produced. An `inf` in a
gradient tensor becomes `NaN` after the first optimiser step that touches it. The silent `0` at the
other end is the worse of the two, because the run does not die — it just stops learning that
parameter, with nothing pointing back here.

Neither end is exotic input. Small divisors are what a normalisation term looks like once the
gradients have shrunk; large ones are what a variance or accumulator term looks like before they
do. And because the failure is one step upstream of the value being read, the usual PCC check on
the op that consumed it reports a plausible-looking number.

The band between `2⁶³` and `2⁶⁴` is worth flagging separately for anyone auditing this class of
expression elsewhere: the square is representable there, so a guard written against the square
overflowing will look correct and still lose the value.

A change for the four sites is written and ready — same op count, one op cheaper in `addcdiv_bw` —
along with a test that fails on `main` and passes with it, covering both ends and both dtypes. It
will be opened against this issue.

EazyHood:
### Ticket

Fixes #52376.

### What's wrong

`rdiv_bw`, `reciprocal_bw`, `div_bw` and `addcdiv_bw` evaluate a term of the form `k / x²` by
forming `x²` and inverting it. That intermediate has a much narrower usable range than the result
does, so the gradient is lost at **both** ends of the divisor's range:

```
rdiv_bw(grad=1, input=1e-19, scalar=1e-30)  ->  -inf      exact: -1e8
rdiv_bw(grad=1, input=1e20,  scalar=1e20)   ->  0         exact: -1e-20
```

| condition | what happens | result |
|---|---|---|
| `\|x\| < 1.0842e-19` | `x²` falls below the smallest normal and flushes to zero | `inf` |
| `\|x\| > 2⁶³` | `1/x²` falls below the smallest normal — **the reciprocal** flushes | `0` |
| `\|x\| > 2⁶⁴` | `x²` exceeds the largest normal, so the reciprocal receives `inf` | `0` |

`2⁶³` is exact: the reciprocal underflows once `x² > 1/2⁻¹²⁶`, and `√2¹²⁶ = 2⁶³`. Between `2⁶³`
and `2⁶⁴` the square is still perfectly representable — `(9.22e18)² = 8.5e37` — and the loss is in
the reciprocal, so a bound written against the square's overflow alone would miss half the range.

### What changed

Invert `x` first and fold the numerator in between the two multiplies:

```diff
- ttnn::multiply(ttnn::reciprocal(ttnn::square(input, mc)), scalar, std::nullopt, mc)
+ Tensor recip_input = ttnn::reciprocal(input, mc);
+ ttnn::multiply(ttnn::multiply(recip_input, scalar, std::nullopt, mc), recip_input, std::nullopt, mc)
```

**The op count is unchanged** — one reciprocal and two multiplies either way, same dependency
depth, no extra tensor. `addcdiv_bw` comes out one op cheaper: the reciprocal of `tensor2` was
already being computed for the other gradient, so it is now shared rather than squared and
inverted a second time.

Folding the numerator *between* the two multiplies is load-bearing. Computing
`reciprocal(x) * reciprocal(x)` first overflows, and fails 174 of the same cases the current code
fails; this ordering is the one that keeps both ends in range.

### What is deliberately not changed

`atan2_bw` has the same `SQUARE`→`RECIP` chain, on `hypot(a, b)`, and the same failure — 592 of
2078 representable cases. It is **not** in this PR: its reciprocal is shared between both outputs,
so the same reformulation costs one extra multiply per element there. That is a tradeoff rather
than a free correction, and it belongs in a decision rather than in a bug fix. Happy to add it if
you want it.

The scalar overload of `div_bw` is unaffected — it computes `1.0f / scalar` on the host and never
forms a tensor square.

### Measured

Every input `float32`, with the device's flush-to-zero behaviour modelled, against the same
expression in `float64` rounded to `float32`. Cases whose exact answer is not representable are
excluded — infinity is correct there. Cases whose exact answer is subnormal are excluded and
counted separately: the device flushes those to zero identically before and after.

| site | representable cases | wrong before | wrong after | regressions |
|---|---|---|---|---|
| `rdiv_bw` | 1440 | 182 | **0** | **0** |
| `div_bw` (`other_grad`) | 1670 | 282 | **0** | **0** |
| `reciprocal_bw` | 952 | 112 | **0** | **0** |
| `addcdiv_bw` | 750 | 90 | **0** | **0** |

`rdiv_bw`'s 182 split 72 / 110 between the small and large ends, so neither end is a rounding
detail on top of the other.

The control that made those numbers worth printing: three deliberately broken variants of the
replacement — numerator not interleaved, sign flipped, one reciprocal instead of two — were run
through the same sweep and reported 174, 1188 and 1328 failures. The checker is not silently
passing everything.

### Test

`tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_small_divisor.py`, one
case per op across four divisor magnitudes at each end and both dtypes. `1e-18` and `1e18` sit
just inside the working band at either end and are included as controls that it is untouched.

The registered goldens for these ops are torch's own backward passes, which evaluate the same
derivative as `-grad * ((a / b) / b)` — dividing twice rather than squaring — so they return the
finite value and the comparison is meaningful. Verified rather than assumed: `torch 2.13`,
`float32`, `torch.div(1e-30, x).backward()` gives `-1e8` at `x = 1e-19` and `-1e10` at
`x = 1e-20`.

`reciprocal_bw` gets a shorter list on purpose, and the file says why: torch evaluates *that*
derivative as `-grad * (1/x) * (1/x)`, so its own reference leaves the range below `5.4e-20` and
above roughly `2.7e22` and stops being usable as a golden there. What remains still straddles the
working band on both sides.

Both dtypes are covered because `float32` and `bfloat16` share an exponent field and therefore
share all three thresholds — a test that ran only one of them would not be narrower in an
interesting way, but it would let a later change regress the other quietly.


EazyHood:
Merged current `main`, so this is mergeable again. Two conflicts, both from changes that landed next to this fix, and both kept:

- `div_bw`: #55901 removed the `where()` blocks whose results were discarded. That removal stays, along with this PR's `recip_other.deallocate()`.
- `addcdiv_bw`: #53899 forwards `output_mem_config` on the outer multiply of `grad_a`. That stays too, now around the shared reciprocal.

The change itself is the same four files: the divisor is inverted instead of its square in `rdiv_bw`, `reciprocal_bw`, `div_bw` and `addcdiv_bw`, plus `test_backward_small_divisor.py`.

PR #52666, opened 2026-08-09
Make log2 exact for powers of two

### LINKED ISSUE #52665: [Bug] ttnn.log2 is not exact for powers of two in float32 (46 of 254 exponents off by an ULP)
### Component / Area

ML operations (SFPU / LLK), `ttnn.log2`

### Issue Type (optional)

Bad Outputs

### Observed

`ttnn.log2` does not return the exact exponent for an exact power of two in float32.
**46 of the 254 representable exponents come back an ULP low.**

Measured on an N300 (Wormhole B0), sweeping `2**k` for k in [-126, 127]:

```
powers of two returning the exact exponent: 208/254

log2(2**-125)  device -124.99999237   expected -125
log2(2**-124)  device -123.99999237   expected -124
log2(2**-120)  device -119.99999237   expected -120
log2(2**-119)  device -118.99999237   expected -119
```

**Root cause.** There is no dedicated log2 kernel. `ttnn.log2` is dispatched as natural
log multiplied by a base constant
(`ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_op_utils.cpp`,
`log_with_base_tile(idst, 0x3fb8aa3bu)` where `0x3fb8aa3b` is 1/ln(2)). In
`ckernel_sfpu_log.h` the natural log is formed first and the base change is applied to
the finished sum:

```cpp
result = e_float * sfpi::vConstFloatPrgm0 + r;   // vConstFloatPrgm0 = ln(2) * 2^-23

if constexpr (HAS_BASE_SCALING) {
    result *= sfpi::as<sfpi::vFloat>(sfpi::vUInt(log_base_scale_factor));
}
```

That multiply also scales the exponent contribution, so the exponent path effectively
gets multiplied by `ln(2) * (1/ln(2))`. Those are two separately rounded float
constants and their product does not round to exactly 1, so the exponent term is
perturbed. For an exact power of two the mantissa term `r` is 0 and the entire result
is that perturbed exponent, which is why the error is visible precisely at the inputs
where the answer should be exact.

The mantissa term does not have this problem, because scaling it is the mathematically
correct operation. Only the exponent term needs to stay unscaled.

Note bfloat16 is unaffected (254/254 exact): the error is below bf16 resolution.

### Expected

`log2(2**k) == k` exactly for every representable exponent, as it is on the CPU
reference and in bfloat16 on device today.

### 1. Steps (exact commands)

```python
import torch
import ttnn

ks = list(range(-126, 128))
buf = torch.ones((1, 1, 32, 256), dtype=torch.float32)
buf[0, 0, 0, : len(ks)] = torch.tensor([2.0**k for k in ks], dtype=torch.float32)

t = ttnn.from_torch(buf, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=dev)
got = ttnn.to_torch(ttnn.log2(t))[0, 0, 0, : len(ks)].float()

bad = [(ks[i], float(got[i])) for i in range(len(ks)) if float(got[i]) != float(ks[i])]
print("exact:", len(ks) - len(bad), "/", len(ks))
print("first few wrong:", bad[:4])
ttnn.close_device(dev)
```

No input data required. Any exact power of two in the affected set reproduces it.

### 3. Frequency

Consistent and deterministic.

### 1. Software Versions

tt-metal main at 003feb6663.

### 2. Hardware Details

Wormhole B0, N300. The same code path is present in the `blackhole` copy of
`ckernel_sfpu_log.h`.

### Is this a regression?

Unknown. The base-scaling multiply has been applied to the finished sum for as long as
the current log kernel has existed.

### Logs & Diagnostics

Applying the base change to the mantissa term only, for base 2, makes the remaining
exponent factor exactly `2^-23`. Since the exponent arrives as `k << 23` and `2^-23` is
a power of two, the product is exact. Measured on the same N300 after that change:

| metric (fp32) | before | after |
| --- | --- | --- |
| powers of two exact | 208/254 | **254/254** |
| max relative error vs float64 | 1.026e-07 | **6.764e-08** |
| mean relative error vs float64 | 3.080e-08 | **2.370e-08** |

`log10` and plain `log` are unaffected by that change (every metric identical before and
after), as are all special values. Instruction count is unchanged, and on a 4096x4096
fp32 tensor `log2` measures 0.8505 ms/call against 0.8576 ms/call for `log10`, which
still uses the old base-scaled path in the same build.

Worth noting for anyone tempted to generalise it: the equivalent base 10 constant
`log10(2) * 2^-23` is not exactly representable, and folding it in measured **worse**
than the existing path (powers of ten exact dropped from 59/76 to 47/76). The fix
should stay scoped to base 2.

A fix is proposed in the linked PR.

### Priority

Low

### Impact

Correctness and IEEE conformance rather than a large numerical error: the existing worst
case is about 1 ULP. It matters for code that relies on `log2` of a power of two being
exact, for example deriving an integer shift or bit width, where an answer of
-124.99999237 truncates to a different integer than -125.

AJ0070:
### Ticket

Fixes #52665

### Problem description

`ttnn.log2` does not return the exact exponent for an exact power of two in float32:
**46 of the 254 representable exponents come back an ULP low**, for example
`log2(2**-125)` returns `-124.99999237`.

There is no dedicated log2 kernel. log2 is dispatched as natural log times 1/ln(2), and
`ckernel_sfpu_log.h` applies that base change to the finished sum:

```cpp
result = e_float * sfpi::vConstFloatPrgm0 + r;   // vConstFloatPrgm0 = ln(2) * 2^-23
if constexpr (HAS_BASE_SCALING) {
    result *= sfpi::as<sfpi::vFloat>(sfpi::vUInt(log_base_scale_factor));
}
```

That multiply also scales the exponent contribution, so the exponent path is effectively
multiplied by `ln(2) * (1/ln(2))`. Those are two separately rounded float constants and
their product does not round to exactly 1. For an exact power of two the mantissa term
`r` is zero and the whole result is that perturbed exponent, which is why the error shows
up precisely where the answer should be exact.

### What's changed

For base 2 only, apply the base change to the mantissa term and leave the exponent term
scaled by `2^-23`. The exponent arrives as `k << 23` and `2^-23` is a power of two, so the
product is exact for every k.

Scoped through a new defaulted `base_is_two` template parameter on `log_with_base_tile`,
set from the `LOG2` dispatch. Both the `wormhole_b0` and `blackhole` kernel copies are
updated.

**Deliberately not generalised to other bases.** The equivalent base 10 constant
`log10(2) * 2^-23` is not exactly representable, and folding it in measured *worse* than
the existing path: powers of ten returning the exact exponent dropped from 59/76 to 47/76,
and mean relative error rose from 3.50e-08 to 4.80e-08. log10 keeps the current path.

### Verification on N300 (Wormhole B0)

Accuracy, fp32, against a float64 reference:

| metric | before | after |
| --- | --- | --- |
| log2 powers of two exact | 208/254 | **254/254** |
| log2 max relative error | 1.026e-07 | **6.764e-08** |
| log2 mean relative error | 3.080e-08 | **2.370e-08** |

Nothing else moves. These were measured before and after and are **identical**:

- `log10`: powers of ten 59/76, max rel err 1.3055e-07, mean 3.4974e-08
- `log`: max rel err 6.2245e-08, mean 2.0345e-08
- log2 in bfloat16: 254/254 exact both before and after
- special values for all three ops in both dtypes: `0.0, -0.0, inf, -inf, nan, -1.0, 1.0`
  produce identical results, and log2 fp32 matches torch exactly
  (`-inf, -inf, inf, nan, nan, nan, 0.0`)

Cost is unchanged, one multiply plus one multiply-add either way. On a 4096x4096 fp32
tensor, averaged over 30 calls after warmup:

| op | ms/call |
| --- | --- |
| log | 0.8468 |
| **log2 (new path)** | **0.8505** |
| log10 (old base-scaled path, same build) | 0.8576 |

log2 is 0.83% faster than log10, which still takes the old path in the same build, so the
restructure is at worst neutral.

Existing tests, all passing: `test_math.py` 6, `test_unary.py` 77, `test_unary_ops_ttnn.py`
78, `test_unary_fp32.py` 1, filtered to log-related cases. The one unrelated failure in
that filter is `test_unary_logical_not[BFLOAT8_B]`, matched only because `logical` contains
`log`; it uses a different kernel and is unaffected by this change.

Adds `test_log2_exact_powers_of_two`, which asserts exactness across all 254 exponents and
would have caught the 46 failures reported in the issue.

### Checklist

- [x] Tested on Wormhole B0 (N300) hardware
- [x] Both wormhole_b0 and blackhole kernel copies updated
- [x] log10 and log verified byte-identical before and after


KalaivaniMCW:
Summary of my exhaustive testing for this branch : 

image

Key Takeaways

The PR makes 99.55% of all float32 log2 outputs bit-exact(up from 70.1%), a gain of 627 million values.
- Mean relative error improves 63x.
- Max ULP stays at 2, so the fix doesn't introduce any new worst-case outliers.
- One notable point: the ULP=2 count went from 9,390 to 38,278. This is because inputs that were previously 1 ULP off in the wrong direction can now land 2 ULP away from the torch reference when the rounding goes the other way on the mantissa path. So we end up with  (0.002%) in ULP 2. 

Note: I have not tested the the spl cases ie. 0/inf/nan here

Aswinmcw:
BH sanity - https://github.com/tenstorrent/tt-metal/actions/runs/31463566564
All models - https://github.com/tenstorrent/tt-metal/actions/runs/31463578190
Single card models - https://github.com/tenstorrent/tt-metal/actions/runs/31463586542

AJ0070:
@Aswinmcw Can you please trigger the CI?

AJ0070:
@Aswinmcw CI is green!

KalaivaniMCW (approved):

Aswinmcw:
WH BH sanity - https://github.com/tenstorrent/tt-metal/actions/runs/32211998570
All models - https://github.com/tenstorrent/tt-metal/actions/runs/32212009774
Frequent models - https://github.com/tenstorrent/tt-metal/actions/runs/32212017324

AJ0070:
@Aswinmcw Failures are unrelated to this PR

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_log.h
```diff
@@ -40,7 +40,7 @@
 namespace ckernel {
 namespace sfpu {
 
-template <bool FAST_APPROX, bool HAS_BASE_SCALING, bool is_fp32_dest_acc_en>
+template <bool FAST_APPROX, bool HAS_BASE_SCALING, bool is_fp32_dest_acc_en, bool BASE_IS_TWO = false>
```
fvranicTT:
```suggestion
template <bool FAST_APPROX, bool HAS_BASE_SCALING, bool is_fp32_dest_acc_en, bool IS_BASE_TWO = false>
```
This is more in line with other parameters in the signature: is_fp32..., has_base_scaling, etc.

fvranicTT (approved):

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_log.h
AJ0070: Done in b2d2d04ba9, thanks. Note the suggestion had only been applied to this one blackhole template declaration, which left the usages below (lines 98/141/145) and the whole wormhole_b0 twin still on `BASE_IS_TWO` (inconsistent, would not compile). I've completed the rename to `IS_BASE_TWO` across every occurrence in both files.

AJ0070:
@fvranicTT

Issue #51227, opened 2026-07-27
[ttnn] 3 high-severity defects in typecast / kv_cache / randn


amahmudTT:
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

Three unrelated high-severity defects, one each in `copy/typecast`, `kv_cache` and `randn`.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

Item 3 is the odd one out and the easiest to confirm: unseeded `ttnn::randn` re-seeds a fresh `std::mt19937` from `std::time(nullptr)` per call, so two calls in the same wall-clock second return bitwise-identical tensors. Every sibling RNG op already seeds once from a persistent generator, so there is an in-tree pattern to copy.

| # | Sev | Location | Defect |
|---|---|---|---|
| 1 | H | [`copy/typecast/device/typecast_rm_chunked_program_factory.cpp:40`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/copy/typecast/device/typecast_rm_chunked_program_factory.cpp#L40) | CB page size is padded only to 32 elements, but the compute kernel packs a full 32x32 tile per page, so any row narrower than 1024 elements makes pack_tile write... |
| 2 | H | [`kv_cache/device/update_cache_multi_core_program_factory.cpp:278`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/kv_cache/device/update_cache_multi_core_program_factory.cpp#L278) | u_count = u_range / granularity truncates when the cache batch size is odd and >= 3, so the reader/compute/writer kernels iterate over only (Bcache-1) users per b... |
[...]

PR #52667, opened 2026-08-09
#51227: fix unseeded randn repeating

AJ0070:
### Ticket

Item 3 of #51227.

### Problem description

Unseeded `randn` repeats within a wall-clock second. Both `create()` and
`override_runtime_arguments()` built a fresh `std::mt19937(std::time(nullptr))` per
dispatch. `std::time` has one-second resolution, so every unseeded call made in the
same second received an identical seed and returned a bitwise-identical tensor. An
explicit seed was unaffected.

### What's changed

One persistent generator for the unseeded path, so consecutive calls advance instead
of restarting from the same seed. This matches what `rand`, `uniform` and `bernoulli`
already do. An explicit seed still constructs its own generator and still reproduces
exactly. The generator has internal linkage, as in `rand_program_factory.cpp`.

### Measured

Four unseeded 256x256 calls on an N300:

- before: 3 identical pairs across 1.97 s
- after: 0 identical pairs

`test_randn_unseeded_calls_differ` and `test_randn_same_seed_is_reproducible` cover
both directions. The existing matrix only ever passed an explicit seed, which is why
this was latent.

### Scope note

This PR originally also carried item 2 of #51227, the `update_cache` granularity
truncation. That has been dropped: #52789 fixes the same lines more generally and
also covers `Bcache > 32`, which the change here did not. Item 1 of #51227
(`typecast_rm_chunked` CB page size) is not addressed by either.

### Checklist


ttnn/cpp/ttnn/operations/kv_cache/device/update_cache_multi_core_program_factory.cpp
```diff
@@ -154,7 +154,13 @@ ProgramDescriptor UpdateCacheMultiCoreProgramFactory::create_descriptor(
 
     uint32_t B = input_tensor.padded_shape()[-2];
     uint32_t Bcache = cache_tensor.padded_shape()[0];
-    const uint32_t granularity = std::min(static_cast<uint32_t>(2), Bcache);  // granularity = 2 best for performance
+    // Each kernel walks u_count iterations of an inner loop of `granularity` user rows, so
+    // granularity must divide u_range exactly or the trailing rows are never visited.
+    // Deriving granularity from u_range preserves that invariant. The previous
+    // std::min(2, Bcache) left u_count = u_range / 2 truncating for odd Bcache >= 3, which
+    // silently skipped the last user's cache update.
```
vtsilytskyiTT: Last sentence belongs to PR description and shouldn't stay in the code. Comments must reflect current state, not all the history of changes.

ttnn/cpp/ttnn/operations/randn/device/randn_program_factory.cpp
```diff
@@ -16,6 +17,13 @@ using namespace tt::tt_metal;
 
 std::uniform_int_distribution distribution(1, std::numeric_limits<int32_t>::max());
 
+// Persistent generator for the unseeded path.  Constructing a fresh
+// std::mt19937(std::time(nullptr)) per dispatch gave every call made within the same
+// wall-clock second an identical seed sequence, so consecutive unseeded randn calls
+// returned bitwise-identical tensors.  Seeding one generator once and letting it advance
+// matches rand, uniform and bernoulli.
+std::mt19937 default_rng(std::random_device{}());
```
vtsilytskyiTT: Same. You may leave only first sentence here, rest is too verbose and obsolete to stay here.

vtsilytskyiTT (approved):
Approve, overall looks good. Please address 2 minor comments.

ttnn/cpp/ttnn/operations/kv_cache/device/update_cache_multi_core_program_factory.cpp
AJ0070: Done in 62831be1a3. Dropped the sentence about the previous `std::min(2, Bcache)` behaviour and kept only the invariant a reader needs: granularity has to divide `u_range` exactly or the trailing rows are never visited. The before-and-after belongs in the PR description, agreed.

ttnn/cpp/ttnn/operations/randn/device/randn_program_factory.cpp
AJ0070: Done in 62831be1a3. Reduced to one line: `// Persistent generator for the unseeded path, so consecutive calls advance instead of repeating.` That keeps the reason the generator is file-scope, so it does not get moved back inside the function, without the history of what the old code did.

AJ0070:
@vtsilytskyiTT addressed both comments!

AJ0070:
Rebased onto current main and squashed to a single commit. The fix itself is unchanged.

AJ0070:
Reduced to the randn half. The update_cache granularity change is superseded by #52789, which fixes the same lines more generally and also covers Bcache > 32. The randn seeding fix is independent of that work.

vtsilytskyiTT:
> Reduced to the randn half. The update_cache granularity change is superseded by #52789, which fixes the same lines more generally and also covers Bcache > 32. The randn seeding fix is independent of that work.

Please update PR description as well. Git commit message depends on the description.

ttnn/cpp/ttnn/operations/randn/device/randn_program_factory.cpp
```diff
@@ -16,6 +17,9 @@ using namespace tt::tt_metal;
 
 std::uniform_int_distribution distribution(1, std::numeric_limits<int32_t>::max());
 
+// Persistent generator for the unseeded path, so consecutive calls advance instead of repeating.
+std::mt19937 default_rng(std::random_device{}());
+
```
vtsilytskyiTT, line 24: One small style note: `default_rng` sits in `ttnn::operations::randn` with external linkage. rand/uniform keep theirs in an anonymous namespace. Worth matching, so nothing else can link to this symbol.

AJ0070: Good catch, fixed in 9d6f833. Wrapped it in an anonymous namespace to match `rand_program_factory.cpp`.

AJ0070:
Description updated to cover item 3 only, with a note on why the update_cache half was dropped.

vtsilytskyiTT (approved):

AJ0070:
@vtsilytskyiTT Can you please trigger the CI again?

bbradelTT (approved):

vtsilytskyiTT:
@AJ0070 there is a performance problem:

```
  =================================== FAILURES ===================================
  ___________ test_dram_group_norm_vae_welford_reciprocal_performance ____________
  models/demos/stable_diffusion_xl_base/tests/test_sdxl_op_unit_test_perf.py:150: in test_dram_group_norm_vae_welford_reciprocal_performance
      assert (
  E   AssertionError: Performance outside expected range. Got 1353513.00 ns, expected 1331396 ± 1.5% (1311425.06-1351366.94 ns)
  E   assert 1353513.0 <= 1351366.94
```

bbradelTT:
Looks like Sanity is broken on main: https://github.com/tenstorrent/tt-metal/actions/runs/33613433877
We'll need to wait until things are fixed.

bbradelTT:
Sanity seems green again. @AJ0070 please rebase when you get the chance and we'll try re-running the workflows.

AJ0070:
rebased

vtsilytskyiTT:
[wormhole_b0 cdxl op test](https://github.com/tenstorrent/tt-metal/actions/runs/33708393683/job/100572332451?pr=52667) failure happens in other PRs (https://github.com/tenstorrent/tt-metal/actions/runs/33736322634/job/100614772715?pr=55306), not related to the current change.

Can't confirm [DeepSeek](https://github.com/tenstorrent/tt-metal/actions/runs/33708393683/job/100572285016?pr=52667).

AJ0070:
DeepSeek: the two failures are `test_decoder_block.py::test_forward_pass[...MoEDecoderBlock2D-model.layers.3...]` and `test_mode_decode_forward_pass_batch_8_users_per_row[...]`, both PCC ~0.002. #39159 (open since March) names the same test on Galaxy.

This PR touches only `randn_program_factory.cpp` and `test_randn.py`, and `models/demos/deepseek_v3/` has no `ttnn.randn` usage at all — it uses `torch.randn` on the host, so there is no path from this change to those tests.

bbradelTT:
Margins have been updated for sdxl via https://github.com/tenstorrent/tt-metal/pull/55470
Please rebase [again].

AJ0070:
@bbradelTT rebased

ijankowskiTT (approved):

Issue #51218, opened 2026-07-27
[ttnn/eltwise] 6 high-severity defects: inverted nextafter, 3 program-cache-key omissions, 2 missing packer restores


amahmudTT:
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

Six high-severity defects in `ttnn/cpp/ttnn/operations/eltwise/`, each producing silently wrong results (or, for item 3/4, a wrongly-formatted output tensor) on a supported configuration.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS: @tenstorrent/metalium-developers-eltwise

Three of the six fall into two shared fix patterns, so they are cheaper together than apart:

- **Program-cache key omissions (items 2, 5, 6)** — `compute_program_hash` leaves out state that the descriptor bakes into compile-time args, so two dispatches differing only in that state collide and the second silently reuses the first one's kernels. Same defect in three ops; the fix is to add the missing term to each hash.
- **Missing packer restore (items 3, 4)** — after an LLK row-broadcast packs into `c_4`/`c_5`/`c_6`, the final result is packed into `c_3` with the packer still configured for the broadcast CB's data format. Byte-for-byte the same omission in two ternary compute kernels; fix both.
- **Item 1** is standalone: `ttnn::nextafter` moves the result the wrong way.

| # | Sev | Location | Defect |
|---|---|---|---|
| 1 | H | [`eltwise/binary/device/binary_composite_op.cpp:40`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/eltwise/binary/device/binary_composite_op.cpp#L40) | ttnn::nextafter nudges the result AWAY from input_b instead of toward it: both branches of the where-chain have the wrong sign. |
[...]

PR #52668, opened 2026-08-09
fix(eltwise): step one ULP toward the target in nextafter, not a fixed epsilon away from it

EazyHood:
### Ticket

Addresses item 1 of #51218 (`ttnn::nextafter` moves away from `input_b`). Deliberately not using a closing keyword — that issue tracks six defects and this PR only covers the first.

### What's wrong

#51218 has the diagnosis right: both arms of the `where` chain are swapped, so the op steps away from `input_b` instead of toward it. The suggested fix there is to swap the two value arms back.

I would have sent exactly that patch, so I measured it first, and it does not fix the op. There is a second, independent defect in the same function: **`eps` is not a ULP.** It is `hal::get_eps()`, which the source documents as `FLT_EPSILON`:

```c
// tt_metal/hw/ckernels/{blackhole,wormhole_b0}/metal/llk_api/llk_math_common_api.h:17
#define EPS 1.19209e-07  // std::numeric_limits::epsilon() for FP32
```

That is the ULP at `1.0`, used as an absolute increment at every magnitude. It is one ULP only inside `[1, 2)`. Above that band `a + eps` rounds straight back to `a`; below it the step overshoots by several ULPs.

Reimplementing the composite as written — same branches, same constant — in float32 and bfloat16 with round-to-nearest-even, over 120 (magnitude, direction) pairs from `1e-6` to `1e6` in both directions:

| | float32 | bfloat16 |
|---|---|---|
| today | **0 / 120 correct** | **0 / 120 correct** |
| swapping the two arms, as #51218 suggests | **3 / 120** | **6 / 120** |
| this PR | **120 / 120** | **120 / 120** |

The middle row is the reason this is not a two-line change. With the step size left alone, the direction fix leaves the op wrong in 117 of 120 cases; in bfloat16 the composite is the *identity* in 102 of the 120, a bfloat16 ULP at 1.0 being `2^-8` against the `2^-23` being added.

### What changed

The magnitude's bit pattern is stepped by one, which is exactly one ULP at every magnitude, subnormals included. Working on the magnitude and restoring the sign afterwards keeps the sign handling in one place and leaves `input_a == 0` as the only special case: zero has no magnitude bits to walk down, and its neighbour in either direction is the smallest subnormal, carrying the sign of `input_b`.

`ttnn::bitcast` already supports the two pairs this needs — `UINT32 <-> FLOAT32` and `UINT16 <-> BFLOAT16` — and `where` and the integer `add`/`subtract` paths already accept `UINT32`/`UINT16`, so no new device support is involved.

### Measured

The composite as written in this PR, emulated in float32 and bfloat16 against `torch.nextafter`: **245 of 245 exact**, across twelve decades in both directions and both signs, plus `input_a == 0` in both directions and `input_a == input_b`.

### Test

#51218 also notes that the existing test cannot catch this, and that is worth acting on rather than just recording. The test was not using the wrong golden — `torch.nextafter` is correct and already attached — it was asserting with `assert_with_pcc`. Since `nextafter` differs from its input by one ULP, a PCC threshold cannot separate `a + 1ulp` from `a - 1ulp`, nor either from `a` left alone. The test passes today against an implementation that is wrong in every case.

It now asserts equality against `torch.nextafter`: across binade boundaries, at magnitudes where a fixed epsilon vanishes entirely, with negative operands in both directions, with `input_a == 0` in both directions, and with `input_a == input_b`.

### What I could not check

**This has not been compiled or run on a device.** I have no Tenstorrent hardware; every number above comes from reimplementing the arithmetic on x86 and comparing against `torch`. The first thing I would look at on real silicon is the `input_a == 0` path, which produces the smallest subnormal — if the device flushes it, that case needs handling closer to `nan_to_num` than a plain bitcast. I would rather say so than have a reviewer find it.

If someone already has a fix in flight for this item, say so and I will close this.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
@@ -48,8 +48,9 @@ def test_binary_nextafter_ttnn(input_shapes, device):
     golden_function = ttnn.get_golden_function(ttnn.nextafter)
     golden_tensor = golden_function(in_data1, in_data2)
 
-    comp_pass = compare_pcc([output_tensor], [golden_tensor])
-    assert comp_pass
+    # nextafter moves its input by one ULP, so a PCC check cannot referee it: it cannot
+    # separate a+1ulp from a-1ulp, nor either of them from a left alone.
+    assert_with_ulp(golden_tensor, output_tensor, 0)
 
 
 @pytest.mark.parametrize(
--- a/tests/ttnn/unit_tests/operations/eltwise/test_nextafter.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_nextafter.py
@@ -8,20 +8,66 @@ import torch
 
 import ttnn
 
-from tests.ttnn.utils_for_testing import assert_with_pcc
+
+# nextafter moves its input by exactly one ULP, so a correlation check cannot referee it: a PCC
+# threshold cannot separate `a + 1ulp` from `a - 1ulp`, nor either of them from `a` left alone.
+# These assert equality against torch.nextafter instead.
 
 
-@pytest.mark.parametrize("shape", [(1, 1, 32, 32)])
-def test_nextafter(device, shape):
-    torch.manual_seed(0)
+@pytest.mark.parametrize("dtype", [torch.float32, torch.bfloat16])
+@pytest.mark.parametrize(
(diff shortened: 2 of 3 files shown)


mateusznowakTT:
@EazyHood : nextafter is being reworked from composite to SFPU with #56050 making this PR irrelevant.

PR #52669, opened 2026-08-09
[Feature] Add native 1D and 2D DWT/IDWT operations to TTNN

### LINKED ISSUE #40494: [Bounty $2.5k] Implement Lifting Wavelet Transform (LWT) and Inverse LWT (ILWT)
# Implement Lifting Wavelet Transform (LWT) and Inverse LWT (ILWT)

Implement an optimized LWT and ILWT kernels for TT-Metal/TTNN on Wormhole/Blackhole hardware, which would enable usage of discrete wavelet transforms (DWTs) on device.

## 📝 Background

Discrete wavelet transform will help us enable multi-resolution signal processing directly on Wormhole/Blackhole hardware. This unlocks:
- JPEG2000 lossy compression.
- Real-time Digital Signal Processing pipelines.
- Audio, biosignals, radar, and seismic data processing applications.

## 🎯 What Success Looks Like

Requirements
- Support 1D/2D inputs with batch dimension (B, 1, H, W).
  - Layout can be both row major and tile. Preferably RM for 1d input, and tile layout for 2d input.
- Support the wavelets from PyWavelets catalog:
  - Haar: db1
  - Daubechies: dbN
  - Biorthogonal (used in JPEG2000): biorN.M
  - Symlets: symN
  - Coiflets: coifN
  - Discrete Meyer: dmey
- Support fp32 (SFPU-based implementation)
- Should be implemented as ttnn operation (`ttnn.dwt`/`ttnn.idwt`)
  - Should be implemented as optimized multicore kernel (not a naive composite version)
- Precision targets for the SFPU fp32 (Wavelet categories are defined in [the table below](#wavelet-categories)):
- Compact lifting schemes:
  - Definition: `tap_size <= 12`, `num_steps <= 11`, and `max_abs_coeff < 1e2`.
  - Forward transformation precision should be measured against PyWavelets using L2 relative error:
    - target median L2 relative error < 1e-7;
    - target 95th percentile L2 relative error < 1e-5.
  - Reconstruction precision should be measured against the original signal using L2 relative error:
    - target median L2 relative error < 1e-7;
    - target 95th percentile L2 relative error < 1e-5.

- Medium lifting schemes:
  - Definition: `12 < tap_size <= 24`, `12 <= num_steps <= 23`, and `max_abs_coeff < 1e4`.
  - Forward transformation precision should be measured against PyWavelets using L2 relative error:
    - target median L2 relative error < 1e-6;
    - target 95th percentile L2 relative error < 1e-4.
  - Reconstruction precision should be measured against the original signal using L2 relative error:
    - target median L2 relative error < 1e-6;
    - target 95th percentile L2 relative error < 1e-4.

- Large / numerically sensitive lifting schemes:
  - Definition: `tap_size > 24`, or `num_steps > 23`, or `max_abs_coeff >= 1e4`.
  - Forward precision is best effort.
  - Reconstruction precision is best effort.
  - These wavelets should report per-wavelet median / p95 / max L2 relative error and known outliers, but should not be treated as hard-gating failures.

- Precision reports should always include a per-wavelet breakdown and explicitly call out known SFPU fp32 outliers.

- No strict requirements for potential lower precision implementation.

- Optionally
  - Investigate FPU implementation (im2col/conv based) with tf32/bf16 + acc fp32 precision
  - Support l1 sharded inputs
  - Support arbitraty number of dimenions
  - Support for multi device support (i.e. device-sharded input)

## 💡 Problem to Solve

Implementation will
SFPU implementation will require adding a new stencil LLK function.
- Y stencil for 2d input should be a bit easier. Adjacent dst_reg rows are neighbors.
- X stenctil will require `subvec_shflror1` / `subvec_shflshr1` for lane-level neighbor access
- reader will require a careful halo management for fused hor/ver implementation (Maybe `untilize_with_halo_v2` will be useful)

Horizontal and vertical uplifting kernels might be fused. Cycle over levels is not required to be fused, but this can be investigated. Host pseudocode will look something like that:

```
for level in 0 .. J-1:
    active = top-left LL region from previous level

    kernel A: horizontal lifting on active region
    kernel B: vertical lifting on active region
    # Or kernel AB for both

    active = LL(active)
```

## ⏳Deadline

31.05.2026

## 🧭 Guidance & Starting Points

Lifting Wavelet Transform is based on the approach proposed by [Sweldens et al. (1996)](https://doi.org/10.1006/acha.1996.0015), we have already factorized PyWavelets catalog of filter banks into [lifting steps](https://github.com/draklowell/lifting-factorization) using modified algorithm that is based on [Daubechies et al. (1998)](https://doi.org/10.1007/bf02476026).

---
NOTE THAT THIS BOUNTY IS SUBJECT TO THE [TENSTORRENT BOUNTY PROGRAM TERMS & CONDITIONS](https://docs.tenstorrent.com/bounty_terms.html)

## 📚 Resources

- Sweldens, W. (1996). "The Lifting Scheme: A Custom-Design Construction of Biorthogonal Wavelets" - [Paper](https://doi.org/10.1006/acha.1996.0015)
- Daubechies, I., Sweldens, W. (1998). "Factoring Wavelet Transforms into Lifting Steps" - [Paper](https://doi.org/10.1007/bf02476026)
- FFT Feature Request on tt-metal - [#21412](https://github.com/tenstorrent/tt-metal/issues/21412)
- Nick Brown et al. (2025). "Exploring Fast Fourier Transforms on the Tenstorrent Wormhole" - [Paper](https://arxiv.org/abs/2506.15437)
- PyWavelets - [GitHub](https://github.com/PyWavelets/pywt)
- Lifting Factorization - [GitHub](https://github.com/draklowell/lifting-factorization)

## Wavelet Categories

| Wavelet | Category | tap_size | num_steps | max_abs_coeff |
|---|---:|---:|---:|---:|
| `bior1.1` | Compact | 2 | 5 | `1.41421` |
| `bior1.3` | Compact | 6 | 4 | `1.41421` |
| `bior1.5` | Compact | 10 | 4 | `1.41421` |
| `bior2.2` | Compact | 6 | 5 | `1.41421` |
| `bior2.4` | Compact | 10 | 5 | `1.41421` |
| `bior2.6` | Large / Sensitive | 14 | 5 | `1.41421` |
| `bior2.8` | Large / Sensitive | 18 | 5 | `1.41421` |
| `bior3.1` | Compact | 4 | 5 | `1.125` |
| `bior3.3` | Compact | 8 | 6 | `2.12132` |
| `bior3.5` | Compact | 12 | 6 | `2.12132` |
| `bior3.7` | Large / Sensitive | 16 | 6 | `2.12132` |
| `bior3.9` | Large / Sensitive | 20 | 6 | `2.12132` |
| `bior4.4` | Compact | 10 | 7 | `1.58613` |
| `bior5.5` | Compact | 12 | 7 | `5.58579` |
| `bior6.8` | Large / Sensitive | 18 | 9 | `1.15131` |
| `coif1` | Compact | 6 | 7 | `91.8006` |
| `coif10` | Large / Sensitive | 60 | 33 | `2437.12` |
| `coif11` | Large / Sensitive | 66 | 37 | `16170.8` |
| `coif12` | Large / Sensitive | 72 | 39 | `2116.57` |
| `coif13` | Large / Sensitive | 78 | 43 | `23489.6` |
| `coif14` | Large / Sensitive | 84 | 45 | `3417.27` |
| `coif15` | Large / Sensitive | 90 | 49 | `36169.8` |
| `coif16` | Large / Sensitive | 96 | 51 | `4375.36` |
| `coif17` | Large / Sensitive | 102 | 55 | `34912.6` |
| `coif2` | Compact | 12 | 9 | `3.57823` |
| `coif3` | Medium | 18 | 13 | `42.7863` |
| `coif4` | Medium | 24 | 15 | `14.799` |
| `coif5` | Large / Sensitive | 30 | 19 | `283.731` |
| `coif6` | Large / Sensitive | 36 | 21 | `93.4704` |
| `coif7` | Large / Sensitive | 42 | 25 | `1724.88` |
| `coif8` | Large / Sensitive | 48 | 27 | `828.314` |
| `coif9` | Large / Sensitive | 54 | 31 | `306199` |
| `db1` | Compact | 2 | 5 | `1.41421` |
| `db10` | Medium | 20 | 13 | `21.6996` |
| `db11` | Medium | 22 | 15 | `1039.05` |
| `db12` | Medium | 24 | 15 | `43.1261` |
| `db13` | Large / Sensitive | 26 | 17 | `2596.18` |
| `db14` | Large / Sensitive | 28 | 17 | `85.6274` |
| `db15` | Large / Sensitive | 30 | 19 | `6283.63` |
| `db16` | Large / Sensitive | 32 | 19 | `170.158` |
| `db17` | Large / Sensitive | 34 | 21 | `15893.3` |
| `db18` | Large / Sensitive | 36 | 21 | `487.219` |
| `db19` | Large / Sensitive | 38 | 23 | `105101` |
| `db2` | Compact | 4 | 5 | `1.11536` |
| `db20` | Large / Sensitive | 40 | 23 | `609.252` |
| `db21` | Large / Sensitive | 42 | 25 | `2710.08` |
| `db22` | Large / Sensitive | 44 | 25 | `15693.6` |
| `db23` | Large / Sensitive | 46 | 27 | `1207.55` |
| `db24` | Large / Sensitive | 48 | 27 | `840.317` |
| `db25` | Large / Sensitive | 50 | 29 | `6606.6` |
| `db26` | Large / Sensitive | 52 | 29 | `55897.7` |
| `db27` | Large / Sensitive | 54 | 31 | `285776` |
| `db28` | Large / Sensitive | 56 | 31 | `13871.4` |
| `db29` | Large / Sensitive | 58 | 33 | `2.02289e+06` |
| `db3` | Compact | 6 | 7 | `14.932` |
| `db30` | Large / Sensitive | 60 | 33 | `132646` |
| `db31` | Large / Sensitive | 62 | 35 | `57852.8` |
| `db32` | Large / Sensitive | 64 | 35 | `2473.4` |
| `db33` | Large / Sensitive | 66 | 37 | `9.28142e+06` |
| `db34` | Large / Sensitive | 68 | 37 | `2.2496e+06` |
| `db35` | Large / Sensitive | 70 | 39 | `7.97156e+06` |
| `db36` | Large / Sensitive | 72 | 39 | `54462.6` |
| `db37` | Large / Sensitive | 74 | 41 | `2.55517e+07` |
| `db38` | Large / Sensitive | 76 | 41 | `170662` |
| `db4` | Compact | 8 | 7 | `2.63378` |
| `db5` | Compact | 10 | 9 | `46.0839` |
| `db6` | Compact | 12 | 9 | `5.42251` |
| `db7` | Large / Sensitive | 14 | 11 | `143.003` |
| `db8` | Large / Sensitive | 16 | 11 | `10.8886` |
| `db9` | Medium | 18 | 13 | `398.18` |
| `dmey` | Large / Sensitive | 62 | 33 | `28.028` |
| `haar` | Compact | 2 | 5 | `1.41421` |
| `rbio1.1` | Compact | 2 | 5 | `1.41421` |
| `rbio1.3` | Compact | 6 | 5 | `1.41421` |
| `rbio1.5` | Compact | 10 | 5 | `1.41421` |
| `rbio2.2` | Compact | 6 | 5 | `1.41421` |
| `rbio2.4` | Compact | 10 | 5 | `1.41421` |
| `rbio2.6` | Large / Sensitive | 14 | 5 | `1.41421` |
| `rbio2.8` | Large / Sensitive | 18 | 5 | `1.41421` |
| `rbio3.1` | Compact | 4 | 5 | `2.12132` |
| `rbio3.3` | Compact | 8 | 5 | `2.12132` |
| `rbio3.5` | Compact | 12 | 5 | `2.12132` |
| `rbio3.7` | Large / Sensitive | 16 | 5 | `2.12132` |
| `rbio3.9` | Large / Sensitive | 20 | 5 | `2.12132` |
| `rbio4.4` | Compact | 10 | 7 | `1.58613` |
| `rbio5.5` | Compact | 12 | 7 | `5.58579` |
| `rbio6.8` | Large / Sensitive | 18 | 9 | `1.15131` |
| `sym10` | Medium | 20 | 13 | `26.3969` |
| `sym11` | Medium | 22 | 15 | `94.636` |
| `sym12` | Medium | 24 | 15 | `32.8267` |
| `sym13` | Large / Sensitive | 26 | 17 | `4.89679e+06` |
| `sym14` | Large / Sensitive | 28 | 17 | `2592.45` |
| `sym15` | Large / Sensitive | 30 | 19 | `68.4772` |
| `sym16` | Large / Sensitive | 32 | 19 | `19.8683` |
| `sym17` | Large / Sensitive | 34 | 21 | `1070.37` |
| `sym18` | Large / Sensitive | 36 | 21 | `28.6658` |
| `sym19` | Large / Sensitive | 38 | 23 | `1849.14` |
| `sym2` | Compact | 4 | 5 | `1.11536` |
| `sym20` | Large / Sensitive | 40 | 23 | `16728.7` |
| `sym3` | Compact | 6 | 7 | `14.932` |
| `sym4` | Compact | 8 | 7 | `5.87693` |
| `sym5` | Compact | 10 | 9 | `3.32658` |
| `sym6` | Compact | 12 | 9 | `13.9887` |
| `sym7` | Large / Sensitive | 14 | 11 | `17127.2` |
| `sym8` | Large / Sensitive | 16 | 11 | `19.0019` |
| `sym9` | Medium | 18 | 13 | `309.77` |

ke1rro:
## PR Category

Feature

## Summary

This PR implements the TTNN lifting-wavelet bounty described in
[#40494](https://github.com/tenstorrent/tt-metal/issues/40494). It adds native
FP32 1D and separable 2D DWT/IDWT operations for Wormhole and Blackhole:

- `ttnn.dwt_coeff_len`
- `ttnn.dwt`
- `ttnn.idwt`
- `ttnn.dwt_2d`
- `ttnn.idwt_2d`

Closes #40494

## Notes for reviewers

Please focus on the public API and output contract, particularly the 1D
stick-native valid-prefix representation described below.

## API and usage

```python
ttnn.dwt_coeff_len(input_length, wavelet)

ttnn.dwt(
    input,
    wavelet,
    *,
    boundary_mode="symmetric",
    memory_config=None,
    output_tensors=None,
)

ttnn.idwt(
    approximation,
    detail,
    wavelet,
    original_length,
    *,
    boundary_mode="symmetric",
    memory_config=None,
    output_tensor=None,
)

ttnn.dwt_2d(
    input,
    wavelet,
    *,
    boundary_mode="symmetric",
    memory_config=None,
    output_tensors=None,
)

ttnn.idwt_2d(
    ll,
    lh,
    hl,
    hh,
    wavelet,
    output_shape,
    *,
    boundary_mode="symmetric",
    memory_config=None,
    output_tensor=None,
)
```

The examples below assume that `device` is an open single-device

`ttnn.MeshDevice`.

### 1D DWT and IDWT

```python
import torch
import ttnn

signal = torch.sin(torch.arange(65, dtype=torch.float32) * 0.1)
input_tensor = ttnn.from_torch(
    signal,
    dtype=ttnn.float32,
    layout=ttnn.ROW_MAJOR_LAYOUT,
    device=device,
    memory_config=ttnn.DRAM_MEMORY_CONFIG,
)

wavelet = "db4"
coefficient_count = ttnn.dwt_coeff_len(signal.shape[-1], wavelet)

approximation, detail = ttnn.dwt(
    input_tensor,
    wavelet,
    boundary_mode="symmetric",
)

# DWT output is consumed directly by IDWT; no repacking is required.
reconstructed = ttnn.idwt(
    approximation,
    detail,
    wavelet,
    signal.shape[-1],
    boundary_mode="symmetric",
)

approximation_values = ttnn.to_torch(approximation).flatten()[:coefficient_count]
reconstructed_values = ttnn.to_torch(reconstructed).flatten()[: signal.shape[-1]]
```

### 2D DWT and IDWT

```python
image = torch.randn((33, 35), dtype=torch.float32)
input_tensor_2d = ttnn.from_torch(
    image,
    dtype=ttnn.float32,
    layout=ttnn.TILE_LAYOUT,
    device=device,
    memory_config=ttnn.DRAM_MEMORY_CONFIG,
)

bands = ttnn.dwt_2d(
    input_tensor_2d,
    "bior3.9",
    boundary_mode="reflect",
)
ll, lh, hl, hh = bands

reconstructed_2d = ttnn.idwt_2d(
    ll,
    lh,
    hl,
    hh,
    "bior3.9",
    image.shape,
    boundary_mode="reflect",
)
```

The 2D band order is `(LL, LH, HL, HH)`. Exact-spec preallocated storage can
be supplied through `output_tensors` or `output_tensor`.

## Supported wavelets

The generated registry contains 106 discrete wavelet schemes grouped as:

- Haar: `haar`
- Daubechies: `db1` through `db38`
- Symlets: `sym2` through `sym20`
- Coiflets: `coif1` through `coif17`
- Biorthogonal: `bior1.1`, `bior1.3`, `bior1.5`, `bior2.2`, `bior2.4`,
  `bior2.6`, `bior2.8`, `bior3.1`, `bior3.3`, `bior3.5`, `bior3.7`,
  `bior3.9`, `bior4.4`, `bior5.5`, and `bior6.8`
- Reverse biorthogonal: `rbio1.1`, `rbio1.3`, `rbio1.5`, `rbio2.2`,
  `rbio2.4`, `rbio2.6`, `rbio2.8`, `rbio3.1`, `rbio3.3`, `rbio3.5`,
  `rbio3.7`, `rbio3.9`, `rbio4.4`, `rbio5.5`, and `rbio6.8`
- Discrete Meyer: `dmey`

## Boundary modes

The supported boundary-extension modes are:

- `zero`
- `constant`
- `symmetric` (default)
- `reflect`
- `periodic`
- `smooth`
- `antisymmetric`
- `antireflect`

## Input and output contract

- Operations support FP32 tensors on one physical device.
- 1D DWT inputs use ROW_MAJOR layout with shape `[W]` or `[B, 1, 1, W]`.
- 1D IDWT accepts equal-shaped coefficient tensors in either representation:
  - stick-native output produced directly by `ttnn.dwt`: `[ceil(C / 32), 32]`
    or batched `[B, 1, ceil(C / 32), 32]`;
  - externally supplied canonical coefficients: `[Wc]` or batched
    `[B, 1, 1, Wc]`.
- 2D operations use the standard 32x32 tile layout with shape `[H, W]` or
  `[B, 1, H, W]`.
- Inputs must be non-sharded INTERLEAVED tensors in DRAM or L1. Inverse inputs
  may use different supported placements while remaining on the same device.
- Outputs are currently non-sharded INTERLEAVED DRAM tensors. Preallocated
  outputs must match the inferred tensor specification and must not alias an
  input or another output.
- `original_length` and `output_shape` recover exact odd or even inverse output
  dimensions.

## 1D coefficient representation

The 1D path uses 32 FP32 values as one 128-byte interleaved ROW_MAJOR page. For
an input width `W`, let:

```text
C = ttnn.dwt_coeff_len(W, wavelet)
```

The physical DWT output capacity is:

```text
[ceil(C / 32), 32]
```

or, for batched input:

```text
[B, 1, ceil(C / 32), 32]
```

Only the first `C` flattened values per batch item are defined. Lanes after
that valid prefix in the final physical page are unspecified. IDWT consumes
this representation directly and similarly returns a stick-native tensor in
which the first `original_length` values are defined.

`ttnn.dwt_coeff_len` exposes the mathematical coefficient count separately
from the physical output capacity. This is a performance-motivated contract:
callers can identify the valid prefix without requiring `ttnn.dwt` to repack
every result into canonical `[C]` storage.

Canonical IDWT input is useful when coefficients originate outside
`ttnn.dwt`. It does not change the optimized direct path: output produced by
`ttnn.dwt` remains stick-native and is consumed by `ttnn.idwt` without
repacking.

The operation does not silently repack DWT output to canonical `[C]` storage.
Keeping the 128-byte stick representation preserves the existing memory path
without adding a copy, canonicalization kernel, or dispatch, and keeps DWT
output directly consumable by IDWT.

### Maintainer question

For the long-term public TTNN contract, which representation should the API
use? Canonical output may be preferable for API consistency, but approaches
that repack the result or change its physical page geometry may reduce
performance and should be evaluated against the current stick-native path.

1. Keep stick-native output with an explicit valid length.
2. Add an explicit canonicalization operation.
3. Add an output-format option to the wavelet API.
4. Extend TTNN so logical ROW_MAJOR width can differ from physical page width.

## SFPI kernels

This PR also adds dedicated FP32 SFPI kernel routines for the horizontal and
vertical lifting stencils and per-tile scaling. The stencil routines rotate
source lanes and accumulate coefficient-weighted values directly in SFPU
registers; scaling supports both full 32x32 tiles and the 32x16 narrow tiles
used by the 1D path. Horizontal lane-shift handling remains explicitly
specialized for Wormhole and Blackhole.

## Testing

Coverage includes:

- focused 1D and 2D forward/inverse comparisons;
- odd and non-aligned 1D inputs, odd non-square 2D inputs, and batching;
- supported interleaved DRAM/L1 input combinations;
- explicit runtime regressions for both unbatched and batched canonical 1D
  coefficient tensors supplied directly to IDWT;
- exact-spec preallocated outputs and alias validation;
- program-cache address rebinding and operation specialization;
- all 106 schemes across all eight boundary modes for 1D and 2D execution;
- physical Wormhole and Blackhole execution.

This section describes local and hardware validation; it does not claim that
the upstream PR's GitHub CI has already run.


ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
```diff
@@ -0,0 +1,589 @@
...
+
+struct SmoothIndexI32 {
+    uint32_t source_index{0};
+    uint32_t auxiliary_index{0};
+    uint32_t distance{0};
+    bool affine{false};
+};
+
+[[nodiscard]] WAVELET_EXTENSION_ALWI uint32_t
+extension_positive_mod_i32(const int32_t index, const uint32_t period) noexcept {
```
vtsilytskyiTT, line 46: These 3 functions seems to be trivial enough to become `constexpr`.

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
```diff
@@ -0,0 +1,589 @@
...
+        .remainder = static_cast<uint64_t>(remainder),
+    };
+}
+
+[[nodiscard]] WAVELET_EXTENSION_ALWI uint32_t
+extension_positive_mod(const int64_t index, const uint32_t length) noexcept {
+    return static_cast<uint32_t>(decompose_extension_period(index, length).remainder);
+}
+
+template <BoundaryMode Mode>
```
vtsilytskyiTT: Template Specialization is designed to prevent `else if constexpr (Mode == *)` chain. Try to keep sanity checks common and apply specialization for the rest of the logic..

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
```diff
@@ -0,0 +1,589 @@
...
+    const bool left = index < 0;
+    return SmoothIndexI32{
+        .source_index = left ? 0U : length - 1U,
+        .auxiliary_index = left ? 1U : length - 2U,
+        .distance = left ? 0U - static_cast<uint32_t>(index) : static_cast<uint32_t>(index) - (length - 1U),
+        .affine = true,
+    };
+}
+
+template <BoundaryMode Mode>
```
vtsilytskyiTT, line 320: Same here

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
vtsilytskyiTT, line 1: What is the compilation time for this header? I'd expect [Combinatorial explosion](https://en.wikipedia.org/wiki/Combinatorial_explosion) and a very slow compilation...

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
```diff
@@ -0,0 +1,148 @@
...
+    TiledShape2D band{};
+    bool padding_precedes_split{true};
+};
+
+[[nodiscard]] constexpr bool is_tile_aligned_2d(const Shape2D shape) noexcept {
+    return !shape.empty() && shape.height % kTileHeight2D == 0 && shape.width % kTileWidth2D == 0;
+}
+
+[[nodiscard]] inline size_t checked_shape_area_2d(const Shape2D shape, const char* label) {
+    TT_FATAL(!shape.empty(), "{} shape must be positive, got {}x{}", label, shape.height, shape.width);
```
vtsilytskyiTT:
```suggestion
    TT_FATAL(!shape.empty(), "{} shape can't be empty, got {}x{}", label, shape.height, shape.width);
```

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
```diff
@@ -0,0 +1,148 @@
...
+    friend constexpr bool operator==(const TiledShape2D&, const TiledShape2D&) = default;
+};
+
+struct Lwt2DTilingContract {
+    TiledShape2D input{};
+    TiledShape2D band{};
+    bool padding_precedes_split{true};
+};
+
+[[nodiscard]] constexpr bool is_tile_aligned_2d(const Shape2D shape) noexcept {
```
vtsilytskyiTT: Can be a structure member

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
```diff
@@ -0,0 +1,148 @@
...
+        shape.height <= std::numeric_limits<size_t>::max() / shape.width,
+        "{} shape {}x{} overflows size_t",
+        label,
+        shape.height,
+        shape.width);
+    return shape.height * shape.width;
+}
+
+[[nodiscard]] inline size_t round_up_to_tile_2d(const size_t value, const size_t tile_extent, const char* label) {
+    TT_FATAL(value > 0, "{} logical extent must be positive", label);
```
vtsilytskyiTT:
```suggestion
    TT_FATAL(value > 0, "{} logical extent must be positive, got {}", label, value);
```

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
```diff
@@ -0,0 +1,148 @@
...
+    TT_FATAL(
+        value <= std::numeric_limits<size_t>::max() - (tile_extent - 1),
+        "{} logical extent {} cannot be rounded to a {}-element tile",
+        label,
+        value,
+        tile_extent);
+    return ((value + tile_extent - 1) / tile_extent) * tile_extent;
+}
+
+[[nodiscard]] inline TiledShape2D make_tiled_shape_2d(const Shape2D logical) {
```
vtsilytskyiTT: This is basically a constructor, isn't it?

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,389 @@
...
+        source0_buffer.pop_front(1);
+        scale_tile(0, coefficient);
+        tile_regs_commit();
+        tile_regs_wait();
+        output_buffer.reserve_back(1);
+        pack_tile(0, cb_output);
+        output_buffer.push_back(1);
+        tile_regs_release();
+    }
+}
```
vtsilytskyiTT, line 156: This approach is slow. Try processing multiple DST tiles per acquire/commit cycle.
block_size depends on architecture and data type (fp32 vs bf16) and set to 8 just to show the idea.

```c++
constexpr uint32_t block_size = 8;
for (uint32_t base = 0; base < tile_count; base += block_size) {
    const uint32_t count = std::min(block_size, tile_count - base);
    source0_buffer.wait_front(count);
    tile_regs_acquire();
    copy_block(cb_source0, 0, 0, count);
    for (uint32_t i = 0; i < count; ++i) {
        scale_tile(i, coefficient);
    }
    source0_buffer.pop_front(count);
    tile_regs_commit();
```

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,389 @@
...
+        }
+        if (base_scale_bits != 0) {
+            scale_tile(base, base_scale_bits);
+        }
+    }
+};
+
+template <bool ScaleSource, bool ScaleBase, uint32_t SourceScaleBits, uint32_t BaseScaleBits>
+struct SpecializedScalePolicy {
+    inline void apply(const uint32_t source0, const uint32_t source1, const uint32_t base) const {
```
vtsilytskyiTT: Strange API. `apply` takes 3 arguments, but never uses all of them, which is clear on compilation step.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,389 @@
...
+    }
+}
+
+}  // namespace
+
+void kernel_main() {
+    constexpr uint32_t cb_source0 = get_compile_time_arg_val(0);
+    constexpr uint32_t cb_source1 = get_compile_time_arg_val(1);
+    constexpr uint32_t cb_base = get_compile_time_arg_val(2);
+    constexpr uint32_t cb_output = get_compile_time_arg_val(3);
```
vtsilytskyiTT, line 356: This is a new kernel. I suggest to focus on Metal 2.0 API. It is object-oriented and more robust. Check docs/source/tt-metalium/tt_metal/apis/kernel_apis/data_movement/device_api_migration_guide.md . For example for reading arguments you should use `get_arg`.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,424 @@
...
+#endif
+
+// copy_tile_init_short() still configures the unpack MOP from the 32x16 CB
+// metadata (two faces). Only the FP32 math-side Dst address needs overriding:
+// Metalium's generic direct-copy path currently hardcodes a 32x32 Dst stride.
+ALWI void copy_narrow_tile(const uint32_t cb, const uint32_t input_tile_index, const uint32_t dst_tile_index) {
+    UNPACK((llk_unpack_A<BroadcastType::NONE, false, EltwiseBinaryReuseDestType::NONE, UnpackToDestEn>(
+        cb, input_tile_index)));
+    MATH((copy_narrow_tile_math(dst_tile_index)));
+}
```
vtsilytskyiTT: This function belongs to LLK. With custom LLKs in kernels we'll get maintenance problems.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,424 @@
...
+        for (uint32_t tile = 0; tile < 3; ++tile) {
+            tile_regs_acquire();
+            copy_tile_to_dst_init_short(cb_input);
+            copy_narrow_tile(cb_input, tile, kDstScale);
+            scale_narrow_tile(kDstScale, scalar_packed);
+            tile_regs_commit();
+            tile_regs_wait();
+            pack_tile(kDstScale, cb_output, tile);
+            tile_regs_release();
+        }
```
vtsilytskyiTT: As above. Extract some steps out of the loop. You need to acquire once, init copy once, commit, wait, release also once. `3` should be a named const ~~and used as the last argument in `pack_tile`~~.
DST register can fit 4 FP32 tiles or 8 BF16 tiles.
UPD: pack_tile should be in a separate loop.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
```diff
@@ -0,0 +1,1468 @@
...
+    uint32_t x_length;
+};
+
+#if defined(LWT_2D_COMPACT_BOUNDARY_CODE) || defined(ILWT_2D_COMPACT_BOUNDARY_CODE)
+#define LWT_2D_BOUNDARY_FUNCTION __attribute__((noinline))
+#else
+#define LWT_2D_BOUNDARY_FUNCTION ALWI
+#endif
+
+[[nodiscard]] ALWI Rect load_rect(const uint32_t* words, const uint32_t offset) {
```
vtsilytskyiTT: 1. Conceptually it is a `Rect's` constructor. Could you make it as such?
2. These arguments are C-style. Modern C++ approach would be an `std::span` or something similar.

vtsilytskyiTT (changes requested):
Thanks for your contribution!

There are 2 things that I worry about:
1. Too many templates, especially in kernels. They cause 2 problems: long compilation time (it is still runtime, just on host) and hash misses (just a tiny change in template argument will cause the full recompilation, which is, again, slow due to templates). Try to find a balance between templates and runtime computation.
2. Upd: ~~Old Metal API. As this is a new code, please utilize Metal 2.0 API.~~ We still use old API for the new code, so no worries

Minor things - non-canonical OOP. Many functions belongs to their structs (constructors, area calculation, etc.).

Also, try to split PR to a series of smaller PRs. This will help in manual and automatic code review.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_program_factory.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+struct Ilwt2DInputs {
+    const Tensor& ll;
+    const Tensor& lh;
+    const Tensor& hl;
+    const Tensor& hh;
+    const std::optional<Tensor>& preallocated_output;
+};
+
+struct Ilwt2DDeviceOperation {
```
ayerofieiev-tt: lets please have each device operation in its own .hpp/.cpp

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_program_factory.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+struct Lwt2DInputs {
+    const Tensor& input;
+    const std::optional<std::array<Tensor, 4>>& preallocated_outputs;
+};
+
+struct Lwt2DDeviceOperation {
+    using operation_attributes_t = Lwt2DParams;
+    using tensor_args_t = Lwt2DInputs;
+    using spec_return_value_t = std::
```
ayerofieiev-tt: is this the order ll, lh, hl, hh?

ttnn/cpp/ttnn/operations/wavelet/generated/schemes/bior1_1.hpp
ayerofieiev-tt, line 1: trying to find code that was used to generate these schemes

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_program_factory.hpp
```diff
@@ -0,0 +1,106 @@
...
+    MemoryConfig output_memory_config;
+};
+
+struct Ilwt1DInputs {
+    const Tensor& approximation;
+    const Tensor& detail;
+    const std::optional<Tensor>& preallocated_output;
+};
+
+struct Ilwt1DDeviceOperation {
```
ayerofieiev-tt: we ask Program Factory and Device Operation to be separated into their own files

ttnn/cpp/ttnn/operations/wavelet/generated/schemes/bior1_1.hpp
ayerofieiev-tt: @wilderfield @afuller-TT should we instead commit source python script and jsons and let cmake generate these sources on the fly?

nsextonTT: I'd prefer that tbh

ke1rro: Another option could be to keep the generation code and the 106 scheme definitions in a separate repository and use it as a submodule here. Then CMake could generate the .hpp sources during the build, so we wouldn’t need to commit all 106 generated headers into tt-metal.

This might actually be better for users as well, since the source scheme definitions could keep the filter coefficients in a human-readable floating-point form. Right now the generated headers store them as hex values to avoid runtime fp32 -> uint32_t conversion, which makes the actual coefficients much harder to inspect.

afuller-TT: Is the generation deterministic?  Is it slow?

As an example: we commit `.proto` files as the source of truth.  We generate the corresponding headers and source files during the build.  Git is a "source" control.  Only the sources (not derived works) get tracked in Git.

If something is crazy slow, that might be a different matter.  If it's "generated" by an LLM, then a true (deterministic) source doesn't exist.

So the question is does it quack more like protobuf or like Claude?

draklowell: The generation algorithm is fully deterministic and reproducible, you can use scripts at https://github.com/draklowell/lifting-factorization to reproduce the json files. The process is very slow (more than 4 hours wall-time on consumer-grade CPU).

I think this might be a case of protobuf. In our original work we use json files, from which sources are generated, but we can probably just move to protobuf and commit `.proto` files of schemes + generation as a part of compilation. Will this be sufficient?

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
vtsilytskyiTT: Sorry, this is irrelevant. Keep using current API as you do now.

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
ke1rro: @vtsilytskyiTT Thanks for the detailed review. I took a look at this and measured it against some popular ttnn ops, rather than trying to estimate it from the number of templates. 

As `signal_extension.hpp` is a part of the reader kernel, I measure the actual NCRISC reader compilation that includes it. 

For example for the `symmetric`, `reflect` and `antireflect` specializations the compiler subprocess was around 0.58 s, with the linker stage 0.93-0.96 s. 

For the table below, `Cold API` is the synchronized first call which includes program creation/cache lookup, JIT compilation/linking if needed, dispatch, device execution, and `ttnn.synchronize_device()`.
`Compiler max` and `Linker max` are the longest individual compiler/linker subprocesses measured directly around the TT-Metal JIT commands; they are not derived from cold - warm.
`Warm API` is an immediate repeat in the same process with the program/JIT caches populated, so it still includes dispatch, actual device execution, and synchronization.

Here some table with separated compiler and linker time for a ttnn ops and representative wavelet schemes:

Operation | Cold API | Compiler max | Linker max | Warm API
-- | -- | -- | -- | --
ttnn.add | 0.79 s | 0.58 s | 0.29 s | 0.40 ms
ttnn.softmax | 1.54 s | 1.07 s | 0.45 s | 0.28 ms
ttnn.matmul | 1.07 s | 0.73 s | 0.34 s | 0.33 ms
ttnn.conv2d | 3.13 s | 0.74 s | 0.31 s | 1.08 ms
DWT db1 | 1.53 s | 0.68 s | 0.93 s | 0.45 ms
DWT bior3.9 | 3.11 s | 0.68 s | 2.43 s | 0.50 ms
DWT coif17 | 1.97 s | 0.77 s | 1.20 s | 3.05 ms

As we see for example 3.11 s first call for `bio3.9` is not 3 s of template compilation the most of the extra cost comes from the TRISC1 linker. 

About the [Combinatorial explosion](https://en.wikipedia.org/wiki/Combinatorial_explosion): I also checked the hash directly. A full `106 schemes (filters) * 8 boundary modes = 848` calls produced 124 logical kernel hashed: 

- 106 compute hashes - oner per scheme 
- 16 reader - 8 boundary mode * 2 finite dataflow policies 
- 2 writer hashes - one per dataflow policy 

So there isnt a `106 * 8 * reader/compute/writer` cross product in the current code. When the user selects the scheme on `ttnn.dwt(input, wavelet)`  the scheme name is resolved on the host and only that selected scheme is used to instantiate the compute path. The device compute kernel then includes only the generated header for that scheme, so calling `db1`, for example, does not compile all 106 wavelets. 

When another scheme is used for the first time, it normally adds a new compute kernel hash, while the existing reader and writer kernels can be reused. Boundary mode works independently: changing it adds a reader specialization without rebuilding the compute or writer kernels.

I did find one real cache sensitivity while checking this changing the 1D ROW_MAJOR width currently changes the reader hash because the input page size is part of its compile-time arguments. I'm looking at that separately, since that seems like a more concrete place where moving something out of the compile key may be useful.

Does this address the concern you had in mind, or were you referring to a different part of the compilation path?

jberkowitzTT:
@ke1rro - let me know once the comments above are resolved & you're ready for final review

ke1rro:
> @ke1rro - let me know once the comments above are resolved & you're ready for final review

Okay, I almost finished with all fixes but didn't push. The last thing need to figure out is the placement of all 106 wavelet schemes.

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
ke1rro: Implemented in [2f55c3fa95](https://github.com/tenstorrent/tt-metal/pull/52669/changes/2f55c3fa9558c2ec5c5498de3d02818d65acb748).

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
ke1rro: Implemented in 2f55c3fa95. Tile alignment is `Shape2D::is_tile_aligned()`.

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
ke1rro: Implemented in 2f55c3fa95. Replaced `make_tiled_shape_2d` with the narrow named factory `TiledShape2D::from_logical()`

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
ke1rro: Implemented in 62f7a193d5. API is now split into `scale_sources(source0, source1)` and `scale_base(base)` operation receives only the operands it uses.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
ke1rro: Implemented in 2a0c4e9742. The path now names `kScaleTileCount = 3`, acquires/commits/waits/releases once, processes all three 32x16 FP32 tiles, and packs each output tile separately. `Tile32x16` destination-capacity `static_assert` guards the mapping.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
ke1rro: Implemented in 62f7a193d5. I tested `std::span` first, but the actual TT-Metal device JIT command explicitly uses `riscv-tt-elf-g++ -std=c++17`.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_program_factory.hpp
ke1rro: Implemented in 1311ad0d3a

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_program_factory.hpp
ke1rro: made explicit in 1311ad0d3a: the forward 2D output tuple is ordered `(LL, LH, HL, HH)`.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_program_factory.hpp
ke1rro: Implemented in 1311ad0d3a.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
vtsilytskyiTT: ConfigWords is a better abstraction, thanks. Due to span, C++17 indeed doesn't have anything similar, so let it be a pointer+offset.

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
vtsilytskyiTT: I see it is addressed in https://github.com/tenstorrent/tt-metal/pull/52669/changes/2f55c3fa9558c2ec5c5498de3d02818d65acb748 , thanks

ttnn/cpp/ttnn/operations/wavelet/common/tiling_2d.hpp
vtsilytskyiTT: Agree, resolved

ttnn/cpp/ttnn/operations/wavelet/device/kernels/primitives/interleave.hpp
```diff
@@ -0,0 +1,208 @@
...
+            output_tiles = output_buffer.get_read_ptr();
+        }
+
+        const uint32_t group_signal_offset = group * signal_group_elements;
+        const uint32_t group_output_length = output_length - group_signal_offset < signal_group_elements
+                                                 ? output_length - group_signal_offset
+                                                 : signal_group_elements;
+        const uint32_t output_end = output_begin + output_length;
+        const uint32_t first_stick = output_begin / ttnn::operations::wavelet::kStickWidth;
+        const uint32_t stick_count =
```
ayerofieiev-tt: **Bug (confirmed): multi-group direct interleave overwrites earlier groups' output with stale workspace data.**

`write_direct_interleaved_signal` computes `group_output_length` (line 135) but never uses it: the stick loop runs over `stick_count` derived from the full `output_length`, so every group iteration rewrites the entire chunk range. Updated-parity lanes outside the current group's `cb_output` window fall to the workspace fallback, but on the direct-interleave path the final lifting step is never written back to workspace (`lwt_writer.cpp:341-359` replaces `write_local_output_groups`), so the last group's full-range rewrite replaces all earlier groups' updated-parity output with pre-final-step values.

Reachable on Blackhole tile-native (`policy.hpp` sets `final_interleave_direct = true`) whenever a chunk's output length exceeds `kIlwtGroupOutputElements` (3072): `output_group_count > 1` and the reconstruction is silently wrong. The unused `group_output_length` looks like the intended loop bound - restricting each group's stick range to its own span should fix it.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1490 @@
...
+            std::get<0>(*tensor_args.preallocated_outputs),
+            std::get<0>(expected_specs),
+            tensor_args.input.device(),
+            "DWT approximation output");
+        validate_preallocated_output(
+            std::get<1>(*tensor_args.preallocated_outputs),
+            std::get<1>(expected_specs),
+            tensor_args.input.device(),
+            "DWT detail output");
+        TT_FATAL(
```
ayerofieiev-tt: **Bug (confirmed): 1D forward validate never checks that preallocated outputs don't alias the input.**

This guard only compares the two outputs against each other. The 2D forward path has the input check (`wavelet_2d_operation_impl.cpp:996` "2D DWT outputs must not alias the input") and both inverse paths check it too, so the omission here looks unintentional.

Exploitable from Python: `dwt(x, 'haar', output_tensors=(x, d))` where `x` is a `[1,32]` FLOAT32 tensor recycled from a prior wavelet output - its spec exactly equals what `output_spec_1d` computes for length 32 (coefficient length 16 -> 1 stick), so the spec-equality check passes and the reader streams the input while the writer overwrites the same DRAM buffer. Silent corruption; a one-line `TT_FATAL` against `tensor_args.input.buffer()` closes it.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1490 @@
...
+        };
+    }
+    TT_FATAL(
+        shape.rank() == 4,
+        "{} must have shape [S,32], [W], [B,1,S,32], or [B,1,1,W], got rank {}",
+        tensor_name,
+        shape.rank());
+    TT_FATAL(shape[0] > 0, "{} batch dimension must be positive", tensor_name);
+    TT_FATAL(shape[1] == 1, "{} requires C == 1, got {}", tensor_name, shape[1]);
+    if (shape[3] == kStickWidth && shape[2] > 1) {
```
ayerofieiev-tt, line 141: **API hazard (confirmed): stick-native inputs silently treat padding lanes as signal - multi-level DWT composition produces silently wrong results.**

`logical_1d_shape` accepts `[S,32]` / `[B,1,H,32]` and interprets them as one signal of length `S*32` per batch item, including the final-stick lanes the PR's own docstring declares "unspecified" (`wavelet_nanobind.cpp:44-45`).

Two consequences:
1. The standard composition `a1, d1 = dwt(x, 'db2')` with `len(x)=100` gives `a1` as `[2,32]` with only 51 valid lanes; `dwt(a1, 'db2')` passes all validation and transforms length 64 including 13 garbage lanes -> silently wrong level-2 coefficients. There is no `valid_length` parameter to exclude the padding.
2. A user's `[64,1,4,32]` batch of length-32 signals is silently reinterpreted as 64 signals of length 128 (only the `W==32 && H>1` combination hits this branch; `W==31` errors cleanly).

Either reject stick-native inputs that weren't documented (`docstring says [W] or [B,1,1,W]`), or add an explicit length parameter so multi-level use is possible correctly.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1490 @@
...
+        "LWT padded input length exceeds the device signed-index range");
+
+    const uint32_t max_cores = core_limit(mesh_device);
+    const ArchitecturePolicy architecture_policy = make_architecture_policy(mesh_device.arch());
+    const std::optional<WorkspaceLayout> workspace_override = workspace_layout_override();
+    const WorkspaceLayout initial_layout = workspace_override.value_or(WorkspaceLayout::kRowMajor);
+    const bool initial_hybrid_tile_mirror =
+        supports_hybrid_tile_mirror(architecture_policy.architecture, initial_layout);
+    const uint32_t signal_budget_bytes =
+        planner_signal_budget_bytes(mesh_device, architecture_policy, initial_hybrid_tile_mirror, 1U);
```
ayerofieiev-tt: **Robustness (confirmed): execution plan depends on live L1 occupancy, but L1 state is not in the program cache key.**

`planner_signal_budget_bytes` (and the 2D equivalent at `wavelet_2d_operation_impl.cpp:689-720`) queries `lowest_occupied_compute_l1_address()`, and the resulting chunk structure / CB sizes are baked into the cached program. No `compute_program_hash` is defined, so the default hash (attributes + tensor args) omits L1 state.

Failure mode: run the op with L1 empty (caches a plan with ~768KB workspace CBs), allocate a large persistent L1 tensor, run the same spec again -> cache hit replays the oversized CB layout and enqueue's `validate_circular_buffer_region` throws "circular buffers clash with L1 buffers", even though a fresh re-plan would have shrunk the CBs and succeeded. Conversely the inverse ops re-plan inside validate on every cache hit and can TT_FATAL against a cached program that still fits. Either include the effective L1 budget in a `compute_program_hash`, or quantize the budget so plans are L1-state-independent.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1490 @@
...
+void validate_preallocated_output(
+    const Tensor& output,
+    const tt::tt_metal::TensorSpec& expected_spec,
+    const tt::tt_metal::distributed::MeshDevice* expected_device,
+    const char* output_name) {
+    validate_1d_tensor(output, output_name);
+    validate_output_memory_config(output.memory_config(), output_name);
+    TT_FATAL(output.device() == expected_device, "{} must be on the same device as the inputs", output_name);
+    TT_FATAL(
+        output.tensor_spec() == expected_spec,
```
ayerofieiev-tt: **Usability bug (confirmed): the preallocated-output spec cannot be constructed through the Python API.**

`output_spec_1d` builds the expected spec with a non-default `Alignment{kStorageAlignmentBytes}`, and `TensorSpec::operator==` is member-wise, so `Alignment({32}) != Alignment({1})` even though the physical 128-byte-stick layout is byte-identical. Ordinary creation (`ttnn.from_torch`, row-major default) yields the default alignment, so every user-preallocated output fails this check with "tensor spec does not match the wavelet output specification" and no hint about the hidden alignment requirement.

The PR's own tests never construct a preallocated output from scratch - they all recycle prior op outputs (`test_wavelet.py:286-292, 605-614`), which is the only working path. Consider comparing the properties that actually matter (dtype/layout/logical shape/memory config/physical page size) instead of full spec equality, or documenting an API that produces a conforming tensor.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_operation_impl.cpp
```diff
@@ -0,0 +1,1152 @@
...
+        physical_elements % batch_count == 0, "{} physical volume is not divisible by its batch count", tensor_name);
+    const uint64_t elements_per_batch = physical_elements / batch_count;
+    TT_FATAL(
+        elements_per_batch % (kTileHeight2D * kTileWidth2D) == 0,
+        "{} physical batch stride is not tile aligned",
+        tensor_name);
+    return checked_u32(elements_per_batch / (kTileHeight2D * kTileWidth2D), "2D wavelet tiles per batch item");
+}
+
+[[nodiscard]] uint32_t static_workspace_address_2d(
```
ayerofieiev-tt: **Latent hazard (confirmed): hand-replayed CB allocator layout with no assert tying it to the framework-assigned address.**

`static_workspace_address_2d` re-simulates CB placement (base allocator address + the 9 CBs in declaration order, rounded with the DRAM alignment) and the resulting raw plane addresses are passed to kernels as runtime args. It is currently correct - `allocate_circular_buffers` places locally-allocated CBs in declaration order with exactly that alignment - but any CB reorder/size change in `create_program_descriptor` (lines 452-461), or an allocator packing change, silently shifts the real workspace while kernels keep the stale mirrored addresses: reads/writes land in neighboring CBs with no assert firing.

The reader already derives other addresses kernel-side via `CircularBuffer(cb).get_write_ptr()`, so the plane base could come from the workspace CB directly; failing that, a single `TT_FATAL` comparing the mirrored address to the framework-assigned address of `kLwt2DWorkspaceCb` converts silent corruption into a clean failure.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+template <typename Scheme>
+void lwt_compute() {
+    constexpr uint32_t cb_input0 = get_compile_time_arg_val(0);
+    constexpr uint32_t cb_input1 = get_compile_time_arg_val(1);
+    constexpr uint32_t cb_base = get_compile_time_arg_val(2);
+    constexpr uint32_t cb_output = get_compile_time_arg_val(3);
+    static_assert(
+        kInlineTerminalScale != kInlineInverseScale, "Exactly one scale policy must be selected for LWT or ILWT");
+    constexpr uint32_t route_count =
```
ayerofieiev-tt, line 391: **Latent hazard (confirmed): host and compute kernel derive the per-chunk runtime-arg stride from two independent sources with no cross-check.**

The kernel computes `route_count` from the compile-time scheme (`executable_step_count<Scheme>() - 1/-2`; the 2D kernel uses `4 * Scheme::num_steps` at `lwt_2d_compute.cpp:366-367`), while the host strides runtime args by the planner's `plan.chunks.front().routes.size()`. The planner's skip rules (drop swaps, inline terminal scale forward / two leading scales inverse) exactly mirror the kernel's constexpr skips today, but the two rule sets live in different files with no shared constant, and grep shows zero references to `executable_step_count`/`num_steps` in either host impl file.

Any asymmetric change shifts every chunk>0 arg slot by one: compute runs steps with a neighboring route's group count and deadlocks against the reader with no diagnostic. A host-side `TT_FATAL(plan.chunks.front().routes.size() == <scheme-derived count>)` at arg-packing time makes divergence a clean error.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_reader.cpp
```diff
@@ -0,0 +1,836 @@
...
+                {});
+        }
+    }
+}
+
+template <typename ConfigAccessor>
+ALWI const uint32_t* load_config_page(
+    const ConfigAccessor& config, const uint32_t config_addr, const uint32_t cb_config, const uint32_t page_index) {
+    const auto page_accessor =
+        TensorAccessor(config, config_addr, ttnn::operations::wavelet::device_protocol::kRouteConfigPageBytes);
```
ayerofieiev-tt, line 217: **Latent hazard (confirmed): chunk-config pages are read with the route-config page size; the two constants are only coincidentally equal.**

`load_config_page` hardcodes `kRouteConfigPageBytes` for its accessor but is also called with `chunk_config_addr` (line ~720; same pattern in `lwt_writer.cpp:36/316`), while the host allocates and writes chunk config with `kLwtChunkConfigPageBytes`. Both are 16 words today, but they are defined independently in `lwt_config.hpp` (lines 26-27, 53-54) with no `static_assert` tying them, and the ILWT chunk record already uses 14 of the 16 words.

The first person who adds a 17th chunk-config word bumps the host page to 128B while kernels keep striding `page_id` in 64B units - every chunk with `global_chunk > 0` reads the wrong half of a config page, yielding garbage stream bounds and out-of-bounds L1 writes. A one-line `static_assert(kLwtChunkConfigPageBytes == kRouteConfigPageBytes)` (or passing the page size in) closes it.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_operation_impl.cpp
```diff
@@ -0,0 +1,1152 @@
...
+        band_shape.height == expected_height && band_shape.width == expected_width,
+        "2D IDWT band shape {}x{} does not match expected shape {}x{} for output {}x{}",
+        band_shape.height,
+        band_shape.width,
+        expected_height,
+        expected_width,
+        operation_attributes.output_height,
+        operation_attributes.output_width);
+
+    dispatch_scheme(operation_attributes.scheme_id, [&]<typename Scheme>() {
```
ayerofieiev-tt: **Efficiency (confirmed): inverse validate builds and discards the full execution plan on every invocation, including program-cache hits; cache misses then plan twice.**

`validate_inverse_inputs_2d` runs the complete templated `make_inverse_plan_2d` (O(tiles_y x tiles_x) candidate search, heap-allocated chunk/route vectors, two allocator queries) purely for its internal shape checks, and both ILWT device operations wire `validate_on_program_cache_hit` to this same function (`ilwt_2d_device_operation.cpp:16-19`, same in 1D). On a miss, `build_inverse_workload_2d` rebuilds the identical plan again.

Every cached IDWT launch pays full host replanning, defeating the cache fast path. The shape guarantee validate actually needs is O(1) arithmetic via `scheme_info().tap_size` / coefficient-length math (the forward validates already do exactly that); alternatively compute the plan once and hand it to workload creation.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
```diff
@@ -0,0 +1,1469 @@
...
+            if (exact_full_tile) {
+                const uint32_t source_tile = (static_cast<uint32_t>(full_canonical_y) / kTileSide) * band_tile_columns +
+                                             static_cast<uint32_t>(full_canonical_x) / kTileSide;
+                noc.async_read(
+                    band,
+                    CoreLocalMem<uint32_t>(destination_addr),
+                    kTileBytes,
+                    {.page_id = band_tile_base + source_tile},
+                    {});
+                noc.async_read_barrier();
```
ayerofieiev-tt: **Efficiency (confirmed): per-tile read barrier serializes every DRAM fetch in the dominant inverse band-init path.**

In `initialize_inverse_band_plane`'s `exact_full_tile` fast path, each band tile gets one `noc.async_read` immediately followed by `noc.async_read_barrier()` inside the tile loop, so the interior (common) case runs at single-transaction NoC latency instead of pipelined bandwidth. Each read targets a distinct workspace `destination_addr`, so nothing forces the serialization - unlike the boundary-gather path below, which legitimately barriers per iteration over shared scratch.

`stage_split_source_tiles` in this same file (lines 202-230) already uses the batched pattern: issue all reads, one trailing barrier. Applying it here is a small, contained change.

ayerofieiev-tt (commented):
Automated deep review (8 finder angles, per-candidate verification pass; 6 candidates were refuted against planner/host guarantees before posting). 10 findings below as inline comments, ranked roughly by severity: 2 confirmed correctness bugs (stale-data overwrite in the ILWT direct-interleave path; missing input/output aliasing check in 1D forward validate), 3 API/robustness issues (stick-padding lanes silently treated as signal, program-cache key omitting L1 state, preallocated-output spec unattainable via the Python API), 3 latent cross-file contract hazards, and 2 host/device efficiency issues.

Verified clean along the way, for the record: CB push/pop pairing across all six kernels, NOC barriers before CB pushes, host/kernel runtime-arg order for all four reader/writer kernels, all 8 boundary-extension modes differentially fuzzed against PyWavelets semantics (zero mismatches), and planner sweeps across 5 schemes x 8 modes x n=1..500 forward/inverse (zero failures).

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
ke1rro: Done in [1199f2a](https://github.com/tenstorrent/tt-metal/pull/52669/commits/1199f2ad0c77712f63d94c1c6723b8e582be5c6c)

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
ke1rro: Done in [1199f2a](https://github.com/tenstorrent/tt-metal/pull/52669/commits/1199f2ad0c77712f63d94c1c6723b8e582be5c6c)

ttnn/cpp/ttnn/operations/wavelet/device/kernels/primitives/interleave.hpp
ke1rro: fixed in 12e988f. `write_direct_interleaved_signal` now check first and last stick and lane valid interval from the current groups `[group_begin, group_end)` range, so a later group cannot rewrite earlier output with stale workspace values

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: fixed in d5cc8bd. 1d forward validator now rejects either preallocated output when its buffer aliases the input

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: fixed in d5cc8bd. DWT use a separate signal shape and accepts only what is documented `[W]` and `[B, 1, 1, W]` inputs. In IDWT uses existing canonical and stick native coefficients forms because `original_length` provides the valid prefix contract.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: Fixed in 6fa2383

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: Fixed in d5cc8bd

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_operation_impl.cpp
ke1rro: Fixed in fa33f3c. Host pass workspace plane byte offsets instead of replaying circular buffer allocation to construct absolute L1 addresses

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
ke1rro: Fixed in a4a2e9cc75

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_reader.cpp
ke1rro: Fixed in a4a2e9cc75. `static_assert` now requires `kLwtChunkConfigPageBytes == kRouteConfigPageBytes`.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_operation_impl.cpp
ke1rro: Fixed in d5cc8bdf33

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
ke1rro: Fixed in bb79cefed0. Full tile reads now remain in flight and are drained by one barrier at plane completion

ttnn/cpp/ttnn/operations/wavelet/generated/schemes/bior1_1.hpp
ke1rro: @ayerofieiev-tt @nsextonTT @afuller-TT Thanks for the feedback. I made PR with suggested fix: https://github.com/ke1rro/tt-metal/pull/2

This PR no longer tracks the 106 generated scheme headers or the generated registry.  Change now tracks only:

 - JSON catalog wtih all 106 lifting schemes
 - deterministic Python generator
 - CMake integration that generates the 106 headers and `registry.hpp` as part of build.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
ke1rro: @vtsilytskyiTT 
I intentionally chose the 32x16 tile representation because the tt-metal program descriptor accepts explicit tile dimensions and the lower LLK stack already defines and tests the 32x16 tensor shape, including compact Tile32x16 destination addressing (`tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/ckernel_defs.h`). What I missed that support is not exposed end to end through the public `copy_tile()` path. This 32x16 provided a real performance boost for 1D DWT. The idea of using a narrow tile came from the current lifting-step dataflow, which keeps four source tiles and three base tiles live in DST at the same time. Seven 32x16 FP32 tiles occupy the destination space of three and a half full 32x32 tiles, so they fit within the four full-tile FP32 slots available in one acquired DST section. This allows the predict/update step to load all seven inputs, execute the stencil, and commit the result within one acquire/commit cycle. Preserving the same dataflow with full 32x32 DST slots would not fit it would require restructuring the step around additional DST cycles or staging, with a possible performance cost.

To summarize the problem: 

The CB and unpacker correctly handle the 32x16 geometry, and the lower LLK already provides `set_dst_write_addr<DstTileShape::Tile32x16>()`. However, the math side of public `copy_tile()` selects `Tile32x32` internally and does not expose the destination shape. I directly replaced my function with `copy_tile()`, compiled and ran, but it produced incorrect coefficients because the tiles were placed using a full tile DST stride. Thats why I created this bridge as workaround. 

I thought about fixing this, but doing so in the shared API would mean adding a compile-time shape parameter for example, to `copy_tile()`, with `Tile32x32` as default for existing callers. `copy_tile()` would then pass shape into the math side unary-datacopy path, where it would be used both for the DST capacity check and for selecting the destination stride. The Wormhole and Blackhole wrappers would need to propagate the param to their architecture specific implementations. 

I can work on this fix, but I would like to hear what you recommend before going further. There may be a better approach or an existing solution that I missed. I’ll wait for your feedback. Thank you.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
ke1rro: I tried this approach, but the naive two tile version stalled on batched inputs.

Here is some of my finds: 

The actual problem was CB wrapping. `wait_front(2)` / `reserve_back(2)` may
succeed even when two available pages straddle the physical end of the
ring, while `copy_block()` and pack loop use linear pointer arithmetic
and do not wrap in the middle of the block.
I got it working by calculating the contiguous page count from the UNPACK
read pointer and the PACK write pointer. Because those pointers are owned by
different kernel threads, I had to send both counts through the ckernel
mailboxes. The block size then became the minimum of two, the remaining
tiles, and the contiguous space on both sides, falling back to one tile at a
wrap boundary.
That version fixed the stall and passed the large batched case plus all focused tests. However, it did not improve performance the median change
across DWT2D/IDWT2D cases was +0.001%. mailbox and wrap handling logic adds complexity without gain, so I don't keep this change.

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
vtsilytskyiTT: Yes, it does, thank you! I think numbers are acceptable.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
vtsilytskyiTT: Kernel stuck when your symmetric invocations wait_front/pop_front are not balanced. One kernel waits for more data than another kernel provides. Batch processing may be tricky and there are 2 approaches to that:

1. If your op is resistant to zeroes (like elementwise add/sub, accumulation) - just pad your tensor with zeroes to align with batch size.
2. If not, you may keep current approach, or hybrid (function for batch-sized data and another function to process rest tile by tile).

Anyway, I do not insist. Treat it as a room for improvement.

vtsilytskyiTT:
@ke1rro thanks for the updates! Is it possible to split this giant change into series of PRs? This allows automatic code review by copilot + facilitates manual code review. I didn't reach the end of this PR to be honest.
Something like this should work: https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
vtsilytskyiTT: I don't mind keeping this function. The problem is the location. That is definitely an LLK code and should be placed there. You are mixing abstraction layers.

ke1rro:
> @ke1rro thanks for the updates! Is it possible to split this giant change into series of PRs? This allows automatic code review by copilot + facilitates manual code review. I didn't reach the end of this PR to be honest. Something like this should work: https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/

It looks like the native feature of stacked PRs aren't supported for forks, but I'll split this into a few smaller PRs instead.

ke1rro:
> @ke1rro thanks for the updates! Is it possible to split this giant change into series of PRs? This allows automatic code review by copilot + facilitates manual code review. I didn't reach the end of this PR to be honest. Something like this should work: https://github.blog/changelog/2026-07-30-stacked-pull-requests-are-now-in-public-preview/

I split the implementation into two dependent PRs:

PR1: https://github.com/tenstorrent/tt-metal/pull/54038
PR2: https://github.com/tenstorrent/tt-metal/pull/54039

I don't think splitting it further would help much because there are already a lot of dependencies between the remaining parts.

PR2 still shows ~15k-line diff against `main` because it depends on PR1.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
ke1rro: I moved the 32x16 copy helper out of the kernel into the shared LLK/internal compute layer.

For now `copy_tile_to_dst_32x16()` lives under `internal/compute`, while the architecture specific DST addressing is implemented in the Wormhole and Blackhole LLK math wrappers. I kept it internal because Wavelet is currently the only caller. Would you prefer to keep this as an internal interface for now, or expose it through the public compute API in this PR like `copy_tile()`?

jberkowitzTT:
Hi @ke1rro - let me know if there's anything I can do to help here & let me know once it's ready for review

Thanks!

ke1rro:
> Hi @ke1rro - let me know if there's anything I can do to help here & let me know once it's ready for review
> 
> Thanks!

Hi @jberkowitzTT, thanks! As I mentioned earlier, I split #52669 into two smaller PRs. The first part has already gone through review, and the second one is #54039.

From my side, #54039 is fully ready for review and has been for a while. It seems the process has stalled a bit there, so if you could help connect it with the right reviewer or help move the review forward, I’d really appreciate it.

Thanks!

Issue #38306, opened 2026-02-23
Matmul operation on Wormhole produces some incorrect outputs with HiFi4 and fp32 accumulation


sosborne-TT:
### Component / Area

ML operations, possibly LLK or hardware

### Issue Type (optional)

Bad Outputs

### Observed

Observations from @nmauriceTT : 
- The bug happens most frequently when combining `HiFi4` and `fp32_accumulation` with bfloat16 data.
- It's only a few data points (1 bad data point for a full [4096, 4096] @ [4096, 4096] matmul). In these cases, the accuracy is fine everywhere other than these data points. 
- It is deterministic: it fails on different boards, and at the same location and with the same error. Other people have reproduced it.
- The outputs are always off by roughly powers of two: I've seen absolute errors of ~0.5, 1, 2 and ~64 (bit flip ?). The difference is too large to be explained by rounding errors.
- So far, I've only seen the bug on Wormhole. Blackhole numerics look different from Wormhole, but math fidelity results remain consistent. ([@Izajasz Wrosz](https://tenstorrent.enterprise.slack.com/team/U08UWPPSYAH) told me that FPU differed slightly between WH and BH, which could explain it)

### Expected

Worst-case error for HiFi4 should be much lower.

### 1. Steps (exact commands)

Script from @nmauriceTT to reproduce the matmul error: [test_matmul_dot_product.py](https://github.com/user-attachments/files/25475816/test_matmul_dot_product.py)

> In this case, I get an outlier at [1531, 2136]  ref=-11.0259  tt=-74.5000  err=63.4741 for HiFi4 + fp32_acc (absolute error of ~63). In comparison, all absolute errors for HiFi3 + fp32_acc are below 1.7; for HiFi4 without fp32_acc it's below 30.

### 2. Input data / link or description

Worst-case error for Conv3D (which uses matmul) on Wormhole for different math fidelities. We see similar errors for the matmul operation.

<img width="4130" height="1758" alt="Image" src="https://github.com/user-attachments/assets/2a2fb850-f3a2-4938-97c5-bba5a4468d5d" />

### 3. Frequency

Consistent and deterministic.

### 1. Software Versions

Latest tt-metal main branch.

### 2. Hardware Details

So far it has only been seen on Wormhole devices (single and multi-card).

### Is this a regression?

Unknown

### Priority

None

sosborne-TT:
cc @bbradelTT @rmillerTT

PR #52676, opened 2026-08-09
[Feature] Kimi K3 AttnRes on Blackhole: the read in one fused dispatch

nmilicevicTT:
### PR Category

Feature.

### Summary

Brings up Kimi K3's attention residuals (AttnRes) on Blackhole: a new module under
`models/demos/deepseek_v3_d_p/`, and the fused TTNN op it is built around —
`ttnn.experimental.attn_res_gather_merge`, filed under `ttnn::experimental::ccl`.

Layout: torch references in `reference/attn_res/`, the ttnn composite in `tt/attn_res/`,
tests in `tests/attn_res/` split `model/` vs `perf/`. The op's suites live in the ttnn trees.

**Background for reviewers:** a shared workbook covering the math, the shape walk, the
roofline, and the full optimization series —
https://claude.ai/code/artifact/e3ff58d2-a56a-40e7-8c0b-47ba68026527

AttnRes reads back a shelf of frozen residual snapshots. Per read site it scores each
candidate against a folded query, softmaxes, and mixes. The sealed half of a block runs
once for all 24 read sites (`inter_block`); only the live half runs per site (`merge`).
The hidden axis is TP-sharded, so the live score's statistics have to cross a collective
before they can be scaled.

#### The op

`attn_res_gather_merge` is one read site's whole path — from a TP-sharded residual stream
to the mixed hidden state — in **one dispatch**. It replaces a four-program chain:
statistics, the cross-rank collective, the score chain, and the online-softmax fold.

Each rank holds `d / ring_size` of every token, so the live score needs a dot and a sum of
squares summed across the ring before it can be scaled:

```
live_scores = sum_p dots_p * rsqrt(sum_p sum_squares_p * inv_hidden_size + eps)
m = max(shift, live_scores)
h = (partial * exp(shift - m) + prefix_sum * exp(live_scores - m))
    / (mass * exp(shift - m) + exp(live_scores - m))
```

Two design points worth a reviewer's attention:

- **It gathers rather than reduces.** The wire carries each rank's raw statistics, not
  their sum. Two scalars per token cross either way, and summing them on arrival is a pair
  of dest-register adds inside a pass that already holds those tiles — reducing on the wire
  would cost a second device program, which is the thing being removed.
- **The residual add is inside.** Given `pending`, the op scores and folds
  `prefix_sum + pending` and hands the sum back alongside `h`. That deletes a whole
  `ttnn.add` dispatch, and is *also* cheaper on device: 98.8 us fused against
  85.4 + 18.1 = 103.5 us separate.

One `Program`, four kernels (reader, writer, compute, gather — 874 kernel lines), one
`EnqueueProgram`, one row in the profiler CSV.

The block's sealed half still composes: `rms_norm_pre_all_gather` + `matmul` +
`ttnn.all_reduce` + `fast_weighted_reduce_nc` (a dim-1 reduce that MACs its weight straight
into the accumulator instead of materializing the product, batched over R weight sets
against one shared input).

#### Perf

Blackhole LoudBox, `(2, 4)` mesh, 640 rows per chip, `d/tp = 1792`, TP factor 4.

The governing number is the **untraced** walk — production does not capture a trace, so
host dispatch is what AttnRes actually costs a forward. 93 layers, 186 reads, S ramping
1 -> 8:

| | untraced | traced |
|---|---|---|
| four-op composed chain | 194.72 ms | 27.54 ms |
| **fused** | **105.71 ms** | **26.54 ms** |

**1.84x on the untraced walk.** Enqueue is 105.60 ms of the 105.71 ms total, so the walk is
host-bound end to end and the win is dispatch count, exactly as intended. Traced barely
moves (1.04x), which is the point: this is not an arithmetic optimization.

Host-side dispatch for a single read: 366.7 us for the three-op path, 205.3 us fused —
**1.79x**.

Isolated device time per read site, traced, from a profile containing only this op:

| S | us |
|---|---|
| 1 | 99.7 |
| 4 | 110.6 |
| 8 | 130.5 |

`kFoldCoresPerRow = 2` (41 fold+gather cores) is measured-optimal, and the curve is not
monotonic in core count:

| cores | 21 | **41** | 61 | 81 | 101 |
|---|---|---|---|---|---|
| us | 100.8 | **98.6** | 113.4 | 111.4 | 119.5 |

`split_work_to_cores` splits total output tiles, so the per-core share is even only when
the per-row core count divides `Wt = 56`. 61 and 101 cores do not, and their ragged groups
straddle row boundaries and re-fetch that row's scalars. Widening the fold does not help —
it is not core-starved at 41.

#### Correctness

- `tests/ttnn/unit_tests/operations/ccl/test_attn_res_gather_merge.py` — the op against a
  torch reference at 640 and 128 tokens per chip, `plain` and `settle` (fused-add) arms.
  Gate 0.9999. The settle arm also asserts `torch.equal` between the op's carried-forward
  sum and a separate `ttnn.add`, so the fused add is bit-identical, not merely PCC-equal.
- `tests/attn_res/model/test_attn_res.py` — the composite read at S=1 and S=8, PCC and worst
  relative error. S=8 over 24 sites at 1280 tokens: worst PCC **0.9999825**, worst rel err
  1.180e-02. Plus `test_sequence_axis_communicates_nothing`, which catches a collective
  leaking across the sequence axis (`max|delta| = 0.000e+00`) — PCC against torch cannot see
  that class of bug.
- `tests/attn_res/perf/test_attn_res_e2e.py::test_walk_matches_reference` — the device walk
  against the torch reference walk over all 93 layers / 186 reads: PCC **0.9999897**. This is
  the only thing in the tree that exercises the seal cadence and per-site bookkeeping across
  a stack; a driver that batches the wrong sites or seals on the wrong layer is fast and
  wrong, and the single-read suites cannot see it.

The gated command as CI runs it: **8 passed, 4 skipped in 79 s** on a LoudBox. The 4 skips
are the `(8, 4)` Galaxy arms, which need 32 chips.

### Notes for reviewers

**Blackhole only, and enforced in C++.** The op hardcodes `HiFi4 + fp32_dest_acc_en`, the
combination flagged by **#38306** on Wormhole — upstream `generic_reductions` downgrades to
HiFi3 there, this does not. It would otherwise compile, run, and be silently wrong.
Validation rejects a non-Blackhole arch outright, so the guarantee does not rest on the
suites' `skipif`: the op is bound on every arch through the nanobind module, and a caller
reaching it from model code hits a `TT_FATAL` instead of bad numbers. Downgrading fidelity
on Wormhole is the real fix and stays a follow-up — it changes numerics and needs
validating on a card nobody has run this on.

**`tp_factor == 1` is unsupported, by construction.** The constructor asserts `tp_factor > 1`.
The cross-rank exchange is what the single dispatch is built around; at TP=1 the reduce is
an identity and there is nothing to fuse. No suite has a TP=1 arm.

**`stats` is caller-allocated scratch.** It is not read on entry and holds nothing on exit —
it is an operand so a caller walking many read sites allocates it once. It must be
allocated across the whole mesh, because the exchange addresses a peer's slot by page of
that peer's own copy.

**`site` is a runtime argument, not part of the program hash.** A walk over R read sites
reuses one cached program.

**Where the CI rows live and what they cost.**

| test | pipeline | manifest | sku | timeout |
|---|---|---|---|---|
| `test_attn_res_gather_merge.py` + `tests/attn_res/model/` + the walk gate | Blackhole e2e | `blackhole_e2e_tests.yaml` | `bh_loudbox` | 3 min |
| `test_fast_weighted_reduce_nc.py` | L2 nightly | `ops_unit_tests.yaml` | `bh_p150b_civ2` | 2 min |

The gated e2e command measures 79 s, inside its 3-minute slot. `models/e2e/bh_loudbox` is
saturated at **193/193** — the next row anyone adds there needs a bump in the same PR or
`verify_time_budget.py` rejects it. `ttnn/unit/bh_p150b_civ2` goes 1045 -> **1047** for the
one op job.

**Why the op suite rides the e2e row rather than the directory-wide experimental suites:**
those run `-x`, so an untried arch would abort the whole experimental run instead of
skipping. The op is also multi-device — it reduces across TP — so a single-card SKU would
run a green arm over a computation the model never executes.

**The perf files are bringup-only and gate nothing.** No timing assertion appears in either;
a wall-clock threshold on a shared box is a flake generator. They report through loguru, so
`pytest -s` is what makes them visible. Only `test_walk_matches_reference` is registered in
CI, by node id.

**Open levers, not taken:**

- `FOLD_MIN_CANDIDATES = 4` gates the candidate-axis fold, so many of the 186 sites still
  run an unfused `ttnn.add`. Worth revisiting now that the add can ride inside the op.
- The wire format is padded 32x. Stat tiles are fp32 32x32 = 4096 B but carry real data
  only in column 0 (128 B).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-2026-07-30)


nmilicevicTT:
Implemented in https://github.com/tenstorrent/tt-metal/pull/53114, https://github.com/tenstorrent/tt-metal/pull/53116, https://github.com/tenstorrent/tt-metal/pull/53318

PR #52697, opened 2026-08-10
[model_tracer] Include hardware and mesh in configuration identity

Aswinmcw:
### 1. Hardware and mesh now count towards configuration identity

Identity is documented as `operation + arguments + hardware + mesh`, but the mesh was dropped: the same op with the same arguments on a `[1, 8]` and a `[2, 4]` mesh produced the **same** `config_hash` and collapsed into one configuration.

- `_extract_hardware_and_mesh()` only read the mesh from the legacy `machine_info["tensor_placements"]`, so for traces written in the current format it came back `None`. It now falls back to the promoted `mesh_device_shape` / `device_count`, normalised so `[1, 8]` and `"[1, 8]"` are one identity.
- `update_master_file()` matched on arguments alone; it now also requires hardware/mesh to agree.
- `recompute_config_hashes()` used to rehash mixed configurations from `executions[0]`, silently picking a winner. It now leaves them alone and reports them for splitting.

Model output is unaffected. The damage was downstream: anything joining on `config_hash` (results DB, per-mesh perf) was mixing topologies.

### 2. A dead ETH heartbeat is classified as infrastructure

UMD's `EthFirmwareHeartbeatError` ("Timed out waiting for ETH heartbeat…") is raised during topology discovery, so no op kernel ever runs — but it matched no signature and was booked as a test failure. Added to `_FABRIC_INFRA_SIGNATURES`, matching the invariant phrase only (ASIC ID and core coordinate vary, and the text says "to start" when the value is still 0).

In scheduled run 32208327970 these were the only 2 failures of 1273 records, both against `add_model_traced`. With this they become NOT_RUN: **1266 pass / 0 failures**, jobs still red as infra.

### Verification

Host-only, no device: mixed-mesh configs now hash apart and stay separate; identical `operation + arguments + hardware + mesh` still dedupes into one configuration with both executions kept. For the signature: the real message and its "to start" variant classify as infra, existing fabric/`run_mailbox`/TLB signatures still match, and real failures (PCC `0.0002…`, bare `0.0`, L1 OOM) still classify as failures.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:Aswinmcw/tracer-config-identity-mesh)


Aswinmcw:
Both findings were real. Fixed in 0920754c392 (gather) and 085014783e1 (identity).

**Do not collapse a distributed reduction to device 0** (`fast_reduce_nc_model_traced.py`) — correct, and the module says so itself: *"the output is sharded across chips and must be reassembled with mesh_tensor_to_torch (taking just device_tensors[0] returns only the chip-0 slice and leaves the rest of the global tensor empty)"*. It also preallocates its output with its own `ot_placement`, so input A's construction path is not evidence about how the reduction distributed its result. Reverted to the plain gather.

Auditing the other 36 wired modules for the same shape of problem found **two more you didn't flag**, both reverted:

- `linear` — the "Partial-reduce fallback" right after the gather depends on the reassembler *concatenating* per-chip partials so it can reshape+sum them into the global result. Collapsing would hand it a single partial.
- `rotary_embedding_llama` — in prefill the golden is built GLOBAL, explicitly "to match what `mesh_tensor_to_torch` reassembles".

Each carries a comment recording why it is deliberately not wired, so the omission doesn't look accidental to the next reader.

Where a module *does* preallocate its output on the mesh (`clamp`, `gt`, `multiply`, `slice`, `where`), I applied your argument rather than just reverting: those now pass `(ot_placement or input_a_tensor_placement)`, so the decision follows the tensor whose distribution is actually in question, falling back when the traced config carries no output placement. Net: 51 → 48 wired sites across 35 modules, re-audited per call site against its own enclosing function.

**Preserve device-count-only mesh identity** (`generic_ops_tracer.py:712`, and the sibling sites) — also correct. The fallback now reads `device_count` when the shape is absent or unusable, via a separate `_normalize_device_count`. The count is recorded **as a count, not converted into a shape**: 8 devices could be `[1, 8]`, `[2, 4]`, `[4, 2]` or `[8, 1]`, and picking one would assert a topology the trace never stated.

One scope note in the interest of accuracy: this changes nothing for current data. Every execution in both corpora I can check carries **both** fields — 647/647 locally and 2499/2499 in the CI master JSON from run 32238707094 — so there are no count-only records today and no hashes move. It closes the gap rather than fixing an active collision.

Verified host-only: shape+count and shape-only take the shape path unchanged (no `device_count` key added, so existing identities are untouched); count-only separates 1 device from 8; absent, non-numeric and zero counts yield `None` rather than a bogus identity; `[4, 8]` and `[8, 4]` stay distinct.

bbradelTT (approved):

Issue #51209, opened 2026-07-27
[Minimax M3] moe optimizations


philei-tt:
hide shared expert, revisit moe

PR #52698, opened 2026-08-10
[MiniMax-M3] MoE prefill: op fusion, TP-sharded residual, and a gate-bias bug fix

vmelnykovTT:
### PR Category

related https://github.com/tenstorrent/tt-metal/issues/51209

### Summary

Fuses the MoE gate (8 ops → 1 `moe_grouped_topk`) and the clamped swigluoai activation (7 ops → 1),
switches the residual stream to `emb/tp`-sharded with one shared all-gather per norm, widens dispatch
to 3 workers/sender, and bounds the gate and dispatch to a chunk's real tokens. Also fixes a `glob`
bug that was silently dropping the MoE gate's `e_score_correction_bias` on the production cache-only
path — that fix alone is what moves V PCC above its gate.

### Results

Single Blackhole Galaxy, mesh 8×4 (SP=8 / TP=4 / EP=32), 5120-token chunks, bf4 expert weights.

Throughput, last chunk, median of 5 iters (run-to-run drift ~1–2 %):

| config | main | branch | Δ |
|---|---:|---:|---:|
| Cold 5k | 5029.4 | 5679.2 | +12.9% |
| 25k @ 20k cache | 4712.6 | 5242.3 | +11.2% |
| 50k @ 46k cache | 4616.4 | 5316.4 | +15.2% |
| 55k @ 51k cache | 4695.5 | 5324.3 | +13.4% |

Min PCC across 60 layers, `longbook_10240` chunked 2×5120:

| tensor | before | after | gate |
| --- | --- | --- | --- |
| K | 0.96352 | 0.96731 | — |
| V | 0.87963 | 0.88728 | 0.88 — was failing, now clears by 0.007 |
| index_k | 0.97586 | 0.97875 | — |

Worth noting for future work: the gain came from removing op launches (2965 → 2505 per chip per
chunk, −15.5 %), not faster kernels — total device kernel time moved only −2.4 %. M3 prefill has no
trace capture, so every op is host-enqueued each chunk

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vmelnykov/minimax_m3_moe_prefill_perf)


models/demos/minimax_m3/tt/moe/activation.py
```diff
@@ -3,14 +3,125 @@
...
+
+_swiglu_path_logged = False
+
+
+def use_fused_swiglu() -> bool:
+    """True -> one multiply with activation spans; False -> the readable 7-op chain."""
+    v = os.environ.get("M3_FUSED_SWIGLU")
+    if v is None:
+        return DEFAULT_USE_FUSED_SWIGLU
+    return v.strip().lower() in ("1", "true", "yes", "on")
```
philei-tt: Should we always just use fused swiglu? Any reason not to use it? I think we have way too many undocumented env variables already...

models/demos/minimax_m3/tt/moe/tt_minimax_moe.py
```diff
@@ -19,17 +19,55 @@
...
+# parallelize the DRAM read and the untilize but not the allocation. A flat result at N=3 is therefore
+# evidence that the baton — not the core count — is the limit, which is the premise of the per-worker
+# offsets rework.
+DEFAULT_DISPATCH_WORKERS_PER_SENDER = 3
+
+
+def dispatch_workers_per_sender() -> int:
+    """Workers per dispatch sender. See above — values above 3 do not fit L1 at M3's emb."""
+    return int(os.environ.get("M3_DISPATCH_WORKERS", DEFAULT_DISPATCH_WORKERS_PER_SENDER))
+
```
philei-tt: Let's be honest nobody is going to read this. Can you summarize in a single paragraph? Also, why do we need to configure this with env variable. Just hardcode to what works best

models/demos/minimax_m3/tt/moe/tt_minimax_moe.py
```diff
@@ -63,16 +71,26 @@ def __init__(
...
+        # small device tensors to host every layer, which is a stall mid-forward and illegal under trace.
+        self.check_dispatch_overflow = check_dispatch_overflow or bool(int(os.environ.get("M3_MOE_AUDIT", "0")))
+        self._overflow_reported = False
 
         # MiniMax routing: sigmoid + e_score_correction_bias, no groups -> n_group=1.
+        # route_scale MUST match the model's routed_scaling_factor (2.0 for M3), not default to 1.0:
+        # the internal gate applies it to the returned top-k weights, so a 1.0 here silently halves
+        # every routed contribution the moment `gate_fallback_mode` selects the internal gate over the
+        # caller-supplied topk. Harmless while the production path injects topk_indices/topk_weights
+        # (tt/mlp.py -> TopKRouter already scales), which is exactly why it went unnoticed.
```
philei-tt: this comment is mentioning your debugging process. This is a complete noise after this pr gets merged

models/demos/minimax_m3/tt/moe/tt_minimax_moe.py
```diff
@@ -63,16 +71,26 @@ def __init__(
         self.seq_len_per_chip = seq_len_per_chip
         self.experts_per_chip = experts_per_chip
         self.emb_dim = emb_dim
+        self.max_dispatch_buffer_token_size = max_dispatch_buffer_token_size
+        # Host-readback overflow audit (see _check_dispatch_overflow). Off by default: it composes two
+        # small device tensors to host every layer, which is a stall mid-forward and illegal under trace.
+        self.check_dispatch_overflow = check_dispatch_overflow or bool(int(os.environ.get("M3_MOE_AUDIT", "0")))
```
philei-tt: Is this supposed to be committed? If so, if it's incompatible with trace, you should assert that this venv is unset when trace is enabled

models/demos/minimax_m3/tt/moe/tt_minimax_moe.py
```diff
@@ -207,23 +293,32 @@ def forward(self, x, topk_indices=None, topk_weights=None):
...
-
         with zone("experts_mm"):
+            # Hand the ROW_MAJOR dispatch buffer straight to the composite, as DeepSeek does. The
+            # layout selects the strategy (see TtRoutedExpert.forward): ROW_MAJOR streams sticks into
+            # cb_x_rm and tilizes each K-block in-kernel, so only each expert's real token region is
+            # converted. A standalone to_layout instead tilized all max_dispatch_buffer_token_size
+            # rows — 10240 at chunk 5120/SP=8, of which only ~640 are occupied (4 local experts), a
+            # 16x amplification costing ~0.54 ms/layer at zero cross-chip variance. Upstream #49744.
+            # The ROW_MAJOR input stays live for the duration of the call and the composite returns a
+            # fresh output, so free it after, not before.
```
philei-tt: Are you sure this comment is useful? It oddly specific...

models/demos/minimax_m3/tt/runners/adapters/minimax_m3.py
```diff
@@ -82,9 +82,15 @@ class MiniMaxM3PrefillAdapter(PrefillModelAdapter):
 
     l1_small_size = 0
 
-    # M3's sequence-parallel residual keeps the full embedding on every TP col (emb replicated, seq
-    # SP-sharded), so the D2D hidden state ships emb-replicated across TP.
-    pipeline_activation_emb_tp_sharded = False
+    # The D2D hidden state must ship in whatever layout the residual stream uses at a layer boundary:
+    # emb/tp-sharded by default, emb-replicated under M3_SHARDED_RESIDUAL=0. Read as a
+    # property so one env var moves the layout and the socket mapper together (a mismatch would hand the
+    # next rank a wrong-width activation).
```
philei-tt: Idk, I think that code is self-descriptive enough. Feel free to ignore if you disagree

models/demos/minimax_m3/tt/moe/tt_reduce.py
```diff
@@ -0,0 +1,126 @@
...
+
+M3's own copy of DeepSeek's TtReduceModule (deepseek_v3_d_p/tt/moe/tt_reduce.py), which we do not
+modify. The compute half is identical — the same shared `deepseek_prefill.post_combine_reduce` kernel.
+The difference is the collective:
+
+  DeepSeek  ttnn.reduce_scatter                    -- the plain prim, with no barrier semaphore
+  here      caller-supplied reduce_scatter_fn      -- M3 passes MeshConfig.reduce_scatter, i.e.
+                                                      reduce_scatter_minimal_async with the ping-pong
+                                                      + barrier semaphores every other M3 collective
+                                                      already uses (tt/config.py)
```
philei-tt: Are you sure you want to refer to details of other model implementation and files in comments? This comment might become stale very quickly, and deepseek folks won't be updating it.

models/demos/minimax_m3/tt/moe/tt_reduce.py
```diff
@@ -0,0 +1,126 @@
...
+
+  1. EP all-reduce. A mesh column owns only 32 of the 128 experts, so post_combine_reduce's output is a
+     PARTIAL sum over that column's experts. Summing across the 4 columns is what completes top-4.
+  2. TP scatter. The same op scatters emb across those 4 columns, handing the next stage emb/tp.
+
+Expert parallelism is folded onto the tensor-parallel axis. That is also why this collective shows the
+largest barrier wait in the block (measured 145 us work + 879 us wait): unlike `combine`, which
+synchronises 8 chips that share a token set, this synchronises 4 chips whose expert loads are
+unrelated, so four independent whale-expert skews meet here.
+"""
```
philei-tt: > Why bother, given both reduce over the same axis and measure the same ~145 us of work. 

Again, you mention specific numbers from debuggin. This is noise.

models/demos/minimax_m3/tt/residual.py
```diff
@@ -0,0 +1,132 @@
...
+#       result. 4 ops per norm; the big ops touch only emb/tp.
+#
+# MEASURED on the 8x4 galaxy (Tracy capture, LEVEL=3 LAYERS=6 CACHE=25600, 42 layer-visits, against
+# the 46a19f8 baseline capture 2026_08_05_18_07_13) — "distributed" loses on BOTH axes:
+#
+#                       device-op launches      DEVICE kernel time (max-across-chips)   wall-clock
+#   baseline                  2785                        309.53 ms                        --
+#   gather_first              2757  (-1.0%)               302.61 ms  (-2.2%)             -0.1%
+#   distributed               2925  (+5.0%)               308.90 ms  (-0.2%)             -4.6%
+#
```
philei-tt: 😫

models/demos/minimax_m3/tt/residual.py
```diff
@@ -0,0 +1,132 @@
...
+    v = os.environ.get("M3_SHARDED_RESIDUAL")
+    if v is None:
+        return DEFAULT_USE_SHARDED_RESIDUAL
+    return v.strip().lower() in ("1", "true", "yes", "on")
+
+
+def norm_mode() -> str:
+    """Either ``gather_first`` (default) or ``distributed`` — see DEFAULT_NORM_MODE. Only meaningful
+    when the residual is sharded; the replicated scheme always norms full emb in one pass."""
+    v = (os.environ.get("M3_SHARDED_RESIDUAL_NORM") or DEFAULT_NORM_MODE).strip().lower()
```
philei-tt, line 47: Can you please document ALL variables you added?

models/demos/minimax_m3/tt/residual.py
```diff
@@ -0,0 +1,132 @@
...
+    single-pass on a gathered full-emb input (either the replicated scheme, or gather_first)."""
+    return use_sharded_residual() and norm_mode() == "distributed"
+
+
+def gather_before_norm() -> bool:
+    """True -> the layer's per-norm all-gather runs BEFORE the norm (gather_first); False -> after it."""
+    return use_sharded_residual() and norm_mode() == "gather_first"
+
+
+def log_scheme_once(mesh_config=None) -> bool:
```
philei-tt: I don't understand why you need this complication. Why can't you just log this once at model creation time, why do you need to check on this global variable. Why do even need to move this in a separate function

models/demos/minimax_m3/tt/topk.py
```diff
@@ -113,10 +133,93 @@ def __init__(self, mesh_device, hf_config, state_dict, tensor_cache_path=None):
             else None
         )
 
+        # The fused gate needs the bias at the FULL [num_tokens, num_experts] shape: the op asserts
+        # bias.logical_shape() == scores.logical_shape(). Built ONCE here, never per call, and never via
+        # a host round-trip — one 640x128 bf16 tensor (~164 KiB) per layer, constant for the whole run.
```
philei-tt: Why do you need to mention host round-trip if you don't even do it? Also this commetn assumes 5k chunk with 8 sp, which is not true in general case

philei-tt (commented):
PLEASE remove those useless bloated ai comments, they trash the code and artificially increase it's isze. I didn't flag all places, but pretty much all I checked are not good.

philei-tt (commented):
A few notes from a review pass, mostly around the new sharded-residual plumbing.

**Question (not a suggested fix):** `TopKRouter._padding_config_cache` (`topk.py`, `build_padding_config`) memoizes one device tensor per distinct `actual_isl` value it sees, keyed by real-token count, and the dict is never evicted for the life of the process.

Worked out the actual bound: a full chunk always short-circuits to `None` before touching the cache (`build_padding_config`'s early return), so only a *ragged final chunk* ever populates it — and a chunk's raggedness is bounded by `chunk_size`, not by `max_seq_len`/context length (a longer context just adds more full chunks, which are no-ops here). So with the default `chunk_size=5120` and 57 MoE layers, the theoretical ceiling is `chunk_size - 1` ≈ 5,119 entries per layer (~292k device tensors total, model-wide) — small in absolute device-memory terms, and it saturates rather than growing without bound.

Given that, this doesn't look like a memory problem in practice. Mostly want to confirm: is an unbounded-dict-that-happens-to-have-a-low-ceiling the intended design here, or would you rather see an LRU/cap on `_padding_config_cache` regardless of how small the current ceiling works out to be? Not suggesting it be removed — just want to make sure the tradeoff is deliberate rather than incidental.

models/demos/minimax_m3/tt/topk.py
```diff
@@ -113,10 +133,93 @@ def __init__(self, mesh_device, hf_config, state_dict, tensor_cache_path=None):
...
+        [c*tokens_per_chip, (c+1)*tokens_per_chip) and its real count is the clamped remainder. That is
+        DeepSeek's non-rotated branch; M3 needs neither its zigzag nor its rotated-chunk case.
+
+        NOTE this ends in a ttnn.from_torch, i.e. a host->device write. Fine untraced — it is memoized,
+        so it costs one tiny [sp, 2] write per distinct count — but a traced runtime must derive the
+        row on-device instead (DeepSeek's build_padding_config_device / the moe_padding_config op).
+        Port that alongside any trace work.
+        """
+        tokens_per_chip = self.num_tokens
+        sp_factor = self.mesh_device.shape[0] if isinstance(self.mesh_device, ttnn.MeshDevice) else 1
```
philei-tt: `sp_factor` is read straight off `self.mesh_device.shape[0]`, hardcoding SP on mesh axis 0. Everywhere else touched by this PR (`layer.py`, `mlp.py`, `attention/operations.py`, ...) resolves the axis via `mesh_config.tp_axis`/`sp_axis` instead of assuming a fixed one — `MeshConfig` already derives `sp_axis` as the complement of `tp_axis` (`config.py`). `TopKRouter` isn't constructed with a `mesh_config` at all here (only a raw `mesh_device`), so if a future deployment ever sets `tp_axis=0` (SP on axis 1) — a configuration `MeshConfig` already supports — `build_padding_config` would compute the per-chip real-token split against the wrong axis and mis-mark which rows are padding (real tokens sentinel-marked as pad, or vice versa).

Today's only deployed config (`tp_axis=1`, SP on axis 0) happens to make this correct, so it's silent rather than broken right now. `MLP.__init__` (the only caller, `mlp.py`) already has `mesh_config` in scope when it builds `TopKRouter` — might be worth threading it through and using `mesh_device.shape[mesh_config.sp_axis]` here instead, for consistency with the rest of the PR.

models/demos/minimax_m3/tt/dense_mlp.py
```diff
@@ -81,13 +88,23 @@ def __call__(self, x):
...
         act.deallocate(True)
-        # down is row-parallel: each TP device holds a partial sum over the intermediate
-        # shard -> all-reduce to complete the hidden output.
+        # down is row-parallel: each TP device holds a partial sum over the intermediate shard, so a TP
+        # collective is required either way. Sharded residual -> reduce-scatter only (emb/tp out, which
+        # the caller adds straight into its residual); replicated residual -> full all-reduce (RS + AG).
         if self.mesh_config.tp > 1:
-            with zone("tp_allreduce"):
-                out = self.mesh_config.allreduce(out, self.ccl_manager, axis=self.mesh_config.tp_axis)
+            if self.scatter_output:
```
philei-tt, line 100: This reduce-scatter (and the analogous one in `mlp.py`'s `moe_reduce_scatter`, ~line 160) closes the sharded-residual contract the same way `attention/operations.py::apply_reduce_scatter` does, but skips the `assert_sharded_residual_unpadded` check that function calls right before scattering. That assert exists specifically because a non-tile-aligned `hidden_size / tp` would land the o_proj-style output padding inside only the last TP column's slice after a reduce-scatter, silently corrupting the residual in a way nothing downstream can trim.

M3's current shape (6144/4=1536) is tile-aligned, so this is a non-issue today — but as written, `DenseMLP`'s (and `mlp.py`'s) reduce-scatter would fail silently instead of loudly on a future config that isn't tile-aligned, which is exactly the failure mode this PR hardened attention against. Might be worth calling the same guard here for parity.

models/demos/minimax_m3/tests/unit/test_model_sp_vs_ref.py
```diff
@@ -209,15 +210,21 @@ def test_model_sp_dense_vs_ref(mesh_device, device_params, seq_len, reset_seeds)
...
-    # --- SP input shard: one sequence, S/sp rows per device, hidden replicated across TP cols ---
+    # --- SP input shard: one sequence, S/sp rows per device. Hidden is replicated across the TP cols
+    # under the default residual scheme and emb/tp-sharded under M3_SHARDED_RESIDUAL — the model's
+    # residual contract, which this test bypasses the embedding to feed directly. The RoPE tables keep
+    # the replicated-hidden dims either way (rotary_dim is never TP-sharded), hence the separate list. ---
     in_dims = [None, None]
     in_dims[sp_axis] = 2  # seq -> SP rows
+    x_dims = list(in_dims)
+    if use_sharded_residual():
+        x_dims[mesh_config.tp_axis] = 3  # emb -> TP cols
```
ayerofieiev-tt, line 221: **[1/10] Only the dense SP test got the `x_dims` emb-sharding fix** — `test_model_sp_moe_vs_ref` (line 379) and `test_model_sp_msa_vs_ref` (line 666) still build `x_tt` with `dims=in_dims` (emb replicated).

Both run only on (8,4) (tp=4). With the new default `M3_SHARDED_RESIDUAL=1`, layer 0's `_norm_to_full_emb` all-gathers the already-full `[1,1,s_local,6144]` input to `[1,1,s_local,24576]`; the q/k/v matmul (K=6144) then hits a shape fatal. The model-level MoE and MSA PCC gates — the checks that guard exactly this PR's data flow — cannot run. This hunk shows the required fix; the same `x_dims` treatment is needed at lines 372-379 and 659-666.

models/demos/minimax_m3/tt/mlp.py
```diff
@@ -158,27 +186,47 @@ def __init__(
...
-            idx, wts = self.router(hidden_states, True)  # per-row top-k on [1,1,S,H]
+            idx, wts = self.router(hidden_states, padding_config=padding_config)  # per-row top-k
         x3d = ttnn.squeeze(hidden_states, dim=0)  # [1,1,S,H] -> [1,S,H] per device
-        out = self.experts(x3d, topk_indices=idx, topk_weights=wts)  # -> [1,S,H/tp] reduce-scattered
+        out = self.experts(
+            x3d, topk_indices=idx, topk_weights=wts, padding_config=padding_config
+        )  # -> [1,S,H/tp] reduce-scattered
         out = ttnn.unsqueeze(out, dim=0)  # -> [1,1,S,H/tp]
-        if self.mesh_device.shape[1] > 1 and out.shape[-1] < Hfull:
+        if not self.sharded_residual and self.mesh_device.shape[1] > 1 and out.shape[-1] < Hfull:
```
ayerofieiev-tt, line 229: **[2/10] `test_ep_moe_vs_ref.py:139` breaks under this new default** — gating the trailing all-gather on `not self.sharded_residual` means `mlp(tt_x)` now returns `[1,1,S,1536]` by default, but the test (runs only on (8,4)) still reads TP column 0 and does `.reshape(-1, HIDDEN=6144)`: it either raises on element count or garbles S/4 rows of interleaved shards. The only end-to-end MoE-vs-reference check — the one guarding the fused gate, padding config, and `TtMiniMaxReduce` — fails before any PCC comparison. The test needs to either concatenate the TP column shards or construct the MLP for the replicated layout.

models/demos/minimax_m3/tt/residual.py
```diff
@@ -0,0 +1,61 @@
...
+Every consumer (embedding, norms, attention's closing collective, both MLPs, the LM-head tail) reads
+these helpers so a run cannot end up half-sharded.
+"""
+
+import os
+
+# Default ON: measured better on device time and op launches than the replicated layout (KV PCC
+# bit-identical), and it is the layout the sibling models use. M3_SHARDED_RESIDUAL=0 restores the
+# replicated layout, kept for bisecting against baselines measured on it.
+DEFAULT_USE_SHARDED_RESIDUAL = True
```
ayerofieiev-tt, line 26: **[3/10] Three more pre-existing module tests encode the old replicated-residual contract and break under this flipped default**: `test_attention_vs_ref.py:178` (and `test_attention_chunked_vs_ref.py:175/340`), `test_dense_mlp_vs_ref.py:86`, `test_parallel_embedding_vs_ref.py:77`.

All parametrize (8,4). `attention_forward` returns emb/tp via `apply_reduce_scatter`, `DenseMLP` takes the `tp_reduce_scatter` branch (`scatter_output=None` derives True), and `TtParallelEmbedding` skips its closing TP all-gather — each test then reshapes/compares a `[.,.,S,1536]` tensor against full 6144 width and raises. Additionally, `test_dense_mlp_vs_ref` parametrizes small hidden sizes for which `assert_sharded_residual_unpadded` fires at construction. None of these files were updated; only a manual `M3_SHARDED_RESIDUAL=0` run passes, so CI coverage of both the new and old paths is silently lost.

models/demos/minimax_m3/tt/moe/tt_minimax_moe.py
```diff
@@ -207,33 +220,35 @@ def forward(self, x, topk_indices=None, topk_weights=None):
         # Dispatch -> per-expert buffers (NO shared expert)
         with zone("dispatch"):
             dispatched_buffer, metadata = self.dispatch_module(
-                x, scores, indices, tt_expert_offsets, self.tt_expert_dispatch_table
+                x,
+                scores,
+                indices,
+                tt_expert_offsets,
+                self.tt_expert_dispatch_table,
+                padding_config=padding_config,
```
ayerofieiev-tt, line 228: **[4/10] `padding_config` reaches dispatch unconditionally but never the internal DeepSeek gate** on the `topk_indices=None` branch (line 196), breaking the gate/dispatch pairing the docstring itself calls "only consistent together".

A caller on the internal-gate path (the documented standalone/op-test path) with a non-None `padding_config` gets: the gate routes the padded tail rows as real (no sentinel marking) while dispatch shortens its token loop to `num_real_tokens`; `post_combine_reduce` then reads combine slots that were never written because `indices`/`scores` claim they were routed → uninitialized/1e38 values in the tail rows of `routed_output`. Thread `padding_config` into `self.gate(...)` or `assert padding_config is None` when `topk_indices is None`.

models/demos/minimax_m3/tt/rms_norm.py
```diff
@@ -26,73 +57,111 @@ def __init__(self, mesh_device, hf_config, state_dict, tensor_cache_path=None, m
...
+            # cache-only path there is no torch source, so the gain is recovered from the replicated
+            # device tensor with one tiny init-time D2H (12 KiB per norm) — no new cache entry, hence no
+            # dependence on the (often other-user-owned, read-only) weight cache being writable.
+            self.tt_weight = ttnn.from_torch(
+                torch_weight if torch_weight is not None else ttnn.to_torch(ttnn.get_device_tensors(replicated)[0]),
+                device=mesh_device,
+                dtype=ttnn.bfloat16,
+                layout=ttnn.ROW_MAJOR_LAYOUT,
+                memory_config=ttnn.DRAM_MEMORY_CONFIG,
+                mesh_mapper=self.mesh_config.shard_mapper(mesh_device, mesh_dims=(None, -2)),
```
ayerofieiev-tt, line 90: **[5/10] Distributed-norm gain sharded with hardcoded `mesh_dims=(None, -2)`** (always mesh axis 1) while this same function's stats all-gather and stats memory-config were generalized to `mesh_config.tp_axis`.

`MeshConfig` explicitly supports `tp_axis=0` (config.py:24-36), and `shard_mapper(tensor_dim=-2)` would honor it — the explicit `mesh_dims=(None,-2)` bypasses that. On a `tp_axis=0` mesh with `M3_SHARDED_RESIDUAL_NORM=distributed`, the gain shards across the SP axis while stats gather across the TP axis: every device scales its emb/tp slice by the wrong 1/tp slice of the gain — silently wrong output, with no shape error when rows==cols. The new test only covers (8,4) with tp_axis=1, where the two coincide.

models/demos/minimax_m3/tt/parallel_embedding.py
```diff
@@ -99,6 +100,10 @@ def __init__(
         self.sp_axis = mesh_config.sp_axis
         self.dtype = dtype
         self.shard_vocab_on_sp = shard_vocab_on_sp
+        # Residual-stream layout (tt/residual.py). Both lookup paths already produce emb/tp naturally —
+        # the 1D path from its local emb-dim slice, the 2D path out of its SP reduce-scatter — so the
+        # sharded scheme is a matter of NOT running the closing TP all-gather, not of adding anything.
+        self.gather_emb = not use_sharded_residual()
```
ayerofieiev-tt, line 106: **[6/10] The residual layout is re-derived from the env at 9 sites with inconsistent guards and lifetimes**: this line keys on `use_sharded_residual()` alone plus mesh shape, most modules latch `... and mesh_config.tp > 1` at init, and `attention/prefill.py:358` re-reads the env on every forward.

Two constructible half-sharded runs: (a) `MeshConfig((8,4), tp=1)` — valid per config.py — with the default env: the embedding stays emb/4-sharded (it keys on `mesh_device.shape`) and skips its gather, while `DecoderLayer.sharded_residual` is False (tp==1) so no gather ever restores full width → K-dim fatal in layer 0's q/k/v. (b) Any process that mutates `M3_SHARDED_RESIDUAL` between model construction and forward: attention alone re-derives per call and would reduce-scatter into a full-emb residual. Resolve the scheme once (e.g. a field on MeshConfig) and have every consumer read it.

models/demos/minimax_m3/tt/moe/tt_reduce.py
```diff
@@ -0,0 +1,105 @@
...
+            expert_dispatch_table: expert id -> chip id, INT32, sharded per dispatch group. Together
+                with `indices` this is what lets the kernel SKIP the ~3 of 4 slots belonging to other
+                dispatch groups, so the untouched slots are never read.
+
+        Returns:
+            [seq_len, emb_dim / num_chips_in_cluster_axis]
+        """
+        if weights is not None:
+            # [..., topk] -> [..., topk, 1] so the multiply broadcasts along emb.
+            if weights.shape[-1] != 1:
```
ayerofieiev-tt, line 75: **[7/10] `TtMiniMaxReduce` is a near-verbatim fork of DeepSeek's `TtReduceModule` made solely to inject `reduce_scatter_fn`, and it is already stale**: it carries the naive `weights.shape[-1] != 1` check that upstream replaced with `_weights_have_output_channel_dim` (models/demos/deepseek_v3_d_p/tt/moe/tt_reduce.py:29,107).

With `num_experts_per_tok==1`, scores reshape to `[1,S,1]`; the unsqueeze is skipped and the while-loop left-pads to `[1,1,1,S,1]` against `combine_output [1,1,S,1,emb]` → misaligned weighted sum. M3 is top-4 so it does not bite today, but every future fix to the shared reduce path now silently misses this copy. Prefer adding an optional `reduce_scatter_fn=None` parameter to the shared `TtReduceModule` (default preserves DeepSeek behavior) and deleting this file.

models/demos/minimax_m3/tt/topk.py
```diff
@@ -113,10 +122,83 @@ def __init__(self, mesh_device, hf_config, state_dict, tensor_cache_path=None):
...
+                layout=ttnn.TILE_LAYOUT,
+                device=mesh_device,
+                memory_config=ttnn.DRAM_MEMORY_CONFIG,
+            )
+
+        # Padding-config memo, keyed by real-token count (see build_padding_config). Deliberately
+        # never evicted: only a ragged final chunk populates it, so it saturates at chunk-size-many
+        # tiny [sp, 2] tensors rather than growing with context length.
+        self.mesh_device = mesh_device
+        self._padding_config_cache = {}
```
ayerofieiev-tt, line 151: **[8/10] `build_padding_config` re-implements `TtMoEGatePrefill.build_padding_config`** (reachable as `self.experts.gate.build_padding_config`, models/demos/deepseek_v3_d_p/tt/moe/tt_moe_gate_prefill.py:596), and its per-router, never-evicted cache multiplies cost and memory by the layer count.

Each of the ~57 MoE layers owns a TopKRouter with its own `_padding_config_cache`, and mlp.py calls `build_padding_config` inside every layer's forward — so every ragged final chunk pays 57 identical torch builds + H2D writes + `logger.info` calls on the dispatch-critical host path, and a long-lived serving process accumulates up to 5120 x 57 never-freed device tensors (the docstring's "chunk-size-many" bound counts one router, not all). The copy also hardcodes right-padding and derives sp from mesh rows, which the shared helper parameterizes. Build the config once per chunk at the runtime/model level and pass the tensor down.

models/demos/minimax_m3/tt/mlp.py
```diff
@@ -138,6 +149,23 @@ def __init__(
...
+        # semaphore), so every M3 collective goes through one managed path.
+        # Contract: reduce over the TP axis, scatter on the last dim.
+        # dim is resolved from the tensor's rank rather than passed as -1: MeshConfig.reduce_scatter
+        # forwards straight to reduce_scatter_minimal_async, which (unlike the ttnn.reduce_scatter
+        # wrapper it replaces) does not normalize a negative dim.
+        moe_reduce_scatter = None
+        if mesh_config is not None and ccl_manager is not None and mesh_config.tp > 1:
+            # Same guard attention's apply_reduce_scatter runs: a non-tile-aligned hidden/tp would land
+            # output-dim padding inside one TP column's residual slice after the scatter.
+            assert_sharded_residual_unpadded(mesh_config, hf_config.hidden_size)
```
ayerofieiev-tt, line 164: **[9/10] The assert+reduce-scatter body is inlined three times instead of calling the `apply_reduce_scatter` helper this same PR adds, and this assert runs unconditionally at tp>1** — including under the replicated layout the env flag is documented to fall back to.

`attention/operations.py::apply_reduce_scatter`, `dense_mlp.py:100-107`, and this lambda re-inline the same guard+collective and already diverge (dim=3 vs `len(t.shape)-1`; only the helper deallocates its input). Worse: on a future config with non-tile-aligned hidden/tp, setting `M3_SHARDED_RESIDUAL=0` — the documented escape hatch for exactly that situation — still hard-fails at MLP construction with a message about the sharded residual. The MoE/dense MLPs also now import a residual invariant from the attention package; it belongs in tt/residual.py, checked once at Model init.

models/demos/minimax_m3/utils/general_utils.py
```diff
@@ -17,12 +17,31 @@ def cache_file_exists(cache_file_name):
...
         return False
-    import glob
+    import os
 
-    return bool(glob.glob(f"{cache_file_name}*.tensorbin"))
+    directory, prefix = os.path.split(str(cache_file_name))
+    if not directory or not prefix:
+        return False
+    try:
+        return any(e.startswith(prefix) and e.endswith(".tensorbin") for e in os.listdir(directory))
```
ayerofieiev-tt, line 42: **[10/10] This glob-metacharacter fix was applied only to minimax_m3's copy of `cache_file_exists`** — the byte-identical buggy copy remains at models/demos/gpt_oss_d_p/utils/general_utils.py:24, and minimax's own weight_cache.py already contains a third, correct implementation of the same probe.

The bug class is "cache-presence probe built from a path containing glob metacharacters" (`MeshShape([8, 4])` parsed as a character class), and its failure mode is silent — an optional cached weight reported absent, i.e. exactly the unbiased-routing regression this PR chased. The gpt_oss_d_p twin will reproduce it the moment anything calls it. Fix the twin now; longer-term, collapse the three probes into one shared directory-scan helper so the check cannot drift per model.

ayerofieiev-tt (commented):
Automated review (recall-biased, 8 finder angles + per-candidate verification). 10 findings, ranked most severe first. The dominant class: the PR flips `M3_SHARDED_RESIDUAL` to default-on while updating only one of the ~7 pre-existing tests that encode the old replicated-residual contract, so most of the multi-device PCC gates covering this change cannot pass. Findings 4-6 are silent-wrong-numbers risks on reachable-but-untested configurations; 7-10 are verified reuse/altitude debts with concrete correctness edges.

philei-tt (approved):

Issue #52675, opened 2026-08-09
ttnn.pow (bfloat16 path): relative error grows linearly with exponent (~0.2%*|y|; 18% at y=100); overflow returns NaN instead of inf


JerodLee:
### Summary

`ttnn.pow(tensor, float_scalar)`, `ttnn.pow(tensor, tensor)` and `ttnn.rpow` on non-fp32-dest tensors (BFLOAT16 / BFLOAT8_B / BFLOAT4_B) run `_sfpu_unary_power_21f_` / `_sfpu_binary_power_21f_`, which compute `pow(x,y) = 2**(y * log2(x))` with log2 coming from a single 3rd-order minimax polynomial on the mantissa ([`ckernel_sfpu_unary_power.h` L60](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_unary_power.h#L60)). That polynomial's absolute error (max ≈ 2.8e-3, worst as mantissa → 2⁻) is multiplied by the exponent before the `2**z` step, so the **result's relative error grows ≈ 2.0e-3 · |y| with no bound** — silent, finite, wrong values across the documented input domain:

| y | max rel err | median rel err (bf16-exact bases in [0.001, 1000]) |
|---|---|---|
| 2 | 0.5% | 0.16% |
| 10 | 2.0% | 0.66% |
| 32 | 5.8% | 1.8% |
| 100 | 18.4% | 5.8% |
| 500 | 133.6% | 29.3% |

Concrete: `ttnn.pow(x=0.996094, 100.0)` returns **0.8008** where the true value is **0.6761** (18% off). Negative exponents mirror this (−100 → 17.5%). The bf16 output ULP floor is ~0.39%, so the algorithm noise dominates the format for |y| ≳ 3.

The fp32 path was completely rebuilt under the pow accuracy bounty #49625 (NR reciprocal + atanh-series log + hi/lo split); the bf16 path — the default dtype for most users — still runs the original algorithm.

A sharp edge for users: the `int32_t` exponent overload uses exact repeated squaring, so `ttnn.pow(x, 100)` (Python int) is accurate while `ttnn.pow(x, 100.0)` (float) and tensor exponents are not.

### Secondary defect: overflow returns NaN instead of +inf

[...]

PR #52707, opened 2026-08-10
fix(sfpu): saturate pow 21f overflow to +inf instead of NaN (bf16 dest path)

JerodLee:
### Problem

The bf16/non-fp32-dest pow path (`_sfpu_unary_power_21f_` / `_sfpu_binary_power_21f_`) clamps `z = y*log2(x)` only on the low side (`z < -127`). For `z >= 128` the biased-exponent reconstruction wraps instead of saturating, so large **finite** results come back as **NaN**:

```
ttnn.pow(x, 128.0)  with x=2.0     -> NaN   (want +inf)
10.0 ** 39.0                        -> NaN   (want +inf)
1000.0 ** 15.0                      -> NaN   (want +inf)
```

This is part 1 of #52675, per the suggested two-PR split there (the accuracy fix for the same path is the follow-up).

### Fix

Add the high-side guard the fp32 path already has (`v_if(out_exp >= 255) { y = infinity; }` in `_sfpu_pow2_f32_accurate_hilo_`), adapted to the 21f path:

- Predicated on the **live `z_f32` register**, which at that point holds `z * 2**23` (after the `addexp(z_f32, 23)` rescale) — hence the `2**30` threshold. No extra LReg is consumed, no new `v_if` cascade beyond the single guard.
- Placed **before** the zero-base NaN override and the negative-base handling, so `pow(0, -p) -> NaN` still wins and the sign for negative bases with odd integer exponents is still applied via `setsgn2` (`(-2)**129` -> `-inf`).

Applied to the Wormhole and Blackhole copies of both the unary and binary kernels (the four files share the algorithm).

### Testing

- Float32-faithful simulation of the kernel algorithm (same model that reproduces the error tables in #52675 exactly): all 1263 overflow cases in a sweep over bf16-exact bases in [0.001, 1000] × y in [-1000, 1000] now return correctly-signed inf; non-overflow results are bit-identical to the current kernel.
- I don't have silicon; happy to run ttsim per `tt_metal/tt-llk/tests/TTSIM.md` if maintainers want device confirmation, and the guard mirrors an existing in-repo pattern.

### DIFF
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
@@ -113,6 +113,14 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_21f_(sfpi::vFloat base, sfpi::vFloat
 
     sfpi::vFloat y = sfpi::as<sfpi::vFloat>(zii);
 
+    // High-side guard, mirroring the fp32 path: z >= 128 means the result's biased
+    // exponent overflows 8 bits, so the exponent-field reconstruction above wraps
+    // instead of saturating and returns NaN for large finite results
+    // (e.g. pow(2.0f, 128.0f), pow(1000.0f, 15.0f)). z_f32 still holds z * 2**23
+    // (addexp above), so the threshold is 128 * 2**23 = 2**30.
+    v_if(z_f32 >= 0x1p30f) { y = std::numeric_limits<float>::infinity(); }
+    v_endif;
+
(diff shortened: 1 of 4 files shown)


JerodLee:
Closing this: I measured the overflow cases on ttsim (details in https://github.com/tenstorrent/tt-metal/issues/52675#issuecomment-5256429748) and the NaN-instead-of-inf behavior this PR guards against does not reproduce — the measured kernel already saturates to ±inf with correct sign. My static analysis of the reconstruction was wrong on this point, so this guard has no demonstrated defect to fix.

The real issue on this path is the accuracy defect (measured +23.1% at y=100, 2^15 → 31744), which needs the log2-stage rework tracked in #52675 — that fix's reconstruction saturates naturally, so a separate guard wouldn't be needed anyway.

Issue #48267, opened 2026-06-26
[Bug] tensor to_string mismatches buffers and coordinates for multi-host mesh tensors (wrong device labels)


cmaryanTT:
## Summary

Sub-issue of #48188 (Bug #14 from the AI-harvested bugs table).

**File:** `ttnn/core/tensor/tensor_impl.cpp:318`
**Priority:** Low
**Category:** Index/coordinate desync (iterating two containers with different lengths)

## What's wrong

In `to_string_impl` for device tensors, the loop iterates over `buffers` (index `i`) and advances `coords_it` once per buffer, assuming `buffers[i]` and `coords[i]` are 1:1.

- `coords = storage.get_coords()` returns **ALL** device coordinates.
- `buffers` is built by `get_host_buffers()` → `DistributedHostBuffer::apply()`, which (per its header comment) only visits **populated/local** shards and skips non-local ones.

On a multi-host mesh where some shards are remote, `buffers.size() < coords.size()`, so the per-buffer `*coords_it++` pairs each local buffer with the wrong coordinate, printing the wrong `device_id` / coordinate label. The `mesh_device.is_local(coord)` guard checks the mis-paired coordinate, compounding the mislabeling.

Single-host (all shards local) is unaffected. This is a display-correctness bug limited to multi-host prints.

## Suggested Fix

Iterate coordinates and only consume a buffer for coordinates that actually have a populated/local shard, e.g.:

```cpp
const auto& dist_buffer = row_major_tensor.host_storage().buffer();
std::stringstream ss;
bool first = true;
for (const auto& coord : dist_buffer.shard_coords()) {
    auto maybe_shard = dist_buffer.get_shard(coord);
    if (!maybe_shard.has_value() || !mesh_device.is_local(coord)) continue;
    if (!first) ss << std::endl;
    first = false;
    ss << "device_id: " << mesh_device.get_device(coord)->id() << ", " << coord << std::endl;
    detail::to_string(ss, maybe_shard->view_as<T>(), shape, strides, tensor.dtype(), tensor.layout());
}
return ss.str();
```

amahmudTT:
@azaytsev-epam  is this closed, i see you merged a pr, could you close this if done ?

azaytsev-epam:
> [@azaytsev-epam](https://github.com/azaytsev-epam) is this closed, i see you merged a pr, could you close this if done ?

@amahmudTT You are right - need to be closed. closed

azaytsev-epam:
[...]

Issue #48188, opened 2026-06-25
Bugs harvested by AI


amahmudTT:
While looking for races, found that AI is not bad at finding existing bugs in the repo. And I verified one bug it found (a loop running 0 times and totally skipping the work.). This is an attempt to find as many bugs as possible. Thus creating a master issue regarding bugs marked by AI. Since there are many bugs I will label it as P1. The suggested solutions are also in the table. 

### Component / Area

tt-metal

### Issue Type (optional)

Multiple bugs, Master Issue

## Confirmed correctness bugs (adversarially verified)

| # | Priority | File:line | Category | What's wrong (and correct behavior) | Suggested fix |
|---|---|---|---|---|---|
| 1 | High | `ttnn/cpp/ttnn/operations/eltwise/unary_backward/unary_backward.cpp:88` | Inverted condition / wrong comparison (copy-paste slip) | In the two-bounds branch of the tensor-overload clamp_bw (min and max both tensors), the in-range mask is computed with inverted comparisons: minT = le(input, min) and maxT = ge(input, max), then anded together. The clamp backward must pass the gradient only where the input was NOT clamped, i.e. where min <= input <= max, so it should use minT = ge(input, min) and maxT = le(input, max). As written, the mask is (input <= min) AND (input >= max), which is empty for any normal min < max and therefore produces an all-zero gradient. This is confirmed by the sibling float-overload of clamp_bw (lines 58-59), which correctly uses ge(input,min) and le(input,max), and by the single-bound branches in this same tensor overload (lines 77 and 83), which also correctly use ge for the min bound and le for the max bound. Only the both-bounds branch is inverted. clip_bw delegates to clamp_bw, so it is affected too. Result: wrong (zero) gradients for clamp/clip backward when both min and max are passed as tensors. | At /localdev/amahmud/tt-metal/ttnn/cpp/ttnn/operations/eltwise/unary_backward/unary_backward.cpp line 88 change to `Tensor minT = ttnn::ge(input, min.value(), std::nullopt, output_memory_config);` and line 89 to `Tensor maxT = ttnn::le(input, max.value(), std::nullopt, output_memory_config);`, so the in-range mask is (input >= min) AND (input <= max), matching the float overload (lines 58-59) and the single-bound branches (lines 77, 83). |
[...]

PR #52708, opened 2026-08-10
Bug #48267  Pair printed shards with their own coordinates in to_string

azaytsev-epam:
### PR Category
Bug fix
   **Note: T3000 tests** Four of the twelve test jobs failed, none of them related to this PR. The same four fail identically on the most recent scheduled run of this workflow on main ([run 31078748491](https://github.com/tenstorrent/tt-metal/actions/runs/31078748491), and all five most recent main runs are red), and that baseline additionally fails t3k_tt_metal_multiprocess_tests, which passed on the branch.
   
 ### Summary
Fix `ttnn::to_string` mislabeling and silently omitting shards when printing a
device tensor on a multi-host mesh (#48267, sub-issue of #48188).
#### Root cause
The multi-device print path zipped two containers that agree in neither length
nor order. `DistributedHostBuffer::apply()` visits only shards that are both
local and populated, in ascending flat index over the global mesh, and passes no
coordinate to its callback. `DeviceStorage::get_coords()` returns every
coordinate the tensor spans, with no locality filtering. The loop advanced the
coordinate iterator once per buffer, so it consumed a prefix of the coordinate
list rather than the local subset.
Two things followed. A buffer paired with a remote coordinate was skipped by the
`is_local` guard and never printed at all, so locally owned shards silently
vanished from the output. A buffer paired with a local coordinate belonging to a
different device was printed under that wrong coordinate and device_id. The
separator was also emitted for skipped shards, leaving stray blank lines.
On the T3K dual-rank 2x4 binding, where each rank owns a 2x2 block, printing a
full-mesh tensor showed only 2 of each rank's 4 local shards. On rank 1 both
printed shards were mislabeled: the data of (1,2) and (1,3) appeared under
coordinates (0,2) and (0,3).
Single-host meshes were unaffected, because every coordinate is local there and
the two containers coincide.
 #### Fix
Drive the loop from the host buffer's own coordinates and consume a buffer only
for coordinates that actually have a materialized shard, so a coordinate can
never be attached to another device's data.
This is sound because both device-to-host read-back paths build the host buffer
from the mesh view and record every requested coordinate in `shard_coords()`,
local and remote alike, while populating only the local ones. `get_shard()`
therefore returns a value exactly for the shards `apply()` used to visit, and
`std::set<MeshCoordinate>` order is lexicographic, which over a fixed-shape mesh
is the same total order as the row-major flat index that `apply()` used. The
sequence of printed buffers is unchanged; only the coordinate attached to each
one is now guaranteed correct. The row-major and dtype conversions applied before
printing go through `DistributedHostBuffer::transform()`, which forwards
`shard_coords()` and the local/remote partitioning verbatim.
Incidentally this also removes the print path's dependence on the ordering of
`DeviceStorage::coords_`, which has no sortedness invariant: the explicit-coords
constructors store the caller's vector as given.
Output is byte-for-byte unchanged on single-host meshes, so existing golden print
tests are unaffected. The `first` flag replaces the trailing-separator logic and
reproduces the old spacing exactly when no shard is skipped, while no longer
emitting blank lines for skipped shards.
 #### Tests
Add `BigMeshDualRankTest2x4.ToStringPairsEveryLocalShardWithItsOwnCoordinate` in
a new `unit_tests_dual_rank_2x4_to_string` target. It shards a tensor so each
device holds a value unique to it, prints the device tensor, and parses the
`device_id`/coordinate headers back out. Ground truth is taken per coordinate via
`get_shard()` on the read-back tensor, so the test depends on neither the
mapper's shard ordering nor the pairing logic under test. It asserts one printed
block per local shard and that each block's coordinate, device_id and payload
agree.
Note the test only reproduces the bug under a multi-rank launch. Run as a single
rank, every coordinate is local, the two containers coincide, and the old code
passes.

### Before fix, rank 1
<- blank line, skipped shard (0,0)
<- blank line, skipped shard (0,1)
device_id: D(0,2), MeshCoordinate([0, 2]) ttnn.Tensor([[[[106.0000, 106.0000, 106.0000, 106.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR) device_id: D(0,3), MeshCoordinate([0, 3]) ttnn.Tensor([[[[107.0000, 107.0000, 107.0000, 107.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR)

Two independent defects in six lines. The block labelled `(0, 2)` carries
`106`, which is the data of `(1, 2)`, and the block labelled `(0, 3)` carries
`107`, the data of `(1, 3)` — both mislabeled, under the wrong `device_id`. And
the shards that really live at `(0, 2)` and `(0, 3)`, payloads `102` and `103`,
never appear at all, as do the coordinates `(1, 2)` and `(1, 3)`. Half of the
rank's four local shards are silently gone, and the two blank lines are the
separators emitted for the shards that were skipped.

### After fix, rank 1
device_id: D(0,2), MeshCoordinate([0, 2]) ttnn.Tensor([[[[102.0000, 102.0000, 102.0000, 102.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR) device_id: D(0,3), MeshCoordinate([0, 3]) ttnn.Tensor([[[[103.0000, 103.0000, 103.0000, 103.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR) device_id: D(1,2), MeshCoordinate([1, 2]) ttnn.Tensor([[[[106.0000, 106.0000, 106.0000, 106.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR) device_id: D(1,3), MeshCoordinate([1, 3]) ttnn.Tensor([[[[107.0000, 107.0000, 107.0000, 107.0000]]]], shape=Shape([1, 1, 1, 4]), dtype=DataType::FLOAT32, layout=Layout::ROW_MAJOR)

All four local shards, each under its own coordinate, and `100 + 4r + c` matches
the coordinate in every header.
Rank 0 shows the omission without the mislabeling. Before, it printed only
`(0, 0)` → `100` and `(0, 1)` → `101`, correctly labelled by luck because its
local coordinates happen to be a prefix of the coordinate list, then emitted a
trailing blank line where `(1, 0)` → `104` and `(1, 1)` → `105` should have
been. After, all four appear.

[(T3K) T3000 unit tests](https://github.com/tenstorrent/tt-metal/actions/workflows/t3000-unit-tests.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:AlexeyZaytsev-epam/48267_bug_tensor_to_string_missmatches)


vtsilytskyiTT (commented):
Can you show how values were printed before and after change? Somewhere in PR description would be OK.

azaytsev-epam:
> Can you show how values were printed before and after change? Somewhere in PR description would be OK.

Added

bbradelTT (approved):

afuller-TT (approved):

vtsilytskyiTT (approved):
Thanks, looks good to me.

PR #48896, opened 2026-07-03
Functional SFPU implementation for mean

ign-sampanna:
### PR Category

Bug fix

### Summary

This PR fixes high ULP errors in `ttnn.mean` for `FLOAT32` reductions by introducing an accurate FP32 reduction path using the SFPU.

Previously, `FLOAT32` mean reductions used the FPU reduction path, which internally truncates FP32 inputs to TF32, resulting in large ULP errors for models such as LLaMA and DeepSeek.

This change:
- Routes `FLOAT32` mean through the SFPU reduction path for full FP32 accumulation.
- Lowers `mean` to `sum + 1/N` while preserving FP32 accuracy.
- Reuses the existing SFPU reduction infrastructure with architecture guards for Quasar.
- Updates the host post-processing path to correctly handle block-float (`BFLOAT8_B`/`BFLOAT4_B`) outputs.
- Preserves existing behavior for other reduction operations and data types.

Fixes #25779.

### Notes for reviewers

- **Scope**: Accurate FP32 mean only — `ttnn.mean` with `fast_and_approximate_mode=False` (the default). Other FP32 reductions (`sum`, `max`, etc.) stay on the existing FPU path.
- **Mechanism**: Host lowers AVG → SUM and applies `1/N` via post-mul; device runs FP32 SUM on the SFPU with `add_binary_tile` cross-tile accumulation and `UnpackToDestFp32` to avoid TF32 truncation.
- **Opt-out**: `fast_and_approximate_mode=True` restores the prior FPU (TF32) behavior for users who prefer speed over accuracy.
- **Limits**:
  - Quasar lacks `add_binary_tile`; host falls back to FPU and the kernel is guarded with `ARCH_QUASAR`.
- **Touch points**: `reduce_op.cpp` (routing/lowering), `common.hpp` (`use_sfpu_reduce_path` / `requires_post_mul`), W/H program factories (`REDUCE_SFPU_FP32`, `UnpackToDestFp32`), `reduce_helpers_compute.inl` (`add_binary_tile`), nanobind (`fast_and_approximate_mode` on `mean` only).
- **Tests**: `tests/ttnn/nightly/unit_tests/operations/reduction/test_mean_ulp.py` — `test_mean_fp32_fast_and_approximate_mode` compares accurate vs fast path ULP against an FP64 reference.

## Performance Results

[...]

PR #52718, opened 2026-08-10
[Feature] Add accurate fp32 SFPU path for ttnn.max via fast_and_approximate_mode

### LINKED ISSUE #51889: ttnn.max on float32 returns a value not present in the input (global reduce, no dim)
### Summary

`ttnn.max` with no `dim` (global reduce) returns a value that **is not present in the input tensor** when the input is `float32`. `max` is a selection operation, so its result must be one of the input elements.

On `bfloat16` the same op returns the exact element. Only `float32` is affected.

### Minimal repro

```python
import torch, ttnn

device = ttnn.open_mesh_device(mesh_shape=ttnn.MeshShape(1, 1))
torch.manual_seed(0)
src = torch.randn(1, 1, 64, 64, dtype=torch.float32)
tt = ttnn.from_torch(src, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)

got = ttnn.to_torch(ttnn.max(tt)).flatten()[0]      # global reduce: no `dim`
print("torch.max        :", torch.max(src).item())
print("ttnn.max         :", got.item())
print("present in input?", bool((src.flatten() == got).any()))
ttnn.close_mesh_device(device)
```

### Expected

```
present in input? True
```

### Actual

```
torch.max        : 4.1014933586120605
ttnn.max         : 4.09814453125
present in input? False
```

### Scope

6/6 float32 cases return a value absent from the input; 6/6 bfloat16 cases return the exact element.

| dtype | shape | seed | `torch.max` | `ttnn.max` | in input? | rel err |
|---|---|---|---|---|---|---|
| bfloat16 | (16, 2, 32, 3) | 0 | 4.093750000 | 4.093750000 | **True** | 0 |
| bfloat16 | (16, 2, 32, 24) | 1 | 3.796875000 | 3.796875000 | **True** | 0 |
| bfloat16 | (1, 1, 64, 64) | 0 | 4.093750000 | 4.093750000 | **True** | 0 |
| float32 | (16, 2, 32, 3) | 0 | 4.101493359 | 4.098144531 | **False** | 8.16e-04 |
| float32 | (16, 2, 32, 3) | 1 | 3.362807751 | 3.361572266 | **False** | 3.67e-04 |
| float32 | (16, 2, 32, 24) | 1 | 3.802687168 | 3.801269531 | **False** | 3.73e-04 |
| float32 | (1, 1, 64, 64) | 1 | 3.405680895 | 3.406005859 | **False** | 9.54e-05 |

The error magnitude (1e-4 … 8e-4) is consistent with bfloat16 precision, which suggests the reduce runs at reduced internal precision and returns the rounded value rather than the selected element.

A fuller script covering both dtypes and several shapes is at `repro_max_float32_selection.py` (exits 1 when reproduced).

### Why this isn't caught today

`tests/ttnn/unit_tests/operations/reduce/test_max.py::test_max_global` is the only test covering the no-`dim` path, and it uses **bfloat16** input with **`assert_equal`**:

```python
torch_input_tensor = torch_random((batch_size, h, w), -100, 100, dtype=torch.bfloat16)
torch_output_tensor = torch.max(torch_input_tensor)
output_tensor = ttnn.max(input_tensor)
assert_equal(torch_output_tensor, output_tensor)
```

That combination passes trivially, so the float32 path is never exercised.

### How it surfaced

Found in the model-traced sweeps, where FLOAT32 `max` vectors with `dim` absent reported `PCC 0.0` (shapes `(16,2,32,{3,6,12,24})`, e.g. vector `035bf99f081b`). PCC is undefined for a scalar result, so the comparison falls back to `torch.allclose`, and the discrepancy above exceeds its tolerance.

### Question for op owners

Is reduced-precision internal accumulation expected for `float32` reductions? If so, callers need to know `max` may return a value not in the input, which is surprising for a selection op and means exact-equality assertions are invalid on `float32`. If not, this looks like a genuine precision bug in the reduce path.

### Environment

- commit `97897756cde`
- `wormhole_b0`, N300 (2 chips), single-device mesh (1, 1)
- firmware bundle 19.7.1, KMD 2.5.0
- torch 2.11.0+cpu

ign-sudha:
### PR Category
- Feature

### Summary
Fixes #51889.

FLOAT32 `ttnn.max` could return a value that was not present in the input. The FPU/GMPOOL path truncates operands to TF32 before comparing, so the packed result can be a rounded value rather than an actual input element.

This mirrors #48896, `ttnn.mean`: add `fast_and_approximate_mode` to `ttnn.max`.
- `False` (default): accurate SFPU path at full float32 (selection must match the input element exactly).
- `True`: existing faster FPU/TF32 path (approximate).

Falls back to FPU when `fp32_dest_acc_en=False` or on Quasar. No effect for non-FLOAT32.

### Notes for reviewers

**Kernel** (`reduce_helpers_common.hpp`, `reduce_helpers_compute.inl`)
- FLOAT32 MAX is eligible for the accurate SFPU path.
- Cross-tile fold uses `binary_max_tile` (same pattern as FLOAT32 SUM / Int32 MAX).
- Within-face reload transpose stays FPU-only (`!is_sfpu`); SFPU does not use GMPOOL's face layout.

**Host** (`reduce_op.cpp`, `common.hpp`)
- Accurate path: FLOAT32 + `fast_and_approximate_mode=False` + `fp32_dest_acc_en` + non-Quasar.
- AVG and MAX share that eligibility; MAX also requires `!negate` (float MIN is still `-max(-x)` on fused-negate kernels with no SFPU path).

**Multi-axis / no-dim** (`generic_reductions.cpp`)
- No-dim and other multi-axis reduces go through `reduce_nd_loop`.
- `Max` forwards the caller's `fast_and_approximate_mode` (max-of-maxes is exact).
- `Mean` keeps FPU for those sub-steps (each axis applies its own `1/N`).

**API**
- `ttnn.max(..., fast_and_approximate_mode=False)` by default.
- Mean and max share `generic_reduction_with_fast_mode` in nanobind.

**Not in this PR**
- FLOAT32 `min` stays on FPU.
- Accurate path is slower; use `fast_and_approximate_mode=True` to opt out.

**Tests**
- `test_max.py`: FLOAT32 coverage on the three #51889 shapes, both mode values (`assert_equal` for accurate, `assert_allclose` for fast).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/sudha-max-fp32-sfpu)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/sudha-max-fp32-sfpu)

### DIFF
--- a/tests/ttnn/unit_tests/operations/reduce/test_max.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_max.py
@@ -10,7 +10,7 @@ import torch
 
 import ttnn
 from models.common.utility_functions import torch_random
-from tests.ttnn.utils_for_testing import assert_equal
+from tests.ttnn.utils_for_testing import assert_allclose, assert_equal
 
 TEST_PADDING_VALUE = -42
 
@@ -142,3 +142,31 @@ def test_max_dim(device, input_shape_and_dim, keepdim):
 
     # test for equivalance
     assert_equal(torch_output_tensor, output_tensor)
+
+
+@pytest.mark.parametrize("input_shape", [(16, 2, 32, 3), (16, 2, 32, 24), (1, 1, 64, 64)])
+@pytest.mark.parametrize("dim", [None, -1, -2])
+@pytest.mark.parametrize("fast_and_approximate_mode", [False, True], ids=["accurate", "fast"])
+def test_max_fp32_fast_and_approximate_mode(device, input_shape, dim, fast_and_approximate_mode):
+    """FLOAT32 max with both values of fast_and_approximate_mode.
+    - False (default): accurate SFPU path - result matches torch exactly.
+    - True: faster FPU/TF32 path - result is approximate.
+    """
+    torch.manual_seed(1)
+
+    torch_input_tensor = torch.randn(input_shape, dtype=torch.float32)
+    if dim is None:
+        torch_output_tensor = torch.max(torch_input_tensor)
+    else:
+        torch_output_tensor, _ = torch.max(torch_input_tensor, dim=dim)
+
+    input_tensor = ttnn.from_torch(torch_input_tensor, layout=ttnn.TILE_LAYOUT, device=device, dtype=ttnn.float32)
+    input_tensor = ttnn.fill_implicit_tile_padding(input_tensor, TEST_PADDING_VALUE)
+
+    output_tensor = ttnn.max(input_tensor, fast_and_approximate_mode=fast_and_approximate_mode, dim=dim)
+    output_tensor = ttnn.to_torch(ttnn.from_device(output_tensor)).reshape(torch_output_tensor.shape)
+
+    if fast_and_approximate_mode:
+        assert_allclose(torch_output_tensor, output_tensor, rtol=1e-3, atol=1e-2)
+    else:
+        assert_equal(torch_output_tensor, output_tensor)
--- a/tests/ttnn/unit_tests/operations/reduce/test_max.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_max.py
@@ -146,8 +146,9 @@ def test_max_dim(device, input_shape_and_dim, keepdim):
 
 @pytest.mark.parametrize("input_shape", [(16, 2, 32, 3), (16, 2, 32, 24), (1, 1, 64, 64)])
 @pytest.mark.parametrize("dim", [None, -1, -2])
+@pytest.mark.parametrize("scalar", [1.0, 2.5, -2.5])
 @pytest.mark.parametrize("fast_and_approximate_mode", [False, True], ids=["accurate", "fast"])
-def test_max_fp32_fast_and_approximate_mode(device, input_shape, dim, fast_and_approximate_mode):
+def test_max_fp32_fast_and_approximate_mode(device, input_shape, dim, scalar, fast_and_approximate_mode):
     """FLOAT32 max with both values of fast_and_approximate_mode.
     - False (default): accurate SFPU path - result matches torch exactly.
     - True: faster FPU/TF32 path - result is approximate.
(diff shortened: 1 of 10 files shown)


tests/ttnn/unit_tests/operations/reduce/test_max.py
```diff
@@ -142,3 +142,31 @@ def test_max_dim(device, input_shape_and_dim, keepdim):
 
     # test for equivalance
     assert_equal(torch_output_tensor, output_tensor)
+
+
+@pytest.mark.parametrize("input_shape", [(16, 2, 32, 3), (16, 2, 32, 24), (1, 1, 64, 64)])
+@pytest.mark.parametrize("dim", [None, -1, -2])
```
iwroszTT, line 148: Add a post-reduce scalar parameter (None, positive, negative)

ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op.cpp
```diff
@@ -124,10 +124,15 @@ Tensor reduce(
...
-    const bool use_sfpu_fp32_mean =
-        !fast_and_approximate_mode && input_tensor.dtype() == tt::tt_metal::DataType::FLOAT32 &&
-        reduce_math == tt::tt_metal::ReduceOpMath::AVG && arch != tt::ARCH::QUASAR && config.fp32_dest_acc_en;
+    // Accurate fp32 SFPU path (FPU truncates to TF32). Falls back to FPU without fp32_dest_acc_en or on Quasar.
+    const bool fp32_sfpu_eligible = !fast_and_approximate_mode &&
+                                    input_tensor.dtype() == tt::tt_metal::DataType::FLOAT32 &&
+                                    arch != tt::ARCH::QUASAR && config.fp32_dest_acc_en;
+
+    const bool use_sfpu_fp32_mean = fp32_sfpu_eligible && reduce_math == tt::tt_metal::ReduceOpMath::AVG;
+    const bool use_sfpu_fp32_max = fp32_sfpu_eligible && reduce_math == tt::tt_metal::ReduceOpMath::MAX && !negate;
```
iwroszTT, line 141: Why is negate excluded?

ign-sudha: This PR scopes the `fp32`→`SFPU` routing to `ttnn.max` via `fast_and_approximate_mode`. 

`negate=true` is not a real max - it is the float/bf16 min lowering. Non-Int32 min is rewritten as `-max(-x)` in `reduce_min()`, which calls `detail::reduce(..., MAX, ..., /*negate=*/true)`. That selects the negate kernels (`reduce_h_neg.cpp` / `reduce_w_neg.cpp`), which are on FPU.

So `!negate` keeps the accurate SFPU path for true fp32 max only. If we allowed negate, we would route the min lowering into the SFPU eligibility flags even though it still has to run on those FPU negate kernels.

Accurate `fp32` min (direct SFPU MIN, skipping `-max(-x)`) is intentionally out of scope for this PR; that is the follow-up in the all-ops work.

tests/ttnn/unit_tests/operations/reduce/test_max.py
ign-sudha: Added

ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op.cpp
ign-sudha: | Op | `negate` | Path today (this PR) | Notes |
|----|----------|----------------------|--------|
| `MAX` | `false` | Accurate SFPU (when eligible) | True `ttnn.max` |
| `MAX` | `true` | FPU only via `-max(-x)` | Float/bf16 `min` lowering (`reduce_{h,w}_neg`) |
| `MIN` | `false` | SFPU (Int32 today) | Direct LLK `MIN` |

Follow-up: accurate fp32 `min` will take the `MIN` + `negate=false` SFPU path (same as Int32).

iwroszTT (approved):

bbradelTT (approved):

s-jovic (approved):

PR #52719, opened 2026-08-10
[Bug fix] GPT-OSS: fix garbage output on multi-device Blackhole (trace allocation) (#52176)

### LINKED ISSUE #52176: [GPT-OSS-120B] Garbage output
Recently identified and has been going unnoticed since at least July 30: https://github.com/tenstorrent/tt-metal/actions/runs/30510488367/job/90770618918#step:6:384

TODO:
- Find first point of failure.
- Fix.
- Add a top-1/top-5 test to CI (should already exist somewhere)

mtairum:
## ✅ Status: validated on the rebased branch

Rebased onto latest main. All regressions this branch introduced have been found and fixed, and
GPT-OSS is verified clean on both Blackhole SKUs from #52176 — including `bh_galaxy` — **after** the
rebase.

| Check | Result |
|---|---|
| GPT-OSS-120B `bh_galaxy` | **clean** — `splits=8`, `On-device sampling unavailable` count 0 |
| GPT-OSS-120B `bh_quietbox_2` | clean |
| GPT-OSS-120B `wh_galaxy_perf` | clean |
| Llama-3.1-8B `bh_quietbox_2` / DP / llmbox / p300 | pass |
| Mistral-7B `wh_n150` | pass |
| Gemma-4-12B, Gemma-3-4B, Gemma-3-27B, Qwen VL family | pass |
| Tier 1 E2E | 20 passed |
| Tier 2 E2E | 24+ passed |

Runs: [Tier 1](https://github.com/tenstorrent/tt-metal/actions/runs/31811096809) ·
[Tier 2](https://github.com/tenstorrent/tt-metal/actions/runs/31811102778) ·
[bh_galaxy](https://github.com/tenstorrent/tt-metal/actions/runs/31818550602) ·
[Mistral](https://github.com/tenstorrent/tt-metal/actions/runs/31816528778)

Remaining red jobs in those sweeps are all pre-existing or infrastructure, itemised at the bottom.

## Ticket

Fixes #52176 — GPT-OSS garbage output on multi-device Blackhole. **Still reproducing on main** as of
2026-08-14 (`bh_quietbox_2` scheduled run), i.e. live for over two months.

## Attribution

The first four commits are **@ifabijanic-tt's work**, taken from `llama_8b_tracing_fix` and rebased
with authorship preserved. That branch had no PR. The remaining commits are follow-on fixes for
regressions the full sweeps surfaced. @ifabijanic-tt please take this over or say if you'd rather
close it in favour of your own.

## Root cause

On-device sampling intermittently returned an **out-of-range token ID** while a decode trace was
live. Measured by reading logits to host and comparing host argmax against the device-sampled token:

```
step=0   host_argmax=35644    device_sampled=35644        AGREE
step=1   host_argmax=200008   device_sampled=3214098612   DIVERGE
```

`vocab_size` is 201088, so `3214098612` is uninitialized memory read as `uint32`. The logits are
correct at every step — the sampling op's output buffer is what gets corrupted. One bad token fed
back gives an out-of-range embedding lookup that poisons the context, which is why the symptom looked
like gradual degradation rather than a single fault.

`2910e4aa1c8` addresses exactly this: buffers otherwise *"allocated during the decode loop, inside
this trace's scratch, and be overwritten on every replay"*.

## Rebase: how this interacts with main's `_mark_trace_buffers_corruptible`

Main has since grown its own mechanism in the same code, so it's worth stating how the two relate.
They are **complementary, not competing**, and both are preserved here:

- `_mark_trace_buffers_corruptible` is opt-in behind `_tt_allow_decode_trace_buffer_reuse`, and only
  `Qwen36ForCausalLM` sets it — for decode bucketing, where several traces stay live and buffer
  overlap is *deliberate*. It calls `ttnn.mark_corruptible` only if that attribute exists, and it does
  not exist in ttnn yet, so it is currently a no-op on hardware. It **annotates** intended overlap.
- This branch **prevents** long-lived buffers from being allocated inside a live trace's scratch in
  the first place. That is the actual #52176 fix, and it is why main's mechanism does not fix the bug.

Conflict resolutions (only Ivan's first commit conflicted; the other ten applied cleanly):

- kept main's `skip_precompile` parameter and guard around the compile pass, wrapped over the
  `compile_output` capture. `compile_output` is now `None` on the skip path, so the sampling
  pre-compile loop is guarded against it.
- kept the prealloc-reuse staging and re-added main's `_mark_trace_buffers_corruptible` call on the
  staged trace inputs.
- `capture_trace(..., skip_precompile=True)` is correct in both cases: either
  `_prepare_decode_trace_text` pre-compiled the sampling pipeline before any trace was live, or the
  caller passed `skip_precompile` and is asserting the cache is already warm.
- destructor cleanup is a union: main releases bucket-namespace sampling traces, this branch releases
  the post-prefill traces.

## Follow-on fixes in this PR

The restructuring touches shared code (`models/tt_transformers/tt/generator.py`,
`models/common/sampling/`), so it reaches every tt_transformers model. `model=all` sweeps found four
regressions, each fixed here:

1. **Trace capture** — compile pass ran against transient buffers, capture against persistent ones, so
   uncached programs loaded inside the capture window: `mesh_workload.cpp:153 !is_capturing_trace`,
   "Cannot load new binaries during trace capture". Hit Gemma-4. Fixed by warming the program cache
   over the persistent inputs before capture.
2. **Trace region** — the added post-prefill trace raises trace-region demand for every model.
   Mistral-7B `wh_n150` fell 2048 B per bank short; raised 25 MB → 32 MB.
   **Reviewers: any model close to its trace-region limit may need the same. The model/SKU matrix is
   wider than these sweeps cover.**
3. **Prefetcher, PREFILL mode** — the deferred capture restored `switch_mode(Mode.PREFILL)`, reaching
   `Prefetcher.init()`'s PREFILL branch, which has never executed on main (main only ever calls
   `switch_mode(Mode.DECODE)`) and is not functional.
4. **Prefetcher, sub-devices** — hoisting decode setup into the prefill phase is fundamentally
   incompatible with the prefetcher's per-mode sub-device managers: staying in DECODE leaves prefill
   running against decode sub-devices (`program.cpp:2205`, "Kernel group cores do not match sub device
   cores"). Gated the hoist off for prefetcher-driven models, which keep main's original path. Costs
   nothing this branch targets — the prefetcher is Blackhole-only, GPT-OSS does not use it, and the
   Llama-70B-Galaxy work here is Wormhole. **Scope limit worth knowing: Blackhole prefetcher models
   get none of the trace-allocation benefit.**

The prefetcher's dead `Mode.PREFILL` sub-device path is untouched and worth a separate fix by its
owner (unassigned `all_core_range_set`; supplying one trips a CB/L1 clash).

## Validation (pre-rebase)

| Check | Result |
|---|---|
| GPT-OSS-120B `bh_galaxy` | clean ×2 — the SKU from the original report |
| GPT-OSS-120B `bh_quietbox_2` | clean ×5 |
| GPT-OSS-120B `wh_galaxy_perf` | clean — `batch128` + `prefill_128`, no regression |
| GPT-OSS-20B (p150 / quietbox / llmbox) | pass |
| Llama-3.1-8B (quietbox, DP, llmbox, p300) | pass |

Every clean GPT-OSS run was checked against a silent fallback: `On-device sampling unavailable` count
0, `On-device sampling initialized (vocab_size=201088, splits=4)` present. These are the genuinely
failing configuration — traced decode **and** on-device sampling — now producing correct output.

## Remaining failures, all pre-existing or infrastructure

- `Gemma-4-12B [bh_quietbox_2]` — `PCC too low: 0.9378978280024141`, byte-identical on main
- `Gemma-4-31B [wh_llmbox_perf]` — job timeout; fails on main in 3 of 4 recent scheduled runs
- `Shallow UNet`, TT-DiT Flux / LTX / Wan, SDXL — import none of the changed modules
- `Llama-3.3-70B Galaxy` — read-only `/mnt/MLPerf` cache write; fails identically on main
- Various galaxy jobs — `TT_FATAL: requested_size <= system_size` (4 devices for a 32-device mesh)

Infrastructure noise was heavy: container-startup faults (`Value cannot be null (ContainerId)`),
`HttpClient.Timeout`, 20-minute checkout timeouts, and one runner (`qb2-120-p06t02`) that failed nine
jobs across two sweeps. Worth attention independently of this PR — especially the `bh_galaxy` label,
which repeatedly schedules 32-device jobs onto 4-device hosts, and which is part of why this bug went
unnoticed for eight weeks.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/52176-gpt-oss-bh-trace-alloc-fix)

### DIFF
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -544,6 +544,12 @@ def format_sampling_params(sampling_params, max_batch_size):
             # Device sampling treats p=0 as a first-token cutoff; with k=1
             # this is the compact argmax representation for greedy rows.
             top_p[i] = 0.0
+            # A greedy row has a single candidate and never draws from the RNG, so a seed on it cannot
+            # change the sampled token. Clear it: otherwise a caller-supplied default (seed=0 rather
+            # than None is enough) marks the request "explicitly seeded", which forces sampling off its
+            # captured trace and re-runs the whole sampling graph every token -- allocating device
+            # buffers behind live traces on every step, where they can be corrupted on replay.
+            seed[i] = None
         else:
             temperature[i] = 1 / temperature[i]
 
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -311,7 +311,48 @@ class SamplingGenerator:
         if penalties_on:
             logits = self.tt_penalties.apply(logits)
         tt_tokens, tt_log_probs = self.tt_sampling(logits, tt_out_tok=tt_out_tok)
+        if penalties_on:
+            # Fold the penalty bookkeeping into the sampled step rather than running it afterwards in
+            # sample(). The order is unchanged -- penalties are applied to this step's logits from the
+            # previous steps' counts, then the new token is counted -- but doing it here means it is part
+            # of whatever trace captures this, instead of a handful of scatter/tilize/reshape allocations
+            # on every decode step behind a live trace. Those ops take no preallocated output tensor, so
+            # tracing them is the only way to stop them allocating.
+            self.tt_penalties.update_output_tokens(tt_out_tok if tt_out_tok is not None else tt_tokens)
         return tt_tokens, tt_log_probs
+
+    def reset_penalty_counts(self):
+        """Zero the output-token penalty counters, if penalties are active.
+
+        Needed after any pass that runs the sampling pipeline for its side effects rather than its result:
+        _run_sampling now counts the sampled token, so a pre-compile or a trace capture over dummy inputs
+        would otherwise leave those phantom tokens in the counts and skew the next real step's penalties.
+        In-place, so it allocates nothing.
+        """
+        if self._penalties_active:
+            self.tt_penalties.reset_output_tokens()
+
+    def precompile(
+        self,
+        logits: ttnn.Tensor,
+        *,
+        tt_out_tok: Optional[ttnn.Tensor] = None,
+    ) -> None:
+        """Run the sampling pipeline once without capturing, to compile it and size its scratch.
+
+        This is the pre-compile step :meth:`capture_trace` would otherwise do inline. Callers that capture
+        the sampling trace behind another trace (e.g. right after the decode trace) should run it earlier,
+        while no trace is live on device, and then pass ``skip_precompile=True`` to :meth:`capture_trace`;
+        left inline, this pass allocates device buffers that a live trace can corrupt on replay.
+
+        ``logits`` only has to match the spec of the tensor that will later be captured, not be it.
+        """
+        self._run_sampling(
+            logits,
+            penalties_on=self._penalties_active,
+            tt_out_tok=tt_out_tok,
+        )
+        self.reset_penalty_counts()
 
     def capture_trace(
         self,
@@ -347,6 +388,7 @@ class SamplingGenerator:
         )
         ttnn.end_trace_capture(self.mesh_device, trace_id, cq_id=self.cq_id)
         ttnn.synchronize_device(self.mesh_device)
+        self.reset_penalty_counts()
 
         if tt_out_tok is not None:
             if isinstance(sampled, tuple):
@@ -412,11 +454,8 @@ class SamplingGenerator:
             self._validate_trace_inputs(slot, logits, tt_out_tok)
             tt_out = self._execute_trace(key)
 
-        if penalties_on and tt_out is not None:
-            if isinstance(tt_out, tuple):
-                self.tt_penalties.update_output_tokens(tt_out[0])
-            else:
-                self.tt_penalties.update_output_tokens(tt_out)
+        # The penalty update now runs inside _run_sampling, so it is captured with the rest of the sampled
+        # step and replayed with it -- there is nothing to do here.
         return tt_out
 
 
--- a/models/common/sampling/tt_penalties.py
+++ b/models/common/sampling/tt_penalties.py
@@ -145,6 +145,9 @@ class TTPenalties(LightweightModule):
             host=torch.ones(self._total_batch, 1), shard_dims=shard_dims_gathered, layout=ttnn.ROW_MAJOR_LAYOUT
         )
         self.zeros = self._alloc_int_buffer(shard_dims=shard_dims_gathered, layout=ttnn.ROW_MAJOR_LAYOUT)
+        # All-ones scatter sources for the non-fast update path, cached by token width. See
+        # update_output_tokens: allocating these per call put device buffers behind live traces.
+        self._scatter_src_by_width = {}
         self.presence_penalties = self._alloc_bf16_buffer(shard_dims=shard_dims_bf16)
         self.frequency_penalties = self._alloc_bf16_buffer(shard_dims=shard_dims_bf16)
         self.repetition_penalties = self._alloc_bf16_buffer(shard_dims=shard_dims_bf16)
@@ -294,11 +297,18 @@ class TTPenalties(LightweightModule):
             new_tokens = ttnn.reshape(new_tokens, [batch, 1], **self._op_kwargs)
             src = self.decode_src
         else:
-            src = self._alloc_int_buffer(
-                host=torch.ones(self._total_batch, new_tokens.shape[-1]),
-                shard_dims=self._shard_dims_gathered,
-                layout=ttnn.ROW_MAJOR_LAYOUT,
-            )
+            # Cache per width: this used to allocate a fresh all-ones buffer on every call, which on the
+            # prefill-sampling path lands after the trace captures and is exactly the kind of allocation a
+            # live trace can corrupt. Its contents are constant, so one buffer per width is enough.
+            width = int(new_tokens.shape[-1])
+            src = self._scatter_src_by_width.get(width)
+            if src is None:
+                src = self._alloc_int_buffer(
+                    host=torch.ones(self._total_batch, width),
+                    shard_dims=self._shard_dims_gathered,
+                    layout=ttnn.ROW_MAJOR_LAYOUT,
+                )
+                self._scatter_src_by_width[width] = src
         self.token_bin_counts_and_mask(
             new_tokens=new_tokens,
             counts=self.output_counts_gathered,
--- a/models/demos/gpt_oss/tt/model.py
+++ b/models/demos/gpt_oss/tt/model.py
@@ -148,6 +148,10 @@ class Model:
         self.cos_matrix = self.rope_setup.cos_matrix
         self.sin_matrix = self.rope_setup.sin_matrix
         self.transformation_mats = self.rope_setup.get_both_trans_mats()
+        # Prefill rope slices, keyed by sequence length. See prepare_inputs_prefill: the slice is a device
+        # op, so recomputing it per call allocates -- and on a traced prefill it allocates with the trace
+        # live, then throws the result away (the replay uses the matrices bound at capture time).
+        self._prefill_rope_slices = {}
 
         if state_dict:
             embedding_weight = substate(state_dict, "model.embed_tokens")["weight"]
@@ -598,6 +602,21 @@ class Model:
 
         return logits
 
+    def slice_last_token_block(self, x, last_token_idx, out=None):
+        """Slice out the 32-token tile containing ``last_token_idx``.
+
+        ``out`` writes into a caller-owned tensor instead of allocating a fresh one. The generator relies
+        on that to run this slice with a trace live on device: the tile is the input buffer of the traced
+        post-prefill tail, so the slice must not allocate (see Generator._run_post_prefill).
+        """
+        get_last_token = (last_token_idx // 32) * 32
+        return ttnn.slice(
+            x,
+            (0, 0, get_last_token, 0),
+            (1, 1, get_last_token + 32, x.shape[-1]),
+            output_tensor=out,
+        )
+
     def process_logits_after_prefill_trace(self, logits, last_token_idx):
         """
         Post-process traced prefill output to the 32-token tile containing `last_token_idx`.
@@ -605,13 +624,7 @@ class Model:
         Unlike tt_transformers `Transformer`, GPT-OSS `ttnn_prefill_forward` already
         applies final norm + lm_head, so this method only slices logits.
         """
-        get_last_token = (last_token_idx // 32) * 32
-        logits = ttnn.slice(
-            logits,
-            (0, 0, get_last_token, 0),
-            (1, 1, get_last_token + 32, logits.shape[-1]),
-        )
-        return logits
+        return self.slice_last_token_block(logits, last_token_idx)
 
     def prepare_row_sharded_prefill_iter(
         self, tokens, page_table, prompt_lens, iter_idx, max_padded_len, max_num_blocks, users_per_row_per_iter=1
@@ -709,8 +722,20 @@ class Model:
         model_args=None,
         trace_cache=None,
         empty_slots=None,
+        before_capture=None,
     ):
         """Row-parallel batched prefill: 1 user per row per iteration.
+
+        ``before_capture``: optional callback invoked once, immediately before the trace capture and after
+        this path's own trace inputs are allocated. It is the last point at which a caller can allocate
+        long-lived buffers with no trace live; the generator uses it to stage the decode trace's inputs,
+        which would otherwise be allocated during the decode loop, inside this trace's scratch, and be
+        overwritten on every replay.
+
+        Allocation only -- do not compile from here. Compiling a decode program at this point makes the
+        capture below deadlock in the MoE dispatch, and compiling one before any prefill has run deadlocks
+        in the MoE combine's global-semaphore setup instead. That is why the rest of decode trace setup
+        still happens later, in the decode loop, with this trace already live.
 
         ``empty_slots``: optional list of decoder slot ids, one per input user.
         If supplied, users are first reordered so that array-index ``i`` lands
@@ -843,6 +868,11 @@ class Model:
                 ho = self.prepare_inputs_prefill(t0, page_table=p0, trace_enabled=True, batched_prefill=True)
                 hi = (ho[0], ho[3], ho[4])
                 di = copy_host_to_device(hi, mesh_device=mesh_device)
+                # Last moment at which a caller can allocate long-lived buffers with no trace live.
+                # Deliberately after ``di``: the trace binds to those buffers, and letting the callback
+                # allocate first would move them off the addresses the compile pass just used.
+                if before_capture is not None:
+                    before_capture()
                 tid = ttnn.begin_trace_capture(mesh_device, cq_id=0)
                 tr = self.transform_and_embed_prefill_inputs_device(*di)
                 tt_out = self.ttnn_prefill_forward(
@@ -1096,12 +1126,18 @@ class Model:
             if len(tokens_embd.shape) == 3:
                 tokens_embd = ttnn.unsqueeze_to_4D(tokens_embd)
 
-        # Prepare rotation matrices (slice from rope_setup like tt-transformers model.py lines 156-159)
+        # Prepare rotation matrices (slice from rope_setup like tt-transformers model.py lines 156-159).
+        # Cached per sequence length: the slice runs on device, so recomputing it allocates. On a traced
+        # prefill that allocation happens with the trace live -- and is then discarded, since the replay
+        # uses the matrices that were bound when the trace was captured.
         seq_len = self.args.max_seq_len if trace_enabled else tokens_embd.shape[-2]
-        rot_mats_global = [
-            self.rope_setup.cos_matrix_prefill[:, :, :seq_len, :],
-            self.rope_setup.sin_matrix_prefill[:, :, :seq_len, :],
-        ]
+        rot_mats_global = self._prefill_rope_slices.get(seq_len)
+        if rot_mats_global is None:
+            rot_mats_global = [
+                self.rope_setup.cos_matrix_prefill[:, :, :seq_len, :],
+                self.rope_setup.sin_matrix_prefill[:, :, :seq_len, :],
+            ]
+            self._prefill_rope_slices[seq_len] = rot_mats_global
         rot_mats_local = None
 
         # Prepare page tables if provided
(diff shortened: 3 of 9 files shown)


mtairum:
## CI validation

Dispatched on this branch via `models-t1-e2e-tests.yaml` (all SKUs per model):

- GPT-OSS-120B: https://github.com/tenstorrent/tt-metal/actions/runs/31392032377
- Llama-3.1-8B: https://github.com/tenstorrent/tt-metal/actions/runs/31392035770

Earlier validation runs on equivalent branches (same commits, pre-rebase and two-commit subset):

| Run | Content | Result |
|---|---|---|
| [31186076704](https://github.com/tenstorrent/tt-metal/actions/runs/31186076704) | all 4 commits, GPT-OSS-120B `bh_quietbox_2` | clean |
| [31190537570](https://github.com/tenstorrent/tt-metal/actions/runs/31190537570) | all 4 commits, all SKUs | `bh_quietbox_2` clean; `bh_galaxy` cancelled (no runner); `wh_galaxy_perf` undersized runner |
| [31197163136](https://github.com/tenstorrent/tt-metal/actions/runs/31197163136) | 2-commit subset, GPT-OSS-120B `bh_quietbox_2` | clean |
| [31197165711](https://github.com/tenstorrent/tt-metal/actions/runs/31197165711) | all 4 commits, `wh_galaxy_perf` | clean — `batch128` + `prefill_128` |
| [31264223325](https://github.com/tenstorrent/tt-metal/actions/runs/31264223325) | rebased, GPT-OSS-120B | `bh_quietbox_2` clean |
| [31264225911](https://github.com/tenstorrent/tt-metal/actions/runs/31264225911) | rebased, Llama-3.1-8B | `bh_p150` + `wh_n150` pass; DP-Galaxy undersized runner |
| [31264224886](https://github.com/tenstorrent/tt-metal/actions/runs/31264224886) | rebased, Llama-3.3-70B-Galaxy | read-only NAS cache write — [fails identically on main](https://github.com/tenstorrent/tt-metal/actions/runs/31235603679) |

Reproducer for anyone re-checking: `pytest models/demos/gpt_oss/demo/text_demo.py -k prefill_128` on `bh_quietbox_2`, ~15 min end to end. Before these commits it emits token salad; after, coherent output with on-device sampling active.

Caveat repeated from the description: **no 32-device path has been exercised.** Every galaxy-class job either got cancelled waiting for a runner or died at device open with `requested_size <= system_size`. That is an infrastructure gap, not a property of this change, but it does mean `bh_galaxy` — the SKU in the original report — is unverified.

mtairum:
## Blast radius and full CI sweep

Two of the seven changed files are **shared infrastructure**, so this is much wider than GPT-OSS:

| Changed file | Scope |
|---|---|
| `models/tt_transformers/tt/generator.py` (+870) | every tt_transformers-based model |
| `models/tt_transformers/tt/model.py` | every tt_transformers-based model |
| `models/common/sampling/generator.py` | every model using on-device sampling |
| `models/common/sampling/tt_penalties.py` | every model using on-device sampling |
| `models/demos/gpt_oss/tt/model.py` | GPT-OSS only |
| `models/demos/llama3_70b_galaxy/tt/generator.py`, `demo/text_demo.py` | Llama-70B-Galaxy only |

Consumers of the two shared modules, by directory:

`gpt_oss`, `gpt_oss_d_p`, `llama3_70b_galaxy`, `tt_transformers` (so Llama-3.1-8B and friends), `gemma4`, `multimodal/gemma3`, `blackhole/qwen36`, `qwen25_vl`, `qwen3_vl`, `wormhole/qwen3_embedding_8b`, `deepseek_v3`, `minimax_m3`, `experimental/llama32_1b_quasar`.

Note `qwen25_vl` and `qwen3_vl` matter specifically: they are the two models that already set `_tt_disable_sampling_trace` + `_tt_vllm_always_refresh_decode_trace_inputs` for a related trace-allocation failure, so they exercise nearby code paths.

### Sweeps dispatched on this branch (`model=all`, `sku=all`)

- (Tier 1) Models E2E: https://github.com/tenstorrent/tt-metal/actions/runs/31402892028
- (Tier 2) Models E2E: https://github.com/tenstorrent/tt-metal/actions/runs/31402895078
- (Tier 1) Models Unit: https://github.com/tenstorrent/tt-metal/actions/runs/31402898862
- (Tier 2) Models Unit: https://github.com/tenstorrent/tt-metal/actions/runs/31402901938

Tier 2 picks up GPT-OSS-20B and the Qwen-VL models; the unit sweeps pick up the sampling unit tests (`models/demos/gpt_oss/tests/unit/test_sampling.py`, `models/common/tests/test_sampling.py`, and the Llama-70B sampling unit tests).

I will post results here. Expect some failures unrelated to this change — the known ones so far are undersized galaxy runners (`TT_FATAL: requested_size <= system_size`) and read-only `/mnt/MLPerf` tensor-cache writes; both reproduce on plain main.

mtairum:
## ⚠️ Full-sweep triage: this PR regresses other tt_transformers models — not ready to merge

The four sweeps finished. Comparing each failing job against the same job in today's scheduled main runs (baselines: 31453445386, 31455834844, 31458779844, 31449433959), most failures are pre-existing. **But some jobs that pass on main fail here**, and two have been verified as genuine regressions caused by these commits.

### Verified regressions

**1. Trace capture — cannot load binaries during capture**

`Gemma-4-12B e2e tests [bh_quietbox_2]` (main = **success**, here = failure), in `test_packed_verify_matches_sequential[blackhole-1x4]`:

```
TT_FATAL @ tt_metal/distributed/mesh_workload.cpp:153: !is_capturing_trace
info: Cannot load new binaries during trace capture. This program is not yet in program cache.
      Warm up before capturing a trace.
```

This is the exact hazard `901737e5cd5` warns about in its own comment — *"Allocation only -- do not compile from here. Compiling a decode program at this point makes the capture below deadlock in the MoE dispatch"*. The `before_capture` staging moves work relative to trace capture, and on Gemma-4 a program that was previously warmed up before capture is now compiled during it.

Same-signature failures, same baseline-passing status, not yet individually confirmed: `Gemma-4-31B [bh_quietbox_2]`, `Qwen3.6-27B [bh_quietbox_2]`, `Gemma-4-31B [wh_llmbox_perf]`, `Qwen2.5-32B [bh_quietbox_2]`, `Qwen2.5-VL-72B [bh_quietbox_2]`, `Qwen3-32B (Blackhole) [bh_quietbox_2]`, `Llama 3.1-8B [bh_quietbox_2]`, `Llama 3.3-70B [bh_quietbox_2]`.

**2. Allocator assertion on a different machine**

`Mistral-7B e2e tests [wh_n150]` (main = **success**, here = failure), in `test_demo_text[...eval-32]`:

```
TT_FATAL @ tt_metal/impl/allocator/bank_manager.cpp:462: false
```

Different SKU and different arch from the trace failures, so this is not one unhealthy machine. Consistent with the earlier staging of long-lived buffers changing the allocation profile.

### Not caused by this PR

- `bh_galaxy`, `wh_galaxy_perf`, DP-Galaxy, `bh_sc1`, `bh_sc4-mgd` jobs — fail on main too (undersized runners, `requested_size <= system_size`)
- Llama-70B-Galaxy and Llama-8B DP — read-only `/mnt/MLPerf` tensor-cache writes, fail identically on main
- `TT-DiT Flux.1 schnell [bh_quietbox_2]` — read-only cache; TT-DiT does not use `tt_transformers` or `models/common/sampling`, so it cannot be affected by this change

### Where this leaves the PR

The GPT-OSS fix itself is sound — four clean `bh_quietbox_2` runs with on-device sampling active, and the root cause is confirmed. But the trace-allocation restructuring in `models/tt_transformers/tt/generator.py` needs warm-up handling for the other models that share it before this can land. My earlier validation only covered GPT-OSS and Llama-3.1-8B, which is why I missed this; the `model=all` sweeps are what surfaced it.

@ifabijanic-tt this is your area — the `before_capture` hook needs the affected models warmed up before capture, or the staging needs to be opt-in per model. Happy to run more sweeps against any revision.

I am **not** recommending merge in the current state.

mtairum:
## Green — all four regressions closed

Superseding my earlier "not ready to merge" comment. The final gate fix landed and the sweeps are clean.

**Tier 2** ([31745844955](https://github.com/tenstorrent/tt-metal/actions/runs/31745844955)): **32 passed**, 2 failures — `Gemma-4-31B [wh_llmbox_perf]` (timeout, fails on main 3 of 4 recent runs) and `Shallow UNet [bh_p150]` (imports none of the changed modules).

`Llama 3.1-8B [bh_quietbox_2]` — the job that failed four different ways as each layer was peeled back — now **passes**, along with every other Llama-3.1-8B and GPT-OSS-20B variant.

**Tier 1** ([31738847902](https://github.com/tenstorrent/tt-metal/actions/runs/31738847902)): GPT-OSS-120B clean on `bh_quietbox_2` **and `bh_galaxy`**, the SKU from the original report, which took several attempts to get onto a correctly sized runner.

The last fix needed two commits because my first attempt gated only `_prepare_decode_trace_once`, while `_prepare_decode_trace_for_warmup` has a second caller that bypassed it. Gate now lives inside the function so both paths are covered.

### Notes for reviewers

- The trace-region bump (Mistral 25 MB → 32 MB) is a **general property of adding a trace**, not specific to Mistral. Any model near its limit will need the same, and the model/SKU matrix is larger than these sweeps cover.
- The prefetcher exclusion means Blackhole prefetcher-driven models keep main's non-hoisted path and get none of the trace-allocation benefit. That is deliberate — the two are incompatible — but it is a scope limit worth knowing.
- `Prefetcher.init()`'s `Mode.PREFILL` branch is dead and non-functional (unassigned `all_core_range_set`; supplying one trips a CB/L1 clash). Untouched here, worth a separate fix by its owner.

mtairum:
## Post-rebase validation complete — including `bh_galaxy`

Rebased onto latest main and re-validated. The last gap is now closed: **GPT-OSS-120B on `bh_galaxy` is clean after the rebase** ([run](https://github.com/tenstorrent/tt-metal/actions/runs/31818550602), `g13blx01`), with device sampling genuinely active — `On-device sampling initialized (vocab_size=201088, splits=8)` and `On-device sampling unavailable` count 0.

```
<|channel|>analysis<|message|>The user asks: "How many moons does Earth have?" Straightforward:
Earth has one natural satellite, commonly called the Moon. Some might mention quasi-satellites
like 2020 CD3, but generally one.
```

### One extra fix the rebase required

My conflict resolution kept main's `skip_precompile` threading but left it off the wrapper signature after Ivan's prepare/record split, giving `Generator._capture_decode_trace_text() got an unexpected keyword argument 'skip_precompile'`. That took out six models across two arches (Llama-3.1-8B, Mistral-7B, Gemma-3-4B, Gemma-3-27B, Qwen2.5-VL-72B, Qwen3-VL-32B) — a good illustration of this shared code's reach. Fixed by forwarding the parameter, and verified with an AST pass that no other `self.<method>(...)` call in the touched files passes a keyword its signature rejects.

### Infrastructure notes for whoever owns the fleet

- **`g07blx03` (bh_galaxy) needs attention.** Five consecutive GPT-OSS failures there, all at pytest collection: `TLB window allocation failed ... error code -12 (ENOMEM) ... likely caused by a stale or hung process from a previous run still holding the TLB resources`. `g13blx01` passes the same job. Three parallel dispatches were needed before one landed on the healthy host.
- Also seen across these sweeps, none related to this change: container-startup faults (`Value cannot be null (ContainerId)`), `HttpClient.Timeout`, 20-minute checkout timeouts, `codeload` archive-download failures, job-level timeouts, and one runner (`qb2-120-p06t02`) that failed nine jobs across two sweeps.
- `bh_galaxy` repeatedly schedules 32-device jobs onto 4-device hosts (`requested_size <= system_size`), and GPT-OSS-120B `bh_galaxy` **unit** tests still ERROR at setup for that reason. That gap is part of why #52176 went unnoticed for eight weeks and is worth a ticket of its own.

@ifabijanic-tt this is ready for your review — the first four commits are yours, rebased with authorship preserved.

mtairum:
## One thing for reviewers to decide: Qwen2.5-VL-72B trips its perf **upper** bound

Final Tier 2 sweep ([31811102778](https://github.com/tenstorrent/tt-metal/actions/runs/31811102778)) came back **31 passed**. All six failures are pre-existing or infrastructure, but one deserves a decision rather than a dismissal.

`Qwen2.5-VL-72B [wh_llmbox_perf]`: the tests **passed** (`2 passed`). The job failed in step 9, "Report and Validate Perf and Accuracy targets", because throughput came in *above* the allowed range:

```
decode_t/s:   measured=434.1392716522318  > upper_bound
decode_t/s/u: measured=13.566852239132244 > upper_bound
PerfRegressionWarning: Perf drift detected against expected metrics
  {'prefill_time_to_first_token': 1.163, 'decode_t/s/u': 10.x}
```

Measured 13.57 t/s/u against an expected ~10.x — roughly 30% faster than the recorded target.

That is plausibly this change *helping*: Qwen2.5-VL goes through the same shared generator, and the point of the trace-allocation work is to stop redundant allocation under a live trace. But I have not verified that, and the alternative — a measurement artifact, or the model taking a different path than before — is equally consistent with one data point.

Either way it is not a correctness problem, and it is not a reason to hold the PR. It does mean **this model's perf targets in `models/model_targets.yaml` will need updating** if the speedup is real, otherwise the job will keep failing its upper bound on every scheduled run after this merges.

I have deliberately not changed the targets myself: adjusting a perf bound to make a gate pass is the sort of thing that should be a conscious decision by someone who can confirm the speedup is genuine, not a side effect of the PR that caused it.

### Remaining red jobs in the final sweeps, for completeness

| Job | Cause |
|---|---|
| Qwen2.5-32B, Qwen3-32B, Qwen2.5-Coder-32B `[bh_quietbox_2]` | job-level timeouts before any test, three different hosts |
| Gemma-4-31B `[wh_llmbox_perf]` | pre-existing timeout; fails on main in 3 of 4 recent scheduled runs |
| Mistral-7B `[wh_n150]` | `codeload` archive-download timeout — **passed on targeted retry** ([31816528778](https://github.com/tenstorrent/tt-metal/actions/runs/31816528778)) |
| GPT-OSS-120B `[bh_galaxy]` | TLB exhaustion on `g07blx03`; **passed on `g13blx01`** ([31818550602](https://github.com/tenstorrent/tt-metal/actions/runs/31818550602)) |

mtairum:
## Rebased onto latest main + Tier 1/2/3 e2e sweep for trace-region coverage

Rebased (was 107 behind, zero conflicts, all 13 commits applied with @ifabijanic-tt's authorship intact) and dispatched all three e2e tiers, `model=all` / `sku=all`:

| Pipeline | Run |
|---|---|
| (Tier 1) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32127333188 |
| (Tier 2) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32127336097 |
| (Tier 3) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32127339773 |

### Why Tier 3 is included this time

I audited trace-region headroom across `models/model_trace_region_sizes.yaml`, because the post-prefill trace this PR adds raises trace-region demand for **every** tt_transformers model — Mistral-7B `wh_n150` needed 25 MB → 32 MB, and that is a general property of adding a trace, not something specific to Mistral.

The audit: **114 model/SKU entries, 112 affected by this change, and only 32 were exercised green by my earlier Tier 1+2 sweeps. 80 were unverified.** Tier 3 is where most of the gap lives, including the group with the least headroom:

| Region | Model / SKU |
|---|---|
| dynamic (0) | `deepseek-v3` × `wh_llmbox_perf`, `wh_galaxy_perf`, `p300x2` |
| 10.0 MB | `llama3.2-1b`, `llama3.2-3b`, `phi-3-mini` (`wh_n150`); `qwen2.5-7b`, `qwen2.5-vl-7b` (`wh_n300`) |
| 14.2 MB | `llama3.1-70b-legacy` `wh_llmbox_perf` |
| 17.4 MB | `llama3.2-11b-vision` `wh_llmbox_perf` |
| 21.4 MB | `gemma-3-27b-vision` `wh_n150` |
| 30.0 MB | `gpt-oss-20b` (`wh_n150`, `p300x2`), `gemma-4-e2b`, `gemma-3-27b` ×3, `mistral-small-3.1-24b` |

The signature to look for is unambiguous and distinguishable from CI noise:

```
TT_FATAL Out of Memory: Not enough space to allocate <N> B TRACE buffer across <M> banks
```

Any model showing it needs its `trace_region_size` raised the way Mistral's was.

### Caveats

- Tier 3 has never run on this branch, so I have no baseline for which of its jobs are already red on main. I will compare against main's scheduled Tier 3 runs before attributing any failure to this PR.
- Even all three tiers will not cover the full 112-entry matrix — several SKUs in that YAML (`p300x2`, `wh_n300`, some galaxy variants) are not exercised by these pipelines. This narrows the gap substantially rather than closing it.

### Also in this push

`_uses_prefetcher()` now emits a one-shot `logger.info` when it skips the hoisted decode trace setup. The prefetcher exclusion was otherwise a silent capability regression: Blackhole prefetcher-driven models keep main's non-hoisted path and get none of the trace-allocation benefit, with nothing in the logs saying so.

Issue #47554, opened 2026-06-19
[LLK] Test All Experimental LLKs


ncvetkovicTT:
Test all experimental LLKs, both the ones from tt-metal experimental folder and the ones migrated from tt-blaze.

ncvetkovicTT:
Some items that are the part of migration/contract enforcement:
* SFPU Layer cleanup
* Dest bit 11 removal
* Some other state that doesn't follow, like strides, disable_src_zero_flags
* layer violation fixups

PR #52720, opened 2026-08-10
[LLK][test] Advance-test scaffolds for 8 demo-fork experimental LLKs (Blackhole)

ncvetkovicTT:
## What
**Advance-test scaffolds** for 8 deepseek demo-fork experimental LLKs (SDPA/matmul family), pinned to the **primitive symbols** against the demo-fork headers (`models/demos/deepseek_v3_b1/kernel_includes/`) via a shadow `-I`.

`custom_mm`, `compressed_custom_mm`, `sdpa_custom_mm`, `sdpa_custom_mm_reuse_dest_srcb`, `sdpa_bcast_col_srcb_reuse`, `sdpa_bcast_col_srca_srcb_reuse`, `unpack_A_sdpa`, `sdpa_reduce_row`.

## Why
Decouple test coverage from the tt-blaze migration. Every one of these was diffed against tt-blaze `main`: **0 numeric forks** — the only divergences are FPU↔SFPU signalling cadence (`signal_granularity`/`input_granularity`/`output_granularity`), which collapse to the demo behavior at granularity=1 and are orthogonal to a numeric golden. So they're **internal-move candidates** (demo → `experimental/`), no blaze dependency for correctness.

## Status
- **On-device verified on a p100a (Blackhole): all 8 files green, 200/200 hardware variants** (80 on the merge gate, the rest the nightly `compressed_custom_mm` grid). Run with no marker filter, so the nightly `compressed_custom_mm` full grid (120 cases over 3 BFP formats) is included in that count. The most layout-sensitive goldens are negative-controlled (perturb the golden → the test fails), so nothing passes vacuously.
- Silicon changed real things in three of the eight files and corrected the in0 L1 layout in three more — see the two status comments below, which record the DEST/face-layout facts no header states. Landed in `76126f77f`, `f047dcb3e`, `cd950f78b`, `ff794aab4` and `ad651a47a`, with `2526ede69f` and `eb5c83ea3a` as no-behaviour-change cleanups on top (one definition each of the face row dim and the K-aware matmul atol).
- Open before this is worth reviewing: the `mask_chunk` coverage gap on `sdpa_custom_mm` (documented in the test, tracking issue still to be filed) and a second opinion on the 36 dropped `in0_rows` variants on the two sdpa matmuls.

## At promotion (per LLK)
- Delete the matching shadow `-I` in `helpers/test_config.py` (canonical `-I` then resolves the header).
- Add the defaulted granularity template arg to the primitive call (`signal_granularity=1` / `input_granularity=1`); for `sdpa_bcast_col_srca_srcb_reuse` drop `fused_signalling`.
- Make the promoted headers `-Werror`-clean and remove the localized `#pragma GCC diagnostic` shims.
- `compressed_custom_mm`'s undocumented `base_address_meta` arg is now driven for real: packed 3-bit per-tile compression-format codes, routed through the harness's optional third operand. Worth documenting on promotion that it is **control flow, not just numerics** — code 0 means "zero tile", for which the unpacker emits no UNPACR at all and math takes the `STALLWAIT(SRCB_VLD)` branch, so an all-zero meta buffer hangs Math/Packer forever. The dropped `CT_DIM` in its unpack init is unchanged.

Part of tenstorrent/tt-metal#47554 / tt-blaze#1971.

cc @ldjurovic — this is the SDPA/matmul slice of the uplifted-kernel test coverage; complements the promotion work.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ncvetkovic/experimental-llk-advance-scaffolds)


ncvetkovicTT:
Follow-up promotion tracked in #52721 — moves these 8 LLKs from the demo-fork shadow tree into `experimental/` (internal move; 0 numeric forks). Once that lands, the shadow `-I` here gets dropped and the primitive calls gain the defaulted granularity template args.

ldjurovicTT:
## Review comments addressed — status and what's left

Pushed two commits (`76126f77f`, `f047dcb3e`). Everything below was worked out against a p100a (Blackhole), not from reading alone.

**Baseline before these commits:** 212 variants, all failing, **20 `TENSIX TIMED OUT`s** across three files.

### The `sdpa_bcast_col_srcb_reuse` "deadlock" is not an LLK defect — retracting the writeup

Comment on `sdpa_bcast_col_srcb_reuse_test.cpp:115` is correct and it settles the open question in `docs/tests/sdpa_bcast_col_srcb_reuse_deadlock.md`. The demo issues **two** `llk_unpack_A` calls per math call (`compute_kernel_api/sdpa.h:56-57`), so the duplicated `ckernel_template::run()` is deliberate: the op is a two-operand-tile op computing `cb_l1*P1 + cb_l2*P2`. The test was supplying 2 SrcA dvalids against 4 `CLR_A` ELWMULs.

So the writeup's §3/§4 should be struck: neither Option 1 (edit the LLK) nor Option 2 (skip the tests) is needed. It was a test-side fix.

Three things measured on silicon that no header states, and that between them fixed four of the eight tests:

- The op's tile is **8x32** (two 8x16 faces), not 16x32 — each ELWMUL writes 8 dest rows.
- The ELWMULs **accumulate into DEST**, so `clear_dest` must be true or the column-source seed still sitting in `DEST[dst]` leaks into column 0.
- The demo's DEST convention throughout is an 8x32 tile packed into **one 16x16 face** (rows 0-7 = logical cols 0-15, rows 8-15 = cols 16-31). That single fact fixed `srca_srcb`, `sdpa_reduce_row` and `sdpa_custom_mm_reuse_dest_srcb` as well.

### Verified passing on p100a (5 files)

`sdpa_bcast_col_srcb_reuse`, `unpack_A_sdpa`, `sdpa_bcast_col_srca_srcb_reuse`, `sdpa_reduce_row`, `sdpa_custom_mm_reuse_dest_srcb`. All three hang classes are gone. The two most layout-sensitive goldens were negative-controlled (perturb the golden → they fail), so they are not passing vacuously.

Notes worth flagging from those:

- `srca_srcb` sources **both** operands from DEST (MOP is `[MOVD2A, MOVD2A, ELWMUL]`, every addrmod has `.srca.incr == 0`), so the unpacked `buffer_A` tile was discarded and the `Elwmul(src_A, src_B_bcast)` golden was unreachable. It computes `X + X*bcast_col(P)` — which is exactly why the demo subtracts 1 from its SFPU scale (`sdpa.h:207-210`).
- `sdpa_custom_mm_reuse_dest_srcb` needed all of: the `UNPACK_MATH_DONE` posts hoisted above the unpack execute, a real in0 unpack (the datacopy was seeding DEST from the ZEROSRC dummy, so in0 was identically zero), the SDPA init moved below the datacopy init that clobbers `ADDR_MOD_0/2/3`, and **exactly one** SrcB dummy dvalid — zero hangs both `kt_dim`, two wedges `kt_dim == 4`.

### ~~Not yet verified (3 files) — this is the main thing left~~ — resolved, all three now verified

> **Update:** the board came back and all three pass on a p100a. The in0 layout fix in `f047dcb3e` also had a host-side bug that made every variant of these three files fail in stimulus write, before the kernel ran (`Data to write must not be empty`); fixed in `ad651a47a`. Details in the follow-up comment below. Original text kept for the record:

`custom_mm`, `compressed_custom_mm`, `sdpa_custom_mm`. The in0 layout fix is in and **all 176 variants compile clean**, but the numerics are **unverified**: the board went into `PcieHangError` mid-run and `tt-smi -r` cannot re-initialise it (`Read 0xffffffff over PCIe`, ARC unresponsive). Needs a PCIe remove/rescan or a host reboot, then:

```
pytest -q --tb=short -p no:randomly test_custom_mm.py test_compressed_custom_mm.py test_sdpa_custom_mm.py
```

I'd expect one or two more iterations there — those goldens are reasoned from the silicon-validated `matmul_custom_compressed` sibling rather than measured, and every other file in this batch took a device round-trip to land.

**Please sanity-check one judgement call:** I dropped 36 variants, the `in0_rows` sweep on both sdpa matmuls. Not a coverage cut to hide failures — `sdpa_custom_mm_configure_addrmod` opens with `constexpr std::uint32_t face_r_dim = 8;` and derives every dest step from it, and `sdpa_custom_mm_reuse_dest_srcb` hardcodes the same 8-row step, so rows in `{1,2,4}` were never implemented by either primitive. The `{1,2,4,8}` in the banners is the *plain* `custom_mm` contract, which takes `face_r_dim` as an argument. If you read the sdpa variants as supporting all four, that's the call to revisit.

### Comment-by-comment

| Comment | Status |
|---|---|
| 1 — sem posts above the unpack execute | done, verified |
| 2 — SDPA init below the datacopy init | done, verified |
| 3 — real in0 unpack, dummy after datacopy | done, verified |
| 4 — srca_srcb sources both operands from DEST | done, verified |
| 5 — srcb_reuse is a two-operand-tile op | done, verified |
| 6 — `num_faces_A`/`num_faces_B` crossed | done, all 4 call sites |
| 7 — in0 dense-face L1 layout | done, verified (`ad651a47a`) |
| 8 — `@blackhole_only` | done, all 8 |
| 9 — scoped `-I` fixture | done (`helpers/advance_llk_includes.py`) |
| 10 — reuse `IN_FACE_DIMS` | done |
| 11 — `mask_chunk` coverage | **not done — needs a tracking issue** |
| 12 — factor the cloned goldens/grids | done (`helpers/sdpa_bcast_utils.py`, `helpers/custom_mm_utils.py`) |
| 13 — SrcB-only vs SrcA+SrcB dummy helper | done, verified |
| 14 — dangling `rmsnorm_test.cpp` reference | done |
| 15 — `TensorShape` / bare arg comments | done |
| 16 — `NUM_FACES` TemplateParameter | done, as `MATH_NUM_FACES` |

Two deliberate deviations:

- **16**: named `MATH_NUM_FACES`, not `NUM_FACES`. After the srca_srcb fix it is no longer the same value as the host/pack count — that test drives a single 16x16 face on the unpack/pack side while its mop still needs 2 (the inner-loop count for the two 8-row ELWMUL chunks). `NUM_FACES` would have been misleading.
- **11**: left as a documented gap rather than added unverified. `mask_chunk=true` is well contained — unpack pulls an extra SrcB tile through config context 1 (`llk_unpack_AB_sdpa_custom_mm.h:43-47`) and MATH seeds DEST with `ct_dim*2` `MOVB2D MOV_8_ROW_BRCST`s instead of ZEROACCing (`llk_math_sdpa_custom_mm.h:102-109`), so the op computes `mask + in0 @ in1`. But pinning the broadcast semantics needs the board, and shipping an unverified masked variant seemed worse than the gap. `test_sdpa_custom_mm.py` now spells out exactly what is untested. **A tracking issue still needs filing.**

### Unrelated harness bug fixed along the way

`ttexalens` 0.3.29 (what `tests/requirements.txt` pins) moved `CallstackEntry`'s file/line/column onto a nullable `file_info`, so `helpers/device.py:_print_callstack` raised `AttributeError` instead of printing the assert — every LLK assertion was unreportable. Fixed to handle both shapes and frames with no DWARF line info.

ldjurovicTT:
## On-device verification of the last 3 files — all green on p100a

Follow-up to the status comment above. `tt-smi -r` brought the board back, and the three files that comment listed as unverified — `custom_mm`, `compressed_custom_mm`, `sdpa_custom_mm` — now pass on Blackhole silicon. **Comment 7 (in0 dense-face L1 layout) moves from _compile-only_ to _verified_**, which was the last open item from that round.

### The layout fix had a host-side bug, so none of those variants had ever reached the device

`f047dcb3e` gave in0 the layout the unpacker actually reads — `kt_dim*2` densely packed `in0_rows x 16` faces, an in0 k-tile being `64 * in0_rows` bytes — but kept handing that buffer to `StimuliConfig.write_matrix`, which strides the *host* buffer at `MAX_TILE_ELEMENTS` (1024) per tile and packs `num_faces * face_r_dim * FACE_C_DIM` datums out of each stride. With `tile_count_A = kt_dim` over a buffer of `32 * kt_dim * in0_rows` datums, tile 1's slice is empty, the packer returns `b""`, and ttexalens rejects the write:

```
Error type: TTException
Data to write must not be empty.
  test_compressed_custom_mm.py:162 -> helpers/stimuli_config.py:392 -> helpers/device_io.py:112
```

Every variant of all three files died there, in stimulus write, before the kernel ran — so "all 176 variants compile clean" was as far as that commit could have got, and the goldens it shipped had never been compared against anything. `test_compressed_custom_mm` carried a second instance of the same bug: `write_matrix` also walked its `buffer_B` / `buffer_C` `torch.zeros(1)` placeholders at `tile_count_B = kt_dim * ct_dim` before the real pre-packed bytes were written over the top.

Fixed in `ad651a47a`. An operand that is already its own exact L1 image has no tile geometry to stride by, so it now bypasses the packing path: `StimuliConfig._write_prepacked` sends a `bytes` / `bytearray` / `memoryview` buffer to L1 verbatim and both `write_matrix` variants return early on it. This is what the silicon-validated `compressed_utils.CompressedStimuliConfig` achieves by overriding `write()` wholesale. Tensor buffers are untouched, so no other test changes behaviour. `pack_in0_faces` returns those packed bytes now, and `_CompressedMMStimuli` is gone — in0, the BFP-compressed in1 stream and the meta words all go into a plain `StimuliConfig` as bytes.

### Results

| Suite | Result |
|---|---|
| `test_compressed_custom_mm.py` | **144 passed** — the 24 merge-gate corners plus the 120-case nightly full grid, all 3 BFP formats |
| all 8 files in this PR (`test_compressed_custom_mm`, `test_custom_mm`, `test_sdpa_custom_mm*`, `test_sdpa_bcast_col_src*`, `test_sdpa_reduce_row`, `test_unpack_A_sdpa`) | **200 passed** |

No marker filter on that run, so the nightly full grid is included in the 200. Negative-controlled the same way as the earlier five files: scaling `test_custom_mm`'s golden by 1.5 makes it fail on `matmul_acc_atol` (0.24), so these are real device reads, not a compare that short-circuits. So the whole batch of 8 is now verified on silicon rather than compile-green.

The two goldens I was least sure of both held without adjustment: the dense partial-tile readback (`dense_result_rowmajor`, 2 faces of `in0_rows x 16` per tile plus padding, `dense_packing` on the math init) and the BFP-quantization fold on `compressed_custom_mm` (pack in1 to its BFP format, unpack back to bf16, multiply *that* in the golden, so BFP rounding is not charged against PCC). Both were reasoned from the silicon-validated `matmul_custom_compressed` sibling; the round trip confirmed them.

### Still open — unchanged from the previous comment

- **11 — `mask_chunk`**: still a documented gap in `test_sdpa_custom_mm.py`, and the tracking issue still needs filing.
- **The dropped `in0_rows` sweep** (36 variants on the two sdpa matmuls) still wants the second opinion I asked for above — the primitives hardcode `face_r_dim = 8`, so I read rows in `{1,2,4}` as never implemented, but that is the one judgement call in this batch I would not want to be alone on.

tt_metal/tt-llk/tests/python_tests/test_compressed_custom_mm.py
```diff
@@ -0,0 +1,251 @@
...
+        super().write(location)
+        write_to_device(location, self.buf_b_addr, self.packed_b)
+        write_to_device(location, self.buf_c_addr, self.meta_bytes)
+
+
+@blackhole_only
+@parametrize(
+    formats=COMPRESSED_MM_FORMATS,
+    ct_kt_rows=matmul_grid(),
+)
```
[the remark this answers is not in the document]
ldjurovicTT: Addressed in `ff794aab4` (pushed after this review ran). The merge-gate `test_compressed_custom_mm` keeps all three BFP formats but only the grid corners — `ct_dim` and `in0_rows` at their extremes, both `kt_dim` values, 24 cases — and the full 120-case sweep moved to `test_compressed_custom_mm_full_grid` under `@pytest.mark.nightly`. Same body via `_run_compressed_custom_mm`, so the two cannot drift. All 144 pass on a p100a.

tt_metal/tt-llk/tests/sources/sdpa_bcast_col_srca_srcb_reuse_test.cpp
```diff
@@ -0,0 +1,242 @@
...
+#include "llk_pack_common.h"
+#include "params.h"
+
+void run_kernel(RUNTIME_PARAMETERS params)
+{
+#if defined(RUNTIME_FORMATS) && !defined(SPEED_OF_LIGHT)
+    const FormatConfig& formats = params.formats;
+#endif
+    _llk_pack_hw_configure_wrapper_<is_fp32_dest_acc_en, PackMode::Default>(formats.pack_src, formats.pack_dst, params.TILE_SIZE_PACK);
+    _llk_pack_init_wrapper_<PackMode::Default, false /* zero_output */>(formats.pack_dst);
```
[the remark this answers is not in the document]
ldjurovicTT: Already fixed at head, in `cd950f78b` (pushed after this review ran): both wrapper calls now pass the geometry explicitly — `_llk_pack_hw_configure_wrapper_(..., params.TILE_SIZE_PACK, FACE_R_DIM, FACE_C_DIM, params.num_faces)` and `_llk_pack_init_wrapper_(formats.pack_dst, FACE_R_DIM, FACE_C_DIM, params.num_faces)` — with a comment recording exactly the overrun you describe: the Default-mode pack MOP loops `MOP_OUTER_LOOP == num_faces * num_tiles`, so the defaults would retire 4 faces per `_llk_pack_` and write 4 x 512 B into a 512 B `buffer_Res` slot. `test_sdpa_bcast_col_srca_srcb_reuse` passes on a p100a.

ldjurovicTT:
## What is done, what is covered, and what is not

Two follow-up commits since the verification comment above, both no-behaviour-change cleanups, plus the coverage picture in one place so it does not have to be reconstructed from eight files.

### Cleanups

**`2526ede69f1` — one definition of the face row dim.** `FACE_R_DIM = 16` had been declared four times over (`custom_mm_utils`, `test_sdpa_reduce_row`, `test_sdpa_bcast_col_srca_srcb_reuse`, plus `SDPA_FACE_R_DIM = 16` as a fifth alias) while `tile_constants.MAX_FACE_R_DIM` is the canonical one `helpers/unpack.py` and `helpers/tilize_untilize.py` already use. All sites import it now, and `FACE_C_DIM` comes from `tile_constants` directly instead of being re-exported through `custom_mm_utils`. `SDPA_LOGICAL_ROWS` was dead and is deleted.

**`eb5c83ea3a9` — one K-aware matmul atol helper.** The LoFi K-deep atol floor existed twice: `custom_mm_utils.matmul_acc_atol` and the same computation inline in `compressed_utils.run_compressed`, whose comment carried the calibration detail the other copy only referenced. Both hardcoded `0.05`, which is a third copy of a number `helpers/utils.py` already owns in `tolerances`. Both now call `helpers.utils.matmul_acc_atol`, sitting next to `tolerances` and the `passed_test` that consumes it; the floor reads `tolerances[fmt].atol` instead of a literal, so it is correct for a non-bf16 output rather than silently bf16-only.

The shared helper went into `helpers/utils.py` rather than into either family module on purpose: had `compressed_utils` imported from `custom_mm_utils`, silicon-validated code would depend on scaffolding that promotion deletes.

Verified rather than assumed: a property check over shapes, `kt ∈ {2,4,16}`, magnitudes spanning 1e-3..1e2, a golden with structural zeros and the all-zero golden returns **bit-identical** values to the previous inline calibration; `test_matmul_custom_compressed` + `test_matmul_face_compressed` are **1750 passed** against a baseline captured *before* the change on the same board (406 s vs 405 s); the eight advance files are **200 passed**, unchanged.

### Coverage

200 hardware variants, 80 of them on the merge gate:

| Test | Variants (gate / total) | Swept | Fixed |
|---|---|---|---|
| `test_custom_mm` | 40 / 40 | ct_dim {1,2,4,8,16} × kt_dim {2,4} × in0_rows {1,2,4,8} | bf16 in/out, rt_dim 1 |
| `test_compressed_custom_mm` | 24 / 144 | in1 {Bfp8_b, Bfp4_b, Bfp2_b} × the same 40-point grid; gate keeps all 3 formats at the grid corners, full sweep nightly | bf16 in0/out, rt_dim 1, uniform format per run |
| `test_sdpa_custom_mm` | 10 / 10 | ct_dim {1,2,4,8,16} × kt_dim {2,4} | in0_rows 8, `mask_chunk` false |
| `test_sdpa_custom_mm_reuse_dest_srcb` | 2 / 2 | kt_dim {2,4} | in0_rows 8, nt_dim 1 |
| `test_sdpa_bcast_col_srcb_reuse` | 1 / 1 | — | one 8x32 tile (2 faces), 2 operand tiles + 2 column sources |
| `test_sdpa_bcast_col_srca_srcb_reuse` | 1 / 1 | — | one 16x16 DEST face, 8 logical rows |
| `test_unpack_A_sdpa` | 1 / 1 | — | same op/golden as `srcb_reuse`, different primitive pinned |
| `test_sdpa_reduce_row` | 1 / 1 | — | one 32x32 tile, one 8x32 reduce span, ReducePool.Max |

The four single-variant tests are single-variant because the primitives fix the geometry, not because the grid was trimmed: their MOP/addrmod helpers hardcode the face count and row step, so there is no shape axis to sweep. Only a format axis could grow them, and these ops are bf16-only.

**Not covered, deliberately:**

- **`mask_chunk` on `sdpa_custom_mm`** — the SrcB-mask / `MOVB2D MOV_8_ROW_BRCST` branch. Documented in the test banner; pinning the broadcast semantics needs on-silicon work. Tracking issue still to be filed, and still the one real gap in the batch.
- **`in0_rows` on the two sdpa matmuls** — `sdpa_custom_mm_configure_addrmod` opens with `constexpr std::uint32_t face_r_dim = 8` and derives every DEST step from it, so rows in {1,2,4} are not implemented by the primitive. This is the judgement call I would still like a second opinion on.
- **`nt_dim > 1` on `sdpa_custom_mm_reuse_dest_srcb`**, and **`ReducePool` other than Max** on `sdpa_reduce_row` — both single values today.
- **32-bit DEST accumulation** — every file runs `DestAccumulation.No`. **Fidelity** is LoFi everywhere, which is the contract (the math inits take no `MathFidelity`), and **throttle** is unsupported by these primitives.
- **Bfp8_b in0** on `custom_mm` — allowed by the LLK, not swept; the grid is bf16-natural.

### On the three new helper modules

`helpers/advance_llk_includes.py`, `helpers/custom_mm_utils.py`, `helpers/sdpa_bcast_utils.py` — asked internally whether these should be one file. They stay three:

- Each is shared by more than one test file: the include fixture by all 8, `custom_mm_utils` by the 4 matmul tests, `sdpa_bcast_utils` by 2.
- Per-family helper modules are the existing convention in `helpers/` — `compressed_utils`, `matmul_sweep`, `bfp_format_utils`, `sfpu_domains`. `sdpa_bcast_utils.run_sdpa_bcast_col_srcb_reuse` is the same shape as `compressed_utils.run_compressed`.
- The two utils modules have disjoint consumers, hold different kinds of thing (pure layout/golden functions vs a driver that runs the kernel), and would have collided on `SDPA_FACE_R_DIM`, which meant 16 in one and 8 in the other.
- `advance_llk_includes` standalone means promotion deletes one file plus one import line per test. Moving the fixture to `conftest.py` would park scaffolding in a file the whole suite shares and still need a `usefixtures` line in all 8 tests.

ldjurovicTT:
### CI note: bit-exact failures are not from this PR

**Rebased** onto `main` (`d144573` → `be8a814`, now on `85a8c1b`). Two conflicts in
`helpers/test_config.py`, both from main's new `llk_api/llk_sfpu` include landing in the same
block; resolved keeping both sides. Branch net delta is byte-identical pre/post rebase — same
25 files, no content change.

**Round 1 — fixed.** The 4 `test_generalized_moe_gate_*` failures were the known
[#52699](https://github.com/tenstorrent/tt-metal/pull/52699) bug (uninitialized SFPU LReg residue
in the tile padding; run 0 cold vs runs 1..N warm). Branch was cut Aug 9, fix landed Aug 10. The
rebase picks it up and those 4 now pass — at **499** re-runs, not the 100 that previously caught
them.

**Round 2 — not ours.** [Run 31583572316](https://github.com/tenstorrent/tt-metal/actions/runs/31583572316)
shows 7 `test_deepseek_moe_gate_*` failures. Same defect, sibling primitive:

- This branch does not touch those files (`git diff --name-only origin/main HEAD | grep deepseek_moe_gate` → empty).
- The tests are ~2.5 h older than the run: added by [#52827](https://github.com/tenstorrent/tt-metal/pull/52827) (`55aede79fd7`, Aug 12 07:12 UTC). They arrived **with the rebase** — `merge-base --is-ancestor 55aede79fd7 d144573c38a` is false.
- **Main has never bit-exact tested them.** The job is `schedule` (`0 3 * * *`) + `workflow_dispatch` only — it never runs on merge gates. Main's last run was the Aug 12 **03:00** nightly, 4 h before #52827 merged. This run is the first execution of these tests anywhere.
- `deepseek_moe_gate` runs the same SFPU `sum_top2` / `sort_top4_groups` / `top8` stages, and #52699's `gmg_sanitize_scratch()` was **never ported to it** — grep for `sanitize|scratch|residue|dst_reg[` in `deepseek_moe_gate_test.cpp` returns nothing. Identical cold-vs-warm signature at the same byte offsets (40/42/43).
- Ruled out the only way this branch could reach them: the `advance_llk_includes.py` autouse fixture is module-scoped and restores on teardown, and none of its 3 shadow roots can satisfy any `#include` in `deepseek_moe_gate_test.cpp` (they are fully qualified under `deepseek/moe/deepseek_moe_gate/device/kernel_includes/...`). No shadowing possible.

**Heads-up:** the Aug 13 03:00 UTC nightly on `main` will fail the same way. The fix belongs on
`main` — port `gmg_sanitize_scratch()` to `deepseek_moe_gate_test.cpp`, re-deriving the guard for
`DMG_MODE`/`DMG_SUB_OP` (the DEST layout already matches: `NUM_DEST_TILES=4`, `SCORES_TILE=0`,
`IDS_TILE=1`). Tracked separately; not in scope for this PR.

**This PR's own state:** all 8 advance-test files byte-identical across 499 runs; `test_perf_header_gate`
green since `d144573`.

tt_metal/tt-llk/tests/sources/sdpa_custom_mm_reuse_dest_srcb_test.cpp
```diff
@@ -0,0 +1,261 @@
...
+//   SFPU op, so MATH would block forever on the wait_on_zero. We fake the SFPU side on the UNPACK thread (Tensix
+//   semaphores are shared across all three threads): after its unpack work, UNPACK posts semaphore::UNPACK_MATH_DONE
+//   exactly kt_dim times. MATH's per-K wait_on_zero then sees non-zero and proceeds, and MATH's per-K get decrements
+//   it, so posts (kt_dim) == gets (kt_dim) and nothing leaks. kt_dim <= 4 here, well under SEMAPHORE_MAX_VALUE (15).
+//   We instantiate signal_output == false, so MATH never POSTs semaphore::FPU_SFPU and there is nothing to drain on
+//   that channel (matches the demo default; mirrors sdpa_custom_mm_test.cpp's "MATH never WAITS on SFPU" reasoning,
+//   here inverted: MATH waits, UNPACK fakes the post). See harness_needs in the comparison verdict.
+//
+// Blackhole-only. Deliverable here is compile-green (compile-producer). On-device numerical verification is pending
+// Blackhole hardware/CI; this host is Wormhole. The MatmulGolden below validates standard LoFi matmul numerics and,
```
ncvetkovicTT: Stale status banner. This says "compile-green (compile-producer)", verification "pending Blackhole hardware/CI; this host is Wormhole" — but the PR body says all 8 files are on-device verified on p100a (200/200), and this file's own comments describe fixes "measured on p100a" (the semaphore-count and single-dummy-SrcB-valid findings). If the p100a run covered this op, please drop the compile-green/pending/Wormhole line so it stops contradicting the body and the in-body p100a notes.

tt_metal/tt-llk/tests/sources/sdpa_custom_mm_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+// (orthogonal to this numerical golden).
+//
+// The primitive under test is the demo-fork sdpa_custom_mm matmul (in0 [{1,2,4,8},32] -> SrcB, in1 [32,32] -> SrcA,
+// rt_dim==1, ct_dim 1..16, kt_dim even 2..256, LoFi-only). We instantiate the mask_chunk=false path (per the
+// comparison verdict) so no SrcB-mask / MOVB2D branch is taken and no base_address_mask buffer is required. MATH only
+// POSTs semaphore::FPU_SFPU (non-blocking t6_semaphore_post) and never WAITS on SFPU, so the isolated UNPACK/MATH/PACK
+// kernel does not deadlock; the un-drained FPU_SFPU increments are harmless.
+//
+// Blackhole-only. Deliverable here is compile-green (compile-producer). On-device numerical verification is pending
+// Blackhole hardware/CI; this host is Wormhole.
```
ncvetkovicTT: Same stale "compile-green / verification pending / this host is Wormhole" banner as the reuse_dest_srcb file. Contradicts the PR body's "all 8 files green, 200/200" on p100a. Reconcile: either this op was silicon-verified (drop these two lines) or the body overstates coverage.

tt_metal/tt-llk/tests/sources/sdpa_reduce_row_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+// stream byte-identical between the DEMO and tt-blaze headers. The numerical golden below therefore covers the shared
+// (non-signalling) compute path.
+//
+// The op writes results via SFPU LREG stores to DEST addressed by TT_SETC16(DEST_TARGET_REG_CFG_MATH_Offset...), so we
+// drive it from the thread that owns DEST. Mirroring the analog sfpu_reduce_sdpa_test.cpp, the SFPU reduce runs on the
+// PACK thread with its own MATH/SFPU init, fed a Float16_b 8x32 (single-face) input tile seeded into DEST by an A2D
+// datacopy on the MATH thread.
+//
+// Blackhole-only. Deliverable here is compile-green (compile-producer). On-device numerical verification is pending
+// Blackhole hardware/CI; this host is Wormhole.
```
ncvetkovicTT: This banner says verification is "pending Blackhole hardware/CI; this host is Wormhole", but the paired test_sdpa_reduce_row.py says the opposite — "The golden below is verified on Blackhole silicon (p100a), not compile-green only" and "Verified on p100a". The two files can't both be right. Please make this banner match (drop the compile-green/pending/Wormhole line if the p100a run covered it).

tt_metal/tt-llk/tests/sources/sdpa_reduce_row_test.cpp
```diff
@@ -0,0 +1,172 @@
...
+                tile, formats.math, formats.math);
+        }
+
+        _llk_math_dest_section_done_<DstSync::SyncHalf, is_fp32_dest_acc_en>();
+    }
+}
+
+#endif
+
+#include "llk_math_eltwise_unary_sfpu.h"
```
ncvetkovicTT: Stray file-scope `#include "llk_math_eltwise_unary_sfpu.h"` between the MATH and PACK `#ifdef` blocks. It's here so the PACK block gets the SFPU init symbols, but every other include in all eight files lives inside its thread's `#ifdef`. Harmless (guards), but move it inside `#ifdef LLK_TRISC_PACK` to match the rest.

tt_metal/tt-llk/tests/python_tests/test_sdpa_custom_mm_reuse_dest_srcb.py
```diff
@@ -0,0 +1,172 @@
...
+        in1,
+        formats.output_format,
+        MathFidelity.LoFi,
+        input_A_dimensions=in0_dimensions,
+        input_B_dimensions=in1_dimensions,
+        tilize=False,
+        input_A_format=formats.input_format,
+        input_B_format=formats.input_format,
+    ).reshape(in0_rows, NT_DIM * 32)
+    golden_tensor = sdpa_dest_tile_golden(matmul_rowmajor, torch_format)
```
ncvetkovicTT: This inlines `MatmulGolden()(...).reshape(in0_rows, NT_DIM*32)`, which is byte-identical to `custom_mm_utils.matmul_lofi_golden` — its reshape `(in0_dimensions[0], in1_dimensions[1])` equals `(in0_rows, NT_DIM*32)` here. The other three matmul tests all call the helper; this is the only one that doesn't. Switching to `matmul_lofi_golden(in0, in1, formats, in0_dimensions, in1_dimensions)` also drops the now-unneeded `MatmulGolden`/`MathFidelity` imports.

tt_metal/tt-llk/tests/python_tests/test_sdpa_custom_mm_reuse_dest_srcb.py
```diff
@@ -0,0 +1,172 @@
...
+    # and drop the three faces of L1 the kernel never touches. StimuliConfig's num_faces would narrow the readback
+    # for us, but it narrows the buffer_A / buffer_B *writes* too, which would truncate the in1 K-tiles.
+    face_datums = MAX_FACE_R_DIM * FACE_C_DIM
+    tile_datums = len(res_tensor) // NT_DIM
+    res_tensor = torch.cat(
+        [
+            res_tensor[n * tile_datums : n * tile_datums + face_datums]
+            for n in range(NT_DIM)
+        ]
+    )
```
ncvetkovicTT: This leading-face readback is identical to `custom_mm_utils.face_result_leading(res_tensor, NT_DIM)`, whose docstring even says "Used by the two sdpa matmuls" — but only test_sdpa_custom_mm.py uses it. Replace this block with the helper so both sdpa matmuls share it (also lets you drop the `FACE_C_DIM`/`MAX_FACE_R_DIM` imports if they're otherwise unused).

ncvetkovicTT (commented):
Consistency/correctness pass. Main item is the verification-status banners below — they contradict the PR description ("all 8 files green, 200/200" on p100a) and, in two cases, other files in this same PR. Plus a helper-dedup follow-up and one stray include.

tt_metal/tt-llk/tests/sources/sdpa_reduce_row_test.cpp
```diff
@@ -0,0 +1,174 @@
...
+        // file banner) so the isolated PACK-thread kernel does not deadlock and the compute path is byte-identical to
+        // tt-blaze.
+        _llk_math_eltwise_unary_sfpu_init_<SfpuType::reduce>();
+        ckernel::sfpu::_init_sdpa_reduce_row_8x32_<DataFormat::Float16_b>();
+        ckernel::sfpu::_init_sdpa_reduce_max_row_8x32_replay_buffers_();
+        _llk_math_eltwise_sfpu_start_(0);
+        ckernel::sfpu::_calculate_sdpa_reduce_max_row_8x32_<DataFormat::Float16_b, BLOCK_WIDTH, true /* skip_signalling */>(
+            SRC_INDEX, DST_INDEX, false /* prev_max */);
+        _llk_math_eltwise_sfpu_done_();
+
```
markoradosavljevicTT: I think we need STALLWAIT here to wait for all SFPU instructions to complete before we start pack.

tt_metal/tt-llk/tests/sources/unpack_A_sdpa_test.cpp
```diff
@@ -0,0 +1,220 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
iklikovacTT: This file is identical to sdpa_bcast_col_srcb_reuse_test.cpp, save for differences in comments. Can we delete one of them?

tt_metal/tt-llk/tests/python_tests/test_sdpa_bcast_col_srca_srcb_reuse.py
```diff
@@ -0,0 +1,140 @@
...
+    #   out[d, c] = X[d, c] * (1 + P[d % 8, 0])
+    # The scale repeats every 8 dest rows because srcb.incr == 0 across the MOP's two 8-row chunks.
+    x = src_A.reshape(MAX_FACE_R_DIM, FACE_C_DIM).to(torch_format)
+    p_col0 = src_B.reshape(MAX_FACE_R_DIM, FACE_C_DIM).to(torch_format)[
+        :LOGICAL_ROWS, 0
+    ]
+    scale = 1.0 + p_col0.repeat(MAX_FACE_R_DIM // LOGICAL_ROWS).reshape(
+        MAX_FACE_R_DIM, 1
+    )
+    golden_tensor = (x * scale).flatten()
```
iklikovacTT: Can we use EltwiseBinaryGolden here?

tt_metal/tt-llk/tests/python_tests/helpers/test_variant_parameters.py
```diff
@@ -67,6 +67,33 @@ def convert_to_struct_fields(self) -> tuple[str, str]:
 # === TEMPLATE PARAMETER IMPLEMENTATIONS ===
 
 
+@dataclass
+class MATH_NUM_FACES(TemplateParameter):
```
markoradosavljevicTT: Can we drop MATH_NUM_FACES and just pass NUM_FACES in templates instead of runtimes? It already emits constexpr std::uint32_t num_faces, which is the compile-time constant the MATH thread needs.

tt_metal/tt-llk/tests/python_tests/helpers/test_config.py
```diff
@@ -477,6 +477,10 @@ def setup_compilation_options(
                 # resolves with this on the path. Listed last so the tt-llk copy still
                 # wins the basenames that exist in both trees.
                 "-I../../hw/ckernels/blackhole/metal/llk_api/llk_sfpu",
+                # NOTE: the demo-fork experimental-LLK ADVANCE TEST shadow roots deliberately do NOT live here.
```
iklikovacTT: Maybe reword or remove this comment before this gets merged into main? I think it would just confuse people reading it later on.

tt_metal/tt-llk/tests/python_tests/test_sdpa_custom_mm_reuse_dest_srcb.py
```diff
@@ -0,0 +1,150 @@
...
+    formats=SDPA_REUSE_FORMATS,
+    kt_dim=KT_DIMS,
+)
+def test_sdpa_custom_mm_reuse_dest_srcb(
+    formats,
+    kt_dim,
+    boot_mode=BootMode.DEFAULT,
+):
+    if isinstance(formats, tuple):
+        (formats,) = formats
```
iklikovacTT: If I'm not mistaken, there's no need for this unwrap since this is a parametrize with more than 1 axis.

tt_metal/tt-llk/tests/python_tests/helpers/custom_mm_utils.py
```diff
@@ -0,0 +1,171 @@
...
+    nt_dim = out.shape[1] // 32
+    return torch.cat(
+        [
+            _to_dest_face(out[:, n * 32 : (n + 1) * 32], torch_format).reshape(-1)
+            for n in range(nt_dim)
+        ]
+    )
+
+
+def matmul_lofi_golden(in0, in1, formats, in0_dimensions, in1_dimensions):
```
markoradosavljevicTT: ● I don't think the matmul golden needs a wrapper here, it looks like boilerplate

tt_metal/tt-llk/tests/python_tests/helpers/sdpa_bcast_utils.py
```diff
@@ -0,0 +1,185 @@
...
+    operands = src_A.reshape(SDPA_TILE_PAIRS * rows, 32).to(torch_format)
+
+    golden_tensor = torch.zeros(rows, 32, dtype=torch_format)
+    for t in range(SDPA_TILE_PAIRS):
+        scale = _bcast_col_untilized(
+            tilized_B.flatten()[t * per_tile : (t + 1) * per_tile],
+            formats.input_format,
+            broadcast_golden,
+        ).to(torch_format)
+        golden_tensor += operands[t * rows : (t + 1) * rows, :] * scale
```
iklikovacTT: Does it make sense to use EltwiseBinaryGolden here?

tt_metal/tt-llk/tests/python_tests/helpers/stimuli_config.py
```diff
@@ -418,6 +418,26 @@ def get_packer(data_format):
         }
         return packers.get(data_format)
 
+    @staticmethod
+    def _write_prepacked(buffer, base_address: int, location: str) -> bool:
```
iklikovacTT: Maybe it would be a good idea to add some bounds checking to this function

iklikovacTT (approved):

ncvetkovicTT:
Reconciling the overlap with #53361 (per review feedback — yes, these two PRs duplicate a chunk of tests).

#53361 targets the now-landed **promoted `experimental/` headers**, is HW-validated on a Blackhole p150b, and carries the fixes these tests need (custom_mm fp32 dense-packing pack-stride → 96/96 incl. Float32; sdpa_custom_mm `read_transposed`). To remove the duplication, **this PR should drop the 5 tests that #53361 owns** (both the `test_*.py` and the matching `sources/*_test.cpp`):

- `test_custom_mm.py` + `sources/custom_mm_test.cpp`
- `test_sdpa_custom_mm.py` + `sources/sdpa_custom_mm_test.cpp`
- `test_sdpa_custom_mm_reuse_dest_srcb.py` + `sources/sdpa_custom_mm_reuse_dest_srcb_test.cpp`
- `test_sdpa_reduce_row.py` + `sources/sdpa_reduce_row_test.cpp`
- `test_unpack_A_sdpa.py` + `sources/unpack_A_sdpa_test.cpp`

This PR keeps what's unique to it: `compressed_custom_mm`, plus `sdpa_bcast_col_srcb_reuse` and `sdpa_bcast_col_srca_srcb_reuse` (already removed from #53361 and ceded here).

(This also resolves the identical-`unpack_A_sdpa_test.cpp` observation from the inline review.)

tt_metal/tt-llk/tests/sources/unpack_A_sdpa_test.cpp
ncvetkovicTT: Yes — unpack_A_sdpa_test.cpp is one of the 5 duplicates this PR is dropping (covered by #53361 against the promoted experimental/ headers; see the dedup comment above), so it's removed rather than deduped in place.

tt_metal/tt-llk/tests/python_tests/test_sdpa_custom_mm_reuse_dest_srcb.py
ncvetkovicTT: This file is also in the drop-5 set ceded to #53361, so it's being removed here. FYI #53361's owned copy uses a clean 3-axis parametrize with no such unwrap, so the nit doesn't carry over.

ldjurovicTT:
> Reconciling the overlap with #53361 (per review feedback — yes, these two PRs duplicate a chunk of tests).
> 
> #53361 targets the now-landed **promoted `experimental/` headers**, is HW-validated on a Blackhole p150b, and carries the fixes these tests need (custom_mm fp32 dense-packing pack-stride → 96/96 incl. Float32; sdpa_custom_mm `read_transposed`). To remove the duplication, **this PR should drop the 5 tests that #53361 owns** (both the `test_*.py` and the matching `sources/*_test.cpp`):
> 
> * `test_custom_mm.py` + `sources/custom_mm_test.cpp`
> * `test_sdpa_custom_mm.py` + `sources/sdpa_custom_mm_test.cpp`
> * `test_sdpa_custom_mm_reuse_dest_srcb.py` + `sources/sdpa_custom_mm_reuse_dest_srcb_test.cpp`
> * `test_sdpa_reduce_row.py` + `sources/sdpa_reduce_row_test.cpp`
> * `test_unpack_A_sdpa.py` + `sources/unpack_A_sdpa_test.cpp`
> 
> This PR keeps what's unique to it: `compressed_custom_mm`, plus `sdpa_bcast_col_srcb_reuse` and `sdpa_bcast_col_srca_srcb_reuse` (already removed from #53361 and ceded here).
> 
> (This also resolves the identical-`unpack_A_sdpa_test.cpp` observation from the inline review.)

Dropped the 5 tests as suggested — agreed on the split, and #53361's versions are broader (`custom_mm` covers Float32 and the odd `ct` 7/9/11; `sdpa_custom_mm_reuse_dest_srcb` sweeps fidelity/dest_acc too).

One thing needs to carry over before the drop is safe. @markoradosavljevicTT flagged a missing SFPU drain on my `sdpa_reduce_row_test.cpp`, and **#53361's copy has the same gap** — so dropping mine loses the fix rather than moving it.

`tests/sources/sdpa_reduce_row_test.cpp`, ~line 174: `_llk_math_eltwise_sfpu_done_()` only clears the dest write address (`math::clear_dst_reg_addr()`) — it issues no wait. The SFPU is still retiring the reduce's stores when the pack loop below it starts reading DEST.

Please add, between `_llk_math_eltwise_sfpu_done_()` and the `_llk_pack_` loop:

```cpp
    _llk_math_eltwise_sfpu_done_();

    // Hold the packer until the reduce's SFPU stores drain.
    TTI_STALLWAIT(p_stall::STALL_PACK, p_stall::WAIT_SFPU);

    for (std::uint32_t tile = 0; tile < params.TILE_CNT; ++tile)
    {
        _llk_pack_<...>(tile, L1_ADDRESS(params.buffer_Res[tile]));
    }
```

Same guard the promoted SFPU-on-PACK primitives already use — `ckernel_sfpu_topk_xl.h:2408`. It applies to both the Max and Sum paths.

Worth a re-run on the p150b: this is a timing race, so "passes today" doesn't rule it out — it can hide behind pipeline latency and surface later on a different build or arch.

Two smaller notes:

- #53361 is still open, so there's a coverage gap between my PR merging and yours. Fine by me, just flagging it.
- The headers behind the 3 tests I keep (`llk_math_sdpa_bcast_col_srcb_reuse.h`, `llk_math_sdpa_bcast_col_srca_srcb_reuse.h`, `llk_math_compressed_custom_mm.h`, `llk_unpack_A_sdpa.h`) have **all landed in `experimental/` on main** too. So my shadow-`-I` scaffold is stale as well and I should repoint those at the canonical headers — which would also let the `-Werror` suppressions in the two bcast kernels be reviewed against the real headers, and connects to your #53295 note for @pmilenkovicTT about the 5 unused symbols in `llk_math_sdpa_bcast_col_srca_srcb_reuse.h`.

tt_metal/tt-llk/tests/python_tests/helpers/custom_mm_utils.py
```diff
@@ -0,0 +1,78 @@
...
+"""
+
+import torch
+
+from .tile_constants import FACE_C_DIM
+
+# Header contract: ct_dim 1..16, kt_dim even (2..256), in0 rows in {1, 2, 4, 8}.
+CT_DIMS = [1, 2, 4, 8, 16]
+KT_DIMS = [2, 4]
+IN0_ROWS = [1, 2, 4, 8]
```
ndivnicTT: Does it make sense to rename IN0_ROWS to RT_DIMS?

tt_metal/tt-llk/tests/python_tests/helpers/utils.py
```diff
@@ -98,6 +98,34 @@ def format_tile_row(
...
+    The mean excludes zeros so a golden carrying structural zeros -- the bfp0 "zero tile"
+    of the compressed matmuls -- cannot deflate it.
+    """
+    active = golden_tensor.abs().flatten()
+    active = active[active > 0]
+    mean_active = active.mean().item() if active.numel() else 0.0
+    return max(
+        tolerances[output_data_format].atol,
+        MATMUL_ACC_ATOL_PER_KT * kt_dim * mean_active,
+    )
```
ndivnicTT, line 142: Why not just tolerances[output_data_format].atol * kt_dim if it scales with kt_dim?

tt_metal/tt-llk/tests/python_tests/test_compressed_custom_mm.py
```diff
@@ -0,0 +1,263 @@
...
+# MOP config supports.
+IN1_COMPRESSED_FORMATS = [DataFormat.Bfp8_b, DataFormat.Bfp4_b, DataFormat.Bfp2_b]
+
+_FMT_CODE = {
+    DataFormat.Bfp8_b: FMT_CODE["bfp8"],
+    DataFormat.Bfp4_b: FMT_CODE["bfp4"],
+    DataFormat.Bfp2_b: FMT_CODE["bfp2"],
+}
+
+TILE_DIM = 32
```
ndivnicTT: Don't we have these as some utils? Does it make sense to move them to utils, it's not custom matmul specific

tt_metal/tt-llk/tests/sources/compressed_custom_mm_test.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+// PRIMITIVE symbol under test (NOT the forked _api.h wrapper / compute_kernel_api entry).
+// Resolved from the promoted experimental/ copy (landed on main via #53295); the demo-fork shadow tree this test was
+// originally written against is gone. The promoted header still takes params that the instantiation we drive does not
+// read, which -Werror -Wunused-parameter flags. Scoped push/pop is enough here (unlike the sdpa siblings, the
+// offending params are not on template bodies). Drop it once the promoted header is warning-clean.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunused-parameter"
+#include "experimental/llk_unpack_AB_compressed_custom_mm.h"
+#pragma GCC diagnostic pop
```
ndivnicTT: Or just add [[maybe_unused]] in .h file, if that's a valid option

ndivnicTT (approved):
LGTM!

Issue #52630, opened 2026-08-08
`ValueError`: Missing DeepSeek quantization_config.weight_block_size.


kgrujcic-tt:
https://github.com/tenstorrent/tt-metal/actions/runs/31241236255/job/93073764769

PR #52744, opened 2026-08-10
#52630: Split state-dict conversion from dequantization in hf_model_utils

kgrujcic-tt:
### Ticket

Related to #52630 (the CI fix itself is #52743, which does not depend on this PR).

### Problem description

`dequantize_state_dict` has always done two jobs: dequantize tensors that carry a `_scale_inv` companion, and pass everything else through with a dtype cast. But it reads `quantization_config.weight_block_size` up front, before it looks at a single tensor:

```python
def dequantize_state_dict(state_dict, hf_config, dtype=torch.bfloat16):
    block_shape = get_weight_block_shape(hf_config)   # raises here
    ...
```

So the pass-through half is unreachable for a checkpoint whose config carries no quantization metadata at all — the function raises before it can discover that there was nothing to dequantize:

```
ValueError: Missing DeepSeek quantization_config.weight_block_size.
```

Every model here ships as both a quantized and a dequantized variant, and only the config distinguishes them, so callers can't know in advance which function to reach for. This surfaced as a Kimi CI failure (#52630, fixed independently in #52743); this PR addresses the shared helper so any consumer of `hf_model_utils` can load either variant.

### What's changed

`models/demos/deepseek_v3/utils/hf_model_utils.py`:

- **`passthrough_state_dict`** — the no-scale half of `dequantize_state_dict`, extracted verbatim. Raises on a float8 tensor, which means quantized weights whose config lost its metadata.
- **`convert_state_dict`** — dequantizes when `quantization_config` is present, passes through when it isn't, and logs which conclusion it reached (once per process, since callers invoke it per module and per layer). An optional `dequantizer` hook lets a caller substitute its own dequantizer.
- **`get_quantization_config`** — resolves the config, also looking under `text_config` for multimodal checkpoints (Kimi K2.5/K2.6) that nest the LM config there. Without this a quantized Kimi reads as dequantized and its fp8 weights would be passed through raw.

No existing caller changes behavior:

- `dequantize_state_dict` keeps its signature and semantics; only its two shared bodies are factored into `_cast_unquantized_tensor` / `_iter_weight_names`.
- `get_weight_block_shape` only gains the `text_config` fallback — a strict widening of what it already accepted.

The pre-existing `pytest.raises` calls in `test_hf_model_utils.py` are converted to the `expect_error` fixture, since the `prefer-expect-error` pre-commit hook blocks any commit touching that file otherwise.

### Checklist

- `models/demos/deepseek_v3/tests/test_hf_model_utils.py` — **24 passed** (20 before, 4 new covering pass-through, dequantize, `text_config` nesting, and the float8-without-config error).

The four new tests pin both branches of the decision and the nesting case; the existing 20 pin that nothing else moved.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kgrujcic/52630-convert-state-dict)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kgrujcic/52630-convert-state-dict)

### DIFF
--- a/models/demos/deepseek_v3/tests/test_hf_model_utils.py
+++ b/models/demos/deepseek_v3/tests/test_hf_model_utils.py
@@ -15,6 +15,7 @@ import torch
 
 from models.demos.deepseek_v3.utils.hf_model_utils import (
     _default_quad_ring_shard_maps,
+    convert_state_dict,
     default_quad_ring_model_path,
     default_stacked_dequantized_model_path,
     index_model_weights,
@@ -412,7 +413,7 @@ def test_save_dequantized_hf_checkpoint_rewrites_dequantized_shards_when_stackin
     )
 
 
-def test_save_dequantized_hf_checkpoint_rejects_incomplete_stacked_expert_groups(tmp_path: Path):
+def test_save_dequantized_hf_checkpoint_rejects_incomplete_stacked_expert_groups(tmp_path: Path, expect_error):
     source_dir = tmp_path / "deepseek-source"
     output_dir = default_stacked_dequantized_model_path(source_dir)
     source_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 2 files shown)


kgrujcic-tt:
Split between `deepseek_v3` and `deepseek_v3_d_p`. [The other PR has been merged](https://github.com/tenstorrent/tt-metal/pull/52743).

PR #52747, opened 2026-08-10
Retire the demo deepseek_moe_gate fork onto canonical generalized_moe_gate

pmilenkovicTT:
### PR Category

Cleanup

### Summary

Delete the `deepseek_v3_b1` copies of the `deepseek_moe_gate` LLK family (13 headers, Blackhole + Wormhole) — a stale fork of what upstream productized as `generalized_moe_gate` — and rewire the unified gate kernel onto the canonical family. This is the same migration tt-blaze made in tenstorrent/tt-blaze#2890, now applied to the demo tree.

Relates to tenstorrent/tt-blaze#1971 / tenstorrent/tt-blaze#2002.

`unified_kernels/deepseek_moe_gate.hpp` changes:

- includes the canonical `api/compute/experimental/generalized_moe_gate.h`, with `GMG_UNGROUPED_TOP8` defined to `0` (grouped DeepSeek gate) before the include, per the header's explicit path-select contract;
- `deepseek_moe_gate_init` / `deepseek_moe_gate` calls renamed to their `generalized_moe_gate` counterparts (identical template and value arguments).

Out of scope: the ttnn `deepseek_moe_gate` op's own vendored copies are a separate fork, tracked for a later batch.

### Notes for reviewers

- The canonical family has genuinely diverged from this fork (transpose init, SFPU macro layer, multi-block helpers), so validation is by golden tests rather than binary identity: `test_deepseek_moe_gate` checks routed indices exactly and scores to tolerance, and passes.
- `moe_routed_expert` / `moe_mlp` / `decoder_block` compile the same rewired unified header; their runtime suites are galaxy-only (all pre-existing platform skips on single-chip) and covered by CI.

Verified on BH p100a: `test_deepseek_moe_gate` + `test_bcast_moe_reduce_pipeline` — 13 passed, 2 pre-existing skips.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:pmilenkovic/retire-demo-deepseek-moe-gate)


tt-aho (approved):

pmilenkovicTT:
Following up on the review questions with measured data (device-vs-device A/B, fork vs. this PR):

### Outputs — are they exactly the same?

**Routed indices: bit-identical. Scores: not bit-exact, ≤2 bf16 ULPs.**

Methodology: both sides ran on the same BH p100a, same host, with seeded identical inputs across all 12 `test_deepseek_moe_gate` parameter combos (batch ∈ {1,2} × sigmoid ∈ {on,off} × seeds {42, 201, 512}). Raw device output tensors were dumped via `ttnn.to_torch`, canonicalized the same way the test does (slice to top-8, sort by index), and compared bitwise (`int16` view of the bf16 buffers).

- Indices: equal in all 12 combos — the gate routes to exactly the same experts.
- Scores: 62.5% of the 144 top-8 score elements differ, max distance **2 bf16 ULPs** (max abs 3.9e-3 at ~0.5 magnitude). Expected: the canonical family reordered the math (new transpose-dest init, SFPU macro layer), so the rounding path differs — inherent to adopting the productized rewrite rather than a copy.

### Perf — is it the same?

**Device-side: sigmoid path +35–46 cycles (~+30 ns, +2.5–3.3%); non-sigmoid path −21–29 cycles (~−2%). End-to-end wall time unchanged.**

Methodology: `TT_METAL_DEVICE_PROFILER=1`, measuring the **TRISC kernel zone span** (earliest `ZONE_START` to latest `ZONE_END` across TRISC0-2 per program launch) from `profile_log_device.csv`, at the profiler-reported 1350 MHz. Per combo: 5 warm-up launches (JIT + caches), then 100 measured launches; the reported number is the **median** (p10–p90 spread ~±30 cy; medians reproduce within 1 cycle across independent runs). All 12 combos measured on both sides, same card, same host session.

| path | fork (median cy) | canonical (median cy) | delta |
|---|---:|---:|---:|
| sigmoid, b1 | 1403–1405 | 1442–1444 | +2.8% |
| sigmoid, b2 | 1415–1419 | 1454–1463 | +2.5–3.3% |
| no-sigmoid, b1 | 1096–1101 | 1074–1080 | −2.0% |
| no-sigmoid, b2 | 1104–1112 | 1076–1089 | −2.4% |

The sign flips cleanly with `enable_sigmoid` across both batch sizes and all seeds, so it's structural, not noise. Traced: the wrapper-level sigmoid branch is instruction-equivalent between the two families (alias/dead-parameter differences only); the delta lives in the rewritten shared SFPU sort/merge functors (restructured load/store helpers with different ADDR_MOD assignments, plus the templated `topk`/`output_softmax` rank-masking in finalize) interacting differently with each path's pre-stage.

A note on wall-clock: deliberately **not** used as evidence. Host dispatch dominates this op (~130 µs wall vs ~1.1–1.4 µs device), and back-to-back runs drift ±3% with the direction flipping based on run order — measured both orders to confirm. Only the device-profiler numbers are stable.

Context for the +30 ns: the gate runs once per layer per token, so the production-path cost is ~30 ns × layers per token.

@tt-aho Whenever you get the chance to read this, let me know if this makes sense and if we're comfortable with the numbers.

tt-aho:
sigmoid regression is fine to me. We don't use that path in production, as the sigmoid is fused to a preceding MM.

pmilenkovicTT:
Post-rebase verification on the current head (`938a9b66dc1`): [blackhole-sanity run 31725888398](https://github.com/tenstorrent/tt-metal/actions/runs/31725888398).

The only failing job is `sdpa nightly tests (QB2 only)` — identical to main's current failure ([main run 31720311397](https://github.com/tenstorrent/tt-metal/actions/runs/31720311397), same job, same step), so pre-existing and unrelated to this PR.

All four b1 jobs are green, and the tests exercising this PR's rewired kernel pass in both dispatch modes (verified in the job logs, zero skips):
- `test_deepseek_moe_gate` — all 12 parametrizations (batch {1,2} × sigmoid {on,off} × seeds {42,201,512})
- `test_deepseek_moe_gate_uint16_high_bit`

Job sessions: 475 passed (slow dispatch) / 471 passed (fast dispatch); the JIT compile lines confirm the kernels built from the canonical `generalized_moe_gate` includes.

PR #52757, opened 2026-08-10
[Bug fix] Metal 2.0: record dataflow buffers in graph-capture CB accounting

### LINKED ISSUE #51674: Graph capture reports incorrect CB sizes for ops ported to Metal 2.0
Metal 2.0 dataflow buffers (DataflowBufferSpec) are not recorded by TTNN graph capture, so cb_peak_size_per_core reports 0 for ported ops. This issue appeared after PR #51068. The repeat op reports 0 as the cb size. This issue is blocking uplift of metal into tt-mlir.

dgomezTT:
### PR Category

Bug fix

### Summary

Graph capture built its L1 nodes only from `Program::circular_buffers()`. A Metal 2.0 program has no circular buffers — its per-core L1 is dataflow buffers and kernel scratchpads — so nothing was recorded and `cb_peak_size_per_core` came out 0 for every ported op. That zero blocks the metal uplift in tt-mlir.

Fixes #51674.

### What changed

Each kind of memory reports on its own channel, rather than being flattened into the CB bucket:

| kind | tracking call | trace node | metric |
| --- | --- | --- | --- |
| circular buffer | `track_allocate_cb` (unchanged) | `circular_buffer_allocate` | `peak_cb` |
| dataflow buffer | `track_allocate_dataflow_buffer` (new) | `dataflow_buffer_allocate` | `peak_dataflow_buffer` (new) |
| kernel scratchpad | `track_allocate_scratchpad` (new) | `scratchpad_allocate` | `peak_scratchpad` (new) |

`peak_total` and `extract_peak_L1_memory_usage` include all three. The walk lives in `tt_metal/impl/graph/graph_tracking.cpp`, gated on `hook_program()`: a hooked program never runs, so nothing else reports its L1. That is the estimation path (`RunMode::NO_DISPATCH`); a real run's trace is unchanged.

Skipped from the totals: alias secondaries, which share the primary's region, and borrowed dataflow buffers, whose bytes belong to a tensor that is counted in its own right.

### Reviewer notes

**No Metal API change.** `graph_tracking.hpp` grows two virtuals — graph tracking is a TTNN concept that already lives in Metalium. `Program`, `experimental/metal2_host_api/` and the rest of `tt_metal/api/tt-metalium/` are untouched. An earlier revision added a `DataflowBufferFootprint` struct and accessor to the Metal API; that is gone.

**Scratchpads were a second under-report.** `allocate_scratchpads` stacks them on the same L1 region as the dataflow buffers and nothing reported them, so a scratchpad-bearing op (Quasar `pad`, `padded_slice`, `tilize`, `reshape_view`, `matmul`, `interleaved_to_sharded`) claimed less L1 than it uses — the wrong direction for a "does this fit" query. `size_bytes` is known at program-build time, so the estimation path can report it.

**For tt-mlir:** `peakL1MemorySize` becomes correct for ported ops with no change on their side. `cbL1PeakSize` stays 0 — by definition now, not by omission — so reading the per-kind fields is a follow-up there.

**`pretty_format` raised `ValueError` on unknown node types**, so it had to learn the two new ones.

### Testing

- **Estimation path** (`unit_tests_ttnn_mock_allocator`, mock Wormhole): `repeat` reports `dataflow_buffer_peak_size_per_core == 768` with `cb_peak_size_per_core == 0`, that peak reaches `peak_memory_usage_per_core`, two repeats in one capture do not stack, and `relu` is unaffected.
- **Tracking contract** (`unit_tests_api`, mock Quasar): a recording `IGraphProcessor` and blocking `IGraphHooks` assert alias collapse, the borrowed flag, a scratchpad on its own channel (2048 DFB + 1024 scratchpad, no CB), and that an un-hooked program reports nothing.
- **Accounting** (`test_graph_trace_utils.py`, hand-built traces): the three kinds stay separate and all reach `peak_total`, borrowed buffers are excluded, all three are released between programs, and the legacy API counts them. Hand-built because no Wormhole-reachable op produces a scratchpad.
- **End to end** (`test_graph_trace_utils.py`, gated by `ttnn_sanity_tests.yaml`): `dataflow_buffer_allocate` sizes are exactly `[384, 384]`, no `circular_buffer_allocate` node appears, `peak_dataflow_buffer == 768`, `peak_cb == 0`.

Green on silicon, rebased onto current main: 66/66 Quasar, 28/28 mock-allocator, 32/32 graph gtests (6 pre-existing skips), 19/19 `test_graph_trace_utils.py`, 173/173 with `test_graph_report.py`. Negative-controlled by disabling the gate and the scratchpad accumulation: the relevant tests fail, then pass once restored.

Six ported ops go from reporting nothing to reporting under `peak_dataflow_buffer`, with `peak_total` up by the same amount and legacy CB ops untouched: repeat(native) 768, typecast 12288, clone 4096, permute-RM 12288, sort 41984, split 4096.

**Not included:** the `RunMode::NORMAL` trace still omits both kinds. That needs real addresses and changes ttnn-visualizer output, so it is a separate change.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/dfb-graph-capture-no-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/dfb-graph-capture-no-api)

### DIFF
--- a/tt_metal/api/tt-metalium/graph_tracking.hpp
+++ b/tt_metal/api/tt-metalium/graph_tracking.hpp
@@ -65,10 +65,7 @@ public:
         bool /*is_globally_allocated*/,
         const IDevice* /*device*/) {};
 
-    // Metal 2.0 program-scope L1. Reported apart from circular buffers, and apart from each other,
-    // so a consumer can tell which kind of memory it is looking at.
-    // borrows_memory: the buffer is a view onto memory owned elsewhere (a tensor's buffer), which is
-    // tracked in its own right, so summing it again would double-count that L1.
+    // borrows_memory: a view onto a tensor's buffer, so counting its bytes again double-counts.
     virtual void track_allocate_dataflow_buffer(
         const CoreRangeSet& /*core_range_set*/,
         uint64_t /*addr*/,
--- a/tt_metal/api/tt-metalium/graph_tracking.hpp
+++ b/tt_metal/api/tt-metalium/graph_tracking.hpp
@@ -65,6 +65,21 @@ public:
         bool /*is_globally_allocated*/,
         const IDevice* /*device*/) {};
 
+    // Metal 2.0 program-scope L1. Reported apart from circular buffers, and apart from each other,
+    // so a consumer can tell which kind of memory it is looking at.
+    // borrows_memory: the buffer is a view onto memory owned elsewhere (a tensor's buffer), which is
+    // tracked in its own right, so summing it again would double-count that L1.
+    virtual void track_allocate_dataflow_buffer(
+        const CoreRangeSet& /*core_range_set*/,
+        uint64_t /*addr*/,
+        uint64_t /*size*/,
+        bool /*borrows_memory*/,
+        const IDevice* /*device*/) {};
+
+    virtual void track_allocate_scratchpad(
+        const CoreRangeSet& /*core_range_set*/, uint64_t /*addr*/, uint64_t /*size*/, const IDevice* /*device*/) {};
+
+    // Releases every kind of program-scope L1 above: they share one lifetime.
     virtual void track_deallocate_cb(const IDevice* /*device*/) {};
 
     virtual void track_program(tt::tt_metal::Program* /*program*/, const IDevice* /*device*/) {};
@@ -138,6 +153,12 @@ public:
         uint64_t size,
         bool is_globally_allocated,
         const IDevice* device);
+
+    void track_allocate_dataflow_buffer(
+        const CoreRangeSet& core_range_set, uint64_t addr, uint64_t size, bool borrows_memory, const IDevice* device);
+
+    void track_allocate_scratchpad(
+        const CoreRangeSet& core_range_set, uint64_t addr, uint64_t size, const IDevice* device);
 
     void track_deallocate_cb(const IDevice* device);
 
--- a/tt_metal/impl/graph/graph_tracking.cpp
+++ b/tt_metal/impl/graph/graph_tracking.cpp
@@ -9,7 +9,9 @@
 #include <tt_stl/assert.hpp>
 
 #include <tt-metalium/program.hpp>
+#include "impl/context/metal_context.hpp"
 #include "impl/dataflow_buffer/dataflow_buffer_impl.hpp"
+#include "impl/kernels/kernel.hpp"
 #include "impl/program/program_impl.hpp"
 
 namespace tt::tt_metal {
@@ -77,6 +79,26 @@ void GraphTracker::track_allocate_cb(
     }
 }
 
+void GraphTracker::track_allocate_dataflow_buffer(
+    const CoreRangeSet& core_range_set, uint64_t addr, uint64_t size, bool borrows_memory, const IDevice* device) {
+    if (processors.empty()) {
+        return;
+    }
+    for (auto& it : processors) {
+        it->track_allocate_dataflow_buffer(core_range_set, addr, size, borrows_memory, device);
+    }
+}
+
+void GraphTracker::track_allocate_scratchpad(
+    const CoreRangeSet& core_range_set, uint64_t addr, uint64_t size, const IDevice* device) {
+    if (processors.empty()) {
+        return;
+    }
+    for (auto& it : processors) {
+        it->track_allocate_scratchpad(core_range_set, addr, size, device);
+    }
+}
+
 void GraphTracker::track_deallocate_cb(const IDevice* device) {
     if (processors.empty()) {
         return;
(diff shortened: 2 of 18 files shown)


dgomezTT:
Heads-up for the tt-mlir OpModel side, since this changes what a Metal 2.0 op reports in graph capture: @arminaleTT @sgholamiTT @djstefanovicTT

**Decision:** Metal 2.0 memory is reported as its own kinds rather than folded into the circular-buffer bucket. A dataflow buffer and a kernel scratchpad are not circular buffers, so hiding them under `circular_buffer_allocate` would make the trace lie about what the memory is. The capture will gain distinct node types (`dataflow_buffer_allocate`, `scratchpad_allocate`) and distinct per-kind peaks, and `cb_*` goes back to meaning circular buffers only.

**What this means for `OpConstraints`:**

- `cbL1PeakSize` — the value populated from ttnn's `cb_peak_size_per_core` — will legitimately be **0** for any Metal 2.0-ported op, because those programs contain no circular buffers. Today it is 0 too, which is what #51674 reports; the difference is that it will now be 0 *by definition* rather than by omission, with the bytes reported under their own kind.
- `peakL1MemorySize` — from `peak_memory_usage_per_core` — **will** include dataflow buffers and scratchpads. So "does this fit in L1" stays correct without any change on your side. This is the part that was actually broken: for a ported op it was previously missing that L1 entirely.
- New per-kind peaks will be available if you want the breakdown rather than just the total.

**The ask:** if any optimizer or validation decision keys off `cbL1PeakSize` specifically (rather than `peakL1MemorySize`), that will need a small change on the tt-mlir side to also read the dataflow-buffer and scratchpad peaks. Worth confirming before this lands, and happy to align on field names and timing.

For context on why not the simpler route: an earlier revision reported these bytes as circular buffers, which fixed the total in one line but left the trace unable to distinguish the memory kinds — no good for tt-visualizer, and no good for anyone making an informed decision from the numbers.

arminaleTT:
@rpavlovicTT @bmalesevicTT  @odjuricicTT 
Fyi, this might require opModel and/or optimizer changes

tt_metal/impl/graph/graph_tracking.cpp
```diff
@@ -91,6 +113,11 @@ void GraphTracker::track_program(Program* program, const IDevice* device) {
     for (auto& it : processors) {
         it->track_program(program, device);
     }
+    // A hooked program never runs, so no allocation will report its L1 later. Addresses are not
+    // assigned yet either, matching how a capture reports this program's circular buffers.
+    if (hook_program(program)) {
```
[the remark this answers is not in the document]
dgomezTT: The premise is right — `hook_program()` is now called twice per program, once inside `track_program()`
and once in `track_workload()` — but the consequence does not hold for any hook that exists.

There is exactly one implementation in the tree:

```cpp
bool ProcessorHooks::hook_program(tt::tt_metal::Program*) { return do_block; }   // graph_processor.cpp:914
```

It ignores its argument and returns a member flag, so it is a pure predicate: idempotent, no recorded
events, and both calls necessarily agree. `GraphTracker::hook_program` adds only a null check. So there
is no path today where DFB accounting and dispatch can see different answers.

You are right that the interface (`IGraphHooks::hook_program`, a pure virtual) does not *require*
purity, so the redundancy is a latent fragility rather than a current bug. Removing the second call
would mean either changing `GraphTracker::track_program` to return the hook result — a signature change
on a public metalium API header — or exposing the L1 accounting helper so the ttnn side can call it
after its own `hook_program()`. Both are API-shaped changes that are out of scope for a bug fix, so I
am leaving the call structure as is and noting it here.

ttnn/core/graph/graph_trace_utils.cpp
```diff
@@ -83,6 +84,14 @@ uint32_t extract_peak_L1_memory_usage(const nlohmann::json& trace) {
             current_op.push_back(v[kParams][kName]);
         } else if (v[kNodeType] == kNodeCBAllocate) {
```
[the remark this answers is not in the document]
dgomezTT: Fixed in 21ab0cd2b6e. The `kNodeScratchpadAllocate` branch had the same body as the `kNodeCBAllocate`
one (`total_cb += json_to_int(v[kParams][kSize])`), so the two conditions are folded into one — both
are program-scope L1, so the shared body reads correctly for either node type.

tt_metal/api/tt-metalium/graph_tracking.hpp
```diff
@@ -65,6 +65,21 @@ class IGraphProcessor {
         bool /*is_globally_allocated*/,
         const IDevice* /*device*/) {};
 
+    // Metal 2.0 program-scope L1. Reported apart from circular buffers, and apart from each other,
```
akerteszTT: This is a very AI-ful comment for a public header, even one that has no right to exist :).
Consider rejiggering. No strong feelings though, given that this is not an API we expect non-TTNN-infra users to ever invoke.

tt_metal/api/tt-metalium/graph_tracking.hpp
```diff
@@ -65,6 +65,21 @@ class IGraphProcessor {
...
+    // so a consumer can tell which kind of memory it is looking at.
+    // borrows_memory: the buffer is a view onto memory owned elsewhere (a tensor's buffer), which is
+    // tracked in its own right, so summing it again would double-count that L1.
+    virtual void track_allocate_dataflow_buffer(
+        const CoreRangeSet& /*core_range_set*/,
+        uint64_t /*addr*/,
+        uint64_t /*size*/,
+        bool /*borrows_memory*/,
+        const IDevice* /*device*/) {};
```
akerteszTT, line 74: Just a heads up, we're in the process of looking into killing off `IDevice`.

No objection to this new usage here, as it's obviously symmetric with the existing `track_allocate_cb`

akerteszTT (approved):
Thanks for getting keeping the blast radius confined to graph tracking!

tt_metal/api/tt-metalium/graph_tracking.hpp
dgomezTT: Fair, and trimmed in 7da5791ac1e — it's one line now, keeping only the bit a caller can't infer:

```cpp
// borrows_memory: a view onto a tensor's buffer, so counting its bytes again double-counts.
```

The rest was restating the signature. I checked the other comments this PR adds and cut a three-line
block in the run-args test down to two; the `SECTION` banners are the existing convention in those files
so I left those alone.

tt_metal/api/tt-metalium/graph_tracking.hpp
dgomezTT: Understood, thanks for the heads up. This mirrors `track_allocate_cb`'s signature deliberately, so
whenever `IDevice` goes it can be swapped out for both in the same pass rather than leaving the two
hooks inconsistent. Leaving this open in case you want it tracked against that work.

TT-BrianLiu (approved):

Issue #52756, opened 2026-08-10
Host-side EnqueueMeshWorkload cost scales linearly with mesh size (~19 us/device), dominating op2op gap on 32-chip meshes


jbaumanTT:
Sub-issue of #50932, which measured op2op gap at **51.6 %** of warm-iteration time for Kimi K2.6 chunked prefill on the 8×4 (32-chip) Blackhole mesh. This issue covers one specific contributor to that gap: the host-side cost of `EnqueueMeshWorkload` grows **linearly with the number of devices the workload covers**, so a 32-chip Galaxy pays ~4× what an 8-chip box pays for the same op.

## Where the time goes

`FDMeshCommandQueue::enqueue_mesh_workload` (`tt_metal/distributed/fd_mesh_command_queue.cpp`) writes the finalized program command sequence into **each device's hugepage issue queue, one after another, on the calling thread**:

```
for each program:
    update_program_dispatch_commands(...)          # patch the cached cmd sequence, once
    for each local device in the program's range:  # serial, N devices deep
        write_program_command_sequence(...)        # memcpy into that device's issue queue
for each device with no program:
    write_go_signal(...)                           # also serial
```

Every op in the profile in #50932 reports `ndev 32`, i.e. one program replicated across all 32 chips — exactly the shape where this loop is longest.

## Measurements

8×p150b Blackhole host, 2×4 mesh, `main`. Host-side time of `EnqueueMeshWorkload` only (non-blocking, workload reused so nothing recompiles), 200 iterations per sample, 3 repeats. Programs come from `create_benchmark_programs` (max CBs, max RTAs), so per-device cost is on the heavy side. p90 is omitted — it is issue-queue backpressure (~7.6 ms), not host time.

| workload shape | min | p50 |
|---|---|---|
| 1 program on 1 device (+7 go signals) | 28.9 µs | 29.5 µs |
| 1 program on 4 devices (+4 go signals) | 85.3 µs | 86.7 µs |
| 1 program on all 8 devices | 161.0 µs | 162.9 µs |
| 8 distinct programs, one per device | 205.2 µs | 268.0 µs |

The replicated-program row fits `≈ 10 µs + 19.4 µs × devices_covered` almost exactly (predicted 29.4 / 87.6 / 165.2 vs measured 29.5 / 86.7 / 162.9). Go signals to uncovered devices are nearly free; the per-device **program write** is the whole slope.

[...]

Issue #50932, opened 2026-07-23
Kimi K2.6 chunked prefill — per-layer device-time vs op-to-op-gap (L10, first chunk)


pavlepopovic:
# Kimi K2.6 chunked prefill — per-layer device-time vs op-to-op-gap (L10, first chunk)

## Summary
Profiled the Kimi K2.6 chunked-prefill transformer with the Tracy device profiler to
separate **on-device kernel time** from **host-dispatch op-to-op gap**, per layer, for the
first chunk (KV cache empty, `kv_actual=0`), 10 layers, on the 8×4 Blackhole mesh. Each op
runs on all 32 devices at once, so per op we report min/max across devices; per layer we sum
the per-op **max** (critical path).

Headline (warm iteration): **device kernel 175.6 ms**, **op2op gap 187.5 ms**, total **363.0 ms**,
op2op = **51.6 %**. The eager path spends roughly half its time in device compute and half in
per-op host-dispatch gaps. The sum (363 ms) matches the measured warm-iter wall clock (0.356 s),
so the breakdown is self-consistent.

Branch: `ppopovic/op2opgaps` (commit `8cb5cfc1cc4`).

## Setup
- Test: `test_kimi_prefill_transformer_chunked_no_pcc[blackhole-kimi-mesh-8x4-L10-preload0-chunks1-two_iters-margin5pct]`
- Config: Kimi K2.6, `GateComputeMode.DEVICE_FP32`, mesh 8×4 (32 chips) Blackhole, FABRIC_1D.
- 2 iterations: iter 0 = cold compile (discarded), iter 1 = warm (reported).
- Each iteration is bracketed with `synchronize_device()` + `iter_{it}_start/_end` signposts;
  the device profiler buffer is drained once per iteration.
- Profiling instrumentation: deeper nested signposts + in-forward `logger.debug/warning` are
  commented out to cut host-side noise; `forward_layer_{i}_*`, `MLA_START/END`, `MoE_START/END`
  signposts are kept for segmentation. (These edits are on the branch; not intended for `main`.)

## Results (warm iter, 8×4 BH; per-layer = sum of per-op MAX across 32 devices = critical path)

| Layer | ops | device (ms) | op2op (ms) | total (ms) | MLA device (ms) | FFN/MoE device (ms) |
|------:|----:|------------:|-----------:|-----------:|----------------:|--------------------:|
[...]

PR #52775, opened 2026-08-11
#52756: Performance: cut host-side EnqueueMeshWorkload cost on multi-device meshes

jbaumanTT:
### PR Category

Performance

### Summary

Dispatching a `MeshWorkload` cost more the more devices it covered, because every per-device step ran serially on the calling thread. On an 8-device Blackhole mesh a workload spanning the whole mesh took 163 µs of host time, and one with a distinct program per device took 265 µs. That cost lands directly in the eager-mode op-to-op gap.

Relates to #52756 (scaling data), which is a sub-issue of #50932 (op2op gap measured at 51.6 % of warm-iteration time for Kimi K2.6 chunked prefill on a 32-chip mesh).

Four changes:

1. **Skip `compile_and_allocate` once a program is laid out.** It runs on every enqueue, not just the first, and after the first every step inside it early-returns anyway. A flag records that the program is laid out for a device; the `invalidate_{circular,dataflow}_buffer_allocation` hooks and the CB/DFB creation paths clear it, so updating a circular buffer or resizing a dataflow buffer between enqueues still re-runs the layout.
2. **Keep `MeshWorkloadImpl::compile` off the thread pool once the workload is finalized.** Before, more than one program always meant a fan-out and a join, even when every `compile_program` call was about to return immediately. Only the first compile of a multi-program workload does enough work to pay for a worker wake-up.
3. **Block the thread-pool workers on the task counter** instead of polling it with a growing sleep. A worker idle for as little as one dispatch's worth of work was already sleeping in 5 µs steps, so a task could sit tens of microseconds before being noticed.
4. **Write each device's program commands from its own pool task** rather than in a loop. Each device has its own sysmem manager, so the writes are independent; a program's shared dispatch commands are updated once, by whichever of its devices claims it first, while the others wait on that.

Measured on an 8× p150b host (2×4 mesh, 200 enqueues per sample, median of 3 runs, p50 of per-call host time):

| workload shape | before | after | |
|---|---|---|---|
| 1 program, full mesh (8 devices) | 163.1 µs | 92.4 µs | −43% |
| 2 programs, full mesh | 197.0 µs | 95.8 µs | −51% |
| 4 programs, full mesh | 244.5 µs | 98.3 µs | −60% |
| 8 programs, one per device | 264.8 µs | 101.2 µs | −62% |
| 1 program, half the mesh (4 devices) | 86.1 µs | 73.9 µs | −14% |
| 1 program, 1 device | 29.7 µs | 28.3 µs | −5% |

No shape regressed. The program-count penalty is essentially gone: before, going from 1 to 8 programs added 102 µs; now it adds 9 µs.

### Notes for reviewers

**Where to focus — change 1 (the layout flag) is the one that can go wrong.** It is only correct if every path that invalidates a program's CB/DFB layout clears the flag. I hooked all four sites that exist today (`invalidate_circular_buffer_allocation`, `invalidate_dataflow_buffer_allocation`, `add_dataflow_buffer`, and the globally-allocated branch of `add_circular_buffer_`), and the flag is also keyed on the device so a program laid out for one device is not skipped for another. A mutator I have not found would silently skip a needed re-layout. `MeshWorkloadTestSuite.MeshWorkloadCBUpdate` and `MeshWorkloadFactoryHWTest.MakeMeshWorkloadFromSpecSupportsDfbResizeBetweenEnqueues` are the tests that guard this, and both pass — but a second pair of eyes on the invalidation set is worth more than the tests here.

**Change 3 touches shared infrastructure.** `NumaAwareExecutor` backs every user of the dispatch and reader thread pools — buffer reads/writes, event records, program compile, the profiler — not just program dispatch. It also changes idle behavior: workers now block in the kernel rather than polling, which I did not measure the CPU-utilization effect of. `notify_all` rather than `notify_one` is deliberate: both the worker (waiting for work) and a thread in `wait()` (waiting for the count to reach zero) block on the same counter, so waking one could wake the wrong one and strand the worker.

**Behavior change worth knowing about:** a program whose range is entirely on another host previously had its dispatch commands updated during the per-program loop. It still does, but now explicitly on the calling thread, since no device task exists to claim it.

**On the remaining cost.** The per-device work that is left is a bandwidth-bound copy of the command stream into each device's hugepage — 443 KB for these benchmark programs, 99.5 % of it runtime args, at ~25 GB/s per core. Parallelism lifts the aggregate to ~41 GB/s and then hits a platform ceiling, so more threads will not help further; sending fewer runtime-arg bytes is the next lever. Note also that the ~19 µs/device figure quoted in #52756 is specific to these deliberately RTA-heavy benchmark programs — the general rule is command-stream bytes ÷ ~25 GB/s, so a smaller op costs proportionally less.

**Testing.** On an 8× p150b host: `unit_tests_api` 1443/1443, and the `MeshWorkload`, `MeshTrace`, `MeshEvents` and `MeshSubDevice` suites in `distributed_unit_tests` 41/41. The benchmark used for the numbers above is not checked in — it is a `MeshDevice2x4Fixture` variant of the existing `MeshDeviceFixture4x8DispatchAgnostic.ParallelizationBenchmark` that times `EnqueueMeshWorkload` calls individually and reports min/p50. Happy to add it if reviewers want it in-tree.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jbauman/parallel-mesh-workload-dispatch)

### DIFF
--- a/tt_metal/distributed/fd_mesh_command_queue.cpp
+++ b/tt_metal/distributed/fd_mesh_command_queue.cpp
@@ -15,7 +15,9 @@
 #include <tt-metalium/tt_metal.hpp>
 #include <algorithm>
 #include <array>
+#include <exception>
 #include <functional>
+#include <mutex>
 #include <optional>
 #include <type_traits>
 #include <utility>
@@ -100,6 +102,24 @@ void record_program_sub_device_for_range(
             num_available_worker_cores);
     });
 }
+
+// Per-program state shared by the per-device dispatch tasks targeting that program. A program's
+// cached dispatch commands are common to every device it runs on, so they are updated exactly once
+// per workload: the first task to reach the program takes ownership through `prepared` and the rest
+// block there until the update lands. If the owner throws, the flag stays clear and the next waiter
+// retries, so a waiter can never block forever.
+struct ProgramDispatchEntry {
+    Program* program = nullptr;
+    ProgramCommandSequence* cmd_seq = nullptr;
+    std::once_flag prepared;
+};
+
+// One local device and the program it runs. A MeshWorkload's programs cover disjoint device ranges,
+// so a device appears in at most one of these; devices in none of them are sent a go signal instead.
+struct DeviceDispatchTask {
+    IDevice* device = nullptr;
+    ProgramDispatchEntry* program_entry = nullptr;
+};
 
 [[maybe_unused]] MeshCoordinate get_local_start_coord(MeshDevice* mesh_device, const MeshCoordinateRange& range) {
     for (const auto& coord : range) {
@@ -486,8 +506,6 @@ void FDMeshCommandQueue::enqueue_mesh_workload(MeshWorkload& mesh_workload, bool
         program_ordering_sync_count,
         dispatch_metadata);
 
-    std::unordered_set<uint32_t> chip_ids_in_workload = {};
-
     auto max_program_kernels_sizeB = mesh_workload.impl().max_program_kernels_sizeB_;
     bool use_prefetcher_cache = mesh_workload.impl().use_prefetcher_cache_;
     if (use_prefetcher_cache) {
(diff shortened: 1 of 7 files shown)


sidnadTT (approved):

arikTT (approved):

abhullar-tt (approved):

msudumbrekar-TT (approved):
Non blocking nit : BACKOFF_END_ATTEMPT and BACKOFF_FACTOR_MICROSECONDS at thread_pool.cpp:283-284 have no references left so the comment above them still describes the linear backoff. worth dropping both or changing the comment

PR #52776, opened 2026-08-11
Feature: Adding CrossNodeDFB support for WH/BH

abhullar-tt:
### PR Category
Feature

### Summary
GlobalCB concept is split into CrossNodeDFB and GlobalDFB. This PR introduces CrossNodeDFB for WH/BH (Quasar to follow up). 

A CrossNodeDFB is a remote dataflow buffer to sync between nodes within a single program. It is shared across kernels on a sender↔receiver topology, but it is not persistent. Firmware resets rd/wr pointers and credits on every program init. GlobalDFB allow for cross-program persistence (follow-up).

**Memory**
Config buffer: runtime-allocated, height sharded over senders ∪ receivers (one config page per core). Holds topology, FIFO geometry, noc xy / sender coords, and credit slots. Matches GlobalCB
Data buffer: the ring FIFO. Either allocated by CreateCrossNodeDFB, or supplied as borrowed sharded L1 matching that layout. UpdateDynamic retargets the data base (and any relay borrowed base) without reallocating config.

**Setup**
The CrossNodeDFB constructor reads local L1 config and tracks a CrossNodeDFBInterface rather than FW reading the config and writing to global table. This is because adding another global table on WH was causing ncrisc code size to overflow. Per previous convo with @jbaumanTT this should be okay because we don't update the config space before a program is done (this might change in the future)

CrossNode is same-program and reset-on-init, there is no FW/init handshake today. Peers do not need to publish “I’m ready” before kernels run — credit sync begins when user kernels push_back / pop_front.

**Relay DFB (creates remote DM <-> local DM <-> local Compute bridge)**
Host declares a normal local DFB that aliases the CrossNode data ring (experimental API CreateCrossNodeRelayDataflowBuffer). Intended device API:
```
// DM receiver
experimental::CrossNodeDFB cn(remote_dfb_id);
auto relay = cn.bind_relay();
relay.reserve_back(n);
cn.wait_front(n);
relay.push_back(n);
relay.wait_consumed(n);   // wait until TRISC freed these slots
cn.pop_front(n);          // only then ack remote sender

// TRISC. normal local DFB, no CrossNode knowledge
DataflowBuffer relay(dfb::relay_name);
relay.wait_front(n);
// consume ...
relay.pop_front(n);
```
bind_relay() exists so DM does not take a DFBBindingToken for the relay. The host wires the relay device slot into the CrossNode receiver interface; DM opens only that typed relationship. TRISC still uses the normal generated token.

**Posted Writes**
Sender payload writes are posted. After posting, the sender flush_writes() then publishes credits (push_back / NOC-inc pages_sent). Completions are not waited per write; visibility is ordered by the credit path: receivers only wait_front after credits arrive, and credits are published only after the payload flush. Receiver pop_front NOC-acks the sender, freeing ring space. That matches GlobalCB-style remote credit sync and keeps the fast path free of write-ack round trips.

### Notes for reviewers
- Tests don't completely use Metal 2.0. Follow up PR will propagate CrossNodeDFB support to public APIs (at this point `experimental::` will be dropped from CrossNodeDFB on device as well
- GlobalDFB and Quasar support will be done in follow up. 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abhullar/cross-node-dfbs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abhullar/cross-node-dfbs)


tt_metal/api/tt-metalium/mesh_buffer.hpp
```diff
@@ -211,6 +211,10 @@ class AnyBuffer {
     static AnyBuffer create(
         const tt::tt_metal::InterleavedBufferConfig& config, std::optional<uint64_t> address = std::nullopt);
 
+    // Borrow an existing allocation (shares ownership via shared_ptr; no new alloc).
+    static AnyBuffer borrow(std::shared_ptr<Buffer> buffer);
+    static AnyBuffer borrow(std::shared_ptr<MeshBuffer> buffer);
```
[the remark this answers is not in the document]
abhullar-tt: addressed

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -0,0 +1,505 @@
...
+            uint32_t src_addr = src_l1_addr + recv_src_offset;
+            noc.set_async_write_state<NocOptions::POSTED>(
+                dst, coalesced_page_size, {.noc_x = noc_x, .noc_y = noc_y, .addr = dest_addr});
+            for (uint32_t h = 0; h < num_rows; ++h) {
+                const uint32_t row_src_start = src_addr;
+                for (uint32_t w = 0; w < coalesced_num_pages_per_row; ++w) {
+                    noc.async_write_with_state<NocOptions::POSTED>(
+                        CoreLocalMem<uint32_t>(src_addr), dst, coalesced_page_size, {}, {.addr = dest_addr});
+                    src_addr += coalesced_page_size;
+                    dest_addr += coalesced_page_size;
```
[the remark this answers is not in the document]
abhullar-tt: we don't do partial entries - added checks

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -0,0 +1,505 @@
...
+    // Write n entries from src_l1_addr to a single receiver (receiver_idx) at its
+    // current fifo_wr_ptr.  Does NOT advance wr_ptr or increment credits.
+    // Pair with push_back() (collective credit after all per-receiver writes) or
+    // push_back_to_receiver() (per-receiver credit) as appropriate.
+    FORCE_INLINE void write_to_receiver(
+        uint32_t receiver_idx, uint32_t src_l1_addr, uint32_t num_entries, const Noc& noc = Noc{}) {
+        CrossNodeSenderDFBInterface& iface = get_cross_node_sender_dfb_interface(id_);
+        const uint32_t entry_size = iface.fifo_page_size;
+        const uint32_t len_bytes = num_entries * entry_size;
+        const uint32_t dest_l1_base = iface.fifo_wr_ptr;
```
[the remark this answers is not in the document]
abhullar-tt: we don't do partial entries - addressed with check

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -0,0 +1,505 @@
...
+
+    // Broadcast: write n entries of identical data from src_l1_addr to all receivers
+    // at their current fifo_wr_ptr. Uses loop-unicast (hardware NOC multicast requires
+    // a rectangular destination grid). For different bytes per receiver, use
+    // write_to_receiver / write_strided instead.
+    FORCE_INLINE void write_broadcast(uint32_t src_l1_addr, uint32_t num_entries, const Noc& noc = Noc{}) {
+        CrossNodeSenderDFBInterface& iface = get_cross_node_sender_dfb_interface(id_);
+        const uint32_t entry_size = iface.fifo_page_size;
+        const uint32_t len_bytes = num_entries * entry_size;
+        const uint32_t dest_l1_base = iface.fifo_wr_ptr;
```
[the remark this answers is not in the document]
abhullar-tt: we don't split like this - addressed with check

tt_metal/impl/program/program.cpp
```diff
@@ -1288,12 +1289,155 @@ CBHandle detail::ProgramImpl::add_circular_buffer(
...
+                 gdfb.config_address(),
+                 gdfb.credit_reset_address(),
+                 gdfb.credit_reset_size(),
+                 gdfb.entry_size(),
+                 std::numeric_limits<uint8_t>::max()});
+            TT_FATAL(
+                remote_dfb_id == static_cast<uint8_t>(attachments.size() - 1),
+                "CrossNodeDFB remote_dfb_id {} must be dense (expected {}) on core {}",
+                remote_dfb_id,
+                attachments.size() - 1,
```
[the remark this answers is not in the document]
abhullar-tt: addressed

tt_metal/impl/program/dispatch.cpp
```diff
@@ -1431,6 +1494,98 @@ class DataflowBufferCommandGenerator {
...
+            "CrossNodeDFBCommandGenerator: unexpected CROSS_NODE_DFB_OFFSET_NONE with attachments present");
+
+        for (const auto& kg : kernel_groups) {
+            for (const CoreRange& core_range : kg->core_ranges.ranges()) {
+                // Use the first core's attachments as representative for the range.
+                // In typical usage, all cores in a kernel group have identical CrossNodeDFB attachments.
+                const CoreCoord& first_core = core_range.start_coord;
+                auto it = per_core_cross_node_dfbs.find(first_core);
+                if (it == per_core_cross_node_dfbs.end() || it->second.empty()) {
+                    continue;
```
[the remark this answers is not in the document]
abhullar-tt: addressed

tt_metal/impl/program/dispatch.cpp
```diff
@@ -352,6 +354,67 @@ void finalize_dfb_masks(
...
+// If none are attached, launch-msg cross_node_dfb_offset is CROSS_NODE_DFB_OFFSET_NONE.
+uint32_t finalize_cross_node_dfbs(
+    std::vector<std::shared_ptr<KernelGroup>>& kernel_groups, ttsl::Span<ProgramImpl*> programs, uint32_t base_offset) {
+    using AttachmentMap = std::unordered_map<CoreCoord, std::vector<ProgramImpl::CrossNodeDFBAttachment>>;
+    AttachmentMap merged;
+    for (ProgramImpl* program : programs) {
+        for (const auto& [core, attachments] : program->get_per_core_cross_node_dfbs()) {
+            auto& v = merged[core];
+            v.insert(v.end(), attachments.begin(), attachments.end());
+        }
```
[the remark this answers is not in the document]
abhullar-tt: addressed

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -0,0 +1,505 @@
...
+        const uint32_t entry_size = iface.fifo_page_size;
+        const uint32_t fifo_size = get_config_word(iface.config_ptr, 3);
+
+        uint32_t len_bytes = num_entries * entry_size;
+        uint32_t new_wr = iface.fifo_wr_ptr + len_bytes;
+        if (new_wr >= iface.fifo_limit_page_aligned) {
+            len_bytes += iface.fifo_start_addr + fifo_size - iface.fifo_limit_page_aligned;
+            new_wr = iface.fifo_start_addr + (new_wr - iface.fifo_limit_page_aligned);
+        }
+        iface.fifo_wr_ptr = new_wr;
```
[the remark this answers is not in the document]
abhullar-tt: addressed

kstevensTT (approved):

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -0,0 +1,540 @@
...
+//
+// Writes are contiguous: a reserve/write/push of n entries must fit from the current
+// write position to fifo_limit without straddling the wrap (same rule as local CBs). The
+// derived position wraps only when the credited entries reach a multiple of the ring
+// depth. If fewer than n slots remain before the limit, issue a smaller batch (or drain
+// to wrap), then continue.
+//
+// ═══════════════════════════════════════════════════════════════════════
+//  SENDER FLOWS
+// ═══════════════════════════════════════════════════════════════════════
```
jbaumanTT, line 66: Beautiful!

arikTT (approved):

tt_metal/impl/buffers/cross_node_dfb.cpp
```diff
@@ -0,0 +1,450 @@
...
+    // Config uses the same L1 bank class as the borrowed data buffer.
+    allocate_config_and_write_pages(data_buffer.buffer_type());
+}
+
+void CrossNodeDFB::retarget_data_buffer(Buffer& data_buffer) {
+    validate_data_buffer(data_buffer);
+    TT_FATAL(config_buffer_.get_buffer() != nullptr, "CrossNodeDFB config buffer must already exist for retarget");
+    set_data_address(static_cast<uint32_t>(data_buffer.address()));
+    // Keep the existing config allocation; rewrite pages with the new data base address.
+    write_config_pages();
```
jbaumanTT: Could an existing program still be using this config buffer on the device at this point? (potentially launched from a different CQ, which isn't ordered with respect to this one).

tt_metal/impl/program/dispatch.cpp
```diff
@@ -2251,13 +2450,24 @@ void assemble_device_commands(
...
     batched_transfer_generator.construct_commands(batched_transfers, program_config_buffer_calculator);
+    BatchedTransferGenerator cross_node_credit_reset_generator;
+    cross_node_credit_reset_generator.construct_commands(
+        cross_node_credit_reset_transfers, program_config_buffer_calculator);
 
     program_command_sequence.program_config_buffer_command_sequence =
         HostMemDeviceCommand(program_config_buffer_calculator.write_offset_bytes());
     batched_transfer_generator.assemble_commands(
         program_command_sequence, program_command_sequence.program_config_buffer_command_sequence);
+    cross_node_credit_reset_generator.assemble_commands(
```
jbaumanTT: Writing outside of the ringbuffer from a program dispatch command is problematic, since these writes will land before the previous program completes. If two programs in a row use the same program config buffer, this may inadvertently stomp over it while the previous program is using it. We either need to add an explicit wait command to ensure this blocks on the previous program completing, or we need to move the config buffer into the ringbuffer so we can have one copy per program launch.

tt_metal/impl/buffers/cross_node_dfb.cpp
abhullar-tt: yeah this is a real race. Ill update so that we only write the config before program launch

tt_metal/impl/program/dispatch.cpp
abhullar-tt: good point, Ill move this into the ring buffer as well

abhullar-tt: actually I think inserting the wait would be better because adding the config pages there would bloat the ring buffer too much

tt_metal/distributed/fd_mesh_command_queue.cpp
```diff
@@ -461,13 +463,28 @@ void FDMeshCommandQueue::enqueue_mesh_workload(MeshWorkload& mesh_workload, bool
     expected_num_workers_completed_[*sub_device_id] = updated_worker_counts.current;
     expected_num_workers_completed = updated_worker_counts.previous;
 
+    // Only order CrossNode config Buffer rewrites when this workload re-launches a
+    // program that previously used CrossNode on this CQ. Non-zero sync_count means "wait until that prior launch."
+    uint32_t program_ordering_sync_count = 0;
+    for (const auto& [device_range, program] : mesh_workload.get_programs()) {
```
jbaumanTT, line 468: This code only applies to non-trace cases. You'll need to modify SimpleTraceAllocator::allocate_trace_programs_on_subdevice to have similar logic. Also cross_node_program_completion_counts_ will need to be updated update_worker_state_post_trace_execution - either (ideally) with the correct state from the trace, or potentially with expected_num_workers_completed at that point.

tt_metal/impl/program/dispatch.cpp
```diff
@@ -2439,6 +2675,16 @@ void reserve_space_in_kernel_config_buffer(
         reservation = config_buffer_mgr.reserve(program_config_sizes);
     }
 
+    if (program_ordering_sync_count != 0) {
```
jbaumanTT, line 2678: Add this code to handle wrapping the stream counters on the device - so we don't accidentally wait for a sync that's actually intended to be against the distant past, but looks like it's in the future due to wrapping:
```
    // The dispatcher compares stream counters mod 2^MEM_WORD_ADDR_WIDTH with a
    // 2^16 window (stream_wrap_ge), so a count that is far behind â or that
    // predates a counter reset by trace replay / reset_worker_state â reads as
    // "in the future" and stalls forever. Fall back to a full barrier on all
    // prior work, which is always in-window and always at least as strong.
    constexpr uint32_t safe_ordering_window = 1u << 15;
    const int32_t age = static_cast<int32_t>(expected_num_workers_completed - program_ordering_sync_count);
    if (age < 0 || static_cast<uint32_t>(age) > safe_ordering_window) {
        program_ordering_sync_count = expected_num_workers_completed;
    }
```

tt_metal/impl/program/dispatch.cpp
```diff
@@ -2589,6 +2835,20 @@ void update_program_dispatch_commands(
         }
     }
 
+    // Refresh launch-time writes to the dedicated CrossNode config Buffers. This
+    // picks up host-only UpdateDynamic changes and resets credits from the zeroed
+    // host page image on every ordered re-launch.
+    for (const auto& update : cached_program_command_sequence.cross_node_config_updates) {
```
jbaumanTT, line 2851: You also need to modify the TraceNode to store this config_page data on creation, and modify update_traced_program_dispatch_commands to update the commands using the stored information (similar to what we do for CBs).

tt_metal/distributed/fd_mesh_command_queue.cpp
jbaumanTT: Actually we should just set cross_node_program_completion_counts_  = expected_num_workers_completed in update_worker_state_post_trace_execution. We already do mark_completely_full there, so there's no point trying to do anything more fine-grained.

jbaumanTT (approved):

PR #51910, opened 2026-08-03
models: add Kimi K3 delta attention

mvasilijevicTT:
## Intent

Add the Kimi-K3 delta-attention prefill layer with a small model API, MLA-aligned checkpoint/tensor-cache construction, explicit recurrent state, and reviewable correctness and Blackhole LoudBox performance contracts.

## High-level changes

- Add KDA reference, checkpoint, configuration, layer, operation, correctness, and performance-test packages.
- Support 2-D sequence/tensor parallelism. LoudBox is validated at SP1xTP8, SP2xTP4, and SP4xTP2; Galaxy is an intended target but is not yet validated.
- Compute the sequence-parallel affine prefix and causal-convolution halo with ordered linear-topology all-gathers.
- Use standalone output matmul followed by reduce-scatter, matching MLA's execution shape.
- Use fused four-tap QKV convolution and fused gated RMS norm.
- Keep vector-gate KDA scan tuning local. Qwen scalar GDN retains its existing public API and execution contracts.
- Validate PCC and absolute/relative NaN, +Inf, and -Inf counts for both the independent CPU reference and device results. Production T=5120 CPU references are persistently cached with input/configuration validation.

## Architecture and performance decisions

All measurements use pinned real Kimi-K3 layer-1 weights, B=1, T=5120, and 10 synchronized trace replays on Blackhole LoudBox. Candidate PCC is measured against the immediately preceding configuration; latency deltas below are isolated component A/B results, not end-to-end layer gains.

| Decision | SP1xTP8 | SP2xTP4 | SP4xTP2 | Candidate PCC |
|---|---:|---:|---:|---:|
| BF16 affine-summary storage | +0.173% | +0.176% | +0.687% | 0.999429 / 0.999545 / 0.999702 |
| HiFi2 affine-prefix math | -0.228% | +0.299% | -0.276% | 1.0 / 1.0 / 1.0 |
| BF16 grouped-scan output | -0.289% | -0.347% | -0.199% | 0.999989 / 0.999990 / 0.999984 |
| HiFi2 grouped final scan | -0.365% | -0.473% | +0.094% | 0.995901 / 0.995789 / 0.997248 |
| BF16 gated-RMS output | -1.060% | -0.655% | -1.339% | 1.0 / 1.0 / 1.0 |
| HiFi2 output projection | -4.879% | -5.165% | -3.580% | 0.999987 / 0.999988 / 0.999990 |

Additional scoped decisions:

[...]

PR #52783, opened 2026-08-11
[KDA 1/8] Add sigmoid-gated RMSNorm

mvasilijevicTT:
Adds `ttnn.experimental.kda.sigmoid_gated_rms_norm`, one operation from the [full change PR #51910](https://github.com/tenstorrent/tt-metal/pull/51910).

The operation applies per-head RMSNorm, learned scaling, and sigmoid output gating:

`y = (x / sqrt(mean_V(x²) + ε)) · w · sigmoid(g)`

It accepts recurrence output `x[B·H, T, V]`, gate `g[B, T, H·V]`, and weight `w[V]`, and produces `y[B, T, H·V]`. The input keeps `(B·H)` outermost because the recurrence evolves each head independently; the output restores time-first, flattened-head order for the following output projection. This conversion is performed inside the fused kernel rather than through a separate permutation.

The fused implementation measured a 92.72–93.92% latency reduction over the equivalent unfused operations, with output PCC ≥ 0.999990.

### Test plan

[L2 nightly • Blackhole experimental](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch%3Akda-split%2F01-sigmoid-gated-rms-norm)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kda-split/01-sigmoid-gated-rms-norm)


ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+    cb_push_back(cb_out, Vt);
+}
+}  // namespace
+
+void kernel_main() {
+    constexpr uint32_t Vt = get_compile_time_arg_val(0);
+    constexpr uint32_t epsilon_bits = get_compile_time_arg_val(1);
+    constexpr uint32_t inv_v_bits = get_compile_time_arg_val(2);
+    const uint32_t count = get_arg_val<uint32_t>(0);
+    compute_kernel_hw_startup(cb_x, cb_scaler, cb_out);
```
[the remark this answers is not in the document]
mvasilijevicTT: compute_kernel_hw_startup is the first Compute API called, no nede for change.

mbezuljTT (approved):

dpopovTT (approved):

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+    pack_tile(0, cb_inv, 0);
+    tile_regs_release();
+    cb_push_back(cb_inv, 1);
+}
+
+void normalize(uint32_t Vt) {
+    cb_reserve_back(cb_norm, Vt);
+    pack_reconfig_data_format(cb_norm);
+    reconfig_data_format(cb_x, cb_inv);
+    mul_bcast_cols_init_short(cb_x, cb_inv);
```
skrsticTT: this one is deprecated and will be removed soon, can we switch to `mul_bcast_cols_init` to avoid warnings in logs
same for `mul_bcast_rows_init_short` and `mul_tiles_init`

skrsticTT (approved):
attach successful l2 nightly run

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "api/compute/eltwise_unary/rsqrt.h"
+#include "api/compute/eltwise_unary/binop_with_scalar.h"
+#include "api/compute/bcast.h"
+#include "api/compute/tile_move_copy.h"
+#include "api/compute/reconfig_data_format.h"
+#include "api/dataflow/circular_buffer.h"
+#include "ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.hpp"
+
+namespace {
+constexpr uint32_t cb_x = 0, cb_gate = 1, cb_weight = 2, cb_tmp = 3;
```
pavlejosipovic: don't hardcode cb ids these need to passed in through the factor as constexpr args

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+}
+}  // namespace
+
+void kernel_main() {
+    constexpr uint32_t Vt = get_compile_time_arg_val(0);
+    constexpr uint32_t epsilon_bits = get_compile_time_arg_val(1);
+    constexpr uint32_t inv_v_bits = get_compile_time_arg_val(2);
+    const uint32_t count = get_arg_val<uint32_t>(0);
+    compute_kernel_hw_startup(cb_x, cb_scaler, cb_out);
+    CircularBuffer(cb_weight).wait_front(Vt);
```
pavlejosipovic: don't wrap cb handles with ctors, make objects once

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+    rsqrt_tile_init();
+    rsqrt_tile(0);
+    tile_regs_commit();
+    tile_regs_wait();
+    pack_tile(0, cb_inv, 0);
+    tile_regs_release();
+    cb_push_back(cb_inv, 1);
+}
+
+void normalize(uint32_t Vt) {
```
pavlejosipovic: feels more like multiply than normalize?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+
+void inverse_rms(uint32_t epsilon_bits, uint32_t inv_v_bits) {
+    cb_reserve_back(cb_inv, 1);
+    pack_reconfig_data_format(cb_inv);
+    reconfig_data_format_srca(cb_stats);
+    copy_tile_to_dst_init_short(cb_stats);
+    tile_regs_acquire();
+    copy_tile(cb_stats, 0, 0);
+    binop_with_scalar_tile_init();
+    mul_unary_tile(0, inv_v_bits);
```
pavlejosipovic: I would expect to have something along these lines here
init add
add
init mul
mul
init rsqrt
rsqrt

I'm not sure what is binop_with_scalar_tile_init doing?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+#include "api/compute/eltwise_unary/eltwise_unary.h"
+#include "api/compute/eltwise_unary/rsqrt.h"
+#include "api/compute/eltwise_unary/binop_with_scalar.h"
+#include "api/compute/bcast.h"
+#include "api/compute/tile_move_copy.h"
+#include "api/compute/reconfig_data_format.h"
+#include "api/dataflow/circular_buffer.h"
+#include "ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.hpp"
+
+namespace {
```
pavlejosipovic: non need for anonymous namespaces in kernels

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+        CircularBuffer(cb_gate).wait_front(Vt);
+        square(Vt);
+        CircularBuffer(cb_tmp).wait_front(Vt);
+        compute_kernel_lib::reduce<
+            ckernel::PoolType::SUM,
+            ckernel::ReduceDim::REDUCE_ROW,
+            cb_tmp,
+            cb_scaler,
+            cb_stats,
+            compute_kernel_lib::ReduceInputPolicy::NoWaitNoPop,
```
pavlejosipovic: I think you should use default policy for wait and pop (remove explict wait pop and use default policy)

pavlejosipovic: maybe you can use more defaults here in general

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,167 @@
...
+            compute_kernel_lib::ReduceDataFormatReconfigMode::INPUT_AND_OUTPUT,
+            ReduceFp32Mode::Accurate>(compute_kernel_lib::ReduceInputBlockShape::of(1, Vt));
```
pavlejosipovic: I don't think we need ReduceFp32Mode::Accurate, that is probably an overkill?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,34 @@
...
+        input.storage_type() == StorageType::DEVICE && input.buffer() != nullptr,
+        "sigmoid_gated_rms_norm: input must be an allocated device tensor");
+    const auto output_memory_config = memory_config.value_or(ttnn::DRAM_MEMORY_CONFIG);
+    const auto kernel_config = init_device_compute_kernel_config(
+        input.device()->arch(),
+        compute_kernel_config,
+        MathFidelity::HiFi4,
+        /*default_approx_mode=*/true,
+        /*default_fp32_acc=*/true,
+        /*default_l1_acc=*/true);
```
pavlejosipovic: I haven't seen usage of l1 acc in compute kernle?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/sigmoid_gated_rms_norm.cpp
```diff
@@ -0,0 +1,34 @@
...
+    DataType output_dtype) {
+    TT_FATAL(
+        input.storage_type() == StorageType::DEVICE && input.buffer() != nullptr,
+        "sigmoid_gated_rms_norm: input must be an allocated device tensor");
+    const auto output_memory_config = memory_config.value_or(ttnn::DRAM_MEMORY_CONFIG);
+    const auto kernel_config = init_device_compute_kernel_config(
+        input.device()->arch(),
+        compute_kernel_config,
+        MathFidelity::HiFi4,
+        /*default_approx_mode=*/true,
```
pavlejosipovic, line 27: is this apporox here affecting rsqrt accuracy?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "sigmoid_gated_rms_norm_program_factory.hpp"
+
+#include <cstring>
+#include <vector>
+
+#include <tt-metalium/buffer.hpp>
+#include <tt-metalium/constants.hpp>
+#include <tt-metalium/tensor_accessor_args.hpp>
+
+#include "../../device/kda_factory_utils.hpp"
```
pavlejosipovic: we don't use relative includes

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+using namespace tt::tt_metal;
```
pavlejosipovic, line 20: avoid using ..

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+    ProgramDescriptor desc;
+    auto add_cb = [&](uint32_t idx, uint32_t tiles, tt::DataFormat format, uint32_t buffers = 1) {
+        const uint32_t tile_size = tt::tile_size(format);
+        desc.cbs.push_back(CBDescriptor{
+            .total_size = tiles * buffers * tile_size,
+            .core_ranges = cores,
+            .format_descriptors = {{CBFormatDescriptor{
+                .buffer_index = static_cast<uint8_t>(idx), .data_format = format, .page_size = tile_size}}}});
+    };
+    add_cb(0, Vt, datatype_to_dataformat_converter(in.input.dtype()));
```
pavlejosipovic: don't hardcode cb ids, make a local var and keep incremeting it on aeach add_cb call

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+    // Use the fewest workers that preserve the all-core maximum items/worker.
+    const auto grid = in.input.device()->compute_with_storage_grid_size();
+    const uint32_t max_items_per_core = tt::div_up(total, grid.x * grid.y);
+    const uint32_t rms_core_limit = tt::div_up(total, max_items_per_core);
+    auto dist =
+        kda_factory_detail::distribute_prep(in.input.device()->compute_with_storage_grid_size(), total, rms_core_limit);
+    const auto& cores = dist.core_set;
+
+    ProgramDescriptor desc;
+    auto add_cb = [&](uint32_t idx, uint32_t tiles, tt::DataFormat format, uint32_t buffers = 1) {
```
pavlejosipovic: it doesn't seem like you have double buffering on CBs anywhere?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+    auto* output_buffer = outputs[0].buffer();
+    for (uint32_t i = 0; i < dist.cores.size(); i++) {
+        const auto& core = dist.cores[i];
+        reader.emplace_runtime_args(
+            core, {dist.wi_start[i], dist.wi_count[i], input_buffer, gate_buffer, weight_buffer});
+        writer.emplace_runtime_args(core, {dist.wi_start[i], dist.wi_count[i], output_buffer});
+        compute.emplace_runtime_args(core, {dist.wi_count[i]});
+    }
+
+    desc.kernels.push_back(std::move(reader));
```
pavlejosipovic: push_back enforces a copy, if you want move semantics do emplace_back

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+    auto* input_buffer = in.input.buffer();
+    auto* gate_buffer = in.gate.buffer();
+    auto* weight_buffer = in.weight.buffer();
+    auto* output_buffer = outputs[0].buffer();
+    for (uint32_t i = 0; i < dist.cores.size(); i++) {
+        const auto& core = dist.cores[i];
+        reader.emplace_runtime_args(
+            core, {dist.wi_start[i], dist.wi_count[i], input_buffer, gate_buffer, weight_buffer});
+        writer.emplace_runtime_args(core, {dist.wi_start[i], dist.wi_count[i], output_buffer});
```
pavlejosipovic: do we need any runtime args/tensor patching on program cache hit?

ttnn/cpp/ttnn/operations/experimental/kda/factory/kda_factory_utils.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include <vector>
+
+#include <tt-metalium/core_coord.hpp>
+#include <tt-metalium/program_descriptors.hpp>
+
+#include "ttnn/operations/core/compute_kernel/compute_kernel_config.hpp"
+
+namespace ttnn::experimental::prim::kda_factory_detail {
```
pavlejosipovic, line 10: maybe rename device folder to some sort of kda factory utils?

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
```diff
@@ -0,0 +1,268 @@
...
+    epsilon: float,
+    output_dtype: ttnn.DataType,
+    message: str,
+) -> None:
+    _, (input_tt, gate_tt, weight_tt) = _device_inputs(device)
+    with expect_error(RuntimeError, message):
+        _run(input_tt, gate_tt, weight_tt, num_heads=num_heads, epsilon=epsilon, output_dtype=output_dtype)
+
+
+def test_sigmoid_gated_rms_norm_rejects_sharded_output(device: ttnn.Device, expect_error: Callable) -> None:
```
pavlejosipovic, line 583: can we add determinism/perf coverage?

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: mul_unary_tile and add_unary_tile are scalar SFPU operations on DST. They share binop_with_scalar_tile_init().

pavlejosipovic: lets try to avoid SFPU, this should be doable on FPU

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: Migrated the operation fully to Metal 2.0 `ProgramSpec` and `TT_KERNEL`. CB identity now comes from named DFB bindings, so numeric CB IDs no longer cross the host/kernel boundary as compile-time arguments.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: Switched to the default `WaitAndPopPerTile` policy. It tied `BulkWaitBulkPop` within measurement noise and was faster than the manual `NoWaitNoPop` path on all three production shapes, with unchanged accuracy.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: A/B testing showed `Fast` was about 5% faster than `Accurate` on all three production shapes with no accuracy loss, so the kernel now uses the default `Fast` mode.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: This kernel packs each final output tile once and has no partial-output accumulation lifecycle, so `packer_l1_acc=true` is unsupported and now rejected during validation. An omitted config uses the production defaults; supported explicit compute settings are honored.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: A/B testing showed approximate math was about 2.5% faster than exact math without an accuracy regression, so the default approximate mode is retained. A forced-JIT regression also verifies that exact mode produces a distinct program and output.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
mvasilijevicTT: The Metal 2.0 migration removes host-assigned numeric CB IDs entirely. The factory now declares semantic DFBs, and kernels access them through named `DataflowBuffer(dfb::<name>)` bindings.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
mvasilijevicTT: A/B testing selected double buffering for the streamed `x`, `gate`, and `out` buffers. Double-buffering every CB was tied within noise but used an additional 52 KiB/core, because setup-only and compute-local scratch buffers have no concurrent producer to overlap.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/sigmoid_gated_rms_norm_program_factory.cpp
mvasilijevicTT: With Metal 2.0, the standard adapter patches fresh named tensor bindings through `UpdateTensorArgs` on cache hits, so no custom override is needed.

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
mvasilijevicTT: Added separate eager accuracy/determinism and performance coverage for the three production workloads. The current forced-JIT module passes 32/32, repeated outputs are bit-identical, and the performance checks use seven-sample device-time baselines with a 5% margin. CI configuration remains outside this PR.

ttnn/cpp/ttnn/operations/experimental/kda/sigmoid_gated_rms_norm/device/kernels/compute/sigmoid_gated_rms_norm.cpp
mvasilijevicTT: Updated. `sigmoid` and `rsqrt` are the only operations on SFPU now. Everything else goes through FPU.

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
```diff
@@ -0,0 +1,569 @@
...
+    *,
+    count: int = 3,
+) -> tuple[list[ttnn.Tensor], list[torch.Tensor]]:
+    device_outputs = []
+    host_outputs = []
+    cache_entries = None
+    for _ in range(count):
+        output_tt = run()
+        ttnn.synchronize_device(device)
+        output = ttnn.to_torch(output_tt).clone()
```
pavlejosipovic: Could we mirror the device-side determinism check in `tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py` here? Keep the first output device-resident, compare subsequent outputs with `ttnn.ne`, reduce to an on-device mismatch marker, and transfer only that scalar after the loop. This avoids three full output D2H copies for the production shapes; one host copy is still enough for the accuracy assertion and mismatch diagnostics.

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
```diff
@@ -0,0 +1,569 @@
...
+            ttnn.WormholeComputeKernelConfig(
+                math_fidelity=ttnn.MathFidelity.HiFi4,
+                packer_l1_acc=True,
+            ),
+            "packer_l1_acc=true is unsupported",
+        ),
+        (
+            ttnn.WormholeComputeKernelConfig(
+                math_fidelity=ttnn.MathFidelity.HiFi4,
+                throttle_level=ttnn.ThrottleLevel.LEVEL_1,
```
pavlejosipovic: Let’s remove the op-specific `throttle_level` validation and this test case. The kernel does not consume this setting, and it is not translated into the Metal 2 compute hardware configuration, so it does not belong to the sigmoid-gated RMSNorm contract.

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
```diff
@@ -0,0 +1,569 @@
...
+    exact = ttnn.to_torch(_run(input_tt, gate_tt, weight_tt, compute_kernel_config=exact_config))
+    assert device.num_program_cache_entries() == entries + 1
+    assert not torch.equal(approximate, exact)
+
+
+@pytest.mark.parametrize(
+    ("compute_kernel_config", "message"),
+    [
+        (
+            ttnn.WormholeComputeKernelConfig(
```
pavlejosipovic: Use `ttnn.BlackholeComputeKernelConfig` here instead of `ttnn.WormholeComputeKernelConfig`; this is a Blackhole nightly test and the current name is misleading.

pavlejosipovic (approved):

tests/ttnn/nightly/unit_tests/operations/experimental/kda/test_sigmoid_gated_rms_norm.py
```diff
@@ -0,0 +1,569 @@
...
+        num_heads=case.num_heads,
+        value_dim=_PRODUCTION_VALUE_DIM,
+        output_dtype=_PRODUCTION_OUTPUT_DTYPE,
+    )
+    _assert_accurate_and_deterministic(expected, host_outputs, name=case.case_id)
+    _assert_inputs_unchanged(input_snapshots, device_inputs)
+
+
+@pytest.mark.parametrize("case", _PRODUCTION_CASES, ids=lambda case: case.case_id)
+@skip_with_llk_assert("No need to verify LLK asserts for performance tests.")
```
pavlejosipovic, line 348: Please add `@pytest.mark.requires_host_iommu`, following `test_topk_large_indices_production_perf_check`. This routes the test to the marker-selected `bh_p150b_civ2_viommu` job, where the real-time profiler is expected to be active; the regular Blackhole experimental job excludes `performance` tests.

Issue #12349, opened 2024-09-06
[Blackhole Ops] Fast Dispatch Post Commit Issue Master List


abhullar-tt:
This is a master list of all tests that needed to be disabled to enable passing `fast-dispatch-build-and-unit-tests.yaml` on BH

Goal: Remove instances of `skip_for_blackhole` that are not skipped for WH

**Matmul**
- [x] test_bert_ops.py::test_bert_linear_batch4 hanging #12597
- [x] sweep_tests/pytests/helper_funcs/test_linear.py::test_linear_with_bias PCC failure with input shape [1, 1, 64, 128] #12539
- [x] misc/test_bert_ops.py::test_bert_linear_batch4_fp32_input_output hanging #12598
- [x] misc/test_attn_matmul.py hanging #12599
- [x] test_sharded.py::test_sharded_matmul_2d hanging #12600
- [x] misc/test_tilize_hpadding_matmul.py hanging #12601
- [x] misc/test_tilize_matmul.py hanging #12602
- [x] misc/test_matmul_dram_sharded.py segfault #12603
- [x] sweep_tests/pytests/tt_dnn/test_matmul.py::test_run_matmul_test some configs hanging #12604
- [x] unit_testing/misc/test_sharded.py::test_resharded_binary_to_matmul hanging - will be fixed with other matmul hangs with #13981 

**Softmax**
- [x] misc/test_softmax_sharded.py::test_scale_mask_softmax_rm mismatch https://github.com/tenstorrent/tt-metal/issues/14350
- [x] misc/test_single_core_fused_ops.py::test_softmax mismatch https://github.com/tenstorrent/tt-metal/issues/14351

**LayerNorm**
- [x] misc/test_distributed_layernorm_post_allgather.py mismatching https://github.com/tenstorrent/tt-metal/issues/14354
- [x] misc/test_single_core_fused_ops.py::test_layernorm mismatching https://github.com/tenstorrent/tt-metal/issues/14355

**Moreh Ops**
- [ ] misc/test_moreh_getitem.py mismatching
- [x] misc/test_moreh_sum.py::test_moreh_sum_integer mismatching
- [ ] misc/test_moreh_layernorm.py::test_moreh_layernorm_backward mismatching
- [ ] misc/test_moreh_nll_loss.py mismatching
- [ ] misc/test_moreh_nll_loss_unreduced.py mismatching

**Loss Ops**
- [ ] unit_testing/loss_ops/test_loss_mae.py mismatching
[...]

PR #52789, opened 2026-08-11
Fix update_cache for non-tile-aligned batch sizes

### LINKED ISSUE #52671: [ttnn/kv_cache]: update_cache silently corrupts the cache when the batch is above 32 and not a multiple of 32
## Component / Area

ops (ttnn/kv_cache)

## Issue Type

Bad Outputs (non-deterministic)

## Observed

`ttnn.update_cache` writes the input tensor's tile padding over real users whenever the
cache batch `Bcache = cache.padded_shape()[0]` is greater than 32 and not a multiple of 32.
Affected user rows come back as zeros. Nothing rejects the shape, and nothing warns.

For several sizes the result is also non-deterministic: the same call with the same inputs
returns a different number of corrupted rows on each dispatch, and occasionally returns the
fully correct answer, which makes it very easy to miss in testing.

Probe: user row `u` of the input is filled with the constant `u + 1`, the cache is
pre-filled with `-1`, and after the update `cache[u, 0, 0, 0]` says what landed on user `u`
(`u+1` correct, `0` the input's padding, `-1` never written).

| Bcache | input rows after tile padding | correct rows | rows overwritten with padding zeros |
| --- | --- | --- | --- |
| 32 | 32 | 32 | 0 |
| 33 | 64 | varies, 2 to 33 | varies, 0 to 31 |
| 34 | 64 | 4 | 30 |
| 40 | 64 | 16 | 24 |
| 48 | 64 | 32 | 16 |
| 63 | 64 | 62 | 1 |
| 64 | 64 | 64 | 0 |
| 65 | 96 | varies, 34 to 65 | varies, 0 to 31 |
| 80 | 96 | 64 | 16 |
| 96 | 96 | 96 | 0 |
| 97 | 128 | 67 | 30 |

Only the input's padding lands in the wrong place: no other user's data is written to the
wrong row, and nothing outside the target sequence position is touched. `num_heads` 1, 2
and 4 behave identically.

## Expected

Either every user's row is updated correctly for any `Bcache` the validator accepts, or
`validate_on_program_cache_miss` rejects the unsupported shape with a clear `TT_FATAL`.
Silently returning a partially zeroed KV cache is the worst of the three.

## Root cause

`update_cache_multi_core_program_factory.cpp`:

```cpp
uint32_t B = input_tensor.padded_shape()[-2];                                  // line 155
uint32_t Bcache = cache_tensor.padded_shape()[0];                              // line 156
uint32_t num_batched_heads = input_tensor.padded_shape()[1] * B / TILE_HEIGHT; // line 158
...
const uint32_t u_range = std::min(static_cast<uint32_t>(32), Bcache);          // line 278
const uint32_t u_count = u_range / granularity;                                // line 279
```

`B` is the input's *padded* row count, so it is `ceil(Bcache / 32) * 32` once `Bcache > 32`,
and `num_batched_heads` counts `ceil(Bcache / 32)` groups per head. Every group walks
`u_count * granularity == min(32, Bcache) == 32` user rows, so the kernels write

```
num_heads * ceil(Bcache / 32) * 32
```

user rows into a cache that only holds `num_heads * Bcache` of them. The writer receives
`Bcache` as its `B` runtime argument (`writer_desc.emplace_runtime_args(core, {dst_buffer, Wt, Bcache, ...})`)
and wraps with `if (b == B) { b = 0; ... }`, so the surplus
`ceil(Bcache / 32) * 32 - Bcache` rows wrap back onto users 0, 1, 2 ... carrying the input
tensor's tile padding. That surplus is exactly the count of corrupted rows measured above
for the deterministic sizes: 34 gives 30, 40 gives 24, 48 gives 16, 63 gives 1, 80 gives 16.

The last group must be short. Nothing currently makes it short.

This is a different root cause from item 2 of #51227, which is the `granularity` /
`u_range` truncation for odd `Bcache < 32`. The two are independent: the fix for that one
(#52667) cannot change this path, because for any `Bcache >= 32` it still yields
`u_range == 32` and `granularity == 2`.

## Steps to reproduce

```python
import torch
import ttnn

dev = ttnn.open_device(device_id=0, l1_small_size=2**15)

num_users, num_heads, head_dim, max_seq_len, cache_idx = 34, 1, 64, 64, 0

cache = torch.full([num_users, num_heads, max_seq_len, head_dim], -1.0)
cachett = ttnn.Tensor(cache.bfloat16().float(), ttnn.bfloat16).to(ttnn.TILE_LAYOUT).to(dev)

x = torch.zeros([num_users, num_heads, 1, head_dim])
for u in range(num_users):
    x[u, :, 0, :] = float(u + 1)
xt = ttnn.Tensor(x.permute(2, 1, 0, 3), ttnn.bfloat16).to(ttnn.TILE_LAYOUT).to(dev)

cachett = ttnn.update_cache(cachett, xt, cache_idx)
got = cachett.cpu().to(ttnn.ROW_MAJOR_LAYOUT).to_torch().float()

landed = [float(got[u, 0, cache_idx, 0]) for u in range(num_users)]
print("correct:", sum(1 for u in range(num_users) if landed[u] == u + 1), "of", num_users)
print("zeroed :", [u for u in range(num_users) if landed[u] == 0.0])

ttnn.close_device(dev)
```

Prints `correct: 4 of 34` and `zeroed : [1, ..., 30]`.

## Input data

None. The snippet above generates everything it needs.

## Frequency

`Bcache` of 34, 40, 48, 63, 80 and 97: always, and identical on every repeat. `Bcache = 34`
gave the same 4 correct of 34 on five consecutive dispatches.

`Bcache` of 33 and 65: intermittent. Five consecutive dispatches each gave a different
answer.

| repeat | Bcache 33, correct of 33 | Bcache 65, correct of 65 |
| --- | --- | --- |
| 1 | 31 | 65 |
| 2 | 9 | 34 |
| 3 | 32 | 56 |
| 4 | 33 | 43 |
| 5 | 2 | 46 |

`Bcache` of 32, 64 and 96: always correct.

## Why this is latent

`test_update_cache_decode` and `test_update_cache_decode_program_cache_hits` both skip the
whole range:

```python
if num_users > 32 or (num_users + batch_offset) > 32:
    pytest.skip("Batch offset is only used when num_users < 32 and batch_offset + num_users <= 32")
```

and `UpdateKVCacheOperation::validate_on_program_cache_miss` only requires
`cache_tensor.padded_shape()[0] <= input_tensor.padded_shape()[-2]`, plus
`batch_offset + Bcache <= 32` when `Bcache < 32`. No parity and no divisibility constraint,
so every shape in the table above passes validation.

## Suggested fix

Minimal and safe: reject the shape in `validate_on_program_cache_miss`.

```cpp
TT_FATAL(
    Bcache <= 32 || Bcache % 32 == 0,
    "update_cache requires the cache batch to be at most 32 or a multiple of 32, got {}",
    Bcache);
```

Full support: make the per-group user count depend on the group index so the last group is
short, i.e. `min(32, Bcache - group * 32)` instead of `min(32, Bcache)`, and pass it through
to the reader, compute and writer as a per-core runtime argument rather than a single
compile-time `u_count`.

## System details

Software: Ubuntu 22.04, tt-metal built from source, `ttnn` Python API.
Hardware: Wormhole B0, N300 (2 chips), device 0.

Verified against code byte-identical to `main` for the lines quoted above.

kpaigwar:
## Summary
- Fixes silent cache corruption in `ttnn.update_cache` when `Bcache > 32` and not a multiple of `TILE_HEIGHT` (e.g. 34), and when odd `Bcache` left a granularity remainder unupdated.
- Reader / compute / writer now derive per-group user count at runtime as `min(32, Bcache - batch_idx)` and drain remainders with `g = min(granularity, users_remaining)`.
- Validates that input batch dim equals `round_up(Bcache, 32)` and adds decode coverage for batches `[7, 33, 34, 40]`.

Fixes #52671

## Test plan
- [ ] `pytest tests/tt_eager/python_api_testing/unit_testing/misc/test_update_cache.py::test_update_cache_decode_non_tile_batch -v`
- [ ] `pytest tests/tt_eager/python_api_testing/unit_testing/misc/test_update_cache.py::TestUpdateCache::test_update_cache_decode -v` (tile-aligned / `<32` regression)
- [ ] Confirm Wormhole and Blackhole if available (existing decode path was previously skipped on BH for #12349)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:fix/update-cache-non-tile-batch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:fix/update-cache-non-tile-batch)

### DIFF
--- a/tests/tt_eager/python_api_testing/unit_testing/misc/test_update_cache.py
+++ b/tests/tt_eager/python_api_testing/unit_testing/misc/test_update_cache.py
@@ -10,6 +10,23 @@ from tests.tt_eager.python_api_testing.sweep_tests.comparison_funcs import comp_
 from models.common.utility_functions import skip_for_blackhole
 
 
+def _pad_decode_input_batch(x, num_users, num_heads, head_dim, batch_offset):
+    """Pad decode input batch on dim0 to round_up(num_users, 32) for tilize.
+
+    When batch_offset > 0 (only valid for num_users < 32), real users sit at
+    [batch_offset, batch_offset + num_users) inside a 32-wide tile.
+    """
+    x_new = x.clone()
+    pad_users = ((num_users + 31) // 32) * 32
+    if batch_offset > 0:
+        x_new = torch.cat((torch.zeros(batch_offset, num_heads, 1, head_dim), x_new), dim=0)
+    pad_after = pad_users - x_new.shape[0]
+    if pad_after > 0:
+        x_new = torch.cat((x_new, torch.zeros(pad_after, num_heads, 1, head_dim)), dim=0)
+    assert x_new.shape[0] == pad_users, f"Expected x.shape[0] to be {pad_users}, got {x_new.shape[0]}"
+    return x_new, pad_users
+
+
 @pytest.mark.parametrize("head_dim", [64])
 @pytest.mark.parametrize("max_seq_len", [4096])
 @pytest.mark.parametrize("num_users", [8, 16, 32, 64])
@@ -120,7 +137,7 @@ class TestUpdateCache:
         cache_dtype,
         device,
     ):
-        if num_users > 32 or (num_users + batch_offset) > 32:
+        if batch_offset != 0 and (num_users >= 32 or (num_users + batch_offset) > 32):
             pytest.skip("Batch offset is only used when num_users < 32 and batch_offset + num_users <= 32")
         if cache_dtype != ttnn.bfloat16:
             pytest.skip(
@@ -131,16 +148,11 @@ class TestUpdateCache:
         cache = torch.randn(cache_shape).bfloat16().float()
         cachett = ttnn.Tensor(cache, cache_dtype).to(ttnn.TILE_LAYOUT).to(device)
         x = torch.randn(input_shape).bfloat16().float()
-        # pad dim0 of x to 32 if batch size is less than 32, make 0-batch_offset elements 0, batch_offset-batch_offset+num_users elements non-zero, and rest 0
-        x_new = x.clone()
-        if num_users < 32:
-            x_new = torch.cat((torch.zeros(batch_offset, num_heads, 1, head_dim), x_new), dim=0)
-            x_new = torch.cat((x_new, torch.zeros(32 - num_users - batch_offset, num_heads, 1, head_dim)), dim=0)
-            assert x_new.shape[0] == 32, f"Expected x.shape[0] to be 32, got {x_new.shape[0]}"
+        x_new, pad_users = _pad_decode_input_batch(x, num_users, num_heads, head_dim, batch_offset)
         xt = ttnn.Tensor(x_new.permute(2, 1, 0, 3), input_dtype).to(ttnn.TILE_LAYOUT)
         if in_sharded:
             compute_grid_size = device.compute_with_storage_grid_size()
-            num_cores = min(max(num_users, 32) // 32 * num_heads, compute_grid_size.x * compute_grid_size.y)
+            num_cores = min(pad_users // 32 * num_heads, compute_grid_size.x * compute_grid_size.y)
             shard_grid = ttnn.num_cores_to_corerangeset(num_cores, compute_grid_size, True)
             input_shard_spec = ttnn.ShardSpec(
                 shard_grid,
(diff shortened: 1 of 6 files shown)


AJ0070:
Desk-checked the host arithmetic against the kernel-side derivation for Bcache 2..65. The granularity guard and the `users_full`/`users_last` split come out correct in every case I tried, including the awkward ones: Bcache=33 gives g=1, full=32, last=1, and Bcache=63 gives g=1, full=32, last=31.

One overlap worth flagging. #52667 touches the same granularity lines to fix a narrower case: for odd Bcache in 3..31, `min(2, Bcache)` leaves `u_count * granularity == u_range - 1`, so the last user's row is never written. Your `users_full % granularity` guard covers that case as well, and also covers Bcache > 32, which #52667 does not. So yours is the more general fix and the two will conflict on this file. The randn half of #52667 is independent of this.

Happy to re-verify on a p150b once this lands.

PR #52811, opened 2026-08-11
ttnn.div int32: guard INT32_MIN residual conversion (fixes #51476)

### LINKED ISSUE #51476: ttnn.div int32 (trunc/floor): dividend -2**31 returns -1 — the INT32_MIN guard is missing at one of three identical conversion sites
# ttnn.div int32 (trunc/floor): dividend -2**31 returns -1 — the INT32_MIN guard is missing at one of three identical conversion sites

*(WH and BH, up to 1023 absolute / 99.90% relative error, inside the range the op's own $10000 bounty
specified as "the full 32 bit range")*

`calculate_div_int32_body()` knows that `sfpi::abs(-2**31)` returns `0x80000000` and that converting
that to float yields `-0.0f`. Its own comment says so, and it clamps the two conversions that come
from it. There is a third conversion of exactly the same kind, and it has no clamp:

```c
sfpi::vFloat b_f = sfpi::convert<sfpi::vFloat>(b, sfpi::RoundMode::Nearest);
v_if(b_f < 0.0f) { b_f = 2147483648.0f; }                            // line 38  guarded
v_endif;
...
sfpi::vFloat a_f = sfpi::convert<sfpi::vFloat>(sfpi::abs(a_s), sfpi::RoundMode::Nearest);
v_if(a_f < 0.0f) { a_f = 0x1.0p31f; }                                // line 74  guarded
v_endif;
...
sfpi::vFloat r_f = sfpi::convert<sfpi::vFloat>(sfpi::abs(r), sfpi::RoundMode::Nearest);
                                                                     // line 109 NOT guarded
```

`r_f` becomes `-0.0f`, the residual correction is computed as `-0.0f * inv_b_f`, rounds to the
integer `0`, and the entire quotient is lost. What survives is only the ±1 the recovery step adds,
so for `a = -2147483648` the op returns `-1` (trunc) or `-2` (floor) whatever the divisor, where the
true quotient runs up to `±1024`:

```
ttnn.div(-2147483648, 2097152, rounding_mode="trunc")    ->  -1      torch: -1024
ttnn.div(-2147483648, 2097153, rounding_mode="trunc")    ->  -1      torch: -1023
ttnn.div(-2147483648, 239823930, rounding_mode="trunc")  ->  -1      torch:    -8
ttnn.div(-2147483648, -2097152, rounding_mode="trunc")   ->   1      torch:  1024
ttnn.div(-2147483648, 2097152, rounding_mode="floor")    ->  -2      torch: -1024
```

This is not a rounding argument. Across the whole affected band the returned magnitude is pinned at
`1` while the true magnitude is anything from `2` to `1024` — a 99.90% relative error at the worst
point, on ordinary in-range integers, against an op whose unit test asserts bit-exact equality with
the torch golden. Both Wormhole and Blackhole are affected (the same missing guard, at different
thresholds — see below).

**Against the op's own stated acceptance bar.** [#33241](https://github.com/tenstorrent/tt-metal/issues/33241)
("[Bounty $10000] Full 32 bit int32 div with rounding (trunc/floor) support") asks for "precise
results for the full 32 bit range when using the rounding modes". The delivering PR
[#33398](https://github.com/tenstorrent/tt-metal/pull/33398) enumerates the inputs it considers
undefined, and it lists exactly two: `div(a, 0)`, and `div(-2147483648, -1)`. Neither covers
`div(-2147483648, b)` for the other 4.29 billion divisors — `-2147483648 / 2097153 = -1023` is
perfectly representable in int32. The op's unit test asserts `assert_equal`, i.e. bit-exact
agreement with the torch golden, so there is no accuracy budget to spend here either.

---

### Summary

Files (both are current `main`, and both carry the defect):

- [`tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_div_int32_floor.h`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_div_int32_floor.h) — guards at **L38** and **L74**, missing at **L109**
- [`tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_div_int32_floor.h`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_div_int32_floor.h) — guards at **L33** and **L47**, missing at **L68**

The two files are *not* the same implementation — WH uses a software reciprocal with 11-bit chunked
`SFPMAD` products and a `2**34` mantissa alignment, BH uses `SFPARECIP` + `SFPMUL24` and a `2**33`
alignment — but the guard pattern, and the omission, are identical in both.

Every public entry point reaches this kernel. Both `ttnn.div` overloads route int32 +
`rounding_mode` to the same op, so there is no alternative path that gets it right:

- `div(Tensor, ScalarVariant, ..., rounding_mode)` → `BinaryOpType::DIV_FLOOR` / `DIV_TRUNC`
  — [`binary_composite_op.cpp:149-215`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/binary/device/binary_composite_op.cpp#L149-L215)
- `div(Tensor, Tensor, ..., rounding_mode)` → the same two
  — [`binary_composite_op.cpp:272-340`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/binary/device/binary_composite_op.cpp#L272-L340)

(This is worth stating because the analogous scalar/tensor split *is* where other eltwise defects
hide. Here it is not: the two overloads carry the same `rounding_mode` branch and reach the same
kernel, so the scalar path is not a workaround.)

→ `invoke_binary_ng` → [`binary_ng_utils.cpp:433-434`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/binary_ng/device/binary_ng_utils.cpp#L433-L434)
`{"div_int32_floor_tile_init();", "div_int32_floor_tile"}` → `calculate_div_int32_floor` /
`calculate_div_int32_trunc` → `calculate_div_int32_body<floor>`.

### Root cause

Walking WH with `a = -2147483648`, `b = 2097153`, `trunc`:

1. **L80** `q_f = a_f * inv_b_f + vConstFloatPrgm0`, with `vConstFloatPrgm0 = 2**34` (L190). The ULP
   of `2**34` is `2**11 = 2048`, so `q_m = exman(q_f)` (L83) is the quotient divided by 2048 and
   rounded to nearest, and `q = q_m << 11` (L95) puts it back. With `|a| = 2**31` the true quotient
   is at most 1024 once `|b| >= 2**31 / 1024 = 2**21`, so `q_m` rounds to **0**, and `q = 0`,
   `qb = 0` (L102-103). The whole quotient is now owed to the residual-correction step — which is
   by design, and works fine for every other dividend.
   BH uses `2**33` (L146), ULP `2**10`, so its collapse threshold is `|b| >= 2**22`.
2. **L107-108** `a_s = sfpi::abs(a_s); r = a_s - qb;`. `sfpi::abs(INT32_MIN)` returns the unchanged
   bit pattern `0x80000000` (`sfpi_lib.h:284-287` in the SFPI toolchain headers — the comment there
   says so explicitly: *"Even though mostneg returns unchanged bit pattern, this returns vMag"*),
   so `r = 0x80000000 - 0 = 0x80000000`. As an unsigned magnitude this is *correct* — the true
   remainder is `2**31`. As a signed `vInt` it reads as `-2147483648`.
3. **L109** `r_f = convert<vFloat>(sfpi::abs(r))`. `convert<vFloat>(vMag)` goes through
   `convert<ToType>(as<vSMag>(val))` (`sfpi_lib.h:817-830`) — a sign-magnitude reinterpretation, so
   `0x80000000` becomes sign=1, magnitude=0, i.e. **`-0.0f`**, not `2**31`.
4. **L112-114** `correction_f = -0.0f * inv_b_f = -0.0f`; `convert<vUInt16>(-0.0f) = 0`. The
   correction is gone.
5. **L126** `v_if(r >= 0)` reads the same `0x80000000` as negative, so the lane also takes the
   "quotient was over-estimated" branch — the wrong direction. (This matters for the fix; see below.)
6. **L136** the one-ULP fallback `v_if(r < 0 && (r - 1) < 0)` cannot rescue it: `r - 1` overflows
   `INT_MIN` to `INT_MAX`, so the condition is false. The source comment on L135 says the `(r - 1)`
   test "is required to handle r=INT_MIN" — the one remaining path was deliberately closed against
   exactly this bit pattern.

Result: the computed quotient collapses to zero and only the recovery step's ±1 survives — `trunc`
returns `-1`, `floor` sees `r != 0` and returns `-2`.

**The control isolates it to exactly that one conversion.** Change the dividend by one, to
`a = -2147483647`, and keep `b = 2097153`. Everything before L109 is identical — `q_m = 0`,
`q = 0`, `qb = 0`, so the residual step still owes the entire quotient. But now `abs(a)` is
representable, `r = 2147483647` converts to a real float, `correction = 1024`, and the kernel
returns the correct `-1023`. The same lane with `a = -2147483648` gets `correction = 0` and
returns `-1`. One bit of input, same code path, same branch, and the quotient's entire magnitude.

### Measured

One step downstream of the defect depends on how the toolchain lowers `v_elseif(r >= b)` at WH L140 /
BH L99 (`sfpi.h:529`, `operator>=(vInt a, vMag b) { return a >= vInt(b); }`, lowered through
`__builtin_rvtt_sfpxicmpv`). That expansion is in the GCC backend, and
[`tenstorrent/sfpi-gcc`](https://github.com/tenstorrent/sfpi-gcc) settles it —
`gimple-rvtt-expand.cc`, `copy_and_replace_icmp()`:

> *"The sfpxicmpv insn is used to compare two vectors. The hardware doesn't have a compare insn,
> instead the compiler emits a subtract and tests the sign bit of the result."*

So every `sfpxicmpv` becomes `sfpxiadd_v | SFPXIADD_MOD1_IS_SUB`: a subtract whose sign bit is
tested, which wraps on overflow. The numbers below are computed under that lowering.

`a = -2147483648`, Wormhole:

| b | rounding | kernel | torch | abs err |
|---:|---|---:|---:|---:|
| 2097152 (= 2^21) | trunc | **-1** | -1024 | 1023 |
| 2097152 | floor | **-2** | -1024 | 1022 |
| 2097153 | trunc | **-1** | -1023 | 1022 |
| 4194305 | trunc | **-1** | -511 | 510 |
| 239823930 | trunc | **-1** | -8 | 7 |
| 1073741824 (= 2^30) | trunc | **-1** | -2 | 1 |
| -1073741824 | trunc | **1** | 2 | 1 |
| 2147483647 | trunc | -1 | -1 | 0 — above `2**30` the true quotient is ±1, so the broken path is accidentally right |
| 2097151 | trunc | -1024 | -1024 | 0 — control, just under the threshold |
| 2097151 | floor | -1025 | -1025 | 0 — control |

Blackhole is identical in shape with the threshold moved to `2**22`:

| b | rounding | kernel | torch | abs err |
|---:|---|---:|---:|---:|
| 4194303 | trunc | -512 | -512 | 0 — control |
| 4194304 (= 2^22) | trunc | **-1** | -512 | 511 |
| 4194305 | trunc | **-1** | -511 | 510 |
| 239823930 | trunc | **-1** | -8 | 7 |
| 1073741824 | trunc | **-1** | -2 | 1 |

**Threshold is exact.** Binary search over `b` for `a = INT32_MIN` puts the first failure at
`b = 2097152 = 2**21` on WH and `b = 4194304 = 2**22` on BH — matching the analytic prediction from
the `2**34` / `2**33` alignment constants.

**Size of the affected set.** With `a = INT32_MIN` on WH, every divisor in the band
`2**21 <= |b| <= 2**30` returns a wrong quotient — **49.90%** of the 4,294,967,295 non-zero int32
divisors, counted exactly rather than sampled. Above `2**30` the true quotient is `±1`, which is
what the broken path happens to return, so those divisors are accidentally correct. BH has the same
shape with the band starting at `2**22`.

**Model fidelity.** The instruction-level emulations of both kernels return the exact torch result on
260,840 inputs that contain no `INT32_MIN` — an exhaustive `|a|,|b| <= 120` grid, 40,000 random
full-range int32 pairs, and every `±2**e + d` cross product — on both platforms. They disagree with
torch only where `INT32_MIN` is involved.

**BH reciprocal.** `SFPARECIP`'s table is not public, so it is modelled as the correctly rounded
reciprocal. This does not matter: the kernel refines it with a Halley step (`e = -inv*b + 1;
e = e*e + e; inv = e*inv + inv`), and re-running with the seed perturbed by ±2^-4, ±2^-8 and 2^-12
relative error leaves every result in the tables above bit-identical, because the failure is triggered
by `q_m` collapsing to zero, not by the precision of the quotient estimate.

### Why the test suite cannot see this

This is the interesting part: `INT32_MIN` **is** in the test data, and the test still cannot fail.

[`tests/ttnn/unit_tests/operations/eltwise/test_binary_int32.py::test_div_int32_rounding_modes`](https://github.com/tenstorrent/tt-metal/blob/main/tests/ttnn/unit_tests/operations/eltwise/test_binary_int32.py)
has two parametrizations whose dividend range starts at `-2147483648`, and the `pytest.skip` for them
only applies to `rounding_mode=None`:

```python
(-2147483648, 2147483647, -2147483648, -2),
(-2147483648, 2147483647, 1, 2147483647),
...
torch_input_tensor_a = torch.linspace(high_a, low_a, num_elements, dtype=torch.int32)
torch_input_tensor_b = torch.linspace(high_b, low_b, num_elements, dtype=torch.int32)
```

Both operands are **monotone descending linspaces over the same index**, and `linspace` includes both
endpoints. `a == INT32_MIN` therefore occurs in exactly one lane — the last one — and that lane is
structurally locked to `b == low_b`, which is `-2147483648` in the first parametrization and `1` in
the second. Neither lane can fail, but for two different reasons:

- `b = 1` is far below the `2**21` threshold, so the quotient never collapses.
- `b = -2147483648` **is** above the threshold, but in that lane `a == b`, so the residue `r` and the
  divisor `b` are both `0x80000000`. `v_elseif(r >= b)` then holds by *equality* — no overflow, no
  reinterpretation, true under any lowering — and the `+1` restores the correct quotient of `1`.

Note that this second lane is safe by *equality*, not by any reinterpretation of `b`: `r` and `b`
are the same bit pattern, so the comparison holds under any lowering at all. Replaying both
parametrizations at all three tested shapes through the emulator:

| parametrization | lanes with `a == INT32_MIN` | divisor paired with it | failing lanes |
|---|---:|---:|---:|
| `a in [-2^31, 2^31-1]`, `b in [-2^31, -2]` | 1 | -2147483648 | 0 |
| `a in [-2^31, 2^31-1]`, `b in [1, 2^31-1]` | 1 | 1 | 0 |

`test_div_edge_cases` in the same file lists 18 explicit pairs and every extreme in it is
`±2147483647`; `-2147483648` does not appear at all.

At the LLK level the gap is structural and documented. `tt_metal/tt-llk/tests/python_tests/
test_sfpu_binary.py` sweeps `SfpuDivInt32` and `SfpuDivInt32Floor` over `(1.0, 8_000_000.0)`, under
the comment *"trunc/floor division < 2\*\*24: exact int->fp32 reciprocal, trunc == floor, and the
sign-magnitude pack path can't round-trip the negatives these kernels would emit."* So the LLK
harness cannot exercise a negative operand for this op at all — every input is positive and below
`2**24`, i.e. below the `2**21`/`2**22` thresholds *and* on the wrong side of zero. The ttnn-level
test is the only thing that could have caught it, and its linspace pairing prevents it.

Shuffling one of the two tensors, or adding explicit `(INT32_MIN, b)` pairs with `|b| >= 2**22`, is
enough to turn the existing `assert_equal` red.

### Proposed fix

**The obvious one-line patch does not work, and makes the error larger.** Adding only the missing
clamp —

```c
v_if(r_f < 0.0f) { r_f = 0x1.0p31f; }      // necessary, not sufficient
v_endif;
```

— restores the correction magnitude but leaves step 5 above intact: `v_if(r >= 0)` (WH L126) /
`v_if(r < 0)` (BH L83) still reads `0x80000000` as a negative remainder, so the now-correct
correction is applied in the wrong direction. Emulated, the worst absolute error goes from **1024 to
2048** on WH and 512 to 1024 on BH. Both the value and the sign of `r` need the guard.

The working patch keeps the existing idiom and adds one predicate:

```c
sfpi::vInt r = a_s - qb;
sfpi::vFloat r_f = sfpi::convert<sfpi::vFloat>(sfpi::abs(r), sfpi::RoundMode::Nearest);

// `r` holds the magnitude |a| - q*|b|, which ranges over [0, 2**31].  The single
// value 2**31 -- reached only when a == INT32_MIN and the aligned quotient
// underflowed to zero -- has the bit pattern 0x80000000, so it converts to -0.0f
// and reads as a negative remainder.  Same edge case as b_f and a_f above; the
// bit pattern of `r` itself is already the right two's complement operand.
sfpi::vInt r_sign = r;
v_if(r_f < 0.0f) {
    r_f = 0x1.0p31f;
    r_sign = 0;
}
v_endif;
...
v_if(r_sign >= 0) { tmp = -tmp; cor = -cor; }   // WH L126, was v_if(r >= 0)
v_endif;
```

and on Blackhole the same, with `v_if(r_sign < 0) { ... } v_else { ... }` at L83.

Note that `r`'s own bit pattern is left alone — as an unsigned value `0x80000000` *is* `2**31`, so
`r += tmp` with `tmp = -(correction * b)` lands on the correct small remainder in two's complement
without any further change. Only the two places that interpret `r` as *signed* need fixing.

Emulated results, `a = INT32_MIN` against 20492 divisors × {trunc, floor} = 40984 results per
platform/model combination:

| patch | WH wrong | BH wrong | worst abs err |
|---|---:|---:|---:|
| current | 12711 | 12367 | 1023 (WH), 511 (BH) |
| clamp only | 24608 | 24264 | **2048** (WH), 1024 (BH) |
| clamp + sign | **0** | **0** | **0** |

A wider sweep — 699,990 divisors (both signs; every `|b| <= 120000`, both threshold neighbourhoods
±5000, all `2**e + d`, plus 350k uniform random up to `2**31-1`), 1,399,980 results per combination
— gives the same answer on Wormhole: 480,151 wrong today, worst error 1023, smallest failing
positive divisor exactly `2097152`; **0 wrong** with the clamp + sign patch. (That divisor set deliberately over-samples small `|b|`, where
the kernel is correct, so those counts are not the population rate — the population rate is the
49.90% above.)

The patch is also *neutral*: on a regression corpus of 190,320 results (exhaustive `|a|,|b| <= 90`,
30,000 random full-range pairs, all `±2**e + d` cross products) it changes **no** result that does
not involve `INT32_MIN`, on either platform under either compare model.

Two properties worth checking in review, both of which hold by construction:

- `correction` is a `vUInt16`, so it must stay under 65536. When the guard fires,
  `correction ≈ 2**31 / |b|`, and the guard can only fire when `|b| >= 2**21` (WH) / `2**22` (BH),
  bounding it at 1024 / 512 — a factor of 64 of headroom.
- After the correction, `|r| <= |b| / 2` by the rounding of `correction`, so the existing ±1 fixup
  at L136/L140 covers the remainder exactly as it does for every other input.

Cost is roughly three extra SFPU ops per element (one compare, one predicated load, one move) on top
of ~60. That is an instruction count, not a measurement.

### Relationship to existing issues

- [**#33241**](https://github.com/tenstorrent/tt-metal/issues/33241) (closed) — the `$10000` bounty
  that specified this op: "precise results for the full 32 bit range". This is inside that scope.
- [**#33398**](https://github.com/tenstorrent/tt-metal/pull/33398) (merged) — the implementing PR.
  Its undefined-input list is `div(a, 0)` and `div(-2147483648, -1)` only.
- [**#33334**](https://github.com/tenstorrent/tt-metal/issues/33334) (closed) — "Typecasting two's
  complement int32 to float32 fails for -2**31", the same `-0.0f` conversion, filed as a bug by the
  same author. It is why `rounding_mode=None` is skipped for `a == INT32_MIN` in the test above; the
  `trunc`/`floor` paths are not skipped, because the kernel is supposed to have handled it.
- [**#20852**](https://github.com/tenstorrent/tt-metal/issues/20852) (open, `do-not-close`) — "SFPABS
  + integer overflow", which asks for the `0x80000000` case to be handled "for any kernel that uses
  SFPABS + integers". This kernel is one, at three sites.
- [**#44013**](https://github.com/tenstorrent/tt-metal/pull/44013) (merged) — fixed `calculate_abs_int32`
  for `INT32_MIN`, and its own description says the sfpi form it adopted is "matching the pattern
  already used in `ckernel_sfpu_div_int32_floor.h`". The pattern was taken from this file while this
  file was still applying it at two sites out of three.
- [**#33698**](https://github.com/tenstorrent/tt-metal/issues/33698) (closed, `bounty`) —
  "minimum/maximum(-2**31, ...) fails for int32", whose author added "I believe there are several
  areas that have incorrect handling of int32 with similar issues". This is one of them.
- [**#44750**](https://github.com/tenstorrent/tt-metal/issues/44750) /
  [**#49803**](https://github.com/tenstorrent/tt-metal/issues/49803) (both open) — int32 reduce
  min/max wrong on `INT32_MIN`. Different ops; same family.
- [**#48732**](https://github.com/tenstorrent/tt-metal/issues/48732) /
  [**#48736**](https://github.com/tenstorrent/tt-metal/pull/48736) — the most recent correctness work
  on int32 division, but on `ckernel_sfpu_div_int32.h` (`rounding_mode=None`, float output), a
  different kernel. It does not touch `ckernel_sfpu_div_int32_floor.h`.
- [**#48763**](https://github.com/tenstorrent/tt-metal/issues/48763) (closed) — "Wormhole vs
  Blackhole numerical mismatches: int32 div truncates low". The repo's existing WH/BH divergence
  report for int32 division, root-caused there to reciprocal truncation on *exact* quotients. That
  is a different mechanism from the one reported here — an off-by-one on exact divisions, versus the
  residual-correction term collapsing to zero for one specific dividend — and it does not cover
  `a == INT32_MIN`. Cited because the success criteria below ask for a WH/BH check, and this is the
  prior art for that question.

No open issue and no open bounty covers this. I searched `div_int32`, `INT32_MIN`, `2147483648`,
`rounding_mode`+`div`+`int32`, and the full `label:bounty state:open` list (14 issues, none on
integer division).

### Success criteria

- [ ] `ttnn.div(-2147483648, b, rounding_mode="trunc"/"floor")` is bit-exact against
      `torch.div(..., rounding_mode=...)` for every `b` except the two cases the op already declares
      undefined (`b = 0`, and the single pair `(-2147483648, -1)` whose result is not representable)
- [ ] the invariant `|q| >= 1` whenever `0 < |b| <= |a|` holds for all int32 inputs
- [ ] WH and BH agree on every input (they currently diverge for `2**21 <= |b| < 2**22` with
      `a = INT32_MIN`: WH is already wrong there, BH is not)
- [ ] `test_div_int32_rounding_modes` gains INT32_MIN pairings that are not index-locked to
      `low_b` — shuffling one operand tensor, or explicit pairs, is enough
- [ ] `test_div_edge_cases` gains `(-2147483648, b)` rows with `|b|` above both thresholds
      (e.g. 2097152, 4194304, 239823930, 2^30, -2^30, 2147483647)
- [ ] the LLK sweep for `SfpuDivInt32` / `SfpuDivInt32Floor` can express a negative operand — it
      currently runs `(1.0, 8_000_000.0)` because the sign-magnitude Dst pack path in that harness
      cannot round-trip negatives, which means this whole op family is untested below zero at LLK
      level (a separate, larger piece of work — flagging it, not scoping it in)
- [ ] no change to any result that does not involve `INT32_MIN` (the patch above is verified
      neutral on 190320 emulated results)
- [ ] the `b == INT32_MIN` half resolved either way, on device

### Why this matters

`-2**31` is not an exotic input. It is the identity element for an int32 max-reduction, the
conventional sentinel for "unset" in index and mask tensors, and where int32 arithmetic lands when it
wraps — which it does here, as `test_binary_int32.py` itself notes ("2147483647 * 2 outputs -2").
That it turns up in real int32 tensors is not hypothetical: two currently-open issues,
[#44750](https://github.com/tenstorrent/tt-metal/issues/44750) and
[#49803](https://github.com/tenstorrent/tt-metal/issues/49803), exist precisely because int32
reductions misbehave *when the input contains `INT32_MIN`*.

The failure mode is the bad kind: silent, finite, plausible-looking, confined to the one lane holding
the value, at up to 99.90% relative error. Nothing downstream can tell it apart from a real quotient, and
because both `ttnn.div` overloads land on the same kernel there is no scalar-vs-tensor fallback.

The op is also the deliverable of a `$10000` bounty whose acceptance criterion was the *full* 32-bit
range, and the guard that is missing is one the author already wrote twice, earlier in the same
function, for the same bit pattern, from the same `sfpi::abs`.

---

*Method: the numbers are instruction-level float32/bit-exact emulations of both kernels' exact
instruction sequences, written from the two sources; `sfpi::abs`, `convert<vFloat>(vMag)`, `exman`,
`SFPMUL24` and the `vUInt16` conversion semantics were read out of `sfpi_lib.h` and cross-checked
against `ckernel_sfpu_mul_int32.h`'s 23-bit split. **I do not have a device**, so no row here is a
silicon measurement. The emulations reproduce torch exactly on 260,840 non-`INT32_MIN` inputs, which
is the basis for trusting them. Two things are explicitly model-dependent and flagged as such: the
lowering of `vInt >= vMag` (both candidates reported; the main claim holds under both) and the
accuracy of BH's `SFPARECIP` (shown irrelevant by perturbation). The reference is
`torch.div(a, b, rounding_mode="trunc"/"floor")` on int32, which for integers is exactly C
truncation and Python floor division; results are wrapped to int32 the way the op does.*

I can prepare a PR with the kernel change and a pytest covering the special-value cases above, if this is accepted.

Gauravraj004:
### Problem

`ttnn.div` on int32 returns wrong quotients when the dividend is `INT32_MIN` (`-2147483648`). For example `ttnn.div(-2147483648, 2097152, rounding_mode="trunc")` returns `-1` where the true quotient is `-1024` — up to **99.90% relative error** for divisors in the normal range (e.g. `-2147483648 / 239823930 = -8`).

### Root cause

`calculate_div_int32_body` already knows that `sfpi::abs(-2**31)` returns `0x80000000` (sign-magnitude), which converts to `-0.0f`. It clamps the two conversions derived from that pattern (`b_f`, `a_f`), but a **third conversion of the same kind** — the residual `r_f = convert(abs(r))` at the remainder-correction step — is **unguarded**. When the dividend is `INT32_MIN`, `r_f` becomes `-0.0f`, the correction is computed as `-0.0f * inv_b_f`, and the whole quotient is lost, leaving only the ±1 the recovery step adds.

### Fix

Apply the same `v_if(r_f < 0.0f) { r_f = 0x1.0p31f; }` clamp used at the two existing sites to `r_f`, in both the Wormhole and Blackhole kernels. The code change is identical in pattern to the guards already present a few lines above.

### Tests

Extended `test_div_edge_cases` in `test_binary_int32.py` with INT32_MIN dividend cases across both `trunc` and `floor` rounding modes:

- `(-2147483648, 2097152)` → -1024
- `(-2147483648, 2097153)` → -1023
- `(-2147483648, 239823930)` → -8
- `(-2147483648, -2097152)` → 1024
- `(-2147483648, 1)` → -2147483648
- `(-2147483648, 2147483647)` → -1
- `(-2147483648, -2147483647)` → 1
- `(-2147483648, 1073741824)` → -2

Verified with an instruction-level emulation of the kernel sequence that mirrors the issue author's methodology.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_binary_int32.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_binary_int32.py
@@ -869,6 +869,16 @@ def test_div_edge_cases(rounding_mode, device):
         (2147483647, 1073741823),
         (1073741823, -2147483647),
         (1073741824, -2147483647),
+        # INT32_MIN dividend: abs(-2**31) is sign-magnitude INT32_MIN, which
+        # converts to -0.0f at the unguarded third conversion site (issue #51476).
+        (-2147483648, 2097152),
+        (-2147483648, 2097153),
+        (-2147483648, 239823930),
+        (-2147483648, -2097152),
+        (-2147483648, 1),
+        (-2147483648, 2147483647),
+        (-2147483648, -2147483647),
+        (-2147483648, 1073741824),
     ]
 
     numerators, denominators = zip(*pairs)
--- a/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
@@ -227,6 +227,44 @@ def test_binary_sfpu_accuracy(device, dtype):
         assert_allclose(torch_output_tensor, output, rtol=0.005, atol=1e-3)  # Ensures > 99.5% accuracy
 
 
+# Regression for issue #52675 (tensor-exponent path): the bf16 binary power ran
+# the same single-cubic log2 as the unary path, so a tensor exponent with large
+# magnitude produced silently wrong results (e.g. 0.99609375 ** 100 was 0.8008
+# instead of 0.6761) and overflow wrapped to NaN instead of +inf. bf16 now routes
+# through the accurate fp32 algorithm; verify large-magnitude tensor exponents
+# stay within a few ULP and overflow saturates to +inf.
+@pytest.mark.parametrize("exponent", [32.0, 100.0, 500.0])
+def test_binary_pow_bf16_large_exponent(device, exponent):
+    torch.manual_seed(1)
+    bases = torch.logspace(-3, 3, 4096, dtype=torch.float32).reshape(64, 64).to(torch.bfloat16)
+    exps = torch.full([64, 64], exponent, dtype=torch.bfloat16)
+    torch_output = torch.pow(bases, exps)
+
+    ttnn_base = ttnn.from_torch(bases, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_exp = ttnn.from_torch(exps, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    output = ttnn.to_torch(ttnn.pow(ttnn_base, ttnn_exp))
+
+    finite = torch_output.isfinite() & (torch_output.abs() >= torch.finfo(torch.bfloat16).tiny)
+    assert finite.any()
+    assert_with_ulp(torch_output[finite], output[finite], 3)
+
+
+@pytest.mark.parametrize("exponent", [128.0, 39.0, 15.0])
+def test_binary_pow_bf16_overflow_to_inf(device, exponent):
+    torch.manual_seed(0)
+    base = torch.full([1, 1, 32, 32], 10.0, dtype=torch.bfloat16)
+    exp = torch.full([1, 1, 32, 32], exponent, dtype=torch.bfloat16)
+    golden = torch.pow(base, exp)
+    assert torch.isinf(golden).all()  # sanity: this exponent really overflows
+
+    ttnn_base = ttnn.from_torch(base, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_exp = ttnn.from_torch(exp, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    result = ttnn.to_torch(ttnn.pow(ttnn_base, ttnn_exp))
+
+    assert torch.isinf(result).all(), "overflow wrapped to a finite value instead of +inf"
+    assert (result > 0).all(), "overflow produced -inf/NaN instead of +inf"
+
+
 def test_special_input_fp32(device):
     a = torch.tensor(
         [[1.0, 0.999, 0.999, 0.999, 0.999, 0.234, 0.985, 1.456, 0.0, -1.0, 1.2, -5.3, 6.7, 9.8, -10.9, 5.999]],
--- a/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
@@ -245,19 +245,28 @@ def test_binary_pow_bf16_large_exponent(device, exponent):
     output = ttnn.to_torch(ttnn.pow(ttnn_base, ttnn_exp))
 
     finite = torch_output.isfinite() & (torch_output.abs() >= torch.finfo(torch.bfloat16).tiny)
-    assert finite.any()
+    assert finite.any(), f"exponent {exponent}: no finite, non-denormal golden outputs to compare"
     assert_with_ulp(torch_output[finite], output[finite], 3)
 
 
-@pytest.mark.parametrize("exponent", [128.0, 39.0, 15.0])
-def test_binary_pow_bf16_overflow_to_inf(device, exponent):
+@pytest.mark.parametrize(
+    "base,exponent",
+    [
+        (2.0, 128.0),  # 2**128 overflows fp32 (issue repro)
+        (10.0, 39.0),  # 10**39 overflows fp32 (issue repro)
+        (1000.0, 15.0),  # 1000**15 overflows fp32; 10**15 would not
+    ],
+)
+def test_binary_pow_bf16_overflow_to_inf(device, base, exponent):
     torch.manual_seed(0)
-    base = torch.full([1, 1, 32, 32], 10.0, dtype=torch.bfloat16)
+    base_t = torch.full([1, 1, 32, 32], base, dtype=torch.bfloat16)
     exp = torch.full([1, 1, 32, 32], exponent, dtype=torch.bfloat16)
-    golden = torch.pow(base, exp)
-    assert torch.isinf(golden).all()  # sanity: this exponent really overflows
+    golden = torch.pow(base_t, exp)
+    assert torch.isinf(
+        golden
+    ).all(), f"base={base} exponent={exponent}: golden does not overflow (got {golden[0,0,0,0]})"
 
-    ttnn_base = ttnn.from_torch(base, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_base = ttnn.from_torch(base_t, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
     ttnn_exp = ttnn.from_torch(exp, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
     result = ttnn.to_torch(ttnn.pow(ttnn_base, ttnn_exp))
 
--- a/tests/ttnn/unit_tests/operations/eltwise/test_unary_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_unary_pow.py
@@ -94,3 +94,55 @@ def test_power_as_activation(device, op_type, exponent):
     tt_out = ttnn.to_torch(z_tt)
 
     assert_with_ulp(z_torch, tt_out, 1)
+
+
+# Regression for issue #52675: the bf16 power path ran a single cubic log2 whose
+# ~2.8e-3 absolute error was amplified by |exponent| (result rel err ~= 0.2%*|y|,
+# unbounded: 18% at y=100, 133% at y=500). bf16 now runs the accurate fp32
+# algorithm (atanh-series log2 + hi/lo argument split + explicit overflow guard),
+# so large exponents must stay within a few ULP of the bf16 golden and overflow
+# must saturate to +inf instead of wrapping to NaN/garbage.
+@pytest.mark.parametrize(
+    "base,exponent",
+    [
+        (0.99609375, 100.0),  # issue repro: 0.8008 (18% off) -> ~0.6761
+        (0.99609375, 500.0),  # issue repro: 0.3295 (133% off) -> ~0.1413
+        (0.99609375, -100.0),
+        (1.125, 64.0),
+        (0.875, -64.0),
+        (2.0, 128.0),  # overflow -> +inf
+        (10.0, 39.0),  # overflow -> +inf
+    ],
+)
+def test_pow_bf16_large_exponent(device, base, exponent):
+    torch.manual_seed(0)
+    torch_base = torch.full([1, 1, 32, 32], base, dtype=torch.bfloat16)
+    torch_output = torch.pow(torch_base, exponent)
+
+    ttnn_base = ttnn.from_torch(torch_base, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_output = ttnn.to_torch(ttnn.pow(ttnn_base, exponent))
+
+    if torch_output.isinf().all():
+        assert ttnn_output.isinf().all(), f"pow({base}, {exponent}) should overflow to +inf, got {ttnn_output[0,0,0,0]}"
+        assert (ttnn_output > 0).all(), "overflow produced -inf/NaN instead of +inf"
+    else:
+        assert_with_ulp(torch_output, ttnn_output, 3)
+
+
+# Sweep over a range of bases and large exponents to prove the error no longer
+# grows with |exponent|. Before the fix the median error was 5.8% at y=32 and
+# 29.3% at y=500; after the fix every point is within 3 ULP of the bf16 golden.
+@pytest.mark.parametrize("exponent", [32.0, 100.0, 500.0])
+def test_pow_bf16_large_exponent_sweep(device, exponent):
+    torch.manual_seed(1)
+    bases = torch.logspace(-3, 3, 4096, dtype=torch.float32).reshape(64, 64).to(torch.bfloat16)
+    torch_output = torch.pow(bases, exponent)
+
+    ttnn_base = ttnn.from_torch(bases, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_output = ttnn.to_torch(ttnn.pow(ttnn_base, exponent))
+
+    # Skip points where the golden overflows or denormal-flushes; those are
+    # correctly handled but outside the accuracy comparison.
+    finite = torch_output.isfinite() & (torch_output.abs() >= torch.finfo(torch.bfloat16).tiny)
+    assert finite.any()
+    assert_with_ulp(torch_output[finite], ttnn_output[finite], 3)
--- a/tests/ttnn/unit_tests/operations/eltwise/test_unary_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_unary_pow.py
@@ -144,5 +144,5 @@ def test_pow_bf16_large_exponent_sweep(device, exponent):
     # Skip points where the golden overflows or denormal-flushes; those are
     # correctly handled but outside the accuracy comparison.
     finite = torch_output.isfinite() & (torch_output.abs() >= torch.finfo(torch.bfloat16).tiny)
-    assert finite.any()
+    assert finite.any(), f"exponent {exponent}: no finite, non-denormal golden outputs to compare"
     assert_with_ulp(torch_output[finite], ttnn_output[finite], 3)
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
@@ -286,7 +286,15 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_(sfpi::vFloat base, sfpi::vFloat pow
 // is_fp32_dest_acc_en == false
 template <>
 sfpi_inline sfpi::vFloat _sfpu_binary_power_<false>(sfpi::vFloat base, sfpi::vFloat pow) {
-    return _sfpu_binary_power_21f_<false>(base, pow);
+    // Use the accurate fp32 algorithm for bf16 too. _sfpu_binary_power_21f_
+    // computed log2 with a single cubic on the mantissa, so pow*log2(base)
+    // amplified the ~2.8e-3 log2 error by |pow| and bf16 (the default dtype)
+    // returned silently wrong values for large |pow| plus NaN on overflow. The
+    // fp32 path (atanh-series log2 + hi/lo argument split + explicit overflow
+    // guard) is exact to ~1 ULP, and the final round-to-nearest bf16 conversion
+    // keeps the stored bf16 result correctly rounded.
+    sfpi::vFloat y = _sfpu_binary_power_f32_(base, pow);
+    return sfpi::convert<sfpi::vFloat16b>(y, sfpi::RoundMode::Nearest);
 }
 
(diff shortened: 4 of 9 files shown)


Gauravraj004:
Hi team — this is a small, low-risk fix for the `INT32_MIN` div edge case (#51476): one guarded conversion in the remainder-correction step, CI fully green, 4 commits. Could someone review it when you get a chance? Happy to address any feedback.

Issue #51216, opened 2026-07-27
[ttnn/data_movement] 4 host/kernel contract mismatches: 3 CB-credit deadlocks + 1 bad fast path


amahmudTT:
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

## Host and kernel disagree about the contract — 4 findings

The program factory and the kernel it launches hold different assumptions about how much work exists, or about what format flows between them.

**Manifests as:** Items 1-3 hang (deadlock). Item 4 hands the writer a data format it cannot consume.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS for the paths below: @tenstorrent/metalium-developers-ops-data-movement

**These are the cheapest of the 30 to prove.** A CB-credit imbalance deadlocks deterministically on the triggering shape, so each of items 1-3 gives a clean fail-without / pass-with regression test (the test hangs before the fix, passes after) rather than needing a numerical-accuracy comparison.

| # | Location | Defect |
|---|---|---|
| 1 | [`bcast/device/kernels/dataflow/reader_bcast_h_sharded.cpp:32`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/data_movement/bcast/device/kernels/dataflow/reader_bcast_h_sharded.cpp#L32) | Sharded bcast-H reader produces exactly Ht*Wt tiles per core, but the compute kernel it is paired with consumes NC*Ht*Wt; whenever NC (=N*C of input_a) > 1 the pr... |
| 2 | [`reshape_on_device/device/reshape_tile_program_factory.cpp:35`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/data_movement/reshape_on_device/device/reshape_tile_program_factory.cpp#L35) | The tile reshape writer's tile count is derived from the INPUT tensor's padded volume while the reader produces tiles counted from the OUTPUT tensor's padded shap... |
[...]

PR #52813, opened 2026-08-11
[Bug Fix] bcast sharded-H deadlock

abdulla-TT:
### PR Category
Bug fix

### Summary
Fixes item 1 of #51216 — a host/kernel contract mismatch that deadlocks `ttnn.bcast(..., BcastOpDim::H)` on width/block-sharded input whenever `input_a` has `N*C > 1`.

**Root cause.** `select_program_factory` (`bcast_device_operation.cpp`) only compares `padded_shape()[0]` of the two operands. When `input_a`'s batch lives in the C dim (e.g. `a = [1, 2, 1024, 1280]`, so `padded_shape()[0] == 1` but `N*C == 2`) and `input_b` has `padded_shape()[0] == 2`, it selects the non-optimised `BcastShardedHProgramFactory`. That factory handed the compute kernel `B = N*C` while the reader (`reader_bcast_h_sharded.cpp`) produces exactly `Ht*Wt` tiles per core. `Ht` is the shard height in tiles and already spans the full flattened `N*C*H`, so the extra `N*C` factor has no producer: `bcast_h.cpp` loops `b<B, h<Ht, w<Wt` and waits forever for `N*C*Ht*Wt` tiles (and reserves `N*C*Ht*Wt` pages in an output CB that only holds `Ht*Wt`) → device hang.

**Fix.** Pass `B = 1` to the compute kernel in `BcastShardedHProgramFactory`, since `Ht` already covers the full shard height. This makes the compute tile count match what the reader produces. The sibling optimised factory was already safe (it reads the `NC` arg but never uses it as a loop bound). Also removed the now-unused `N`/`C`/`NC` locals.

**Test.** Added `test_bcast_h_width_sharded_batched_channel` in `tests/tt_eager/python_api_testing/unit_testing/misc/test_bcast.py`. It puts the batch in the C dim (`a = [1, batch, H, W]`) with `b = [batch, 1, 1, W]`, which is the exact configuration that routes to the non-optimised factory. The test **hangs before this fix and passes after** (PCC vs. torch reference), for both ADD and MUL. Existing coverage only used `[N, 1, H, W]` shapes, so `N*C == 1` on this path and the bug was never exercised.

Relates to #51216 (the remaining items 2–4 are tracked separately in that issue).

### Notes for reviewers
- The behavioral change is a single runtime arg: compute `B` goes from `NC` → `1`. Please confirm the reasoning that `Ht` (= `shard_spec.shape[0] / TILE_HEIGHT`) already spans the full flattened `N*C*H` for both WIDTH_SHARDED and BLOCK_SHARDED layouts, so no outer batch multiplier is needed.
- The reader's separate runtime arg named `"NC"` is unrelated — it is fed `Ht_per_core` and only used for `input_b` wrap-around; it is untouched.
- The regression test is intentionally a hang-before/pass-after case (no numeric-only failure mode), consistent with the issue's description of these CB-credit imbalances.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/bcast_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/bcast_deadlock_fix)


tests/tt_eager/python_api_testing/unit_testing/misc/test_bcast.py
```diff
@@ -149,3 +149,70 @@ def test_bcast(
...
+        a_torch, device=device, layout=ttnn.TILE_LAYOUT, dtype=ttnn.bfloat16, memory_config=ttnn.L1_MEMORY_CONFIG
+    )
+
+    shard_height = batch * height_per_batch
+    shard_width = width // num_cores
+    core_grid = get_shard_grid_from_num_cores(num_cores, device)
+    sharded_mem_config = ttnn.create_sharded_memory_config(
+        shape=(shard_height, shard_width),
+        core_grid=core_grid,
+        strategy=ttnn.ShardStrategy.WIDTH,
```
Aswinmcw, line 195: Only WIDTH_SHARDED is covered, but pre-fix `B = N*C` was passed for BLOCK_SHARDED too, so block sharding with `N*C > 1` deadlocked identically — and the factory's block-sharded `offset` math already contemplates `bN > 1`, so it's a reachable path. A block-sharded parametrization would close the other half of the regression.

ttnn/cpp/ttnn/operations/data_movement/bcast/device/bcast_sharded_h_program_factory.cpp
```diff
@@ -210,7 +206,11 @@ ttnn::device_operation::ProgramArtifacts BcastShardedHProgramFactory::create_pro
             core,
             {{"Ht", Ht}, {"Wt", Wt}, {"offset", offset}, {"NC", Ht_per_core}, {"batch_offset", tile_offset}});
 
-        AddRuntimeArgsForNode(compute_args.runtime_arg_values, core, {{"B", NC}, {"Ht", Ht}, {"Wt", Wt}});
+        // Compute 'B' must be 1: the reader produces exactly Ht*Wt tiles per core and Ht is the shard
+        // height in tiles, which already spans the full flattened NC*H. Passing the input's N*C here
+        // would make the compute kernel (bcast_h.cpp loops b<B, h<Ht, w<Wt) wait for NC*Ht*Wt tiles
+        // that no producer ever pushes, deadlocking whenever N*C > 1.
+        AddRuntimeArgsForNode(compute_args.runtime_arg_values, core, {{"B", 1u}, {"Ht", Ht}, {"Wt", Wt}});
```
Aswinmcw, line 218: The issue's companion suggestion — a `TT_FATAL` asserting the compute tile count matches what the reader produces — isn't implemented, so this stays a comment-enforced invariant. `TT_FATAL(Ht * Wt == num_tile_per_core, ...)` here is a real cross-check (`num_tile_per_core` is computed independently from the shard shape with a `ceil`) and would surface a future shard-shape or arg-plumbing change as an error instead of a device hang.

Aswinmcw (approved):
On the reviewer question: `B = 1` is right, but the invariant that justifies it is layout-independent — the reader pushes exactly `Ht * Wt` tiles into both DFBs, so compute must consume exactly that. It does not depend on `Ht` spanning the full flattened `N*C*H` (for BLOCK_SHARDED it doesn't; `Ht` is the per-core shard height there).

VirdhatchaniKN (approved):

ttnn/cpp/ttnn/operations/data_movement/bcast/device/bcast_sharded_h_program_factory.cpp
```diff
@@ -92,6 +88,18 @@ ttnn::device_operation::ProgramArtifacts BcastShardedHProgramFactory::create_pro
         (shard_spec.shape[0] % TILE_HEIGHT == 0) && (shard_spec.shape[0] % TILE_WIDTH == 0),
         "Shard shapes must be multiple of TILE_HEIGHT ");
 
+    // Cross-check the host/kernel tile-count contract: the reader pushes Ht*Wt tiles per core and the
+    // compute kernel consumes (and the output CB holds) num_tile_per_core. These are derived
+    // independently (Ht/Wt from the tile-aligned shard dims, num_tile_per_core with a ceil), so a
+    // future shard-shape or arg-plumbing change that breaks the equality surfaces here as an error
+    // instead of a silent device hang.
```
amorrisonTT: Nit: unnecessary Claude-slop comment.

ttnn/cpp/ttnn/operations/data_movement/bcast/device/bcast_sharded_h_program_factory.cpp
```diff
@@ -210,7 +218,12 @@ ttnn::device_operation::ProgramArtifacts BcastShardedHProgramFactory::create_pro
             core,
             {{"Ht", Ht}, {"Wt", Wt}, {"offset", offset}, {"NC", Ht_per_core}, {"batch_offset", tile_offset}});
 
-        AddRuntimeArgsForNode(compute_args.runtime_arg_values, core, {{"B", NC}, {"Ht", Ht}, {"Wt", Wt}});
+        // Compute 'B' must be 1: the reader pushes exactly Ht*Wt tiles per core into both input DFBs,
+        // so the compute kernel (bcast_h.cpp loops b<B, h<Ht, w<Wt) must consume exactly Ht*Wt. This
+        // is layout-independent (Ht is the per-core shard height in tiles). Passing the input's N*C
+        // here would make compute wait for N*C*Ht*Wt tiles that no producer ever pushes, deadlocking
+        // whenever N*C > 1.
+        AddRuntimeArgsForNode(compute_args.runtime_arg_values, core, {{"B", 1u}, {"Ht", Ht}, {"Wt", Wt}});
```
amorrisonTT, line 218: This actually feels kind of backwards? Shouldn't fix the Op so that it just works for all upper dims, rather than forcing Shape[0]==1?

abdulla-TT: Not a restriction — `Shape[0]` can be anything. `B` is the compute kernel's per-core outer loop count, and batch is already folded into `Ht` (the per-core shard tile height). The reader pushes exactly `Ht*Wt` tiles per core, so passing `N*C` as `B` was counting batch twice and made compute wait on tiles nobody pushes. Multi-batch still works.

amorrisonTT: I see, thanks

amorrisonTT (approved):

PR #52816, opened 2026-08-11
[tt-train] Add bf16_ulp: bf16 ULP error metrics for tests

imichalakTT:
Part of #52951.

### PR Category

Feature

### Why

Kernel error scales with the magnitude of the output, so a fixed `atol` is either too
loose for small values or too tight for large ones. Measuring in bf16 ULP lets one
bound hold across the whole range.

Two numbers, because they catch different failures:

- **peak** — largest error, in ULP at `max |expected|`. Catches a single corrupt
  element that a percentile would discard.
- **p99** — 99th percentile of per-element error, each in ULP at its own element.
  Catches flat absolute noise: small next to the peak, huge next to small elements.
  `p99_floor_binades=k` measures elements smaller than `peak / 2**k` at the spacing of
  `peak / 2**k` instead, i.e. to an absolute tolerance; larger elements stay relative.

### Contents

`tt-train/tests/python/bf16_ulp.py`, a helper next to the tests rather than part of the `ttml`
package. Tests import it as `from bf16_ulp import ...`; `conftest.py` adds the directory to
`sys.path`, since the repo's pytest configuration does not put test directories there.

- `bf16_spacing(x)` — one bf16 ULP at `x`'s magnitude
- `bf16_ulp_error(got, expected, p99_floor_binades=None)` — `(peak, p99)`; rejects shape
  mismatches, empty and non-finite oracles
- `assert_within_bf16_ulp(...)` — asserts both, naming both numbers and both limits on failure

### Tests

`test_bf16_ulp.py` checks the spacing grid against `np.spacing` as an independent
oracle — subnormals, the binade a value rounds into, and the top of the range where
`nextafter` would return `inf`. Also pins what the bound accepts
and rejects: bf16 round-trip passes, while a uniform rescale, one corrupt element in
4096, flat noise at one peak ULP, and non-finite elements each fail on the expected one
of the two numbers. For the floor: what it measures, what it leaves alone, and what it hides.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:imichalak/llama-sp/1-testing-helpers)

### DIFF
--- a/tt-train/sources/ttml/ttml/testing.py
+++ b/tt-train/sources/ttml/ttml/testing.py
@@ -1,0 +1,44 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Shared helpers for TTML Python tests: bf16 error metrics."""
+
+from __future__ import annotations
+
+import numpy as np
+
+BF16_MANTISSA_BITS = 7
+BF16_MIN_NORMAL = 2.0**-126  # bf16 goes subnormal below this and the spacing stops shrinking
+
+
+def bf16_spacing(x):
+    """Gap between adjacent bf16 values where ``x`` lands, i.e. one bf16 ULP at that magnitude."""
+    x = np.maximum(np.abs(np.asarray(x, np.float64)), BF16_MIN_NORMAL)
+    mantissa, exponent = np.frexp(x)
+    binade = np.where(mantissa >= 1.0 - 2.0**-9, exponent, exponent - 1)
+    return 2.0 ** (binade - BF16_MANTISSA_BITS)
+
+
+def ulp_error(got, expected) -> tuple[float, float]:
+    """``|got - expected|`` in bf16 ULP.
+
+    Returns ``(peak_ulp, p99_ulp)``: the max error in ULP at ``max |expected|``, then the p99 of
+    the per-element errors, each in ULP at its own element.
+    """
+    got, expected = np.asarray(got, np.float64), np.asarray(expected, np.float64)
+    err = np.abs(got - expected)
+    scale = float(np.abs(expected).max())
+    return (
+        float(err.max() / bf16_spacing(scale)),
+        float(np.percentile(err / bf16_spacing(expected), 99)),
+    )
+
+
+def assert_within_ulp(got, expected, label: str, max_ulp: float, max_ulp_p99: float = np.inf) -> None:
+    got, expected = np.asarray(got, np.float64), np.asarray(expected, np.float64)
+    assert got.shape == expected.shape, f"{label}: shape {got.shape} != {expected.shape}"
+    ulp, ulp_p99 = ulp_error(got, expected)
+    detail = f"{label}: ulp={ulp:.2f} (limit {max_ulp}), ulp_p99={ulp_p99:.2f} (limit {max_ulp_p99})"
+    assert ulp <= max_ulp, detail
+    assert ulp_p99 <= max_ulp_p99, detail
--- a/tt-train/tests/python/test_ttml_testing.py
+++ b/tt-train/tests/python/test_ttml_testing.py
@@ -1,0 +1,195 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for ``ttml.testing``: the bf16 ULP metric."""
+
+from __future__ import annotations
+
+import ml_dtypes
+import numpy as np
+import pytest
+
+from ttml.testing import assert_within_ulp, bf16_spacing, ulp_error
+
+BF16_MAX = float((2.0 - 2.0**-7) * 2.0**127)
+BF16_MIN_SUBNORMAL = 2.0**-133
+
+
+def as_bf16(x):
+    return np.asarray(x).astype(ml_dtypes.bfloat16)
+
+
+class TestBf16Spacing:
+    @pytest.mark.parametrize(
+        "x,expected",
+        [(1.0, 2.0**-7), (1.5, 2.0**-7), (2.0, 2.0**-6), (4.0, 2.0**-5), (100.0, 0.5), (0.1, 2.0**-11)],
+    )
+    def test_known_magnitudes(self, x, expected):
+        assert float(bf16_spacing(x)) == expected
+
+    def test_matches_nextafter_across_the_whole_range(self):
+        """Independent oracle: the gap ml_dtypes reports above each value."""
+        mantissas = np.array([1.0, 1.0 + 2.0**-7, 1.5, 2.0 - 2.0**-7, 2.0 - 2.0**-8, 2.0 - 2.0**-9])
+        grid = np.outer(2.0 ** np.arange(-140.0, 127.0), mantissas).ravel()
+        rng = np.random.default_rng(0)
+        sample = rng.uniform(-1, 1, 20000) * 2.0 ** rng.integers(-140, 126, 20000)
+        x = np.abs(np.concatenate([grid, sample]).astype(np.float32)).astype(np.float64)
+        x = x[(x > 0) & (x < BF16_MAX)]
+
+        oracle = np.spacing(as_bf16(x)).astype(np.float64)
+        mismatch = np.flatnonzero(bf16_spacing(x) != oracle)
+        assert mismatch.size == 0, f"{mismatch.size}/{x.size} differ, first at x={x[mismatch[:1]]}"
+
+    @pytest.mark.parametrize("x", [0.0, BF16_MIN_SUBNORMAL, 2.0**-130, 1e-40, 2.0**-126, 2.0**-125 - 2.0**-133])
+    def test_floors_at_the_subnormal_spacing(self, x):
+        """Below ``2**-125`` the grid stops halving, so every magnitude there shares one gap."""
+        assert float(bf16_spacing(x)) == BF16_MIN_SUBNORMAL
+
+    def test_resumes_halving_above_the_subnormal_range(self):
+        assert float(bf16_spacing(2.0**-125)) == 2.0**-132
+
+    @pytest.mark.parametrize(
+        "x,expected",
+        [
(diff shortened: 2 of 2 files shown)


tt-train/sources/ttml/ttml/testing.py
```diff
@@ -0,0 +1,43 @@
...
+    return 2.0 ** (binade - BF16_MANTISSA_BITS)
+
+
+def ulp_error(got, expected) -> tuple[float, float]:
+    """``|got - expected|`` in bf16 ULP.
+
+    Returns ``(peak_ulp, p99_ulp)``: the max error in ULP at ``max |expected|``, then the p99 of
+    the per-element errors, each in ULP at its own element.
+    """
+    err = np.abs(got - expected)
```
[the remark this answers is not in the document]
imichalakTT: fixed

abustamanteTT (approved):

philei-tt (approved):
Approving. The math checks out (I verified `bf16_spacing` against `ml_dtypes` as an oracle and the binade / ties-to-even rule by hand), the 36 tests pass locally with `RuntimeWarning` as error, and `train_unit_tests` collects `tt-train/tests/python` wholesale so the new file runs in CI. A few comments, none blocking.

**1. p99 has no floor at small/zero oracle values (design, suggest a follow-up)**

`ulp_error` divides each element's error by `bf16_spacing(expected)` at that element, and nothing stops that denominator from bottoming out at the subnormal spacing (2**-133). Wherever the oracle is exactly 0 or tiny, any residual becomes an astronomical per-element ULP, and if more than 1% of elements are like that the p99 itself blows up:

- oracle with 2% exact zeros + noise at 1e-6 of the peak → p99 ≈ 2e34
- same noise, no zeros → p99 ≈ 0.011

#52817 already hits this: where `silu'(gate) → 0`, a correct kernel's dgate reaches ~164 ULP per element and needs its own `MAX_ULP_P99_DGATE`. Every future caller with padding, masks, ReLU/SiLU tails or cancelling reductions will rediscover it the same way. Suggest either flooring the per-element scale at a fraction of the peak, e.g. `bf16_spacing(np.maximum(np.abs(expected), scale * 2.0**-k))`, or the near-zero split that `tests/ttnn/utils_for_testing.py::measure_ulp_with_near_zero_atol` uses (scaled atol below 1% of max, ULP over the rest). At minimum, please document the limitation on `assert_within_ulp`, which has no docstring and is the entry point #52817 / #52820 use.

**2. Placement (question)**

`tt-train/sources/ttml/CMakeLists.txt:504` installs every `ttml/*.py`, and `ttml/__init__.py` imports `_ttml` unconditionally, so `ttml.testing` ships as runtime API that can't be imported without the compiled extension, even though it is pure numpy. Nothing else under `ttml/` is test-only. `torch.testing` is a fine precedent if the intent is that model authors call this; if it is only for our tests, `tt-train/tests/python/` (two files there already do the `sys.path` insert) or next to `assert_with_ulp` / `ulp_distance` / `measure_ulp_with_near_zero_atol` in `tests/ttnn/utils_for_testing.py` would avoid a fourth, differently-defined ULP metric in the repo (ttnn's p99 is integer bit-distance, this one is fractional at the oracle's magnitude; on identical data they report 0.0 vs 0.50).

**3. Small simplification**

`scale` equals `bf16_spacing(expected).max()` by monotonicity (fuzz-checked incl. zeros, subnormals, `BF16_MAX`, nan/inf), so `ulp_error` can be:

```python
spacing = bf16_spacing(expected)
return float(err.max() / spacing.max()), float(np.percentile(err / spacing, 99))
```

which drops one full-size `np.abs` pass and the `scale` local; all 36 tests still pass.

**4. Nits**

- `bf16_spacing` docstring: worth saying why frexp instead of `np.spacing(as_bf16(x))`: (a) ml_dtypes returns inf/nan for anything that rounds to `BF16_MAX` or above, (b) it double-rounds f64 via f32, so values within half an f32 ulp below a binade-top midpoint land one binade higher. That second point is what the `np.nextafter(2.0 - 2.0**-8, 0.0)` test row pins, but nothing says so. Also `1.0 - 2.0**-(BF16_MANTISSA_BITS + 2)` instead of the literal `1.0 - 2.0**-9`, so the threshold can't drift from the constant.
- `test_matches_nextafter_across_the_whole_range`: the `x < BF16_MAX` filter does not exclude `[BF16_MAX - 2**119, BF16_MAX)`, where ml_dtypes says `inf` and `bf16_spacing` says `2**120`. It passes today only because the random exponents stop at 2**125 and the grid's top-binade points are all ≥ `BF16_MAX`. Tighten to `x < BF16_MAX - 2.0**119` or comment why the range stops there.
- Non-finite values in `expected` (e.g. a matched `-inf` in masked logits) give `(nan, nan)` and always fail. Fails closed, so harmless for current callers, but undocumented; an `assert np.all(np.isfinite(expected))` with a message would make it explicit.
- `assert_within_ulp(np.zeros(4), np.zeros(5), "lbl", max_ulp=0.0)  # never reached` reads as if the call is dead code; it means `max_ulp` is never evaluated.

tt-train/sources/ttml/ttml/testing.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def ulp_error(got, expected) -> tuple[float, float]:
+    """``|got - expected|`` in bf16 ULP.
+
+    Returns ``(peak_ulp, p99_ulp)``: the max error in ULP at ``max |expected|``, then the p99 of
+    the per-element errors, each in ULP at its own element.
+    """
+    got, expected = np.asarray(got, np.float64), np.asarray(expected, np.float64)
+    err = np.abs(got - expected)
```
mdragulaTT, line 30: Shape check belongs here, not only in `assert_within_ulp`: `ulp_error(np.ones(4), np.ones((4, 1)))` broadcasts to (4, 4) and returns `(0.0, 0.0)`. This is public and called directly in tests. Empty arrays also die in `.max()` with a numpy error rather than a clear one.

tt-train/sources/ttml/ttml/testing.py
```diff
@@ -0,0 +1,44 @@
...
+    scale = float(np.abs(expected).max())
+    return (
+        float(err.max() / bf16_spacing(scale)),
+        float(np.percentile(err / bf16_spacing(expected), 99)),
+    )
+
+
+def assert_within_ulp(got, expected, label: str, max_ulp: float, max_ulp_p99: float = np.inf) -> None:
```
mdragulaTT, line 38: Whatever the placement decision (Pavlo's 2), the name should carry bf16: ttnn's `assert_with_ulp` already means something different, and a 0.06 error on a ~9.4 fp32 loss passes here as 0.96 ULP. `assert_within_bf16_ulp` / `bf16_ulp_error`, or a dtype argument.

tt-train/sources/ttml/ttml/testing.py
```diff
@@ -0,0 +1,44 @@
...
+    got, expected = np.asarray(got, np.float64), np.asarray(expected, np.float64)
+    assert got.shape == expected.shape, f"{label}: shape {got.shape} != {expected.shape}"
+    ulp, ulp_p99 = ulp_error(got, expected)
+    detail = f"{label}: ulp={ulp:.2f} (limit {max_ulp}), ulp_p99={ulp_p99:.2f} (limit {max_ulp_p99})"
+    assert ulp <= max_ulp, detail
```
mdragulaTT, line 43: Use `raise AssertionError(detail)`: bare `assert` is stripped under `python -O`, and pytest's assert rewriting does not reach library modules, so this becomes a no-op check. Same for the shape assert on L40.

mdragulaTT (commented):
Nice, tight helper; the sensitivity tests pin exactly what the metric catches. +1 to Pavlo's 1 (p99 floor) and 2 (placement). On 2, my preference is a tests-side module under `tt-train/tests/python/`: nothing else under `ttml/` is test-only, and PR4 adds mesh helpers to this file that need `ttml` anyway. Two small ones inline: `ulp_error` broadcasts silently on mismatched shapes, and bare `assert` is stripped under `-O`. Not blocking, but merge only once placement/name is settled and those two are in. PR2–5 import these names.

tt-train/sources/ttml/ttml/testing.py
imichalakTT: addressed in https://github.com/tenstorrent/tt-metal/pull/52816/changes/bd61cd083dad8f193df7e5e46cc131dc9910fe8e

added a shape check and check against empty arrays

tt-train/sources/ttml/ttml/testing.py
imichalakTT: addressed in https://github.com/tenstorrent/tt-metal/pull/52816/changes/bd61cd083dad8f193df7e5e46cc131dc9910fe8e

assert_within_bf16_ulp / bf16_ulp_error

tt-train/sources/ttml/ttml/testing.py
imichalakTT: addressed in https://github.com/tenstorrent/tt-metal/pull/52816/changes/bd61cd083dad8f193df7e5e46cc131dc9910fe8e

switched to an AssertionError

mdragulaTT (approved):
Thanks, LGTM ✅

Issue #52395, opened 2026-08-07
Replace mid-kernel compute_kernel_hw_startup re-inits with a targeted DST re-arm


ncvetkovicTT:
`compute_kernel_hw_startup` is a documented **call-once** API (idle-unit assumptions; see `tt_metal/hw/inc/api/compute/compute_kernel_hw_startup.h`), but a number of compute kernels call it **mid-kernel / in a loop** as a deliberate re-init/re-arm of the MATH↔PACK DST semaphore + PACK after a disrupting op (tilize_uninit, reduce, SFPU, matmul, etc.).

This pattern is **pre-existing** — those sites called the old full inits (`binary_op_init_common` / `init_bcast` / 2-arg `unary_bcast_init`) at the same spots; the eltwise-binary/bcast init cleanup (#50745) renamed those to `compute_kernel_hw_startup`, which made the call-once contract violation *visible* (it did not introduce new mid-kernel startups — verified 0 net-new).

**Ask:** replace these mid-kernel `compute_kernel_hw_startup` re-inits with a lighter, mid-kernel-safe **targeted re-arm** (just `llk_math_pack_sync_init` + `llk_pack_dest_init`, exposed as a small compute-API helper) instead of the full hardware startup. Until then, #50745 marks each such site with a `TODO` referencing this issue.

Flagged by @bbradelTT on #50745.

ncvetkovicTT:
Scope note: 'mid-kernel' here = `compute_kernel_hw_startup` is **not the first Tensix-engine consumer** in the kernel (per @bbradelTT). This covers both (a) in-loop / after-`tile_regs_acquire` re-arms after a disrupting op, and (b) startups placed after `init_sfpu` or a prior startup. #50745 tags 63 such sites with `TODO(#52395)`; branch-exclusive `#if/#else` firsts are excluded.

iwroszTT:
@ncvetkovicTT @bbradelTT 
I've created a draft PR with the helper API, so that we don't need to add more TODOs in the ports that we are working on.

It's not yet ready for a formal review, but please take a look when you have time if it's something we want to have.

Issue #32533, opened 2025-11-17
[tt-train] Refactor `assign_per_core_runtime_args` to common utils module


mdragulaTT:
## Description

Refactor the `assign_per_core_runtime_args` function to a common utils module with parameterized kernel handles and buffer configurations. Different operations have varying numbers of input/output buffers and kernel configurations.

## Impact

Duplicated pattern across tt-train operations:
- **SwiGLU**: 4 input + 1 output buffers
- **SiLU**: 2 input + 1 output buffers  
- **RMSNorm Forward**: 2 input + 2 output buffers
- **RMSNorm Backward**: 4 input + 2 output buffers

## TODO Checklist

- [ ] **Analyze common patterns** in existing `assign_per_core_runtime_args` functions
- [ ] **Create generic interface** supporting variable buffer counts and core distribution
- [ ] **Implement common utils** function in `metal/ops/common/runtime_args_utils.hpp`
- [ ] **Migrate operations** to use shared implementation

## Benefits

- **Reduce code duplication** across tt-train operations
- **Standardized patterns** for runtime argument assignment  
- **Easier implementation** of new operations

## Context

**Location**: `swiglu_fw_program_factory.cpp` (lines 63-109)  
**Affected**: SwiGLU, SiLU, RMSNorm, LayerNorm, and other tt-train kernels

mdragulaTT:
Picking this up, in batches rather than one sweep.

- Helper: `for_each_core_with_work` and `for_each_core` in `metal/common/program_utils.hpp`. Same core order and group lookup as today; the op keeps only its argument vectors, in a lambda at the call site.
- Batch 1, #56524: rmsnorm fw/bw, layernorm fw/bw, silu_bw, swiglu_elemwise_bw, profiler_no_op. The ops whose runtime args are buffers plus count plus offset.
- Batch 2: cross_entropy fw/bw, polynorm fw/bw, softmax_backward.
- Batch 3: sdpa fw/bw, mla_q_rope, mla_kv_assemble fw/bw, adamw, sgd. Their per-core arguments are computed rather than plain offsets, which fits the same lambda.
- New ops adopt from day one. #52817 adds two program factories with the hand-written walk and is asked to switch.

The location in the description is stale: `swiglu_fw_program_factory.cpp` no longer exists. The pattern lives in every `*_program_factory.cpp` under `metal/ops`, 20 at last count.

Sibling cleanup with the same shape, the per-op tensor validation lambda: #56523.

[...]

PR #52817, opened 2026-08-11
[tt-train] Add packed-SwiGLU forward and backward metal ops

imichalakTT:
Part of #52951.

### Summary

`swiglu_packed_fw` / `swiglu_packed_bw` compute the SwiGLU gating over one packed
`[.., R, 2I]` tensor laid out `[gate | up]`, instead of two separate `[.., R, I]` tensors.

#52818 fuses the gate and up projections (w1 and w3) into one matmul whose output is exactly that
layout. Feeding it to the existing elementwise path would mean slicing the halves apart first — the
`slice+comp` arm below, +123% to +257% on the forward. These kernels read both halves in place and
write `[dgate | dup]` into the two halves of one output, so there is no slice and no concat.

- `swiglu_packed_fw(packed, preallocated_output = nullopt) -> h [.., R, I]`
- `swiglu_packed_bw(packed, dL_dh, preallocated_dL_dpacked = nullopt) -> dL/dpacked [.., R, 2I]`
- `ops::swiglu_packed` wires them into autograd; both are exposed through nanobind.

The gating backward math moved to `metal/common/swiglu_gate_bw_compute.hpp`, now shared with
`swiglu_elemwise_bw` — which loses 143 lines rather than gaining a second copy. Work is
distributed by block rather than by tile-row, so shapes with few tile-rows still fill the grid.

Against the separate-tensor composite the packed forward is 2–9% faster on five of the mid and
large shapes, 2% slower on `i5632_4k_rows`, and 49% faster on the smallest, where the composite is
dispatch-bound and its forward varies between runs; the backward is 18–51% faster on five shapes and
within 3% on `i5632_4k_rows` and `i3584_4k_rows`. Its backward output is 2I wide, so DRAM peak doubles unless
the caller hands over a destination — only a fused-MLP caller has a dead buffer to give, which is
what the `packed+pre` arm prices.

`tests/python/test_swiglu_packed.py` checks both directions against a float64 oracle within ULP
bounds, using `bf16_ulp` from #52816.

### Benchmark results

warmup=3 measure=20, median of 3 runs; µs are host time per launch; Fwd %/Bwd % are each direction against the composite arm.
DRAM columns are the peak of one captured invocation of that direction.

#### `i1024_1k_rows` — 1024 rows (32 tile-rows), I = 1024, packed = 2048

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |               56.8 |   +0.00 |               77.6 |   +0.00 |      2048.0 |      2048.0 |
| packed     |               28.9 |  -49.12 |               43.8 |  -43.56 |      2048.0 |      4096.0 |
| packed+pre |               29.0 |  -48.94 |               43.0 |  -44.59 |      2048.0 |         0.0 |
| slice+comp |              202.6 | +256.69 |              166.2 | +114.18 |      6144.0 |      8192.0 |

#### `i1024_4k_rows` — 4096 rows (128 tile-rows), I = 1024, packed = 2048

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |               73.7 |   +0.00 |              176.6 |   +0.00 |      8192.0 |      8192.0 |
| packed     |               71.2 |   -3.39 |              137.9 |  -21.91 |      8192.0 |     16384.0 |
| packed+pre |               71.2 |   -3.39 |              137.1 |  -22.37 |      8192.0 |         0.0 |
| slice+comp |              164.7 | +123.47 |              268.3 |  +51.93 |     24576.0 |     32768.0 |

#### `i5632_1k_rows` — 1024 rows (32 tile-rows), I = 5632, packed = 11264

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |               98.2 |   +0.00 |              392.0 |   +0.00 |     11264.0 |     11264.0 |
| packed     |               96.5 |   -1.73 |              192.0 |  -51.02 |     11264.0 |     22528.0 |
| packed+pre |               96.1 |   -2.14 |              192.6 |  -50.87 |     11264.0 |         0.0 |
| slice+comp |              221.3 | +125.36 |              517.3 |  +31.96 |     33792.0 |     45056.0 |

#### `i5632_4k_rows` — 4096 rows (128 tile-rows), I = 5632, packed = 11264

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |              365.4 |   +0.00 |              846.7 |   +0.00 |     45056.0 |     45056.0 |
| packed     |              373.0 |   +2.08 |              824.8 |   -2.59 |     45056.0 |     90112.0 |
| packed+pre |              373.9 |   +2.33 |              844.2 |   -0.30 |     45056.0 |         0.0 |
| slice+comp |              856.3 | +134.35 |             1317.1 |  +55.56 |    135168.0 |    180224.0 |

#### `i14336_1k_rows` — 1024 rows (32 tile-rows), I = 14336, packed = 28672

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |              236.4 |   +0.00 |              986.9 |   +0.00 |     28672.0 |     28672.0 |
| packed     |              220.6 |   -6.68 |              528.0 |  -46.50 |     28672.0 |     57344.0 |
| packed+pre |              220.6 |   -6.68 |              524.6 |  -46.84 |     28672.0 |         0.0 |
| slice+comp |              551.2 | +133.16 |             1303.4 |  +32.07 |     86016.0 |    114688.0 |

#### `i14336_4k_rows` — 4096 rows (128 tile-rows), I = 14336, packed = 28672

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |              917.4 |   +0.00 |             2241.7 |   +0.00 |    114688.0 |    114688.0 |
| packed     |              838.5 |   -8.60 |             1834.5 |  -18.16 |    114688.0 |    229376.0 |
| packed+pre |              841.7 |   -8.25 |             1769.3 |  -21.07 |    114688.0 |         0.0 |
| slice+comp |             2161.8 | +135.64 |             3512.1 |  +56.67 |    344064.0 |    458752.0 |

#### `i3584_4k_rows` — 4096 rows (128 tile-rows), I = 3584, packed = 7168

max abs diff vs reference: fw 0.03125, bw 0

| Arm        | Fwd host µs/launch | Fwd %   | Bwd host µs/launch | Bwd %   | Fwd DRAM KB | Bwd DRAM KB |
|:-----------|-------------------:|--------:|-------------------:|--------:|------------:|------------:|
| composite  |              238.2 |   +0.00 |              537.8 |   +0.00 |     28672.0 |     28672.0 |
| packed     |              218.5 |   -8.27 |              527.8 |   -1.86 |     28672.0 |     57344.0 |
| packed+pre |              218.7 |   -8.19 |              530.4 |   -1.38 |     28672.0 |         0.0 |
| slice+comp |              551.1 | +131.36 |              859.9 |  +59.89 |     86016.0 |    114688.0 |

<details>
<summary>Backward before/after issuing each init once per block (review thread on the shared header)</summary>

Host µs per launch. Before: one run of the committed header with per-tile inits; after: median of the 3 runs above. The forward does not use this header and was unchanged within noise.

| Case | composite | packed | packed+pre |
|:--|--:|--:|--:|
| `i1024_1k_rows` | 78.3 → 77.6 | 44.0 → 43.8 | 43.1 → 43.0 |
| `i1024_4k_rows` | 176.0 → 176.6 | 138.8 → 137.9 | 137.9 → 137.1 |
| `i5632_1k_rows` | 394.9 → 392.0 | 193.2 → 192.0 | 193.8 → 192.6 |
| `i5632_4k_rows` | 845.4 → 846.7 | 852.3 → 824.8 | 852.0 → 844.2 |
| `i14336_1k_rows` | 994.0 → 986.9 | 534.0 → 528.0 | 534.6 → 524.6 |
| `i14336_4k_rows` | 2229.7 → 2241.7 | 1909.6 → 1834.5 | 1913.5 → 1769.3 |
| `i3584_4k_rows` | 542.9 → 537.8 | 532.0 → 527.8 | 532.9 → 530.4 |

</details>

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:imichalak/llama-sp/2-swiglu-packed)


tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
```diff
@@ -0,0 +1,338 @@
...
+        c.tile_rows(),
+        c.inner,
+        2U * c.inner,
+        result.max_abs_diff.forward,
+        result.max_abs_diff.backward);
+    fmt::print(
+        "+------------+-----------+-----------+-----------+-----------+-------------+-------------+\n"
+        "| Arm        | Fwd µs    | Fwd %     | Bwd µs    | Bwd %     | Fwd DRAM KB | Bwd DRAM KB |\n"
+        "+------------+-----------+-----------+-----------+-----------+-------------+-------------+\n");
+    for (const auto& r : result.arms) {
```
vtsilytskyiTT, line 300: Nit: This is almost a markdown table. MD tables are convenient if you want to copy-paste results somewhere, like GitHub. Did you consider markdown formatting?

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
```diff
@@ -0,0 +1,338 @@
...
+    std::string name;
+    uint32_t rows;
+    uint32_t inner;
+
+    uint32_t tile_rows() const {
+        return rows / tt::constants::TILE_HEIGHT;
+    }
+};
+
+const std::vector<Case>& all_cases() {
```
vtsilytskyiTT, line 59:
```suggestion
constexpr std::array<Case, 7> all_cases {
```
Nit: there is no need in function and in memory allocation.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
```diff
@@ -0,0 +1,156 @@
...
+    const auto& out_memcfg = tensor_args.preallocated_dL_dpacked.has_value()
+                                 ? tensor_args.preallocated_dL_dpacked->memory_config()
+                                 : packed.memory_config();
+    return tt::tt_metal::operation::hash_operation<SwigluPackedBwDeviceOperation>(
+        args,
+        packed.dtype(),
+        packed.logical_shape(),
+        packed.padded_shape(),
+        packed.memory_config(),
+        tensor_args.dL_dh.logical_shape(),
```
philei-tt, line 134: The hash includes packed's dtype and both shapes but only `dL_dh`'s logical shape and memory config, not its dtype or padded shape. `swiglu_elemwise_bw` hashes dtype + both shapes for every input. Since validation is skipped on a cache hit, a float32 upstream grad with a matching shape would reuse the bf16 program. Low practical risk in an all-bf16 stack, but worth matching the sibling op.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
```diff
@@ -0,0 +1,156 @@
...
+        packed_padded[-1],
+        two_tiles_w);
+    TT_FATAL(
+        packed.logical_shape()[-1] % two_tiles_w == 0U,
+        "SwigluPackedBw: packed last logical dim {} must be a multiple of {} so the gate|up split "
+        "lands on the tile boundary where the kernel splits the padded row",
+        packed.logical_shape()[-1],
+        two_tiles_w);
+    TT_FATAL(
+        dh_padded[-1] * 2U == packed_padded[-1] && dh_padded[-2] == packed_padded[-2] &&
```
philei-tt, line 73: Rank of `dL_dh` isn't checked before indexing dims 0, 1, -1, -2 of its padded shape, so a 3D grad gives a confusing dimension-mismatch message rather than a rank error. Also only the padded row count is compared to packed's, not the logical one. Both are one-liners.

tt-train/tests/python/test_swiglu_packed.py
```diff
@@ -0,0 +1,180 @@
...
+MAX_ULP = 2.5
+# Per-element ULP is unbounded where the oracle vanishes: silu'(gate) -> 0 for negative gate
+# drives dgate's per-element error to ~164 on a correct kernel, while its p99 stays near 6.
+MAX_ULP_P99 = 2.5
+MAX_ULP_P99_DGATE = 10.0
+
+# (batch, S, I). I/32 of 1..4 covers every block size get_block_size(Wt, 4) can pick; I=512
+# adds a row spanning many blocks; the last is the only case with batch > 1 and with
+# total_rows (128) above the core count, which is what splits work into two compute groups.
+SHAPES = [(1, 32, 32), (1, 64, 64), (1, 32, 96), (1, 128, 128), (1, 64, 512), (2, 2048, 128)]
```
philei-tt, line 34: The comment says `(2, 2048, 128)` splits work into two compute groups, but I don't think it does. With I=128, block_size=4 and one block per tile-row, total work is 128 blocks. On an 8x8 grid that's exactly 2 per core; on 13x10 it's 128 cores with 1 block each. Neither produces a second group, so the second compute kernel and the runtime-arg walk across the group boundary are untested. A block count not divisible by the core count (e.g. 65 or 130 tile-rows of I=32) would cover it.

tt-train/sources/ttml/metal/common/swiglu_gate_bw_compute.hpp
```diff
@@ -0,0 +1,129 @@
...
+    uint32_t cb_scratch,
+    uint32_t cb_silu_grad,
+    uint32_t block_size>
+inline void swiglu_gate_bw_block() {
+    constexpr uint32_t one = 0x3F800000;  // 1.0f bits
+
+    // sigmoid(gate) -> cb_sigmoid, stored for reuse in both gradients.
+    tile_regs_acquire();
+    for (uint32_t i = 0; i < block_size; ++i) {
+        copy_tile_init(cb_gate);
```
philei-tt, line 46: Perf nit, not blocking: `copy_tile_init` and `sigmoid_tile_init` are called inside the per-tile loop here (carried over from the old elemwise kernel). The new forward kernel hoists them out of the loop; doing the same here would be a small win for both consumers.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/compute/swiglu_packed_fw_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+    copy_tile_init(cb_gate);
+    for (uint32_t i = 0; i < block_size; ++i) {
+        copy_tile(cb_gate, i, i);
+    }
+    silu_tile_init();
+    for (uint32_t i = 0; i < block_size; ++i) {
+        silu_tile(i);
+    }
+    tile_regs_commit();
+    pack_and_push_block(cb_silu, block_size);
```
philei-tt, line 37: Perf nit, not blocking: `silu(gate)` round-trips through an L1 CB to be multiplied with `up`. With fp32 dest acc on, DST holds 4 tiles, so a DST-resident SFPU multiply would need block_size 2. Probably not worth it given the forward is already at parity with the fused ttnn multiply on large shapes, just noting the option.

philei-tt (approved):
Kernels and math look correct: tile indexing in reader/writer, the backward derivative, CB pop-before-pack ordering, and the shared-header extraction all check out. The logical-vs-padded width validation (100 -> 128 case) is a nice catch. A few small things inline, none blocking.

One note: because the PR targets a stacked branch, no hardware pipeline ran here (only pre-commit, docs, ASan build). Worth running tt-train post-commit once the stack is rebased onto main.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/compute/swiglu_packed_fw_kernel.cpp
mdragulaTT: +1, not blocking. At parity with the fused ttnn multiply on large shapes, fine to defer.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/compute/swiglu_packed_fw_kernel.cpp
```diff
@@ -0,0 +1,65 @@
...
+    tile_regs_commit();
+    cb_pop_front(cb_silu, block_size);
+    pack_and_push_block(cb_out, block_size);
+}
+
+void kernel_main() {
+    init_sfpu(cb_gate, cb_out);
+    // TODO(#52395): compute_kernel_hw_startup is a call-once API and should be the kernel's first Tensix-engine call,
+    // but here it follows another engine op (init_sfpu / a prior startup); see the issue.
+    compute_kernel_hw_startup(cb_gate, cb_up, cb_out);
```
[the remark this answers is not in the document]
mdragulaTT: Repo-wide pattern; every tt-train compute kernel carries the same TODO, tracked in #52395. Not for this PR.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
mdragulaTT: +1. This also covers Copilot's comment below: compare the full logical shape with the last dim halved, not only `[-1]`.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
```diff
@@ -0,0 +1,156 @@
...
+            dh_padded[0] == packed_padded[0] && dh_padded[1] == packed_padded[1],
+        "SwigluPackedBw: dL_dh padded shape {} must be packed {} with the last dim halved",
+        dh_padded,
+        packed_padded);
+    TT_FATAL(
+        dL_dh.logical_shape()[-1] * 2U == packed.logical_shape()[-1],
+        "SwigluPackedBw: dL_dh last logical dim {} must be half packed's {}",
+        dL_dh.logical_shape()[-1],
+        packed.logical_shape()[-1]);
```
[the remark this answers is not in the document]
mdragulaTT: +1, see Pavlo's comment above on L73.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
mdragulaTT: +1.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/kernels/compute/swiglu_packed_bw_kernel.cpp
```diff
@@ -0,0 +1,50 @@
...
+constexpr uint32_t cb_dup = tt::CBIndex::c_4;    // out: grad wrt up branch
+constexpr uint32_t cb_sigmoid = tt::CBIndex::c_5;
+constexpr uint32_t cb_scratch = tt::CBIndex::c_6;
+constexpr uint32_t cb_silu_grad = tt::CBIndex::c_7;
+
+void kernel_main() {
+    init_sfpu(cb_gate, cb_dgate);
+    // TODO(#52395): compute_kernel_hw_startup is a call-once API and should be the kernel's first Tensix-engine call,
+    // but here it follows another engine op (init_sfpu / a prior startup); see the issue.
+    compute_kernel_hw_startup(cb_gate, cb_up, cb_dgate);
```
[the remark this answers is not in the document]
mdragulaTT: Same as fw: #52395.

tt-train/sources/ttml/metal/common/swiglu_gate_bw_compute.hpp
mdragulaTT: +1.

tt-train/sources/ttml/CMakeLists.txt
```diff
@@ -180,6 +181,12 @@ set(METAL_OPS_FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_elemwise_bw/swiglu_elemwise_bw.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_elemwise_bw/device/swiglu_elemwise_bw_device_operation.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_elemwise_bw/device/swiglu_elemwise_bw_program_factory.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_packed_fw/swiglu_packed_fw.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_packed_fw/device/swiglu_packed_fw_device_operation.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_packed_fw/device/swiglu_packed_fw_program_factory.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_packed_bw/swiglu_packed_bw.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
```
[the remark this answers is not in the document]
mdragulaTT: Pre-existing for every tt-train op; kernels resolve through the runtime root. Not for this PR.

tt-train/tests/CMakeLists.txt
```diff
@@ -171,6 +171,14 @@ target_link_libraries(
...
         TTNN::TTNN
 )
+
+add_executable(swiglu_packed_benchmark benchmark/swiglu_packed_benchmark.cpp)
+target_link_libraries(
+    swiglu_packed_benchmark
+    PUBLIC
+        ttml
+        TTNN::TTNN
+)
```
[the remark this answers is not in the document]
mdragulaTT: Matches the two sibling benchmark targets above. PCH for all three would be its own PR.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
mdragulaTT: +1. `std::string_view` for the name makes it constexpr.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
mdragulaTT: +1. You already hand-converted these to markdown for the PR body. Also label the columns as host µs per launch; readers will compare them to device profiler numbers.

tt-train/tests/python/test_swiglu_packed.py
mdragulaTT: +1, checked: 128 blocks is one group on 8x8 and on 13x10.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/swiglu_packed_fw_device_operation.cpp
```diff
@@ -0,0 +1,145 @@
...
+// the corresponding shape of packed with the last dim halved.
+ttnn::Shape halve_last_dim(const ttnn::Shape& shape) {
+    return ttnn::Shape({shape[0], shape[1], shape[2], shape[-1] / 2U});
+}
+
+}  // namespace
+
+void SwigluPackedFwDeviceOperation::validate_on_program_cache_miss(
+    const operation_attributes_t& args, const tensor_args_t& tensor_args) {
+    auto check_tensor = [](const ttnn::Tensor& tensor, const std::string& name) {
```
mdragulaTT, line 26: Third identical copy of this lambda in the diff (bw L16, elemwise_bw L16); frobenius has the same checks inline and two more ops carry dtype variants. #56523 adds `check_device_tensor` / `check_same_device` in `metal/common/tensor_validation.hpp` for exactly this. Use it after rebase.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/swiglu_packed_fw_device_operation.cpp
```diff
@@ -0,0 +1,145 @@
...
+            name,
+            name);
+    };
+
+    const auto& packed = tensor_args.packed;
+    check_tensor(packed, "packed");
+
+    const auto& packed_padded = packed.padded_shape();
+    TT_FATAL(packed_padded.rank() == 4U, "SwigluPackedFw: packed must be 4D, got rank {}", packed_padded.rank());
+    const uint32_t two_tiles_w = 2U * tt::constants::TILE_WIDTH;
```
mdragulaTT, line 66: Rank, padded %64 and logical %64 are repeated verbatim in bw L59–71. One `validate_packed_gate_up(packed, op_name)` shared by both, next to the shared compute header.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/swiglu_packed_fw_program_factory.cpp
```diff
@@ -0,0 +1,172 @@
...
+    const operation_attributes_t& args, const tensor_args_t& tensor_args, tensor_return_value_t& output) {
+    const auto& packed = tensor_args.packed;
+
+    auto* const device = packed.device();
+    tt::tt_metal::Program program{};
+
+    const tt::DataFormat data_format = datatype_to_dataformat_converter(packed.dtype());
+    const uint32_t tile_size_bytes = tt::tile_size(data_format);
+
+    const auto padded_shape = packed.padded_shape();
```
mdragulaTT, line 82: L82–99 and bw L92–109 are the same work split, and `assign_per_core_runtime_args` plus the override loop are the walk #32533 tracks. #56524 adds `for_each_core_with_work` / `for_each_core` to `program_utils.hpp`; these two factories should be the first new ops on it. The Wt/Ht/blocks arithmetic can still be one small function shared by fw and bw.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/dataflow/reader_swiglu_packed_fw_interleaved.cpp
```diff
@@ -0,0 +1,38 @@
...
+    constexpr auto packed_args = TensorAccessorArgs<2>();
+    const auto packed_gen = TensorAccessor(packed_args, packed_addr);
+
+    // Packed row = 2*Wt tiles: gate [0, Wt), up [Wt, 2*Wt).
+    constexpr uint32_t packed_row_tiles = 2U * Wt;
+    constexpr uint32_t blocks_per_row = Wt / block_size;
+
+    const uint32_t end_block = start_block + num_blocks_to_process;
+    for (uint32_t b = start_block; b < end_block; ++b) {
+        const uint32_t gate_start = (b / blocks_per_row) * packed_row_tiles + (b % blocks_per_row) * block_size;
```
mdragulaTT, line 30: Same formula in bw reader L33 and bw writer L29. The `[gate | up]` layout is now stated in three kernels, two host validators and PR4's loader. One inline in `dataflow_utils.hpp`, or a `swiglu_packed_layout.hpp` beside the compute header, is the single place to get it wrong.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
```diff
@@ -0,0 +1,156 @@
...
+#include <enchantum/enchantum.hpp>
+
+#include "swiglu_packed_bw_program_factory.hpp"
+#include "ttnn/device_operation.hpp"
+
+namespace ttml::metal::ops::swiglu_packed_bw::device {
+
+void SwigluPackedBwDeviceOperation::validate_on_program_cache_miss(
+    const operation_attributes_t& args, const tensor_args_t& tensor_args) {
+    auto check_tensor = [](const ttnn::Tensor& tensor, const std::string& name) {
```
mdragulaTT, line 16: Same lambdas as fw L26; see the comment there.

tt-train/sources/ttml/metal/common/swiglu_gate_bw_compute.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include "api/compute/reg_api.h"
+#include "api/compute/tile_move_copy.h"
+#include "tt-train/sources/ttml/metal/common/compute_utils.hpp"
+
+// Compute one block of the SwiGLU-gating backward. The caller must have already
+// waited `block_size` tiles on cb_gate / cb_plain / cb_dh; this function fully manages the internal
+// cb_sigmoid / cb_scratch / cb_silu_grad and leaves cb_grad_gate / cb_grad_plain pushed. Input CBs
+// are popped by the caller.
+template <
+    uint32_t cb_gate,        // gate branch (silu'd)
```
mdragulaTT, line 31: Naming: this header says gate/plain, the packed ops say gate/up, and elemwise says linear1/gate where "gate" is the plain branch. Three vocabularies for two tensors. Use gate/up here to match the ops that call it; elemwise already maps its names at the call site.

tt-train/sources/ttml/metal/ops/swiglu_elemwise_bw/device/kernels/compute/swiglu_elemwise_bw_kernel.cpp
```diff
@@ -150,12 +33,18 @@ void kernel_main() {
             cb_wait_front(cb_gate, block_size);
             cb_wait_front(cb_dL_dprod, block_size);
 
-            compute_sigmoid();
-            compute_dL_dgate();
-            compute_silu_grad();
-            compute_dL_dlinear1();
+            // linear1 is the silu'd (gate) branch; cb_gate here is the plain branch.
+            swiglu_gate_bw_block<
```
mdragulaTT, line 37: The bw wins in the PR table on 32-tile-row shapes come from block distribution, not from packing; the math here is identical. Give this op the same `total_blocks` split. It is the MoE path via `moe_ffn_swiglu_op`. Here or a filed follow-up, but one of the two.

tt-train/sources/ttml/ops/swiglu_packed_op.cpp
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "swiglu_packed_op.hpp"
+
+#include "autograd/auto_context.hpp"
```
mdragulaTT, line 7: Unused include.

tt-train/sources/ttml/ops/swiglu_packed_op.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "metal/ops/swiglu_packed_fw/swiglu_packed_fw.hpp"
+
+namespace ttml::ops {
+
+autograd::TensorPtr swiglu_packed(const autograd::TensorPtr& packed) {
+    auto h = ttml::metal::swiglu_packed_fw(packed->get_value());
+    auto out = autograd::create_tensor(h);
+
+    autograd::GradFunction grad = [packed, out]() {
+        auto dL_dpacked = ttml::metal::swiglu_packed_bw(packed->get_value(), out->get_grad());
```
mdragulaTT, line 20: `preallocated_dL_dpacked` is priced in the benchmark and never used. `packed->get_value()` is dead after this node when `retain_graph` is off, the same situation as `linear1` in `swiglu_op.cpp:161`, and the kernel is in-place safe per block. Either wire it under the same condition or state here why not. Either way the 2x bw peak belongs in the PR body.

tt-train/sources/ttml/nanobind/nb_ops.cpp
```diff
@@ -554,6 +558,16 @@ void py_module(nb::module_& m) {
             nb::arg("use_per_device_seed") = true);
     }
 
+    {
+        auto py_swiglu_packed = static_cast<nb::module_>(m.attr("swiglu_packed"));
+        py_swiglu_packed.def(
+            "swiglu_packed",
+            &ttml::ops::swiglu_packed,
+            nb::arg("packed"),
+            "Packed-SwiGLU gating: `packed` [.., .., R, 2*I] (gate|up) -> h = silu(gate) * up "
```
mdragulaTT, line 567: `[.., .., R, 2*I]` here, `[.., R, 2*I]` everywhere else.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
```diff
@@ -0,0 +1,338 @@
...
+    const uint32_t inner = shape[-1] / 2U;
+    const ttsl::SmallVector<uint32_t> step = {1U, 1U, 1U, 1U};
+    const ttsl::SmallVector<uint32_t> gate_start = {0U, 0U, 0U, 0U};
+    const ttsl::SmallVector<uint32_t> gate_end = {1U, 1U, rows, inner};
+    const ttsl::SmallVector<uint32_t> up_start = {0U, 0U, 0U, inner};
+    const ttsl::SmallVector<uint32_t> up_end = {1U, 1U, rows, 2U * inner};
+    return {ttnn::slice(packed, gate_start, gate_end, step), ttnn::slice(packed, up_start, up_end, step)};
+}
+
+template <typename Fn>
```
mdragulaTT, line 163: `time_avg_us` and `capture_dram_peak` are the third private copy of these two; `polynorm_fusion_benchmark` and `swiglu_fusion_benchmark` have their own. Move both into `benchmark_utils.hpp`. `measure_average_iteration_time_s` there has no device sync, so it measures launch time and should probably be replaced by this one.

tt-train/tests/python/test_swiglu_packed.py
```diff
@@ -0,0 +1,180 @@
...
+
+def assert_backward(got, gate, up, dh, inner, label):
+    dgate, dup = reference_backward(as_stored(gate), as_stored(up), as_stored(dh))
+    assert_within_ulp(got[..., :inner], dgate, f"{label} dgate", MAX_ULP, MAX_ULP_P99_DGATE)
+    assert_within_ulp(got[..., inner:], dup, f"{label} dup", MAX_ULP, MAX_ULP_P99)
+
+
+class TestForward:
+    @pytest.mark.parametrize("batch,seq,inner", SHAPES)
+    def test_matches_reference(self, batch, seq, inner):
```
mdragulaTT, line 90: Same shape twice with different data would pin `override_runtime_arguments`. Today it only runs because two autograd shapes happen to repeat `SHAPES`.

mdragulaTT (changes requested):
Kernels read correctly and the A/B is what I asked for, thanks. Before I approve, three things at the top level:

- Duplication. The fw and bw ops repeat each other verbatim on packed-layout validation, the work split, runtime-arg assignment and the packed tile-index formula, and the `check_tensor` / `check_same_device` lambda is now its third identical copy in tt-train. Two helpers now exist for exactly this: `check_device_tensor` in #56523 and `for_each_core_with_work` in #56524 (tracked in #32533). Please rebase onto them and make these two factories the first new ops that use them; the remaining fw/bw pairs (layout validation, Wt/Ht/blocks split, tile-index formula) can be one small shared function each.
- Benchmark utils. `time_avg_us` and `capture_dram_peak` are the third private copy of those two helpers across `tests/benchmark`. Move them into `benchmark_utils.hpp` so the next benchmark does not roll a fourth.
- Block distribution. The bw gains in your table on short-row shapes come from splitting work by block rather than by tile-row; the math is shared. `swiglu_elemwise_bw` still splits by row and is the MoE path. Give it the same split here or file the follow-up, either is fine, but one of the two.

Also: the preallocated bw output is priced in the benchmark and never wired into the autograd op; decide and say why. +1 to Pavlo's inline set. Copilot's two red comments are repo-wide (#52395 and kernel packaging), not this PR's. No hardware job ran on this branch; please run tt-train post-commit once it is retargeted to main.

tt-train/sources/ttml/metal/ops/swiglu_elemwise_bw/device/kernels/compute/swiglu_elemwise_bw_kernel.cpp
imichalakTT: Filed this as an issue: https://github.com/tenstorrent/tt-metal/issues/56760.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
imichalakTT: Added shape and memory config to the hash computation in https://github.com/tenstorrent/tt-metal/issues/56760.

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/compute/swiglu_packed_fw_kernel.cpp
imichalakTT: Leaving it for now

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/compute/swiglu_packed_fw_kernel.cpp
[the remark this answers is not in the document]
imichalakTT: I removed the `init_sfpu` call. It is also considered deprecated. Now `compute_kernel_hw_startup` is called first.

https://github.com/tenstorrent/tt-metal/pull/52817/changes/688c7a358f1f1e262914ba79e25f1307a327743a

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/kernels/compute/swiglu_packed_bw_kernel.cpp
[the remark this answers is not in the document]
imichalakTT: I removed the `init_sfpu` call. It is also considered deprecated. Now `compute_kernel_hw_startup` is called first.

https://github.com/tenstorrent/tt-metal/pull/52817/changes/688c7a358f1f1e262914ba79e25f1307a327743a

tt-train/sources/ttml/metal/common/swiglu_gate_bw_compute.hpp
imichalakTT: Solved in https://github.com/tenstorrent/tt-metal/pull/52817/changes/42a327fe5d48c7c5c818a9af5335e2cc6099e81c.

I attached a table with results to the PR. TLDR:

Backward, median of 3 runs vs. the previous header: −9.6% at I=14336 / 4k rows, −3.0% at I=5632 / 4k rows, within noise (±1.5%) on the other five shapes; −2.1% mean across all seven.

mdragulaTT (approved):
Thanks for the round, approving. The shared header and the relaunch and two-group tests cover most of my review; those threads are resolved. Four are still open, please address them before merging, a line or a small fix each:
- [gate/plain vs gate/up naming in the shared bw header](https://github.com/tenstorrent/tt-metal/pull/52817#discussion_r4008089752)
- [the packed tile-index formula, still in three kernels](https://github.com/tenstorrent/tt-metal/pull/52817#discussion_r4008089740)
- [`time_avg_us` / `capture_dram_peak` into `benchmark_utils.hpp`, plus Vlad's two nits](https://github.com/tenstorrent/tt-metal/pull/52817#discussion_r4008089784)
- [`preallocated_dL_dpacked`: wire it or say why not](https://github.com/tenstorrent/tt-metal/pull/52817#discussion_r4008089776)

FYI, #56523 and #56524 add `check_device_tensor` and `for_each_core_with_work` for exactly what `swiglu_packed_common.hpp` and the two factories do by hand. If this merges first, keep it as is and we migrate it with the rest; if those land first, please switch to them here.

CI runs `test_swiglu_packed.py` only after merge, in the unit pipeline on Blackhole, so a run on a Blackhole box before merging would be good.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
imichalakTT: Changed to Markdown in https://github.com/tenstorrent/tt-metal/pull/52817/changes/fd3627aa8a8efcd6a611953b3067fb3c77628ecc

imichalakTT: Also added the note: 'µs are host time per launch'.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
imichalakTT: addressed here https://github.com/tenstorrent/tt-metal/pull/52817/changes/fd3627aa8a8efcd6a611953b3067fb3c77628ecc

tt-train/sources/ttml/metal/ops/swiglu_packed_fw/device/kernels/dataflow/reader_swiglu_packed_fw_interleaved.cpp
imichalakTT: Promoted gate tile calculation to a shared helper in https://github.com/tenstorrent/tt-metal/pull/52817/changes/3e69e503e41cbfbae544c0b8ee65d57b43910cb5.

tt-train/sources/ttml/metal/common/swiglu_gate_bw_compute.hpp
imichalakTT: Addressed in https://github.com/tenstorrent/tt-metal/pull/52817/changes/3e69e503e41cbfbae544c0b8ee65d57b43910cb5.

tt-train/tests/benchmark/swiglu_packed_benchmark.cpp
imichalakTT: Moved them into utils. https://github.com/tenstorrent/tt-metal/pull/52817/changes/fd3627aa8a8efcd6a611953b3067fb3c77628ecc

tt-train/sources/ttml/ops/swiglu_packed_op.cpp
imichalakTT: The saving should be around 224 MB in total allocations for Llama 8B, or 56 MB per device with TP=4. It would not affect the peak memory which sits at the LM head. I would leave this out of this PR.

We could do:

```cpp
// With `grad_overwrites_input` the backward writes dL/dpacked into packed's own buffer instead of a
// new 2I-wide one; only for a caller that is packed's sole consumer and does not retain the graph.
autograd::TensorPtr swiglu_packed(const autograd::TensorPtr& packed, bool grad_overwrites_input = false);
```

But this extends the API and the behavior would rely on implicit assumptions.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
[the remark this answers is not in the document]
imichalakTT: Fixed in https://github.com/tenstorrent/tt-metal/pull/52817/changes/1689a2a936e8e6ec20f026590c23035a922de25d together with Pavlo's comment on L73: `validate_half_of_packed` now compares `dL_dh`'s full padded and logical shapes against `packed` with the last dim halved, so a different logical row count, or a different rank, fails the shape check.

tt-train/sources/ttml/CMakeLists.txt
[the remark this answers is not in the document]
imichalakTT: Pre-existing for every tt-train op: all 30 program factories reference kernels by a `tt-train/sources/...` path resolved through the runtime root. Not changed in this PR.

tt-train/tests/CMakeLists.txt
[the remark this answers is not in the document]
imichalakTT: Added in https://github.com/tenstorrent/tt-metal/pull/52817/changes/688c7a358f1f1e262914ba79e25f1307a327743a: `target_precompile_headers(swiglu_packed_benchmark REUSE_FROM ttml_pch)`, the same treatment main gave the other benchmark targets in #53733.

tt-train/sources/ttml/metal/ops/swiglu_packed_bw/device/swiglu_packed_bw_device_operation.cpp
imichalakTT: Done in https://github.com/tenstorrent/tt-metal/pull/52817/changes/1689a2a936e8e6ec20f026590c23035a922de25d. `validate_half_of_packed` compares `dL_dh`'s full padded and logical shapes with `halve_last_dim(packed)`; a 3D grad now fails that comparison with both shapes in the message, and the logical row count is checked too. The same check covers Copilot's thread below.

tt-train/tests/python/test_swiglu_packed.py
imichalakTT: You're right, 128 blocks is one group on both 8x8 and 13x10. Corrected the comment and added `(1, 4192, 32)`: 131 tile-rows at Wt = 1 give 131 blocks, which neither 64 nor 130 cores divide, so both compute kernels run and the runtime-arg walk crosses the group boundary. https://github.com/tenstorrent/tt-metal/pull/52817/changes/1689a2a936e8e6ec20f026590c23035a922de25d

tt-train/sources/ttml/ops/swiglu_packed_op.cpp
mdragulaTT: Ack, thanks for the numbers, agreed. One non-blocking ask: a one-line comment at the call site saying `packed` is not ours to overwrite and that a fused caller which owns it can pass `preallocated_dL_dpacked`, so the next reader does not re-ask.

PR #49504, opened 2026-07-09
Add GRPO example with remote rollout

ichovpanTT:
Move GRPO rollout to a separate subset of devices that use tt-transformers. 

## Summary of changes + tradeoffs (very useful)

https://claude.ai/code/artifact/17a00c1c-b442-4a73-b6cd-e5cf63819968?via=auto_preview

## Architecture image

Screenshot 2026-07-09 at 9 50 12 AM

## Architecture
2 processes involved: one for training, weight updates, and user code (rank 0), another for token generation (rank 1).

Rank 0 uses 1 n300 (2 wh devices), rank 1 also uses one n300. Tested on wh loudbox & bh loudbox.

Using 4 devices for weight updates and 4 for token generation might result in a hang.

Other splits (like 2-6) are not supported. (#50279, #50297)

### Rank 0 (training + user code)
The user of the GRPOTrainer passes LlamaCompleterRemoteRollout instance, which inherits GRPOCompleter and provides 3 methods: generate(), generate_str(), compute_nlog_probs(). compute_nlog_probs is computed locally (on the ttml rank), but generate() and generate_str() call remote token generation using MPIRolloutClient.

The requests for generating tokens and the prompt completions are passed through MPI, which is implemented in mpi_rollout.py. See MPIRolloutClient and MPIRolloutServer classes for more details.

The logic for weight updates, advantage & loss computations, loss.backward() & optimizer.step() calls remain in the GRPOTrainer.

### Weight updates
are done on the user side as a GRPOTrainer callback after each step.

### Rank 1 (token generation)
Starts MPIRolloutServer, which waits indefinitely for commands over MPI. Supported commands:
* Generate a completion for a prompt -- calls multiple TTTGenerationWorker instances to produce completions on multiple devices
* Request weight transfer (use WeightBridge for further communication)
* Shutdown of the server

### TTTGenerationWorker
Contains prefill & decode calls of the tt-transformers library + various constants.

Each TTTGenerationWorker owns a submesh of [1, 1].  

Right now TTTGenerationWorker.generate() calls are done sequentially, bringing the speed down. Migrating to use 1 instance for all prefills and decodes together

### WeightBridge

Abstract class that defines the methods for transferring model weights. Main assumption: weights can be stored as a python dictionary with string keys & tensor values.

#### WeightBridge methods
connect(), send_weights(), receive_weights(), barrier().

#### Only WeightBridge implementation: HostWeightBridge

[...]

PR #52818, opened 2026-08-11
[tt-train] Fuse the QKV and gate/up projections

imichalakTT:
Fuse Llama's Q+KV projections into `qkv_linear` and gate+up into `w_gate_up`, feeding packed SwiGLU. This removes one projection matmul from attention and one from the MLP on both single-device and tensor-parallel paths.

Updates GQA head splitting, LoRA examples/tests, Muon optimizer routing, and GRPO attention overrides to use the fused projections. Part of #52951; builds on the packed SwiGLU change.

For sequence parallelism, merging each projection pair enables one fused all-gather+matmul, overlapping communication and compute without duplicate collectives. Plain TP can gather activations once before separate projections, but that shared gather sits outside the matmuls and prevents MM+CCL fusion. Single-device execution has no collectives, so its benefit comes from the compute changes.

## Benchmarks

Baseline: parent commit with no projection fusion. The four variants use the same build and vary only `gqattn.py` (Q+KV) and `transformer.py` (gate+up). TPS = tokens/s.

### Llama-3 8B — TP=4, Blackhole QuietBox

mesh `[1,4]` · bs 1 · grad accum 1 · seq 2048 · AdamW 3e-5 · 10 steps

| variant | ms/step | TPS | ΔTPS | MFU | peak DRAM | Δpeak |
|---|--:|--:|--:|--:|--:|--:|
| baseline | 616.4 | 3323 | — | 27.0% | 21,462.27 MB | — |
| Q+KV fused | 601.5 | 3405 | +2.5% | 27.7% | 21,462.19 MB | −0.0% |
| gate+up fused | 592.5 | 3456 | +4.0% | 28.1% | 21,014.27 MB | −2.1% |
| **both — this PR** | **578.1** | **3543** | **+6.6%** | **28.8%** | **21,014.19 MB** | **−2.1%** |

### TinyLlama 1.1B — single device

mesh `[1,1]` · bs 5 · seq 2048 · AdamW 3e-4 · 20 steps

| variant | ms/step | TPS | ΔTPS | MFU | peak DRAM |
|---|--:|--:|--:|--:|--:|
| baseline | 1521.0 | 6733 | — | 33.2% | 26,666.16 MB |
| Q+KV fused | 1513.3 | 6767 | +0.5% | 33.3% | 26,666.17 MB |
| gate+up fused | 1513.7 | 6765 | +0.5% | 33.3% | 26,666.16 MB |
| **both — this PR** | **1504.3** | **6807** | **+1.1%** | **33.5%** | **26,666.16 MB** |

### Loss across six seeds — Llama-3 8B

Baseline versus both fusions; positive deltas mean higher loss.

| seed | baseline@10 | both@10 | Δ@10 | Δmean over 10 steps |
|---|--:|--:|--:|--:|
| 5489 | 9.4375 | 9.7500 | +0.3125 | +0.0500 |
| 1234 | 9.6250 | 9.6875 | +0.0625 | +0.0312 |
| 777 | 9.3125 | 9.5000 | +0.1875 | +0.1875 |
| 4242 | 9.5625 | 9.3125 | −0.2500 | −0.2375 |
| 31337 | 9.8125 | 9.9375 | +0.1250 | +0.0312 |
| 99 | 9.8750 | 9.8125 | −0.0625 | −0.2000 |

### Review notes

- Parameter names and layouts change to `qkv_linear` and `w_gate_up`; checkpoint consumers and LoRA targets must account for them.
- Fused weights use AdamW under `MuonWithAdamW`; their component matrices need separate orthogonalization.
- Dropout seeding is unchanged: the pre-fusion TP path already shared one seed across the mesh, and `use_per_device_seed=not use_tp` only makes that explicit. The DP+TP mask correlation it leaves predates this PR and is filed as #55947.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:imichalak/llama-sp/3-qkv-gate-up-fusion)

### DIFF
--- a/tt-train/sources/examples/grpo/utils/llama_overrides.py
+++ b/tt-train/sources/examples/grpo/utils/llama_overrides.py
@@ -2,93 +2,17 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import ttnn
 import ttml
 
-from ttml.models.llama.gqattn import GroupedQueryAttention
 from ttml.models.llama.transformer import LlamaBlock
 from ttml.models.llama import Llama
-from ttml.modules import RunMode
-
-
-class GroupedQueryAttentionCompositeKV(GroupedQueryAttention):
-    def forward_no_kv(self, input: ttml.autograd.Tensor, mask: ttml.autograd.Tensor) -> ttml.autograd.Tensor:
-        q = self.q_linear(input)
-        kv = self.kv_linear(input)
-
-        q_heads, k_heads, v_heads = ttml.ops.multi_head_utils.grouped_heads_creation(
-            q, kv, self.num_heads, self.num_groups
-        )
-
-        q_heads = ttml.ops.rope.rope(q_heads, self.rope_params)
-        k_heads = ttml.ops.rope.rope(k_heads, self.rope_params)
-
-        # Composite SDPA supports non-broadcast masks like (B, 1, S, S)
-        attention = ttml.ops.attention.scaled_dot_product_attention_composite(q_heads, k_heads, v_heads, mask)
-        attention = ttml.ops.multi_head_utils.heads_fusion(attention)
-
-        out = self.out_linear(attention)
-
-        # Match base behavior in training mode
-        if self.get_run_mode() == RunMode.TRAIN and self.dropout_prob > 0.0:
-            out = ttml.ops.dropout.dropout(out, self.dropout_prob)
-
-        return out
-
-    def forward_kv(
-        self,
-        input: ttml.autograd.Tensor,
-        mask: ttml.autograd.Tensor,
-        kv_cache: ttml.models.KvCache,
-        layer_idx: int,
-        new_tokens: int,
-    ) -> ttml.autograd.Tensor:
-        q = self.q_linear(input)
-        kv = self.kv_linear(input)
-
-        q_heads, k_heads, v_heads = ttml.ops.multi_head_utils.grouped_heads_creation(
-            q, kv, self.num_heads, self.num_groups
-        )
-
-        token_pos = kv_cache.get_cache_position()
-        q_heads = ttml.ops.rope.rope(q_heads, self.rope_params, token_pos)
-        k_heads = ttml.ops.rope.rope(k_heads, self.rope_params, token_pos)
-
-        kv_cache.update(layer_idx, k_heads.get_value(), v_heads.get_value(), new_tokens)
-
-        k_cache = kv_cache.get_k_cache(layer_idx)
-        v_cache = kv_cache.get_v_cache(layer_idx)
-
-        token_end = [k_cache.shape[0], k_cache.shape[1], mask.shape()[-1], k_cache.shape[3]]
-        step = [1, 1, 1, 1]
-        k_cache_slice = ttnn.slice(k_cache, [0, 0, 0, 0], token_end, step)
-        v_cache_slice = ttnn.slice(v_cache, [0, 0, 0, 0], token_end, step)
-
-        k_cache_to_process = ttml.autograd.create_tensor(k_cache_slice)
-        v_cache_to_process = ttml.autograd.create_tensor(v_cache_slice)
-
-        # Change: composite SDPA in decode path
-        attention = ttml.ops.attention.scaled_dot_product_attention_composite(
-            q_heads, k_cache_to_process, v_cache_to_process, mask
-        )
-        attention = ttml.ops.multi_head_utils.heads_fusion(attention)
-        out = self.out_linear(attention)
-        return out
 
 
 class LlamaBlockCompositeKV(LlamaBlock):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-        # preserve existing config values
-        self.attention = GroupedQueryAttentionCompositeKV(
-            embedding_size=self.attention.embedding_size,
-            num_heads=self.attention.num_heads,
-            num_groups=self.attention.num_groups,
-            dropout=self.attention.dropout_prob,
-            rope_params=self.attention.rope_params,
-            bias_linears=False,  # match your model config if needed
-        )
+        # Composite SDPA accepts non-broadcast masks like (B, 1, S, S).
+        self.attention.sdpa = ttml.ops.attention.scaled_dot_product_attention_composite
 
 
 class LlamaCompositeKV(Llama):
@@ -96,14 +20,6 @@ class LlamaCompositeKV(Llama):
         super().__init__(config)
         self.create_name("Llama")
 
-        # do NOT rebuild self.blocks; patch attention in-place
-        for i, block in enumerate(self.blocks):
-            old = block.attention
-            block.attention = GroupedQueryAttentionCompositeKV(
-                embedding_size=old.embedding_size,
-                num_heads=old.num_heads,
-                num_groups=old.num_groups,
-                dropout=old.dropout_prob,
-                rope_params=old.rope_params,
-                bias_linears=config.attention_bias,
-            )
+        # Composite SDPA accepts non-broadcast masks like (B, 1, S, S).
+        for block in self.blocks:
+            block.attention.sdpa = ttml.ops.attention.scaled_dot_product_attention_composite
--- a/tt-train/sources/ttml/ttml/models/llama/gqattn.py
+++ b/tt-train/sources/ttml/ttml/models/llama/gqattn.py
@@ -16,10 +16,10 @@ from ttml.modules import AbstractModuleBase, LinearLayer, ColumnParallelLinear, 
 class GroupedQueryAttention(AbstractModuleBase):
     """Grouped-query attention (GQA) with optional tensor-parallel linear layers.
 
-    When ``use_tp=True`` the Q and KV projections use ``ColumnParallelLinear``
-    (output features sharded) with ``gather_output=False``, and the output
-    projection uses ``RowParallelLinear`` with ``input_is_parallel=True``.
-    This avoids redundant communication between the two matmuls.
+    Under tensor parallelism the fused QKV projection uses ``ColumnParallelLinear``
+    (output features sharded) with ``gather_output=False``, and the output projection
+    uses ``RowParallelLinear`` with ``input_is_parallel=True``. This avoids redundant
+    communication between the two matmuls.
     """
 
     def __init__(
@@ -44,14 +44,15 @@ class GroupedQueryAttention(AbstractModuleBase):
         self.dropout_prob = dropout
         self.rope_params = rope_params
 
-        # concat_kv_dim uses GLOBAL head counts because the weight matrices are
-        # created at full size and then sharded by ColumnParallelLinear.
-        concat_kv_dim = 2 * num_groups * (embedding_size // num_heads)
+        head_dim = embedding_size // num_heads
+        qkv_dim = (num_heads + 2 * num_groups) * head_dim  # == embedding_size + 2 * num_groups * head_dim
 
-        # Head/group counts stored here are LOCAL (per-device) so that reshaping
-        # in grouped_heads_creation matches the sharded activation width.
         if use_tp:
             tp_size = ttml.mesh().axis_size("tp")
+            if num_heads % tp_size != 0:
+                raise ValueError(f"num_heads ({num_heads}) must be divisible by the tensor-parallel size ({tp_size})")
+            if num_groups % tp_size != 0:
+                raise ValueError(f"num_groups ({num_groups}) must be divisible by the tensor-parallel size ({tp_size})")
             self.num_heads = num_heads // tp_size
             self.num_groups = num_groups // tp_size
         else:
@@ -59,17 +60,9 @@ class GroupedQueryAttention(AbstractModuleBase):
             self.num_groups = num_groups
 
         if use_tp:
-            self.q_linear = ColumnParallelLinear(
+            self.qkv_linear = ColumnParallelLinear(
                 embedding_size,
-                embedding_size,
-                has_bias=bias_linears,
-                bias_init=ttml.init.zeros(),
-                gather_output=False,
-                axis_name="tp",
-            )
-            self.kv_linear = ColumnParallelLinear(
-                embedding_size,
-                concat_kv_dim,
+                qkv_dim,
                 has_bias=bias_linears,
                 bias_init=ttml.init.zeros(),
                 gather_output=False,
@@ -84,15 +77,9 @@ class GroupedQueryAttention(AbstractModuleBase):
                 axis_name="tp",
             )
         else:
-            self.q_linear = LinearLayer(
+            self.qkv_linear = LinearLayer(
                 embedding_size,
-                embedding_size,
-                bias_linears,
-                bias_init=ttml.init.zeros(),
-            )
-            self.kv_linear = LinearLayer(
-                embedding_size,
-                concat_kv_dim,
+                qkv_dim,
                 bias_linears,
                 bias_init=ttml.init.zeros(),
             )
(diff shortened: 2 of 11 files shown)


tt-train/sources/ttml/ttml/models/llama/gqattn.py
```diff
@@ -103,25 +93,32 @@ def __init__(
...
-        q_heads, k_heads, v_heads = ttml.ops.multi_head_utils.grouped_heads_creation(
-            q, kv, self.num_heads, self.num_groups
-        )
+        q_heads, k_heads, v_heads = ttml.ops.multi_head_utils.heads_creation(qkv, self.num_heads, self.num_groups)
 
         q_heads = ttml.ops.rope.rope(q_heads, self.rope_params)
         k_heads = ttml.ops.rope.rope(k_heads, self.rope_params)
 
-        attention = ttml.ops.attention.scaled_dot_product_attention(q_heads, k_heads, v_heads, mask)
+        attention = self.sdpa(q_heads, k_heads, v_heads, mask)
```
philei-tt, line 111: I would prefer to explicitly call scaled dot product attention here, instead of adding a separate function just to shorten name

tt-train/sources/ttml/ttml/models/llama/transformer.py
```diff
@@ -73,25 +73,33 @@ def __init__(
...
-                intermediate_size,
-                has_bias=False,
-                gather_output=False,
-                axis_name="tp",
+        # Fused gate+up: one [2*I, E] weight, rows [0:I) = gate, [I:2*I) = up.
+        gate_up_size = 2 * intermediate_size
+
+        # Each device's gate|up half is I/tp wide and must stay tile-aligned, or the packed
+        # SwiGLU op's two halves straddle a tile boundary.
+        tp_size = ttml.mesh().axis_size("tp") if use_tp else 1
```
philei-tt, line 86: (nit) feels a bit error prone. What if tp axis doesn't exist and use_tp is passed or vise-versa? Either need additional verification or maybe just get rid of use_tp, and use tp_size = ttml.mesh().axis_size("tp") if ttml.mesh().axis_size("tp") exists and > 1?

tt-train/sources/examples/grpo/utils/llama_overrides.py
```diff
@@ -2,108 +2,24 @@
...
-            block.attention = GroupedQueryAttentionCompositeKV(
-                embedding_size=old.embedding_size,
-                num_heads=old.num_heads,
-                num_groups=old.num_groups,
-                dropout=old.dropout_prob,
-                rope_params=old.rope_params,
-                bias_linears=config.attention_bias,
-            )
+        # Composite SDPA accepts non-broadcast masks like (B, 1, S, S).
+        for block in self.blocks:
```
ichovpanTT, line 24: `tt-train/sources/examples/grpo_remote_rollout/utils/llama_overrides.py` — the sibling of this file — isn't touched by this PR or any later branch in the stack (I diffed through `5-seq-parallelism`), and both of its exports break as soon as this PR lands:

**1. `GroupedQueryAttentionCompositeKV` — first forward crashes.** It still subclasses `GroupedQueryAttention` and does its own projection:

```python
def forward_no_kv(self, input, mask):
    q = self.q_linear(input)
    kv = self.kv_linear(input)
    q_heads, k_heads, v_heads = ttml.ops.multi_head_utils.grouped_heads_creation(
        q, kv, self.num_heads, self.num_groups
    )
    ...
```

`self.q_linear` / `self.kv_linear` no longer exist after this PR, so first forward raises `AttributeError`. `forward_kv` has the same two lines. Suggest converting it to the exact pattern you're using here — drop the subclass entirely and have `LlamaCompositeKV.__init__` do

```python
for block in self.blocks:
    block.attention.sdpa = ttml.ops.attention.scaled_dot_product_attention_composite
```

The ~80-line subclass collapses to two lines and the two sibling files stop drifting.

**2. `weights_ref_hf_dict()` — crashes on first call, and layout is wrong under TP.** It reads `parameters()[...]` by literal string key:

```python
out[f"model.layers.{i}.self_attn.q_proj.weight"] = get(f"{p}/attention/q_linear/weight")
...
kv = get(f"{p}/attention/kv_linear/weight")
out[f"model.layers.{i}.self_attn.k_proj.weight"] = ttnn.slice(kv, [0,0,0,0], [1,1,kv_dim,H])
out[f"model.layers.{i}.self_attn.v_proj.weight"] = ttnn.slice(kv, [0,0,kv_dim,0], [1,1,2*kv_dim,H])
out[f"model.layers.{i}.mlp.gate_proj.weight"] = get(f"{p}/mlp/w1/weight")
out[f"model.layers.{i}.mlp.up_proj.weight"] = get(f"{p}/mlp/w3/weight")
```

`get()` raises `RuntimeError` on the first missing key (`q_linear/weight`) at layer 0. Fixing the rename means one slice becomes three (Q/K/V out of `qkv_linear`) and gate+up becomes two slices out of `w_gate_up`. The docstring's "single-device assumption" keeps slice-based reconstruction correct on a unit mesh; any TP mesh would need to first reverse the per-shard interleave (see comment on `gqattn.py`) before slicing.

Both surfaces are live: `benchmark_weight_export.py` (`from utils.llama_overrides import LlamaCompositeKV`) and `llama_grpo_completer.py` (`from .llama_overrides import LlamaCompositeKV`) both instantiate the class and rely on `weights_ref_hf_dict()` for the ttml → tt-transformers weight sync in remote rollout. Currently blocked out of the box after this PR.

Unrelated aside: `tt-train/tools/weight-utils/llama_export.py` also maps `attention/q_linear/weight → self_attn.q_proj.weight` etc.; no commits on it in ~12 months so it may be dead, but worth confirming rather than leaving a stale reverse map next to the loader.

ichovpanTT (commented):
Can you please take a look at these comments?

tt-train/sources/ttml/ttml/models/llama/gqattn.py
```diff
@@ -44,32 +44,25 @@ def __init__(
...
-                has_bias=bias_linears,
-                bias_init=ttml.init.zeros(),
-                gather_output=False,
-                axis_name="tp",
-            )
-            self.kv_linear = ColumnParallelLinear(
-                embedding_size,
-                concat_kv_dim,
+                qkv_dim,
                 has_bias=bias_linears,
```
ichovpanTT, line 66: Documenting the fused-weight TP layout invariant here since it's easy to miss and the same trap sits in three separate places (the loader on branch #4, the two `grpo_remote_rollout` files in the other comment, and any future exporter).

`ColumnParallelLinear` shards `(out_features, in_features)` **contiguously** along dim 2 via `mesh.axis_mapper("tp", tdim=2)` — no reordering, no interleaving. Meanwhile the per-device consumer interprets its local width as `[Q_local | K_local | V_local]` (via `heads_creation(qkv, self.num_heads, self.num_groups)` with LOCAL counts) or `[gate_local | up_local]` (via `swiglu_packed`). Those two facts only line up when the **global** layout of the fused weight is per-shard interleaved:

```
qkv_linear:  [ Q_s0 | K_s0 | V_s0 | Q_s1 | K_s1 | V_s1 | … | Q_s{tp-1} | K_s{tp-1} | V_s{tp-1} ]
w_gate_up:   [ gate_s0 | up_s0 | gate_s1 | up_s1 | … | gate_s{tp-1} | up_s{tp-1} ]
```

not the natural `[Q_all | K_all | V_all]` / `[gate_all | up_all]`.

Concrete numbers, Llama-3-8B TP=4: `num_heads=32`, `num_groups=8`, `hidden=4096`, `head_dim=128` → `qkv_dim=6144`, 1536 rows/device. Under naive `[Q|K|V]`:

| device | rows | actual content | what local `heads_creation` labels it as |
| --- | --- | --- | --- |
| 0 | `[0, 1536)` | 12 Q heads | `[8 Q \| 2 K \| 2 V]` |
| 1 | `[1536, 3072)` | 12 Q heads | `[8 Q \| 2 K \| 2 V]` |
| 2 | `[3072, 4608)` | 8 Q + 4 K | `[8 Q \| 2 K \| 2 V]` |
| 3 | `[4608, 6144)` | 4 K + 8 V | `[8 Q \| 2 K \| 2 V]` |

So on device 0, four of its 12 real Q heads get wired into the K and V paths of attention. Same story for `w_gate_up`: `[gate_all | up_all]` puts device 0 on pure gate columns, which then get split into `[gate_local | up_local]` by `swiglu_packed`.

The reason this doesn't show up in the loss curves is that at initialization every element of the fused weight is drawn i.i.d. from `Uniform(-k, k)` (`k = sqrt(1/in_features)`), so no row *is* Q head 3 until forward decides it is. Any consistent labeling produces a valid model. The failure mode only appears the moment real semantics enter the tensor (HF checkpoint load, weight sync back to tt-transformers, HF export).

Branch `4-safetensors-loader`'s `_assemble()` already handles this for the HF loader — it splits each source block into `mesh_size` slices and lays them out as `[block_i, rank_r]` in the right nested order, so the interleave is correct for any number of blocks. Just calling it out here because:

- the two surfaces in the `llama_overrides.py` comment both currently assume a `[Q | K | V]` / `[gate | up]` layout and would need the same interleave (or unshuffle, in the export direction) before they're TP-safe;
- if the C++ Llama loader ever moves to `qkv_linear`, same story.

A one-line comment right here — something along the lines of "under TP the sharded output is `[Q_local | K_local | V_local]`; any code that materializes this weight from a fused source must interleave per-shard" — would save the next person a debugging session.

ichovpanTT:
Please run the grpo_remote_rollout/boolq example to see if it works, the reward should go up. The reference graphs are in this PR (https://github.com/tenstorrent/tt-metal/pull/49504)

When running the example, make sure the temperature is set to 1.0, not 1.5 (more details here https://github.com/tenstorrent/tt-metal/pull/56016)

philei-tt (commented):
Please rebase and add tests for a new heads creation variant

tt-train/sources/ttml/ttml/models/llama/gqattn.py
```diff
@@ -45,32 +45,25 @@ def __init__(
...
+                raise ValueError(f"num_groups ({num_groups}) must be divisible by the tensor-parallel size ({tp_size})")
             self.num_heads = num_heads // tp_size
             self.num_groups = num_groups // tp_size
         else:
             self.num_heads = num_heads
             self.num_groups = num_groups
 
         if use_tp:
-            self.q_linear = ColumnParallelLinear(
+            self.qkv_linear = ColumnParallelLinear(
```
nmauriceTT, line 64: **[P1] Keep pretrained loading functional in this PR**

This rename leaves the current safetensors loader requesting `attention/q_linear/weight`, `attention/kv_linear/weight`, `mlp/w1/weight`, and `mlp/w3/weight`, none of which exist after this change.

I reproduced this by adding `TestLlama::test_safetensors_loader_fused_projection_names` and running:

```bash
scripts/run_safe_pytest.sh \
  tt-train/tests/python/test_llama.py::TestLlama::test_safetensors_loader_fused_projection_names -q
```

It fails with:

```text
FAILED tt-train/tests/python/test_llama.py::TestLlama::test_safetensors_loader_fused_projection_names
RuntimeError: Parameter Llama/blocks/0/attention/q_linear/weight not found in model
```

This breaks the GRPO and LoRA entrypoints that call this loader. Although #52819 contains the fused-weight assembly, merging this PR independently leaves the branch broken. Please include that loader update and this regression coverage here, or otherwise ensure the changes land atomically.

tt-train/sources/ttml/ttml/models/llama/gqattn.py
imichalakTT: I have added `.sdpa` as an override point for GRPO.

GRPO replaces fused sdpa with a composite implementation
```
for block in self.blocks:
  block.attention.sdpa = ttml.ops.attention.scaled_dot_product_attention_composite
```

Citing a comment from that file:

```
1. Swaps every block's attention kernel for scaled_dot_product_attention_composite: 
the completer emits non-broadcast masks ((B, 1, S, S)) that only the composite kernel accepts,
and it makes the decode path (KV-cache slice -> attention) actually work.
```

tt-train/sources/ttml/ttml/models/llama/transformer.py
imichalakTT: `ttml.mesh().axis_size("tp")` raises if mesh is not open or there is no axis named 'tp'. So use_tp=True fails with a clear error.

```
no mesh open:

RuntimeError: Device mesh is not initialized.
Use ttml.open_device_mesh(ttml.Mesh(shape, axis_names)) to initialize the mesh.

no tp axis:

RuntimeError: Mesh has no axis named 'tp'.
  Shape: (1, 2)
  Axis names: ('dp', 'x')
```

The opposite scenario, meaning 'tp' axis exists, but use_tp=False, is just treated as replicated setup.

tt-train/sources/ttml/ttml/models/llama/gqattn.py
imichalakTT: I see. Thanks for pointing this out. I will see what is the best option.

tt-train/sources/examples/grpo/utils/llama_overrides.py
imichalakTT: Fixed in this PR. I have implemented a matching SDPA override and updated `weights_ref_hf_dict()`.

tt-train/sources/ttml/ttml/models/llama/gqattn.py
imichalakTT: I have added comments for qkv_linear and w_gate_up.

```
# Each shard's rows are [Q_local | K_local | V_local], so a weight built from a fused
# source must be interleaved per shard, not laid out [Q | K | V] globally.

# Each shard's rows are [gate_local | up_local], so a weight built from a fused source
# must be interleaved per shard, not laid out [gate | up] globally.
```

abustamanteTT (approved):

tt-train/sources/examples/grpo_remote_rollout/utils/llama_overrides.py
```diff
@@ -150,6 +74,9 @@ def get(name: str) -> ttnn.Tensor:
                 )
             return params[name].get_value()
 
+        def rows(weight: ttnn.Tensor, start: int, end: int) -> ttnn.Tensor:
```
jmalone-tt, line 77: could become a problem since weight sync peak allocation grows due to all slices being live at once. Every slice will be retained until after `update_weights`
Slicing per-layer and freeing as we go might help if needed in the future
Might be worth fixing if it's quick, or at least leave a comment to look here in the future if we OOM

jmalone-tt (commented):
Just added one comment, please finish addressing Nathan's comment as well and then I can re-review

tt-train/sources/ttml/ttml/models/llama/gqattn.py
imichalakTT: I have a plan to squash those two PRs just before merging (loader into fusion). I don't want to disrupt the review process at this stage.

tt-train/sources/examples/grpo_remote_rollout/utils/llama_overrides.py
imichalakTT: I have decided to add a comment above rows fn in https://github.com/tenstorrent/tt-metal/pull/52818/changes/3d2e4fc3ddeb476aecd74c59f9fce33fdd6954e3.

```
# Each call allocates a fresh device copy, and all of them (5 per layer, 80 in
# total, ~1.2 GB for Llama-3.2-1B) stay live until push_weights drops the dict,
# which happens only after the receiver has finished update_weights.
# Fine today: the sync runs after the optimizer step, when the chip holds ~2.3 GB
# of weights and ~4.6 GB of AdamW state out of 32 GB. If a bigger model ever OOMs
# here, cut the slices on access instead of up front: send_weights visits one key
# at a time, so only one slice would be live.
```

I have tested a solution locally, where the weights are sliced lazily, but I am not sure if the solution belongs in this PR.

cc @ichovpanTT

jmalone-tt (approved):

PR #52819, opened 2026-08-11
[tt-train] Load fused QKV and gate/up weights via a declarative rule table

imichalakTT:
Part of #52951.

**Behaviour changes.** Inputs `main` tolerated that now fail, and one detection change:

- The loader no longer crops a checkpoint larger than the parameter. Both LoRA examples sized the model from `tokenizer.vocab_size` (128000 for Llama-3) against 128256-row checkpoints; on `main` the load silently cropped 256 rows and BOS id 128000 fell off the end. They now take the architecture, `vocab_size` included, from the checkpoint's `config.json` and only check that the tokenizer fits.
- A transposed `[in, out]` linear weight is rejected instead of transposed.
- A tied checkpoint on a `weight_tying=Disabled` model raises, with a hint, instead of warning that `fc` was not loaded. The reverse, an untied checkpoint on a tied model, raises too.
- A tensor named twice across shards, or one that is not 1-D or 2-D, raises.
- Tensor parallelism is detected from the mesh's `tp` axis and each parameter's placement, not from `config.use_tp`. A parameter sharded over any other axis raises.
- `Sharding.from_tensor` raises rather than returning `None` placements.

## Why

The PR below this one fuses QKV and gate/up into single parameters. The old loader could not express that. It walked the checkpoint tensor by tensor and pushed each one into a parameter, so there was no place for a parameter fed by three tensors. K and V already needed a nested `try_combine_kv` that stashed one half until the other arrived. QKV and gate/up would each have needed the same treatment: staging, shape checks, TP placement, error paths. That is close to as much code again as the loader contains today.

So it was rewritten to pull instead of push.

## What

The driver walks the model's parameters and pulls the checkpoint tensors each one needs. One rule describes one parameter:

```python
_Rule(
    f"{param}/attention/qkv_linear/weight",
    tuple(f"{hf}.self_attn.{p}_proj.weight" for p in ("q", "k", "v")),
    transform=rope_unpermute,
)
```

A fused parameter is a rule with more than one source. The sources stay separate blocks as far as `_assemble`, which interleaves them per device so every rank gets a slice of Q, K and V. Concatenating any earlier would hand rank 0 part of Q and nothing else.

Pulling rather than pushing also buys two checks:

- Every parameter has to be claimed by some rule, verified at load time. A renamed or newly fused module fails the load instead of silently staying at its initial values.
- Placement is read from the destination parameter, so the loader never guesses how a weight is sharded.
- The checkpoint's vocab must equal the config's; only the model's tile padding may grow it. A short checkpoint fails instead of loading noise where its missing tokens should be.

The loader drops from 455 to 349 lines while gaining both fusions. `load_from_safetensors` goes from about 330 lines to about 50.

## Tests

A new suite checks rule coverage against the real model and loads a synthetic HF checkpoint, in float32 and in bf16, end to end. Each parameter is gathered back off the mesh through `Sharding.gather` after asserting its placement, and compared bit-exactly.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:imichalak/llama-sp/4-safetensors-loader)

### DIFF
--- a/tt-train/sources/examples/lora_llama/train_lora_llama.py
+++ b/tt-train/sources/examples/lora_llama/train_lora_llama.py
@@ -235,12 +235,12 @@ def main():
             print(f"Weights downloaded to: {pretrained_path}")
 
         hf_tokenizer = AutoTokenizer.from_pretrained(src)
-        vocab_size = (hf_tokenizer.vocab_size + 31) // 32 * 32
+        vocab_size = len(hf_tokenizer)  # base vocab + added tokens: the checkpoint's row count
         ids = np.array(hf_tokenizer.encode(text), dtype=np.uint32)
-        print(f"Using HF BPE tokenizer, vocab_size={hf_tokenizer.vocab_size} (padded to {vocab_size})")
+        print(f"Using HF BPE tokenizer, vocab_size={vocab_size}")
     else:
         tokenizer = CharTokenizer(text)
-        vocab_size = (tokenizer.vocab_size + 31) // 32 * 32
+        vocab_size = tokenizer.vocab_size
         ids = np.array(tokenizer.encode(text), dtype=np.uint32)
 
     n_train = int(len(ids) * 0.9)
--- a/tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
+++ b/tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
@@ -4,14 +4,20 @@
 
 """Load HuggingFace Llama safetensors weights into a Python Llama model.
 
-Port of the C++ loader in tt-train/sources/ttml/models/llama.cpp.
+:func:`_rules` names the checkpoint tensors feeding each parameter; the driver walks parameters
+and pulls, so a fused parameter fetches its sources together and nothing needs staging.
+A parameter with no rule is an error.
+
+Placement is read from the destination parameter, and fused block sizes from the source tensors,
+so the only thing stated twice anywhere is the block *order*.
 """
 
 from __future__ import annotations
 
 import os
+from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict
+from typing import Callable, Iterator, Sequence
 
 import ml_dtypes
 import numpy as np
@@ -19,7 +25,55 @@ import numpy as np
 import ttnn
 import ttml
 
+from .. import WeightTyingType
 from . import LlamaConfig
+
+# TTML stores a weight as 4-D (1, 1, out_features, in_features).
+ROW_DIM, COL_DIM = 2, 3
+
+# Noise rather than zeros avoids dead neurons; the fixed seed keeps two loads identical.
+_PAD_SEED = 0
+_PAD_STDDEV = 0.02
+
+# Collapsed on read so the rules name each tensor once.
+_NAME_ALIASES = {
+    "wte.weight": "embed_tokens.weight",
+    "transformer.wte.weight": "embed_tokens.weight",
+}
+
+# Shipped by some checkpoints; TTML derives these at runtime.
+_NOT_WEIGHTS = ("rotary_emb.inv_freq",)
+
+# Tying makes these one parameter, so the model keeps whichever name it walks first.
+_TIED_NAMES = ("Llama/fc/weight", "Llama/tok_emb/weight")
+
+
+def _canonical(name: str) -> str:
+    name = name.removeprefix("model.")
+    return _NAME_ALIASES.get(name, name)
+
+
+def _read_checkpoint(directory: str | os.PathLike) -> dict[str, np.ndarray]:
+    """Every tensor in *directory*, keyed by canonical name and shaped 2-D as ``[out, in]``."""
+    from safetensors.numpy import load_file
+
+    files = sorted(Path(directory).glob("*.safetensors"))
+    if not files:
+        raise FileNotFoundError(f"No .safetensors files found in {Path(directory)}")
+
+    tensors: dict[str, np.ndarray] = {}
+    for path in files:
+        print(f"Loading safetensors file: {path}")
+        for name, array in load_file(str(path)).items():
+            if array.ndim == 1:  # norm gammas; a 1-row weight downstream
+                array = array.reshape(1, -1)
+            elif array.ndim != 2:
+                raise RuntimeError(f"{name}: expected a 1-D or 2-D tensor, got shape {array.shape}")
+            canonical = _canonical(name)
+            if canonical in tensors:
+                raise RuntimeError(f"{name}: collides with another tensor already read as {canonical}")
+            tensors[canonical] = array
+    return tensors
 
 
 def _unpermute_proj_rows(w: np.ndarray, n_heads: int) -> np.ndarray:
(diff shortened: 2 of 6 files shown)


tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -2,123 +2,297 @@
...
 
 import ttnn
 import ttml
+from ttml.common.utils import resolve_padded_load_shape
 
+from .. import WeightTyingType
 from . import LlamaConfig
 
+if TYPE_CHECKING:
+    from safetensors import safe_open
```
awliu-TT, line 26: Is there a reason why this is gated behind TYPE_CHECKING? It's also imported in the `_Checkpoint` constructor.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -2,123 +2,297 @@
...
+def _check_coverage(parameter_names: set[str], rules: Sequence[_Rule], checkpoint_names: frozenset[str]) -> None:
+    """Every parameter must be fed by a rule or be a bias the checkpoint does not carry, and every
+    rule must land. A renamed or newly fused module shows up here instead of as a quietly untrained
+    weight; so does a checkpoint that ships the biases the model was built with.
+    """
+    targets = {rule.param for rule in rules}
+    checkpoint_has_biases = any(name.endswith(".bias") for name in checkpoint_names)
+    exempt = set() if checkpoint_has_biases else _biases(parameter_names)
+    uncovered = sorted(parameter_names - targets - exempt)
+    unknown = sorted(targets - parameter_names)
```
awliu-TT, line 271: `targets` is a set. If two targets cover one param, we wouldn't catch it here. Not a problem today, since we have each rule mapping to a unique param name, but worth noting.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -128,328 +302,48 @@ def load_from_safetensors(
...
-        print(f"Note: {len(ignored_hf)} HF tensors were ignored (no mapping):")
-        for n in sorted(ignored_hf):
-            print(f"  - {n}")
+        leftover = checkpoint.names - {source for rule in rules for source in rule.sources}
+
+    print(f"Loaded {len(rules)} parameters from {len(checkpoint.names) - len(leftover)} checkpoint tensors.")
+    if biases := _biases(parameter_names):
+        print(f"Left at initial values: {len(biases)} biases the checkpoint does not carry.")
+    if unused := sorted(n for n in leftover if not n.endswith(_NOT_WEIGHTS)):
+        print(f"Note: {len(unused)} checkpoint tensors were not used:")
```
awliu-TT, line 347: This would let an untied checkpoint through if we were loading it on a weight-tied model. We should error on this or at least have a warning like at the end of `_check_sources`.

tt-train/sources/ttml/ttml/sharding.py
```diff
@@ -20,40 +20,39 @@ def _mesh_device():
...
-        try:
-            # NATIVE: read topology without coercing precision (avoids a float32/bf16 typecast + cache).
-            topology = tensor.get_value(ttml.autograd.PreferredPrecision.NATIVE).tensor_topology()
-            placements = list(topology.placements())
-            dist_shape = list(topology.distribution_shape())
-        except Exception:
-            placements, dist_shape = None, None  # no topology (unit mesh / older ttnn build)
-        return cls(placements, dist_shape)
+        """Read off the live topology; a unit mesh reads back as 1-D [Replicate], so this raises only
+        for a tensor with no topology to read, which nothing downstream can place correctly anyway."""
```
awliu-TT, line 30: The `from_tensor` docstring names the wrong failure mode. It says this "raises only for a tensor with no topology to read." A tensor with truly no topology is an uninitialized `ttnn::Tensor()` — which `get_value(NATIVE)` will hand back happily, since `autocast_tensor.cpp:53` returns the native slot without an `is_tensor_initialized` check — and `tensor_attributes` is then null, so `tensor.cpp:524` segfaults rather than raising. The old except Exception never covered that either, so nothing is lost; the rationale just doesn't describe a reachable raise.

tt-train/sources/ttml/ttml/sharding.py
```diff
@@ -20,40 +20,39 @@ def _mesh_device():
...
+        return cls(list(topology.placements()), list(topology.distribution_shape()))
 
     @property
-    def placements(self) -> list | None:
-        """Per-mesh-axis ttnn placements (``PlacementShard`` / ``PlacementReplicate``), or None on a unit mesh."""
+    def placements(self) -> list:
+        """Per-mesh-axis ttnn placements (``PlacementShard`` / ``PlacementReplicate``).
+
+        A fully replicated tensor flattens to a single ``Replicate``, so this can be shorter
+        than the mesh rank: never index it by mesh axis.
```
awliu-TT, line 39: The `placements` docstring overstates the flattening. "A fully replicated tensor flattens to a single Replicate" holds for the default topology and replicate_tensor_to_mesh_mapper (`distributed_tensor.cpp:683-689`), but not for tensors built by derive_mapper itself or by ttml.mesh().axis_mapper, which keep full mesh rank with Replicate on unsharded axes (`distributed_tensor.cpp:706-710`). "may flatten" would be accurate. This matters because the consumers that legitimately index by mesh axis (`safetensors_loader.py:184`), the embedding checks, (`test_llama_safetensors_loader.py:420`) are correct only because mapper-created placements are full-rank — the absolute phrasing makes that read as always-wrong.

tt-train/sources/ttml/ttml/sharding.py
```diff
@@ -20,40 +20,39 @@ def _mesh_device():
...
 
-    def __init__(self, placements: list | None, dist_shape: list[int] | None) -> None:
+    def __init__(self, placements: list, dist_shape: list[int]) -> None:
         self._placements = placements
         self._dist_shape = dist_shape
 
     @classmethod
     def from_tensor(cls, tensor: ttml.autograd.Tensor) -> Sharding:
-        try:
-            # NATIVE: read topology without coercing precision (avoids a float32/bf16 typecast + cache).
```
awliu-TT, line 30: Keep this comment.

tt-train/tests/python/test_llama_safetensors_loader.py
```diff
@@ -0,0 +1,628 @@
...
+        model.eval()
+
+        ids = np.random.default_rng(7).integers(0, VOCAB, (1, 1, 1, SEQ_LEN)).astype(np.uint32)
+        mask = np.tril(np.ones((1, 1, SEQ_LEN, SEQ_LEN), dtype=np.float32))
+        logits = model(
+            ttml.autograd.Tensor.from_numpy(ids, ttnn.Layout.ROW_MAJOR, ttnn.DataType.UINT32),
+            ttml.autograd.Tensor.from_numpy(mask, ttnn.Layout.TILE, ttnn.DataType.BFLOAT16),
+        )
+
+        values = gather(logits, COL_DIM)
```
awliu-TT, line 622: `gather(logits, COL_DIM)` requires `Sharding.from_tensor(logits).placements[tp]` to be `Shard(dim=3)`, but the LM-head logits cannot carry that placement. ttnn stamps every new-infra op output with placements derived from its inputs: `launch` calls `update_output_tensor_topologies` (`ttnn/api/ttnn/mesh_device_operation_utils.hpp:126`), which falls back to `compute_output_placements_and_shape` (`ttnn/core/device_operation_detail.cpp:45`) because `MatmulDeviceOperation` defines no `compute_output_topologies`. That function copies each sharded input's `Shard` placement per distribution axis. The only sharded input to the `fc` matmul is its weight, `[Replicate, Shard(2)]`, so the logits report `Shard(dim=2)` and the assert at lines 420-423 fails before the finiteness and variance checks run. Even without the assert, `Sharding.gather` would build a composer that concatenates on dim 2.

Suggest gathering the logits with an explicit composer, e.g. `ttml.core.distributed.concat_mesh_to_tensor_composer(device, 3)` as `test_qwen3_tied_tp_integration.py:188` does, or an `all_gather` on dim 3, and keeping `Sharding.from_tensor`-based placement asserts for mapper-created parameters only. If you have a device run where this test passes, please paste it and I will withdraw this.

tt-train/tests/python/test_llama_safetensors_loader.py
```diff
@@ -0,0 +1,628 @@
...
+
+
+def read_param(params, name: str, shard_dim: int | None) -> np.ndarray:
+    """A parameter as a 2-D ``[out_features, in_features]`` array."""
+    full = gather(params[name], shard_dim)
+    return full.reshape(full.shape[-2], full.shape[-1])
+
+
+@pytest.mark.requires_device
+@pytest.mark.usefixtures("tp_mesh")
```
awliu-TT, line 434: All 18 device tests in this file are gated on `tp_mesh`, which `pytest.skip`s when a `[1, 2]` mesh cannot open. The `tt-train python unit tests` job (`tests/pipeline_reorg/train_unit_tests.yaml:39`) runs `pytest tt-train/tests/python` on `bh_p100` and `bh_p150b_civ2` only, both single-chip, so every device test here skips in CI and the job stays green. The device-free tests never reach the driver: swapping the K and V sources, dropping the unpermute, swapping gate and up, or deleting `_check_sources` all still give 45 passed, 18 skipped. Six other files share this gating, but for this PR it means the load path has zero CI coverage.

`TestConsumerContract`, the non-TP cases of `TestCoverageAgainstRealModels`, `test_without_tensor_parallelism_is_a_plain_concat` and `test_rejects_a_checkpoint_missing_a_fused_source` do not need TP. Running them under a single-device fixture would give CI a real end-to-end load.

tt-train/tests/python/test_llama_safetensors_loader.py
```diff
@@ -0,0 +1,628 @@
...
+            assert np.array_equal(got, as_bf16(hf[f"{pfx}.self_attn.o_proj.weight"])), f"layer {layer}: o_proj"
+            got = read_param(params, f"Llama/blocks/{layer}/mlp/w2/weight", COL_DIM)
+            assert np.array_equal(got, as_bf16(hf[f"{pfx}.mlp.down_proj.weight"])), f"layer {layer}: down_proj"
+
+        embedding_shard = {
+            EmbeddingPlacement.Replicated: None,
+            EmbeddingPlacement.VocabParallel: ROW_DIM,
+            EmbeddingPlacement.FeatureParallel: COL_DIM,
+        }[placement]
+        got = read_param(params, "Llama/tok_emb/weight", embedding_shard)
```
awliu-TT, line 539: `read_param` is called for qkv, w_gate_up, out_linear, w2 and tok_emb only (lines 521-539 and 580-595). `Llama/fc/weight` and the three gamma kinds are never compared with their sources, and those are exactly the parameters whose sources are same-shaped and swappable: swapping `input_layernorm.weight` and `post_attention_layernorm.weight` in `_rules`, or feeding `ln_fc/gamma` from a block norm, passes coverage, `_check_sources`, `_fit`, the clean-load report, this test and the forward test.

A loop over `list(_rules(config, names))` comparing each rule's assembled sources against `read_param` would close this and also cover the tile-padded LM head. The PR body's "each parameter is gathered back ... and compared bit-exactly" would then hold.

tt-train/tests/python/test_llama_safetensors_loader.py
```diff
@@ -0,0 +1,628 @@
...
+        got, half = _unpermute_proj_rows(w, n_heads), per_head // 2
+        for head in range(n_heads):
+            base = head * per_head
+            for i in range(half):
+                assert np.array_equal(got[base + 2 * i], w[base + i])
+                assert np.array_equal(got[base + 2 * i + 1], w[base + half + i])
+
+
+def save_checkpoint(directory, tensors: dict[str, np.ndarray]) -> None:
+    pytest.importorskip("safetensors.numpy").save_file(tensors, str(directory / "model.safetensors"))
```
awliu-TT, line 236: Every test writes a single `model.safetensors`, so `_Checkpoint`'s multi-file path (shared `ExitStack`, per-file `(file, raw)` indirection, cross-file collision) is exercised against one file only, while the real 8B target ships four. I ran the index and source checks by hand against the cached Llama-3.1-8B shards and they are correct; a two-file case here would keep them that way.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -2,123 +2,297 @@
...
+    uncovered = sorted(parameter_names - targets - exempt)
+    unknown = sorted(targets - parameter_names)
+    if not uncovered and not unknown:
+        return
+
+    detail = "".join(f"\n  no rule feeds       {name}" for name in uncovered)
+    detail += "".join(f"\n  no such parameter   {name}" for name in unknown)
+    raise RuntimeError(
+        f"the loader and this Llama disagree about its parameters:{detail}\n"
+        f"Update _rules() in {Path(__file__).name} to match the model; a weight_tying or attention_bias "
```
awliu-TT, line 279: This message says an `attention_bias` mismatch between the config and the model "also lands here", but nothing in the loader reads `config.attention_bias`, and `_biases` exempts every `/bias` parameter unless the checkpoint ships a `.bias`. Both mismatch directions pass silently: `LlamaConfig(attention_bias=True)` with a bias-free model, and `attention_bias=False` with a biased model, both return from `_check_coverage` without error (checked device-free at this commit). Only `weight_tying` actually lands here.

Drop the clause, or make `_rules`/`_check_coverage` cross-check `config.attention_bias` against the presence of `/bias` parameters.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -2,123 +2,297 @@
...
+    return {name for name in parameter_names if name.endswith("/bias")}
+
+
+def _check_coverage(parameter_names: set[str], rules: Sequence[_Rule], checkpoint_names: frozenset[str]) -> None:
+    """Every parameter must be fed by a rule or be a bias the checkpoint does not carry, and every
+    rule must land. A renamed or newly fused module shows up here instead of as a quietly untrained
+    weight; so does a checkpoint that ships the biases the model was built with.
+    """
+    targets = {rule.param for rule in rules}
+    checkpoint_has_biases = any(name.endswith(".bias") for name in checkpoint_names)
```
awliu-TT, line 268: The exemption is withdrawn by any `.bias` name anywhere in the checkpoint, not by the biases that would feed this model. A checkpoint with all required weights plus an unrelated `lm_head.bias` makes a model built with `attention_bias=True` fail with "no rule feeds .../qkv_linear/bias ... Update _rules()", and that hint is a dead end because no rule can feed a bias. Keying the withdrawal on the specific sources (`layers.{i}.self_attn.{q,k,v}_proj.bias`, `o_proj.bias`) would make the check do what the `_biases` docstring says. Low priority since HF Llama ships no biases.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -2,123 +2,297 @@
...
-      - sharded on dim 2 (rows / output features) -> ``"col_w"``
-      - sharded on dim 3 (cols / hidden features)  -> ``"row_w"``
-      - replicated, or no 'tp' axis                -> ``None``
-    """
-    placements = ttml.Sharding.from_tensor(param).placements
-    if placements is None:  # unit mesh / no topology
+def _sharded_dim(param, tp: _TpAxis | None, subject: str) -> int | None:
+    """Which tensor dim *param* shards over 'tp', or ``None`` if replicated."""
+    if tp is None:
         return None
```
awliu-TT, line 179: The "sharded over a mesh axis other than tp" check at lines 181-183 only runs when the mesh has a `tp` axis. On a DP-only mesh (`--mesh_shape 1,8 --dp_axis 1` opens 8 devices with no `tp`), a parameter sharded over `dp` or by `fully_shard` gets `shard_dim=None`, `mesh_size=1`, and `param.assign` receives a full-size replicated tensor with `mapper=None`, which is the case the Raises list at line 314 says raises. Either iterate placements whenever a mesh is open and raise on any `PlacementShard`, or reword that bullet. The `_tp_axis` docstring at line 169 also calls the no-`tp` case "the single-device case", which the DP-only mesh contradicts.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -128,328 +302,48 @@ def load_from_safetensors(
...
-    - Embedding padding when model vocab_size differs from HF
-    - Shape conversion from HF 2D to TTML 4D
+    *safetensors_path* is a directory of ``.safetensors`` files holding one whole model in HF's
+    canonical form. Every check that needs only names runs before the first tensor is read.
 
-    Args:
-        model: A Python Llama model instance.
-        safetensors_path: Path to directory containing .safetensors file(s).
-        config: The LlamaConfig used to build the model.
+    Raises:
```
awliu-TT, line 308: Beyond the two documented changes, the rewrite turns several previously tolerated inputs into hard errors, and none appear in the PR body or this Raises list:

- transposed `[in, out]` linear weights: the old loader accepted `c == tr and r == tc` and transposed; `_fit` now raises "does not match the config-implied shape";
- a tied checkpoint on a `weight_tying=Disabled` config: the old loader loaded `tok_emb` and warned that `fc` was not loaded; `_check_sources` now raises with the hint (verified against the cached Llama-3.2-1B-Instruct, and the hint is correct);
- duplicate canonical names across files (line 74) and tensors that are not 1-D or 2-D (line 93);
- TP is now detected from the mesh having a `tp` axis rather than from `config.use_tp`.

`FileNotFoundError` from line 63 is also missing from the list. Worth adding to the PR description and here.

tt-train/sources/examples/lora_llama/train_lora_llama.py
```diff
@@ -235,12 +235,12 @@ def main():
...
         hf_tokenizer = AutoTokenizer.from_pretrained(src)
-        vocab_size = (hf_tokenizer.vocab_size + 31) // 32 * 32
+        vocab_size = len(hf_tokenizer)  # base vocab + added tokens: the checkpoint's row count
         ids = np.array(hf_tokenizer.encode(text), dtype=np.uint32)
-        print(f"Using HF BPE tokenizer, vocab_size={hf_tokenizer.vocab_size} (padded to {vocab_size})")
+        print(f"Using HF BPE tokenizer, vocab_size={vocab_size}")
     else:
         tokenizer = CharTokenizer(text)
-        vocab_size = (tokenizer.vocab_size + 31) // 32 * 32
+        vocab_size = tokenizer.vocab_size
```
awliu-TT, line 243: This drops the round-up-to-32 on the CharTokenizer path, which is independent of the loader change and not mentioned in the PR. With the Shakespeare text the vocab is 65 or 66, so the default `train_lora_llama.py` run (no `--pretrained`) now builds a non-TP Llama with `padded_vocab_size != vocab_size` for the first time, and `Llama.forward` routes through `SliceLastDim` plus `cross_entropy_loss` over a partial-tile logits tensor, with the backward re-padding the grad. No test exercises that path (`grep SliceLastDim tt-train/tests` is empty), and it also changes the softmax width, so char-path loss curves are not comparable with `main`.

Keeping the round-up here (and in `train_lora_llama_sft.py:239`) is the safe choice; otherwise it needs a device test and a note in the PR body.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
```diff
@@ -128,328 +302,48 @@ def load_from_safetensors(
...
+            param = parameters[rule.param]
+            blocks = [checkpoint[name] for name in rule.sources]
+            if rule.transform:
+                blocks = rule.transform(*blocks)
+            shard_dim = _sharded_dim(param, tp, rule.param)
+            host = _assemble(blocks, shard_dim, mesh_size, rule.param)
+            host = _fit(host, _global_shape(param, shard_dim, mesh_size), rule.source_shape, rule.param)
+
+            mapper = tp.mesh.axis_mapper("tp", tdim=shard_dim) if tp and shard_dim is not None else None
+            param.assign(
```
awliu-TT, line 336: Optional host-side points at real sizes:

- `blocks` is still alive when `from_numpy` runs, so multi-source parameters peak at three copies (blocks, assembled `host`, the C++ HostBuffer) rather than two. A `del blocks` after `_assemble` saves about 235 MB per layer on 8B `w_gate_up` and about 940 MB on a 70B-class model.
- For fp32 checkpoints every intermediate stays fp32 until `_to_bf16_4d`. Casting right after `checkpoint[name]` halves them and is bit-identical, since the cast commutes with permutation and concatenation.
- `safe_open(...).get_slice(name).get_shape()` gives shapes without reading data, so a `_check_shapes` pass before the loop would let a config mismatch fail whole instead of after `fc`, `tok_emb`, `ln_fc` and earlier layers were already assigned.

awliu-TT (commented):
Reviewed at 461360f. The loader core checks out: I ran the name and shape stages against the cached Llama-3.2-1B-Instruct (tied, one file) and Llama-3.1-8B (untied, four files), and the rule table, source hints and per-rank `[Q|K|V]` / `[gate|up]` layout are correct at TP 1, 2, 4 and 8. `nlp_create_qkv_heads` infers `head_dim` from the fused width over `num_q_heads + 2 * num_kv_heads`, matching the block order. Mapper-created parameters carry full-rank placements, so the axis indexing in `_sharded_dim` is safe. Device-free suite: 45 passed, 18 skipped locally. No tt-train CI ran on this PR.

The inline comments are ordered by importance. The first one, the forward test's placement assert, looks like it cannot pass on a TP mesh; please confirm with a device run either way.

Two items outside the diff:

- `tt-train/sources/ttml/ttml/modules/embedding.py:140` and `:264` still test `placements is None`, which can no longer happen after the `Sharding` change; the length guard is the part that still matters.
- `tt-train/sources/examples/lora_llama/README.md:58` shows `vocab_size: 32000` in the yaml example, which `llama_config_from_yaml` ignores, and the README does not say that `--pretrained` now requires `len(tokenizer)` to equal the checkpoint's embedding rows (the loader errors instead of cropping).

tt-train/tests/python/test_llama_safetensors_loader.py
awliu-TT: This is an existing issue. We will be moving towards multi-device enablement for tt-train CI.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: The idea was avoiding loading hf transformers library (where safetensors comes from) on `import ttml`. But I measured how much this import takes and it turns out to be less than 1 ms, so I don't think there is a reason to keep this. I am simplifying to a single top-level import.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: addressed in https://github.com/tenstorrent/tt-metal/pull/52819/changes/d54236d35c50e3250060311fef9d0941be2c1428.

now if there is more than one rule targeting a param, an error raises.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: Fixed. Weight tying mismatch now errors out: https://github.com/tenstorrent/tt-metal/pull/52819/changes/4b61cf68af53b35711584ac353a035416b643be7.

tt-train/sources/ttml/ttml/sharding.py
imichalakTT: addressed in https://github.com/tenstorrent/tt-metal/pull/52819/commits/1f358ced345

tt-train/sources/ttml/ttml/sharding.py
imichalakTT: updated the docstrings in https://github.com/tenstorrent/tt-metal/pull/52819/commits/1f358ced345.

tt-train/sources/ttml/ttml/sharding.py
imichalakTT: Restored it in https://github.com/tenstorrent/tt-metal/pull/52819/commits/1f358ced345.

tt-train/tests/python/test_llama_safetensors_loader.py
imichalakTT: I switched to an explicit composer. https://github.com/tenstorrent/tt-metal/pull/52819/commits/cac6778d532

tt-train/tests/python/test_llama_safetensors_loader.py
imichalakTT: I split some of the tests into TP and non-TP, without the fixture.
https://github.com/tenstorrent/tt-metal/pull/52819/changes/d31160d66596cfce54a94432149dd5c75f41e877

tt-train/tests/python/test_llama_safetensors_loader.py
imichalakTT: Addressed that here:
https://github.com/tenstorrent/tt-metal/pull/52819/commits/11faa439a29

tt-train/tests/python/test_llama_safetensors_loader.py
imichalakTT: the synthetic checkpoint now ships as two shards with every fused group straddling them
https://github.com/tenstorrent/tt-metal/pull/52819/changes/c7b16bf9e67d0930ee93d38d01e11218cf4649ab

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: Dropped the clause in https://github.com/tenstorrent/tt-metal/pull/52819/commits/39c9d259a25.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: fixed here: https://github.com/tenstorrent/tt-metal/pull/52819/changes/0fb3efb21d427142a5fe4d5deabfe911df273169

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: Now the helper raises on any PlacementShard other than TP, even with no axis.
https://github.com/tenstorrent/tt-metal/pull/52819/commits/98522d530a3

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: I updated the docstring with a more up-to-date list, and mentioned all of the behavior changes in the PR body. https://github.com/tenstorrent/tt-metal/pull/52819/changes/996dc2d4260fff6a5e6877899b1ed48d9ec7e386

tt-train/sources/examples/lora_llama/train_lora_llama.py
imichalakTT: Restored in https://github.com/tenstorrent/tt-metal/pull/52819/commits/7ee6b1f72f9, in both examples.

tt-train/sources/ttml/ttml/models/llama/safetensors_loader.py
imichalakTT: Addressed first two of your points in https://github.com/tenstorrent/tt-metal/pull/52819/changes/ec816453826dc85e0dc6e4e33e4e0f43d8149457.

abustamanteTT (approved):

jmalone-tt (commented):
Claude turned up another issue:
`tt-train/sources/ttml/ttml/models/llama/__init__.py:185`: `LlamaConfig.from_hf `breaks tied checkpoints when TP is larger than 1

- What happens: when a checkpoint has `tie_word_embeddings: true`, `from_hf` turns on `weight_tying=Enabled`. It never sets `embedding_placement`, so that stays at its default, `Replicated`. With `use_tp` set, `__post_init__` then rejects the config, because tied weights need `VocabParallel`.
- Who hits it: the LoRA scripts, at `examples/lora_llama/train_lora_llama.py:294` and `train_lora_llama_sft.py:298`. Running them with `--pretrained meta-llama/Llama-3.2-1B-Instruct` on a mesh with tp > 1 now fails with a `ValueError` before the model is built.

Do you think it makes sense to have the LoRA scripts pass that placement themselves?

PR #36659, opened 2026-01-28
[tt-train] ring attention

dyurshevichTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/35827)

### Problem description
Added composite trainable ring attention (only causal mask is supported now) with context parallel. All single and double parallelism modes are supported from dp, tp, cp for Llama families only.

### Some numbers

Here is the table of training step times and peak memory consumption across different parallelism configurations, model sizes and other hyperparameters.
[distributed_config_comparison.xlsx](https://github.com/user-attachments/files/25492899/distributed_config_comparison.xlsx)

### Checklist

- [x] [All post-commit tests](https://github.com/tenstorrent/tt-metal/actions/runs/22313488715/job/64553005571)
- [ ] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:{{branch_name}})
- [ ] [cpp-unit-tests](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:{{branch_name}})
- [ ] New/Existing tests provide coverage for changes

#### Model tests

If your changes cover model-related code, you should run tests corresponding to affected models and platforms (Single card, T3K, Galaxy). "Choose your pipeline" workflows facilitate running multiple kinds of tests in a single run. Each offers `models-mandatory` and `models-extended` presets.
The former includes a minimal set of tests, to be run always. The latter extends that with additional ones - use your best judgement in deciding which is the most appropriate for your PR.

- [ ] [(Single) Choose your pipeline](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:{{branch_name}})
  - [ ] `models-mandatory` preset (runs: [Device perf regressions](https://github.com/tenstorrent/tt-metal/actions/workflows/perf-device-models.yaml) and [Frequent model and ttnn tests](https://github.com/tenstorrent/tt-metal/actions/workflows/fast-dispatch-full-regressions-and-models.yaml))
  - [ ] `models-extended` preset (runs: the mandatory tests, plus [Demo](https://github.com/tenstorrent/tt-metal/actions/workflows/single-card-demo-tests.yaml) and [Model perf](https://github.com/tenstorrent/tt-metal/actions/workflows/perf-models.yaml) tests)
  - [ ] other selection - specify runs

[...]

PR #52820, opened 2026-08-11
[tt-train] Add Megatron sequence parallelism

imichalakTT:
Part of #52951.

Adds Megatron sequence parallelism to the Python Llama model as `TPStrategy.TENSOR_SEQUENCE`, enabled with `enable_sp: true` (requires `enable_tp`). Between the tensor-parallel matmuls the residual stream is sharded along the sequence instead of replicated: the column-parallel input becomes an all-gather over the sequence, the row-parallel output a reduce-scatter, and the norms, dropout and residual adds run on `S / tp` tokens per rank. Same bytes on the fabric, `tp` times less activation memory in that region.

SP also puts a collective directly on both sides of every matmul, an all-gather before the column-parallel one and a reduce-scatter after the row-parallel one, so both can be fused into matmul+CCL kernels; classic TP has one only on the output side.

- `TPStrategy` (`NONE` / `TENSOR` / `TENSOR_SEQUENCE`) replaces `use_tp` on `LlamaConfig`, the Llama modules and the train-script builders.
- `ColumnParallelLinear` and `RowParallelLinear` take `sequence_parallel`. LoRA reuses their collectives and works unchanged.
- `VocabParallelEmbedding` reduce-scatters into the sharded layout; the other placements emit the full sequence and the model slices it locally.
- Parameters on the sharded stream (RMSNorm gains, row-parallel bias, `lora_B`) get partial gradients. They are flagged with `mark_sequence_parallel`, and `ttml.sync_gradients` averages over `dp`/`fsdp` then sums the flagged ones over `tp`.
- `scatter` is now a local `mesh_partition`; it was a reduce-scatter plus `1/n` on a replicated input.

Llama only. Sequence length must be divisible by `32 * tp`; the KV-cache path is refused.

## Tests

`test_sequence_parallel.py` compares SP against TP built from one seed: logits and synced gradients for all three embedding placements, two AdamW steps (the optimizer rewrites parameter topology in place, which broke an earlier topology-based sync), LoRA, and the exact set of flagged parameters. At tp=2 logits match bitwise and gradients within 1 ULP.

## Benchmarks

Llama 3 8B, `tp=4` on 2×p300 (1×4 ring), seq 2048, bs 1 × 4 grad-accum, AdamW 3e-5, seed 5489, no grad checkpointing, replicated embeddings. `training_llama8b_tp4.yaml` vs `training_llama8b_tp4_sp.yaml`, 15 steps.

| Metric (steady state) | `tp=4` | `tp=4` + SP | Δ |
| --- | --: | --: | --: |
| Step time | 2454.2 ± 6.5 ms | 2321.8 ± 7.3 ms | **−5.4 %** |
| Throughput | 3338 tok/s | 3528 tok/s | **+5.7 %** |
| Achieved TFLOPS | 161.2 | 170.3 | +5.6 % |
| MFU (of 594 TFLOPS bf16) | 27.12 % | 28.67 % | **+1.55 pp** |
| Peak DRAM per chip | 25 611 MB | 24 333 MB | **−1 278 MB (−5.0 %)** |
| Average loss (15 steps) | 9.876042 | 9.862500 | −0.0135 |

`MemoryUsageTracker` over the first step, per chip. Weights, gradients and optimizer state are identical; only activations move.

| DRAM snapshot | `tp=4` | `tp=4` + SP | Δ |
| --- | --: | --: | --: |
| `MODEL_CREATION` | 4606.18 MB | 4606.18 MB | identical |
| `OPTIMIZER_CREATION` | 9193.53 MB | 9193.53 MB | identical |
| Resident, steady | 18 390.11 MB | 18 390.11 MB | identical |
| Forward-pass segment peak | 6738.32 MB | 5460.23 MB | **−19.0 %** |
| Backward-pass allocations | 18 781.05 MB | 13 333.05 MB | **−29.0 %** |
| Cumulative peak | 25 611.37 MB | 24 333.28 MB | **−5.0 %** |

The reduction is smaller than the sharded region predicts. Every collective strands its full-size input until the autograd graph unwinds (#52851), about 1 GB per chip here, and SP has the same number of them.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:imichalak/llama-sp/5-seq-parallelism)

### DIFF
--- a/tt-train/sources/examples/lora_llama/train_lora_llama.py
+++ b/tt-train/sources/examples/lora_llama/train_lora_llama.py
@@ -30,6 +30,7 @@ from ttml.models.llama import (
     LlamaRopeScalingConfig,
     load_from_safetensors,
 )
+from ttml.parallel import TPStrategy
 from ttml.modules import LoraConfig, LoraModel
 
 MemoryUsageTracker = ttml.core.utils.MemoryUsageTracker
@@ -87,7 +88,7 @@ def llama_config_from_yaml(yaml_config: dict, vocab_size: int, use_tp: bool = Fa
         runner_type=runner_type,
         weight_tying=weight_tying,
         rope_scaling=rope_scaling,
-        use_tp=use_tp,
+        tp_strategy=TPStrategy.from_flags(use_tp),
     )
 
 
@@ -292,7 +293,7 @@ def main():
             vocab_size=vocab_size,
             max_position_embeddings=256,
             rope_theta=500000.0,
-            use_tp=use_tp,
+            tp_strategy=TPStrategy.from_flags(use_tp),
         )
 
     seq_len = llama_cfg.max_position_embeddings
--- a/tt-train/sources/examples/train/model_builders.py
+++ b/tt-train/sources/examples/train/model_builders.py
@@ -14,6 +14,7 @@ from ttml.common.utils import round_up_to_tile
 from ttml.models.deepseek import DeepSeek, DeepSeekConfig
 from ttml.models.deepseek.flops import calculate_flops_per_token as _deepseek_flops
 from ttml.models.llama import Llama, LlamaConfig, LlamaRopeScalingConfig
+from ttml.parallel import TPStrategy
 from ttml.models.llama.flops import calculate_flops_per_token as _llama_flops
 from ttml.models.nanogpt import NanoGPT, NanoGPTConfig, NanoGPTExperimentalConfig, create_nanogpt
 from ttml.models.nanogpt.flops import calculate_flops_per_token as _gpt2_flops
@@ -142,8 +143,8 @@ def _parse_gpt2(tc: dict) -> GPT2Spec:
     return spec
 
 
-def _build_gpt2(cfg: ModelConfig, use_tp: bool) -> Model:
-    if use_tp:
+def _build_gpt2(cfg: ModelConfig, tp_strategy: TPStrategy) -> Model:
+    if tp_strategy.tensor_parallel:
         raise ValueError("model_type=gpt2 has no TP path; use model_type=llama for DP+TP")
     assert isinstance(cfg.spec, GPT2Spec)
     spec = cfg.spec
@@ -181,7 +182,7 @@ def _parse_llama(tc: dict) -> LlamaSpec:
     return spec
 
 
-def _build_llama(cfg: ModelConfig, use_tp: bool) -> Model:
+def _build_llama(cfg: ModelConfig, tp_strategy: TPStrategy) -> Model:
     assert isinstance(cfg.spec, LlamaSpec)
     spec = cfg.spec
     if spec.num_groups <= 0:
@@ -208,7 +209,7 @@ def _build_llama(cfg: ModelConfig, use_tp: bool) -> Model:
                 low_freq_factor=spec.low_freq_factor,
                 original_context_length=spec.original_context_length,
             ),
-            use_tp=use_tp,
+            tp_strategy=tp_strategy,
             embedding_placement=spec.embedding_placement,
         )
     )
@@ -244,10 +245,11 @@ def _parse_deepseek(tc: dict) -> DeepSeekSpec:
 _MOE_TYPES = ("dense", "sparse_ep")
 
 
-def _build_deepseek(cfg: ModelConfig, use_tp: bool) -> Model:
+def _build_deepseek(cfg: ModelConfig, tp_strategy: TPStrategy) -> Model:
     # DeepSeek integrates with the named-mesh TP path (MLA, dense MLP, LM head, and — under
     # full-model TP — sparse MoE all shard across the "tp" axis). moe_type selects the MoE FFN
     # variant; sparse_ep additionally partitions experts across a mesh axis.
+    use_tp = tp_strategy.tensor_parallel
     assert isinstance(cfg.spec, DeepSeekSpec)
     spec = cfg.spec
     if spec.moe_type not in _MOE_TYPES:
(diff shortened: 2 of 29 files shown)


tt-train/sources/ttml/ttml/_mesh.py
```diff
@@ -288,3 +288,56 @@ def _param_is_fsdp_sharded(param, axis_index: int) -> bool:
...
+    """Sum the gradients of TP-replicated parameters across the tensor-parallel axis.
+
+    Under Megatron sequence parallelism the residual stream is sharded along the
+    sequence across ``axis_name`` (the TP axis), so a parameter that is *replicated*
+    on that axis -- the RMSNorm ``gamma`` weights, and any bias added in a
+    sequence-sharded region -- only accumulates the gradient from its rank's slice of
+    the sequence. The full gradient is the SUM over ranks, so we all-reduce with **no
+    averaging** (unlike :func:`sync_gradients`, which means-reduces over the data
+    axes). TP-*sharded* parameters are skipped: each rank already holds the correct grad
+    for its shard.
```
dyurshevichTT: Can we have a single sync_gradients which will do all_reduce along the axis based on sharding of gradients and weights? If we fixed the tensor layout propagation for all ops, we need to all_reduce along the axis iff weights are replicated and gradients for those weights are sharded right?

tt-train/sources/ttml/ttml/modules/embedding.py
```diff
@@ -294,9 +316,14 @@ def forward(self, x: ttml.autograd.Tensor) -> ttml.autograd.Tensor:
...
+        emb = ttml.ops.distributed.all_gather(
             emb,
             3,
             self.cluster_axis,
             ttml.ops.distributed.GradOutputType.REPLICATED,
         )
+
+        if self.sequence_parallel:
+            emb = ttml.ops.distributed.scatter(emb, 2, self.cluster_axis)
+
```
dyurshevichTT: If this scatter a slice or reduce_scatter*tp_size? If latter, can you add a scatter version which will call slice with different compile time parameters on each device using create_mesh_workload like here https://github.com/tenstorrent/tt-metal/pull/36659/files#diff-d5b17773ee998b41b02337cd4792913edf5ad5d1c1ae861a7d7389433ef2d7a7?

imichalakTT: Used to be reduce_scatter. I updated the implementation of scatter in this PR to use ttnn::mesh_partition, which removes the unnecessary overhead.

tt-train/sources/ttml/ttml/_mesh.py
imichalakTT: Topology propagation support is still inconsistent across the project. I started with that, but there were some edge cases + explicitly marking parameters with SP markers (for gradient summation over TP axis) aligns better with the existing convention established by FSDP. I see your point though.

imichalakTT: I made existing sync_gradients to handle both jobs based on those markers, so the user API is aligned with that idea.

dyurshevichTT: could you create tickets on the edge cases? If its layout that doesn't propagate properly through the ops, we should fix it.

tt-train/sources/examples/train/model_builders.py
```diff
@@ -387,19 +391,23 @@ def parse_model_config(yaml_config: dict) -> ModelConfig:
...
 
 
-def create_model(cfg: ModelConfig, use_tp: bool = False) -> Model:
+def create_model(cfg: ModelConfig, tp_strategy: TPStrategy = TPStrategy.NONE) -> Model:
     """Dispatch on `cfg.model_type` to the registered builder.
 
     `cfg.vocab_size` is the raw (unpadded) vocab; each builder tile-pads it to a 32-multiple
     for the model, since the embedding/LM-head weights must be tile-aligned.
+
+    ``tp_strategy`` selects tensor / sequence parallelism (llama only; see ``TPStrategy``).
```
jmalone-tt, line 400: Is there a reason to keep this llama only? Or will we add support for other models at some point?

tt-train/sources/examples/lora_llama/train_lora_llama.py
jmalone-tt, line 1: I think we can fix a bug in this file. Right now, lora_A gradients aren't summed across TP ranks so they will gradually drift since they will have separate slices of lora_B. I believe we can mark_sequence_parallel on lora_A and make sure to sync_gradients.
I am not super familiar with this code, so I might be off here. What do you think?

PR #52828, opened 2026-08-11
[Bug Fix] reshape_on_device tile-reshape deadlock

abdulla-TT:
### PR Category
Bug fix

### Summary
Fixes item 2 of #51216 — a host/kernel contract mismatch that deadlocks `ttnn.reshape_on_device(...)` on the TILE path whenever the input and output have different tile padding.

**Root cause.** In `ReshapeTileProgramFactory` the writer's tile count was derived from the **input** tensor's padded volume (`input_tensor.physical_volume() / TILE_HW`), while the reader produces exactly `output_N*output_C*output_Ht*output_Wt` tiles counted from the **output** tensor's padded shape. Nothing constrains the two to match — `validate` only checks `input.physical_volume() % TILE_HW == 0`, and `reshape_on_device` only guarantees equal *logical* volumes. When the input and output round to different tile-padded shapes the counts disagree:
- e.g. logical `(1,1,1,1024)` (padded `(1,1,32,1024)` = 32 tiles) reshaped to `(1,1,32,32)` (1 tile): the writer waited for 32 tiles while the reader pushed only 1 → device hang.
- the mirror case `(1,1,32,32)` → `(1,1,1,1024)` inverts it: the reader pushes 32 tiles, the writer pops 1 → the reader wedges and most of the output is left uninitialized.

**Fix.** Derive the writer's tile count from the output tensor (`output_tensor.physical_volume() / TILE_HW`), which is exactly what the reader produces, so the producer/consumer counts always match.

**Test.** Added `test_reshape_on_device_tile_padding_mismatch` in `tests/tt_eager/python_api_testing/unit_testing/misc/test_reshape.py`, covering both directions and asserting exact (`torch.equal`) results. It hangs before the fix and passes after. Verified on device (all 4 tests in the file pass). Existing tests only use shapes whose H and W are already multiples of 32, where the two counts coincide, so the bug was never exercised.

Relates to #51216 (the remaining items are tracked separately in that issue).

### Notes for reviewers
- Single-line behavioral change: the writer's `num_tiles` now comes from the output instead of the input; the reader side is unchanged.
- The reshape reader (`reader_unary_reshape_interleaved.cpp`) remaps output sticks from the input by logical row-major position, so the results are numerically correct for both directions post-fix (padding rows in the mirror case are stripped by `to_torch`).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/reshape_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/reshape_deadlock_fix)


tests/tt_eager/python_api_testing/unit_testing/misc/test_reshape.py
```diff
@@ -110,3 +110,29 @@ def test_tile_major_reshape_var(device):
...
+    xtt = ttnn.reshape_on_device(xtt, 1, 1, 32, 32)
+    assert list(xtt.padded_shape) == [1, 1, 32, 32]
+    tt_got_back = ttnn.to_torch(xtt)
+    assert torch.equal(x.reshape(1, 1, 32, 32), tt_got_back)
+
+    # mirror: single input tile -> 32 output width tiles (only row 0 is real, rest is padding).
+    # Logical output is (1,1,1,1024); its padded shape rounds H up to a tile -> (1,1,32,1024).
+    y = torch.randn((1, 1, 32, 32), dtype=torch.bfloat16)
+    ytt = ttnn.from_torch(y, layout=ttnn.TILE_LAYOUT, device=device, dtype=ttnn.bfloat16)
+    ytt = ttnn.reshape_on_device(ytt, 1, 1, 1, 1024)
```
Aswinmcw: This direction reads out of bounds on the input buffer, and the assert can't see it. Input `(1,1,32,32)` → `input_Wt = 1`, `num_sticks_per_input_tile_row = 32`, buffer = 1 tile; output padded `(1,1,32,1024)` → `output_Ht = 1`, `output_Wt = 32`. In `reader_unary_reshape_interleaved.cpp` the stick id is `w + 32*tile_h`, reaching 1023, so `input_tile_row_to_read = 1023/32 = 31`, `input_tile_col_to_read = 1023 % 1 = 0`, `input_tile_to_read = 31` — pages 1..31 don't exist. Only `tile_h == 0` is real data; rows 1..31 land in output padding that `to_torch` strips, so the OOB NoC reads never reach the comparison. The forward case above is in bounds (`input_tile_to_read` maxes at 31 of a 32-tile input).

ttnn/cpp/ttnn/operations/data_movement/reshape_on_device/device/reshape_tile_program_factory.cpp
```diff
@@ -32,10 +32,12 @@ ProgramDescriptor ReshapeTileProgramFactory::create_descriptor(
 
     Buffer* src0_buffer = input_tensor.buffer();
 
-    uint32_t num_tiles = input_tensor.physical_volume() / tt::constants::TILE_HW;
-
     auto output_shape = output_tensor.padded_shape();
 
+    // Tile count must come from the output tensor to match what the reader produces; deriving it
+    // from the input deadlocks when the input and output have different tile padding.
+    uint32_t num_tiles = output_tensor.physical_volume() / tt::constants::TILE_HW;
```
Aswinmcw, line 39: The other half of the issue's suggested fix — a `TT_FATAL` in `validate_on_program_cache_miss` that input and output physical volumes match on the TILE path — isn't here. Worth adding: it would also reject the mirror shape added in the test, which is the correct outcome given the reader's OOB reads there.

Aswinmcw (approved):

VirdhatchaniKN (approved):

ttnn/cpp/ttnn/operations/data_movement/reshape_on_device/device/reshape_tile_program_factory.cpp
amorrisonTT, line 39: I just want to double check that this `num_tiles` doesn't get used in the context of the input tensor, eg the reader kernel reading input pages.

abdulla-TT: No — `num_tiles` only feeds the writer's runtime args; the reader gets no tile count, it loops `N × C × Ht × Wt` over the output shape. That's the bug: input-derived count made the writer pop fewer tiles than the reader pushed → hang.

abdulla-TT: No — `num_tiles` only feeds the writer's runtime args; the reader gets no tile count, it loops `N × C × Ht × Wt` over the output shape. That's the bug: input-derived count made the writer pop fewer tiles than the reader pushed → hang.

amorrisonTT (approved):

PR #52839, opened 2026-08-11
[feature] A single fused TTNN:QR-Decomposition Op. 

cfarm6:
Op structure modeled on bernoulli (dataflow writer does the math) and topk (two-output device operation); registered in the ttnn umbrella and nanobind.

### Summary

This provides a QR decomposition OP for use in linear algebra and scientific computing applications. Currently required for use in porting kUPS to use the TT backend. This performs a  fused Householder QR for rank-2 fp32 matrices up to 32x32 (one tile): reader kernel streams the input tile to L1, the writer kernel computes reduced QR in-register (fp32, IEEE soft-float; bit-seed Newton sqrt, sign(0)=1 zero guard), and both Q (m x k) and R (k x n) tiles are written straight to DRAM in a single dispatch.

### Notes for reviewers

Currently only focuses on using a single 32x32. Further improvements are needed for the Op to generalize but this should serve as a starting point. I'll likely update with an improved algorithm later.


cfarm6:
My coding agent added too many local commits into this PR. I'll clean this up and resubmit and clean PR without the extra CMake config for my local setup.

PR #52844, opened 2026-08-11
Add port-measure.yaml, the device half of the agentic porting pipeline

ebanerjeeTT:
## What this is

`port-op.md` is an agentic workflow that ports a tt-dm-codegen `generic_op` into tt-metal as a native C++ program factory and grades it on hardware. This PR adds the workflow it dispatches to do the building and measuring.

## Why it is a separate workflow, and why it has to be on main

The agent cannot run in-cluster. CIv2 reaches the network through `proxy.restricted-proxy.svc.cluster.local`, which answers 403 to `api.githubcopilot.com`, so every model request fails there. Running the agent on a GitHub-hosted runner solves that and creates a new problem: 4 cores and no Tenstorrent card. A probe measured the build reaching ~1050 of 1403 targets in 15 minutes, with the rate collapsing from ~145 targets/minute to 10-15 as it crossed into the ttnn unity builds — which is exactly the part an edit to an op touches. So the hosted runner builds nothing, and every build and every measurement is dispatched here, onto the pool where the build already happens with a warm Garage ccache.

`workflow_dispatch` only sees workflow files on the default branch, so this is not dispatchable at all until it lands on `main`. That is the only reason it is going in ahead of the rest of the pipeline.

It is a near-copy of [`test-dispatch.yaml`](.github/workflows/test-dispatch.yaml), which already solves the parts that are easy to get wrong: the dev-image container, `--device /dev/tenstorrent`, the viommu hugepages conditional, and the artifact-plus-wheel install. It is not simply `test-dispatch.yaml` with a `command` because the port harness needs `tt-dm-codegen` checked out, and a free-text `command` input cannot carry `CODEGEN_REPO_TOKEN` without putting it in run metadata and logs.

## Shape

Three modes, all `workflow_dispatch` only:

- `build` — `check-harbor` then `build-artifact` with `tracy: true` and `build-wheel: true`. A compile check needs no card, so it stops there and takes no device queue.
- `baseline` — adds a device job that emits the generated routing test, checks the resolved golden against native, and measures a native baseline. Run once before the agent starts.
- `verify` — the same device job running `gate.py` for the correctness and/or performance bands.

All three upload a results artifact with the verdict, the per-case measurements and the emitted test.

## Risk

- Nothing dispatches this yet. It has no `push`, `pull_request` or `schedule` trigger, so merging it changes no existing behaviour and consumes no CI capacity until someone dispatches it by hand or the porting pipeline does.
- The scripts it calls (`.github/scripts/port/*.py`) come from the checkout of the `ref` input, not from `main`, so this does not depend on anything else landing first.

## Test plan

- [ ] Dispatch with `mode: build` against a known-good ref and confirm the build job is the whole run.
- [ ] Dispatch with `mode: baseline` for `untilize` and confirm the device job produces a `port-results-baseline-*` artifact containing the emitted routing test and a non-`blocked` baseline verdict.
- [ ] Dispatch with `mode: verify` and confirm `gate.json` comes back in the artifact.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ebanerjee/port-measure-workflow)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ebanerjee/port-measure-workflow)

### DIFF
--- a/.github/workflows/port-measure.yaml
+++ b/.github/workflows/port-measure.yaml
@@ -16,73 +16,129 @@
 # is `CODEGEN_REPO_TOKEN`: the port harness needs the generator repo checked out, and a free-text
 # `command` input cannot carry a secret without putting it in run metadata and logs.
 #
-# `workflow_dispatch` only fires for workflow files on the default branch, so this has to be merged
-# to main before the agent can call it, regardless of which branch the port is developed on.
+# Triggered by the push of the scratch ref itself, not by `workflow_dispatch`. That is deliberate and
+# it is what lets the whole pipeline be developed on a branch: `workflow_dispatch` only fires for
+# workflow files that are already on the default branch, whereas a `push` event runs the workflow as
+# it exists *in the commit that was pushed*. The launcher has to push the code under test anyway, so
+# the push doubles as the trigger and this file needs no privileged position to be exercised.
+#
+# Two consequences worth knowing before editing this. A push made with `GITHUB_TOKEN` never starts a
+# workflow -- GitHub suppresses that to stop runs from triggering runs -- so the launcher must push
+# with a PAT or nothing happens here at all. And nothing else in the repository watches these refs:
+# every other `push` trigger is either pinned to a branch list that excludes them or filtered to
+# paths a scratch commit does not touch, so a dispatch costs exactly this workflow and no other.
 name: "(port) Build and measure a codegen port"
 
 on:
-  workflow_dispatch:
-    inputs:
-      mode:
-        description: "baseline proves the harness before the agent starts; build is a compile check; verify grades the port"
-        required: true
-        type: choice
-        options:
-          - baseline
-          - build
-          - verify
-      band:
-        description: "For verify: correctness, performance, or both"
-        required: false
-        type: string
-        default: both
-      op:
-        required: false
-        type: string
-        default: untilize
-      category:
-        required: false
-        type: string
-        default: data_movement
-      codegen-ref:
-        required: false
-        type: string
-        default: codegen_agentic_port
-      perf-limit:
-        description: "In-scope cases measured per performance band"
-        required: false
-        type: string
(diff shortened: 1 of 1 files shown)


ebanerjeeTT:
This is no longer a prerequisite for anything, and it should not be merged until the pipeline has actually run.

It was opened only because `workflow_dispatch` fires exclusively for workflow files already on the default branch, which made merging this a precondition for testing the pipeline at all — a bad ordering, since it meant merging an unproven workflow to `main` in order to find out whether it worked.

The workflow now triggers on the push of the scratch ref that the launcher has to make anyway. A `push` event runs the workflow as it exists in the pushed commit, so the whole pipeline can be developed and proven on a branch, with `main` untouched. Pushed here so the PR is not stale; leaving it as a draft until there is a green run behind it.

PR #52852, opened 2026-08-11
#51235: Fix reduce_to_root intermediate tensor spec and c_18 circular buffer overflow

### LINKED ISSUE #51235: [ttnn/ccl] reduce_to_root: wrong intermediate tensor spec and a 2-page CB overflow
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

Two high-severity defects in `ttnn/cpp/ttnn/operations/ccl/reduce_to_root/`, one host-side and one device-side.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS: @tenstorrent/metalium-developers-ops-data-movement

The two are independent, but both are in the same op so a single pass over `reduce_to_root` covers them. Item 1 is host-side (the intermediate tensor spec is one shard wide instead of the whole per-device tensor, and reuses the wrong shard shape); item 2 is a device-side CB overflow that overruns `c_18` by 2 pages — 1 KB with the shipped config — corrupting whatever L1 follows.

Prior art in the same op, neither overlapping these: #50308 (closed) fixed a deferred `noc_async_read_barrier` in `root_receive_reader_kernel.cpp`, and #48586 discusses the op in passing. Different files, different defects.

| # | Sev | Location | Defect |
|---|---|---|---|
| 1 | H | [`ccl/reduce_to_root/device/reduce_to_root_op.cpp:99`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_op.cpp#L99) | The auto-computed intermediate tensor spec is one shard wide (shard_w + 2*s_shard_w) instead of the whole per-device tensor, and it reuses the l tensor's layout (... |
| 2 | H | [`ccl/reduce_to_root/device/reduce_to_root_program.cpp:358`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_program.cpp#L358) | CB c_18 (packet_cb_id_2, used by root2_writer_kernel) is allocated only packet_size_bytes, but the writer fills it with packet_size_bytes + 2*aligned_page_size_by... |

---

### 1. `ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_op.cpp:99` — wrong-shape/units

**The auto-computed intermediate tensor spec is one shard wide (shard_w + 2*s_shard_w) instead of the whole per-device tensor, and it reuses the l tensor's layout (shard shape [8,128]) instead of a [8,192] shard, so each core's intermediate allocation is smaller than the packet the kernels write into it.**

**How it fails.** Call ttnn.reduce_to_root without intermediate_tensor (exactly what tests/ttnn/docs_examples/test_ccl_examples.py::test_reduce_to_root does): l is [8, 128*8] width-sharded [8,128] over 8 cores, s/m [8, 32*8] sharded [8,32]. compute_output_specs builds intermediate_shape = Shape{8, 128 + 2*32} = [8,192] and TensorSpec(intermediate_shape, l.tensor_layout()), i.e. shard spec still [8,128]. That tensor is only 6 pages wide => 2 shards distributed round-robin over the 8-core grid, so each L1 bank reserves 1 shard = 4 pages = 2048 B. Every one of the 8 shard cores then has 3072 B (packet_size_bytes + 2*aligned_page = 2048+1024) written to intermediate_tensor.buffer()->address() by the fabric fused write (sender_writer / root2_writer) and read back with the same size by the receive readers, overrunning the 2048 B allocation by 1 KB per core and silently corrupting adjacent L1.

**Evidence.** reduce_to_root_op.cpp lines 95-101: `uint32_t shape_0 = final_output_spec[0].memory_config().shard_spec()->shape[0]; uint32_t shape_1 = final_output_spec[0].memory_config().shard_spec()->shape[1] + (2 * final_output_spec[1].memory_config().shard_spec()->shape[1]); Shape intermediate_shape = Shape{shape_0, shape_1}; TensorSpec intermediate_spec(intermediate_shape, final_output_spec[0].tensor_layout());` -- shape built from SHARD dims (no num_cores factor) while the layout carries the l tensor's shard spec. The intended spec is spelled out by the caller that supplies its own intermediate in tests/.../nightly/test_reduce_to_root_trace.py: `intermediate_shapes = [[8, 192 * num_cores]]` with `shard_spec_int_l = ttnn.ShardSpec(shard_grid, [8, 192], ...)`. Kernel-side size: root_receive_reader_kernel.cpp l.95 `new_packet_size_bytes = packet_size_bytes + 2 * align(page_size_bytes, alignment)` read from `intermediate_base_addr`, and sender_writer_kernel.cpp l.53 writes the same size to the remote intermediate address.

**Suggested fix.** Build the intermediate spec for the full per-device tensor and give it its own shard spec: shape {shard_h, (shard_w + 2*s_shard_w) * num_shard_cores} with a MemoryConfig whose shard shape is {shard_h, shard_w + 2*s_shard_w} on the same grid (mirror what the trace test constructs), instead of reusing final_output_spec[0].tensor_layout().

<sub>CODEOWNERS: `@tenstorrent/metalium-developers-ops-data-movement`</sub>

- [ ] fixed

---

### 2. `ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_program.cpp:358` — buffer-overflow

**CB c_18 (packet_cb_id_2, used by root2_writer_kernel) is allocated only packet_size_bytes, but the writer fills it with packet_size_bytes + 2*aligned_page_size_bytes, writing 2 pages (1 KB with the shipped config) past the end of the circular buffer.**

**How it fails.** 4-device line, root at (1,0) (the shipped test config): l shard [8,128], tile (8,32), bf16 => input_page_size_bytes=512, input_num_tiles=4, packet_size_bytes=2048. cb_packet_config_2 total size = 2048 B (1 page). root2_writer_kernel computes new_payload_size_bytes = payload_size_bytes + 2*aligned_page_size_bytes = 2048 + 1024 = 3072 and tt_memmoves the s page to packet_base_addr+2048 and the m page to packet_base_addr+2560, i.e. 1024 bytes beyond the CB, then DMAs 3072 bytes from that base over fabric. The 1 KB tail lands in whatever CB the L1 allocator placed after c_18 (for the root2 device the next CB created is c_11 / compute_out_cb_l), silently corrupting another kernel's L1 region; any change in dtype/tile/shard size makes the clobbered region live.

**Evidence.** Program factory: `constexpr auto packet_cb_id_2 = tt::CBIndex::c_18; CircularBufferConfig cb_packet_config_2 = CircularBufferConfig(packet_size_bytes, {{packet_cb_id_2, input_dataformat}}).set_page_size(packet_cb_id_2, packet_size_bytes);` (lines 356-360) versus the sibling sized correctly for exactly the same packet layout: `auto total_pkt_size = packet_size_bytes + 1024; ... CircularBufferConfig(2 * total_pkt_size, {{packet_cb_id, input_dataformat}}).set_page_size(packet_cb_id, total_pkt_size)` (lines 278-283). Consumer root2_writer_kernel.cpp: `const uint32_t new_payload_size_bytes = payload_size_bytes + 2 * aligned_page_size_bytes;` (l.108) with `packet_base_addr = get_write_ptr(packet_cb_id)` (l.112), writes at `packet_base_addr + payload_size_bytes` (l.127) and `+ payload_size_bytes + aligned_page_size_bytes` (l.134), then `send_payload_without_header_non_blocking_from_address(packet_base_addr, new_payload_size_bytes)` (l.149). writer_ct_args for root2 pass packet_cb_id_2 as the packet CB and packet_size_bytes as payload_size_bytes (lines 584-594).

**Suggested fix.** Size c_18 like c_7: `CircularBufferConfig(total_pkt_size, {{packet_cb_id_2, input_dataformat}}).set_page_size(packet_cb_id_2, total_pkt_size)` (and prefer `packet_size_bytes + 2 * aligned_input_page_size_bytes` over the hard-coded +1024 in total_pkt_size, which only happens to be right for bf16 8x32 tiles).

<sub>CODEOWNERS: `@tenstorrent/metalium-developers-ops-data-movement`</sub>

- [ ] fixed

---

azaytsev-epam:
### Ticket
Closes #51235

### Problem description
`reduce_to_root` moves one packet per shard core, and every producer sizes that packet as
`packet_size_bytes + 2 * aligned_page_size_bytes` (the l payload plus the s and m pages) —
3072 B for the shipped 4-device bf16 config with `(8,32)` tiles. Two destinations reserved
only `packet_size_bytes` (2048 B), so on each of the 8 shard cores 1 KB was written past
the end of a live L1 region, silently corrupting whatever followed:

1. `compute_output_specs` built the auto-computed intermediate tensor from *shard*
   dimensions with no `num_cores` factor, and passed `final_output_spec[0].tensor_layout()`
   straight through, so the tensor was `[8, 192]` carrying l's `[8, 128]` shard spec.
   That reserves 2048 B per bank on 2 of the 8 cores, while the fabric fused write
   (`sender_writer_kernel.cpp:53`) and the receive readers
   (`root_receive_reader_kernel.cpp:95`) use 3072 B at `intermediate_tensor.buffer()->address()`
   on all 8 cores.
2. CB `c_18` (`packet_cb_id_2`) was created with `packet_size_bytes`, while
   `root2_writer_kernel` fills it with `payload_size_bytes + 2 * aligned_page_size_bytes`
   and sends from its base, overrunning it by 2 pages into the next CB created for the
   root2 device, `c_11` / `compute_out_cb_l`.

### What's changed
- `reduce_to_root_op.cpp`: build the intermediate spec for the whole per-device tensor —
  `{shard_h, (shard_w + 2 * s_shard_w) * num_shard_cores}` — with its own
  `{shard_h, shard_w + 2 * s_shard_w}` shard spec on the l grid, keeping l's dtype, page
  config and buffer type via `TensorLayout::with_memory_config`. The result is identical to
  the intermediate `test_reduce_to_root_trace.py` passes in explicitly, so the
  caller-supplied and auto-computed paths now agree.
- `reduce_to_root_program.cpp`: size `c_18` for a full packet (`total_pkt_size`), the way
  its sibling `c_7` already was, and derive `total_pkt_size` from
  `2 * aligned_input_page_size_bytes` rather than a hard-coded `+ 1024`. The constant is
  unchanged for bf16 `(8,32)` tiles and now also correct for other dtypes and tile sizes.

No kernel, API or tensor-spec change is visible to callers; the op needs 1 KB more L1 per
shard core for the intermediate and 1 KB more for `c_18` on root2 devices.

**Proof the fix works**
(TM-Fabric) Fabric Tests:

Result before fix (main @ 06aaedf1f): [run 31446969553](https://github.com/tenstorrent/tt-metal/actions/runs/31446969553) — [BH-LoudBox ccl nightly](https://github.com/tenstorrent/tt-metal/actions/runs/31446969553/job/93848739913) 193 passed, 159 skipped, [BH-Quietbox-2](https://github.com/tenstorrent/tt-metal/actions/runs/31446969553/job/93848739234), [P300-viommu](https://github.com/tenstorrent/tt-metal/actions/runs/31446969553/job/93848739387). The op had no live coverage:
test_reduce_to_root_trace.py::test_reduce_to_root_with_trace[...fabric_1d_linear_trace] SKIPPED
SKIPPED [1] test_reduce_to_root_trace.py:90: Consistently failing with data mismatch after trace replay on all BH multi-chip setups. See #39098
Results after fix (tree ca8333c7e, run on 1b2cb3343): [run 31588558302](https://github.com/tenstorrent/tt-metal/actions/runs/31588558302) — [BH-LoudBox ccl nightly](https://github.com/tenstorrent/tt-metal/actions/runs/31588558302/job/94091046207) 195 passed, 158 skipped, [P300-viommu ccl nightly](https://github.com/tenstorrent/tt-metal/actions/runs/31588558302/job/94091015927) 54 passed, 299 skipped:
PASSED test_reduce_to_root_trace.py::test_reduce_to_root_with_trace[...fabric_1d_linear_trace]
PASSED test_reduce_to_root_trace.py::test_reduce_to_root_auto_intermediate[...fabric_1d_linear]

Nightly tt-metal L2 tests, same SKU before and after — [ccl nightly tests [bh_quietbox_2]](https://github.com/tenstorrent/tt-metal/actions/runs/31597345244/job/94118661339) went from **1 failed, 280 passed** on `dc23d69` to **281 passed** on `38b3c7f`. On `dc23d69` the re-enabled trace test aborted in `end_trace_capture` with `Out of Memory: Not enough space to allocate 139264 B TRACE buffer across 8 banks, where each bank needs to store 24576 B, but bank size is 32768 B`, before it could compare any output. That was the test's own budget, not the fix: it keeps the 15-iteration warmup trace (12288 B/bank) and the 30-iteration perf trace (24576 B/bank) live simultaneously, needing 36864 B/bank against the 32768 B/bank that `trace_region_size: 225280` provides. The trace buffer is in DRAM while this fix only grows L1, and the CB portion of a dispatch command is sized by CB index count rather than CB bytes (`tt_metal/impl/program/dispatch.cpp:319`), so bytes-per-captured-op are identical either way. `38b3c7f` raises the region to 524288, after which the test passes its data comparison — so the mismatch reported in #39098 no longer reproduces.

Remaining red on this branch is pre-existing on `main` or infrastructure, none of it touching `reduce_to_root`: `Galaxy CCL tests [wh_galaxy]` and `BH Galaxy CCL tests [bh_galaxy]` fail identically on the `main` nightly ([run 31569965089](https://github.com/tenstorrent/tt-metal/actions/runs/31569965089)) — the former on `strided_reduce_scatter_async` ncrisc build errors, the latter requesting a 32-device mesh on a 16-device host; the BH-Quietbox-2 fabric jobs never started (`Value cannot be null. (Parameter 'ContainerId')`); and `wh-6u-fabric-tests` plus the T3K throughput ubench are unrelated Wormhole failures.

### Verification pipelines

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tm-fabric-tests.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pr-gate.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-static-checks.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/merge-gate.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:AlexeyZaytsev-epam/51235_bug_reduce_to_root)

### DIFF
--- a/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
+++ b/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
@@ -91,7 +91,9 @@ def compute_reference_reduce_to_root(
 @pytest.mark.parametrize(
     "device_params",
     [
-        ({"fabric_config": ttnn.FabricConfig.FABRIC_1D, "trace_region_size": 225280}),
+        # The warmup (15 iters) and perf (30 iters) traces are both live at end_trace_capture,
+        # which needs 36864 B per bank; 225280 only budgets 32768 B per bank.
+        ({"fabric_config": ttnn.FabricConfig.FABRIC_1D, "trace_region_size": 524288}),
     ],
     indirect=["device_params"],
     ids=["fabric_1d_linear_trace"],
--- a/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
+++ b/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
@@ -91,8 +91,6 @@ def compute_reference_reduce_to_root(
 @pytest.mark.parametrize(
     "device_params",
     [
-        # The warmup (15 iters) and perf (30 iters) traces are both live at end_trace_capture,
-        # which needs 36864 B per bank; 225280 only budgets 32768 B per bank.
         ({"fabric_config": ttnn.FabricConfig.FABRIC_1D, "trace_region_size": 524288}),
     ],
     indirect=["device_params"],
--- a/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
+++ b/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
@@ -9,6 +9,7 @@ from loguru import logger
 from tracy import signpost
 from models.perf.benchmarking_utils import BenchmarkProfiler
 from models.common.utility_functions import skip_for_wormhole_b0
+from tests.tt_eager.python_api_testing.sweep_tests.comparison_funcs import comp_pcc
 from tests.ttnn.unit_tests.operations.ccl.blackhole_CI.box.nightly.test_all_gather_nightly import validate_test
 
 
@@ -85,6 +86,24 @@ def compute_reference_reduce_to_root(
         m_final_cores.append(m_final)
 
     return torch.cat(l_final_cores, dim=1), torch.cat(s_final_cores, dim=1), torch.cat(m_final_cores, dim=1)
+
+
+# Tolerances for bfloat16 with exponentials.
+RTOL = 0.01
+ATOL = 0.06
+PCC = 0.99
+
+
+def assert_output_matches(name, output, reference, context=""):
+    """Element-wise closeness plus PCC, so a systematic shift and a few corrupted elements both fail."""
+    suffix = f" {context}" if context else ""
+    assert torch.allclose(
+        output, reference, rtol=RTOL, atol=ATOL
+    ), f"{name} tensor output does not match reference{suffix}"
+
+    passing, message = comp_pcc(reference, output, PCC)
+    logger.info(f"{name} tensor{suffix}: {message}")
+    assert passing, f"{name} tensor output failed PCC check{suffix}: {message}"
 
 
 @skip_for_wormhole_b0("This test is for blackhole")
@@ -308,27 +327,12 @@ def test_reduce_to_root_with_trace(bh_2d_mesh_device):
     out_s_root = out_s_torch[root_device_idx]
     out_m_root = out_m_torch[root_device_idx]
 
-    # Tolerances for bfloat16 with exponentials
-    rtol = 0.01
-    atol = 0.06
+    # S and M only carry a meaningful value in the first column of each core's shard.
+    cols_to_check = [i * 32 for i in range(num_cores)]
 
-    # Check L tensor
-    l_match = torch.allclose(out_l_root, l_ref, rtol=rtol, atol=atol)
-    assert l_match, "L tensor output does not match reference after trace execution"
-
-    # Check S tensor (only column 0)
-    s_cols_to_check = [i * 32 for i in range(8)]
-    s_output_col0 = out_s_root[:, s_cols_to_check]
-    s_ref_col0 = s_ref[:, s_cols_to_check]
-    s_match = torch.allclose(s_output_col0, s_ref_col0, rtol=rtol, atol=atol)
-    assert s_match, "S tensor output does not match reference after trace execution"
-
-    # Check M tensor (only column 0)
-    m_cols_to_check = [i * 32 for i in range(8)]
-    m_output_col0 = out_m_root[:, m_cols_to_check]
-    m_ref_col0 = m_ref[:, m_cols_to_check]
-    m_match = torch.allclose(m_output_col0, m_ref_col0, rtol=rtol, atol=atol)
-    assert m_match, "M tensor output does not match reference after trace execution"
+    assert_output_matches("L", out_l_root, l_ref, "after trace execution")
+    assert_output_matches("S", out_s_root[:, cols_to_check], s_ref[:, cols_to_check], "after trace execution")
+    assert_output_matches("M", out_m_root[:, cols_to_check], m_ref[:, cols_to_check], "after trace execution")
 
 
 @skip_for_wormhole_b0("This test is for blackhole")
@@ -427,14 +431,8 @@ def test_reduce_to_root_auto_intermediate(bh_2d_mesh_device):
     out_s_root = ttnn.to_torch(out_s, mesh_composer=composer)[root_device_idx]
     out_m_root = ttnn.to_torch(out_m, mesh_composer=composer)[root_device_idx]
 
-    rtol = 0.01
-    atol = 0.06
     cols_to_check = [i * 32 for i in range(num_cores)]
 
-    assert torch.allclose(out_l_root, l_ref, rtol=rtol, atol=atol), "L tensor output does not match reference"
-    assert torch.allclose(
-        out_s_root[:, cols_to_check], s_ref[:, cols_to_check], rtol=rtol, atol=atol
-    ), "S tensor output does not match reference"
-    assert torch.allclose(
-        out_m_root[:, cols_to_check], m_ref[:, cols_to_check], rtol=rtol, atol=atol
-    ), "M tensor output does not match reference"
+    assert_output_matches("L", out_l_root, l_ref)
+    assert_output_matches("S", out_s_root[:, cols_to_check], s_ref[:, cols_to_check])
+    assert_output_matches("M", out_m_root[:, cols_to_check], m_ref[:, cols_to_check])
--- a/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
+++ b/tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
@@ -87,9 +87,6 @@ def compute_reference_reduce_to_root(
     return torch.cat(l_final_cores, dim=1), torch.cat(s_final_cores, dim=1), torch.cat(m_final_cores, dim=1)
 
 
-@pytest.mark.skip(
-    reason="Consistently failing with data mismatch after trace replay on all BH multi-chip setups. See #39098"
-)
 @skip_for_wormhole_b0("This test is for blackhole")
 @pytest.mark.parametrize(
     "device_params",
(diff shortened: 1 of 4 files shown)


tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
```diff
@@ -87,14 +87,13 @@ def split_by_cores(tensor_list, core_width):
...
-@pytest.mark.skip(
-    reason="Consistently failing with data mismatch after trace replay on all BH multi-chip setups. See #39098"
-)
 @skip_for_wormhole_b0("This test is for blackhole")
 @pytest.mark.parametrize(
     "device_params",
     [
-        ({"fabric_config": ttnn.FabricConfig.FABRIC_1D, "trace_region_size": 225280}),
+        # The warmup (15 iters) and perf (30 iters) traces are both live at end_trace_capture,
+        # which needs 36864 B per bank; 225280 only budgets 32768 B per bank.
```
jvegaTT: The comment is unnecessary

tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_reduce_to_root_trace.py
```diff
@@ -332,3 +331,112 @@ def test_reduce_to_root_with_trace(bh_2d_mesh_device):
...
+    atol = 0.06
+    cols_to_check = [i * 32 for i in range(num_cores)]
+
+    assert torch.allclose(out_l_root, l_ref, rtol=rtol, atol=atol), "L tensor output does not match reference"
+    assert torch.allclose(
+        out_s_root[:, cols_to_check], s_ref[:, cols_to_check], rtol=rtol, atol=atol
+    ), "S tensor output does not match reference"
+    assert torch.allclose(
+        out_m_root[:, cols_to_check], m_ref[:, cols_to_check], rtol=rtol, atol=atol
+    ), "M tensor output does not match reference"
```
jvegaTT: Can we also validate PCC

ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_program.cpp
```diff
@@ -276,7 +276,8 @@ ttnn::device_operation::CachedProgram<ReduceToRootOp::ReduceToRoot::shared_varia
             .set_page_size(packet_header_cb_id, packet_header_size_bytes)
             .set_tile_dims(packet_header_cb_id, stats_tile);
 
-    auto total_pkt_size = packet_size_bytes + 1024;
+    // The writers append the s and m pages after the l payload, so a packet is 2 aligned pages larger than the payload.
+    const uint32_t total_pkt_size = packet_size_bytes + (2 * aligned_input_page_size_bytes);
```
jvegaTT, line 280: Please verify if this will work for large row major pages. Fabric has a maximum packet size I believe it is 4KB but it may be larger nowadays but with this change we run the risk that depending on the input config of the tensor we may exceed the maximum allowed by fabric. This may be a non issue, I know Sean was working on a change on the fabric side to break apart large packets in fabric so we can issue any size transfer and his APIs do the splitting but I don't know if that is done/merged. Just try this with for instance a 32x1000000 RM tensor and see if it works.

Another issue with this is it forces a whole two pages to fit in the CB space which is another possible issue introduced with large pages.

jvegaTT (commented):
I am worried this PR will introduce an issue with large pages. As I am going on PTO soon, I am not blocking but other reviewers please verify that the PR author has validated this before approving

amahmudTT:
> I am worried this PR will introduce an issue with large pages. As I am going on PTO soon, I am not blocking but other reviewers please verify that the PR author has validated this before approving

Github actually did not assign proper reviewers, codeowner-bypass is just for bypassing review. Who should review this one ?

amahmudTT:
@mateusznowakTT do you know who should review this pr ? Github did not auto assign properly.

ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_program.cpp
azaytsev-epam: device/reduce_to_root_program.cpp — added a TT_FATAL immediately after total_pkt_size, checking it against tt::tt_fabric::get_tt_fabric_max_payload_size_bytes(). 

device/reduce_to_root_op.cpp — added two TT_FATALs in validate: all three inputs must be Layout::TILE, and all three must be sharded.

reduce_to_root_nanobind.cpp — documented the per-shard-core packet budget and the 4352 B default in the docstring

ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_op.cpp
```diff
@@ -63,6 +63,21 @@ void ReduceToRootOp::validate(const operation_attributes_t& operation_attributes
             optional_output_tensor_m.value().device() == mesh_device,
             "Output tensor must be allocated on same mesh device as input tensor");
     }
+    const auto& input_s = tensor_args.input_tensor_s;
+    const auto& input_m = tensor_args.input_tensor_m;
+
+    TT_FATAL(
+        input_l.layout() == tt::tt_metal::Layout::TILE && input_s.layout() == tt::tt_metal::Layout::TILE &&
```
amorrisonTT, line 70: This should probably go in `validate_on_program_cache_miss`

ttnn/cpp/ttnn/operations/ccl/reduce_to_root/device/reduce_to_root_op.cpp
```diff
@@ -63,6 +63,21 @@ void ReduceToRootOp::validate(const operation_attributes_t& operation_attributes
...
+            input_m.layout() == tt::tt_metal::Layout::TILE,
+        "ReduceToRoot requires TILE layout for all three input tensors, got l: {}, s: {}, m: {}",
+        input_l.layout(),
+        input_s.layout(),
+        input_m.layout());
+
+    TT_FATAL(
+        input_l.is_sharded() && input_s.is_sharded() && input_m.is_sharded(),
```
amorrisonTT, line 78: Same

amorrisonTT (approved):
Approving to unblock

PR #52870, opened 2026-08-11
Support padding on every dimension of a rank > 4 tensor in ttnn.pad

### LINKED ISSUE #38144: ttnn::pad only supports padding on the lowest 3 dimensions for tensors with rank > 4
ttnn::pad only supports padding on the lowest 3 dimensions for tensors with rank > 4. Ideally it should support all dimensions. 
Following error occurs:
Always | TT_FATAL: ttnn::pad only supports padding on the lowest 3 dimensions for tensors with rank > 4

Python script to reproduce the issue:
```
import ttnn
import torch

input_torch = torch.randn(1, 4, 4, 4, 128, dtype=torch.bfloat16)
input_ttnn = ttnn.from_torch(
    input_torch, dtype=ttnn.DataType.BFLOAT16, layout=ttnn.Layout.ROW_MAJOR, device=device
)

padding = [(0, 0), (1, 1), (0, 0), (0, 0), (0, 0)]
output_ttnn = ttnn.pad(input_ttnn, padding, 0.0)

print(output_ttnn)

ttnn.close_device(device)
```

### LINKED ISSUE #53461: pad: rank > 4 shape restore drops padding on upper dimensions
item 2 from master ticket

## Problem

After device padding on a rank > 4 tensor, `update_original_shape()` rebuilds the output logical shape by copying **input** sizes for every dimension above the last two, and only taking the padded sizes from the **last two** dimensions.

Padding applied on dimension `rank-3` (or any upper dim the op allows) is **silently discarded** on restore. The returned tensor can report the wrong shape and volume.

## Example

Rank-5 tensor shape `(2, 3, 4, 5, 6)`, pad dimension `-3` by 1:

| | Shape |
|--|-------|
| Expected padded logical shape | `(2, 3, 5, 5, 6)` — volume 900 |
| Current restore | `(2, 3, 4, 5, 6)` — volume 720 |

The buffer holds 900 elements but the tensor is described as 720 — wrong answer or volume mismatch on downstream ops.

## Location

`ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp` — `update_original_shape()`

## Suggested fix

When rebuilding the logical shape, apply the full padding spec across all dimensions — not just the last two. For dims above the bottom three, derive sizes from the padded shape (or compute logical shape directly from input shape + padding).

## Test plan

Add unit tests in `tests/ttnn/unit_tests/operations/data_movement/test_pad.py` for rank ≥ 5 padding on dimension `rank-3` and other upper dims, comparing shape, volume, and values against PyTorch.

VirdhatchaniKN:
Fixes https://github.com/tenstorrent/tt-metal/issues/38144.

`ttnn.pad` rejected padding on any dimension above the lowest three once rank exceeded four, even though the op is documented as rank-generic. The device path already squeezed to 4D; the real gaps were (1) leading dimensions merged by `squeeze_from_ND_to_4D` had no padding mechanism, and (2) ND shape restore dropped padding on upper dims. This PR removes the rank > 4 restriction for device tensors, adds a reshape-based path for leading dims, fixes logical shape restore, and keeps the Quasar mirror in sync.

Host tensors with rank > 4 still only support padding on the lowest three dimensions (`TT_FATAL`), since host tile padding is not fixed here. When dim `rank-4` maps cleanly to 4D axis 0 after squeeze, that dim is handled in the final 4D pass to avoid an extra copy; otherwise it goes through the reshape path. Added rank > 4 unit tests (row-major, tile, dtypes, memory configs, multicore).

Pipeline status

- [x]  Sanity
- [x]  (Runtime) Sanity Test
- [x]  BHPC
- [x]  Nightly for DM Category
- [x]  All model test
- [x]  Single card
- [x]  T3K
- [x]  Galaxy
- [x]  (Tier 1) Models Device Perf Tests
- [x]  (TM-DM) Data Movement Test Suite

 
### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/38144)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/38144)

### DIFF
--- a/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
@@ -1341,6 +1341,36 @@ def _ttnn_padding_to_torch(padding):
     return tuple(v for i in reversed(range(len(padding))) for v in padding[i])
 
 
+def _assert_pad_fill_regions(output_tensor, input_shape, padding, value):
+    """Exact check that every padded slice equals the fill value."""
+    for dim, (before, after) in enumerate(padding):
+        if before:
+            slices = [slice(None)] * len(input_shape)
+            slices[dim] = slice(0, before)
+            assert torch.all(output_tensor[tuple(slices)] == value)
+        if after:
+            slices = [slice(None)] * len(input_shape)
+            slices[dim] = slice(before + input_shape[dim], before + input_shape[dim] + after)
+            assert torch.all(output_tensor[tuple(slices)] == value)
+
+
+def _assert_pad_body_matches(output_tensor, input_tensor, padding, *, atol=0.0):
+    """Check the unpadded region matches the input."""
+    slices = []
+    for dim, (before, _after) in enumerate(padding):
+        slices.append(slice(before, before + input_tensor.shape[dim]))
+    body_out = output_tensor[tuple(slices)]
+    if atol == 0.0:
+        assert torch.equal(body_out, input_tensor)
+    else:
+        assert_allclose(body_out.float(), input_tensor.float(), rtol=0, atol=atol)
+
+
+# Rank > 4 regression coverage below exercises ttnn.pad on device. The Quasar mirror
+# (ttnn.experimental.quasar.pad / experimental/quasar/pad/pad.cpp) carries the same
+# logic and is verified-by-inspection, same as the Quasar slice fix in #52902.
+
+
 @pytest.mark.parametrize(
     "shape,padding",
     [
@@ -1480,31 +1510,41 @@ def test_pad_rank_gt4_tile_bfloat8_b(device, padding):
     """bfloat8_b is padded via a bfloat16 round trip; make sure the rank > 4 path preserves that."""
     torch.manual_seed(0)
     shape = (1, 4, 4, 32, 128)
+    value = 7.0
 
     torch_input_tensor = torch.randn(shape, dtype=torch.bfloat16)
-    torch_output_tensor = torch.nn.functional.pad(torch_input_tensor, _ttnn_padding_to_torch(padding), value=0.0)
+    torch_output_tensor = torch.nn.functional.pad(torch_input_tensor, _ttnn_padding_to_torch(padding), value=value)
 
     input_tensor = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat8_b, layout=ttnn.TILE_LAYOUT, device=device)
-    output_tensor = ttnn.to_torch(ttnn.pad(input_tensor, padding, 0.0))
+    output_tensor = ttnn.to_torch(ttnn.pad(input_tensor, padding, value))
 
     assert output_tensor.shape == torch_output_tensor.shape
-    assert_with_pcc(torch_output_tensor.float(), output_tensor.float(), 0.999)
+    _assert_pad_fill_regions(output_tensor, shape, padding, value)
+    _assert_pad_body_matches(output_tensor, torch_input_tensor, padding, atol=0.05)
 
 
 @pytest.mark.parametrize("layout", [ttnn.ROW_MAJOR_LAYOUT, ttnn.TILE_LAYOUT])
 @pytest.mark.parametrize("memory_config", [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG])
-def test_pad_rank_gt4_memory_config(device, layout, memory_config):
+@pytest.mark.parametrize(
+    "shape,padding",
+    [
+        # Single leading dim: trailing squeeze path still applies memory_config.
+        ((1, 4, 4, 32, 128), ((0, 0), (0, 1), (0, 0), (0, 0), (0, 0))),
+        # Multiple leading dims only: memory_config must be applied after the reshape passes.
+        ((2, 4, 4, 4, 128), ((0, 1), (0, 2), (0, 0), (0, 0), (0, 0))),
+    ],
+    ids=["single_leading_dim", "all_leading_dims"],
+)
+def test_pad_rank_gt4_memory_config(device, layout, memory_config, shape, padding):
     """The requested output memory config must survive the reshape/pad/reshape round trip."""
     torch.manual_seed(0)
-    shape = (1, 4, 4, 32, 128)
-    # Tile layout does not support front padding on device.
-    padding = ((0, 0), (0, 1), (0, 0), (0, 0), (0, 0))
+    value = 0.0
 
     torch_input_tensor = torch.randn(shape, dtype=torch.bfloat16)
-    torch_output_tensor = torch.nn.functional.pad(torch_input_tensor, _ttnn_padding_to_torch(padding), value=0.0)
+    torch_output_tensor = torch.nn.functional.pad(torch_input_tensor, _ttnn_padding_to_torch(padding), value=value)
 
     input_tensor = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=layout, device=device)
-    output = ttnn.pad(input_tensor, padding, 0.0, memory_config=memory_config)
+    output = ttnn.pad(input_tensor, padding, value, memory_config=memory_config)
 
     assert output.memory_config().buffer_type == memory_config.buffer_type
     output_tensor = ttnn.to_torch(output)
--- a/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
@@ -1545,3 +1545,29 @@ def test_pad_rank_gt4_tile_front_padding_not_supported(device, padding, expect_e
 
     with expect_error(RuntimeError, "ttnn.pad: on device tile padding does not support front padding"):
         ttnn.pad(input_tensor, padding, 0.0)
+
+
+@pytest.mark.parametrize("use_multicore", [True, False], ids=["multicore", "singlecore"])
+@pytest.mark.parametrize(
+    "shape,padding",
+    [
+        # Leading dim (reshape path) and dim rank-3 (squeeze-to-4D path).
+        ((2, 4, 4, 4, 128), ((0, 1), (0, 0), (0, 0), (0, 0), (0, 0))),
+        ((2, 4, 4, 4, 128), ((0, 0), (0, 1), (0, 2), (0, 0), (0, 0))),
+    ],
+    ids=["dim0", "dim1_and_dim2"],
+)
+def test_pad_rank_gt4_use_multicore(device, shape, padding, use_multicore):
+    """The rank > 4 path must agree with torch for both the multicore and single-core kernels."""
+    torch.manual_seed(0)
+
+    torch_input_tensor = torch.randn(shape, dtype=torch.bfloat16)
+    torch_output_tensor = torch.nn.functional.pad(
+        torch_input_tensor, _ttnn_padding_to_torch(padding), mode="constant", value=0.0
+    )
+
+    input_tensor = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.ROW_MAJOR_LAYOUT, device=device)
+    output_tensor = ttnn.to_torch(ttnn.pad(input_tensor, padding, 0.0, use_multicore=use_multicore))
+
+    assert output_tensor.shape == torch_output_tensor.shape
+    assert torch.equal(torch_output_tensor, output_tensor)
--- a/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_pad.py
@@ -665,9 +665,7 @@ def _unsqueeze(smaller, larger, fill):
 )
 @pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.int32, ttnn.uint16, ttnn.bfloat8_b])
 def test_pad_tile(shape, padding, dtype, device):
-    if (shape, padding) in [([5, 4, 3, 2, 1], [1, 0, 0, 0]), ([5, 4, 3, 2, 1], [32, 32, 32, 64])]:
-        pytest.xfail("Can't pad upper dims with rank>4")
-
+    # Padding upper dims of a rank>4 tensor used to xfail; supported since issue #38144.
     if len(shape) < len(padding):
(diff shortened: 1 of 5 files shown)


ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -85,15 +85,78 @@ ttnn::Tensor pad_via_interleaved_composite(
...
+
+    uint32_t inner_extent = 1;
+    for (int i = dim + 1; i < rank - 2; i++) {
+        inner_extent *= shape[i];
+    }
+
+    ttsl::SmallVector<PadSpecDim> padding_4d = {
+        {0, 0}, {pad_spec.before_elements * inner_extent, pad_spec.after_elements * inner_extent}, {0, 0}, {0, 0}};
+
+    auto padded = ttnn::pad(reshaped, padding_4d, value, use_multicore, memory_config_arg, sub_core_grids);
```
Aswinmcw: `memory_config_arg`/`sub_core_grids` are forwarded to every intermediate pad, and this runs once per padded leading dim (`rank5_all_dims`, `rank6_dim0_dim2_dim5`) — a sharded config is sized for the final rank-N output, so it is wrong for the 4D intermediates, and when this path consumes all the padding the trailing `reshape` returns without re-applying it. Suggest `std::nullopt` for the intermediates and applying the requested config once at the end; `test_pad_rank_gt4_memory_config` only uses interleaved DRAM/L1 and only asserts `buffer_type`, so it won't catch this.

tests/ttnn/unit_tests/operations/data_movement/test_pad.py
```diff
@@ -1284,3 +1277,245 @@ def test_pad_nd_sharded_front_padding_tile_layout_not_supported(device, dtype):
...
+    shape = (1, 4, 4, 32, 128)
+
+    torch_input_tensor = torch.randn(shape, dtype=torch.bfloat16)
+    torch_output_tensor = torch.nn.functional.pad(torch_input_tensor, _ttnn_padding_to_torch(padding), value=0.0)
+
+    input_tensor = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat8_b, layout=ttnn.TILE_LAYOUT, device=device)
+    output_tensor = ttnn.to_torch(ttnn.pad(input_tensor, padding, 0.0))
+
+    assert output_tensor.shape == torch_output_tensor.shape
+    assert_with_pcc(torch_output_tensor.float(), output_tensor.float(), 0.999)
```
Aswinmcw: PCC 0.999 over the whole tensor is a weak assertion for a pure data-movement op, and both parametrized cases fill with 0.0, so a misplaced block would still correlate. Prefer a bit-exact/ULP comparison of the copied region plus an exact check that the padded region equals the fill value — every other new test in this block uses `torch.equal`.

Aswinmcw (approved):

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -85,15 +85,73 @@ ttnn::Tensor pad_via_interleaved_composite(
...
+
+    uint32_t inner_extent = 1;
+    for (int i = dim + 1; i < rank - 2; i++) {
+        inner_extent *= shape[i];
+    }
+
+    ttsl::SmallVector<PadSpecDim> padding_4d = {
+        {0, 0}, {pad_spec.before_elements * inner_extent, pad_spec.after_elements * inner_extent}, {0, 0}, {0, 0}};
+
+    auto padded = ttnn::pad(reshaped, padding_4d, value, use_multicore, std::nullopt, std::nullopt);
```
Aswinmcw: Dropping `memory_config` here is right, but `sub_core_grids` is a core-allocation constraint rather than a shape-dependent config, so it should still be forwarded — as written, a caller that restricts pad to a sub-grid gets intermediate leading-dim pads running on the full grid. Suggest keeping the `sub_core_grids` parameter on `pad_leading_dimension_via_reshape`/`pad_leading_dimensions` and passing only `std::nullopt` for the memory config.

Aswinmcw (approved):
Both points addressed — memory config is now applied once on the final tensor (with an `all_leading_dims` case added), and the bfloat8_b test asserts the fill regions exactly plus a bounded body comparison.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -85,15 +85,75 @@ ttnn::Tensor pad_via_interleaved_composite(
...
+        dim,
+        rank);
+
+    uint32_t before = 1;
+    for (int i = 0; i < dim; i++) {
+        before *= shape[i];
+    }
+
+    uint32_t middle = 1;
+    for (int i = dim; i < rank - 2; i++) {
```
amorrisonTT: Nit: these loops are a good opportunity to use std::accumulate

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -268,6 +328,22 @@ ttnn::Tensor pad_impl(
         padding_size = padding.size();
         extra_index = 0;
     }
+    if (original_rank > 4) {
+        // Only padding[extra_index .. extra_index + 3] reaches the 4D kernel. Dims below extra_index have no
+        // slot in the 4D spec at all, and axis 0 of the squeezed tensor may be several original dims folded
+        // together -- padding applied there would stretch the whole merged axis instead of one dimension.
+        // pad_leading_dimensions() consumes those dims before we get here; this guard is what the removed
```
amorrisonTT: Let's drop the Claude-slop historical context from this comment. It can be found via git history if necessary

ttnn/cpp/ttnn/operations/experimental/quasar/pad/pad.cpp
```diff
@@ -28,15 +28,77 @@ using ttnn::operations::data_movement::squeeze_or_unsqueeze_shape_to_ND;
...
+    const int leading_dims_end = rank - 3;
+    for (int dim = 0; dim < leading_dims_end; dim++) {
+        if (padding[dim].before_elements == 0 && padding[dim].after_elements == 0) {
+            continue;
+        }
+        result = pad_leading_dimension_via_reshape(result, dim, padding[dim], value, use_multicore, sub_core_grids);
+        padding[dim] = {0, 0};
+    }
+    return result;
 }
```
amorrisonTT, line 108: Same comments for the quasar version

amorrisonTT (approved):

bbradelTT (approved):

VirdhatchaniKN:
https://github.com/tenstorrent/tt-metal/actions/runs/32034379617 - passed

fplavecTT:
Please don't leave AI-generated PR description as is, because it is usually overly verbose. For example, there is almost never a need for PR description to list all the changed files, since GH provides that functionality. There is also no need to have two different checklists for pipelines. In general, 2-3 paragraphs is plenty for a PR description.

tests/ttnn/unit_tests/operations/data_movement/test_pad.py
```diff
@@ -613,9 +608,7 @@ def _unsqueeze(smaller, larger, fill):
 )
 @pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.int32, ttnn.uint16, ttnn.bfloat8_b])
 def test_pad_tile(shape, padding, dtype, device):
-    if (shape, padding) in [([5, 4, 3, 2, 1], [1, 0, 0, 0]), ([5, 4, 3, 2, 1], [32, 32, 32, 64])]:
-        pytest.xfail("Can't pad upper dims with rank>4")
-
+    # Padding upper dims of a rank>4 tensor used to xfail; supported since issue #38144.
```
fplavecTT: These kind of AI-generated historical comments are not useful. Comments should describe current state of code and tests (not trace its history), so they are useful to a future reader.

tests/ttnn/unit_tests/operations/data_movement/test_pad.py
```diff
@@ -1284,3 +1277,280 @@ def test_pad_nd_sharded_front_padding_tile_layout_not_supported(device, dtype):
             use_multicore=True,
             memory_config=output_memory_config,
         )
+
+
+def _ttnn_padding_to_torch(padding):
+    """Convert ttnn per-dim (before, after) padding to torch.nn.functional.pad format."""
+    return tuple(v for i in reversed(range(len(padding))) for v in padding[i])
```
fplavecTT, line 1341: This one liner could use a code comment explaining why this non-trivial operation is needed.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -454,23 +532,26 @@ ttnn::Tensor pad(
         return input_tensor;
     }
 
+    ttnn::Tensor working_tensor = input_tensor;
     if (original_rank > 4) {
-        const auto first_pad_idx =
-            std::find_if(
-                working_padding.begin(), working_padding.end(), [](auto& p) { return p.after_elements != 0; }) -
-            working_padding.begin();
-        TT_FATAL(
```
fplavecTT, line 462: Isn't this `TT_FATAL` now removed even for host tensors, which are not fixed (according to PR description)? If so, it would be better to leave `TT_FATAL` active for host tensors.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -85,15 +86,69 @@ ttnn::Tensor pad_via_interleaved_composite(
 
 bool eq_spans(const auto a, const auto b) { return std::equal(a.begin(), a.end(), b.begin(), b.end()); }
 
-ttnn::Shape update_original_shape(const ttnn::Shape& padded_shape, const ttnn::Shape& input_shape) {
-    ttsl::SmallVector<uint32_t> updated_shape;
-    size_t input_rank = input_shape.rank();
-    for (size_t i = 0; i < input_rank - 2; i++) {
-        updated_shape.push_back(input_shape[i]);
+ttnn::Shape compute_padded_logical_shape(const ttnn::Shape& input_shape, const ttsl::SmallVector<PadSpecDim>& padding) {
```
fplavecTT, line 87: All the newly added helpers that are not exposed in the header should go into the anonymous namespace above.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
```diff
@@ -268,6 +323,19 @@ ttnn::Tensor pad_impl(
         padding_size = padding.size();
         extra_index = 0;
     }
+    if (original_rank > 4) {
+        // Only padding[extra_index .. extra_index + 3] reaches the 4D kernel. Dims below extra_index have no
+        // slot in the 4D spec, and axis 0 of the squeezed tensor may fold several original dims together.
+        const auto is_unpadded = [](const PadSpecDim& p) { return p.before_elements == 0 && p.after_elements == 0; };
+        const bool axis0_is_merged =
+            input_tensor_4D.logical_shape()[0] != input_tensor.logical_shape()[static_cast<int>(extra_index)];
+        TT_FATAL(
```
fplavecTT, line 350: The leading-dimension path does one full device pad per dimension, so padding two leading dims plus a trailing dim costs three complete copies of the data. However, the 4D kernel accepts padding amounts for all four slots in a single call, but each round here fills in only slot 1 and passes {0,0} for the rest, and the final trailing-dims round fills slots 1 - 3 while leaving slot 0 at zero. So there's a free slot in the last round.
Could we take advantage of this to make it more efficient and reduce number of copies?

Same would apply to Quasar version.

tests/ttnn/unit_tests/operations/data_movement/test_pad.py
VirdhatchaniKN: Removed historical/issue-reference comments from the rank > 4 test section and trimmed verbose docstrings.

tests/ttnn/unit_tests/operations/data_movement/test_pad.py
VirdhatchaniKN: Added a comment explaining that torch.nn.functional.pad expects padding pairs in reverse dimension order (last dim first).

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
VirdhatchaniKN: Restored TT_FATAL for host tensors with rank > 4; device tensors use the leading-dim reshape path.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
VirdhatchaniKN: Moved eq_spans, compute_padded_logical_shape, pad_leading_dimension_via_reshape, and pad_leading_dimensions into an anonymous namespace. Exposed apply_leading_dimension_padding as the single detail entry point.

ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp
VirdhatchaniKN: Dim rank-4 is deferred to the final 4D pass only when squeezed axis 0 equals shape[rank-4] (not merged with earlier dims). Otherwise it still uses the reshape path. Mirrored in Quasar.

VirdhatchaniKN:
Addressed the latest review feedback in 26ee65fb9a8 and shortened the PR description.

fplavecTT (approved):

VirdhatchaniKN:
https://github.com/tenstorrent/tt-metal/actions/runs/32350462318 - passed

PR #52872, opened 2026-08-11
Add opt-in file-export transport for prefill migration device map

dmadicTT:
## Ticket
tenstorrent/tt-d-gen#270

## Problem description
The KV-migration bring-up delivers each rank's device map to the co-located migration worker over shmem queues and blocks on WORKER_READY. For the KV manager (tt-d-gen) we instead want the runner to simply drop the device map (and leave the KV chunk table) on the filesystem on each host. An earlier quick-and-dirty branch (`dmadicTT/devmap-to-file`) did this but removed the shmem-queue transport entirely.

## What's changed
- The shmem-queue transport remains the **default**; behavior is unchanged when the toggle is off.
- New **feature toggle** `PREFILL_MIGRATION_EXPORT_TO_FILE=1` enables file-export mode:
  - every rank writes its local device map to a host-local text file (`PREFILL_MIGRATION_DEVICE_MAP_PATH`, default `/tmp/prefill_device_map.txt`; one `mesh_id chip_id umd_id` line per chip, atomic write),
  - rank 0 builds + serializes the merged KV chunk table to `PREFILL_MIGRATION_TABLE_PATH` as before, but skips SET_TABLE / WORKER_READY (no `MigrationLayerClient`),
  - incompatible combinations (`PREFILL_MIGRATION_SELFTEST`, `PREFILL_MIGRATION_INTERLEAVED`) are rejected up front since they need the worker client to issue `migrate()`.
- The two modes are kept in clearly separated branches (`migration.py` helpers + single `_file_export` forks in `prefill_runner.py`) so either can be removed wholesale later.

## Checklist
- [ ] New/Existing tests provide coverage for changes

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dmadicTT/devmap-to-file-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dmadicTT/devmap-to-file-toggle)


idjuricTT (approved):

philei-tt:
`export_device_map_to_file` (`migration.py:291-300`) writes each rank's *local* device map by overwriting a shared, non-rank-namespaced path (`path.tmp` -> `os.replace`). If two ranks are co-located on the same host, this loses data rather than merging it — worst case (overlapping writes racing on the same `path.tmp` inode) it can even produce a garbled file, not just clean last-writer-wins. Compare with the decode-side PR (tenstorrent/tt-blaze#3073), which handles multiple co-located ranks via flock-serialized append (+ a once-per-run truncate to avoid stale entries across relaunches).

Today most prefill models run a single rank per host, so this likely isn't hit yet. But #51874 (intra-galaxy pipeline-parallel prefill) will put multiple ranks on one host, at which point this will silently drop device-map entries. Worth fixing before that lands, or at least tracking it.

philei-tt (approved):

PR #52893, opened 2026-08-12
[Feature] Support read-only pinned host memory

### LINKED ISSUE #46132: Support copying read-only data to device using pinned memory
The KMD recently added support for pinning memory for read only, which is particularly helpful when reading data from disk. We should add support to this in metal.

jbaumanTT:
### PR Category
Feature

### Summary
Closes #46132.

Pins host memory device-**read-only** on the H2D path, so buffers the process cannot write can still take the pinned fast path.

Uploading a tensor only ever needs the device to *read* host memory, but until now every pin was read/write. That forecloses the case that motivates this change: `ttnn.load_tensor()` opens a tensor file `O_RDONLY` and maps it `PROT_READ | MAP_PRIVATE`, and a read/write pin of that mapping cannot be created at all — so a large file-backed upload falls off the pinned path and gets copied instead.

KMD 2.9.0 added a pinning mode where the IOMMU mapping permits device reads and faults device writes. This threads that mode through `PinnedMemory` and the pin cache and uses it for H2D transfers. D2H stays read/write, since the device writes into the host mapping there.

The paired UMD change, tenstorrent/tt-umd#3156, has merged, and `main` already points at a umd that carries it. This PR therefore contains **no submodule bump** and has no outstanding dependency — it is rebased onto `main` and builds against the umd pinned there.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jbauman/read-only-page-pinning)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jbauman/read-only-page-pinning)

### DIFF
--- a/tests/pipeline_reorg/runtime_unit_tests.yaml
+++ b/tests/pipeline_reorg/runtime_unit_tests.yaml
@@ -210,6 +210,7 @@
   cmd: |
     TT_METAL_SLOW_DISPATCH_MODE=1 ./build/test/tt_metal/unit_tests_device --gtest_filter='*PinnedMemory*'
     ./build/test/tt_metal/distributed/distributed_unit_tests --gtest_filter='*PinnedMemory*'
+    ./build/test/tt_metal/unit_tests_api --gtest_filter='*PinnedMemory*'
     ./build/test/tt_metal/unit_tests_tensor --gtest_filter='*ReadOnly*:*UniformCopyToDevice_ReusesPinnedMemoryCacheEntries:*UniformCopyToHost_ReusesPinnedMemoryCacheEntries'
   skus:
     wh_n150_civ2:
--- a/tests/pipeline_reorg/runtime_unit_tests.yaml
+++ b/tests/pipeline_reorg/runtime_unit_tests.yaml
@@ -202,12 +202,15 @@
 # complete in time on 8 devices. Smaller SKUs above skip the multi-device (2x4/4x8) suites.
 
 # --- IOMMU PinnedMemory Tests (SD + FD) ---
+# This job installs only the build artifact, not the Python wheel (see setup-job's wheel-artifact-name input,
+# which runtime-unit-tests-impl.yaml does not pass), so every command here must be a prebuilt C++ binary. The
+# matching pytest, test_large_read_only_file_backed_tensor_upload, runs from the ttnn sanity job instead, which
+# already covers all of tests/ttnn/unit_tests/tensor and does install the wheel.
 - name: runtime_iommu_pinned_memory
   cmd: |
     TT_METAL_SLOW_DISPATCH_MODE=1 ./build/test/tt_metal/unit_tests_device --gtest_filter='*PinnedMemory*'
     ./build/test/tt_metal/distributed/distributed_unit_tests --gtest_filter='*PinnedMemory*'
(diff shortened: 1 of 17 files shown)


tdowdallTT (approved):

bbradelTT (approved):

sidnadTT (approved):

riverwuTT (approved):

ruizhangTT (approved):

tt_metal/distributed/pinned_memory.cpp
```diff
@@ -381,7 +414,8 @@ std::shared_ptr<PinnedMemory> PinnedMemory::Create(
     void* host_ptr = static_cast<void*>(bytes.data());
     size_t buffer_size = bytes.size();
 
-    auto pinned_memory = std::shared_ptr<PinnedMemory>(new PinnedMemory(devices, host_ptr, buffer_size, map_to_noc));
+    auto pinned_memory =
+        std::shared_ptr<PinnedMemory>(new PinnedMemory(devices, host_ptr, buffer_size, map_to_noc, access));
```
vtsilytskyiTT, line 418:
```suggestion
        std::make_shared<PinnedMemory>(devices, host_ptr, buffer_size, map_to_noc, access);
```

vtsilytskyiTT (approved):
Looks good.

tests/pipeline_reorg/runtime_unit_tests.yaml
```diff
@@ -202,10 +202,15 @@
 # complete in time on 8 devices. Smaller SKUs above skip the multi-device (2x4/4x8) suites.
 
 # --- IOMMU PinnedMemory Tests (SD + FD) ---
+# This job installs only the build artifact, not the Python wheel (see setup-job's wheel-artifact-name input,
+# which runtime-unit-tests-impl.yaml does not pass), so every command here must be a prebuilt C++ binary. The
+# matching pytest, test_large_read_only_file_backed_tensor_upload, runs from the ttnn sanity job instead, which
+# already covers all of tests/ttnn/unit_tests/tensor and does install the wheel.
```
abhullar-tt, line 208: @riverwuTT should we move these tensor tests into a runtime pipeline

abhullar-tt (approved):

tt_metal/distributed/pinned_memory.cpp
jbaumanTT: The pinned memory constructor is private, so shared_ptr can't call into it.

Issue #51214, opened 2026-07-27
[ttnn/data_movement] 11 wrong-index / wrong-shape defects causing silent wrong data


amahmudTT:
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

## Wrong index, shape or guard arithmetic — 11 findings

The computation is well-formed but addresses the wrong element, reports the wrong shape, or checks the wrong quantity.

**Manifests as:** Silent wrong data. Every access stays in bounds, so nothing asserts and nothing crashes — the op just returns the wrong answer.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS for the paths below: @aczajkowskiTT @sjameelTT @bbradelTT @nsorabaTT, @tenstorrent/metalium-developers-ops-data-movement

**Items 1-5 share one review lens:** a 4D (or rank-3) assumption leaking into a rank-generic entry point. They sit in five different ops, so they can be fixed and closed independently.

**Items 6-7** are the same logical-vs-padded-shape confusion in `split`, so they are one fix unit.

| # | Location | Defect |
|---|---|---|
| 1 ✅ | [`expand/expand.cpp:19`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/data_movement/expand/expand.cpp#L19) | create_repetition_vector left-aligns the input shape against the requested output shape; torch.expand (and ttnn::repeat's own match_input_rank) right-align. Any e... |
| 2 ✅ | [`pad/pad.cpp:91`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp#L91) | update_original_shape() rebuilds the post-pad shape for rank>4 tensors by taking only the last TWO dims from the padded shape, so padding applied to the third-fro... |
[...]

PR #52902, opened 2026-08-12
Fix slice tensor-args reader kernel upper-dim offset (Horner loop order)

### LINKED ISSUE #52901: slice: upper-dimension start offset uses reverse Horner order in TILE reader kernel
## Parent

Sub-issue of #51214 (item 5).

## Summary

In the TILE slice reader kernel `reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp`, the loop that computes the **upper-dimension start tile offset** iterates dimensions in **reverse** order. That applies column-major flattening instead of the row-major layout used by the host-side offset helper and the rest of the slice stack.

When slicing a rank ≥ 4 TILE tensor with a non-zero start index on an upper dimension (not the last two), the reader begins at the wrong tile page. Access stays in bounds — **no crash, silent wrong data**.

## Location

`ttnn/cpp/ttnn/operations/data_movement/slice/device/kernels/dataflow/reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp` (~line 103)

## Buggy code

```cpp
uint32_t upper_dims_offset = 0;
for (int32_t i = num_dims - 3; i >= 0; i--) {
    upper_dims_offset = upper_dims_offset * input_shape_args[i] + start_indices[i];
}
start_offset += upper_dims_offset * multiplier;
```

The loop runs from the **innermost upper dim outward** (reverse). Horner evaluation for row-major linearization must run from the **outermost upper dim inward** (forward).

## How it fails

Example: TILE rank-4 tensor shape `[2, 3, 64, 64]`, slice start `[1, 0, 0, 0]`.

| | Tile-page offset (upper dims only) |
|--|-------------------------------------|
| Host `get_upper_start_offset` (correct) | **12** |
| Kernel (buggy reverse loop) | **4** |

The reader starts **8 tile pages too early** and returns wrong slice contents.

### Additional cross-checks (kernel vs host)

| Shape | Start | Correct | Kernel (buggy) |
|-------|-------|---------|----------------|
| [2,3,64,64] | [1,0,0,0] | 12 | 4 |
| [4,8,64,64] | [1,3,0,0] | 44 | 52 |
| [2,5,32,32] | [1,1,0,0] | 6 | 3 |
| [3,2,4,64,64] | [1,1,1,0,0] | 52 | 40 |

**Why CI missed it:** rank == 3 has only one upper dimension, so reverse and forward loops coincide. Existing tests likely only hit that degenerate case or start `[0, …]` on upper dims.

## Suggested fix

Forward Horner loop (row-major):

```cpp
for (uint32_t i = 0; i + 2 < num_dims; ++i) {
    upper_dims_offset = upper_dims_offset * input_shape_args[i] + start_indices[i];
}
```

## Test plan

Add a unit test (e.g. in slice nightly/unit tests) that:

1. Creates a TILE tensor with rank ≥ 4 and **distinct** values per upper-dim slice (so wrong offset is detectable).
2. Calls `ttnn.slice` with non-zero start on dim 0 (or another upper dim).
3. Compares against PyTorch / host reference slice — PCC or exact match.

Minimal repro shape: `[2, 3, 64, 64]`, start `[1, 0, 0, 0]`, end `[2, 3, 64, 64]`.

## Scope

- **1-line kernel fix** (+ test).
- No API change.
- Single-card Wormhole test sufficient.

## Related

- Parent audit: #51214
- Same bug class as split items 6–7 (logical vs padded confusion) but **different root cause** — index arithmetic order, not shape metadata.

VirdhatchaniKN:
### Issue
#52901
Sub-issue of #51214 (item 5)

### Summary

- Fix reverse Horner loop in the TILE slice tensor-args reader kernel (`reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp`). The loop computed upper-dimension tile-page offsets in column-major order instead of row-major, causing **silent wrong data** (no crash) for rank ≥ 4 tensors with non-zero start on upper dimensions.
- Apply the same 1-line fix to the Quasar mirror kernel.
- Add device regression test `test_slice_tensor_args_upper_dim_offset` (6 cases) with deterministic `offset_increment_tensor` input and exact `assert_equal`.

---

### Root cause

The kernel Horner loop iterated upper dimensions **innermost → outermost** (reverse / column-major):

```cpp
for (int32_t i = num_dims - 3; i >= 0; i--) {
    upper_dims_offset = upper_dims_offset * input_shape_args[i] + start_indices[i];
}
```

The host helper `get_upper_start_offset()` and the rest of the slice stack use **outermost → innermost** (forward / row-major).

Example: shape `[2, 3, 64, 64]`, start `[1, 0, 0, 0]`

| | Upper tile-page offset |
|--|------------------------|
| Host (correct) | **12** |
| Buggy reverse loop | **4** |
| Fixed forward loop | **12** |

**Why CI missed it:**

- Rank 3 has only one upper dimension → forward and reverse coincide.
- Many tests use `start [0, …]` on upper dims → offset 0 either way.
- Bug only shows for rank ≥ 4 with non-zero start on upper dims (not H/W).

The non-tensor-args TILE reader is unaffected — host precomputes `start_id` via `get_tiled_start_offset()`.

---

### Changes

| File | Change |
|------|--------|
| `slice/.../reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp` | Reverse → forward Horner loop + cross-ref comment |
| `experimental/quasar/slice/.../reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp` | Same fix + comment (cross-ref to quasar `slice_device_operation.cpp`) |
| `tests/.../test_slice.py` | `test_slice_tensor_args_upper_dim_offset` — 6 device cases |

### Kernel fix

Changed the upper-dimension Horner loop in both tensor-args reader kernels (data_movement and Quasar mirror) from **reverse** iteration (innermost upper dim first, column-major) to **forward** iteration (outermost upper dim first, row-major). Added a cross-reference comment tying the kernel loop to `get_upper_start_offset()` in the respective `slice_device_operation.cpp`. The host helper was already correct; only the on-device kernel math diverged.

---

### Test plan

*Device pytest*

```bash
pytest tests/ttnn/unit_tests/operations/data_movement/test_slice.py::test_slice_tensor_args_upper_dim_offset -v
```

- [x] 6 parametrized cases (rank 3, 4, 5 TILE tensor-args path)
- [x] Deterministic `offset_increment_tensor` (float32) + `assert_equal`
- [x] Full slice unit suite: **547+ passed**, 0 regressions

| Case | Shape | Start |
|------|-------|-------|
| Rank 3 invariant | `[4, 64, 64]` | `[2, 0, 0]` |
| Minimal repro | `[2, 3, 64, 64]` | `[1, 0, 0, 0]` |
| Asymmetric dims | `[4, 2, 32, 32]` | `[2, 0, 0, 0]` |
| Strong Horner check | `[6, 4, 32, 32]` | `[3, 0, 0, 0]` |
| Rank 5 | `[3, 2, 4, 64, 64]` | `[1, 0, 0, 0, 0]` |
| Rank 5 large offset | `[6, 3, 2, 32, 32]` | `[3, 0, 0, 0, 0]` |

**Note:** The tensor-args device path only slices one dimension (`slice_dim` / `num_devices`). Multi-upper-dim non-zero starts (e.g. `[1,3,0,0]`) are not exercised by the Python API; the kernel fix for those cases is verified-by-inspection (same forward Horner change as the tested data_movement path).

---

**Quasar note:** The Quasar tensor-args kernel mirror is fixed (same 1-line change as data_movement). This PR does not add a Quasar device test; the kernel fix is verified-by-inspection (identical diff to the tested data_movement path).

---

### Pipeline status

- [x]  Sanity
- [x]  (Runtime) Sanity Test
- [x]  BHPC
- [x]  Nightly for DM Category
- [x]  All model test
- [x]  Single card
- [x]  T3K
- [x]  Galaxy
- [x] (TM-DM) Data Movement Test Suite

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/slice_52901)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/slice_52901)


tests/ttnn/unit_tests/gtests/data_movement/test_slice_upper_start_offset.cpp
```diff
@@ -0,0 +1,134 @@
+// SPDX-FileCopyrightText: © 2025 Tenstorrent USA, Inc.
```
Aswinmcw: Copyright year should be 2026 on a new file.

tests/ttnn/unit_tests/gtests/data_movement/test_slice_upper_start_offset.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+#include <tt-metalium/constants.hpp>
+#include "ttnn/operations/data_movement/slice/device/slice_device_operation.hpp"
+#include "ttnn/tensor/shape/shape.hpp"
+#include "ttnn/tensor/types.hpp"
+
+namespace ttnn::operations::data_movement::test {
+namespace {
+
+// Mirror of the fixed forward Horner loop in
```
Aswinmcw: This gtest compares the host helper against a *copy* of the loop that lives in this file, so it cannot catch a regression of this fix — revert the kernel and `KernelAndHostUpperStartOffsetAgree` still passes (and `EXPECT_NE(reverse_offset, host_offset)` is a property of two functions defined here, not of production code). That matters because the PR body says the multi-upper-dim tuples (`[4,8,64,64]` / `[1,3,0,0]`, `[3,2,4,64,64]` / `[1,1,1,0,0]`) are "covered by the host gtest" — against the kernel they aren't covered anywhere. Worth noting in the comment that the replica can drift, and reconsidering the `UNIT_TESTS_TTNN_SMOKE_SOURCES` tier for a test with no production-code coupling.

ttnn/cpp/ttnn/operations/experimental/quasar/slice/device/kernels/dataflow/reader_unary_unpad_dims_interleaved_start_id_tensor_args.cpp
```diff
@@ -105,7 +105,8 @@ void kernel_main() {
             uint32_t upper_dims_offset = 0;
             uint32_t multiplier = (input_height / tile_height) * num_pages_width;
 
-            for (int32_t i = num_dims - 3; i >= 0; i--) {
+            // Row-major Horner; must match get_upper_start_offset() in slice_device_operation.cpp
```
Aswinmcw: The cross-reference is ambiguous here: Quasar has its own `get_upper_start_offset` in `ttnn/cpp/ttnn/operations/experimental/quasar/slice/device/slice_device_operation.cpp` (namespace `ttnn::operations::experimental::quasar`), which is the helper this kernel actually has to match. Point at that path rather than the data_movement one.

Aswinmcw (approved):

Aswinmcw (approved):
All three points addressed. One leftover: the description still documents the dropped work — the Changes table lists `slice_device_operation.hpp/.cpp`, `gtests/sources.cmake` and `test_slice_upper_start_offset.cpp`, and the Test plan reports gtest results, but the diff is now just the two kernel loops plus the device pytest. Worth trimming so the body matches.

amorrisonTT (approved):

bbradelTT (approved):

VirdhatchaniKN:
Sanity passed : https://github.com/tenstorrent/tt-metal/actions/runs/31762480137

Issue #50682, opened 2026-07-22
ttnn.group_norm fused kernel silently produces wrong output for non-tile-aligned H*W


ashokkumarkannan1:
### Observed

- `ttnn.group_norm` (fused kernel) **silently produces numerically wrong output** when the per-sample flattened height `H*W` is not a multiple of the tile height (32). No exception is raised — the op runs and returns a plausible-looking tensor.
- The fused kernel computes each group's mean/variance over the **tile-padded** `H*W` rows (rounded up to the next multiple of 32) instead of the real `H*W` rows, so the padding rows contaminate the statistics. A correct affine-free group_norm normalizes every group to **mean 0, std 1**; on non-tile-aligned `H*W` the per-group mean drifts from 0 and std from 1.
- The shape is **not rejected**: a logically non-tile-aligned `H*W` is silently accepted and executed.
- Important measurement note: **PCC vs the correct reference stays ≈ 1.0**, because the contamination is a per-group *affine* (scale + shift) transform, and Pearson correlation is invariant to affine transforms. PCC therefore does **not** detect this bug — the per-group **mean/std** and the **absolute error** do.

### Error (no exception — silently wrong output)

`GroupNorm(num_groups=32, C=1024)` on Blackhole, vs `torch.nn.functional.group_norm` (tt-metal's own group_norm test tolerance is `atol=0.065`):

```
TILE-ALIGNED H*W (control - correct):
  H*W=256 (pad  0.0%) | per-group mean=0.0257 (want ~0)  std=0.9857 (want ~1)  maxErr_vs_torch=0.059  PCC=0.99999

NON-TILE-ALIGNED H*W (BUG - fused kernel reduces over tile-padding rows):
  H*W=200 (pad 10.7%) | per-group mean=0.1701 (want ~0)  std=0.8969 (want ~1)  maxErr_vs_torch=0.361  PCC=0.99999
  H*W= 40 (pad 37.5%) | per-group mean=0.5485 (want ~0)  std=0.8313 (want ~1)  maxErr_vs_torch=0.884  PCC=0.99993
```

- The error **scales monotonically with the padding fraction** (0% → within tolerance; 37.5% → `maxErr 0.884`), and the wrong output matches a "reduce-over-tile-padded-rows" reference to within bf16 noise — confirming the fused kernel reduces over the tile-padding.
[...]

PR #51159, opened 2026-07-27
Fix ttnn.group_norm fused kernel for non-tile-aligned H*W (#50682)

ashokkumarkannan1:
### Ticket
- https://github.com/tenstorrent/tt-metal/issues/50682
- tt-mlir https://github.com/tenstorrent/tt-mlir/pull/8935 (merged) works around this by decomposing non-tile-aligned `group_norm`; once this lands that gate can be relaxed, which is what unblocks the fused kernel for coqui/XTTS-v2.

### Summary

`ttnn.group_norm`'s fused kernel reduced over the **tile-padding rows** when the per-sample `H*W` is not a multiple of 32, so every group's mean/variance used the padded row count instead of the real one. No exception, and **PCC stayed ~1.0** — the contamination is a per-group affine transform, and Pearson correlation is invariant to those, so neither PCC nor the op model flagged it. At `H*W=200`: per-group mean 0.1701 (want 0), std 0.8969 (want 1), max abs err 0.361 vs `torch` — 5.5x the op's own 0.065 tolerance.

Fixed analytically, no extra pass. With `L = logical_hw`, `P = padded_hw`, `K = P/L - 1`:

| | Problem | Fix |
|---|---|---|
| **Mean** | padding contributes 0 to the sum, only the divisor is wrong | writer rescales the reduce scaler to `1/sqrt(reduce_factor_w * L/P)` (`sqrt` because the AVG/`REDUCE_SCALAR` LLK applies it twice). Exact |
| **Variance** | pass 2 centers each padding row to `0 - E[x]` and squares it, biasing by exactly `K*E[x]^2` | compute kernel subtracts `K*E[x]^2` before `(Var + eps)` |

Applied to all three two-pass factories (`mcast`, `no_mcast`, `sharded`), with the shared derivation in `GroupNormPadCorrection`. **Tile-aligned inputs are untouched**: `P == L` makes the flag a compile-time `false`, so the correction is `if constexpr`'d out, the three extra CBs are not allocated, and the writer takes the original path. Welford cannot express these shapes (it transposes `H*W` into tile columns and counts in tile units), so `ttnn::group_norm` routes non-tile-aligned Welford to the two-pass path and warns once per process.

### Results

Per-group **mean 0 / std 1** is the invariant that actually detects this bug. Post-fix, blackhole:

| `H*W` | K | \|mean\| | std |
|---|---|---|---|
| 224 (aligned control) | 0.00 | 0.0155 | 0.9886 |
[...]

PR #52924, opened 2026-08-12
[group_norm] Mask the tile-padding rows instead of back-correcting for them

### LINKED ISSUE #52685: ttnn.group_norm silently wrong for non-tile-aligned H*W when its input's tile padding is non-zero
Follow-up to #51159 (which fixed #50682).

**Two defects are recorded here.** The first is the subject of this issue; the second is a distinct bug in another op, included because it constrains the fix for the first.

1. **`ttnn.group_norm`** is silently wrong for non-tile-aligned `H*W` when its input's tile padding is non-zero — both interleaved and block-sharded, the latter catastrophically. *Fixed by the masking work tracked here.*
2. **`ttnn.fill_implicit_tile_padding`** corrupts BLOCK_SHARDED tensors. **Out of scope for this issue and NOT fixed by masking** — masking merely removes `group_norm`'s need to call it. Any other caller is still affected, so this needs its own fix.

### Observed

The #50682 correction reduces over the tile-padding rows and back-corrects for them **assuming those rows are zero**. Nothing in ttnn guarantees that, so `ttnn.group_norm` is silently wrong for non-tile-aligned `H*W` whenever its input's tile padding is non-zero. No exception is raised.

Non-zero padding is normal, and for some producers it is *correct*:

| producer | padding it leaves |
|---|---|
| `exp` | exactly `1.0` — this is `exp(0)`, the correct elementwise result |
| `softmax` | uniform non-zero weight |
| `slice` | the sliced-away source rows, still physically present |
| `reshape` | source spill / stale bytes (shape-dependent) |

Only host-upload paths (`from_torch`, `tilize_with_zero_padding`) reliably zero it — and **every existing group_norm test builds its input that way**, which is why the whole suite passes against this bug.

A producer-side fix is not viable: `reshape` deliberately only fills padding when the caller passes `pad_value` (a default-on fill was found to clobber logical lanes of zero-cost-aliased buffers — see the fill rules in `reshape.cpp`), and for `exp` there is nothing to fix.

### Error (no exception — silently wrong output)

**Interleaved**, `GroupNorm(num_groups=32, C=1024)`, `H*W=259` (pads to 288). Identical data, only the padding rows differ:

```
padding = 0.0      maxErr=0.0960  PCC=1.00000  worst per-group std ratio=0.972 (want 1.000)
padding = 0.5      maxErr=0.1903  PCC=0.99989  worst per-group std ratio=0.958
padding = 2.0      maxErr=0.7338  PCC=0.99999  worst per-group std ratio=0.834
padding = 7.0      maxErr=2.3203  PCC=0.99997  worst per-group std ratio=0.402
```

**PCC does not detect this** (stays >=0.9998 while per-group std collapses to 0.40) — the contamination is a per-group affine transform, the same blindness as #50682.

**Block-sharded is far worse**, `C=256`, `H*W=100` (pads to 128), grid 2x2. Dirty padding survives the interleaved -> sharded conversion intact:

```
padding = zeros    maxErr=     0.0630   worst per-group std ratio=0.976
padding = 7.0      maxErr=     2.2920   worst per-group std ratio=0.094
padding = 1.0      maxErr=786430.25     worst per-group std ratio=0.000
```

At `1.0` the corrected variance crosses zero and `rsqrt` explodes — a blow-up, not a degradation. **`1.0` is exactly what `exp()` leaves**, so the worst case is reachable from an ordinary `exp -> group_norm` graph.

Real-model impact — XTTS-v2 conditioning encoder (`GroupNorm(32, 1024)`, mel length 259), fed from `reshape`: encoder PCC **0.9996 -> 0.8223**, max err 9.18. Zero-filling that same tensor's padding restores it.

The Welford path is **not** affected: `group_norm` already forces `use_welford=false` for non-tile-aligned `H*W`.

### Expected

`ttnn.group_norm` must be correct for any legal input, independent of its producer's tile-padding contents.

The padding's contribution has to be either made zero or excluded — unknown values cannot be back-corrected. Excluding it (**masking rows `>= logical_hw` in the kernel, in BOTH accumulation passes**) is the only approach that also covers sharded, since it never touches the input tensor and so is layout-independent. It additionally removes the `Var - K*E[x]^2` subtraction that produces the 786430 above.

Masking only the variance pass is **not** sufficient: pass 1 sums the padding rows before centering, so the mean is contaminated too and rescaling the divisor cannot recover it.

The mechanism still needs design work: the correction lives in three program factories (`mcast`, `no_mcast`, `sharded`), and it is worth checking whether other kernels that reduce over a padded dim and back-correct share this assumption.

Whatever lands needs a test that supplies **dirty** padding — every current test zeroes it, so none of them guard this.

#### Defect 2 — `ttnn.fill_implicit_tile_padding` corrupts BLOCK_SHARDED tensors (separate; NOT fixed by masking)

Zero-filling the input is a viable fix for **interleaved** only, and this is why. Feeding a BLOCK_SHARDED tensor (`C=256`, `H*W=100`, grid 2x2) with padding seeded to 7.0 through `fill_implicit_tile_padding(t, 0.0)`:

```
                                      logical data err   padding after fill
BLOCK_SHARDED, before fill                     0.0000           7.000
BLOCK_SHARDED, after  fill(0)                  0.9961           7.000   <-- both contracts broken
INTERLEAVED control, after fill(0)             0.0000           0.000
```

It corrupts the logical region **and** fails to zero the padding, silently, while the interleaved equivalent is correct. This affects **any** caller of that public op, not just `group_norm`.

Masking does not fix this — it only removes `group_norm`'s need to call it. **This defect requires its own fix and must not be considered resolved when the masking work lands.**

### Steps (exact commands)

Save as `repro_group_norm_dirty_padding.py` and run with the tt-metal Python env:

```bash
python repro_group_norm_dirty_padding.py
```

```python
# ttnn.group_norm's non-tile-aligned correction is only correct when the implicit
# tile-padding rows are zero. Identical data, only the padding differs.
import math

import torch
import ttnn

C, HW, G = 1024, 259, 32
PAD = math.ceil(HW / 32) * 32  # 288


def pcc(a, b):
    a, b = a.flatten().float(), b.flatten().float()
    a, b = a - a.mean(), b - b.mean()
    d = a.norm() * b.norm()
    return float((a @ b) / d) if d > 0 else float("nan")


def main():
    torch.manual_seed(0)

    x = torch.randn((1, C, 1, HW), dtype=torch.bfloat16)
    ref = (torch.nn.functional.group_norm(x.float(), G)
           .permute(0, 2, 3, 1).reshape(1, 1, HW, C))
    xin = x.permute(0, 2, 3, 1).reshape(1, 1, HW, C)

    print(f"GroupNorm(num_groups={G}, C={C}), H*W={HW} -> tile-padded to {PAD}")
    for fill in (0.0, 0.5, 2.0, 7.0):
        t = ttnn.from_torch(xin, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
        t = ttnn.fill_implicit_tile_padding(t, fill)
        o = ttnn.group_norm(t, num_groups=G, inplace=False,
                            memory_config=ttnn.DRAM_MEMORY_CONFIG)
        ttnn.synchronize_device(dev)
        out = ttnn.to_torch(ttnn.from_device(o)).float().reshape(1, 1, HW, C)
        ratio = (out.reshape(1, HW, G, C // G).std(dim=(1, 3), unbiased=False) /
                 ref.reshape(1, HW, G, C // G).std(dim=(1, 3), unbiased=False))
        print(f"  padding = {fill:<4} maxErr={(out - ref).abs().max():7.4f}  "
              f"PCC={pcc(ref, out):.5f}  worst per-group std ratio={ratio.min():.3f}")
    ttnn.close_device(dev)


if __name__ == "__main__":
    main()
```

### Software Versions

- Ubuntu 24.04
- tt-metal @ `1a28ab18deefcb9bca72be573e081a31933a7b64` (includes #51159, merged as `ecc15a361295c31b579e2a8408e6425bc3f43948`)
- Reproduced on Wormhole B0 and Blackhole
- Surfaced from tt-xla XTTS-v2 bring-up (tenstorrent/tt-xla#5376) via tt-mlir's group_norm decomposition gate (tenstorrent/tt-mlir#8935). The gate relaxation that would route models onto the fused path is parked on branch `groupnorm-gate-relax` pending this issue.

ashokkumarkannan1:
### Ticket
Closes #52685 (follow-up on #50682 / #51159)

### Summary

- The non-tile-aligned correction added in #51159 reduces over the implicit tile-padding rows and back-corrects for them **assuming they hold zeros**. ttnn does not guarantee that, and no producer can reasonably be asked to.

- Measured by poisoning DRAM and reading each producer's output padded region:

| producer | H\*W=259 | H\*W=100 | verdict |
|---|---|---|---|
| `from_torch`, `tilize_with_zero_padding` | 0.0% | 0.0% | clean |
| `reshape` (2D→2D) | **4.0%** | 0.0% | dirty, shape-dependent |
| `reshape` (split dim) | **32.5%** | **41.7%** | dirty |
| `slice` | **100%** | **100%** | dirty (retains the sliced-away rows) |
| `exp` | **100%** | **100%** | exactly 1.0 — the *correct* elementwise result |
| `softmax` | **100%** | **100%** | uniform weight — also correct |
| `permute`, `typecast`, `concat`, `add`, `matmul`, `layer_norm` | 0.0% | 0.0% | clean |

- `exp` and `softmax` are not defects: `exp(0) = 1` is the right answer on zero padding. Producer-side filling is also known-harmful — `reshape` deliberately fills only when the caller passes `pad_value`, because a default-on fill "can be a zero-cost alias of the input buffer … clobbers logical lanes of aliased tensors" (see `reshape.cpp`'s fill rules). **Tile padding is de facto "don't care", so `group_norm` must own its precondition.**

- Every existing group_norm test builds its input via `from_torch` / `tilize_with_zero_padding` — the two producers that *do* zero padding — which is why the whole suite passed against this.

- **Fix: exclude the padding rows instead of back-correcting for them.** The input mask is already multiplied into both accumulation passes and its tile index depends only on the column, never the row-tile. So `create_group_norm_input_mask` gained an optional `rows_in_last_tile` that appends a **second set of groups** — the mask becomes `[1, 2*num_groups, 32, 32*block_wt]`, identical to the first with rows `>= logical_hw % 32` zeroed — and the kernels select it on the final row-tile of each batch.

### Results

- **Padding independence is bit-exact.** Output is *identical*, not merely within tolerance, for tile padding `0.0 / 7.0 / 1.0 / -3.5 / 0.5 / 100.0`, on both interleaved and block-sharded.

| case | max abs err | PCC |
|---|---|---|
| interleaved, tile-aligned control H\*W=128 | 0.047208 | 0.999981 |
| interleaved non-aligned H\*W=100 (any padding value) | 0.036495 | 0.999987 |
| block-sharded, tile-aligned control H\*W=128 | 0.047620 | 0.999981 |
| block-sharded non-aligned H\*W=100 (any padding value) | 0.050061 | 0.999980 |

Non-aligned now sits **at or below** its tile-aligned control.

- **The K-dependent error term is gone.** #51159's correction scaled with the padding fraction `K = padded/logical - 1`; that subtraction no longer exists. Re-measured on `wormhole_b0` at input mean 0, against tile-aligned controls:

| H\*W (K) | measured | aligned control |
|---|---|---|
| 16 (1.00) | 0.0447 | 0.0436 |
| 17 (0.88) | 0.0473 | 0.0436 |
| 24 (0.33) | 0.0415 | 0.0436 |
| 33 (0.94) | 0.0456 | 0.0488 |
| 50 (0.28) | 0.0507 | 0.0488 |
| 100 (0.28) | 0.0406 | 0.0549 |

- Error no longer tracks K at all — H\*W=16 at K=1.00 beats H\*W=50 at K=0.28 — so #51159's "only H\*W ≤ 16 exceeds tolerance" caveat is obsolete.

- **Downstream:** the XTTS-v2 conditioning encoder (`GroupNorm(32,1024)`, `H*W=259`, fed from `reshape`) recovered from PCC **0.822335 → 0.999100**, against a decomposed reference of 0.999552.

- **This flips a strict xfail that main added to pin this defect.** `test_group_norm_non_tile_aligned_garbage_padding_DRAM` was `@pytest.mark.xfail(strict=True)` with the reason *"…measured ~1.92 vs ~0.045. If this XPASSes, the padding rows are being masked and the precondition can be dropped from the kernel docs."* It now passes, and is converted to a plain test.

- **Perf:** block-sharded non-aligned is **108.9 µs/call vs 108.5 µs aligned** — no cost, provided the mask is built with `rows_in_last_tile`. Interleaved with an auto-created mask is 157.9 → 254.7 µs/call for non-aligned; that is the pre-existing per-call mask rebuild now carrying 2× the bytes, and callers in a hot loop should hoist the mask. A caller-supplied mask *without* `rows_in_last_tile` still works — group_norm derives the second set with a device-side multiply+concat — but pays ~4× per call, so both paths are test-covered.

### Caveats

- **Tile-aligned shapes are unchanged.** Everything is gated on `pad.active`; with it false, `mask_sets = 1` keeps `in_mask_CB_size` and `num_tiles_input_mask` byte-identical, and the kernel branch is an `if constexpr` rather than a foldable ternary, so aligned shapes emit the same instructions rather than relying on the optimizer.
- **Welford is unaffected**, worth stating explicitly since all three program factories are shared with it: `pad.active` is false whenever `use_welford` is true (`ttnn::group_norm` already routes non-tile-aligned Welford to the two-pass path); the Welford sharded compute kernel reads positional compile args up to index 25 while the new `has_row_mask` lands at 26–28; the Welford writers read runtime args up to 7 (sharded) and 9 (interleaved) while the new ones are 9 and 10. No Welford kernel file is touched.
- **New rejection:** the device op now requires mask `dim1 == 2*num_groups` for non-tile-aligned two-pass. Deliberate — fail loud rather than read past the mask. No in-tree caller can hit it: there are no direct `ttnn::prim::group_norm` callers and it is not exposed to Python.
- **`create_group_norm_input_mask` gained a trailing optional `rows_in_last_tile = 0`.** All 42 existing call sites pass at most 4 positional args and are unaffected.
- **Non-tile-aligned models cannot regress** — that path was silently wrong before whenever the producer left dirty padding, so only improvement is available there.

### Implementation notes

- Nothing writes to the input tensor, which is what makes this layout-independent. A host-side zero-fill was tried first and only works interleaved: `fill_implicit_tile_padding` addresses padding from the logical shape and corrupts a BLOCK_SHARDED tensor whose padding tail sits on the last M-core (took `test_group_norm_block_sharded_non_tile_aligned` from 6 passed to 3 failed). Zeroing inside the kernel is also unsafe — a sharded input CB can alias the caller's shard.

- Which core applies the row mask is a **per-core** property (only the last core row holds the batch tail when a batch is split across rows), and these compute kernels take no runtime args. So the **writer** selects the set from a per-core runtime offset and hands non-owning cores a duplicate of the normal set, keeping the compute-side switch compile-time and unconditional. Uniform rule across all three factories: `owns = (m_index % num_cores_per_batch) == num_cores_per_batch - 1`.

- The final row-tile is identified by its **global index within the batch**, not by "last out-block, last row": `num_out_blocks` need not divide `block_h`, and the last out-block can legitimately have zero rows.

- Removes the `Var - K*E[x]^2` back-correction and its three scratch CBs per path — that subtraction was a bfloat16 catastrophic cancellation which, on a block-sharded input with padding `1.0`, drove the variance negative and `rsqrt` to **786430**. The corrected reduce scaler (divisor `L`, not `P`) is now the only remaining correction.

### Coverage notes

Run locally on `wormhole_b0` — **1,281 passed, 0 regressions**:

| suite | result |
|---|---|
| `unit_tests/…/test_group_norm_DRAM.py` | 193 passed, 2 skipped |
| `unit_tests/…/test_group_norm.py` | 358 passed, 29 skipped |
| `nightly/…/test_group_norm{,_DRAM}.py` + `operations_compute_only` | 566 passed, 19 skipped |
| `docs_examples/test_normalization_examples.py` | 13 passed |
| `models/tt_dit/tests/unit/test_normalization.py` | 117 passed, 214 skipped |
| `models/experimental/retinanet/tests` | 22 passed |
| `tests/tt_eager/…/fallback_ops/test_group_norm.py` | 12 passed |

New coverage targets what the old suite structurally could not reach:
- interleaved dirty padding over **7 grid / batch / `num_out_blocks` combinations** — this is what exercises the per-core ownership rule and the empty-last-out-block case, and it hits all three program factories (no_mcast at grid 1x8, mcast at 2x8 and 4x8);
- sharded dirty padding over **4 grids × 4 padding values × 2 mask-construction paths**;
- a bit-exactness test asserting the output is *identical* across padding values rather than merely within tolerance.

### CI runs

Single-card SKUs only — galaxy / llmbox / loudbox, T3000 and multichip legs were deliberately left off, since this is an op-level change with no multi-device behaviour.

| workflow | result |
|---|---|
| Sanity tests | pass — https://github.com/tenstorrent/tt-metal/actions/runs/31594923582 |
| (Runtime) Sanity Tests | pass — https://github.com/tenstorrent/tt-metal/actions/runs/31594928368 |
| Nightly tt-metal L2 tests | pass — https://github.com/tenstorrent/tt-metal/actions/runs/31594935663 |
| Blackhole sanity tests | pass — https://github.com/tenstorrent/tt-metal/actions/runs/31671123487 |
| All Model Tests | pass — https://github.com/tenstorrent/tt-metal/actions/runs/31671126991 |
| (Single-card) Frequent model and ttnn tests | `bge_m3` x2 only — https://github.com/tenstorrent/tt-metal/actions/runs/31676768042 |

**All results match main.** *All Model Tests* covers the model consumers and is green, including 26 Stable Diffusion XL jobs (base / VAE / LoRA / refiner / unet-loop / img2img / tensor-parallel) across `wh_n150`, `wh_n300`, `bh_p150` and `bh_p150b_civ2`.

The only red is `bge_m3` on n150 and n300, and it fails identically **on main at the same commit**: main's nightly for `556a284a809` — the exact base this branch sits on — fails those two jobs and passes `efficientdetd0`, which is precisely what this branch reports. `bge_m3` references `group_norm` in zero files.

Nothing red is attributable to this PR.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:groupnorm-dirty-padding-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:groupnorm-dirty-padding-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:groupnorm-dirty-padding-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:groupnorm-dirty-padding-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:groupnorm-dirty-padding-fix)

### DIFF
--- a/tests/ttnn/unit_tests/operations/fused/test_group_norm.py
+++ b/tests/ttnn/unit_tests/operations/fused/test_group_norm.py
@@ -1896,3 +1896,90 @@ def test_group_norm_sharded_all_config(
         atol=atol,
         frobenius_threshold=frobenius_threshold,
     )
+
+
+@pytest.mark.parametrize("device_params", DEVICE_PARAMS_L1_SMALL_SIZE, indirect=True)
+@pytest.mark.parametrize(
+    # 1.0 is the exp() case, and the worst case for the back-correction this replaced (it drove the
+    # variance negative there, and rsqrt to 786430).
+    "padding_value",
+    [7.0, 1.0, -3.5, 0.5],
+)
+@pytest.mark.parametrize(
+    # grid_x is the M split. With grid_x=2 the batch spans two cores and only the second holds the
+    # padding row-tile; with grid_x=1 every core holds its own batch tail.
+    "grid_y, grid_x",
+    [(2, 1), (2, 2), (2, 4), (4, 2)],
+)
+@pytest.mark.parametrize(
+    # prebuilt_mask=True is what real code should do (free); False makes group_norm derive the
+    # second set per call (~4x slower), covered so the fallback cannot rot.
+    "prebuilt_mask",
+    [True, False],
+)
+def test_group_norm_sharded_dirty_padding(device, grid_y, grid_x, padding_value, prebuilt_mask):
+    # Mirror of test_group_norm_non_tile_aligned_garbage_padding_DRAM for the BLOCK-SHARDED path:
+    # supply non-zero tile padding and require the result to be independent of it.
+    #
+    # A host-side zero-fill of the input only worked interleaved, since fill_implicit_tile_padding
+    # corrupts a block-sharded tensor whose padding tail sits on the last M-core. In-kernel masking
+    # never touches the input, so it covers block-sharded too.
+    if device.core_grid.x < grid_x or device.core_grid.y < grid_y:
+        pytest.skip(f"device grid too small for {grid_x}x{grid_y}")
+
+    torch.manual_seed(0)
+    N, C, HW, G, padded = 1, 256, 100, 32, 128
+    grid_size = ttnn.CoreGrid(y=grid_y, x=grid_x)
+
+    real = torch.rand((N, 1, HW, C), dtype=torch.bfloat16)
+    ref = torch.nn.functional.group_norm(real.view(N, HW, C).permute(0, 2, 1).reshape(N, C, 1, HW).float(), G)
+    ref = ref.permute(0, 2, 3, 1).reshape(N, 1, HW, C)
+
+    buf = torch.zeros((N, 1, padded, C), dtype=torch.bfloat16)
+    buf[:, :, :HW, :] = real
+    buf[:, :, HW:, :] = padding_value  # dirty padding, as reshape / slice / exp would leave
+
+    tt = ttnn.from_torch(
+        buf, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=ttnn.DRAM_MEMORY_CONFIG
+    )
+    tt = ttnn.reshape(tt, ttnn.Shape([N, 1, HW, C]), ttnn.Shape([N, 1, padded, C]))
+
+    grid_coord = ttnn.CoreCoord(grid_size.x - 1, grid_size.y - 1)
+    shard_grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), grid_coord)})
+    shard_spec = ttnn.ShardSpec(shard_grid, (padded // grid_size.x, C // grid_size.y), ttnn.ShardOrientation.COL_MAJOR)
+    sharded_mem_config = ttnn.MemoryConfig(
+        ttnn.types.TensorMemoryLayout.BLOCK_SHARDED, ttnn.types.BufferType.L1, shard_spec
+    )
+    tt = ttnn.to_memory_config(tt, sharded_mem_config)
+
+    mask = ttnn.to_device(
+        ttnn.create_group_norm_input_mask(
+            C,
+            G,
+            grid_size.y,
+            ttnn.DataType.BFLOAT8_B,
+            rows_in_last_tile=(HW % 32) if prebuilt_mask else 0,
+        ),
+        device,
+    )
+    out = ttnn.group_norm(
+        tt,
+        num_groups=G,
+        input_mask=mask,
+        memory_config=sharded_mem_config,
+        core_grid=grid_size,
+        inplace=False,
+    )
+    out = ttnn.to_torch(ttnn.from_device(ttnn.to_memory_config(out, ttnn.DRAM_MEMORY_CONFIG))).float()[:, :, :HW, :]
+
+    max_abs_err = (out - ref).abs().max().item()
+    pcc = torch.corrcoef(torch.stack([out.flatten(), ref.flatten()]))[0, 1].item()
+    logger.info(
+        f"block-sharded dirty padding grid={grid_x}x{grid_y} padding={padding_value} "
+        f"prebuilt_mask={prebuilt_mask}: max_abs_err={max_abs_err} pcc={pcc}"
+    )
+    assert max_abs_err < 0.08, (
+        f"max abs error {max_abs_err} with tile padding = {padding_value} on a {grid_x}x{grid_y} "
+        f"block-sharded grid; group_norm must be independent of its padding (see #52685)"
+    )
+    assert pcc > 0.999, f"pcc {pcc} with tile padding = {padding_value} (see #52685)"
--- a/tests/ttnn/unit_tests/operations/fused/test_group_norm_DRAM.py
+++ b/tests/ttnn/unit_tests/operations/fused/test_group_norm_DRAM.py
@@ -701,17 +701,9 @@ def test_group_norm_non_tile_aligned_shifted_input_DRAM(device, input_shift, max
 
 
 @pytest.mark.parametrize("device_params", DEVICE_PARAMS_L1_SMALL_SIZE, indirect=True)
-@pytest.mark.xfail(
-    strict=True,
-    reason="tt-metal #50682: the analytical correction requires the tile-padding rows to hold "
-    "exact zeros (every TILE-layout producer supplies them). With garbage in the padding the "
-    "reduce ingests it and the result is meaningless -- measured ~1.92 vs ~0.045. If this XPASSes, "
-    "the padding rows are being masked and the precondition can be dropped from the kernel docs.",
-)
 def test_group_norm_non_tile_aligned_garbage_padding_DRAM(device):
-    # Documents (as a strict xfail) the precondition the correction depends on. The op reduced over
-    # the padding rows before the fix too, so the dependence is not new -- but the fix is what makes
-    # "correct for non-tile-aligned H*W" a claim, and this is its boundary.
+    # Was a strict xfail while group_norm reduced over the tile padding (measured ~1.92 vs ~0.045);
+    # the padding rows are masked out now, so it is a plain test.
     if device.core_grid.y == 7:
         pytest.skip()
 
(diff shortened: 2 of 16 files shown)


kamalrajkannan78 (approved):

bbradelTT (approved):
For "fill_implicit_tile_padding addresses padding from the logical shape and corrupts a BLOCK_SHARDED tensor whose padding tail sits on the last M-core", is this a bug? If yes, please file an issue.

ashokkumarkannan1:
Yes, it is a bug — filed as #53110.

Reproduced on current main (`556a284a809`), same tensor and same call, only the memory config differing:

```
INTERLEAVED     logical region max err = 0.0000   padding non-zero =   0.0%
BLOCK_SHARDED   logical region max err = 0.9961   padding non-zero =  50.0%  absmax = 7.0000
```

So on block-sharded it is worse than just failing to zero the padding — it also overwrites the logical region. Likely cause is that `fill_implicit_tile_padding` reshapes to rank 3 and dispatches `prim::fill_pad`, which addresses the padding from the *logical* shape; on a block-sharded tensor the padding tail sits on the last M-core, so those offsets land on the wrong cores.

This PR does not depend on that op — masking never writes to the input tensor — but the defect affects any caller, so it is filed separately with a standalone reproducer.

PR #52931, opened 2026-08-12
[Bug Fix] untilize sub_core_grids deadlock for tall tensors

abdulla-TT:
### PR Category
Bug fix

### Summary
Fixes item 3 of #51216 — a host/kernel contract mismatch that deadlocks `ttnn.untilize(..., sub_core_grids=...)` whenever the input tensor is taller than one tile row (`ntiles_per_column > 1`).

**Root cause.** `UntilizeMultiCoreSubCoreGridsProgramFactory` shares its writer kernel (`writer_unary_stick_layout_split_rows_interleaved_parallel_columns`) with the `parallelize-column` factory. That kernel loops `(num_sticks / TILE_HEIGHT) * num_tiles_per_core` times calling `wait_front(1)`, but reader/compute only push `ntiles_per_core` tiles per core. The factory passed `nsticks = ntiles_per_column * TILE_HEIGHT`, so when `ntiles_per_column > 1` the writer waited for `ntiles_per_column * ntiles_per_core` tiles from a pipeline that only ever produced `ntiles_per_core` → device hang. At the same time, `offset_within_stick` accumulated by a factor of `ntiles_per_column` too much per core, corrupting all write addresses.

The sibling `parallelize-column` factory already defends against this: `get_pf_type()` only selects it when `num_tiles_per_col == 1` (comment: *"currently multi_core parallelization on column only works for single tile height tensors"*). The `sub_core_grids` path lacked any such gate.

**Fix.** Add `TT_FATAL(ntiles_per_column == 1)` in `validate_on_program_cache_miss` inside the `sub_core_grids` check block, mirroring the identical precondition that already protects the `parallelize-column` path. This converts the silent hang into a clear `RuntimeError` and is consistent with every real-world caller (LLM inference decode steps), which always operate on single-tile-height tensors.

**Test.** Added two tests in `tests/ttnn/unit_tests/operations/data_movement/test_untilize.py`:
- `test_untilize_sub_core_grids_multi_tile_height_rejected` — parametrized over two tall shapes (`[1,1,64,1024]` and `[1,1,96,512]`). **Hangs before this fix, raises `RuntimeError` after.**
- `test_untilize_sub_core_grids_single_tile_height` — confirms the valid single-tile-row path (`[1,1,32,1024]`) still passes.

### Notes for reviewers
- Behavioral change is a single `TT_FATAL` guard; no kernel or factory logic is modified.
- The error message text matches the `expect_error` fixture call in the test (as required by the CI log triager).
- Remaining items 1-4 of #51216 tracked separately.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/untilize_subcoregrids_deadlock_fix)


ttnn/cpp/ttnn/operations/data_movement/untilize/device/untilize_device_operation.cpp
```diff
@@ -115,6 +115,21 @@ void UntilizeDeviceOperation::validate_on_program_cache_miss(
...
             operation_attributes.use_multicore,
             "sub_core_grid implementation only supported when use_multicore flag argument is set to true");
+        // The sub_core_grids factory uses the same writer kernel as the parallelize-column
+        // factory (writer_unary_stick_layout_split_rows_interleaved_parallel_columns). That kernel
+        // loops `(num_sticks / TILE_HEIGHT) * num_tiles_per_core` times calling wait_front(1), but
+        // reader/compute only push `ntiles_per_core` tiles. When ntiles_per_column > 1 the counts
+        // disagree and the writer deadlocks. The parallelize-column factory guards against this with
+        // the same precondition (see get_pf_type: "currently multi_core parallelization on column
+        // only works for single tile height tensors").
+        const uint32_t ntiles_per_column = tensor_height / TILE_HEIGHT;
```
Aswinmcw: The quasar clone has the identical defect and this guard doesn't reach it: `experimental/quasar/untilize/device/factories/untilize_multi_core_sub_core_grids_program_factory.cpp:162` passes `nsticks_per_core = ntiles_per_column * TILE_HEIGHT` to the same writer, and the `sub_core_grids` block in its `validate_on_program_cache_miss` (lines 107-117) has no equivalent precondition. Worth mirroring the FATAL there, or filing a follow-up so #51216 item 3 isn't half-closed.

Aswinmcw (approved):

sjameelTT (approved):

VirdhatchaniKN (approved):

bbradelTT (approved):

PR #52969, opened 2026-08-12
MiniMax-H3 audio decode: fix conv_pre under T-sharding, 1.25 s to 281.6 ms at unchanged 49.45 dB

rouzbehTT:
## MiniMax-H3 audio decode: 1.25 s → 281.6 ms at unchanged accuracy

Decodes 207 latents (5.17 s of audio, batch 2) in **281.6 ms** at **49.45 dB PSNR vs the CPU
reference** — the same accuracy as the single-device path, per-clip identical. Since only 6.5 chips worth of computation is happening, we maybe able to do `5.17 s x 32÷(207÷32) =  25.5 s` worth of audio still in `~280 ms`

| config | latency | PSNR vs CPU |
|---|---|---|
| single device (before) | 1.25 s | 49.45 dB |
| **4x8 mesh, `t_factor=8 axis=1`, traced** | **0.2816 s** | **49.45 dB** |

    CVD_MESH=4x8 CVD_T_FACTOR=8 CVD_MESH_AXIS=1 CVD_TRACED=1 \
      python models/tt_dit/tests/models/minimax_h3/audio_perf/cpu_vs_device.py

T-parallel sharding and tracing already existed on `cglagovich/minimax-h3`; what
was missing was a correctness fix that made the sharded path usable, and the measurement to prove it.

### The bug fix

`Vocoder.conv_pre` (2048 → 1024, k=7) — by far the widest conv in the decode — **returned uninitialized
memory under T-sharding**: absmax 2.6e+11 against a reference absmax of 1.391, i.e. −204 dB. That made
*every* T-sharded decode return wrong audio (−10.1 dB at factor 4, −11.0 dB at factor 8), which had been
recorded as the sharded path being broken in general.

It is specific to that one conv. A sharded-vs-unsharded sweep found `1024→512 k=7`, `256→256 k=7` and
`64→64` at kernels 1/3/7 and dilation 5 all **bit-exact** under sharding, and `_partition_t` /
`_t_neighbor_pad` verified exact at the affected shape. The cause is that conv3d's blocking config is
derived at `__init__` from the unsharded shape, and this is the shape whose size trips the DRAM
auto-slice fallback.

`conv_pre` now runs unsharded on the full sequence, with T partitioned immediately after — the same
pattern `ConvTranspose1dViaConv3d` already uses for its inner conv ("Inner conv stays UNSHARDED"). It is
close to free: `conv_pre` runs on the *input* sequence, 1/800th of the final waveform length. Guarded so
channel-TP keeps its original ordering, where `conv_pre`'s own gather is what reconstructs full C_in.

A second, independent fix in `conv3d.py`: the last-resort blocking used `C_in_block = in_channels`, which
overflows L1 outright for a 2048-channel conv in bf16 ("circular buffers grow to 1981312 B" against
1572864). It now falls back to the tuned fp32 blocking, which cannot fail to fit because a bf16 tile is
half the size.

### Test changes, in `test_audio_decode_t_parallel`

* **`KNOWN_BROKEN` is now empty.** It held `(8, 1)`, which *silenced the PSNR assert on the very
  configuration that was broken* — the test passed while returning wrong audio. Both factors now measure
  78.7 dB against the unsharded path and the assert is enforced.
* **A baseline-substitution guard.** `assert baseline_ran` catches "everything failed" but not "factor 1
  died and factor 4 was promoted to baseline, scoring PSNR inf against itself" — which is how a −10.1 dB
  configuration read as correct.

Suite result: **25 passed** (~11 min). It was 24 passed / 1 failed, with the failure described as
inherited; it was not inherited, and it is fixed.

`test_audio_decode_t_parallel` takes ~9 minutes on a 4x8 mesh and `pytest.ini` sets `timeout = 300`, so
run it with `--timeout 2400` or it is killed and looks like a hang.

### Why tracing matters here

Tracing is worth ~1.04x on a single device, which is why it had been recorded as a dead end. On 32 chips
it is worth ~3x: `factor=1` on a 32-chip mesh costs 1.4409 s against 1.0980 s on one chip — **+343 ms of
pure 32-wide dispatch** — and trace removes ~331 ms of it. Both readings are right about different
configurations, so any untraced multi-chip measurement understates its configuration by ~3x.

Note the 281.6 ms uses **8-way parallelism**; `t_factor=8 axis=1` shards the 8-wide mesh axis and
replicates across the 4-wide one. That cannot be improved at this sequence length — T pads to
`32 * factor`, so factor 8 and factor 32 both give 32 rows per chip and therefore the same latency. The
idle chips are worth ~4x *throughput* (concurrent clips), not lower latency.

### `audio_perf/`

Three files: `cpu_vs_device.py` (latency and PSNR against the CPU reference, PASS/FAIL, plus WAVs so the
difference can be heard), `decode_bench.py` (medians — timing one decode per process measures build and
JIT noise instead) and a README with the reproduce command.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rouzbeh/minimax-audio)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rouzbeh/minimax-audio)


rouzbehTT:
Superseded by #53073, which targets `main`.

Coleman's work moved to `cglagovich/minimax-h3-functional`, so this branch's base (`cglagovich/minimax-h3`) is no longer the integration point. #53073 is the same work re-derived against `main`, which required more than a rebase:

- **main replaced the env-var levers with constructor parameters and flipped the defaults to the accurate path** (`split_mode="full"`, `tap_matmul=True`, `prefer_mac=True`). The perf work here was written against the fast defaults.
- That default makes `_forward_tap_matmul` live, and it pads shard boundaries locally instead of through the halo — **a second T-sharding bug that this MR's `conv_pre` fix alone would not have covered**. #53073 fixes both.
- The `conv3d.py` blocking fix here is **obsolete**: main's `register_h3_audio_blockings` + `max_c_in_block` solves the same L1 overflow properly.
- The **281.6 ms at 49.45 dB** result was measured with the fast levers and does not carry over to main's accurate defaults, so #53073 quotes no performance numbers until they are re-measured there.

Nothing is lost — the branch `rouzbeh/minimax-audio` still exists at the state this PR described.

PR #52977, opened 2026-08-12
[Cleanup] Remove data_types.hpp from the tt-metalium API surface

riverwuTT:
### PR Category
Cleanup

### Summary

Runtime team is performing surface area reduction, the goal of the project is to cleanup the internal / stale headers and symbols out of the API folder to cleanup it's surface area. This effort enables further cleanups for our API.

`tt-metalium/data_types.hpp` has been a pure forwarding header since #38965 (merged 2026-03-06).

This PR removes this header to reduce bloat in the api folder.

### Notes for reviewers

- All use sites have been udpated.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/rm-data-types-hpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/rm-data-types-hpp)

PR #52999, opened 2026-08-12
ci: use system SFPI toolchain in CI-internal builds to avoid redundant network downloads

nsextonTT:
## Summary

CMake only skips downloading the SFPI toolchain from `github.com/tenstorrent/sfpi/releases` (via `FetchContent`) when `-DTT_USE_SYSTEM_SFPI=ON` is passed, in which case it uses the copy already staged at `/opt/tenstorrent/sfpi` in the docker image instead. Before this PR, no CI workflow set that flag (only `setup.py`'s pip/wheel path did), so every CMake-driven CI build silently re-downloaded the identical, version-pinned SFPI tarball on every run. That download flaking on a proxy blip is what originally broke [run 31637228026](https://github.com/tenstorrent/tt-metal/actions/runs/31637228026/job/94251823873?pr=52977) (PR #52977).

This PR wires `-DTT_USE_SYSTEM_SFPI=ON` into CI-internal builds only, and fixes the fallout that surfaced along the way.

## What changed

**Core plumbing:**
- `build_metal.sh`: new `--use-system-sfpi` flag (mirrors the existing `--ttnn-shared-sub-libs` pattern).
- `clang-tidy-reusable.yaml`: passes `-DTT_USE_SYSTEM_SFPI=ON` directly to its `cmake --preset` call (never packages, so no downstream impact).
- `build-artifact.yaml`: new opt-in `use-system-sfpi` `workflow_call`/`workflow_dispatch` boolean input (**default `false`**, preserving the original download behavior for every caller unless explicitly opted in — this workflow is called from ~70 other workflows, so a global default flip was not safe).

**Opted in** (CI-internal only — packages only ever consumed inside docker images that stage SFPI):
- `pr-gate.yaml`: `asan-build`
- `merge-gate.yaml`: `build`, `build-tsan`, `build-asan`
- `sanity-tests.yaml`: `build-artifact`

**Explicitly left off:**
- `release-build-test-publish.yaml`'s three build jobs (feeding `package-and-release.yaml`) are explicitly set to `use-system-sfpi: false`, with a comment. Released packages get installed on arbitrary external machines with no `/opt/tenstorrent/sfpi` to fall back to, so they must bundle the SFPI JIT toolchain themselves.
- Every other `build-artifact.yaml` caller not listed above is unaffected (stays on the default).

**Packaging fallout, fixed:**
`CMakeLists.txt:333`'s `install(DIRECTORY runtime/sfpi ...)` — which populates the `jit-build` CPack component (`tt-metalium-jit` .deb) — only runs `if(NOT TT_USE_SYSTEM_SFPI)`. `tt_metal/jit_build/build.cpp:148` already has a documented runtime fallback to `/opt/tenstorrent/sfpi` when that package isn't bundled, so the missing `.deb` is harmless **in containers that stage SFPI**. A repo-wide grep found four workflow call sites that hard-required `tt-metalium-jit_*.deb` to exist:

| File | Fix |
|---|---|
| `smoke.yaml` | tolerant glob (`shopt -s nullglob`) |
| `basic.yaml` | tolerant glob |
| `sdk-examples.yaml` | tolerant glob (needed the Dockerfile fix below first) |
| `ttsim-sanity-tests-impl.yaml` | tolerant glob (needed the Dockerfile fix below first) |

`sdk-examples.yaml`/`ttsim-sanity-tests-impl.yaml` run in `basic-dev-docker-image`, which — unlike `ci-build`/`ci-test`/`dev` — deliberately didn't stage SFPI (only its sibling `basic-ttnn-runtime` stage did), so the runtime fallback couldn't apply there and a workflow-level glob fix alone wasn't enough. Fixed in `dockerfile/Dockerfile.basic-dev` by moving the SFPI copy step from `basic-ttnn-runtime` up into the shared `base` stage, so every `basic-dev`-derived image inherits it. Also brought all of `Dockerfile.basic-dev`'s tool-layer `COPY` instructions (uv, cmake, openmpi, ccache, sfpi) in line with the `--link` pattern already used in the main `Dockerfile`, for cache-layer independence.

## Test plan
- [x] `bash -n build_metal.sh` passes.
- [x] YAML syntax validated for every workflow file touched.
- [x] Traced Dockerfile / docker-bake stage inheritance for `ci-build`, `ci-test`, `dev`, `basic-dev`, `basic-ttnn-runtime`.
- [x] `runtime-smoke-tests` + `ttnn-smoke-tests` passed on this PR with `tt-metalium-jit` legitimately absent ([run 31647896511](https://github.com/tenstorrent/tt-metal/actions/runs/31647896511?pr=52999)).
- [ ] CI re-running with the full `merge-gate`/`sanity-tests`/`basic-dev` changes — confirming clean end-to-end.

🤖 Generated with BrAIn

### DIFF
--- a/.github/workflows/basic.yaml
+++ b/.github/workflows/basic.yaml
@@ -118,7 +118,12 @@ jobs:
         timeout-minutes: 10
         run: |
           # Ideally only ${{ inputs.product }}-validation, but APT doesn't resolve dependencies from files on disk without being told about them.
-          apt install -y ./pkgs/tt-metalium_*.deb ./pkgs/tt-metalium-jit_*.deb ./pkgs/tt-metalium-validation_*.deb ./pkgs/${{ inputs.product }}_*.deb ./pkgs/${{ inputs.product }}-validation_*.deb
+          # tt-metalium-jit is optional: TT_USE_SYSTEM_SFPI builds don't produce it, since the
+          # dev/ci-test containers this workflow runs in already have /opt/tenstorrent/sfpi for
+          # the JIT-build runtime fallback (see tt_metal/jit_build/build.cpp).
+          shopt -s nullglob
+          pkgs=(./pkgs/tt-metalium_*.deb ./pkgs/tt-metalium-jit_*.deb ./pkgs/tt-metalium-validation_*.deb ./pkgs/${{ inputs.product }}_*.deb ./pkgs/${{ inputs.product }}-validation_*.deb)
+          apt install -y "${pkgs[@]}"
 
       - name: Set up env vars for watcher
         if: ${{ inputs.enable-watcher == true }}
--- a/.github/workflows/build-artifact.yaml
+++ b/.github/workflows/build-artifact.yaml
@@ -648,7 +648,6 @@ jobs:
             ${args_fixme}
             "--enable-ccache"
             "--configure-only"
-            "--use-system-sfpi"
           )
 
           echo "Build tracy: ${{ inputs.tracy }}"
--- a/.github/workflows/build-artifact.yaml
+++ b/.github/workflows/build-artifact.yaml
@@ -48,6 +48,11 @@ on:
         type: boolean
         default: true
         description: "Make resulting debian packages available in the workflow"
+      use-system-sfpi:
+        required: false
+        type: boolean
+        default: false
+        description: "Use the SFPI toolchain already staged in the docker image instead of downloading it. Only safe for internal CI consumption: skips producing the tt-metalium-jit package, so leave false for release/publish builds whose packages need to be self-contained."
       build-ttnn-import-artifact:
         required: false
         type: boolean
@@ -222,6 +227,11 @@ on:
         type: boolean
         default: false
         description: "Package the wheel from the compiled tree (no second compile) instead of the manylinux build-wheel job; in-family CI consumers only"
+      use-system-sfpi:
+        required: false
+        type: boolean
+        default: false
+        description: "Use the SFPI toolchain already staged in the docker image instead of downloading it. Only safe for internal CI consumption: skips producing the tt-metalium-jit package, so leave false for release/publish builds whose packages need to be self-contained."
       architecture:
         required: false
         type: string
@@ -648,8 +658,12 @@ jobs:
             ${args_fixme}
             "--enable-ccache"
             "--configure-only"
-            "--use-system-sfpi"
           )
+
+          echo "Use system SFPI: ${{ inputs.use-system-sfpi }}"
+          if [ "${{ inputs.use-system-sfpi }}" = "true" ]; then
+            build_args+=("--use-system-sfpi")
+          fi
 
           echo "Build tracy: ${{ inputs.tracy }}"
           if [ "${{ inputs.tracy }}" = "false" ]; then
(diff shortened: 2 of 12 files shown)


nsextonTT:
The `build-artifact.yaml` change caused a real regression, caught by [runtime-smoke-tests failing on this PR](https://github.com/tenstorrent/tt-metal/actions/runs/31644360016/job/94280813563?pr=52999):

`TT_USE_SYSTEM_SFPI=ON` skips `CMakeLists.txt:333`'s `install(DIRECTORY runtime/sfpi ...)` step, which populates the `jit-build` CPack component (the `tt-metalium-jit` .deb). Downstream jobs that `apt install ./pkgs/tt-metalium-jit_*.deb` then fail because no such package exists — `build-artifact.yaml`'s packages run on machines that don't have SFPI pre-installed system-wide, so they need the toolchain bundled regardless of what the *build host* had.

Reverted the `build-artifact.yaml` part of this change in e6f50050bb. **This PR now only fixes `clang-tidy-reusable.yaml`**, which is safe since it never runs CPack/packaging.

nsextonTT:
Follow-up: re-enabled \`--use-system-sfpi\` for \`build-artifact.yaml\` (commit d4aab9a686). The actual blocker turned out to be a stricter-than-necessary requirement in \`smoke.yaml\`, not a packaging gap — see updated PR description for the full explanation. \`tt-metalium-jit\` is now treated as optional in the "Install packages" step, since \`tt_metal/jit_build/build.cpp\` already falls back to \`/opt/tenstorrent/sfpi\` at runtime, which the smoke-test container has.

tdowdallTT (approved):

nsextonTT:
verified \`build-artifact.yaml\` is called from ~70 workflows, including the release pipeline (\`release-build-test-publish.yaml\`, invoked from \`package-and-release.yaml\`). My earlier commits added \`--use-system-sfpi\` unconditionally inside \`build-artifact.yaml\`, which silently affected every one of those callers — including release builds, which **must** bundle the SFPI JIT toolchain since they're installed on machines with no \`/opt/tenstorrent/sfpi\` to fall back to.

Fixed by turning it into a proper opt-in \`workflow_call\`/\`workflow_dispatch\` input (default \`false\`, i.e. original download behavior). Explicitly set \`true\` only on \`pr-gate.yaml\`'s \`asan-build\` (the job this whole fix targets, fully verified end-to-end this session). Explicitly set \`false\` (documenting intent) on all three \`release-build-test-publish.yaml\` build calls. Every other caller — including \`merge-gate.yaml\`'s analogous \`build\`/\`build-tsan\`/\`build-asan\` jobs — is left on the default for now since it hasn't been verified in this round; can be opted in later the same way once confirmed.

nsextonTT:
extended \`use-system-sfpi: true\` to the remaining CI-internal builds — \`merge-gate.yaml\`'s \`build\`/\`build-tsan\`/\`build-asan\` and \`sanity-tests.yaml\`'s \`build-artifact\` job. Same rationale as \`pr-gate.yaml\`'s \`asan-build\`: all consumers run in \`ci-build\`/\`ci-test\`/\`dev\`/\`basic-dev\`/\`basic-ttnn-runtime\` images, which now all stage SFPI. \`release-build-test-publish.yaml\` stays explicitly \`false\` — untouched.

Commit 05372e6bc7.

Issue #52909, opened 2026-08-12
[Bounty $5,000] Perf - top-k/sort/moe/sampling index tile is written with 1024 scalar stores per tile on a DM core


kinginu:
**Component / Area:** kernels

**Issue Type:** Performance

## Observed

`generate_index_tile()` writes **1024 scalar values per tile** on a data-movement core, one store at a time, in four copies of the same loop.

`ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/dataflow/topk_dataflow_common.hpp:48`:

```cpp
for (uint32_t i = 0; i < tile_faces; ++i) {
    for (uint32_t j = 0; j < tile_faces; ++j) {
        for (uint32_t k = 0; k < face_size; ++k) {
            for (uint32_t l = 0; l < face_size; l++) {
                const T value = l + face_size * j + w;   // no dependence on k
                ptr[count] = value;
                count++;
            }
        }
    }
}
```

The value depends on the column within the face and on the face column, not on the row. All 16 rows of a face are identical, the two lower faces repeat the two upper ones, and the 1024-entry tile holds 32 distinct values. Measured cost of the stores: **4.1 µs per tile**, about 5 cycles per 16-bit store at 1.35 GHz.

The construction that avoids this is already in the tree. `ttnn/cpp/ttnn/operations/experimental/reduction/deepseek_grouped_gate/device/kernels/dataflow/writer_deepseek_grouped_gate.cpp:30` writes one face line with the RISC and replicates it with local NoC reads, carrying the store-visibility guard #50374 added for exactly that loop-back. `ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/moe_grouped_topk/device/kernels/dataflow/writer_moe_grouped_topk.cpp:9` is the same construction without that guard. The four shipping copies have neither:

* `ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/dataflow/topk_dataflow_common.hpp:27`
* `ttnn/cpp/ttnn/operations/data_movement/sort/device/kernels/dataflow/sort_dataflow_common.hpp:11`
* `ttnn/cpp/ttnn/operations/reduction/moe/device/kernels/dataflow/reader_create_index_tensor.cpp:18`
* `ttnn/cpp/ttnn/operations/reduction/sampling/device/kernels/dataflow/reader_values_indices_tensor.cpp:21`

[...]

PR #48886, opened 2026-07-03
[Performance] Enable force_argmax fast-path on non-Galaxy (single-chip P150/N150)

morjun:
### Ticket
Closes #48885

### Problem
On single-chip Blackhole P150, Llama-3.1-8B-Instruct batch-1 greedy decode runs at ~22 t/s/u, well below what the same board achieves on the pre-#31046 path (~35 t/s/u, e.g. `v0.64.0-dev20251030`).

### Root cause
On-device sampling (#31046) routes greedy decode (`temperature=0` → rewritten to `top_k=1`) through the full sampling op (3 all-gathers + top-k + softmax + RNG over the padded vocab), which dominates per-token decode latency. The cheaper in-trace argmax fast-path (1 all-gather + on-device argmax) already exists but `allow_force_argmax` was enabled only for Galaxy: the non-Galaxy `default_sampling_force_argmax` in `model_config.py` set `"allow_force_argmax": False`, so single-chip P150 never took the fast path for greedy decode.

### Change
Set `allow_force_argmax = True` in the non-Galaxy `default_sampling_force_argmax` config (Linear topology, 1 link).

### Results (P150, Llama-3.1-8B-Instruct, batch-1)
~22 → ~38.5 t/s/u steady (1st/128th token 38.7/38.5), exceeding the v0.64 reference (~35.5). No output change — `temperature=0` is argmax either way.

### Test plan
`pytest models/tt_transformers/demo/simple_text_demo.py -k "batch-1 and performance"` on a single P150 (Blackhole).

PR #35153, opened 2025-12-29
Implement skip list for device sampling

pprajapatiTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/34763

### Problem description
Device sampling is not working with Mistral7B. For now skip sampling on device for Mistral7B. This needs proper fixing later.

### Checklist

- [ ] [All post-commit tests](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml?query=branch:pprajapati/dev_sampling_skip)
- [ ] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:pprajapati/dev_sampling_skip)
- [ ] [cpp-unit-tests](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pprajapati/dev_sampling_skip)
- [ ] New/Existing tests provide coverage for changes

Issue #34763, opened 2025-12-19
Investigate mistral7b PCC and repeat batch errors


astancovTT:
Mistral7b is failing on (Single-card) Demo tests (no perf)
examples:
https://github.com/tenstorrent/tt-metal/actions/runs/20278067177/job/58232833711
https://github.com/tenstorrent/tt-metal/actions/runs/20346619208/job/58462348571
It needs to be investigated

env: N150

skrsticTT:
N300 mistral7b is failing as well https://github.com/tenstorrent/tt-metal/actions/runs/20478791650/job/58848650977

pprajapatiTT:
I have narrowed down the failure of mistral7b to this commit `58ac27a91255db122255e3c30278b660b816c8f1` . `device_sampling_params` seems to have some issues for mistal7b. further debug WIP.

pprajapatiTT:
I have a hack for now, https://github.com/tenstorrent/tt-metal/pull/35153, to unblock people. It worked locally. I am running CI also.

I would check with @sraizada-tt , who is original author of this feature, for proper fix.

pprajapatiTT:
CI: https://github.com/tenstorrent/tt-metal/actions/runs/20602482936 passing with the hack. Seems like sampling on device was not tested for Mistral7b.

pprajapatiTT:
@sraizada-tt is looking into the actual fix.

PR #53001, opened 2026-08-12
[Bug fix] sampling: TTSampling hands ttnn.topk an indices_tensor half the width of the logits

### LINKED ISSUE #52928: [Bug] Single-device TTSampling returns wrong tokens: the iota indices_tensor it hands ttnn.topk is half the width of the logits
**Component / Area:** models / sampling

**Issue Type:** Bug

## Observed

On a 1×1 mesh `TTSampling` returns tokens that are **not the model's argmax** whenever sampling takes the top-k path. Through `TTSampling.decode_forward` with `k=1`, 8 trials of 32 users, counting a token wrong only when its logit is strictly below the row maximum:

| padded vocab | half | top-k factory | wrong tokens |
|---|---|---|---|
| 1024 | 512 | single-core | 0/256 |
| 4096 | 2048 | single-core | 0/256 |
| 32768 | 16384 | **multi-core** | **126/256**, some outside the vocabulary |
| 65536 | 32768 | **multi-core** | **191/256** |

End to end on `mistralai/Mistral-7B-Instruct-v0.3`, `MESH_DEVICE=P150`, `simple_text_demo.py -k "performance-ci-token-matching"`, 500 teacher-forced tokens against the repo's own `reference_outputs/Mistral-7B-Instruct-v0.3.refpt`:

| sampling params | path taken | Top-1 | Top-5 |
|---|---|---|---|
| `temperature=0` — the config as committed | argmax fast path | 96.40% | 99.40% |
| `k=1, p=0.5, temp=1` | **top-k** | **83.20%** | **85.40%** |
| `k=1, p=0.5, temp=1`, with the fix below | top-k | **96.40%** | **99.40%** |

`p=0.5` is the whole difference between the first two rows: `_is_force_argmax_sampling` requires `p ∈ {0.0, 1.0}`, so `p=0.5` keeps the run on the top-k path while `k=1` keeps the result deterministic. The committed CI config uses `temperature=0`, which #48886 now routes to the argmax fast path on single-chip parts — which is why CI is green and this is not visible today.

This is also the root cause of #34763, open since December 2025, whose workaround disables on-device sampling for the whole Mistral-7B family.

### Root cause

Both samplers build a `[1, 1, 32, W]` buffer whose contents are `0..W-1` and hand it to `ttnn.topk` as `indices_tensor`.

`models/common/sampling/tt_sampling.py:342`:

```python
indices_tensor_torch = torch.zeros(1, 1, self.max_batch_size, padded_per_device, dtype=torch.int32)
for i in range(padded_per_device):
    indices_tensor_torch[:, :, :, i] = i
```

`padded_per_device` is `padded_vocab_size // _get_num_sampling_shards()`, and `_get_num_sampling_shards()` returns **2** on a 1×1 mesh (`tt_sampling.py:307`), so the buffer is `V/2` wide. `decode_forward` then splits it in half again and hands each `V/4` half to a `ttnn.topk` whose logits half is `V/2`. Every call passes an indices tensor half the width of its input.

Nothing catches that. `TopKDeviceOperation::validate_on_program_cache_miss` checks the indices tensor's *dtype* and never its shape against the input (`topk_device_operation.cpp:177`). Both readers then page the indices accessor with `page_id = i * Wt + j`, where `Wt` is the **input's** width in tiles (`reader_create_index_local_topk.cpp:65`, `reader_create_index_tensor.cpp`), so the second half of every row is read past the end of the buffer. The values are unaffected — they never come from that buffer — but the indices returned alongside them no longer belong to them. Feeding an 8192-wide iota to a 16384-wide input directly:

| | full-width iota | half-width iota |
|---|---|---|
| values | correct | identical |
| `gather(input, idx) == vals` | True | **False** |
| indices out of range | 0 | 274 |
| top-1 index wrong | 0/32 | 17/32 |

`topk_single_core_program_factory.cpp:212` hardcodes `GENERATE_INDICES = "1"` and ignores the argument entirely (#51921, #36329) — the comment at `:186` already records that "the reader's sole use sits behind `#if not GENERATE_INDICES` and that define is" never cleared. So the corruption only surfaces when a half reaches the multi-core factory: a power of two in `[8192, 65535)`.

### Which configurations reach it

On one device `padded_vocab_size = ceil(vocab/32)*32`, so the reachable vocabularies are 16384, 32768 and 65536. Of the models tt_transformers supports on a single card, only `mistralai/Mistral-7B-Instruct-v0.3` lands on one — 32768, halving to 16384. v0.1 and v0.2 are 32000, which halves to 16000 and falls back to the single-core factory.

Since #48886 set `allow_force_argmax` to `self.num_devices == 1`, single-chip **greedy** decode takes the argmax fast path and never reaches this. What is left exposed is **non-greedy sampling** on a single chip — any `top_k > 1`, any `top_p` other than 0 or 1, any `temperature` other than 1 — which is the ordinary vLLM path. #48886 narrowed the blast radius without touching the defect.

`Sampling1D` is not affected: its `_make_local_indices` already used the full width on the split path and records that as a fix over the TTTv1 port.

It survived because `TTSampling`'s 1×1 top-k path has no test at all. `models/common/tests/test_sampling.py` is TG-only, the gpt_oss and gemma4 unit tests are (4,8) and (1,8), and `test_sampling_1d.py` covers `Sampling1D`.

### Why this is #34763

`simple_text_demo.py:35` disables on-device sampling for the whole Mistral-7B family, pointing at #34763. #35153, which added that skip list, describes itself as a hack — *"Device sampling is not working with Mistral7B. For now skip sampling on device for Mistral7B. This needs proper fixing later."* Its author bisected the failure to `58ac27a9`, *"[TT-Transformers] Support on-device sampling (#31046)"* — the commit that introduced this buffer — and noted *"Seems like sampling on device was not tested for Mistral7b."* The last update on #34763 is 2026-01-07.

With the defect fixed, the skip list can come off: the same demo, skip list lifted, is 96.40% Top-1 on the committed greedy config (argmax fast path) and 96.40% on the top-k path, against 96.40% for today's host-sampling default. Decode is 24.78 ms/token today, 24.51 with device sampling on the greedy path.

## Expected

Drop the argument and the buffers rather than widening them. The buffer holds exactly what `generate_index_tile()` writes when the argument is absent, so the width the kernel generates for itself is the correct one by construction. Implemented and measured; a PR can follow.

**On today's main this costs a little at the op level and nothing at the model level.** `ttnn.topk`, `[1, 1, 32, W]`, `bfloat16`, 50 ops enqueued back to back, median of 30, the two branches interleaved in one process:

| W | read `indices_tensor` | generate | |
|---|---|---|---|
| 8192 | 96.93 | 107.84 | +11.3% |
| 16384 | 109.60 | 130.21 | +18.8% |
| 32768 | 128.37 | 186.95 | +45.6% |

`Sampling1D.decode_forward` on a 1×1 mesh, median of 30 samples of 10 calls, two interleaved runs: vocab 32768 goes 352.9 → 378.2 µs (+7.2%), vocab 1024 goes 269.2 → **264.9** µs (−1.6%, since the single-core path never read the buffer and the per-call `ttnn.split` and two `deallocate` go away).

At the model level that disappears into the step: Mistral-7B batch-1 decode on the top-k path is **24.71 ms/token before and 24.73 after**, +0.08%, which is noise. The sampler is roughly 1.4% of a decode step.

It also drops a resident `uint16` tensor per device — 1 MiB at a 16384-wide shard — re-read every decode step.

The op-level cost inverts with #52909, which builds the index tile with the NoC: with that change generation becomes the cheaper branch (105.83 µs against 109.81 at W=16384, −3.6%) and `Sampling1D.decode_forward` at vocab 32768 drops to 347.5 µs, −6.7% against today. That is a reason to want #52909, not a reason to hold this.

The per-shard offset that makes gathered indices global comes from `index_offsets` / `tt_indices_device_offsets`, a separate buffer, and is unaffected.

### Correctness

The generated index tile is bit-identical to the iota buffer it replaces. `ttnn.topk` with and without `indices_tensor`, values and indices compared bitwise and both compared against `torch.topk` on the bf16-rounded input, at W = 64, 128, 1024, 4096, 8192, 16384, 32768, 65536, 131072, 1536, 3000 and 64128 — both program factories, the UINT16→UINT32 index boundary at 65535, and non-power-of-2 widths. Identical on every check. The output index dtype is chosen from the input width inside the op, never from the passed tensor, so dropping the argument cannot change it.

`models/common/tests/modules/sampling/test_sampling_1d.py`: 92 passed, 48 skipped, identical before and after. `tests/ttnn/unit_tests/operations/reduce/test_topk.py`, `test_sampling.py` and `tests/ttnn/unit_tests/operations/data_movement/test_sort.py`: 345 passed, 8 skipped, 80 xfailed.

## Scope

| file | |
|---|---|
| `models/common/sampling/tt_sampling.py` | `_create_indices_tensors`, two `ttnn.topk` calls, `_select_topk_indices_dtype` becomes unused |
| `models/common/modules/sampling/sampling_1d.py` | `local_indices` config field and factory, two `ttnn.topk` calls |
| `models/common/tests/modules/sampling/test_sampling_1d.py` | one buffer assertion |
| `models/common/tests/test_sampling.py` | new 1×1 top-k-vs-argmax case at vocab 32768 and 32000, covering both factories; fails 21/32 users without the fix |
| `models/tt_transformers/demo/simple_text_demo.py` | drop `"Mistral-7B"` from `models_not_supported_for_device_sampling` |

`models/experimental/llama32_1b_quasar/` carries copies of both samplers with the same defect and is left alone; its vocab 128256 halves to 64128, so it is latent there.

The change leaves `ttnn.topk`'s parameter in place, but it is worth recording that afterwards **it has no production caller left in the repo**: the only remaining callers are two direct `ttnn.topk` calls in `test_sampling_1d.py` and the `pass_indices_tensor` parametrization in `tests/ttnn/unit_tests/operations/reduce/test_topk.py`, both of which test the op itself. The parameter is currently worse than unused — the multi-core factory reads it with no shape validation, and the single-core factory appends its `TensorAccessorArgs` and then ignores it. Deleting it removes roughly 60 lines across nine files and resolves #51921 by subtraction rather than by making the two factories agree. That is a public-API change and belongs in its own PR.

A shape check against the input in `validate_on_program_cache_miss` would have turned this into a loud failure instead of wrong tokens, and is worth adding whether or not the parameter survives.

## Environment

Blackhole P300, one chip visible, Ubuntu 22.04, main at `03b9d08219`, built Release. 1×1 and 1×2 meshes.

## Not established

* **Different part.** #34763's CI failures were N150 and N300; the reproduction above is Blackhole P150. Same code path, different silicon.
* **The N300 report is not explained by this defect.** On a 1×2 mesh the buffer width already matches the input width, and 1×2 measures 0 mismatches before and after the change. Either something else in the same feature is also broken, or the 2025-12-24 comment refers to a different test in the same job.
* 1×8 and 4×8 are unmeasured. 1×2 is the largest mesh available here, and multi-device is where the buffer is largest and the read most expensive.
* Whether the vLLM integration applies the same skip list is outside this repo. It is the path most exposed, since it is the one that uses real sampling parameters rather than greedy.

## Related

* #34763 — *Investigate mistral7b PCC and repeat batch errors*, the symptom.
* #35153 — the skip list added to contain it, self-described as a hack.
* #31046 / `58ac27a9` — introduced on-device sampling and this buffer.
* #48886 — routed single-chip greedy to the argmax fast path, which is why this is currently invisible in CI.
* #51921, #36329 — the single-core factory ignores `indices_tensor`.
* #25728, #36136 — added the buffer to take the read branch.
* #52909 — builds the index tile with the NoC; inverts the op-level cost above.

kinginu:
### PR Category

Bug fix

### Summary

Closes #52928, which carries the full analysis and the measurements.

On a 1×1 mesh both samplers hand `ttnn.topk` an iota `indices_tensor` half the width of the logits, and nothing validates the two shapes against each other, so the multi-core reader pages index tiles past the end of the buffer and returns indices that do not belong to the values beside them. 126 of 256 sampled tokens are wrong at `padded_vocab_size` 32768. End to end, `Mistral-7B-Instruct-v0.3` on the top-k path drops from 96.40% to 83.20% Top-1 against the repo's own reference tokens.

The buffer holds exactly what `generate_index_tile()` writes when the argument is absent, so it is dropped rather than widened.

Two commits, separable:

1. **the fix**, plus a 1×1 top-k-vs-argmax regression test — `TTSampling`'s single-device top-k path had no test at all, which is how this survived.
2. **`Re-enable on-device sampling for Mistral-7B`** — one line off the demo's skip list, which #35153 added to contain #34763 and called a hack at the time. Drop this commit if CI disagrees; the fix does not depend on it.

### Notes for reviewers

**This is invisible in CI today.** Since #48886 routed single-chip greedy to the argmax fast path, the committed `ci-token-matching` config (`temperature=0`) never reaches the broken path. Reproducing it needs `p ∉ {0.0, 1.0}` so `_is_force_argmax_sampling` declines — `k=1, p=0.5, temp=1` keeps the result deterministic and gives the 83.20% above. What stays exposed is non-greedy sampling on a single chip, which is the ordinary vLLM path.

**Cost.** On today's main, generating the index tile is dearer than reading it: +18.8% on `ttnn.topk` at `[1, 1, 32, 16384]`, which is +7.2% on `Sampling1D.decode_forward` at vocab 32768 and −1.6% at vocab 1024. At the model level it vanishes into the step — Mistral-7B batch-1 decode on the top-k path is 24.71 ms/token before and 24.73 after, +0.08%. #52909 inverts the op-level number to −3.6%.

Verified on Blackhole P150, main at `03b9d08219`:

- [x] New test `test_ttsampling_topk_matches_argmax_on_single_device`, vocab 32768 and 32000 to cover both program factories. Fails 21/32 users at 32768 without the fix.
- [x] `test_sampling_1d.py`: 92 passed / 48 skipped, identical before and after. `test_topk.py`, `test_sampling.py`, `test_sort.py`: 345 passed, 8 skipped, 80 xfailed.
- [x] Generated index tile bit-identical to the iota buffer at twelve widths — both factories, the UINT16→UINT32 boundary at 65535, non-power-of-2 widths.
- [x] `simple_text_demo.py` end to end on the real model: 83.20% → 96.40% Top-1.
- [ ] **1×8 / 4×8 unmeasured** — 1×2 is the largest mesh I have, and multi-device is where the buffer is largest.
- [ ] **N150 unverified** — #34763's failures were N150 and N300. CI is what confirms commit 2.
- [ ] **The N300 report is not explained by this defect** — on 1×2 the buffer width already matches the input width, and 1×2 measures 0 mismatches before and after.

### DIFF
--- a/models/common/modules/sampling/sampling_1d.py
+++ b/models/common/modules/sampling/sampling_1d.py
@@ -80,9 +80,6 @@ class Sampling1DConfig:
     # --- Persistent buffer specs (LazyBuffer | ttnn.Tensor | None) ---
     # Static index buffers (computed from vocab_size + num_devices, never mutated)
     index_offsets: LazyBuffer | ttnn.Tensor | None = None  # [1,1,32,max_top_k*num_devices], int32, TILE
-    local_indices: LazyBuffer | ttnn.Tensor | None = (
-        None  # [1,1,32,W], uint16, TILE (W=vocab for 1x1, per_dev_vocab otherwise)
-    )
     invalid_vocab_mask: LazyBuffer | ttnn.Tensor | None = None  # Full fallback mask, sharded like logits
     invalid_vocab_tail_mask: LazyBuffer | ttnn.Tensor | None = None  # Compact [1,1,32,tail] local mask
     invalid_vocab_tail_width: int = 0
@@ -103,7 +100,7 @@ class Sampling1DConfig:
             return False
         if self.mesh_device.get_num_devices() > 1 and self.tt_ccl is None:
             return False
-        required_buffers = ["index_offsets", "local_indices", "seeds", "user_ids"]
+        required_buffers = ["index_offsets", "seeds", "user_ids"]
         if not all(self._buf_resolved(getattr(self, f)) for f in required_buffers):
             return False
         if self.valid_vocab_size is not None and self.valid_vocab_size < self.vocab_size:
@@ -195,7 +192,6 @@ class Sampling1D(LightweightModule):
         cfg = self.config
 
         self._index_offsets = _materialize(cfg.index_offsets)
-        self._local_indices = _materialize(cfg.local_indices)
         self._invalid_vocab_mask = _materialize(cfg.invalid_vocab_mask) if cfg.invalid_vocab_mask is not None else None
         self._invalid_vocab_tail_mask = (
             _materialize(cfg.invalid_vocab_tail_mask) if cfg.invalid_vocab_tail_mask is not None else None
@@ -522,7 +518,6 @@ class Sampling1D(LightweightModule):
         """Split vocab in half → two topk → concat. Port of tt_sampling.py:346-371."""
         cfg = self.config
         x_list = ttnn.split(x_bf16, x_bf16.shape[-1] // 2, dim=3)
-        indices_list = ttnn.split(self._local_indices, self._local_indices.shape[-1] // 2, dim=3)
 
         values_parts = []
         indices_parts = []
@@ -532,12 +527,10 @@ class Sampling1D(LightweightModule):
                 k=cfg.max_top_k,
                 dim=-1,
                 sub_core_grids=cfg.sub_core_grid_topk,
-                indices_tensor=indices_list[i],
             )
             values_parts.append(vals)
             indices_parts.append(idxs)
             x_list[i].deallocate()
-            indices_list[i].deallocate()
 
         gathered_values = ttnn.concat(values_parts, dim=3)
         gathered_indices = ttnn.concat(indices_parts, dim=3)
@@ -554,8 +547,7 @@ class Sampling1D(LightweightModule):
         cluster_shape = cfg.mesh_device.shape
 
         # Pad the per-device shard up to the next power of 2 so ttnn.topk hits its fast path.
-        # Padded entries get -inf so they are never selected; the pre-padded _local_indices buffer
-        # is widened to match (see _resolve_sampling1d_config). Mirrors tt_sampling.py:451-458.
+        # Padded entries get -inf so they are never selected. Mirrors tt_sampling.py:451-458.
         if cfg.pad_to_power_of_2 and not _is_power_of_2(x_bf16.shape[-1]):
             padded_width = _upper_power_of_2(x_bf16.shape[-1])
             x_bf16 = ttnn.pad(
@@ -570,7 +562,6 @@ class Sampling1D(LightweightModule):
             k=cfg.max_top_k,
             dim=-1,
             sub_core_grids=cfg.sub_core_grid_topk,
-            indices_tensor=self._local_indices,
         )
 
         # For 1D meshes use cluster_axis=None
@@ -755,40 +746,6 @@ def _resolve_sampling1d_config(config: Sampling1DConfig) -> Sampling1DConfig:
         memory_config=ttnn.DRAM_MEMORY_CONFIG,
     )
     to_set["index_offsets"] = _resolve_buf(config.index_offsets, idx_defaults, _make_index_offsets)
-
-    # local_indices: [1, 1, B, local_indices_width]
-    # For multi-device: width = per_device_vocab (each device's shard)
-    # For single-device split (multi_step_reduction): width = V (full vocab), so that
-    #   after ttnn.split(..., V//2, dim=3) each half has width V//2 = per_device_vocab,
-    #   matching the logits half width. Each half contains a 0-based range [0..V//2-1].
-    #   Bug fix: TTTv1 used per_device_vocab here too, causing a 2x width mismatch
-    #   between indices_tensor and logits in ttnn.topk on single-device.
-    local_indices_width = V if multi_step_reduction else per_device_vocab
-
-    def _make_local_indices():
-        if multi_step_reduction:
-            half = local_indices_width // 2
-            r = torch.arange(half, dtype=torch.int32)
-            row = torch.cat([r, r])
-        else:
-            row = torch.arange(local_indices_width, dtype=torch.int32)
-        out = row.unsqueeze(0).unsqueeze(0).expand(1, 1, B, -1).contiguous()
-        # Pad the indices buffer to match the power-of-2-padded topk input width (multi-device
-        # path only — strict TTTv1 parity, so the 1×1 split path is never padded). Fill with -1
-        # (invalid index) so the padded slots are never used. Mirrors tt_sampling.py:277-284.
-        if config.pad_to_power_of_2 and not multi_step_reduction and not _is_power_of_2(local_indices_width):
-            padded_width = _upper_power_of_2(local_indices_width)
-            out = torch.nn.functional.pad(out, (0, padded_width - local_indices_width), mode="constant", value=-1)
-        return out
-
-    local_idx_defaults = dict(
-        dtype=ttnn.uint16,
-        layout=ttnn.TILE_LAYOUT,
-        device=mesh_device,
-        mesh_mapper=replicate_mapper,
-        memory_config=ttnn.DRAM_MEMORY_CONFIG,
-    )
-    to_set["local_indices"] = _resolve_buf(config.local_indices, local_idx_defaults, _make_local_indices)
 
     vocab_shard_dims = get_vocab_shard_dims(cluster_shape)
     invalid_vocab_defaults = dict(
--- a/models/common/modules/sampling/sampling_1d.py
+++ b/models/common/modules/sampling/sampling_1d.py
@@ -267,7 +267,7 @@ class Sampling1D(LightweightModule):
         """Dispatcher."""
         return self.decode_forward(logits, **kwargs)
 
-    # -- Argmax path (port from tt_sampling.py:310-341) -----------------------
+    # -- Argmax path (port of the argmax branch of TTSampling.forward) --------
 
     def _sample_argmax(self, logits, tt_out_tok):
         slice_valid_vocab = self._can_slice_valid_vocab_for_argmax()
@@ -364,7 +364,7 @@ class Sampling1D(LightweightModule):
         """DRAM memory config: no extra round-trip needed."""
         return topk_values, topk_indices_int32
 
-    # -- Top-k sampling (port from tt_sampling.py:343-481) --------------------
+    # -- Top-k sampling (port of the top-k branch of TTSampling.forward) ------
 
     def _sample_topk(self, logits, k, p, temp, seeds, tt_out_tok):
         cfg = self.config
@@ -515,7 +515,7 @@ class Sampling1D(LightweightModule):
     # -- Top-k strategies (bound at init, no if-else in forward) --------------
 
     def _topk_single_device(self, x_bf16):
-        """Split vocab in half → two topk → concat. Port of tt_sampling.py:346-371."""
+        """Split vocab in half → two topk → concat. Port of the single-device branch of TTSampling.forward."""
         cfg = self.config
         x_list = ttnn.split(x_bf16, x_bf16.shape[-1] // 2, dim=3)
 
@@ -542,12 +542,12 @@ class Sampling1D(LightweightModule):
         return gathered_values, gathered_indices
 
     def _topk_multi_device(self, x_bf16):
-        """Local topk → all_gather across devices. Port of tt_sampling.py:372-421."""
+        """Local topk → all_gather across devices. Port of the multi-device branch of TTSampling.forward."""
         cfg = self.config
         cluster_shape = cfg.mesh_device.shape
 
         # Pad the per-device shard up to the next power of 2 so ttnn.topk hits its fast path.
-        # Padded entries get -inf so they are never selected. Mirrors tt_sampling.py:451-458.
+        # Padded entries get -inf so they are never selected. Mirrors the padding in TTSampling.forward.
         if cfg.pad_to_power_of_2 and not _is_power_of_2(x_bf16.shape[-1]):
             padded_width = _upper_power_of_2(x_bf16.shape[-1])
             x_bf16 = ttnn.pad(
@@ -597,7 +597,7 @@ class Sampling1D(LightweightModule):
     def _perform_all_gather(self, tensor, dim, cluster_axis, memory_config, num_links, buffer_key=None, dtype=None):
         """Flexible all-gather: prefer line_all_gather if available, else ttnn.all_gather.
 
-        Port of TTSampling._perform_all_gather (tt_sampling.py:231-259).
+        Port of TTSampling._perform_all_gather.
         """
         if callable(self._line_all_gather):
             kwargs = {
--- a/models/common/modules/sampling/sampling_1d.py
+++ b/models/common/modules/sampling/sampling_1d.py
@@ -167,15 +167,11 @@ class Sampling1D(LightweightModule):
         # CCL introspection (port from TTSampling.__init__ lines 77-91)
         self._line_all_gather = getattr(self.config.tt_ccl, "line_all_gather", None) if self.config.tt_ccl else None
         self._line_all_gather_supports_buffer_key = False
-        self._line_all_gather_supports_dtype = False
         if callable(self._line_all_gather):
             try:
                 sig = inspect.signature(self._line_all_gather)
                 params = sig.parameters
                 self._line_all_gather_supports_buffer_key = "buffer_key" in params or any(
-                    p.kind == inspect.Parameter.VAR_KEYWORD for p in params.values()
-                )
-                self._line_all_gather_supports_dtype = "dtype" in params or any(
                     p.kind == inspect.Parameter.VAR_KEYWORD for p in params.values()
                 )
             except (TypeError, ValueError):
@@ -586,7 +582,6 @@ class Sampling1D(LightweightModule):
             memory_config=ttnn.DRAM_MEMORY_CONFIG,
             num_links=cfg.num_gather_links,
             buffer_key="SAMPLING_INDICES",
-            dtype=ttnn.uint16,
         )
         ttnn.deallocate(topk_indices)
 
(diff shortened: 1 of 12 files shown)


mtairum:
**Review summary** (fix for #52928)

The core fix is sound: dropping the half-width `indices_tensor` and letting `ttnn.topk` generate indices in-kernel is correct — verified the `-inf` padding, index dtype auto-selection, and downstream offset math all hold.

A few things worth addressing:

1. **Root cause is still armed in `ttnn.topk`** — `validate()` in `topk_device_operation.cpp` checks a caller-supplied `indices_tensor` only for dtype, never width vs the input. Any other caller can silently reproduce this bug. Suggest a `TT_FATAL` width check so the footgun is closed for everyone, not just this call site.
2. **Quasar Llama-3.2-1B port still has the bug** — `models/experimental/llama32_1b_quasar/sampling/tt_sampling.py:645` (and its `sampling_1d.py`) is a verbatim copy of the pre-fix code. Latent only because vocab 128256 routes to the single-core factory, which ignores caller indices (#36329).
3. **Mistral-7B re-enable** — the exclusion removal assumes #34763 (PCC + repeat-batch errors) is fully explained by this bug, but the new regression test covers `TTSampling` in isolation. Worth running the N150 Mistral demo before merge and closing/updating #34763.

Minor: index dtype is now derived from the post-power-of-2-pad width, so per-device shards in (32768, 65535] flip uint16→uint32 (2× index gather traffic; the `dtype=ttnn.uint16` gather request at `tt_sampling.py:858` is dead weight); `sampling_1d.py` topk still lacks `stable=`/tie-break unlike the sibling TTSampling path (#52177); the `tt_sampling.py:NNN` cross-reference comments in `sampling_1d.py` now point at shifted line ranges; `models_not_supported_for_device_sampling` is now always empty along with its dead fallback branch.

mtairum:
Kicked off a Mistral-7B N150 CI run on the PR head (`011d6b9`, pushed to `mtairum/pr53001-mistral-n150-ci` for dispatch), to validate commit 2 (the skip-list removal) against #34763's original N150 failure mode.

Here's the run:
**(Tier 2) Models End-To-End Tests — `model=mistral-7b`, `sku=wh_n150`:** https://github.com/tenstorrent/tt-metal/actions/runs/31718556276

Will report results here.

mtairum:
**Mistral-7B N150 CI result** (run: https://github.com/tenstorrent/tt-metal/actions/runs/31718556276)

The job is red, but not for a functional reason — the demo ran end to end (prefill + decode for 32 users, TTFT target **pass**), and the only failure is the decode perf target being exceeded on the *good* side:

```
decode_t/s/u: measured=25.31 > upper_bound=25.18 (expected=21.9, tolerance=0.15)
[much better than expected, update target]
```

That is consistent with commit 2 re-enabling on-device sampling for Mistral-7B: decode is now ~15% faster than the target that was set while host sampling was forced. No PCC/accuracy or repeat-batch failures in the log — #34763's original failure mode did not reproduce on N150 with the fix in place.

Suggested follow-up for this PR: bump the mistral-7b wh_n150 `decode_t/s/u` target (21.9 → ~25.3) so the tier-2 job goes green with device sampling on.

mtairum:
Pushed two commits to this branch (maintainer edit):

- `01da1c9` — apply the same indices_tensor removal to the Quasar Llama-3.2-1B port (`models/experimental/llama32_1b_quasar/`), which was a verbatim copy of the pre-fix code.
- `0e0e5f0` — raise the mistral-7b `wh_n150` `decode_t/s/u` target 21.9 → 25.3 in `models/model_targets.yaml`, since device sampling makes decode faster than the old host-sampling target's +15% upper bound (see CI run above).

kinginu:
Thanks for the two commits, and for running the N150 job.

1. Agreed, I will add the width check to `validate()`. Nothing under `models/` passes an `indices_tensor` to `ttnn.topk` once this PR drops the sampling call — the MoE routers, mixtral, grok, gpt_oss, minimax and gemma4 all take the two-return form — so the check closes the footgun without touching a live caller. It is separate from #52909, which proposes generating the index tile in the kernel: that removes the need to pass indices but not the parameter, so the guard is still wanted either way.

2. Covered by your `01da1c9`. I had not looked under `models/experimental/` and would have missed it.

3. Your run answers the functional half. #34763 is assigned to @sraizada-tt, so rather than close it I will post the demo result there and let them decide.

Minor:

- The `dtype=ttnn.uint16` on the gather is ignored rather than narrowing. `_line_all_gather_supports_dtype` comes from inspecting the CCL signature, and the only production `line_all_gather` (`llama_ccl.py:1244`) declares neither `dtype` nor `**kwargs`, so the kwarg is dropped; the `ttnn.all_gather` fallback has no `dtype` parameter either. It is true only against the mock at `test_sampling_1d.py:211`. I will remove the argument — getting the traffic back needs `line_all_gather` to accept a dtype, which is a CCL change rather than one for this PR.
- `sampling_1d.py`'s topk lacking `stable=` is a real gap against the TTSampling path after #52177. I will file it as its own issue: it changes tie-breaking on a path this PR does not otherwise touch, so it wants its own test rather than riding along here.
- The `tt_sampling.py:NNN` cross-references in `sampling_1d.py` are stale by the lines this PR removes. Will re-point them.
- `models_not_supported_for_device_sampling` and the branch at `simple_text_demo.py:1206` are both dead now. Will remove.

Will push the validate check and the three cleanups shortly.

kinginu:
Pushed. Two things came out differently from what I said above.

**The check is in `ttnn::topk`, not the device op's `validate()`.** I put it in `validate()` first and it failed five existing `pass_indices_tensor=True` cases. By the time `validate()` runs, the front end has transposed the input to put the reduced dim last, reshaped it to 4D and padded it up to `min_dim_per_core`, and it does none of that to the indices tensor — at `W=20` the input arrives 64 wide against the caller's 32, so the two shapes are not comparable there. At the `ttnn::topk` boundary they are the caller's own, next to the identical check the preallocated output tensors already get.

**Matching shapes are not sufficient.** With the reduced dim not already last, the transpose above breaks the correspondence just as badly. At `[1, 1, 8192, 32]` with `dim=2` on Blackhole P300, every returned index came back rounded down to a multiple of the tile height — 6272 against 6288, 4864 against 4884 — while the same call with the argument dropped agreed with `torch.topk`. `175cfee` rejects that case too. Transposing the indices tensor alongside the input would make it work rather than fail, which is the better fix if anything ever wants it; nothing does today.

The `min_dim_per_core` padding is the third instance of the same asymmetry, but it is unreachable: anything padded up to 64 is below `multi_core_min_width`, and the single-core factory ignores caller indices entirely.

`tests/ttnn/unit_tests/operations/reduce/test_topk.py` gets a case for each. The file is green on Blackhole P300 — 193 passed, 8 skipped, 80 xfailed.

`82d1af7` is the three cleanups. The `dtype=ttnn.uint16` is gone from both copies; `_perform_all_gather`'s `dtype` parameter stays, since `test_sampling_1d.py` exercises it directly and it would carry a dtype the day a CCL accepts one.

mtairum:
Pushed two more commits addressing the remaining review concerns:

- `d210061` — **finish the uint16 cleanup consistently**: the `dtype=ttnn.uint16` removal from TTSampling's SAMPLING_INDICES gather had left the same line in both Sampling1D ports (models/common + quasar), and with it the last caller of `_perform_all_gather`'s `dtype` parameter. Removed the kwarg from the gathers and deleted the dead dtype plumbing (parameter, forwarding branch, signature introspection) in all four copies, with the mock-CCL unit tests updated to match.
- `da0ccb4` — **close the `ttnn::prim::topk` bypass**: the new shape guards in the composite `topk()` front-end don't cover the prim entry point, so the device op's `validate_on_program_cache_miss` now enforces padded-shape equality between a caller-supplied indices tensor and the input — the reader kernels page indices at the input's stride (`page_id = i * Wt + j`), so this is the device-op's own invariant.

mtairum:
This is looking good now and is passing all the required CI tests:

- Sanity tests ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31806916206
- Static checks / linters ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31806915133
- PR Gate —  https://github.com/tenstorrent/tt-metal/actions/runs/31806915843

Model related tests:
- mistral-7b · wh_n150 ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31799628113
- llama3.2-1b · wh_n150 ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31799633387
- qwen2.5-7b · wh_n300 ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31799635450
- llama3.3-70b · wh_galaxy_perf ✅ — https://github.com/tenstorrent/tt-metal/actions/runs/31799881719
- gpt-oss-120b · wh_galaxy_perf ✅  — https://github.com/tenstorrent/tt-metal/actions/runs/31806929252

I've requested reviews.

Thank you for your contribution!

mtairum (approved):

sraizada-tt (approved):

bbradelTT (approved):

yieldthought:
Codex review:

[ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp:376](https://github.com/tenstorrent/tt-metal/blob/c2911221a947f188811bdc28ad664af61597926d/ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp#L376)
The padding applied to a caller-supplied indices_tensor uses the 4D padding vector built for transformed_tensor, but padded_indices_tensor is still the caller's original-rank tensor. ttnn.pad rejects padding vectors longer than the tensor rank ([ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp:445](https://github.com/tenstorrent/tt-metal/blob/c2911221a947f188811bdc28ad664af61597926d/ttnn/cpp/ttnn/operations/data_movement/pad/pad.cpp#L445)), so ttnn.topk now regresses for lower-rank valid calls such as a 2D [32, 32] input with a same-shape indices tensor: pad_amount is nonzero, line 376 passes four pad dims to a rank-2 tensor, and the op raises before prim::topk. Suggestion: transform the indices tensor in the same way as the input before padding, or build the indices padding vector with the indices tensor's rank.

yieldthought (approved):
Codex: Approved per Mark's explicit Miguel vacation instruction after posting coreview results. Miguel should use his judgment on whether to address any findings before submission.

Issue #51835, opened 2026-08-01
[Cleanup] Remove `IDevice` from public API


obacherikovTT:
**Reason**

Abstract class `IDevice` at this point has a single derived class that is used in the public API: `MeshDevice`.
Thus, it's better to remove it and instead directly use `MeshDevice`.

**Scope**

Deprecate and later delete all the public APIs that reference `IDevice`.

**Notes**
Functional changes should be minimized.

PR #53007, opened 2026-08-12
[Cleanup] Refactoring device ID access in dispatch tests

obacherikovTT:
### PR Category
Cleanup, Test Only

### Summary

We're aiming to remove `IDevice` from the public API.
It's used, in particular, in `MeshDevice::get_devices()` method.
A lot of its usage appears in tests just to log the device ID.

To reduce the scope of the complete cleanup, we:
- Remove the device ID logs from a single-device tests.
- Switch to `get_device_ids()[0]` in multi-device tests.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:obacherikov-device-id-logs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:obacherikov-device-id-logs)

### DIFF
--- a/tests/tt_metal/tt_metal/dispatch/dispatch_buffer/test_EnqueueWriteBuffer_and_EnqueueReadBuffer.cpp
+++ b/tests/tt_metal/tt_metal/dispatch/dispatch_buffer/test_EnqueueWriteBuffer_and_EnqueueReadBuffer.cpp
@@ -1399,8 +1399,7 @@ TEST_F(UnitMeshCQSingleCardSharedBufferFixture, TestWrapCompletionQOnInsufficien
 TEST_F(UnitMeshMultiCQMultiDeviceBufferFixture, WriteOneTileToDramBank0) {
     TestBufferConfig config = {.num_pages = 1, .page_size = 2048, .buftype = BufferType::DRAM};
     for (const auto& mesh_device : devices_) {
-        auto* device = mesh_device->get_devices()[0];
-        log_info(tt::LogTest, "Running On Device {}", device->id());
+        log_info(tt::LogTest, "Running On Device {}", mesh_device->get_device_ids()[0]);
         distributed::MeshCommandQueue& a = mesh_device->mesh_command_queue(0);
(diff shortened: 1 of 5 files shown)


tests/tt_metal/tt_metal/dispatch/dispatch_buffer/test_EnqueueWriteBuffer_and_EnqueueReadBuffer.cpp
```diff
@@ -715,7 +715,6 @@ namespace dram_tests {
 TEST_F(UnitMeshCQSingleCardSharedBufferFixture, WriteOneTileToDramBank0) {
     TestBufferConfig config = {.num_pages = 1, .page_size = 2048, .buftype = BufferType::DRAM};
     for (const auto& mesh_device : devices_) {
-        log_info(tt::LogTest, "Running On Device {}", mesh_device->id());
```
obacherikovTT, line 718: For tests like this that have "SingleCard" in the name distinguishing device doesn't make sense, because there's just one.

jbaumanTT:
These IDs can be useful to determine which device the test is failing on. Some tests may only fail on specific devices, because those devices have different properties (different harvesting, tunneled vs. non-tunneled). Removing the logging would make debugging that harder.

jbaumanTT (approved):

PR #52202, opened 2026-08-05
[Bug fix] SFPU: ttnn.i1 returns a constant above |x| = 88.5

kinginu:
## PR Category

Bug fix.

## Summary

Closes #52174.

`i1` clamps its input at 88.5 and returns a constant above it. The bound is `exp`'s, not `i1`'s: the kernel forms `exp(x)/sqrt(2*pi*x) * P(1/x)`, and the divisor is about 23.6 there, so `exp` leaves FP32 at 88.72284 while `i1` does not until 91.90626. Between those two numbers the op has an answer and returns a saturated one instead.

The two unsafe `exp` helpers take a compile-time `EXP2_DOWNSCALE`, `i1`'s asymptotic moves into a new `ckernel_sfpu_bessel_common.h` that applies the matching rescale to its caller's coefficients, and the 88.5 clamp goes away.

The rescale is applied inside `exp` rather than to its result, because by then the exponent field has already wrapped: rescaling afterwards returns `Inf` at `x = 89` while returning correct values at 90 and 91.5. Inside, it folds into the 21f path's bias constant or the FP32 path's exponent recombination, so it is exact and costs no instruction.

The FP32 path guards the rescale behind `if constexpr` rather than folding it into the exponent arithmetic, because at the default of 0 the folded spelling still perturbs codegen — measured at +32 bytes of `.text` in `tanhshrink`, the only other caller of the unsafe path. With the guard, every existing caller is byte-identical.

`ckernel_sfpu_bessel_common.h` lands with one caller. It is free — with `i1` routed through it, all 29 math kernels in the JIT cache are byte-identical to this PR without it — and it exists because `i0`'s asymptotic is the same shape, differing only in the coefficients and `i1`'s sign fix-up, while the open `i0` work carries its own copy of that block including the 88.5 bound.

## Accuracy

Exhaustive over all 65536 bfloat16 patterns, reference `torch.special.i1` in float64, max BF16 ULP. **P300 and N300 give the same figures.**

| region | patterns | before | after |
|---|---|---|---|
| `\|x\| <= 10` | 33346 | | bit-identical |
| `10 < \|x\| <= 88.5` | 802 | 0.8582 | 0.8582 |
| `88.5 < \|x\| <= 91.90626` | 12 | **218.1142** | **0.5643** |
| true value overflows FP32 | 30364 | 30364 finite | **30364 `±Inf`** |

[...]

PR #53022, opened 2026-08-13
[Bug fix] SFPU: ttnn.erfc returns a constant above |x| = 5

### LINKED ISSUE #51137: erfc returns a constant for |x| >= 5 (input clamp), and segment 1 [2.5, 5.0] fails the LLK domain test
### Describe the bug

`calculate_erfc` clamps `|x|` to `5.0` before evaluating its rational LUT, so
**every input with `|x| >= 5` returns the same constant** — the LUT value at
`5.0`, roughly `2.4e-12`. `erfc` does not saturate; it keeps decaying, so this
is a growing error rather than a bounded one:

| x | kernel returns | `torch.erfc` | rel. error |
|---|---|---|---|
| 5 | 2.415845e-12 | 1.534772e-12 | 5.7e-01 |
| 6 | 2.415845e-12 | 2.146720e-17 | 1.1e+05 |
| 8 | 2.415845e-12 | 1.124127e-29 | 2.2e+17 |
| 20 | 2.415845e-12 | 5.4e-176 | 4.5e+163 |

The clamp was introduced in #41850, which applied the same treatment to both
ops: *"Both erf and erfc clamp input `|x|` using branchless `vec_min_max`
(erf: |x| ≤ 10, erfc: |x| ≤ 5)"*. For **erf** that is correct — `erf`
genuinely saturates at `1.0`, so clamping the input preserves the output. For
**erfc** it is not: `erfc(x) -> 0`, and pinning the input pins the output at a
nonzero constant instead of letting it decay.

Separately, the second LUT segment `[2.5, 5.0]` is inaccurate enough to fail
the LLK domain test on its own, independent of the clamp (details below).

### Why CI has never caught either problem

Both the kernel's accuracy validation and the LLK sweep stop exactly where the
problem starts:

- **#41850's own accuracy table** measured erf over `[-10, 10]` but erfc
  *"exhaustive, 33,090 values in **[-5, 5]**"* — precisely the clamp boundary,
  so no sample ever crossed it.
- **`sfpu_domains.py`** samples `Erfc` on `[-3.0, 3.0]`. That never reaches the
  clamp, and only barely enters segment 1 (`[2.5, 5.0]`).

There is also a tolerance effect worth flagging, because it means simply
widening the domain will not reliably catch this. `passed_test` uses
`atol=0.05, rtol=0.05` plus `PCC > 0.99`, and skips the PCC gate entirely when
`golden.abs().max() < PCC_SIGNAL_FLOOR` (1e-6). Since `erfc(x) < 1e-6` for
`x > 3.5`, a stimulus placed **entirely** above the clamp has no signal, PCC is
skipped, and `atol=0.05` passes trivially against values of order `1e-12`. That
is why `[5.0, 10.0]` **passes** in the sweep below despite the kernel being
wrong by 17 orders of magnitude there.

### To Reproduce

On `ttsim` (no TT hardware needed), widening only the `Erfc` stimulus range in
`tt_metal/tt-llk/tests/python_tests/helpers/sfpu_domains.py`:

```
### erfc
  low=-3.0, high=3.0    1 passed     <- current CI domain
  low=2.5, high=5.0     1 failed
  low=2.5, high=5.0     1 failed     <- repeated, deterministic
  low=2.5, high=3.5     1 failed
  low=3.0, high=4.0     1 passed
  low=5.0, high=10.0    1 passed     <- passes only because PCC is skipped

### erf control -- same clamp shape, saturation IS correct for erf
  low=-3.0, high=3.0    1 passed
  low=2.5, high=5.0     1 passed
  low=-10.0, high=10.0  1 passed
```

Command used for each row:

```bash
export TT_METAL_SIMULATOR=/path/to/libttsim_wh.so
export TT_METAL_DISABLE_SFPLOADMACRO=1
python3 -m pytest --run-simulator \
  "test_sfpu_unary.py::test_eltwise_unary_sfpu_domain[formats:Float16_b->Float16_b-approx_mode:No-mathop:Erfc-dest_acc:No-input_dimensions:[64, 64]]"
```

The **erf control row matters**: erf goes through the identical
clamp-then-rational structure and passes on `[2.5, 5.0]`, which rules out "the
domain test is simply strict in that band" as an explanation for the erfc
failures.

### Segment 1 accuracy, measured separately

Evaluating `ERFC_LUT` on the host in fp64 (a model of the kernel, used only to
localize the failure the simulator already demonstrated):

```
max bf16 ULP error by band
  [0.0, 2.5]   61 ULP     <- segment 0, the band CI actually samples
  [2.5, 3.0]   61 ULP
  [3.0, 5.0]  146 ULP     <- segment 1
  [5.0, 8.0]  5.5e19 ULP  <- the clamp
```

`Q(x)` has no root in `[2.5, 5.0]` (`Q(2.5) = -2.7e-3`, `Q(5) = -9.9`), so this
is fit quality in segment 1, not a pole.

The same host model reproduces all five simulator verdicts, which is what makes
it usable for localizing the cause:

| domain | golden max | PCC gate | PCC | verdict |
|---|---|---|---|---|
| [-3.0, 3.0] | 2.0e+00 | applied | 1.0000 | pass |
| [2.5, 5.0] | 4.1e-04 | applied | 0.9857 | **fail** |
| [2.5, 3.5] | 4.1e-04 | applied | 0.9821 | **fail** |
| [3.0, 4.0] | 2.2e-05 | applied | 0.9958 | pass |
| [5.0, 10.0] | 1.5e-12 | **skipped** | n/a | pass |

`[2.5, 5.0]` sits at PCC ~0.986 against a 0.99 threshold — far enough below it
to fail on every draw rather than intermittently, which matches the repeated
runs above.

### Expected behavior

`erfc` should keep decaying past `|x| = 5` instead of returning a constant.
`torch.erfc` underflows bf16 to `0.0` between `x = 9.489` and `x = 9.490`, so in
bf16 everything above the fit range is either tiny or exactly zero — returning
`0.0` there would be correct to well within a bf16 ULP, and is the same shape as
erf's saturation, just to `0` rather than to a constant. (fp32 needs more care:
`erfc(8) ~ 1.1e-29` is representable there.) The negative side is already fine —
`2 - erfc(|x|)` saturates to `2.0` correctly.

### Please complete the following environment information

- OS: Ubuntu 22.04 (container)
- Device: **ttsim** simulator (`libttsim_wh.so`), Wormhole B0 target
- Measured on `upstream/main` @ `6b291d0c107`

### Additional context

I have not opened a PR for this. The fix touches a shipped op with a
performance budget attached (#41850 tracked erfc device time to ±3%), and there
are at least two reasonable choices — clamp the *output* to `0` above the fit
range, or extend the fit — that trade accuracy against those FMA counts. That
call belongs to whoever owns the kernel. Happy to prepare the patch and re-run
the ttsim sweep if a direction is picked.

One follow-up worth considering regardless of the fix: the erfc entry in
`sfpu_domains.py` cannot detect a regression above `x ~ 3.5`, because the
golden values fall under `PCC_SIGNAL_FLOOR` and `atol=0.05` dominates. A
relative check (or a tighter `custom_atol` via `DOMAIN_CUSTOM_TOLERANCES`) would
be needed for the widened domain to actually assert anything.

kinginu:
### PR Category

Bug fix.

### Summary

Closes #51137.

The rational and its 25-entry LUT are replaced by a degree-5 polynomial in `x²` below 1.5 and the asymptotic `exp(-x²)/x · P(1/x²)` above it, and the input clamp moves from 5 to 9.3.

The handover sits at 1.5 rather than at 5 because the rational's second piece is already wrong before its own boundary — 50.6 BF16 ULP at 2.5 and 197.6 at 5, against 1.07 for the first piece — so extending it would have shipped a 1.9× step at the seam. The asymptotic holds far below where one is normally trusted: 0.0073 BF16 ULP from 1.5, and still 0.27 from 1.0, measured in float64.

The clamp at 9.3 is not a smaller version of the reported defect. The discriminator is whether the true function still moves by more than one output ULP across the clamped run, and `erfc` leaves the smallest FP32 normal at about 9.25 and is 2.1e-45 at 10, so every input above the bound has the same correct answer, 0. `ckernel_sfpu_erf.h` clamps at 10 on the same grounds.

### Accuracy

Blackhole P300, exhaustive over all 65536 bfloat16 patterns, reference `torch.special` in float64, max BF16 ULP:

| region | before | after |
|---|---|---|
| `\|x\| <= 5` | 197.6 | 2.007 |
| `5 < x <= 9.1945` | 3.2e28 | 2.686 |
| `x < -5` | 1.4e-10 | 1.4e-10 (unchanged) |
| true value underflows | 0 of 15976 correct | 15976 of 15976 |
| distinct outputs above 5 | 1 | 116 |

Known residual: for `x` in `(9.25, 9.45]` the true value is an FP32 denormal and the device flushes it, so 4 bfloat16 patterns return 0 where a vanishingly small answer exists.

### Cost

JIT-compiled TRISC1 math ELF `.text`, each arm built in its own `TT_METAL_CACHE` and compared with `objcopy --only-section=.text` (a whole-ELF hash is useless here — the ELF embeds its own build path, which differs between the two caches):

| kernel | main | this PR | |
|---|---|---|---|
| `erfc` BF16 | 1080 B | 1056 B | **−24 B** |
| `erfc` FP32 | 1148 B | 1124 B | **−24 B** |

Smaller in both, because the rational and its 25-entry LUT go away. Wall clock at `[16, 8, 192, 384]`, 200 reps × 5 runs: 170.3 → 172.9 µs. `grep -rn "ttnn.erfc" models/` returns 0.

### Tests

`erfc`'s existing coverage draws from `[0, 1)`, inside the accurate first segment, so it could not see this; the new `test_unary_erfc.py` covers the decay, the underflow and the negative tail, and fails on unmodified `main`. `test_math.py` already checks `ttnn.erfc` at `ulp=1` and still passes. The two files give 49 passed, 0 failures.

### One accuracy/speed knob

The asymptotic takes the reciprocal at zero Newton iterations. That, not the polynomial fits, is what sets the residual: the low path uses no reciprocal and measures 186 FP32 ULP, while the asymptotic sits at 8.9e4. Two iterations measure:

| | this PR | 2 iterations |
|---|---|---|
| BF16 max ULP, `\|x\| <= 5` | 2.007 | 1.261 |
| BF16 max ULP, `5 < x <= 9.19` | 2.686 | 1.349 |
| FP32 max ULP, `5 <= x <= 9.19` | 1.15e5 | 2.88e4 |
| wall clock | 173.5 µs | 182.4 µs |
| `.text`, FP32 kernel | 1124 B | 1160 B |

Two iterations buy 1.6x on the BF16 tail and 4x on FP32, for +5% of wall clock and +36 bytes of `.text`. The cheaper one is in because the reported defect is a constant tail rather than an accuracy shortfall, and 2 ULP is already 100x better than the 197.6 it replaces. Happy to switch.

### Notes for reviewers

- **Wormhole is not measured** — there is no Wormhole hardware on this host. The kernel changes identically and the WH copy stays byte-equivalent to the BH one.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_unary_erfc.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_unary_erfc.py
@@ -1,0 +1,86 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tail coverage for ttnn.erfc.
+
+`test_math.py::test_erfc` draws random bf16 inputs from [0, 1), which is inside the
+rational fit's accurate first segment. It therefore cannot see either of the two
+things measured here: that the rational's second segment [2.5, 5] is inaccurate,
+and that past the fit domain the op used to return a single constant for every
+input. Both are asserted against a golden computed from the *unclamped* input.
+"""
+
+import torch
+
+import ttnn
+
+# Measured on Blackhole P300, exhaustive over all 65536 bfloat16 patterns:
+# 2.007 max BF16 ULP for |x| <= 5 and 2.686 for the asymptotic tail. 4 leaves
+# headroom without admitting the 197 ULP the rational produced at x = 5.
+_MAX_BF16_ULP = 4.0
+
+# erfc leaves bfloat16 near 9.45, but the last stretch of that is denormal in FP32,
+# where the device flushes to zero. 9.2 is the largest input whose true value is
+# still an FP32 normal.
+_LARGEST_NORMAL_INPUT = 9.2
+
+
+def _run(values, device, dtype=ttnn.bfloat16):
+    padded = torch.zeros((1, 1, 32, 32), dtype=torch.float32)
+    padded[0, 0, 0, : values.numel()] = values
+    t = ttnn.from_torch(padded, layout=ttnn.TILE_LAYOUT, dtype=dtype, device=device)
+    return ttnn.to_torch(ttnn.erfc(t))[0, 0, 0, : values.numel()].float()
+
+
+def _bf16_ulp(got, ref):
+    ulp = torch.exp2(torch.floor(torch.log2(ref.abs())) - 7.0)
+    return ((got - ref).abs() / ulp).max().item()
+
+
+def test_erfc_decays_past_the_fit_domain(device):
+    """erfc keeps decaying past the rational's domain; it must not go flat."""
+    values = torch.arange(2.5, _LARGEST_NORMAL_INPUT + 0.01, 0.25, dtype=torch.float32)
+    ref = torch.special.erfc(values.to(torch.bfloat16).double()).float()
+    assert (ref > 0).all(), "every reference here must still be a normal number"
+
+    got = _run(values, device)
+
+    # Distinctness first: the failure this guards against is a constant, and a
+    # tolerance alone would report it only as a large number.
+    assert got.unique().numel() == values.numel(), (
+        f"erfc returned {got.unique().numel()} distinct values for {values.numel()} distinct "
+        f"inputs in [2.5, {_LARGEST_NORMAL_INPUT}]; it is constant where it should be decaying"
+    )
+    max_ulp = _bf16_ulp(got.double(), ref.double())
+    assert max_ulp <= _MAX_BF16_ULP, f"MaxULP={max_ulp:.2f} exceeds {_MAX_BF16_ULP}"
+
+
+def test_erfc_underflows_to_zero(device):
+    """Past the point erfc leaves the format, the only correct answer is 0."""
+    values = torch.tensor([9.5, 10.0, 20.0, 100.0, 3.0e38, float("inf")], dtype=torch.float32)
+    got = _run(values, device)
+    assert torch.equal(
+        got, torch.zeros_like(got)
+    ), f"erfc must underflow to 0 once it leaves the format; got {got.tolist()} for {values.tolist()}"
+
+
+def test_erfc_negative_tail_is_two(device):
+    """erfc(-x) = 2 - erfc(x), and for x <= -5 that is exactly 2 in both formats."""
+    values = torch.tensor([-5.0, -6.0, -10.0, -20.0, -3.0e38, float("-inf")], dtype=torch.float32)
+    got = _run(values, device)
+    assert torch.equal(got, torch.full_like(got, 2.0)), f"got {got.tolist()}"
+
+
+def test_erfc_bfloat8_b_tail_is_a_format_limit_not_a_kernel_one(device):
+    """erfc's tail is not representable in bfloat8_b, and that is the format's doing.
+
+    bfloat8_b shares one exponent across a block, so a block that contains erfc(0) = 1
+    flushes everything below roughly 2^-8 of it to zero. Measured: the same stimuli that
+    give 4.04e-04 ... 4.16e-23 in bfloat16 give all zeros in bfloat8_b. The op is bound
+    for bfloat8_b, so this is asserted rather than left as folklore -- it is why the
+    decay is checked in bfloat16 only.
+    """
+    values = torch.arange(2.5, 7.01, 0.5, dtype=torch.float32)
+    assert (_run(values, device, dtype=ttnn.bfloat16).unique().numel()) == values.numel()
+    assert torch.equal(_run(values, device, dtype=ttnn.bfloat8_b), torch.zeros(values.numel()))
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_erfc.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_erfc.h
@@ -4,71 +4,75 @@
 
 #pragma once
 
-#include <array>
-
 #include "ckernel.h"
 #include "ckernel_defs.h"
 #include "cmath_common.h"
-#include "sfpu/ckernel_sfpu_converter.h"
-
-#include "ckernel_sfpu_piecewise_rational.h"
+#include "ckernel_sfpu_exp.h"
+#include "ckernel_sfpu_recip.h"
+#include "sfpu/ckernel_sfpu_polyval.h"
 
 namespace ckernel::sfpu {
 
 // ======================================================================
-// LUT-based erfc via piecewise rational P(x)/Q(x)
+// erfc(x), two plain polynomials and no rational.
 //
-// Uses abs(x) symmetry: erfc(-x) = 2 - erfc(x)
-// Fit on [0, 5.0] only, 2 segments with n4/d5 rational per segment.
-// BF16 MaxULP=118 (was 128 with 3-seg n4/d4 on [-5,5])
-// FP32 MaxULP≈9M  (was 1.47B)
-// 18 FMAs          (was 24)
+// Uses abs(x) symmetry: erfc(-x) = 2 - erfc(x).
+//   |x| <  1.5:  erfc(x) = 1 - x*g(x^2), g a degree-5 minimax in x^2
+//   |x| >= 1.5:  erfc(x) = exp(-x^2)/x * P(1/x^2), P absorbing 1/sqrt(pi)
+//
+// The asymptotic is accurate far below where an asymptotic expansion is
+// normally trusted: measured in float64 it holds to 0.0073 BF16 ULP from 1.5,
+// and still to 0.27 BF16 ULP from 1.0. Letting it take everything above 1.5
+// leaves the low region an easy fit, which is what removes the piecewise
+// rational and its LUT.
+//
+// Inputs are clamped to 9.3 before squaring, so every lane -- including the ones
+// whose branch result is discarded -- satisfies the unsafe exp's precondition,
+// which holds while -x^2 stays above -88.03. Past ~9.25 the result underflows to
+// 0 on its own, which is the correct saturation for a function decaying to 0, so
+// the bound costs nothing: erfc(9.3) is already below the smallest FP32 normal.
 // ======================================================================
 
-constexpr uint32_t ERFC_NUM_DEGREE = 4;
-constexpr uint32_t ERFC_DEN_DEGREE = 5;
-constexpr uint32_t ERFC_NUM_SEGMENTS = 2;
-constexpr uint32_t ERFC_LUT_SIZE = 25;
-constexpr std::array<float, ERFC_LUT_SIZE> ERFC_LUT = {{// Breakpoints
-                                             0.0000000000e+00f,
-                                             2.5000000000e+00f,
-                                             5.0000000000e+00f,
-                                             // Segment 0 [0, 2.5]: numerator (degree 4)
-                                             1.0000233650e+00f,
-                                             -1.3375675678e+00f,
-                                             6.8185544014e-01f,
-                                             -1.5691982210e-01f,
-                                             1.3746744953e-02f,
-                                             // Segment 0 [0, 2.5]: denominator (degree 5)
-                                             1.0000000000e+00f,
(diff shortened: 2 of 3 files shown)


fvranicTT:
I have a proposal to the current state of this PR (rev1), proposing the ordering + CReg version (rev2).
Accuracy is untouched, this is purely a cycles change.

**Accuracy: bit-identical.** Not just close. Checked against the disassembly rather than assuming: the CReg split keeps the same Horner association (`C5*x2+C4`, then `*x2+C3`, then the outer three), and hoisting `xlog2` out of exp21f emits the same SFPMUL/SFPADDI pair in the same order. Every ULP number from rev1 carries over unchanged.

**Perf**, per row of 32 datums:

|                               | main | rev1 | rev2 |
|-------------------------------|------|------|------|
| SFPU instrs                   | 77   | 72   | 67   |
| SFPLOADI                      | 40   | 33   | 27   |

The reason this is worth doing: rev1 drops 5 instructions but goes from 5 to 13 back-to-back dependent ops, which on a +1 cycle stall model puts it at 85 vs main's 82. So rev1 is arguably a small regression, and the instruction count alone hides that. rev2 lands at 76, roughly 7% under main and 11% under rev1.

What changed:
- three highest coefficients move to CRegs, set once in `erfc_init` (init goes from 1 to 10 SFPU instrs, pays for itself inside the first 8-row block). Drops 6 SFPLOADI per row.
- exp21f split so the caller forms `xlog2`, which lets independent work sit in the MUL/ADDI shadow.
- Asymptotic path reordered: recip issues before the square, P(t) before the exp tail.
- `abs_x - thresh` computed early so the compare covers the first CReg MAD.

```cpp
constexpr float ERFC_ASYMPTOTIC_THRESHOLD = 1.5f;
constexpr float ERFC_MAX_INPUT = 9.3f;

// Low-path g(u), u = x^2. Highest-degree three live in CRegs (see erfc_init).
constexpr float ERFC_G_C0 = 1.1283125367e+00f;
constexpr float ERFC_G_C1 = -3.7556339817e-01f;
constexpr float ERFC_G_C2 = 1.1125811263e-01f;
constexpr float ERFC_G_C3 = -2.4775019023e-02f;
constexpr float ERFC_G_C4 = 3.7510146011e-03f;
constexpr float ERFC_G_C5 = -2.8441375985e-04f;

// Asymptotic branch, outlined so its intermediates do not stay live across the
// polynomial path. Instruction order is the BH schedule: 1/ln2 LOADIs cover
// SFPSWAP, recip sits before the square, and P(t) covers the exp21f integer
// conversion's SFPMAD/SFPADDI shadows.
sfpi_inline sfpi::vFloat calculate_erfc_asymptotic_(const sfpi::vFloat abs_x) {
    const sfpi::vFloat axc = sfpi::min(abs_x, ERFC_MAX_INPUT);
    constexpr float ONE_LN2 = 1.4426950216293334961f;  // matches exp21f
    const sfpi::vFloat one_ln2 = ONE_LN2;
    // Zero Newton iterations, spelled as the iteration count rather than as
    // sfpu_reciprocal<true>: the reciprocal refines an asymptotic whose residual is set by
    // other terms, so the approximate seed costs ~2 ULP rather than orders of magnitude.
    const sfpi::vFloat inv = sfpu_reciprocal_iter<0>(axc);
    const sfpi::vFloat neg_x2 = -(axc * axc);
    const sfpi::vFloat t = inv * inv;
    const sfpi::vFloat xlog2 = neg_x2 * one_ln2 + 127.f;
    // P(t), t = 1/x^2, degree 4, minimax on [1/100, 1/2.25]; 0.0073 BF16 ULP
    // measured after rounding the coefficients to float32. Carries 1/sqrt(pi).
    const sfpi::vFloat correction = PolynomialEvaluator::eval(
        t, 5.6414234638e-01f, -2.7857559919e-01f, 3.5396602750e-01f, -4.4328993559e-01f, 2.8342172503e-01f);
    const sfpi::vFloat exp_value = _sfpu_exp_21f_bf16_from_xlog2_<true>(xlog2);
    return exp_value * inv * correction;
}

template <int ITERATIONS = 8>
inline void calculate_erfc() {
    for (int d = 0; d < ITERATIONS; d++) {
        sfpi::vFloat x = sfpi::dst_reg[0];
        const sfpi::vFloat abs_x = sfpi::abs(x);

        // Low path, computed unconditionally: erfc(x) = 1 - x*g(x^2).
        // g is fitted on [0, 2.25] in x^2; 0.0035 BF16 ULP measured.
        const sfpi::vFloat x2 = x * x;
        // Innermost three of g live in CRegs. The 1.5 LOADI covers x^2, and
        // the compare covers the first CReg MAD, so those two MADs are not
        // back-to-back with each other or with the square (BH SFPMAD/SFPMUL
        // result latency is 2).
        const sfpi::vFloat thresh = ERFC_ASYMPTOTIC_THRESHOLD;
        sfpi::vFloat g = sfpi::vConstFloatPrgm0 * x2 + sfpi::vConstFloatPrgm1;
        const sfpi::vFloat d_thresh = abs_x - thresh;
        g = g * x2 + sfpi::vConstFloatPrgm2;
        g = PolynomialEvaluator::eval(x2, ERFC_G_C0, ERFC_G_C1, ERFC_G_C2, g);
        sfpi::vFloat r = 1.0f - abs_x * g;

        v_if(d_thresh >= 0.0f) { r = calculate_erfc_asymptotic_(abs_x); }
        v_endif;
        // erfc(-x) = 2 - erfc(x)
        v_if(x < 0.0f) { r = 2.0f - r; }
        v_endif;
        sfpi::dst_reg[0] = r;
        sfpi::dst_reg++;
    }
}

template <bool APPROXIMATION_MODE>
void erfc_init() {
    math::reset_counters(p_setrwc::SET_ABD_F);
    // Reciprocal uses 0 Newton iterations, so Prgm0 is not the NR constant 2.
    // CRegs hold g's three highest-degree coefficients. Each assignment is
    // SFPLOADI lo/hi + SFPCONFIG, once per op; the row loop drops those 6
    // SFPLOADIs.
    sfpi::vConstFloatPrgm0 = ERFC_G_C5;
    sfpi::vConstFloatPrgm1 = ERFC_G_C4;
    sfpi::vConstFloatPrgm2 = ERFC_G_C3;
}
```

kinginu:
Thanks — I built rev2 and measured it. Bit-identical confirmed, all 65536 bfloat16 patterns. Your `main` and rev1 counts reproduce exactly, SFPLOADI 40 and 33, so we are counting the same way.

Wall clock, `[16, 8, 192, 384]` DRAM, 100 reps × 9, median, both measurement orders, Blackhole P300 at `1227e18213e`:

| | wall clock | SFPU | SFPLOADI |
|---|---|---|---|
| `main` | 169.4 µs | 78 | 40 |
| rev1 | 167.1 | 73 | 33 |
| **rev2a** — reorder + exp split, no CRegs | **164.8** | 74 | 33 |
| rev2b — CRegs only | 168.3 | 76 | 33 |
| rev2 | 188.3 | 77 | 33 |

The reordering is a real win — rev2a is the fastest arm, 2.7% under `main`, and bit-identical to rev1. I would like to take it.

The CRegs drop no SFPLOADI here: it stays at 33, and the only disassembly difference between rev2a and rev2 is `sfpconfig` 1 → 4. The halves are also not additive — reorder alone −2.3 µs, CRegs alone +1.2, both +21.2.

The profiler inverts it. TRISC_1 median is 220973 cycles for rev1 and 200037 for rev2, so rev2 is 9.5% faster inside the math zone and 12% slower end to end. Same binary either way, checked with `PROFILE_KERNEL` on and off. I cannot account for that.

One flag: rev2a needs `_sfpu_exp_21f_bf16_from_xlog2_`, so `erfc` would touch `ckernel_sfpu_exp.h` again. This PR was split off from a wider one on review feedback about scope, and `erfc` was made independent of that header on purpose — clamping at 9.3 instead of taking the parameter the `i1` PR (#52202) adds there, bit-identical either way. Taking rev2a puts a merge order back between #52202 and this one. Your call, happy either way.

Would you take rev2a? And if you have a build where the CRegs do drop the six SFPLOADI, I would like to see it.

PR #53040, opened 2026-08-13
[Port] Migrate `moreh_layer_norm_backward`, `moreh_group_norm_backward` to Metal 2.0

### LINKED ISSUE #52408: Metal 2.0 port: moreh/moreh_layer_norm_backward

### LINKED ISSUE #53043: Metal 2.0 port: moreh/moreh_group_norm_backward

VirdhatchaniKN:
### Issue
https://github.com/tenstorrent/tt-metal/issues/53043 , https://github.com/tenstorrent/tt-metal/issues/52408

### Summary

This is the result of AI-driven porting to Metal 2.0 APIs of
`ttnn/cpp/ttnn/operations/moreh/moreh_layer_norm_backward` and
`ttnn/cpp/ttnn/operations/moreh/moreh_group_norm_backward`. No functional or performance change
intended. All four backward factories are ported to `ProgramSpecFactoryConcept`; nothing is left on the
legacy `ProgramDescriptor` concept.

Both ops are in one PR because `moreh_group_norm_backward` owns no compute kernel — all three come from
`moreh_layer_norm_backward`. A shared kernel converts to Metal 2.0 once, and at that moment every
factory binding it must already speak Metal 2.0 or the build breaks, so the atomic unit is one shared
kernel plus all of its binders: gamma_beta_grad (both ops, 5 kernel sources) and input_grad (both ops,
8 sources, reader and compute runtime-selected on `use_large_algorithm`). The three shared kernels are
converted in place rather than forked — the census finds exactly two consumers each and both convert
here, so no `_metal2` fork is needed and no binder is left behind.

Full porting reports:
Linked in first commit that added them, since commit 2 removes them from the branch tip (they are
vetting artifacts, not shipping code). Port plans and pre-port audits sit beside them at the same SHA.

### Notes for reviewers

Gen1 only; no Gen2/Quasar concerns addressed. One asymmetry worth recording for whoever runs the Gen2
uplift: layer-norm reaches `unpack_modes` through the `std::visit`-based free accessors, so it needs no
work there, while group-norm's hand-built `ComputeGen1Config{}` is a site that pass will have to touch.

Three silent-perf settings are worth a look, since they pass every test whether right or wrong:
`opt_level` is pinned to `O3` on all four compute specs (legacy `ComputeConfig` defaulted to `O3`,
Metal 2.0 defaults to `O2`); layer-norm's compute config goes through `ttnn::to_compute_hardware_config`
while group-norm's all-default `ComputeConfigDescriptor{}` builds `ComputeGen1Config{}` directly, since
the TTNN helper's defaults would silently flip every field; and Float32 `unpack_modes` entries were
added where the validator now requires them, all `UnpackToSrc` to match the legacy default. Details and
the five pre-existing findings shipped forward unfixed are in the reports.

`tests/ttnn/nightly/unit_tests/operations/moreh/test_moreh_layer_norm.py` and `test_moreh_group_norm.py`:
**108 passed, 123 skipped**, against a pre-port baseline of 101 passed / 122 skipped — the delta is the
new coverage described below, and every pre-existing case matches the baseline exactly. The full moreh
nightly suite (all 25 files) is 3581 passed, 3171 skipped, 0 failures, run because `ttnn_op_moreh` is a
unity-build target and these factories share translation units with other moreh ops.

The last commit adds coverage for three paths that had **none before this PR** — layer-norm never
selected its large input_grad algorithm on any tested shape, and group-norm reached it only through
tile-aligned shapes, so `reader_moreh_layer_norm_backward_input_grad_large.cpp` was never executed and
neither large-algorithm kernel ever ran with a mask buffer bound. Since that was already true of the
pre-port baseline, "matches the baseline" would not have defended those files. Two parametrizations
sized past the L1 selection threshold, with unaligned extents so both mask predicates fire, close all
three.

One finding carried forward is worth its own issue rather than a report paragraph: the input_grad
**small** kernel ends with a second `wait_front(2)` on the mask buffer where the **large** kernel has
`pop_front(2)`. It is harmless today (fill-once/read-many reuse means the entries are already
available), and was preserved verbatim because a port must not change behavior — but it reads as a
typo. An issue will be filed against the op owners and addressed separately after this merges.

Pipeline status
 

- [x] Sanity
- [x]  (Runtime) Sanity Test
- [x]  Nightly for moreh Category
- [x] T3K
- [x] Single card

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/moreh_bw_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/moreh_bw_port)


ttnn/cpp/ttnn/operations/moreh/moreh_group_norm_backward/device/input_grad/kernels/dataflow/reader_moreh_group_norm_backward_input_grad_large.cpp
```diff
@@ -63,56 +44,67 @@ void kernel_main() {
...
+#ifdef GAMMA_HAS_VALUE
+    const auto gamma_tile_bytes = dfb_gamma.get_tile_size();
+#endif
 
-    const auto output_grad_tile_bytes = get_tile_size(cb_id_output_grad);
-    const auto input_tile_bytes = get_tile_size(cb_id_input);
-    const auto rstd_tile_bytes = get_tile_size(cb_id_rstd);
-    const auto gamma_tile_bytes = get_tile_size(cb_id_gamma);
+    const auto mean_dtype_bytes = mean_tile_bytes / (TILE_H * TILE_W);
+    const auto rstd_dtype_bytes = mean_tile_bytes / (TILE_H * TILE_W);
```
Aswinmcw: `rstd_dtype_bytes` is derived from `mean_tile_bytes` rather than `rstd_tile_bytes` (same in `reader_moreh_group_norm_backward_input_grad_small.cpp:101`). Harmless today since mean and rstd always share a dtype, but worth correcting while these lines are being rewritten.

Aswinmcw (approved):

Aswinmcw (approved):

mcw-anasuya (approved):

aczajkowskiTT (approved):

bbradelTT (approved):

PR #53054, opened 2026-08-13
[Performance] Give Quasar dispatch engines their own L1 memory layout

sagarwalTT:
### PR Category

Performance

### Summary

Quasar dispatch engines were deriving their L1 memory map from the Tensix one and patching a few fields afterwards. A dispatch engine has no TRISCs, so that borrowed layout reserved L1 space that could never be used, and all of it came out of the L1 budget available to DE kernels.

This change derives the DE layout independently instead of inheriting it, reclaiming ~88 KB per dispatch engine:
- The unused portion of the mailbox, which was sized for the 16 TRISC processors in addition to the 8 DM processors, instead of just being sized for the 8 DM processors (~35 KB)
- TRISC firmware (20 KB)
- TRISC globals (16 KB)
- The TRISC portion of the init-local staging area (16 KB)
- The LLK debug region (1 KB)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:sagarwal/qsr_de_optimize_l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:sagarwal/qsr_de_optimize_l1)


arikTT (approved):

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -51,15 +51,25 @@
...
-#define MEM_TRISC_KERNEL_SIZE (1024 * 24)
 #define MEM_DM_GLOBAL_SIZE (1024 * 2)
-#define MEM_TRISC_GLOBAL_SIZE (1024 * 2)
 #define MEM_DM_LOCAL_SIZE (1024 * 8)
-#define MEM_TRISC_LOCAL_SIZE (1024 * 4)
+#define MEM_TRISC_KERNEL_SIZE (1024 * 24)
 #define MEM_TRISC_LOCAL_OFFSET (0x2000)
 
+// A dispatch engine has no TRISCs and nothing on it uses the LLK debug region.
+#if defined(COMPILE_FOR_DISPATCH_ENGINE)
```
jbaumanTT: On other platforms we don't make these mem map defines depend on what sort of risc they're being compiled on.  You really shouldn't need to do these redefinitions since you already have "MEM_DISPATCH_INTERRUPT_TABLE_BASE" that doesn't depend on MEM_LLK_DEBUG_SIZE, for example. Are there any cases where having them always be the correct size would cause a problem?

sagarwalTT: These defines are used during linking for `kn_data`, `text_size`, etc. in `qa_hal.cpp`, however, I'm not sure if having these value be non-zero when linking while not actually reserving anything on L1 would cause any issues. The main reason I implemented different values for these defines based on `COMPILE_FOR_DISPATCH_ENGINE` is just for consistency.

jbaumanTT: The trisc defines shouldn't actually be used when linking  DM kernels or firmware, so their values shouldn't really matter.

tt-vjovanovic (approved):

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
sagarwalTT: @jbaumanTT Can you take another look at this PR before I merge it

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -224,8 +257,53 @@
...
+#undef MEM_FABRIC_COUNTER_BASE
+#define MEM_FABRIC_COUNTER_BASE MEM_DISPATCH_FABRIC_COUNTER_BASE
+#undef MEM_FABRIC_CONNECTION_LOCK_BASE
+#define MEM_FABRIC_CONNECTION_LOCK_BASE MEM_DISPATCH_FABRIC_CONNECTION_LOCK_BASE
+#undef MEM_TENSIX_ROUTING_TABLE_BASE
+#define MEM_TENSIX_ROUTING_TABLE_BASE MEM_DISPATCH_TENSIX_ROUTING_TABLE_BASE
+#undef MEM_TENSIX_ROUTING_PATH_BASE
+#define MEM_TENSIX_ROUTING_PATH_BASE MEM_DISPATCH_TENSIX_ROUTING_PATH_BASE
+#undef MEM_TENSIX_ROUTING_PATH_BASE_1D
+#define MEM_TENSIX_ROUTING_PATH_BASE_1D MEM_DISPATCH_TENSIX_ROUTING_PATH_BASE
```
jbaumanTT: Modify ttt_metal/hostdevcommon/api/hostdevcommon/fabric_common.h to reference the correct define on DE rather than change the define here.

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -224,8 +257,53 @@
...
+#undef MEM_MAILBOX_SIZE
+#define MEM_MAILBOX_SIZE MEM_DISPATCH_MAILBOX_SIZE
+#undef MEM_MAILBOX_END
+#define MEM_MAILBOX_END MEM_DISPATCH_MAILBOX_END
+#undef MEM_INTERRUPT_TABLE_BASE
+#define MEM_INTERRUPT_TABLE_BASE MEM_DISPATCH_INTERRUPT_TABLE_BASE
+#undef MEM_DM_FIRMWARE_BASE
+#define MEM_DM_FIRMWARE_BASE MEM_DISPATCH_DM_FIRMWARE_BASE
+#undef MEM_DM_GLOBAL_BASE
+#define MEM_DM_GLOBAL_BASE MEM_DISPATCH_DM_GLOBAL_BASE
```
jbaumanTT: This is only referenced on the host in tt_metal/llrt/hal/tt-2xx/quasar/qa_hal.cpp. You shouldn't need to redefine it here.

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -224,8 +257,53 @@
...
+#undef MEM_DM_LOCAL_BASE
+#define MEM_DM_LOCAL_BASE MEM_DISPATCH_DM_LOCAL_BASE
+#undef MEM_NOC_COUNTER_BASE
+#define MEM_NOC_COUNTER_BASE MEM_DISPATCH_NOC_COUNTER_BASE
```
jbaumanTT: Don't redefine things. If you need to have different definitions, make a separate define like CURRENT_MEM_NOC_COUNTER_BASE that's defined on both dispatch engine and non-dispatch-engine, and reference it at the users.

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -224,8 +257,53 @@
...
+#undef MEM_MAP_END
+#define MEM_MAP_END MEM_DISPATCH_MAP_END
+#undef MEM_KERNEL_CONFIG_SIZE
+#define MEM_KERNEL_CONFIG_SIZE MEM_DISPATCH_KERNEL_CONFIG_SIZE
+#undef MEM_DM0_INIT_LOCAL_L1_BASE_SCRATCH
+#define MEM_DM0_INIT_LOCAL_L1_BASE_SCRATCH MEM_DISPATCH_DM0_INIT_LOCAL_L1_BASE_SCRATCH
+#undef MEM_BANK_TO_NOC_SCRATCH
+#define MEM_BANK_TO_NOC_SCRATCH MEM_DISPATCH_BANK_TO_NOC_SCRATCH
+#undef MEM_LOGICAL_TO_VIRTUAL_SCRATCH
+#define MEM_LOGICAL_TO_VIRTUAL_SCRATCH MEM_DISPATCH_LOGICAL_TO_VIRTUAL_SCRATCH
```
jbaumanTT: Is this even used anywhere?

jbaumanTT (approved):

ubcheema (approved):

PR #53059, opened 2026-08-13
Uncollecting deepseek prefill combine test cases instead of skipping them

dlukicTT:
### Summary
Moving skip tests logic to collection phase, so that test cases we know in CT we want to skip are not even collected, thus giving us more realistic test case count when --collect-only'ing.

Also, removing extended-model flag (set in combine parametrization and consumed in CI yaml files), and removing unintended mesh configs for non dsv3 models, instead.

Change also removes 1/2 links parametrization and uses 2 links always, but we want to add some of 1link test cases to cover minimal GptOss support. Once that is in place the PR will be published.

For easier overview of effects on the collection - https://claude.ai/code/artifact/4edcf7fd-25c0-4dec-b612-45ff687f1080

### CI Status
[Blaze CI - affected jobs are green](https://github.com/tenstorrent/tt-metal/actions/runs/32731794058)
[BH e2e CI - as green as on main](https://github.com/tenstorrent/tt-metal/actions/runs/32744761770)
[BH e2e CI for reference](https://github.com/tenstorrent/tt-metal/actions/runs/32722640876)


models/demos/deepseek_v3_d_p/tests/conftest.py
```diff
@@ -371,20 +399,21 @@ def _get_requested_fabric_cfg(item):  # returns fabric cfg that a particular tes
...
-        if cluster_type in CI_ALLOWED_FABRICS.keys():
-            allowed_fabric_dct = CI_ALLOWED_FABRICS[cluster_type]
-            if mesh_shape in allowed_fabric_dct.keys():
-                allowed_fabric_cfgs = allowed_fabric_dct[mesh_shape]
-
-        if requested_fabric_cfg not in allowed_fabric_cfgs:
-            item.add_marker(
-                pytest.mark.skip(
-                    reason="requested combination of fabric config and mesh, unfeasible on the given hardware"
+        if on_ci:
```
pmilojevicTT: Why is this skipped only in CI, but not locally. Do CI machines only have problem with unsupported fabric rings, while local machines (exabox also) don't have this problem?

models/demos/deepseek_v3_d_p/tests/op_unit_tests/test_prefill_combine.py
```diff
@@ -432,17 +404,12 @@ def _fabric_cfg_to_fabric_topo_for_cmb_op(fabric_cfg):
 
 def _cross_product_conflated_cmb_test_dimensions():
     params = []
-    for model_name, model_config_class, is_extended_model, test_meshes in COMBINE_MODELS:
```
pmilojevicTT, line 420: Just please be sure to also remove `extended_model` from whole CI for combine test only.

models/demos/deepseek_v3_d_p/tests/conftest.py
dlukicTT: I wanted to allow people more freedom in local testing, where (my thingking) hang affects only them so if they want to try out something that is not completely safe, they can.
But I can remove it. It's not like I'm sure it's a good idea.

models/demos/deepseek_v3_d_p/tests/op_unit_tests/test_prefill_combine.py
dlukicTT: That would be a follow up. I thought that flag is in CI for the whole MoE, so removing "if not extended model" would have to come after we remove the flag from other test too. I'm not sure my logic is sound there. We can definitely talk more about that one.

mbezuljTT (approved):

mbezuljTT:
Assuming green CI

PR #53066, opened 2026-08-13
Imporving MGD Pinning algoirthm to allow for more fine tuned control

Riddy21:
## Summary

Adds **4×4 SplitHost Middle/Edge** galaxy PGD groupings (revC + revAB) and makes the PGD↔MGD pinning path place split-host and galaxy meshes correctly, with CPU-only test coverage.

### What changed
- **New PGD groupings:** 4×4 SplitHost Middle/Edge galaxy groupings for revC and revAB (a 4×4 galaxy slice split across two host ranks uses disjoint ASIC-location columns — "Middle" {3,4,7,8} and "Edge" {1,2,5,6}).
- **Profile-aware PGD matching:** a candidate grouping is committed only when **every** pinning group of a profile fully lands on it (was: kept if ≥1 pin landed). A mesh may carry two profiles (Middle **and** Edge) and bind to whichever physical column a placement provides; a placement that fits no profile is rejected. This stops single-profile descriptors (e.g. `two_4x4` M0→Middle / M1→Edge) from over-committing hundreds of groupings.
- **Intra-mesh pinning application** (`apply_pinning_groups`): skip a pin group whose ASIC column is absent on the candidate placement, and require at least one profile to apply — so shared 4×4 split-host descriptors match either column.
- **Galaxy corner pins are now ONE many-to-many group** (`get_galaxy_fixed_asic_position_pinnings_for_mesh`): the four corner fabric nodes map bijectively to the four tray-corner ASICs in a single group, instead of four separate single-corner groups. As separate groups the profile layer collapsed the (same-position) corners to one, leaving the mesh orientation under-determined and smearing the 2×4 host-rank slice across all four trays (`TestGalaxyLayoutCheck` failures on `dual_4x16`, `llama_8b_*`). One group applies all four corner constraints together and fixes the orientation.

### Tests (CPU-only, `run_fabric_cpu_only_unit_tests.sh`)
- Added `gemma_4galaxy` (8× 4×4 RING/RING) and single-profile `two_4x4` across the SC4 revAB/revC subtorus quad mocks.
- 2-profile pipeline MGDs for the 8/32/40-stage 4×4 split-host pipelines.
- Fixed the `llama_8b` galaxy-layout test to reference the existing `llama_8b_4galaxy` MGD (CI referenced a non-existent `3galaxy` descriptor).

### Validation
Rebased onto latest `main`. Full local CPU-only fabric suite is green (Release) on latest HEAD `bce57fbd035`: 3847 gtest blocks passed, 0 failures — gemma_4galaxy, two_4x4, 8/32/40-stage pipelines, `dual_4x16`, `llama_8b_1x2/2x1_pod`, `llama_8b_4galaxy`, and all other groups pass.

CPU-only CI run for latest HEAD: https://github.com/tenstorrent/tt-metal/actions/runs/32213979961

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riddy21/fabric-4x4-splithost-middle-edge)


p1-0tr (approved):
LGTM :)

tt_metal/fabric/physical_grouping_descriptor_matching.cpp
```diff
@@ -834,35 +822,39 @@ std::map<LogicalChipId, tt::tt_metal::ASICPosition> compose_mesh_node_to_asic_po
...
-        for (const auto& fabric_node : group.fabric_nodes) {
-            if (applicable_mesh_ids.contains(fabric_node.mesh_id.get())) {
-                filtered_group.fabric_nodes.push_back(fabric_node);
-            }
-        }
-        if (!filtered_group.fabric_nodes.empty()) {
-            filtered.push_back(std::move(filtered_group));
+    for (const auto& group : pinnings) {
+        if (!group.fabric_nodes.empty() && applicable_mesh_ids.contains(group.fabric_nodes.front().mesh_id.get())) {
+            filtered.push_back(group);
```
Riddy21: This will filter out pinnings to make sure only apply to this mesh

tt_metal/fabric/physical_grouping_descriptor_matching.cpp
```diff
@@ -879,7 +871,9 @@ std::size_t add_mgd_to_pgd_asic_position_pinning_constraints(
             pgd_nodes.insert(found_pgd_nodes.begin(), found_pgd_nodes.end());
         }
         if (!mgd_nodes.empty() && !pgd_nodes.empty()) {
-            constraints.add_required_constraint(mgd_nodes, pgd_nodes);
+            if (!constraints.add_required_constraint(mgd_nodes, pgd_nodes)) {
+                return 0;
+            }
```
Riddy21, line 858: All pinnings must apply, if one is missing then it will skip this, this makes profiles more strict to rule out more unsatisfactory cases, makes solving easier

blozano-tt (approved):

tt_metal/fabric/physical_grouping_descriptor_matching.cpp
```diff
@@ -1000,188 +1013,194 @@ ValidGroupingsMap PhysicalGroupingDescriptor::get_valid_groupings_for_mgd(
...
-                    const std::size_t pinning_constraints_added = add_mgd_to_pgd_asic_position_pinning_constraints(
-                        constraints, grouping_info, *applicable_pinnings);
-                    // Ok if no pinning constraints were added, means this grouping might not be best fit
-                    if (pinning_constraints_added == 0) {
+        for (const auto& active_pinnings : pin_set_variants) {
+            for (const auto& [node_diff, name_idx_pairs] : candidates_by_diff) {
+                best_matches_topology.clear();
+                best_matches_psd_placed.clear();
+                best_matches_topology.reserve(name_idx_pairs.size());
+
```
Riddy21, line 1029: Removing a layer of the loop made the diff WAY bigger than it needed to be, very small change actually, just with extra indent

tt_metal/fabric/physical_grouping_descriptor_matching.cpp
p1-0tr: Making the policy on adding constraints makes a lot of sense. One thing we should follow up with (does not need to happen in this PR), is a richer return type, such that we can tell the user which constraints didn't apply and why. If we could get to errors that report which constraint failed to apply, down to line number in the MGD (if it was a `*.textproto`), and why (what was missing in the physical topo, that'd be ideal.

p1-0tr (approved):

PR #50945, opened 2026-07-23
[ttnn] Add validate_program_args config (skip Metal 2.0 host-side arg validation off-CI)

dgomezTT:
### What

Adds a `ttnn.CONFIG` flag **`validate_program_args`** (default `false`) that gates the Metal 2.0 host-side program-arg validation. The spec-factory adapters pass `skip_validation = !validate_program_args` to all three call sites:

- `MakeProgramFromSpec` — cache miss, the Step 1b semantic pass (`ValidateProgramSpec`),
- `SetProgramRunArgs` — cache miss,
- `UpdateProgramRunArgs` — cache hit.

- **Off by default** → prod, models, and perf tests skip the host-side validation overhead.
- **Device-sanity CI turns it on** via `TTNN_CONFIG_OVERRIDES={"validate_program_args": true}` — the same mechanism CI already uses to flip `enable_fast_runtime_mode`. That keeps the validation safety net exactly where regressions are caught, so a bad spec can't silently ship.
- The field is reflection-bound (`reflect::for_each` over `Config::attributes_t`), so it auto-exposes on `ttnn.CONFIG` and via `TTNN_CONFIG_OVERRIDES` / the JSON config with no extra pybind. `Config::validate()` only special-cases named fields, so the new flag falls through with no extra handling.

Note the cache-miss path keeps its structural diagnostics regardless of the flag: `MakeProgramFromSpec` Step 1a (`CollectSpecData`) and the Gen1 multi-binding risc-mask check in Step 2b always run. Only the Step 1b semantic pass is gated.

### CI wiring

The flag is set to `true` in the five device-sanity workflows:

- `.github/workflows/galaxy-sanity-impl.yaml`
- `.github/workflows/ops-sanity-impl.yaml`
- `.github/workflows/runtime-sanity-tests-impl.yaml`
- `.github/workflows/t3000-sanity-tests-impl.yaml`
- `.github/workflows/ttnn-sanity-tests-impl.yaml` (both the job-level env and the `fast_runtime_mode_off` step, which merges both keys into `$GITHUB_ENV` rather than clobbering)

### Perf — validation cost

Host-bound `ttnn.rand` (32×32-tile grid, **cache-hit dispatch**, median of 5 trials × 300 iters), spec-override path:

| validate_program_args | 512² µs/op |
|---|---|
| `true` (CI mode) | 75.8 |
| `false` (default) | 60.7 |

[...]

PR #53067, opened 2026-08-13
[ttnn] Parse TTNN_CONFIG_OVERRIDES in C++ so pure C++ consumers honor it

dgomezTT:
### What

`TTNN_CONFIG_PATH` and `TTNN_CONFIG_OVERRIDES` were both handled in `ttnn/__init__.py`, so they only ever
applied to Python processes. The five device-sanity workflows set
`TTNN_CONFIG_OVERRIDES='{"validate_program_args": true}'` (#50945) at job level — but the C++ gtests in those
same jobs silently ran with validation off. Same trap for the pre-existing `fast_runtime_mode_off` wiring, and
for any C++ consumer of a `TTNN_CONFIG_PATH` file.

This moves both configuration sources into C++ and has Python consume them:

- **`Config::apply_json_overrides(json_text, strict, source)`** in `ttnn/core/config.cpp`: nlohmann parse plus
  the same `reflect::for_each` walk the class already uses for printing/Inspector. Keys are applied into a copy
  of `attributes_t` and committed once, so a strict rejection changes nothing at all, while a non-strict caller
  (a config file) keeps every key that parsed and warns about the rest. `validate()` runs once per applied key.
  `source` names the env var or file path in every message.
- **`Config::load_from_file` / `Config::save_to_file`**: the reader/writer pair for the `TTNN_CONFIG_PATH`
  format. Serialization used to exist only in Python; it now has one owner.
- **`Config::keys()`**: the single enumeration of overridable attributes, so nothing has to filter
  `dir(CONFIG)` to find them.
- A static initializer applies the file first and the env var on top — the order `__init__.py` used — so pure
  C++ consumers (gtests, standalone binaries) get the same configuration as Python.
- nanobind exposes all four. `load_config_from_dictionary` becomes a one-line delegation that keeps raising
  `ValueError`; `load_config_from_json_file` / `save_config_to_json_file` delegate too, and the whole
  `if CONFIG_PATH ... / if CONFIG_OVERRIDES ...` block in `__init__.py` is gone.

### Behavior changes

- A bad `TTNN_CONFIG_OVERRIDES` (malformed JSON, unknown key, wrong-typed value) fails the process at library
  load: the reason is logged and the process exits `1`. Previously this was a Python exception at
  `import ttnn`, and C++-only processes ignored the variable entirely. Deliberate: a typo in a workflow file
  must not silently disable the validation safety net. The exit is `_Exit`, not a throw — an escaping throw
  ends in SIGABRT before `main`, and a plain `std::exit` deadlocks against the loader lock.
- A `TTNN_CONFIG_PATH` file is now read and written by any process that links ttnn, not just Python. In
  particular, pointing it at a missing path makes that process create the file, seeded from default
  configuration values. Nothing in CI sets `TTNN_CONFIG_PATH` (only the visualizer and model-bringup docs).
- A stale or corrupt config file warns and applies whatever parsed, rather than failing the load — including
  the computed `report_path` entry that older saved files contain. Newly saved files no longer contain it.
- Config attributes are typed strictly now: `{"enable_model_cache": 1}` is rejected instead of coerced. Every
  `TTNN_CONFIG_OVERRIDES` value in the repo (workflows, scripts, docs) was audited and parses.

### Testing

`tests/ttnn/unit_tests/base_functionality/test_config.py`, 22 tests: parser semantics (types, `null` →
`nullopt`, unknown-key strict/non-strict, non-object, malformed JSON, source naming), strict-applies-nothing
vs non-strict-keeps-what-parsed, the `ValueError` contract, saved keys == `Config.keys()`, and a
`TTNN_CONFIG_PATH` round trip. Seven subprocess tests load `_ttnn.so` directly via `spec_from_file_location`,
bypassing `ttnn/__init__.py`, to pin the pure-C++ path: env override applies, config file applies, a missing
file is seeded with defaults (and the env override is *not* baked into it), a stale key warns and the rest
applies, a corrupt file does not fail the load, and a bad env var exits `1` with the message.

Also verified against a real pure-C++ binary rather than only through `dlopen`:
`unit_tests_ttnn --gtest_list_tests` reads a `TTNN_CONFIG_PATH` file, seeds a missing nested path with all 17
keys at their defaults, still lets the env var win over the file, and exits `1` on a bad env var with no core
dump.

Full `test_graph_report.py` (exercises the delegation and `save_config_to_json_file`): 158 passed, 1 skipped.
### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/ttnn-config-overrides-cpp)

### DIFF
--- a/tests/ttnn/unit_tests/base_functionality/test_config.py
+++ b/tests/ttnn/unit_tests/base_functionality/test_config.py
@@ -16,11 +16,7 @@ import ttnn
 @pytest.fixture(autouse=True)
 def restore_config():
     config = ttnn._ttnn.CONFIG
-    original = {
-        key: getattr(config, key)
-        for key in dir(config)
-        if not key.startswith("_") and key != "report_path" and not callable(getattr(config, key))
-    }
+    original = {key: getattr(config, key) for key in ttnn.Config.keys()}
     yield
     for key, value in original.items():
         setattr(config, key, value)
@@ -50,9 +46,14 @@ def test_apply_json_overrides_null_clears_optional():
 
 
 def test_apply_json_overrides_unknown_key(expect_error):
-    with expect_error(RuntimeError, "Unknown configuration key"):
+    with expect_error(RuntimeError, "Unknown configuration key: no_such_key"):
         ttnn.CONFIG.apply_json_overrides('{"no_such_key": 1}')
     ttnn.CONFIG.apply_json_overrides('{"no_such_key": 1}', strict=False)
+
+
+def test_apply_json_overrides_names_the_source(expect_error):
+    with expect_error(RuntimeError, "from /some/config.json"):
+        ttnn.CONFIG.apply_json_overrides('{"no_such_key": 1}', source="/some/config.json")
 
 
 def test_apply_json_overrides_rejects_non_object(expect_error):
@@ -60,12 +61,47 @@ def test_apply_json_overrides_rejects_non_object(expect_error):
         ttnn.CONFIG.apply_json_overrides('["validate_program_args"]')
 
 
+def test_apply_json_overrides_rejects_malformed_json(expect_error):
+    with expect_error(RuntimeError, "not valid JSON"):
+        ttnn.CONFIG.apply_json_overrides('{"validate_program_args": true,}')
+
+
+@pytest.mark.parametrize(
+    "bad",
+    [
+        '{"enable_model_cache": true, "no_such_key": 1}',
+        '{"enable_model_cache": true, "comparison_mode_pcc": "0.5"}',
+    ],
+)
+def test_strict_failure_applies_nothing(bad, expect_error):
+    ttnn.CONFIG.enable_model_cache = False
+    with expect_error(RuntimeError, "configuration key"):
+        ttnn.CONFIG.apply_json_overrides(bad)
+    assert ttnn.CONFIG.enable_model_cache is False
+
+
+def test_non_strict_keeps_the_keys_that_parsed():
+    ttnn.CONFIG.enable_model_cache = False
+    pcc = ttnn.CONFIG.comparison_mode_pcc
+    ttnn.CONFIG.apply_json_overrides(
+        '{"enable_model_cache": true, "comparison_mode_pcc": "0.5", "no_such_key": 1}', strict=False
+    )
+    assert ttnn.CONFIG.enable_model_cache is True
+    assert ttnn.CONFIG.comparison_mode_pcc == pytest.approx(pcc)
+
+
 def test_load_config_from_dictionary(expect_error):
     ttnn.load_config_from_dictionary({"enable_model_cache": True})
     assert ttnn.CONFIG.enable_model_cache is True
-    with expect_error(RuntimeError, "Unknown configuration key"):
+    with expect_error(ValueError, "Unknown configuration key"):
         ttnn.load_config_from_dictionary({"no_such_key": 1})
     ttnn.load_config_from_dictionary({"no_such_key": 1}, from_file=True)
+
+
+def test_saved_config_holds_exactly_the_overridable_keys(tmp_path):
+    path = tmp_path / "config.json"
+    ttnn.save_config_to_json_file(path)
+    assert set(json.loads(path.read_text())) == set(ttnn.Config.keys())
 
 
 def test_config_file_round_trip(tmp_path):
@@ -78,17 +114,25 @@ def test_config_file_round_trip(tmp_path):
     assert ttnn.CONFIG.report_name == saved_report_name
 
 
+def run_python(script, extra_env):
+    return subprocess.run(
+        [sys.executable, "-c", script],
+        env={**os.environ, **extra_env},
+        capture_output=True,
+        text=True,
+        timeout=600,
+    )
+
+
 def load_raw_ttnn_module(extra_env):
     """Load _ttnn directly, bypassing ttnn/__init__.py, as a pure C++ consumer would."""
-    script = (
+    return run_python(
         "import importlib.util, sys\n"
         f"spec = importlib.util.spec_from_file_location('_ttnn', {ttnn._ttnn.__file__!r})\n"
         "m = importlib.util.module_from_spec(spec)\n"
         "spec.loader.exec_module(m)\n"
-        "print(m.CONFIG.validate_program_args)\n"
-    )
-    return subprocess.run(
-        [sys.executable, "-c", script], env={**os.environ, **extra_env}, capture_output=True, text=True, timeout=120
+        "print(m.CONFIG.validate_program_args)\n",
+        extra_env,
     )
 
 
@@ -98,6 +142,72 @@ def test_env_overrides_apply_without_python_init():
     assert result.stdout.strip() == "True"
 
 
-def test_env_overrides_unknown_key_fails_at_load():
-    result = load_raw_ttnn_module({"TTNN_CONFIG_OVERRIDES": '{"no_such_key": true}'})
-    assert result.returncode != 0
+@pytest.mark.parametrize(
+    "overrides, message",
+    [
+        ('{"no_such_key": true}', "Unknown configuration key: no_such_key"),
+        ('{"validate_program_args": true,}', "not valid JSON"),
+        ('{"validate_program_args": 1}', "Bad value for configuration key validate_program_args"),
+    ],
+)
+def test_bad_env_overrides_exit_cleanly_at_load(overrides, message):
+    result = load_raw_ttnn_module({"TTNN_CONFIG_OVERRIDES": overrides})
+    # Exit code, not a signal: the load-time throw must not reach std::terminate/SIGABRT.
+    assert result.returncode == 1, result.stdout + result.stderr
+    assert message in result.stdout
+    assert "TTNN_CONFIG_OVERRIDES" in result.stdout
+
+
+def test_config_file_applies_without_python_init(tmp_path):
+    path = tmp_path / "config.json"
+    path.write_text(json.dumps({"validate_program_args": True}))
+    result = load_raw_ttnn_module({"TTNN_CONFIG_PATH": str(path)})
+    assert result.returncode == 0, result.stdout + result.stderr
+    assert result.stdout.strip().splitlines()[-1] == "True"
+
+
+def test_missing_config_file_is_seeded_with_defaults_without_python_init(tmp_path):
+    path = tmp_path / "nested" / "config.json"
+    result = load_raw_ttnn_module(
+        {"TTNN_CONFIG_PATH": str(path), "TTNN_CONFIG_OVERRIDES": '{"validate_program_args": true}'}
+    )
+    assert result.returncode == 0, result.stdout + result.stderr
+    saved = json.loads(path.read_text())
+    assert set(saved) == set(ttnn.Config.keys())
+    # The env override applies to the process but must not be baked into the file as a default.
+    assert saved["validate_program_args"] is False
+    assert result.stdout.strip().splitlines()[-1] == "True"
+
+
+def test_stale_config_file_key_warns_and_keeps_the_rest(tmp_path):
+    path = tmp_path / "config.json"
+    path.write_text(json.dumps({"report_path": "generated", "validate_program_args": True}))
+    result = load_raw_ttnn_module({"TTNN_CONFIG_PATH": str(path)})
+    assert result.returncode == 0, result.stdout + result.stderr
+    assert "Unknown configuration key: report_path" in result.stdout
+    assert result.stdout.strip().splitlines()[-1] == "True"
+
+
+def test_corrupt_config_file_does_not_fail_the_load(tmp_path):
+    path = tmp_path / "config.json"
+    path.write_text("{not json")
+    result = load_raw_ttnn_module({"TTNN_CONFIG_PATH": str(path)})
+    assert result.returncode == 0, result.stdout + result.stderr
+    assert "Failed to load ttnn configuration" in result.stdout
+
+
+def test_env_overrides_take_precedence_over_the_config_file(tmp_path):
+    path = tmp_path / "config.json"
+    path.write_text(json.dumps({"enable_model_cache": True, "validate_program_args": True}))
+    result = run_python(
+        "import ttnn; print(ttnn.CONFIG.enable_model_cache, ttnn.CONFIG.validate_program_args)",
+        {"TTNN_CONFIG_PATH": str(path), "TTNN_CONFIG_OVERRIDES": '{"enable_model_cache": false}'},
+    )
+    assert result.returncode == 0, result.stderr
+    assert result.stdout.strip().splitlines()[-1] == "False True"
+
+
+def test_env_overrides_are_validated_once_per_process():
+    result = run_python("import ttnn", {"TTNN_CONFIG_OVERRIDES": '{"enable_logging": true}'})
+    assert result.returncode == 0, result.stderr
+    assert result.stdout.count("Logging cannot be enabled in fast runtime mode") == 1
--- a/tests/ttnn/unit_tests/base_functionality/test_config.py
+++ b/tests/ttnn/unit_tests/base_functionality/test_config.py
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import os
+import pathlib
+import subprocess
+import sys
+
+import pytest
+
+import ttnn
+
+
+@pytest.fixture(autouse=True)
+def restore_config():
+    config = ttnn._ttnn.CONFIG
+    original = {
+        key: getattr(config, key)
+        for key in dir(config)
+        if not key.startswith("_") and key != "report_path" and not callable(getattr(config, key))
+    }
+    yield
+    for key, value in original.items():
+        setattr(config, key, value)
+
+
+def test_apply_json_overrides():
+    ttnn.CONFIG.apply_json_overrides(
+        json.dumps(
+            {
+                "validate_program_args": True,
+                "comparison_mode_pcc": 0.5,
+                "tmp_dir": "/tmp/ttnn-test",
+                "report_name": "config test",
+            }
+        )
+    )
+    assert ttnn.CONFIG.validate_program_args is True
+    assert ttnn.CONFIG.comparison_mode_pcc == pytest.approx(0.5)
+    assert ttnn.CONFIG.tmp_dir == pathlib.Path("/tmp/ttnn-test")
+    assert ttnn.CONFIG.report_name == pathlib.Path("config test")
+
+
+def test_apply_json_overrides_null_clears_optional():
+    ttnn.CONFIG.apply_json_overrides('{"report_name": "something"}')
+    ttnn.CONFIG.apply_json_overrides('{"report_name": null}')
+    assert ttnn.CONFIG.report_name is None
+
+
+def test_apply_json_overrides_unknown_key(expect_error):
+    with expect_error(RuntimeError, "Unknown configuration key"):
+        ttnn.CONFIG.apply_json_overrides('{"no_such_key": 1}')
+    ttnn.CONFIG.apply_json_overrides('{"no_such_key": 1}', strict=False)
+
+
+def test_apply_json_overrides_rejects_non_object(expect_error):
+    with expect_error(RuntimeError, "must be a JSON object"):
+        ttnn.CONFIG.apply_json_overrides('["validate_program_args"]')
+
+
+def test_load_config_from_dictionary(expect_error):
+    ttnn.load_config_from_dictionary({"enable_model_cache": True})
+    assert ttnn.CONFIG.enable_model_cache is True
+    with expect_error(RuntimeError, "Unknown configuration key"):
+        ttnn.load_config_from_dictionary({"no_such_key": 1})
+    ttnn.load_config_from_dictionary({"no_such_key": 1}, from_file=True)
+
+
+def test_config_file_round_trip(tmp_path):
+    path = tmp_path / "config.json"
+    ttnn.save_config_to_json_file(path)
+    saved_report_name = ttnn.CONFIG.report_name
+    ttnn.CONFIG.enable_model_cache = True
+    ttnn.load_config_from_json_file(path)
+    assert ttnn.CONFIG.enable_model_cache is False
+    assert ttnn.CONFIG.report_name == saved_report_name
+
+
+def load_raw_ttnn_module(extra_env):
+    """Load _ttnn directly, bypassing ttnn/__init__.py, as a pure C++ consumer would."""
+    script = (
+        "import importlib.util, sys\n"
+        f"spec = importlib.util.spec_from_file_location('_ttnn', {ttnn._ttnn.__file__!r})\n"
+        "m = importlib.util.module_from_spec(spec)\n"
+        "spec.loader.exec_module(m)\n"
+        "print(m.CONFIG.validate_program_args)\n"
+    )
+    return subprocess.run(
+        [sys.executable, "-c", script], env={**os.environ, **extra_env}, capture_output=True, text=True, timeout=120
+    )
+
+
+def test_env_overrides_apply_without_python_init():
+    result = load_raw_ttnn_module({"TTNN_CONFIG_OVERRIDES": '{"validate_program_args": true}'})
+    assert result.returncode == 0, result.stderr
+    assert result.stdout.strip() == "True"
+
+
+def test_env_overrides_unknown_key_fails_at_load():
+    result = load_raw_ttnn_module({"TTNN_CONFIG_OVERRIDES": '{"no_such_key": true}'})
+    assert result.returncode != 0
--- a/ttnn/core/config.cpp
+++ b/ttnn/core/config.cpp
@@ -24,33 +24,50 @@ namespace ttnn::core {
 Config CONFIG{};
 
 void Config::apply_json_overrides(const std::string& json_text, bool strict) {
-    auto json = nlohmann::json::parse(json_text);
+    nlohmann::json json;
+    try {
+        json = nlohmann::json::parse(json_text);
+    } catch (const nlohmann::json::exception& e) {
+        // Reached from a static initializer, so name the source instead of surfacing a bare parser error.
+        TT_THROW("TTNN config overrides are not valid JSON: {} (input: {})", e.what(), json_text);
+    }
     TT_FATAL(json.is_object(), "TTNN config overrides must be a JSON object, got: {}", json_text);
-    reflect::for_each(
-        [&](auto I) {
-            auto name = reflect::member_name<I>(this->attributes);
-            auto it = json.find(std::string(name));
-            if (it == json.end()) {
-                return;
+
+    // Restore on any failure: a rejected override must not leave process-wide settings half applied.
+    attributes_t previous = this->attributes;
+    try {
+        reflect::for_each(
+            [&](auto I) {
+                auto name = reflect::member_name<I>(this->attributes);
+                auto it = json.find(std::string(name));
+                if (it == json.end()) {
+                    return;
+                }
+                auto& member = reflect::get<I>(this->attributes);
+                using T = std::decay_t<decltype(member)>;
+                if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, float>) {
+                    member = it->template get<T>();
+                } else if constexpr (std::is_same_v<T, std::optional<std::filesystem::path>>) {
+                    member = it->is_null() ? T{} : T{it->template get<std::string>()};
+                } else {
+                    member = std::filesystem::path(it->template get<std::string>());
+                }
+                this->validate(name);
+                json.erase(it);
+            },
+            this->attributes);
+        for (const auto& [key, value] : json.items()) {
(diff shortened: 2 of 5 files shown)


ttnn/core/config.cpp
```diff
@@ -6,20 +6,63 @@
...
+#include <tt_stl/assert.hpp>
 #include <tt_stl/reflection.hpp>
 #include <tt-metalium/experimental/inspector_config.hpp>
 
 namespace ttnn::core {
 
 Config CONFIG{};
 
+void Config::apply_json_overrides(const std::string& json_text, bool strict) {
+    auto json = nlohmann::json::parse(json_text);
```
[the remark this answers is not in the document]
dgomezTT: Fixed in a606be6662c. The parse is wrapped and rethrown with the source named, which matters because
this runs from a static initializer:

```
TTNN config overrides are not valid JSON: <parser detail> (input: <the env var>)
```

Bad *values* (right key, wrong type) get the same treatment rather than escaping as a bare
`nlohmann` type error. Verified against a local build: `apply_json_overrides("{not json", true)` now
reports "not valid JSON" instead of the raw parser exception.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,63 @@
...
+        [&](auto I) {
+            auto name = reflect::member_name<I>(this->attributes);
+            auto it = json.find(std::string(name));
+            if (it == json.end()) {
+                return;
+            }
+            auto& member = reflect::get<I>(this->attributes);
+            using T = std::decay_t<decltype(member)>;
+            if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, float>) {
+                member = it->template get<T>();
```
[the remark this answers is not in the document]
dgomezTT: Fixed in a606be6662c. The attributes are snapshotted before anything is written and restored if any
field conversion, `validate()` or the unknown-key check throws, so a rejected override leaves
process-wide settings untouched.

I used snapshot-and-restore rather than building a detached `attributes_t` because `validate()` is a
const member that inspects `this->attributes`, so a candidate object would not be validated by it.

Verified on a local build: `{"enable_model_cache": true, "nope": 1}` with strict=true throws and
`enable_model_cache` stays `False` — before the fix it stayed applied.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,80 @@
...
+        this->attributes = std::move(previous);
+        throw;
+    }
+}
+
+// Parse TTNN_CONFIG_OVERRIDES at load so pure C++ consumers (e.g. the sanity-pipeline gtests) honor
+// it too; ttnn/__init__.py applies overrides through the same function.
+static const int apply_env_config_overrides = [] {
+    if (const char* overrides = std::getenv("TTNN_CONFIG_OVERRIDES")) {
+        CONFIG.apply_json_overrides(overrides, /*strict=*/true);
```
ayerofieiev-tt: **[high] TT_THROW escaping a static initializer terminates the process at dlopen.**

This initializer runs during dlopen of `_ttnn.so` (and at process start of any binary linking libttnn), and `apply_json_overrides` throws on invalid JSON, non-object input, unknown key, and wrong-typed value. There is no C++ handler in the loader/CPython import path, so a malformed or stale `TTNN_CONFIG_OVERRIDES` (typo'd key, `0`/`1` for a bool, trailing comma) now ends in `std::terminate`/SIGABRT before `main`, where pre-PR it was a catchable `ValueError`/`JSONDecodeError` at `import ttnn` with a clean traceback.

`test_env_overrides_unknown_key_fails_at_load` only asserts `returncode != 0`, which SIGABRT (-6) also satisfies, so the test cannot distinguish a clean error from an abort.

Consider the repo's existing pattern for fail-fast env parsing: lazy first-use (`RunTimeOptions` is constructed inside `MetalContext::instance()`, so its TT_THROWs surface on a catchable call stack), or catch here and fail with a clear message (`log_fatal` + `std::exit`) instead of an unwinding throw.

ttnn/ttnn/__init__.py
```diff
@@ -72,8 +63,9 @@ def save_config_to_json_file(json_path):
         save_config_to_json_file(CONFIG_PATH)
```
ayerofieiev-tt: **[high] Env overrides get baked into the freshly-created `TTNN_CONFIG_PATH` file.**

The C++ static initializer applies `TTNN_CONFIG_OVERRIDES` at `import ttnn._ttnn` (line 17), so when `TTNN_CONFIG_PATH` points at a missing file, this save now writes the already-overridden CONFIG to disk as the "default" file. First run with `TTNN_CONFIG_PATH=~/ttnn.json` and `TTNN_CONFIG_OVERRIDES='{"enable_fast_runtime_mode": false}'` permanently bakes slow mode into the file; every later run without the env var silently inherits it. Pre-PR the save ran before overrides were applied, so the file captured pristine defaults.

Fix: serialize a default-constructed `Config()` here rather than the live CONFIG.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,80 @@
...
+            [&](auto I) {
+                auto name = reflect::member_name<I>(this->attributes);
+                auto it = json.find(std::string(name));
+                if (it == json.end()) {
+                    return;
+                }
+                auto& member = reflect::get<I>(this->attributes);
+                using T = std::decay_t<decltype(member)>;
+                if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, float>) {
+                    member = it->template get<T>();
```
ayerofieiev-tt: **[medium] Strict `get<T>()` drops the old type coercion, and one bad value now discards the whole config file.**

Old Python did `value = type(getattr(CONFIG, key))(value)`, so `{"enable_model_cache": 1}` and `{"comparison_mode_pcc": "0.5"}` worked. nlohmann `get<bool>()` throws `type_error.302` on JSON `0`/`1` and `get<float>()` throws on strings. Per path: env var -> the static-init abort; `load_config_from_dictionary` -> RuntimeError; and on the lenient `load_config_from_json_file` path the snapshot/restore means a single bad value now rolls back **every valid key in the file** with only a warning — old code applied keys up to the failure.

Stricter typing is defensible (the old `bool("false") == True` was its own bug), but silently discarding a whole config file over one stale value is a regression; consider per-key application when `strict=false`.

ttnn/ttnn/__init__.py
```diff
@@ -57,6 +46,8 @@ def save_config_to_json_file(json_path):
             if re.match("^_.+_$", key):
                 continue
             value = getattr(CONFIG, key)
+            if callable(value):
```
ayerofieiev-tt: **[medium] `report_path` is still serialized, so every load of a self-saved config warns — and stale files now apply keys they never applied before.**

`report_path` is a read-only nanobind property (a value, not callable, not matching `^_.+_$`), and it is not an `attributes_t` member, so every `load_config_from_json_file` of a self-saved file logs `Unknown configuration key: report_path` at import. The new test fixture special-cases `report_path` by name, so this was already encountered.

Also a behavior flip worth knowing about: pre-PR, loading a saved file hit `AttributeError` on the read-only `report_path` setter mid-loop (swallowed by the broad except), so the alphabetically-later keys — `root_report_path`, `throw_exception_on_fallback`, `tmp_dir`, `validate_program_args` — were silently never applied. Now they are. That is arguably a fix, but users with stale saved files will see settings take effect that never did before.

Skipping `report_path` in this loop is a one-liner.

ttnn/ttnn/__init__.py
```diff
@@ -26,18 +26,7 @@
...
-            if getattr(CONFIG, key) is not None:
-                value = type(getattr(CONFIG, key))(value)
-            setattr(CONFIG, key, value)
-        elif from_file:
-            logger.warning(
-                f"Unknown configuration key: {key}. Please update your configuration file: {CONFIG_PATH}. Or delete it to get the new default config"
-            )
-        else:
-            raise ValueError(f"Unknown configuration key: {key}")
+    CONFIG.apply_json_overrides(json.dumps(config, default=str), strict=not from_file)
```
ayerofieiev-tt: **[low] Public-API contract change: `ValueError` -> `RuntimeError`, and the remediation hint is gone.**

Unknown keys now surface as `RuntimeError` (TT_THROW -> `std::runtime_error`) where this public function raised `ValueError` before — external `except ValueError` callers break silently (no in-repo caller does, verified). The `from_file` warning also lost its actionable text ("Please update your configuration file: {CONFIG_PATH}. Or delete it to get the new default config"); the bare C++ `Unknown configuration key: X` no longer says which file to fix. Worth a release note at minimum; consider re-adding the file hint.

ttnn/ttnn/__init__.py
```diff
@@ -72,8 +63,9 @@ def save_config_to_json_file(json_path):
         save_config_to_json_file(CONFIG_PATH)
 
 if CONFIG_OVERRIDES is not None:
+    # Already applied at _ttnn load; re-applied here so the env keeps precedence over TTNN_CONFIG_PATH.
     logger.debug(f"Loading ttnn configuration overrides from environment variable TTNN_CONFIG_OVERRIDES")
-    load_config_from_dictionary(json.loads(CONFIG_OVERRIDES))
+    CONFIG.apply_json_overrides(CONFIG_OVERRIDES)
```
ayerofieiev-tt: **[low] Overrides apply twice per Python process, duplicating `validate()` warnings.**

Static init applies them at line 17, then this line re-applies unconditionally, so per-key `validate()` warnings (e.g. the fast-runtime + logging conflict) print twice on every import. The re-apply only changes state when `CONFIG_PATH.exists()` actually loaded a file — guarding it on that branch removes the duplicate side effects without touching precedence.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,80 @@
...
+        reflect::for_each(
+            [&](auto I) {
+                auto name = reflect::member_name<I>(this->attributes);
+                auto it = json.find(std::string(name));
+                if (it == json.end()) {
+                    return;
+                }
+                auto& member = reflect::get<I>(this->attributes);
+                using T = std::decay_t<decltype(member)>;
+                if constexpr (std::is_same_v<T, bool> || std::is_same_v<T, float>) {
```
ayerofieiev-tt: **[low] This dispatch chain re-implements `ttsl::json::from_json<T>`.**

`tt_stl/reflection.hpp` (already included here) ships `from_json_t` specializations for bool, arithmetic, enums, strings, and `std::optional<T>` with exactly the null->nullopt semantics hand-rolled above. Only `std::filesystem::path` lacks a specialization; add that one-liner and this whole chain becomes `member = ttsl::json::from_json<T>(*it);`. As written, the else-branch is a trap: a future `std::string` attribute silently routes through `std::filesystem::path` construction, and an `int`/enum attribute fails to compile deep inside the lambda.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,80 @@
...
+    try {
+        json = nlohmann::json::parse(json_text);
+    } catch (const nlohmann::json::exception& e) {
+        // Reached from a static initializer, so name the source instead of surfacing a bare parser error.
+        TT_THROW("TTNN config overrides are not valid JSON: {} (input: {})", e.what(), json_text);
+    }
+    TT_FATAL(json.is_object(), "TTNN config overrides must be a JSON object, got: {}", json_text);
+
+    // Restore on any failure: a rejected override must not leave process-wide settings half applied.
+    attributes_t previous = this->attributes;
```
ayerofieiev-tt: **[low] Copy-commit is still available despite the `validate()` coupling noted in a606be6.**

`validate()` never throws — it only log_warns — so the detached-candidate objection has a cheap answer: mutate a local `attributes_t next`, commit `this->attributes = std::move(next)` after the unknown-key check, then run the validate warnings once against the committed state. That deletes both restore blocks and the `catch (...)`, decouples `json.erase` from the leftover scan, and makes strict rejection touch no process-wide state at all.

tests/ttnn/unit_tests/base_functionality/test_config.py
```diff
@@ -0,0 +1,103 @@
...
+    original = {
+        key: getattr(config, key)
+        for key in dir(config)
+        if not key.startswith("_") and key != "report_path" and not callable(getattr(config, key))
+    }
+    yield
+    for key, value in original.items():
+        setattr(config, key, value)
+
+
```
ayerofieiev-tt, line 24: **[low] Third hand-rolled "enumerate settable config attributes" filter.**

`save_config_to_json_file` has another (this PR added its `callable()` check precisely because binding `apply_json_overrides` broke the old filter — direct evidence of the fragility). A `keys()`-style helper bound next to the existing C++ `get_config_entries()` would let both this fixture and the save path enumerate `attributes_t` members directly, so the next bound method or computed property cannot silently break either filter.

ayerofieiev-tt (commented):
Automated deep review (recall-biased): 8 finder angles, 10 candidates verified individually — 9 confirmed, 1 refuted (tt-logger use during static init is safe: lazy function-local singleton). Inline comments below, ranked high to low.

ttnn/core/config.cpp
dgomezTT: Fixed in 3371b4bd94e. The initializer now catches, logs the reason and `_Exit(1)`s, so the failure is an exit code with a message rather than `std::terminate`/SIGABRT.

Worth recording: `std::exit` here does not work. It runs atexit handlers and static destructors, which re-enter the loader while `dlopen` holds its lock — the process printed the critical line and then **hung** (killed at 30s; the three pytest cases sat at their 300s timeout). `_Exit` skips all of that, with `fflush(nullptr)` first so the logged message is out.

Lazy first-use I did not take: `CONFIG` is an `extern Config` global, not a function-local singleton, so making it lazy means changing every `ttnn::CONFIG` user, and a gtest that never calls an init hook would still miss the override.

The tests now pin the failure mode instead of just `returncode != 0` — parametrized over unknown key, malformed JSON and wrong-typed value, asserting `returncode == 1` (a signal gives a negative code) plus the message text. Verified on a real pure-C++ binary too: `TTNN_CONFIG_OVERRIDES='{"no_such_key": true}' unit_tests_ttnn --gtest_list_tests` → EXIT=1, message logged, no core dump.

ttnn/ttnn/__init__.py
dgomezTT: Fixed in 3371b4bd94e, and the orchestration moved: C++ now owns `TTNN_CONFIG_PATH` as well (your Slack point), so the seed file is written by the static initializer from a **default-constructed `Config{}`** — never the live config. `nb::init<>()` is bound for the Python side.

Test `test_missing_config_file_is_seeded_with_defaults_without_python_init` runs a pure `_ttnn.so` load with `TTNN_CONFIG_PATH` at a missing nested path plus `TTNN_CONFIG_OVERRIDES={"validate_program_args": true}`, then asserts the file says `false` while the live config says `True`.

ttnn/core/config.cpp
dgomezTT: The whole-file discard is fixed in 3371b4bd94e: application is per key now. `attributes_t next` takes each key, and on a bad value strict throws (nothing committed at all) while non-strict warns and skips **that key only**, so a stale value in a config file no longer costs the keys that did parse. `test_non_strict_keeps_the_keys_that_parsed` and `test_strict_failure_applies_nothing` pin both halves.

Exact typing kept, as you say it is defensible — and I re-audited every `TTNN_CONFIG_OVERRIDES` value in the repo (workflows, scripts, docs): all are correctly typed JSON. The messages name the key and the source now, e.g. `Bad value for configuration key validate_program_args: [json.exception.type_error.302] type must be boolean, but is number (from TTNN_CONFIG_OVERRIDES)`.

ttnn/ttnn/__init__.py
dgomezTT: Fixed in 3371b4bd94e, by deleting the loop rather than special-casing: serialization moved to C++ (`Config::save_to_file`, a `reflect::for_each` writer over `attributes_t`), so `report_path` and bound methods are structurally unable to reach a saved file. `save_config_to_json_file` is now a one-line delegation.

`test_saved_config_holds_exactly_the_overridable_keys` asserts the saved key set equals `Config.keys()`.

The stale-file flip you describe is now covered explicitly: `test_stale_config_file_key_warns_and_keeps_the_rest` writes a file with `report_path` in it and checks the load warns `Unknown configuration key: report_path (from <path>; update or delete it to get the new default config)` and still applies the rest.

ttnn/ttnn/__init__.py
dgomezTT: Both fixed in 3371b4bd94e. `load_config_from_dictionary` catches the `RuntimeError` and re-raises `ValueError`, so the public contract is unchanged (`test_load_config_from_dictionary` asserts `ValueError`).

The hint is back via a new `source` argument on `apply_json_overrides`, which names the origin in every message. `Config::load_from_file` passes `<path>; update or delete it to get the new default config`, so the warning again says which file to fix.

ttnn/ttnn/__init__.py
dgomezTT: Fixed in 3371b4bd94e — there is now exactly one application site. The file and the env var are both applied by the static initializer (file first, then env, same precedence), and `ttnn/__init__.py` no longer re-applies anything, so `validate()` runs once per applied key per process.

`test_env_overrides_are_validated_once_per_process` asserts the fast-runtime/logging warning appears exactly once.

ttnn/core/config.cpp
dgomezTT: Not adopted, and the reason is worth recording: `from_json_t` is `noexcept`. tt_stl/reflection.hpp:1623 defines it for integral/floating/enum as `T operator()(const nlohmann::json& j) noexcept { return j.get<T>(); }`, and `get<bool>()` on JSON `1` throws `type_error.302` — inside a `noexcept` function that is `std::terminate`. I checked with a standalone probe: the direct call throws catchably, the same call through a noexcept wrapper aborts (exit 134). So routing user input through `from_json` reintroduces exactly the uncatchable abort the comment above asks me to remove.

What I did instead (3371b4bd94e) is fix the trap you identified by inverting the chain: the two `std::filesystem::path` shapes are the explicit branches, and the fallback is `value.get<T>()` guarded by

```cpp
static_assert(std::is_arithmetic_v<T> || std::is_same_v<T, std::string>,
              "Config attribute type needs a JSON conversion in apply_json_overrides");
```

so a future `std::string` attribute can no longer silently become a path, and an enum/vector attribute fails at that line with a readable message instead of deep inside the lambda.

ttnn/core/config.cpp
dgomezTT: Done in 3371b4bd94e. `attributes_t next = this->attributes`, keys applied into `next`, commit `this->attributes = std::move(next)` after the unknown-key check, then `validate()` once per applied key against the committed state. Both restore blocks, the `catch (...)` and the `json.erase` are gone, and a strict rejection touches no process-wide state.

tests/ttnn/unit_tests/base_functionality/test_config.py
dgomezTT: Done in 3371b4bd94e: `Config::keys()` (static, next to `get_config_entries`, bound via `def_static`) is the single enumeration of `attributes_t` members. The fixture uses it, and the save path no longer needs a filter at all — serialization moved into C++ (`Config::save_to_file`), so the `dir(CONFIG)` + regex + `callable()` walk is deleted along with the now-unused `import re`.

dgomezTT:
CI triage for head `3371b4bd9` (I dispatched the full set of pipelines since none had run against this commit).

| Pipeline | Result |
|---|---|
| Sanity tests (WH + real BH + Sim) | green |
| Nightly tt-metal L2 tests | green |
| (Runtime) Sanity Tests | green |
| All Model Tests — Tiers[1,2] unit+e2e, all SKUs | 23 failures |

Of the 23, **19 also fail on `main`** (baseline = union of failing jobs across 14 recent `main` runs of the same workflow, paginated; `All Model Tests` is red on main every run, so a raw count means nothing on its own).

The 4 not in that baseline, with their failure signatures:

```
Timeout                                  t1-e2e-tests  / Qwen3.6-27B e2e            [bh_quietbox_2]
Value cannot be null (ContainerId)       t2-unit-tests / SDXL LoRA unit             [wh_n150]
"much higher than expected"              t2-e2e-tests  / Qwen2.5-VL-72B e2e         [wh_llmbox_perf]
PCC                                      t2-unit-tests / SDXL base unet-loop unit   [wh_n300]
```

- The first two are infrastructure: a job timeout, and a runner whose container never attached (no test ran).
- Qwen2.5-VL-72B fails for being **faster** than its expected target, not slower — the same stale perf baseline that flagged on #53084. Its own log says "much higher than expected. Please update the expected perf."
- **SDXL base unet-loop `[wh_n300]` PCC is the only one I would not dismiss.** It is not in the main baseline and the signature is numerical rather than infra. Given this commit moves config ownership into C++, worth confirming it reproduces before assuming noise — a rerun of that single job is the cheapest discriminator.

Sanity/L2/Runtime Sanity being green means the config-ownership change is clean on the paths those cover.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,159 @@
...
+    for (std::string_view name : applied) {
+        this->validate(name);
+    }
+}
+
+std::vector<std::string> Config::keys() {
+    std::vector<std::string> names;
+    reflect::for_each<attributes_t>([&](auto I) { names.emplace_back(reflect::member_name<I, attributes_t>()); });
+    return names;
+}
```
vtsilytskyiTT, line 92: This function is used in tests only. It shouldn't belong to external API.
In perfect case we shouldn't add functionality for debug-only purpose.
Can you replace "keys" test with following:
1. Dump config into a temporary json file;
2. Apply changes;
3. Dump config second time;
4. Compare files.

If you had `to_json` & `from_json` functions/constructors, you could do the same in memory, without dumping data into files.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,159 @@
...
 #include <tt-logger/tt-logger.hpp>
+#include <tt_stl/assert.hpp>
 #include <tt_stl/reflection.hpp>
 #include <tt-metalium/experimental/inspector_config.hpp>
 
 namespace ttnn::core {
 
 Config CONFIG{};
 
+void Config::apply_json_overrides(const std::string& json_text, bool strict, const std::string& source) {
```
vtsilytskyiTT, line 27: Nitpick: `json_text` can be an `std::string_view` (which is already extensively used in this function). Not sure about `source`.

ttnn/core/config.cpp
```diff
@@ -6,20 +6,159 @@
 
 #include <algorithm>
 #include <chrono>
+#include <cstdlib>
+#include <fstream>
```
vtsilytskyiTT, line 10: IMHO config shouldn't work with filesystem. Let it be a pure config.
I suggest to implement `from_json` and `to_json`, which consumes/produces either `std::stringstream` or`nlohmann::json` object.
Then file operations becomes just an extra line in the `ttnn/__init__.py`. Plus some extra effort inside tests.

tests/ttnn/unit_tests/base_functionality/test_config.py
```diff
@@ -0,0 +1,213 @@
...
+    ttnn.save_config_to_json_file(path)
+    saved_report_name = ttnn.CONFIG.report_name
+    ttnn.CONFIG.enable_model_cache = True
+    ttnn.load_config_from_json_file(path)
+    assert ttnn.CONFIG.enable_model_cache is False
+    assert ttnn.CONFIG.report_name == saved_report_name
+
+
+def run_python(script, extra_env):
+    return subprocess.run(
```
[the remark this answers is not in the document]
dgomezTT: False positive. `run_python` passes a **list** to `subprocess.run` with no `shell=True`, so the arguments go straight to `execve` with no shell parsing — there is no command line for metacharacters to break out of. `sys.executable` is hardcoded, and `script` is a Python string literal written in this test module, not external or user-supplied input. The only thing a caller controls is Python source that the test itself authors and intends to execute.

The construct is also load-bearing for what this PR proves: it runs `_ttnn.so` in a fresh interpreter, bypassing `ttnn/__init__.py`, which is the only way to test that a pure C++ consumer honors the config. That requires a subprocess.

ttnn/core/config.cpp
dgomezTT: Keeping `Config::keys()` bound, deliberately — it exists to satisfy an earlier review point on this PR.

The ask was for a single enumeration of `attributes_t` so the "which attributes are settable" filter stops being hand-rolled in three places; `keys()` (static, next to `get_config_entries`) is that single point. Dropping the binding would not just move the test — it would cost `test_saved_config_holds_exactly_the_overridable_keys`, which is the assertion that pins `report_path` and bound methods out of a saved config file. That guard is what closed the earlier finding that every load of a self-saved config warned about `report_path`.

The dump/apply/dump/compare shape you describe tests round-tripping, which is a different property, and `test_config_file_round_trip` already covers it. It would not catch a newly bound method leaking into the saved key set, which is the regression the `keys()` test exists to catch.

ttnn/core/config.cpp
dgomezTT: Not adopting this one, because moving file I/O to `ttnn/__init__.py` would defeat what the PR exists to do.

The point of the change is that a **pure C++ consumer** — anything that loads `_ttnn.so` or links `libttnn` without importing the Python package — honors the config file. C++ owning `TTNN_CONFIG_PATH` (load and seed) is exactly what makes that true, and it was an explicit review request earlier in this PR. `test_missing_config_file_is_seeded_with_defaults_without_python_init` pins it by loading `_ttnn.so` directly with `ttnn/__init__.py` bypassed; if the file operations lived in `__init__.py`, that test could not pass and the C++ path would silently fall back to defaults.

The "pure config core" half of your suggestion is already how it is built: `apply_json_overrides` takes text and knows nothing about paths, and `load_from_file`/`save_to_file` are thin edges over it. Happy to name those `from_json`/`to_json` if the naming is the sticking point, but the filesystem entry points need to stay on the C++ side.

ttnn/core/config.cpp
dgomezTT: Leaving this as `const std::string&`. The function is a nanobind-bound entry point called about twice per process (static init for the file, then the env var), so there is no throughput to win, and `string_view` across a Python binding boundary buys a lifetime question in exchange for nothing measurable.

dgomezTT:
All review threads are resolved. Full triage of the `All Model Tests` reds at `3371b4bd9` — 23 failures, all 23 accounted for, none attributable to this PR.

| Bucket | Jobs | Evidence |
| --- | --- | --- |
| Runner infra — Weka mount | 11 | Every one failed at the **`Set up runner`** step, so no test executed. `mnt-MLPerf.mount - Mount Weka filesystem MLPerf ... Active: failed` — the model-weights share did not mount. All `[wh_n150]`. |
| Pre-existing | 10 | Same jobs fail on unrelated branches at matching scope. |
| Hardware link fault | 1 | TT-DiT Flux.1-dev `[bh_quietbox_2]`: `Fabric Router Sync: Timeout after 10000 ms ... Ethernet handshake likely failed -- the link may not be healthy.` |
| Flake | 1 | SDXL `test_unet_loop[...512x512]` `[wh_n300]`, PCC 0.8077. **Re-ran the job: passes.** |

The SDXL one was the only red that needed real work: it passed on `main` in an era-matched run two days later, so "pre-existing" was not available as an explanation and it had to be either a flake or a regression from this PR. Two things settle it as a flake — the rerun is green on the same commit, and `config.hpp` is `+15/-0`, so this PR changes no default value; it moves *where* overrides are parsed, not what any setting is.

Worth noting for anyone reading the red badge: 11 of the 23 never ran a line of test code, so the raw failure count badly overstates what actually broke.

dgomezTT:
/codeowners ping All review threads resolved and the All Model Tests reds are fully triaged (11 runner-infra, 10 pre-existing, 1 link fault, 1 flake that passes on rerun). Ready for codeowner review.

ttnn/core/config.cpp
vtsilytskyiTT: No, naming was not the point. I wanted to see `config.cpp` as a pure data container without IO.

ttnn/core/config.cpp
vtsilytskyiTT: OK, so you deliberately expose API for future use (and test it, of course). Then no questions.

vtsilytskyiTT (approved):
I'm OK with the change.
Would be great to split data container `config.cpp` and IO operations on different abstractions layer, but acceptable as is.

PR #53073, opened 2026-08-13
MiniMax-H3 audio: fix conv_pre and tap_matmul under T-sharding (rebased onto main)

### LINKED ISSUE #52260: Enhance Audio Decode Kernels
Right now audio decode kernels for the MinMax-H3 takes about 1506ms. Goal is decrease this by orders of magnitude

rouzbehTT:
Three correctness fixes for the MiniMax-H3 audio decode under T-sharding, the test that gates them, and the harness that measures latency and PSNR. Rebased onto `main`.

## The bugs

**1. `Vocoder.conv_pre` returned uninitialized memory when T-sharded.** At `(C_in=2048, C_out=1024, k=7)` -- the widest conv in the decode -- the sharded path produced absmax `2.6e+11` against a reference absmax of `1.391`, i.e. **-204 dB**. This is what made *every* T-sharded decode return wrong audio.

A sharded-vs-unsharded sweep over conv shapes found this to be the only shape affected: `1024->512 k=7`, `256->256 k=7`, and `64->64` at kernels 1/3/7 with dilation 5 are all bit-exact under T-sharding, and both `_partition_t` and `_t_neighbor_pad` were verified exact at this shape. `conv_pre` now runs unsharded and `_forward_device` partitions T immediately afterwards -- close to free, because `conv_pre` consumes the *input* sequence (207 latents, 256 padded), ~1/800th of the final waveform length. Under channel-TP the original ordering is preserved, since there `conv_pre`'s own `gather_channel_to_full` reconstructs full `C_in`.

Note this is a **workaround, not a root cause**: a 2048-wide k=7 conv returning uninitialized memory under T-sharding presumably still affects anyone else who shards a conv that wide.

**2. The shifted-matmul (tap) path was disabled under T-sharding, and broken underneath.** `Conv1dViaConv3d` gated `tap_matmul` on `not sharded`, so sharded convs always took conv3d. Two bugs were hiding behind that gate:

* **Halo width.** The taps consume `eff_k - 1` rows, but `halo_pad_left/right` are both `eff_k // 2`, which sums to that only for *odd* `eff_k`. At the even ones -- the K=12 anti-aliased downsamplers -- the output came out one row too long and the resblock's residual add rejected it ("Invalid subtile broadcast type"). `pad_right` is now derived as `(eff_k - 1) - pad_left`, leaving `halo_pad_*` alone so the conv3d path is untouched.
* **Freeing a buffer we did not own.** The tail did `if x_padded is not x_BTC: deallocate(x_padded)`, which is right unsharded (`_zero_pad_t` allocated it) and wrong sharded, where `x_padded` **is** the `ccl_manager`'s ping-pong buffer, handed back as `neighbor_pad_async`'s output. Freeing it left a dead buffer in a cache keyed on `(shape, dims, pads, dtype)`, so a *later* halo call with the same shape failed `input_tensor.is_allocated()` inside an unrelated op -- identically traced and untraced, with nothing in the message pointing at the real culprit. Ownership is now tracked explicitly.

With both fixed the gate is removed, which is what makes **sharded output equal unsharded output at the same levers**: `test_audio_decode_t_parallel` now measures **83.4 dB** agreement, up from 57.4 dB. The old 57.4 dB was not noise -- with tap gated off under sharding, the two paths were computing different math, so "sharding buys latency only" only held when tap was off. It now holds with tap on too. The cost is that the tap path is slower than conv3d: T-parallel speedups fall from 2.51x/3.24x to 1.84x/2.14x, and `tap_matmul` costs ~80 ms more (see the table).

**3. `CCLManager.get_np_ping_pong_buffer` handed out deallocated buffers.** It cached two buffers per `(shape, dtype)` and never checked them again, so any consumer that freed one poisoned the cache for every later caller -- surfacing as `is_allocated()` in an unrelated op. It now validates and re-creates with a warning. That cache has other callers; this is a latent trap for them too. The warning stayed silent across every run here, which is how we know fix 2 was complete and no other caller does the same thing.

## Supporting changes

- **C-chunked `ttnn.conv1d` recovery.** At `T_pad=166, C=512, K=7` the DRAM slicer runs out of L1 (the activation block is `C*K` wide) and conv1d fails. Depthwise means channel slices are independent, so it retries in chunks of 128/64/32 and concatenates. Hit 72 times per sharded decode. The assert on `(chunk * 4) % 64 == 0` is load-bearing: a last-dim concat is lossy in fp32 unless the row is a multiple of the 64B buffer alignment, so a future chunk size that broke it would silently truncate the mantissa to TF32.
- **`test_audio_decode_t_parallel`, resurrected** with `KNOWN_BROKEN` empty. Three guards, each for a way it previously passed while broken: the factor-1 baseline must have run (all three factors raising on a dirty device once reported green); factor 1 must *be* the baseline (when it crashed, factor 4 was promoted to reference and scored PSNR `inf` against itself, hiding a -10.1 dB configuration); and a listed factor that raises now fails the test instead of being recorded as "unsupported" (factor 4 could regress while factor 8 kept the run green). Carries `@pytest.mark.timeout(2400)` -- it takes ~14 min against `pytest.ini`'s 300 s default.
- **`audio_perf/cpu_vs_device.py`** -- encodes four real clips with the torch/diffusers reference, decodes on CPU and device, scores device against CPU, prints latency and accuracy verdicts. All knobs are env vars. `decode_bench.py` is the median-latency harness. Both default their output under `$TT_METAL_HOME/generated/audio_perf`.

## Measured

`test_audio_minimax_h3.py`: **10 passed**, with one known flake (below).

T-parallel, untraced, versus the single-device path:

| factor | axis | time | speedup | PSNR vs 1-device |
|---|---|---|---|---|
| 1 | 1 | 6.42 s | 1.00x | baseline |
| 4 | 0 | 3.49 s | 1.84x | 83.4 dB |
| 8 | 1 | 3.01 s | 2.14x | 83.4 dB |

Decode latency and accuracy on a 4x8 Blackhole mesh, `t_factor=8 axis=1`, traced:

| levers | latency | PSNR vs CPU |
|---|---|---|
| all fast | 292.6 ms | 45.80 dB |
| `tap_matmul=True` | 391.5 ms | 46.74 dB |
| `split_mode="weight"` | 367.4 ms | 48.26 dB |
| **`split_mode="weight"` + `tap_matmul=True`** | **448.4 ms** | **50.77 dB** |
| `split_mode="full"` | 484.9 ms | 50.22 dB |
| `prefer_mac=True` | 840.9 ms | 49.45 dB |
| all three = the constructed defaults | 1103.8 ms | **81.89 dB** |

Same settings on a single device, all fast: 1247 ms / 45.80 dB -- so the 8-way T-shard plus trace is worth ~4x, and it buys **latency only**: PSNR matches the unsharded path at the same levers, which is the point of fixes 1 and 2. At the far end, a single device with the constructed defaults measures 5902 ms / **81.99 dB**.

That last pair is the clearest measure of what fix 2 bought. At the constructed defaults, sharded now scores **81.89 dB against the unsharded 81.99 dB -- a 0.1 dB gap, where it was 14.6 dB (67.37 vs 81.99) while the tap path was gated off under sharding.** A 4x speedup that quietly cost 14.6 dB was not "latency only"; this is.

Trace matters much more than it looks on one chip: ~1.04x single-device, ~3x on a sharded mesh, because a 32-chip mesh has to dispatch to every chip.

Two dead ends recorded so nobody re-sweeps them: `max_c_in_block=256` measures 287.0 ms / 45.82 dB (0.02 dB over the default 128, i.e. nothing) and 512 exceeds L1 during program build. And 32-way T-sharding cannot beat 8-way at this length: `_upload_BCT` pads T to `32 * factor`, so factor 8 and factor 32 both land 32 rows per chip -- identical per-chip work, 4x the wasted padding. The idle 24 chips are worth 4x *throughput*, not lower latency.

**No configuration reaches both ~280 ms and ~50 dB.** Inside a 300 ms budget the choice is 292.6 ms / 45.80 dB; ~50 dB costs ~450 ms.

## Reproducing any of the above

Setup, once per shell:

    source python_env/bin/activate
    export TT_METAL_HOME=$W
    export PYTHONPATH=$W/ttnn:$W                 # $W/ttnn FIRST -- otherwise the venv's ttnn-custom.pth
                                                 # silently wins and you measure a different checkout
    export MINIMAX_H3_MODEL_PATH=/path/to/MiniMax-H3-diffusers
    export CVD_OUT_DIR=/tmp/h3_clips             # optional; defaults under generated/audio_perf
    S=models/tt_dit/tests/models/minimax_h3/audio_perf/cpu_vs_device.py
    SHARD="CVD_MESH=4x8 CVD_T_FACTOR=8 CVD_MESH_AXIS=1 CVD_TRACED=1"

| latency | PSNR | command |
|---|---|---|
| 292.6 ms | 45.80 dB | `env $SHARD CVD_SPLIT_MODE=off CVD_TAP_MATMUL=0 CVD_PREFER_MAC=0 python $S` |
| 391.5 ms | 46.74 dB | `env $SHARD CVD_SPLIT_MODE=off CVD_TAP_MATMUL=1 CVD_PREFER_MAC=0 python $S` |
| 367.4 ms | 48.26 dB | `env $SHARD CVD_SPLIT_MODE=weight CVD_TAP_MATMUL=0 CVD_PREFER_MAC=0 python $S` |
| **448.4 ms** | **50.77 dB** | `env $SHARD CVD_SPLIT_MODE=weight CVD_TAP_MATMUL=1 CVD_PREFER_MAC=0 python $S` |
| 484.9 ms | 50.22 dB | `env $SHARD CVD_SPLIT_MODE=full CVD_TAP_MATMUL=0 CVD_PREFER_MAC=0 python $S` |
| 840.9 ms | 49.45 dB | `env $SHARD CVD_SPLIT_MODE=off CVD_TAP_MATMUL=0 CVD_PREFER_MAC=1 python $S` |
| 1103.8 ms | 81.89 dB | `env $SHARD python $S` -- no lever vars: the constructed defaults |
| 1247 ms | 45.80 dB | `env CVD_SPLIT_MODE=off CVD_TAP_MATMUL=0 CVD_PREFER_MAC=0 python $S` -- single device |
| **5902 ms** | **81.99 dB** | `python $S` -- single device, constructed defaults, nothing set at all |

The last two rows are the equivalence check worth re-running after any change to the sharded paths: 81.89 dB sharded against 81.99 dB unsharded at the same levers. The remaining 0.1 dB is the T-shard's halo/pad residual, and it only becomes visible once every other error term has been spent away.

`CVD_MAX_C_IN_BLOCK=512` and `=2048` exceed L1 during program build (2048 asks for 4.25 MB of circular buffers against 1.57 MB). `CVD_T_FACTOR` must equal the length of the axis it shards: on a 4x8 mesh only `(4, axis 0)` and `(8, axis 1)` work. Accuracy is scored against the CPU/diffusers reference, so it is absolute; pass `CVD_BASELINE_PSNR=<dB>` to assert rather than print. Read the lever set from each run's own `levers:` line -- the output filenames encode only the mesh/trace configuration.

The suite, and the T-parallel table above:

    pytest models/tt_dit/tests/models/minimax_h3/test_audio_minimax_h3.py -q

## Known flake

`test_roundtrip[blackhole-single_device]` intermittently fails with
`TT_FATAL ... Can't get device from mesh buffer, already deallocated`, in roughly **1 of 3 full-suite
runs** (rarely when `t_parallel` is deselected), and it passes in isolation and on re-run. It is not
introduced here: it reproduced across every code state tested, including one where the only suspect --
a `SnakeBeta` timestep fold returning a view of a freed temporary -- had been removed. It is
single-device, so it does not touch the sharded paths or the CCL cache these fixes change. Re-run before
assuming a real failure; the residual cause is not identified.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rouzbeh/audio-decode-main)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rouzbeh/audio-decode-main)


models/tt_dit/tests/models/minimax_h3/test_audio_minimax_h3.py
```diff
@@ -401,3 +459,219 @@ def test_audio_decode_traced(mesh_device):
     decoder.release_trace()
 
     assert psnr_db > 60.0, f"traced output diverges from untraced: PSNR {psnr_db:.2f} dB"
+
+
+# -------------------------------------------------------------------- T-parallel audio decode
+#
+# Resurrected per this file's own instruction: the module docstring said to bring this back "if it
```
sadesoyeTT: AI code comments are unfortunately geared for themselves. We should clean up.

models/tt_dit/layers/audio_ops.py
```diff
@@ -1049,7 +1157,10 @@ def _forward_tap_matmul(self, x_BTC: ttnn.Tensor) -> ttnn.Tensor:
                 ttnn.deallocate(accumulator)
                 ttnn.deallocate(term)
                 accumulator = new_accumulator
-        if x_padded is not x_BTC:
+        # Only free what we allocated. Sharded, `x_padded` is the ccl_manager's cached ping-pong buffer,
```
sadesoyeTT: Do we need to explicitly do this?

models/tt_dit/parallel/manager.py
```diff
@@ -643,6 +644,19 @@ def get_np_ping_pong_buffer(
 
         cache_key = ("np", tuple(output_shape), dtype)
 
+        # These buffers are handed out as `neighbor_pad_async`'s output, so a caller that treats the
```
sadesoyeTT: Do you know why we were deallocating?

models/tt_dit/layers/audio_ops.py
```diff
@@ -856,17 +950,11 @@ def __init__(
...
-        # route owns the halo exchange and the C_in gather, and duplicating that here would be a second
-        # place for the same invariant to be wrong. Sharded audio decode is off by default.
-        self.tap_matmul = (
-            tap_matmul
-            and dtype == ttnn.float32
-            and stride == 1
-            and not sharded
-            and channel_axis(parallel_config) is None
-        )
+        # Needs stride 1 to index taps directly and fp32 to be worth doing. T-sharding is supported --
```
sadesoyeTT: Cleanup AI comments.

sadesoyeTT (approved):

models/tt_dit/parallel/manager.py
rouzbehTT: This is a left over. Good catch!
The audio code was freeing the manager's buffer, leaving a dead entry in a cache

models/tt_dit/layers/audio_ops.py
rouzbehTT: The old code was if x_padded is not x_BTC:, which correctly excluded the case below:
`sharded path`
and
`the ccl_manager ping-pong buffer from _t_neighbor_pad`

In this case, it is borrowed and we shouldn't be deallocating.

I changed removed this specific flag and change the logic a little bit

jonathansuTT (approved):

PR #53075, opened 2026-08-13
Replace the GMG_UNGROUPED_TOP8 define with a defaultless template parameter

pmilenkovicTT:
### PR Category

Cleanup / API improvement

### Summary

Follow-up to #52747: `generalized_moe_gate` path select becomes a **defaultless template parameter** instead of a preprocessor define.

- `generalized_moe_gate<bool ungrouped_top8, ...>` — the `#if GMG_UNGROUPED_TOP8 / #else / #endif` block becomes `if constexpr`. The "no silent default" property is preserved and strengthened: the parameter has no default value, so the choice is enforced by the type system at every call site (the old `#error` guard is gone because it is no longer reachable).
- **ttnn op**: the descriptor builder passes `moe_gate_ungrouped_top8` as a named compile-time arg (replacing the `GMG_UNGROUPED_TOP8` defines injection on all three kernels); the kernel threads it through `ComputeCTArgs` to both call sites.
- **b1 demo**: drops `#define GMG_UNGROUPED_TOP8 0` and passes `/*ungrouped_top8=*/false` explicitly.
- **tt-blaze** keeps its vendored define-based copy until its rewire, which flips call sites the same way (mechanical, tracked in tt-blaze#1971).

**Stacked on #52747** (currently in the merge queue) — this PR shows both commits until it lands, then collapses to one.

### Verification

- **Instruction-identical codegen**: the moe-gate kernel TU was compiled at the base commit (define version) and at this PR's commit (template version) with the production Blackhole JIT flag set, for all three TRISC contexts; the disassemblies are identical to the instruction (the only diff is the object filename in the objdump header). `if constexpr` produces exactly the code `#if` did.
- **Device bit-exact**: raw device outputs across all 12 `test_deepseek_moe_gate` parameter combos (batch × sigmoid × 3 seeds) are bit-identical between the define and template versions on BH p100a.
- No file this PR touches has changed on main since the change was authored (verified per-file).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:pmilenkovic/gmg-template-toggle)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:pmilenkovic/gmg-template-toggle)

### DIFF
--- a/models/common/tests/modules/moe/test_generalized_moe_gate.py
+++ b/models/common/tests/modules/moe/test_generalized_moe_gate.py
@@ -390,7 +390,7 @@ def test_generalized_moe_gate_512_global(device, batch_size, enable_sigmoid, see
 def test_generalized_moe_gate_grouped(device, batch_size, enable_sigmoid, seed):
     """DeepSeek GROUPED gate via ``generalized_moe_gate(grouped=True)``: 256 experts = 8 groups × 32 ->
     top-2-sum per group -> top-4 groups -> top-8, linear renorm + scale. Confirms the unified op's grouped
-    path (compiled with GMG_UNGROUPED_TOP8=0) matches the grouped golden — the path the standalone
+    path (ungrouped_top8=false via the moe_gate_ungrouped_top8 CT arg) matches the grouped golden — the path the standalone
     ``deepseek_moe_gate`` op used to own. grouped fixes top-8 + linear renorm, so topk / output_softmax are
     not swept (the op rejects other values in that mode)."""
     eps, scaling_factor = 1e-20, 2.5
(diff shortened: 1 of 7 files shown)


ndivnicTT (approved):
LGTM!

mtairum (approved):

tt-aho (approved):

bbradelTT (approved):

PR #53053, opened 2026-08-13
Fix ring-joint SDPA split-forwarding synchronization

pavlejosipovic:
## Summary

Fixes the even-ring split-forwarding synchronization introduced by PR #52730 (`c080ab0cfc94`, Ring joint SDPA and the LTX-2.3 distilled AV pipeline).

The feature splits the diametric KV shard across both fabric links so ring-joint SDPA can overlap communication and compute. The fused receiver waited for `backward_writes_expected + 1` direction-1 signals, but that semaphore includes an initial local-write signal. This released the diametric shard one remote arrival early, allowing SDPA to consume it while its second half was still being written.

This change waits for `backward_writes_expected + 2`, then makes the split-forwarding decision a single host-derived compile-time protocol flag shared by the ring-joint receiver and both all-gather directions. It keeps latent-V, sliding-window, and indexer-score paths on their established unsplit protocols, preserves legacy standalone all-gather gating, and keys sliding-window configuration in the program cache.

## CI failures

The regression first appeared after `c080ab0cfc94` in scheduled Blackhole sanity runs. The representative failure was [run 31699848471, job 94452565434](https://github.com/tenstorrent/tt-metal/actions/runs/31699848471/job/94452565434):

- `test_ring_joint_attention_chunked_nd_sharded_indexed_kv_cache_accuracy[bf16_acc]`
- `test_ring_joint_attention_chunked_nd_sharded_indexed_kv_cache_accuracy[fp32_acc]`

Both failed on the first chunk with PCC around 0.979 in CI. The same split protocol could deadlock latent-V cache replay in `test_ring_mla_chunked_kv_actual_isl_indexed_reuse_max_accuracy_and_determinism`.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pjosipovic/fix-ring-joint-sdpa-split-forwarding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pjosipovic/fix-ring-joint-sdpa-split-forwarding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:pjosipovic/fix-ring-joint-sdpa-split-forwarding)
[...]

PR #52730, opened 2026-08-10
[Feature] Ring joint SDPA and the LTX-2.3 distilled AV pipeline

nwoodallTT:
# [Feature] Ring joint SDPA and the LTX-2.3 distilled AV pipeline

### PR Category

Feature

---

> **Stacked PR 3 of 3.** Base: PR 2 (`nwoodall/ltx-neighbor-pad-halo-conv3d`, #52514),
> which is itself based on PR 1 (#52513, merged). This PR is the model-side wiring
> and **genuinely depends on both**: the VAE calls `neighbor_pad_halo` /
> `halo_scatter` from PR 2, and the transformer calls the MMRS and strided-AGMM ops
> from PR 1. It must merge last.
>
> Until #52514 lands, this PR's diff against `main` also contains PR 2's commits.
> The delta that is *this* PR's own work is **15 commits, 27 files changed,
> 2830 insertions(+), 196 deletions(-)**. Commits are referenced by subject below
> rather than by hash, since the stack is rebased as the lower PRs land.

## Summary

Brings the LTX-2.3 distilled two-stage audio-video pipeline up on a Blackhole
4x8 galaxy (32 chips, `FABRIC_1D_RING`, `num_links=2`) at **6.1 s end-to-end
compute** for a 6 s 1920x1088 video, by routing the model onto the ops added in
PRs 1 and 2.

### Ring-attention all-gather and cross-padded joint SDPA

Cross-padded joint ring SDPA, plus even-ring split-forwarding for the
ring-attention all-gather (the diametric slice is split across both links, with
direction 1 receiving and relaying one extra half). This lets the VAE's joint
attention run on the same ring topology as the rest of the pipeline instead of
falling back to a different collective shape.

*Commits:* `sdpa: ring-attention all-gather and cross-padded joint SDPA` ·
`test: coverage for ring joint SDPA including cross-padded`

### LTX model wiring

VAE conv3d routes through `neighbor_pad_halo` + `halo_scatter` into a standalone
conv3d. `get_conv3d_config` carries a per-shape table marking shapes `halo_last`
(interior-then-boundary two-pass) or `force_spatial_parallel`, so the interior
conv pass hides the halo exchange. Those entries carry their measured numbers
in-comment.

The per-head attention gate folds into the QKV/Q matmul via variable-width
chunks, honored only on the fused AGMM path (TP>1 on Ring) with a standalone gate
op elsewhere; the gate is `num_heads/TP` columns per device (sub-tile), padded to
[...]

PR #38256, opened 2026-02-20
A balanced traffic pattern for AG minimal

llongTT:
### Ticket
[Link to Github Issue](https://docs.google.com/spreadsheets/d/1M3x8flyi5hjsSqHCw9DRmMBYV_KLWVjgdayAG0VZY3Y/edit?gid=0#gid=0)

### Problem description
AG minimal has low fabric utilization, especially for device ring = 4 cases.
WAN 2.2 model perf requires AG to reach 92% fabric utilization to achieve the target.

### What's changed
1. Balance the packet traffic between forward worker and backward worker such that they split the last slice. It saves 25% of fabric latency by going from "2 slice forward, 1 slice backward" to "1 and half slice both forward and backward" in device ring = 4.
2. If balanced traffic feature enabled, also split the local writes half/half, to reduce the noc traffic on backward worker.
3. Extend the feature to all device rings >2 and ring size even cases.

### Checklist

- [x] [All post-commit tests](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml?query=branch:llong/dit_ag_min)
- [x] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:llong/dit_ag_min)
- [x] [cpp-unit-tests](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:llong/dit_ag_min)
- [ ] New/Existing tests provide coverage for changes

- [x] [(Galaxy) integration tests]https://github.com/tenstorrent/tt-metal/actions/runs/22455835923
- [x] [(Galaxy) unit tests]
Feature branch: https://github.com/tenstorrent/tt-metal/actions/runs/22455806968
Main branch: https://github.com/tenstorrent/tt-metal/actions/runs/22455802699

- [x] [(Galaxy) demo tests]
Feature branch: https://github.com/tenstorrent/tt-metal/actions/runs/22455908585
Main branch: https://github.com/tenstorrent/tt-metal/actions/runs/22455904714

- [x] [(Galaxy) perf tests]
Feature branch: https://github.com/tenstorrent/tt-metal/actions/runs/22464529049
Main branch: https://github.com/tenstorrent/tt-metal/actions/runs/22464525251

- [x] [vLLM nightly tests]
Feature branch: https://github.com/tenstorrent/tt-metal/actions/runs/22455918368
[...]

PR #53076, opened 2026-08-13
sdpa: revert even-ring split forwarding from ring-joint SDPA

rsalmanTT:
## Summary

Reverts the even-ring **split-forwarding** feature that #52730 (`c080ab0cfc94`) added to ring-joint SDPA. This is the surgical alternative to #53053, which Pavle has handed off to us.

Split-forwarding is the *entirety* of what #52730 changed in these 6 files (3 hunks per all-gather kernel, all split-related; the metadata / KV-cache-slot features in the same files predate it), and **nothing has touched them since** on `main`. So the revert is `git checkout c080ab0cfc94~1 -- <6 files>`: an exact, conflict-free inverse, `+35/-110` against `+110/-35`.

## What was failing

`blackhole-multi-card-fast-sanity-tests / sdpa nightly tests (QB2 only)` — 10 test failures, which are **2 real bugs plus 8 consequences**:

| Group | Tests | Signature |
|---|---|---|
| Hang | `test_ring_mla_chunked_accuracy[kimi_k3, kimi50k]` | `TIMEOUT: device timeout ... device is unrecoverable` |
| Cascade | 6 tests | `Device 0 init: failed to initialize FW!` — downstream of the hang |
| PCC | `..._chunked_nd_sharded_indexed_kv_cache_accuracy[fp32_acc, bf16_acc]` | PCC 0.9794 / 0.9796 vs 0.997 / 0.994 |

The PCC failures sit earlier in the file than the hang, so they ran before the device died and are independent of it. `fp32` and `bf16` give near-identical PCC (0.97944 / 0.97957), which reads as systematic error rather than a race.

## Why revert rather than patch again

1. **The design requires a cross-op mirror.** The all-gather in `experimental/ccl` decides to split; a consumer in `transformer/sdpa` must independently reproduce that decision. #52730's own comment says it: *"Must match the all-gather kernels' split-forwarding gate exactly."* #53053 improves this by passing the flag instead of re-deriving it, but the underlying wait is still a single coarse threshold (`split_second_half_wait = backward_writes_expected + 1`) for the whole shard rather than a per-chunk event count — which is exactly why `+1` vs `+2` was an open question that consumed two CI cycles.

2. **Two further consumers of the same rt-arg block were never updated**, both currently latent:
   - `fused_op_indexer.hpp:36` advances `rt_args_idx += 2` while `push_ring_sdpa_fused_op_rt_args` now pushes 5. Harmless only because that constructor has no callers.
   - `get_next_ring_id_and_consume_one_signal()` never received the split-shard second-half wait its sibling `get_next_ring_id_and_sync()` got, and it uses `down(1)` rather than `wait_min`. Masked because only the sliding-window path calls it, which #53053 gates off.

   Same shape of defect twice: the feature added protocol state that not every consumer learned about.

3. **Prior art keeps failing.** The optimization originated as Ligang Long's AG-minimal balanced traffic pattern and was **reverted twice** (`#36607` → `#37832`, `#37878` → `#38202`) before landing on the third attempt in `#38256`.

## What is NOT reverted

**#52513's split forwarding in strided AGMM stays**, and it is a genuinely independent, correct implementation — not a copy. It splits at *chunk* granularity with byte-identical producer/consumer predicates (`minimal_default_writer.cpp:330` vs `minimal_default_mm_signal_aggregator.cpp:81`) and waits on a monotonic event counter (`event_target++` then `noc_semaphore_wait_min` per worker per band). There is no threshold to get wrong, so the miscount here cannot occur there. LTX's strided-AGMM path is therefore unaffected by this revert.

## Verification status — please read

**This revert cannot be verified outside CI, and is not yet verified.**

The suite runs on exactly one SKU (`tests/pipeline_reorg/blackhole_multi_card_sanity_tests.yaml:51`, `skus:` lists only `bh_quietbox_2`), and the failing shapes require `sp=4`:

| | QB2 (fails) | 4x8 galaxy (dev box) |
|---|---|---|
| devices | 4 | 32 |
| ring (`sp_size`) | 4 | 8 |
| `tp_size` | 1 | 4 |
| SDPA cores | 100 | 110 |
| chunk-0 `logical_n` | **256** | 512 |

`MeshConfig.detect()` derives all of this from the `/dev/tenstorrent` count with no override, and sequence length scales as `128 x sp_size`. CI's failing case `test_ring_mla_chunked_accuracy[kimi_k3-all-qk-chunk2560]` **does not exist** on a galaxy — it parametrizes as `[kimi_k3-q32-k512-chunk5120]`. The failing shape is unreachable there.

What local galaxy testing *can* show is absence of ring8 regression:

- `determinism or nd_sharded_indexed_kv_cache_accuracy`: **27/27 passed** (720s)
- chunked accuracy batch: in progress at time of opening
- non-chunked accuracy batch: pending
- LTX distilled e2e vs a 26.9 s/gen baseline: pending

Caveat: there is **no pre-revert ring8 baseline** for those 27 (the baseline job was lost to a tooling error on my side). They pass with the revert applied; that is not the same claim as "unchanged by the revert".

**The gate is a `workflow_dispatch` of `blackhole-sanity-tests.yaml` against this branch with `run-blackhole-multi-card-fast-unit-tests` enabled.** It does not fire automatically on PRs. Pass condition: the 2 real failures go green, the 6 cascade failures recover, and Minimax reuse-KV does not regress (that is what #53053's `+2` experiment broke, per QB2 job 94493466874).

## Follow-ups (not in this PR)

- Fix the two latent defects above so they cannot bite whoever reintroduces split forwarding.
- If split forwarding is wanted back, the AGMM shape is the model: per-chunk split, monotonic event counter, producer and consumer deriving the predicate from the same constants in the same op.
- No galaxy SKU runs `tests/nightly/blackhole/sdpa/`, which is how this reached `main`. Worth deciding whether that suite should also run on a 4x8.

cc @cglagovichTT @pavlejosipovic @nwoodallTT

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rsalman/revert-ring-joint-split-forwarding-2026-08-13)

### DIFF
--- a/ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_reader.cpp
+++ b/ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_reader.cpp
@@ -185,13 +185,6 @@ void kernel_main() {
         }
     }
 
-    // Mirror the writer's split-forwarding (see ring_attention_all_gather_writer.cpp): on an even ring the diametric
-    const bool split_forwarding_enabled = (topology == Topology::Ring) && (ring_size % 2 == 0) && (ring_size > 2);
-    if (split_forwarding_enabled && direction == 1) {
-        slices_expected++;
-        writes_expected++;
-    }
-
     while (slices_received < slices_expected) {
         // Do i expect more from the backward direction?
         // In the linear case, I expect num_targets_backward_direction slices from the left
@@ -231,32 +224,16 @@ void kernel_main() {
         // In the ring case, if I have received on the right less than my targets on the left, forward
         if ((topology == Topology::Linear && writes_expected > 0) ||
             (topology == Topology::Ring && (slices_received < (writes_expected + 1)))) {
-            // The last slice we relay is the diametric shard of our downstream neighbor
-            const bool is_split_forwarded_slice = split_forwarding_enabled && (slices_received == writes_expected);
             for (uint32_t input_idx = 0; input_idx < num_inputs; input_idx++) {
+                uint32_t tiles_read = input_tile_id_start[input_idx];
+                uint32_t tiles_to_read = input_tile_id_end[input_idx];
+
+                uint32_t output_tile_id_start = 0;
+                uint32_t pages_read_in_row = input_tile_id_start[input_idx] % input_tensor_Wt[input_idx];
+                uint32_t row_offset =
+                    (input_tile_id_start[input_idx] / input_tensor_Wt[input_idx]) * output_tensor_Wt[input_idx];
                 uint32_t slice_Wt = input_tensor_Wt[input_idx];
                 uint32_t stride_Wt = output_tensor_Wt[input_idx];
-
-                // Packet-aligned midpoint of this input's per-batch-head page range (matches the writer)
-                const uint32_t total_pages = input_tile_id_end[input_idx] - input_tile_id_start[input_idx];
-                const uint32_t num_packets = (total_pages + packet_size_in_pages - 1) / packet_size_in_pages;
-                const uint32_t first_half_pages = (num_packets / 2) * packet_size_in_pages;
-                const bool split_this_input = is_split_forwarded_slice;
-                uint32_t relay_start = input_tile_id_start[input_idx];
-                uint32_t relay_end = input_tile_id_end[input_idx];
-                if (split_this_input) {
-                    if (direction == 0) {
-                        relay_end = input_tile_id_start[input_idx] + first_half_pages;
-                    } else {
-                        relay_start = input_tile_id_start[input_idx] + first_half_pages;
-                    }
-                }
-
-                uint32_t tiles_read = relay_start;
-                uint32_t tiles_to_read = relay_end;
-                uint32_t output_tile_id_start = 0;
-                uint32_t pages_read_in_row = relay_start % slice_Wt;
-                uint32_t row_offset = (relay_start / slice_Wt) * stride_Wt;
                 if (gather_dim == 3) {
                     output_tile_id_start = actual_sender_chip_id * input_tensor_Wt[input_idx];
                 } else {
@@ -285,10 +262,11 @@ void kernel_main() {
                             }
                             return pid;
                         });
-                    pages_read_in_row = relay_start % slice_Wt;
-                    row_offset = (relay_start / slice_Wt) * stride_Wt;
-                    tiles_read = relay_start;
-                    tiles_to_read = relay_end;
+                    pages_read_in_row = input_tile_id_start[input_idx] % input_tensor_Wt[input_idx];
+                    row_offset =
+                        (input_tile_id_start[input_idx] / input_tensor_Wt[input_idx]) * output_tensor_Wt[input_idx];
+                    tiles_read = input_tile_id_start[input_idx];
+                    tiles_to_read = input_tile_id_end[input_idx];
                     output_tile_id_start += output_tensor_Wt[input_idx] * output_tensor_Ht[input_idx];
                 }
             }
--- a/ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_writer.cpp
+++ b/ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_writer.cpp
@@ -294,12 +294,6 @@ void kernel_main() {
         }
     }
 
-    // On an even ring the terminal relayed slice
-    const bool split_forwarding_enabled = (topology == Topology::Ring) && (ring_size % 2 == 0) && (ring_size > 2);
-    if (split_forwarding_enabled && direction == 1) {
-        writes_expected++;
-    }
-
     while (slice_writes < writes_expected) {
         // Direction == backward
         // Did I get something from my left to send to my right?
@@ -321,7 +315,6 @@ void kernel_main() {
             slice_chip_id = my_chip_id - slice_writes - 1;
             actual_slice_chip_id = (slice_chip_id < 0) ? ring_size + slice_chip_id : slice_chip_id;
         }
-        const bool is_split_forwarded_slice = split_forwarding_enabled && (slice_writes == writes_expected - 1);
         for (uint32_t input_idx = 0; input_idx < num_inputs; input_idx++) {
             uint32_t tiles_read = input_tile_id_start[input_idx];
             uint32_t tiles_to_read = input_tile_id_end[input_idx];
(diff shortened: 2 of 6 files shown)


rsalmanTT:
## Verified on QB2 — the failing suite is green

The gate for this revert is the one SKU that reproduces the failures. It passes:

**✅ [`sdpa nightly tests (QB2 only)` [bh_quietbox_2]](https://github.com/tenstorrent/tt-metal/actions/runs/31733956385/job/94563823330)** — run [31733956385](https://github.com/tenstorrent/tt-metal/actions/runs/31733956385) on this branch, `workflow_dispatch` of `blackhole-sanity-tests.yaml`.

That single job is easy to miss: there are 129 check-runs on the head commit and the two red ones sort to the top, so the direct link above is the fastest route.

### Before / after, same job, same SKU

| Branch | `sdpa nightly tests (QB2 only)` |
|---|---|
| `main` @ `8a934ae0186` | ❌ [10 failures](https://github.com/tenstorrent/tt-metal/actions/runs/31720311397) |
| #53053 (`+1` threshold attempt) | ❌ [10 failures](https://github.com/tenstorrent/tt-metal/actions/runs/31719235944/job/94514373094) |
| **this branch** | ✅ [**success**](https://github.com/tenstorrent/tt-metal/actions/runs/31733956385/job/94563823330) |
| last known-good before #52730 | ✅ [success](https://github.com/tenstorrent/tt-metal/actions/runs/31550656743) |

All three failure groups are cleared: the `test_ring_mla_chunked_accuracy` device hang, the `mla_indexed_nd_sharded_kv` PCC drop (0.9794 vs 0.997), and the 6 cascade `Device 0 init: failed to initialize FW!` failures downstream of the hang. Minimax reuse-KV also passes — that is the workload #53053's earlier `+2` experiment broke (QB2 job 94493466874), so it was the specific regression risk here.

Sibling QB2 jobs in the same run are green too: `TTNN multi-device CCL, SDPA PERF, and indexer tests`, `fabric fast unit tests (1D, 2D, single erisc)`, `ttnn ops fast unit tests (ccl, DRAM prefetcher)`.

### Regression window, independently confirmed

Scheduled QB2 nightlies on `main`, 2026-08-12 (`c080ab0cfc94` landed 11:26:34Z):

| Time | sha | Has #52730 | sdpa job |
|---|---|---|---|
| 00:34 | `ea042c4ad62` | no | ✅ success (ran, not skipped) |
| 04:38 | `69096826694` | no | ❌ *infra*: runner lost communication |
| 08:35 | `b161df2172b` | no | ❌ *infra*: `Value cannot be null (ContainerId)` |
| **12:24** | `4a89582b458` | **yes** | ❌ **10 real test failures** |
| 16:22 → | | yes | ❌ 10 real test failures |

The two pre-#52730 failures were infrastructure, not code. The first nightly *containing* #52730 is the first with genuine test failures.

### LTX perf impact: none material

Traced steady state on this branch (gen #2, pure replay; 1088x1920, 145 frames, mesh (4,8) sp=8 tp=4 Ring, `IN0_SUB_CHUNKS=1`):

| Stage | Time |
|---|---|
| Encoder | 0.18 s |
| Stage 1 denoise | 2.35 s |
| Latent upsample | 0.14 s |
| Stage 2 denoise | 2.73 s |
| VAE decode | 0.59 s |
| Audio decode | 0.42 s |
| **Total** | **6.41 s** |

Against the **6.1 s** advertised in #52730, roughly 5%. Trace capture and replay are unaffected by the revert (46.9 s capture → 9.7 s → 6.41 s replay), which was the plausible way removing split-forwarding could have broken tracing.

Caveat, stated plainly: 6.1 s comes from #52730's description, measured on a different commit under conditions I have not reproduced. This is **not** a controlled A/B. A pre-revert traced run on the same tree would settle it and is cheap; happy to add it if the number matters for the decision.

### Local galaxy testing (weaker evidence, included for completeness)

41 ring-joint SDPA accuracy + determinism tests pass at ring8/tp4 on a 4x8 galaxy, 0 genuine failures. This only demonstrates absence of ring8 regression — the failing configs need `sp=4` and are unreachable on a galaxy (`MeshConfig.detect()` derives ring size from the `/dev/tenstorrent` count, and sequence length scales as `128 x sp_size`, so CI's `[kimi_k3-all-qk-chunk2560]` parametrizes as `[kimi_k3-q32-k512-chunk5120]` there).

Two caveats on that local run: there is no pre-revert ring8 baseline (that job was lost to a tooling error on my side), so the claim is "passes with the revert applied", not "unchanged by the revert". And two of the four `test_ring_mla_chunked_accuracy` cases went unrun — the broker reaps on stdout silence, and these tests compute a torch CPU reference for minutes while the device idles. The device-side watchdog (`TT_METAL_OPERATION_TIMEOUT_SECONDS=300` with a triage callback) never fired and wrote no artifacts, confirming no op was stuck on device.

### Not covered by this run

I disabled `run-ttnn-unit-tests`, `run-profiler-tests`, `run-umd-sanity-tests` and `run-ops-unit-tests` on the dispatch to get the answer quickly, so nothing outside the multi-card path has been exercised on this branch. A full-default dispatch is worth doing before merge.

The two red checks on this PR (`code-analysis / Clang Tidy`, and `PR Gate Status` which merely aggregates it) are a runner checkout timeout — `The action 'Check out repo' has timed out after 20 minutes` then `exit code 127`, so clang-tidy never ran. Unrelated to this diff; being re-run.

rsalmanTT:
## Perf: controlled A/B — the revert costs LTX nothing measurable

This supersedes the perf paragraph in my previous comment, which compared against the 6.1 s figure published in #52730 and estimated ~5%. That was not a controlled comparison and it overstated the cost. Here is the real measurement.

Both arms on the same machine (BH 4x8 galaxy), same commit apart from the 6 reverted files, same env file, same harness, same seed, same measurement point (gen #2, traced pure replay). 1088x1920, 145 frames, mesh (4,8) sp=8 tp=4 Ring, `LTX_TRACED=1`, `RUN_WARMUP=1`, `IN0_SUB_CHUNKS=1`.

| Stage | Pre-revert (split-fwd ON) | Post-revert | Δ |
|---|---|---|---|
| Encoder | 0.18 s | 0.18 s | 0 |
| Stage 1 denoise | 2.32 s | 2.35 s | +0.03 |
| Latent upsample | 0.13 s | 0.14 s | +0.01 |
| Stage 2 denoise | 2.72 s | 2.73 s | +0.01 |
| VAE decode | 0.59 s | 0.59 s | 0 |
| Audio decode | 0.42 s | 0.42 s | 0 |
| **Total** | **6.36 s** | **6.41 s** | **+0.05 s (0.8%)** |

Trace capture progression, for completeness: 45.6 → 9.0 → 6.36 s pre-revert vs 46.9 → 9.7 → 6.41 s post-revert.

**Read:** split-forwarding buys LTX nothing measurable at steady state. Per-stage deltas are ≤0.03 s, at or below single-sample resolution, and the pre-revert arm was also faster on the two capture generations — the wrong direction for a real effect, which points at run-to-run noise rather than signal. One sample per arm, so treat 0.05 s as "indistinguishable from zero", not as a measured 0.8% regression.

Two side findings:

- Trace capture and replay work in **both** arms. The traced-mode capture hang noted in our LTX harness (2026-07-23, strided-AGMM `to_out` swap) did not reproduce either way, so it is unrelated to split-forwarding.
- The **6.1 s** in #52730's description is not reproducible on this box in either arm; both land at ~6.4 s. Different commit and measurement conditions, so no contradiction implied — just don't use 6.1 s as the baseline for judging this revert.

cglagovichTT (approved):

PR #53084, opened 2026-08-13
[ttnn] Always validate Metal 2.0 program args on cache miss, gate only hit-path re-validation

dgomezTT:
### Ticket
Follow-up to #50945.

### Problem description
#50945 added the `validate_program_args` config (off by default, enabled in the sanity pipelines) to keep Metal 2.0 host-side arg validation off the user hot path. As wired, the polarity is inverted on both ends:

- **Cache miss (cold path) was gated.** With the default config, `MakeProgramFromSpec` skips `ValidateProgramSpec` — the semantic legality checks — and `SetProgramRunArgs` skips its validation. A malformed spec goes straight into the program cache at the one point where validation is nearly free (amortized against a full program build).
- **Base-concept cache hit (hot path) was unconditional.** `UpdateTensorArgs` has no skip parameter and calls `ValidateTensorArgs` on every cache hit, so the steady-state dispatch path pays the validation cost everywhere — including off-CI, where the config was meant to remove it. Only the custom-factory hit path (`UpdateProgramRunArgs`) was gated correctly.

### What's changed
- Cache-miss build in `mesh_device_operation_adapter.hpp` always validates: `MakeProgramFromSpec` / `SetProgramRunArgs` are called without `skip_validation`.
- `UpdateTensorArgs` gains a defaulted `bool skip_validation = false` (mirroring `SetProgramRunArgs` / `UpdateProgramRunArgs`), gating only the `ValidateTensorArgs` call; the base-concept cache-hit path passes it from `validate_program_args`, same as the custom-factory hit path already did.
- `validate_program_args` now purely controls hit-path re-validation: CI sanity re-validates every dispatch, everything else validates once per cache entry at build time. All pipelines now get spec validation on cache miss, including those that don't set the config.

### Checklist
- [ ] CI passes

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/validate-args-hit-path-only)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/validate-args-hit-path-only)

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_run_args.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_run_args.cpp
@@ -1798,6 +1798,40 @@ TEST_F(ProgramRunArgsTestGen1, UpdateTensorArgs_TensorSpecMismatchFails) {
             "TensorArgument for binding 'input_tensor' supplied a MeshTensor whose TensorSpec does not match")));
 }
 
+// Locks the gate's polarity: the three tests above prove UpdateTensorArgs validates by default, so
+// this one proves skip_validation=true is what turns it off — and still performs the address patch.
+TEST_F(ProgramRunArgsTestGen1, UpdateTensorArgs_SkipValidationBypassesSpecCheck) {
+    NodeCoord node{0, 0};
+    ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
+    auto binding = MakeMinimalTensorParameter("input_tensor");  // shape {1, 32}
+    spec.tensor_parameters = {binding};
+    BindTensorParameterToKernel(spec.kernels[0], "input_tensor", "input_ta");
+
+    Program program = MakeProgramFromSpec(*mesh_device_, spec);
+
+    MeshTensor tensor = AllocateTensorForBinding(*mesh_device_, binding);
+    auto params = MakeRunArgsForMinimalSpec(node, {}, {});
+    params.tensor_args = {
+        {TensorParamName{"input_tensor"}, TensorArgument{tensor}},
+    };
+    SetProgramRunArgs(program, params);
+
+    auto page_config = tt::tt_metal::PageConfig(tt::tt_metal::Layout::ROW_MAJOR);
+    auto memory_config =
+        tt::tt_metal::MemoryConfig{tt::tt_metal::TensorMemoryLayout::INTERLEAVED, tt::tt_metal::BufferType::DRAM};
+    auto tensor_layout = tt::tt_metal::TensorLayout(tt::tt_metal::DataType::BFLOAT16, page_config, memory_config);
+    auto wrong_spec = tt::tt_metal::TensorSpec(tt::tt_metal::Shape{1, 64}, tensor_layout);
+    MeshTensor wrong_tensor = MeshTensor::allocate_on_device(*mesh_device_, wrong_spec);
+
+    Table<TensorParamName, TensorArgument> tensor_args{
+        {TensorParamName{"input_tensor"}, TensorArgument{wrong_tensor}},
+    };
+    EXPECT_NO_THROW(UpdateTensorArgs(program, tensor_args, /*skip_validation=*/true));
+    EXPECT_EQ(
+        ReadBindingAddressFromCRTA(program, "dm_kernel", "input_tensor"), static_cast<uint32_t>(wrong_tensor.address()))
+        << "skip_validation should bypass only the checks, not the address patch";
+}
+
 TEST_F(ProgramRunArgsTestGen1, UpdateTensorArgs_PatchesBindingAddress) {
     NodeCoord node{0, 0};
     ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/program.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/program.hpp
@@ -102,9 +102,15 @@ void UpdateProgramRunArgs(Program& program, const ProgramRunArgs& params, bool s
 // ProgramSpec, exactly once. The supplied MeshTensor's TensorSpec must match the
 // TensorParameter's declared spec.
 //
+// PRE-CONDITION: If skip_validation is true, the caller guarantees the completeness and
+// spec-match requirements above.
+//
 // USE CASE: Program re-enqueue loops where the only per-enqueue ProgramRunArgs variation
 // is in the tensor args (i.e. which specific MeshTensors are operated on by the Program).
-void UpdateTensorArgs(Program& program, const Table<TensorParamName, ProgramRunArgs::TensorArgument>& tensor_args);
+void UpdateTensorArgs(
+    Program& program,
+    const Table<TensorParamName, ProgramRunArgs::TensorArgument>& tensor_args,
+    bool skip_validation = false);
 
 }  // namespace tt::tt_metal::experimental
 
--- a/tt_metal/impl/metal2_host_api/program_run_args.cpp
+++ b/tt_metal/impl/metal2_host_api/program_run_args.cpp
@@ -793,7 +793,8 @@ void SetProgramRunArgs(Program& program, const ProgramRunArgs& params, bool skip
     program_impl.mark_program_run_args_initialized();
 }
 
-void UpdateTensorArgs(Program& program, const Table<TensorParamName, ProgramRunArgs::TensorArgument>& tensor_args) {
+void UpdateTensorArgs(
+    Program& program, const Table<TensorParamName, ProgramRunArgs::TensorArgument>& tensor_args, bool skip_validation) {
     log_debug(tt::LogMetal, "Updating tensor args (partial fast-path)");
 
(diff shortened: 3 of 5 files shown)


dgomezTT:
CI status: **91 of 92 sanity jobs pass.**

The one red is `blackhole-multi-card-fast-sanity-tests / sdpa nightly tests (QB2 only) [bh_quietbox_2]`,
and it is a runner fault rather than anything from this change:

- `TT_THROW @ risc_firmware_initializer.cpp:1530: Device 0 init: failed to initialize FW! Try resetting
  the board.` Every later test in the job cascades from that, which is where the 73 failures / 19 errors
  come from.
- The same job with the same board-init failure has now hit #53065 twice (Device 0, then Device 2) and
  **main** (Sanity run 31727898275), so that quietbox needs a reset — it is failing repo-wide.
- Nothing in the failure involves program-arg validation. The only validation-related line in the log is
  the config dump reading `validate_program_args | false`, i.e. this PR's intended default.

akerteszTT (approved):

dgomezTT:
Added `ProgramRunArgsTestGen1.UpdateTensorArgs_SkipValidationBypassesSpecCheck` for the requested polarity lock.

Worth noting the request was already half-covered: `UpdateTensorArgs_{MissingTensorArg,UnknownTensorParameter,TensorSpecMismatch}Fails` all call `UpdateTensorArgs` with the default argument, so "the hit path validates unless told otherwise" is asserted three times over. What nothing covered was the new parameter itself — an inert `skip_validation` that never disabled anything would have passed the whole existing suite.

The new test supplies a tensor whose spec does not match the binding (same input the sibling test uses to assert a throw), passes `skip_validation=true`, and asserts both that it does *not* throw and that the binding address is still patched — so the flag cannot degrade into a full no-op and stay green.

Verified locally rather than by inspection (mock cluster, no HW needed, ~2s):
- restored build: 26/26 `ProgramRunArgsTestGen1.*` pass
- negative control: with the gate neutered to `ValidateTensorArgs` unconditionally, the new test **fails** on `TensorSpec does not match`, and passes again once restored

Cache-miss coverage is unchanged and stays where it was — `MakeProgramFromSpec`/`SetProgramRunArgs` are now called with their defaults, which the existing `TensorSpecMismatchFails`/`MissingTensorArgFails` tests already pin.

dgomezTT:
Correcting my earlier CI note on this PR: the failure was **not** the QB2 board firmware-init fault I attributed it to. That fault was real earlier in the day but the board has since recovered, and the actual failure underneath was different.

The real failure is an accuracy miss in the ring-joint SDPA / ring-MLA chunked tests:

```
test_ring_joint_attention_chunked_nd_sharded_indexed_kv_cache_accuracy[bf16_acc]
Chunked prefill mla_indexed_nd_sharded_kv-q32-k32 PCC/RMSE failures
(pcc_threshold=0.994): chunk 0 (logical_n=256): PCC=0.9795682179361413, RMSE=0.066283
```

Root cause is a **stale base**, not this diff. Two independent PRs (#53065 and #53084) were producing bit-identical PCC/RMSE numbers, which ruled out flakiness and pointed at something shared. Main landed `d3f62089783 sdpa: revert even-ring split forwarding from ring-joint SDPA (#53076)` after both branches merge-bases, so both still carried the reverted code while main did not — hence green on main, red here, identical digits on both.

Rebased onto current main (which includes the revert) and re-dispatched sanity.

ayerofieiev-tt (approved):

dgomezTT:
Full pipeline verification on the rebased head \`61f2950c400\` (this PR changes the program-cache path, so Sanity alone is not sufficient coverage).

| Pipeline | Result |
|---|---|
| Sanity tests (WH + real BH + Sim) | green |
| Nightly tt-metal L2 tests | green |
| (Runtime) Sanity Tests (all WH + BH SKUs) | green |
| All Model Tests — Tiers[1,2] Tests[unit,e2e] Model[all] SKUs[all] | 76 success / 7 failure / 17 skipped |

The 7 remaining All-Model-Tests failures all fail on `main` as well:

```
t1-e2e-tests   / Llama 3.3-70B e2e (Galaxy) [wh_galaxy_perf]
t1-e2e-tests   / Qwen3-32B e2e (Galaxy)     [bh_galaxy]
t1-unit-tests  / Qwen3-32B unit (Galaxy)    [wh_galaxy_perf]
t1-unit-tests  / GPT-OSS 120B unit          [bh_galaxy]
t1-unit-tests  / Qwen3.6 unit               [bh_quietbox_2]
t1-unit-tests  / TT-DiT Flux.1-dev unit     [bh_quietbox_2]
t1-unit-tests  / TT-DiT Wan2.2-I2V-A14B unit [wh_galaxy]
```

Baseline for that claim is the union of failures across recent `main` runs of the same workflow (same scope: Tiers[1,2], unit+e2e, all SKUs, 100 jobs). A single `main` run is not a fair baseline here — these Galaxy/quietbox failures are intermittent, and `main` run 31561177068 alone had 19 failures. Overlap with the surviving 7 is 100%.

First attempt had 22 failures; `rerun-failed-jobs` cleared 15 of them, which were dominated by `codeload.github.com` archive-download timeouts (the job never fetched the source) and quietbox action timeouts.

One failure in the first attempt deserves a note because it is *not* reproducible: `Gemma-3-4B e2e [wh_n150]` hit `TT_THROW: Statically allocated circular buffers in program 488 clash with L1 buffers`. It passed on rerun. It is also not reachable from this diff — `validate_circular_buffer_region` is called unconditionally at `tt_metal.cpp:1059` and `program.cpp:2666` with no `validate_program_args` gate, and this PR touches neither file nor any allocation code.

No performance claim is being made here: a Qwen2.5-VL-72B perf test failed the first attempt for exceeding its tolerance *upper* bound and passed on rerun, so that measurement is noise near the bound, not evidence of a speedup.

dgomezTT:
Correction to my earlier CI summary on this PR — that comment understated the failure count and drew two wrong conclusions. Cause was mine: I queried the run's jobs with `per_page=100`, but the run has **147 jobs**, so the list silently truncated.

Corrected numbers for All Model Tests run 31811478383 (paginated): **120 success / 10 failure / 17 skipped**, not the 76/7/17 I reported.

Re-triaged against the union of failures across **12** recent `main` runs of the same workflow (also paginated):

**9 of the 10 also fail on `main`:**
```
t1-e2e-tests  / Llama 3.3-70B e2e (Galaxy)   [wh_galaxy_perf]
t1-e2e-tests  / Qwen3-32B e2e (Galaxy)       [bh_galaxy]
t1-unit-tests / GPT-OSS 120B unit            [bh_galaxy]
t1-unit-tests / Qwen3-32B unit (Galaxy)      [wh_galaxy_perf]
t1-unit-tests / Qwen3.6 unit                 [bh_quietbox_2]
t1-unit-tests / TT-DiT Flux.1-dev unit       [bh_quietbox_2]
t1-unit-tests / TT-DiT Wan2.2-I2V-A14B unit  [wh_galaxy]
t2-e2e-tests  / Gemma-3-4B e2e               [wh_n150]
t2-e2e-tests  / Gemma-4-31B e2e              [wh_llmbox_perf]
```

Two specific corrections:

1. **Gemma-3-4B e2e [wh_n150]** (the `Statically allocated circular buffers ... clash with L1 buffers` throw): I previously said it passed on rerun. It does not — I had checked the wrong job (`Gemma-3-4B *unit* [wh_n150]`). It still fails, **but it also fails on `main`**, so it is pre-existing rather than introduced here. That matches the mechanism: `validate_circular_buffer_region` is called unconditionally at `tt_metal.cpp:1059` and `program.cpp:2666` with no `validate_program_args` gate, and this PR touches neither file nor any allocation code.

2. **Qwen2.5-VL-72B e2e [wh_llmbox_perf]** — the one failure **not** seen on `main` in 12 runs. I previously retracted the claim that this reflects a real speedup, calling it noise near the tolerance bound. **That retraction was wrong**, and it was based on the same truncated job list. The failure is reproducible and the log is explicit:

```
decode_t/s   431.66 vs expected 347.6  -- "much higher than expected. Please update the expected perf."
decode_t/s/u  13.49 vs expected 10.86  -- same
```

Decode is the cache-hit path, and this PR stops re-validating tensor args on every cache hit (`UpdateTensorArgs` is now gated by `validate_program_args`, default off). So the measured improvement is the expected consequence of the change, and the test fails because its expected values are stale — which is what its own warning says.

Net: 9 pre-existing failures, and 1 that is this PR making the hit path faster.

PR #53109, opened 2026-08-14
[Test Only] Trim redundant test_split/test_repeat_interleave/test_slice_for_conv parametrization to fix nightly data_movement timeout

### LINKED ISSUE #53460: [CI Failure] Nightly tt-metal L2 tests - ttnn nightly data_movement tests timing out at 40 min
### Pipeline Name
Nightly tt-metal L2 tests

### Test Name
ops-unit-tests / ttnn nightly data_movement tests [wh_n150_civ2] (also affects wh_n300_civ2, bh_p100, bh_p150b_civ2)

### Path to Test in Repo
tests/pipeline_reorg/ops_unit_tests.yaml (runs `pytest tests/ttnn/nightly/unit_tests/operations/data_movement`)

### Non-Deterministic (ND) failure?
No

### Failure type
Timeout

### Link to Failing Pipeline
- https://github.com/tenstorrent/tt-metal/actions/runs/31728596960/job/94545754267
- https://github.com/tenstorrent/tt-metal/actions/runs/32011689997/job/95337710167

### Failure Signature Snippet
```
##[error]The action 'ttnn nightly data_movement tests [wh_n150_civ2]' has timed out after 40 minutes.
```
pytest ran continuously for ~39-40 minutes with no hang/deadlock; the last test started with only milliseconds left on the clock. This is a pure runtime-growth problem, not a flaky/ND failure.

### Additional info
Root cause: the `tests/ttnn/nightly/unit_tests/operations/data_movement` suite grew heavier over time and several test files carry a large amount of redundant combinatorial parametrization (crossing axes like `dtype`/`layout`/`orientation`/`pad_value` fully against the primary `shape`/`dim` sweep, even when those secondary axes don't interact with shape/dim handling).

Re-profiling a failing run (correcting for CI log lines that don't carry a per-test timestamp, which had previously caused runtime to be silently misattributed to neighboring files) identified the two biggest contributors:
- `test_slice_for_conv.py`: ~505-520s (~8.6 min, roughly a third of the 40-minute budget)
- `test_repeat_codegen_routing.py`: ~370s (auto-generated correctness gate, intentionally left untouched)

Being tracked/fixed in [PR #53109](https://github.com/tenstorrent/tt-metal/pull/53109), which trims redundant parametrization in `test_split.py`, `test_repeat_interleave.py`, and `test_slice_for_conv.py` (1359 -> 354 test cases across the three files) while keeping the job timeout unchanged at 40 minutes, to validate whether the trim alone is sufficient before considering a timeout/budget bump.

abdulla-TT:
### PR Category
Test Only

Fixes #53460

### Summary
The nightly `ttnn nightly data_movement tests` job (`tests/pipeline_reorg/ops_unit_tests.yaml`) has been timing out at its 40 minute budget, e.g. [run 31728596960](https://github.com/tenstorrent/tt-metal/actions/runs/31728596960/job/94545754267) on `wh_n150_civ2`, with no hang/deadlock — pytest ran continuously for ~39m47s and the last test started with just 3ms left on the clock.

Rather than raising the timeout/budget ceiling, this PR reduces the actual runtime by trimming combinatorial-explosion parametrization in the three heaviest test files in this suite:

- **`test_split.py::test_split`**: was fully crossing `dtype × layout` (4 combos) against the entire `shape × dim × chunksize` sweep (64 combos) = 256 test cases, even though dtype/layout don't interact with shape/dim/chunksize handling in this op. Now the full 64-case sweep runs once at a default `bfloat16`/`TILE_LAYOUT`, and a new `test_split_dtype_layout_coverage` spot-checks the other 3 dtype/layout combinations against one representative shape per rank (12 cases). **256 -> 76 test cases.**
- **`test_repeat_interleave.py`**: both functions are regression tests for #41631 (integer/fp32 dtypes must not round-trip through bf16) on a fixed `(1,1,32,32)` input, but were fully crossing `dim × repeats` against every dtype. That `dim × repeats` genericity is already swept in full (with `bfloat16`/`uint16`) by the post-commit `tests/ttnn/unit_tests/operations/data_movement/test_repeat_interleave.py` on every PR, so `dim`/`repeats` are trimmed to 2 representative values each here while all 4 regression dtypes are kept in full. **75 -> 20 test cases.**
- **`test_slice_for_conv.py`**: its 6 functions were each fully crossing `dims × slice_dim × layout` against `orientation`/`dtype`/`pad_value` (1028 test cases total). Each function now keeps a full `dims × slice_dim × layout` sweep (the axis that drives distinct shard/compile behavior) at one representative `dtype`/`orientation`/`pad_value`, plus a small coverage test that spot-checks the remaining `dtype`/`orientation`/`pad_value` values against a representative subset of shapes. No `(dims, slice_size, cores)` entry lost coverage — every shape is still exercised at least once in the primary sweep. **1028 -> 258 test cases.**

Combined: **1359 -> 354 test cases (-74%)** across the three files, with rank/axis/dtype/shape coverage that isn't duplicated elsewhere left fully intact.

### Notes for reviewers
- **The job timeout is deliberately left at the original 40 minutes** (not bumped, and `.github/time_budget.yaml` is untouched) so this PR can be validated as a proper CI run: if trimming these 1359 -> 354 test cases is enough to bring the job back under the existing 40-minute budget on its own, that confirms the fix without needing any budget/config changes at all.
- The `test_split`/`test_repeat_interleave` trim alone was **not sufficient**: a validation run on [4a2d155](https://github.com/tenstorrent/tt-metal/actions/runs/32011689997/job/95337710167) still timed out at 40 minutes. Re-profiling that run (accounting for CI log lines that don't carry a per-test timestamp, which had previously caused their runtime to be silently misattributed to neighboring files) showed `test_slice_for_conv.py` was actually the single biggest contributor at ~505-520s (~8.6 min, roughly a third of the whole budget) — bigger than `test_split.py` and `test_repeat_interleave.py` combined. That's why this PR now also trims `test_slice_for_conv.py`.
- One more heavy file, `test_repeat_codegen_routing.py` (~370s), was **intentionally left untouched**: it's auto-generated by a codegen pipeline with an explicit "edit the emitter, not this file" comment, and exhaustively checks that out-of-scope shapes correctly fall back to the native implementation instead of being mis-routed — a correctness gate, not incidental coverage bloat. Hand-trimming it risks silently weakening that check.
- These are intended to be behavior-preserving reductions, not coverage cuts: the shape/dims/dim/chunksize/slice_dim/layout sweeps (the axes that actually drive distinct code paths and compiled kernels) are kept in full for every op; only the secondary dtype/orientation/pad_value crossing that's redundant or covered elsewhere was de-crossed from the primary sweep.
- Verified with `py_compile`/AST-based case counting and no linter errors, but not run on hardware from this environment — please validate in CI/on device before merging. If the job is still too close to the 40-minute edge after this lands, a follow-up timeout bump can be layered on top with a much smaller margin needed.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/DM_ops_nightly_CI_timeout_fix)


abdulla-TT:
[Nightly tt-metal L2 tests](https://github.com/tenstorrent/tt-metal/actions/runs/31773610455) - now runs without timeout error.

.github/time_budget.yaml
```diff
@@ -92,8 +92,8 @@ ttnn:
     bh_galaxy: 40
     bh_p100: 805
     bh_p150b_civ2: 1045
-    wh_n150_civ2: 1205
-    wh_n300_civ2: 1180
+    wh_n150_civ2: 1215
```
Aswinmcw: Re-ran the `verify_time_budget.py` sums against main: ttnn `unit` is n150 1205/1205, n300 1180/1180, bh_p150b_civ2 **1035/1045**, bh_p100 760/805 - so bh_p150b_civ2 had exactly 10 min of slack and passes the +10 on its own, which contradicts the description's "previously failing for wh_n150_civ2/wh_n300_civ2/bh_p150b_civ2". Worth correcting, and worth calling out that this lands bh_p150b_civ2 at zero slack too, so it joins n150/n300 as a SKU where the next test added to `ops_unit_tests.yaml` fails the budget check outright.

Aswinmcw (approved):

.github/time_budget.yaml
roseli-TT: please change either the pr description or the numbers here to match

roseli-TT:
Please do some investigation (and document in pr description) of what tests were added to warrant this increase, and if any older tests can be deprecated to make room.

sjameelTT (approved):

Issue #51887, opened 2026-08-03
Explore Kimi-K3 Attention Residuals


nmilicevicTT:
K3 uses a specialized residuals strategy https://www.kimi.com/en/showcases/deep-research/attention-residuals-interactive-demo

mbezuljTT:
Let's PR functional version without CCL fusing; and continue with resolving CCL AR in the functional-followup

nmilicevicTT:
Closed by https://github.com/tenstorrent/tt-metal/pull/53114, https://github.com/tenstorrent/tt-metal/pull/53116, https://github.com/tenstorrent/tt-metal/pull/53318

PR #53114, opened 2026-08-14
[Kimi K3] AttnRes 1/4 — torch references for attention residuals

nmilicevicTT:
### Ticket

Part of Kimi K3 attention-residual bring-up (#51887).

### What's changed

The torch references for **AttnRes**, the residual scheme Kimi K3 uses in place of a plain
residual stream. No device code, no ttnn import, no CI job. This is the first of four PRs;
the device modules and the two ops land on top of it and are all scored against these files.

**What AttnRes changes.** A plain transformer hands each module the residual stream itself —
one running sum every module output is folded into. AttnRes cuts that sum into 12-layer blocks
and hands the module a **read** instead: a softmax mixture of the current block's partial sum
with one summed representation per earlier block, weighted by RMS-normalized scores against a
per-site learned query. Writing is untouched — a module's output is still added into the
partial sum with weight one. A 93-layer stack performs 186 reads.

**Ten files.**

```
models/demos/deepseek_v3_d_p/reference/kimi_k3/attn_res/
    hf_attn_res.py    HuggingFace's _apply_attn_res, vendored byte-identical
    hf_walk.py        the seal/read schedule, driven by that vendored read
    attn_res.py       the folded form the device mirrors; every device gate scores against it
    weights.py        checkpoint -> per-site queries
    __init__.py       what a read is, and why the three files are not one
    LICENSE-Kimi-K3
    tests/test_attn_res_reference.py
    tests/test_weights.py
models/demos/deepseek_v3_d_p/tests/attn_res/
    assertions.py       assert_accurate / assert_equal / assert_bit_identical
    test_assertions.py  those three, against inputs they must reject
```

### The question a reviewer will ask first: why three references?

`attn_res.py` is not the definition. It is the definition with two algebraic shortcuts already
taken, both of which the device depends on:

* **the folded query** — `res_norm.weight` and `res_proj.weight` only ever meet as a product
  against the candidate, so they collapse into one `[d]` vector;
* **the hoisted `rsqrt`** — the normalizer is a per-(token, candidate) scalar, so it comes out
  of the dot product;

and it is structured around a third, **the split read**: the sealed blocks are scored and mixed
separately from the live partial sum, then merged. That is what the fused op implements.

A device test cannot catch an error in any of the three, because it compares against the file
that already contains them. So each is checked here against a form that does not share it:

| shortcut | checked against | why nothing else reaches it |
|---|---|---|
| folded query | `hf_attn_res` | the vendored function applies the two weights as separate factors |
| split read | the one-shot read in `attn_res.py` | HuggingFace takes one softmax over the whole candidate set and exposes no seam |
| the schedule | `hf_walk.hf_stack` | which layers seal and which reads see how many candidates lives in the model's layer loop, not in any HuggingFace function |

`hf_attn_res.py` is the one read in this PR that nobody here wrote — vendored byte-identical
from `modeling_kimi_linear.py` (`moonshotai/Kimi-K3`, sha256 recorded in the file header,
`_apply_attn_res` at lines 1075-1088). It is the only evidence that the definition was read
correctly; every comparison between two files written on this side could agree and both be wrong.

The seal schedule is therefore written twice, `hf_walk.HfStream` and `attn_res.AttnResStream`,
and that is deliberate rather than a duplicate to clean up. The two differ in exactly two places:
which read they call, and whether the query arrives folded or as the two weights the checkpoint
stores. `seal`, `accumulate` and `num_sealed` are identical, so the pair does not gate the
schedule against itself — what it gates is the folded query at *every* site of a whole stack,
rather than at the single read the other tests cover. Collapsing them into one class parametrized
by a read function would leave `test_stack_matches_hf_walk` comparing a walk to itself.
`reference/kimi_k3/attn_res/__init__.py` says the same thing at the top of the package.

The schedule row of that table is the weakest of the three, and worth naming as such: both walks
were transcribed from the same reading of the model's layer loop, so a misreading agrees with
itself. What holds it is a written-down trace — 5 layers at block 2 seal after `[1,1,2,2,3]`
layers and issue reads over `[2,2,2,2,3,3,3,3,4,4]` candidates; the production 93 layers at block
12 seal 8 times and issue 186 reads, the last over 9 candidates. That catches a boundary that
later moves, not one read wrong to begin with, and it is what a reviewer checks against the model
source by eye.

`reference/kda/tests/` in this package is laid out the same way.

### Tests

```
pytest models/demos/deepseek_v3_d_p/reference/kimi_k3/attn_res/tests/ \
       models/demos/deepseek_v3_d_p/tests/attn_res/test_assertions.py
```

**48 passed in ~7s.** Pure torch — no device, no ttnn import, no checkpoint, no network. Measured
agreement between the folded/split forms and the vendored one is under `1.2e-6` at fp32 against a
`1e-5` tolerance.

Every new gate was mutation-tested rather than assumed: moving the seal boundary by one, dropping
the length guard, weakening the shape check to an element count, and swapping either the two read
sites or the two query factors each fail (3, 4, 2, 1 and 6 tests respectively). The site swap is
caught by exactly one test, which is why the fold's ordering is checked and not just its names.

No CI row is added. Nothing in `tests/pipeline_reorg/` sweeps host-only model tests, and booking a
LoudBox for 7 seconds of CPU arithmetic is not worth a job; these run in the AttnRes e2e job that
arrives with PR 4/4.

The 23 tests in `test_assertions.py` cover the assertion helpers themselves, which every later
device gate calls. The one worth reading is `test_assertions_reject_nonfinite_values`: PCC of two
NaN tensors is NaN and `NaN < threshold` is false, so without an explicit finiteness gate a NaN
read passes against a NaN oracle silently.

The 12 in `test_weights.py` cover the checkpoint-to-query fold, where a wrong answer is invisible
downstream: swap a layer's two read sites, or a query's two factors, and every later device gate
still compares against an oracle built from the same wrong queries. `validate_query_weights` now
checks shape rather than presence, because `fold_query` flattens — a transposed `res_proj.weight`
folded to the same numbers and passed. `[1, d]` is what the model's projection to a scalar score
stores, and what the vendored read's `proj.weight.squeeze(0)` requires.

### What lands on top

| PR | contents | depends on |
|---|---|---|
| **1/4** | this one — torch references | — |
| **2/4** | `ttnn.experimental.attn_res_weighted_reduce_nc` (batched weighted reduce over dim 1) | — |
| **3/4** | `ttnn.experimental.attn_res_gather_softmax` (fused gather + split read) | — |
| **4/4** | `tt/attn_res/` device modules, PCC gates, e2e CI row | 1/4, 2/4, 3/4 |

### Background for reviewers

Sheets **01 What it is**, **02 Inside a layer** and **03 Shapes** of the AttnRes workbook cover
everything this PR needs — what a read mixes, where the seals land, and the shapes involved:
https://claude.ai/code/artifact/e3ff58d2-a56a-40e7-8c0b-47ba68026527

### Checklist

- [x] New/existing tests provide coverage for changes — 48 host-only tests, all new
- [x] No device required; no CI job added

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nmilicevic/bringup/kimi-k3-attnres-0-reference)


mbezuljTT:
Thank you @nmilicevicTT could you please clarify license terms - I can see some files have TT copyright while some are not?

mbezuljTT:
Also, for the announced PRs I would want ops built and maintained by prefill team to stay under experimental/deepseek_prefill - we should rename that to experimental/prefill at some point.

nmilicevicTT:
> Thank you @nmilicevicTT could you please clarify license terms - I can see some files have TT copyright while some are not?

[The file](https://huggingface.co/moonshotai/Kimi-K3/blob/main/modeling_kimi_linear.py) I took a part of reference implementation contains this: 

```
# Licensing Information:
--
# - Code adapted from DeepSeek-V3 (DeepSeek-V3/modeling_deepseek.py) is licensed under the Apache License, Version 2.0.
# - Other parts of the code are licensed under the Kimi K3 License (see the LICENSE file in this repository).
```

And indeed there is a [Kimi license](https://huggingface.co/moonshotai/Kimi-K3/blob/main/LICENSE) in there which doesn't belong to any standard SPDX-License-Identifier like Apache-2.0 that we can reference. So we are obligated to place their license. The file `models/demos/deepseek_v3_d_p/reference/kimi_k3/attn_res/hf_attn_res.py` contains the licensed code, the rest are our files with TT copyright.

> Also, for the announced PRs I would want ops built and maintained by prefill team to stay under experimental/deepseek_prefill - we should rename that to experimental/prefill at some point.

Got it, will move ops there.

jjovicicTT (approved):

nbabinTT (approved):

PR #53124, opened 2026-08-14
[tt-train] DiT diffusion-model training example (first non-LLM example)

ayerofieiev-tt:
### Description
Adds a class-conditional **DiT (Diffusion Transformer)** training example to tt-train — the first non-LLM training workload on the stack. The model, DDPM pipeline, trainer, and sampler are implemented entirely in Python on `ttml`; CIFAR-10 pixel-space; scales from one Blackhole chip to a 32-chip Galaxy via DDP.

Measured: DiT-S (32.6M) at ~1,050–1,300 img/s on one chip (batch 64–256); patch-2 variant at ~2,380 img/s at global batch 2,048 on a 32-chip Galaxy (6.9× one chip). A tiny-model 5k-step run and a DiT-S 50k-step run (53 min) complete end-to-end with sample grids and checkpoints.

### Changes Made
- [x] **New Feature:** `tt-train/sources/examples/dit/` — model (`dit_ttml.py`, incl. a custom Python autograd `SliceLastDim` op), host-side diffusion utilities, config-driven trainer (weight EMA, cosine LR, DDP via `dp` mesh axis + `sync_gradients`), offline DDIM sampler with CFG, on-device primitive gate test, and a 1:1 PyTorch golden reference. Configs for tiny/S/patch-2/galaxy variants.

The example README documents the framework workarounds it needed, which double as a gap list: host-side patchify (no autograd permute/transpose), one-hot labels through a linear (`embedding_backward` tile-width constraint on ids), composite SDPA for non-causal attention (fused kernel treats `mask=None` as causal — consider exposing `AttentionMaskType::None`), and checkpoint reads via `PreferredPrecision.NATIVE` because the default `to_numpy` fp32 cache (#41657) silently freezes checkpoint/EMA reads at their first-read values.

### Testing
- [x] Manual testing conducted: `test_device_primitives.py` (9 on-device checks incl. an end-to-end train-step loss-decrease gate) green on Blackhole p150/Galaxy; `test_reference.py` (CPU golden: patchify round-trip, overfit, sampling) green; full training runs on 1/4/32 chips with matching loss trajectories; DDP checkpoint save/load verified.
- [ ] CI integration for the example (suggestions welcome — the primitive gate is designed to be cheap enough for post-commit)

### Review Checklist
- [x] No breaking changes introduced (new files only)
- [x] Code complies with project style guidelines (config-driven like existing examples)

### Additional Context
Perf notes from bringup: the training step is device-bound at batch ≥64 (~7–11% MFU, dispatch fully overlapped), so op fusion is the main lever; a fused adaLN-modulate kernel prototyped in tt-lang measured 1.5–2.6× vs the composed ops and is a candidate follow-up, as is flattening the DDP gradient all-reduce (~150 sequential CCL calls dominate multi-chip step time). Trace-replay of the full training step was validated (1.0× at batch 64, ~1.6× at batch 8).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:tt-train-dit-example)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:tt-train-dit-example)

### DIFF
--- a/tt-train/configs/training_configs/training_cifar10_dit_s_p2_galaxy.yaml
+++ b/tt-train/configs/training_configs/training_cifar10_dit_s_p2_galaxy.yaml
@@ -1,0 +1,39 @@
+training_config:
+  project_name: "tt_train_dit"
+  seed: 5489
+  batch_size: 2048
+  max_steps: 6250
+  model_save_interval: 1000
+  data_path: "data/cifar-10"
+  output_dir: "runs/dit"
+  model_config: "${TT_METAL_RUNTIME_ROOT}/tt-train/configs/model_configs/dit_s_p2.yaml"
+  lr_schedule: cosine
+  warmup_steps: 250
+  ema_decay: 0.9995
+  ema_interval: 25
+  optimizer:
+    type: AdamW
+    lr: 0.0008
+    beta1: 0.9
+    beta2: 0.999
+    epsilon: 1.0e-8
+    weight_decay: 0.0
+
+diffusion_config:
+  timesteps: 1000
+  beta_start: 0.0001
+  beta_end: 0.02
+  cfg_drop_prob: 0.1
+
+eval_config:
+  sample_interval: 0
+  sample_steps: 50
+  cfg_scale: 2.0
+
+logging_config:
+  log_interval: 100
+  wandb: false
+
+device_config:
+  enable_ddp: false
+  mesh_shape: [32, 1]
--- a/tt-train/sources/examples/dit/diffusion.py
+++ b/tt-train/sources/examples/dit/diffusion.py
@@ -1,0 +1,94 @@
+# SPDX-FileCopyrightText: (c) 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Host-side diffusion utilities (numpy) shared by the ttml training loop.
+
+Same math as the torch versions in reference_torch.py; kept torch-free so the
+device training loop only needs numpy on the host.
+"""
+
+from __future__ import annotations
+
+import numpy as np
+
+
+def patchify(images: np.ndarray, patch: int) -> np.ndarray:
+    """[B, C, H, W] -> [B, 1, T, patch*patch*C]."""
+    b, c, h, w = images.shape
+    gh, gw = h // patch, w // patch
+    x = images.reshape(b, c, gh, patch, gw, patch)
+    x = x.transpose(0, 2, 4, 3, 5, 1)
+    return x.reshape(b, 1, gh * gw, patch * patch * c)
+
+
+def unpatchify(tokens: np.ndarray, patch: int, channels: int, height: int, width: int) -> np.ndarray:
+    """[B, 1, T, patch*patch*C] -> [B, C, H, W]."""
+    b = tokens.shape[0]
+    gh, gw = height // patch, width // patch
+    x = tokens.reshape(b, gh, gw, patch, patch, channels)
+    x = x.transpose(0, 5, 1, 3, 2, 4)
+    return x.reshape(b, channels, height, width)
+
+
+def timestep_features(t: np.ndarray, dim: int, max_period: int = 10000) -> np.ndarray:
+    """[B] int -> [B, 1, 1, dim] fp32 sinusoidal features."""
+    half = dim // 2
+    freqs = np.exp(-np.log(max_period) * np.arange(half, dtype=np.float32) / half)
+    args = t.astype(np.float32)[:, None] * freqs[None]
+    emb = np.concatenate([np.cos(args), np.sin(args)], axis=-1).astype(np.float32)
+    return emb.reshape(t.shape[0], 1, 1, dim)
+
+
+class DiffusionSchedule:
+    """Linear-beta DDPM schedule, fp32."""
+
+    def __init__(self, timesteps: int = 1000, beta_start: float = 1e-4, beta_end: float = 2e-2):
+        self.timesteps = timesteps
+        self.betas = np.linspace(beta_start, beta_end, timesteps, dtype=np.float32)
+        self.alphas_bar = np.cumprod(1.0 - self.betas, axis=0)
+        self.sqrt_alphas_bar = np.sqrt(self.alphas_bar)
+        self.sqrt_one_minus_alphas_bar = np.sqrt(1.0 - self.alphas_bar)
+
+    def add_noise(self, x0: np.ndarray, noise: np.ndarray, t: np.ndarray) -> np.ndarray:
+        a = self.sqrt_alphas_bar[t].reshape(-1, 1, 1, 1)
+        s = self.sqrt_one_minus_alphas_bar[t].reshape(-1, 1, 1, 1)
+        return (a * x0 + s * noise).astype(np.float32)
+
+
+def make_training_batch(
+    images: np.ndarray,  # [B, C, H, W] fp32 in [-1, 1]
+    labels: np.ndarray,  # [B] int
+    schedule: DiffusionSchedule,
+    patch: int,
+    model_dim: int,
+    rng: np.random.Generator,
+    cfg_drop_prob: float = 0.1,
+    null_class: int | None = None,
+):
+    """Returns (tokens, t_feats, labels_onehot, target) as numpy arrays ready
+    for Tensor.from_numpy: [B,1,T,in], [B,1,1,dim], [B,1,1,null_class+1] fp32
+    one-hot, [B,1,T,in]. One-hot (not ids): the model embeds labels via
+    one-hot @ W because ttnn embedding_backward can't take a single id."""
+    b = images.shape[0]
+    t = rng.integers(0, schedule.timesteps, size=b)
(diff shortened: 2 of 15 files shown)


tt-train/sources/examples/dit/train_dit_cifar.py
```diff
@@ -0,0 +1,324 @@
...
+def _param_to_numpy(t, composer=None, num_devices: int = 1):
+    # precision=NATIVE reads the live bf16 buffer directly. The default
+    # precision=FULL runs a device-side typecast whose fp32 view is CACHED and
+    # never refreshed (tt-train issue #41657) — with it, every checkpoint/EMA
+    # read after the first silently returns stale (init-time) weights.
+    arr = t.to_numpy(None, composer, ttml.autograd.PreferredPrecision.NATIVE).astype(np.float32)
+    if composer is not None and num_devices > 1:
+        # Replicated param read via concat composer stacks the per-device
+        # copies along dim 0; keep the first replica.
+        arr = arr[: arr.shape[0] // num_devices]
```
ayerofieiev-tt, line 88: **Replica extraction is wrong for tp>1 and wasteful on the galaxy config.** `concat_mesh_to_tensor_composer` concatenates *all* mesh devices along dim 0, but this divides by `num_devices=dp_size`. The mesh is declared `ttml.Mesh(mesh_shape, ("dp","tp"))`, so a legal `mesh_shape: [2,2]` yields 4 stacked copies sliced to 2 -> every checkpoint/EMA entry has a doubled leading dim, and `sample_from_ckpt.load_weights` then crashes or loads garbage. Even on the shipped `[32,1]` config this pulls 32 full copies (~4GB for the ~130MB state) over ethernet per EMA update/checkpoint and discards 31.

`ttml.sharding.Sharding.from_tensor(t).gather(t)` (already on this branch) reads exactly one replica via a size-1 mesh override, follows the live placement (safe if a param ever becomes dim-0-sharded), and applies the `PreferredPrecision.NATIVE` workaround internally — this whole helper can collapse to it.

tt-train/sources/examples/dit/diffusion.py
```diff
@@ -0,0 +1,94 @@
...
+    noisy = schedule.add_noise(images, noise, t)
+
+    if null_class is not None and cfg_drop_prob > 0:
+        drop = rng.random(b) < cfg_drop_prob
+        labels = np.where(drop, null_class, labels)
+
+    tokens = patchify(noisy, patch)
+    target = patchify(noise, patch)
+    t_feats = timestep_features(t, model_dim)
+    labels_onehot = one_hot(labels, (null_class if null_class is not None else int(labels.max())) + 1)
```
ayerofieiev-tt, line 85: **Batch-dependent one-hot width when `null_class=None`.** The fallback `int(labels.max())+1` makes the width depend on batch contents: a batch whose max label is 7 produces `[B,1,1,8]`, the next batch `[B,1,1,10]`, while `DiT.label_emb` is `LinearLayer(num_classes+1, dim)` — shape mismatch crash (or a per-batch reinterpretation of table columns) the moment someone trains unconditionally, which is this parameter's documented meaning. Pass the width in explicitly instead of inferring it.

tt-train/sources/examples/dit/sample_from_ckpt.py
```diff
@@ -0,0 +1,78 @@
...
+import ttnn
+
+import ttml
+from ttml.common.config import load_config
+
+from diffusion import DiffusionSchedule
+from train_dit_cifar import build_model, sample_grid
+
+
+def load_weights(model, npz_path: str):
```
ayerofieiev-tt, line 29: **Consider `ttml.checkpointing` instead of the bespoke npz format.** `ttml/ttml/checkpointing.py` (present on this branch) already provides `save_checkpoint`/`load_checkpoint` with per-tensor streaming, atomic tmp+`os.replace` writes (npz here can corrupt the newest checkpoint on a crash mid-write), missing/unexpected-param validation, `Sharding`-based replica gathering (subsuming the NATIVE-precision workaround), resharding onto any mesh on load, and optimizer-state round-trip for resume. Only the EMA numpy state genuinely needs a custom container.

Same applies to `cosine_lr()` in the trainer: `ttml/common/schedulers.py` has `SequentialScheduler(LinearScheduler, CosineAnnealingScheduler)` with state save/restore — the stateless `opt.set_lr(cosine_lr(...))` loses LR state on any future resume support.

tt-train/configs/training_configs/training_cifar10_dit_s_p2_galaxy.yaml
```diff
@@ -0,0 +1,39 @@
...
+  sample_interval: 0
+  sample_steps: 50
+  cfg_scale: 2.0
+
+logging_config:
+  log_interval: 100
+  wandb: false
+
+device_config:
+  enable_ddp: false
```
ayerofieiev-tt, line 38: **`enable_ddp` is a dead key — and it's `false` in the 32-chip DDP config.** `train_dit_cifar.py` derives DDP purely from `mesh_shape[0]>1` and never reads this, so the reference galaxy config ships `enable_ddp: false` while running DDP, and setting it `true` on a `[1,1]` mesh is silently ignored. Either honor it via `ttml.common.utils.build_mesh(device_config)` (the existing facility that consumes `enable_ddp`) or drop the key from all four YAMLs.

tt-train/sources/examples/dit/train_dit_cifar.py
```diff
@@ -0,0 +1,324 @@
...
+
+        loss_val = float(loss.to_numpy(ttnn.DataType.FLOAT32, composer=loss_composer).mean()) if loss_composer \
+            else float(loss.to_numpy(ttnn.DataType.FLOAT32).reshape(-1)[0])
+        if ema is not None:
+            ema.maybe_update(step)
+        ctx.reset_graph()
+        ema_loss = loss_val if ema_loss is None else 0.98 * ema_loss + 0.02 * loss_val
+
+        if step % log_every == 0:
+            ips = batch_size * log_every / (time.time() - t0)
```
ayerofieiev-tt, line 288: **img/s window absorbs sampling + checkpoint time.** `t0` resets at the log line, but `sample_grid`/`save_checkpoint` run *after* it in the same iteration, and `sample_interval` is a multiple of `log_interval` in every shipped config — so the next window's `ips` includes a 50-step DDIM sample (100 CFG forwards) plus two checkpoint saves, e.g. reporting ~3 img/s instead of ~300 and writing periodic bogus dips into the wandb `images_per_sec` curve. Reset `t0` after the sampling/checkpoint block (or time those separately).

ayerofieiev-tt (commented):
Self-review notes (Claude-assisted pass over the diff). Inline comments below cover the actionable items; softer golden-reference parity caveats worth a README line rather than code changes:

- **GELU variant**: `reference_torch.py` uses `nn.GELU(approximate="tanh")` but `ttml.ops.unary.gelu` is exact erf GELU (`ttnn::gelu` accurate fwd; bwd hardcodes `approx_mode="none"` in `ops/unary_ops.cpp`). Systematic ~3e-3 offset, compounds over 12 blocks — switch the reference to `nn.GELU()` or note the divergence.
- **LN affine + init distributions**: ttml `LayerNorm` registers learnable gamma/beta vs the reference's `elementwise_affine=False` (grads flow from step 1, param sets differ by name), and non-zero-init layers use `normal(0,0.02)` vs torch `nn.Linear` kaiming defaults / `nn.Embedding` N(0,1) / `trunc_normal_` pos_emb — from-scratch loss curves diverge for init reasons unrelated to device bugs.
- **Baked `(1+scale)` bias init**: with `weight_decay>0`, ttml AdamW decays *every* param (`adamw_composite.cpp:191-196`), pulling the effective multiplier toward 0 (torch parameterization decays toward 1); also makes ttml checkpoints compute `2+Wx` if loaded into the torch reference. Shipped configs use wd=0.0, so this is a trap, not a live bug — deserves a warning next to the wd knob.
- `Attention._ones_masks` (unreachable path today): class-level device-tensor cache would dangle across mesh close/reopen; the real fix is plumbing `AttentionMaskType::None` through `ttml::ops::scaled_dot_product_attention` (enum already exists in `ttml/metal/common/const_utils.hpp`).
- Minor: `Modulation`/`broadcast_rows`/`SliceLastDim` are unused on the ttml side; `--max-steps 0` falls through `args.max_steps or ...`; `loss.to_numpy` forces a device sync every step though it's logged every `log_interval`.

PR #53129, opened 2026-08-14
Track mesh buffers in Inspector

onenezicTT:
This PR enables Inspector to track `MeshBuffer`s and their lifecycles which will be used in future debug features (prerequisite for https://github.com/tenstorrent/tt-metal/pull/52841).

Adds `~50-70ns` to `MeshBuffer` allocate and deallocate paths. Closes https://github.com/tenstorrent/tt-metal/pull/53129

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:onenezic/inspector-buffer-tracking)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:onenezic/inspector-buffer-tracking)


tt_metal/distributed/mesh_buffer.cpp
```diff
@@ -101,6 +102,7 @@ std::shared_ptr<MeshBuffer> MeshBuffer::create(
         auto mesh_buffer =
             std::shared_ptr<MeshBuffer>(new MeshBuffer(mesh_buffer_config, device_local_config, 0, 0, mesh_device));
         mesh_buffer->initialize_device_buffers();
+        Inspector::mesh_buffer_created(mesh_buffer.get());
```
tt-vjovanovic: Why do we have `Inspector::mesh_buffer_created()` call here?
I would say that having ones in constructor is enough. Plant events in constructor and destructor instead of other wrapper functions.

tt_metal/distributed/mesh_buffer.cpp
```diff
@@ -258,6 +265,7 @@ MeshBuffer& MeshBuffer::operator=(MeshBuffer&& other) noexcept {
         other.state_ = DeallocatedState{};
         other.address_ = 0;
         other.device_local_size_ = 0;
+        Inspector::mesh_buffer_created(this);
```
tt-vjovanovic: Hm... This one is a bit mid ground. What do we want to track in Inspector? Buffers that are allocated, or all objects that exist? If it is allocated buffers, then attach to allocate/deallocate. You cannot call `operator=` on object that doesn't exist. Move constructor/operator= are good to cover if you want to track elements that are allocated in Inspector.

tt_metal/impl/debug/inspector/types.hpp
```diff
@@ -21,6 +21,7 @@ namespace tt::tt_metal {
     class MetalContext;
 
     namespace distributed {
+    class MeshBuffer;
```
tt-vjovanovic: Is this needed?

tt-vjovanovic (commented):
Are we planing to expose buffers with this PR or next one?

tt_metal/impl/debug/inspector/types.hpp
onenezicTT: Its probably a leftover from somewhere

tt_metal/distributed/mesh_buffer.cpp
onenezicTT: I was worried about leaving it the constructor as its private and only called by `MeshBuffer::create` for one and all the device initialization stuff happens after so there might be a chance we catch on rpc read a buffer that's not initialized and doesn't have stuff in it. 

Maybe it makes sense what you mentioned as realistically that race could not happen I think.

tt_metal/distributed/mesh_buffer.cpp
onenezicTT: I was thinking all buffer objects that are alive, then we can easily use those to figure out which ones are allocated and which ones are not. Maybe it makes sense to hook onto allocate/deallocate as well for the purpose of logging them.

onenezicTT: `You cannot call operator= on object that doesn't exist.` what do you mean by this in this context?

onenezicTT: On the question of exposing this info via rpc, I haven't really figured how exactly could that look like yet so I purposely left that as a follow up and this part only in the PR to unblock the socket dump PR.

tt_metal/impl/debug/inspector/types.hpp
onenezicTT: Removed

tt_metal/distributed/mesh_buffer.cpp
onenezicTT: Aligned offline on tracking allocated buffers rather than all buffer objects.

tt_metal/distributed/mesh_buffer.cpp
onenezicTT: Aligned offline to keep it here

tt_metal/impl/debug/inspector/types.hpp
```diff
@@ -23,7 +23,7 @@ namespace tt::tt_metal {
     namespace distributed {
     class MeshDeviceImpl;
     class MeshWorkloadImpl;
-    }
+    }  // namespace distributed
```
onenezicTT:
```suggestion
    }
```

tt_metal/distributed/mesh_buffer.cpp
```diff
@@ -239,9 +242,11 @@ MeshBuffer::MeshBuffer(MeshBuffer&& other) noexcept :
     device_local_size_(other.device_local_size_),
     buffers_(std::move(other.buffers_)),
     state_(std::move(other.state_)) {
+    Inspector::mesh_buffer_deallocated(&other);
```
tt-vjovanovic: Small note, this could potentially represent race condition between Inspector read and this move constructor as you first move values and then notify Inspector about deallocated buffer. It should go other way around.

tt_metal/distributed/mesh_buffer.cpp
```diff
@@ -255,14 +260,22 @@ MeshBuffer& MeshBuffer::operator=(MeshBuffer&& other) noexcept {
         buffers_ = std::move(other.buffers_);
         state_ = std::move(other.state_);
 
+        Inspector::mesh_buffer_deallocated(&other);
```
tt-vjovanovic: Same as in move constructor, you should first notify Inspector to deallocate buffer (remove it from its list) and then do the update.

tt_metal/impl/debug/inspector/inspector.cpp
```diff
@@ -295,6 +295,45 @@ void Inspector::mesh_device_initialized(const distributed::MeshDeviceImpl* mesh_
...
+    if (!data) {
+        // Inspector failed to initialize, no need to print failure message again.
+        return;
+    }
+    try {
+        std::lock_guard<std::mutex> lock(data->mesh_buffers_mutex);
+        data->mesh_buffers_data.insert(mesh_buffer);
+        if (data->mesh_buffer_logging_enabled) {
+            data->logger.log_mesh_buffer_allocated(mesh_buffer);
+        }
```
tt-vjovanovic, line 310: Do this outside the lock scope. You can do it before locking.

tt_metal/impl/debug/inspector/inspector.cpp
```diff
@@ -295,6 +295,45 @@ void Inspector::mesh_device_initialized(const distributed::MeshDeviceImpl* mesh_
...
+    auto* data = get_inspector_data();
+    if (!data) {
+        return;
+    }
+    try {
+        std::lock_guard<std::mutex> lock(data->mesh_buffers_mutex);
+        data->mesh_buffers_data.erase(mesh_buffer);
+        if (data->mesh_buffer_logging_enabled) {
+            data->logger.log_mesh_buffer_deallocated(mesh_buffer);
+        }
```
tt-vjovanovic, line 329: Do this outside the lock scope. You can do it before locking.

tt-vjovanovic (approved):

jbaumanTT (approved):

PR #53162, opened 2026-08-14
[Bug fix] emule: converge the runner with the emule-blaze fork

arminaleTT:
### PR Category

Bug fix

### Summary

Brings `main`'s copy of the emule runner up to what the emule fork runs. Nothing builds this
code here — `TT_METAL_USE_EMULE` defaults to `OFF` and there is no emule job — so it drifts
from the fork, where it is the only place it gets compiled and tested, and every fork rebase
has to reconcile the difference by hand. After this, the two hold identical emule sources.

All of it sits in `TT_METAL_USE_EMULE`-gated code or an emule-only branch of a mock predicate,
so a default build is unaffected.

**A dataflow buffer on Wormhole/Blackhole is a circular buffer.** Tile-counter setup is gated
on `hal().has_tile_counter_registers()`, the predicate `finalize_dataflow_buffer_configs()`
already splits on. Those arches have no tile counters, so the previous code applied a Quasar
limit — 8 DFBs per core — to hardware without the feature, and `ttnn::batch_norm` (9 DFBs since
its Metal 2.0 port) could not be allocated. Also closes a latent OOB: Blackhole allows 64 DFB
slots against a `MAX_DFBS` of 32.

**A resolved NOC offset is bounded by the target core's size.** `Core::l1_ptr` already
bounds-checks and takes `uint64_t` so a GB-scale DRAM bank needs no truncation; the resolvers
now pass the offset through unmodified and treat one the target cannot hold as a resolve miss.
Previously a worker offset was masked into a 2 MB slot and a DRAM offset truncated to 32 bits.
That mattered because UMD hands kernels `device_io_addr = pcie_base + arena_offset`, which
carries no NOC coordinates, so an unmapped host-window address decodes to `(0, 0)` — where
Wormhole's soc descriptor has a DRAM core — and the truncation pulled a >32 GB offset into
range and returned a valid pointer instead of `nullptr`. Applies to
`__emule_resolve_noc_addr` and `__emule_fabric_resolve_remote`;
`__emule_multicast_write` likewise passes its offset through, so `l1_ptr` surfaces a bad one.

**Wormhole on-chip NOC addresses resolve.** Mapped-buffer registry membership is the
host-facing discriminator, with `noc_addr >= pcie_base` kept only as a pre-filter. The bare
threshold only works when `pcie_base` sits above the NOC-encoded range; on Wormhole it is
`0x8'0000'0000`, *below* the bit-36 coordinate field, so every on-chip address took the host
path, missed the registry and returned `nullptr`. Callers guard with `if (ptr)`, so transfers
were silently skipped and kernels read stale L1 — wrong data, no crash, on every Wormhole test
moving data over the NOC.

**DFB entry metadata and `PROCESSOR_INDEX`.** `build_kernel_defines` walks
`dataflow_buffers_on_core` and derives tile metadata from `data_format` + `tile`, skipping
`DataFormat::Invalid` slots, matching `ProgramImpl::set_dfb_data_fmt_and_tile`; `entry_size` is
the NOC-facing stride and stays in the sync state. Without the walk a program declaring DFBs
instead of CBs (`ttnn::typecast`) reported page size 0 to its kernels. `PROCESSOR_INDEX` backs
`get_hw_thread_idx()`, which the debug headers need to compile. Two silent `EMULE_NUM_CBS`
skips are `TT_FATAL`s.

**`PinnedMemory`'s no-mapping path is `Mock`-only.** `SWEmuleChip` has a real
`SimulationSysmemManager`, so dummy addresses stranded the host-facing socket counters an
emulated kernel writes back. Same narrowing already at `distributed.cpp:121-125`.

Four commits are comment or whitespace only.

### Notes for reviewers

**tt-metal CI cannot exercise any of this** — it builds with `TT_METAL_USE_EMULE=OFF`, so these
files are not compiled here and a break would merge silently. The evidence is emule-side, on a
build of this branch's exact file content:

| Check | Result |
|---|---|
| Blackhole ttnn PR gate set (48 entries incl. typecast, batch_norm, topk) | 48 pass, 0 fail |
| C++ regression: wormhole / quasar / blackhole | 39/0, 73/0, 20/0 |
| `Core::l1_ptr` OOB aborts across every log | none |
| Compile + link, `TT_METAL_USE_EMULE=ON` | clean |
| tt-emule-blaze CI at this content (17 checks: both builds, p150 + loudbox blaze gates, TTNN smoke, bh_galaxy) | green |

The fork carries byte-identical files and its divergence from `main` is patch-id-identical to
this diff, so those results apply directly here. Companion PR:
tenstorrent/tt-emule-blaze#315

Worth a look:

- The zero OOB-abort count is the load-bearing evidence for the bounds change. Removing the
  mask and the truncation means a previously-silent wrap now trips `Core::l1_ptr`'s existing
  loud check, and nothing in ttnn or three arches of C++ tripped it.
- Per-CB face geometry (#52742) now reaches emule's `CBSyncState`, which `pack_untilize` reads
  — the fork had never populated it, so it ran the full-tile 16/4 default throughout.
- `main`'s removal of the pass-2 global CB fill (#52726) is adopted; all three `topk` entries,
  the path it existed for, pass.
- Checking whether an address maps to a PCIe core is not possible against what UMD supplies,
  since `device_io_addr` carries no coordinates. Encoding the PCIe tile's coordinates, or
  giving `SimulationSysmemManager` a simulation-only `pcie_base` above the NOC range (which
  would also fix Quasar's `default: return 0`), are tt-umd changes and are not here.

### DIFF
--- a/tt_metal/distributed/pinned_memory.cpp
+++ b/tt_metal/distributed/pinned_memory.cpp
@@ -96,10 +96,14 @@ void PinnedMemoryImpl::initialize_from_devices(
         unique_mmio_devices.insert(mmio_device_id);
     }
 
-    // Mock/emulated devices have no SysmemManager, so there is no host-memory
-    // mapping to perform. Record the device map and keep the host pointer only;
-    // device/NOC address queries return dummies (see get_device_addr / get_noc_addr).
-    if (cluster.is_mock_or_emulated()) {
+    // MockChip has no SysmemManager, so there is no host-memory mapping to perform. Record the
+    // device map and keep the host pointer only; device/NOC address queries return dummies (see
+    // get_device_addr / get_noc_addr).
+    //
+    // Mock only, not is_mock_or_emulated(): SWEmuleChip has a real SimulationSysmemManager, so a
+    // dummy address would strand the host-facing socket counters an emulated kernel writes back.
+    // See tt-emule docs/socket-emulation.md §7.
+    if (cluster.get_target_device_type() == tt::TargetDevice::Mock) {
         is_mock_ = true;
         mock_host_ptr_ = host_buffer;
         host_offset_ = 0;
--- a/tt_metal/distributed/pinned_memory.cpp
+++ b/tt_metal/distributed/pinned_memory.cpp
@@ -96,13 +96,10 @@ void PinnedMemoryImpl::initialize_from_devices(
         unique_mmio_devices.insert(mmio_device_id);
     }
 
-    // MockChip has no SysmemManager, so there is no host-memory mapping to perform. Record the
-    // device map and keep the host pointer only; device/NOC address queries return dummies (see
-    // get_device_addr / get_noc_addr).
-    //
-    // Mock only, not is_mock_or_emulated(): SWEmuleChip has a real SimulationSysmemManager, so a
-    // dummy address would strand the host-facing socket counters an emulated kernel writes back.
-    // See tt-emule docs/socket-emulation.md §7.
+    // MockChip has no SysmemManager, so there is nothing to map: keep the host pointer and let
+    // device/NOC queries return dummies (see get_device_addr / get_noc_addr). Mock only, not
+    // is_mock_or_emulated() — SWEmuleChip has a real SimulationSysmemManager, and a dummy address
+    // would strand an emulated kernel's socket counters. See tt-emule docs/socket-emulation.md §7.
     if (cluster.get_target_device_type() == tt::TargetDevice::Mock) {
         is_mock_ = true;
         mock_host_ptr_ = host_buffer;
--- a/tt_metal/distributed/pinned_memory.cpp
+++ b/tt_metal/distributed/pinned_memory.cpp
@@ -99,7 +99,7 @@ void PinnedMemoryImpl::initialize_from_devices(
     // MockChip has no SysmemManager, so there is nothing to map: keep the host pointer and let
     // device/NOC queries return dummies (see get_device_addr / get_noc_addr). Mock only, not
     // is_mock_or_emulated() — SWEmuleChip has a real SimulationSysmemManager, and a dummy address
-    // would strand an emulated kernel's socket counters. See tt-emule docs/socket-emulation.md §7.
+    // would strand the host-facing socket counters an emulated kernel writes back.
     if (cluster.get_target_device_type() == tt::TargetDevice::Mock) {
         is_mock_ = true;
         mock_host_ptr_ = host_buffer;
--- a/tt_metal/impl/emulation/emulated_program_runner.cpp
+++ b/tt_metal/impl/emulation/emulated_program_runner.cpp
@@ -1529,6 +1529,23 @@ static std::map<std::string, std::string> build_kernel_defines(
                 }
             }
         }
+        // A dataflow buffer carries the same LLK-facing entry metadata as a circular buffer, keyed
+        // by the same device slot, so it feeds the same tables — without this a program declaring
+        // DFBs instead of CBs (ttnn::typecast) reports page size 0 to its kernels.
+        // See tt-emule docs/cb-dataformat.md.
+        for (const auto& dfb_impl : impl.dataflow_buffers_on_core(first_core)) {
+            const uint32_t slot = dfb_impl->device_slot;
+            if (slot >= EMULE_NUM_CBS) {  // no CB view above the ceiling, so no metadata either
+                continue;
+            }
+            const auto& dfb_cfg = dfb_impl->config;
+            tile_sizes[slot] = dfb_cfg.entry_size;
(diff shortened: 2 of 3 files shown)


tt_metal/impl/emulation/emulated_program_runner.cpp
```diff
@@ -1529,6 +1532,26 @@ static std::map<std::string, std::string> build_kernel_defines(
...
+            const uint32_t slot = dfb_impl->device_slot;
+            TT_FATAL(
+                slot < EMULE_NUM_CBS,
+                "DFB device slot {} exceeds the emulated CB ceiling ({}); the host assigns slots below the arch's "
+                "NUM_CIRCULAR_BUFFERS ({}).",
+                slot,
+                EMULE_NUM_CBS,
+                MetalContext::instance().hal().get_arch_num_circular_buffers());
+            const auto& dfb_cfg = dfb_impl->config;
+            tile_sizes[slot] = dfb_cfg.entry_size;
```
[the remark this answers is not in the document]
arminaleTT: Good catch — this was a real bug and I've fixed it.

You're right on both counts, and silicon confirms it: `ProgramImpl::set_dfb_data_fmt_and_tile` derives the descriptor from `data_format` + `tile` + `unpack_face_geometry` and never touches `entry_size`, and it explicitly skips `DataFormat::Invalid` slots rather than deriving a tile size from them. The CB pass a few lines above this one in `build_kernel_defines` was already doing it correctly via `Tile::get_tile_size(data_format)`; the DFB loop I added was the odd one out.

Now derived the same way as the CB pass, with the `Invalid` skip, and `entry_size` stays only in the runtime sync state via `init_cb_sync` — which is the split you suggested.

Verified at the fixed commit on the emule side (tt-metal CI can't exercise this, `TT_METAL_USE_EMULE=OFF`): blackhole ttnn PR-gate set 48/48, including `elt_test_typecast_int` 13 passed and `elt_test_typecast_sharded` 151 passed — typecast being the op whose `entry_size` is deliberately aligned, so exactly the case you flagged — plus `reduce_test_moe` 2 passed and `test_batch_norm_program_cache_and_default` 18 passed/18 xfailed.

sgholamiTT (approved):

cfjchu (approved):

tt_metal/distributed/pinned_memory.cpp
```diff
@@ -96,10 +96,11 @@ void PinnedMemoryImpl::initialize_from_devices(
         unique_mmio_devices.insert(mmio_device_id);
     }
 
-    // Mock/emulated devices have no SysmemManager, so there is no host-memory
-    // mapping to perform. Record the device map and keep the host pointer only;
-    // device/NOC address queries return dummies (see get_device_addr / get_noc_addr).
-    if (cluster.is_mock_or_emulated()) {
+    // MockChip has no SysmemManager, so there is nothing to map: keep the host pointer and let
+    // device/NOC queries return dummies (see get_device_addr / get_noc_addr). Mock only, not
+    // is_mock_or_emulated() — SWEmuleChip has a real SimulationSysmemManager, and a dummy address
```
jbaumanTT, line 101: Is emulated only for SWEmule, or are the other types of emulator we need to be careful about?

tt_metal/impl/emulation/emulated_program_runner.cpp
```diff
@@ -339,22 +339,25 @@ static uint64_t get_pcie_base_cached(uint32_t device_id) {
...
     emule_require_self(__func__);
 
-    // Host-facing (PCIe) address: SimulationSysmemManager's device_io_addr space starts at
-    // pcie_base_ and shares the same 64-bit range as a real on-chip NOC address, so this branch
-    // must run FIRST, before any noc_x/noc_y/local_addr decomposition below.
+    // pcie_base alone cannot tell host-facing from on-chip: on Wormhole it is 0x8'0000'0000,
+    // below the bit-36 coordinate field, so every on-chip address clears it. Registry
+    // membership is the discriminator; the threshold is only a pre-filter.
     uint32_t device_id = __emule_self->chip_id;
     if (noc_addr >= get_pcie_base_cached(device_id)) {
```
jbaumanTT, line 336: Shouldn't we be looking at whether the NOC address is mapped to a PCIe core?

tt_metal/distributed/pinned_memory.cpp
arminaleTT: It is just tt-emule, which uses the SWEmule chip

tt_metal/impl/emulation/emulated_program_runner.cpp
arminaleTT: Yes, but emule does not model PCIe cores, so this gets complicated. What is here is basically a hack, but I have opened an issue to fix it. https://github.com/tenstorrent/tt-emule-blaze/issues/317

Issue #20966, opened 2025-04-22
Cleanup Tasks Post TT-Mesh Merge


tt-asaigal:
Tracking all cleanup tasks to keep the codebase healthy once TT-Mesh is merged

- [ ] Remove Single Device APIs from TTNN: This allows us to perform the next task in the list
- [ ] Remove handling of single device in TTNN (if/else on `IDevice` type)
- [ ] Remove `set_mesh_device` and `get_mesh_device` APIs in `MeshDevice` and the `mesh_device` member in `Device`. This is an API violation
- [ ] Have the Device Profiler Natively Support `MeshDevice`
- [ ] Remove all instances of `dynamic_cast<Device*>` and `dynamic_cast<MeshDevice*>`
- [ ] Remove all instances of `push_work` and `enable_async`, since these are nops

fyi @ayerofieiev-tt

tt-asaigal:
Assigning @tt-dma, based on an offline discussion, for cleanup of single device APIs from TTNN. This is on the roadmap for his work related to cleaning up device init and teardown, which requires the removal of existing device APIs from TTNN and TT-Metal.

cc @pgkeller

tt-asaigal:
Most of this has been completed and we have more dedicated issues tracking the remaining items.

RE: Distributed Namespace Purge
Not sure if this is the right time to do this, given that we're trying to keep a stable API surface. Idea is on runtime team radar - to be discussed further.

PR #53199, opened 2026-08-14
[Cleanup] GlobalSemaphore and GlobalCircularBuffer creation made mesh-based

obacherikovTT:
### PR Category
Cleanup

### Summary

`MeshDevice` at this point is the only kind of device that can be created publicly. So, we're gradually updating all the APIs to take it as a parameter instead of ambiguous `IDevice`. In this PR:
- `MeshDevice` overloads are added for `GlobalSemaphore` and `GlobalCircularBuffer` constructor, `CreateGlobalSemaphore` and `CreateGlobalCircularBuffer` free functions, and the alternatives are deprecated.
- Python bindings are changed to only call `MeshDevice` overloads. Other devices don't exist in Python, see:
https://github.com/tenstorrent/tt-metal/blob/ac2e6ad488d969d2097d361d20c855e9df2455d8/ttnn/ttnn/device.py#L13

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:obacherikov-create-global)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:obacherikov-create-global)

### DIFF
--- a/tests/tt_metal/tt_metal/dispatch/sub_device_test_utils.hpp
+++ b/tests/tt_metal/tt_metal/dispatch/sub_device_test_utils.hpp
@@ -6,6 +6,7 @@
 
 #include <tt-metalium/host_api.hpp>
 #include <tt-metalium/global_semaphore.hpp>
+#include <tt-metalium/mesh_device.hpp>
 #include "impl/context/metal_context.hpp"
 #include "impl/kernels/kernel.hpp"
 #include "sub_device.hpp"
@@ -13,10 +14,10 @@
 namespace tt::tt_metal {
 
 inline std::tuple<Program, CoreCoord, GlobalSemaphore> create_single_sync_program(
-    IDevice* device, const SubDevice& sub_device) {
+    distributed::MeshDevice* device, const SubDevice& sub_device) {
     auto syncer_coord = sub_device.cores(HalProgrammableCoreType::TENSIX).ranges().at(0).start_coord;
     auto syncer_core = CoreRangeSet(CoreRange(syncer_coord, syncer_coord));
-    auto global_sem = CreateGlobalSemaphore(device, sub_device.cores(HalProgrammableCoreType::TENSIX), INVALID);
+    auto global_sem = GlobalSemaphore(*device, sub_device.cores(HalProgrammableCoreType::TENSIX), INVALID);
 
     Program syncer_program = CreateProgram();
     auto syncer_kernel = CreateKernel(
@@ -30,7 +31,7 @@ inline std::tuple<Program, CoreCoord, GlobalSemaphore> create_single_sync_progra
 }
 
 inline std::tuple<Program, Program, Program, GlobalSemaphore> create_basic_sync_program(
-    IDevice* device, const SubDevice& sub_device_1, const SubDevice& sub_device_2) {
+    distributed::MeshDevice* device, const SubDevice& sub_device_1, const SubDevice& sub_device_2) {
     auto waiter_coord = sub_device_2.cores(HalProgrammableCoreType::TENSIX).ranges().at(0).start_coord;
     auto waiter_core = CoreRangeSet(CoreRange(waiter_coord, waiter_coord));
     auto waiter_core_physical = device->worker_core_from_logical_core(waiter_coord);
@@ -39,7 +40,7 @@ inline std::tuple<Program, Program, Program, GlobalSemaphore> create_basic_sync_
     auto syncer_core = CoreRangeSet(CoreRange(syncer_coord, syncer_coord));
     auto syncer_core_physical = device->worker_core_from_logical_core(syncer_coord);
     auto all_cores = waiter_core.merge(incrementer_cores).merge(syncer_core);
-    auto global_sem = CreateGlobalSemaphore(device, all_cores, INVALID);
+    auto global_sem = GlobalSemaphore(*device, all_cores, INVALID);
 
     Program waiter_program = CreateProgram();
     auto waiter_kernel = CreateKernel(
@@ -77,7 +78,7 @@ inline std::tuple<Program, Program, Program, GlobalSemaphore> create_basic_sync_
 }
 
 inline std::tuple<Program, Program, Program, GlobalSemaphore> create_basic_eth_sync_program(
-    IDevice* device, const SubDevice& sub_device_1, const SubDevice& sub_device_2, DataMovementProcessor dm_processor) {
+    distributed::MeshDevice* device, const SubDevice& sub_device_1, const SubDevice& sub_device_2, DataMovementProcessor dm_processor) {
     auto waiter_coord = sub_device_2.cores(HalProgrammableCoreType::ACTIVE_ETH).ranges().at(0).start_coord;
     auto waiter_core = CoreRangeSet(CoreRange(waiter_coord, waiter_coord));
     auto tensix_waiter_coord = sub_device_2.cores(HalProgrammableCoreType::TENSIX).ranges().at(0).start_coord;
@@ -88,7 +89,7 @@ inline std::tuple<Program, Program, Program, GlobalSemaphore> create_basic_eth_s
     auto syncer_core = CoreRangeSet(CoreRange(syncer_coord, syncer_coord));
     auto syncer_core_physical = device->worker_core_from_logical_core(syncer_coord);
     auto all_cores = tensix_waiter_core.merge(incrementer_cores).merge(syncer_core);
-    auto global_sem = CreateGlobalSemaphore(device, all_cores, INVALID);
+    auto global_sem = GlobalSemaphore(*device, all_cores, INVALID);
 
     Program waiter_program = CreateProgram();
     auto waiter_kernel = CreateKernel(
--- a/tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
+++ b/tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include <optional>
+
 #include <tt-metalium/global_semaphore.hpp>
 
 namespace tt::tt_metal::experimental {
@@ -16,13 +18,23 @@ namespace tt::tt_metal::experimental {
  *
  * | Argument       | Description                                            | Type                                                      | Valid Range  | Required |
  * |----------------|--------------------------------------------------------|-----------------------------------------------------------|--------------|----------|
- * | device         | The device to create the semaphore on                  | IDevice*                                                  |              | Yes      |
+ * | device         | The mesh device to create the semaphore on             | distributed::MeshDevice&                                  |              | Yes      |
  * | cores          | Range of the Tensix coordinates using the semaphore    | const CoreRangeSet &                                      |              | Yes      |
  * | initial_value  | Initial value of the semaphore                         | uint32_t                                                  |              | Yes      |
  * | buffer_type    | Buffer type to store the semaphore                     | BufferType                                                | L1 types     | No       |
  * | address        | Address of the semaphore to create                     | uint64_t                                                  |              | Yes      |
  */
 // clang-format on
+GlobalSemaphore CreateGlobalSemaphore(
+    distributed::MeshDevice& device,
+    const CoreRangeSet& cores,
+    std::optional<uint32_t> initial_value,
+    BufferType buffer_type,
+    uint64_t address);
+
+[[deprecated(
+    "Use CreateGlobalSemaphore(distributed::MeshDevice&, ...) instead. "
+    "CreateGlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
 GlobalSemaphore CreateGlobalSemaphore(
     IDevice* device,
     const CoreRangeSet& cores,
--- a/tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
+++ b/tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
@@ -34,7 +34,7 @@ GlobalSemaphore CreateGlobalSemaphore(
 
 [[deprecated(
     "Use CreateGlobalSemaphore(distributed::MeshDevice&, ...) instead. "
-    "CreateGlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
+    "CreateGlobalSemaphore(IDevice*, ...) will be removed after 2026-09-20.")]]
 GlobalSemaphore CreateGlobalSemaphore(
     IDevice* device,
     const CoreRangeSet& cores,
--- a/tt_metal/api/tt-metalium/global_circular_buffer.hpp
+++ b/tt_metal/api/tt-metalium/global_circular_buffer.hpp
@@ -33,6 +33,24 @@ struct GlobalCircularBufferDramSenderInternals;
 
 class GlobalCircularBuffer {
 public:
+    /**
+     * @brief Allocates a global circular buffer in L1 on the device.
+     *
+     * @param device Mesh device to create the global circular buffer on.
+     * @param sender_receiver_core_mapping The mapping of remote sender to remote receiver cores for the circular
+     * buffer.
+     * @param size Size of the global circular buffer per core in bytes.
+     * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.
+     */
+    GlobalCircularBuffer(
+        distributed::MeshDevice& device,
+        const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
+        uint32_t size,
+        BufferType buffer_type = BufferType::L1);
+
+    [[deprecated(
+        "Use GlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
+        "GlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
     GlobalCircularBuffer(
         IDevice* device,
         const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
@@ -126,6 +144,9 @@ private:
  * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.\
  * @return The allocated global circular buffer.
  */
+[[deprecated(
+    "Use GlobalCircularBuffer(distributed::MeshDevice&, ...) constructor instead. "
+    "CreateGlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
 GlobalCircularBuffer CreateGlobalCircularBuffer(
     IDevice* device,
     const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
--- a/tt_metal/api/tt-metalium/global_circular_buffer.hpp
+++ b/tt_metal/api/tt-metalium/global_circular_buffer.hpp
@@ -138,14 +138,29 @@ private:
 /**
  * @brief Allocates a global circular buffer in L1 on the device.
  *
+ * @param device Mesh device to create the global circular buffer on.
+ * @param sender_receiver_core_mapping The mapping of remote sender to remote receiver cores for the circular buffer.
+ * @param size Size of the global circular buffer per core in bytes.
+ * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.
+ * @return The allocated global circular buffer.
+ */
+GlobalCircularBuffer CreateGlobalCircularBuffer(
+    distributed::MeshDevice& device,
+    const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
+    uint32_t size,
+    BufferType buffer_type = BufferType::L1);
+
+/**
+ * @brief Allocates a global circular buffer in L1 on the device.
+ *
  * @param device The device to create the global circular buffer on.
  * @param sender_receiver_core_mapping The mapping of remote sender to remote receiver cores for the circular buffer.
  * @param size Size of the global circular buffer per core in bytes.
- * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.\
+ * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.
  * @return The allocated global circular buffer.
  */
 [[deprecated(
-    "Use GlobalCircularBuffer(distributed::MeshDevice&, ...) constructor instead. "
+    "Use CreateGlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
     "CreateGlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
 GlobalCircularBuffer CreateGlobalCircularBuffer(
     IDevice* device,
--- a/tt_metal/api/tt-metalium/global_circular_buffer.hpp
+++ b/tt_metal/api/tt-metalium/global_circular_buffer.hpp
@@ -50,7 +50,7 @@ public:
 
     [[deprecated(
         "Use GlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
-        "GlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
+        "GlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-20.")]]
     GlobalCircularBuffer(
         IDevice* device,
         const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
@@ -161,7 +161,7 @@ GlobalCircularBuffer CreateGlobalCircularBuffer(
  */
 [[deprecated(
     "Use CreateGlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
-    "CreateGlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
+    "CreateGlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-20.")]]
 GlobalCircularBuffer CreateGlobalCircularBuffer(
     IDevice* device,
     const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
--- a/tt_metal/api/tt-metalium/global_semaphore.hpp
+++ b/tt_metal/api/tt-metalium/global_semaphore.hpp
@@ -18,15 +18,38 @@ namespace tt::tt_metal {
 class IDevice;
 class GlobalSemaphore;
 class GlobalSemaphoreImpl;
+namespace distributed {
+class MeshDevice;
+}  // namespace distributed
 }  // namespace tt::tt_metal
 
 namespace tt::tt_metal {
 
 class GlobalSemaphore {
 public:
+    /**
+     * @brief Allocates a global semaphore in L1 on the mesh device.
+     *
+     * @param device Mesh device to create the semaphore on.
+     * @param cores Range of Tensix coordinates using the semaphore.
+     * @param initial_value Initial value of the semaphore.
+     * @param buffer_type Buffer type to store the semaphore. Can only be an L1 buffer type.
+     */
+    GlobalSemaphore(
+        distributed::MeshDevice& device,
+        CoreRangeSet cores,
+        uint32_t initial_value,
+        BufferType buffer_type = BufferType::L1);
+
+    [[deprecated(
+        "Use GlobalSemaphore(distributed::MeshDevice&, ...) instead. "
+        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
     GlobalSemaphore(
         IDevice* device, const CoreRangeSet& cores, uint32_t initial_value, BufferType buffer_type = BufferType::L1);
 
+    [[deprecated(
+        "Use GlobalSemaphore(distributed::MeshDevice&, ...) instead. "
+        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
     GlobalSemaphore(
         IDevice* device, CoreRangeSet&& cores, uint32_t initial_value, BufferType buffer_type = BufferType::L1);
 
--- a/tt_metal/api/tt-metalium/global_semaphore.hpp
+++ b/tt_metal/api/tt-metalium/global_semaphore.hpp
@@ -43,13 +43,13 @@ public:
 
     [[deprecated(
         "Use GlobalSemaphore(distributed::MeshDevice&, ...) instead. "
-        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
+        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-20.")]]
     GlobalSemaphore(
         IDevice* device, const CoreRangeSet& cores, uint32_t initial_value, BufferType buffer_type = BufferType::L1);
 
     [[deprecated(
         "Use GlobalSemaphore(distributed::MeshDevice&, ...) instead. "
-        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-17.")]]
+        "GlobalSemaphore(IDevice*, ...) will be removed after 2026-09-20.")]]
     GlobalSemaphore(
         IDevice* device, CoreRangeSet&& cores, uint32_t initial_value, BufferType buffer_type = BufferType::L1);
 
--- a/tt_metal/impl/buffers/global_circular_buffer.cpp
+++ b/tt_metal/impl/buffers/global_circular_buffer.cpp
@@ -101,6 +101,13 @@ void initialize_global_circular_buffer(
 }  // namespace
 
 GlobalCircularBuffer::GlobalCircularBuffer(
+    distributed::MeshDevice& device,
+    const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
+    uint32_t size,
+    BufferType buffer_type) :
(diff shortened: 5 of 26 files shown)


ttnn/api/ttnn/global_semaphore.hpp
```diff
@@ -17,31 +17,13 @@ struct MultiDeviceGlobalSemaphore {
     auto attribute_values() const { return std::forward_as_tuple(this->global_semaphores); }
 };
 
-// Single Device Creation API
-GlobalSemaphore create_global_semaphore(
-    IDevice* device, const CoreRangeSet& cores, uint32_t initial_value, BufferType buffer_type = BufferType::L1);
-
 // MeshDevice Creation API
```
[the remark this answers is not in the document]
obacherikovTT: I've heard that ttnn doesn't guarantee backward compatibility in C++. Please, let me know if I still need to follow the deprecation process here.

ttnn/api/ttnn/global_circular_buffer.hpp
```diff
@@ -13,13 +13,6 @@
...
 namespace ttnn::global_circular_buffer {
 
-// Single Device APIs
-GlobalCircularBuffer create_global_circular_buffer(
-    IDevice* device,
-    const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
-    uint32_t size,
-    BufferType buffer_type = BufferType::L1);
-
 // Multi Device APIs
```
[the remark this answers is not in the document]
obacherikovTT: I've heard that ttnn doesn't guarantee backward compatibility in C++. Please, let me know if I still need to follow the deprecation process here.

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
```diff
@@ -20,8 +20,7 @@ void py_module(nb::module_& mod) {
     // Single Device Creation API
     mod.def(
         "create_global_semaphore",
-        nb::overload_cast<IDevice*, const CoreRangeSet&, uint32_t, BufferType>(
-            &ttnn::global_semaphore::create_global_semaphore),
+        &ttnn::global_semaphore::create_global_semaphore,
```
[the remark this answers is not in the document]
obacherikovTT: I'm not sure if this is in scope of this PR. But I'm open to follow specific suggestions from code owners.

ttnn/cpp/ttnn-nanobind/global_circular_buffer.cpp
```diff
@@ -33,8 +33,7 @@ void py_module(nb::module_& mod) {
     // Single Device APIs
     mod.def(
         "create_global_circular_buffer",
-        nb::overload_cast<IDevice*, const std::vector<std::pair<CoreCoord, CoreRangeSet>>&, uint32_t, BufferType>(
-            &ttnn::global_circular_buffer::create_global_circular_buffer),
+        &ttnn::global_circular_buffer::create_global_circular_buffer,
```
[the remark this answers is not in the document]
obacherikovTT: I'm not sure if this is in scope of this PR. But I'm open to follow specific suggestions from code owners.

tt_metal/impl/buffers/global_circular_buffer.cpp
```diff
@@ -100,6 +100,13 @@ void initialize_global_circular_buffer(
 
 }  // namespace
 
+GlobalCircularBuffer::GlobalCircularBuffer(
+    distributed::MeshDevice& device,
+    const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
+    uint32_t size,
+    BufferType buffer_type) :
+    GlobalCircularBuffer{&device, sender_receiver_core_mapping, size, buffer_type} {}
```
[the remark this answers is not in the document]
obacherikovTT: It's simpler to implement this way, and later remove them together.

tt_metal/api/tt-metalium/global_semaphore.hpp
```diff
@@ -18,15 +18,38 @@ namespace tt::tt_metal {
...
 
 class GlobalSemaphore {
 public:
+    /**
+     * @brief Allocates a global semaphore in L1 on the mesh device.
+     *
+     * @param device Mesh device to create the semaphore on.
+     * @param cores Range of Tensix coordinates using the semaphore.
+     * @param initial_value Initial value of the semaphore.
+     * @param buffer_type Buffer type to store the semaphore. Can only be an L1 buffer type.
```
obacherikovTT, line 36: It's confusing why we have `buffer_type` at all if it can only be L1.
I've kept it to be consistent with the old API, but please let me know if I should remove it.

tt_metal/api/tt-metalium/global_circular_buffer.hpp
```diff
@@ -33,6 +33,24 @@ struct GlobalCircularBufferDramSenderInternals;
...
+    GlobalCircularBuffer(
+        distributed::MeshDevice& device,
+        const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
+        uint32_t size,
+        BufferType buffer_type = BufferType::L1);
+
+    [[deprecated(
+        "Use GlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
+        "GlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
     GlobalCircularBuffer(
```
obacherikovTT, line 54: This class is actually in `namespace experimental`.
Does this mean that I can just change its API without deprecation?

ttnn/cpp/ttnn-nanobind/global_circular_buffer.cpp
```diff
@@ -53,8 +52,7 @@ void py_module(nb::module_& mod) {
     // Multi Device APIs
     mod.def(
         "create_global_circular_buffer",
```
obacherikovTT: @nsextonTT afaik Python doesn't support function overloads, and in this file we bind two functions with the same name. Does it mean one of them was just overriding another?

ttnn/api/ttnn/global_semaphore.hpp
```diff
@@ -17,31 +17,13 @@ struct MultiDeviceGlobalSemaphore {
     auto attribute_values() const { return std::forward_as_tuple(this->global_semaphores); }
 };
```
ayerofieiev-tt: `MultiDeviceGlobalSemaphore` is orphaned by this PR: both factories that produced it are deleted, but the struct, its `size_t` ctor, the `get_global_semaphore_address`/`reset_global_semaphore_value` overloads, and the CCL op overloads that consume it (`all_gather_async.hpp:55`, `all_reduce_async.hpp:70`) all remain. Repo-wide grep confirms nothing constructs it after this diff, and the consuming CCL overloads are not nanobind-bound — permanently unreachable dead code. Either delete the type plus its consumers here, or keep a mesh-based factory.

tt_metal/impl/buffers/global_circular_buffer.cpp
ayerofieiev-tt, line 108: The new non-deprecated ctor delegates to the `IDevice*` ctor this same PR marks `[[deprecated]]` (removal promised after 2026-09-17). That means the deprecated ctor can't actually be deleted on the promised date without first re-homing its body. It's only warning-clean because of the repo-wide `-Wno-deprecated-declarations` (CMakeLists.txt:207). The same PR already shows the right pattern: `GlobalSemaphore(MeshDevice&)` bypasses its deprecated sibling and constructs the impl directly — invert the delegation here to match.

ttnn/core/global_circular_buffer.cpp
```diff
@@ -17,22 +17,13 @@
...
-}
-
 GlobalCircularBuffer create_global_circular_buffer(
     MeshDevice* device,
     const std::vector<std::pair<CoreCoord, CoreRangeSet>>& sender_receiver_core_mapping,
     uint32_t size,
     BufferType buffer_type) {
-    return tt::tt_metal::experimental::CreateGlobalCircularBuffer(
-        device, sender_receiver_core_mapping, size, buffer_type);
+    TT_FATAL(device != nullptr, "MeshDevice cannot be null");
```
ayerofieiev-tt: This `TT_FATAL(device != nullptr)` is hand-copied at three pointer-to-reference bridge sites (here, `ttnn/core/global_semaphore.cpp`, `tt_metal/impl/experimental/udm/mesh_semaphore.cpp`), while sibling functions in this same file (`create_global_circular_buffer_for_tensor_prefetcher`, `build_matmul_1d_gcb_krow_major`, `build_matmul_1d_gcb_recv_contig`) deref the same `MeshDevice*` unguarded. Since forming `*device` from null is UB at the call site, these bridge guards are the only functioning null check on the chain — every future bridge site has to remember the boilerplate. Consider migrating the ttnn signatures to `MeshDevice&` instead.

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
ayerofieiev-tt, line 22: Both `mod.def("create_global_semaphore", ...)` registrations now bind the identical function, differing only in kwarg name (`device=` vs `mesh_device=`), and this first one keeps a stale 'Single Device Creation API' docstring describing `device (Device)`. `help()` and nanobind TypeError messages will advertise two identical overloads, one documenting semantics that no longer exist. No in-repo Python caller uses either keyword (all positional). Collapse to one def with an accurate docstring.

ttnn/cpp/ttnn-nanobind/global_circular_buffer.cpp
ayerofieiev-tt, line 35: Same duplicate-registration issue as global_semaphore.cpp: both `create_global_circular_buffer` defs bind the identical function; this surviving first def still says 'on a single device' / `device (Device)`, and both carry the stale `nb::keep_alive<0, 1>(),  // test` comment. No Python call site uses `device=`/`mesh_device=` keywords. Collapse to a single def, fix the docstring, drop `// test`.

tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
```diff
@@ -4,25 +4,38 @@
 
 #pragma once
 
+#include <optional>
+
 #include <tt-metalium/global_semaphore.hpp>
+#include <tt-metalium/mesh_device.hpp>
```
ayerofieiev-tt: The full `mesh_device.hpp` include isn't needed here — the only new use is a `distributed::MeshDevice&` parameter, and the forward declaration added to `tt-metalium/global_semaphore.hpp` in this same PR is already visible via the include above. As-is, every TU including this header (tt-mlir is the stated audience) transitively compiles the mesh_device.hpp graph and gains a rebuild dependency on MeshDevice internals; once downstream relies on the transitive include, removing it becomes a breaking cleanup. The .cpp already includes mesh_device.hpp itself. Drop this include, keep `<optional>`.

ttnn/core/global_semaphore.cpp
```diff
@@ -7,6 +7,7 @@
 #include <algorithm>
```
ayerofieiev-tt: Dead includes after the deletions: `<algorithm>` (`std::all_of/max/min` lived only in the deleted `create_global_semaphore_with_same_address`), `<tt-metalium/host_api.hpp>` (`CreateGlobalSemaphore` is no longer called — and it's a heavyweight include), and `<tt_stl/span.hpp>` (unused even before this PR). Post-diff this file only needs the ttnn header, `<tt-metalium/global_semaphore.hpp>`, and `<tt_stl/assert.hpp>`.

tt_metal/impl/buffers/global_semaphore.cpp
```diff
@@ -105,6 +105,15 @@ void GlobalSemaphoreImpl::setup_buffer(
 
 namespace experimental {
 // Forge backdoor API.
+GlobalSemaphore CreateGlobalSemaphore(
+    distributed::MeshDevice& device,
+    const CoreRangeSet& cores,
+    std::optional<uint32_t> initial_value,
+    BufferType buffer_type,
+    uint64_t address) {
```
ayerofieiev-tt, line 113: This new forge-backdoor overload duplicates the `IDevice*` overload's body verbatim (modulo `&device` vs `device`) instead of sharing one construction path — any future change to backdoor semaphore construction has to be made twice. Have the deprecated `IDevice*` overload delegate to this one (deprecated → new is the right direction), matching the delegation style used elsewhere in this PR.

ayerofieiev-tt (commented):
Automated review (Claude Code) — 8 findings, each adversarially verified against the diff and repo. No blockers among them; the first three are worth addressing before merge, the rest are cleanups.

ttnn/api/ttnn/global_semaphore.hpp
obacherikovTT: I deleted the factories because they were using `create_global_semaphore` overload based on the deprecated constructor, preventing its deletion.

`all_gather_async` and `all_reduce_async` are pretty large, so I'd prefer to exclude them from the scope of this PR. I can remove them in a separate PR, or include in this one if it's preferred - please let me know.

tt_metal/impl/buffers/global_circular_buffer.cpp
obacherikovTT: I prefer re-homing the body when deleting the deprecated overload. That's because at that point we can be sure we're only getting `MeshDevice` and not other kind of `IDevice`, and can simplify the body accordingly.

`GlobalSemaphore` is easier to handle because it just forwards its parameters to `GlobalSemaphoreImpl`.

ttnn/api/ttnn/global_semaphore.hpp
obacherikovTT: I opened a separate PR to address this, aiming to merge it before this PR
https://github.com/tenstorrent/tt-metal/pull/53213

ttnn/core/global_circular_buffer.cpp
obacherikovTT: I'll remove the checks for consistency.

Migrating ttnn signatures is out of scope for this PR because they are involved in Python bindings.

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
obacherikovTT: I kept the one with the "device" parameter name according to your preference.

tt_metal/impl/buffers/global_semaphore.cpp
obacherikovTT: I'd prefer not to add extra complexity here, because it's unlikely that we'll change this code before deleting the deprecated version.

nsextonTT (approved):

tt_metal/api/tt-metalium/global_circular_buffer.hpp
riverwuTT: That's a good question. @akerteszTT

ayerofieiev-tt (approved):

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
```diff
@@ -17,17 +17,15 @@ namespace ttnn::global_semaphore {
 void py_module_types(nb::module_& mod) { nb::class_<GlobalSemaphore>(mod, "global_semaphore"); }
 
 void py_module(nb::module_& mod) {
-    // Single Device Creation API
     mod.def(
         "create_global_semaphore",
-        nb::overload_cast<IDevice*, const CoreRangeSet&, uint32_t, BufferType>(
-            &ttnn::global_semaphore::create_global_semaphore),
+        &ttnn::global_semaphore::create_global_semaphore,
```
ayerofieiev-tt, line 22: **Python keyword break.** Pre-PR, the MeshDevice overload was the only Python-reachable one (IDevice has no nanobind class, and MeshDevice is bound baseless so it never converted to `IDevice*`), and its kwarg was `mesh_device`. The sole surviving binding renames it to `device` with no shim, so external Python calling `ttnn.create_global_semaphore(mesh_device=md, cores=crs, initial_value=0)` — legal and documented before — now raises `TypeError` at runtime. All in-repo callers are positional, so CI won't catch the external breakage. Consider keeping `nb::arg("mesh_device")` or adding a kwarg alias.

ttnn/cpp/ttnn-nanobind/global_circular_buffer.cpp
```diff
@@ -30,18 +30,16 @@ void py_module_types(nb::module_& mod) {
...
 
 void py_module(nb::module_& mod) {
-    // Single Device APIs
     mod.def(
         "create_global_circular_buffer",
-        nb::overload_cast<IDevice*, const std::vector<std::pair<CoreCoord, CoreRangeSet>>&, uint32_t, BufferType>(
-            &ttnn::global_circular_buffer::create_global_circular_buffer),
-        nb::keep_alive<0, 1>(),  // test
+        &ttnn::global_circular_buffer::create_global_circular_buffer,
+        nb::keep_alive<0, 1>(),
```
ayerofieiev-tt, line 36: **Same Python keyword break as `create_global_semaphore`**: the removed MeshDevice overload's `mesh_device` kwarg is replaced by `device`. External `ttnn.create_global_circular_buffer(mesh_device=md, ...)` callers get `TypeError` post-PR; positional callers are unaffected, so it breaks silently at first keyword call.

tt_metal/api/tt-metalium/global_semaphore.hpp
```diff
@@ -18,15 +18,38 @@ namespace tt::tt_metal {
...
+     */
+    GlobalSemaphore(
+        distributed::MeshDevice& device,
+        CoreRangeSet cores,
+        uint32_t initial_value,
+        BufferType buffer_type = BufferType::L1);
+
+    [[deprecated(
```
ayerofieiev-tt, line 44: This deprecation note promises support until 2026-09-17 for a path that only works when the `IDevice*` is secretly a MeshDevice. `CreateDevice` still returns a plain `Device*`, and with one of those this ctor crashes instead of diagnosing: `initial_value` is non-optional so `setup_buffer` always calls `reset_semaphore_value`, which does `buffer_.get_mesh_buffer()` then `mesh_buffer->device()` — for a non-mesh `IDevice`, `AnyBuffer::create` stores a plain `Buffer` and `get_mesh_buffer()` returns nullptr → null shared_ptr deref/segfault with no message. Pre-existing behavior, but now that it's stamped with a support window: add a `TT_FATAL` with a "use the MeshDevice& ctor" message, or scope the note to mesh-backed devices.

tt_metal/impl/buffers/global_circular_buffer.cpp
ayerofieiev-tt, line 108: The new non-deprecated `GlobalCircularBuffer(MeshDevice&)` ctor delegates to the `GlobalCircularBuffer(IDevice*)` ctor this same PR marks `[[deprecated]]` for removal after 2026-09-17 — the recommended replacement is implemented in terms of the API scheduled for deletion, so removing it on schedule breaks the mesh ctor's own implementation. It compiles warning-clean today only because of the repo-wide `-Wno-deprecated-declarations` (CMakeLists.txt:207). Sibling `GlobalSemaphore`'s mesh ctor avoids this by calling `GlobalSemaphoreImpl` directly — invert the delegation the same way.

ttnn/api/ttnn/global_semaphore.hpp
```diff
@@ -17,31 +17,13 @@ struct MultiDeviceGlobalSemaphore {
     auto attribute_values() const { return std::forward_as_tuple(this->global_semaphores); }
```
ayerofieiev-tt: `MultiDeviceGlobalSemaphore` becomes an unconstructible API island: this PR deletes its only factories, but the struct, its `size_t` ctor, the MultiDevice overloads of `get_global_semaphore_address` / `reset_global_semaphore_value` (ttnn/core/global_semaphore.cpp:25,38), and the C++ overloads of `all_gather_async` (all_gather_async.hpp:55) and `all_reduce_async` (all_reduce_async.hpp:71) all survive with zero producers repo-wide (nanobind never binds the type either). Readers will assume a live multi-device path exists; external users of those CCL overloads have no way to build the argument. #53213 deletes most of these consumers — worth coordinating so nothing is left stranded between the two PRs.

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
```diff
@@ -17,17 +17,15 @@ namespace ttnn::global_semaphore {
...
         nb::arg("device"),
         nb::arg("cores"),
         nb::arg("initial_value"),
         nb::arg("buffer_type") = nb::cast(tt::tt_metal::BufferType::L1),
         R"doc(
-            Create a GlobalSemaphore Object on a single device.
+            Create a GlobalSemaphore Object on a device.
 
             Args:
                 device (Device): The device on which to create the global semaphore.
```
ayerofieiev-tt: The consolidation kept the wrong half of the docstring: it still reads `device (Device)` while the only accepted type is now MeshDevice (same in global_circular_buffer.cpp:45). At merge-base the MeshDevice overloads correctly documented `mesh_device (MeshDevice)`. As written, `help(ttnn.create_global_semaphore)` tells users a single Device is accepted; passing one raises an opaque nanobind `TypeError`. One-word fix.

ayerofieiev-tt (commented):
Automated review (Claude Code, requested by @ayerofieievTT). The migration itself is clean — all in-repo C++ and Python call sites updated correctly, `tt_metal/api` deprecations carry a consistent dated window, no efficiency regressions. 10 findings inline; the two that matter most before merge:

1. **Silent Python kwarg break**: `ttnn.create_global_semaphore` / `ttnn.create_global_circular_buffer` lose the `mesh_device=` kwarg (renamed to `device=` with no shim). External keyword callers get a runtime `TypeError`, and in-repo callers are all positional so CI won't surface it.
2. **The replacement is built on the thing being deleted**: the new `GlobalCircularBuffer(MeshDevice&)` ctor delegates to the `IDevice*` ctor this PR deprecates for removal.

Also note the `MultiDeviceGlobalSemaphore` finding overlaps with #53213 — worth coordinating which PR deletes the remaining consumers.

ttnn/core/global_semaphore.cpp
```diff
@@ -4,99 +4,18 @@
...
-
-GlobalSemaphore create_global_semaphore(
-    IDevice* device, const CoreRangeSet& cores, uint32_t initial_value, BufferType buffer_type) {
-    return CreateGlobalSemaphore(device, cores, initial_value, buffer_type);
-}
-
 GlobalSemaphore create_global_semaphore(
     MeshDevice* mesh_device, const CoreRangeSet& cores, uint32_t initial_value, BufferType buffer_type) {
-    return CreateGlobalSemaphore(mesh_device, cores, initial_value, buffer_type);
+    return GlobalSemaphore(*mesh_device, cores, initial_value, buffer_type);
```
ayerofieiev-tt, line 15: `*mesh_device` dereferences the raw pointer before any validation, losing the old descriptive `TT_FATAL("Device cannot be null")` path (same at ttnn/core/global_circular_buffer.cpp:25). Binding a reference from nullptr is immediate UB — UBSan builds trap, and the compiler may legally elide the downstream null check since `device_` is sourced from a reference. Low severity (nullptr was always a caller bug; no in-repo nullptr caller exists), but a diagnostics regression — a `TT_FATAL(mesh_device != nullptr, ...)` first line restores it.

tt_metal/impl/buffers/global_semaphore.cpp
```diff
@@ -117,6 +126,10 @@ GlobalSemaphore CreateGlobalSemaphore(
 
 // GlobalSemaphore implementation
 
+GlobalSemaphore::GlobalSemaphore(
```
ayerofieiev-tt, line 129: Altitude note: the new mesh ctors decay `MeshDevice&` to `IDevice*` one line in — `GlobalSemaphoreImpl` still stores `IDevice*` (global_semaphore_impl.hpp:60) and mesh-ness is re-derived via `dynamic_cast` in `AnyBuffer::create` (mesh_buffer.cpp:382-384, TODO #20966). Once the deprecated ctors are deleted, every object is mesh-backed by construction yet the impl and accessors still traffic in `IDevice*`, so #20966's cleanup stays blocked and a second migration pass over the same files is needed. Non-blocking, but pushing `MeshDevice` into the impl now would finish the thought.

ttnn/core/global_semaphore.cpp
```diff
@@ -4,99 +4,18 @@
 
 #include "ttnn/global_semaphore.hpp"
 
-#include <algorithm>
-#include <tt-metalium/host_api.hpp>
 #include <tt-metalium/global_semaphore.hpp>
-#include <tt_stl/span.hpp>
+#include <tt_stl/assert.hpp>
```
ayerofieiev-tt: This include is dead on arrival: the only `TT_*` user in this file was the `TT_THROW` in the deleted `create_global_semaphore_with_same_address`. No `TT_THROW`/`TT_FATAL`/`TT_ASSERT` remains — the surviving code needs only `ttnn/global_semaphore.hpp` and `tt-metalium/global_semaphore.hpp`. (Moot if you add the null `TT_FATAL` suggested below.)

tests/tt_metal/tt_metal/dispatch/sub_device_test_utils.hpp
```diff
@@ -6,17 +6,18 @@
...
 #include <tt-metalium/host_api.hpp>
 #include <tt-metalium/global_semaphore.hpp>
+#include <tt-metalium/mesh_device.hpp>
 #include "impl/context/metal_context.hpp"
 #include "impl/kernels/kernel.hpp"
 #include "sub_device.hpp"
 
 namespace tt::tt_metal {
 
 inline std::tuple<Program, CoreCoord, GlobalSemaphore> create_single_sync_program(
```
ayerofieiev-tt, line 16: Minor: these helpers are retyped to `distributed::MeshDevice*` and then unconditionally dereferenced (`*device`), while the API convention this PR establishes is `MeshDevice&`. Callers already pass `.get()` from a shared_ptr, so taking a reference (`create_single_sync_program(*mesh_device, ...)` at call sites) makes null unrepresentable and moves any crash to the call boundary instead of the hidden deref.

ttnn/cpp/ttnn-nanobind/global_semaphore.cpp
ayerofieiev-tt: can omit

ttnn/cpp/ttnn-nanobind/global_circular_buffer.cpp
obacherikovTT: I followed the previous recommendation to keep the "device" name, but I'm happy to switch to the "mesh_device".

tt_metal/api/tt-metalium/global_semaphore.hpp
obacherikovTT: I'd prefer not to modify the current (erroneous) behavior and simply remove it after the deprecation window to avoid extra risks in this PR.

ttnn/api/ttnn/global_semaphore.hpp
obacherikovTT: MultiDeviceGlobalSemaphore was removed in a separate PR, and I rebased on top of it https://github.com/tenstorrent/tt-metal/pull/53213

tt_metal/impl/buffers/global_circular_buffer.cpp
obacherikovTT: I already replied to the same comment here https://github.com/tenstorrent/tt-metal/pull/53199#discussion_r3788067056

ttnn/core/global_semaphore.cpp
obacherikovTT: I followed the suggestion from the previous review https://github.com/tenstorrent/tt-metal/pull/53199#discussion_r3787989514

tests/tt_metal/tt_metal/dispatch/sub_device_test_utils.hpp
obacherikovTT: I kept the pointer here to avoid changing extra files in this PR and keep it more focused.

tt_metal/impl/buffers/global_semaphore.cpp
obacherikovTT: Will address this after the deprecation window.

tt_metal/api/tt-metalium/experimental/forge_backdoor/global_semaphore.hpp
```diff
@@ -4,25 +4,37 @@
...
  */
 // clang-format on
+GlobalSemaphore CreateGlobalSemaphore(
+    distributed::MeshDevice& device,
+    const CoreRangeSet& cores,
+    std::optional<uint32_t> initial_value,
+    BufferType buffer_type,
+    uint64_t address);
+
+[[deprecated(
```
akerteszTT, line 35: Technically speaking, the `experimental` directory is exempted from the deprecation policy. You could just change this directly.

That said, this particular header is used by Forge (a TT compiler in a separate repo). They periodically sync up to the latest tt-main. So, in actuality, your deprecation choice is appropriate and quite polite; they'll see it on their next sync-to-tt-metal. But of you prefer, you could simply give them a heads up. Either is fine.

tt_metal/api/tt-metalium/global_circular_buffer.hpp
akerteszTT: Stuff that's in the `experimental` namespace but not the `experimental` directory predates our stability policy. In the case of GlobalCircularBuffer, it was created in the `experimental` namespace but has been kicking around the public API for a long time.... in my view, we should treat it as part of the public API.

We ought at some point to clean up these long-dead `experimental` namespace usages littered around the API dir. They all predate when I started policing the API headers.

Heads up, `GlobalCircularBuffer` will in time be deprecated in favor of `GlobalDataflowBuffer`, which is currently in development. It'll be a long deprecation timeline though.

tt_metal/api/tt-metalium/global_circular_buffer.hpp
```diff
@@ -117,15 +135,33 @@ class GlobalCircularBuffer {
...
+/**
+ * @brief Allocates a global circular buffer in L1 on the device.
+ *
+ * @param device Mesh device to create the global circular buffer on.
+ * @param sender_receiver_core_mapping The mapping of remote sender to remote receiver cores for the circular buffer.
+ * @param size Size of the global circular buffer per core in bytes.
+ * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.
+ * @return The allocated global circular buffer.
+ */
+GlobalCircularBuffer CreateGlobalCircularBuffer(
```
akerteszTT, line 147: Thanks for adding the free function! I know it's superfluous, but it's canonical. 

I'd envisioned having it in `host_api.hpp` to keep the legacy API self-contained there. (It's a little goofy to have it in the same header.) That said, that is awkward given that GlobalCB is `experimental`, and you're sticking with precedent.

tt_metal/api/tt-metalium/global_circular_buffer.hpp
```diff
@@ -117,15 +135,33 @@ class GlobalCircularBuffer {
...
  * @param device The device to create the global circular buffer on.
  * @param sender_receiver_core_mapping The mapping of remote sender to remote receiver cores for the circular buffer.
  * @param size Size of the global circular buffer per core in bytes.
- * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.\
+ * @param buffer_type Buffer type to store the global circular buffer. Can only be an L1 buffer type.
  * @return The allocated global circular buffer.
  */
+[[deprecated(
+    "Use CreateGlobalCircularBuffer(distributed::MeshDevice&, ...) instead. "
+    "CreateGlobalCircularBuffer(IDevice*, ...) will be removed after 2026-09-17.")]]
```
akerteszTT: Remember to update the date based on when the PR merges.

(I'm usually too lazy to be specific on the data for this reason.)

tt_metal/api/tt-metalium/global_semaphore.hpp
akerteszTT: I vote remove the superfluous argument. 

I'm guessing someone thought this should be extensible, then never extended it? A GlobalSemaphore in DRAM would be absurd though.

akerteszTT (approved):
Looks good, thanks! Left some minor comments; mostly just FYIs.

vtsilytskyiTT (approved):
LGTM

tt_metal/api/tt-metalium/global_semaphore.hpp
obacherikovTT: I checked, and this parameter is actually used because apparently we have multiple types of L1
https://github.com/tenstorrent/tt-metal/blob/6b35ca8d35e722fe6ce4579db5708311f498ac92/ttnn/cpp/ttnn/operations/ccl/all_gather/device/all_gather_unicast_factory.cpp#L51

jbaumanTT (approved):

Issue #46121, opened 2026-06-04
[tt-train] Accuracy issues with a set of tt-train tests


kevinwuTT:
At least two runs on main have reported failures with the following tests. They are all accuracy related issues. However, the failures are not deterministic. 

June 1: https://tenstorrent.slack.com/archives/C09CK9093LH/p1780371472701349?thread_ts=1780371224.610849&cid=C09CK9093LH
June 3: https://tenstorrent.slack.com/archives/C09CK9093LH/p1780506960499709?thread_ts=1780506696.203329&cid=C09CK9093LH

## Failing tests grouped by similar errors
| Test | Detail |
|------|--------|
| `SDPAForwardTest.SDPAForwardTest_CausalMask_MHA_Batch4_Seq256` | MSE inf (result + intermediate) |
| `SDPAForwardTest.SDPAForwardTest_CausalMask_GQA_Batch16_Seq512` | MSE inf (result + intermediate) |
| `SDPAForwardTest.ValidationTest_EdgeCaseDimensions` | MSE inf (result) |
| `SDPAForwardTest.SDPAForwardTest_DifferentVDim_MultiBatch` | MSE inf (result) |

| Test | Detail |
|------|--------|
| `SDPABackwardTest.SmallBatch` | NaN/Inf in composite_dV; fwd output + kernel dQ/dK/dV mismatch |
| `SDPABackwardTest.TinyLlamaConfig` | NaN/Inf in kernel_dK; fwd output + kernel dQ/dK/dV mismatch |
| `SDPABackwardTest.CausalMask_MHA` | NaN/Inf in composite_dV; fwd output mismatch |
| `SDPABackwardTest.CausalMask_GQA` | NaN/Inf in composite_dQ and composite_dV; fwd output mismatch |
| `SDPABackwardTest.DiffVDim_Causal_SmallV` | NaN/Inf in composite_dV; fwd output + kernel dK mismatch |
| `SDPABackwardTest.DiffVDim_Causal_LargeV` | NaN/Inf in composite_dV; fwd output mismatch |
| `SDPABackwardTest.DiffVDim_ArbitraryMask` | NaN/Inf in composite_dV; fwd output + kernel dQ/dK/dV mismatch |
| `SDPABackwardTest.DiffVDim_GQA_Causal` | NaN/Inf in composite_dV; fwd output mismatch |
[...]

PR #53217, opened 2026-08-15
[tt-train] Fix sdpa_fw mask=None hang, WH scale truncation, and sdpa_bw validation gaps

### LINKED ISSUE #53207: [tt-train] sdpa_fw hangs with AttentionMaskType::None; WH fw/bw softmax-scale mismatch; sdpa_bw validation gaps and NaN on fully-masked rows
Found in a systematic audit of tt-train's fused metal kernels.

## Bug 1: `sdpa_fw` with `AttentionMaskType::None` deadlocks

`sdpa_fw_compute_kernel.cpp:194-224` — the `#else` branch (mask_type == None) unconditionally calls `apply_mask_on_reg(matmul_accum_reg, cb_attn_mask, ...)` (:212) and `cb_pop_front(cb_attn_mask, Sk_chunk_t)` (:223), but for None:

- the host never creates CB c_3 (`sdpa_fw_program_factory.cpp:401` guards on `mask_type != None`),
- the reader never reads mask tiles (all mask reads under `#ifdef USE_ATTN_MASK`),
- the writer only generates causal tiles under CAUSAL/BALANCED.

`apply_mask_on_reg` does `cb_wait_front` on a CB that is never configured or pushed → hang or garbage. Validation accepts None (`sdpa_fw_device_operation.cpp:128-136`); the host comment ("kernel will just scale attention scores", factory:520) shows the no-mask path was intended but never implemented. The bw kernels do this correctly (`sdpa_bw_q_compute_kernel.cpp:152-155` has a proper scale-only `#else`).

**Trigger:** any `metal::sdpa_fw(..., None, ...)` call — exactly what the ring-attention merge simulation test does (`sdpa_bw_op_test.cpp:1137,1225`); that test is GTEST_SKIPped, so zero coverage. Fix is a 2-line `#ifdef USE_ATTN_MASK` guard.

## Bug 2: fw/bw softmax-scale mismatch on Wormhole

`sdpa_compute_utils_common.hpp:174,189` — the WH path of `sdpa_exp_tile_scaled` uses `scaler_bf16 = (uint16_t)(scaler_fp32 >> 16)`: a bf16 **truncation** (not round-to-nearest) of the scale. So fw computes P with the bf16 scale, but fw's LSE uses the full fp32 scale (`compute_and_pack_lse`, `sdpa_compute_utils.hpp:459-461`) and bw recomputes P = exp(scale_fp32·s − lse) with the fp32 scale (`sdpa_bw_compute_utils.hpp:61`). For head dims where 1/√d is not bf16-exact (d=128: rel. error ~2⁻⁹; d=64 is exact, masking this in most tests), bw's recomputed P deviates from fw's actual P by several percent on low-probability entries — a systematic fw/bw inconsistency biasing dS. BH is unaffected (fp32 scale in LREG12). Plausibly part of why D=128 bw configs need 3e-2 tolerances.

## Bug 3: bw host validation gaps → silent gradient corruption

- **dtype never pinned to BF16:** both bw factories size all CBs with `bfloat16_single_tile_size_bytes` but set data_format from the actual dtype (`sdpa_bw_q_program_factory.cpp:252-254,327-351`); validation only checks dtypes are equal to each other. FLOAT32 inputs → 4KB tiles in 2KB pages → garbage. fw validates BF16 explicitly.
- **`intermediates` completely unvalidated** in both bw ops — must be the fw (B,H,S,32) FP32 lse tensor; reader assumes FP32 tile size. Wrong tensor → silent garbage gradients.
- **dropout silently ignored in bw** (fw TT_FATALs on dropout≠0; bw accepts and ignores).
- No tile-alignment validation: `scaler = 1/√(padded qEmbd)` (`sdpa_fw_program_factory.cpp:271,293-294`) — logical head_dim 40 → padded 64 → wrong scale, no error.

## Bug 4: fully-masked rows (Arbitrary mask) → uniform attention in fw, inf/NaN in bw

fw masks by adding −1e9 before the row-max reduce, so a fully-masked row gets exp(scale·(s−max))=1 everywhere → output = mean(V), lse ≈ −1e8. bw computes P = exp(scale·s − lse) = exp(+1e8) = inf on any position not masked identically — and bw uses true −inf (`hal::get_inf()`, `sdpa_bw_q_program_factory.cpp:279`) vs fw's −1e9, a second fw/bw asymmetry. Padding masks that fully mask a row (padded batches) hit NaN gradients. No test covers this.

## Context: backward is effectively unverified

10 of 12 bw accuracy tests are DISABLED_ citing #46121 (non-deterministic accuracy failures); the only always-on passing bw test is B=1, S=32, single core, single tile. The two largest fw causal tests (the only ones exercising balanced parallelism + Sk_chunk_t=4 with the packer-L1-accumulate mask stamp) are also disabled. GQA, causal, and diff-vdim gradients should be treated as unknown-correct until #46121 is root-caused.

## Fix order

1. Guard the fw None path (trivial).
2. Pin bw dtypes, validate intermediates, TT_FATAL on dropout≠0 in bw, TT_FATAL on padded≠logical dims.
3. Use round-to-nearest (or fp32) for the WH exp scale to match LSE/backward.
4. Root-cause #46121 before re-enabling the disabled suites.

ayerofieiev-tt:
Fixes #53207

## What

- **mask=None hang**: `sdpa_fw_compute_kernel.cpp` unconditionally waited on / popped the mask CB (c_3), which the host never creates for `AttentionMaskType::None` — guaranteed hang. Mask application now compiled out on the None path (scaling was never in `apply_mask_on_reg`, so no substitute needed). Guarded by always-on regression tests (`SDPAForwardTest_NoMask_Small`, `SDPAForwardTest_NoMask_TwoTileRows`) — previously the only None-path caller was a permanently skipped NIGHTLY test.
- **WH scale truncation**: the WH exp path built its bf16 scale by truncation (`>>16`); now round-to-nearest-even. Reduces the systematic fw/bw P mismatch for head dims where 1/√d isn't bf16-exact (d=128). BH path untouched.
- **bw validation** (both sdpa_bw ops): inputs pinned to BFLOAT16 (CB pages are bf16-sized; FP32 inputs previously produced silent garbage — including `attn_output`, which was never dtype-checked at all); `intermediates` validated (FP32/TILE/INTERLEAVED/(B,H,S,32)) — INTERLEAVED means DRAM **or** L1: an earlier revision said DRAM-only, but `sdpa_fw` allocates intermediates with the query's memory config, so an L1-interleaved query legitimately yields L1-interleaved intermediates and the reader's TensorAccessor handles both; only sharded layouts are rejected; `TT_FATAL(dropout == 0)` (fw already had it; bw silently ignored dropout). Covered by four negative validation tests (`Validation_RejectsFP32Inputs`, `Validation_RejectsMalformedIntermediates`, `Validation_RejectsNonzeroDropout`, `Validation_RejectsPaddedQK`).
- **Padded-dim scale**: fw and bw now reject padded ≠ logical on S and head_dim — the softmax scaler was computed from the padded dim (head_dim 40 silently used 1/8 instead of 1/√40). The alignment check is restricted to **Q/K head_dim only**: V with a non-tile-aligned vdim is a supported DiffVDim config (V's vdim never feeds the scaler) and the first revision wrongly rejected it — caught on device, fixed in 504173ce67e.

Remaining work under #53207 (design decision, not taken here): fully-masked rows produce uniform attention in fw and inf/NaN in bw.

## Validation checklist

- [x] sdpa fw/bw suites on BH — full single-chip round green (605 tests passed) after 504173ce67e, DiffVDim configs included
- [ ] sdpa fw/bw suites on WH — pending
- [x] `metal::sdpa_fw(..., None, ...)` on device — previously a guaranteed hang; no hang on BH (single-chip suite green), plus the two always-on no-mask regression tests pass. WH run pending for full coverage
- [ ] Check whether RTN scaling tightens the D=128 bw tolerances (needs the WH run)
- [x] Negative tests: FP32 bw inputs, wrong intermediates, dropout≠0, non-tile-aligned Q/K dims TT_FATAL as expected — all four device-green in the 605-test round
- [ ] Do NOT re-enable the 10 DISABLED_ bw suites until #46121 is root-caused

## Device verification (BH galaxy, bh_sc5 quad)

All runs on Blackhole galaxies; BH was the test vehicle for this PR.

- The new tile-alignment validation initially rejected V with non-tile-aligned vdim — a supported DiffVDim config. Caught on device; fixed by restricting the check to Q/K (504173ce67e). Rerun confirmed: `SDPAForwardTest`/`SDPABackwardTest` green in the full 605-test single-chip round, including the no-mask regression tests and the four negative validation tests.
- mask=None path: single-chip BH suite green — no hang.
- Coverage caveat: the None-path hang fix and the WH RTN scale change still need a WH run for full coverage (both code paths were exercised on BH only).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ayerofieiev/kernel-fix/sdpa-2026-08-14)

### DIFF
--- a/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
+++ b/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
@@ -9,19 +9,6 @@
 #include "api/compute/compute_kernel_api.h"
 #include "api/compute/eltwise_unary/exp.h"
 
-// FP32 bits → BF16 bits with round-to-nearest-even. A plain `>> 16` truncation
-// can be off by a full BF16 ulp (~2^-8 relative), which matters here because the
-// WH exp path applies the BF16 scale to P while the LSE and the backward pass use
-// the full FP32 scale — the rounding error would become a systematic fw/bw
-// mismatch instead of noise. RTN halves the worst-case error; carrying full FP32
-// through the WH SFPU scale multiply would remove it entirely but needs LLK work.
-// Unlike the host-side bfloat16 conversion this has no NaN guard: the bias carry
-// would turn a NaN into Inf, but the only input is 1/sqrt(head_dim), always finite.
-inline constexpr uint16_t sdpa_fp32_bits_to_bf16_rne(uint32_t fp32_bits) {
-    const uint32_t rounding_bias = 0x7FFFU + ((fp32_bits >> 16) & 1U);
-    return static_cast<uint16_t>((fp32_bits + rounding_bias) >> 16);
-}
-
 #ifdef TRISC_MATH
 #ifdef ARCH_BLACKHOLE
 
@@ -179,15 +166,17 @@ inline void sdpa_exp_tile_init() {
 }
 
 // Arch-dispatched scaled exp: exp(scale * x) over 8 face groups.
-//   WH: pre-multiplies by scale via sfpi then calls accurate exp directly.
-//   BH: folds scale into LREG12 at init time — one SFPU op fewer per element.
-template <uint32_t scaler_fp32>
+//   WH: pre-multiplies by the BF16 scale via sfpi then calls accurate exp directly.
+//       `scaler_bf16` carries the host-side RNE conversion of `scaler_fp32` (low 16 bits).
+//   BH: folds the full FP32 scale into LREG12 at init time — one SFPU op fewer per element.
+template <uint32_t scaler_fp32, uint32_t scaler_bf16>
 inline void sdpa_exp_tile_scaled(uint32_t idst) {
 #ifdef ARCH_WORMHOLE
-    constexpr uint16_t scaler_bf16 = sdpa_fp32_bits_to_bf16_rne(scaler_fp32);
 #ifdef TRISC_MATH
     _llk_math_eltwise_unary_sfpu_params_(
-        _sdpa_detail::mul_then_sfpi_exp</*ITERATIONS*/ 8, DST_ACCUM_MODE, scaler_bf16>, idst, VectorMode::RC);
+        _sdpa_detail::mul_then_sfpi_exp</*ITERATIONS*/ 8, DST_ACCUM_MODE, static_cast<uint16_t>(scaler_bf16)>,
+        idst,
+        VectorMode::RC);
 #endif
 #elif defined(ARCH_BLACKHOLE)
     sdpa_exp_tile_init</*approx*/ false, /*SCALE_EN*/ true, scaler_fp32>();
@@ -196,15 +185,15 @@ inline void sdpa_exp_tile_scaled(uint32_t idst) {
 }
 
 // Arch-dispatched scaled first-column exp: exp(scale * x) over column 0 only (4 iterations).
-template <uint32_t scaler_fp32>
+template <uint32_t scaler_fp32, uint32_t scaler_bf16>
 inline void sdpa_exp_tile_first_column_scaled(uint32_t idst) {
 #ifdef ARCH_WORMHOLE
-    constexpr uint16_t scaler_bf16 = sdpa_fp32_bits_to_bf16_rne(scaler_fp32);
 #ifdef TRISC_MATH
     constexpr int ITERATIONS_HALF_FACE = 4;
     constexpr int DST_STRIDE = 2;
     _llk_math_eltwise_unary_sfpu_params_(
-        _sdpa_detail::mul_then_sfpi_exp<ITERATIONS_HALF_FACE, DST_ACCUM_MODE, scaler_bf16, DST_STRIDE>,
+        _sdpa_detail::
+            mul_then_sfpi_exp<ITERATIONS_HALF_FACE, DST_ACCUM_MODE, static_cast<uint16_t>(scaler_bf16), DST_STRIDE>,
         idst,
         VectorMode::C);
 #endif
--- a/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
+++ b/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
@@ -9,6 +9,17 @@
 #include "api/compute/compute_kernel_api.h"
 #include "api/compute/eltwise_unary/exp.h"
 
+// FP32 bits → BF16 bits with round-to-nearest-even. A plain `>> 16` truncation
+// can be off by a full BF16 ulp (~2^-8 relative), which matters here because the
+// WH exp path applies the BF16 scale to P while the LSE and the backward pass use
+// the full FP32 scale — the rounding error would become a systematic fw/bw
+// mismatch instead of noise. RTN halves the worst-case error; carrying full FP32
+// through the WH SFPU scale multiply would remove it entirely but needs LLK work.
+inline constexpr uint16_t sdpa_fp32_bits_to_bf16_rne(uint32_t fp32_bits) {
+    const uint32_t rounding_bias = 0x7FFFU + ((fp32_bits >> 16) & 1U);
+    return static_cast<uint16_t>((fp32_bits + rounding_bias) >> 16);
+}
+
 #ifdef TRISC_MATH
 #ifdef ARCH_BLACKHOLE
 
@@ -171,7 +182,7 @@ inline void sdpa_exp_tile_init() {
 template <uint32_t scaler_fp32>
 inline void sdpa_exp_tile_scaled(uint32_t idst) {
 #ifdef ARCH_WORMHOLE
-    constexpr uint16_t scaler_bf16 = static_cast<uint16_t>(scaler_fp32 >> 16);
+    constexpr uint16_t scaler_bf16 = sdpa_fp32_bits_to_bf16_rne(scaler_fp32);
 #ifdef TRISC_MATH
     _llk_math_eltwise_unary_sfpu_params_(
         _sdpa_detail::mul_then_sfpi_exp</*ITERATIONS*/ 8, DST_ACCUM_MODE, scaler_bf16>, idst, VectorMode::RC);
@@ -186,7 +197,7 @@ inline void sdpa_exp_tile_scaled(uint32_t idst) {
 template <uint32_t scaler_fp32>
 inline void sdpa_exp_tile_first_column_scaled(uint32_t idst) {
 #ifdef ARCH_WORMHOLE
-    constexpr uint16_t scaler_bf16 = static_cast<uint16_t>(scaler_fp32 >> 16);
+    constexpr uint16_t scaler_bf16 = sdpa_fp32_bits_to_bf16_rne(scaler_fp32);
 #ifdef TRISC_MATH
     constexpr int ITERATIONS_HALF_FACE = 4;
     constexpr int DST_STRIDE = 2;
--- a/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
+++ b/tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
@@ -15,6 +15,8 @@
 // the full FP32 scale — the rounding error would become a systematic fw/bw
 // mismatch instead of noise. RTN halves the worst-case error; carrying full FP32
 // through the WH SFPU scale multiply would remove it entirely but needs LLK work.
+// Unlike the host-side bfloat16 conversion this has no NaN guard: the bias carry
+// would turn a NaN into Inf, but the only input is 1/sqrt(head_dim), always finite.
 inline constexpr uint16_t sdpa_fp32_bits_to_bf16_rne(uint32_t fp32_bits) {
     const uint32_t rounding_bias = 0x7FFFU + ((fp32_bits >> 16) & 1U);
     return static_cast<uint16_t>((fp32_bits + rounding_bias) >> 16);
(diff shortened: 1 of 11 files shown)


tt-train/sources/ttml/metal/ops/sdpa_fw/device/kernels/compute/sdpa_fw_compute_kernel.cpp
```diff
@@ -209,7 +213,9 @@ FORCE_INLINE void process_single_row(uint32_t global_row_idx) {
                     tile_idx * Sk_chunk_t + n,
                     /* dst */ matmul_accum_reg);
             }
+#ifdef USE_ATTN_MASK
             apply_mask_on_reg(matmul_accum_reg, cb_attn_mask, minus_one_bits, custom_inf_bits, /* mask_tile_idx */ n);
+#endif
```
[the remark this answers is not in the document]
ayerofieiev-tt: Agreed — real gap. Added always-on forward tests `SDPAForwardTest_NoMask_Small` and `SDPAForwardTest_NoMask_TwoTileRows` in 63e0cf1: they run `sdpa_fw` with `AttentionMaskType::None` and check numerical output against the composite and float references (the reference mask is all-ones for the None path, since the kernel computes unmasked attention). They complete-or-hang, so a regression of this fix now fails the regular suite instead of only the skipped NIGHTLY ring test. Device run pending (tracked in the PR checklist).

tt-train/sources/ttml/metal/ops/sdpa_bw/device/sdpa_bw_q_device_operation.cpp
```diff
@@ -47,21 +47,92 @@ void SDPABackwardQDeviceOperation::validate_on_program_cache_miss(
...
-    // Validate data formats
+    // Validate data formats. The program factory sizes every input CB for BFLOAT16 tiles,
+    // so any other dtype would overrun CB pages and silently corrupt gradients.
     TT_FATAL(
-        grad_output.dtype() == query.dtype() && query.dtype() == key.dtype() && key.dtype() == value.dtype(),
-        "All input tensors must have the same data type");
+        grad_output.dtype() == tt::tt_metal::DataType::BFLOAT16 &&
+            tensor_args.attn_output.dtype() == tt::tt_metal::DataType::BFLOAT16 &&
+            query.dtype() == tt::tt_metal::DataType::BFLOAT16 &&
+            key.dtype() == tt::tt_metal::DataType::BFLOAT16 && value.dtype() == tt::tt_metal::DataType::BFLOAT16,
```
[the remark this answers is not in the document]
ayerofieiev-tt: Added in 63e0cf1 — focused `EXPECT_ANY_THROW` negative tests in `sdpa_bw_op_test.cpp` covering the new contract: `Validation_RejectsFP32Inputs` (FP32 grad_output and FP32 attn_output), `Validation_RejectsMalformedIntermediates` (wrong shape (B,H,S,64) and wrong dtype BF16), `Validation_RejectsNonzeroDropout` (dropout=0.5), and `Validation_RejectsPaddedQK` (head_dim=48, tile-padded to 64). Each starts from a minimal valid input set and corrupts exactly one property. Device run pending (tracked in the PR checklist).

tt-train/sources/ttml/metal/ops/sdpa_bw/device/sdpa_bw_q_device_operation.cpp
```diff
@@ -47,21 +47,92 @@ void SDPABackwardQDeviceOperation::validate_on_program_cache_miss(
...
+        "intermediates must have TILE layout, got {}",
+        intermediates.layout());
+    TT_FATAL(
+        intermediates.dtype() == tt::tt_metal::DataType::FLOAT32,
+        "intermediates must be FLOAT32, got {}",
+        intermediates.dtype());
+    TT_FATAL(
+        intermediates.buffer() != nullptr &&
+            intermediates.memory_config().memory_layout() == tt::tt_metal::TensorMemoryLayout::INTERLEAVED,
+        "intermediates must be interleaved");
```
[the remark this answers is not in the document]
ayerofieiev-tt: Good catch — the PR description was stale. INTERLEAVED-only (DRAM or L1) is the intended contract: the original DRAM-only pin was relaxed after device evidence, because `sdpa_fw` allocates intermediates with query's memory config, so an L1-interleaved query legitimately yields L1-interleaved intermediates, and the reader's TensorAccessor handles both. Only sharded layouts are rejected. Updated the PR description to match the code.

tt-train/sources/ttml/metal/ops/sdpa_fw/device/kernels/compute/sdpa_fw_compute_kernel.cpp
```diff
@@ -192,10 +192,14 @@ FORCE_INLINE void process_single_row(uint32_t global_row_idx) {
...
 #else
-        // USE_ATTN_MASK (provided mask) — uses a per-`n` matmul_tiles + apply_mask_on_reg path.
+        // USE_ATTN_MASK (provided mask) and no-mask — a per-`n` matmul_tiles path.
         // The per-n mask is unique (no reusable transformed tiles) and apply_mask_on_reg operates
         // on a single DST tile + scratch, which fits comfortably here. K is laid out col-major
         // in cb_key (uniform reader layout), so the K tile index is `feat*Sk_chunk_t + n`.
+        //
+        // Without USE_ATTN_MASK (AttentionMaskType::None) the host never creates cb_attn_mask
+        // and the reader never fills it, so every mask touch must be compiled out; scaling
+        // still happens later inside the scaled exp of the softmax.
```
vtsilytskyiTT: This is obvious

tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
```diff
@@ -9,6 +9,19 @@
...
 
+// FP32 bits → BF16 bits with round-to-nearest-even. A plain `>> 16` truncation
+// can be off by a full BF16 ulp (~2^-8 relative), which matters here because the
+// WH exp path applies the BF16 scale to P while the LSE and the backward pass use
+// the full FP32 scale — the rounding error would become a systematic fw/bw
+// mismatch instead of noise. RTN halves the worst-case error; carrying full FP32
+// through the WH SFPU scale multiply would remove it entirely but needs LLK work.
+// Unlike the host-side bfloat16 conversion this has no NaN guard: the bias carry
+// would turn a NaN into Inf, but the only input is 1/sqrt(head_dim), always finite.
+inline constexpr uint16_t sdpa_fp32_bits_to_bf16_rne(uint32_t fp32_bits) {
```
vtsilytskyiTT: We already have `fp32_to_bf16_bits_round_to_nearest_even` in the tt_metal/impl/data_format/bfloat16.cpp

vtsilytskyiTT (changes requested):
I suggest to extract existing function from anonymous namespace and reuse it.

tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
ayerofieiev-tt: Done in 823e246d6e3. The duplicate existed because this is a device-kernel header — the kernel TU can't include the host impl — so instead I moved the conversion host-side: extracted `fp32_to_bf16_bits_round_to_nearest_even` from the anonymous namespace (declared in `tt-metalium/bfloat16.hpp`), the sdpa_fw program factory now converts once and passes the BF16 bits as an extra compile-time arg, and the kernel helper is deleted.

tt-train/sources/ttml/metal/ops/sdpa_fw/device/kernels/compute/sdpa_fw_compute_kernel.cpp
ayerofieiev-tt: Removed in 823e246d6e3.

vtsilytskyiTT (approved):

tt-train/sources/ttml/metal/common/sdpa_compute_utils_common.hpp
vtsilytskyiTT: Yes, this is fine. I'm surprised that copilot didn't spot the code duplication.

ncvetkovicTT (approved):

riverwuTT (approved):

vmelnykovTT (approved):
LGTM!

PR #53263, opened 2026-08-15
Bug fix: validate sharded buffer cores against the allocator's banks

jbaumanTT:
### PR Category

Bug fix

### Summary

A sharded buffer's cores were never validated against the banks the allocator
actually has, so a buffer sharded onto a core with no bank was constructed
successfully and only went wrong later, if at all.

Nothing further down the allocation path can catch it. `AllocatorImpl::allocate_buffer()`
is handed `buffer->num_cores()` -- a shard *count*, not the coordinates -- so it
has nothing to check. `BankManager::allocate_buffer()` does bound that count, but
against `compute_grid.num_cores()` for every buffer type, which for DRAM means an
8-shard grid is compared against 120 L1 compute cores and passes. Ops resolve bank
ids ad hoc via `get_bank_ids_from_logical_core()`, so whether an invalid core is
caught depends on which op touches the buffer.

Resharding to a harvested DRAM grid shows what that costs. Of 16 failures on a
Blackhole part with one GDDR channel harvested: 9 aborted in the channel lookup,
1 in the allocator, and **6 silently produced wrong data** (PCC 0.8657-0.8779).
Same invalid buffer in every case -- only the path that happened to touch it
differed.

So `validate_buffer_parameters()` now asks the allocator whether each shard core
has a bank, via a new `AllocatorImpl::has_bank()`. Construction is the one place
that runs for every buffer regardless of which op later uses it.

Relates to #53264

### Notes for reviewers

**Why the allocator rather than a bank-count comparison.** An earlier revision of
this PR compared `core.x < num_dram_banks`. That only works because DRAM banks are
1D; there is no equivalent for L1's 2D grid, so it had no L1 arm at all. Asking the
allocator works for both and cannot drift from how the allocator was configured --
including sub-device allocators, whose worker grid is narrowed to the sub-device's
compute cores.

**L1_SMALL is asked about the L1 bank map, not its own.** One loop over the logical
cores fills both maps, so they agree on which coordinates are legal -- but the
L1_SMALL map is left empty entirely when the device is opened without a small
region, and `DEFAULT_L1_SMALL_SIZE` is 0. Asking it directly answers "no bank here"
for every core on such a device, which is a fact about the allocator's
configuration and not about the shard grid. Whether a small region exists stays the
allocator's to report, and it does. A buffer that is constructed but never
allocated -- graph capture hooks the allocation out -- legitimately never asks.

This one is worth a look because it is the failure that dequeued the first revision
of this PR from the merge queue: `Conv2dOpIfTest.Conv2d` builds an L1_SMALL
sliding-window config tensor under `query_op_constraints`, and pool, upsample and
the quasar factories build the same kind of tensor.

**`has_bank()` skips the mutex.** It is the same lookup as
`get_bank_ids_from_logical_core()` but reports absence instead of throwing. The bank
map is populated during init and never mutated afterwards, which `get_num_banks()`
already relies on for the same reason. This matters because the check is per shard
core, not per buffer -- up to 130 lookups for a full L1 grid.

**The DRAM `y == 0` check stays separate**, ahead of the lookup. A DRAM core off row
0 is a real coordinate (logical y indexes a view's subchannels), so "no bank here"
would describe it misleadingly and would drop the part that matters: banks are keyed
`{bank_id, 0}`, so such a core aliases onto bank x and corrupts it.

**This is a behaviour change for L1.** The L1 bank map covers only
`AllocCoreType::ComputeAndStore` cores, so it is a proper subset of the Tensix grid --
dispatch cores are excluded. Sharding L1 onto a dispatch core previously succeeded,
because sharded L1 I/O addresses cores by NOC coordinate and never resolves a bank id;
a raw write lands on L1 the allocator does not manage. That case now throws. I did not
find any existing test relying on it, but this is the part of the change most worth a
second opinion.

It has no test of its own either. `ShardGridValidationL1RejectsCoreOutsideGrid` reaches
a dispatch core only incidentally, and only on parts where dispatch takes the row or
column immediately past the compute grid -- elsewhere its coordinates are off the Tensix
grid entirely. Say the word and I will add a test that finds a dispatch core explicitly.
It is also in the merge gate's blind spot -- see below -- so the queue is otherwise the
first place it gets exercised.

### Testing

`test_shard_grid_validation.cpp`, 7 tests covering all three buffer types and both
failure modes, deriving coordinates from the device rather than hardcoding them
(both grids shrink under harvesting, so a core that is invalid on a harvested part
is legal on a stock one). The two L1_SMALL tests construct without allocating, via
the explicit-address `CreateBuffer` overload, because the fixture's device has no
small region -- an allocation would fail on that alone and say nothing about the
grid.

Run on a Blackhole Galaxy with one GDDR channel soft-harvested per chip, 7 of 8
banks, all 32 chips:

- `unit_tests_api`, filtered to `*Shard*:*Buffer*:*Alloc*:*L1*:*Dram*:*DRAM*` --
  133 passed, 0 failed. Every rejection logged traced to the new tests' own
  `EXPECT_ANY_THROW`; no other suite tripped the check.
- `test_reshard.py` -- 16 failed, 93 passed, 32 skipped. The 16 are exactly the
  set that already failed on this configuration, now all failing at
  `buffer.cpp: allocator.has_bank(...)` with `shard core (7, 0) has no DRAM bank
  on this device, which has 7 of them`. No test that passed before is now
  rejected, and the 6 that previously corrupted data silently now fail loudly at
  construction.

The first revision of this PR was dequeued from the merge queue by the L1_SMALL
case described above -- `Conv2dConfigVariations/Conv2dOpIfTest.Conv2d/{0,1}` in
`ttnn-basic-tests`, failing identically on wh_n150 and bh_p150b. Worth knowing for
anyone reading a green PR gate here: `merge-gate.yaml`'s seed jobs carry
`github.event_name != 'pull_request'` and everything `needs:` them, so
`ttnn-basic-tests` and its siblings report `skipping` on a PR and only really run
once enqueued.

After the fix, on an N300:

- `unit_tests_api --gtest_filter=*ShardGridValidation*` -- 7 of 7 pass.
- `unit_tests_ttnn --gtest_filter=*Conv2dOpIfTest*` -- 2 of 2 pass, i.e. the tests
  that dequeued it.
- `unit_tests_ttnn --gtest_filter='*QueryOp*:*Conv2dOpIfTest*'` -- 41 passed, 2
  failed. The 2 are `SoftmaxOpIfTest.Softmax/{0,1}` on
  `cb_peak_size_per_core == 0`; they are gated to `BoardType::N300` so CI on n150
  skips them, and they are not this change: the new assert text does not appear in
  the run and the preceding `query.status == Success` expectation passes, so nothing
  threw.

Not yet run against an unharvested part, where every grid is full and the check
should be inert.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jbauman/validate-dram-shard-grid-bank-count)


sidnadTT (approved):

abhullar-tt (approved):

pavlejosipovic:
Let pls put a proper codeowenr for tests/ttnn/nightly/unit_tests/operations/experimental/test_moe_gpt_e2e.py

pavlejosipovic (approved):

sraizada-tt (approved):

jonathansuTT (approved):

PR #53335, opened 2026-08-17
[skip ci] fix(docs): keep Python API list tables visible

### LINKED ISSUE #53333: [docs]: Figma theme CSS hides Python API list-tables
## Issue Metadata

**Component / Area:** docs
**Issue Type:** Other

### Observed

On a local `make html` (and any build with the custom top nav), dtype/layout `list-table`s in TTNN API pages are invisible. View Source still has a correct 3-column table. The rendered page only shows the surrounding Note text.

Cause: `docs/source/common/_static/tt_theme.css` does:

```css
body:has(.tt-top-nav) .rst-content dl.py.function table.docutils { display: none !important; }
```

The comment says `api_style.js` replaces those tables. That script is in still-open #49958. Layout #49959 already injects `.tt-top-nav` (which trips the CSS) and a `<script>` for `docs-toc.js` (404s locally: `GET /ttnn/_static/docs-toc.js`). `api_style.js` is not loaded at all.

Even if #49958 merged as written, it would not fix this. `api_style.js` restyles C++ param tables and autosummary cards. It does not restyle Python `Note` `list-table`s, so `dl.py.function table.docutils` stays hidden.

Related:
- #49957 (2026-07-23) introduced the hide rule, gated on `.tt-top-nav`
- #49959 (2026-07-24) shipped the nav without the companion scripts
- #49958 still open (`api_style.js` + `docs-toc.js`)

### Expected

Dtype/layout tables in API Notes stay visible on local `make html` / `make server`. Hide-and-replace should only apply to tables the JS actually transforms, and only when that JS is loaded.

### 1. Steps (exact commands)

1. `cd docs && make html && make server`
2. Open `http://localhost:8888/ttnn/ttnn/api/ttnn.experimental.gelu_bw.html` (any op with a Note `list-table` works)
3. Note section shows the heading and following sentence; no table
4. View Source: `<table class="docutils">` with Dtypes / Layouts / Ranks is present
5. DevTools: the rule above is `display: none !important`

Do not click TT-NN in the top nav — that URL is hardcoded to `docs.tenstorrent.com`.

n/a — docs build only. Example page: `ttnn.experimental.gelu_bw` (any Python API page whose docstring contains `.. list-table::` under `Note:`).

### 3. Frequency

Always, on any page with `.tt-top-nav` plus a `dl.py.function` `list-table`.

### 1. Software Versions

Local Sphinx `make html` from current main. No device required.

### 2. Hardware Details

n/a

## Regression Info

**Is this a regression?** Yes

- First bad: #49957 (hide rule). Fully triggered once #49959 added `.tt-top-nav`.
- Last known good: docs builds without `.tt-top-nav` (pre-#49959), where the `:has()` gate does not apply.

## Suggested fix

Stop hiding Python docstring tables. Drop `dl.py.function` / `dl.py.class` from that rule, or exclude `.admonition table`. Do not treat landing #49958 as the fix; it neither loads `api_style.js` from the merged layout nor handles these tables.

This is a docs-frontend issue, not a docstring/content bug.

breezeFur:
Fixes #53333

## Summary

- Keep Python API list-tables visible in the custom documentation shell.
- Continue hiding only C++ docutils tables that are replaced by `api_style.js`.

## Root cause

The shared CSS selector hid `dl.py.function` and `dl.py.class` tables whenever `.tt-top-nav` was present. The custom JavaScript does not replace Python docstring list-tables, so those tables disappeared from rendered pages.

## Validation

Before:
image

After:
image

### DIFF
--- a/docs/source/common/_static/tt_theme.css
+++ b/docs/source/common/_static/tt_theme.css
@@ -1323,11 +1323,9 @@ h2.tt-api-section-heading {
   margin: 0 !important;
 }
 
-/* Hide raw docutils table (replaced by api_style.js list) — gated on the custom
-   shell that loads api_style.js so plain builds keep native API tables. */
-body:has(.tt-top-nav) .rst-content dl.cpp.function table.docutils,
-body:has(.tt-top-nav) .rst-content dl.py.function table.docutils,
-body:has(.tt-top-nav) .rst-content dl.py.class table.docutils {
+/* Hide only the C++ tables replaced by api_style.js. Python API tables are
+   rendered natively and must remain visible. */
+body:has(.tt-top-nav) .rst-content dl.cpp.function table.docutils {
   display: none !important;
 }
 


breezeFur:
Closing this because upstream PR #49958 has now merged and supersedes this change. It removes the table-hiding rules entirely and moves the C++ parameter-table transformation to build time, which also leaves Python list tables visible as requested by #53333. Rebasing this one-line CSS fix would reintroduce a C++ hide rule that the new build-time implementation no longer needs.

PR #53352, opened 2026-08-17
[Feature] [Codegen] Guard blind regeneration runs against git-history recovery

jmacanovicTT:
### Summary

A blind run (`HIDE_EXISTING_KERNEL`, `REMOVE_TESTS`) deletes the target op's files and
commits that on the worktree branch, so the agent finds nothing at `HEAD`. But a worktree
shares the parent repo's object store, so the deleted code is still reachable —
`git show HEAD~1:<path>`, `git rev-list --all --objects`, or reading `.git/logs/HEAD`
recovers it. The only protection was an instruction in the prompt, and afterwards there
was no way to tell whether a run had looked.

Adds enforcement plus a per-run audit trail:

- **`.claude/hooks/git-guard.sh`** — `PreToolUse` hook on `Bash|Write|Edit`. Logs every
  command with an `OK`/`BLOCK` verdict; in blind mode denies history reads (`log`, `show`,
  `reflog`, `rev-list`, `cat-file`, `blame`, `stash`, `fsck`, `archive`), older-revision
  references (`HEAD~`, `@{`, bare SHA), and any `.git` path. It also scans **written
  scripts**, so a git command cannot be smuggled into a file and then executed.
- **Two `settings.json`** (`codegen/`, `tt-llk/`) — wire the hook and deny
  `Read(//**/.git/**)`, covering the `Read`/`Grep`/`Glob` path the hook cannot see.
- **`orchestrator_steps.sh`** — the hide step arms the guard, and **fails before hiding
  anything** if it cannot; the transcripts step copies the log to `$LOG_DIR/git-guard.log`
  and records `GUARD_BLOCKS` in `state.json`.
- **`tt-llk/CLAUDE.md`** — documents the policy.

Auditing a run is one field:

| `GUARD_BLOCKS` | meaning |
|---|---|
| `0` | the guard ran and the run never reached for history |
| `n` | `n` attempts — named in `git-guard.log` |
| `null` | not measured — **not** evidence of a clean run |

**Default is off.** Outside a blind run the hook only logs, so the router keeps the
read-only git it documents. `git status`/`add`/`commit`, bare `git diff`, and
`git worktree add -b <branch>` stay allowed even in blind mode — the pipeline needs them.

### Notes for reviewers

**Marker file, not an env var.** `HIDE_EXISTING_KERNEL` is chosen after Claude starts, and
you can't change a running process's environment. An `export` in the pipeline wouldn't
survive either — each Bash call is a fresh shell. So the hide step touches
`$TMPDIR/codegen-blind-run-<session-id>` and the hook keys off that.

**Two settings files.** Settings load from the launch directory. Runs launch from
`codegen/`, and `tt-llk/.claude/settings.json` is verifiably not picked up there — the test
command returned `RAN` before `codegen/.claude/settings.json` existed, `BLOCKED` after.

**Threat model.** This stops accidental and casual history access — the failure mode we
actually see. It does not defeat deliberate evasion: a command assembled letter-by-letter,
base64-decoded, or built from concatenated strings passes any text check, wherever that
check lives. Verified, and deliberately not chased — heuristics there buy false confidence
rather than coverage. `GUARD_BLOCKS` and the command log are the durable part, and the
structural fix (a checkout whose history never contained the hidden files) is the only
version that is immune.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jmacanovic/codegen-git-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jmacanovic/codegen-git-guard)

### DIFF
--- a/tt_metal/tt-llk/.claude/hooks/git-guard.sh
+++ b/tt_metal/tt-llk/.claude/hooks/git-guard.sh
@@ -80,7 +80,10 @@ SUBCMD='(^|[^[:alnum:]_-])git([[:space:]]+(-[^[:space:]]+|-C[[:space:]]+[^[:spac
 # bare-SHA pattern is what stops `git worktree add <dir> <old-sha>` and
 # `git checkout <old-sha> -- <path>`; `worktree` itself stays allowed because the
 # pipeline creates its own worktrees.
-REVREF='HEAD~|HEAD\^|@\{|\.git($|[^[:alnum:]_-])|(^|[^[:alnum:]])[0-9a-f]{7,40}([^[:alnum:]]|$)'
+#
+# Bare SHAs are handled by _has_sha_token below rather than by this pattern — see there for
+# the two false-positive classes the first real run exposed.
+REVREF='HEAD~|HEAD\^|@\{|\.git($|[^[:alnum:]_-])'
 
 # Same, minus the bare-SHA clause, for scanning file content: a 7+ hex-digit token is
 # common in source (constants, masks) and would false-block legitimate writes.
@@ -94,6 +97,21 @@ SUBCMD_LOOSE='(^|[^[:alnum:]_-])git[^[:alnum:]_]+(log|reflog|rev-list|cat-file|b
 
 # Cheap gate: only inspect text that mentions git or a .git path at all.
 GITISH='(^|[^[:alnum:]_-])git([^[:alnum:]_-]|$)|\.git($|[^[:alnum:]_-])'
+
+# A bare commit SHA, as an argument rather than as a fragment of something else. Two
+# false-positive classes from the first real run drove this shape:
+#   * hex inside a path — run ids are 8 hex chars, so every command naming LOG_DIR
+#     (".../2026-08-20_relu_quasar_8c1bcf65/...") matched. Hence: whitespace-delimited only.
+#   * pure-decimal numbers — every digit is a valid hex digit, so Confluence page ids like
+#     1173618806 matched. Hence: the token must contain at least one a-f.
+# A short SHA that happens to be all digits is therefore missed; that is an accepted gap,
+# since this clause is only a backstop behind SUBCMD and the HEAD~/.git patterns.
+_has_sha_token() {
+    printf '%s' "$1" \
+        | grep -oE '(^|[[:space:]])[0-9a-f]{7,40}([[:space:]]|:|$)' \
+        | tr -d '[:space:]:' \
+        | grep -qE '[a-f]'
+}
 
 # Writing a script and then running it defeats a command-text check: the Bash hook only
 # sees `bash x.sh`. Authoring the script *through bash* is already caught (the git text is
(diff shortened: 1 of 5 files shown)


tt_metal/tt-llk/CLAUDE.md
```diff
@@ -6,6 +6,8 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 Read-only git commands are allowed (`git rev-parse`, `git log`, `git status`, `git diff`, `git show`). **NEVER push, commit, checkout, restore, reset, or otherwise modify** the repo via git.
 
+**Blind codegen runs are stricter.** A run that exports `CODEGEN_BLIND_RUN=1` regenerates an implementation that was deliberately hidden, so recovering it from history would invalidate the result. In that mode the `PreToolUse` hook `.claude/hooks/git-guard.sh` denies history reads (`log`, `show`, `reflog`, `rev-list`, `cat-file`, `blame`, `stash`, `fsck`, `archive`), any reference to an older revision (`HEAD~`, `HEAD^`, `@{`, a bare SHA), and direct reads under `.git/`. `git status`, `git add`, `git commit`, a bare `git diff`, and `git worktree add` on a branch stay available. Outside a blind run the hook only logs commands and the policy above applies unchanged. Every bash command is logged either way — set `CODEGEN_GUARD_LOG` to keep that log with the run's artifacts.
```
njokovicTT: This is only explanation of what has been changed in the PR. I don't see how will this affect agent's decision on what to do.
I would remove this

tt_metal/tt-llk/.claude/settings.json
```diff
@@ -0,0 +1,24 @@
+{
+    "$schema": "https://json.schemastore.org/claude-code-settings.json",
+    "permissions": {
+        "deny": [
+            "Read(//**/.git/**)",
+            "Read(//localdev/jmacanovic/tt-metal/.git/**)"
```
njokovicTT: I think you can remove this, since you have it covered in the line above
`"Read(//**/.git/**)"`

tt_metal/tt-llk/.claude/hooks/git-guard.sh
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env bash
```
njokovicTT, line 1: FYI, this will only work for Claude Code harness. We can later change this and port it to custom implementation such as using `syslog-ng` since we are running on Linux machines

tt_metal/tt-llk/.claude/settings.json
```diff
@@ -0,0 +1,24 @@
...
+        ]
+    },
+    "hooks": {
+        "PreToolUse": [
```
njokovicTT, line 9: There is a potential workaround to this that agents can try and execute and that is creating a temporary script that contains simple git  command that will output the history.
One of the ways to avoid it could be by using `PostToolUse` and altering `updatedToolOutput` by removing predefined hidden commit name in one of the orchestrator steps

PR #53354, opened 2026-08-17
examples: consolidate the perf catalog and add distribution_gate

dstoiljkovicTT:
`examples/master.md` becomes one catalog in two parts — Part 1 the realized, measured examples; Part 2 the lever propositions, absorbing the standalone data-movement checklist. Levers that have an example now link to it.

**New example:** `distribution_gate` (kernels, CLI, README, measured report, test). A height split strands a wide-short tensor on one core and a width split strands a tall-narrow one — the trap is symmetric, so what fills the grid is gating the axis on the shape. Reproduced on a real op: tilize run 923 measures **+272%** with the gate off (110 cores → 1).

**Catalog changes.** A0 becomes a per-regime correctness check: the bandwidth knee is a hard core-count ceiling, and only for *bandwidth*-bound ops — issue-rate and small-page ops still want the full grid. B0 notes that levers adding fixed per-core setup are a per-regime tradeoff, not a free win. New group F covers precision cost — the knobs whose default is the expensive one.

`tile_reorder` is removed — its lesson, coalesce into whole-page transactions, is lever B5, and the checklist it was paired with is now Part 2. Its Part 1 entry and B5's example pointer go with it, so every catalog link still resolves.

Ported from `dstoiljkovic/llk_helper_tm_nuked` (13 commits). Excluded: the submodule bumps riding along in those commits, and a `golden_functions` guard that exists only because that branch nukes tilize. Two of that branch's examples are also left out, both misses against real op runs — `dram_saturation`, whose bandwidth knee did not transfer to tilize (a small-page, issue-rate-bound op wants the whole grid, and the lever it backs went unmeasured), and `zero_copy_fold`, whose fold measured slower than the separate reader/compute/writer on every shape. The A0 knee and the C14 aliasing text stay as propositions; only their example pointers go.

17 files, +987/−424. Examples and docs only — no library or op code touched.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dstoiljkovic/perf-examples-catalog)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dstoiljkovic/perf-examples-catalog)


ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,3 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+  layout match; fall back to the generic streaming factory only when nothing matches.
+  `untilize/device/untilize_device_operation.cpp:285,310,315-316,346-349`.
+- **D21. Precompute per-core indexing host-side; `InterleavedAddrGenFast` (shifts, not multiplies) for
+  pow2 pages** *(proposition)*. `untilize_multi_core_program_factory.cpp:335,396`;
+  `dataflow_api_addrgen.h:349`.
+
+## E. Host-dispatch overlap (whole-model level)
+- **E22. Metal Trace + multiple command queues + events** *(proposition — whole-model, usually outside
+  perf-lab's single-op scope)* — remove per-op host dispatch; overlap input I/O (CQ1) with execution
+  (CQ0). `AdvancedPerformanceOptimizationsForModels.md:33,157,161`.
```
s-jovic, line 473: Do we need this for op codegen?

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,3 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+| Args | runtime address-gen | compile-time `TensorAccessorArgs`, cached program |
+| Precision knobs | precise packer / fp32 DEST / HiFi4 on by default "to be safe" | cheapest config that clears the measured accuracy gate; caller's contract untouched |
+
+## Notes
+- Sections A–B are grounded in `Saturating_DRAM_bandwidth.md` (>92% DRAM BW on Wormhole). Use
+  `/perf-ceiling-dm` to turn a proposed transfer scheme into a predicted target and `/perf-measure` to
+  measure the real number on device.
+- Some ops referenced by earlier drafts (`interleaved_to_sharded`, `sharded_to_interleaved`, `reshard`)
+  are absent on this branch (nuked for agent eval); the surviving `untilize` / `tilize` / `transpose` /
+  `concat` / `move` factories illustrate the same host-side patterns.
```
s-jovic, line 546: This should be removed

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,3 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+  `untilize_multi_core_program_factory.cpp:103-116`; `tilize/device/tilize_device_operation.cpp:22`.
+  Aliasing has two
+  degrees: removing the *NoC traffic* (the CB is the shard) and removing the *kernel* (fold the
+  dataflow away entirely, so the program is compute-only). The second is a separate, measurable step
+  — a resident path whose reader still exists to run the CB handshake has taken only the first.
+- **C15. Prefer sharded (L1-resident) over interleaved for DRAM-bound ops** *(proposition)* — each reader
+  its own bank, >92% BW vs interleaved congestion. `Saturating_DRAM_bandwidth.md`.
+- **C16. Double-buffer CBs (depth 2) — but only when it pays** — single-block cores skip it to save L1.
+  `concat/device/concat_program_factory.cpp:111`; `untilize_multi_core_program_factory.cpp:132`.
+  **→ example: `double_buffer`**.
```
s-jovic, line 454: let's not reference ttnn ops

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,3 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+stateful NoC B13, shard-aligned core groups) add a fixed per-core setup/issue cost that pays off when
+each core has enough work to amortize it — and *regresses* the smallest-shard / lowest-work-per-core
+regime, where that fixed cost dominates the ~1–8 tiles of real work. So a lever's counterfactual
+(Mode C) must be measured on the **smallest regime it will run in**, not only the aggregate or a large
+shape: a lever that is net-positive on big/interleaved shapes can be net-negative on tiny sharded
+ones. "Missed lever" (Mode D) is real headroom **only in the regime the lever would actually run** —
+gate it on a work-per-core threshold rather than applying it globally.
+
+- **B5. Coalesce into whole-page transactions; don't scatter sub-tile faces** — bigger transactions hit
+  higher achieved BW. `dataflow_api.h:566`.
```
s-jovic, line 416: I'd also avoid references to code that will become obsolete very soon

ttnn/ttnn/operations/examples/master.md
```diff
@@ -1,22 +1,30 @@
-# Performance examples — catalog
+# Performance catalog — examples + propositions
```
djordjenTT, line 1: In general, this file is becoming big and hardly maintainable. It should serve a single purpose - an agent should use it to answer "is there something interesting for my problem that I can look up in more detail in a per-example dedicated .md?" So we should heavily filter the entries only to answer "what the example is about" and "when to consider it" - the measurements (expressed in ns - I guess it is ok to keep relative speedups) hardly mean anything (eg. "gated 90665 ns vs height 90666 ns"). Possibly out of scope for this review, but we need to address this.

ttnn/ttnn/operations/examples/master.md
```diff
@@ -1,22 +1,30 @@
...
-concepts and **measure** them on device (real `ns`, never a claimed speedup). Use them to learn a
-pattern, then re-measure on your own shapes with each example's CLI / test.
+One catalog, two parts:
 
-**Reading order:** this file → the example's `README.md` → read the code and run the test only if
-you need to. For the ⭐ Starter examples, the *gist* below is often enough to act immediately.
+- **Part 1 — Realized examples**: short, self-contained, **runnable** ops that each isolate one or two
+  kernel-level performance concepts and **measure** them on device (real `ns`, never a claimed speedup).
+  Learn a pattern, then re-measure on your own shapes with the example's CLI / test.
+- **Part 2 — Propositions**: the cross-codebase "optimal vs non-optimal" data-movement lever checklist —
```
djordjenTT, line 8: To me, the Part 2 looks like a mixed bag, and I strongly propose spilling the parts where they truly belong.
- Examples that are built (eg. pointing to noc_placement example)
- Examples that are not (yet) built - let's build them (eg. a bunch of them pointing to Untilize op, which may or may not be nuked from the repo)
- General performance instructions - TBH I can't justify why those belong here and not in the (blocking) agent definitions, or an .md (in tt_ops_codegen) repo that those agents refer to (eg. blocking-mode.md where blocking doctrine is being explained).
This approach has a big potential to create a maintenance nightmare (eg. "F24." referenced from a matching change in tt_ops_codegen repo), and I just can't see anything solid enough to justify that.

There are other classes as well (eg. instructions that are already covered somewhere else). Claude can help with that eg. this is what I got (up to debate of course) after a discussion about Marko's recent changes followed by a discussion about this PR.
https://claude.ai/code/artifact/bdf94963-c7bb-4801-8f59-fb8b67cb6bba

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,32 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+  FIRST**, because the knee only exists for a **bandwidth-bound** op:
+  - **Bandwidth-bound** (big pages / coalesced reads that actually saturate DRAM/L1 BW) → the knee is a
+    **hard ceiling**: once achieved bandwidth plateaus,
+    extra cores add no bandwidth and only add dispatch/NoC overhead. If the knee is **below**
+    `total_tiles`, stop at the knee — more cores past it can be *slower*, not just wasteful.
+  - **Read/write-transaction-rate bound** (small pages — e.g. a 32-row stick at ≤128 B/page, the
+    typical layout-conversion reader) → there **is no reachable knee**: the op cannot hit DRAM
+    bandwidth at *any* core count because it is issue-rate limited, and the sync/dispatch floor scales
+    with **blocks-per-core**, so shedding cores *adds* cost. Here `bandwidth_knee = full grid` — use
+    **all** the cores. **Do not cap.** (Measured on tilize: applying a 16-core knee cap was **~2.4×
```
djordjenTT, line 379: This should not be here - we should build an op-agnostic example if we don't have it.

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,32 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+on WH) and `tech_reports/AdvancedPerformanceOptimizationsForModels/AdvancedPerformanceOptimizationsForModels.md`
+(host-dispatch overlap, E). API surface: `tt_metal/hw/inc/api/dataflow/dataflow_api.h`.
+
+## A. Core / grid placement — the biggest single lever
+
+**A0. Active-core count is a per-regime correctness check, not a one-time choice.** Before the
+placement levers below, the *number* of active cores must be right for **every shape regime the op
+accepts** — a single distribution scheme can be optimal in one regime and pathological in another.
+The criterion:
+- **interleaved** → `active_cores == min(grid, total_tiles, bandwidth_knee)` — **but classify the bound
```
djordjenTT, line 369: A good example for my general comment - in my understanding, the "bandwidth knee" procedure is a guidance to the planner and kind of fits next to the split ranking in blocking-model.md (introduced by #140). But then it seems that Marko deleted the occupancy-based guidance in #156 though I need to understand why) so this might be conflicting with that now. Different repo, different name, no reference to that deletion here. A good illustration for why we need to keep these things consolidated.

djordjenTT (changes requested):

ttnn/ttnn/operations/examples/master.md
```diff
@@ -321,32 +338,209 @@ tile-index `reduce_scatter_mcast` wins instead (**~2×** over root) by paralleli
...
+below (coalesced-block reads B5, deferred barriers B7, trid double-issue B8, per-reader VCs B10,
+stateful NoC B13, shard-aligned core groups) add a fixed per-core setup/issue cost that pays off when
+each core has enough work to amortize it — and *regresses* the smallest-shard / lowest-work-per-core
+regime, where that fixed cost dominates the ~1–8 tiles of real work. So a lever's counterfactual
+(Mode C) must be measured on the **smallest regime it will run in**, not only the aggregate or a large
+shape: a lever that is net-positive on big/interleaved shapes can be net-negative on tiny sharded
+ones. "Missed lever" (Mode D) is real headroom **only in the regime the lever would actually run** —
+gate it on a work-per-core threshold rather than applying it globally.
+
+- **B5. Coalesce into whole-page transactions; don't scatter sub-tile faces** — bigger transactions hit
```
djordjenTT, line 415: We removed the example (tile_reorder) from index (but the actual code still stays!) and replaced it with B5 prose. Is it because you considered that example to be too weak to be a separate entity?

Issue #49625, opened 2026-07-10
[Bounty $2,500] Optimise/improve accuracy for pow(x, y) fp32 (non-integer exponent)


aliraza556:
## Summary

The `ttnn.pow(x, y)` fp32 implementation (the `exp_21f`-based path added in PR #25025, which closed #23529) was a big improvement over the old kernel, but it **still lost up to 27 ULP for non-integer exponents**, while integer exponents stayed bit-exact (0 ULP). The op is the composite `y * log2(x)` followed by `2^(...)`, so the fp32 log2 and exp2 rounding errors compound. The CogVideo case that motivated #23529 (`10000 ** 1.7984`) was exactly one of those weak spots.

This PR is a **follow-up accuracy improvement on top of the accepted #25025 fix**, in the same class as the transcendental accuracy bounties exp2 #46006, expm1 #45046, atan2 #41029. It brings **all** tested non-integer exponents to within 3 ULP — including the long-mantissa `1.7984` CogVideo case (19 → 2 ULP) — while keeping integer exponents bit-exact.

## Current state and fix (measured on Wormhole B0)

Dense sweep, 1000 log-spaced bases `x ∈ [0.5, 50000]`, `fp32` input, `fp32_dest_acc_en=true`, `math_approx_mode=false`. Golden = `torch.pow` (fp32). Both columns are hardware-measured on Wormhole B0 with a confirmed clean JIT recompile: "before" from the unmodified #25025 kernel, "after" with this fix applied.

| exponent `y`      | Max ULP (before) | Max ULP (after) | after: points > 3 ULP | after: points > 10 ULP |
| ----------------- | ---------------: | --------------: | --------------------: | ---------------------: |
| 2.0  (integer)    |            **0** |           **0** |               0 / 1000 |               0 / 1000 |
| 3.0  (integer)    |            **0** |           **0** |               0 / 1000 |               0 / 1000 |
| 0.5               |                4 |           **1** |               0 / 1000 |               0 / 1000 |
| 1.5               |               19 |           **2** |               0 / 1000 |               0 / 1000 |
| 1.7984 (CogVideo) |               19 |           **2** |               0 / 1000 |               0 / 1000 |
| **2.5**           |           **27** |           **3** |               0 / 1000 |               0 / 1000 |

`2.5` has a single point at exactly 3.0 ULP (p99 = 2.0); every other non-integer exponent is ≤ 2 ULP. bf16 is unaffected (≤ 2 ULP, unchanged).

[...]

PR #53367, opened 2026-08-17
[Bug fix] ttnn.pow(0, p) returns 2^(-127p) instead of 0 (#51468)

### LINKED ISSUE #51468: `ttnn.pow(0, p)` returns `2^(-127p)` instead of `0` for every `p` in `(0, 1)` — the zero-base special case was written for negative exponents only
# `ttnn.pow(0, p)` returns `2^(-127p)` instead of `0` for every `p` in `(0, 1)` — the zero-base special case was written for negative exponents only

Both `pow` SFPU kernels compute `x**y` as `2**(y · log2(x))`, and both obtain `log2` by a bit-field
range reduction (`setexp` / `exexp`) that has no meaning for `x = 0`. `setexp(0, 127)` produces the bit
pattern `0x3F800000` = `1.0f` and `exexp(0)` returns `0 - 127 = -127`, so **`log2(0)` is computed as the
finite value `-127.0`** and the kernel goes on to return `2^(-127·p)`.

For `p >= 1` that lands at or below `2^-127` and washes out; for `0 < p < 1` it does not. Every exponent
in `(0, 1)` returns a wrong, finite, plausible-looking number:

```
ttnn.pow(0.0, 0.0001) = 0.99123567     torch: 0.0
ttnn.pow(0.0, 0.01)   = 0.41465977     torch: 0.0
ttnn.pow(0.0, 0.25)   = 2.7688385e-10  torch: 0.0
ttnn.pow(0.0, 0.5)    = 7.6664670e-20  torch: 0.0     <- this is sqrt(0)
```

The last line is the one that matters most: `ttnn.sqrt(0.0)` returns exactly `+0.0` (bits `0x00000000`),
but `ttnn.pow(0.0, 0.5)` — the same mathematical value, same library, same op family — returns
`0x1FB504F3`. Anything downstream that relies on `x == 0 ⇒ sqrt(x) == 0` (norms, distances, RMS,
variance, `1/sqrt`) silently gets `1.3e19` instead of `inf`, or a nonzero denominator where it expected
a guarded zero.

**The author did consider `base == 0`, and implemented exactly half of it.** Both kernels contain a
special case, and both cover only `pow < 0`:

`ckernel_sfpu_binary_pow.h:256` (fp32 path; `:123` is the same guard in the bf16 path, spelled
`v_if((absbase == 0.f) && pow < 0.f)`):

```c
    // Division by 0 when base is 0 and pow is negative => set to NaN (only for negative exponents)
    v_if(base == 0.f && pow < 0.f) {
        y = std::numeric_limits<float>::quiet_NaN();
    }
    v_endif;
```

`ckernel_sfpu_unary_power.h:118-123` and `:299-304` go further and compile the check *out* when the
exponent is non-negative, with a doc comment at `:30` stating the assumption:

```c
/**                                                                    // :30
 * @tparam IS_POSITIVE_EXPONENT If true, assumes exponent >= 0 (skips zero-base check for optimization)
 */
...                                                                    // :118
    if constexpr (!IS_POSITIVE_EXPONENT) {
        v_if(abs_base == 0.f) {
            y = sfpi::vConstFloatPrgm2;  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
        }
        v_endif;
    }
```

`IS_POSITIVE_EXPONENT` is selected inside the LLK itself, by `if (pow_scalar >= 0.0f)` in
`_sfpu_unary_power_bf16_` / `_sfpu_unary_power_fp32_` (`:335`, `:358`) — a scalar branch evaluated on
the TRISC at kernel entry, with `pow_scalar` coming from `Converter::as_float(exponent)`. So for
`ttnn.pow(x, 0.5)` there is no zero-base handling in the emitted kernel at all. The stated premise —
*exponent >= 0 means the zero base needs no special treatment* — is false for every `p` in `(0, 1)`,
which is exactly the range `sqrt`, `cbrt` and RMS-style normalisations live in.

`pow(0, p<0) -> NaN` is **documented behaviour** (`ckernel_sfpu_binary_pow.h:34`, "base = 0, pow < 0:
Returns NaN (undefined)"), and this proposal does not ask for it to change. The missing half is
`base == 0 && pow > 0 -> +0`.

### Which public calls are affected

All four `ttnn.pow` overloads reach one of the two broken kernels, plus `ttnn.rpow`:

| call | dispatch | kernel |
|---|---|---|
| `ttnn.pow(t, 0.5)` | `binary_composite_op.cpp:714` (non-integral) → `ttnn::power` → `UnaryOpType::POWER` → `power_tile` | `ckernel_sfpu_unary_power.h` → `_sfpu_unary_power_61f_updated_<true>` / `_sfpu_unary_power_21f_<true>` |
| `ttnn.pow(t, 4)` | `binary_composite_op.cpp:728` (integral, not 0–3) → `ttnn::power` | same as above |
| `ttnn.pow(t, exp_tensor)` | `binary_composite_op.cpp:742` → `invoke_binary_ng(POWER)` | `ckernel_sfpu_binary_pow.h` → `_sfpu_binary_power_<...>` |
| `ttnn.pow(0.0, exp_tensor)` | `binary_composite_op.cpp:764` → `full_like` → tensor–tensor | `ckernel_sfpu_binary_pow.h` |
| `ttnn.rpow(t, 0.0)` | `unary.cpp:208` → `RPOW` → `calculate_rpow` | `ckernel_sfpu_rpow.h` → `_sfpu_binary_power_<...>` |

`ttnn.pow(t, 0)`, `(t, 1)`, `(t, 2)`, `(t, 3)` are routed to `power_iterative` (exact multiply loop) and
are correct — which is a large part of why this survived.

### Measured

`pow(0.0, p)`. Truth is exactly `+0.0` for every `p > 0` in IEEE-754, so the absolute error equals the
returned value. Both dest-accumulate paths and both dispatch paths are shown; the four are independent
algorithms (Moroz `exp_21f` vs. Dekker/Veltkamp + Cody-Waite) that fail identically because they share
the `setexp`/`exexp` range reduction.

| exponent `p` | unary fp32 | unary bf16 | binary fp32 | binary bf16 | torch |
|---:|---:|---:|---:|---:|---:|
| 1e-4 | **0.991235673** | 0.98828125 | 0.991235673 | 0.98828125 | 0 |
| 0.01 | **0.414659768** | 0.416015625 | 0.414659768 | 0.416015625 | 0 |
| 0.05 | 0.0122591257 | 0.0122680664 | 0.0122591257 | 0.0122680664 | 0 |
| 0.25 | 2.76883849e-10 | 2.76486389e-10 | 2.76883849e-10 | 2.76486389e-10 | 0 |
| 1/3 (`cbrt`) | 1.80466454e-13 | 1.81188398e-13 | 1.80466454e-13 | 1.81188398e-13 | 0 |
| **0.5 (`sqrt`)** | **7.6664670e-20** | 7.6656482e-20 | 7.6664670e-20 | 7.6656482e-20 | 0 |
| 0.75 | 2.12272101e-29 | 2.12992444e-29 | 2.12272101e-29 | 2.12992444e-29 | 0 |
| 0.99 | 1.41741914e-38 | 1.41426664e-38 | 1.41741914e-38 | 1.41426664e-38 | 0 |

`pow(0, 0.5)` is bit pattern **`0x1FB504F3`** = `7.6664669522108749e-20` — literally `float32(sqrt(2))`
with its exponent field rewritten to 63, i.e. `2^-63.5`. It is stated as bits because the shortest
9-digit decimals `7.66646695e-20` and `7.66646708e-20` both round to this same float32.

**Full sweep**, `p` over 200001 points in `[1e-5, 0.999]`:

| path | nonzero results | max returned value |
|---|---:|---:|
| unary fp32 | **200001 / 200001 (100%)** | 0.999120116 @ `p` = 1e-5 |
| unary bf16 | **200001 / 200001 (100%)** | 0.99609375 |
| binary fp32 | **200001 / 200001 (100%)** | 0.999120116 |
| binary bf16 | **200001 / 200001 (100%)** | 0.99609375 |

**This is not hidden by flush-to-zero.** The returned value is `2^(-127p)`, which is a *normal* float32
whenever `p < 126/127 = 0.99213`. Over the sweep, **198624 / 200001 = 99.31%** of the wrong results are
normal floats that no FTZ behaviour can turn back into zero. That is also the explanation for why the
`p >= 1` cases look fine: there the residue is `<= 2^-127`, subnormal, and disappears.

**The two paths already disagree at `base = 0`.** The binary fp32 kernel clamps to exactly `-127.0`
(`ckernel_sfpu_binary_pow.h:231`) and does return exact `0` for `p >= 1`. The unary fp32 kernel clamps
inside `_sfpu_pow2_f32_accurate_hilo_` to `-0x7e.ffff8p0` = `-126.99999237`, so it returns
`6.16571324e-44` for *every* `p >= 1` — never exact `+0.0`:

| call | unary fp32 | binary fp32 | torch |
|---|---:|---:|---:|
| `pow(0, 1.5)` | 6.16571324e-44 | 0 | 0 |
| `pow(0, 4.0)` | 6.16571324e-44 | 0 | 0 |
| `pow(0, 8.0)` | 6.16571324e-44 | 0 | 0 |

That residue is subnormal and very likely flushed on device, so I am *not* claiming it as an
observable bug — but it does mean the unary fp32 kernel has no path that produces exact `+0.0` for a
zero base, at any exponent.

`base = -0.0` behaves identically to `+0.0` (the `v_if(base < 0.0f)` sign branch does not fire, since
`-0.0 < 0.0` is false), so `pow(-0.0, 0.5)` is also `7.67e-20` instead of `0`.

### Root cause

Step by step for `_sfpu_unary_power_61f_updated_`, `base = 0.0f`, `pow = 0.5f`
(traced instruction-by-instruction; `ckernel_sfpu_unary_power.h:237-296`):

1. `abs_base = sfpi::abs(0.0f)` → `0x00000000`.
2. `m = sfpi::setexp(abs_base, 127)` — `setexp` writes only the 8-bit exponent field, leaving the
   mantissa at 0 → `0x3F800000` = **`1.0f`**. (The file's own comment at `:244-246` already notes that
   "setexp writes the 8-bit exponent field and wraps instead of saturating".)
3. `exp = sfpi::exexp(abs_base)` — default `ExponentMode::Unbiased`, so `0 - 127` = **`-127`**.
4. `m = 1.0` is below `SQRT2`, so no range-reduction adjustment. `z = (m-1)/(m+1) = 0` exactly, the
   `atanh` series gives `p = 1.0`, and `ln_m = 2·(0·1) = 0` **exactly**.
5. Therefore `log2(0)` is computed as `exp_f32 + ln_m·ln2inv = -127.0 + 0` = **`-127.0`, exactly**.
   `2^(-127p)` is a closed-form result here, not an approximation — which is why every one of the
   200001 sweep points is wrong rather than merely inaccurate.
6. `z_hi = pow_hi · exp_f32 = 0.5 · -127 = -63.5`, `z_lo = 0`. The only guard is
   `s = max(s, -126.99999237)`, which `-63.5` does not come near. `2^-63.5 = 7.666467e-20`.
7. The zero-base special case is `if constexpr`-ed out because `pow_scalar >= 0`.

The bf16/`exp_21f` path reaches the same place by different arithmetic: the 3rd-order minimax series
evaluated at `x = 1.0` is `((1·0x2.44734p-4 − 0xd.e712ap-4) + 0x2.4f5388p+0) − 0x1.952992p+0`, which is
**exactly `0.0f` in float32**, again giving `log2 = -127.0` exactly, then the `-127` clamp at `:77` is
missed for `p < 1`.

### Three sibling kernels in this repo already handle exactly this

This is not an unconsidered edge case elsewhere in the tree — it is handled everywhere *except* inside
`pow`'s inlined `log2`:

- **`ckernel_sfpu_log.h:49`** — the production `ttnn.log`, which does the same job (`log` of a possibly
  zero/subnormal input), normalises first and documents both classes:
  ```c
      // normalise a (-0.0 and subnormals become +0.0)
      a = a * 1.0f + 0.0f;
  ...
      // Handle special cases:
      //   input 0.0  -> -inf
      //   input +inf -> +inf
      //   input NaN  -> NaN
      a = sfpi::addexp(a, -1);                                  // :94
  ```
  `pow` needs `log2(0) = -inf` for the same reason `log` needs `log(0) = -inf`, and gets `-127` instead.
- **`ckernel_sfpu_sqrt.h`** — the plain-sqrt path returns exact `+0.0` for a zero input *by
  construction*: in the `SQRT_23` body `xy = 0`, so `y = one_minus_xyy * half_xy + xy = +0.0`. It
  needs no zero guard because the algorithm degenerates correctly. (The explicit `x_bits != 0`
  branches at `:40` and `:67` belong to the `if constexpr (RECIPROCAL)` path — they special-case
  `rsqrt(0) = inf`, not `sqrt`.)
- **`ckernel_sfpu_sqrt_custom.h`** (WH `:24`, BH `:18` — the two copies differ in layout, the WH one
  wrapping its refinements in a `NEWTON_ITERATIONS` loop) — a second, independent sqrt implementation,
  whose entire Newton-Raphson body sits inside `v_if(val != 0.0f)` with `out` pre-initialised to
  `val`, so a zero input is passed straight through as `+0.0`.

So `ttnn.sqrt(0) == 0` and `ttnn.log(0) == -inf`, while `ttnn.pow(0, 0.5) == 7.67e-20`. The library
disagrees with itself about the same special value.

### Test coverage

`tests/ttnn/unit_tests/operations/eltwise/test_pow.py` does contain a zero base — exactly one, and it
is paired with the one exponent that the `-127` clamp rescues. `test_special_input_fp32` (`:230`):

```python
a = torch.tensor([[1.0, 0.999, ..., 0.0, -1.0, ...]])   # index 8 is the only 0.0
b = torch.tensor([[0.999, 1.0,  ..., 1.0,  1.0, ...]])  # its exponent is 1.0
...
assert_with_ulp(torch_output_tensor, output, ulp_threshold=2)
```

Move that one exponent and the assertion fails immediately:

| `a[8]` | `b[8]` | result | torch |
|---:|---:|---:|---:|
| 0.0 | **1.0** (as tested) | 0 | 0 |
| 0.0 | 0.999 | 1.0816819e-39 | 0 |
| 0.0 | 0.5 | 7.6664670e-20 | 0 |
| 0.0 | 0.25 | 2.7688385e-10 | 0 |

The rest of the file cannot see it either:

- `test_pow_arange_masking_fp32` (`:418`) is parametrised over exactly the broken range —
  `[0.25, 0.5, 0.75, -0.25, -0.5, -0.75]` — but its input never contains a zero: it comes from the
  shared `test_unary_pow.py::generate_clean_bf16_tensor`, whose mask
  `torch.isfinite(fp32) & ~neg_zero_mask & (fp32.abs() >= tiny)` strips `±0` and all subnormals
  before the tensor reaches this file. Its own `in_range` filter and the `flush_subnormal_values_to_zero`
  call at `:421` narrow further, but the zeros are already gone by then. Its assertion is also
  `assert_allclose(atol=5e-4, rtol=8e-7)`, so even if a zero were injected, the wrong results
  (`2.77e-10`, `7.67e-20`, `2.12e-29`) all sit well inside that tolerance and it would still pass.
- `test_unary_pow.py::test_pow_arange_masking` parametrises `[0.0, 1.0, 2.0, 3.0, -1.0]` — no
  exponent in `(0, 1)`. `test_eltwise_pow_float.py` draws bases from `uniform(-100, 100)` and asserts
  `assert_with_pcc(0.99)`, which cannot see this either.
- `test_unary_pow_fp32_ulp_noninteger` / `..._integer_exact` (added by #49625) sweep
  `torch.logspace(-0.3, 4.7)`, i.e. `x ∈ [0.5, 50000]` — no zero, no small base.
- `test_binary_pow_scalar_input` (`:51`) parametrises `input` over `[10.0, 5.5, -5.0, ...]` — no zero.
- `test_backward_rpow.py` uses `exponent ∈ {1.5, 5.7, 0.0, 15.2}` (all `>= 1` or `0`, i.e. all inside
  the clamp's reach), inputs from `data_gen_with_range(shapes, -201, 199)` (probability of an exact
  `0.0` is ~0), and compares with `compare_pcc` — which cannot see "should be 0, got 7.7e-20" at all.

No test in the file asserts `pow(0, p) == 0` for any `p` in `(0, 1)`.

### Relationship to existing issues

- **#49625 ($2,500 bounty, closed 2026-07-18) / #49649** rewrote the fp32 `pow` path — the Veltkamp
  split and three-NR reciprocal now in `_sfpu_unary_power_61f_updated_` and
  `_sfpu_binary_power_f32_` are that work. Its success criteria were stated over
  `x ∈ [0.5, 50000]`, so zero and small bases were out of scope by construction. The code I quote above
  is post-#49625 `main`, verified byte-identical to my local copy. This proposal is disjoint from it:
  it touches only lanes where `base == 0`, and does not move a single ULP anywhere #49625 measured.
- **#34129 (open draft PR, "Pow binary testing")** is the official exhaustive binary-`pow` ULP sweep.
  It is worth being precise about what it does and does not cover, because at first glance it looks
  like it must already have found this.

  Its base tensor comes from `test_unary_pow.py::generate_clean_bf16_tensor`, whose mask is
  `torch.isfinite(fp32) & ~neg_zero_mask & (fp32.abs() >= tiny)` with `tiny = 2⁻¹²⁶`. That removes
  `+0`, `-0` **and** every subnormal, leaving 65024 of the 65536 bf16 values. So the 65024 × 65024
  sweep never uses zero as a base. Its companion special-value file pairs
  `[inf, nan, -inf, -0.0, 1.0, 0.0]` against each other — no exponent in `(0, 1)` appears there
  either. The two files between them therefore cannot reach `pow(0, p)` for `0 < p < 1`.

  On subnormals specifically, the PR states *"Subnormal values (< 2⁻¹²⁶) are excluded as they are not
  supported by the hardware (510 values)"*. **Zero is not a subnormal**, and `-0.0` is in that PR's own
  special-value list — so a zero base is inside the scope it claims, and is missing only because the
  shared generator filters it out. Whatever that sweep recorded, no
  code change followed — the guard on `main` today is still the `pow < 0` half only, and there is no
  regression test for a zero base at a fractional exponent. I have taken the subnormal exclusion at
  face value and demoted the subnormal part of this report accordingly (see "Known limits").
- **#40692** (open issue) is accuracy work on binary fp32 `pow`; **#37023** and **#31933** are open
  draft PRs on validation-range testing. None proposes a special-value fix. Two of these being
  in-flight test PRs rather than reports is worth noting: it means the gap here is that nobody has
  *reported* the zero base, not that nobody is working on `pow` coverage.
  **#21785** ("ttnn.pow return inf instead of nan", closed) and **#23529** (closed) are unrelated cases.
- I found no open issue or PR covering a zero or near-zero base, and no **open** `label:bounty` issue
  touching `pow` (the 14 open bounty issues are model bring-ups plus i0/quantize/matmul/Nix). The one
  closed `label:bounty` issue on `pow` is #49625 above.

### Files

| file | what |
|---|---|
| `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_unary_power.h` | `:118-123`, `:299-304` guard; `:335`, `:358` host branch |
| `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_unary_power.h` | `:116-118`, `:291-293`; `:327`, `:350` |
| `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h` | `:123`, `:256` guard |
| `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h` | `:123`, `:255` |

`ckernel_sfpu_rpow.h` is affected through `_sfpu_binary_power_` and needs no edit of its own. WH and BH
are arithmetically identical for all four files (differences are `ADDR_MOD`, one `sfpi::` qualifier and
comment placement).

Fix templates already in-repo: `ckernel_sfpu_log.h:49` (normalises `-0.0` and subnormals, maps
`0.0 -> -inf`) and `ckernel_sfpu_sqrt_custom.h` (`v_if(val != 0.0f)`, WH `:24` / BH `:18`).

### Proposed fix

**Unary kernels.** The scalar exponent is already branched on at `:335` / `:358`; make it three-way so
the common case costs nothing. Note the NaN exponent: today `pow_scalar >= 0.0f` is false for NaN, so
it selects the `<false>` instantiation and a zero base is filled with NaN. The ordering below keeps
that (`NaN` falls through both `> 0` and `== 0` into the final branch):

```c
    if (pow_scalar > 0.0f) {          // NEW instantiation
        ... _sfpu_unary_power_61f_updated_<ZeroBase::ToZero>(base, pow) ...
    } else if (pow_scalar == 0.0f) {
        ... <ZeroBase::None>   ...    // 0**0 == 1 already falls out
    } else {
        ... <ZeroBase::ToNaN>  ...    // pow < 0 *and* NaN — matches today exactly
    }
```

The `== 0` test must come before the final `else`, not after it: a NaN exponent fails both `> 0` and
`== 0` and so reaches `ToNaN`, which is what today's `pow_scalar >= 0.0f` already does for NaN.

with, in the kernel body, in place of the current `if constexpr (!IS_POSITIVE_EXPONENT)` block:

```c
    if constexpr (ZERO_BASE == ZeroBase::ToZero) {
        v_if(abs_base == 0.f) { y = 0.0f; }
        v_endif;
    } else if constexpr (ZERO_BASE == ZeroBase::ToNaN) {
        v_if(abs_base == 0.f) { y = sfpi::vConstFloatPrgm2; }
        v_endif;
    }
```

Cost for `pow > 0`: one `v_if` on a value (`abs_base`) that is already live, plus a load-immediate —
the same instruction count the `pow < 0` instantiation already pays today. Nothing added for `pow == 0`.

**On code size**, since this kernel is not far from the compiler's limits — #49625 records a
`maximum number of generated reload insns` ICE on Wormhole when a full two-product was tried on the
16×-inlined body. The sketch above turns two template instantiations into three, each carrying its
own `#pragma GCC unroll 8`. If that increment is unwelcome, the same fix collapses to **one**
instantiation instead of three: drop the template parameter, pass the fill value as an ordinary
runtime `vFloat` (`0.0f` when `pow > 0`, NaN when `pow < 0`, and skipped when `pow == 0`), and
execute `v_if(abs_base == 0.f) { y = zero_fill; }` unconditionally. That is strictly smaller than
today's two instantiations. I have no way to measure either variant without a device; whichever the
maintainers prefer is fine, and the correctness argument is the same for both.

**Binary kernels.** `pow` is per-lane, so the guard must be vectorised. Replace
`v_if(base == 0.f && pow < 0.f)` with a strict superset:

```c
    v_if(base == 0.f) {
        v_if(pow < 0.f)     { y = std::numeric_limits<float>::quiet_NaN(); }  // unchanged
        v_elseif(pow > 0.f) { y = 0.0f; }                                     // NEW
        v_endif;                                    // pow == 0 keeps y == 1 (0**0)
    }
    v_endif;
```

One extra compare and one extra predicated move on top of a guard that already exists, in a kernel
measured at ~8000 ns per fp32 tile in #49625.

**Validation of the fix** (emulated, same harness as the tables above):

| check | result |
|---|---|
| `pow(0, p)`, 200001 points in `[1e-5, 0.999]`, unary fp32 + bf16 | **0 / 200001 nonzero** (was 200001 / 200001) |
| `pow(0, p)` for `p ∈ {1.5, 4, 8}` | exact `+0.0` (was `6.17e-44`) |
| `pow(0, 0)` | `1.0` preserved |
| `pow(0, p<0)` | `NaN` preserved (documented behaviour untouched) |
| `pow(-0.0, 0.5)` | `+0.0` |
| **regression: 600005 non-zero bases** spanning `1e-37 … 1e37` and both signs, exponents in `[-8, 8]` | **0 outputs changed, bit-for-bit**, both paths |

The last row is the important one: the fix is confined to lanes where `base == 0`, so nothing #49625
measured can move.

### Known limits (stated up front)

- **Subnormal bases are a real but secondary case, and I am deliberately not leading with them.** The
  same `exexp`/`setexp` reduction mis-scales subnormal bases (`exexp` returns `-127` for every
  subnormal, and `setexp(·,127)` fabricates the implicit leading 1), giving e.g.
  `pow(2^-149, 0.5)` = `7.6664676e-20` against `torch.sqrt`'s float32 result `3.7433921e-23`
  (`0x1A3504F3`; the exact value `2^-74.5` is `3.7433921305...e-23`, and the float32 result is that
  correctly rounded) — a ratio of exactly
  **2048×**; `pow(1e-40, 0.5)` = `7.6990075e-20` vs `9.9999730e-21`, ratio **7.699×**.
  But #34129 states subnormals are excluded as unsupported by the hardware, and `ckernel_sfpu_sqrt.h`
  is inaccurate on subnormal inputs too (`sqrt(1e-40)` → `3.46e-21` vs `1.0e-20`), so this is a
  tree-wide policy question rather than a `pow` defect. **Either way the zero fix covers it**: if the
  device flushes subnormal inputs to zero — the convention `test_pow.py` itself applies via
  `flush_subnormal_values_to_zero` — the subnormal base becomes `+0.0` and the fixed kernel returns
  `+0.0`, matching the flushed torch golden. If subnormals do survive into the SFPU, a proper fix needs
  a normalise-then-compensate step (multiply by `2^24`, subtract 24 from `exp`), whose correctness
  depends on whether an SFPU multiply accepts a subnormal operand — which I cannot determine without a
  device. I would scope that separately.
- **`pow(0, p<0)` is left as `NaN`**, not IEEE's `±inf`. That is the documented contract at
  `ckernel_sfpu_binary_pow.h:34` and I am not proposing to change it here.
- **The `6.17e-44` residue for `p >= 1` on the unary fp32 path is subnormal** and probably flushed on
  device; I have not counted it as an observable defect, only noted that it exists and that the fix
  removes it.
- **No hardware.** All numbers are float32 emulations of each kernel's exact instruction sequence
  (details in the footnote). Steps 1–3 and 7 of the root cause are properties of the source and hold
  regardless of any numeric model.

### Success criteria

- [ ] `ttnn.pow(x, p) == 0` exactly, for `x == ±0.0` and every `p > 0`, on both fp32 and bf16 dest
      accumulate, both dispatch paths, both architectures
- [ ] `ttnn.pow(x, 0.5) == ttnn.sqrt(x)` at `x == ±0.0` — by value, not by bit pattern. The sign of a
      zero result is genuinely ambiguous here: `torch.sqrt(-0.0)` is `-0.0`, and `torch.pow` itself is
      inconsistent (`torch.pow(-0.0, 0.5)` returns `-0.0` with a scalar exponent, because it takes a
      `sqrt` shortcut, but `+0.0` with a tensor exponent). The repo's own
      `utils_for_testing.py::ulp_distance` already documents that some SFPU pack/convert paths
      canonicalise `-0` to `+0` and treats `±0` as 0 ULP apart, so pinning the sign here would be
      asking for something the hardware does not promise. The fix as sketched returns `+0.0` in all
      cases; if `-0.0` for a negative zero base is wanted, that is a separate, larger question
- [ ] `ttnn.pow(zeros, 4.0)` (unary) and `ttnn.pow(zeros, full_like(zeros, 4.0))` (binary) agree
- [ ] `pow(0, 0) == 1` and `pow(0, p<0) == NaN` unchanged
- [ ] bit-for-bit no change for any non-zero base — in particular the #49625 sweeps
      (`test_unary_pow_fp32_ulp_noninteger` `<= 3` ULP, `test_unary_pow_fp32_ulp_integer_exact` `0` ULP)
      still pass unmodified
- [ ] regression tests added with `p` in `(0, 1)` at a zero base. This needs a **new** case rather than
      a tweak to an existing one: the shared input generator
      `test_unary_pow.py::generate_clean_bf16_tensor` masks with
      `torch.isfinite(fp32) & ~neg_zero_mask & (fp32.abs() >= tiny)`, so `+0`, `-0` and every
      subnormal are removed before any of the callers see them. Relaxing a downstream filter cannot
      reintroduce them.
- [ ] the new case asserts **exact equality**, not `assert_allclose`. Today's wrong values are tiny
      (`2.77e-10`, `7.67e-20`, `2.12e-29` for `p` = 0.25, 0.5, 0.75), so any absolute tolerance at or
      above `5e-4` — which is what `test_pow_arange_masking_fp32` uses — passes on the current kernel.
      A test that cannot fail on today's code would not be a regression test for this
- [ ] assertion is exact equality or ULP, not PCC — a PCC threshold over a tile cannot see a handful
      of wrong lanes, which is why `test_backward_rpow.py` never caught this
- [ ] confirmed on Blackhole

### Why this matters

`x == 0 ⇒ sqrt(x) == 0` is not an accuracy target, it is an invariant, and there is no tolerance to
negotiate: the correct answer is exactly representable and the kernel returns something else for
**100%** of the `(0, 1)` exponent range. The failure is silent — finite, positive, and small enough to
look like rounding — and it is worst where it is most likely to be hit: as `p → 0+` the returned value
approaches `1.0`.

The affected exponents are the common ones. `p = 0.5` is `sqrt` (norms, distances, RMSNorm, attention
scaling, standard deviation). `p = 1/3` is `cbrt`. `p ∈ (0,1)` is the whole family of fractional-power
normalisations and learning-rate schedules. Zero inputs in those tensors are ordinary — padded tiles,
masked positions, ReLU outputs, freshly initialised buffers — and each one becomes `7.7e-20` instead
of `0`, which then flows into a reciprocal or a division and produces `1.3e19`.

It is also cheap and low-risk to fix: a few SFPU instructions, confined to lanes where `base == 0`, on
a `pow` implementation that was just brought to `<= 3` ULP by #49625 and deserves a correct zero.

---

*Method: numbers are float32 emulations of each kernel's exact instruction sequence in numpy, replaying
`ckernel_sfpu_unary_power.h` and `ckernel_sfpu_binary_pow.h` operation by operation with every
intermediate rounded to float32 (the SFPU has no fused multiply-add, so each `sfpmul`/`sfpadd` rounds
separately); `setexp`, `exexp`, `addexp`, `setsgn` and the `RoundMode`/`ExponentMode` defaults were read
from `sfpi_lib.h`. **I do not have a device — these are simulations, not silicon measurements.** The
unary and binary paths were replayed independently and agree, which is a cross-check on the harness but
not a substitute for hardware. `pow(0, 0.5)` is quoted as bits (`0x1FB504F3`) because it is a closed
form — `float32(sqrt(2))` with exponent field 63 — reachable by hand from the source without any
simulator. Reference values were checked directly against `torch` 2.13 (CPU), the golden this repo's
tests compare to: `torch.pow(0.0, p)` returns exactly `0.0` for every `p` tested in `(0, 1)`, and
`torch.sqrt(0.0)` returns `0.0`. Each input is taken at its float32 value before the reference is
computed. Sources were fetched from `tenstorrent/tt-metal` `main` and the local copies used for the
replay were diffed against them.*

I can prepare a PR with the kernel change and a pytest covering the special-value cases above, if this is accepted.

iklemenskiTT:
### PR Category
Bug fix

### Summary
Fixes #51468. `setexp`/`exexp` map `log2(0)` to `-127`, so `ttnn.pow(0, p)` returned `2**(-127p)` for every `p`. `2**(-127p)` is a normal float only for `p < 126/127 ≈ 0.992`, so small exponents return visibly wrong values (`pow(0, 0.5)` = `7.67e-20`) while larger ones underflow and flush to `0` — which is why `pow(t, 4)` always looked correct and `pow(t, 0.5)` did not. The zero-base guard only covered `pow < 0 → NaN`.

Unary and binary SFPU kernels (WH + BH) now fill `±0 ** p>0 → +0`, after the sign branch. `rpow` inherits the binary kernel. `0**0 → 1` unchanged. `0**negative` stays NaN on fp32 (documented behaviour); bf16 dest packs that NaN as inf.

Also fixes a pre-existing case on binary fp32: `pow(-0, negative even integer)` returned inf (e.g. `pow(-0, -2)`), because the guard's `base == 0.f` is bit-exact and misses `-0`. The zero-base guard now uses a single `abs`, so it returns NaN like every other path. Non-integer negative `p` was already correct — the sign branch caught it.

Issue #51468 is wrong that `-0` behaves like `+0`. SFPU `<` is sign-bit based, so `v_if(base < 0)` fires on `-0`: pre-fix `pow(-0, 0.5)` was NaN, not `7.67e-20`.

### Notes for reviewers
- The zero fill runs after the negative-base branch. SFPU `<` is sign-bit based, so `v_if(base < 0)` classes `-0` as a negative base and would give `-0 ** 0.5` the complex-result NaN. `-0` is a zero, not a negative base — `torch.pow(-0.0, 0.5)` is `+0` — so the zero fill has to run last and win.
- Do not drop `abs_end` and reuse the top-of-function `abs_base` in these guards: that keeps `abs_base` live across the whole log2/exp2 body and spills `_sfpu_binary_power_f32_` on Blackhole (`sfpi_classes.h:131: internal compiler error: cannot store sfpu register`). Recomputing `sfpi::abs(base)` at the point of use costs one instruction and frees the register.
- `v_and` narrowing rather than two flat `v_if`s: `v_and` tightens the active lane mask in
  place, so it needs no `sfppushc`/`sfppopc`/`sfpcompc` for a second predicate scope.
  144 vs 147 SFPU instructions on Blackhole, 173 vs 176 on Wormhole — the same
  `sfpsetcc −2 / sfpencc −1` exchange with nothing added. A nested `v_if`/`v_elseif`
  measures 148, worse than either.
- `pow(0, NaN)` on the binary path returns `+0` where it previously returned `inf`; both
  are wrong (IEEE says NaN) and the guard shape does not affect it — flat, nested and
  `v_and` all measure `0x00000000`. Tracked in #53922 with the cost of fixing it.
- No third unary template instantiation, to stay clear of the reload-insn ICE from #49625. (Not measured; that ICE is a different failure mode from the spill above.)
- Binary fp32 `pow(-0, odd integer)` was IEEE `-0` pre-fix and is `+0` post-fix (zero fill runs after `setsgn2`). Issue success criteria allow `+0`. Unary was already `+0`.
- On bf16 dest, the NaN the kernel writes for `0**negative` reads back as inf
  (tenstorrent/tt-llk#675 — packers flush NaN to infinity on the BF16→BF16 path), so the
  test asserts `isnan` on fp32 and only `~isfinite` on bf16.
- Scalar `pow(t, 0.0)` floors to the integer overload and routes to `power_iterative`, not the SFPU kernel; SFPU `0**0` is covered by the tensor-exponent path.

### Test plan
- [x] `test_pow_zero_base_special_cases` (new) — exact `0` for `±0` × `{1e-4, 0.01, 0.25, 0.5, 1/3, 0.75, 0.99, 1.5, 3, 4}` on unary, binary and `rpow`; `0**0 == 1`; fp32 `isnan` / bf16 `~isfinite` for `±0 ** -1.5`. Asserts exact equality, not tolerance — a regressed `2**(-127p)` is ~`7.7e-20` at `p=0.5` and passes any `allclose` this file uses.
- [x] `test_pow.py` + `test_unary_pow.py` full — 594 passed, 3 xfailed (pre-existing), Blackhole. Includes `test_special_input_fp32`, `test_unary_pow_fp32_ulp_noninteger` (≤3 ULP) and `test_unary_pow_fp32_ulp_integer_exact` (0 ULP), so #49625's accuracy gates are unmoved.
- [x] Wormhole 
- [ ] CI post-commit

### Performance

Blackhole `p100a`, binary `pow`, device kernel time via tracy
(`ttnn-eltwise-op-tester/bench.py --type binary -k pow`).Run-to-run spread
is 0.002% on bf16 and bit-identical on fp32, so these deltas are well outside noise.

| state | bf16 cycles/tile | fp32 cycles/tile |
|---|---:|---:|
| pre-fix (`d7f3415767d`) | 3263.71 | 5606.21 |
| flat `v_if` pair | 3391.79 | 5735.90 |
| `v_and` narrowing | 3295.69 | 5639.71 |

| comparison | bf16 | fp32 |
|---|---:|---:|
| cost of the fix (pre-fix → flat pair) | +128.08 (+3.92%) | +129.69 (+2.31%) |
| `v_and` vs flat pair | **−96.10 (−2.83%)** | **−96.19 (−1.68%)** |
| net cost vs pre-fix (`v_and`) | +31.98 (+0.98%) | +33.50 (+0.60%) |

`v_and` recovers ~75% of the fix's cost. Cycles agree exactly with the
disassembly: `cycles_per_datum × 32 lanes` gives 3.00 (bf16) and 3.01 (fp32) for
the `v_and` saving, matching the 147 → 144 SFPU instruction count in
`trisc1.elf`, and 4.00 / 4.05 for the fix itself. One SFPU instruction = one
cycle across 32 lanes on this kernel.

`v_and` is behaviourally identical, not just correct: base ∈ {`+0`, `−0`} × all
65536 bf16 exponent patterns × both dtypes = 262,144 lanes, **0 bitwise
differences** against the flat pair. All 32,640 positive exponent patterns
return exact `+0`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:iklemenskiTT/51468/fix-pow-zero-base)

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_pow.py
@@ -249,6 +249,74 @@ def test_special_input_fp32(device):
 
 
 @pytest.mark.parametrize("dtype", ["float32", "bfloat16"])
+def test_pow_zero_base_special_cases(device, dtype):
+    # Needs its own inputs: the shared generators strip ±0 before any caller sees them.
+    # Asserts exact equality, not allclose — a regressed 2**(-127p) is ~7.7e-20 at p=0.5,
+    # well inside any absolute tolerance this file uses elsewhere.
+    torch_dtype = getattr(torch, dtype)
+    ttnn_dtype = getattr(ttnn, dtype)
+    shape = [1, 1, 32, 32]
+    positive_exponents = (1e-4, 0.01, 0.25, 0.5, 1.0 / 3.0, 0.75, 0.99, 1.5, 3.0, 4.0)
+
+    def assert_zero_power_undefined(out):
+        if dtype == "float32":
+            assert torch.isnan(out).all()
+        else:
+            # bf16 dest reads back as inf rather than the NaN the kernel writes
+            # (tenstorrent/tt-llk#675), so only non-finiteness is assertable here.
+            assert (~torch.isfinite(out)).all()
+
+    for exponent in positive_exponents:
+        exp_t = torch.full(shape, exponent, dtype=torch_dtype)
+        tt_exp = ttnn.from_torch(exp_t, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+        # rpow scalar base is always +0; one run per exponent covers that kernel.
+        rpow_out = ttnn.to_torch(ttnn.rpow(tt_exp, 0.0))
+        assert torch.equal(
+            rpow_out, torch.zeros_like(rpow_out)
+        ), f"rpow({exponent}, 0.0) = {rpow_out.flatten()[0].item()}"
+
+        for base_val in (0.0, -0.0):
+            zeros = torch.full(shape, base_val, dtype=torch_dtype)
+            tt_zeros = ttnn.from_torch(zeros, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+            unary_out = ttnn.to_torch(ttnn.pow(tt_zeros, exponent))
+            assert torch.equal(
+                unary_out, torch.zeros_like(unary_out)
+            ), f"unary pow({base_val}, {exponent}) = {unary_out.flatten()[0].item()}"
+
+            binary_out = ttnn.to_torch(ttnn.pow(tt_zeros, tt_exp))
+            assert torch.equal(
+                binary_out, torch.zeros_like(binary_out)
+            ), f"binary pow({base_val}, {exponent}) = {binary_out.flatten()[0].item()}"
+
+    tt_zeros = ttnn.from_torch(
+        torch.zeros(shape, dtype=torch_dtype), dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    # Scalar 0.0 is an integer exponent → power_iterative, not the SFPU kernel.
+    unary_zero_exp = ttnn.to_torch(ttnn.pow(tt_zeros, 0.0))
+    assert torch.equal(unary_zero_exp, torch.ones_like(unary_zero_exp))
+
+    tt_zero_exp = ttnn.from_torch(
+        torch.zeros(shape, dtype=torch_dtype), dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    binary_zero_exp = ttnn.to_torch(ttnn.pow(tt_zeros, tt_zero_exp))
+    assert torch.equal(binary_zero_exp, torch.ones_like(binary_zero_exp))
+
+    unary_neg = ttnn.to_torch(ttnn.pow(tt_zeros, -1.5))
+    assert_zero_power_undefined(unary_neg)
+
+    tt_neg_zero = ttnn.from_torch(
+        torch.full(shape, -0.0, dtype=torch_dtype), dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    tt_neg_exp = ttnn.from_torch(
+        torch.full(shape, -1.5, dtype=torch_dtype), dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    binary_neg = ttnn.to_torch(ttnn.pow(tt_zeros, tt_neg_exp))
+    binary_neg_zero = ttnn.to_torch(ttnn.pow(tt_neg_zero, tt_neg_exp))
+    assert_zero_power_undefined(binary_neg)
+    assert_zero_power_undefined(binary_neg_zero)
+
+
+@pytest.mark.parametrize("dtype", ["float32", "bfloat16"])
 def test_pow_determinism(device, dtype):
     torch.manual_seed(0)
 
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
@@ -31,6 +31,7 @@ namespace sfpu {
  * @return sfpi::vFloat Result of base**pow
  *
  * Special Cases:
+ * - base = 0, pow > 0: Returns 0
  * - base = 0, pow < 0: Returns NaN (undefined)
  * - base < 0, pow = integer: Returns proper signed result (negative if odd power)
  * - base < 0, pow = non-integer: Returns NaN (complex result)
@@ -119,12 +120,6 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_21f_(sfpi::vFloat base, sfpi::vFloat
         pow, sfpi::RoundMode::Nearest);  // int16 should be plenty, since large powers will approach 0/Inf
     auto pow_rounded = sfpi::convert<sfpi::vFloat>(pow_int, sfpi::RoundMode::Nearest);
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN
-    v_if((absbase == 0.f) && pow < 0.f) {
-        y = std::numeric_limits<float>::quiet_NaN();  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-    }
-    v_endif;
-
     v_if(base < 0.0f) {  // negative base
         // If pow is odd integer then result is negative
         // If power is even, then result is positive
@@ -132,6 +127,24 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_21f_(sfpi::vFloat base, sfpi::vFloat
 
         // Check for integer power, if it is not then overwrite result with NaN
         v_if(pow_rounded != pow) {  // negative base and non-integer power => set to NaN
+            y = std::numeric_limits<float>::quiet_NaN();
+        }
+        v_endif;
+    }
+    v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the negative-base branch: SFPU `<` is sign-bit based, so
+    // v_if(base < 0) classes -0 as negative and would overwrite the 0 with the
+    // complex-result NaN; -0 is a zero, not a negative base.
+    // Fill 0 for every non-zero exponent, then narrow to the negative ones.
+    // v_and tightens the enclosing predicate in place, so it costs one compare
+    // where a second flat v_if would also save and restore the lane mask.
+    // pow == 0 fails the != 0 gate and keeps 1.
+    v_if(absbase == 0.f) {
+        v_if(pow != 0.f) {
+            y = 0.0f;
+            v_and(pow < 0.f);
             y = std::numeric_limits<float>::quiet_NaN();
         }
         v_endif;
@@ -251,12 +264,6 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
     v_if(out_exp >= 255) { y = std::numeric_limits<float>::infinity(); }
     v_endif;
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN (only for negative exponents)
-    v_if(base == 0.f && pow < 0.f) {
-        y = std::numeric_limits<float>::quiet_NaN();  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-    }
-    v_endif;
-
     v_if(base < 0.0f) {  // negative base
         // Post-processing: ensure that special values (e.g. 0**0, -1**0.5, ...) are handled correctly
         // Check valid base range
@@ -271,6 +278,28 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
 
         // Check for integer power, if it is not then overwrite result with NaN
         v_if(pow_rounded != pow) {  // negative base and non-integer power => set to NaN
+            y = std::numeric_limits<float>::quiet_NaN();
+        }
+        v_endif;
+    }
+    v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the negative-base branch: SFPU `<` is sign-bit based, so
+    // v_if(base < 0) classes -0 as negative and would overwrite the 0 with the
+    // complex-result NaN; -0 is a zero, not a negative base.
+    // SFPU `== 0` is bit-exact, so matching -0 needs an abs. Recompute it here
+    // rather than reusing abs_base: that live range would span the log2/exp2 body
+    // and spills this kernel.
+    // Fill 0 for every non-zero exponent, then narrow to the negative ones.
+    // v_and tightens the enclosing predicate in place, so it costs one compare
+    // where a second flat v_if would also save and restore the lane mask.
+    // pow == 0 fails the != 0 gate and keeps 1.
+    sfpi::vFloat abs_end = sfpi::abs(base);
+    v_if(abs_end == 0.f) {
+        v_if(pow != 0.f) {
+            y = 0.0f;
+            v_and(pow < 0.f);
             y = std::numeric_limits<float>::quiet_NaN();
         }
         v_endif;
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
@@ -31,6 +31,7 @@ namespace sfpu {
  * @return sfpi::vFloat Result of base**pow
  *
  * Special Cases:
+ * - base = 0, pow > 0: Returns 0
  * - base = 0, pow < 0: Returns NaN (undefined)
  * - base < 0, pow = integer: Returns proper signed result (negative if odd power)
  * - base < 0, pow = non-integer: Returns NaN (complex result)
@@ -119,12 +120,6 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_21f_(sfpi::vFloat base, sfpi::vFloat
         pow, sfpi::RoundMode::Nearest);  // int16 should be plenty, since large powers will approach 0/Inf
     auto pow_rounded = sfpi::convert<sfpi::vFloat>(pow_int, sfpi::RoundMode::Nearest);
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN
-    v_if((absbase == 0.f) && pow < 0.f) {
-        y = std::numeric_limits<float>::quiet_NaN();  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-    }
-    v_endif;
-
     v_if(base < 0.0f) {  // negative base
         // If pow is odd integer then result is negative
         // If power is even, then result is positive
@@ -133,6 +128,24 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_21f_(sfpi::vFloat base, sfpi::vFloat
 
         // Check for integer power, if it is not then overwrite result with NaN
         v_if(pow_rounded != pow) {  // negative base and non-integer power => set to NaN
+            y = std::numeric_limits<float>::quiet_NaN();
+        }
+        v_endif;
+    }
+    v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the negative-base branch: SFPU `<` is sign-bit based, so
+    // v_if(base < 0) classes -0 as negative and would overwrite the 0 with the
+    // complex-result NaN; -0 is a zero, not a negative base.
+    // Fill 0 for every non-zero exponent, then narrow to the negative ones.
+    // v_and tightens the enclosing predicate in place, so it costs one compare
+    // where a second flat v_if would also save and restore the lane mask.
+    // pow == 0 fails the != 0 gate and keeps 1.
+    v_if(absbase == 0.f) {
+        v_if(pow != 0.f) {
+            y = 0.0f;
+            v_and(pow < 0.f);
             y = std::numeric_limits<float>::quiet_NaN();
         }
         v_endif;
@@ -252,12 +265,6 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
     v_if(out_exp >= 255) { y = std::numeric_limits<float>::infinity(); }
     v_endif;
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN (only for negative exponents)
-    v_if(base == 0.f && pow < 0.f) {
-        y = std::numeric_limits<float>::quiet_NaN();  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-    }
-    v_endif;
-
     v_if(base < 0.0f) {  // negative base
         // Post-processing: ensure that special values (e.g. 0**0, -1**0.5, ...) are handled correctly
         // Check valid base range
@@ -271,6 +278,28 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
 
         // Check for integer power, if it is not then overwrite result with NaN
         v_if(pow_rounded != pow) {  // negative base and non-integer power => set to NaN
+            y = std::numeric_limits<float>::quiet_NaN();
+        }
+        v_endif;
+    }
+    v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the negative-base branch: SFPU `<` is sign-bit based, so
+    // v_if(base < 0) classes -0 as negative and would overwrite the 0 with the
+    // complex-result NaN; -0 is a zero, not a negative base.
+    // SFPU `== 0` is bit-exact, so matching -0 needs an abs. Recompute it here
+    // rather than reusing abs_base: that live range would span the log2/exp2 body
+    // and spills this kernel.
+    // Fill 0 for every non-zero exponent, then narrow to the negative ones.
+    // v_and tightens the enclosing predicate in place, so it costs one compare
+    // where a second flat v_if would also save and restore the lane mask.
+    // pow == 0 fails the != 0 gate and keeps 1.
+    sfpi::vFloat abs_end = sfpi::abs(base);
+    v_if(abs_end == 0.f) {
+        v_if(pow != 0.f) {
+            y = 0.0f;
+            v_and(pow < 0.f);
             y = std::numeric_limits<float>::quiet_NaN();
         }
         v_endif;
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_unary_power.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_unary_power.h
@@ -25,11 +25,12 @@ namespace sfpu {
  *
  * @param base The base value (sfpi::vFloat vector), can be any floating point number
  * @param pow The exponent/power value (sfpi::vFloat vector), can be any floating point number
- * @tparam IS_POSITIVE_EXPONENT If true, assumes exponent >= 0 (skips zero-base check for optimization)
+ * @tparam IS_POSITIVE_EXPONENT If true, assumes exponent >= 0 (zero base maps to 0 rather than NaN)
  *
  * @return sfpi::vFloat Result of base**pow
  *
  * Special Cases:
+ * - base = 0, pow > 0: Returns 0
  * - base = 0, pow < 0: Returns NaN (undefined)
  * - base < 0, pow = integer: Returns proper signed result (negative if odd power)
  * - base < 0, pow = non-integer: Returns NaN (complex result)
@@ -112,14 +113,6 @@ sfpi_inline sfpi::vFloat _sfpu_unary_power_21f_(sfpi::vFloat base, sfpi::vFloat 
 
     sfpi::vFloat y = sfpi::as<sfpi::vFloat>(zii);
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN (only for negative exponents)
-    if constexpr (!IS_POSITIVE_EXPONENT) {
-        v_if(abs_base == 0.f) {
-            y = sfpi::vConstFloatPrgm2;  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-        }
-        v_endif;
-    }
-
     // Negative base handling (for both positive and negative exponents)
     v_if(base < 0.0f) {
         // Post-processing: ensure that special values (e.g. 0**0, -1**0.5, ...) are handled correctly
@@ -139,6 +132,20 @@ sfpi_inline sfpi::vFloat _sfpu_unary_power_21f_(sfpi::vFloat base, sfpi::vFloat 
         v_endif;
     }
     v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the sign branch: SFPU `<` is sign-bit based, so v_if(base < 0)
+    // fires on -0 and would replace the 0 result with NaN for non-integer p.
+    // pow == 0 falls through the guard and keeps 1.
+    if constexpr (IS_POSITIVE_EXPONENT) {
+        v_if(abs_base == 0.f && pow > 0.f) { y = 0.0f; }
+        v_endif;
+    } else {
+        v_if(abs_base == 0.f) {
+            y = sfpi::vConstFloatPrgm2;  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
+        }
+        v_endif;
+    }
 
     // LRegs work on float32 data. If DST is bfloat16 then SFPSTORE will truncate it.
     // This can reduce accuracy: for instance, 9**2 = 80.8 gets round to 80.5
@@ -287,14 +294,6 @@ sfpi_inline sfpi::vFloat _sfpu_unary_power_61f_updated_(const sfpi::vFloat& base
     sfpi::vFloat z_lo = pow_lo * exp_f32 + pow * (ln_m * vConst1Ln2);
     sfpi::vFloat y = _sfpu_pow2_f32_accurate_hilo_(z_hi, z_lo);
 
-    // Division by 0 when base is 0 and pow is negative => set to NaN (only for negative exponents)
-    if constexpr (!IS_POSITIVE_EXPONENT) {
-        v_if(abs_base == 0.f) {
-            y = sfpi::vConstFloatPrgm2;  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
-        }
-        v_endif;
-    }
-
     v_if(base < 0.0f) {  // negative base
         // Post-processing: ensure that special values (e.g. 0**0, -1**0.5, ...) are handled correctly
         // Check valid base range
@@ -314,6 +313,20 @@ sfpi_inline sfpi::vFloat _sfpu_unary_power_61f_updated_(const sfpi::vFloat& base
         v_endif;
     }
     v_endif;
+
+    // setexp/exexp map 0 to log2 = -127, so 0**p evaluates as 2**(-127p).
+    // Must follow the sign branch: SFPU `<` is sign-bit based, so v_if(base < 0)
+    // fires on -0 and would replace the 0 result with NaN for non-integer p.
+    // pow == 0 falls through the guard and keeps 1.
+    if constexpr (IS_POSITIVE_EXPONENT) {
+        v_if(abs_base == 0.f && pow > 0.f) { y = 0.0f; }
+        v_endif;
+    } else {
+        v_if(abs_base == 0.f) {
+            y = sfpi::vConstFloatPrgm2;  // negative powers of 0 are NaN, e.g. pow(0, -1.5)
+        }
+        v_endif;
+    }
 
     return y;
 }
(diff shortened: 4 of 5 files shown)


tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
```diff
@@ -277,6 +275,19 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
...
+    // fires on -0 and would replace the 0 result with NaN for non-integer p.
+    // SFPU `== 0` is bit-exact, so the guards need an abs to match -0. Recompute it
+    // here rather than reusing abs_base: that live range would span the log2/exp2
+    // body and spills this kernel.
+    // pow == 0 falls through both guards and keeps 1.
+    sfpi::vFloat abs_end = sfpi::abs(base);
+    v_if(abs_end == 0.f && pow < 0.f) { y = std::numeric_limits<float>::quiet_NaN(); }
+    v_endif;
+    v_if(abs_end == 0.f && pow > 0.f) { y = 0.0f; }
+    v_endif;
```
nmauriceTT: Nit: If sfpi compiler does not optimize the following, then this would get translated into two separate branches (2 x `SFPGT`/`SFPLE`/`SFPSETCC` + `SFPENCC`), and you'd be doing lane-enable/disable twice + check twice if `abs_end == 0.f`. 

Instead, I think the following may be _slightly_ faster:

```suggestion
    v_if(abs_end == 0.f) { 
        v_if (pow < 0.f) {
            y = std::numeric_limits<float>::quiet_NaN(); 
        }
        v_elseif (pow > 0.f) {
            y = 0.0f;
        }
        v_endif;
    v_endif;
```

tests/ttnn/unit_tests/operations/eltwise/test_pow.py
```diff
@@ -248,6 +248,74 @@ def test_special_input_fp32(device):
...
+    shape = [1, 1, 32, 32]
+    positive_exponents = (1e-4, 0.01, 0.25, 0.5, 1.0 / 3.0, 0.75, 0.99, 1.5, 3.0, 4.0)
+
+    def assert_zero_power_undefined(out):
+        if dtype == "float32":
+            assert torch.isnan(out).all()
+        else:
+            # SFPSTORE's FP32->BF16 conversion truncates the mantissa, which turns
+            # some NaN encodings into inf, so only non-finiteness is assertable here.
+            assert (~torch.isfinite(out)).all()
```
nmauriceTT: Nit: AFAIK, NaN issue is related to a packer LLK bug rather than a SFPU hardware issue with SFPSTORE: https://github.com/tenstorrent/tt-llk/issues/675

While quiet NaNs are implementation defined, implementation should use `0x7FC00000`. Truncating this to bfloat16 should yield 0x7FC0 which should also be a NaN.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
iklemenskiTT: Thanks for the call-out, I'm going with a `v_and` refactoring instead, which cuts the fix overhead from 4 to 1 instructions on Blackhole. Added the measurement results in the PR description, and pushed an update.

mmoscickiTT (approved):

nmauriceTT (approved):

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
```diff
@@ -277,6 +284,28 @@ sfpi_inline sfpi::vFloat _sfpu_binary_power_f32_(sfpi::vFloat base, sfpi::vFloat
...
+    // SFPU `== 0` is bit-exact, so matching -0 needs an abs. Recompute it here
+    // rather than reusing abs_base: that live range would span the log2/exp2 body
+    // and spills this kernel.
+    // Fill 0 for every non-zero exponent, then narrow to the negative ones.
+    // v_and tightens the enclosing predicate in place, so it costs one compare
+    // where a second flat v_if would also save and restore the lane mask.
+    // pow == 0 fails the != 0 gate and keeps 1.
+    sfpi::vFloat abs_end = sfpi::abs(base);
+    v_if(abs_end == 0.f) {
+        v_if(pow != 0.f) {
```
Aswinmcw, line 300: `pow != 0.f` is true for a NaN exponent and `v_and(pow < 0.f)` is false for it, so a NaN exponent over a zero base now falls into the `y = 0.0f` fill.

Measured on p150b — this PR's BH headers JIT'd over `main`'s host build, cache cleared:

| | `main` | this PR | torch |
|---|---|---|---|
| `pow(0, nan)` fp32 | `inf` | **`0.0`** | `nan` |
| `pow(0, nan)` bf16 | `inf` | **`0.0`** | `nan` |

Both are wrong, so this isn't a correctness regression so much as a change of failure mode — but `inf` is loud and `+0` is silent, and a NaN lane over a zero base is exactly the masked-tensor shape this fix is aimed at. Binary only: unary is unaffected, since a NaN scalar fails `pow_scalar >= 0.0f` and takes the `<false>` NaN fill (measured `nan` on fp32, `inf` on bf16 via tt-llk#675).

Note this is specific to the `v_and` refactor — the `v_if(pow < 0)` / `v_elseif(pow > 0)` shape it replaced leaves NaN on the computed path, because neither arm fires. Swapping `!= 0.f` for `> 0.f` on its own won't do it: the `v_and` narrows inside that predicate, so the negative arm would end up empty. If the 3-instruction saving is worth keeping, the NaN has to be excluded before the fill rather than after.

Aswinmcw (approved):
Approving — the fix is right and the measurements back it. One behavioural note below, not a blocker.

VirdhatchaniKN (approved):

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary_pow.h
iklemenskiTT: Thanks for pointing this out. I'll merge as is and we can track it as follow-up in #53922.

Worth noting the guard shape isn't what determines this. I measured all three: the flat `v_if` pair, the nested `v_if`/`v_elseif` from #51468, and the merged `v_and` all return `0x00000000` for a NaN exponent. In the nested form the else-if arm does fire — `pow > 0.f` is a bit-pattern test and positive quiet NaN satisfies it — so the lane doesn't stay on the computed path. Pre-#53367, with no guard at all, that path gives `inf` on fp32 binary. So `inf → 0` traces to the base fix, and no restructuring recovers the louder value.

On excluding NaN before the fill: `pow != pow` doesn't work — it fires for `±inf` too and would turn the currently-correct `pow(0, +inf) = 0` into NaN. An integer-magnitude test (`as<vInt>(abs(pow)) > 0x7F800000`) does work and passes the full suite, but costs +7 SFPU instructions / ~224 cycles per tile (+4–8%), which is 2.3× what the refactor saved. Trade-off is laid out in #53922. Also worth noting #53367 incidentally fixed `pow(0, +inf)`, which returned `inf` before.

PR #50178, opened 2026-07-17
[codegen] Port untilize (build_untilize_tile) — validated on wormhole_b0 + blackhole

ebanerjeeTT:
## Ported: `untilize`

Automated port of `untilize` from tt-dm-codegen into a registered `ttnn.DeviceOperation` + `ProgramFactory` that reuses the generated kernels. Opened as a **draft** by the agentic port pipeline. Confirmed on 2 archs — `wormhole_b0` and `blackhole`. The full CI sweep runs via ordinary tt-metal CI.

### Performance
**wormhole_b0** — device time: median **0.30x** of native across 112 config(s) (best 0.23x, worst 0.82x). Ratio = ported device time / native device time; < 1.0 = ported is faster.

untilize device time — wormhole_b0 (device time ratio ported/native, sorted best→worst)

<details>
<summary>All 112 configs — wormhole_b0</summary>

| # | layout | shape | kwargs | native device µs | ported device µs | speedup (ported/native) |
|---:|---|---|---|---:|---:|---:|
| 1 | tile | `[4, 6, 106, 59]` | `memory_config=MemoryConfig(memory_layout=TensorMemoryLayout::INTERLEAVED,buffer_type=BufferType::DRAM,shard_spec=std::nullopt,nd_shard_spec=std::nullopt,created_with_nd_shard_spec=0)` | 28.154 | 6.494 | 0.231 |
| 2 | tile | `[102, 165]` | `memory_config=MemoryConfig(memory_layout=TensorMemoryLayout::INTERLEAVED,buffer_type=BufferType::DRAM,shard_spec=std::nullopt,nd_shard_spec=std::nullopt,created_with_nd_shard_spec=0)` | 17.132 | 4.039 | 0.236 |
| 3 | tile | `[3, 8, 103, 44]` | `memory_config=MemoryConfig(memory_layout=TensorMemoryLayout::INTERLEAVED,buffer_type=BufferType::DRAM,shard_spec=std::nullopt,nd_shard_spec=std::nullopt,created_with_nd_shard_spec=0)` | 28.205 | 6.654 | 0.236 |
| 4 | tile | `[132, 246]` | `memory_config=MemoryConfig(memory_layout=TensorMemoryLayout::INTERLEAVED,buffer_type=BufferType::DRAM,shard_spec=std::nullopt,nd_shard_spec=std::nullopt,created_with_nd_shard_spec=0)` | 18.912 | 4.462 | 0.236 |
[...]

PR #53399, opened 2026-08-17
[Bug Fix] untilize: budget codegen CB plan against live L1 occupancy

ebanerjeeTT:
## Problem

`ttnn.untilize()` aborts during program creation when there are L1-resident buffers (model weights, trace buffers) sitting low in L1:

```
RuntimeError: TT_THROW @ tt_metal/impl/program/program.cpp:1919: tt::exception
info:
Statically allocated circular buffers in program 488 clash with L1 buffers on core range [0-0 - 7-7].
L1 buffer allocated at 556864 and static circular buffer region ends at 914752
```

Stack: `ttnn::untilize` → `ttnn::prim::untilize_codegen` → `UntilizeCodegenDeviceOperation` → `ProgramImpl::validate_circular_buffer_region`.

**Reproducer (real hardware, N150):**
```
HF_MODEL=google/gemma-3-4b-it pytest \
  'models/demos/multimodal/gemma3/demo/vision_demo.py::test_multimodal_demo_text[wormhole_b0-performance-device_params0-1-batch1-multi-image-trace-normal-1]'
```

**Bisected on N150** to #50178:
- parent `70285f9c14890c1faf6a90580608c3b715e9ec8f` — **passes**
- `7c70c5a816472f77a817172dc62d11d7dd71cf47` — **fails** with the TT_THROW above

That commit rewired `ttnn::untilize()` to route eligible tensors through the new `untilize_codegen` program factory.

## Root cause

`usable_l1_bytes()` in `untilize_codegen_supported.cpp` computed the codegen CB budget as

```cpp
device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(HalMemType::L1)
```

i.e. against the whole of L1 above the allocator base. That is only correct on an otherwise-idle device.

Statically allocated circular buffers grow **upward** from the allocator's base L1 address, while L1 buffers are allocated **downward** from the top of L1. The space actually available to a program's CBs is therefore `lowest_occupied_compute_l1_address() - base` — the exact quantity `ProgramImpl::validate_circular_buffer_region()` checks the CB region end against.

The native path never had this problem: it lets the runtime CB allocator find free space. The codegen path pre-computes a client-side budget, and when that budget ignores already-resident L1 buffers, `plan_cb_depths()` happily selects a double-buffered plan that provably overlaps them — and the run dies in `validate_circular_buffer_region()` rather than degrading to a shallower plan.

With the CI numbers above: `lowest_occupied = 556864`, so the real budget is `556864 - base`, but the gate/factory planned against `1499136 - base` and produced a CB region ending at `914752`.

## Fix

Budget `usable_l1_bytes()` against the same frontier the validator uses:

```cpp
const uint32_t base = device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
const auto lowest_occupied = device->lowest_occupied_compute_l1_address();
const uint32_t ceiling =
    lowest_occupied.has_value() ? static_cast<uint32_t>(*lowest_occupied) : device->l1_size_per_core();
return ceiling > base ? ceiling - base : 0;
```

`std::nullopt` (nothing resident in L1) preserves the previous whole-of-L1 behaviour, so an idle device sees no change. This mirrors existing in-repo precedent — `get_max_l1_space()` in `ttnn/cpp/ttnn/operations/data_movement/common/common.cpp`, plus similar accounting in `matmul_utilities.cpp`, `transpose.cpp` and `reshard_rm_program_factory.cpp`.

Applied to the failing case, the plan now degrades from the double-buffered-input tier to the single-buffer tier and no longer clashes.

### Secondary hardening (strictly additive)

Because the budget can now be *smaller* than `plan_cb_depths()`'s floor tier (`cb_in + cb_out` at one slot each) — which has nothing left to degrade to and `TT_THROW`s — `supported_by_codegen()` gains a `min_cb_plan_fits()` floor check on **every** builder path, not just the column-parallel one it previously covered. A case that cannot fit even the floor now routes to `untilize_native` instead of aborting inside program creation, mirroring how the gate already defers to native for other out-of-scope shapes.

Notes on the bound used: `build_2d_column` plans on `wt/ncol` and `build_main_split` on `wt`, so `Wt` bounds both. `choose_2d_ncol()` lives in the factory's anonymous namespace, so the gate uses the looser `Wt` bound — which can only over-route to native, never under-route into a plan the factory cannot build.

**`is_demoted()` is unchanged.**

## Test

`tests/ttnn/unit_tests/operations/data_movement/test_untilize.py::test_untilize_codegen_with_resident_l1_buffers`

Reproduces the Gemma trace-mode scenario at the op level: allocates a persistent interleaved-L1 tensor (via `ttnn.allocate_tensor_on_device`, so there is no host-side cost) sized from `ttnn._ttnn.reports.get_device_info(device)` so that the remaining headroom above the CB base fits the single-buffer CB plan but **not** the double-buffered one, then runs `ttnn.untilize` and checks bit-exactness against the input.

The test asserts on `resident.buffer_address()` before running the op, so it fails loudly rather than passing vacuously if the reservation does not land where intended. It is device-agnostic (skips if L1 is too small) and cleans up in a `finally`.

Without the fix this test hits the same `clash with L1 buffers` `TT_THROW`; with it, the op selects the single-buffer plan and passes.

## Validation status — draft pending CI

This was developed by reading the allocator/program-validator code; **no hardware run has been done on this branch yet**. Before merge this still needs:

1. The specific failing Gemma-3-4B test above, on N150.
2. The full untilize performance sweep documented in the comments on #50178 — the gate is now sensitive to live L1 occupancy, so it is worth confirming that under normal free-L1 conditions nothing silently re-routes to native. (It should not: `kWideChunkThreshold` (800000) already bounds `2*Wt*2048` well below the full free-L1 budget of ~1391232 on WH.)

Marking draft until both are green.

Related: #50178

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:fix/untilize-codegen-l1-overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:fix/untilize-codegen-l1-overlap)

### DIFF
--- a/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
@@ -3056,8 +3056,9 @@ def test_untilize_sub_core_grids_single_tile_height(device):
     assert_equal(input_torch, ttnn.to_torch(output))
 
 
+@pytest.mark.parametrize("warmup", [False, True], ids=["cold", "warmup_then_constrain"])
 @pytest.mark.parametrize("headroom_regime", ["shallower_cb_plan", "no_cb_plan_fits"])
-def test_untilize_codegen_with_resident_l1_buffers(device, headroom_regime):
+def test_untilize_codegen_with_resident_l1_buffers(device, headroom_regime, warmup):
     """Regression: the codegen CB plan must be budgeted against LIVE L1 occupancy.
 
     Statically allocated CBs grow upward from the allocator's base L1 address while L1 tensors are
@@ -3079,9 +3080,11 @@ def test_untilize_codegen_with_resident_l1_buffers(device, headroom_regime):
       - no_cb_plan_fits: not even the single-buffered codegen plan fits, so the program factory
         must build the native-equivalent program instead of failing.
 
-    This asserts only on the result, never on which implementation served it: the choice is made
-    inside the program factory (once per program-cache miss, against one L1 snapshot) precisely so
-    that no observer outside it -- including this test -- has to reason about live L1 state.
+    This asserts only on the result, never on which implementation served it.
+
+    warmup=True first untilizes with free L1 (caches a roomier plan), deallocates that
+    output, then constrains L1. The CB-tier / Native-block-split is in the program-cache
+    key, so the second call must miss and rebuild rather than clash.
     """
     torch.manual_seed(42)
 
@@ -3123,6 +3126,11 @@ def test_untilize_codegen_with_resident_l1_buffers(device, headroom_regime):
     # untilize only relayouts values, so the input is its own golden.
     input_torch_tensor = torch.randn([1, 1, 32 * height_tiles, 32 * wt], dtype=torch.bfloat16)
     input_ttnn_tensor = ttnn.from_torch(input_torch_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    if warmup:
+        warmup_output = ttnn.untilize(input_ttnn_tensor)
+        assert_equal(input_torch_tensor, ttnn.to_torch(warmup_output))
+        ttnn.deallocate(warmup_output)
 
     # Interleaved L1 spreads pages round-robin over every bank, so tiles_per_bank tiles per bank
     # pushes the lowest occupied L1 address down on all of them. Allocated directly on device: the
--- a/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
@@ -3054,7 +3054,10 @@ def test_untilize_sub_core_grids_single_tile_height(device):
     sub_core_grids = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(num_cores - 1, 0))})
     output = ttnn.untilize(input_ttnn, use_multicore=True, sub_core_grids=sub_core_grids)
     assert_equal(input_torch, ttnn.to_torch(output))
-def test_untilize_codegen_with_resident_l1_buffers(device):
+
+
+@pytest.mark.parametrize("headroom_regime", ["shallower_cb_plan", "no_cb_plan_fits"])
+def test_untilize_codegen_with_resident_l1_buffers(device, headroom_regime):
     """Regression: the codegen CB plan must be budgeted against LIVE L1 occupancy.
 
     Statically allocated CBs grow upward from the allocator's base L1 address while L1 tensors are
@@ -3069,6 +3072,16 @@ def test_untilize_codegen_with_resident_l1_buffers(device):
     rather than degrading to a shallower CB plan (or falling back to native). Seen end-to-end on
     Gemma-3-4B in trace mode; reproduced here at the op level by pinning a persistent interleaved
     L1 tensor before untilizing.
+
+    Two regimes, both of which must simply produce the right answer:
+      - shallower_cb_plan: enough headroom for the single-buffered codegen plan but not the
+        double-buffered one, so the program factory must degrade the tier it picks.
+      - no_cb_plan_fits: not even the single-buffered codegen plan fits, so the program factory
+        must build the native-equivalent program instead of failing.
+
+    This asserts only on the result, never on which implementation served it: the choice is made
+    inside the program factory (once per program-cache miss, against one L1 snapshot) precisely so
+    that no observer outside it -- including this test -- has to reason about live L1 state.
     """
     torch.manual_seed(42)
 
@@ -3087,9 +3100,20 @@ def test_untilize_codegen_with_resident_l1_buffers(device):
     if info.cb_limit < double_in_bytes:
         pytest.skip(f"needs at least {double_in_bytes} B of CB space, device offers {info.cb_limit} B")
 
-    # Leave a headroom window that fits the single-buffer plan but NOT the double-buffered one, so
-    # the only non-clashing plan is unambiguous and an unbudgeted build provably overruns it.
-    headroom_target = single_buffer_bytes + 64 * 1024
+    if headroom_regime == "shallower_cb_plan":
+        # A headroom window that fits the single-buffer plan but NOT the double-buffered one, so
+        # the only non-clashing codegen plan is unambiguous and an unbudgeted build provably
+        # overruns it.
+        headroom_target = single_buffer_bytes + 64 * 1024
+        expected_headroom = (single_buffer_bytes, double_in_bytes)
+    else:
+        # Below every codegen tier, so the factory has to hand off to native. Still generous
+        # enough for native's own blocked CBs, which it sizes against the same live L1 space.
+        headroom_target = 256 * 1024
+        if headroom_target >= single_buffer_bytes:
+            pytest.skip("device L1 headroom cannot be driven below the smallest codegen CB plan")
+        expected_headroom = (0, single_buffer_bytes)
+
     tiles_per_bank = (info.cb_limit - headroom_target) // TILE_BYTES
     if tiles_per_bank <= 0:
         pytest.skip("device L1 is too small to leave a meaningful headroom window")
@@ -3114,9 +3138,10 @@ def test_untilize_codegen_with_resident_l1_buffers(device):
         # Guard against a vacuous pass: if the reservation did not land where intended there is no
         # L1 pressure left to regress against.
         actual_headroom = resident.buffer_address() - info.address_at_first_l1_cb_buffer
-        assert single_buffer_bytes <= actual_headroom < double_in_bytes, (
-            f"resident L1 buffer left {actual_headroom} B above the CB base; the test needs it in "
-            f"[{single_buffer_bytes}, {double_in_bytes}) so that only the single-buffer CB plan fits"
+        low, high = expected_headroom
+        assert low <= actual_headroom < high, (
+            f"resident L1 buffer left {actual_headroom} B above the CB base; the {headroom_regime} "
+            f"regime needs it in [{low}, {high})"
         )
 
         output = ttnn.untilize(input_ttnn_tensor)
--- a/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
@@ -3054,3 +3054,73 @@ def test_untilize_sub_core_grids_single_tile_height(device):
     sub_core_grids = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(num_cores - 1, 0))})
     output = ttnn.untilize(input_ttnn, use_multicore=True, sub_core_grids=sub_core_grids)
     assert_equal(input_torch, ttnn.to_torch(output))
+def test_untilize_codegen_with_resident_l1_buffers(device):
+    """Regression: the codegen CB plan must be budgeted against LIVE L1 occupancy.
+
+    Statically allocated CBs grow upward from the allocator's base L1 address while L1 tensors are
+    allocated downward from the top of L1, so the space actually available to a program's CBs is
+    lowest_occupied_compute_l1_address() - base, not the whole of L1. The codegen path originally
+    planned against the whole of L1, which is only correct on an otherwise-idle device; with a
+    model's weights/trace buffers resident in L1 it planned a CB region that overlapped them and
+    ProgramImpl::validate_circular_buffer_region() aborted the run with
+
+        Statically allocated circular buffers in program N clash with L1 buffers on core range ...
+
+    rather than degrading to a shallower CB plan (or falling back to native). Seen end-to-end on
+    Gemma-3-4B in trace mode; reproduced here at the op level by pinning a persistent interleaved
+    L1 tensor before untilizing.
+    """
+    torch.manual_seed(42)
+
+    TILE_BYTES = 2048  # bfloat16 tile
+    info = ttnn._ttnn.reports.get_device_info(device)
+
+    # Wt wide enough that the whole-of-L1 budget selects a double-buffered CB plan, but still
+    # inside the gate's wide-chunk threshold (2 * Wt * 2048 <= 800_000 => Wt <= 195). Two tile
+    # rows keep this on the row-parallel writer, whose CB plan is sized by Wt.
+    wt = 192
+    height_tiles = 2
+    single_buffer_bytes = 2 * wt * TILE_BYTES  # cb_in + cb_out, one slot each
+    double_in_bytes = 3 * wt * TILE_BYTES  # 2x cb_in + 1x cb_out
+
+    # cb_limit is exactly the whole-of-L1 budget the buggy plan used: worker L1 minus the base.
+    if info.cb_limit < double_in_bytes:
+        pytest.skip(f"needs at least {double_in_bytes} B of CB space, device offers {info.cb_limit} B")
+
+    # Leave a headroom window that fits the single-buffer plan but NOT the double-buffered one, so
+    # the only non-clashing plan is unambiguous and an unbudgeted build provably overruns it.
+    headroom_target = single_buffer_bytes + 64 * 1024
+    tiles_per_bank = (info.cb_limit - headroom_target) // TILE_BYTES
+    if tiles_per_bank <= 0:
+        pytest.skip("device L1 is too small to leave a meaningful headroom window")
+
+    # Built before the L1 reservation so that any device-side tilize in from_torch runs with the
+    # usual amount of L1 available; only the untilize under test should see the pressure.
+    # untilize only relayouts values, so the input is its own golden.
+    input_torch_tensor = torch.randn([1, 1, 32 * height_tiles, 32 * wt], dtype=torch.bfloat16)
+    input_ttnn_tensor = ttnn.from_torch(input_torch_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    # Interleaved L1 spreads pages round-robin over every bank, so tiles_per_bank tiles per bank
+    # pushes the lowest occupied L1 address down on all of them. Allocated directly on device: the
+    # contents are irrelevant, only the occupancy matters.
+    resident = ttnn.allocate_tensor_on_device(
+        ttnn.Shape([1, 1, 32 * tiles_per_bank, 32 * info.l1_num_banks]),
+        ttnn.bfloat16,
+        ttnn.TILE_LAYOUT,
+        device,
+        ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.L1),
+    )
+    try:
+        # Guard against a vacuous pass: if the reservation did not land where intended there is no
+        # L1 pressure left to regress against.
+        actual_headroom = resident.buffer_address() - info.address_at_first_l1_cb_buffer
+        assert single_buffer_bytes <= actual_headroom < double_in_bytes, (
+            f"resident L1 buffer left {actual_headroom} B above the CB base; the test needs it in "
+            f"[{single_buffer_bytes}, {double_in_bytes}) so that only the single-buffer CB plan fits"
+        )
+
+        output = ttnn.untilize(input_ttnn_tensor)
+
+        assert_equal(input_torch_tensor, ttnn.to_torch(output))
+    finally:
+        ttnn.deallocate(resident)
--- a/ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_cb_plan.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_cb_plan.cpp
@@ -1,0 +1,262 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#include "untilize_codegen_cb_plan.hpp"
+
+#include <algorithm>
+
+#include <tt-metalium/constants.hpp>
+#include <tt-metalium/core_coord.hpp>
+#include <tt-metalium/host_api.hpp>
+#include <tt-metalium/tt_backend_api_types.hpp>
+#include <tt-metalium/work_split.hpp>
+
+#include "ttnn/operations/core/work_split/work_split_tilize.hpp"
+#include "ttnn/operations/data_movement/common/common.hpp"
+
+using namespace tt::constants;
+using namespace tt::tt_metal;
+
+namespace ttnn::prim::untilize_codegen_detail {
+
+uint32_t compute_block_ct_dim(uint32_t wt, bool fp32) {
+    uint32_t max_bct = fp32 ? 4 : 8;
+    for (uint32_t bct = max_bct; bct >= 1; --bct) {
+        if (wt % bct == 0) {
+            return bct;
+        }
+    }
+    return 1;
+}
+
+uint32_t choose_2d_ncol(uint32_t total_tile_rows, uint32_t wt, uint32_t valid_cores) {
+    if (total_tile_rows >= valid_cores || wt < 2) {
+        return 1;
+    }
+    uint32_t max_ncol = std::min(valid_cores / total_tile_rows, wt);
+    uint32_t best = 1;
+    for (uint32_t d = 2; d <= max_ncol; ++d) {
+        if (wt % d == 0) {
+            best = d;
+        }
+    }
+    return best;
+}
+
+std::optional<CbPlan> plan_cb_depths(
+    uint64_t usable_l1, uint32_t pages_per_unit, uint32_t page_size, uint32_t block_units) {
+    uint64_t p = pages_per_unit;
+    uint64_t ts = page_size;
+    uint64_t double_both = (2 * p + 2 * p) * ts;
+    uint64_t double_in = (2 * p + p) * ts;
+    uint64_t single_both = (p + p) * ts;
+    if (double_both <= usable_l1) {
+        return CbPlan{static_cast<uint32_t>(2 * p), static_cast<uint32_t>(2 * p), pages_per_unit};
+    }
+    if (double_in <= usable_l1) {
+        return CbPlan{static_cast<uint32_t>(2 * p), pages_per_unit, pages_per_unit};
+    }
+    if (single_both <= usable_l1) {
+        return CbPlan{pages_per_unit, pages_per_unit, block_units};
+    }
+    return std::nullopt;
+}
+
+namespace {
+
+bool needs_dst_accum(DataType dtype) {
+    return dtype == DataType::FLOAT32 || dtype == DataType::INT32 || dtype == DataType::UINT32;
+}
+
+struct PaddedGrid {
+    uint32_t wt;
+    uint32_t total_tile_rows;
+    bool tile_aligned;
+};
+
+PaddedGrid padded_grid(const Tensor& input) {
+    const auto& padded_shape = input.padded_shape();
+    uint32_t rank = padded_shape.rank();
+    uint32_t w = padded_shape[-1];
+    uint32_t h = padded_shape[-2];
+    uint32_t nc = 1;
+    for (uint32_t i = 0; i + 2 < rank; ++i) {
+        nc *= padded_shape[i];
+    }
+    uint32_t wt = w / TILE_WIDTH;
+    uint32_t ht = h / TILE_HEIGHT;
+    const auto& logical_shape = input.logical_shape();
+    bool tile_aligned = logical_shape[-2] % TILE_HEIGHT == 0 && logical_shape[-1] % TILE_WIDTH == 0;
+    return PaddedGrid{wt, nc * ht, tile_aligned};
+}
+
+struct PagesAndBlock {
+    uint32_t pages_per_unit;
+    uint32_t block_units;
+};
+
+PagesAndBlock pages_for_builder(const Tensor& input, bool fp32) {
+    auto g = padded_grid(input);
+    auto* device = input.device();
+    auto grid = device->compute_with_storage_grid_size();
+    uint32_t valid_cores = static_cast<uint32_t>(grid.x) * static_cast<uint32_t>(grid.y);
+
+    if (!g.tile_aligned) {
+        return {g.wt, compute_block_ct_dim(g.wt, fp32)};
+    }
+    if (g.total_tile_rows == 1 && g.wt > 1) {
+        auto [_num_cores, _core_range, cg1, cg2, tpc1, tpc2] =
+            tt::tt_metal::split_work_to_cores(grid, g.wt, /*row_wise=*/true);
+        uint32_t max_tpc = std::max(tpc1, cg2.empty() ? 0u : tpc2);
+        return {max_tpc, compute_block_ct_dim(max_tpc, fp32)};
+    }
+    if (g.wt > 1) {
+        uint32_t ncol = choose_2d_ncol(g.total_tile_rows, g.wt, valid_cores);
+        if (ncol >= 2) {
+            uint32_t tpc = g.wt / ncol;
+            return {tpc, compute_block_ct_dim(tpc, fp32)};
+        }
+    }
+    return {g.wt, compute_block_ct_dim(g.wt, fp32)};
+}
+
+CodegenCbPlan tier_from_depths(const CbPlan& plan, uint32_t pages_per_unit) {
+    if (plan.cb_in_depth == 2 * pages_per_unit && plan.cb_out_depth == 2 * pages_per_unit) {
+        return CodegenCbPlan::DoubleBoth;
+    }
+    if (plan.cb_in_depth == 2 * pages_per_unit && plan.cb_out_depth == pages_per_unit) {
+        return CodegenCbPlan::DoubleIn;
+    }
+    return CodegenCbPlan::SingleBoth;
+}
+
+NativeCacheIdentity identity_from_ncores(
+    const ttnn::NcoresWHsb& sb, uint32_t width_tiles, uint32_t height_tiles) {
+    NativeCacheIdentity id;
+    id.enough_space_height = false;
+    id.split_valid = true;
+    id.ncores = sb.ncores;
+    id.nblocks_per_core = sb.nblocks_per_core;
+    id.single_block_size = sb.single_block_size;
+    if (id.single_block_size == 0) {
+        id.split_valid = false;
+        return id;
+    }
+    uint32_t total_blocks_width = sb.total_blocks_width;
+    uint32_t total_blocks_height = sb.total_blocks_height;
+    id.full_cores_per_row = width_tiles / id.single_block_size;
+    id.has_cliff_row = (id.full_cores_per_row < total_blocks_width);
+    id.full_cores_per_col = height_tiles / id.single_block_size;
+    id.has_cliff_col = (id.full_cores_per_col < total_blocks_height);
+    id.single_block_size_cliff_row = width_tiles - (id.full_cores_per_row * id.single_block_size);
+    id.single_block_size_cliff_col = height_tiles - (id.full_cores_per_col * id.single_block_size);
+    id.single_sub_block_size = sb.single_sub_block_size;
+    return id;
+}
+
+NativeCacheIdentity block_split_identity(
+    uint32_t grid_area,
+    uint32_t nblocks,
+    uint32_t width_tiles,
+    uint32_t height_tiles,
+    uint32_t cb_block_size_limit) {
+    auto wh = ttnn::compute_ncores_wh(grid_area, nblocks, width_tiles, height_tiles);
+    ttnn::NcoresWHsb sb{
+        wh.ncores,
+        wh.nblocks_per_core,
+        wh.total_blocks_width,
+        wh.total_blocks_height,
+        wh.single_block_size,
+        wh.single_block_size};
+    if (cb_block_size_limit >= 1 && wh.single_block_size > cb_block_size_limit) {
+        auto maybe = ttnn::try_compute_ncores_wh_sb(
+            grid_area, nblocks, width_tiles, height_tiles, cb_block_size_limit);
+        if (!maybe.has_value()) {
+            NativeCacheIdentity sentinel;
+            sentinel.enough_space_height = false;
+            sentinel.split_valid = false;
+            return sentinel;
+        }
+        sb = *maybe;
+    }
+    return identity_from_ncores(sb, width_tiles, height_tiles);
+}
+
+}  // namespace
+
+ChosenCodegenCbPlan choose_codegen_cb_plan(
+    const UntilizeCodegenOperationAttributes& attrs, const UntilizeCodegenTensorArgs& tensor_args) {
+    const Tensor& input = tensor_args.input;
+    auto out_spec = UntilizeCodegenDeviceOperation::compute_output_specs(attrs, tensor_args);
+    DataType in_dtype = input.dtype();
+    DataType out_dtype = out_spec.data_type();
+    bool fp32 = needs_dst_accum(in_dtype);
+    auto in_fmt = datatype_to_dataformat_converter(in_dtype);
+    auto out_fmt = datatype_to_dataformat_converter(out_dtype);
+    uint32_t tile_size_for_planning = std::max(tt::tile_size(in_fmt), tt::tile_size(out_fmt));
+    uint64_t usable_l1 = ttnn::operations::data_movement::get_max_l1_space(input);
+    auto pages = pages_for_builder(input, fp32);
+    auto depths = plan_cb_depths(usable_l1, pages.pages_per_unit, tile_size_for_planning, pages.block_units);
+    if (!depths.has_value()) {
+        return ChosenCodegenCbPlan{CodegenCbPlan::Native, std::nullopt};
+    }
+    return ChosenCodegenCbPlan{tier_from_depths(*depths, pages.pages_per_unit), depths};
+}
+
+NativeCacheIdentity native_cache_identity(
+    const UntilizeCodegenOperationAttributes& attrs,
+    const UntilizeCodegenTensorArgs& tensor_args,
+    CodegenCbPlan plan) {
+    (void)attrs;
+    NativeCacheIdentity zeros;
+    if (plan != CodegenCbPlan::Native) {
+        return zeros;
+    }
+
+    namespace dm = ttnn::operations::data_movement;
+    const Tensor& input = tensor_args.input;
+    auto in_fmt = datatype_to_dataformat_converter(input.dtype());
+    uint32_t single_tile_size = tt::tile_size(in_fmt);
+    uint32_t num_tiles_per_row = input.padded_shape()[-1] / TILE_WIDTH;
+    const bool enough_space_height =
+        dm::is_enough_space(input, single_tile_size, single_tile_size, num_tiles_per_row);
+    if (enough_space_height) {
+        NativeCacheIdentity id;
+        id.enough_space_height = true;
+        return id;
+    }
+
+    auto out_spec = UntilizeCodegenDeviceOperation::compute_output_specs(attrs, tensor_args);
+    auto out_fmt = datatype_to_dataformat_converter(out_spec.data_type());
+    uint32_t output_single_tile_size = tt::tile_size(out_fmt);
+    uint32_t max_l1_size = dm::get_max_l1_space(input);
+    uint32_t denom = single_tile_size + output_single_tile_size;
+    uint32_t cb_block_size_limit = denom == 0 ? 0 : max_l1_size / denom;
+
+    const auto& logical_shape = input.logical_shape();
+    const bool tile_aligned = logical_shape[-2] % TILE_HEIGHT == 0 && logical_shape[-1] % TILE_WIDTH == 0;
+    auto* device = input.device();
+    CoreCoord grid_size = device->compute_with_storage_grid_size();
+
+    if (tile_aligned) {
+        uint32_t a_tile_width = input.tensor_spec().tile().get_width();
+        uint32_t a_tile_height = input.tensor_spec().tile().get_height();
+        uint32_t width_tiles = input.padded_shape()[-1] / a_tile_width;
+        uint32_t height_tiles = input.padded_shape()[-2] / a_tile_height;
+        uint32_t nblocks = (input.padded_shape()[-1] * input.padded_shape()[-2]) / (a_tile_height * a_tile_width);
+        uint32_t grid_area = static_cast<uint32_t>(grid_size.x) * static_cast<uint32_t>(grid_size.y);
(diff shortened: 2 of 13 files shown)


ebanerjeeTT:
Pushed `68f60f8ee5`, which replaces the approach in `bdafe915`.

**Why the first attempt failed.** `bdafe915` made `supported_by_codegen()` consult live L1 occupancy. That predicate is evaluated independently at three sites in one dispatch — `ttnn::untilize`'s routing gate, `untilize_force_codegen`'s `TT_FATAL`, and `validate_on_program_cache_miss` — and they only agree because it is a pure function of static properties. Once it depended on mutable device state, routing saw `true`, dispatched to codegen, and validate rejected the same tensor moments later (the op's own `create_output_tensors()` allocates the output and moves `lowest_occupied_compute_l1_address` in between), giving the `TT_FATAL @ untilize_codegen_device_operation.cpp:28` reported on N150.

**What this commit does instead.** `supported_by_codegen()`/`usable_l1_bytes()` go back to the pure static form, with the invariant documented. The live-L1 decision moves to the single point that can observe it once, consistently, after the op has committed and allocated its output: `UntilizeCodegenProgramFactory::create_descriptor`. It samples `get_max_l1_space(input)` once, threads it through `CommonArgs`, and if no codegen CB tier fits below the resident L1 buffers it emits the native-equivalent program for the same tensors (delegating to `UntilizeDeviceOperation`/`UntilizeWithUnpaddingDeviceOperation`'s own `select_program_factory` + `create_descriptor` rather than duplicating them) instead of failing.

**No fast-path cost.** `create_descriptor` runs only on a program-cache *miss*. Every codegen builder and every native factory the fallback can reach emplaces runtime args with a `Buffer*` first argument, so buffer bindings always resolve and cache *hits* take `apply_descriptor`'s `apply_resolved_bindings` path without re-entering the factory. And no previously-working program changes shape: the three CB tiers are strictly ordered by size and the live gap is always <= the whole-L1 budget planned against before, so the only plans that differ are exactly the ones `ProgramImpl::validate_circular_buffer_region()` would have rejected anyway.

**Not yet validated on hardware** — no device available here. Needs a run of `test_untilize_codegen_with_resident_l1_buffers` (now parametrized over both L1-pressure regimes) plus the Gemma-3-4B trace-mode e2e test and the untilize perf sweep.

One known interaction to be aware of: under the opt-in `ENABLE_DESCRIPTOR_PATCHING_PARITY_CHECK` build (OFF by default), `create_descriptor` is re-invoked on cache hits and diffed against the cached descriptor; if L1 occupancy shifts enough between miss and hit to flip a tier, that debug check could report a spurious mismatch. It never runs in production dispatch.

ebanerjeeTT:
## Cross-arch re-verification of the untilize codegen sweep — no regressions

Item 2 of this PR's own validation list. Re-ran the full 112-config untilize codegen sweep from #50178 on both archs at this PR's head `68f60f8e`, through `tt-dm-codegen`'s cross-arch framework.

The framework refuses this PR directly — `cross_arch.guard_ref` rejects any head branch that is not
`port/*`, and `dispatch` reads the op and manifest from a `codegen-port-results:v2` block this body
does not carry. So the same commit was pushed under a `port/*` name and dispatched from #53420
(scaffold only, no code differs, results below). Runs:
[wormhole_b0/N300](https://github.com/tenstorrent/tt-metal/actions/runs/32065863722) ·
[blackhole/P150](https://github.com/tenstorrent/tt-metal/actions/runs/32065876271), both at
`68f60f8e`, 112 cases each.

### Correctness

| | board | bit-exact vs native | in-tree unit tests | routing matrix |
|---|---|---|---|---|
| **wormhole_b0** | N300 | 112 / 112 | `unit_tests_exit=0` (920 passed, 4 skipped, 2 xfailed) | passed |
| **blackhole** | P150 | 112 / 112 | `unit_tests_exit=0` (922 passed, 4 skipped) | passed |

The unit-test leg includes this PR's new test. Both regimes passed on both archs:
`test_untilize_codegen_with_resident_l1_buffers[headroom_regime=shallower_cb_plan]` and
`[headroom_regime=no_cb_plan_fits]`.

### Performance — device time, ported / native (lower is faster)

| | board | median | best | worst | p25 / p75 | at or above parity |
|---|---|---|---|---|---|---|
| **wormhole_b0** | N300 | **0.285x** | 0.226x | 0.866x | 0.255 / 0.347 | 0 of 112 |
| **blackhole** | P150 | **0.260x** | 0.223x | 0.799x | 0.243 / 0.313 | 0 of 112 |

Against the post-review baseline in [this comment](https://github.com/tenstorrent/tt-metal/pull/50178#issuecomment-5258454601) on #50178 (wormhole_b0 median 0.286x, best 0.225x, worst 0.868x, p25/p75 0.252/0.346; blackhole median 0.261x, best 0.223x, worst 0.800x, p25/p75 0.244/0.312), every statistic is within 0.002 on both archs.

Per-config, joining all 112 configs against the last CI tables with per-case numbers (`3c131242`,
[wormhole_b0](https://github.com/tenstorrent/tt-metal/pull/50178#issuecomment-5182879173) ·
[blackhole](https://github.com/tenstorrent/tt-metal/pull/50178#issuecomment-5182878977)):

- median delta −0.016 on both archs, i.e. slightly faster overall;
- range −0.057 to +0.057, which is the run-to-run noise floor this framework already documents for
  a byte-identical binary;
- the only configs moving more than 0.05 in the slower direction are the four single-tile shapes
  (`[32, 32]` and `[2, 32, 32]`, DRAM and L1). Those are 2–3.5 µs ops where the codegen path is
  still 0.4–0.6 µs faster than native in absolute terms, so a 0.05 ratio move is roughly 150 ns of
  fixed dispatch delta. Blackhole moved 4 configs the same distance in the *faster* direction.

The 56 configs whose output lives in L1 — the half this change actually touches, since the output
tensor itself lowers the ceiling `get_max_l1_space()` now budgets against — show no separate
signature: wormhole_b0 median 0.308x, blackhole 0.273x, in line with their DRAM counterparts.

Wall-clock flags were the usual CI noise and no worse than the baseline run: wormhole_b0 flagged 1
case (4 more cleared on the in-job retest), blackhole 3 (4 cleared), against 1 and 10 respectively
at `3c131242`. Every flagged case still beats native on device.

### What this run does and does not cover

No config re-routed. Nothing in the sweep landed at or above device parity, which is what a
fall-through to the native-equivalent program would look like, so on an idle device the tightened
budget selects the same CB tier it did before — consistent with `kWideChunkThreshold` bounding
`2*Wt*2048` (max `Wt` here is 8, so 64 KB) far below the ~1.39 MB of free L1.

That is also the limit of what a sweep can show: it runs on a near-idle device, so it does not
exercise the new `std::nullopt` → `build_native_equivalent` path. That path is covered by the
`no_cb_plan_fits` regime of the new unit test, which passed on both archs above.

Item 1 of the validation list — the Gemma-3-4B trace-mode test on N150 — is not covered here.

ebanerjeeTT:
## Why this PR exists, and what it costs the sweep

Cross-posting the per-config charts from the cross-arch re-verification [above](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320293410), plus the nightly history that pins the break to #50178. Together: the failure is real and still happening on `main`, and fixing it costs the untilize codegen sweep nothing.

### The break

`Gemma-3-4B e2e tests [wh_n150]` in the (Tier 2) Models E2E nightly, either side of #50178 (merged `7c70c5a81`, 2026-08-14 13:56 UTC):

| nightly | commit | contains #50178 | result |
|---|---|---|---|
| 2026-08-14 (03:46 UTC) | `bef164b30` | no | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/31767822219/job/94670230682) |
| 2026-08-15 | `3ce2764d9` | yes | [failed](https://github.com/tenstorrent/tt-metal/actions/runs/31861408948/job/94957983588) |
| 2026-08-17 | `e16f6cbfb` | yes | [failed](https://github.com/tenstorrent/tt-metal/actions/runs/31990919029/job/95290902099) — most recent |

The last green nightly is the last one whose commit does not contain #50178, and the first red one is the first that does. The most recent failure carries the same signature as the original report — `Statically allocated circular buffers in program 488 clash with L1 buffers on core range [0-0 - 7-7]`, `ttnn::prim::untilize_codegen` in the stack, same test id `test_multimodal_demo_text[wormhole_b0-performance-device_params0-1-batch1-multi-image-trace-normal-1]` — so this is still broken on `main` today, not a one-night flake.

That same job now **passes** on this branch at `68f60f8e` — [run 32070288203](https://github.com/tenstorrent/tt-metal/actions/runs/32070288203/job/95513658884), details [below](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320588100).

### The cost, per config

Both charts are the full 112-config sweep at this PR's head `68f60f8e`, ported vs native. Bars under the grey parity line are configs where the codegen path is faster. Nothing touches the line on either arch, which is also how you can see that no config fell through to the native-equivalent program the fix adds — that would read as 1.0.

#### wormhole_b0 / N300 — [run](https://github.com/tenstorrent/tt-metal/actions/runs/32065863722)

```mermaid
%%{init: {"themeVariables": {"xyChart": {"plotColorPalette": "#d9d9d9, #2f6feb, #8b949e, #d1642c"}}}}%%
xychart-beta
    title "wormhole_b0/N300 — device time and wall clock vs native"
    x-axis "config # (sorted best → worst by device time)" 1 --> 112
    y-axis "ported / native  (under 1.0 = ported faster)" 0 --> 1.13
    line "re-measured" [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.13, 0, 1.13, 0, 0, 0, 1.13, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.13, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.13, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
    bar "device time" [0.226, 0.227, 0.227, 0.228, 0.228, 0.228, 0.228, 0.228, 0.229, 0.229, 0.229, 0.229, 0.230, 0.230, 0.231, 0.232, 0.232, 0.233, 0.233, 0.234, 0.235, 0.239, 0.243, 0.243, 0.245, 0.246, 0.250, 0.254, 0.257, 0.258, 0.258, 0.264, 0.265, 0.265, 0.266, 0.270, 0.271, 0.271, 0.275, 0.275, 0.276, 0.276, 0.278, 0.278, 0.278, 0.280, 0.280, 0.280, 0.281, 0.281, 0.281, 0.282, 0.283, 0.285, 0.285, 0.285, 0.286, 0.286, 0.286, 0.288, 0.288, 0.293, 0.294, 0.297, 0.302, 0.303, 0.304, 0.305, 0.308, 0.309, 0.312, 0.314, 0.318, 0.320, 0.321, 0.321, 0.324, 0.325, 0.326, 0.329, 0.330, 0.335, 0.337, 0.345, 0.348, 0.351, 0.357, 0.369, 0.370, 0.371, 0.380, 0.382, 0.398, 0.406, 0.417, 0.433, 0.437, 0.439, 0.443, 0.447, 0.763, 0.770, 0.779, 0.779, 0.780, 0.780, 0.783, 0.785, 0.846, 0.846, 0.861, 0.866]
    line "parity 1.0" [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
    line "wall clock" [0.728, 0.965, 0.958, 0.946, 0.963, 0.842, 0.965, 0.964, 0.972, 0.799, 0.983, 0.951, 0.906, 0.980, 0.947, 0.956, 0.954, 0.982, 0.980, 0.911, 0.988, 0.710, 0.934, 1.024, 0.931, 0.822, 0.878, 0.978, 0.943, 0.822, 0.968, 0.870, 0.804, 0.784, 0.824, 0.895, 0.826, 0.925, 1.022, 0.824, 0.973, 0.971, 0.973, 0.706, 0.977, 0.760, 0.980, 0.962, 0.950, 0.964, 0.884, 0.956, 0.721, 0.962, 0.973, 1.018, 0.966, 0.901, 0.981, 0.984, 0.962, 0.756, 0.939, 0.927, 1.013, 0.804, 0.917, 0.956, 0.963, 0.824, 0.541, 0.793, 0.831, 0.903, 0.787, 0.962, 0.973, 0.710, 0.893, 0.847, 0.843, 0.852, 0.966, 0.654, 0.960, 0.845, 0.748, 0.930, 0.700, 0.916, 0.893, 0.899, 0.819, 0.614, 0.791, 0.999, 1.006, 0.542, 1.003, 0.993, 0.988, 1.007, 0.986, 1.001, 0.989, 0.975, 0.984, 0.991, 0.973, 0.980, 0.981, 0.976]
```
**Legend** — `2f6feb` bars: device time; `8b949e` flat line: parity with native; `#d1642c` line: wall clock, which includes host overhead and is noisier; `#d9d9d9` vertical marks: the 5 case(s) the in-job retest re-measured, whose wall figure is a median over several windows, not a single shot. Those are configs 19, 21, 25, 42, 61, marked † in the table.

#### blackhole / P150 — [run](https://github.com/tenstorrent/tt-metal/actions/runs/32065876271)

```mermaid
%%{init: {"themeVariables": {"xyChart": {"plotColorPalette": "#d9d9d9, #2f6feb, #8b949e, #d1642c"}}}}%%
xychart-beta
    title "blackhole/P150 — device time and wall clock vs native"
    x-axis "config # (sorted best → worst by device time)" 1 --> 112
    y-axis "ported / native  (under 1.0 = ported faster)" 0 --> 1.31
    line "re-measured" [1.31, 0, 0, 0, 0, 1.31, 0, 0, 0, 1.31, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.31, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.31, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1.31, 1.31, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
    bar "device time" [0.223, 0.225, 0.228, 0.228, 0.228, 0.229, 0.229, 0.229, 0.230, 0.230, 0.232, 0.232, 0.232, 0.233, 0.234, 0.234, 0.236, 0.237, 0.238, 0.239, 0.239, 0.240, 0.241, 0.242, 0.242, 0.242, 0.243, 0.243, 0.243, 0.244, 0.245, 0.245, 0.246, 0.247, 0.248, 0.248, 0.248, 0.249, 0.250, 0.251, 0.251, 0.252, 0.253, 0.253, 0.253, 0.253, 0.253, 0.254, 0.254, 0.254, 0.255, 0.256, 0.258, 0.258, 0.259, 0.260, 0.261, 0.264, 0.264, 0.268, 0.270, 0.273, 0.273, 0.274, 0.274, 0.275, 0.276, 0.276, 0.277, 0.277, 0.280, 0.283, 0.283, 0.285, 0.290, 0.294, 0.296, 0.300, 0.301, 0.305, 0.305, 0.310, 0.310, 0.312, 0.313, 0.316, 0.316, 0.325, 0.329, 0.332, 0.335, 0.341, 0.356, 0.366, 0.404, 0.424, 0.426, 0.432, 0.433, 0.541, 0.751, 0.752, 0.755, 0.755, 0.756, 0.757, 0.763, 0.764, 0.774, 0.776, 0.797, 0.799]
    line "parity 1.0" [1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
    line "wall clock" [1.026, 0.945, 0.950, 0.948, 0.945, 0.959, 0.998, 0.906, 0.948, 0.869, 0.930, 0.936, 0.907, 0.928, 0.773, 1.032, 0.953, 0.457, 0.961, 0.892, 0.947, 0.658, 0.942, 0.952, 0.919, 0.924, 0.934, 0.927, 0.942, 0.975, 0.948, 0.460, 0.869, 0.979, 0.300, 0.928, 0.941, 0.957, 0.717, 0.935, 0.934, 0.930, 0.858, 0.932, 0.899, 0.732, 0.965, 0.342, 0.906, 0.414, 0.972, 0.913, 0.921, 0.941, 0.400, 0.873, 0.548, 0.929, 0.815, 0.962, 0.903, 0.977, 0.926, 0.848, 0.937, 0.934, 0.905, 0.957, 0.931, 0.663, 0.924, 0.830, 0.557, 0.769, 0.739, 0.844, 0.511, 0.665, 0.493, 0.924, 0.874, 0.922, 0.430, 0.881, 0.917, 0.879, 0.543, 0.530, 0.525, 0.806, 0.655, 0.568, 0.865, 0.648, 0.895, 0.979, 1.193, 1.019, 1.005, 0.520, 0.664, 0.977, 0.979, 0.971, 0.998, 0.983, 1.006, 0.960, 0.932, 0.972, 0.968, 0.574]
```
**Legend** — `2f6feb` bars: device time; `8b949e` flat line: parity with native; `#d1642c` line: wall clock, which includes host overhead and is noisier; `#d9d9d9` vertical marks: the 7 case(s) the in-job retest re-measured, whose wall figure is a median over several windows, not a single shot. Those are configs 1, 6, 10, 66, 85, 97, 98, marked † in the table.

Median device time is 0.285x of native on wormhole_b0 and 0.260x on blackhole, both within 0.002 of the pre-fix baseline, and the per-config deltas sit inside the run-to-run noise floor. The full comparison, including the four single-tile configs that moved and why 150 ns reads as a 0.05 ratio move on a 3 µs op, is in the [comment above](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320293410).

ebanerjeeTT:
## Gemma-3-4B passes on this branch — validation item 1 closed

[Run 32070288203](https://github.com/tenstorrent/tt-metal/actions/runs/32070288203) — `models-e2e-tests / Gemma-3-4B e2e tests [wh_n150]`, [job](https://github.com/tenstorrent/tt-metal/actions/runs/32070288203/job/95513658884) — **success** at `68f60f8e`, which is this PR's head. Same workflow, same job name, and same N150 SKU as the nightly that has been failing on `main` since #50178 merged.

The specific test from the bug report passed:

```
PASSED models/demos/multimodal/gemma3/demo/vision_demo.py::test_multimodal_demo_text[wormhole_b0-performance-device_params0-1-batch1-multi-image-trace-normal-1]
```

The job log contains no `clash with L1 buffers`, and no `TT_THROW` or `TT_FATAL` of any kind — against 3 occurrences of the clash in [yesterday's nightly](https://github.com/tenstorrent/tt-metal/actions/runs/31990919029/job/95290902099) on `main`.

So the before/after is now complete on the same job, same SKU:

| commit | contains #50178 | contains this fix | Gemma-3-4B e2e [wh_n150] |
|---|---|---|---|
| `bef164b30` (nightly, 08-14) | no | no | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/31767822219/job/94670230682) |
| `e16f6cbfb` (nightly, 08-17) | yes | no | [failed](https://github.com/tenstorrent/tt-metal/actions/runs/31990919029/job/95290902099) |
| `68f60f8e` (this PR) | yes | yes | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/32070288203/job/95513658884) |

Note this is the first hardware run of *this* approach. The N150 `TT_FATAL @ untilize_codegen_device_operation.cpp:28` reported earlier was against `bdafe915`, the routing-gate version this commit replaced.

Both items on the PR's validation list are now covered: this run, and the [cross-arch sweep re-verification](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320293410) showing no regressions on either arch.

ebanerjeeTT:
### Response to the Copilot review — all three findings are pre-existing; the change is monotone-safe

Copilot raised three things: a cache-hit staleness bug on the factory's live-L1 sample, a test-coverage gap for that same sequence, and (suppressed) a HYBRID-allocator blind spot in the budget. All three describe real properties of the code. None of them is introduced by this PR, and the first two share a single root cause that lives in the dispatch framework rather than in untilize. Filed as #53537 and #53538 respectively; details below.

#### 1. Live-L1-dependent plans frozen on cache hits — confirmed, pre-existing, and this PR is strictly safer than the status quo

The mechanism is exactly as described. `handle_mesh_adapter_cache_hit` goes `apply_descriptor` → `enqueue_mesh_workload` without passing through `compile_and_allocate`, so `ProgramImpl::validate_circular_buffer_region()` never runs a second time (`ttnn/api/ttnn/device_operation.hpp:254-290`, `tt_metal/impl/program/program.cpp:2662`). Nothing invalidates the program cache when an L1 buffer is allocated — only device close and an explicit `clear_program_cache()`. A CB plan chosen at first miss is frozen and never re-checked.

Two corrections to the framing, both of which matter for what the fix should be.

**The symptom is not the clash error.** On a cache hit the validator is not running, so there is nothing to throw. Statically allocated CBs simply overlap the resident L1 buffer and silently corrupt it. That is a worse failure mode than the one this PR fixes, and it is the reason the suggested regression test (below) would not detect it.

**This PR cannot make it worse, and strictly narrows it.** `plan_cb_depths`' three tiers are strictly ordered by size and the live gap is always `<=` the whole-of-L1 budget the old code planned against, so the tier selected here is never larger than the tier `main` would select for the same case. Any cache entry that clashes on a hit after this change would also have clashed before it; the converse is false. The change removes stale-plan cases, it does not add any.

**For the workload that motivated the PR, the stale path is demonstrably not what was breaking.** The original failure was a `TT_THROW` from inside `validate_circular_buffer_region`, which only runs on a cache *miss*. So the Gemma-3-4B miss provably happens with the weights already resident — the shallower tier is what gets cached, and the hit path replays a plan that already fit.

What this PR does newly introduce, and which is worth stating plainly since Copilot only brushes it: one cache key can now map to structurally different programs (three CB tiers plus the native-equivalent program) depending on transient device state. That is a reproducibility property, separate from any clash.

**Why the fix is not here.** The obvious per-op fix is to fold the selected tier into the cache identity — precedent exists one directory over, where native untilize already carries its live-L1 decision as `UntilizeOperationAttributes::enough_space_height` and the default reflection hash picks it up. It works, and it works for a clean reason: with the tier in the key, the program becomes a pure function of the key, and crossing a tier boundary produces a *miss*, which forces the rebuild and the revalidation. Within a tier the cached plan still fits by construction, because tier `i` is chosen exactly when `size(tier_i) <= H < size(tier_{i-1})`.

I am not doing that here, for three reasons.

- It fixes one op. Native untilize has the same exposure through `enough_space_height`, and every op with static CBs has it in the weaker form where the plan is fixed but the frontier still moves underneath it. This is a framework problem in an op-shaped costume.
- The framework fix is much cheaper and strictly better: on a cache hit, compare the cached program's CB region end against the current lowest occupied L1 address. One integer comparison against a value the program already holds, and it converts silent corruption into a loud throw — or a forced rebuild — for every op at once, including the native path.
- The per-op version puts a new allocator query on every dispatch of an op whose smallest configs already show measurable sensitivity to ~150 ns of fixed dispatch delta, and it changes the cache-key contract, which invalidates the 112-config cross-arch sweep just re-run at this head.

Filed as #53537.

#### 2. Test coverage for changing occupancy on a cache hit — real gap, but the suggested test would pass vacuously

Correct that both regimes reserve L1 before the single `ttnn.untilize` call, so only the miss-with-pressure sequence is covered.

The suggested test as described would not catch the bug, though. The memory the oversized CBs overwrite is the *resident* tensor, not the output — the output here lands in DRAM and comes back bit-exact either way. Detecting it requires filling `resident` with known data and asserting it survives the cached run.

Written that way, the test fails on `main`, fails on this branch, and fails for native untilize, because none of them revalidate on a hit. It is a bug-exposing test for #53537, not a validation test for this change, so it belongs with that fix rather than here.

#### 3. HYBRID per-core allocations (suppressed comment) — valid, pre-existing, and not local to this op

Confirmed, and worth being precise about where the gap is. `AllocatorImpl::get_lowest_occupied_l1_address(bank_id)` *does* consult the per-core allocator in HYBRID mode (`tt_metal/impl/allocator/allocator.cpp:425-439`) — but only for the bank it is asked about, and the device-level query only ever asks about bank 0 on the default sub-device manager's allocator:

```cpp
// tt_metal/impl/sub_device/sub_device_manager_tracker.cpp:160
constexpr uint32_t global_bank_id = 0;
const auto& global_allocator = default_sub_device_manager_->allocator(SubDeviceId{0});
auto found_addr = global_allocator->get_lowest_occupied_l1_address(global_bank_id);
```

The validator instead walks every core in the CB's range across every *physical device* allocator (`program.cpp:1895-1911`). So per-core allocations on any core other than bank 0, or living on a per-device rather than the mesh allocator, are invisible to `get_max_l1_space()`. The budget can overestimate and the clash can still occur — Copilot's "can continue throwing" is the right characterisation, this is a residual rather than a regression.

This is not specific to untilize: `get_max_l1_space()` is the shared helper `transpose`, `matmul_utilities` and `reshard_rm_program_factory` all budget against, and they share the blind spot. Fixing it correctly means an allocator-mode-aware and core-range-aware frontier that the validator and the ttnn helpers both call, which is a `tt_metal` change with several callers to re-verify. Filed as #53538, noting there that untilize codegen is maximally exposed since every builder places work over the full compute-with-storage grid.

-Opus 5

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_program_factory.cpp
```diff
@@ -469,6 +575,30 @@ ProgramDescriptor UntilizeCodegenProgramFactory::create_descriptor(
...
+    // uses: (lowest_occupied_compute_l1_address ?: l1_size_per_core) - allocator base, i.e. the
+    // gap between where statically allocated CBs start growing up and where the lowest resident
+    // L1 buffer sits. Taken after create_output_tensors() has already allocated the output, so it
+    // accounts for that too.
+    //
+    // Cost on the hot path: zero. create_descriptor runs only on a program-cache MISS. Every
+    // builder below (and every native factory the fallback can delegate to) emits
+    // emplace_runtime_args with a Buffer* first argument on every core, so buffer bindings are
+    // always populated and resolve; cache HITS therefore take apply_descriptor's
+    // apply_resolved_bindings path, which patches addresses in the cached Program without ever
```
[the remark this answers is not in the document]
ebanerjeeTT: Confirmed, and pre-existing rather than introduced here — the cache-hit path never re-enters `compile_and_allocate`, so `validate_circular_buffer_region` does not run a second time for any op. Two corrections: the symptom on a hit is silent corruption of the resident buffer rather than the clash throw (the validator is not running), and this change is monotone-safe — the tier selected against live headroom is never larger than the whole-of-L1 tier `main` always selects, so it removes stale-plan cases without adding any. Also, the Gemma failure was thrown from inside the validator, which only runs on a miss, so that workload's miss provably already sees the resident weights.

Fixing it in the cache identity would fix one op; the framework-level hit-path guard is one integer comparison and fixes every op including native untilize. Tracked in #53537, with the full reasoning in [this comment](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5329856859).

-Opus 5

tests/ttnn/unit_tests/operations/data_movement/test_untilize.py
```diff
@@ -3009,3 +3009,98 @@ def test_pc_untilize_codegen(device, tensor_shape):
...
+        # Guard against a vacuous pass: if the reservation did not land where intended there is no
+        # L1 pressure left to regress against.
+        actual_headroom = resident.buffer_address() - info.address_at_first_l1_cb_buffer
+        low, high = expected_headroom
+        assert low <= actual_headroom < high, (
+            f"resident L1 buffer left {actual_headroom} B above the CB base; the {headroom_regime} "
+            f"regime needs it in [{low}, {high})"
+        )
+
+        output = ttnn.untilize(input_ttnn_tensor)
```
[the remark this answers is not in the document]
ebanerjeeTT: Right that both regimes only cover the miss-with-pressure sequence. As described the test would pass vacuously, though: the memory the oversized CBs overwrite is `resident`, not the output, and the output is bit-exact either way — you would need to fill `resident` with known data and assert it survives. Written that way it also fails on `main` and for native untilize, so it is a regression test for #53537 rather than coverage for this change. Details in [this comment](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5329856859).

-Opus 5

ronnie00TT:
`enqueue_mesh_workload` is called on a hit — and it does reach `compile_and_allocate`, three hops down: device_operation.hpp:210 → `distributed::EnqueueMeshWorkload` → `mesh_workload.impl().compile()` unconditionally under fast dispatch (distributed.cpp:117) → `compile_program` (mesh_workload.cpp:112) → `compile_and_allocate` (mesh_workload.cpp:102).

ebanerjeeTT:
Seems Opus hallucinated some stuff. doing a deeper dive now

ebanerjeeTT:
Opus made some hallucinations, but I think the PR is still valid.

AFAICT, the specific case that copilot was worried about is that: 

1. Untilize will be invoked on the codegen path 
2. The program will be cached with a large circular buffer for input type X when there's a lot of free space in L1
3. Then after the program runs, buffers are added in L1, eating into the space that was previously marked as free for the program with input type X
4. Then the op is invoked again on input type X using the cached version of the program 

Because this cached program defines circular buffers in space that now is being used for other buffers, it leads to a failure when compile_and_allocate checks whether there's space for the circular buffers

I'm 90% sure this exact risk already exists on main today (and existed even before the original codegen untilize PR landed) because there's no logic to make sure that adding new buffers in L1 doesn't invalidate circular buffer configurations in already cached programs. while that probably should be fixed, I think it can be defined as a problem outside the scope of this PR

Opus hallucinated that compile_and_allocate isn't run during cache hits, which would cause silent memory overwrites in this specific edge case. That's wrong, but the implementation itself should still work.

ronnie00TT:
@ebanerjeeTT That makes sense. But I'm still afraid that the following case would still case error after getting this merged:
1. L1 empty -> cache miss -> pick size1 = 1,179,648  (double_in tier)
2. program cached; CB size frozen. cb_base = 105,664
   => CB region end = 105,664 + 1,179,648 = 1,285,312
3. an L1 tensor is allocated; lowest occupied L1 address drops to 368,640
   => gap above cb_base is only 262,976 < size1
4. same shape called again -> cache HIT (num_program_cache_entries 1 -> 1)
5. validate_circular_buffer_region: 1,285,312 > 368,640
6. TT_THROW @ program.cpp:1919
   "Statically allocated circular buffers in program 2 clash with L1 buffers
    on core range [0-0 - 7-7]. L1 buffer allocated at 368640 and static
    circular buffer region ends at 1285312"

ronnie00TT:
Should we log a warning when the codegen path falls back to native? And for `untilize_force_codegen` it probably needs to be a hard error rather than a log — untilize_force.hpp:35-37 promises that entry never serves native, so an A/B against the native prim can't silently end up measuring native twice.

Note a log/throw inside the factory alone wouldn't cover the forced leg: it shares its cache key with a routed call, so once a routed call has cached the native-equivalent program for that shape, `create_descriptor` never runs again and nothing fires. Adding `forced` to `UntilizeCodegenOperationAttributes` handles both — it also lands in the default hash, so the two legs stop sharing an entry.

ebanerjeeTT:
## Gemma-3-4B still passes after the CB-tier hash — validation item 1 still closed

[Run 32293879602](https://github.com/tenstorrent/tt-metal/actions/runs/32293879602) — `models-e2e-tests / Gemma-3-4B e2e tests [wh_n150]`, [job](https://github.com/tenstorrent/tt-metal/actions/runs/32293879602/job/96202589894) — **success** at `a2cb3b0c`, this PR's current head (live-L1 factory budget at `68f60f8e` plus hashing the codegen CB tier / Native block-split identity). Same workflow (`models-t2-e2e-tests.yaml`), same job name, and same N150 SKU as the nightly that has been failing on `main` since #50178 merged, and as the earlier pass on this PR.

The specific test from the bug report passed:

```
PASSED models/demos/multimodal/gemma3/demo/vision_demo.py::test_multimodal_demo_text[wormhole_b0-performance-device_params0-1-batch1-multi-image-trace-normal-1]
```

The job log contains no `clash with L1 buffers`, and no `TT_THROW` or `TT_FATAL` of any kind — against 3 occurrences of the clash in [the 08-17 nightly](https://github.com/tenstorrent/tt-metal/actions/runs/31990919029/job/95290902099) on `main`.

So the before/after, including the hash-key follow-up, is:

| commit | contains #50178 | live-L1 CB budget | CB-tier / Native split in program hash | Gemma-3-4B e2e [wh_n150] |
|---|---|---|---|---|
| `bef164b30` (nightly, 08-14) | no | no | no | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/31767822219/job/94670230682) |
| `e16f6cbfb` (nightly, 08-17) | yes | no | no | [failed](https://github.com/tenstorrent/tt-metal/actions/runs/31990919029/job/95290902099) |
| `68f60f8e` (this PR, factory-only) | yes | yes | no | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/32070288203/job/95513658884) |
| `a2cb3b0c` (this PR, current head) | yes | yes | yes | [passed](https://github.com/tenstorrent/tt-metal/actions/runs/32293879602/job/96202589894) |

The hash change does not reintroduce the Gemma clash. Cross-arch sweep re-verification at `68f60f8e` is [here](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320293410); that still needs a re-run at `a2cb3b0c`.

ebanerjeeTT:
## Cross-arch re-verification of the untilize codegen sweep at `a2cb3b0c` — no regressions

Re-ran the full 112-config untilize codegen sweep from #50178 on both archs at this PR's current
head `a2cb3b0c` (the CB-tier hash follow-up on top of the `68f60f8e` live-L1 budget fix, [re-verified
separately](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5320293410)), through
`tt-dm-codegen`'s cross-arch framework.

Same mechanism as before: the framework refuses this PR directly (`cross_arch.guard_ref` rejects any
head that isn't `port/*`, and `dispatch` reads the op/manifest from a `codegen-port-results:v2`
block this body doesn't carry), so `a2cb3b0c` was pushed under a `port/*` name and dispatched from
#53697 (scaffold only, no code differs, results below). Runs:
[wormhole_b0/N300](https://github.com/tenstorrent/tt-metal/actions/runs/32297304723) ·
[blackhole/P150](https://github.com/tenstorrent/tt-metal/actions/runs/32297316426), both at
`a2cb3b0c`, 112 cases each.

### Correctness

| | board | bit-exact vs native | in-tree unit tests | routing matrix |
|---|---|---|---|---|
| **wormhole_b0** | N300 | 112 / 112 | `unit_tests_exit=0` (922 passed, 4 skipped, 2 xfailed) | passed |
| **blackhole** | P150 | 112 / 112 | `unit_tests_exit=0` (924 passed, 4 skipped) | passed |

The unit-test leg includes this PR's regression test at both headroom regimes and both the cold and
`warmup_then_constrain` parametrizations, all passing on both archs:
`test_untilize_codegen_with_resident_l1_buffers[headroom_regime=shallower_cb_plan-cold]`,
`[headroom_regime=shallower_cb_plan-warmup_then_constrain]`,
`[headroom_regime=no_cb_plan_fits-cold]`, `[headroom_regime=no_cb_plan_fits-warmup_then_constrain]`.
`test_untilize_codegen_routing.py` — including `test_forced_codegen_refuses_out_of_scope_case` and
both `test_untilize_execution_controls_route_to_native` variants — passed in full on both archs.

### Performance — device time, ported / native (lower is faster)

| | board | median | best | worst | p25 / p75 | at or above parity |
|---|---|---|---|---|---|---|
| **wormhole_b0** | N300 | **0.284x** | 0.226x | 0.861x | 0.256 / 0.347 | 0 of 112 |
| **blackhole** | P150 | **0.261x** | 0.223x | 0.800x | 0.243 / 0.312 | 0 of 112 |

Against the `68f60f8e` re-verify (wormhole_b0 median 0.285x, best 0.226x, worst 0.866x, p25/p75
0.256/0.345; blackhole median 0.260x, best 0.223x, worst 0.799x, p25/p75 0.243/0.312) and against
the #50178 post-review baseline (wormhole_b0 median 0.286x, best 0.225x, worst 0.868x, p25/p75
0.252/0.346; blackhole median 0.261x, best 0.223x, worst 0.800x, p25/p75 0.244/0.312): every
statistic is within 0.002 of both prior runs, on both archs. The hash-key change touches program
cache identity, not the budget arithmetic or the plan the factory selects, and that's exactly what
these numbers show — no measurable device-time effect from `68f60f8e` to `a2cb3b0c`.

Per-config, joining all 112 configs against the `68f60f8e` re-verify's per-case numbers:

- median delta essentially zero on both archs (wormhole_b0 −0.0001, blackhole 0.0000);
- range −0.017 to +0.008 on wormhole_b0 and −0.003 to +0.003 on blackhole — tighter than the noise
  band #53420 measured against #50178 (−0.057 to +0.057), consistent with comparing two CI runs of
  the *same* budget arithmetic rather than a code change;
- no config's ratio moved enough on either arch to be distinguishable from run-to-run noise.

The 56 configs whose output lives in L1 — the half `get_max_l1_space()` actually touches — again show
no separate signature: wormhole_b0 median 0.308x, blackhole 0.273x, matching `68f60f8e`'s 0.308x /
0.273x exactly to three decimals.

Wall-clock flags were the usual CI noise, similar order to both prior runs: wormhole_b0 flagged 1
case (5 more cleared on the in-job retest), blackhole 1 (4 cleared) — against 1/4-cleared and
3/4-cleared respectively at `68f60f8e`, and against 1/4-cleared (wormhole_b0 only) at `3c131242`.
Every flagged case still beats native on device.

### What this run does and does not cover

No config re-routed on either arch (`demoted_count: 0` both), and nothing in the sweep landed at or
above device parity — the signature a fall-through to the native-equivalent program would leave. On
a near-idle device the CB-tier hash is inert: it changes which cache key a program resolves under,
not whether a plan fits, so this run cannot exercise the `std::nullopt` → `build_native_equivalent`
path any more than the `68f60f8e` re-verify could. That path stays covered by the `no_cb_plan_fits`
regime of the unit test above, which passed on both archs at this commit too.

Item 1 of the validation list — the Gemma-3-4B trace-mode test on N150 — was
[re-run separately at this same commit](https://github.com/tenstorrent/tt-metal/pull/53399#issuecomment-5347310258)
and still passes; not duplicated here.

Scaffold: #53697 (closing now that both runs are collected).

vtsilytskyiTT:
Apparently, CB budget calculation is not trivial. Did you consider making nice common functions for it? I think other OPs would be happy to reuse the solid verified API to get CB budget.

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_cb_plan.cpp
```diff
@@ -0,0 +1,262 @@
...
+            best = d;
+        }
+    }
+    return best;
+}
+
+std::optional<CbPlan> plan_cb_depths(
+    uint64_t usable_l1, uint32_t pages_per_unit, uint32_t page_size, uint32_t block_units) {
+    uint64_t p = pages_per_unit;
+    uint64_t ts = page_size;
```
amorrisonTT: Nit: this variable copying is not really necessary and some const qualifiers would help with readability

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_cb_plan.cpp
```diff
@@ -0,0 +1,262 @@
...
+    bool tile_aligned;
+};
+
+PaddedGrid padded_grid(const Tensor& input) {
+    const auto& padded_shape = input.padded_shape();
+    uint32_t rank = padded_shape.rank();
+    uint32_t w = padded_shape[-1];
+    uint32_t h = padded_shape[-2];
+    uint32_t nc = 1;
+    for (uint32_t i = 0; i + 2 < rank; ++i) {
```
amorrisonTT: Nit: nice place for std::accumulate

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_device_operation.cpp
```diff
@@ -65,6 +68,30 @@ UntilizeCodegenDeviceOperation::tensor_return_value_t UntilizeCodegenDeviceOpera
...
 
+ttsl::hash::hash_t UntilizeCodegenDeviceOperation::compute_program_hash(
+    const operation_attributes_t& operation_attributes, const tensor_args_t& tensor_args) {
+    const auto chosen = untilize_codegen_detail::choose_codegen_cb_plan(operation_attributes, tensor_args);
+    const auto native =
+        untilize_codegen_detail::native_cache_identity(operation_attributes, tensor_args, chosen.tier);
+    return ttsl::hash::hash_objects_with_default_seed(
+        ttsl::hash::type_hash<UntilizeCodegenDeviceOperation>,
+        operation_attributes,
+        tensor_args,
```
amorrisonTT, line 79: Wait, should we be hashing on tensor args?

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_program_factory.cpp
```diff
@@ -446,6 +408,82 @@ ProgramDescriptor build_with_unpadding(
...
+// Delegating (rather than duplicating native's kernels) keeps the two in lockstep by
+// construction. It is sound because the output tensor the codegen op already allocated is
+// byte-identical in spec to the one native would allocate for the same case:
+//   - tile-aligned  -> UntilizeDeviceOperation::compute_output_specs (same logical shape, same
+//     ROW_MAJOR page config, same dtype demotion bf8_b->bf16, same padded shape carried through)
+//   - non-aligned   -> UntilizeWithUnpaddingDeviceOperation::compute_output_specs with
+//     output_tensor_end = logical_shape - 1, i.e. the compact interleaved spec, which is exactly
+//     what UntilizeCodegenDeviceOperation::compute_output_specs's non-aligned branch declares.
+// The native factories only read `output` (buffer address, spec) to emit their descriptor, so
+// handing them the already-allocated tensor is exactly what their own op would have done.
```
amorrisonTT: This comment is a bit superfluous

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_device_operation.cpp
ebanerjeeTT: AFAICT, we always hash on tensor args. This custom hash function entirely replaces the default one for the codegen path. Omitting tensor_args would drop shape/dtype/layout from the key, and Including it is just copying the default. The only new addition to the key is what appears after the tensor_args line, which are:

chosen.tier,
native.split_valid,
native.ncores,
native.nblocks_per_core,
native.single_block_size,
native.single_block_size_cliff_row,
native.single_block_size_cliff_col,
native.has_cliff_row,
native.has_cliff_col,
native.full_cores_per_row,
native.full_cores_per_col,
native.single_sub_block_size

excluding native.enough_space_height since that also exists on the normal untilize key

ebanerjeeTT:
> Apparently, CB budget calculation is not trivial. Did you consider making nice common functions for it? I think other OPs would be happy to reuse the solid verified API to get CB budget

I'm fairly certain the CB budget api already exists as `get_max_l1_space` in data_movement/common. It's what this PR uses to fix the gemma test failure. The only other new functionality for CB planning added in this PR is `plan_cb_depths`, which is only relevant for codegen untilize's reader and writer kernels.

ttnn/cpp/ttnn/operations/data_movement/untilize/codegen/untilize_codegen_device_operation.cpp
amorrisonTT: Yes, you're right this is handled properly by the reflection

amorrisonTT (approved):

vtsilytskyiTT:
> > Apparently, CB budget calculation is not trivial. Did you consider making nice common functions for it? I think other OPs would be happy to reuse the solid verified API to get CB budget
> 
> I'm fairly certain the CB budget api already exists as `get_max_l1_space` in data_movement/common. It's what this PR uses to fix the gemma test failure. The only other new functionality for CB planning added in this PR is `plan_cb_depths`, which is only relevant for codegen untilize's reader and writer kernels.

Agree. Interesting that there are 2 implementations: in data_movement and in matmul_utilities.

vtsilytskyiTT (approved):

bbradelTT (approved):

PR #53404, opened 2026-08-17
Set KERNEL_COMPILE_TIME_ARG_MAP via a generated header instead of -D define

esmalTT:
## Problem

Every named compile-time arg a kernel registers is packed into one long
`-DKERNEL_COMPILE_TIME_ARG_MAP={...}` flag. Linux limits any single command-line argument
to 128 KB, so once a kernel's arg list is big enough, it simply cannot be compiled. The
only symptom points at the compiler:

```
posix_spawnp failed for .../riscv-tt-elf-g++: Argument list too long (jit_build_utils.cpp:156)
→ ncrisc build failed
→ Failed to generate binaries for gemma4_sliding_window_decoder_layer
```

Nothing there mentions compile-time args or a size limit. That one flag is essentially the
whole command line — the next longest argument on a failing compile is under 200 bytes.

This is not specific to one model. Large fused ops emit hundreds to thousands of named
args, and each arg name repeats the op's full naming scope, so names average ~75 bytes and
the ceiling lands around 1,750 args. Several tt-blaze models are already close:

| kernel | named args | % of limit |
|---|---|---|
| `gemma4_global_decoder_layer` | 1,293 | ~74% |
| `gemma4_sliding_window_decoder_layer` | 1,256 | ~72% |
| `dflash_decoder` | 1,098 | ~63% |
| `minimax_m2_global_layer` | 1,089 | ~62% |

The reported failure measured 131,755 bytes against the 131,072 limit — over by 0.5%.

Note that the 128 KB limit applies to each argument individually, so switching the JIT
build to `posix_spawn` did not help here: it removed the cap on total command length, but
this single flag exceeds the per-argument cap on its own.

## Fix

Write the arg list into a small generated header per kernel and have the compiler include
it automatically, instead of passing it as a flag. File contents have no size limit, so the
command line stays short no matter how many args a kernel has.

Kernels see no difference: `KERNEL_COMPILE_TIME_ARG_MAP` is defined exactly as before, and
`get_named_compile_time_arg_val()` keeps working everywhere it is used.

Three details worth calling out:

- The header is written alongside the other generated per-kernel headers rather than at
  build time. That directory is already uploaded to the remote compile server and already
  folded into preprocessed output, so local builds, remote builds, and preprocessed builds
  all pick it up with no new plumbing.
- The include uses a plain filename, not a full path. The remote compile server lays its
  directories out differently from the local build, and both already search the right
  parent directory, so one setting works on both. A local absolute path would build fine
  here and then fail on the server.
- Args are sorted by name. They come from an unordered map, so their order otherwise varies
  between runs; sorting keeps the generated file identical run to run, which the build
  cache depends on.

Caching is unaffected: the named args are already part of the hash that picks a kernel's
output directory, so a changed arg list lands in a different directory, and the new header
is tracked as a dependency of every object built from it.

The emulator path had the same problem, and worse — it runs its compile through a shell, so
it is limited by *total* command length. It now writes the arg list into the generated
wrapper source, which is already where that path puts kernel defines to keep long or
awkward values off the command line. Both paths share one formatter, so they produce
identical text.

Fixes tenstorrent/tt-blaze#2017
Fixes tenstorrent/tt-blaze#3209

### DIFF
--- a/tests/tt_metal/tt_metal/api/test_compile_time_args.cpp
+++ b/tests/tt_metal/tt_metal/api/test_compile_time_args.cpp
@@ -3,6 +3,8 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <gtest/gtest.h>
+#include <unistd.h>
+#include <cstddef>
 #include <cstdint>
 #include <tt-metalium/host_api.hpp>
 #include <tt-metalium/tt_metal.hpp>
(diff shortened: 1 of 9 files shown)


esmalTT:
## Local validation

Run on Blackhole galaxy hardware (`bh-glx-120-c01u08`), RelWithDebInfo.

### Host unit tests — `unit_tests_jit_build --gtest_filter=NamedCtArgMap.*`

```
[==========] 11 tests from 1 test suite ran. (80 ms total)
[  PASSED  ] 11 tests.
```

### Device tests — `unit_tests_api --gtest_filter=*NamedCompileTimeArgs*`

```
[       OK ] NamedArgsTest.TensixTestNamedCompileTimeArgs (29019 ms)
[       OK ] NamedArgsTest.TensixTestNamedCompileTimeArgsComputeKernel (7471 ms)
[       OK ] CompileTimeArgsTest.TensixTestNamedCompileTimeArgs (2294 ms)
[       OK ] CompileTimeArgsTest.TensixTestNamedCompileTimeArgsExceedingArgvLimit (3288 ms)
[  PASSED  ] 4 tests.
```

The header is really being emitted per kernel, and small maps are unaffected:

```
/tmp/.../kernels/blaze_named_runtime_args_kernel/281751123106870504/named_ct_arg_map_generated.h   137 bytes
/tmp/.../kernels/blaze_named_compile_time_args_compute_kernel/.../named_ct_arg_map_generated.h     137 bytes
```

### Negative control — the new device test does reproduce the bug

To confirm the regression test has teeth, I restored the old `-D` delivery in
`export_target_recipe` (nothing else) and re-ran on a **cold** JIT cache
(`TT_METAL_CACHE` pointed at an empty dir):

```
TEST EXIT=1
error | BuildKernels | posix_spawnp failed for '.../riscv-tt-elf-g++':
        Argument list too long (jit_build_utils.cpp:156)
error | BuildKernels | brisc compile failure -- cmd: .../riscv-tt-elf-g++ -O2 -std=c++17 ...
```

That is the exact failure signature from the two linked issues. Restoring the fix and
re-running under identical cold-cache conditions passes. Matched pair, one variable.

Worth noting for anyone re-checking this by hand: the kernel hash covers the named args
but not the delivery mechanism, so a warm cache will serve the ELF from a previous run
and the test passes either way. The cold cache is what makes the control meaningful.

### `MAX_ARG_STRLEN` boundary

Measured directly with `posix_spawn` on this kernel, and pinned as a test:

| argv element size | result |
|---|---|
| 131,071 bytes | exec OK |
| 131,072 bytes | `E2BIG` |
| 135,168 bytes | `E2BIG` |

Confirms the limit is per-argv-element and that moving to `posix_spawn` did not raise it.

### CI

Dispatched on this branch (draft PRs skip `pr-gate` automatically, so these were
triggered manually):

- **[post-commit] all - Static checks, linters etc.** — ✅ success
- **PR Gate** — running
- **(Runtime) Unit Tests** (`bh_p150b_civ2`, `wh_n150_civ2`) — running; this is the leg
  that runs both `unit_tests_api` and `unit_tests_jit_build`

esmalTT:
## CI results

| Workflow | Result |
|---|---|
| Static checks, linters etc. | ✅ success |
| PR Gate | ✅ success |
| (Runtime) Unit Tests | ❌ failure — pre-existing, see below |

### The Runtime Unit Tests failure is not from this change

Two tests failed, both on `wh_n150_civ2`, neither related to compile-time args:

- `LLKMeshDeviceFixtureSlowDispatchOnly.TensixBinaryComputeColBroadcastThenDestReuseSrcB`
- `MeshTensorDeviceTest.UniformCopyToHost_ReusesPinnedMemoryCacheEntries`

The same two tests fail in the same two jobs on `main` — see nightly run
[31993970889](https://github.com/tenstorrent/tt-metal/actions/runs/31993970889), which
predates this branch. This workflow has been red on `main` for at least its last five
nightly runs.

### The tests that matter here passed

The new oversized-map device test passed on **both** SKUs, on cold CI caches:

```
[ RUN      ] CompileTimeArgsTest.TensixTestNamedCompileTimeArgsExceedingArgvLimit
[       OK ] CompileTimeArgsTest.TensixTestNamedCompileTimeArgsExceedingArgvLimit (2648 ms)
[       OK ] CompileTimeArgsTest.TensixTestNamedCompileTimeArgs (803 ms)
```

Job outcomes for the two jobs covering the changed code:

| Job | wh_n150_civ2 | bh_p150b_civ2 |
|---|---|---|
| `runtime_core` (runs `unit_tests_jit_build` — the 11 host tests) | ✅ | ✅ |
| `runtime_api` (runs `unit_tests_api` — the device test) | ❌ pre-existing | ✅ |

So the new tests pass on Wormhole and Blackhole, and the only red job is red on `main` for
unrelated reasons.

cfjchu (approved):
ship it - thanks @esmalTT

tt-aho (approved):

arminaleTT (approved):

ruizhangTT (approved):

PR #53428, opened 2026-08-17
[Bug fix] Dump the DSA indexer key cache for src-KV validation in GLM5.2

yeongyunTT:
### PR Category
Bug fix

### Summary
`_dump_src_kv` wrote only the KVPE cache, so the decode side had no reference for the indexer cache and prefill src == decode dst covered config 0 only. This adds a config-1 pass so the indexer cache can be validated end to end.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:yeongyun/migration-driver-dump-multi-kv)


nsorabaTT (approved):
This will work for GLM but may need some generalization in a future PR to make it work for Kimi K3, Qwen etc.

Each layer can have a subset of N Caches that are valid. The migration driver should support selectively picking M out of N caches for each layer to be migrated.

philei-tt:
**The `migrated_layers` drop rule disables the indexer dump in exactly the GLM-5.2 reduced P→D flow this PR is for**

`_cache_plan` drops any non-identity cache whenever `migrated_layers` is set. But the dump call site passes `driver.layers` = `PREFILL_MIGRATION_LAYERS`, and every reduced P→D deployment must set it (a full-range `[0, num_layers)` migrate cannot address a 5-row decode table). So in the GLM-5.2 `{0, 4}` run: subset set → config 1 non-identity → dropped → no `ref_index_k_list` in `src_slot<N>.pt` → the decode side's indexer src-KV check still SKIPs. As written, the new dump only fires for full-model / loopback runs.

The blanket drop is also more conservative than the situation requires, and differently so on the two paths:

1. **The dump is safe unconditionally.** It reads the *source* cache on the cache's own axis (the `rows` mapping) — which rows the migrate addressed doesn't change what the source holds. Dumping config-1 layer 0 is precisely the reference the decode check needs. And if decode later compares a row the migrate delivered wrongly (the rank≠layer-id sender behavior your own comment describes), a FAIL is the *correct* verdict — withholding the reference converts a detectable corruption into a SKIP. Suggest removing the `migrated_layers` exclusion from the dump path entirely.

2. **The verify needs a per-layer condition, not per-cache.** A migrated layer `l` is cleanly addressable in a compacted cache iff `rank(l) == l` — true for full-indexer layers 0, 1, 2, exactly the generators the reduced topologies use. Filtering to `{l: r for l, r in rows.items() if r == l}` under a subset migration keeps the coverage the per-cache identity test throws away, with the same safety for rank≠id rows.

Separate observation, not blocking: `_cache_plan` now constructs the exact layer↔rank mapping (`_full_indexer_layer_indices` → rows) whose absence causes the underlying problem — the comment here even names it ("the engine replays the number per cache… so a subset addresses ROW L of every cache"). Applied one file over, in `kv_chunk_table.py`'s table authoring — key config-1 rows by *global layer id* (extent = num_layers, consumer rows left empty, row L pointing at the compacted physical slot `rank(L)`) — it would remove the mismatch at the source: both endpoints would then agree on what "row r" means for every config, full-model indexer migration becomes correct with no `MigrateSpec` change, and this PR's drop/filter logic reduces to the identity case. Worth a tracked issue if it's out of scope here.

Minor: the config-1 `head_dim` fallback dispatches on `why.startswith("DSA index cache")` — behavior keyed to a log-message string that a future reword would silently change; a structured flag on the plan entry would be safer.

jjovicicTT (approved):

PR #53430, opened 2026-08-17
[Feature] Add indexed fused paged KV-cache update

### LINKED ISSUE #53429: [TTNN] Add indexed fused KV-cache update
## Problem

Speculative decoding can produce several packed K/V rows that must be committed to arbitrary physical cache rows. TTNN currently does not provide a direct way to update both K and V caches together for this access pattern, requiring a custom operation or a less efficient decomposition.

## Proposed behavior

Add an experimental TTNN operation that:

- Updates K and V caches in place from packed input rows.
- Uses an index tensor containing the physical destination row for each input row.
- Skips negative indices, allowing variable-length commits.

The exact API, layouts, dtypes, validation, and hardware support are described in the accompanying PR.

## Motivation

This is a general primitive for speculative decoding. DFlash in MuseGlimmer is one demonstrated consumer.

In controlled P150 testing, the native TTNN implementation preserved the performance and output behavior of Cody's original custom kernel.

## Reference

Original [source kernel](https://github.com/Codys12/MuseGlimmer/blob/1d1430465ba98deb70247af1e2c9353ecbcda87e/tt/attention/kernels/packed_kv_update.cpp) authored by Cody Steinmetz (`@Codys12`) in [MuseGlimmer at commit `1d143046`](https://github.com/Codys12/MuseGlimmer/commit/1d1430465ba98deb70247af1e2c9353ecbcda87e). That source file carries an Apache-2.0 SPDX identifier and Tenstorrent copyright. Native TTNN port, generalization, testing, and benchmarking by Samuel Jett (`@sjettTT`).

sjettTT:
### PR Category

Feature

### Summary

Closes #53429

Adds `ttnn.experimental.indexed_fused_update_cache`, an in-place operation that writes packed K/V input rows to arbitrary physical rows in two paged caches. Negative physical indices are no-ops, allowing a speculative decoder to commit a variable number of accepted rows without rebuilding or copying its packed inputs.

The operation is based on the packed KV-cache [source kernel](https://github.com/Codys12/MuseGlimmer/blob/1d1430465ba98deb70247af1e2c9353ecbcda87e/tt/attention/kernels/packed_kv_update.cpp) originally authored by Cody Steinmetz (`@Codys12`) in [MuseGlimmer at commit `1d143046`](https://github.com/Codys12/MuseGlimmer/commit/1d1430465ba98deb70247af1e2c9353ecbcda87e). It was ported to native TTNN, generalized, tested, and benchmarked by Samuel Jett (`@sjettTT`).

## Design decision

This operation uses the TTNN `DeviceOperation` and Metal 2 `ProgramSpec` path. It does not use `GenericOp`, `ProgramDescriptor`, or the legacy `CachedProgram::create/override_runtime_arguments` program-construction path. Its dataflow kernel uses named tensor/DFB bindings and the Metal 2 object-based `DataflowBuffer` and `Noc` APIs.

`ProgramSpec` was selected over `ProgramDescriptor` because this is a new, single-program operation with no global semaphores, per-device program variants, or call-varying scalar arguments. The per-core work assignment is derived deterministically from hashed tensor specs; the only bindings that change between same-spec calls are the five tensors, which the base Spec path handles directly. This follows the current TTNN guidance that Specs are the forward path for new operations whose shape fits the supported Spec model.

The operation is initially exposed separately from `paged_fused_update_cache` because its indexing semantics are different: each packed input row names an arbitrary physical destination row directly, and negative indices represent skipped commits. It does not perform page-table translation.

Keeping these semantics separate avoids complicating the existing paged-cache API. If the paged-cache owners prefer an indexed mode on the existing fused operation, the implementation can be consolidated during review without changing the underlying kernel behavior.

## API and initial support

```python
ttnn.experimental.indexed_fused_update_cache(
    cache_tensor1,
    input_tensor1,
    cache_tensor2,
    input_tensor2,
    physical_update_idxs_tensor,
) -> tuple[ttnn.Tensor, ttnn.Tensor]
```

- Cache tensors: `[pages, heads, rows_per_page, head_dim]`
- Packed inputs: `[1, heads, packed_rows, head_dim]`
- Physical indices: row-major INT32 tensor with at least `packed_rows` entries
- Initial data/layout support: BF16 tiled interleaved K/V tensors
- Negative and out-of-range physical indices are no-ops
- Updates both caches in place and returns the aliased cache tensors
- Physical-row addressing only; no page-table translation
- Replicated tensors and identical tensor-parallel cache/input topologies sharded over heads, head dimension, or both are supported; cache page/row sharding and sharded physical indices are rejected
- Initial hardware support: Wormhole and Blackhole

## Validation

- P150 Watcher unit coverage: 19 single-device cases, 6 two-device mesh cases, and one physical 2x2 mesh case passed, including exact scatter, in-place aliasing, validation errors, runtime-argument repatching, program-cache reuse, trace replay with changed inputs and indices, replicated tensors, head/head-dimension tensor-parallel sharding, and rejection of sharded physical indices, cache page/row sharding, and mismatched data topologies.
- P150 sweep coverage: all 5 vectors passed, including page-boundary crossing, skipped/out-of-range rows, duplicate destinations, and more source rows than worker cores.
- Wormhole Watcher coverage: all 14 single-device cases passed on the latest branch, including exact scatter, program-cache reuse, trace replay, validation, and the larger worker-stride case.
- MuseGlimmer B=1/2/4/8 matched Cody's original kernel exactly for token hashes, decoded outputs, speculative-acceptance histories, and accepted/verification counts. Performance was unchanged within run-to-run noise.

### Notes for reviewers

Review questions:

- Should this remain a separate experimental operation or become an indexed mode of `paged_fused_update_cache`?
- Is `indexed_fused_update_cache` the preferred op name?
- Should physical-row semantics be exposed directly at the TTNN API boundary?
- Is BF16-only K/V support OK for now?

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:sjett/indexed-fused-update-cache)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:sjett/indexed-fused-update-cache)


yieldthought:
Codex review:

1. **Reject custom tile descriptors before dispatch.** [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp:31](https://github.com/tenstorrent/tt-metal/blob/99b6c63343b02e1a0387d1e2c7adda5e29d785ed/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp#L31) checks only `Layout::TILE`, but the factory derives tile counts with fixed 32-row/32-column constants at [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_program_factory.cpp:30](https://github.com/tenstorrent/tt-metal/blob/99b6c63343b02e1a0387d1e2c7adda5e29d785ed/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_program_factory.cpp#L30) and the kernel hard-codes 32x32, non-transposed face geometry at [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp:21](https://github.com/tenstorrent/tt-metal/blob/99b6c63343b02e1a0387d1e2c7adda5e29d785ed/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp#L21). Legal custom tiles therefore pass validation and silently produce wrong results: for example, an input using `ttnn.Tile((16, 32))` with fewer than 16 source rows has padded height 16, so `source_height_tiles = padded_height / 32` becomes zero and the operation returns unchanged caches. Narrow, mismatched, or transposed cache/input tiles can likewise use the wrong page sizes or byte offsets. Suggestion: require all four cache/input tensor specs to use the default non-transposed 32x32 tile and add rejection coverage, or derive the complete geometry from their tile descriptors.

2. **Bound the positions page staged into L1.** [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp:110](https://github.com/tenstorrent/tt-metal/blob/99b6c63343b02e1a0387d1e2c7adda5e29d785ed/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp#L110) only lower-bounds the positions width, and the one-page check at line 113 does not cap it: any row-major `[1, N]` tensor is one device page. The factory then allocates and copies the entire aligned page into a per-core DFB at [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_program_factory.cpp:109](https://github.com/tenstorrent/tt-metal/blob/99b6c63343b02e1a0387d1e2c7adda5e29d785ed/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_program_factory.cpp#L109) even though the kernel consumes at most 256 entries. A large otherwise-valid positions buffer can therefore fail with an internal static-CB/L1 allocation error based entirely on unused trailing entries. Suggestion: cap the accepted width/page bytes at the supported maximum (for example, 256 INT32 entries), or stage only the bytes required for `source_rows`.

sjettTT:
implemented feedback from @amorrisonTT

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,110 @@
...
+    constexpr auto source_height_tiles = get_arg(args::source_height_tiles);
+    constexpr auto cache_page_rows = get_arg(args::cache_page_rows);
+    constexpr auto total_cache_rows = get_arg(args::total_cache_rows);
+    constexpr auto worker_count = get_arg(args::worker_count);
+
+    constexpr uint32_t tile_width = 32;
+    constexpr uint32_t tile_height = 32;
+    constexpr uint32_t face_width = 16;
+    constexpr uint32_t face_height = 16;
+    constexpr uint32_t bytes_per_element = 2;
```
amorrisonTT: Make sure the dtype implications are enforced in `validate_on_program_cache_miss` and documented in the nanobind

sjettTT: Confirmed this, thanks

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,110 @@
...
+    noc.async_read_barrier();
+    positions_dfb.push_back(1);
+    CoreLocalMem<volatile int32_t> physical_positions(positions_dfb.get_read_ptr());
+
+    auto copy_rows = [&](const auto& source, const auto& cache, uint32_t head, uint32_t width_tile) {
+        for (uint32_t source_height_tile = 0; source_height_tile < source_height_tiles; ++source_height_tile) {
+            const uint32_t source_tile = (head * source_height_tiles + source_height_tile) * width_tiles + width_tile;
+
+            scratch_dfb.reserve_back(1);
+            noc.async_read(
```
amorrisonTT: You might be able to push your perf here by pipelining reads using transaction IDs. Up to you whether that is necessary, though

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,110 @@
...
+                const uint32_t source_face_y = tile_row / face_height;
+                const uint32_t source_line = tile_row % face_height;
+                const uint32_t dest_tile_row = row_in_page % tile_height;
+                const uint32_t dest_face_y = dest_tile_row / face_height;
+                const uint32_t dest_line = dest_tile_row % face_height;
+                for (uint32_t face_x = 0; face_x < tile_width / face_width; ++face_x) {
+                    const uint32_t source_offset =
+                        (source_face_y * 2 + face_x) * face_bytes + source_line * face_line_bytes;
+                    const uint32_t dest_offset = (dest_face_y * 2 + face_x) * face_bytes + dest_line * face_line_bytes;
+                    noc.async_write(
```
amorrisonTT: Same thought about pipelining the writes

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_program_factory.cpp
```diff
@@ -0,0 +1,143 @@
...
+             TensorParameter{.unique_id = POSITIONS, .spec = positions.tensor_spec()}},
+        .work_units = {WorkUnitSpec{.name = "main", .kernels = {UPDATE}, .target_nodes = all_cores}},
+    };
+
+    KernelRunArgs update_run_args{.kernel = UPDATE};
+    for (uint32_t worker_start = 0; worker_start < cores.size(); ++worker_start) {
+        AddRuntimeArgsForNode(
+            update_run_args.runtime_arg_values,
+            cores[worker_start],
+            {{"source_rows", source_rows}, {"worker_start", worker_start}, {"worker_stride", num_cores}});
```
amorrisonTT, line 133: Is there any chance of issues if `num_cores >> num_pages` or something like that?

amorrisonTT:
Also, un-draft this if is ready

bbradelTT (approved):
Please comment on or address all AI comments.

I ran this through AI, which had two suggestions that seem reasonable:
- Alias rejection must cover all five bound tensors, including input1 == input2. Otherwise a cached program compiled from aliased inputs can incorrectly bind both inputs to input1 on later distinct-input calls.
- Also cap flattened cache rows at INT32_MAX, not UINT32_MAX, since larger physical indices cannot be represented by the required INT32 index tensor.

sjettTT:
@bbradelTT Addressed both suggestions in `488e09d5b8c`:

- validation now rejects aliasing across every pair of the five bound tensors on both cache-miss and cache-hit validation; regression coverage exercises all 10 pairs, including `input_tensor1 == input_tensor2`, and verifies an aliased first call inserts no program-cache entry;
- flattened cache rows are bounded by the INT32-addressable index range. Since row indices are zero-based, the exact maximum row count is `INT32_MAX + 1`—the last valid row index is `INT32_MAX`.

The Release build passed, and all 36 indexed-cache tests passed on Blackhole, including replicated, head-sharded, head-dimension-sharded, and 2x2 mesh coverage.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+    constexpr auto cache_page_rows = get_arg(args::cache_page_rows);
+    constexpr auto total_cache_rows = get_arg(args::total_cache_rows);
+    constexpr auto worker_count = get_arg(args::worker_count);
+
+    constexpr uint32_t tile_width = 32;
+    constexpr uint32_t tile_height = 32;
+    constexpr uint32_t face_width = 16;
+    constexpr uint32_t face_height = 16;
+    constexpr uint32_t bytes_per_element = 2;
+    constexpr uint32_t face_bytes = face_width * face_height * bytes_per_element;
```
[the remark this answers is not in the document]
sjettTT: This experimental operation intentionally supports BF16. `validate_cache_input_pair` fail-closes all four cache/input tensors to BF16, and the nanobind contract documents the same restriction. Any future dtype expansion must update this kernel geometry together with validation and the API contract.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+        for (uint32_t source_height_tile = 0; source_height_tile < source_height_tiles; ++source_height_tile) {
+            const uint32_t current_slot = source_height_tile % scratch_buffer_depth;
+            const bool has_next_tile = source_height_tile + 1 < source_height_tiles;
+            const uint32_t next_slot = (current_slot + 1) % scratch_buffer_depth;
+            if (has_next_tile) {
+                issue_source_read(source_height_tile + 1, next_slot);
+            }
+
+            bool wrote_current_slot = false;
+            for (uint32_t tile_row = 0; tile_row < tile_height; ++tile_row) {
```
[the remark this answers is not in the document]
sjettTT: Already covered by `test_indexed_fused_update_cache_multi_tile_rows_and_worker_stride`: it uses 96 source rows across three source tiles, maps rows 31, 32, and 64 to the same destination, and compares both caches exactly against the last-source-row-wins reference. This crosses both tile boundaries and the two-slot scratch-buffer wrap.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp
```diff
@@ -0,0 +1,219 @@
...
+    validate_device_tensor(args.input_tensor2, "input_tensor2");
+    validate_device_tensor(args.physical_update_idxs_tensor, "physical_update_idxs_tensor");
+
+    TT_FATAL(
+        args.cache_tensor1.device() == args.input_tensor1.device() &&
+            args.cache_tensor1.device() == args.cache_tensor2.device() &&
+            args.cache_tensor1.device() == args.input_tensor2.device() &&
+            args.cache_tensor1.device() == args.physical_update_idxs_tensor.device(),
+        "all indexed_fused_update_cache tensors must be on the same device");
+    validate_mesh_topologies(args);
```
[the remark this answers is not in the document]
sjettTT: Addressed in `3e0ce7ceb0c`: validation now enforces the documented Wormhole B0 and Blackhole architecture set before program construction.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/indexed_page_cache_nanobind.cpp
```diff
@@ -0,0 +1,60 @@
...
+        mod,
+        R"doc(
+            Writes packed rows from two input tensors into two cache tensors in parallel.
+
+            ``physical_update_idxs_tensor`` is a row-major INT32 tensor with shape
+            ``[1, num_rows]``. Entry ``i`` gives the physical cache row for input
+            row ``i``; negative and out-of-range entries are skipped. Physical rows
+            flatten cache dimensions 0 and 2: ``page = index // cache.shape[2]`` and
+            ``row_in_page = index % cache.shape[2]``. This operation does not perform
+            logical page-table translation.
```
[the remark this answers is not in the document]
sjettTT: Addressed in `3e0ce7ceb0c`: the nanobind documentation now states `1 <= packed_rows <= 256` and `packed_rows <= num_indices <= 256`.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp
```diff
@@ -0,0 +1,219 @@
...
+        "{} and {} must have the same number of heads",
+        cache_name,
+        input_name);
+    TT_FATAL(
+        cache.logical_shape()[3] == input.logical_shape()[3],
+        "{} and {} must have the same head dimension",
+        cache_name,
+        input_name);
+    TT_FATAL(cache.logical_shape()[2] % 32 == 0, "{} rows per physical page must be tile aligned", cache_name);
+    TT_FATAL(cache.logical_shape()[3] % 32 == 0, "{} head dimension must be tile aligned", cache_name);
```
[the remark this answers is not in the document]
sjettTT: The current validation first pins every cache/input descriptor to the default non-transposed `TILE_HEIGHT × TILE_WIDTH`. The following checks state the API requirement for 32-row and 32-column alignment, with custom tile geometry rejected before dispatch.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/indexed_page_cache_nanobind.cpp
```diff
@@ -0,0 +1,60 @@
...
+            is ``[num_pages, num_heads, rows_per_page, head_dim]`` and packed input
+            shape is ``[1, num_heads, num_rows, head_dim]``. The operation updates
+            the cache tensors in place and serializes all rows owned by a head/width
+            worker, so multiple rows may safely target the same physical page.
+
+            Replicated tensors are supported on single- or multi-device meshes. The
+            four cache/input tensors may also use one identical tensor-parallel mesh
+            topology sharded over ``num_heads`` (dimension 1), ``head_dim``
+            (dimension 3), or both. Physical update indices must use the same mesh
+            shape and coordinates and remain replicated across those mesh axes.
```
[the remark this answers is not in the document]
sjettTT: Addressed in the PR description: replicated tensors and identical tensor-parallel cache/input topologies over heads, head dimension, or both are supported. Cache page/row sharding and sharded physical indices remain fail-closed.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+#include <tt-metalium/constants.hpp>
+
+#include "ttnn/device_operation.hpp"
+
+namespace ttnn::experimental::prim::indexed_fused_update_cache {
+
+namespace {
+
+constexpr uint32_t max_update_rows = 256;
```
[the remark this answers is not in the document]
sjettTT: The 256-row constant defines the initial staging and API bound. Every worker stages the complete replicated INT32 positions tensor in one DFB entry, yielding a maximum per-core entry of 1024 bytes and matching the maximum source rows consumed by the kernel.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation_types.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+struct IndexedFusedUpdateCacheInputs {
+    Tensor cache_tensor1;
+    Tensor input_tensor1;
+    Tensor cache_tensor2;
+    Tensor input_tensor2;
+    Tensor physical_update_idxs_tensor;
+};
+
+using IndexedFusedUpdateCacheResultSpec = std::vector<tt::tt_metal::TensorSpec>;
```
vtsilytskyiTT: Why vector, which allocated memory dynamically, if we know upfront that there will be 2 elements? std::pair, std::tuple, std::array should work.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "indexed_fused_update_cache_device_operation_types.hpp"
+#include "indexed_fused_update_cache_program_factory.hpp"
+
+namespace ttnn::experimental::prim::indexed_fused_update_cache {
+
+struct IndexedFusedUpdateCacheDeviceOperation {
+    using operation_attributes_t = IndexedFusedUpdateCacheParams;
+    using tensor_args_t = IndexedFusedUpdateCacheInputs;
+    using spec_return_value_t = IndexedFusedUpdateCacheResultSpec;
+    using topology_return_value_t = std::vector<tt::tt_metal::TensorTopology>;
```
vtsilytskyiTT, line 19: Again, 2 elements inside, no need in std::vector.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation_types.hpp
vtsilytskyiTT: Why vector, which allocated memory dynamically, if we know upfront that there will be 2 elements? std::pair, std::tuple, std::array should work.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/indexed_fused_update_cache/indexed_fused_update_cache_device_operation.cpp
```diff
@@ -0,0 +1,255 @@
...
+        static_cast<uint64_t>(args.cache_tensor1.logical_shape()[0]) * args.cache_tensor1.logical_shape()[2];
+    TT_FATAL(
+        total_cache_rows <= max_indexable_cache_rows,
+        "flattened cache row count exceeds the range addressable by INT32 physical indices");
+}
+
+void IndexedFusedUpdateCacheDeviceOperation::validate_on_program_cache_hit(
+    const operation_attributes_t& attrs, const tensor_args_t& args) {
+    validate_on_program_cache_miss(attrs, args);
+}
```
vtsilytskyiTT: Not required any more. If `validate_on_program_cache_hit` is absent, framework will call `validate_on_program_cache_miss` automatically.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+    constexpr auto source_height_tiles = get_arg(args::source_height_tiles);
+    constexpr auto cache_page_rows = get_arg(args::cache_page_rows);
+    constexpr auto total_cache_rows = get_arg(args::total_cache_rows);
+    constexpr auto worker_count = get_arg(args::worker_count);
+
+    constexpr uint32_t tile_width = 32;
+    constexpr uint32_t tile_height = 32;
+    constexpr uint32_t face_width = 16;
+    constexpr uint32_t face_height = 16;
+    constexpr uint32_t bytes_per_element = 2;
```
vtsilytskyiTT: I saw the question about BF16 support in PR description. Adding FP32 is fairly simple in your case. In kernel you need to read a compile-time argument like this:
```c++
constexpr auto bytes_per_element = get_arg(args::bytes_per_element);
```

In the factory pass this value as:
```c++
const auto bytes_per_element = input.element_size();
// ...
.compile_time_args = {
    {"num_heads", num_heads},
    // ...
    {"bytes_per_element", cache1.element_size()},
},
```
Then relax TT_FATAL restriction.
Maybe some more adjustments here and there and it should work. Just for you to consider :)

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+    // Indices stay on device so a captured trace can target different physical
+    // cache rows on every replay. Negative entries are per-row no-ops.
+    positions_dfb.reserve_back(1);
+    noc.async_read(positions, positions_dfb, positions_dfb.get_entry_size(), {.page_id = 0}, {.offset_bytes = 0});
+    noc.async_read_barrier();
+    positions_dfb.push_back(1);
+    CoreLocalMem<volatile int32_t> physical_positions(positions_dfb.get_read_ptr());
+
+    auto copy_rows = [&](const auto& source, const auto& cache, uint32_t head, uint32_t width_tile) {
+        constexpr uint32_t scratch_buffer_depth = 2;
```
vtsilytskyiTT: Don't hardcode, pass as a compile-time arg. This way it will always stay in sync with factory (already defined there).

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+                    break;
+                }
+                const int32_t physical_row_signed = physical_positions[source_row];
+                if (physical_row_signed < 0 || static_cast<uint32_t>(physical_row_signed) >= total_cache_rows) {
+                    continue;
+                }
+
+                const uint32_t physical_row = static_cast<uint32_t>(physical_row_signed);
+                const uint32_t physical_page = physical_row / cache_page_rows;
+                const uint32_t row_in_page = physical_row % cache_page_rows;
```
vtsilytskyiTT: Nit: This pattern is repeated often (at least 4 times). Python has `divmod`, you could add similar function here.

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
```diff
@@ -0,0 +1,159 @@
...
+            }
+            // Transaction IDs split completion accounting only. These unicast writes share the
+            // same NoC and default write VC, so program order (and last-source-row-wins) is retained.
+            writes_in_flight[current_slot] = wrote_current_slot;
+
+            if (has_next_tile) {
+                noc.async_read_barrier<NocOptions::TXN_ID>({.trid = next_slot + 1});
+                scratch_dfb.push_back(1);
+            }
+            scratch_dfb.pop_front(1);
```
vtsilytskyiTT, line 144: I'm worried that there is no barrier before `pop_front`. Please double check whether we need a `noc.async_write_barrier();` or at least `noc.async_writes_flushed` here.

vtsilytskyiTT (changes requested):

amorrisonTT (approved):

sjettTT:
@vtsilytskyiTT I pushed `263c7570` addressing the review feedback:

- Changed fixed-size result specs to `std::array<TensorSpec, 2>`.
- Removed `validate_on_program_cache_hit`; the framework applies the cache-miss validator on cache hits.
- Passed element size and scratch-buffer depth through compile-time arguments.
- Added a shared `divmod` helper.
- Added tagged `async_writes_flushed` before each scratch DFB slot release, with completion barriers before slot reuse.
- Documented why worker partitioning is independent of the physical page count.

`compute_output_topologies` continues to return `std::vector<TensorTopology>` because the framework recognizes the hook only for that exact return type:
https://github.com/tenstorrent/tt-metal/blob/263c7570b54500992be433e5636dc9dab5f48eee/ttnn/api/ttnn/device_operation.hpp#L511-L518

For the FP32 suggestion, the kernel byte width now comes from `cache1.element_size()`. The operation remains BF16-scoped in this PR so FP32 can receive dedicated validation and hardware coverage.

Validation at `263c7570`:

- Release `ttnn` build passed.
- Blackhole: 36/36 indexed-cache tests passed, including program-cache reuse, trace replay, replicated placement, head/head-dimension sharding, and 2x2 mesh.
- Wormhole B0: 29/29 applicable single-device tests passed with `TT_METAL_WATCHER=1`; seven mesh-only cases were deselected.
- PR checks: 41 passed, with zero failures or pending checks.

Could you take another look?

yieldthought (approved):

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/device/kernels/dataflow/indexed_fused_update_cache.cpp
vtsilytskyiTT: Addressed

vtsilytskyiTT (approved):
Comments addressed - PR approved. Thanks!

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+        FILES ${TTNN_OP_EXPERIMENTAL_INDEXED_PAGE_CACHE_API_HEADERS}
+        FILE_SET kernels
+        TYPE HEADERS
+        BASE_DIRS ${CMAKE_CURRENT_SOURCE_DIR}
+        FILES ${kernels}
+    PRIVATE
+        ${TTNN_OP_EXPERIMENTAL_INDEXED_PAGE_CACHE_SRCS}
+)
+
+target_link_libraries(ttnn_op_experimental_indexed_page_cache PRIVATE TT::Metalium PUBLIC TTNN::Core)
```
tt-rkim, line 33: Does `TTNN::Core` have to be linked as `PUBLIC`?

tt-rkim (approved):

ttnn/cpp/ttnn/operations/experimental/indexed_page_cache/CMakeLists.txt
sjettTT: Yes. `indexed_page_cache.hpp` is installed as a public API header and directly includes `ttnn/tensor/tensor.hpp` and exposes `ttnn::Tensor`. Consumers of this target therefore need `TTNN::Core` usage requirements; `PRIVATE` would cover only this target compilation/link. This is also consistent with the adjacent `paged_cache` and `indexer_score` targets.

PR #53433, opened 2026-08-17
tt-train/grpo: example cleanup, move logging & wandb init into ttml

ichovpanTT:
## Summary

* Move GRPO per-step logging out of the example scripts and into `ttml.trainers`, with a small TRL-style config surface.

### Behaviour changes

- `GRPOTrainer.__init__` auto-appends `GRPOMonitor` unless `disable_default_monitor` is set.
- CSV renames: `reward` → `reward_mean`, `avg_length` → `mean_completion_len`.
- `report_to` is a single string (`"none"` / `"wandb"`), not a TRL list.
- Callers no longer call `wandb.init(...)` themselves; if they do, the monitor logs into their run and leaves its lifecycle alone.
- **`step_time_s` semantics changed**: it is now the full per-step wall time — generation + host post-generation + reference log-probs + training loop + every non-monitor `on_step_end` callback + any checkpoint save on that step. Previously it stopped at `optimizer.step()`, so callback / checkpoint time landed outside `step_time_s`. This matches TRL's `step_time`.
- **`logging_steps` semantics aligned with TRL.** All `TrainerCallback` hooks now fire every optimizer step (previously non-monitor `on_step_end` was gated by `logging_steps`, which silently throttled things like `WeightSyncCallback` at `lcm(logging_steps, weight_sync_every)`). The built-in `GRPOMonitor` accumulates every step in constant-memory running stats (`_MetricStats`) and emits a CSV row / wandb log every `logging_steps` steps: interval-mean per key, except `min_completion_len` / `max_completion_len` (min / max over the window) and `lr` (current value). With `logging_steps = 1` (the default in all shipped yamls) behavior is unchanged; with `logging_steps > 1` each logged row now aggregates the interval instead of sampling one step from it. Booleans are also now accepted by the wandb scalar filter.
- **Multiple reward functions** (`reward_funcs=[...]`, TRL-style). `GRPOTrainer` now accepts either `reward_func` (single callable) or `reward_funcs` (list of callables); passing both or neither raises. Per-completion outputs are summed element-wise for the advantage computation. Per-component means are emitted as `<fn.__name__>_mean` on both the CSV and wandb (only when `len(reward_funcs) > 1`, so single-reward runs are unchanged). Both boolq examples (`grpo/boolq` and `grpo_remote_rollout/boolq`) are migrated to `reward_funcs=[accuracy_reward, brevity_reward]`; the reverse-text example stays on a single `reward_func`.
- **CSV header deferred to the first logging step.** `GRPOMonitor` no longer writes the header at `on_train_begin`; it writes the header and the first data row together on the first logging step, deriving the column set from `trainer.metrics` at that moment. Callbacks that populate `trainer.metrics` only in `on_step_end` (e.g. an eval callback writing `eval_similarity`) now Just Work — no `on_train_begin` seed idiom is needed. Keys that first appear after the header is committed are still dropped with a one-time warning. Internal rename: `_BASE_COLUMNS` → `_CSV_BASE_COLUMNS`.
- **Legacy custom-monitor compatibility.** The auto-append check now also skips when a callback's class name is literally `"GRPOMonitor"`, not just when it's an `isinstance` of the framework class — pre-refactor forks that still define their own local `class GRPOMonitor` are no longer paired with an auto-appended framework monitor writing to the same `grpo_metrics.csv`. Custom loggers using a different class name should still set `GRPOConfig.disable_default_monitor=True`.
- **Console output is now a metrics dict.** `GRPOMonitor._log_console` dumps `{"step": N, **scalars}` verbatim, matching `transformers.PrinterCallback.on_log`'s `print(logs)` shape. Every scalar in `trainer.metrics` (including user-callback extras like `eval_similarity` and per-reward-component means) surfaces automatically; the old fixed `Step N | Reward: … | Len: …` template is gone. `_NON_CSV_KEYS` (`prompts` / `completions` / `rewards`) are filtered out — they still go through `_maybe_log_completions`.
- **`num_completions_to_print` type-validated.** `GRPOConfig.__post_init__` now rejects non-integer values (including `True`) with a `TypeError` up front, instead of crashing later at `prompts_strs[:k]` when the first logging step tries to slice.

## Test plan

- [x] All four training yamls parse; `GRPOConfig` builds with the new `run_name` field and every yaml resolves the expected `model_source` / `model_config`.
- [x] `--help` works for both example scripts (imports resolve, argparse clean).
- [x] Standalone `GRPOMonitor` end-to-end test with a `SlowCallback` sleeping 50 ms in `on_step_end`: CSV records `SlowCallback_time_s = nan` at step 1 and `≈0.050s` at step 2 (one-step lag), plus `GRPOMonitor_time_s`.
- [x] `ReadLints` clean on every modified file.
- [x] `_MetricStats` / `_aggregate` policy table (mean default; `min` / `max` / `last` overrides) smoke-tested in isolation.
- [x] `_derive_reward_names` + `_init_rewards` validation branches (both / neither / empty list / valid single / valid multi) exercised in isolation; element-wise sum of `reward_funcs` verified against per-element addition.
- [x] Deferred-header CSV smoke test: callback writing only in `on_step_end` produces its column in the header; late-added key excluded via one-time warning.
- [x] `num_completions_to_print` type check smoke test: `1.5` / `True` / `"3"` all raise `TypeError`; ints pass.
- [x] End-to-end reverse-text run on p150 — easy to reproduce locally, not gated on this PR.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ichovpan/grpo-logging-framework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ichovpan/grpo-logging-framework)


jmalone-tt:
Change looks good overall to me, and it's nice to see the UX cleaned up here. 

Notes based on claude /review

  Silent behavior change to on_step_end cadence (likely bug). Previously on_step_end fired for every non-monitor callback on every optimizer step. The new code (grpo_trainer.py around line 1447) only fires non-monitor on_step_end when is_logging_step is true (num_steps % grpo_cfg.logging_steps == 0). Consequence:
  - WeightSyncCallback.on_step_end in grpo_remote_rollout/boolq/boolq_training_example.py — which drives cross-rank weight sync every weight_sync_every steps — will now only tick on logging steps. If logging_steps=5 and weight_sync_every=2, weight sync effectively runs at multiples of lcm(5, 2)=10. The PR description doesn't mention this, and the logging_steps docstring says it controls "logging," not callback frequency.
  Either (a) fire on_step_end every step and let callbacks self-gate, or (b) call this out explicitly and document that logging_steps now gates all on_step_end firing.
- should fix this

  on_before_optimizer_step is un-gated but its timings feed a gated dict. _callback_times is reset every step; non-monitor on_before_optimizer_step accumulates into it; but on non-logging steps the accumulated times are discarded silently. That's fine, but combined with the point above it means callback timing accuracy quietly degrades between logging steps.
- should fix too


  _maybe_log_wandb filters out NaN scalars but not bool. isinstance(value, (int, float)) includes bool; harmless today, but if a callback writes trainer.metrics["converged"] = True, it'll show up as grpo/converged: True — probably fine, worth being intentional about.
- for this one, I'm guessing we can just add bool to the isinstance tuple to show it is expected


I will approve once these are addressed

awliu-TT (approved):
Thanks for the cleanup, looks good to me (assuming you've run Qwen3 and Llama GRPO training examples to ensure WandB logging still works). Please address the 3 remaining copilot comments.

ichovpanTT:
Thanks for the reviews! Addressed the comments (incl from Copilot), wandb did work on boolq. Added a few nice things, TRL-style reward print in a dictionary, multiple reward functions support

jmalone-tt (approved):

Issue #52859, opened 2026-08-11
tt-run / generate_rank_bindings: accept FSD argument, map from FSD when available, fall back to PSD discovery


Riddy21:
## Summary

Today, `tt-run`'s auto-mapper flow (`--mesh-graph-descriptor` + `--hosts`) and Fabric Manager's `generate_rank_bindings` path always map against a Physical System Descriptor (PSD) that is **discovered live** from the cluster at runtime. There's no way to instead map against the **expected/as-built topology** (the Factory System Descriptor, FSD) directly.

Add an FSD argument to both `tt-run` and `generate_rank_bindings`, use it to drive the mapping when it's supplied and covers the requested topology, and fall back to today's live PSD discovery otherwise.

## Motivation

- The FSD already exists as the "what the cluster *should* look like" descriptor (vs. PSD, which is "what's live right now") and is used elsewhere (validation tooling's `--factory-descriptor-path`, Fabric Manager telemetry serving).
- Mapping from a known-good FSD avoids depending on live discovery (useful for degraded clusters, "skinny FSD" subsets, offline/CI mapping, and pre-flight checks before hardware is fully up).
- PSD-based live discovery should remain the default/fallback so nothing breaks for users not passing an FSD.

## Steps

- [ ] Add an FSD argument (e.g. `--factory-system-descriptor <path>`) to `tt-run`, and the equivalent to Fabric Manager's `generate_rank_bindings` path, accepting an FSD textproto — mirroring the existing `--factory-descriptor-path` convention used by the validation tooling.
- [ ] When an FSD is supplied, derive the topology graph directly from it (reusing existing FSD conversion utilities, e.g. `generate_cluster_descriptor_from_fsd()` in `tools/scaleout/factory_system_descriptor/utils.cpp`) instead of discovering the PSD live.
- [ ] Wire the FSD-derived topology into the existing topology mapper / rank-binding generation so mapping (Phase 1) runs against it exactly as it does today against a live-discovered PSD.
- [ ] Implement fallback logic: if no FSD is passed, or the supplied FSD doesn't cover the requested hosts/topology, fall back to the current behavior of discovering the PSD live.
- [ ] Add test coverage for both paths: mapping from a supplied FSD, and fallback to live PSD discovery when the FSD is absent/incomplete.
- [ ] Update `README_ttrun.md`, `MGD_README.md`, and the Automapper Guide to document the new argument, when to use it, and the fallback semantics.

## References

[...]

PR #53451, opened 2026-08-18
tt-run: add --factory-system-descriptor argument (plumbing)

Riddy21:
First part of #52859.

Adds a Factory System Descriptor (FSD) argument to `tt-run` and plumbs it end-to-end (RTOption + env-var propagation) **without yet consuming it for mapping**. When an FSD is not supplied, behavior is unchanged (live PSD discovery).

## What is FSD for?
- This PR is added in preparation for Fabric 2.0, in which we will be changing topology solver to be solving from the "Ideal state" (Factory system descriptor) of a cluster. So that we can utilize the dynamic fabric rerouting systems that @ubcheema and team are developing.
- Here are some docs about Factory system descriptor, Its mostly a subject that we've been using in the context of DC bringup, because they are most concerned with the "Ideal state" of the system (Basically exactly what connections we should be expecting between hosts). But basically it is an adjacency graph describing what ethernet connectivity should exist between chips and hosts.

https://docs.tenstorrent.com/cloud-native-support/factory-system-descriptor.html
https://github.com/tenstorrent/tt-metal/blob/main/tools/scaleout/README.md

## What's here
- **RTOption** (`tt_metal/llrt/rtoptions.{hpp,cpp}`): new `TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH` option with getter/setter/has-check.
- **`tt-run`** (`ttnn/ttnn/distributed/ttrun.py`): new `--factory-system-descriptor` flag, managed the same way as `--mock-cluster-rank-binding` — set per-rank in `get_rank_environment()` and blocklisted from parent-env passthrough. Propagated to Phase 1 (`generate_rank_bindings`) and Phase 2 (workload), and folded into the Phase 1 cache fingerprint.
- **`generate_rank_bindings`**: reads the FSD path from RTOptions (env-driven, no new CLI arg per design); currently logs it and falls back to live PSD discovery.
- **Docs/tests**: `README_ttrun.md` entry + unit tests in `test_ttrun.py`.

## What's deferred
Deriving the topology from the FSD (`generate_cluster_descriptor_from_fsd()` → `PhysicalSystemDescriptor` substitution in `generate_rank_bindings::main()`), the coverage/fallback predicate, and Stage 2 tests/docs. See `Plan.md` for the full follow-up plan.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ridvan/ttrun-fsd-argument)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ridvan/ttrun-fsd-argument)


agupta-TT (approved):

jbaumanTT (approved):

Riddy21:
Addressed the review comments in 56745c743f8:

- **🟡 Resolve the FSD path in legacy mode** — `legacy_flow` now `resolve_path(..., must_be_file=True, must_exist=True)`s `--factory-system-descriptor` (matching `new_mode_flow`), so remote MPI workers get an absolute, validated path. Added two regression tests: legacy mode resolves + propagates an absolute FSD path into the Phase-2 command, and a missing FSD errors out.
- **🟢 Keep the parameter contract stage-accurate** — updated the FSD docstrings/comments (`build_generate_rank_bindings_mpi_cmd`, `get_rank_environment`, `new_mode_flow`) to state the path is propagated through RTOptions but **not yet consumed for mapping** (Phase 1 still performs live PSD discovery).
- **🟢 Keep the implementation plan out of the repo root** — `Plan.md` was removed from this PR (moved to the Stage-2 topology PR #53593); also dropped the dangling `Plan.md` reference in `generate_rank_bindings.cpp`, pointing at #52859 instead.

tests/ttnn/distributed/test_ttrun.py
```diff
@@ -41,6 +41,7 @@
     PHASE2_MOCK_MAPPING_FILENAME,
     PHASE1_CACHE_ID_HEX_LEN,
     RANK_BINDINGS_MAPPING_FILENAME,
+    FACTORY_SYSTEM_DESCRIPTOR_ENV_VAR,
```
dgomezTT, line 44: we have ttnn options in a json file inside in a environment variable

tests/ttnn/distributed/test_ttrun.py
```diff
@@ -302,6 +303,37 @@ def test_build_generate_rank_bindings_mpi_cmd_hosts(self, temp_dir):
...
+        fsd_path = temp_dir / "fsd.textproto"
+        fsd_path.touch()
+        output_dir = temp_dir / "output"
+        hosts = ["node1", "node2"]
+
+        cmd = build_generate_rank_bindings_mpi_cmd(
+            executable, mgd_path, hosts, output_dir, factory_system_descriptor=fsd_path
+        )
+
+        assert f"{FACTORY_SYSTEM_DESCRIPTOR_ENV_VAR}={fsd_path}" in cmd
```
dgomezTT, line 321: we should (and I want to stress the word should) have tests for ttnn config. This definitely belongs there

tools/scaleout/src/generate_rank_bindings.cpp
```diff
@@ -676,6 +676,19 @@ int main(int argc, char** argv) {
     try {
         log_info(tt::LogFabric, "Generating rank bindings...");
 
+        // Factory System Descriptor (FSD) path comes from RTOptions, populated from the
+        // TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH environment variable (set by tt-run and propagated to this
+        // MPI worker). NOTE: FSD-driven mapping is not yet implemented; today we always fall back to live PSD
+        // discovery below. See issue #52859 for the follow-up that consumes this.
+        const auto& fsd_path = tt::tt_metal::MetalContext::instance().rtoptions().get_factory_system_descriptor_path();
```
dgomezTT, line 683: is it a ttnn concern or a metal concern?

tt_metal/llrt/rtoptions.cpp
```diff
@@ -58,6 +58,7 @@ enum class EnvVarID {
     TT_METAL_VISIBLE_DEVICES,                 // Comma-separated list of visible device IDs
     ARCH_NAME,                                // Architecture name (simulation mode)
     TT_MESH_GRAPH_DESC_PATH,                  // Custom fabric mesh graph descriptor
+    TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH,  // Factory System Descriptor (FSD) path
```
dgomezTT, line 61: same question here, is it metal or ttnn the one that should handle this?

ttnn/ttnn/distributed/README_ttrun.md
```diff
@@ -225,5 +225,6 @@ Both modes support:
...
 - `--force-rediscovery` – **New mode only.** Always run Phase 1 and refresh the Phase 1 cache even when a cache hit would otherwise skip `generate_rank_bindings`. Ignored with `--rank-binding` (legacy mode). When new mode continues to **Phase 2**, failures there (config, mpirun, non-zero program exit, or Ctrl+C during a hang) log a hint to try **`--force-rediscovery`**.
+- `--factory-system-descriptor` – Path to a Factory System Descriptor (FSD) textproto describing the as-built/expected topology. tt-run exports it to every rank via `TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH`, so both `generate_rank_bindings` (Phase 1) and the workload (Phase 2) read it through RTOptions. The FSD content is also folded into the Phase 1 cache fingerprint. When omitted, tt-run falls back to today's live PSD discovery. **Note:** this release only wires the argument through; FSD-driven mapping is not yet consumed (see `Plan.md`).
```
dgomezTT: where is plan.md?

ttnn/ttnn/distributed/ttrun.py
```diff
@@ -483,13 +483,21 @@ def phase1_rank_bindings_input_path(run_dir: Path, mgd_path_is_mapping_yaml: boo
 # Short prefix of SHA-256 for cache directory names (balance: path length vs collision risk for local cache).
 PHASE1_CACHE_ID_HEX_LEN = 16
 
+# Environment variable that carries the Factory System Descriptor (FSD) path. tt-run sets this per-rank from
+# --factory-system-descriptor (managed the same way as TT_METAL_MOCK_CLUSTER_DESC_PATH: set explicitly in
+# get_rank_environment() and blocklisted from parent-env passthrough). It is propagated to both Phase 1
+# (generate_rank_bindings) and Phase 2 (the workload binary), where it is consumed via RTOptions
+# (RunTimeOptions::get_factory_system_descriptor_path()).
+FACTORY_SYSTEM_DESCRIPTOR_ENV_VAR = "TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH"
```
dgomezTT, line 491: this definitely makes me think is a metal concern

dgomezTT:
Why are @akerteszTT or @riverwuTT not part of this pr?

akerteszTT (changes requested):
`RTOptions` is part of the Metalium public user-facing API.

There is no such thing as a Factory System in Metalium -- this feels very much like a TTNN concept that you are accidentally leaking into the API layer beneath. 

I don't think this is the right solution to your problem. Please follow up with @dgomezTT whether you can address your needs at the TTNN level.

ttnn/ttnn/distributed/README_ttrun.md
Riddy21: Sorry I deleted the plan file, I will be wiring this up very soon, this is in preparation for Fabric 2.0

tests/ttnn/distributed/test_ttrun.py
Riddy21: This is a tt-metal concept that we need in order to enable fabric 2.0 features, we cannot include this at the ttnn level because it deals with topology mapper which operates at tt-metal. If we don't inlcude it in tt-metal it will not be able to be consumed by fabric correctly

tests/ttnn/distributed/test_ttrun.py
Riddy21: See comment above

tools/scaleout/src/generate_rank_bindings.cpp
Riddy21: tt-metal

tt_metal/llrt/rtoptions.hpp
```diff
@@ -184,6 +184,11 @@ class RunTimeOptions {
     bool is_custom_fabric_mesh_graph_desc_path_set = false;
     std::string custom_fabric_mesh_graph_desc_path;
 
+    // Factory System Descriptor (FSD): the "as-built"/expected topology descriptor. When set, tooling
```
akerteszTT: Can you please improve this comment to explain the purpose of this option? 

It should explain what the option is for (from a user-facing perspective), that it is here only temporarily, and what the eventual expected solution is.

The current AI-gen comment doesn't really provide what's needed.

akerteszTT (approved):
My previous concern was based on a misunderstanding.

That said, the user-facing comments currently in the RTOptions files are inadequate to explain what this new RTOption is all about. Please improve it before merging this PR.

Approving now to avoid needing another review round -- but please do not merge without fixing this.

tt_metal/llrt/rtoptions.hpp
Riddy21: Rewrote the comment in `70d46c69` to cover the three points: (1) user-facing purpose — pointing rank-binding generation at the as-built factory topology instead of live PSD discovery; (2) that it's temporary/plumbing-only right now (path carried end-to-end but not yet consumed, just logs the fallback); and (3) the eventual consumption path (build PSD from the FSD-derived cluster descriptor, with coverage check + fallback to live discovery). Framing follows the Stage 1 / Stage 2 rollout in Plan.md.

cfjchu (approved):

PR #53454, opened 2026-08-18
BH LLK topk: replay-recorded step loads/stores for sort phases >= 4 (1.15-1.2x single-core topk)

nkapreTT:
### What

The single-core `ttnn.topk` local sort issues its phase>=4 step loop as discrete TTI load/store instructions on TRISC1. This records the 16-issue body into replay slots once per phase and replays it, removing the RISC issue bottleneck. Blackhole only (`tt_llk_blackhole` header), default-on for the topk step path, compile-time opt-out via `-DTOPK_DISABLE_REPLAY_STEP`.

### Measured

Blackhole p150a, Tracy device-kernel durations (medians), 90-cell three-arm A/B sweep across N=4096..131072 / k=8..512:

| shape | baseline | with replay |
|---|---|---|
| N=4096, k=8 | 661.5 µs | 1.166× |
| N=4096, k=64 | 1.288 ms | 1.202× |
| N=65536, k=32 | 10.96 ms | 1.154× |
| N=65536, k=512 | 158.6 ms | 1.172× |
| N=131072, k=512 | 317.9 ms | 1.172× |
| multi-core topk / ttnn.sort | — | ~1.00× (bound elsewhere) |

Zero correctness violations, zero hangs across the sweep; values/indices bit-compared against torch goldens. Consumers of this header (topk, sort, sampling, MoE gate ops) all pass their existing suites on the development branch carrying this change (`nkapre/sorting`).

A load-side replay variant was also measured and is deliberately NOT included (consistently ~5-8% worse than store-side, plus an unresolved full-battery-only hang signature under Tracy).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/topk-llk-replay-step)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/topk-llk-replay-step)

### DIFF
--- a/tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
@@ -14,6 +14,10 @@
 #include "lltt.h"
 #include "sfpi.h"
 
+// Replay windows for the phase >= 4 step loop's load16/store16
+#define TOPK_STEP_LOAD_REPLAY_START  16
+#define TOPK_STEP_STORE_REPLAY_START 24
+
 namespace ckernel
 {
 namespace sfpu
@@ -538,6 +542,8 @@ inline void _bitonic_topk_phases_steps(const int idir, const int i_end_phase, co
                 switch (ph)
                 {
                     case 0:
+                    {
+                        constexpr int replay_count = STABLE_SORT ? 6 : 4;
                         for (int d = 0; d < 4; d++)
                         {
                             // Groups of 16 datums being sorted at the same time
@@ -550,7 +556,6 @@ inline void _bitonic_topk_phases_steps(const int idir, const int i_end_phase, co
                             {
                                 lltt::replay(0, 8);
                             }
-                            constexpr int replay_count = STABLE_SORT ? 6 : 4;
                             if (init_phase)
                             {
                                 load_replay_buf<Exec>(16, replay_count, [] { bitonic_topk_ph0_st1_to_1<STABLE_SORT>(); });
@@ -571,12 +576,14 @@ inline void _bitonic_topk_phases_steps(const int idir, const int i_end_phase, co
                             }
                         }
                         break;
+                    }
                     case 1:
+                    {
+                        constexpr int replay_count = STABLE_SORT ? 10 : 6;
                         // Groups of 16 datums being sorted at the same time
                         for (int d = 0; d < 4; d++)
                         {
                             lltt::replay(0, 8);
-                            constexpr int replay_count = STABLE_SORT ? 10 : 6;
                             if (init_phase)
                             {
                                 load_replay_buf<Exec>(16, replay_count, [] { bitonic_topk_ph1_st2_to_1<STABLE_SORT>(); });
@@ -589,11 +596,13 @@ inline void _bitonic_topk_phases_steps(const int idir, const int i_end_phase, co
                             lltt::replay(8, 8);
                         }
                         break;
+                    }
                     case 2:
+                    {
+                        constexpr int replay_count = STABLE_SORT ? 14 : 9;
                         for (int d = 0; d < 4; d++)
                         {
                             lltt::replay(0, 8);
-                            constexpr int replay_count = STABLE_SORT ? 14 : 9;
                             if (init_phase)
                             {
                                 load_replay_buf<Exec>(16, replay_count, [] { bitonic_topk_ph2_st3_to_1<STABLE_SORT>(); });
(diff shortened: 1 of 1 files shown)


tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -14,6 +14,108 @@
...
 #include "sfpi.h"
 
+// ---------------------------------------------------------------------------
+// EXPERIMENTAL issue-rate knobs for _bitonic_topk_phases_steps (Blackhole).
+//
+// All three are OFF by default and every use site is #if-guarded, so the
+// default build is byte-identical to the unflagged kernel. They exist to
+// answer one question with measurements instead of assertion: is the local
+// sort limited by the SFPU backend or by the rate at which the math RISC-V
+// can issue into it?
```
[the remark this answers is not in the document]
nkapreTT: Both findings addressed in the amended push: (1) the stale "all knobs off / byte-identical" block was left over from the development branch's experiment era and contradicted the default-on replay path — the experiment knobs (TOPK_HOIST_INIT_GUARDS, TOPK_MOP_INNER_LOOP, TOPK_PROBE_RV_NOPS and their #if-guarded alternative paths, ~180 lines) are now removed entirely rather than documented, which also resolves the suppressed maintainability comment about benchmark-only scaffolding. What remains is exactly the shipping path: TOPK_REPLAY_STEP_STORE/LOAD default-on + the TOPK_DISABLE_REPLAY_STEP opt-out. Re-validated on Blackhole silicon after the strip: reduce/topk suite 224 passed / 0 failed.

rdjogoTT (changes requested):
If it's just better then I don't see a point of adding flags to disable it. 
You can keep the `#define TOPK_STEP_LOAD_REPLAY_START  16` type defines since they make things cleaner, but we don't need to fall back to the slower way.

Also the comments are a bit AI sloppy, need to be cleaned up or even just removed.

nkapreTT:
Done, both points:
- **Flags removed** — the replay path is now unconditional (no `TOPK_DISABLE_REPLAY_STEP` / `TOPK_REPLAY_STEP_*` lattice, no fallback to the discrete-issue path). Kept the `TOPK_STEP_LOAD_REPLAY_START` / `TOPK_STEP_STORE_REPLAY_START` defines as you suggested. Net: the diff shrank from 96 to ~40 added lines.
- **Comments cleaned** — the header block is now one short paragraph (what's recorded where, and the one safety invariant: every consumer of slots ≥ 16 re-records before replaying); the inline step comment is two lines.

LLK topk + moe-gate suites re-run on silicon after the change: 86 passed.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -14,6 +14,17 @@
...
 #include "sfpi.h"
 
+// Replay windows for the phase >= 4 step loop's load16/store16: each step's
+// `dist` is fixed, so the 8-instruction sequences are recorded once per step
+// and replayed afterwards (1 math-RISC issue per iteration instead of 8; the
+// SFPU sees an identical stream). Slots 0-15 hold the kernel-lifetime
+// load16/store16(4, 8) recordings. These windows overlap the phase-3 lattice
+// (slots 16+), which is safe because every consumer of slots >= 16 re-records
+// before it replays: the step loop re-records at the top of every step, and
+// the phase >= 4 tail re-records the lattice under `init_phase`.
```
rdjogoTT:
```suggestion
// Replay windows for the phase >= 4 step loop's load16/store16
```

nkapreTT: Applied verbatim.

rdjogoTT (approved):

PR #53455, opened 2026-08-18
BH LLK topk_xl: unfused SFPLOADMACRO primitives + fused-u16 end-to-end family, with LLK-level tests

nkapreTT:
**Consolidated LLK PR** (restructured per @pavlejosipovic's review direction: LLK changes as their own PR with LLK-level tests, reviewed by the LLK team first; the op-level change (#53457) stacks on this branch).

All Blackhole `topk_xl` LLK deltas of the `topk_large_indices` line, three primitive families in `ckernel_sfpu_topk_xl.h` + the `llk_api` shim + the compute-API header:

1. **Unfused merge/rebuild on SFPLOADMACRO-scheduled compare-exchange** — the InstructionTemplate bodies installed per entry; scheduling notes inline (silicon-validated).
2. **Fused-u16 end-to-end**: `_topk_xl_add_lsb_indices_rt_` (runtime chunk-id stamp into index bits [15:11] via `TT_SFPLOADI`) and the once-per-row global split `_topk_xl_separate_indices_row_major_global_(_init_)` recovering `chunk_id * K + within-chunk` from the fused [bf16|u16] word.
3. **Segmented-fusion support**: `_topk_xl_separate_indices_row_major_global_base_` (runtime 2^(5+log2 K)-aligned segment base OR'd into every decoded index — OR == ADD, disjoint bits). (An earlier draft carried a defensive index-tracking-disable primitive; an adversarial audit proved it unreachable — every fused init resets the whole SFPU LaneConfig via `_init_sfpu_config_reg` — so it was deleted rather than shipped untestable.)

**LLK-level tests** (`tt-llk/tests`, primitives exercised bare through the `topk_xl_test.cpp` harness):
- New `FUSED_E2E` kernel mode: runtime stamp + fused merge/rebuild + global split, `num_chunks` ∈ {2, 4, **32**} — 32 covers the full 5-bit stamp range (ids 0..31), the information-theoretic limit of the fused word.
- Segment-base split at two bases per K; partial `-inf` tail; the index-tracking-disable re-entry pattern; the unfused SFPLOADMACRO macro path (`test_topk_xl_unfused_macro.py`).
- New `planted` stimulus mode: K distinct bf16 highs at seeded positions over constant filler — the only exact-index golden possible past `search_len` 16384 (bf16 has no 2^16 distinct finite positive values; `0x3F80 + 16384` is already `+inf`).

[...]

PR #53458, opened 2026-08-18
topk_large_indices: multi-rectangle trees, hybrid row split, fused-u16 end-to-end

nkapreTT:
### What

**Stacked on #53457** (base branch = its head, so this diff shows only the delta: 11 files, +590/−199). Together with #53457 this takes the production 160-row k=2048 W=65536 call from 712.3 µs (as-shipped) to **388 µs**, with zero call-site changes.

- **Multi-rectangle trees**: the column-parallel factory generalizes from one rectangle to a grid tiling — one P-core in-place merge tree per rectangle, disjoint and concurrent; rows split contiguously as runtime args over capacity-tiled rectangles (empty rectangles run zero rows), so one cached program serves any row count with a given layout. Explicit `num_slices` becomes legal on multi-row shapes (measured: 30 rows @P=4 = **98.8 µs** vs 356.5 row-parallel, 3.6×). The device op never auto-selects the multi-row form: non-stable tie order differs across engines and the TILE/u16 opt-ins are row-parallel-only, so auto-routing would break the opt-ins-change-layout-never-results contract the suite guards.
- **Hybrid row split** (op-internal): canonical calls with rows > worker grid run full row-parallel waves plus a multi-rectangle remainder window over the *same un-sliced input* (new internal `row_start`/`row_count`: readers offset into global rows, writers stay output-relative) and concat the row outputs. 160×65536 k=2048: 712.3 → **467.0 µs**. The remainder-window model requires a ≥12.5% modeled win, so marginal splits never net negative.
- **Fused-u16 end-to-end** (rows of ≤ 32 LLK-window chunks; `FUSED_E2E` factory gate mirrored into the program hash via a single-sourced helper): the runtime chunk-id stamp keeps every merge/rebuild in the fused [bf16|u16] word; one global split per row recovers `chunk_id*K + within-chunk`. Per-row kernel 359.7 → **280.6 µs** (1.28×); the 160-row call lands at **388 µs**. Width > 65536 provably cannot run fused (a u16 must address every global position) and keeps the unfused body, which stays CI-covered via the boundary-width tests.

### Testing

Op suite **181 green on this branch's own build**, including: both fused-gate sides (31-chunk+tail / 32-chunk-exact / >32-chunk fallback widths), poisoned-tail `valid_length` under the fused program, values arm, and program-cache-hit reruns with row redistribution (60→30 rows share one program).

Tie-identity note for reviewers: `valid_length` vs physically-sliced comparisons now span the fused gate (different physical widths choose different engines), so tie order may differ *on exact bf16 ties only*. Audited on silicon: 2,882 index diffs, **0 non-ties**, value multisets bit-exact — and the suite now asserts exactly that contract (per-diff tie proof + bitwise sorted-value equality, not a tolerance).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/topk-large-indices-hybrid)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/topk-large-indices-hybrid)

---

## Perf over array length (requested by @pavlejosipovic)

160 rows, k=2048 (k=1536 identical — same 2048 LLK window), p150a, Tracy device-kernel time:

| W (positions) | main µs | this stack µs | speedup | µs per 64k |
|---|---|---|---|---|
| 5,120 | 69 | 66 | 1.06× | — |
| 25,600 | 295 | 181 | 1.63× | 464 |
| 65,536 | 726 | 388 | 1.87× | 388 |
| 131,072 | 1,447 | 752 | 1.92× | 376 |
| 262,144 | 2,890 | 1,480 | 1.95× | 370 |
| 524,288 | 5,775 | 2,932 | 1.97× | 367 |
| 1,048,576 | 11,545 | 5,837 | 1.98× | 365 |

Linear in width: the µs-per-64k column is flat (the small taper is fixed per-row overhead amortizing out); no cliffs at the old 65536 fused boundary. GLM 5.2 DSA-indexer shape (1M preallocated, 512k valid, k=2048): **5.77 → 2.93 ms (1.97×)**.

## How fused-u16 works past 65k positions (segmented fusion)

A fused [bf16|u16] word can only address 2^16 positions, so u16 fusion above 65k is impossible per se. Instead, rows run as ≤32-chunk **segments** (32 × 2048 = exactly the u16 limit): within a segment the chunk id is stamped at runtime into index bits [15:11] (segment-LOCAL, 5 bits always suffice) and every merge/rebuild stays fused; one split per segment decodes to u32 and ORs in the 65536-aligned segment base (OR == ADD, disjoint bits); segments fold into the running survivor with one unfused u32 merge each — 1 unfused merge per 31 fused, preserving ~98% of the fused rate at any width. For k ≥ 1024 this is the single codepath at every width (a ≤32-chunk row is the single-segment case), so program selection carries no width term and a growing valid_length never recompiles.

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/experimental/test_topk_large_indices.py
+++ b/tests/ttnn/nightly/unit_tests/operations/experimental/test_topk_large_indices.py
@@ -88,6 +88,30 @@ def test_topk_large_indices_row_major_bfloat16_uint32_indices(device, k, num_row
     _assert_topk_matches_torch(torch_input, tt_indices, k)
 
 
+def test_topk_large_indices_random_k2048_multichunk_multirow(device):
+    # PR2 triage repro (minimal): row-parallel factory, llk_k=2048, num_chunks=4,
+    # multiple rows, RANDOM input — the exact op-level combo behind the routed
+    # k=2048 cells that failed on silicon. If this passes while the routed
+    # composite fails, the composite's post-op chain (gather at index width
+    # 2048 -> gather's untested RM multi-core variant) is the culprit, not
+    # this op. Tie-safe value-multiset assertions (random bf16 has duplicates).
+    torch.manual_seed(0)
+    rows, n, k = 2, 8192, 2048
+    torch_input = torch.randn(rows, n, dtype=torch.bfloat16)
+
+    tt_indices = ttnn.experimental.topk_large_indices(_to_device(torch_input, device), k=k)
+    indices = ttnn.to_torch(tt_indices, dtype=torch.uint32).to(torch.int64)
+
+    assert indices.min() >= 0
+    assert indices.max() < n
+    for row_indices in indices:
+        assert row_indices.unique().numel() == k
+
+    actual_values = torch.gather(torch_input.float(), dim=-1, index=indices)
+    ref_values, _ = torch.topk(torch_input.float(), k, dim=-1, largest=True, sorted=True)
+    assert_equal(actual_values.sort(dim=-1).values, ref_values.sort(dim=-1).values)
+
+
 def test_topk_large_indices_random_bfloat16_ties_return_distinct_indices(device):
     torch.manual_seed(0)
     rows = 8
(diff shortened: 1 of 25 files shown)


nkapreTT:
Pushed an addition driven by the #53464 review: a `neginf_sentinel` op attribute (default `true` — existing contract unchanged; program-hashed; C++ + nanobind). `false` keeps each exact bf16 `-inf` lane's real stamped source position instead of the 0xFFFFFFFF sentinel — the routed `ttnn.topk` path uses it for stock/torch index parity. Kernel-side it compiles out `mark_neginf_indices` via a define on all four compute variants.

Also two robustness guards from a corruption found while extending u16 usage: the device op now `TT_FATAL`s when a multi-row rectangle launch is combined with `tile_output` or `index_dtype=UINT16` (those opt-ins are row-parallel/single-row-tree only; previously the u16 output spec under a u32 writer silently overflowed into the values buffer), and `hybrid_row_split` declines u16 (its remainder window is a multi-row rect launch) so u16 callers fall back to the plain row-parallel program.

New tests: sentinel-off on both factories, and a program-cache test proving the flag hashes into a distinct program. Op nightly suite green on p150a (perf pins still skipped pending #53459).

nkapreTT:
**Segmented fusion pushed — the op no longer falls back to the unfused body at any width** (request from @pjosipovicTT via the DSA-indexer scaling test).

Mechanism: rows wider than the 32-chunk fused ceiling run ≤32-chunk fused segments (segment-LOCAL 5-bit chunk-id stamp), each split once with the 65536-aligned segment base OR'd into the decoded index (one small new LLK variant, `_topk_xl_separate_indices_row_major_global_base_`), then folded into the running survivor with one unfused merge per segment — the tree path's silicon-shipped rebuild-ascending mirror pattern, relocated in-DST. DST fits the 8-tile FP32 budget exactly; the program-hash schema is unchanged (the define is a pure function of the already-hashed fused bit + k); one cached program serves every runtime width and valid_length, verified by a cache-reuse test that collapses a 256-chunk program to 1 segment at runtime. Gated on USER_K ≥ 1024: small-k wide rows keep the classic body because segmentation would lose chunk-skip, which is a measured win there. Fused (≤32-chunk) programs are byte-identical — control cell measured unchanged to the tenth of a µs.

Measured (p150a, Tracy device-kernel sums): 160×524288 k=2048 **3562→2932 µs**, 160×1M **7097→5837 µs** (1.22×; **1.97–1.98× vs main**); 640×524288 k=1536 **14139→10987 µs (1.29×)** — the exact width class that previously showed no improvement now wins at the same ratio as the ≤65536 fused class. New tests: 8 segmented-width cells including the 33-chunk single-chunk-tail-segment boundary, W=1M (indices to 2^20), non-chunk-multiple width, with-values, and the valid_length collapse. Full op nightly suite green.

Tie note (documented non-stable contract): in-segment ties resolve in fused stamp order, cross-segment ties in unfused merge order — tie identity at >32-chunk widths differs from the old classic path; values are exact as always.

nkapreTT:
(Correction to the mention above: @pavlejosipovic — mistyped the handle in the previous comment.)

nkapreTT:
**Preallocated-buffer follow-up pushed: tree slices now balance on `valid_length`** — closing the gap on @pavlejosipovic's GLM 5.2 scaling shape (1M preallocated buffer, 512k valid, k=2048, 160 rows).

`compute_slice_runtime` previously split the PHYSICAL chunk range evenly and clipped by the search width, so a short valid prefix emptied the trailing tree slices while the busy ones kept near-full per-slice work — the hybrid's remainder window measured 1424µs there vs 720µs when sized to the valid width. It now splits the VALID chunk range evenly; empty slices keep the existing -inf-fill service, everything stays runtime-only (one cached program serves any valid_length with freshly balanced slices — the hash is untouched), and the hybrid's cost model also evaluates on the searched width.

Measured on p150a at the reported shape: **5774µs (main) → 2932µs (1.97×)**, i.e. wave 2204µs + rebalanced remainder 720µs + concat — under the requested 3.0ms. Full-valid cells are unchanged (identical slice layout when valid == logical width). Full op + routed suites green (437 passed; the 4 int32 failures are pre-existing #53466).

One correction for the record: an earlier comment of mine attributed 2204µs to the remainder window and 1424µs to the wave — that attribution was inverted (the 2-row control probe settled it: the row-parallel wave costs ~2200µs/row at 256 chunks; the remainder trees were the fast side, 720–1424µs depending on balance). The code comments now carry the corrected numbers, and an interim `valid < 7/8·n` hybrid-decline gate built on the inverted reading was reverted before it ever shipped — declining the hybrid measures 4400µs, strictly worse.

nkapreTT:
One-codepath consolidation pushed (per @pavlejosipovic's requirements): for k ≥ 1024 the row-parallel factory now selects the segmented body at EVERY width — a ≤32-chunk row runs the single-segment path, which executes exactly the FUSED_E2E sequence, so the separate fused-e2e variant is unreachable in this class and **the width term leaves program selection entirely**: one cached program serves any width and any runtime valid_length (the extended cache test pins no-recompile across the old 65536 boundary). Selection is single-sourced (`compute_body_mode`) into both the kernel defines and `compute_program_hash` so they can never skew. Smaller k keeps the previous three-way (fused-e2e / classic+chunk-skip) unchanged.

Perf equivalence measured (p150a): 160×65536 k=2048 program medians 189.8µs vs 189.7 pre-consolidation; 640×51200 k=1536 1073.0 vs 1072.2; the 1M-prealloc/512k-valid cell holds 2932µs. Perf-over-width is linear for fixed rows: ~364–388µs per 64k positions from 5k through 1M (constant per-chunk cost across the old fused cliff). Contract untouched: defaults, output specs, hash schema, and the documented non-stable tie behavior are all as before. Suites green (437 passed; the 4 int32 failures are pre-existing #53466).

nkapreTT:
Closing per @pavlejosipovic's review direction ("split this into LLK changes / add LLK level tests / get that reviewed / then get single op level change"). The stack is now:

- **#53455** — ALL topk_xl LLK changes consolidated (unfused SFPLOADMACRO primitives + the fused-u16 end-to-end family: runtime chunk-id stamp, global split, segment-base split, index-tracking disable), with **LLK-level tests** added to `tt-llk/tests` (FUSED_E2E harness mode, num_chunks to 32 = the full 5-bit stamp range, segment-base cells, tail, re-entry pattern; 105 passed on silicon).
- **#53457** — the SINGLE op-level PR, stacked on #53455: everything this PR carried (multi-rect trees, hybrid, segmented fusion, valid_length rebalancing, neginf contract) merged with the 4a content into one op change touching only `ttnn/` + op tests. Byte-identical functional tree to what was validated here (437 passed); the perf table and segmented-fusion write-up moved to its description.
- #53464 (routing) retargeted onto #53457's branch; #53465 unchanged on top.

All discussion threads on this PR (iwroszTT's review, the u16 corruption find, the valid_length rebalance) are preserved here for reference; the code lives on in the two PRs above.

PR #48843, opened 2026-07-02
#48140: Add INT32 index support to topk

vsureshTT:
### Ticket
Closes #48140

### Problem description
`topk` accepted only `UINT16` and `UINT32` index tensors. The op needs to also support `INT32` indices (as the `sampling` op already does).

### What's changed
`INT32` is now a valid index dtype, handled identically to the existing `UINT32` path — both are 4-byte and bit-identical for the non-negative positions `topk` produces — on **both** the single-core and multi-core paths.

- Validation accepts `INT32` for the optional input indices tensor and the preallocated index output.
- The 16-bit-vs-32-bit decision is derived from the resolved output index dtype (`resolve_index_dtype` / `index_cb_data_format_for` helpers) rather than the reduced-dimension size, which also fixes a latent off-by-one at `dim == 65535`.
- `INT32` index circular buffers are configured as `UInt32` for the on-device sort datapath (the transpose/sort/merge LLKs handle `UInt32`, not `Int32`). The writer copies the raw tile bytes into the `INT32`-typed output buffer unchanged.
- 32-bit indices are supported on the multi-core path: the local and final compute kernels enable `fp32_dest_acc_en` for 32-bit indices so the wider index values survive the datapath without truncation, and the multi-core memory-cost analysis now uses the real 4-byte index tile size.
- Fixed a pre-existing bug: a `UINT32` index output on a multi-core-eligible shape previously produced incorrect results.
- Updated the op docstring and stale "multi-core only supports UInt16" comments; added `test_topk_int32_indices` and moved the `(bfloat16, int32)` case out of the "should raise" list.

The default/auto index dtype is unchanged (`UINT16`/`UINT32` chosen by dimension size). `INT32` is obtained by supplying a preallocated `INT32` index tensor (or an `INT32` input `indices_tensor`).

Impact: for a multi-core-eligible shape (`1x1x32x8192`, k=50), the multi-core path is selected for `UINT16`, `UINT32`, and `INT32` indices, all with correct results. The full `tests/ttnn/unit_tests/operations/reduce/test_topk.py` suite passes (113 passed, 76 pre-existing BFLOAT8_B xfails) with no regressions.

### Checklist

- [ ] [All post-commit tests](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml?query=branch:vsureshTT/topk-int32-support)
[...]

Issue #33492, opened 2025-12-01
Sort `ckernel::topk_local_sort` kernel produces wrong results with `stable` flag set to True


mgajewskiTT:
# Problem

This issue is related to the implementation of the `stable` flag described in:
[https://github.com/tenstorrent/tt-metal/issues/20625](https://github.com/tenstorrent/tt-metal/issues/20625)

After exposing the `stable` flag in the LLK and adding support for it in the compute kernel, value sorting appears correct — `torch_sort_values` and `ttnn_sort_values` match. However, when comparing `torch_sort_indices` with `ttnn_sort_indices`, mismatches occur.

For a functional description of the `stable` flag, please refer to the previous ticket.

Support for the `stable` flag on the compute kernel side is available on the branch:
`mgajewskiTT/27223-tracking-stable-flag-support-in-topk_local_sort-llk`.
## Reproduction

```python
@pytest.mark.parametrize(
    "shape, dim, descending, stable",
    [
        ([32, 64], -1, False, True),
        ([32, 64], -1, True, True),
    ],
)
def test_sort_standard(shape, dim, descending, device, stable):
    torch.manual_seed(0)

    torch_dtype = torch.bfloat16
    input = torch.randn(shape, dtype=torch_dtype)

    ttnn_input = ttnn.from_torch(input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
    torch_sort_values, torch_sort_indices = torch.sort(input, dim=dim, descending=descending, stable=stable)
    ttnn_sort_values, ttnn_sort_indices = ttnn.sort(ttnn_input, dim=dim, descending=descending, stable=stable)

    assert torch_sort_values.shape == ttnn_sort_values.shape
    assert torch_sort_indices.shape == ttnn_sort_indices.shape

    assert list(ttnn_sort_values.shape) == shape
    assert list(ttnn_sort_indices.shape) == shape

    if len(shape) == 0 or len(shape) == 1:
        assert torch_sort_values == ttnn.to_torch(ttnn_sort_values)
    else:
        assert_allclose(torch_sort_values, ttnn.to_torch(ttnn_sort_values))
        if stable:
            assert_allclose(torch_sort_indices, ttnn.to_torch(ttnn_sort_indices))
```

vvukomanovicTT:
Hi @mgajewskiTT,

I took a look at this issue .

There are two problems
[...]

PR #53464, opened 2026-08-18
ttnn.topk: route large-k, wide/non-pow2 small-k, and MoE-gate shapes onto topk_large_indices (BH)

nkapreTT:
### What

**Stacked on #53458** (base = its head; the diff here is routing + tests only). Composite-level routing — the device op's program hash is untouched. On Blackhole, bf16 last-dim `largest=True, stable=False` calls with no user indices tensor, no preallocated outputs, and no `sub_core_grids` route through `ttnn.experimental.topk_large_indices` when the shape falls off the stock fast paths:

| arm | shapes | measured |
|---|---|---|
| large-k | 64 < k ≤ 2048 | k=2048 W=65536: 631.5 ms → **51.5 µs (12,265×)**; k=512: 161.6 ms → 20.9 µs |
| small-k structural | k ≤ 64, padded W ≥ 65535 or non-pow2, ≥ 4096 | 40–423× off the linear single-core factory |
| MoE gates | k ≤ 16, padded W 128–512, ≤ 32 rows | 24.2 → 8.2 µs, 77.5 → 8.3 µs (2.95×/9.29×), zero model changes |
| multi-row rectangles | rows fit a rectangle tiling | rows=32 sampling shapes: routed 213.6 → **128.7 µs (74.5× vs stock e2e)** |

Everything else keeps today's engine bit-for-bit — including `stable=True`, `largest=False`, and all disqualifying-argument calls.

### Tie semantics (the deliberate contract)

The routed path returns a correct top-k **set** with deterministic-but-unspecified tie order (the op is non-stable; `stable=True` never routes). The added tests assert value-exactness (both sides sorted descending, bitwise) and index validity (`input[index] == value`, in-range, unique) — never index-order equality.

### Testing

- Routed regression tests (large-k, non-pow2, bfp8-inf fallback predicate) + a sweep-framework `large_k` suite for CI coverage of the routed path.
- Validated on this branch's own build: reduce/topk + op nightly = **414 passed**; the only failures are the four `test_topk_int32_indices[...W=8192...k=50]` cases, which **fail identically on unmodified `main` on Blackhole silicon** (verified by differential run against a pure-main `topk.cpp` build — pre-existing #48843 behavior at that shape, will file separately).
- `topk.cpp` here is a verified 3-way merge of the routing block with main's INT32-index support (disjoint regions, clean merge).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/topk-large-k-routing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/topk-large-k-routing)

### DIFF
--- a/tests/sweep_framework/sweeps/reduction/topk/topk.py
+++ b/tests/sweep_framework/sweeps/reduction/topk/topk.py
@@ -56,11 +56,33 @@ parameters = {
             [0, 1, 2, 3],
         ],
         "largest": [True, False],
-        "k": [32],  # only k = 32 is supported for now
+        "k": [32],  # small-k suite; large k is swept by the "large_k" suite below
         "input_a_dtype": [ttnn.float32, ttnn.bfloat16, ttnn.bfloat8_b],
         "input_layout": [ttnn.TILE_LAYOUT, ttnn.ROW_MAJOR_LAYOUT],
         "input_a_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
         "output_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
+    },
+    "large_k": {
+        # Large-k coverage (64 < k <= 2048). On Blackhole + bfloat16 + last-dim +
+        # largest these route through the ttnn.experimental.topk_large_indices
+        # composite (indices from the op, values gathered back from the
+        # original tensor — exact); everywhere else stock single-core would
+        # take seconds-to-minutes per vector, so those combinations are
+        # invalidated below rather than swept.
+        "input_shape": gen_shapes([1, 1, 1, 8192], [1, 1, 4, 131072], [1, 1, 1, 8192], 20)
+        + gen_shapes([1, 1, 1, 100000], [1, 1, 2, 100000], [1, 1, 1, 1], 2)
+        + gen_shapes([1, 1, 32, 8192], [1, 1, 64, 65536], [1, 1, 32, 8192], 6)
+        # Routed-envelope ceiling is 2^19: the value-recovery gather parks a
+        # full W * 2 B input row-stick in L1, so wider rows fall back to the
+        # stock single-core factory (seconds-to-minutes per vector here).
+        + gen_shapes([1, 1, 4, 262144], [1, 1, 8, 524288], [1, 1, 4, 262144], 4),
+        "dim": [-1],
+        "largest": [True],
+        "k": [96, 128, 256, 512, 1024, 2048],
+        "input_a_dtype": [ttnn.bfloat16],
+        "input_layout": [ttnn.TILE_LAYOUT],
+        "input_a_memory_config": [ttnn.DRAM_MEMORY_CONFIG],
+        "output_memory_config": [ttnn.DRAM_MEMORY_CONFIG],
     },
     "xfail": {
         "input_shape": gen_shapes([1, 1, 32, 64], [6, 12, 256, 1024], [1, 1, 32, 64], 64)
@@ -101,6 +123,18 @@ parameters = {
 def invalidate_vector(test_vector) -> Tuple[bool, Optional[str]]:
     if len(test_vector["input_shape"]) != 4:
         return True, "Input shape must be 4D"
+    if test_vector["k"] > 64:
+        # The large_k suite covers the Blackhole routed path
+        # (ttnn.experimental.topk_large_indices). The standard sweep matrix
+        # schedules every regular module on the wormhole-n150 lane
+        # (compute_sweep_matrix.py: compute_standard_matrix) and has no
+        # Blackhole lane, so these vectors are pre-staged as INVALID: on WH
+        # the stock single-core factory takes seconds-to-minutes per vector
+        # (recorded as hangs, triggering device resets). Routed-path CI
+        # coverage lives in tests/ttnn/unit_tests/operations/reduce/
+        # test_topk.py (Blackhole-gated). Flip this to arch-conditional
+        # sub-checks when a Blackhole sweeps lane exists.
+        return True, "large_k suite is Blackhole-only (routed path); no Blackhole sweep lane yet"
     if test_vector["dim"] != -1:
         return True, "Only the last dim is supported right now"
     if test_vector["dim"] * (-1) > (len(test_vector["input_shape"])):
@@ -134,7 +168,20 @@ def run(
     data_seed = random.randint(0, 20000000)
     torch.manual_seed(data_seed)
 
-    input_shape = sanitize_shape(input_shape, "topk")
+    if k > 64:
+        # The large_k suite exists to cover the Blackhole routed path
+        # (ttnn.experimental.topk_large_indices). Elsewhere these vectors land
+        # on the stock linear single-core factory (seconds-to-minutes per
+        # vector against the 30s timeout, recorded as hangs). Generation-time
+        # invalidation cannot see the target arch (the generator is offline),
+        # so fail fast here instead.
+        assert "blackhole" in ttnn.get_arch_name().lower(), "large_k suite is Blackhole-only (routed path)"
+        # sanitize_shape rounds the last dim up to the next power of two,
+        # which would silently retarget the non-pow2 vectors (e.g. W=100000)
+        # onto already-covered pow2 widths. Non-pow2 width is one of the two
+        # structural reasons the routed path exists — run the declared shape.
+    else:
+        input_shape = sanitize_shape(input_shape, "topk")
 
     torch_input_tensor_a = gen_func_with_cast_tt(
         partial(torch_random, low=-100, high=100, dtype=torch.float32), input_a_dtype
--- a/tests/ttnn/unit_tests/operations/reduce/_topk_route_cells_bench.py
+++ b/tests/ttnn/unit_tests/operations/reduce/_topk_route_cells_bench.py
@@ -1,0 +1,40 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""Per-cell composite bench for the ttnn.topk -> topk_large_indices routing
+boundary questions (review #53464): pow2-4096 arm and tiny-width large-k
+floor. One cell per invocation; run under `python -m tracy -r -v` and sum
+DEVICE KERNEL DURATION over the report CSV.
+
+Usage: _topk_route_cells_bench.py <H> <W> <k> <routed|stock>
+`stock` forces the stock engine by passing sub_core_grids (the routing
+predicate declines whenever sub_core_grids is provided).
+"""
+import sys
+import torch
+import ttnn
+from loguru import logger
+
+H, W, K = int(sys.argv[1]), int(sys.argv[2]), int(sys.argv[3])
+ARM = sys.argv[4]
+WARMUP, ITERS = 3, 30
+
+device = ttnn.open_device(device_id=0)
+device.enable_program_cache()
+
+torch.manual_seed(2005)
+torch_input = torch.randn(1, 1, H, W, dtype=torch.bfloat16) * 0.9
+tt_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+
+kwargs = dict(dim=-1, largest=True, sorted=True)
+if ARM == "stock":
+    grid = device.compute_with_storage_grid_size()
+    kwargs["sub_core_grids"] = ttnn.CoreRangeSet(
+        [ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(grid.x - 1, grid.y - 1))]
+    )
+
+for _ in range(WARMUP + ITERS):
+    values, indices = ttnn.topk(tt_input, K, **kwargs)
+    ttnn.synchronize_device(device)
+
+logger.info(f"BENCH_DONE H={H} W={W} k={K} arm={ARM} iters={WARMUP + ITERS}")
+ttnn.close_device(device)
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -546,3 +546,206 @@ def test_topk_multicore_values_beyond_first_tile_row(num_rows, largest, device):
     assert torch.allclose(
         got_s, ref_s, atol=1e-2
     ), f"values fabricated past first tile row: max_diff={(got_s - ref_s).abs().max():.4f}"
+
+
+# ---------------------------------------------------------------------------
+# Large-k Blackhole routing: ttnn.topk with bf16, dim=-1, largest=True,
+# stable=False and 64 < k <= 2048 is routed at the composite level through
+# ttnn.experimental.topk_large_indices (untilize -> clamp to lowest-finite
+# bf16 -> op -> gather of the ORIGINAL values by index -> tilize -> index
+# dtype match; a slice pair only when k is not a multiple of 16), bypassing
+# the single-core cliff (the device op's own multi-core path is gated at
+# k <= 64 + pow2 width + width < 65536).
+#
+# Tie semantics on the routed path: the returned index SET is a correct top-k
+# set with deterministic-but-unspecified tie order, so these tests assert
+# value-exactness (both sides sorted descending) and index validity
+# (input[index] == value, in-range, unique), never index-order equality.
+# ---------------------------------------------------------------------------
+
+from models.common.utility_functions import is_blackhole
+
+
+def run_topk_large_k_routed_test(N, C, H, W, k, device):
+    torch.manual_seed(2005)
+    shape = [N, C, H, W]
+    torch_input = torch.randn(shape, dtype=torch.bfloat16) * 0.9
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+
+    pyt_values, _ = torch.topk(torch_input, k, dim=-1, largest=True, sorted=True)
+
+    ttnn_values, ttnn_indices = ttnn.topk(ttnn_input, k, dim=-1, largest=True, sorted=True)
+
+    # Index dtype contract must match the stock device op: UINT16 iff the
+    # tile-padded width fits 16 bits, else UINT32.
+    padded_w = 32 * ((W + 31) // 32)
+    uint16_expected = padded_w <= UINT16_MAX
+    assert ttnn_indices.dtype == (ttnn.uint16 if uint16_expected else ttnn.uint32)
+
+    desired_shape = [N, C, H, k]
+    assert list(ttnn_values.shape) == desired_shape
+    assert list(ttnn_indices.shape) == desired_shape
+
+    torch_values = ttnn.to_torch(ttnn_values)
+    torch_indices = ttnn.to_torch(ttnn_indices, dtype=torch.uint16 if uint16_expected else torch.uint32).to(torch.int64)
+
+    # Values: exact, order included -- both sides are sorted descending, so
+    # this holds even under ties.
+    assert_equal(torch_values, pyt_values)
+
+    # Indices: value-based validation (tie order is unspecified).
+    assert torch_indices.min() >= 0
+    assert torch_indices.max() < W
+    gathered = torch.gather(torch_input, -1, torch_indices)
+    assert_equal(gathered, torch_values)
+    for row_indices in torch_indices.reshape(-1, k):
+        assert row_indices.unique().numel() == k
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+@pytest.mark.parametrize("k", [96, 128, 256, 512, 1024, 2048])
+@pytest.mark.parametrize("W", [8192, 32768, 65536, 131072])
+def test_topk_large_k_routed(k, W, device):
+    # All combos take the routed (topk_large_indices) path: bf16, dim=-1,
+    # largest=True, stable=False, 64 < k <= 2048, W <= 2^19.
+    run_topk_large_k_routed_test(1, 1, 32, W, k, device)
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+@pytest.mark.parametrize("k", [512, 2048])
+def test_topk_large_k_routed_non_pow2_width(k, device):
+    # W=100000: neither a power of two nor 16-bit -- the old multi-core gates
+    # excluded it entirely; the routed path supports it natively.
+    run_topk_large_k_routed_test(1, 1, 32, 100000, k, device)
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+@pytest.mark.parametrize("W", [262144, 524288])
+@pytest.mark.parametrize("k", [1536, 2048])
+def test_topk_large_k_routed_wide(k, W, device):
+    # The routed envelope extends to 2^19 — the RM gather parks one full
+    # W * 2 B input row-stick in an L1 CB, so 2^20 would not fit Blackhole's
+    # 1.5 MB budget (see large_k_route_max_width in topk.cpp). Fewer rows
+    # keep host-side reference cost and transfer size sane at these widths.
+    run_topk_large_k_routed_test(1, 1, 8, W, k, device)
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+def test_topk_large_k_routed_single_row(device):
+    # A single logical row engages topk_large_indices' column-parallel
+    # (intra-row multi-core) factory underneath the routed composite.
+    run_topk_large_k_routed_test(1, 1, 1, 65536, 2048, device)
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+def test_topk_large_k_routed_neginf_lanes(device):
+    # Rows whose top-k contains exact -inf: the routed path clamps the op's
+    # input to the lowest finite bf16 (the op then sees no -inf and stamps a
+    # REAL source position for every lane — its 0xFFFFFFFF sentinel never
+    # fires) and gathers values from the ORIGINAL tensor, so -inf values are
+    # bit-exact and -inf lanes carry real positions (stock/torch parity).
+    # W=65536 is tile-aligned, so every returned index must be a real
+    # in-range column.
+    torch.manual_seed(2005)
+    W, k, finite_count = 65536, 512, 100  # W=65536 -> tile-padded > 65535 -> uint32 indices
+    torch_input = torch.full((1, 1, 32, W), -float("inf"), dtype=torch.bfloat16)
+    finite = (torch.randn(1, 1, 32, finite_count) * 0.9).to(torch.bfloat16)
+    torch_input[..., :finite_count] = finite
+
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+    ttnn_values, ttnn_indices = ttnn.topk(ttnn_input, k, dim=-1, largest=True, sorted=True)
+    assert ttnn_indices.dtype == ttnn.uint32
+
+    torch_values = ttnn.to_torch(ttnn_values)
+    torch_indices = ttnn.to_torch(ttnn_indices, dtype=torch.uint32).to(torch.int64)
+    pyt_values, _ = torch.topk(torch_input, k, dim=-1, largest=True, sorted=True)
+
+    # Values match torch exactly, including the -inf tail.
+    assert_equal(torch_values, pyt_values)
+
+    # ALL lanes (finite and -inf alike) carry real, in-range, unique,
+    # self-consistent indices -- no sentinel anywhere.
+    assert torch_indices.min() >= 0
+    assert torch_indices.max() < W
+    gathered = torch.gather(torch_input, -1, torch_indices)
+    assert_equal(gathered, torch_values)
+    for row_indices in torch_indices.reshape(-1, k):
+        assert row_indices.unique().numel() == k
+    # The finite lanes specifically must point into the finite prefix.
+    assert torch_indices[..., :finite_count].max() < finite_count
+
+
+@pytest.mark.skipif(not is_blackhole(), reason="fallback-path guards mirror the Blackhole routing predicate")
+def test_topk_large_k_routing_fallbacks(device):
+    # Shapes/args just OUTSIDE the routing predicate must keep the stock
+    # (device-op) path and stay correct. Kept at W=8192 so the stock
+    # single-core runs are fast.
+    W = 8192
+
+    # k=2049 exceeds the topk_large_indices ceiling (2048) -> stock path.
+    run_topk_test(1, 1, 32, W, 2049, ttnn.bfloat16, 3, True, True, device)
+
+    # largest=False is not supported by topk_large_indices -> stock path.
+    run_topk_test(1, 1, 32, W, 512, ttnn.bfloat16, 3, True, False, device)
+
+    # stable=True promises lowest-index tie-breaking -> stock path. A
+    # constant row makes this a real guard: stable top-96 of an all-equal
+    # row MUST return indices 0..95 in order, which the routed (non-stable)
+    # engine does not promise -- so this asserts INDICES, not just values.
+    # (Value-only asserts pass on either engine and would not catch the
+    # stable guard being deleted from the routing predicate.)
+    torch_input = torch.ones(1, 1, 32, W, dtype=torch.bfloat16)
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+    ttnn_values, ttnn_indices = ttnn.topk(ttnn_input, 96, dim=-1, largest=True, sorted=True, stable=True)
+    assert_equal(ttnn.to_torch(ttnn_values), torch.ones(1, 1, 32, 96, dtype=torch.bfloat16))
+    expected_indices = torch.arange(96, dtype=torch.int64).expand(1, 1, 32, 96)
+    assert_equal(ttnn.to_torch(ttnn_indices).to(torch.int64), expected_indices)
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+def test_topk_routed_k_not_multiple_of_16(device):
+    # k=100 exercises the routed k % 16 != 0 tail: topk_large_indices
+    # requires a multiple of 16, so the op runs at k_rounded=112 and
+    # post_topk_transform slices back down to 100.
+    run_topk_large_k_routed_test(1, 1, 32, 8192, 100, device)
+
+
+
+@pytest.mark.skipif(not is_blackhole(), reason="routing predicate is Blackhole-only")
+def test_topk_large_k_routing_engages(device):
+    # Guards the routing predicate itself: outputs alone are also satisfied
+    # by the stock path, so a silently narrowed predicate would demote the
+    # whole routed suite to stock with everything still green. Graph capture
+    # shows which ops actually ran.
+    torch.manual_seed(2005)
+    torch_input = torch.randn(1, 1, 32, 8192, dtype=torch.bfloat16) * 0.9
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+
+    def ran_large_indices(captured) -> bool:
+        # The device op surfaces as TopkLargeIndices*/topk_large_indices
+        # depending on node type; normalize before matching.
+        return "largeindices" in str(captured).lower().replace("_", "")
+
+    # Routed shape (k=96 > 64): topk_large_indices must appear in the trace.
+    ttnn.graph.begin_graph_capture()
+    ttnn.topk(ttnn_input, 96, dim=-1, largest=True, sorted=True)
+    assert ran_large_indices(ttnn.graph.end_graph_capture())
+
+    # Stock shape (k=32, pow2 width >= 8192 -> multi-core bitonic eligible):
+    # the routed op must NOT appear.
+    ttnn.graph.begin_graph_capture()
+    ttnn.topk(ttnn_input, 32, dim=-1, largest=True, sorted=True)
+    assert not ran_large_indices(ttnn.graph.end_graph_capture())
+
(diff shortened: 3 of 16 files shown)


ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+    // width fits 16 bits (compute_output_specs compares the padded shape).
+    const bool emit_u16 = padded_width <= std::numeric_limits<uint16_t>::max();
+    const bool tile_native_output = !use_rects && k_rounded <= large_k_route_tile_output_max_k;
+
+    // BFLOAT16 values + (UINT32 or UINT16) indices of the top k_rounded per
+    // row, both sorted by value descending; -inf value lanes carry exact
+    // bf16 -inf values and the sentinel index (0xFFFF under the UINT16
+    // contract, exactly what a UINT32 -> UINT16 typecast produces).
+    auto [values, indices] = ttnn::experimental::topk_large_indices_with_values(
+        op_input,
```
iwroszTT: When `k` is a multiple of 32 (every tested k), the user's `memory_config` is silently dropped: the `topk_large_indices` outputs inherit the input tensor's memory config, and `post_topk_transform_tensor` applies `input_memory_config` only inside the slice branches, which are skipped when `adjusted_k == k` and the shape already matches. E.g. `ttnn.topk(dram_input, k=512, memory_config=ttnn.L1_MEMORY_CONFIG)` routes and returns DRAM outputs, where the stock path honors the config via `ttnn::prim::topk`. In-tree MoE-gate callers do pass `memory_config` on routable shapes.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+    // Match the device op's index dtype contract: UINT16 iff the tile-padded
+    // width fits 16 bits (compute_output_specs compares the padded shape).
+    const bool emit_u16 = padded_width <= std::numeric_limits<uint16_t>::max();
+    const bool tile_native_output = !use_rects && k_rounded <= large_k_route_tile_output_max_k;
+
+    // BFLOAT16 values + (UINT32 or UINT16) indices of the top k_rounded per
+    // row, both sorted by value descending; -inf value lanes carry exact
+    // bf16 -inf values and the sentinel index (0xFFFF under the UINT16
+    // contract, exactly what a UINT32 -> UINT16 typecast produces).
+    auto [values, indices] = ttnn::experimental::topk_large_indices_with_values(
```
iwroszTT: Behavior change on a public API: routed calls return the 0xFFFF/0xFFFFFFFF sentinel index for lanes whose value is genuinely bf16 `-inf`, where the stock path (at pow2/32-multiple widths, which need no host `-inf` padding) returns real in-range positions. Masked logits are the common case: `[1,1,32,8192]` with 300 finite tokens, k=512 routes and 212 lanes come back as index 65535. In-tree consumers feed topk indices straight into `ttnn.scatter` (models/demos/gemma4/tt/router.py:126) and `ttnn.gather` (models/common/modules/moe/tt_moe_gate.py:643), where a sentinel index reads/writes out of bounds. Latent today (gates keep >= k finite lanes) but unguarded — worth either mapping sentinels to a safe in-range index on the routed path or documenting + asserting the precondition.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+            return false;
+        }
+    }
+    if (transformed_tensor.dtype() != DataType::BFLOAT16) {
+        return false;
+    }
+    if (transformed_tensor.layout() != Layout::TILE) {
+        return false;
+    }
+    if (transformed_tensor.memory_config().is_sharded()) {
```
iwroszTT, line 323: `should_route` checks the input tensor's memory config but never the `memory_config` parameter: a sharded output config routes and silently returns interleaved outputs, where the stock path raises `TT_FATAL("Sharded implementation not supported yet")` (topk_device_operation.cpp:193). With k%32!=0 it's worse — the sharded config is forwarded as `ttnn::slice`'s output config, producing an output the topk contract rejects or a confusing slice-internal failure. Suggest declining the route (or failing fast) when `memory_config` is sharded.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+// when the device op's index-dtype contract calls for UINT16, emits UINT16
+// indices natively too — the former tilize-values, tilize-indices, and
+// typecast stages are gone. k_rounded == 2048 keeps the RM-output + tilize
+// (+ typecast) chain, which measures faster there.
+std::vector<Tensor> run_topk_large_indices_route(const Tensor& transformed_tensor, const uint32_t k_rounded) {
+    const auto& lshape = transformed_tensor.logical_shape();
+    const uint32_t flattened_rows =
+        ttnn::operations::experimental::topk_large_indices::flattened_rows_excluding_last_dim(lshape);
+    const uint32_t padded_width = transformed_tensor.padded_shape()[-1];
+    const bool tile_native_input = flattened_rows == 1 && padded_width >= large_k_route_tile_input_min_width;
```
iwroszTT: The predicate checks `layout() == TILE` but not `tensor_spec().tile()`: on the `tile_native_input` arm the tensor is fed directly to the `topk_large_indices` tile reader, which hardcodes 32x32 tiles (2048-byte pages, 16x16 face offsets). A bf16 TILE tensor with a non-standard tile (e.g. `Tile([16,32])`, 1024-byte pages) shaped `[1,1,1,65536]` satisfies this arm and is misread page-by-page — silently wrong values and indices, where the stock path derives everything from the real tile spec. Suggest gating the route on the default 32x32 tile.

tests/sweep_framework/sweeps/reduction/topk/topk.py
```diff
@@ -101,6 +118,17 @@
 def invalidate_vector(test_vector) -> Tuple[bool, Optional[str]]:
     if len(test_vector["input_shape"]) != 4:
         return True, "Input shape must be 4D"
+    if test_vector["k"] > 64:
```
iwroszTT, line 126: `invalidate_vector` filters dtype/layout/largest but not architecture, so on non-Blackhole sweep runners every large_k vector runs the stock linear single-core factory — the suite comment above says these combinations "are invalidated below rather than swept", but nothing invalidates them by arch. E.g. `[1,1,64,65536]` k=2048 runs seconds-to-minutes on the single-core kernel against the framework's 30s timeout — recorded as hangs, triggering device resets and starving the WH sweep queue. Suggest invalidating the large_k suite on non-Blackhole.

tests/sweep_framework/sweeps/reduction/topk/topk.py
```diff
@@ -56,12 +56,29 @@
...
         "output_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
     },
+    "large_k": {
+        # Large-k coverage (64 < k <= 2048). On Blackhole + bfloat16 + last-dim +
+        # largest these route through ttnn.experimental.topk_large_indices'
+        # multi-core path (values + indices, exact); everywhere else stock
+        # single-core would take seconds-to-minutes per vector, so those
+        # combinations are invalidated below rather than swept.
+        "input_shape": gen_shapes([1, 1, 1, 8192], [1, 1, 4, 131072], [1, 1, 1, 8192], 20)
+        + gen_shapes([1, 1, 1, 100000], [1, 1, 2, 100000], [1, 1, 1, 1], 2)
```
iwroszTT, line 73: These W=100000 vectors (and any non-pow2 width) don't actually run at their declared width: `run()` passes `input_shape` through `sanitize_shape(input_shape, "topk")`, which rounds the last dim up to the next power of two — so they execute at 131072, duplicating existing pow2 vectors, while the stored vector metadata says 100000. Non-pow2 width handling is one of the two structural reasons the route exists, and only the two fixed W=100000 unit tests cover it. Suggest bypassing the width rounding for this suite — the routed path under test is exactly the thing that makes non-pow2 legal.

tests/ttnn/unit_tests/operations/reduce/test_topk.py
```diff
@@ -546,3 +546,142 @@ def test_topk_multicore_values_beyond_first_tile_row(num_rows, largest, device):
...
+    gathered = torch.gather(torch_input, -1, torch_indices)
+    assert_equal(gathered, torch_values)
+    for row_indices in torch_indices.reshape(-1, k):
+        assert row_indices.unique().numel() == k
+
+
+@pytest.mark.skipif(
+    not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
+)
+@pytest.mark.parametrize("k", [96, 128, 256, 512, 1024, 2048])
```
iwroszTT, line 611: Two newly-routed regions ship with zero coverage: the MoE-gate arm (k<=16, padded width 128-512, <=32 rows — the arm that hits production expert gates) is exercised by no test or sweep vector, and every routed k in the PR is a multiple of 32, so the routed k%32!=0 tail (op at k_rounded, then TILE bf16 + UINT16/UINT32 slice from k_rounded down to k) never executes. Relatedly, none of the positive routed tests asserts the routed engine actually engaged — outputs alone are also satisfied by the stock path, so a silently narrowed predicate (a future constant bump, a padded_shape semantics change) demotes the whole suite to stock with everything still green. Suggest adding e.g. `[1,1,32,512]` k=10 and a routed k=100 case, plus an engagement assertion (program-cache entry inspection or similar).

tests/ttnn/unit_tests/operations/reduce/test_topk.py
```diff
@@ -546,3 +546,142 @@ def test_topk_multicore_values_beyond_first_tile_row(num_rows, largest, device):
...
+    # largest=False is not supported by topk_large_indices -> stock path.
+    run_topk_test(1, 1, 32, W, 512, ttnn.bfloat16, 3, True, False, device)
+
+    # stable=True promises lowest-index tie-breaking -> stock path.
+    torch.manual_seed(2005)
+    torch_input = torch.randn(1, 1, 32, W, dtype=torch.bfloat16) * 0.9
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+    ttnn_values, _ = ttnn.topk(ttnn_input, 96, dim=-1, largest=True, sorted=True, stable=True)
+    pyt_values, _ = torch.topk(torch_input, 96, dim=-1, largest=True, sorted=True)
+    assert_equal(ttnn.to_torch(ttnn_values), pyt_values)
```
iwroszTT: This stable=True arm is vacuous as a routing guard: it asserts values only, and the sorted top-k value multiset is engine-independent — deleting the `if (stable) return false;` guard in `should_route_to_topk_large_indices` routes stable calls onto the non-stable engine, breaks index parity for tied inputs, and this assert still passes. A tie-heavy input (e.g. a constant row, where stable must return indices 0..95 in order) with an index assertion makes it a real guard. Context: #33492 is making stable=True mean exact torch-stable tie order on the stock path, so this guard is what keeps that promise from silently routing onto the non-stable engine.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+    if (k <= large_k_route_min_k_exclusive) {
+        const uint32_t padded_width = transformed_tensor.padded_shape()[-1];
+        // Wide arm: route only cells the device op's multi-core bitonic
+        // cannot take (mirrors select_program_factory requirements #1-#2,
+        // topk_device_operation.cpp:66-72 — padded width must be < 65535 and
+        // a power of two), where stock otherwise falls to the linear
+        // single-core factory. Cells that fail only verify_multi_core_cost
+        // stay on the stock path (unchanged, conservative).
+        const bool is_pow2 = padded_width != 0 && (padded_width & (padded_width - 1)) == 0;
+        const bool multicore_structurally_ineligible = padded_width >= std::numeric_limits<uint16_t>::max() || !is_pow2;
```
iwroszTT: Two envelope boundaries look unintended: (a) this mirror of select_program_factory "requirements #1-#2" omits requirement #1's `multi_core_min_width = 8192` floor, so pow2 padded width 4096 — the only pow2 in [4096, 8192), bitonic-ineligible in stock — stays on the ~0.5 ms single-core factory while non-pow2 4128 routes, contradicting the 4096-floor rationale; (b) the large-k arm below (~line 359) has no width floor at all, only width >= k_rounded, so e.g. `[1,1,32,128]` k=96 routes and pays untilize + op (+2 slices when k%32!=0) where stock single-core at W=128 is ~24 us by this PR's own gate-arm measurement. Was the asymmetry with the small-k arm's 4096 floor deliberate?

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+        // <= 32 flattened rows — the measured decode-gate region (see the
+        // gate_route_* constants above). Everything here is below the
+        // bitonic's 8192 width floor, so only the single-core factory is ever
+        // displaced. transformed_tensor is 4D with the target dim last, so
+        // logical_volume / width is exactly the flattened row count (zero
+        // volume already returned earlier in ttnn::topk).
+        const uint64_t flattened_rows = transformed_tensor.logical_volume() / transformed_tensor.logical_shape()[-1];
+        const bool gate_arm = k <= gate_route_max_k && padded_width >= gate_route_min_padded_width &&
+                              padded_width <= gate_route_max_padded_width &&
+                              flattened_rows <= gate_route_max_flattened_rows;
```
iwroszTT: The <=32-flattened-rows cap makes one model's MoE gate straddle two engines: decode (<=32 rows) routes to topk_large_indices while prefill (hundreds of rows) stays stock, so tie selection among equal expert scores (common in bf16 after sigmoid/softmax rounding) differs between prefill and decode within the same model. Within the documented contract, but it's a silent cross-phase consistency change no model test asserts — worth a comment here documenting the trade-off, or reconsidering the row cap.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+    const Tensor op_input =
+        tile_native_input ? transformed_tensor : ttnn::to_layout(transformed_tensor, Layout::ROW_MAJOR);
+
+    // Multi-row rectangle trees: when every row can own a concurrent P-core
+    // tree (rows <= grid tiling capacity) and the op's cost model says the
+    // tree wins, pass num_slices explicitly. allow_multi_row=true is
+    // ROUTING'S opt-in, not the op's: routed calls already changed engine vs
+    // stock (tie order re-audited at the I5 relaxation), so the rect engine's
+    // different-but-equal tie identity is within the same acceptance class.
+    // rows > capacity come back disabled and keep nullopt — the op's internal
```
iwroszTT: This comment says rows > capacity "come back disabled and keep nullopt — the op's internal hybrid row split handles rows > grid on its own", but `hybrid_row_split` (topk_large_indices_device_operation.cpp:300) returns nullopt whenever `tile_output` is set, and the route sets `tile_output=true` for exactly these cells when k_rounded <= 1024. So a routed large-k call with rows >> grid (the large-k arm has no row cap) runs a plain row-parallel launch with an idle partial last wave instead of the hybrid peel this justification relies on.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,285 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+// expert gates select k <= 16 experts out of a few hundred, on <= 1 tile row
+// of tokens. Measured cells (p150a, bare-op ceilings vs stock single-core):
+//   gpt-oss gate  32x128  k=4  -> 4.07 us op vs 24.2 us stock (5.9-6.3x)
+//   qwen3.5 gate  32x512  k=10 -> 4.04 us op vs 77.5 us stock (19.2-20.2x)
+// Every padded width in [128, 512] is below the multi-core bitonic's
+// multi_core_min_width = 8192 (topk_constants.hpp), so this arm can only ever
+// replace the linear single-core factory, never the bitonic. The bounds are a
+// measured-contiguous-region gate (both endpoints measured, everything
+// between is the same shape class), NOT a hardware limit: k <= 16 keeps
+// k_rounded at exactly one LLK window (16), and the <= 32-row cap pins the
```
iwroszTT: The rationale "k <= 16 keeps k_rounded at exactly one LLK window (16)" doesn't match what ships: the route rounds with `large_k_route_k_multiple = 32`, so gate-arm calls run the op at k=32 (two LLK windows) plus a k_rounded!=k slice tail — a different configuration from the bare-op cells the cited 4 us / 2.95x / 9.29x measurements were taken at. Worth re-measuring the shipped configuration (or rounding gate-arm k to 16). Related shipped-cost note: the RM-tail typecast at ~line 452 looks avoidable — the op supports `index_dtype=UINT16` independent of `tile_output` (program factory line 124), so the k_rounded==2048 / use_rects tail could emit UINT16 natively instead of tilizing UINT32 indices and dispatching a separate typecast.

iwroszTT (commented):
Reviewed the routing layer against the stock path's contract (automated review, findings individually verified). 12 comments: 4 correctness/validation gaps on the routed path (memory_config drop, -inf sentinel indices, sharded-config validation, tile-spec assumption), 2 sweep-infra issues, 2 test-coverage gaps, 2 gate-envelope/behavior questions, 2 comment-accuracy/efficiency notes.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Fixed. The routed return now conforms both outputs to `memory_config.value_or(input.memory_config())` whenever they differ (`ttnn::to_memory_config`), matching the stock device op's output-placement semantics on every routed shape — including the `adjusted_k == k` case that skipped the slice branches.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Fixed, one level deeper than the suggested mapping: the op (in the base PR #53458) grew a `neginf_sentinel` attribute — default `true`, so the experimental op's own contract is unchanged; part of the program hash; threaded through the C++ API and nanobind. The routed path passes `neginf_sentinel=false`, so `-inf` value lanes keep their REAL stamped source positions (the fused [bf16|u16] word already carries them) — no sentinel anywhere on `ttnn.topk`, and the `scatter`/`gather` consumers you cite get in-range indices.

One documented caveat (parity with stock's own looseness): lanes the op itself `-inf`-pads (width not a chunk multiple, short `valid_length` tail) carry their padded position, which can exceed the logical width — same class of behavior as the stock path returning indices into its own `-inf` padding.

Coverage added: op-level tests for both factories plus a program-cache test proving the flag hashes separately; `test_topk_large_k_routed_neginf_lanes` now asserts real, in-range, unique, self-consistent indices on ALL lanes (finite prefix must resolve into the finite region).

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Fixed: `should_route_to_topk_large_indices` now takes the user `memory_config` and declines when it is sharded, keeping the stock path's loud `TT_FATAL` instead of silently returning interleaved outputs (or feeding a sharded config into the slice tail).

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Fixed: the `tile_native_input` arm now additionally requires the standard 32x32 tile spec; non-standard tile specs take the untilize path, which normalizes them before the op's tile reader ever sees the buffer.

tests/sweep_framework/sweeps/reduction/topk/topk.py
nkapreTT: Fixed, with one deviation from the suggested mechanism: `invalidate_vector` runs in the offline generator (`sweeps_parameter_generator.py`), which has no view of the arch the vectors will eventually run on, so generation-time invalidation can't express "Blackhole-only". Instead `run()` fails fast on non-Blackhole for k>64 — a sub-second assert with a self-explanatory message (the pattern already used by e.g. the softplus sweep's `is_wormhole_b0` assert) — so no 30s-timeout "hangs", no device resets, no starved WH queue. If the framework grows a runtime-skip status this can be upgraded to a true skip.

tests/sweep_framework/sweeps/reduction/topk/topk.py
nkapreTT: Fixed: k>64 vectors bypass `sanitize_shape` and run their declared width (non-pow2 handling is exactly what's under test); the small-k suite keeps the rounding.

tests/ttnn/unit_tests/operations/reduce/test_topk.py
nkapreTT: All three added:
- `test_topk_moe_gate_routed`: `[1,1,32,512]` k=10 (gate arm; also a k%32!=0 case — op at k_rounded=32, sliced to 10);
- `test_topk_routed_k_not_multiple_of_32`: `[1,1,32,8192]` k=100 (large-k arm k%32 tail, op at 128 sliced to 100);
- `test_topk_large_k_routing_engages`: `ttnn.graph` capture asserts `TopkLargeIndicesDeviceOperation` appears in the trace for a routed shape (k=96) and does NOT for a stock-eligible one (k=32, pow2 W=8192) — a silently narrowed predicate now fails loudly instead of demoting the suite to stock.

tests/ttnn/unit_tests/operations/reduce/test_topk.py
nkapreTT: Fixed: the stable=True arm now feeds an all-equal row and asserts INDICES == 0..95 in order (plus exact values). Deleting the `stable` guard from the predicate routes onto the non-stable engine, whose tie order is unspecified — this test now catches that, which value-only asserts provably could not. Noted re #33492: this guard is what keeps torch-stable tie order from silently routing away.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Both boundaries were unintended; (a) is fixed and both are now measured (Tracy device-kernel sums over the steady-state composite chain, 33 iterations per cell, p150a):

(a) The small-k arm's structural-ineligibility test now includes requirement #1's floor: pow2 padded width below `multi_core_min_width` (8192) routes. Measured at `[1,1,32,4096]`: k=32 routed **17.4 us** (untilize 2.8 + op 10.1 + tilize 3.3 + typecast 1.2) vs stock single-core **661.5 us** — **38x**; k=64 routed **18.4 us** vs stock **1288.3 us** — **70x**.

(b) The missing large-k width floor turns out not to be needed: `[1,1,32,128]` k=96 measures routed **6.9 us** (untilize 2.2 + op 4.7, TILE-native output, no slice at k%32==0) vs stock single-core **40.7 us** — **5.9x**. The stock single-core at k=96 is substantially slower than the k<=32 gate-arm cells suggested (it scales with k), so even the smallest legal width wins routed. Deltas this size are far outside run-to-run noise.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Documented at the gate constants: the row cap can put one model's MoE gate on two engines across phases (decode routed, prefill stock), so tie selection among equal expert scores can differ between phases. Both engines honor the documented non-stable contract; the comment now states that a model needing cross-phase tie consistency should pass `stable=true` (which never routes) rather than rely on unspecified tie order. Kept the row cap itself: it pins exactly the measured decode-token class, and widening it is a measurement task, not a correctness one.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: Fixed: `tile_native_output` now also requires flattened rows <= the worker grid, so rows>grid cells keep ROW_MAJOR outputs and the op's internal hybrid row split engages as the justification claimed; the false comment is rewritten to match.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
nkapreTT: (a) Measured at the shipped configuration (op at k_rounded=32 + two slices), `[1,1,32,512]` k=10, Tracy device-kernel sums, 33 iters: routed **8.0 us** (untilize 2.3 + op 4.1 + slices 1.6) vs stock **87.9 us** (device op 86.3 + slices) — **11.0x**. The two-window k=32 op run costs ~4.1 us vs the ~4.0 us bare-op k=16 cells, so the 32-rounding is effectively free and keeps `large_k_route_k_multiple` single-sourced; the comment's rationale is updated by these shipped-config numbers.

(b) Taken where the op supports it — with one important discovery from applying it wholesale: native UINT16 now replaces the typecast on the TILE-native arm and on the plain row-parallel RM path. It is NOT applied to rectangle launches or single-row RM trees, because the u16 flex writer is row-parallel/single-row-tree only: requesting u16 under a multi-row rect launch halves the indices page size under a writer that still emits u32, overflowing into the values buffer — silent corruption (we reproduced exactly this on every u16-eligible rect cell while applying the suggestion). The base PR now `TT_FATAL`s on that combo instead of corrupting, and the op's hybrid declines u16 (its remainder window is a multi-row rect launch), falling back to the plain row-parallel program which supports u16 at any row count. Extending the u16 writer to the rect/tree paths is follow-up work.

nkapreTT:
Review fix batch pushed (all 12 threads answered inline). Summary of what changed:

**Correctness fixes** — user `memory_config` honored (conformed post-transform) and sharded configs declined; `tile_native_input` gated on the standard 32x32 tile spec; `tile_native_output` yields to the op's internal hybrid when rows exceed the worker grid; routed calls pass the new `neginf_sentinel=false` (op flag added in #53458) so `-inf` lanes carry real source positions — stock/torch parity, no sentinel on `ttnn.topk`.

**Boundary fixes, measured** — pow2 widths below the 8192 bitonic floor now route (W=4096: 38x/70x at k=32/64); the large-k arm needs no width floor (W=128 k=96: routed 6.9 us vs stock 40.7 us); gate arm re-measured at the shipped k_rounded=32 configuration (W=512 k=10: routed 8.0 us vs stock 87.9 us, 11x).

**Native u16, with a discovery** — the typecast tail is eliminated on the TILE-native and plain row-parallel paths. Applying it to rectangle launches corrupted values (u16 output spec under a u32 writer overflows into the values buffer): #53458 now TT_FATALs on that combo and its hybrid declines u16. Extending the u16 writer to rect/tree paths is follow-up.

**Tests/sweep** — gate-arm and k%32!=0 cells; graph-capture engagement assertion (routed shape must run `TopkLargeIndicesDeviceOperation`, stock-eligible shape must not); tie-heavy stable=True index guard; sweep large_k fails fast on non-Blackhole and bypasses the pow2 `sanitize_shape` rounding.

Validation on p150a: full `test_topk.py` + op nightly suite — 421 passed; the only failures are the 4 pre-existing INT32/UINT32 cells tracked in #53466 (fail on pure main).

nkapreTT:
Routed-width envelope raised: `large_k_route_max_width` 2^19 → 2^20, so `ttnn.topk` no longer falls back to the stock single-core path anywhere in the prefill KV range. Earned by new correctness cells (`test_topk_large_k_routed_wide`: W ∈ {262144, 524288, 1048576} × k ∈ {1536, 2048}), a sweep-suite extension to 1M, and a measured direct-op sweep on p150a: with the base PR's segmented fusion, 160-row cells hold **1.9–2.0× vs main from 100k through 1M** (e.g. 1M k=2048: 11545→5837+tails µs). The sampling mirror constant is updated in the stacked PR.

iwroszTT (approved):

tests/ttnn/nightly/unit_tests/operations/experimental/_topk_li_width_sweep.py
```diff
@@ -0,0 +1,28 @@
...
+torch.manual_seed(2005)
+# Chunked generation to keep host memory sane at 1M width.
+torch_input = torch.randn(ROWS, W, dtype=torch.bfloat16) * 0.9
+tt_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.ROW_MAJOR, device=device)
+
+for _ in range(WARMUP + ITERS):
+    indices = ttnn.experimental.topk_large_indices(tt_input, k=K)
+    ttnn.synchronize_device(device)
+
+print(f"BENCH_DONE rows={ROWS} W={W} k={K} iters={WARMUP + ITERS}")
```
bbradelTT: Please use logger.

tests/ttnn/unit_tests/operations/reduce/_topk_route_cells_bench.py
```diff
@@ -0,0 +1,39 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
bbradelTT: USA, Inc.

tests/sweep_framework/sweeps/reduction/topk/topk.py
```diff
@@ -134,7 +163,20 @@ def run(
...
 
-    input_shape = sanitize_shape(input_shape, "topk")
+    if k > 64:
+        # The large_k suite exists to cover the Blackhole routed path
+        # (ttnn.experimental.topk_large_indices). Elsewhere these vectors land
+        # on the stock linear single-core factory (seconds-to-minutes per
+        # vector against the 30s timeout, recorded as hangs). Generation-time
+        # invalidation cannot see the target arch (the generator is offline),
+        # so fail fast here instead.
+        assert "blackhole" in ttnn.get_arch_name().lower(), "large_k suite is Blackhole-only (routed path)"
```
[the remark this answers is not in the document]
nkapreTT: Verified and fixed — this was real: `compute_standard_matrix` schedules every regular module on the wormhole-n150 lane, so the in-run assert would have turned the topk module red on nightly/comprehensive sweeps, and no Blackhole lane exists to deliver the claimed coverage. The large_k suite is now unconditionally INVALID at generation time (recorded as skipped-with-reason, never executed), pre-staged for a future Blackhole sweeps lane — the note in `invalidate_vector` says exactly when to flip it. Actual routed-path CI coverage is the Blackhole-gated unit tests in `test_topk.py` (which run on the BH lanes, e.g. the viommu nightly), and the PR description's coverage claim now matches that.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
```diff
@@ -178,6 +181,333 @@ std::vector<Tensor> post_topk_transform_tensor(
...
+// ---------------------------------------------------------------------------
+// Large-k routing onto ttnn::experimental::topk_large_indices (Blackhole)
+// ---------------------------------------------------------------------------
+//
+// The device op's multi-core bitonic path is gated at k <= 64 (plus pow2 width
+// and width < 65536), so every larger k runs the single-core factory: k=512 at
+// W=65536 measures ~158 ms on one core. topk_large_indices is a Blackhole-only
+// multi-core top-k that has NONE of those gates (arbitrary width up to 2^30,
+// k up to 2048), but returns only UINT32 row-major indices, sorted by value
+// descending. This composite routes eligible calls through it:
```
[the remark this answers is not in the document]
nkapreTT: Fixed — both sites. The orientation comment now describes the op's current surface (values output, TILE-layout I/O, native UINT16 where the writers implement it, neginf_sentinel=false), and the contract-matching bullets were also stale after the sentinel work landed (-inf lanes now carry real positions, not the sentinel) — updated together.

tests/ttnn/unit_tests/operations/reduce/test_topk.py
```diff
@@ -546,3 +546,206 @@ def test_topk_multicore_values_beyond_first_tile_row(num_rows, largest, device):
...
         got_s, ref_s, atol=1e-2
     ), f"values fabricated past first tile row: max_diff={(got_s - ref_s).abs().max():.4f}"
+
+
+# ---------------------------------------------------------------------------
+# Large-k Blackhole routing: ttnn.topk with bf16, dim=-1, largest=True,
+# stable=False and 64 < k <= 2048 is routed at the composite level through
+# ttnn.experimental.topk_large_indices (untilize -> topk_large_indices ->
+# RM gather -> tilize -> slice), bypassing the single-core cliff (the device
+# op's own multi-core path is gated at k <= 64 + pow2 width + width < 65536).
```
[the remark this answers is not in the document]
nkapreTT: Fixed — the banner now describes the current pipeline: values come directly from the op, layout conversions are conditional (TILE-native where supported, untilize/tilize otherwise), and the slice pair only appears when k is not a multiple of 32. The gather stage it referenced was removed when the op's values output landed.

tests/ttnn/nightly/unit_tests/operations/experimental/_topk_li_width_sweep.py
nkapreTT: Done — both bench scripts now use loguru's logger for their completion line.

tests/ttnn/unit_tests/operations/reduce/_topk_route_cells_bench.py
nkapreTT: Fixed — both new bench scripts now carry the standard `© 2026 Tenstorrent USA, Inc.` header.

bbradelTT (approved):

nkapreTT:
**Heads-up @iwroszTT @bbradelTT: this PR's content changed materially after your approvals** — the op owner froze `topk_large_indices`' contract at main's surface (indices-only; no values/TILE/u16/num_slices/sentinel), so the routed composite was rebuilt as a chain and needs your re-review:

`untilize → clamp to lowest-finite-bf16 → op → chunked gather of the ORIGINAL tensor → tilize → u16 typecast where it fits → slice → memory-config conformance`

Notable mechanics: **-inf index parity survives without any op flag** — clamping the searched tensor means the op never sees -inf and stamps a real position for every lane, while values gather from the unclamped tensor so -inf values stay exact (documented tie caveat: genuine lowest-finite-bf16 elements tie with clamped lanes, within the non-stable contract). The gather is chunked ≤1024 wide (the RM multicore gather variant mis-reads above 1920 index columns — verified still live in source). Width envelope is 2^19, not 2^20: the gather stages a full W×2B row-stick in L1 and 2 MB exceeds the budget. **The MoE-gate arm is removed** — with the chain tail it no longer beats stock at gate-sized tensors.

Re-measured on p150a (composite vs stock, 32 rows): k=2048@65536 **477 µs vs 531 ms (1,114×)**; k=512@65536 407 µs vs 135 ms (333×); k=100@8192 54 µs vs 4.2 ms (79×); k=512@100000 578 µs vs 207 ms (358×). Slower than the feature-era numbers (the op runs plain row-parallel for multi-row calls now — pending the owner's call on internal auto-rects), but the cliff-class wins stand. Suites: 382 passed; only the pre-existing #53466 int32 cells fail. Review-landed behaviors kept: memory_config conformance, sharded decline, stable tie-guard, engagement assertion, real -inf positions.

nkapreTT:
Final form pushed (rebased onto #53457 head `c6f18d1dcc8`, which now carries the auto-selecting rectangle engine).

**What changed since the last review round**
- The routed composite's untilize+clamp and gather/assembly steps are now two routing-internal device ops (`topk_route_prep`, `topk_route_finish`) — C++-private, no Python surface, owned by this composite. `ttnn.topk`'s contract is untouched.
- `topk_route_finish` gathers each selected element with a 64 B face-row read from the original TILE tensor, split across both DM RISCs with trid-tagged double-wave pipelining.
- One input-CB wrap fix in the prep compute kernel (`width_tiles % 8 != 0` with multiple blocks per core) plus its trigger test (224×4999).

**Measured on p150a (Tracy device kernel durations, 33 iters)**

| cell (32 rows) | routed | stock | speedup |
|---|---|---|---|
| W=65536, k=2048 | 240.6 µs | 531.4 ms | 2209× |
| W=65536, k=512 | 152.3 µs | 135.3 ms | 888× |
| W=100000, k=512 | 208.8 µs | 206.7 ms | 990× |
| W=8192, k=100 | 49.3 µs | 4.25 ms | 86× |

Suites: 386 passed / 4 failed — the 4 are `test_topk_int32_indices` W=8192 k=50, which fail identically on main (pre-existing, tracked separately with the int32 owners); all routed-path, clamp/-inf-parity, ragged, and cache-invariant tests green.

nkapreTT:
**Rebased onto `main` and retargeted — this PR no longer depends on #53457.**

`topk_large_indices` already exists in main with the identical public contract (`(input, k, valid_length)`, k ≤ 2048 in multiples of 16, width ≤ 2^30), so the routing works against main's op as-is. #53457 then becomes a pure, transparent perf upgrade underneath this routing whenever it lands — no changes needed here.

Re-measured routed vs stock **on main's op internals** (p150a, Tracy device kernel durations, 33 iters):

| cell (rows×W, k) | routed | stock | win |
|---|---|---|---|
| 32×65536, k=2048 | 465.6 µs | 531.5 ms | 1141× |
| 32×65536, k=512 | 333.0 µs | 135.3 ms | 406× |
| 32×100000, k=512 | 486.9 µs | 206.7 ms | 425× |
| 32×8192, k=100 | 69.0 µs | 4.25 ms | 62× |
| 32×5000, k=32 (small-k arm) | 51.4 µs | 708.8 µs | 14× |
| 1×65536, k=2048 (single row) | 382.3 µs | 531.5 ms | 1390× |

Every routed class wins on main's op; the routing predicate is unchanged. With #53457 the same cells improve further (e.g. anchor 465.6 → 240.6 µs).

Suites on this base: 345 passed; the only 2 failures are the IOMMU-gated production perf checks (environment-only, fail identically on plain main on non-IOMMU boxes).

nkapreTT:
Fix for the `sim_bh_p150` sanity break: #53732.

Root cause: the routing sends BH shapes onto `topk_large_indices` — including the *simulated* BH — and the runner's pinned ttsim build crashes silently on those kernels (exactly the three routing-eligible cells died; their non-routing twins passed). The tests pass on BH silicon and on a current ttsim build (verified locally with the routed path engaged). #53732 skips the routed cells on simulator runners only; the durable fix is updating the runner's ttsim artifact, after which the skips revert.

PR #53465, opened 2026-08-18
Sampling call sites: take the routed ttnn.topk form where the route fires

nkapreTT:
### What

**Stacked on #53464** (base = its head; diff is the three sampling files only). Production sampling call sites pass `indices_tensor`, usually `sub_core_grids`, and (the `tt_sampling` family) `stable=True` — each independently disqualifies the Blackhole `topk_large_indices` route. This drops those arguments **only where a call-site mirror of the routing predicate** (`topk_would_route_to_large_indices`, KEEP-IN-SYNC with `should_route_to_topk_large_indices` in topk.cpp) says the route fires; sub-grid-pinned calls and every non-routing shape keep their exact current form, and routed uint32 indices are typecast back to the call-site dtype.

Mirror drift is **fail-safe by construction**: a stale mirror can only miss the optimization (the call keeps its canonical stock form) — it can never change results. One real drift was found and fixed during assembly: `k_multiple` 16 → 32 (the routed pipeline requests TILE output, which rounds k to 32).

### Measured (Blackhole p150a)

- Qwen3 decode sampling chain: 9,590 → 216.9 µs at the call-site level (44.2×); now **128.7 µs** end-to-end with the rectangle routing in #53464 (**74.5×**).
- Split-vocab sampling: 8,924 → 214.9 µs per half (41.5×).

### Bit-exactness audit (performed before any relaxation)

N=20 per shape, integer-bits domain: top-k **values 100% bit-exact** (multiset and sequence); all **27,756 index differences individually proven bitwise ties** (0 non-tie, 0 padding-lane winners); post-tiebreak greedy pick identical 2,560/2,560; the tp8 pow-2 control bit-identical in every field; end-to-end Sampling1D over 20 fixed-seed steps: greedy tokens 639/640 identical, the single diff a proven tied max.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/sampling-routed-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/sampling-routed-topk)

### DIFF
--- a/models/common/sampling/_utils.py
+++ b/models/common/sampling/_utils.py
@@ -38,3 +38,79 @@ def upper_power_of_2(n: int) -> int:
     if n <= 1:
         return 1
     return 1 << (n - 1).bit_length()
+
+
+# ---------------------------------------------------------------------------
+# ttnn.topk large-k routing predicate (call-site mirror).
+#
+# Mirror of should_route_to_topk_large_indices
+# (ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp) as evaluated for a
+# call site that passes NO indices_tensor, NO preallocated outputs, NO
+# sub_core_grids and stable=False, on a 4D tensor reduced over its LAST dim
+# with largest=True (every production sampling call satisfies the shape/dim/
+# largest part already). True iff dropping those arguments makes ttnn.topk take
+# the Blackhole topk_large_indices composite for `x`.
+#
+# KEEP IN SYNC with the C++ gate. Drift is fail-safe in both directions:
+#  - False while C++ would route: the call keeps today's arguments bit-for-bit
+#    (stock path, no change).
+#  - True while C++ would not route: the call runs the stock factory without
+#    indices_tensor/stable; the stock op then generates the same 0-based
+#    positions itself (dtype handled by the caller's typecast restore), and
+#    greedy determinism is owned by _adjust_values_for_tiebreak downstream.
+# ---------------------------------------------------------------------------
+_TOPK_ROUTE_MIN_K_EXCLUSIVE = 64  # large_k_route_min_k_exclusive
+_TOPK_ROUTE_SMALL_K_MIN_PADDED_WIDTH = 4096  # small_k_route_min_padded_width
+_TOPK_ROUTE_MAX_K = 2048  # large_k_route_max_k
+_TOPK_ROUTE_K_MULTIPLE = 32  # large_k_route_k_multiple (k rounds to the TILE-output
+#   multiple since the routed pipeline requests tile_output; was 16 pre-TILE-native)
+_TOPK_ROUTE_MAX_WIDTH = 1 << 20  # large_k_route_max_width
+_TOPK_ROUTE_UINT16_MAX = 65535  # std::numeric_limits<uint16_t>::max()
+_TOPK_ROUTE_MULTI_CORE_MIN_WIDTH = 8192  # ttnn::prim::constants::multi_core_min_width
+# MoE-gate arm (gate_route_* constants in topk.cpp). UNREACHABLE from every
+# production sampling call site — sampling widths are per-device vocab shards
+# (>= 32768 after pow2 padding), far above the 512-wide gate ceiling — kept
+# here only to honor KEEP IN SYNC. Drift on this arm stays fail-safe exactly
+# as described above.
+_TOPK_ROUTE_GATE_MAX_K = 16  # gate_route_max_k
+_TOPK_ROUTE_GATE_MIN_PADDED_W = 128  # gate_route_min_padded_width
+_TOPK_ROUTE_GATE_MAX_PADDED_W = 512  # gate_route_max_padded_width
+_TOPK_ROUTE_GATE_MAX_ROWS = 32  # gate_route_max_flattened_rows
+
+
+def topk_would_route_to_large_indices(x, k, mesh_device) -> bool:
+    """True iff ttnn.topk(x, k, dim=-1) with no indices_tensor / sub_core_grids /
+    preallocated outputs and stable=False routes to the Blackhole
+    topk_large_indices composite (see module comment above)."""
+    import ttnn
+
+    if not ttnn.device.is_blackhole(mesh_device):  # topk.cpp:311
+        return False
+    if k > _TOPK_ROUTE_MAX_K:  # topk.cpp:285
+        return False
+    if x.dtype != ttnn.bfloat16:  # topk.cpp:302
+        return False
+    if x.layout != ttnn.TILE_LAYOUT:  # topk.cpp:305
+        return False
+    if x.memory_config().is_sharded():  # topk.cpp:308
+        return False
+    padded_width = x.padded_shape[-1]
+    if k <= _TOPK_ROUTE_MIN_K_EXCLUSIVE:  # small-k arm (wide + MoE-gate)
+        pow2 = padded_width > 0 and (padded_width & (padded_width - 1)) == 0
+        structurally_ineligible = (
+            padded_width >= _TOPK_ROUTE_UINT16_MAX or not pow2 or padded_width < _TOPK_ROUTE_MULTI_CORE_MIN_WIDTH
+        )
+        wide_arm = structurally_ineligible and padded_width >= _TOPK_ROUTE_SMALL_K_MIN_PADDED_WIDTH
+        rows = 1
+        for d in tuple(x.shape)[:-1]:
+            rows *= d
+        gate_arm = (
+            k <= _TOPK_ROUTE_GATE_MAX_K
+            and _TOPK_ROUTE_GATE_MIN_PADDED_W <= padded_width <= _TOPK_ROUTE_GATE_MAX_PADDED_W
+            and rows <= _TOPK_ROUTE_GATE_MAX_ROWS
+        )
+        if not (wide_arm or gate_arm):
+            return False
+    width = x.shape[-1]
+    k_rounded = ((k + _TOPK_ROUTE_K_MULTIPLE - 1) // _TOPK_ROUTE_K_MULTIPLE) * _TOPK_ROUTE_K_MULTIPLE
+    return k_rounded <= width <= _TOPK_ROUTE_MAX_WIDTH  # width envelope + k_rounded fit
--- a/models/common/sampling/_utils.py
+++ b/models/common/sampling/_utils.py
@@ -62,20 +62,12 @@ def upper_power_of_2(n: int) -> int:
 _TOPK_ROUTE_MIN_K_EXCLUSIVE = 64  # large_k_route_min_k_exclusive
 _TOPK_ROUTE_SMALL_K_MIN_PADDED_WIDTH = 4096  # small_k_route_min_padded_width
 _TOPK_ROUTE_MAX_K = 2048  # large_k_route_max_k
-_TOPK_ROUTE_K_MULTIPLE = 32  # large_k_route_k_multiple (k rounds to the TILE-output
-#   multiple since the routed pipeline requests tile_output; was 16 pre-TILE-native)
-_TOPK_ROUTE_MAX_WIDTH = 1 << 20  # large_k_route_max_width
+_TOPK_ROUTE_K_MULTIPLE = 16  # large_k_route_k_multiple (as merged in #53464)
+_TOPK_ROUTE_MAX_WIDTH = 1 << 19  # large_k_route_max_width (as merged in #53464)
 _TOPK_ROUTE_UINT16_MAX = 65535  # std::numeric_limits<uint16_t>::max()
 _TOPK_ROUTE_MULTI_CORE_MIN_WIDTH = 8192  # ttnn::prim::constants::multi_core_min_width
-# MoE-gate arm (gate_route_* constants in topk.cpp). UNREACHABLE from every
-# production sampling call site — sampling widths are per-device vocab shards
-# (>= 32768 after pow2 padding), far above the 512-wide gate ceiling — kept
-# here only to honor KEEP IN SYNC. Drift on this arm stays fail-safe exactly
-# as described above.
-_TOPK_ROUTE_GATE_MAX_K = 16  # gate_route_max_k
-_TOPK_ROUTE_GATE_MIN_PADDED_W = 128  # gate_route_min_padded_width
-_TOPK_ROUTE_GATE_MAX_PADDED_W = 512  # gate_route_max_padded_width
-_TOPK_ROUTE_GATE_MAX_ROWS = 32  # gate_route_max_flattened_rows
+# NOTE: merged topk.cpp has no MoE-gate arm; if one lands later, mirror it here
+# in the same PR (KEEP IN SYNC).
 
 
 def topk_would_route_to_large_indices(x, k, mesh_device) -> bool:
@@ -95,21 +87,13 @@ def topk_would_route_to_large_indices(x, k, mesh_device) -> bool:
     if x.memory_config().is_sharded():  # topk.cpp:308
         return False
     padded_width = x.padded_shape[-1]
-    if k <= _TOPK_ROUTE_MIN_K_EXCLUSIVE:  # small-k arm (wide + MoE-gate)
+    if k <= _TOPK_ROUTE_MIN_K_EXCLUSIVE:  # small-k arm (wide structurally-ineligible)
         pow2 = padded_width > 0 and (padded_width & (padded_width - 1)) == 0
         structurally_ineligible = (
             padded_width >= _TOPK_ROUTE_UINT16_MAX or not pow2 or padded_width < _TOPK_ROUTE_MULTI_CORE_MIN_WIDTH
         )
         wide_arm = structurally_ineligible and padded_width >= _TOPK_ROUTE_SMALL_K_MIN_PADDED_WIDTH
-        rows = 1
-        for d in tuple(x.shape)[:-1]:
-            rows *= d
-        gate_arm = (
-            k <= _TOPK_ROUTE_GATE_MAX_K
-            and _TOPK_ROUTE_GATE_MIN_PADDED_W <= padded_width <= _TOPK_ROUTE_GATE_MAX_PADDED_W
-            and rows <= _TOPK_ROUTE_GATE_MAX_ROWS
-        )
-        if not (wide_arm or gate_arm):
+        if not wide_arm:
             return False
     width = x.shape[-1]
     k_rounded = ((k + _TOPK_ROUTE_K_MULTIPLE - 1) // _TOPK_ROUTE_K_MULTIPLE) * _TOPK_ROUTE_K_MULTIPLE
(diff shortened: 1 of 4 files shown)


yieldthought:
Codex review:

1. **[P1] Keep generated indices in the single-device fallback** — [models/common/sampling/tt_sampling.py:807](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/sampling/tt_sampling.py#L807)

On a 1×1 mesh, `_get_num_sampling_shards()` returns 2, so `self.tt_indices_tensor` is already V/2 wide. Splitting it again makes each `indices_tensor_list[i]` V/4 wide while each `x_bf16_list[i]` is V/2. Whenever routing is disabled—for example, the current regression test supplies `sub_core_grid_topk`—the call at [models/common/sampling/tt_sampling.py:857](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/sampling/tt_sampling.py#L857) now fails the equal-shape check at [ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp:590](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp#L590). This directly reintroduces #53001, whose fix deliberately removed the redundant iota buffer; [models/common/tests/test_sampling.py:819](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/tests/test_sampling.py#L819) should catch it. Suggestion: keep the base branch's generated-index fallback and only omit `stable` on the routed branch.

2. **[P1] Preserve 32-bit local indices above the uint16 boundary** — [models/common/modules/sampling/sampling_1d.py:831](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/modules/sampling/sampling_1d.py#L831)

The new local-index buffer is unconditionally UINT16, and routed results are narrowed back to that dtype at [models/common/modules/sampling/sampling_1d.py:552](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/modules/sampling/sampling_1d.py#L552) and [models/common/modules/sampling/sampling_1d.py:610](https://github.com/tenstorrent/tt-metal/blob/73ca135e2dd513a0a288ca7816c7734aa88ae06b/models/common/modules/sampling/sampling_1d.py#L610). `Sampling1D` is enabled for 1×1/1×2 Qwen2/Qwen2.5; a 152064-token vocabulary gives 76032 local entries, so positions 65536–76031 cannot be represented and alias before the global offset is added. The sampler can therefore return a different token whenever a winning candidate lies in that upper region. Suggestion: retain generated indices as the current base does, or select and preserve UINT32 whenever the local shard/half width exceeds the UINT16 range, including through the gather.

nkapreTT:
Verified both — **both P1s are real**, and they share one root cause: this PR's file bodies were assembled from a branch that predates #53001 (merged the same morning), so it resurrects the iota `indices_tensor` machinery that #53001 deliberately removed (P1), and adds a UINT16 local-index buffer + narrowing casts that alias positions ≥65536 on Qwen2.5's 76032-wide local shards (P1). Our local validation didn't catch either because the sampling device tests (including the #53001 regression at test_sampling.py:819) are conftest-skipped on Blackhole hardware.

Fix incoming: the PR will be rebuilt as origin/main's file bodies plus only the minimal routed-relaxation deltas (the route gate + dropping the routing-blocking args on the routed branch + dtype normalization sized to the shard width, never an unconditional u16). Will ping when pushed. Thanks — sharp catches, especially the aliasing bound.

nkapreTT:
Rebuilt and pushed (df5687d). Both P1s are gone by construction, not by patching:

- The PR is now **current main's file bodies + a minimal delta**: `models/common/sampling/_utils.py` (the routing-predicate mirror, new file) and two `stable=False if use_routed_topk else self._topk_stable` gates in `TTSampling` — 105 lines total, down from 346.
- **P1-1**: no `tt_indices_tensor`, no splits, no `indices_tensor=` anywhere — the #53001 world is preserved verbatim; the non-routed branch is bit-for-bit today's call.
- **P1-2**: `Sampling1D` is untouched entirely (zero diff vs main) — its call sites pass only `sub_core_grids`, which already routes when None; no UINT16 buffers or narrowing casts exist. Index dtype is whatever the engine emits, and the routed path's u16/u32 contract matches the stock op's (u16 iff the padded row fits 16 bits), so nothing narrows above the boundary.
- Root cause for the record: the previous revision resurrected pre-#53001 file bodies from the development branch; the local test pass was misleading because the sampling device tests (including the #53001 regression at test_sampling.py:819) are conftest-skipped on Blackhole hardware.

Only remaining relaxation semantics: `stable` is dropped exactly where the route fires (it's best-effort/broken per #33492, and `_adjust_values_for_tiebreak` owns the greedy-pick guarantee — the I5 device study verified values bit-exact and all 27,756 index diffs proven ties).

nkapreTT:
Rebased onto main now that #53464 merged — the diff is down to 3 sampling files (two `stable=False if use_routed_topk else self._topk_stable` gates + the `_utils.py` mirror; `sampling_1d.py` is zero-diff vs main). Checks green.

@mtairum you're in both pending CODEOWNERS groups, so a single approval from you unblocks this one. Evidence for the `stable` relaxation is the I5 bitwise tie study linked in the PR description.

sraizada-tt (approved):

yieldthought:
Codex review:

1. **[P2] Mirror the predicate that actually merged** — [models/common/sampling/_utils.py:75](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/models/common/sampling/_utils.py#L75)

This helper is already a false positive for valid inputs. At this PR head, the C++ predicate has no `gate_route_*` arm: for `k <= 64` it routes only the structurally ineligible wide arm with padded width at least 4096 ([ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp:297](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp#L297)), and its constants are still a 16-wide `k` multiple and a `1 << 19` width ceiling ([ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp:254](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp#L254) and [ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp:259](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp#L259)). The Python copy instead uses 32, `1 << 20`, and the extra 128–512 MoE arm. For example, a bf16 TILE tensor with width 128/256/512 and `k=16` makes `topk_would_route_to_large_indices()` return true even though C++ keeps the stock path. Both callers then pass `stable=False` ([models/common/sampling/tt_sampling.py:876](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/models/common/sampling/tt_sampling.py#L876) and [models/common/sampling/tt_sampling.py:921](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/models/common/sampling/tt_sampling.py#L921)), getting no routed-path performance benefit while relaxing the stock call's tie semantics. The stated “fail-safe” argument does not cover stochastic users: `_adjust_values_for_tiebreak` explicitly leaves `k > 1` rows unchanged ([models/common/sampling/tt_sampling.py:665](https://github.com/tenstorrent/tt-metal/blob/b19b4cfa0dc5f940b016ec0c2d3687f37d4e7436/models/common/sampling/tt_sampling.py#L665)). Suggestion: either restore the intended C++ predicate in the prerequisite/current PR, or make this helper match the predicate that actually merged and add positive/negative boundary parity tests so a false positive cannot silently disable `stable`.

nkapreTT:
Verified — the drift is real and your read of the merged predicate is exactly right: #53464 landed with `k_multiple=16`, a `1<<19` width ceiling, and no gate arm; the mirror was written against a pre-merge revision (same failure family as the earlier round — dev-branch bodies vs merged bodies). Pushed ed6a7aa35bc: mirror now matches merged `topk.cpp` (gate arm removed, 16, `1<<19`) plus `test_topk_route_mirror_parity` — five boundary cells each distinguishing a merged constant from its stale value (your 128/512 k=16 cells, a k=100→112-vs-128 rounding cell, and the width ceiling), verified 5/5 on a merged-main device build, so a re-drift can't silently drop `stable` again.

On blast radius, for the record: no production sampling call reaches the false-positive regions (both call sites pass `k = max_top_k`, typically 32 > 16, on vocab shard/chunk widths ≫512 and ≪2^19; the k-multiple drift errs fail-safe), and the stochastic-user exposure is bounded by `stable=True` being best-effort already (#33492) with `ttnn.sampling`'s own tie-break array-position-dependent — the I5 study's 27,756 routed-vs-stock index diffs were all bitwise ties with values bit-exact. Agreed the mirror must track what merged; if a gate arm or TILE-native rounding lands later, the mirror gets re-synced in that PR and the parity test updated with it.

nkapreTT:
Thanks — all four are fair reads; triaged:

**k-rounding (C3):** the code is the intended state — the mirror carries the *merged* #53464 constants (`k_multiple=16`, `1<<19` ceiling, no gate arm), verified 5/5 on-device by the parity test whose `k=100` row is precisely the 16-vs-32 drift detector. The description's "16 → 32" sentence described an earlier assembly round against a pre-merge #53464 revision — description now updated.

**False-positive drift (C2):** the premises are right (`_adjust_values_for_tiebreak` is a k>1 no-op; the parity test freezes constants), but the harm case doesn't exist: stock `stable=True` never delivered tie-order determinism to stochastic users, because `ttnn.sampling`'s internal re-sort is itself tie-unstable (and topk stable is best-effort per #33492). A stale-True mirror therefore changes nothing that was ever guaranteed; greedy determinism is owned by the tiebreak pass either way. Agreed the KEEP-IN-SYNC discipline is one-directional today — a reciprocal breadcrumb next to `should_route_to_topk_large_indices` goes in the follow-up (it can't ride this PR's sampling-only diff).

**CI selection + routed e2e case (C1, C4):** real coverage gaps, taking both — as an immediate **follow-up PR** rather than a push here (this diff stays frozen so the pending re-review applies to exactly what was reviewed): a `-k` term for the parity test with arch-aware expectations (`expected and is_blackhole`, which also buys WH False-coverage), and a BH-gated TTSampling case with `sub_core_grid_topk=None` at a routing-eligible chunk width. The routed op itself already has op-level CI via the sweep-framework large_k suite, and the e2e path was validated on-device (the Qwen3 numbers in the description).

mtairum:
### Code review

Found 1 issue:

1. `test_topk_route_mirror_parity` has no arch gate, so its two `expected=True` cells fail on Wormhole. The predicate short-circuits to `False` off Blackhole, so on a Wormhole runner every cell evaluates `False` and the `64128/k=32` and `112/k=100` rows assert `False is True`.

https://github.com/tenstorrent/tt-metal/blob/ed6a7aa35bc8a7225d37c587fc2791dfe0500d79/models/common/tests/test_sampling.py#L1136-L1146

The early return that makes this arch-dependent:

https://github.com/tenstorrent/tt-metal/blob/ed6a7aa35bc8a7225d37c587fc2791dfe0500d79/models/common/sampling/_utils.py#L78-L80

This is not only the deselected-by-`-k` case already noted on [models_unit_tests.yaml](https://github.com/tenstorrent/tt-metal/blob/ed6a7aa35bc8a7225d37c587fc2791dfe0500d79/tests/pipeline_reorg/models_unit_tests.yaml#L1393-L1395): the registered `t3k_tttv2_fast_unit_tests` job collects `models/common/tests` with **no** `-k` filter, so the new test is picked up there and runs on `wh_llmbox`.

https://github.com/tenstorrent/tt-metal/blob/ed6a7aa35bc8a7225d37c587fc2791dfe0500d79/tests/scripts/t3000/run_t3000_unit_tests.sh#L321-L323

https://github.com/tenstorrent/tt-metal/blob/ed6a7aa35bc8a7225d37c587fc2791dfe0500d79/tests/pipeline_reorg/t3k_unit_tests.yaml#L73-L78

Unlike the neighbouring tests in this file, the new test also does not pin `mesh_device`, so it takes the full system mesh on T3K. Gating with `skip_for_wormhole_b0` (`models/common/utility_functions.py:1080`), or making the expectation arch-aware (`expected and is_blackhole`), would keep the drift detector meaningful on both architectures.

Everything else checked out: the mirror's constants and small-k wide arm match the merged `should_route_to_topk_large_indices` exactly (`k_multiple=16`, `1<<19` ceiling, no gate arm, `logical` vs `padded` shape used in the same places), and both relaxed call sites genuinely satisfy the docstring's precondition — neither passes `indices_tensor` or preallocated outputs, and `sub_core_grids` is non-`None` only when the `self.sub_core_grid_topk is None` guard has already forced `use_routed_topk` false.

Two smaller things, below the bar to list as issues but worth a follow-up: the five `# topk.cpp:NNN` citations in `_utils.py` are all off (the real checks are at 294/317/320/323/326), which blunts the KEEP-IN-SYNC audit they exist to enable; and the PR description still says the `k_multiple` drift was "16 → 32" when the merged constant is 16 and the code correctly uses 16.

nkapreTT:
Good catch — fixed in 8a064053a7a: the expectations are now arch-aware (`expected and is_blackhole(mesh_device)`), asserted rather than skipped off-BH so the same cells double as "off-Blackhole never routes" coverage on WH runners. Verified 5/5 on a BH device. (This was also Copilot's C1; its sibling C4 — an e2e routed TTSampling case — lands in the follow-up PR noted above so this diff stays minimal.)

mtairum:
### Code review

Re-reviewed at `8a06405`. The arch gate I flagged last round is correctly fixed — `expected_here = expected and is_blackhole(...)` keeps full discriminating power on Blackhole (verified all 5 cells by hand against the merged constants) and collapses to a valid never-routes assertion off it. `_utils.py` and `tt_sampling.py` are byte-identical to the previously reviewed head.

Found 1 issue:

1. The drift detector has no effective coverage in any CI job, so it cannot actually catch a re-drift. Off Blackhole the predicate short-circuits at its first check, so every cell asserts `False is False` and passes even if `_TOPK_ROUTE_K_MULTIPLE` went back to 32 or `_TOPK_ROUTE_MAX_WIDTH` back to `1<<20`.

https://github.com/tenstorrent/tt-metal/blob/8a064053a7af9d5767f8c3d92b1d73fe6a62e953/models/common/tests/test_sampling.py#L1160-L1166

The two jobs that reach this file are disjoint from the one configuration that would give the test teeth:

- `models_unit_tests.yaml` runs on `bh_p150` **and** `wh_n150`, but its `-k` filter selects none of `test_topk_route_mirror_parity` — so the Blackhole runner never collects it. This is the already-deferred selection gap, but combined with the new arch-aware expectation it is now what removes the last path to real coverage.

https://github.com/tenstorrent/tt-metal/blob/8a064053a7af9d5767f8c3d92b1d73fe6a62e953/tests/pipeline_reorg/models_unit_tests.yaml#L1393-L1395

- `t3k_tttv2_fast_unit_tests` collects the directory unfiltered, but only on `wh_llmbox` — where the test is now vacuous.

https://github.com/tenstorrent/tt-metal/blob/8a064053a7af9d5767f8c3d92b1d73fe6a62e953/tests/pipeline_reorg/t3k_unit_tests.yaml#L73-L79

Net: no job both selects this test and runs it on Blackhole. Worth landing the `-k` term in this PR rather than the follow-up, since without it the test is inert. `run_for_blackhole()` (`models/common/utility_functions.py:1092`) would also make the no-coverage state visible as a skip instead of a green pass.

Two things below the bar to list, but worth tightening while you are here:

- The new comment blocks say `_adjust_values_for_tiebreak` guarantees the greedy pick "regardless of per-device tie order" — that phrase is new in this PR, and the function's own KNOWN LIMITATION (`tt_sampling.py:681-684`) says the opposite when one shard holds more than `max_top_k` tied maxima: `stable` governs which candidates survive the local truncation, and the pass cannot recover one that never reached the gathered set. Your own audit's 639/640 greedy result is that case. Suggest scoping it to "among the gathered candidates (see KNOWN LIMITATION)".
- The test does not pin `mesh_device`, unlike every neighbouring test, so it takes the full system mesh and replicates the `1<<20` cell (~67MB) to every chip for what is a pure-metadata check.

Still clean on re-check: the mirror matches `should_route_to_topk_large_indices` on current `origin/main` (constants and both arms), and the routed index dtype is a non-issue — routed and stock share the same u16/u32 width boundary, and `tt_sampling.py:966` typecasts to `int32` before any offset arithmetic on both branches. The stale `# topk.cpp:NNN` citations (real: 294/317/320/323/326) and the PR body's `k_multiple` "16 → 32" sentence are both still unfixed from last round.

nkapreTT:
Fixed in 4dfd2c79a18 — added `topk_route_mirror_parity` to the sampling job's `-k` filter, so the `bh_p150` runner now collects the 5 discriminating cells (verified by collect-only: 50 items / 45 deselected / 5 selected; the WH side of the same job asserts the never-routes collapse). With that, the detector has real teeth on exactly the runner where the constants matter.

One process note: the yaml touch adds `tests/pipeline_reorg/` owners to the required-review set — requesting them now so it doesn't stall.

mtairum (approved):
Approving to unblock, but please run the `On-device sampling unit tests` on tier 3 model unit tests: https://github.com/tenstorrent/tt-metal/actions/workflows/models-t3-unit-tests.yaml

nkapreTT:
Dispatched per the approval note: tier-3 On-device sampling unit tests (`model=tt-model-sampling`, `sku=all`) on this branch — https://github.com/tenstorrent/tt-metal/actions/runs/32389342523. This run also exercises the new parity cells on both arches (discriminating on `bh_p150`, never-routes collapse on `wh_n150`).

nkapreTT:
Tier-3 On-device sampling unit tests: **green** on both SKUs (run 32389342523) — the parity cells exercised on `bh_p150` (discriminating) and `wh_n150` (never-routes). Enqueued.

PR #53474, opened 2026-08-18
[Bug fix] concat failure with Out of Memory issue in L1

### LINKED ISSUE #21358: ttnn.concat fails with Out of Memory issue in L1
### Describe the bug

`ttnn.concat` fails with Out of Memory issue in L1 and passed with DRAM. The issue cannot be reproduced in stand alone unit test for `ttnn.concat`, Hence, created using the model's sub-module test.

1. checkout the branch: [vignesh/yolov10x_concat_issue](https://github.com/tenstorrent/tt-metal/tree/vignesh/yolov10x_concat_issue)
2. Run the following command:  `pytest tests/ttnn/integration_tests/yolov10/test_ttnn_yolov10.py::test_yolov10x_C2f`

### Expected behavior

The test is expected to PASS with the ttnn.concat in L1 memory config. But now it passes only for DRAM memory config.

### Please complete the following environment information

- OS version - Ubuntu 20.04
- Hardware - n150, n300

abdulla-TT:
### PR Category

Bug fix

### Summary

Fixes #21358

**Issue - Root Cause**

`ttnn.concat` on tile-layout tensors with a non-tile-aligned last dim (e.g. width 80, padded to 96) routed through a massaged pipeline: untilize every input to row-major -> transpose -> concat -> transpose back -> retilize. With L1 inputs and an L1 output this failed in two ways:

1. **Circular buffer collision** — `tilize_with_val_padding` sized its circular buffers from the L1 space it believed was free, but the check ran before concat's own output buffer was allocated.
2. **Intermediate blow-up** — the pipeline materializes several full-size copies of the tensor (row-major copy, transposed copy, retilize input) in L1 at once. For shapes like (25600, 100) x 5.

**Fix**

1. **L1 accounting** — new `get_pending_l1_output_reservation()` in `common.cpp` computes the per-core L1 footprint of the op's not-yet-allocated output buffer; `is_enough_space()` now subtracts it. Wired into `tilize`, `tilize_with_val_padding`, `untilize`, and `untilize_with_unpadding`, so factory selection can no longer pick CB layouts that collide with the pending output.
2. **Native unaligned-width concat** — new `ConcatTiledUnalignedProgramFactory` handles last-dim concat of TILE interleaved tensors with tile padding on the concat dim entirely in L1.

### Notes for reviewers

- `sources.cmake`: the new .cpp is deliberately appended last rather than grouped with the other concat sources — inserting it mid-list reshuffles unity-build batches and surfaces duplicate-symbol clashes in unrelated ops (see comment in the file).
- The eligibility check is deterministic (exact CB byte count vs. unreserved L1 minus L1-resident inputs + pending output), so the native path never OOMs at runtime; ineligible cases keep today's behavior.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/concat_L1_mem_issue)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/concat_L1_mem_issue)


Aswinmcw (commented):
One correctness concern below (program-cache reuse vs. the new L1-dependent factory selection); the other two are nits.

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_device_operation.cpp
```diff
@@ -26,6 +26,12 @@ ConcatDeviceOperation::program_factory_t ConcatDeviceOperation::select_program_f
     const auto& input_tensors = tensor_args.input_tensors;
 
     if (const bool input_is_sharded = input_tensors[0].is_sharded(); !input_is_sharded) {
+        // The launch infra allocates the output tensor before factory selection, so the
+        // allocator's free window already accounts for it.
+        if (can_use_tiled_unaligned_concat(
```
Aswinmcw, line 31: The eligibility check reads live L1 occupancy (`get_max_l1_space`), so factory selection is now a function of runtime allocator state — but that state is not in the program hash. `ConcatDeviceOperation` has no `compute_program_hash` override, so it gets the default `hash(type_hash, attributes, tensor_args)`, and on a cache hit `handle_mesh_adapter_cache_hit` replays the *stored* `program_factory_index` (`device_operation.hpp`) rather than re-running this function.

Two identical-spec concats with different L1 occupancy therefore share a cache entry:

* first call tight, later call free → the slow massaged path is pinned forever, silently;
* first call free, later call tight → the cached native program is replayed with the CB addresses baked in at build time, into a window an unrelated live tensor now occupies. That is the same CB/buffer collision this PR is fixing, just moved behind the cache.

So the "Notes for reviewers" claim that the check is deterministic and "the native path never OOMs at runtime" holds only for the cache-miss call. Either fold the budget decision into a `compute_program_hash` override, or make eligibility depend on shape/dtype only and handle the L1 shortfall inside the factory (e.g. clamp bands per core) so it cannot vary per call.

`test_tiled_concat_unaligned_width_program_cache` is reaching for this — it keeps `live_tensors` alive across iterations — but at `[64,100]+[64,60]` the budget can never flip, so it passes either way.

ttnn/cpp/ttnn/operations/data_movement/common/common.cpp
```diff
@@ -437,14 +437,66 @@ uint32_t get_max_l1_space(const Tensor& input_tensor_a) {
...
+
+    size_t total_bytes = 0;
+    size_t page_bytes = 0;
+    try {
+        const tt::tt_metal::TensorSpec output_spec(
+            output_padded_shape,
+            tt::tt_metal::TensorLayout(output_dtype, tt::tt_metal::PageConfig(output_layout), output_memory_config));
+        total_bytes = output_spec.compute_packed_buffer_size_bytes();
+        page_bytes = output_spec.compute_page_size_bytes();
+    } catch (...) {
```
Aswinmcw, line 472: `catch (...) { return 0; }` makes an unconstructible `TensorSpec` reserve nothing, which is exactly the unreserved behaviour that causes the OOM this PR fixes. For the `tilize`/`untilize` callers that is at least status quo, but `can_use_tiled_unaligned_concat` uses the return value to decide eligibility, so a swallowed throw makes it *more* permissive and routes a concat whose output does not fit. Worth letting it propagate (or `TT_FATAL`) so the condition is visible rather than silently downgraded.

ttnn/cpp/ttnn/operations/data_movement/concat/device/kernels/compute/concat_untilize_retilize.cpp
```diff
@@ -0,0 +1,42 @@
...
+                out_wt,
+                cb_asm,
+                cb_out,
+                compute_kernel_lib::tilize_config::InitUninitMode::InitAndUninit,
+                compute_kernel_lib::tilize_config::WaitMode::WaitBlock,
+                compute_kernel_lib::tilize_config::ReconfigureRegisterDatatypeMode::UnpackAndPackReconfigure,
+                compute_kernel_lib::tilize_config::Fp32Mode::Lossless>(1);
+        } else {
+            compute_kernel_lib::tilize<out_wt, cb_asm, cb_out>(1);
+        }
```
Aswinmcw: The three spelled-out template args in the `fp32_lossless` arm are already the helper defaults (`tilize_helpers.hpp`: `InitAndUninit`, `WaitBlock`, `UnpackAndPackReconfigure`), so the two arms differ only in `Fp32Mode`. Collapsing to one call makes that the only visible difference and keeps the arms from drifting if the defaults change:

```cpp
compute_kernel_lib::tilize<
    out_wt,
    cb_asm,
    cb_out,
    compute_kernel_lib::tilize_config::InitUninitMode::InitAndUninit,
    compute_kernel_lib::tilize_config::WaitMode::WaitBlock,
    compute_kernel_lib::tilize_config::ReconfigureRegisterDatatypeMode::UnpackAndPackReconfigure,
    fp32_lossless ? compute_kernel_lib::tilize_config::Fp32Mode::Lossless
                  : compute_kernel_lib::tilize_config::Fp32Mode::Fast>(1);
```

Aswinmcw (approved):
LGTM.

VirdhatchaniKN (approved):
LGTM

bbradelTT (approved):

roseli-TT (approved):

sjameelTT (approved):

tests/ttnn/unit_tests/operations/data_movement/test_concat.py
```diff
@@ -48,6 +48,64 @@ def test_tiled_concat(device, concat_spec, dtype):
...
+        [[50, 100], [50, 100]],  # height not tile-aligned either
+        [[2, 3, 64, 100], [2, 3, 64, 100]],  # rank 4
+        [[1, 7, 100], [1, 7, 100]],  # rank 3, tiny height
+        [[2500, 100]] * 5,  # tall, many inputs, uneven core split
+    ],
+)
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_tiled_concat_unaligned_width(device, shapes, dtype):
+    # L1 memory config only: the native path is buffer-location agnostic, and the DRAM
+    # variant is covered by the issue-21358 regression test.
```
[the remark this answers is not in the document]
abdulla-TT: Updated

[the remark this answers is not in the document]
abdulla-TT: Updated

PR #53481, opened 2026-08-18
[Bug fix] ttnn: derive default mesh composer from tensor topology in to_torch

ctxbot:
## Problem
`Tensor.to_torch()` / `to_numpy()` with `mesh_composer=None` on a tensor distributed on `MeshShape(1, 2)` hits `TT_FATAL` at pytensor.cpp:295:

```
Can't convert a tensor distributed on MeshShape([1, 2]) mesh to row-major logical tensor. Supply a mesh composer to concatenate multi-device shards.
```

## Fix
When every multi-device distribution axis is a Shard placement, derive a concat `MeshToTensor` from the tensor's recorded topology and compose implicitly. Replicated or unrecorded distributions still raise.

## Test
`tests/ttnn/distributed/test_default_mesh_composer.py`. Replicated distribution still raises the original error. A 1-D shard on `MeshShape(1, 2)` host + device roundtrip is the named closer.

## Evidence
- `i-mesh-composer.log` — `I_MESH_COMPOSER_ORIGINAL_FATAL: MeshShape(1,2) dim=3 to_torch(no composer) VERDICT: REPRODUCED`; then `I_MESH_COMPOSER: MeshShape(1,2) dim=3 shape=(1,1,64,128) bf16 RM+TILE host+device sha=f9434d5a0f0 VERDICT: PASS`
- `i-mesh-composer-verify.log` — remesh at this head: `I_MESH_COMPOSER: MeshShape(1,2) dim=3 shape=(1,1,64,128) bf16 RM+TILE host+device sha=a04e6ddf769 VERDICT: PASS`

### DIFF
--- a/tests/ttnn/distributed/test_default_mesh_composer.py
+++ b/tests/ttnn/distributed/test_default_mesh_composer.py
@@ -1,0 +1,64 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""to_torch()/to_numpy() without an explicit mesh_composer on a multi-device tensor.
+
+Repro of record: TT_FATAL at pytensor.cpp ("Can't convert a tensor distributed on
+MeshShape([1, 2]) mesh to row-major logical tensor. Supply a mesh composer to
+concatenate multi-device shards.") fired from Tensor.to_torch() with
+mesh_composer=None.
+
+Part (a): a replicated distribution still raises the original error verbatim --
+the fallback must not silently mask distributions it cannot derive.
+
+Part (b): the same site/mesh/layout with a 1-D shard distribution (the exact
+case the error message names: "concatenate multi-device shards") now derives a
+default composer from the tensor's recorded topology, both host-side and after a
+full device write/readback round trip on MeshShape(1, 2).
+"""
+
+import pytest
+import torch
+import ttnn
+
+MESH_SHAPE = (1, 2)  # P300: 2x p150a
+SHARD_DIM = 3
+SHAPE = (1, 1, 64, 128)  # -> per-device shard (1, 1, 64, 64)
+
+
+def test_sharded_to_torch_without_composer_on_1x2_mesh():
+    device = ttnn.open_mesh_device(ttnn.MeshShape(*MESH_SHAPE))
+    try:
+        torch.manual_seed(0)
+        torch_input = torch.randn(*SHAPE, dtype=torch.float32)
+        torch_bf16 = torch_input.to(torch.bfloat16)
+
+        # (a) Original failure mode preserved: replicated distribution, no composer.
+        replicated = ttnn.from_torch(torch_bf16, mesh_mapper=ttnn.ReplicateTensorToMesh(device))
+        with pytest.raises(RuntimeError, match="Supply a mesh composer"):
+            replicated.to_torch()
+
+        # (b) Fixed path: 1-D sharded distribution, no composer.
+        mapper = ttnn.shard_tensor_to_mesh_mapper(device, dim=SHARD_DIM)
+        sharded = ttnn.from_torch(torch_bf16, mesh_mapper=mapper)
+
+        shards = ttnn.get_device_tensors(sharded)
+        assert len(shards) == 2
+        expected = torch.cat([s.to_torch() for s in shards], dim=SHARD_DIM)
+
+        composed = sharded.to_torch()
+        assert tuple(composed.shape) == SHAPE
+        assert torch.equal(composed, expected)
+        assert torch.equal(composed, torch_bf16)
+
+        # Same site/layout through the device: TILE layout, to_device + from_device.
+        sharded_tile = ttnn.from_torch(
+            torch_bf16, layout=ttnn.TILE_LAYOUT, mesh_mapper=ttnn.shard_tensor_to_mesh_mapper(device, dim=SHARD_DIM)
+        )
+        read_back = ttnn.from_device(ttnn.to_device(sharded_tile, device))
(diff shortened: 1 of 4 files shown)


tests/ttnn/distributed/test_default_mesh_composer.py
```diff
@@ -0,0 +1,64 @@
...
+import pytest
+import torch
+import ttnn
+
+MESH_SHAPE = (1, 2)  # P300: 2x p150a
+SHARD_DIM = 3
+SHAPE = (1, 1, 64, 128)  # -> per-device shard (1, 1, 64, 64)
+
+
+def test_sharded_to_torch_without_composer_on_1x2_mesh():
```
vtsilytskyiTT, line 30:
```suggestion
@pytest.mark.parametrize("mesh_device", [MESH_SHAPE], indirect=True)
def test_sharded_to_torch_without_composer_on_1x2_mesh(mesh_device):
```
What do you think?

ttnn/cpp/ttnn-nanobind/pytensor.cpp
```diff
@@ -223,6 +223,53 @@ RowMajorHostBuffer convert_block_float_to_logical_row_major(const HostBuffer& bu
...
+        return std::nullopt;
+    }
+    ttsl::SmallVector<int> dims;
+    dims.reserve(placements.size());
+    for (size_t axis = 0; axis < placements.size(); ++axis) {
+        if (const auto* shard = std::get_if<MeshMapperConfig::Shard>(&placements[axis])) {
+            if (shard->dim < 0 || shard->dim >= tensor_rank) {
+                return std::nullopt;
+            }
+            dims.push_back(shard->dim);
```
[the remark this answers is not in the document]
vtsilytskyiTT: Agree with copilot. Must fix.

ttnn/cpp/ttnn-nanobind/pytensor.cpp
```diff
@@ -223,6 +223,53 @@ RowMajorHostBuffer convert_block_float_to_logical_row_major(const HostBuffer& bu
...
+            dims.push_back(dims.empty() ? 0 : dims.front());
+        } else {
+            // A replicated axis spanning multiple devices cannot be expressed as a concatenation.
+            return std::nullopt;
+        }
+    }
+
+    const auto composer = ttnn::distributed::MeshToTensor::create(
+        mesh_shape, MeshComposerConfig{.dims = std::move(dims), .mesh_shape_override = distribution_shape});
+    return convert_to_row_major_host_buffer(tt_tensor, composer);
```
vtsilytskyiTT, line 270: This recreates a composer from:

1. physical mesh shape,
2. distribution shape,
3. concatenation dimensions.

It does not use `topology.mesh_coords()`.

For an offset submesh or custom coordinate mapping, it therefore reads shards from the origin/default row-major mapping rather than their recorded coordinates. The topology contains the necessary information, but the reconstruction discards it.

ttnn/cpp/ttnn-nanobind/pytensor.cpp
```diff
@@ -292,6 +339,12 @@ RowMajorHostBuffer convert_to_row_major_host_buffer(const Tensor& tt_tensor, con
     const auto& storage = tt_tensor.host_storage();
     std::vector<HostBuffer> buffers;
     storage.buffer().apply([&buffers](const HostBuffer& shard) { buffers.push_back(shard); });
+    if (buffers.size() > 1) {
```
vtsilytskyiTT, line 342: buffers contains locally populated shards. On a multi-host mesh, one host can have one local shard while the global topology spans many devices. The condition is false, and `to_torch()` returns that local shard as the complete tensor.

The decision must use global topology:

```cpp
topology.distribution_shape().mesh_size() > 1
```
`MeshToTensor::compose()` already performs the host all-gather.

tests/ttnn/distributed/test_default_mesh_composer.py
vtsilytskyiTT, line 1: ### Duplicate active shard dimensions are not validated
A topology such as:

`{Shard{-1}, Shard{3}}`
normalizes to `{3,3}` and reaches the same dims must be unique fatal.

Even if such topology is uncommon, implicit composition should either:

- flatten both mesh axes into one concatenation along dimension 3, or
- reject it cleanly and retain the “supply a composer” error.

### Tests are too narrow
The new test validates only:

1. single-host;
2. physical (1,2) mesh represented as a flattened one-dimensional distribution;
3. positive dimension 3;
4. one sharded axis;
5. logical to_torch().

It does not cover the branches created by the implementation:

1. negative dimensions;
2. genuine two-dimensional distribution;
3. singleton axes;
4. cluster_axis;
5. to_numpy;
6. multi-host;
7. offset submeshes;
8. padded output;
9. duplicate dimensions.

vtsilytskyiTT (changes requested):
I think you can simplify the code. Move topology-to-composer derivation into the distributed tensor layer:

```cpp
std::optional<MeshToTensor>
MeshToTensor::create_from_topology(
    const TensorTopology& topology,
    const MeshShape& physical_shape,
    int tensor_rank);
```

That helper should:

1. Normalize negative dimensions.
2. Ignore identity/singleton axes during concatenation.
3. Reject or flatten duplicate active dimensions.
4. Preserve mesh_coords() rather than reconstructing coordinate mapping.
5. Reject replicated multi-device axes.
6. Be tested directly with C++ topology tests.

Then `pytensor.cpp` only needs:

```cpp
if (is_globally_distributed(tt_tensor) && !padded_output) {
    if (auto composer = MeshToTensor::create_from_topology(...)) {
        return convert_to_row_major_host_buffer(tt_tensor, *composer);
    }
}
```

PR #53551, opened 2026-08-18
[Bug fix] GPT-OSS: minimal fix for garbage output on multi-device Blackhole (#52176)

mtairum:
## Minimal alternative to #52719

Same root cause, same fix mechanism, **3 files instead of 8**. Opened as a draft so the two can be compared side by side; #52719 stays open.

## Root cause (#52176)

On-device sampling intermittently returned an **out-of-range token ID** on multi-device Blackhole while a decode trace was live. Measured by reading logits to host and comparing host argmax against the device-sampled token:

```
step=0   host_argmax=35644    device_sampled=35644        AGREE
step=1   host_argmax=200008   device_sampled=3214098612   DIVERGE
```

`vocab_size` is 201088, so `3214098612` is uninitialized memory read as `uint32`. The logits are correct at every step — only the sampling output is corrupt. One bad token fed back gives an out-of-range embedding lookup that poisons the context, which is why it presented as gradual degradation.

On main, decode trace setup runs on the **first decode step**, by which point the prefill trace is live. Two allocations therefore land behind a live trace:

1. **the decode trace's persistent inputs** — `device_inputs[0]` is the token buffer, and that buffer *is* the sampling op's output, threaded as `tt_out_tok` into `ttnn.sampling(..., output_tensor=tt_out_tok)` (`models/common/sampling/tt_sampling.py:901-909`). That is the buffer observed returning `3214098612`.
2. **the sampling pipeline pre-compile** inside `capture_trace(..., skip_precompile=False)`.

This moves both ahead of every capture (prepare/record split + `SamplingGenerator.precompile` + `capture_trace(skip_precompile=True)`), and caches GPT-OSS's prefill rope slices, which were re-sliced — a device op — on every prefill call behind the live prefill trace.

## What was dropped from #52719, and why it is safe to drop

None of it is load-bearing for #52176:

| Dropped | Why |
|---|---|
| post-prefill (norm + LM head) trace | independent optimization. For GPT-OSS its traced tail is a single `ttnn.untilize`. Removed outright — the model-side `slice_last_token_block` it needs is not added here, and the generator-side machinery has been deleted rather than left dormant. Takes the sampling-mode keying and the `mistral-7b wh_n150` 25→32 MB trace-region bump with it. |
| `before_capture` / `_stage_prealloc_decode_inputs` (both deleted) | reachable only from `_row_sharded_batched_prefill`, which needs `users_row_sharded` **and** batch>1 **and** `sampling_params` — and Blackhole skips every batch>1 GPT-OSS case (`text_demo.py:484-488`). Unreachable on the reproducer. Matters for GPT-OSS batch-128 on Wormhole. |
| all `models/demos/llama3_70b_galaxy/**` | separate deferral implementation with its own latch |
| `tt_penalties.py`, greedy-row `seed[i] = None` | GPT-OSS sets no penalties and passes no seed |

Consequence: several serious findings raised against #52719 do not exist here — the `llama3_70b_galaxy` `KeyError`, the post-prefill sampling-graph keying, the phantom penalty tokens in the post-prefill passes, and the trace-region regression.

## Honest scope statement

- The **hoist is load-bearing**: there is no reordering inside the decode loop that fixes (1), because the prefill trace is already live by then.
- The prepare/record split in `models/tt_transformers/tt/generator.py` is retained in full, so the deferral machinery is still present for models that use `warmup_model_prefill`. This PR does **not** address the concerns raised on #52719 about that path (the `_easy_trace_prefill` `already_pending` reuse, and the flush running after a real prefill when the warmup page table has ≠1 row). GPT-OSS does not take that path — it takes the in-call `elif` branch, where the latch is set and cleared inside a single `prefill_forward_text` call.
- Attribution is from static reading of which allocations moved out from behind live traces. It is **not** a proven single-hunk causal chain.
- The prefill body runs **twice** per trace variant, as on `main`: the single compile pass runs over the persistent trace inputs, and the capture then binds to those same buffers. An earlier revision of this branch compiled over a transient copy and needed a third "warm" forward to close the program-cache gap between the two. That warm pass also allocated its output inside `_record_trace_prefill`, i.e. behind traces already captured for earlier variants — the exact hazard this PR exists to remove. Both are gone. (An earlier claim that this cost ~25% wall clock is retracted: per-iteration throughput varies ~3x across `bh_quietbox_2` hosts, which swamps an effect that size.)

## API note (post-review commits)

`finalize_deferred_traces` went from four parameters to zero (`829b18f5bb4`). The parameters were all ignored after the fallback removal, and the only caller is the internal `prefill_forward_text` wrapper. The method is new in this PR (it does not exist on `main`), so nothing out-of-tree can be calling it yet -- but if you are integrating against this branch, call it bare: `generator.finalize_deferred_traces()`.

### Reproducer and no-regression SKUs

Three parallel runs per SKU, to separate real signal from the infrastructure flake that dominated #52719's validation (container-startup faults, TLB exhaustion, checkout timeouts, undersized galaxy hosts):

- `bh_quietbox_2` — the reproducer
- `bh_galaxy` — the SKU in the original report
- `wh_galaxy_perf` — Wormhole no-regression

Links in the following comment.

Co-authored with @ifabijanic-tt, whose analysis and commits #52719 is built from.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/52176-gptoss-minimal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/52176-gptoss-minimal)
### Timing A/B on `bh_quietbox_2`

Three arms (with warm pass / `main` / warm pass removed) across qwen3.6-27b, qwen3-32b and qwen2.5-32b, to confirm the wall-clock regression is gone. Run links in the review-follow-ups comment below.

### DIFF
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -574,17 +574,20 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
 
         mesh_device = self.model_args[model_id].mesh_device
 
-        # Compile run. Its input buffers are transient; dropping the references frees them again. The
-        # output is a real prefill result for these ids, so deferred warmup callers can consume it instead
-        # of replaying a trace that has not been captured yet.
-        compile_inputs = copy_host_to_device(host_inputs, mesh_device=mesh_device)
-        prepared["compile_output"] = self._prefill_trace_forward(prepared, compile_inputs)
-        ttnn.synchronize_device(mesh_device)
-        del compile_inputs
-        logger.info("Done Compiling Model")
-
         # Persistent trace inputs: these outlive the capture and are refreshed in place on every replay.
+        # Allocated before the compile pass so that the compile pass can run over them directly.
         prepared["device_inputs"] = copy_host_to_device(host_inputs, mesh_device=mesh_device)
+
+        # Compile run, over the *persistent* inputs -- the exact buffers the capture will bind to, so every
+        # program it needs is cached against the specs it will actually see. Compiling over a separate
+        # transient copy would leave any program whose cache key differs between the two uncached, and it
+        # would then load inside the capture window, which the runtime rejects outright: "Cannot load new
+        # binaries during trace capture" (mesh_workload.cpp). Gemma-4's LM head hit this.
+        # The output is a real prefill result for these ids, so deferred warmup callers can consume it
+        # instead of replaying a trace that has not been captured yet.
+        prepared["compile_output"] = self._prefill_trace_forward(prepared, prepared["device_inputs"])
+        ttnn.synchronize_device(mesh_device)
+        logger.info("Done Compiling Model")
         return prepared
 
     def _record_trace_prefill(self, prepared):
@@ -598,13 +601,6 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
         # Release our handle on the compile-pass output before capturing, matching the pre-split behaviour.
         # A deferred warmup caller may still be holding it; that is its own reference to keep or drop.
         prepared.pop("compile_output", None)
-        # Warm the program cache over the *persistent* inputs before capturing. _prepare_trace_prefill's
-        # compile pass ran against transient buffers, so any program whose cache key differs between the two
-        # is still uncached here and would load inside the capture window, which the runtime rejects:
-        # "Cannot load new binaries during trace capture" (mesh_workload.cpp). Gemma-4's LM head hit this.
-        warmup_output = self._prefill_trace_forward(prepared, device_inputs)
-        ttnn.synchronize_device(mesh_device)
-        del warmup_output
         trace_id = ttnn.begin_trace_capture(mesh_device, cq_id=0)
         tt_out_trace = self._prefill_trace_forward(prepared, device_inputs)
         ttnn.end_trace_capture(mesh_device, trace_id, cq_id=0)
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -506,7 +506,17 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
         cores", program.cpp:2205). The prefetcher is Blackhole-only and GPT-OSS does not use it, so
         keeping these models on the original non-hoisted path costs this change nothing it targets.
         """
-        return any(getattr(m, "prefetcher", None) is not None for m in self.model)
+        uses = any(getattr(m, "prefetcher", None) is not None for m in self.model)
+        if uses and not getattr(self, "_logged_prefetcher_hoist_skip", False):
+            self._logged_prefetcher_hoist_skip = True
+            logger.info(
+                "DRAM prefetcher detected: keeping the original (non-hoisted) decode trace setup. "
+                "This model does not get the hoisted trace-allocation path, because the prefetcher's "
+                "per-mode sub-device managers are incompatible with preparing decode setup during the "
+                "prefill phase (see _uses_prefetcher). Trace I/O buffers for this model may still be "
+                "allocated while a trace is live."
+            )
+        return uses
 
     def _prepare_decode_trace_once(self, kv_cache, page_table, on_device_sampling):
         """Prepare the decode trace unless it is already prepared. Safe to call from either hoist point."""
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -455,9 +455,14 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
                 on_device_sampling=on_device_sampling,
             )
         finally:
+            # Restore only the generator's own mode. Deliberately no switch_mode(Mode.PREFILL) here:
+            # switch_mode drives Prefetcher.init(), whose Mode.PREFILL branch has never executed on main
+            # (main only ever calls switch_mode(Mode.DECODE)) and is not functional -- it reads an
+            # unassigned all_core_range_set, and supplying one instead trips
+            # "Statically allocated circular buffers ... clash with L1 buffers on core range".
+            # The prefetcher is not needed for prefill; main runs prefill with it left in whatever mode
+            # it was last initialized for, so leaving it in DECODE matches main's behaviour.
             self.mode = previous_mode
-            for i in range(len(self.model)):
-                self.model[i].switch_mode(Mode.PREFILL)
 
     def _will_row_shard_prefill(self, tokens, sampling_params):
         """Whether this prefill will be dispatched to the model's row-sharded batched path.
@@ -668,9 +673,8 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
             try:
                 trace_ids, tt_out_trace, *device_inputs = self._record_decode_trace_text(prepared)
             finally:
+                # See _prepare_decode_trace_for_warmup: no switch_mode(Mode.PREFILL) here either.
                 self.mode = previous_mode
-                for i in range(len(self.model)):
-                    self.model[i].switch_mode(Mode.PREFILL)
             self.trace_ids_decode[on_device_sampling] = trace_ids
             self.trace_inputs_decode[on_device_sampling] = device_inputs
             self.trace_output_decode[on_device_sampling] = tt_out_trace
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -904,19 +904,6 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
         assert (
             self.data_parallel == 1
         ), "Row-sharded batched prefill requires data_parallel=1 (model handles DP internally)"
-        before_capture = None
-        if enable_trace and self._prealloc_decode_inputs is None and not self._any_trace_captured():
-            # Allocate decode's persistent trace inputs in the one window this path leaves: after its own
-            # trace inputs are staged, before its capture. These are the buffers that matter -- they are
-            # refreshed in place for the entire decode loop, so allocating them later (with the prefill
-            # trace live) leaves them sitting in that trace's scratch, to be overwritten on every replay.
-            # The transient buffers of decode's compile pass are not hoisted with them: compiling decode
-            # here makes the capture below deadlock, so that pass stays in the decode loop. Its allocations
-            # are freed before anything replays, which is why they are the tolerable half of this.
-            # Same prefetcher exclusion as _prepare_decode_trace_once: staging decode inputs from inside
-            # the prefill capture window is not safe for prefetcher-driven models.
-            if not self._uses_prefetcher():
-                before_capture = lambda: self._stage_prealloc_decode_inputs(page_table)  # noqa: E731
         return self.model[0].row_sharded_batched_prefill(
             tokens,
             page_table,
@@ -932,7 +919,6 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
                 "outputs": self.trace_output_prefill,
             },
             empty_slots=empty_slots,
-            before_capture=before_capture,
         )
 
     def _easy_trace_prefill(
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -491,8 +491,22 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
         if self._prealloc_decode_inputs is None:
             self._prealloc_decode_inputs = self._prealloc_decode_trace_inputs(page_table)
 
+    def _uses_prefetcher(self):
+        """Whether any model instance drives the DRAM prefetcher.
+
+        The prefetcher owns sub-device managers that differ per mode, so hoisting decode setup into the
+        prefill phase cannot work for it: switching to DECODE and back reaches the prefetcher's
+        Mode.PREFILL branch, which has never run on main and is not functional, while staying in DECODE
+        leaves prefill running against decode sub-devices ("Kernel group cores do not match sub device
+        cores", program.cpp:2205). The prefetcher is Blackhole-only and GPT-OSS does not use it, so
+        keeping these models on the original non-hoisted path costs this change nothing it targets.
+        """
+        return any(getattr(m, "prefetcher", None) is not None for m in self.model)
+
     def _prepare_decode_trace_once(self, kv_cache, page_table, on_device_sampling):
         """Prepare the decode trace unless it is already prepared. Safe to call from either hoist point."""
+        if self._uses_prefetcher():
+            return
         if self._pending_decode_trace is None:
             self._pending_decode_trace = self._prepare_decode_trace_for_warmup(
                 kv_cache=kv_cache,
@@ -884,7 +898,10 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
             # The transient buffers of decode's compile pass are not hoisted with them: compiling decode
             # here makes the capture below deadlock, so that pass stays in the decode loop. Its allocations
             # are freed before anything replays, which is why they are the tolerable half of this.
-            before_capture = lambda: self._stage_prealloc_decode_inputs(page_table)  # noqa: E731
+            # Same prefetcher exclusion as _prepare_decode_trace_once: staging decode inputs from inside
+            # the prefill capture window is not safe for prefetcher-driven models.
+            if not self._uses_prefetcher():
+                before_capture = lambda: self._stage_prealloc_decode_inputs(page_table)  # noqa: E731
         return self.model[0].row_sharded_batched_prefill(
             tokens,
             page_table,
--- a/models/tt_transformers/tt/generator.py
+++ b/models/tt_transformers/tt/generator.py
@@ -190,6 +190,23 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
         "supports_prefix_caching": True,
     }
 
+    def _any_trace_captured(self):
+        """True once any trace has been captured, i.e. once allocations are no longer unconditionally safe.
+
+        Used to decide whether a prefill call may still arm capture deferral: the point of deferring is to
+        keep every compile pass and every persistent allocation ahead of the first capture, which is only
+        achievable while nothing has been captured yet.
+        """
+        return any(
+            any(store.values())
+            for store in (
+                self.trace_id_prefill,
+                self.trace_id_prefill_sampling,
+                self.trace_id_post_prefill,
+                self.trace_ids_decode,
+            )
+        )
+
     def _get_sampling_contract(self, model_id: int):
         sampling_module = getattr(self.model[model_id], "sampling", None)
         sampling_dp = getattr(self.model[model_id], "sampling_dp", 1)
@@ -263,23 +280,36 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
                 kv_cache=kv_cache, page_table=page_table, can_sample_on_device=can_sample_on_device
             )
 
-    def finalize_deferred_traces(self, kv_cache, page_table, can_sample_on_device):
+    def finalize_deferred_traces(self, kv_cache, page_table, can_sample_on_device, prefill_samples_on_device=None):
         """Prepare the remaining traces and capture everything that is pending.
 
         Called once the prefill that triggered warmup has finished, so every prefill trace variant that run
         actually uses -- including the batched-prefill one, which warmup cannot know about -- has already
         been prepared. Everything below still runs before the first capture, so once this returns nothing
         allocates outside a capture window.
+
+        ``can_sample_on_device`` is decode's sampling mode (it keys the decode trace).
+        ``prefill_samples_on_device`` is prefill's, which keys the post-prefill trace -- the two differ when
+        a caller samples on device during decode but reads logits back on host during prefill (GPT-OSS).
+        Capturing the post-prefill tail in the mode prefill does not use would leave a trace nothing replays.
+        Defaults to ``can_sample_on_device`` for callers that do not distinguish them.
         """
         if not self._defer_trace_recording:
             return
+        if prefill_samples_on_device is None:
+            prefill_samples_on_device = can_sample_on_device
         try:
             # One post-prefill trace per model, shaped from any prepared prefill variant's body output.
             for prepared in self._pending_prefill_traces.values():
                 model_id = prepared["model_id"]
                 if model_id in self._pending_post_prefill_traces:
                     continue
-                sampling_enabled = can_sample_on_device and getattr(self.model[model_id], "sampling", None)
+                # Tracing the tail needs a slice that can write into a pre-allocated buffer; without it the
+                # replay would allocate its own input every call, which is the very thing being avoided.
+                # Models that do not implement it keep the eager tail (and its allocations).
+                if getattr(self.model[model_id], "slice_last_token_block", None) is None:
+                    continue
+                sampling_enabled = prefill_samples_on_device and getattr(self.model[model_id], "sampling", None)
                 self._pending_post_prefill_traces[model_id] = self._prepare_post_prefill_trace(
                     model_id, prepared, bool(sampling_enabled)
                 )
@@ -287,15 +317,15 @@ class Generator(ModelCapabilitiesMixin, WarmupForwardMixin):
             # Decode last, so its compile pass and its capture stay adjacent: capturing a program the
             # cache has not seen fails outright, and the sampling configuration must not drift in between.
             #
-            # Only when this runs straight after warmup. Reached from the batched path -- i.e. after the
-            # caller's own prefill -- the model's sampling state has been reconfigured per prefill user, and
-            # compiling decode against it fails ("Batch size 1 must be equal to max_batch_size"). There the
-            # decode trace stays lazy, exactly as before this change.
-            self._pending_decode_trace = self._prepare_decode_trace_for_warmup(
-                kv_cache=kv_cache,
-                page_table=page_table,
-                on_device_sampling=can_sample_on_device,
-            )
+            # Unless it was already prepared up front. The no-warmup path does that deliberately, because
+            # there the flush happens after a real prefill and the decode compile pass would write mock K/V
+            # over it (see _prefill_forward_text_impl).
+            if self._pending_decode_trace is None:
+                self._pending_decode_trace = self._prepare_decode_trace_for_warmup(
+                    kv_cache=kv_cache,
+                    page_table=page_table,
+                    on_device_sampling=can_sample_on_device,
+                )
         finally:
             self._defer_trace_recording = False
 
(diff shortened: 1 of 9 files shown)


mtairum:
## CI: 3 parallel runs per SKU

Three runs each, deliberately, so a single infrastructure failure does not read as a code failure. #52719's validation was repeatedly derailed by container-startup faults (`Value cannot be null (ContainerId)`), TLB-window exhaustion on one galaxy host, 20-minute checkout timeouts, `codeload` archive failures, and galaxy jobs landing on 4-device machines. Parallel dispatch also spreads across hosts — on #52719 it took three simultaneous attempts before one `bh_galaxy` run landed on a healthy runner.

| SKU | Runs |
|---|---|
| `bh_quietbox_2` (reproducer) | [32158733346](https://github.com/tenstorrent/tt-metal/actions/runs/32158733346) · [32158750807](https://github.com/tenstorrent/tt-metal/actions/runs/32158750807) · [32158767042](https://github.com/tenstorrent/tt-metal/actions/runs/32158767042) |
| `bh_galaxy` (SKU from the report) | [32158738474](https://github.com/tenstorrent/tt-metal/actions/runs/32158738474) · [32158754293](https://github.com/tenstorrent/tt-metal/actions/runs/32158754293) · [32158771016](https://github.com/tenstorrent/tt-metal/actions/runs/32158771016) |
| `wh_galaxy_perf` (no-regression) | [32158741996](https://github.com/tenstorrent/tt-metal/actions/runs/32158741996) · [32158758179](https://github.com/tenstorrent/tt-metal/actions/runs/32158758179) · [32158774750](https://github.com/tenstorrent/tt-metal/actions/runs/32158774750) |

Pass criteria per run, not just a green tick:
- `PASSED ... test_gpt_oss_demo[...prefill_128]`
- coherent generated text for USER 0
- `On-device sampling unavailable` count **0** and `On-device sampling initialized (vocab_size=201088, splits=N)` present — otherwise the run silently used the host-argmax fallback, which is clean by construction and proves nothing

Broader-tier sweeps come next if these hold. Note the wider matrix was already exercised on #52719 at Tier 1/2/3 with no trace-region OOM anywhere, including the 10 MB-region models (`Llama-3.2-1B/3B`, `Phi-3-mini`, `Qwen2.5-7B`) — and this PR drops the post-prefill trace that raised trace-region demand in the first place, so that pressure is lower here than there.

mtairum:
## Tier 1/2/3 e2e on this branch

The GPT-OSS runs above validate the fix, but they say nothing about the other models — and this patch is **not** isolated to GPT-OSS. `models/tt_transformers/tt/generator.py` is still +917/-168, the same scope as #52719, and that file is shared by every tt_transformers-based model (~28 e2e and ~34 unit pipeline entries). The earlier Tier 1/2/3 sweeps I cited were run against #52719, a different tree.

So, full sweeps on this branch, `model=all` / `sku=all`:

| Pipeline | Run |
|---|---|
| (Tier 1) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32241743486 |
| (Tier 2) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32241746090 |
| (Tier 3) Models E2E | https://github.com/tenstorrent/tt-metal/actions/runs/32241749522 |

### What these are specifically checking

The deferral machinery is retained in full here — `_defer_trace_recording`, `_pending_prefill_traces`, `finalize_deferred_traces`, the `_easy_trace_prefill` `already_pending` reuse path, `_warmup_covered_all_variants`. GPT-OSS never enters it: it takes the in-call `elif` branch, where the latch is set and cleared inside a single `prefill_forward_text` call.

**Models that do take the cross-call `warmup_prefill=True` path are therefore still exposed to the two most serious findings raised on #52719** — the `already_pending` re-run returning a *different* prompt's logits, and the flush running after a real prefill (re-executing prefill bodies over warmup's zeroed page table, and running the decode compile pass at `current_pos=0` against the caller's real page table). Those are `models/demos/gemma4/**` and `models/demos/multimodal/gemma3/**`, plus the experimental Quasar and diffusion-gemma paths.

Gemma-4 is the job to watch: it takes the warmup path, and it is where the `mesh_workload.cpp:153 !is_capturing_trace` regression first surfaced during #52719's validation.

### Reading the results

Failures will need per-job triage against main rather than being taken at face value — the recurring non-code classes on this hardware are undersized galaxy hosts (`requested_size <= system_size`), read-only `/mnt/MLPerf` cache writes, container-startup faults (`Value cannot be null (ContainerId)`), TLB-window exhaustion, checkout and `codeload` timeouts, and perf-gate upper-bound trips where the tests themselves pass. I will diff each failure against main's scheduled runs before attributing anything to this PR.

A further reduction is worth considering depending on what comes back: gate the hoist to the in-call `elif` path only and drop the cross-call `warmup_model_prefill` arming entirely. GPT-OSS does not need it, and it is where both severe findings live — that would make this patch genuinely isolated rather than merely smaller.

mtairum:
## Cross-call deferral removed — now genuinely scoped to the decode-trace hoist

Pushed `5b8213d4663`. This addresses the concern that "minimal" was smaller but not isolated.

### What changed

The deferral machinery does not exist on main (0 refs to `_defer_trace_recording`, `_pending_prefill_traces`, `finalize_deferred_traces`) — #52719 introduced it, and independent review found two severe defects in it, both on the cross-call warmup path:

- `_easy_trace_prefill`'s `already_pending` re-run returns a **different prompt's logits** (re-running the body over the first caller's staged inputs), with that request's KV never written
- the flush running after the caller's real prefill re-executes prefill bodies over warmup's zeroed page table, and runs the decode compile pass at `current_pos=0` against the caller's real page table

Both land on callers that prefill **once per request** — i.e. vLLM serving, which is precisely the use case `warmup_model_prefill` exists for. The demos escape only because they prefill twice and the second pass overwrites the damage.

Neither is needed for #52176. What *is* load-bearing is hoisting the decode trace's persistent inputs and the sampling pre-compile ahead of the first capture. So this keeps exactly that, inside a single `prefill_forward_text` call, and drops:

- `warmup_model_prefill` arming the latch across calls, plus the `_warmup_covered_all_variants` early-finalize gate
- staging prefill traces and the `already_pending` re-run — prefill traces now capture immediately, as on main
- staging prefill-sampling traces

### The seqlen warmup benefit is unaffected

Worth stating explicitly, since that is the point of the feature: main's `warmup_model_prefill` already sweeps `get_warmup_prefill_supported_seq_lens()` and captures a prefill trace per length, and it still does here. Only the capture *timing* returns to main's behaviour. Production serving keeps every pre-warmed seqlen.

Fixing the deferral properly — so the warmup path keeps the allocation-safety win too — is separate work and belongs with #52719.

### Validation in flight

GPT-OSS-120B, 2 runs per SKU: [32242824805](https://github.com/tenstorrent/tt-metal/actions/runs/32242824805) · [32242827956](https://github.com/tenstorrent/tt-metal/actions/runs/32242827956) · [32242830666](https://github.com/tenstorrent/tt-metal/actions/runs/32242830666) · [32242833378](https://github.com/tenstorrent/tt-metal/actions/runs/32242833378) · [32242836525](https://github.com/tenstorrent/tt-metal/actions/runs/32242836525) · [32242839796](https://github.com/tenstorrent/tt-metal/actions/runs/32242839796) · [32242842251](https://github.com/tenstorrent/tt-metal/actions/runs/32242842251)

Full sweeps: [Tier 1](https://github.com/tenstorrent/tt-metal/actions/runs/32242824805) · [Tier 2](https://github.com/tenstorrent/tt-metal/actions/runs/32242844943) · [Tier 3](https://github.com/tenstorrent/tt-metal/actions/runs/32242847752)

Gemma-4 and multimodal Gemma-3 are the jobs that matter most now — they are the models that took the cross-call warmup path, so they exercise the code this commit reverted to main's behaviour.

The earlier tier sweeps on this branch ([32241743486](https://github.com/tenstorrent/tt-metal/actions/runs/32241743486), [32241746090](https://github.com/tenstorrent/tt-metal/actions/runs/32241746090), [32241749522](https://github.com/tenstorrent/tt-metal/actions/runs/32241749522)) predate this commit and are superseded.

ifabijanic-tt (approved):
Maybe we can make comments more concise in the fix but otherwise it's OK

mtairum:
### Review follow-ups pushed (`55edecdec47`)

**1. Penalty counters — never reset, just don't count** (`sampling/generator.py`)

The side-effect passes sampled a token off dummy inputs and then called `reset_penalty_counts()` to undo it. That also discarded the caller's real output history: `sample()` enters `capture_trace` lazily on a *real* decode step, so by the time capture runs the counters already hold genuine tokens. Replaced with a `count_tokens` kwarg on `_run_sampling`/`sample()`, passed `False` from the pre-compile and capture passes — nothing phantom is counted, so nothing needs undoing. All three `reset_penalty_counts()` call sites are gone.

**2. Dead post-prefill subsystem removed, rather than described**

`trace_id_post_prefill` was only ever initialized to `None` and read, so `_run_post_prefill` always returned `None` and every prefill took the eager fallback. `_prepare_post_prefill_trace` and `_record_post_prefill_trace` had no callers at all, and the machinery still referenced `slice_last_token_block`, which this branch does not add. Also dropped `_stage_prealloc_decode_inputs` / `_prealloc_decode_trace_inputs` (no callers, so the pickup site always saw `None`) and three unused deferral attributes. None of these symbols exist on `main`. The PR description now matches the code.

**3. Comments trimmed and corrected**

Several blocks had gone stale against the cross-call-arming removal — they still claimed deferral persists across calls, and that the row-sharded path gets pre-allocated inputs. Both corrected, the rest tightened.

**4. `finalize_deferred_traces` is decode-only and exception-safe** (`d6aabbf5f22`)

---

### The warm pass was a real regression — also fixed here

Worth calling out because it was not in the review: the prepare/record split ran the prefill body **three** times per trace variant — compile over a transient copy of the inputs, a warm pass over the persistent inputs to close the program-cache gap between them, then the capture — where `main` runs it twice.

On `bh_quietbox_2` that showed up as roughly **+25% wall clock**: Qwen3-32B's first pytest session went 450s on `main` to 561s here, which pushed several jobs past the GitHub *action* time budget (`##[error]The action has timed out.` — the tests themselves passed).

The fix removes the warm pass rather than optimizing it: allocate the persistent inputs first and run the single compile pass directly over **them**, so the capture binds to the exact buffers the compile pass already exercised. No cache gap to close, two forwards per variant as on `main`.

This also makes `_record_trace_prefill`'s "allocation-free outside the capture window" contract true rather than aspirational — the warm pass had been allocating behind traces captured for earlier variants, which was unsound from the second variant onwards.

### Validation in flight

Three arms on `bh_quietbox_2`, one dispatch each for qwen3.6-27b / qwen3-32b / qwen2.5-32b, to separate the regression from the fix:

| arm | sha | qwen3.6-27b | qwen3-32b | qwen2.5-32b |
|---|---|---|---|---|
| with warm pass | `98de2ecf545` | [32257523438](https://github.com/tenstorrent/tt-metal/actions/runs/32257523438) | [32257533126](https://github.com/tenstorrent/tt-metal/actions/runs/32257533126) | [32257540648](https://github.com/tenstorrent/tt-metal/actions/runs/32257540648) |
| `main` baseline | `b513e875c8e` | [32257528188](https://github.com/tenstorrent/tt-metal/actions/runs/32257528188) | [32257536755](https://github.com/tenstorrent/tt-metal/actions/runs/32257536755) | [32257545316](https://github.com/tenstorrent/tt-metal/actions/runs/32257545316) |
| warm pass removed | `55edecdec47` | [32257717729](https://github.com/tenstorrent/tt-metal/actions/runs/32257717729) | [32257721353](https://github.com/tenstorrent/tt-metal/actions/runs/32257721353) | [32257725395](https://github.com/tenstorrent/tt-metal/actions/runs/32257725395) |

Note on triage for anyone reading the earlier tier runs: the five `wh_n150` failures and Z-Image-Turbo there failed at **"Set up runner" / "Set up job"** — no test ran. Llama-3.3-70B Galaxy hit `tt_tlb_alloc failed with error code -12`, the known galaxy TLB-window infra issue. And qwen3.6-27b's `traced_128` parametrization has never run on `main` in CI (main collects 5 params, this branch 19), so its timeout there is not attributable from history — hence the explicit baseline arm above.

mtairum:
### Correction: retracting the "+25% wall clock" number

The `bh_quietbox_2` pool is far too heterogeneous for the comparison I drew, so the specific figure in my previous comment is not supported. Measured rates for the same 64-iteration loop in the same job:

| where | runner | rate |
|---|---|---|
| scheduled Tier 2 on `main` | (scheduled pool) | ~3.5 s/it |
| this branch, with warm pass | `qb2-120-p05t04` | 5.03 s/it |
| `main` baseline, dispatched | `qb2-120-p02t04` | 11.80 s/it |

Per-iteration throughput varies by roughly 3x across hosts, which swamps any effect of the size I claimed. The 450s-vs-561s comparison behind "+25%" was also across two different runners, so it does not isolate the code change.

Two things follow. First, **`main` itself timed out on this exact job** ([32257536755](https://github.com/tenstorrent/tt-metal/actions/runs/32257536755), `qb2-120-p02t04`) — it never finished its first pytest session, whereas the branch completed one in 561s. So the action time budget for these jobs is marginal on the slower QB2 hosts independent of this PR, and the three timeouts in the earlier tier runs should not be attributed to it.

Second, `55edecdec47` stands, but on the two grounds that are verifiable by reading the code rather than by timing:

- **Forward count.** `main` runs the prefill body twice per trace variant. The branch ran it three times: compile over a transient copy, a warm pass over the persistent inputs to close the program-cache gap, then the capture. The fix restores two, and this is checkable statically — one `_prefill_trace_forward` compile call site and one capture call site.
- **Soundness.** The warm pass allocated its output inside `_record_trace_prefill`, which runs after traces for earlier variants have been captured. That is exactly the allocate-behind-a-live-trace hazard this PR exists to remove, and it contradicted the method's own "allocation-free outside the capture window" contract. Compiling over the persistent inputs removes the allocation entirely instead of relocating it.

Whether the extra forward was ever load-bearing for the CI timeouts is now unproven, and I am not claiming it. The A/B/C arms are still running; I will report what they show, including if they show no timing difference.

mtairum:
### Root cause of the `bh_quietbox_2` timeouts: weight loading, not this PR

The A/B arms settled this. The 64-iteration progress bar I had been timing is **transformer layer loading** — 64 layers — not any trace or warmup work. It is labelled explicitly in the qwen3.6-27b log, and is the same loop in Qwen3-32B (unlabelled, but 64 iterations starting immediately after `Loading weights: 100%|...| 707/707`).

Its throughput varies enormously by host:

| job | runner | s/layer | outcome |
|---|---|---|---|
| Qwen3-32B, branch tier run | (scheduled) | 3.48 → 5.03 | test passed, 561s |
| Qwen3-32B, `main` dispatched | `qb2-120-p02t04` | 11.80 | 12:35 just to load layers, then action timeout |
| Qwen3.6-27B, branch dispatched | `qb2-120-p05t04` | 63.45 | timed out at layer **18 of 64** |

[32257523438](https://github.com/tenstorrent/tt-metal/actions/runs/32257523438) is the clearest case:

```
Loading layers:  28%|██▊       | 18/64 [19:02<48:38, 63.45s/it]
##[error]test_demo_text[blackhole-traced_128-device_params0-mesh_device0]
Failed: Timeout (>1200.0s) from pytest-timeout.
```

It spent the entire 1200s budget loading weights from `/mnt/MLPerf` and **never reached model execution at all** — so no code in this PR ran before the timeout. The same explanation covers the `main` baseline arm, which also timed out ([32257536755](https://github.com/tenstorrent/tt-metal/actions/runs/32257536755)) after 12:35 of layer loading.

So all three `bh_quietbox_2` timeouts in the earlier tier runs are NFS weight-load slowness on specific QB2 hosts, ~18x spread at the extremes, and none of them are attributable to this PR. My "+25% wall clock" claim remains retracted, and I was measuring weight loading throughout.

`55edecdec47` still stands on the two statically-checkable grounds — the branch ran the prefill body three times per trace variant where `main` runs it twice, and the warm pass allocated inside `_record_trace_prefill`, behind traces already captured for earlier variants. Neither claim depends on the timing data.

Separately, this looks worth an infra ticket independent of this PR: a 1200s `pytest-timeout` cannot accommodate a host that needs 68 minutes to load 64 layers, and the failure presents as a model bug rather than as slow storage.

mtairum:
### Addendum: I caused much of that weight-loading slowness

Correcting my own framing above. I dispatched all nine A/B arms at once (13:27–13:43), so up to nine jobs were streaming this model's weights off `/mnt/MLPerf` concurrently. The rates line up with that far better than with host quality:

| arm | runner | s/layer |
|---|---|---|
| qwen3-32b, `main` | `qb2-120-p02t04` | 11.80 |
| qwen3.6-27b, fixed | `qb2-120-p06t03` | 45.70 |
| qwen3.6-27b, warm pass | `qb2-120-p05t04` | 63.45 |
| qwen2.5-32b, `main` | `qb2-120-p02t08` | 70.68 |
| scheduled tier runs | scheduled pool | **3.48–5.03** |

Every dispatched arm is 3–20x slower than the scheduled runs, across four different hosts and both `main` and this branch. Self-inflicted NAS contention is the most likely explanation, so the "~18x host spread" I quoted overstates the infra problem and the A/B was largely self-defeating. I have cancelled the two remaining qwen2.5-32b arms to stop adding load.

What still holds, and is what actually matters for this PR:

- **Both `main` arms that completed failed exactly as the branch arms did**, in weight loading, without reaching model execution ([32257536755](https://github.com/tenstorrent/tt-metal/actions/runs/32257536755), [32257545316](https://github.com/tenstorrent/tt-metal/actions/runs/32257545316) — the latter at layer 20/64). So Qwen2.5-32B and Qwen3-32B on `bh_quietbox_2` fail this way on `main` too, and the earlier tier failures are not attributable to this PR.
- **No arm reached any code this PR touches**, so none of them is evidence for or against it either way.

The earlier tier-run timeouts have their own concurrency to account for (a full tier runs many jobs at once), so I am not claiming the scheduled failures share my specific cause — only that in every case the job died loading weights rather than executing the model.

Functional validation for this PR therefore rests on the Tier 1/2/3 runs already linked, not on these arms.

mtairum:
### A/B arms: final results

**Qwen3.6-27B — fails identically on `main`.** All three arms hit the 1200s `pytest-timeout` inside weight loading, none reaching model execution:

| arm | duration | died at |
|---|---|---|
| warm pass | 1204.64s | layer 18/64 |
| warm pass removed | 1202.42s | layer 26/64 |
| **`main`** (`b513e875c8e`) | **1207.70s** | **layer 17/64**, 66.12 s/layer |

The `main` arm selected the same `test_demo_text[blackhole-traced_128-device_params0-mesh_device0]` (18 deselected, so current `main` does carry `traced_128`) and failed the same way. This test's failure is **not attributable to this PR**, and it now has its first `main` data point — which is a failure under NAS contention, not a pass.

**Qwen3-32B — the fixed arm is the only one that passed.**

| arm | result |
|---|---|
| warm pass | timed out, no session completed |
| `main` | timed out, no session completed |
| **warm pass removed** ([32257721353](https://github.com/tenstorrent/tt-metal/actions/runs/32257721353)) | **success — both sessions: 314.07s and 556.30s** |

So `55edecdec47` completes this SKU cleanly where both the previous revision and `main` timed out under the same self-inflicted contention.

I am deliberately **not** reading that as a proven speedup. Arm 3 was dispatched last and likely got a less contended window, and its two sessions (314s + 556s = 870s) against the scheduled `main` run's (450s + 304s = 754s) do not favour it anyway. What the result does establish is the thing that was actually in doubt: **the fixed code runs Qwen3-32B on `bh_quietbox_2` to completion**, both pytest sessions, with the prefill body compiled over the persistent trace inputs and no warm pass.

Summary of the whole exercise: no arm produced usable timing data, both `main` arms failed the same way the branch arms did, and the one arm that completed is the fixed code. The functional case for this PR rests on the Tier 1/2/3 runs plus this Qwen3-32B pass; the case for `55edecdec47` specifically rests on forward count and allocation soundness, both checkable by reading the diff.

mtairum:
### Validation re-run on the current head (`55edecdec47`)

Everything validated so far was `5b8213d4663`, which is **three commits behind** this branch's head. Two of those three are material — `98de2ecf545` deletes the dead post-prefill/prealloc methods, and `55edecdec47` moves the prefill compile pass onto the persistent trace inputs, which every traced prefill goes through. Neither had tier coverage, so the earlier tier links do not speak for the code as it stands.

Re-dispatched on `55edecdec47`:

| scope | run |
|---|---|
| GPT-OSS 120B `bh_quietbox_2` (reproducer) | [32263071568](https://github.com/tenstorrent/tt-metal/actions/runs/32263071568) |
| GPT-OSS 120B `bh_galaxy` (SKU in #52176) | [32263076256](https://github.com/tenstorrent/tt-metal/actions/runs/32263076256) |
| GPT-OSS 120B `wh_galaxy_perf` (WH no-regression) | [32263081435](https://github.com/tenstorrent/tt-metal/actions/runs/32263081435) |
| Tier 1 e2e, all models / all SKUs | [32263085777](https://github.com/tenstorrent/tt-metal/actions/runs/32263085777) |
| Tier 2 e2e, all | [32263090545](https://github.com/tenstorrent/tt-metal/actions/runs/32263090545) |
| Tier 3 e2e, all | [32263095319](https://github.com/tenstorrent/tt-metal/actions/runs/32263095319) |

Two notes on what these supersede:

- The Tier 1/2/3 runs at `5b8213d4663` ([32242842251](https://github.com/tenstorrent/tt-metal/actions/runs/32242842251), [32242844943](https://github.com/tenstorrent/tt-metal/actions/runs/32242844943), [32242847752](https://github.com/tenstorrent/tt-metal/actions/runs/32242847752)) never completed — dispatched at 10:28 and still showing queued jobs almost four hours later (1, 14 and 6 outstanding respectively). Their partial tallies were 26/6, 25/4 and 26/3 pass/fail, with the failures triaged to infra (see above).
- GPT-OSS at `5b8213d4663` was green on `bh_galaxy` twice, and 1-pass/1-fail on each of `bh_quietbox_2` and `wh_galaxy_perf`.

Deliberate change from the earlier 3-parallel-per-SKU approach: **one** GPT-OSS run per SKU this time. Three concurrent runs of the same large model is what starved `/mnt/MLPerf` earlier and converted nine A/B arms into weight-load timeouts, which produced no signal at all. Any SKU that comes back flaky gets retried individually instead.

mtairum:
### Tier 1/2/3 at `5b8213d4663` — final: 62 pass / 13 fail, none code-attributable

The tier runs that had been stuck queued finished. Totals: Tier 1 [32242842251](https://github.com/tenstorrent/tt-metal/actions/runs/32242842251) 21/6, Tier 2 [32242844943](https://github.com/tenstorrent/tt-metal/actions/runs/32242844943) 20/4, Tier 3 [32242847752](https://github.com/tenstorrent/tt-metal/actions/runs/32242847752) 21/3.

With runner names attached, all 13 failures fall into three causes:

**1. One bad `wh_n150` host — 5 failures, all on `tt-metal-ci-vm-152`, all at the "Set up runner" step, so no test process ever started:** Gemma-3-4B, Stable Diffusion XL (img2img + inpainting), SqueezeBERT, VGGNet, BERT-Tiny. Nothing in common but the machine.

**2. `bh_quietbox_2` weight-load timeouts — 3 failures**, where the job spends its budget loading transformer layers off `/mnt/MLPerf` and never reaches model execution: Qwen3.6-27B (`qb2-120-p06t02`), Qwen3-32B Blackhole (`qb2-120-p06t02`), Qwen2.5-32B (`qb2-120-p05t09`). All three have now been shown to fail the same way on `main` — Qwen3.6-27B at layer 17/64 and Qwen2.5-32B at layer 20/64 in the dispatched baseline arms.

**3. Known-broken or known-flaky — 5 failures:** Llama 3.3-70B Galaxy (`tt_tlb_alloc ... -12` on `OM1-01A01-STGWH03`, a host that has been failing hard since late July), Qwen3-32B Galaxy (fails on `main` in both baselines), TT-DiT LTX-2.3 Pro T2V and Fast I2V on `bh_sc1` (success/failure split across baselines), Z-Image-Turbo (died at "Set up job").

So the reduced patch is clean across all three tiers at that sha. Note these results describe `5b8213d4663`; the re-run on the current head `55edecdec47` is still in flight ([32263085777](https://github.com/tenstorrent/tt-metal/actions/runs/32263085777), [32263090545](https://github.com/tenstorrent/tt-metal/actions/runs/32263090545), [32263095319](https://github.com/tenstorrent/tt-metal/actions/runs/32263095319)) and is what should gate merge.

Two infra items that are independent of this PR and worth someone's attention: `tt-metal-ci-vm-152` failing every job it picks up at runner setup, and `OM1-01A01-STGWH03` still in the galaxy pool.

models/tt_transformers/tt/generator.py
```diff
@@ -430,16 +650,27 @@ def _capture_trace_prefill_sampling(self, model_id, sampling_batch):
...
+    def _record_trace_prefill_sampling(self, prepared):
+        """Capture the batched prefill sampling trace prepared by :meth:`_prepare_trace_prefill_sampling`."""
+        model_id = prepared["model_id"]
+        mesh_device = self.model_args[model_id].mesh_device
+        trace_input = prepared["input"]
 
         trace_id = ttnn.begin_trace_capture(mesh_device, cq_id=0)
         logits = self.model[model_id]._apply_norm_and_lm_head(trace_input)
-        tt_tokens, tt_log_probs = self.model[model_id].sampling.sample(logits, enable_trace=False)
+        tt_tokens, tt_log_probs = self.model[model_id].sampling.sample(logits, enable_trace=False, count_tokens=False)
```
yieldthought: Codex: Keep token accounting in the captured graph. The compile pass above executes against dummy input, so `count_tokens=False` is correct there. This block only records commands; disabling counting here means every real replay omits `update_output_tokens`, so the first decode step does not penalize the token emitted by prefill.

```suggestion
        tt_tokens, tt_log_probs = self.model[model_id].sampling.sample(logits, enable_trace=False)
```

sraizada-tt (approved):

yieldthought (approved):

models/tt_transformers/tt/generator.py
mtairum: Fixed in 829b18f5bb4 — capture window now counts (default), matching SamplingGenerator.capture_trace's convention; only the eager pre-compile pass in _prepare_trace_prefill_sampling keeps count_tokens=False, since it actually executes over dummy logits.

PR #53557, opened 2026-08-18
[Feature] Stable sort and topk

### LINKED ISSUE #33492: Sort `ckernel::topk_local_sort` kernel produces wrong results with `stable` flag set to True

### LINKED ISSUE #54043: ttnn.sort: CrossCoreDataExchange unstable sorts return invalid indices on ties (duplicates in tie groups; out-of-range indices >= logical W at non-pow2 widths)
### Summary

Unstable (`stable=False`) `ttnn.sort` calls routed to the **CrossCoreDataExchange** program factory return **invalid index tensors** whenever a value tie group spans a cross-core tile pair. Two manifestations, same root cause:

1. **Duplicate indices inside tie groups** at CrossCore's native widths (padded W = 4096 / 8192): the index tensor is not a permutation of `[0, W)`. Values are sorted exactly and `torch.gather(input, -1, indices)` still reproduces the sorted values (both duplicated entries carry the tied value), so value-only and gather-parity tests pass while the indices are silently wrong.
2. **Out-of-range indices at non-power-of-2 logical widths**: indices ≥ logical W are returned (they point into the ±inf pad lanes added by the composite padding), and ascending-value gather verification fails. **This is an out-of-bounds hazard for downstream `gather` / `scatter` / embedding-lookup consumers** — the returned index is outside the logical tensor extent.

Observed on Blackhole p150a, current `main` (pre-dates any in-flight sort work — reproduced on `main` as well as feature branches). Affects **both TILE and ROW_MAJOR** layouts, H from 32 to 320, both `descending` values. The stable path (`stable=True`, index-aware comparator `topk_cmp_swap_stable_directional`) is NOT affected — it survived a 4,300+-cell adversarial sweep with exact `torch.sort(stable=True)` parity.

### Minimal repro A — duplicate indices in tie groups (W=4096, also fires at 8192)

```python
import torch, ttnn

x = torch.full((32, 4096), 1.0).bfloat16()  # one giant tie group
tx = ttnn.from_torch(x, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
values, indices = ttnn.sort(tx, dim=-1, stable=False)

idx = ttnn.to_torch(indices).to(torch.int64)
perm = torch.arange(4096).expand(32, -1)
rows_ok = (torch.sort(idx, dim=-1).values == perm).all(dim=-1)
print(rows_ok.sum().item(), "/ 32 rows are a permutation")  # 0 / 32 — every row has duplicates
# values are exact and torch.gather(x, -1, idx) == sorted values, so the
# corruption is invisible to value/gather checks.
```

Same result on `ttnn.Layout.ROW_MAJOR` and with partial tie groups (e.g. 5-level tie-heavy inputs): duplicates appear inside each tie group.

### Minimal repro B — indices ≥ logical W at non-pow2 width (OOB hazard)

```python
import torch, ttnn

W = 3000  # pads to 4096 with ±inf sentinels at the composite layer
x = torch.full((32, W), -2.0)
x[:, 0] = float("inf")
x[:, W - 1] = float("-inf")
tx = ttnn.from_torch(x.bfloat16(), ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
values, indices = ttnn.sort(tx, dim=-1, stable=False)

idx = ttnn.to_torch(indices).to(torch.int64)
print(int(idx.max()), ">= W =", W)  # index 3011 >= 3000, observed in EVERY row
# torch.gather on the logical region also mismatches the returned sorted values
# (a pad-lane index was returned in place of a real-element index).
```

A downstream consumer doing `torch.gather(some_tensor, -1, idx)` (or an on-device gather/scatter) reads out of the logical bounds.

### Mechanism (from code reading of `sort_cross_core_data_exchange.cpp` + `ckernel_sfpu_topk.h`, confirmed on silicon)

- The unstable exchange path uses raw `TTI_SFPSWAP` with hardware index tracking (SFPCONFIG bit 2). On an exact value tie SFPSWAP does **not** swap, so each element's index stays with its DEST slot — **tie resolution is positional**.
- In the cross-core exchange branch, **both peer cores of a spanning tile pair independently run `topk_merge` with swapped operand order**: each loads its LOCAL tile into DEST slot 0 (`input_dest_start`) and the PEER tile into slot 1. The positions of the same two elements are therefore opposite on the two cores.
- Both cores compute the same block direction (`i` and `j` differ only below bit `stage`, so `(i >> stage) & 1` is identical) but keep **opposite halves** (`select_lower = dir ^ (i < j)` flips between the peers).
- Worked example (dir=0, tied lane with `(x, idx_a)` on tile i and `(x, idx_b)` on tile j): core-i's merge leaves `DEST[0]=(x,a), DEST[1]=(x,b)`, `select_lower=1` → keeps `idx_a`; core-j's merge leaves `DEST[0]=(x,b), DEST[1]=(x,a)`, `select_lower=0` → keeps `DEST[1]` = `idx_a`. **Both emit `idx_a`; `idx_b` vanishes.**
- The mechanism is width-independent — it fires whenever CrossCore runs multi-core and a tie spans a pair (`get_number_of_tiles_per_core(USE_AS_MANY_CORES)` = 2 tiles/core at these widths → spanning pairs at every `sub_dist >= 2`). W=4096/8192 are simply the widths the CrossCore factory natively owns in `select_program_factory`.
- Repro B is the same defect escalated by the composite's ±inf pad lanes: pad sentinels tie with each other (and with real ±inf data), and a pad-lane index wins a tie against a real-element index.
- The purely local paths (SingleRowSingleCore; CrossCore's same-core pair branch) move value+index atomically per SFPSWAP and always emit a valid permutation — only the cross-core spanning-pair exchange is affected.

### Expected behavior

`torch.sort(..., stable=False)` makes no promise about *which* index inside a tie group appears at a given output position, but the index tensor must still be a valid permutation of `[0, W)` per row, and every index must be `< W` (logical). Both properties are violated here.

Filing for visibility/tracking — a fix would need either an index-aware compare in the unstable exchange (as the stable comparator does) or a canonical single-peer operand order for spanning pairs.

iwroszTT:
## PR Category

Feature

## Summary

Stable (`stable=True`) topk and sort now return exactly torch's stable tie ordering on Wormhole B0 and Blackhole. Unstable sort indices are now always correct.

Closes #33492 (the issue's repro added as a test). Closes #54043.

Co-authored by: Nachiket Kapre.

Six layers, in review order, bottom-up:

1. **Index-aware comparator: the stable network made correct** (`tt_metal/tt-llk/tt_llk_{blackhole,wormhole_b0}/common/inc/sfpu/ckernel_sfpu_topk.h`, wrappers under `tt_metal/hw/ckernels/*/metal/llk_api/llk_sfpu/`) — the old stable mode incorrectly broke ties with a doubled SFPSWAP. Replaced with a composite (value, index) compare-exchange. Tie polarity is programmed once per kernel from the global sort order; keeping −0.0/+0.0 in one tie class;
2. **Fused-key engine** (same LLK headers; consumed by `topk_multi_core_program_factory.cpp` and the `topk_local`/`topk_final` kernels) — each datum is packed into one `[bf16 value | u16 index]` word and the plain unstable network runs end-to-end: the index half breaks ties for free. E.g., allowing for LLK replay compression. Applicable to bf16 values, u16 indices.
3. **Rank-stamped engine: stable past the u16 index limit (W ≥ 65536)** (LLK stamp/strip primitives; selected in `topk_single_core_program_factory.cpp`) — wide rows need u32 indices, which do not fit the fused word. Local rank tags are stamped into the value words' free low 16 bits, which makes all keys distinct, so the plain unstable network sorts them.
4. **Chain-rank cascade: rank stamping extended to k > 32** (the stamp block in `kernels/compute/topk.cpp`) — re-stamps each accumulator tile every round with that tile's chain-position range. Naive per-level stamps are wrong with k > 32; tests are added.
5. **Stable `ttnn.sort` end-to-end** (`sort_program_factory.cpp`, the three compute kernels under `sort/device/kernels/compute/`, `sort.cpp`) — the comparator network wired through all three sort factories, with a test suite;
6. **Unstable index bugs (#54043)** (CrossCore factory in `sort_program_factory.cpp`; width validation in `sort_device_operation.cpp`) — the CrossCore factory now always runs the index-aware comparator, even for stable=False. When a sorted pair spans two cores, each core merges the same tiles with operands in opposite order, so tie-breaks could disagree, leading to duplicate or out-of-range indices.

Why both the fused-key engine and the rank-stamped are needed?

Fused works by hiding the element's index inside the sort word itself, and a 32-bit word holds only 16 spare bits next to a bf16 value — enough for indices up to 65535. Past that width, or whenever the caller asks for 32-bit index outputs, the true index no longer fits in the word. Rank-stamping is the workaround for exactly that case: the 16 spare bits get short local rank tags instead — just enough to make equal values sort deterministically on the plain network — while the real 32-bit indices travel separately through the index-tracking machinery. The tags are only meaningful within one core's local sort, which is also why rank-stamping cannot go the other way and replace fused on the multicore path: fused's in-word global index is what keeps tie-breaking correct when partial results from different cores merge.

Engine selection for `stable=True`:

| configuration | engine | why |
|---|---|---|
| bf16 values, u16 indices, multicore-eligible | fused-key | the true index rides inside the sort word: cheapest, and correct across cores |
| bf16 values, 32-bit indices (W ≥ 65536, or u32/int32 index outputs) | rank-stamped | local tags make keys distinct; the true u32 indices ride the index-tracking swaps |
| fp32 values | index-aware comparator | the value word has no free bits at all |

## Disclosures (behavior changes)

- No public dtype changes. Index dtypes are unchanged everywhere.
- pow2 width validation at the prim (WH-visible): a direct `ttnn::prim::sort` call with a non-pow2 width (e.g. 192) previously produced silent garbage; now TT_FATALs. The public composite always pads to pow2 and is unaffected.
- Unstable CrossCore sorts: indices are now always a valid permutation (previously duplicates inside tie groups); output equals the torch-stable permutation. fp32 values are canonicalized −0.0 → +0.0 on this path for both stabilities (previously stable-only). Numerically equal; bit-exact consumers of signed zeros will receive +0.0.
- NaN is unsupported (undefined ordering) for bf16 sorts — the datapath canonicalizes NaN to same-sign infinity. Documented, not changed.
- Sampling/TTSampling deliberately stay on the unstable network for decode perf (`_topk_stable = False`); leaving the host-side tie-break machinery

## Follow-up PR

The mergesort row engine, the small-Ht CrossCore reroute, and the single-core fused-key stable sort land in a follow-up stacked on this branch. That PR carries the only public dtype change (stable bf16 → u32 indices at its widths) and all new routing.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:iwroszTT/topk-fused-stable-keys)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:iwroszTT/topk-fused-stable-keys)


nkapreTT:
Downstream #54008 is now rebased onto exact head `6a700bbe6851` and its base-relative pre-commit is green. One parent blocker remains: the current #53557 exact-head pre-commit run reports that `black (llk)` modifies `tt_metal/tt-llk/tests/python_tests/test_topk.py` ([job](https://github.com/tenstorrent/tt-metal/actions/runs/32841707119/job/97782449625)). Please run the LLK black hook (or the full diff-scoped pre-commit) and push the formatting result; I can then restack #54008 again if the base SHA changes.

iwroszTT:
@nkapreTT Thanks for the patience here, and apologies for the slow turnaround — the `black (llk)` failure was fixed back at `e61f152` and pre-commit has been green on every head since, but the branch kept moving while review-readiness work landed.

The branch is merged up to latest main and the head is stable at `6113e887d0b880af2adfac898c506e98b3e43baa`. It was a merge rather than a rebase, so every SHA you previously stacked on is still an ancestor — restacking #54008 should be a single clean rebase onto the new head. No force-push is planned; I'll ping you here first if that ever changes. Thanks again for keeping #54008 in sync!

.github/time_budget.yaml
```diff
@@ -91,7 +91,7 @@ ttnn:
     wh_galaxy: 120
     bh_galaxy: 40
     bh_p100: 805
-    bh_p150b_civ2: 1045
+    bh_p150b_civ2: 1060
```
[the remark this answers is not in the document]
iwroszTT: Time budget change is now reverted

ttnn/cpp/ttnn/operations/data_movement/sort/device/sort_program_factory.cpp
```diff
@@ -1388,6 +1388,15 @@ ttnn::device_operation::ProgramArtifacts SortProgramFactoryCrossCoreDataExchange
...
                 {"ascending", static_cast<uint32_t>(!attributes.descending)},
+                // #54043: this factory ALWAYS runs the index-aware comparator network. The two
+                // peers of a spanning tile pair merge with swapped operand order and keep
+                // opposite halves; a positional (raw SFPSWAP) tie decision can therefore differ
+                // between them, duplicating indices inside tie groups. The comparator's
+                // (value, index) total order is operand-order independent, and a stable ordering
+                // is a valid unstable ordering. Cost is noise here: the factory is
+                // data-movement-bound (~0.3% at W=8192). Pinned host-side so stable=True and
+                // stable=False produce one identical JIT binary.
+                {"stable", 1u},
```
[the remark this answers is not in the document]
iwroszTT: Added a TT_FATAL restricting sort to BH and WH only

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_single_core_program_factory.cpp
```diff
@@ -92,6 +91,21 @@ ttnn::device_operation::ProgramArtifacts TopKDeviceOperation::TopKSingleCoreProg
...
+    // 32-bit indices) sort as [bf16 value | local-rank tag] keys on the plain UNSTABLE network while
+    // the true u32 indices ride the index-tracking swaps — replacing the 7-instruction-per-compare
+    // index-aware comparator. fp32 values keep the comparator (their words have no free low bits for
+    // a tag). Ktiles > 1 (the insertion CASCADE) is covered by chain-rank stamps: each level
+    // re-stamps its accumulator tile with that tile's round-start chain-position range while the
+    // loser tile's tags ride, so a displaced OLD element still outranks NEWER accumulator entries
+    // (see the stamp block in kernels/compute/topk.cpp). The value-side intermediates switch to raw
+    // Float32 transport so the tag bits (and exact bf16 bits) survive the pack/unpack round trips,
+    // exactly like the fused-key engine's packed CBs.
+    const bool rank_stamped_stable = args.stable && !is_fp32_input && !uint16_output;
```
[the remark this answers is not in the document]
iwroszTT: Rejected this case with a TT_FATAL

ttnn/cpp/ttnn/operations/data_movement/sort/sort.cpp
```diff
@@ -274,6 +272,24 @@ std::vector<Tensor> sort(
...
+    // Zero-size tensors: torch.sort returns empty values/indices of the input
+    // shape, so early-exit with empty tensors here. The prim's validate rejects
+    // empty tensors (its program factories cannot build zero-work programs) —
+    // that TT_FATAL stays as the backstop for direct prim callers. Checked
+    // after the dim-range validation above so an out-of-range dim still
+    // raises, matching torch.
+    if (original_lshape.volume() == 0) {
+        auto indices = ttnn::zeros_like(input_tensor, index_dtype);
+        if (operations::data_movement::CMAKE_UNIQUE_NAMESPACE::validate_optional_output_tensors_for_early_exit(
+                optional_output_tensors, original_lshape)) {
```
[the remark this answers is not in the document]
iwroszTT: Fixed

tests/pipeline_reorg/ttsim-skip-list.yaml
```diff
@@ -15,6 +15,11 @@ wormhole_b0:
   - tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py::test_to_memory_config_rm_nd_sharded_to_interleaved_large_row
   - tests/ttnn/unit_tests/operations/conv/test_conv2d.py
   - tests/ttnn/unit_tests/operations/conv/test_conv3d.py
+  # ttsim cannot simulate the index-aware comparator network (SFPSWAP index-operand
+  # gap). Stable sort always runs it, and the all-ones cells route to the CrossCore
+  # factory, which pins it for unstable sorts too.
+  - tests/ttnn/unit_tests/operations/data_movement/test_sort.py::test_sort_stable_index_parity
+  - tests/ttnn/unit_tests/operations/data_movement/test_sort.py::test_sort_unstable_all_ones_permutation
```
mcraigheadTT: Not sure what this is about, but I'm not inclined to approve this not even knowing what it is

mcraigheadTT: Also strange that this is getting added in the middle of a section titled "Job-level timeouts"?

mcraigheadTT: Also, not sure if this is the same as the feature corresponding to the item "Added support for SFPSWAP with ENABLE_DEST_INDEX and VC/VD >= 4" in the release notes for ttsim v1.10.3, but if so, that pointer bump is now in flight: https://github.com/tenstorrent/tt-metal/pull/54496

If it's not the same thing, then I would definitely want to know what this is.

iwroszTT: The first one `test_sort_stable_index_parity` is fixed in the v1.10.3 (it was exactly the SFPSWAP with ENABLE_DEST_INDEX)

The second test looked like a very slow multi-core simulation, but I'm now inclined it's a semaphore sync issue in the kernel. I'm debugging it now.

iwroszTT: The second failure (`test_sort_unstable_all_ones_permutation`) is most likely an existing bug in the sort operation. I've disabled this particular case for BH sim (it passes everywhere else) and filed as a separate issue: https://github.com/tenstorrent/tt-metal/issues/54590

With this and the v1.10.3 update no changes to the ttsim skip list are needed.

mcraigheadTT: Got it, makes sense, thanks

bbradelTT (commented):
Sol analysis:
Blocking: LLK performance schema is incomplete
fused_sort was added to the performance catalog but not wide_schema.DB_SCHEMA. Both Wormhole and Blackhole smoke jobs fail:

[Blackhole failure](https://github.com/tenstorrent/tt-metal/actions/runs/33078206898/job/98538408789)
[Wormhole failure](https://github.com/tenstorrent/tt-metal/actions/runs/33078206898/job/98538408866)
Add fused_sort as a nullable published column or explicitly drop it.

High: public TopK template arguments were silently reordered
Previously:

topk_local_sort<stable_sort, is_fp32_dest_acc_en>
Now fused and rank_stamped precede is_fp32_dest_acc_en. Existing external code such as topk_local_sort<false, true> now enables fused mode instead of fp32 DEST without a compile error. The same problem affects topk_merge and topk_rebuild.

Preserve existing parameter positions and append the new modes.

High risk: missing Quasar rank-stamp wrappers
The architecture-independent API references calculate_topk_stamp_local_positions and calculate_topk_stamp_tile_rank_range, but Quasar defines neither. Quasar has appropriate stubs for fuse/defuse, so equivalent dependent-static_assert stubs should be added for rank stamping.

Current Quasar compile tests pass, but apparently do not compile these API paths.

Correctness trade-off: sampling disables stable TopK globally
TTSampling now sets _topk_stable = False. Its post-gather correction cannot recover the lowest-index maximum when a shard contains more than 32 tied maxima—the source itself documents this limitation.

If lowest-token deterministic greedy behavior is contractual, stable mode should remain enabled on supported routes or tie adjustment must happen before local TopK.

Medium: empty sort(out=...) validates only shape
The new zero-volume early return accepts caller-provided outputs without checking dtype, layout, device, or memory configuration. Wrong outputs are returned unchanged. Reuse normal output validation and add negative tests.

iwroszTT:
> Sol analysis: ...

Addressed all five findings. The point about Sampling - I think the approach taken in this PR is the correct one.

  1. Blocking: LLK performance schema
  It was already fixed, likely concurrently to your review. The PR Gate is now green.
  
  2. High: public TopK template arguments reordered
  Restored the previous order. Tests are still passing.
  
   3. High risk: missing Quasar rank-stamp wrappers
   Added those wrappers for Quasar
   
   4. Sampling disables stable TopK
   Previous to this PR sampling did set stable=True for WH and some cases in BH, even though the result was not stable.
   In all cases, sampling then applied a host-side post-process pass that effectively made its output stable.
   This PR leaves the post-process pass unchanged and sets topk's stable=False in all cases. Note that the now functional stable sort is 2-3x slower in some cases. These cases should be addressed in a follow-up PR (link in description). After that PR it should be safe to replace the host-side pass with a proper stable top-k algorithm.
   
   5. Medium: empty sort(out=...) validates only shape
   Extended the output validation. At the same time fixed pre-existing gaps in input validation.

mtairum (approved):

mbezuljTT:
is there a perf hit?
please run blaze prefill CI

iwroszTT:
> is there a perf hit? please run blaze prefill CI

@mbezuljTT I've started the blaze prefill run: [link](https://github.com/tenstorrent/tt-metal/actions/runs/33385545409)

For the perf impact, this PR does not change the stable=False path. The perf stays the same there.
If some code today applies stable=True (which before this PR is functionally broken) it will see no perf hit with bf16/uint16 values and uint16 indices.
Stable sort/topk on Float32 values and some cases with UINT32 indices are up to 2-3x slower than unstable. There's a follow-up PR which improves the performance of sort/topk in general.

Update: some blaze prefill CI tests (Kimi-K2.6-1T) failed on a timeout due to increased JIT compile time. This PR adds LLK code which caused that. The last commit optimizes LLK header include hierarchy so that we don't include topk code into files that don't need it. In my local tests it now compiles faster than before the PR. I've started a new run of the blaze prefill tests to verify.


@mbezuljTT, timeouts in blaze prefill jobs are fixed now. Unfortunately, I'm still seeing sporadic failures in prefill_runner, kimi_chunked_padded. Although one blaze prefill CI run has completed successfully on this PR ([link](https://github.com/tenstorrent/tt-metal/actions/runs/33677636869)).  I'm looking into the sporadic failures.

bbradelTT (approved):

rdjogoTT (approved):

ttnn/cpp/ttnn/operations/data_movement/sort/device/sort_device_operation.cpp
```diff
@@ -107,20 +117,33 @@ void SortDeviceOperation::validate_on_program_cache_miss(
     // select_program_factory routes UINT16 with Wt > SORT_WT_THRESHOLD to
     // MultiCore to work around that.
 
-    // Width must be a multiple of 64 regardless of layout.
+    // Width must be a power of two >= 64 regardless of layout: the bitonic
+    // engines have no j < Wt partner guard and truncate log2(Wt), so a
+    // non-power-of-two width (e.g. 192 — a multiple of 64) silently produces
+    // garbage rather than failing. The public ttnn.sort composite always pads
+    // the sort dim to the next power of two >= 64 with +/-inf sentinels, so
+    // this only rejects direct prim calls.
```
vtsilytskyiTT: How does `ttnn.sort` discards those `±inf` sentinels? If there are only `+inf`, then it is possible to truncate array by length, but what do we do if valuable data is surrounded by sentinels?

ttnn/cpp/ttnn/operations/data_movement/sort/device/sort_device_operation.cpp
```diff
@@ -107,20 +117,33 @@ void SortDeviceOperation::validate_on_program_cache_miss(
...
     // For TILE the relevant dimension is the padded width; for ROW_MAJOR it is
     // the logical width (padding was already applied in pre_sort_transform_tensor).
     const uint32_t checked_w = is_row_major ? input_lshape[-1] : input_pshape[-1];
     TT_FATAL(
-        checked_w % 64 == 0,
-        "Input shape inner dim {} must be a multiple of 64, pad with +/-infinity if necessary",
+        checked_w >= 64 && (checked_w & (checked_w - 1)) == 0,
```
vtsilytskyiTT:
```suggestion
        checked_w >= 64 && std::has_single_bit(checked_w),
```
Nit: C++20 has a function to verify power-of-2 numbers.
https://en.cppreference.com/cpp/numeric/has_single_bit

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/moe_grouped_topk/device/kernels/compute/moe_grouped_topk.cpp
```diff
@@ -6,6 +6,7 @@
 // normalize -> scale. All stages live in the shared moe_gate_common_compute.hpp blocks, which are also
 // used by moe_hash_gate so the activation/normalize/scale logic exists in exactly one place.
 #include "ttnn/operations/experimental/deepseek_prefill/moe_grouped_topk/device/kernels/compute/moe_gate_common_compute.hpp"
+#include "api/compute/topk.h"
```
vtsilytskyiTT: Why do you need this include?

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_single_core_program_factory.cpp
```diff
@@ -284,14 +298,25 @@ ttnn::device_operation::ProgramArtifacts TopKDeviceOperation::TopKSingleCoreProg
...
 
     KernelSpec compute{
         .unique_id = COMPUTE,
         .source = "ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk.cpp",
         // A compute kernel's legacy default optimization level is O3, while the Metal 2.0
         // default is O2 for every kernel kind, so it has to be stated to keep the level.
-        .compiler_options = {.opt_level = KernelBuildOptLevel::O3},
+        .compiler_options =
+            {.defines = rank_stamped_stable ? KernelSpec::CompilerOptions::Defines{{"TOPK_RANK_STAMPED_STABLE", "1"}}
+                                            : KernelSpec::CompilerOptions::Defines{},
```
vtsilytskyiTT: No need in define, just use compile-time variable `rank_stamped`.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_multi_core_program_factory.cpp
```diff
@@ -36,16 +37,17 @@ namespace ttnn::prim {
  * @param index_tile_size Memory size of index tiles
  * @return Tuple of (num_cores, split_size, remainder, final_input_size, selected_x, selected_y)
  */
-static inline std::tuple<uint16_t, uint16_t, uint16_t, uint16_t, uint16_t, uint16_t> cores_utilized(
-    uint32_t width,
-    uint32_t min_dim,
-    uint32_t max_dim,
-    uint32_t k,
+static inline std::tuple<std::uint16_t, std::uint16_t, std::uint16_t, std::uint16_t, std::uint16_t, std::uint16_t>
```
vtsilytskyiTT: Nit: struct is better than 6-fields-tuple.

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk_local.cpp
```diff
@@ -92,54 +93,72 @@
...
+    constexpr std::uint32_t sorted = get_compile_time_arg_val(13);
     constexpr bool stable_sort = get_compile_time_arg_val(14) == 1;  // Ties keep the lowest index
 
+// Fused-key stable mode (factory-injected define): sort packed [bf16|u16] keys with the unstable
+// network instead of running the comparator-stable network on separate value/index tiles.
+#if defined(TOPK_FUSED_STABLE_KEYS) && TOPK_FUSED_STABLE_KEYS
+    constexpr bool fused_keys = true;
+#else
+    constexpr bool fused_keys = false;
+#endif
```
vtsilytskyiTT: `fused_keys` can be a compile-time argument.

vtsilytskyiTT (approved):
TTNN part looks good.

djordje-tt (approved):

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk_local.cpp
iwroszTT: Changed. It looks simpler now.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_multi_core_program_factory.cpp
iwroszTT: Added `struct TopKCoreDistribution`

ttnn/cpp/ttnn/operations/data_movement/sort/device/sort_device_operation.cpp
iwroszTT: Changed

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/moe_grouped_topk/device/kernels/compute/moe_grouped_topk.cpp
iwroszTT: Removed. The file uses _block_ wrappers, headers of which include the `topk.h` themselves.

ndivnicTT (approved):
LGTM! Thanks for updating LLK perf tests and infra!

ndivnicTT:
@nstojicTT Please have in mind that, in this PR, we got one more column in perf report - fused_sort

tt_metal/tt-llk/tests/python_tests/test_topk.py
```diff
@@ -375,3 +408,436 @@ def test_topk_sfpu(
...
+# Special classes where the two golden hypotheses can diverge
+# (signed_zero, nan_payloads).
+_SPECIAL_CLASSES = tuple(
+    name for name in ADVERSARIAL_STIMULI_CLASSES if name not in _CANON_INVARIANT_CLASSES
+)
+
+# Primary golden hypothesis for the comparator ("stable") engine on the
+# special classes. Flip this single line to "bitexact" if LLK-level
+# characterization shows the comparator sorts bit-exactly.
+_STABLE_SPECIAL_PRIMARY_GOLDEN = "canon"
```
mvlahovicTT: the comment above says the canon behavior was probed at ttnn level but isnt characterized at LLK level, yet this test hard-asserts canon as the primary golden. If the LLK comparator turns out bit-exact, every stable x signed_zero/nan_payloads variant fails and the fix is editing this constant in source. which golden did silicon actually match when you ran these?

tt_metal/tt-llk/tests/sources/eltwise_unary_sfpu_perf.cpp
```diff
@@ -228,7 +228,10 @@ void run_kernel(RUNTIME_PARAMETERS params)
                             ITERATIONS,
                             FAST_MODE,
                             STABLE_SORT,
-                            CLAMP_NEGATIVE>(block_tile, formats.math);
+                            CLAMP_NEGATIVE,
+                            DataFormat::Invalid /* TYPECAST_IN */,
+                            DataFormat::Invalid /* TYPECAST_OUT */,
+                            FUSED_SORT>(block_tile, formats.math);
```
mvlahovicTT, line 234: this passes FUSED_SORT into the network but never calls init_topk_fused or the fuse/defuse sweeps, so the data is never actually packed. fine for timing the network itself, but the fused_sort=Yes row then excludes the per-slab fuse/defuse cost and the replay re-record that fusing forces. maybe add a comment so nobody reads it as end-to-end fused cost.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -193,97 +522,151 @@ inline void bitonic_topk_load16(std::uint32_t dist0, std::uint32_t dist1)
...
+    constexpr std::uint32_t IDX_VD = p_sfpu::LREG4 + (VD & 0x3);
+
+    // Primary key: value compare-exchange.
+    TTI_SFPSWAP(0, VC, VD, MODE);
+
+    // Predicate lanes where compared values are exactly equal. Lanes-on/flags-true CC state
+    // is an entry invariant established once per LLK entry point (see the STABLE_SORT branch
+    // of _bitonic_topk_{phases_steps,merge,rebuild}) and re-established by the trailing
+    // SFPENCC of every comparator body.
+    TTI_SFPXOR(0, VC, VD, 0);
```
mvlahovicTT, line 587: the old sequences double-issued every SFPSWAP, here its a single SFPSWAP with a dependent SFPXOR on the same LREGs right after. SFPSWAP is 2-cycle, and the ISA docs say the BH auto-stall logic misses some SFPSWAP read hazards for modes other than SFPSWAP_MOD1_SWAP, which is exactly the compare modes used here. was this validated on BH silicon, or is there a reason the hazard gap doesnt apply?

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -133,12 +149,319 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+// rank_base + [0, 32): the tile's 32 sequence positions offset by a caller-chosen base. rank_base
+// must be a multiple of 32 and rank_base + 31 must fit 16 bits, so the base ORs into the iota.
+// The k>32 insertion cascade stamps each accumulator tile with its round-start CHAIN position
+// range (32 * level) and the fresh chunk with the top range (32 * output_tiles) exactly once per
+// round; the loser tile's tags then RIDE the cascade unchanged, keeping every tag in the round
+// globally consistent with the true (value, index) order. Same per-vector body and -0.0 fold as
+// the 2-tile local-position stamp below, which is expressed as two calls of this sweep.
+// Clobbers LREG0..2 and the lane enables (left fully enabled).
+template <bool largest>
+inline void _topk_stamp_tile_rank_range_(std::uint32_t dst_tile_index, std::uint32_t rank_base)
```
mvlahovicTT, line 325: the rank_base preconditions (multiple of 32, rank_base + 31 fits 16 bits) are only in the comment. LLK_ASSERT here would protect the k>32 cascade thats going to call this with varying bases.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -25,6 +25,17 @@ namespace sfpu
 
 static std::int32_t topk_replay_init = 0;
 
+// Tie-break polarity for the stable compare-exchange (true = descending / largest first).
+// This is a property of the GLOBAL sort order, not of any one call's sort direction, so the
+// kernel must set it once (after topk init) when STABLE_SORT is used. Default false = ascending.
+// Deliberately not reset by _init_topk.
+static bool topk_stable_descending_mode = false;
+
+TT_ALWAYS_INLINE void set_topk_stable_descending_mode(bool descending)
```
mvlahovicTT, line 34: ambient global that survives _init_topk by design. any future kernel that forgets to call this, or mixes directions, silently gets the wrong tie polarity. did you consider making it a parameter of the stable init instead?

tt_metal/tt-llk/tests/python_tests/test_topk.py
```diff
@@ -375,3 +408,436 @@ def test_topk_sfpu(
...
+#     +0 < positives < +inf < +NaN, i.e. -0 strictly below +0), ties breaking
+#     index-ascending.
+# For sort_mode == "fused" only the canon golden is asserted. For sort_mode ==
+# "stable" on the special classes (signed_zero, nan_payloads) the canon golden
+# is primary, but on a mismatch the bit-exact golden is also evaluated and the
+# failure message reports which golden (if either) matched. The normal-value
+# classes (neg_ties, mixed_sign_ties, tie_straddle_k, infinities) contain only
+# normals and real +-inf, where both engine canonicalization and harness
+# transport are identity, so the canon golden degenerates to a plain
+# torch-stable golden and is asserted strictly for both modes.
```
mvlahovicTT: is this 40 lines comment necessary?

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +152,320 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+//
+// _topk_fuse_tile_ consumes the u16 index tiles at DEST offset 128 (garbage high bits per #50215,
+// masked here) and packs into the value tiles at offset 0; the index region is dead afterwards.
+// _topk_defuse_tile_ restores the pre-fuse layout: [bf16|0x0000] value words in place, u16 indices
+// back at offset 128 (index_store_mode selects raw INT32 or the mode-9 low->high store the packer
+// needs to read u16 from 32-bit DEST). Both are one-shot O(tile) sweeps called explicitly by the
+// kernel (fuse once per fresh slab with the GLOBAL direction — never per network call; defuse once
+// on the final output), and both record over replay slots 0..9, so the fuse poisons
+// topk_replay_init to force the network to re-record its load/store/phase windows.
+// Called from inside the sfpu wrapper (addr_mod_base already 1 — ADDR_MOD_3 = physical incr-0 mod).
```
mvlahovicTT: is this necessary?

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +152,320 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+
+// Stamp ONE value tile (dst_tile_index 0 or 1 within the slab) with sign-conditioned rank tags
+// rank_base + [0, 32): the tile's 32 sequence positions offset by a caller-chosen base. rank_base
+// must be a multiple of 32 and rank_base + 31 must fit 16 bits, so the base ORs into the iota.
+// The k>32 insertion cascade stamps each accumulator tile with its round-start CHAIN position
+// range (32 * level) and the fresh chunk with the top range (32 * output_tiles) exactly once per
+// round; the loser tile's tags then RIDE the cascade unchanged, keeping every tag in the round
+// globally consistent with the true (value, index) order. Same per-vector body and -0.0 fold as
+// the 2-tile local-position stamp below, which is expressed as two calls of this sweep.
+// Clobbers LREG0..2 and the lane enables (left fully enabled).
```
mvlahovicTT: is this necessary?

mvlahovicTT (commented):
please make sure LLK perf CI is green. also i have noticed a lot of unnecessary, over-explained or redundant comments, please make sure to reduce that to the minimum, only add comments if the intention is not clear by looking at code.

tt_metal/tt-llk/tests/python_tests/test_topk.py
iwroszTT: I've ran these LLK tests on Blackhole hardware. With the constant changed to "bitexact" these special cases fail and the error message shows that canon matched, bitexact mismatched.

I've trimmed down the comments and removed the 'fallback mechanism', which was not a fallback really. The bitexact golden, was only used to put additional debug info into the test log.

@nkapreTT

tt_metal/tt-llk/tests/python_tests/test_topk.py
iwroszTT: reduced

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: reduced

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: also reduced

tt_metal/tt-llk/tests/sources/eltwise_unary_sfpu_perf.cpp
iwroszTT: I've added a comment explaining what's measured and which parts are not covered

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: Yes, it was extensively validated on BH silicon.
Looking into the git log history the double-issue pattern was added to implement a stable sort topk behavior (unsuccessfully) see [a0a250a4df3](https://github.com/tenstorrent/tt-metal/commit/a0a250a4df3e377146bcccf971bd6f261a3e46cf) and was later optimized from AABB into ABAB to avoid stalls between instructions targeting the same registers (that is for perf rather than correctness) see [commit](https://github.com/tenstorrent/tt-metal/commit/d162df58f2bcfcd72858a5f682bb993a60b55693).

In isa docs I could find the following caution notes on hazards related to SFPSWAP:
- in [SFPMAD.md](https://github.com/tenstorrent/tt-isa-documentation/blob/main/BlackholeA0/TensixTile/TensixCoprocessor/SFPMAD.md) _If SFPMAD is followed by one of the above cases..._ (e.g. SFPSWAP)  but this code is not using SFPMAD
- is [SFPSWAP.md](https://github.com/tenstorrent/tt-isa-documentation/blob/main/BlackholeA0/TensixTile/TensixCoprocessor/SFPSWAP.md) _Automatic stalling does not apply to SFPSWAP instructions executed as part of an SFPLOADMACRO sequence._ but this code is not using SFMLOADMACRO

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: Fair point. Removed the global, added an `enum class TopkTieOrder`, which is now a statically checked parameter in the ckernel topk functions.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: Asserts added

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,307 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+{
+    TTI_SFPENCC(3, 0, 0, 10);
+    TTI_STALLWAIT(p_stall::STALL_SFPU, p_stall::MATH);
+    set_dst_write_addr(0);
+    TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
+    const std::uint32_t base = dst_tile_index * 64;
+    for (std::uint32_t off = base; off < base + 64; off += 2)
+    {
+        TT_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::INT32, ADDR_MOD_7, off);
+        TT_SFPSTORE(p_sfpu::LREG0, TOPK_SFPSTORE_MODE_PACK_UINT16, ADDR_MOD_7, off);
```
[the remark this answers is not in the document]
iwroszTT: Good catch, thank you. Fixed in f21f1989a4a: the sweep now uses ADDR_MOD_3 inside the SETC16(ADDR_MOD_SET_Base_ADDR32) wrapper, the same way the sibling sweeps in this file do.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -684,12 +1120,54 @@ inline void _bitonic_topk_phases_steps(const int idir, const int i_end_phase, co
...
+
+    static_assert(!(PRE_TAGGED && !RANK_STAMPED), "pre-tagged keys ride the rank-stamped transport");
+    if constexpr (RANK_STAMPED && !PRE_TAGGED)
+    {
+        // Lanes-on FIRST -- the constant programming below goes through the lane-PREDICATED
+        // SFPCONFIG path and transiently clobbers LREG0 (see _topk_fuse_tile_), so it must
+        // run before any load and under fully enabled lanes.
+        TTI_SFPENCC(3, 0, 0, 10);
+        sfpi::vConstIntPrgm0 = 0x0000FFFF;                    // LREG12: tag complement operand
+        _sfpu_load_config32_(p_sfpu::LREG11, 0xFFFF, 0x0000); // lo16 clear mask (SFPAND -- no loads mid-stamp)
```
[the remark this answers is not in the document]
iwroszTT: Thanks. Fixed in 0372af9c716: the lo16 mask now lives in LREG14, and the _init_topk_rank_stamped_ comment is corrected. A regression test (test_topk_const_leak.py) probes -1.0f after the merge.

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_topk.h
```diff
@@ -707,6 +765,11 @@ inline void calculate_bitonic_topk_rebuild(
 // UInt16-in-32b-DEST prepare helper: no-op on Quasar (workaround only required for WH/BH).
 // Defined so compute_kernel_api.h resolves the symbol on all architectures.
 inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t /*tile_index*/) {}
+inline void _topk_uint16_move_dest_tile_to_pack_half_(std::uint32_t /*tile_index*/) {}
+// Rank-stamped strip stub: the mode itself is static_asserted off on Quasar (see topk_init /
+// the network entries), so this is unreachable; it exists only so the arch-independent
+// compute API (topk_strip_rank_tags) compiles.
+inline void _topk_strip_rank_tags_(std::uint32_t /*dst_tile_index*/) {}
```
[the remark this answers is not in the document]
iwroszTT: Thanks, stub added in f21f1989a4a.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -507,15 +851,79 @@ inline void bitonic_topk_inc_x4_dest(std::uint32_t inc, bool cr)
...
+    {
+        // Establish the lanes-on/flags-true CC entry invariant once; every stable comparator
+        // body re-establishes it via its trailing SFPENCC, and the intervening loads/stores/
+        // transposes/SFPCONFIG writes preserve CC state.
+        TTI_SFPENCC(3, 0, 0, 10);
+        if constexpr (is_fp32_dest_acc_en && !TOPK_UINT16_IN_FP32_DEST)
+        {
+            // fp32-family values in 32-bit DEST can carry -0.0: fold it into the +0.0 tie
+            // class before the sign-magnitude network runs (see the sweep's doc comment).
+            topk_stable_canonicalize_negzero_value_tiles();
```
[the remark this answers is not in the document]
iwroszTT: You are right, and the ~0.3% figure was wrong: the sweep costs about 646 cycles per local sort. Hoisted in 067790aae5e. It is now a caller-invoked entry point (topk_canonicalize_negzero_values), called once per fresh slab in every kernel, and the network no longer runs it. The stable local-sort perf row drops from 1,255,104 to 1,172,400 cycles.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,307 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+        for (int parity = 0; parity < 2; parity++) // even / odd Dst columns
+        {
+            TTI_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::INT32, ADDR_MOD_3, 0);
+            TTI_SFPLOADI(p_sfpu::LREG0, sfpi::SFPLOADI_MOD0_LOWER, 0); // clear stale lo16 -> [bf16|0]
+            // -0.0 -> +0.0: (w << 1) == 0 exactly for the two zero encodings.
+            TTI_SFPMOV(0, p_sfpu::LREG0, p_sfpu::LREG1, 0);
+            TTI_SFPSHFT(1, 0, p_sfpu::LREG1, 1);
+            TTI_SFPSETCC(0, p_sfpu::LREG1, 0, sfpi::SFPSETCC_MOD1_LREG_EQ0);
+            TTI_SFPLOADI(p_sfpu::LREG0, sfpi::SFPLOADI_MOD0_UPPER, 0); // zero lanes -> +0.0
+            TTI_SFPENCC(3, 0, 0, 10);
```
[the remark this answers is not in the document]
iwroszTT: Thanks, nice idea. It relies on SFPMAD returning +0.0 for 1.0 * (-0.0) + 0.0 on silicon, so I would like to confirm that first. Leaving it as a follow-up.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,307 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
     }
 }
 
-template <bool is_fp32_dest_acc_en>
+// Ungated variant for the fused-key final extraction: a u16 datum transposed into 32-bit DEST
+// lands in the low half with stale garbage above it, while the packer reads the high half.
+// Strip the garbage and move the datum up (SFPSTORE mode 9). Runs on MATH while DEST is acquired,
+// after the transpose has drained. Called outside sfpu_start: enable addr_mod_base for ADDR_MOD_3.
+inline void _topk_uint16_move_dest_tile_to_pack_half_(std::uint32_t dst_tile_index)
```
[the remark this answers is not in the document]
iwroszTT: Agreed, this most likely folds under IPA-CP, so leaving it as is. Happy to switch to the runtime-base loop if a size check shows otherwise.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,307 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+// fold -0.0 into the +0.0 tie class on the way (torch treats +-0 as ONE tie
+// class broken by index; the raw sign-magnitude compare would otherwise order
+// all -0.0 strictly below all +0.0). Every datum enters the sort through
+// exactly one local-sort call, so this single sweep canonicalizes the whole
+// sort. Clobbers LREG0..2 and the lane enables (left fully enabled).
+template <bool largest>
+inline void _topk_stamp_local_positions_()
+{
+    _topk_stamp_tile_rank_range_<largest>(0, 0);
+    _topk_stamp_tile_rank_range_<largest>(1, 32);
```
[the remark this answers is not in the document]
iwroszTT: Done in f21f1989a4a, on both architectures.

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk.cpp
```diff
@@ -137,9 +144,20 @@ void kernel_main() {
...
+    // The rank tag IS the stable tie-break; the network itself runs unstable in rank-stamped mode.
+    constexpr bool network_stable = stable_sort && !rank_stamped;
+
     // Initialize kernel components
     compute_kernel_hw_startup(input_val_dfb_index, input_ind_dfb_index, output_val_dfb_index);
-    ckernel::topk_tile_init();
+    ckernel::topk_tile_init<false, rank_stamped>();
+    // Tie order follows the GLOBAL sort order, never the per-call idir.
+    constexpr auto tie_order =
+        (largest != 0) ? ckernel::TopkTieOrder::Descending : ckernel::TopkTieOrder::Ascending;
```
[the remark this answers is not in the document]
iwroszTT: Done in f21f1989a4a: topk_tie_order_from_global_direction in api/compute/topk.h, used at all six sites.

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk.cpp
```diff
@@ -353,7 +374,31 @@ void kernel_main() {
...
+                    // transport), so a displaced OLD element still outranks NEWER accumulator
+                    // entries. The stamp also folds -0.0 into +0.0; every fresh datum passes
+                    // through exactly one stamp (level 0 / first sort) before its first compare.
+                    if (first_sort_from_transposed) {
+                        // Both tiles fresh (width chunks 0 and 1): plain positions [0, 64).
+                        ckernel::topk_stamp_local_positions<largest != 0>(0);
+                    } else {
+                        ckernel::topk_stamp_tile_rank_range<largest != 0>(0, 0, 32 * cascade_level);
+                        if (cascade_level == 0) {
+                            ckernel::topk_stamp_tile_rank_range<largest != 0>(0, 1, 32 * output_tiles);
```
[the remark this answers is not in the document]
iwroszTT: Done in f21f1989a4a.

tt_metal/hw/inc/api/compute/topk.h
```diff
@@ -0,0 +1,365 @@
...
+/**
+ * Clears the low 16 bits (stale rank tags) of one rank-stamped value tile in DST, leaving exact
+ * [bf16|0x0000] words so the following Float32->bf16 pack cannot RNE-round on tag bits. Must run
+ * on MATH while DEST is still acquired, after the final transpose back to row layout (same
+ * calling convention as topk_uint16_move_dest_tile_to_pack_half).
+ */
+ALWI void topk_strip_rank_tags(std::uint32_t idst) { MATH((ckernel::sfpu::_topk_strip_rank_tags_(idst))); }
+
+// Before packing a uint16 index tile that lives in 32-bit DEST as a plain integer.
+ALWI void topk_finalize_uint16_indices(std::uint32_t idst) { MATH((ckernel::sfpu::_topk_finalize_hi16_index_tile_(idst))); }
```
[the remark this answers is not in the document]
iwroszTT: Done in f21f1989a4a.

tt_metal/tt-llk/tests/python_tests/test_topk.py
```diff
@@ -375,3 +408,334 @@ def test_topk_sfpu(
...
+
+@parametrize(
+    formats=input_output_formats(
+        [
+            DataFormat.Float16_b,
+        ]
+    ),
+    # Single 2-tile slab per stage; also satisfies the fused-mode W == 128
+    # constraint, so no mode is skipped.
+    input_dimensions=[[32, 128]],
```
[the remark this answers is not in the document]
iwroszTT: Done in f21f1989a4a: formats, input_dimensions and K are now hardcoded in the test body.

tt_metal/tt-llk/tests/python_tests/test_topk.py
mvlahovicTT: the _canon_bits docstring still says the model comes from ttnn-level silicon while the block header now says it was measured on Blackhole at LLK level, please make the two agree

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
mvlahovicTT: _bitonic_topk_phases_steps, _bitonic_topk_merge and _bitonic_topk_rebuild still accept TIE_ORDER = Unset together with STABLE_SORT, and topk_cmp_swap_stable_min_to_vd treats Unset as Ascending, so the silent default survived below the compute API. The LLK test harness in sfpu_operations.h hits it on every stable perf row. Please add `static_assert(!STABLE_SORT || TIE_ORDER != TopkTieOrder::Unset, ...)` to those three entry points and pass an explicit order from sfpu_operations.h

tt_metal/tt-llk/tests/python_tests/perf_eltwise_unary_sfpu.py
```diff
@@ -68,8 +70,15 @@
     MathOperation.TopKRebuild,
 }
 
+# The defuse sweep belongs to the fused engine only. Fused rows: the local-sort row includes the
```
mvlahovicTT: after the last commit the stable fp32 dest_acc=Yes local sort row no longer includes the -0.0 canonicalization sweep, while the fused row does include its per slab fuse sweep. both sweeps are something every real kernel runs once per fresh slab before the first local sort, so the two rows now follow different rules, and the stable rows baseline moved by -6.6% with that commit. either call the canonicalize sweep in the stable branch of sfpu_operations.h the same way the fuse sweep is called (line 1092), or say here that the row is network only and note the baseline change.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,310 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+    TTI_STALLWAIT(p_stall::STALL_CFG, p_stall::WAIT_SFPU);
+    TTI_SETC16(ADDR_MOD_SET_Base_ADDR32, 0);
+    set_dst_write_addr(0);
+}
+
+// uint16 index tile in 32-bit DEST: rotate the [0|idx] integer into the high half the packer reads.
+inline void _topk_finalize_hi16_index_tile_(std::uint32_t dst_tile_index)
+{
+    TTI_SFPENCC(3, 0, 0, 10);
+    TTI_SETC16(ADDR_MOD_SET_Base_ADDR32, 1);
```
mvlahovicTT, line 473: this fixes a path that packed garbage on wh, but the commit message only lists bh runs. no LLK test reaches this sweep - the only caller is the deepseek MoE grouped topk kernel, covered by a ttnn nightly test. was this run on wh silicon, and if so can you add the result to the PR? a small LLK test for the finalize sweep would also close the gap that let the encoding slip through.

tt_metal/hw/inc/api/compute/topk.h
```diff
@@ -0,0 +1,390 @@
...
+#include <cstdint>
+#include "api/compute/compute_kernel_api.h"
+#ifdef TRISC_MATH
+#include "ckernel_sfpu_topk.h"
+#endif
+
+namespace ckernel {
+
+// Mirrors ckernel::sfpu::TopkTieOrder, which is math-TRISC only.
+enum class TopkTieOrder : std::uint8_t { Unset, Ascending, Descending };
```
mvlahovicTT, line 16: nit: this enum is mirrored in ckernel::sfpu for bh, wh and the qsr wrapper and bridged with a static_cast. a static_assert that the enumerator values match would catch a reorder in any of the four copies.

tt_metal/tt-llk/tests/python_tests/perf_eltwise_unary_sfpu.py
iwroszTT: I've changed the test. The stable branch now runs the -0.0 canonicalization sweep.

tt_metal/hw/inc/api/compute/topk.h
iwroszTT: done

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: added

tt_metal/tt-llk/tests/python_tests/test_topk.py
iwroszTT: done

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
iwroszTT: Added the LLK test: test_topk_finalize_uint16.py

mvlahovicTT (approved):

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk_common_funcs.hpp
```diff
@@ -49,69 +67,98 @@ void process_and_sort_tiles(
         if (tiles_to_wait == 2) {
             transpose_tile(index_dfb_index, 1, 3);
         }
+        if constexpr (fused) {
+            // Pack the slab into [bf16|u16] keys once, with the GLOBAL polarity; DEST 2,3 are dead
+            // afterwards and the network below runs unstable on the packed words.
+            ckernel::topk_fuse_tile<fused_largest>(0);
```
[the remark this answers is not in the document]
iwroszTT: The code is correct. Added a comment in the code and test: test_topk_stable_multicore_fused_signed_zero

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -507,15 +854,72 @@ inline void bitonic_topk_inc_x4_dest(std::uint32_t inc, bool cr)
...
+// tie class broken by index. Rewrite -0.0 -> +0.0 in the two freshly loaded value tiles: the kernel
+// runs this once per fresh 2-tile slab before that slab's first local sort, and the fp32
+// pack/unpack transport is a bit identity, so later local sorts and merges see canonical data. The 16-bit-DEST
+// engines need no sweep (their bf16 SrcA datacopy already canonicalizes +-0, silicon-
+// probed), uint16-in-fp32-dest values carry no live sign bit after their own strip, and
+// fused packed keys never take the comparator-stable path.
+// Predicate: (x & 0x7FFFFFFF) == 0 (zero magnitude); action: x &= 0x7FFFFFFF (-> +0.0).
+inline void _topk_canonicalize_negzero_value_tiles_()
+{
+    sfpi::vConstIntPrgm0 = 0x7FFFFFFF;
```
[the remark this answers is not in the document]
iwroszTT: Fixed

tt_metal/hw/inc/api/compute/topk.h
```diff
@@ -0,0 +1,397 @@
...
+ * | fused           | Sort packed [bf16 value | u16 index] keys with the unstable network        | bool     | true, false                                           | False    |
+ * | rank_stamped    | Sort [bf16 value | rank tag] keys with the unstable network (u32 indices)  | bool     | true, false                                           | False    |
+ */
+// clang-format on
+template <
+    bool stable_sort = false,
+    bool is_fp32_dest_acc_en = DST_ACCUM_MODE,
+    bool fused = false,
+    bool rank_stamped = false,
+    TopkTieOrder tie_order = TopkTieOrder::Unset>
```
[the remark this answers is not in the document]
iwroszTT: Fixed

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -197,275 +514,305 @@ inline void bitonic_topk_load16(std::uint32_t dist0, std::uint32_t dist1)
...
+// by index. VD ^= VC leaves 0 only in tied lanes, providing the tie predicate; the second XOR
+// restores VD. INDEX_MIN_TO_VD selects the index-swap operand order to match the sort direction.
+template <std::uint32_t VC, std::uint32_t VD, std::uint32_t MODE, bool INDEX_MIN_TO_VD>
+TT_ALWAYS_INLINE void topk_cmp_swap_stable_directional()
+{
+    constexpr std::uint32_t IDX_VC = p_sfpu::LREG4 + (VC & 0x3);
+    constexpr std::uint32_t IDX_VD = p_sfpu::LREG4 + (VD & 0x3);
+
+    // Primary key: value compare-exchange.
+    TTI_SFPSWAP(0, VC, VD, MODE);
```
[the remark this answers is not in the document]
iwroszTT: This would require detailed perf analysis, e.g. with perf counters to measure stall rate. Skipped.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_topk.h
```diff
@@ -136,13 +150,310 @@ inline void topk_uint16_prepare_value_tile_for_pack(std::uint32_t dst_tile_index
...
+// they are transposed into DEST and before their local sort; defuse each output tile once.
+// Called from inside the sfpu wrapper (addr_mod_base already 1, so ADDR_MOD_3 is the incr-0 mod).
+template <bool largest>
+inline void _topk_fuse_tile_()
+{
+    // Lanes-on FIRST: the constant programming below goes through the SFPCONFIG path, which is
+    // lane-PREDICATED (and clobbers LREG0 transiently) — programmed under a partially-enabled
+    // ambient CC state, disabled lanes would keep stale LREG12 bits and the mask/complement would
+    // silently misfire in exactly those lanes.
+    TTI_SFPENCC(3, 0, 0, 10);
```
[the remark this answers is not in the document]
iwroszTT: Added a macro applied in those 34 sites

ttnn/cpp/ttnn/operations/reduction/topk/device/kernels/compute/topk_common_funcs.hpp
```diff
@@ -163,26 +215,31 @@ void process_tiles(
...
+                reconfig_data_format_srca(input_transposed_dfb_index, index_transposed_dfb_index);
+                copy_init(index_transposed_dfb_index);
+                copy_tile(index_transposed_dfb_index, left_tile_id, index_dest_start);
+                copy_tile(index_transposed_dfb_index, right_tile_id, index_dest_end);
+            }
 
             // merge values - move larger 32 values into 0th dest and lower 32 values into 1st dest
             if (largest) {
-                ckernel::topk_merge<false, stable_sort>(0, m_iter, K);
+                ckernel::topk_merge<false, stable_sort, DST_ACCUM_MODE, fused, false, false, tie_order>(0, m_iter, K);
```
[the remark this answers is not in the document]
iwroszTT: Added annotations

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/moe_grouped_topk/device/kernels/compute/moe_gate_common_compute.hpp
```diff
@@ -170,7 +171,11 @@ void process_and_sort_tiles(
         transpose_tile(cb_expert_index_template_id, wt + 1, 3);
 
         // llk_topk_sort -> inplace
-        ckernel::topk_local_sort<stable_sort>(0, (int)ascending, end_phase);
+        if constexpr (stable_sort) {
+            ckernel::topk_canonicalize_negzero_values(0);
+        }
+        ckernel::topk_local_sort<stable_sort, DST_ACCUM_MODE, false, false, ckernel::TopkTieOrder::Descending>(
```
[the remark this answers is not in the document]
iwroszTT: Fixed. Added constexpr

tt_metal/hw/inc/api/compute/topk.h
```diff
@@ -0,0 +1,397 @@
...
+ * |                 | true-index fuse) — run the rank-stamped transport without re-stamping      |          |                                                       |          |
+ */
+// clang-format on
+template <
+    bool idir = false,
+    bool stable_sort = false,
+    bool is_fp32_dest_acc_en = DST_ACCUM_MODE,
+    bool fused = false,
+    bool rank_stamped = false,
+    bool pre_tagged = false,
```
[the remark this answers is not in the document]
iwroszTT: Removed

nmilicevicTT (approved):

kstevensTT (approved):

ttnn/cpp/ttnn/operations/data_movement/sort/device/kernels/compute/sort_single_row_multi_core.cpp
```diff
@@ -23,10 +24,10 @@ void kernel_main() {
     constexpr uint32_t number_of_available_cores = get_arg(args::number_of_available_cores);
     constexpr uint32_t compute_with_storage_grid_size_x = get_arg(args::compute_with_storage_grid_size_x);
     constexpr bool descending = get_arg(args::descending);
-    constexpr bool stable =
-        get_arg(args::stable);  // TODO: In the future change LLK to have the option or add additional step with
-                                // checking values and indexes after the sorting
-                                // Issue: https://github.com/tenstorrent/tt-metal/issues/20625
+    // Comparator-stable network (issue #33492): on exact value ties the index tiles are
```
scardozaTT, line 27: nit: the Issue# reference here made me think it's an open issue, but I believe this PR would close it, so consider removing the Issue# ?

scardozaTT (approved):

sjameelTT (approved):

PR #53564, opened 2026-08-18
[Feature] #43050 - Generic reduce: DRAM-sharded I/O, and batch NoC reads in reduce readers

### LINKED ISSUE #43050: Reduce ops to support DRAM sharded tensors
Reduce ops currently don't officially support DRAM sharded tensors, as documented in https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/generic_reductions_nanobind.hpp#L62 and https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/std_var_reductions_nanobind.cpp#L61.
Some scenarios may work because we rely on standard APIs that already support DRAM sharding, but not all corner cases are covered.
We should add support and comprehensive tests for DRAM sharding.

When adding this, make sure to address this comment: https://github.com/tenstorrent/tt-metal/pull/42120#discussion_r3133352489

ign-msati:
### PR Category
Feature — DRAM-sharded input and output for generic reduce + batched reader reads (`sum`/`mean`/`max`/`min`/`std`/`var`), issue [[#43050](https://github.com/tenstorrent/tt-metal/issues/43050)]

## Summary

Generic reduce now accepts DRAM-sharded inputs and outputs.

DRAM sharding uses the same `ShardSpec`, but its grid names DRAM banks — a single row at `y=0` — rather than worker cores. Three places in reduce read that grid as worker cores, so each now branches on buffer type:

- **Grid borrowing** — when a sharded output config has no `shard_spec`, `build_reduce_output_tensor_spec` copies the input's grid. It now rejects that when the two sides are in different memory spaces, on the legacy and ND branches alike.
- **Tensix-grid containment** — checking a shard grid against the compute grid is meaningless for bank ids, so it applies to L1 only. `validate_buffer_parameters` already enforces DRAM's bank-grid rules when the buffer is created.
- **The sharded fast paths** — the H factory's width-sharded path and the W factory's height-sharded path both back their circular buffers with the tensor's own buffer, which only L1 can do, so each now needs L1 on both sides. DRAM tensors take the generic `TensorAccessor` path.

Covers `dim=-1`, `dim=-2` and `dim=(-2,-1)`. `std`/`var` need no change: they share the same validator and have no fast path of their own.

Two changes ride along, both on the generic path that DRAM tensors now take:

- **Batched reader reads** — both reduce readers issued one `async_read` per barrier, so every tile paid the full read latency. They now issue a batch behind one barrier (4 tiles for the universal reader, one DEST chunk for the column reader) with the input CB sized at two batches, keeping every reservation contiguous. A program whose smallest core owns less than a batch stays unbatched, as does the fused-negate W path with a single row per core.
- **Negate L1 gate** — `h_reduce_negate_fits_in_l1` now sizes the program the factory actually builds, so `min` stops falling back to `neg + reduce + neg` when the fused kernel fits, and stops asserting when it doesn't.

## Changes Made

**common.cpp / common.hpp**

- `validate_reduce_sharded_buffer_types` accepts DRAM as well as L1 on both sides.
- `build_reduce_output_tensor_spec` rejects borrowing a shard grid across buffer types. The check runs after confirming the input has a grid at all, so an interleaved input still gets the `requires either nd_shard_spec or shard_spec` message.
- `h_reduce_negate_fits_in_l1` takes an `output_mem_config` and mirrors the H factory's fast-path gate, so its buffer sizing matches the program the factory actually builds.
- `kReduceReaderTilesPerBatch`, `reduce_reader_batch` and `reduce_reader_input_cb_tiles` derive the reader batch and input CB depth from the smallest core's tile count.

**reduce_op.cpp**

- Both `h_reduce_negate_fits_in_l1` call sites pass `output_mem_config`.
- The external-negate fallback's final `neg` inherits `h_out` rather than `output_mem_config`. Only `h_out` has a shard shape recomputed for the reduced height.

**Device operation — reduce_op_device_operation.cpp**

- The input-side and output-side grid-containment checks apply to L1 only.

**Reader kernels**

- `reader_unary_reduce_universal_start_id.cpp` — a single loop issues `tiles_per_batch` reads per barrier and publishes them with one `push_back`; the short final batch takes the same path.
- `reader_unary_transpose_wh_universal_input_cols_partitioned.cpp` — reads a whole column chunk per barrier.

**Program factories**

- `reduce_op_multi_core_h_program_factory.cpp` — `use_width_sharding` also requires `is_l1()` on both sides, so DRAM width-sharded tensors use the generic branch. The reader batches a DEST chunk when the smallest core owns one.
- `reduce_op_multi_core_w_program_factory.cpp` — `use_height_sharding` also requires `is_l1()`. The reader batches when the smallest core streams a full batch (`rows/core × Wt`), except on the fused-negate path with a single row per core, which measures 18-32% slower batched.
- `reduce_op_single_core_hw_program_factory.cpp` / `welford_reduce_program_factory.cpp` — pass the unbatched reader.

**Front end — generic_reductions_nanobind.hpp / std_var_reductions_nanobind.cpp**

- Memory Support gains `Sharded (DRAM): Width, Height, and ND sharding` and an explicit unsupported list. The L1 line gains `Block`, which already worked but was undocumented.

## Tests

**test_reduction_sharded_topology.py** (new, 101 cases)

- `test_reduce_dram_sharded` / `_full_hw_reduce` / `_dtypes` — DRAM-sharded in+out across every op, shard layout, dim and dtype. `_full_hw_reduce` covers every output config the op accepts — interleaved, DRAM-sharded and L1-sharded — since a 1×1 result can stay sharded.
- `test_reduce_dram_sharded_full_bank_width_h_reduce` — a WIDTH_SHARDED grid spanning every DRAM bank. On Wormhole that grid is 12 wide against an 8-wide Tensix grid, so it only passes with the containment check gated to L1.
- `test_reduce_h_width_sharded_mixed_l1_dram` and `_l1_and_dram_use_distinct_programs` — the fast-path gate must key off buffer type per side, and must reach the program-cache key.
- `test_reduce_w_height_sharded_mixed_l1_dram` — the W twin, on a geometry whose shards tile the tensor exactly so buffer type is the only term left deciding the fast path; the geometry itself is asserted so the DRAM cases cannot silently stop covering the gate.
- `test_reduce_dram_nd_sharded` — DRAM ND sharding; the shard shape splits C as well as H and W, so the layout cannot normalize to a legacy one.

#### Perf: batched reader reads

Wormhole B0, tile layout, device kernel duration. PR = median of 50 iterations; baseline = median of 27 (3 runs × 9) on a fresh build of the base commit. 
| regime                      | cases | median Δ% | range           |
| --------------------------- | ----: | --------: | --------------- |
| W reader batches            |    28 |    −10.4% | −31.8% .. +4.2% |
| H reader batches            |     4 |     −6.7% | −7.6% .. −6.5%  |
| reader unbatched or Welford |    51 |     −0.0% | −1.8% .. +1.3%  |

<!DOCTYPE html>
S. No | op | dim | config (layout/dtype, in → out) | shape | main µs | PR µs | Δ% | batched by | prog
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | sum | -2 | tile/bf16, WS-L1 → WS-L1 | 1×1×1024×256 | 6.63 | 6.63 | (+0.09%) | H, no batch | Reduce
2 | sum | -1 | tile/bf16, WS-L1 → WS-L1 | 1×1×1024×256 | 14.6 | 14.7 | (+0.65%) | W batched | Reduce
3 | sum | -1 | tile/bf16, WS-L1 → WS-L1 | 1×1×256×2048 | 31.9 | 25.9 | −18.87% | W batched | Reduce
4 | max | -1 | tile/bf16, WS-L1 → WS-L1 | 1×1×256×2048 | 31.9 | 25.7 | −19.70% | W batched | Reduce
5 | sum | -1 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 6.25 | 6.25 | (−0.01%) | W, no batch | Reduce
6 | sum | -2 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 18.19 | 18.19 | (+0.01%) | H, no batch | Reduce
7 | min | -1 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 19.64 | 19.63 | (−0.03%) | W, no batch | Reduce
8 | min | -2 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 25.19 | 25.18 | (−0.06%) | H, no batch | Reduce
9 | std | -1 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 107.17 | 107.23 | (+0.06%) | welford | WelfordReduce
10 | std | -2 | tile/bf16, HS-L1 → HS-L1 | 1×1×1024×256 | 420.54 | 420.47 | (−0.02%) | welford | WelfordReduce
11 | sum | -1 | tile/bf16, BS-L1 → BS-L1 | 1×1×512×512 | 8.4 | 8.5 | (+0.85%) | W batched | Reduce
12 | sum | -2 | tile/bf16, BS-L1 → BS-L1 | 1×1×512×512 | 12.9 | 13.0 | (+0.39%) | H, no batch | Reduce
13 | max | -1 | tile/bf16, BS-L1 → BS-L1 | 1×1×512×512 | 8.4 | 8.5 | +1.22% | W batched | Reduce
14 | std | -1 | tile/bf16, BS-L1 → BS-L1 | 1×1×512×512 | 212.2 | 212.3 | (+0.06%) | welford | WelfordReduce
15 | std | -2 | tile/bf16, BS-L1 → BS-L1 | 1×1×512×512 | 212.3 | 213.7 | (+0.67%) | welford | WelfordReduce
16 | sum | -1 | tile/bf16, ND-L1 → ND-L1 | 1×4×128×128 | 4.7 | 4.9 | +3.96% | W batched | Reduce
17 | sum | -2 | tile/bf16, ND-L1 → ND-L1 | 1×4×128×128 | 5.13 | 5.10 | (−0.58%) | H, no batch | Reduce
18 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 7.5 | 6.3 | −15.47% | W batched | Reduce
19 | sum | HW | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 28.5 | 27.2 | −4.36% | W batched + H, no batch | Reduce+Reduce
20 | mean | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 7.5 | 6.4 | −14.23% | W batched | Reduce
21 | mean | HW | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 24.9 | 23.5 | −5.51% | W batched + H, no batch | Reduce+Reduce
22 | max | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 7.5 | 6.4 | −13.96% | W batched | Reduce
23 | max | HW | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 24.8 | 23.5 | −5.10% | H, no batch | Reduce+Reduce
24 | min | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 8.32 | 8.31 | (−0.17%) | W, no batch | Reduce
25 | min | -2 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 25.6 | 25.5 | (−0.12%) | H, no batch | Reduce
26 | min | HW | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×256 | 33.2 | 33.4 | (+0.35%) | H, no batch | Reduce+Reduce
27 | sum | -1 | tile/fp32, IL-DRAM → IL-DRAM | 1×1×1024×256 | 12.1 | 12.6 | +4.22% | W batched | Reduce
28 | min | -1 | tile/fp32, IL-DRAM → IL-DRAM | 1×1×1024×256 | 12.2 | 12.3 | (+0.97%) | W batched | Reduce
29 | min | -2 | tile/fp32, IL-DRAM → IL-DRAM | 1×1×1024×256 | 27.7 | 27.8 | (+0.40%) | H, no batch | Reduce
30 | sum | -1 | tile/bf8, IL-DRAM → IL-DRAM | 1×1×1024×256 | 5.9 | 4.1 | −31.78% | W batched | Reduce
31 | min | -1 | tile/bf8, IL-DRAM → IL-DRAM | 1×1×1024×256 | 6.81 | 6.80 | (−0.26%) | W, no batch | Reduce
32 | min | -2 | tile/bf8, IL-DRAM → IL-DRAM | 1×1×1024×256 | 24.68 | 24.68 | (+0.02%) | H, no batch | Reduce
33 | min | -1 | tile/bf16, IL-L1 → IL-L1 | 1×1×1024×256 | 6.94 | 6.89 | (−0.68%) | W, no batch | Reduce
34 | min | -2 | tile/bf16, IL-L1 → IL-L1 | 1×1×1024×256 | 25.21 | 25.20 | (−0.01%) | H, no batch | Reduce
35 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×2048 | 47.2 | 42.3 | −10.37% | W batched | Reduce
36 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×8192 | 178.0 | 161.3 | −9.36% | W batched | Reduce
37 | max | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×2048 | 47.3 | 42.3 | −10.45% | W batched | Reduce
38 | sum | -2 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×8192 | 171.5 | 160.2 | −6.54% | H batched | Reduce
39 | sum | -2 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×1024×16384 | 174.5 | 162.4 | −6.91% | H batched | Reduce
40 | sum | -2 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×512×32768 | 179.6 | 165.8 | −7.65% | H batched | Reduce
41 | mean | -2 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×8192 | 171.4 | 160.3 | −6.50% | H batched | Reduce
42 | sum | HW | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×2048 | 88.1 | 83.1 | −5.62% | W batched + H, no batch | Reduce+Reduce
43 | sum | -1 | tile/fp32, IL-DRAM → IL-DRAM | 1×1×2048×2048 | 91.5 | 89.0 | −2.74% | W batched | Reduce
44 | sum | -1 | tile/bf16, IL-L1 → IL-L1 | 1×1×2048×2048 | 46.4 | 40.4 | −12.94% | W batched | Reduce
45 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×128 | 7.6 | 5.7 | −25.01% | W batched | Reduce
46 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×256 | 9.9 | 7.9 | −20.99% | W batched | Reduce
47 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×512 | 15.0 | 12.8 | −14.48% | W batched | Reduce
48 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×1024 | 25.7 | 22.5 | −12.42% | W batched | Reduce
49 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×2048 | 47.4 | 42.3 | −10.81% | W batched | Reduce
50 | sum | -1 | tile/bf16, IL-DRAM → IL-DRAM | 1×1×2048×4096 | 93.0 | 83.3 | −10.49% | W batched | Reduc

#### Perf: negate L1 gate

`ttnn.min` lowers to `-max(-x)` for bf16 and bf8 inputs (int32, and fp32 on the accurate SFPU path, drive a direct MIN instead). Those negations can be applied two ways:

- **Fused** — `reduce_h_neg` negates inline: one device op, at the cost of two circular buffers, `cb_acc` and `cb_ineg`, each `Ht × per-core Wt` tiles.
- **External** — `neg` over the input → plain reduce → `neg` over the output: three device ops, the first reading and rewriting the whole tensor.

`h_reduce_negate_fits_in_l1()` previously used the width-sharded sizing whenever the input was WIDTH_SHARDED L1, even when an interleaved output forced the factory to build the split-across-grid program; matching the estimator to the actual program reduced the estimated CB requirement by 2–4× in the measured cases and enabled the faster fused path.

<!DOCTYPE html>
S. No | op | dim | config (layout/dtype, in → out) | shape | main µs | PR µs | Δ% | batched by | prog
-- | -- | -- | -- | -- | -- | -- | -- | -- | --
1 | min | -2 | tile/bf16, WS-L1 → IL-L1 | 1×1×2048×1024 | 172.5 | 121.6 | −29.49% | H, no batch | Unary+Reduce+Unary → Reduce
2 | min | -2 | tile/bf16, WS-L1 → IL-DRAM | 1×1×2048×1024 | 172.9 | 121.8 | −29.57% | H, no batch | Unary+Reduce+Unary → Reduce
3 | min | -2 | tile/bf16, WS-L1 → IL-L1 | 1×1×2048×1024 | 172.6 | 121.6 | −29.53% | H, no batch | Unary+Reduce+Unary → Reduce
4 | min | -2 | tile/bf16, WS-L1 → IL-DRAM | 1×1×2048×1024 | 173.1 | 121.7 | −29.68% | H, no batch | Unary+Reduce+Unary → Reduce
5 | min | -2 | tile/bf16, WS-L1 → IL-L1 | 1×1×4096×1024 | 172.2 | 133.8 | −22.29% | H, no batch | Unary+Reduce+Unary → Reduce
6 | min | -2 | tile/bf16, WS-L1 → IL-DRAM | 1×1×4096×1024 | 172.5 | 133.6 | −22.54% | H, no batch | Unary+Reduce+Unary → Reduce

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/43050_reduce_dram_shard_support)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/43050_reduce_dram_shard_support)

### DIFF
--- a/tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
@@ -1,0 +1,359 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+# Sharded-topology coverage for generic reduce (sum/mean/max/min/std/var) on natural-axis reduces.
+# L1 and DRAM shard grids are disjoint coordinate spaces - worker-core (x,y) vs bank ids on row
+# y=0 - so buffer type is varied independently on input and output.
+
+import pytest
+
+pytestmark = pytest.mark.use_module_device
+
+import torch
+
+import ttnn
+from tests.ttnn.utils_for_testing import assert_numeric_metrics
+
+REDUCE_OPS = {
+    "sum": (ttnn.sum, lambda t, dim, keepdim: torch.sum(t, dim=dim, keepdim=keepdim)),
+    "mean": (ttnn.mean, lambda t, dim, keepdim: torch.mean(t, dim=dim, keepdim=keepdim)),
+    # amax/amin accept a tuple dim, needed by the full-HW tests.
+    "max": (ttnn.max, lambda t, dim, keepdim: torch.amax(t, dim=dim, keepdim=keepdim)),
+    "min": (ttnn.min, lambda t, dim, keepdim: torch.amin(t, dim=dim, keepdim=keepdim)),
+    # Bessel's correction defaults to True on both ttnn and torch.
+    "std": (ttnn.std, lambda t, dim, keepdim: torch.std(t, dim=dim, keepdim=keepdim)),
+    "var": (ttnn.var, lambda t, dim, keepdim: torch.var(t, dim=dim, keepdim=keepdim)),
+}
+
+
+def test_reduce_dram_block_sharded_construction_is_impossible(device, expect_error):
+    """DRAM's 1D, row-y=0 bank address space cannot hold BLOCK_SHARDED's 2D shard grid, so no
+    positive DRAM BLOCK_SHARDED case exists to test - for reduce or any other op."""
+    shard_grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 3))})
+    shard_spec = ttnn.ShardSpec(shard_grid, (32, 32), ttnn.ShardOrientation.ROW_MAJOR)
+    mem_config = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.BLOCK_SHARDED, ttnn.BufferType.DRAM, shard_spec)
+
+    torch_input_tensor = torch.randn((1, 1, 128, 128), dtype=torch.bfloat16)
+    with expect_error(RuntimeError, "DRAM banks are 1D"):
+        ttnn.from_torch(
+            torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=mem_config
+        )
+
+
+# Geometries satisfying each layout's shard constraints (e.g. WIDTH_SHARDED needs shard height ==
+# full physical height) on a single-row DRAM bank grid. BLOCK_SHARDED is unconstructible on DRAM.
+_DRAM_SHARD_GEOMETRY = {
+    ttnn.TensorMemoryLayout.HEIGHT_SHARDED: {
+        "tensor_shape": (1, 1, 416, 32),
+        "shard_shape": (128, 32),
+        "shard_grid": ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 0))}),
+    },
+    ttnn.TensorMemoryLayout.WIDTH_SHARDED: {
+        "tensor_shape": (1, 1, 32, 128),
+        "shard_shape": (32, 32),
+        "shard_grid": ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 0))}),
+    },
+}
+
+
+def _dram_sharded_memory_config(shard_layout, shard_grid=None, shard_shape=None):
+    geometry = _DRAM_SHARD_GEOMETRY[shard_layout]
+    shard_spec = ttnn.ShardSpec(
+        shard_grid or geometry["shard_grid"],
+        shard_shape or geometry["shard_shape"],
+        ttnn.ShardOrientation.ROW_MAJOR,
+    )
+    return ttnn.MemoryConfig(shard_layout, ttnn.BufferType.DRAM, shard_spec)
+
+
+def _dram_sharded_input(device, shard_layout, dtype=ttnn.bfloat16, torch_input_tensor=None):
+    tensor_shape = _DRAM_SHARD_GEOMETRY[shard_layout]["tensor_shape"]
+    if torch_input_tensor is None:
+        torch_input_tensor = torch.randn(tensor_shape, dtype=torch.bfloat16)
+    interleaved_input = ttnn.from_torch(torch_input_tensor, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    dram_sharded_input = ttnn.interleaved_to_sharded(interleaved_input, _dram_sharded_memory_config(shard_layout))
+    return torch_input_tensor, dram_sharded_input
+
+
+@pytest.mark.parametrize("op_name", list(REDUCE_OPS.keys()))
+@pytest.mark.parametrize("shard_layout", list(_DRAM_SHARD_GEOMETRY.keys()))
+@pytest.mark.parametrize("dim", [-1, -2])
+def test_reduce_dram_sharded(device, op_name, shard_layout, dim):
+    """DRAM-sharded input+output across both reduce-dim program factories: dim=-1 uses the W
+    factory, dim=-2 the H factory, whose width-sharded fast path is L1-only. WIDTH_SHARDED + dim=-1
+    reduces along the tensor's own sharded dimension, collapsing it to a single shard."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+
+    torch_input_tensor, dram_sharded_input = _dram_sharded_input(device, shard_layout)
+    torch_output_tensor = torch_op(torch_input_tensor, dim, True)
+
+    dram_sharded_config = _dram_sharded_memory_config(shard_layout)
+    output_tensor = ttnn_op(dram_sharded_input, dim=dim, keepdim=True, memory_config=dram_sharded_config)
+
+    output_mem_config = output_tensor.memory_config()
+    assert output_mem_config.buffer_type == ttnn.BufferType.DRAM
+    assert output_mem_config.is_sharded()
+
+    output_tensor = ttnn.to_torch(output_tensor)
+    assert_numeric_metrics(
+        torch_output_tensor,
+        output_tensor,
+        pcc_threshold=0.999,
+        rtol=0.05,
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+@pytest.mark.parametrize("op_name", list(REDUCE_OPS.keys()))
+@pytest.mark.parametrize("shard_layout", list(_DRAM_SHARD_GEOMETRY.keys()))
+def test_reduce_dram_sharded_full_hw_reduce(device, op_name, shard_layout):
+    """Full-HW reduce from a DRAM-sharded input; the output is interleaved since a collapsed 1x1
+    result cannot stay sharded across cores. sum/mean/max/min decompose host-side into a W-reduce
+    then an H-reduce, each reading the sharded or intermediate tensor; std/var use one Welford
+    call."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+
+    torch_input_tensor, dram_sharded_input = _dram_sharded_input(device, shard_layout)
+    torch_output_tensor = torch_op(torch_input_tensor, (-2, -1), True)
+
+    output_tensor = ttnn_op(dram_sharded_input, dim=(-2, -1), keepdim=True, memory_config=ttnn.DRAM_MEMORY_CONFIG)
+
+    output_mem_config = output_tensor.memory_config()
+    assert output_mem_config.buffer_type == ttnn.BufferType.DRAM
+    assert not output_mem_config.is_sharded()
+
+    output_tensor = ttnn.to_torch(output_tensor)
+    assert_numeric_metrics(
+        torch_output_tensor,
+        output_tensor,
+        pcc_threshold=0.999,
+        rtol=0.05,
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+@pytest.mark.parametrize("op_name", ["sum", "max"])
+def test_reduce_dram_sharded_full_bank_width_h_reduce(device, op_name):
+    """A WIDTH_SHARDED grid spanning every DRAM bank."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+
+    num_banks = device.dram_grid_size().x
+    shard_grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(num_banks - 1, 0))})
+    tensor_shape = (1, 1, 32, 32 * num_banks)
+    shard_shape = (32, 32)
+
+    torch_input_tensor = torch.randn(tensor_shape, dtype=torch.bfloat16)
+    torch_output_tensor = torch_op(torch_input_tensor, -2, True)
+
+    interleaved_input = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    dram_sharded_config = ttnn.MemoryConfig(
+        ttnn.TensorMemoryLayout.WIDTH_SHARDED,
+        ttnn.BufferType.DRAM,
+        ttnn.ShardSpec(shard_grid, shard_shape, ttnn.ShardOrientation.ROW_MAJOR),
+    )
+    dram_sharded_input = ttnn.interleaved_to_sharded(interleaved_input, dram_sharded_config)
+
+    output_tensor = ttnn_op(dram_sharded_input, dim=-2, keepdim=True, memory_config=dram_sharded_config)
+    output_tensor = ttnn.to_torch(output_tensor)
+    assert_numeric_metrics(
+        torch_output_tensor,
+        output_tensor,
+        pcc_threshold=0.999,
+        rtol=0.05,
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+@pytest.mark.parametrize("op_name", ["sum", "max"])
+@pytest.mark.parametrize("dram_side", ["input", "output"])
+def test_reduce_h_width_sharded_mixed_l1_dram(device, op_name, dram_side):
+    """One side DRAM-WIDTH_SHARDED, the other L1-WIDTH_SHARDED, on H-reduce: the use_width_sharding
+    gate must key off buffer type per side, not just layout."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+
+    tensor_shape = _DRAM_SHARD_GEOMETRY[ttnn.TensorMemoryLayout.WIDTH_SHARDED]["tensor_shape"]
+    shard_shape = _DRAM_SHARD_GEOMETRY[ttnn.TensorMemoryLayout.WIDTH_SHARDED]["shard_shape"]
+    shard_grid = _DRAM_SHARD_GEOMETRY[ttnn.TensorMemoryLayout.WIDTH_SHARDED]["shard_grid"]
+    shard_spec = ttnn.ShardSpec(shard_grid, shard_shape, ttnn.ShardOrientation.ROW_MAJOR)
+    dram_config = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.WIDTH_SHARDED, ttnn.BufferType.DRAM, shard_spec)
+    l1_config = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.WIDTH_SHARDED, ttnn.BufferType.L1, shard_spec)
+
+    torch_input_tensor = torch.randn(tensor_shape, dtype=torch.bfloat16)
+    torch_output_tensor = torch_op(torch_input_tensor, -2, True)
+
+    interleaved_input = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    input_config = dram_config if dram_side == "input" else l1_config
+    output_config = l1_config if dram_side == "input" else dram_config
+    input_tensor = ttnn.interleaved_to_sharded(interleaved_input, input_config)
+
+    output_tensor = ttnn_op(input_tensor, dim=-2, keepdim=True, memory_config=output_config)
+
+    output_mem_config = output_tensor.memory_config()
+    assert output_mem_config.buffer_type == (ttnn.BufferType.L1 if dram_side == "input" else ttnn.BufferType.DRAM)
+
+    output_tensor = ttnn.to_torch(output_tensor)
+    assert_numeric_metrics(
+        torch_output_tensor,
+        output_tensor,
+        pcc_threshold=0.999,
+        rtol=0.05,
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+@pytest.mark.parametrize("op_name", list(REDUCE_OPS.keys()))
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32, ttnn.bfloat8_b])
+def test_reduce_dram_sharded_dtypes(device, op_name, dtype):
+    """Every supported dtype through the H-reduce generic branch, at one fixed geometry."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+    shard_layout = ttnn.TensorMemoryLayout.WIDTH_SHARDED
+
+    torch_dtype = torch.float32 if dtype == ttnn.float32 else torch.bfloat16
+    tensor_shape = _DRAM_SHARD_GEOMETRY[shard_layout]["tensor_shape"]
+    torch_input_tensor = torch.randn(tensor_shape, dtype=torch_dtype)
+    torch_output_tensor = torch_op(torch_input_tensor, -2, True)
+
+    _, dram_sharded_input = _dram_sharded_input(
+        device, shard_layout, dtype=dtype, torch_input_tensor=torch_input_tensor
+    )
+    dram_sharded_config = _dram_sharded_memory_config(shard_layout)
+    output_tensor = ttnn_op(dram_sharded_input, dim=-2, keepdim=True, memory_config=dram_sharded_config)
+    output_tensor = ttnn.to_torch(output_tensor)
+
+    # bfloat8_b's block-float quantization inflates relative error on near-zero sums; absolute
+    # error and PCC stay well within tolerance.
+    assert_numeric_metrics(
+        torch_output_tensor,
+        output_tensor,
+        pcc_threshold=0.999,
+        rtol=0.1,
+        atol=0.2,
+        frobenius_threshold=0.02,
+    )
+
+
+def test_reduce_dram_sharded_non_natural_dim(device, expect_error):
+    """Non-natural-dim reduce moves the axis into H/W position via transpose or permute, whose
+    DRAM-sharded handling is out of scope here. Behavior differs across ops - ttnn.sum succeeds on
+    this input while mean/max/min/std/var fail building the intermediate's TensorSpec - so only
+    ttnn.mean is pinned."""
+    torch.manual_seed(0)
+    shard_grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 0))})
+    shard_spec = ttnn.ShardSpec(shard_grid, (416, 32), ttnn.ShardOrientation.ROW_MAJOR)  # one shard per (n, c) slice
+    dram_sharded_config = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.HEIGHT_SHARDED, ttnn.BufferType.DRAM, shard_spec)
+
+    torch_input_tensor = torch.randn(1, 4, 416, 32, dtype=torch.bfloat16)
+    interleaved_input = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    dram_sharded_input = ttnn.interleaved_to_sharded(interleaved_input, dram_sharded_config)
+
+    with expect_error(RuntimeError, "Number of shards along height"):
+        ttnn.mean(dram_sharded_input, dim=1, keepdim=True)
+
+
+def test_reduce_h_width_sharded_l1_and_dram_use_distinct_programs(device):
+    """L1- and DRAM-WIDTH_SHARDED H-reduce take different branches, so the buffer type must reach
+    the program-cache key as distinct entries."""
+    shard_layout = ttnn.TensorMemoryLayout.WIDTH_SHARDED
+    tensor_shape = _DRAM_SHARD_GEOMETRY[shard_layout]["tensor_shape"]
+    shard_shape = _DRAM_SHARD_GEOMETRY[shard_layout]["shard_shape"]
+    shard_grid = _DRAM_SHARD_GEOMETRY[shard_layout]["shard_grid"]
+    shard_spec = ttnn.ShardSpec(shard_grid, shard_shape, ttnn.ShardOrientation.ROW_MAJOR)
+    l1_config = ttnn.MemoryConfig(shard_layout, ttnn.BufferType.L1, shard_spec)
+    dram_config = ttnn.MemoryConfig(shard_layout, ttnn.BufferType.DRAM, shard_spec)
+
+    torch_input_tensor = torch.randn(tensor_shape, dtype=torch.bfloat16)
+    interleaved_input = ttnn.from_torch(torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    # Ignore whatever earlier tests in this session compiled.
+    device.enable_program_cache()
+    device.clear_program_cache()
+    l1_input = ttnn.interleaved_to_sharded(interleaved_input, l1_config)
+    ttnn.sum(l1_input, dim=-2, keepdim=True, memory_config=l1_config)
+    entries_after_l1 = device.num_program_cache_entries()
+
+    dram_input = ttnn.interleaved_to_sharded(interleaved_input, dram_config)
+    ttnn.sum(dram_input, dim=-2, keepdim=True, memory_config=dram_config)
+    entries_after_dram = device.num_program_cache_entries()
+
+    assert entries_after_dram > entries_after_l1, "expected a new program cache entry for the DRAM-sharded case"
+
+
+def test_reduce_dram_sharded_requires_explicit_output_shard_spec_across_buffer_types(device, expect_error):
+    """A spec-less sharded output may only borrow the input's shard grid when both share a buffer
+    type; a DRAM input with an L1 spec-less output must be rejected."""
+    shard_layout = ttnn.TensorMemoryLayout.HEIGHT_SHARDED
+    torch_input_tensor, dram_sharded_input = _dram_sharded_input(device, shard_layout)
+    output_config_no_spec = ttnn.MemoryConfig(shard_layout, ttnn.BufferType.L1)
+
+    with expect_error(RuntimeError, "requires an explicit shard_spec"):
+        ttnn.sum(dram_sharded_input, dim=-1, keepdim=True, memory_config=output_config_no_spec)
+
+
+# Sharding C as well as H and W keeps the layout ND_SHARDED: no legacy layout can express a
+# split along C, so it cannot normalize to WIDTH/HEIGHT/BLOCK.
+_DRAM_ND_TENSOR_SHAPE = (1, 4, 128, 128)
+_DRAM_ND_SHARD_SHAPE = [1, 2, 64, 64]
+
+
+def _dram_nd_sharded_config():
+    grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 0))})
+    return ttnn.MemoryConfig(ttnn.BufferType.DRAM, ttnn.NdShardSpec(ttnn.Shape(_DRAM_ND_SHARD_SHAPE), grid))
+
+
+@pytest.mark.parametrize("op_name", ["sum", "std"])
+@pytest.mark.parametrize("dim", [-1, -2])
+def test_reduce_dram_nd_sharded(device, op_name, dim):
+    """DRAM ND_SHARDED input and output, covering build_reduce_output_tensor_spec's ND branch."""
+    torch.manual_seed(0)
+    ttnn_op, torch_op = REDUCE_OPS[op_name]
+
+    nd_config = _dram_nd_sharded_config()
+    torch_input_tensor = torch.randn(_DRAM_ND_TENSOR_SHAPE, dtype=torch.bfloat16)
+    torch_output_tensor = torch_op(torch_input_tensor, dim, True)
+
+    input_tensor = ttnn.from_torch(
+        torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=nd_config
+    )
+    assert input_tensor.memory_config().memory_layout == ttnn.TensorMemoryLayout.ND_SHARDED
+
+    output_tensor = ttnn_op(input_tensor, dim=dim, keepdim=True, memory_config=nd_config)
+
+    output_mem_config = output_tensor.memory_config()
+    assert output_mem_config.buffer_type == ttnn.BufferType.DRAM
+    assert output_mem_config.is_sharded()
+
+    assert_numeric_metrics(
+        torch_output_tensor,
+        ttnn.to_torch(output_tensor),
+        pcc_threshold=0.999,
+        rtol=0.05,
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+def test_reduce_dram_nd_sharded_requires_explicit_output_nd_shard_spec(device, expect_error):
+    """ND counterpart of the legacy-fallback guard: an L1 ND output with no nd_shard_spec of its
+    own cannot borrow a DRAM input's bank grid."""
+    torch_input_tensor = torch.randn(_DRAM_ND_TENSOR_SHAPE, dtype=torch.bfloat16)
+    input_tensor = ttnn.from_torch(
+        torch_input_tensor,
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=_dram_nd_sharded_config(),
+    )
+    output_config_no_spec = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.ND_SHARDED, ttnn.BufferType.L1)
+
+    with expect_error(RuntimeError, "requires an explicit nd_shard_spec"):
+        ttnn.sum(input_tensor, dim=-2, keepdim=True, memory_config=output_config_no_spec)
--- a/tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
@@ -107,24 +107,43 @@ def test_reduce_dram_sharded(device, op_name, shard_layout, dim):
     )
 
 
+# Output configs a full-HW reduce accepts.  The 1x1 result can stay sharded in either buffer type;
+# std/var reshard the input to the output config on the way in, so the sharded variants reuse the
+# input's own shard spec.
+_HW_OUTPUT_KINDS = ["interleaved", "dram_sharded", "l1_sharded"]
+
+
+def _hw_output_config(output_kind, shard_layout):
+    if output_kind == "interleaved":
+        return ttnn.DRAM_MEMORY_CONFIG
+    geometry = _DRAM_SHARD_GEOMETRY[shard_layout]
+    buffer_type = ttnn.BufferType.DRAM if output_kind == "dram_sharded" else ttnn.BufferType.L1
+    shard_spec = ttnn.ShardSpec(geometry["shard_grid"], geometry["shard_shape"], ttnn.ShardOrientation.ROW_MAJOR)
+    return ttnn.MemoryConfig(shard_layout, buffer_type, shard_spec)
+
+
 @pytest.mark.parametrize("op_name", list(REDUCE_OPS.keys()))
 @pytest.mark.parametrize("shard_layout", list(_DRAM_SHARD_GEOMETRY.keys()))
-def test_reduce_dram_sharded_full_hw_reduce(device, op_name, shard_layout):
-    """Full-HW reduce from a DRAM-sharded input; the output is interleaved since a collapsed 1x1
-    result cannot stay sharded across cores. sum/mean/max/min decompose host-side into a W-reduce
-    then an H-reduce, each reading the sharded or intermediate tensor; std/var use one Welford
-    call."""
+@pytest.mark.parametrize("output_kind", _HW_OUTPUT_KINDS)
+def test_reduce_dram_sharded_full_hw_reduce(device, op_name, shard_layout, output_kind):
+    """Full-HW reduce from a DRAM-sharded input, across every output config the op accepts.
+    sum/mean/max/min decompose host-side into a W-reduce then an H-reduce, each reading the sharded
+    or intermediate tensor; std/var use one Welford call."""
     torch.manual_seed(0)
     ttnn_op, torch_op = REDUCE_OPS[op_name]
 
     torch_input_tensor, dram_sharded_input = _dram_sharded_input(device, shard_layout)
     torch_output_tensor = torch_op(torch_input_tensor, (-2, -1), True)
 
-    output_tensor = ttnn_op(dram_sharded_input, dim=(-2, -1), keepdim=True, memory_config=ttnn.DRAM_MEMORY_CONFIG)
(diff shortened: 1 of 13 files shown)


iwroszTT:
Please add a few perf cases which compare the new DRAM sharded path agains DRAM Interleaved, using an optimized core-grid.

iwroszTT:
DRAM kernels (both sharded and interleaved) are likely limited by the memory latency, since they fetch just one tile of input at a time.
The reader for w-reduce ([link](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_reduce_universal_start_id.cpp)).
The reader for h-reduce ([link](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp))

The fix would be to increase the input CB size and moving a barrier in the reader kernel. The w-reduce reader would need an inner-loop, something like:
```
for (uint32_t i = start_id; i < end; i += K) {
      dfb_in0.reserve_back(K);
      for (uint32_t k = 0; k < K; ++k)                                                                
          noc.async_read(tensor_accessor, dfb_in0, tile_bytes,
                         {.page_id = i + k}, {.offset_bytes = k * tile_bytes});
      noc.async_read_barrier();                                                                       
      dfb_in0.push_back(K);
 }
```
plus handling of the remainder of tiles at the end.

The h-reader already has the inner loop. Take a look at the comment about CB sizing in the program factory: [link](https://github.com/tenstorrent/tt-metal/blob/1ec5b72398130fab0331d47862cd943131570907/ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op_multi_core_h_program_factory.cpp#L275)

iwroszTT:
On the functional side, can you please run the following test (courtesy of Claude):
```

"""Repro for a gap in PR #53564 (generic reduce: DRAM-sharded input/output support).

PR #53564 gates the H factory's width-sharded fast path on is_l1() for both sides:

    reduce_op_multi_core_h_program_factory.cpp:59-62
        use_width_sharding = ... WIDTH_SHARDED && ... WIDTH_SHARDED
                             && a.memory_config().is_l1() && output.memory_config().is_l1();

The W factory has a structurally identical fast path that did NOT get the same gate:

    reduce_op_multi_core_w_program_factory.cpp:43-49
        use_height_sharding = !rm_path && ... HEIGHT_SHARDED && ... HEIGHT_SHARDED
                              && grids equal && shard heights equal && orientations equal
                              && shard_Ht * grid.num_cores() == NC * Ht;      # no buffer type

Once the gate opens it does what the H factory does:
  :110  all_cores = a.shard_spec().value().grid    -> DRAM bank ids used as Tensix worker coords
  :187  .tensor = &a                               -> CB aliased onto the input buffer
  :213  .tensor = &output                          -> CB aliased onto the output buffer

A CB can only be backed by L1 (circular_buffer_config.cpp:219-223), so a DRAM buffer on
either side throws:

    Only L1 buffers can have an associated circular buffer!

Loud failure rather than corruption, but it is an unhandled config instead of a fallback to
the generic path -- which is exactly what the H factory now does for DRAM.

Two ways in, neither covered by any test:

(a) DRAM on both sides. test_reduce_dram_sharded[*-HEIGHT_SHARDED--1] aims here and misses
    on the exact-tiling term alone: its geometry is (1,1,416,32) with shard (128,32) on 4
    banks, so Ht=13 but shard_Ht * num_cores = 4*4 = 16.  Make Ht 16 and every term holds.

(b) Mixed L1 -> DRAM. The gate compares `a.shard_spec()->grid == output.shard_spec()->grid`,
    but an L1 worker-core grid and a DRAM bank grid with the same coordinates compare EQUAL
    as CoreRangeSets -- (0,0)-(3,0) is legal in both spaces.  PR #53564 added
    test_reduce_h_width_sharded_mixed_l1_dram for this hazard on the H factory; there is no
    W-factory counterpart.

The existing coverage for this fast path -- test_reduce_w_height_sharded in
tests/ttnn/nightly/unit_tests/operations/reduction/test_min_max.py:135 -- exercises every
gate condition except buffer type, because create_sharded_memory_config builds L1 configs.

Expected once fixed: DRAM on either side falls through to the generic TensorAccessor path
and produces correct results, exactly as the H factory does today.

Against main (without PR #53564) these fail earlier and differently, at
validate_reduce_sharded_buffer_types:
    reduce: sharded input memory layout HEIGHT_SHARDED is only supported with L1 buffers

Self-contained: run with  pytest test_reduce_w_dram_height_sharded_repro.py -v
"""
import pytest
import torch

import ttnn

# 4 shards fits every current DRAM bank count (Blackhole 8, Wormhole 12) with room to spare.
NUM_SHARDS = 4
SHARD_HT_TILES = 2
TILE = 32

# Ht is chosen so the fast path's exact-tiling term holds:
#     shard_Ht * num_cores == NC * Ht   ->   2 * 4 == 1 * 8
HEIGHT = SHARD_HT_TILES * NUM_SHARDS * TILE  # 256
WIDTH = 128  # reduced away by dim=-1; any tile-aligned width works
SHARD_SHAPE = (SHARD_HT_TILES * TILE, WIDTH)  # (64, 128)


@pytest.fixture
def device():
    yield dev
    ttnn.close_device(dev)


def _shard_grid():
    """Row y=0 so the same CoreRangeSet is valid as an L1 worker grid AND a DRAM bank grid.

    This overlap is the point of case (b): buffer.cpp:184 requires y == 0 for DRAM, and the
    Tensix grid contains that row too, so one CoreRangeSet value names both.
    """
    return ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(NUM_SHARDS - 1, 0))})


def _height_sharded_config(buffer_type, shard_shape=SHARD_SHAPE):
    return ttnn.MemoryConfig(
        buffer_type,
        ttnn.ShardSpec(_shard_grid(), shard_shape, ttnn.ShardOrientation.ROW_MAJOR),
    )


def _skip_if_too_few_banks(device):
    num_banks = device.dram_grid_size().x
    if num_banks < NUM_SHARDS:
        pytest.skip(f"needs >= {NUM_SHARDS} DRAM banks, device has {num_banks}")


def _check(op_name, torch_input, output_tensor):
    torch_output = (torch.amax if op_name == "max" else torch.sum)(torch_input, dim=-1, keepdim=True)
    actual = ttnn.to_torch(output_tensor)
    if op_name == "max":
        # Selection, not accumulation: bf16 max is exact.
        assert torch.equal(torch_output, actual), f"max mismatch\nexpected {torch_output}\ngot {actual}"
    else:
        assert torch.allclose(torch_output, actual, rtol=0.02, atol=0.02), (
            f"sum mismatch\nexpected {torch_output}\ngot {actual}"
        )


def test_gate_arithmetic_is_satisfied():
    """Documents that the geometry below opens the fast path on purpose, not by accident.

    Mirrors reduce_op_multi_core_w_program_factory.cpp:42-49.  If this ever stops holding,
    the two repros below are silently testing nothing.
    """
    nc = 1
    ht = HEIGHT // TILE
    shard_ht = SHARD_SHAPE[0] // TILE
    assert shard_ht * NUM_SHARDS == nc * ht, (
        f"exact-tiling term must hold: shard_Ht({shard_ht}) * num_cores({NUM_SHARDS}) "
        f"== NC({nc}) * Ht({ht})"
    )
    # The output of dim=-1 is (1, 1, HEIGHT, 1), physically padded to (1, 1, HEIGHT, 32).
    # build_reduce_output_tensor_spec re-derives its shard shape from that physical shape over
    # the same grid, giving shard height HEIGHT/NUM_SHARDS -- equal to the input's, as the gate
    # requires.
    assert HEIGHT // NUM_SHARDS == SHARD_SHAPE[0]


@pytest.mark.parametrize("op_name", ["max", "sum"])
def test_reduce_w_l1_height_sharded_control(device, op_name):
    """Control: identical geometry, L1 on both sides.  Must pass -- isolates buffer type as
    the only difference from the two repros below."""
    torch.manual_seed(0)
    torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)

    l1_config = _height_sharded_config(ttnn.BufferType.L1)
    interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
    l1_input = ttnn.interleaved_to_sharded(interleaved, l1_config)

    l1_out_config = _height_sharded_config(ttnn.BufferType.L1, (SHARD_SHAPE[0], TILE))
    ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
    output = ttnn_op(l1_input, dim=-1, keepdim=True, memory_config=l1_out_config)
    _check(op_name, torch_input, output)


@pytest.mark.parametrize("op_name", ["max", "sum"])
def test_reduce_w_dram_height_sharded_both_sides(device, op_name):
    """(a) DRAM on both sides, dim=-1.

    Expected today: RuntimeError "Only L1 buffers can have an associated circular buffer!"
    Expected after the fix: falls through to the generic path and matches torch.
    """
    _skip_if_too_few_banks(device)
    torch.manual_seed(0)
    torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)

    dram_in_config = _height_sharded_config(ttnn.BufferType.DRAM)
    interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
    dram_input = ttnn.interleaved_to_sharded(interleaved, dram_in_config)
    assert dram_input.memory_config().buffer_type == ttnn.BufferType.DRAM

    # Output shard shape is re-derived from the reduced physical shape; only grid and
    # orientation are taken from this config.
    dram_out_config = _height_sharded_config(ttnn.BufferType.DRAM, (SHARD_SHAPE[0], TILE))

    ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
    output = ttnn_op(dram_input, dim=-1, keepdim=True, memory_config=dram_out_config)

    assert output.memory_config().buffer_type == ttnn.BufferType.DRAM
    _check(op_name, torch_input, output)


@pytest.mark.parametrize("op_name", ["max", "sum"])
def test_reduce_w_height_sharded_l1_in_dram_out(device, op_name):
    """(b) L1 input, DRAM output, dim=-1 -- same shard grid coordinates on both sides.

    The gate's grid-equality check compares two CoreRangeSets drawn from disjoint coordinate
    spaces (worker cores vs bank ids).  They compare equal here, so the fast path opens and
    aliases a CB onto the DRAM *output* buffer.

    Expected today: RuntimeError "Only L1 buffers can have an associated circular buffer!"
    """
    _skip_if_too_few_banks(device)
    torch.manual_seed(0)
    torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)

    interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
    l1_input = ttnn.interleaved_to_sharded(interleaved, _height_sharded_config(ttnn.BufferType.L1))
    assert l1_input.memory_config().buffer_type == ttnn.BufferType.L1

    dram_out_config = _height_sharded_config(ttnn.BufferType.DRAM, (SHARD_SHAPE[0], TILE))

    ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
    output = ttnn_op(l1_input, dim=-1, keepdim=True, memory_config=dram_out_config)

    assert output.memory_config().buffer_type == ttnn.BufferType.DRAM
    _check(op_name, torch_input, output)

```

tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
```diff
@@ -0,0 +1,359 @@
...
+        atol=0.05,
+        frobenius_threshold=0.01,
+    )
+
+
+@pytest.mark.parametrize("op_name", list(REDUCE_OPS.keys()))
+@pytest.mark.parametrize("shard_layout", list(_DRAM_SHARD_GEOMETRY.keys()))
+def test_reduce_dram_sharded_full_hw_reduce(device, op_name, shard_layout):
+    """Full-HW reduce from a DRAM-sharded input; the output is interleaved since a collapsed 1x1
+    result cannot stay sharded across cores. sum/mean/max/min decompose host-side into a W-reduce
```
iwroszTT: What happens if someone uses dim=(-2,-1) and requests sharded output?

ign-msati:
> On the functional side, can you please run the following test (courtesy of Claude):
> 
> ```
> 
> """Repro for a gap in PR #53564 (generic reduce: DRAM-sharded input/output support).
> 
> PR #53564 gates the H factory's width-sharded fast path on is_l1() for both sides:
> 
>     reduce_op_multi_core_h_program_factory.cpp:59-62
>         use_width_sharding = ... WIDTH_SHARDED && ... WIDTH_SHARDED
>                              && a.memory_config().is_l1() && output.memory_config().is_l1();
> 
> The W factory has a structurally identical fast path that did NOT get the same gate:
> 
>     reduce_op_multi_core_w_program_factory.cpp:43-49
>         use_height_sharding = !rm_path && ... HEIGHT_SHARDED && ... HEIGHT_SHARDED
>                               && grids equal && shard heights equal && orientations equal
>                               && shard_Ht * grid.num_cores() == NC * Ht;      # no buffer type
> 
> Once the gate opens it does what the H factory does:
>   :110  all_cores = a.shard_spec().value().grid    -> DRAM bank ids used as Tensix worker coords
>   :187  .tensor = &a                               -> CB aliased onto the input buffer
>   :213  .tensor = &output                          -> CB aliased onto the output buffer
> 
> A CB can only be backed by L1 (circular_buffer_config.cpp:219-223), so a DRAM buffer on
> either side throws:
> 
>     Only L1 buffers can have an associated circular buffer!
> 
> Loud failure rather than corruption, but it is an unhandled config instead of a fallback to
> the generic path -- which is exactly what the H factory now does for DRAM.
> 
> Two ways in, neither covered by any test:
> 
> (a) DRAM on both sides. test_reduce_dram_sharded[*-HEIGHT_SHARDED--1] aims here and misses
>     on the exact-tiling term alone: its geometry is (1,1,416,32) with shard (128,32) on 4
>     banks, so Ht=13 but shard_Ht * num_cores = 4*4 = 16.  Make Ht 16 and every term holds.
> 
> (b) Mixed L1 -> DRAM. The gate compares `a.shard_spec()->grid == output.shard_spec()->grid`,
>     but an L1 worker-core grid and a DRAM bank grid with the same coordinates compare EQUAL
>     as CoreRangeSets -- (0,0)-(3,0) is legal in both spaces.  PR #53564 added
>     test_reduce_h_width_sharded_mixed_l1_dram for this hazard on the H factory; there is no
>     W-factory counterpart.
> 
> The existing coverage for this fast path -- test_reduce_w_height_sharded in
> tests/ttnn/nightly/unit_tests/operations/reduction/test_min_max.py:135 -- exercises every
> gate condition except buffer type, because create_sharded_memory_config builds L1 configs.
> 
> Expected once fixed: DRAM on either side falls through to the generic TensorAccessor path
> and produces correct results, exactly as the H factory does today.
> 
> Against main (without PR #53564) these fail earlier and differently, at
> validate_reduce_sharded_buffer_types:
>     reduce: sharded input memory layout HEIGHT_SHARDED is only supported with L1 buffers
> 
> Self-contained: run with  pytest test_reduce_w_dram_height_sharded_repro.py -v
> """
> import pytest
> import torch
> 
> import ttnn
> 
> # 4 shards fits every current DRAM bank count (Blackhole 8, Wormhole 12) with room to spare.
> NUM_SHARDS = 4
> SHARD_HT_TILES = 2
> TILE = 32
> 
> # Ht is chosen so the fast path's exact-tiling term holds:
> #     shard_Ht * num_cores == NC * Ht   ->   2 * 4 == 1 * 8
> HEIGHT = SHARD_HT_TILES * NUM_SHARDS * TILE  # 256
> WIDTH = 128  # reduced away by dim=-1; any tile-aligned width works
> SHARD_SHAPE = (SHARD_HT_TILES * TILE, WIDTH)  # (64, 128)
> 
> 
> @pytest.fixture
> def device():
>     dev = ttnn.open_device(device_id=0)
>     yield dev
>     ttnn.close_device(dev)
> 
> 
> def _shard_grid():
>     """Row y=0 so the same CoreRangeSet is valid as an L1 worker grid AND a DRAM bank grid.
> 
>     This overlap is the point of case (b): buffer.cpp:184 requires y == 0 for DRAM, and the
>     Tensix grid contains that row too, so one CoreRangeSet value names both.
>     """
>     return ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(NUM_SHARDS - 1, 0))})
> 
> 
> def _height_sharded_config(buffer_type, shard_shape=SHARD_SHAPE):
>     return ttnn.MemoryConfig(
>         ttnn.TensorMemoryLayout.HEIGHT_SHARDED,
>         buffer_type,
>         ttnn.ShardSpec(_shard_grid(), shard_shape, ttnn.ShardOrientation.ROW_MAJOR),
>     )
> 
> 
> def _skip_if_too_few_banks(device):
>     num_banks = device.dram_grid_size().x
>     if num_banks < NUM_SHARDS:
>         pytest.skip(f"needs >= {NUM_SHARDS} DRAM banks, device has {num_banks}")
> 
> 
> def _check(op_name, torch_input, output_tensor):
>     torch_output = (torch.amax if op_name == "max" else torch.sum)(torch_input, dim=-1, keepdim=True)
>     actual = ttnn.to_torch(output_tensor)
>     if op_name == "max":
>         # Selection, not accumulation: bf16 max is exact.
>         assert torch.equal(torch_output, actual), f"max mismatch\nexpected {torch_output}\ngot {actual}"
>     else:
>         assert torch.allclose(torch_output, actual, rtol=0.02, atol=0.02), (
>             f"sum mismatch\nexpected {torch_output}\ngot {actual}"
>         )
> 
> 
> def test_gate_arithmetic_is_satisfied():
>     """Documents that the geometry below opens the fast path on purpose, not by accident.
> 
>     Mirrors reduce_op_multi_core_w_program_factory.cpp:42-49.  If this ever stops holding,
>     the two repros below are silently testing nothing.
>     """
>     nc = 1
>     ht = HEIGHT // TILE
>     shard_ht = SHARD_SHAPE[0] // TILE
>     assert shard_ht * NUM_SHARDS == nc * ht, (
>         f"exact-tiling term must hold: shard_Ht({shard_ht}) * num_cores({NUM_SHARDS}) "
>         f"== NC({nc}) * Ht({ht})"
>     )
>     # The output of dim=-1 is (1, 1, HEIGHT, 1), physically padded to (1, 1, HEIGHT, 32).
>     # build_reduce_output_tensor_spec re-derives its shard shape from that physical shape over
>     # the same grid, giving shard height HEIGHT/NUM_SHARDS -- equal to the input's, as the gate
>     # requires.
>     assert HEIGHT // NUM_SHARDS == SHARD_SHAPE[0]
> 
> 
> @pytest.mark.parametrize("op_name", ["max", "sum"])
> def test_reduce_w_l1_height_sharded_control(device, op_name):
>     """Control: identical geometry, L1 on both sides.  Must pass -- isolates buffer type as
>     the only difference from the two repros below."""
>     torch.manual_seed(0)
>     torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)
> 
>     l1_config = _height_sharded_config(ttnn.BufferType.L1)
>     interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
>     l1_input = ttnn.interleaved_to_sharded(interleaved, l1_config)
> 
>     l1_out_config = _height_sharded_config(ttnn.BufferType.L1, (SHARD_SHAPE[0], TILE))
>     ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
>     output = ttnn_op(l1_input, dim=-1, keepdim=True, memory_config=l1_out_config)
>     _check(op_name, torch_input, output)
> 
> 
> @pytest.mark.parametrize("op_name", ["max", "sum"])
> def test_reduce_w_dram_height_sharded_both_sides(device, op_name):
>     """(a) DRAM on both sides, dim=-1.
> 
>     Expected today: RuntimeError "Only L1 buffers can have an associated circular buffer!"
>     Expected after the fix: falls through to the generic path and matches torch.
>     """
>     _skip_if_too_few_banks(device)
>     torch.manual_seed(0)
>     torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)
> 
>     dram_in_config = _height_sharded_config(ttnn.BufferType.DRAM)
>     interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
>     dram_input = ttnn.interleaved_to_sharded(interleaved, dram_in_config)
>     assert dram_input.memory_config().buffer_type == ttnn.BufferType.DRAM
> 
>     # Output shard shape is re-derived from the reduced physical shape; only grid and
>     # orientation are taken from this config.
>     dram_out_config = _height_sharded_config(ttnn.BufferType.DRAM, (SHARD_SHAPE[0], TILE))
> 
>     ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
>     output = ttnn_op(dram_input, dim=-1, keepdim=True, memory_config=dram_out_config)
> 
>     assert output.memory_config().buffer_type == ttnn.BufferType.DRAM
>     _check(op_name, torch_input, output)
> 
> 
> @pytest.mark.parametrize("op_name", ["max", "sum"])
> def test_reduce_w_height_sharded_l1_in_dram_out(device, op_name):
>     """(b) L1 input, DRAM output, dim=-1 -- same shard grid coordinates on both sides.
> 
>     The gate's grid-equality check compares two CoreRangeSets drawn from disjoint coordinate
>     spaces (worker cores vs bank ids).  They compare equal here, so the fast path opens and
>     aliases a CB onto the DRAM *output* buffer.
> 
>     Expected today: RuntimeError "Only L1 buffers can have an associated circular buffer!"
>     """
>     _skip_if_too_few_banks(device)
>     torch.manual_seed(0)
>     torch_input = torch.randn((1, 1, HEIGHT, WIDTH), dtype=torch.bfloat16)
> 
>     interleaved = ttnn.from_torch(torch_input, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
>     l1_input = ttnn.interleaved_to_sharded(interleaved, _height_sharded_config(ttnn.BufferType.L1))
>     assert l1_input.memory_config().buffer_type == ttnn.BufferType.L1
> 
>     dram_out_config = _height_sharded_config(ttnn.BufferType.DRAM, (SHARD_SHAPE[0], TILE))
> 
>     ttnn_op = ttnn.max if op_name == "max" else ttnn.sum
>     output = ttnn_op(l1_input, dim=-1, keepdim=True, memory_config=dram_out_config)
> 
>     assert output.memory_config().buffer_type == ttnn.BufferType.DRAM
>     _check(op_name, torch_input, output)
> ```

there were failure  4 of 7 there: `test_reduce_w_dram_height_sharded_both_sides[max,sum]` and `test_reduce_w_height_sharded_l1_in_dram_out[max,sum]`, all four with `Only L1 buffers can have an associated circular buffer!`

**Reason:** the W factory's height-sharded fast path aliases the input and output shards directly as circular buffers, which only L1 can back. Its gate checked shard layout and geometry but not buffer type, so DRAM-sharded tensors entered this L1-only path and failed at CB creation (`circular_buffer_config.cpp:222`). Mixed L1→DRAM configurations satisfied the gate for the same reason, since identical worker-core and DRAM-bank coordinates compare equal as `CoreRangeSet`s.

*Fix:* `use_height_sharding` now also requires `is_l1()` on both sides, exactly as the H factory's width-sharded gate does, so any DRAM side falls through to the generic `TensorAccessor` path and returns correct results. Added `test_reduce_w_height_sharded_mixed_l1_dram`, parametrized over which side is DRAM (`neither`/`input`/`output`/`both`).

tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
ign-msati: Yes, that works — a collapsed 1×1 result is a single shard, so it can stay sharded in either buffer type.

 `test_reduce_dram_sharded_full_hw_reduce` is now parametrized over every output config the op accepts — interleaved, DRAM-sharded and L1-sharded — across all six ops and both shard layouts.

ign-msati:
> DRAM kernels (both sharded and interleaved) are likely limited by the memory latency, since they fetch just one tile of input at a time. The reader for w-reduce ([link](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_reduce_universal_start_id.cpp)). The reader for h-reduce ([link](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp))
> 
> The fix would be to increase the input CB size and moving a barrier in the reader kernel. The w-reduce reader would need an inner-loop, something like:
> 
> ```
> for (uint32_t i = start_id; i < end; i += K) {
>       dfb_in0.reserve_back(K);
>       for (uint32_t k = 0; k < K; ++k)                                                                
>           noc.async_read(tensor_accessor, dfb_in0, tile_bytes,
>                          {.page_id = i + k}, {.offset_bytes = k * tile_bytes});
>       noc.async_read_barrier();                                                                       
>       dfb_in0.push_back(K);
>  }
> ```
> 
> plus handling of the remainder of tiles at the end.
> 
> The h-reader already has the inner loop. Take a look at the comment about CB sizing in the program factory: [link](https://github.com/tenstorrent/tt-metal/blob/1ec5b72398130fab0331d47862cd943131570907/ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op_multi_core_h_program_factory.cpp#L275)

added support batch NoC reads in both reduce readers and updated perf table https://github.com/tenstorrent/tt-metal/pull/53564#issue-5184981616

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
```diff
@@ -40,11 +41,17 @@ void kernel_main() {
 
     constexpr uint32_t onetile = 1;
 
+    // A full chunk's reads go out behind a single barrier instead of one barrier per tile, which
+    // would expose the whole read latency on every tile. The host sizes the input CB at two batches
+    // of tiles_per_batch, so a multi-tile reserve is only contiguous while the chunk is exactly that
+    // batch: a row_chunk the host did not predict (the SFPU path shortens it) stays per-tile.
+    constexpr bool batch_reads = row_chunk == tiles_per_batch;
```
iwroszTT: Let's have: `constexpr bool batch_reads = (row_chunk == tiles_per_batch);`

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
```diff
@@ -75,15 +82,31 @@ void kernel_main() {
         uint32_t reset_curr_id = curr_id;
         uint32_t reset_w = w;
         uint32_t reset_col_start = col_start_tile_id;
+        // The tail chunk is shorter than a batch, so it keeps the per-tile path.
+        const bool batch_chunk = batch_reads && (chunk_end - i) == row_chunk;
```
iwroszTT: Let's have: const bool batch_chunk = batch_reads && ((chunk_end - i) == row_chunk);

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
```diff
@@ -75,15 +82,31 @@ void kernel_main() {
         uint32_t reset_curr_id = curr_id;
         uint32_t reset_w = w;
         uint32_t reset_col_start = col_start_tile_id;
+        // The tail chunk is shorter than a batch, so it keeps the per-tile path.
```
iwroszTT: When there is a tail, can we use a smaller batch instead of falling back to the per-tile path?
This way we would not need to maintain both paths in the loop below.

ttnn/cpp/ttnn/operations/reduction/generic/device/common.hpp
```diff
@@ -32,6 +32,23 @@ enum class ReduceOpParallelizationStrategy { MULTI_CORE_H, MULTI_CORE_W, MULTI_C
 
 namespace ttnn::prim {
 
+// Tiles the universal reduce reader fetches per NoC barrier. One read in flight leaves the whole
```
iwroszTT: Let's trim down the comment to: _// Tiles the universal reduce reader fetches per NoC barrier._

ttnn/cpp/ttnn/operations/reduction/generic/device/common.hpp
```diff
@@ -32,6 +32,23 @@ enum class ReduceOpParallelizationStrategy { MULTI_CORE_H, MULTI_CORE_W, MULTI_C
...
+// read latency exposed on every tile, which costs 12-16% on a read-bound reduce; 8 per barrier is
+// 2-3% slower than 4, so 4 is the peak.
+inline constexpr uint32_t kReduceReaderTilesPerBatch = 4;
+
+// Reader batch for a kernel whose smallest core streams min_tiles_per_core tiles: a core with fewer
```
iwroszTT: I think the comment can be removed or needs to be significantly simplified. Currently it's easier to understand the code rather than the comment.

ttnn/cpp/ttnn/operations/reduction/generic/device/common.hpp
```diff
@@ -32,6 +32,23 @@ enum class ReduceOpParallelizationStrategy { MULTI_CORE_H, MULTI_CORE_W, MULTI_C
...
+// tiles than a batch never reaches the batched loop, so the whole program reads unbatched instead of
+// leaving that core on a path the batch size does not fit. 1 is unbatched.
+inline uint32_t reduce_reader_batch(uint32_t min_tiles_per_core) {
+    return min_tiles_per_core < kReduceReaderTilesPerBatch ? 1u : kReduceReaderTilesPerBatch;
+}
+
+// Input CB depth for that batch: two batches, so the reader fills one while compute drains the other,
```
iwroszTT: Simplify the comment. Avoid stating obvious facts.

ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op_single_core_hw_program_factory.cpp
```diff
@@ -81,8 +81,11 @@ tt::tt_metal::ProgramDescriptor ReduceDeviceOperation::ReduceSingleCoreHwProgram
 
     ProgramDescriptor desc;
 
+    // One core streams the whole tensor here, so only a tensor smaller than a batch turns it off.
```
iwroszTT: What's being turned off?

tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
```diff
@@ -0,0 +1,424 @@
...
+    device.clear_program_cache()
+    l1_input = ttnn.interleaved_to_sharded(interleaved_input, l1_config)
+    ttnn.sum(l1_input, dim=-2, keepdim=True, memory_config=l1_config)
+    entries_after_l1 = device.num_program_cache_entries()
+
+    dram_input = ttnn.interleaved_to_sharded(interleaved_input, dram_config)
+    ttnn.sum(dram_input, dim=-2, keepdim=True, memory_config=dram_config)
+    entries_after_dram = device.num_program_cache_entries()
+
+    assert entries_after_dram > entries_after_l1, "expected a new program cache entry for the DRAM-sharded case"
```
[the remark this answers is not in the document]
ign-msati: updated suggested changes.

tests/ttnn/unit_tests/operations/reduce/test_reduction_sharded_topology.py
```diff
@@ -0,0 +1,424 @@
...
+    positive DRAM BLOCK_SHARDED case exists to test - for reduce or any other op."""
+    shard_grid = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 3))})
+    shard_spec = ttnn.ShardSpec(shard_grid, (32, 32), ttnn.ShardOrientation.ROW_MAJOR)
+    mem_config = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.BLOCK_SHARDED, ttnn.BufferType.DRAM, shard_spec)
+
+    torch_input_tensor = torch.randn((1, 1, 128, 128), dtype=torch.bfloat16)
+    with expect_error(RuntimeError, "DRAM banks are 1D"):
+        ttnn.from_torch(
+            torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=mem_config
+        )
```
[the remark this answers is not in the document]
ign-msati: updated

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
ign-msati: updated

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
ign-msati: updated

ttnn/cpp/ttnn/operations/reduction/generic/device/common.hpp
ign-msati: updated

ttnn/cpp/ttnn/operations/reduction/generic/device/common.hpp
ign-msati: updated

ttnn/cpp/ttnn/operations/reduction/generic/device/reduce_op_single_core_hw_program_factory.cpp
ign-msati: updated comment

ttnn/cpp/ttnn/operations/reduction/generic/device/common.cpp
```diff
@@ -272,21 +291,23 @@ void validate_reduce_sharded_buffer_types(
     const tt::tt_metal::MemoryConfig& output_mem_config,
     std::string_view op_name) {
     TT_FATAL(
-        !output_mem_config.is_sharded() || output_mem_config.is_l1(),
-        "{}: sharded output memory layout {} is only supported with L1 buffers, got buffer type {}",
+        !output_mem_config.is_sharded() || output_mem_config.is_l1() || output_mem_config.is_dram(),
```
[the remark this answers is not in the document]
ign-msati: Added an explicit DRAM BLOCK_SHARDED fatal in validate_reduce_sharded_buffer_types (either side), plus test_reduce_dram_block_sharded_is_rejected on a constructible one-row spec. The is_l1() || is_dram() check below is only the L1/DRAM buffer-type gate.

iwroszTT (approved):

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_transpose_wh_universal_input_cols_partitioned.cpp
ign-msati: The W reader can shorten its last batch because that reserve is the last one — one short push, then it stops.

Here the reserve/push pair is inside the j < Ht loop, so a short reserve(T) would repeat Ht times. The CB only wraps when the write pointer lands exactly on its end, and T doesn't divide the depth, so a later reserve(T) crosses the wrap — e.g. dest chunk 8, CB 16, tail 3: the 6th tail row reserves at page 15 and writes 15-17. reserve(1) is the one size that always lands cleanly

bbradelTT (approved):

PR #11940, opened 2024-08-27
Add fold to resnet

yugaoTT:
### Ticket
[Link to Github Issue](https://github.com/tenstorrent/tt-metal/issues/10288)

### Problem description
integrate sharded fold into resnet

### Checklist
- [x] Post commit CI passes https://github.com/tenstorrent/tt-metal/actions/runs/10581779438/job/29320820660
- [x] Model perf pass https://github.com/tenstorrent/tt-metal/actions/runs/10581925776
- [x] Device perf (fail same as main) https://github.com/tenstorrent/tt-metal/actions/runs/10581923082
- [x] t3k model perf pass https://github.com/tenstorrent/tt-metal/actions/runs/10583147082/job/29325093815

PR #53580, opened 2026-08-18
[Cleanup] #52899: Remove inert RuntimeTensorShape accessor args from transpose

### LINKED ISSUE #52899: Remove inert RuntimeTensorShape accessor args from transpose op (Metal 2.0 porting blocker)
## Summary

The transpose op is playing a silly trick on us, making porting seem artificially difficult. A trivial and inert op fix
will remove the blocker.

Four of `ttnn::transpose`'s eight program factories request
`tensor_accessor::ArgConfig::RuntimeTensorShape` when building their `TensorAccessorArgs`. The request
is **inert**: `TransposeDeviceOperation` has no custom program hash, so its cache key pins the full
`TensorSpec` of its input, and a cached program is therefore only ever reused for the exact spec it was
built for. Nothing about the tensor can vary across a cache hit, so runtime tensor shape buys nothing —
it only costs common runtime args on every dispatch.

Please remove the request from those four factories.

## Why this blocks Metal 2.0 porting

Requesting a runtime accessor placement makes transpose present as an op needing a **dynamic tensor-shape
relaxation**. Only a tiny subset of ops do this, and I am carefully analyzing each to ensure a perfectly matching 
Metal 2.0 relaxation to avoid introducing bugs. But transpose is lying about its need!

Remove this blocker and transpose becomes fully portable: it needs no relaxation of any kind.

It is the *only* op in the corpus with this combination (accessor dynamism plus a full-spec cache
key). Every other accessor-dynamic op owns a custom hash, where the dynamism may well be load-bearing.
So this is a single-op cleanup, not a pattern that needs a policy.

Left as-is, transpose either needs a spurious relaxation declared, or the porting recipe needs a
special case instructing porters to ignore a relaxation signal in exactly one op. Both are worse than
deleting eight arguments.

## Why the args are inert

- `TransposeDeviceOperation` declares no `compute_program_hash`, and neither `operation_attributes_t`
  (`= TransposeParams`) nor `tensor_args_t` (`= TransposeInputs`, a bare `{ Tensor input; }`) overrides
  `attribute_values` or `to_hash`. See `transpose_device_operation.hpp:26-27` and
  `transpose_device_operation_types.hpp`. The op therefore takes the default cache key, which reflects
  the `Tensor` structurally and pins its entire `TensorSpec`.
- The same four factories already bake shape-derived values as **compile-time** args in the same
  breath — `N`, `H`, `C`, `W`, `ht`, `wt`, `stick_size`, `aligned_page_size`. Nothing re-applies a
  compile-time arg on a cache hit. If the key admitted shape variation, those CTAs would be stale on
  every hit, so correctness in production already depends on shape being pinned.
- `transpose_hc_tiled_program_factory.cpp:164,167` builds its accessors **without** the request
  (`TensorAccessorArgs(*src0_buffer).append_to(reader_compile_time_args)`) and works correctly. That is
  the target shape for the fix, in the same op.

## How it got this way (not anyone's live mistake)

- `2024-07-31` (`a2830ce061d`) — the TT-Lib → TTNN move gives transpose a custom
  `compute_program_hash` keyed on memory configs, dtype, dim and parallelization strategy. **No shape.**
  A runtime tensor shape is genuinely meaningful under that key.
- `2024-08-29` (`e6693ec26cd`, "Add fold to resnet" #11940) — that hash is deleted as a drive-by change
  in an unrelated PR (changelog line: *"#0: remove transpose attributes"*). The key silently becomes the
  default full-spec hash. This is strictly tighter, so nothing breaks and nothing is noticed.
- `2026-05-08` (`0326f014b52`, #41555 / PR #42155) — "Extend universal input support for transpose" adds
  `is_native_transpose_sharding` so block-sharded, width-sharded and cross-memory configurations fall
  through to the interleaved factories instead of crashing, and adds `RuntimeTensorShape` so those
  factories can cope with sharded geometry.

The dynamism postdates the removal of the loose key by 21 months. It was added into a world where the
cache key already gave every shape its own entry.

## Exactly what to change

Delete the second argument to `TensorAccessorArgs(...)` at these eight sites (line numbers as of this
writing; the call shape is the stable identifier):

| File | Lines |
| --- | --- |
| `ttnn/cpp/ttnn/operations/data_movement/transpose/device/transpose_cn_program_factory.cpp` | 68, 74 |
| `.../transpose_hc_rm_program_factory.cpp` | 152, 159 |
| `.../transpose_hc_tiled_interleaved_program_factory.cpp` | 191, 218 |
| `.../transpose_wh_program_factory.cpp` | 228, 244 |

The kernels need no changes. They construct accessors generically — `TensorAccessorArgs<N>()` followed by
`TensorAccessor(args, addr)` — and derive layout from the compile-time config word. None of them use the
three-argument `TensorAccessor(args, addr, page_size)` form.

## Do NOT change

- `transpose_hc_sharded_program_factory.cpp`, `transpose_wh_sharded_program_factory.cpp`,
  `transpose_wh_sharded_rm_program_factory.cpp` — these use no `TensorAccessorArgs` at all.
- `transpose_hc_tiled_program_factory.cpp` — already correct.
- Any other op. The other accessor-dynamic ops (`eltwise/binary_ng`, `eltwise/ternary`,
  `eltwise/unary`, `transformer/sdpa`'s `SparseSDPAOperation` and `SparseSDPAMsaOperation`) all own
  custom hashes and are out of scope here.

## Implementation hazard worth one check

Dropping the request moves the shape values out of the common-runtime vector and into the compile-time
vector, so the accessor's compile-time block grows. The kernels index that block at a fixed base
(`TensorAccessorArgs<0>`, `<3>`, `<9>`, `<10>`), which is fine — but if any kernel reads compile-time args
positioned *after* the accessor block, those indices shift silently. This is not caught by the compiler.

## Validation

Regression bar is #41555's own coverage, since that PR is why the fall-through path exists:
block-sharded, width-sharded, and cross-memory (sharded ↔ interleaved) transpose configurations, plus
the standard WH / HC / CN suites.

Expected outcome: no behavioural change, and slightly fewer runtime args per dispatch.

vsureshTT:
### Ticket
Closes #52899

### Problem description

Four of `ttnn::transpose`'s eight program factories requested `tensor_accessor::ArgConfig::RuntimeTensorShape` when building their `TensorAccessorArgs`. The request is **inert**, and it made transpose present as an op needing a dynamic tensor-shape relaxation for Metal 2.0 porting — which it does not.

### What's the change?

Dropped the `RuntimeTensorShape` request at eight call sites across four factories:

| File | Sites |
| --- | --- |
| `transpose_cn_program_factory.cpp` | 2 |
| `transpose_hc_rm_program_factory.cpp` | 2 |
| `transpose_hc_tiled_interleaved_program_factory.cpp` | 2 |
| `transpose_wh_program_factory.cpp` | 2 |

The accessor was the only producer of the `*_common_runtime_args` vectors in these factories (`emit_runtime_args_*` never touch them), so the eight now-dead vectors and their `KernelDescriptor` assignments are removed too. This conforms all four to `transpose_hc_tiled_program_factory.cpp`, which already builds its accessors this way and works correctly.

No kernel changes. Net: 8 insertions, 32 deletions.

### Why the args are inert

- `TransposeDeviceOperation` declares no `compute_program_hash`, and neither `operation_attributes_t` (`TransposeParams`) nor `tensor_args_t` (`TransposeInputs`) overrides `attribute_values` or `to_hash`. The op takes the default cache key, which pins the input's entire `TensorSpec`. A cached program is only ever reused for the exact spec it was built for, so runtime tensor shape buys nothing and only costs common runtime args per dispatch. There is already a comment stating as much at `transpose_hc_sharded_program_factory.cpp:468`.
- These same factories already bake shape-derived values (`N`, `H`, `C`, `W`, `ht`, `wt`, `stick_size`, `aligned_page_size`) as **compile-time** args, which nothing re-applies on a cache hit. If the key admitted shape variation those CTAs would be stale on every hit, so correctness already depends on shape being pinned.

The dynamism was added in #41555 (2026-05-08) into a world where the loose custom hash had already been removed as a drive-by in #11940 (2024-08-29), 21 months earlier.

### Hazard checks

- **CTA index shift** (flagged in the issue): dropping the request moves shape values from the common-runtime vector into the compile-time vector, growing the accessor block. In all eight sites the accessor block is appended **last**, and every kernel involved reads its accessor at a fixed base (`TensorAccessorArgs<0>`, `<1>`, `<3>`, `<5>`, `<9>`, `<10>`) with no compile-time args positioned after it. Indices line up exactly with what each factory pushes, so nothing shifts.
- **Generic accessor construction**: no kernel uses the three-argument `TensorAccessor(args, addr, page_size)` form; all derive layout from the compile-time config word, which the CT block still carries first.
- **`TT_FATAL` in the single-arg `append_to`**: it asserts no `Runtime` bits remain in the config. `update_args_config()` only ever sets `Sharded`/`IsDram`, never a Runtime bit, so starting from `ArgConfig::None` it cannot fire — including on the sharded buffers that fall through to these interleaved factories.

### Checklist

Tested on Wormhole n150 (`main` @ `d103fddd8fc`):

- `tests/tt_eager/python_api_testing/unit_testing/misc/test_transpose.py` — **456 passed, 70 skipped** (skips are pre-existing `bfloat8_b` layout restrictions)
- `tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_permute.py` — **95 passed**

The second file is where #41555's block-sharded / width-sharded / cross-memory coverage now lives after `7552777bd5f` folded `test_transpose_universal_io.py` into it, so it is the regression bar the issue asks for.

Expected outcome as stated in the issue holds: no behavioural change, slightly fewer runtime args per dispatch.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vsureshTT/transpose_remove_inert_runtime_tensor_shape)


vsureshTT:
/codeowners ping

Ready for review — `[Cleanup]`, closes #52899. Owner per CODEOWNERS:318 is @tenstorrent/metalium-developers-ops-data-movement.

**What this is:** removes the `tensor_accessor::ArgConfig::RuntimeTensorShape` request at 8 sites in 4 transpose program factories, plus the 8 `*_common_runtime_args` vectors it was the sole producer of. 8 insertions, 32 deletions, no kernel changes, no functional change.

**Why it's a no-op:** `TransposeDeviceOperation` declares no `compute_program_hash`, and neither `TransposeParams` nor `TransposeInputs` overrides `attribute_values`/`to_hash`, so the op takes the default cache key pinning the input's full `TensorSpec`. A cached program is only reused for the exact spec it was built for. These factories also already bake `N`/`H`/`C`/`W`/`ht`/`wt`/`stick_size`/`aligned_page_size` as compile-time args that nothing re-applies on a cache hit — correctness already depends on shape being pinned. `transpose_hc_tiled_program_factory.cpp` already omits the request; this conforms the other four.

**Reviewer shortcut — the two hazards, both cleared:**
1. *CTA index shift.* Dropping the request grows the accessor's compile-time block. In all 8 sites the accessor block is appended **last**, and all 9 kernels involved read their accessor at a fixed base (`TensorAccessorArgs<0>`, `<1>`, `<3>`, `<5>`, `<9>`, `<10>`) with no CT args positioned after it. No kernel uses the 3-arg `TensorAccessor(args, addr, page_size)` form.
2. *`TT_FATAL` in the single-arg `append_to`* (asserts no `Runtime` bits remain). `update_args_config()` only ever sets `Sharded`/`IsDram`, never a Runtime bit, so from `ArgConfig::None` it cannot fire — including for sharded buffers that fall through to these interleaved factories.

**Local testing** (Wormhole n150, branched off `main` @ `d103fddd8fc`):
- `tests/tt_eager/python_api_testing/unit_testing/misc/test_transpose.py` — 456 passed, 70 skipped
- `tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_permute.py` — 95 passed

Note the issue names #41555's `test_transpose_universal_io.py`, but `7552777bd5f` folded that file into `test_universal_input_tm_permute.py`, so the second suite above is where its block/width-sharded and cross-memory coverage now lives.

**CI:** PR Gate auto-fired on ready-for-review; I dispatched Sanity tests SKUs[WH,BH,Sim] (run `32206642043`). Happy to run anything further you'd like to see.

llongTT (approved):

PR #53593, opened 2026-08-18
Extract runtime-free topology into TT::ScaleoutTopology (PSD/MGD/PGD + solver + FSD→PSD)

Riddy21:
Please read: https://github.com/tenstorrent/tt-metal/blob/774b6132a9f1f086f8b1570becdc90bcf32cade2/TOPOLOGY_LIB_PLAN.md

## Problem

The topology **solver + PSD + descriptors (MGD/PGD) + FSD→PSD conversion** are pure host-side graph work, but they
live in tt-metal's `fabric` → compiled into **`libtt_metal.so`** (the full Metalium runtime). There is no lean way
to reuse just this logic. Fabric Manager shows both symptoms today:

**1. It imports the whole runtime** — `tt-fabric-manager/controller/CMakeLists.txt`:
```cmake
target_link_libraries(
    fabric_manager_controller_lib
    PUBLIC
        fabric_manager_common
        TT::Metalium   # the entire libtt_metal: device mgmt · HAL · dispatch · JIT · LLRT · UMD driver
        ...
)
```

## What's in the library (`TT::ScaleoutTopology`, static, hidden-visibility)

**`TT::ScaleoutTopology`** — headers in the tt-metalium API tree, `.cpp` compiled from `tt_metal/fabric/`:
```
tt_metal/api/tt-metalium/experimental/fabric/   (headers, unchanged location)
  physical_system_descriptor.hpp     PSD (proto + C++ class)
  mesh_graph_descriptor.hpp          MGD
  physical_grouping_descriptor.hpp   PGD
  topology_solver.hpp                solver (SAT + CSP)
  topology_mapper_utils.hpp          mapper utils
  physical_descriptor_builder.hpp    FSD → PSD
  (mesh_graph.hpp, topology_mapper.hpp stay with fabric — the runtime side)

tt_metal/fabric/   (.cpp compiled into libscaleout_topology)
  physical_system_descriptor.cpp   mesh_graph_descriptor.cpp
  physical_grouping_descriptor_{core,graph_building,matching}.cpp
  topology_solver{,_sat,_sat_solver}.cpp   topology_mapper_utils.cpp
  physical_descriptor_builder.cpp
```

## Consumers
- Auto Mapper: `generate_rank_bindings.cpp`
- TT-metal: `control_plane.cpp`
- Fabric Manager: https://github.com/tenstorrent/tt-fabric-manager

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ridvan/fsd-to-psd-lib)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ridvan/fsd-to-psd-lib)


Riddy21:
# Plan: `scaleout_topology` — a shared, runtime-free topology library (mesh_graph stays in fabric)

**Goal:** make the **topology solver + PSD** (and the descriptors they need) a shared, runtime-free resource that
both **tt-metal** (tt-run / `generate_rank_bindings`) and **Fabric Manager** can use without linking `libtt_metal`.
**Constraint:** do **not** move `mesh_graph` (`MeshGraph` class stays in `fabric`).

---

## 1. The library organization (what it will look like)

```
┌──────────────────────────────────────────────────────────────────────────────┐
│ TT::ScaleoutTools   (OBJECT, tools/scaleout/)                     UNCHANGED     │
│   board → connector → node → cabling_generator → FSD utils / query             │
│   Cabling descriptor ─► FSD                                                    │
│   OWNS protos: FSD, cabling(node_config/cluster_config), deployment            │
│   deps: TT::STL, umd     (no runtime)                                          │
└───────────────────────────────┬────────────────────────────────────────────── ┘
                                 │  links (for FSD proto + board lib)
                                 ▼
┌──────────────────────────────────────────────────────────────────────────────┐
│ TT::ScaleoutTopology   (STATIC, NEW)                             runtime-free   │
│   PSD (proto + class + serialization)                                          │
│   MGD (proto + class)                                                          │
│   PGD (proto + 3 .cpp)                                                          │
│   topology SOLVER  (SAT engine + PSD/adjacency CSP)                            │
│   topology mapper UTILS  — the MGD / PSD / adjacency paths                     │
│   FSD → PSD builder  (fsd_to_psd)                                              │
│   deps: TT::ScaleoutTools, cadical, protobuf, umd, fmt, tt-logger, TT::STL     │
│         + tt-metalium HEADERS via include-dirs   (NOT libtt_metal)             │
└───────────────────────────────┬────────────────────────────────────────────── ┘
                                 │  links
                                 ▼
┌──────────────────────────────────────────────────────────────────────────────┐
│ fabric   (OBJECT → libtt_metal)                                  runtime        │
│   mesh_graph  (MeshGraph class)                       ◄── STAYS (per constraint)│
│   topology_logical_adapters.cpp  (NEW): the 3 MeshGraph→logical-adjacency fns  │
│   TopologyMapper CLASS · control plane · live PSD discovery · EDM builders     │
│   keeps protos: router_port_directions, port_descriptor_table, intermesh       │
└──────────────────────────────────────────────────────────────────────────────┘
```

Both scaleout libs meet at the **FSD proto**: `CablingGenerator::generate_factory_system_descriptor()` produces
it (ScaleoutTools); `build_psd_from_fsd()` consumes it (ScaleoutTopology). One-way dependency chain, no cycles.

---

## 2. Placement table (answers "where does each piece go")

| Piece | Home | Moves? | Why |
|---|---|---|---|
| **FSD proto** | `scaleout_tools` | No | Output of the descriptor toolchain; the boundary type |
| **FSD `utils.cpp` / `query.cpp`** | `scaleout_tools` | No | Stay with the FSD |
| **MGD** (proto + `mesh_graph_descriptor.cpp`) | `scaleout_topology` | **Yes** | No `MeshGraph` dependency |
| **PSD** (proto + class + serialization) | `scaleout_topology` | **Yes** | No `MeshGraph` dependency; pairs with FSD→PSD |
| **PGD** (proto + 3 `.cpp`) | `scaleout_topology` | **Yes** | `.cpp` uses no `MeshGraph` |
| **topology solver** (`topology_solver_sat*.cpp`, PSD/adjacency part of `topology_solver.cpp`) | `scaleout_topology` | **Yes** | Operates on PSD + `AdjacencyGraph`, not `MeshGraph` |
| **topology mapper utils** (MGD / PSD / adjacency / map fns) | `scaleout_topology` | **Yes** | MGD-based + PSD-based paths are `MeshGraph`-free |
| **`fsd_to_psd`** | `scaleout_topology` | **Yes** | Needs only FSD proto + PSD proto |
| **`mesh_graph`** (`MeshGraph` class) | **fabric** | **No — STAYS** | Per constraint |
| **3 `MeshGraph`→logical-adjacency fns** | **fabric** (new `topology_logical_adapters.cpp`) | relocated | Only things that touch `MeshGraph`; used by the runtime mapper |
| router_port_directions / port_descriptor_table / intermesh protos | fabric | No | Runtime routing/serialization protos |

---

## 3. The one split — the 3 `MeshGraph` adapter functions

These are the **only** `MeshGraph`-coupled definitions across the solver + mapper-utils. They move from their
current files into a new fabric-side TU `tt_metal/fabric/topology_logical_adapters.cpp` (declarations stay in the
shared headers, which reference `MeshGraph` header-only):

| Function | From | To |
|---|---|---|
| `build_adjacency_graph_logical(const MeshGraph&)` | `topology_solver.cpp:90` | `fabric/topology_logical_adapters.cpp` |
| `build_adjacency_map_logical(const MeshGraph&)` | `topology_mapper_utils.cpp:260` | `fabric/topology_logical_adapters.cpp` |
| `build_logical_multi_mesh_adjacency_graph(const MeshGraph&)` | `topology_mapper_utils.cpp:706` | `fabric/topology_logical_adapters.cpp` |

**Verified:** the MGD overload `build_logical_multi_mesh_adjacency_graph(const MeshGraphDescriptor&)`
(`topology_mapper_utils.cpp:697`) does **not** call the `MeshGraph` helper, and `generate_rank_bindings` already
uses that MGD overload (`build_logical_multi_mesh_adjacency_graph(mgd)`) — so the runtime-free path never needs
`MeshGraph`. No constructor/signature change is forced; we only relocate 3 definitions.

---

## 4. Files: new / moved / untouched

**New**
- `tools/scaleout/fsd_to_psd/fsd_to_psd.{hpp,cpp}` — offline FSD→PSD builder (compiled into ScaleoutTopology).
- `tt_metal/fabric/topology_logical_adapters.cpp` — the 3 relocated `MeshGraph` functions (in fabric).
- `tools/tests/scaleout/test_fsd_to_psd.cpp` — gtests (links ScaleoutTopology, **not** libtt_metal).

**Moved from `fabric` → compiled by `scaleout_topology`** (files stay physically in `tt_metal/fabric/`, only the
compiling target changes):
`mesh_graph_descriptor.cpp`, `physical_system_descriptor.cpp`, `serialization/physical_system_descriptor_serialization.cpp`,
`physical_grouping_descriptor_{core,graph_building,matching}.cpp`, `topology_solver.cpp` (minus 1 fn),
`topology_solver_sat.cpp`, `topology_solver_sat_solver.cpp`, `topology_mapper_utils.cpp` (minus 2 fns),
`routing_table_generator.cpp`, `fabric_types.cpp`. Protos: `mesh_graph_descriptor.proto`,
`physical_system_descriptor.proto`, `physical_grouping_descriptor.proto`.

**Untouched (stay in fabric, unchanged)**
- `mesh_graph.cpp` (and its `tt_cluster.hpp` include, `is_ubb_galaxy`, the `MeshGraph(const tt::Cluster&)` ctor).
- `fabric_host_utils.{hpp,cpp}` (no `fabric_type_utils` split needed).
- `topology_mapper.cpp`, `physical_system_discovery.cpp`, control plane, EDM builders.

> Note: keeping `mesh_graph` in fabric **removes** the two hardest changes from the earlier prototype (the
> `mesh_graph` cluster-ctor split and the `fabric_type_utils` split). Net: less code, lower risk.

---

## 5. CMake shape (what the targets look like)

```cmake
# tt_metal/fabric/CMakeLists.txt
add_library(scaleout_topology STATIC ${TOPOLOGY_SOURCES} ${TOPOLOGY_PROTO_SRCS})
add_library(TT::ScaleoutTopology ALIAS scaleout_topology)
target_include_directories(scaleout_topology
    PRIVATE "$<TARGET_PROPERTY:TT::Metalium,INCLUDE_DIRECTORIES>"   # tt-metalium HEADERS only
            ${CMAKE_CURRENT_SOURCE_DIR} .../tools/scaleout .../fsd_to_psd
    SYSTEM PRIVATE ${CMAKE_CURRENT_BINARY_DIR} ${Metalium_BINARY_DIR}/tools/scaleout)
target_link_libraries(scaleout_topology
    PUBLIC TT::STL umd::tt-umd
    PRIVATE TT::ScaleoutTools TT::Metalium::HostDevCommon cadical::cadical
            protobuf::libprotobuf fmt::fmt-header-only tt-logger::tt-logger yaml-cpp::yaml-cpp)
set_target_properties(scaleout_topology PROPERTIES CXX_VISIBILITY_PRESET hidden VISIBILITY_INLINES_HIDDEN TRUE)

# fabric now: drops the moved sources/protos, gains topology_logical_adapters.cpp, and:
target_link_libraries(fabric PUBLIC TT::ScaleoutTopology PRIVATE ...runtime...)
```

---

## 6. How each side consumes it

- **tt-metal:** `fabric` links `TT::ScaleoutTopology`; `generate_rank_bindings`/tt-run reach it transitively.
  Stage-2 of #52859 = call `build_psd_from_fsd()` in `generate_rank_bindings` behind the FSD RTOption.
- **Fabric Manager (recommended, matches its model):** compile the same sources from the vendored tt-metal
  submodule into its controller, replacing FM's FSD→PSD fork — no binary/version coupling (same way FM already
  compiles the vendored proto schemas). If FM instead links a prebuilt tt-metalium, add `install()`/EXPORT so it
  ships `libscaleout_topology.a`.

---

## 7. Change list (implementation steps)

1. Create `tt_metal/fabric/topology_logical_adapters.cpp`; cut the 3 `MeshGraph` fns out of
   `topology_solver.cpp` + `topology_mapper_utils.cpp` into it. (Declarations stay in the shared headers.)
2. `tt_metal/fabric/sources.cmake`: define `TOPOLOGY_SOURCES` (the moved set, minus `mesh_graph.cpp`); add
   `topology_logical_adapters.cpp` to `FABRIC_SOURCES`; remove the moved sources from `FABRIC_SOURCES`.
3. `tt_metal/fabric/CMakeLists.txt`: split protos (topology 3 vs fabric 3); define `scaleout_topology`; make
   `fabric` link `TT::ScaleoutTopology`.
4. `fsd_to_psd.{hpp,cpp}` + `test_fsd_to_psd.cpp` (already written) compiled into ScaleoutTopology / its test.
5. `mesh_graph.cpp`, `fabric_host_utils.*` — revert to untouched (undo the earlier prototype's edits).
6. Add `PRIVATE TT::ScaleoutTopology` to the ~dozen consumer targets that reference the moved types (see §7b).

## 7b. Consumer link updates — the real cost (found by building)

Because `scaleout_topology` uses **hidden visibility** (the `scaleout_tools` convention — its headers return protobuf
message classes by value, so the descriptor symbols must be hidden to avoid cross-module "File already exists in
database" aborts), `libtt_metal` **stops exporting** the moved symbols (MGD/PSD/PGD/solver/mapper-utils). So every
target that references those types must **explicitly link `TT::ScaleoutTopology`** — exactly as the `scaleout_tools`
comment already requires for `run_cluster_validation`, `test_link_retraining`, etc.

**Confirmed by build:** `fabric` compiles clean with the sources removed, but `generate_rank_bindings` fails to
*link* (`undefined reference to MeshGraphDescriptor / PhysicalSystemDescriptor / PhysicalGroupingDescriptor / ...`)
until it links `TT::ScaleoutTopology`.

Targets to update (one line each — `PRIVATE TT::ScaleoutTopology`), from a repo scan for the moved types:
- `tools/scaleout`: `generate_rank_bindings`, `generate_mgd`(_lib), `run_cluster_validation`
- `tools/tests/scaleout`: `test_link_retraining`, `test_cabling_descriptor_mgd_generation`, `test_fsd_to_psd`
- `tests/tt_metal/tt_fabric/fabric_router`: `test_physical_grouping_descriptor`, `test_mesh_graph_descriptor`,
  `test_topology_solver`, `test_routing_tables`, `test_topology_mapper`, `test_topology_mapper_utils`
- `tests/tt_metal/tt_fabric/physical_discovery`: `test_physical_system_descriptor`
- `tests/ttnn/.../multiprocess`, `tests/tt_metal/multihost`, `tests/tt_metal/distributed` (a few multi-host tests)

≈ a dozen targets. Mechanical, but it is the genuine blast radius.

**Visibility tradeoff to decide:**
- **Hidden (recommended, consistent):** the ~dozen explicit link additions above; safe for external proto-compiling
  consumers (FM).
- **Default visibility:** `libtt_metal` keeps exporting the symbols → **zero** consumer churn, but reintroduces the
  protobuf duplicate-descriptor abort risk for any external consumer (FM) that compiles the same schemas. Only viable
  if no such consumer exists.

## 8. Status & checkpoints

- ✅ Extraction feasibility already proven on `ridvan/topology-lib` (a superset that also moved `mesh_graph`):
  `scaleout_topology` + `test_fsd_to_psd` built clean; the test links with **no `libtt_metal`**; 5/5 gtests pass.
- ✅ Checkpoint verified: MGD-based logical builder is `MeshGraph`-free; only 3 fns couple to `MeshGraph`.
- ☐ Rework the branch to this "mesh_graph stays" layout (less code than current) and rebuild `scaleout_topology`,
  `test_fsd_to_psd`, and `fabric` to confirm green on both the moved and the removal side.

TOPOLOGY_LIB_PLAN.md
```diff
@@ -0,0 +1,224 @@
...
+        return tt::scaleout_tools::build_physical_descriptor_from_file(fsd_path);   // path in → C++ PSD out (no protos)
+    }
+    return run_psd_discovery();                                        // live fallback
+}();
+// unchanged downstream:
+auto phys = experimental::tt_fabric::build_physical_multi_mesh_adjacency_graph(psd, pgd, mgd);
+auto res  = experimental::tt_fabric::map_multi_mesh_to_physical(logical_adj, phys, config, ...);
+```
+
+### In Fabric Manager (as a static library, no `libtt_metal`)
```
p1-0tr: One thing to watch out for - we'll likely need a `get_host_rank_for_chip` implementation in `MeshGraphDescriptor`, other than that we should be good to go :)

tests/tt_metal/tt_fabric/fabric_router/test_mesh_graph_descriptor.cpp
```diff
@@ -205,6 +205,58 @@ TEST(MeshGraphDescriptorTests, InfersDeclaredTorusTypeForDegenerateDimensions) {
...
+    const MeshId mesh_id{0};
+
+    // chip_id is the row-major device index; rank = row-major fold of (device_coord / board_size) over host dims.
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 0), MeshHostRankId{0});   // (0,0) -> host (0,0)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 3), MeshHostRankId{1});   // (0,3) -> host (0,1)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 5), MeshHostRankId{0});   // (1,1) -> host (0,0)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 6), MeshHostRankId{1});   // (1,2) -> host (0,1)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 8), MeshHostRankId{2});   // (2,0) -> host (1,0)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 12), MeshHostRankId{2});  // (3,0) -> host (1,0)
+    EXPECT_EQ(desc.get_host_rank_for_chip(mesh_id, 15), MeshHostRankId{3});  // (3,3) -> host (1,1)
```
Riddy21, line 231: @p1-0tr The API you asked for

tools/scaleout/src/generate_rank_bindings.cpp
```diff
@@ -697,7 +697,7 @@ int main(int argc, char** argv) {
             if (!std::filesystem::exists(mgd_path) || !std::filesystem::is_regular_file(mgd_path)) {
                 throw std::runtime_error("Mesh Graph Descriptor file does not exist: " + mgd_path.string());
             }
-            mgds.emplace_back(MeshGraphDescriptor(mgd_path));
+            mgds.emplace_back(MeshGraphDescriptor(mgd_path, /*backwards_compatible=*/false, subctx_id));
```
Riddy21, line 700: Back compatible has always been turned off, addi8ng subctx id so that we don't need Distribute context in MGD

tt_metal/api/tt-metalium/experimental/fabric/topology_mapper_utils.hpp
```diff
@@ -172,19 +172,6 @@ TopologyMappingResult map_mesh_to_physical(
...
- * Creates adjacency maps for each mesh based on the logical connectivity defined in the mesh graph.
- * For each fabric node in a mesh, this function identifies its logical neighbors by examining
- * the intra-mesh connectivity from the mesh graph and creates a mapping of FabricNodeId to
- * its vector of adjacent FabricNodeIds.
- *
- * @param mesh_graph Reference to the mesh graph object containing fabric topology
- * @return std::map<MeshId, LogicalAdjacencyMap> Map from mesh ID to logical adjacency map
- */
-std::map<MeshId, LogicalAdjacencyMap> build_adjacency_map_logical(const ::tt::tt_fabric::MeshGraph& mesh_graph);
-
```
Riddy21, line 187: Moved to seperate components from MeshGraph so that we don't add dependency

tt_metal/fabric/mesh_graph_descriptor.cpp
```diff
@@ -35,22 +34,6 @@ namespace tt::tt_fabric {
...
-        const auto& world = DistributedContext::get_current_world();
-        if (world != nullptr) {
-            const auto sc = world->subcontext_id();
-            if (sc.has_value()) {
-                return *sc.value();
-            }
-        }
-    }
-    return std::nullopt;
-}
```
Riddy21, line 52: Moved to MeshGraph

tt_metal/fabric/mesh_graph_descriptor.cpp
```diff
@@ -262,6 +249,63 @@ uint32_t MeshGraphDescriptor::get_switch_chip_count(const InstanceData& switch_i
     return chip_count;
 }
 
+std::optional<MeshHostRankId> MeshGraphDescriptor::get_host_rank_for_chip(MeshId mesh_id, ChipId chip_id) const {
+    // Locate the mesh instance whose local id matches mesh_id (MeshId{instance.local_id}).
+    const InstanceData* mesh_instance = nullptr;
+    for (const auto& global_id : mesh_instances_) {
```
Riddy21, line 255: Duplicated from MeshGraph, functionally the same

tt_metal/fabric/topology_mapper.cpp
```diff
@@ -1841,4 +1841,51 @@ void TopologyMapper::verify_topology_mapping(const Cluster& cluster) const {
     log_debug(tt::LogFabric, "TopologyMapper: Verification completed successfully for all mapped entries");
 }
 
+// MeshGraph-based logical-adjacency builders (the MGD-based and "parts" overloads they call live in the topology library).
+std::map<MeshId, AdjacencyGraph<FabricNodeId>> build_adjacency_graph_logical(const MeshGraph& mesh_graph) {
```
Riddy21, line 1846: Moved from Topology mapper utils

tests/tt_metal/tt_fabric/fabric_router/test_mesh_graph_descriptor.cpp
p1-0tr: Awesome, with that adoption of the lib in FM should go smoothly :)

p1-0tr (commented):
Looking good :) I'll set aside some time to do a draft integration in FM tomorrow

tt_metal/fabric/sources.cmake
```diff
@@ -18,10 +18,27 @@ set(FABRIC_JIT_API_HEADERS
...
 )
 
-set(FABRIC_SOURCES
-    control_plane.cpp
+# Runtime-free topology sources — compiled into the lean TT::ScaleoutTopology static library
+# (see tt_metal/fabric/CMakeLists.txt). NOTE: mesh_graph.cpp and topology_mapper.cpp stay in FABRIC_SOURCES
+# below (MeshGraph + the runtime TopologyMapper class stay in fabric).
+set(TOPOLOGY_SOURCES
     mesh_graph_descriptor.cpp
     routing_table_generator.cpp
```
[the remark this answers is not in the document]
Riddy21: Fixed. `MeshGraphDescriptor` no longer references `DistributedContext`: the sub-context id used for the `mgd{id}_` instance-name prefix is now an explicit constructor parameter (`name_uniquify_id`), and the ambient `DistributedContext::is_initialized()`/`get_current_world()` query moved to a `runtime_subcontext_id()` helper in `mesh_graph.cpp` (the fabric/runtime side). Verified via `nm` on `libscaleout_topology.a` — no undefined `DistributedContext` symbols remain.

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,105 @@ target_sources(
...
+        TT::ScaleoutTools
+        TT::Metalium::HostDevCommon
+        enchantum::enchantum
+        fmt::fmt-header-only
+        tt-logger::tt-logger
+        yaml-cpp::yaml-cpp
+        protobuf::libprotobuf
+        cadical::cadical
+)
+target_compile_features(scaleout_topology PUBLIC cxx_std_20)
```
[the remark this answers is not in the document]
Riddy21: Done — added `TT_ENABLE_UNITY_BUILD(scaleout_topology)` and `tt_reuse_precompile_headers(scaleout_topology TT::CommonPCH)`. Generated proto sources are excluded from the unity build via `SKIP_UNITY_BUILD_INCLUSION`.

tt_metal/api/tt-metalium/experimental/fabric/physical_descriptor_builder.hpp
```diff
@@ -0,0 +1,70 @@
...
+// topology mapper.
+
+#pragma once
+
+#include <string>
+#include <vector>
+
+#include "protobuf/factory_system_descriptor.pb.h"  // tt::scaleout_tools::fsd::proto (from scaleout_tools)
+#include "protobuf/physical_system_descriptor.pb.h"  // tt::fabric::proto (compiled into this lib)
+#include <tt-metalium/experimental/fabric/physical_system_descriptor.hpp>  // tt::tt_metal::PhysicalSystemDescriptor
```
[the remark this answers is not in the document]
Riddy21: Fixed — the public header now forward-declares `tt::scaleout_tools::fsd::proto::FactorySystemDescriptor` and `tt::fabric::proto::PhysicalSystemDescriptor`, and the generated `.pb.h` includes moved into the `.cpp`. Those types are only named by reference / returned by value, so forward declarations suffice; `validate-metalium-includes` now passes.

tt_metal/fabric/physical_descriptor_builder.cpp
```diff
@@ -0,0 +1,385 @@
...
+    };
+
+    if (fsd.has_eth_connections()) {
+        for (const auto& conn : fsd.eth_connections().connection()) {
+            const uint32_t a = conn.endpoint_a().host_id();
+            const uint32_t b = conn.endpoint_b().host_id();
+            if (a == b) {
+                continue;
+            }
+            if (static_cast<int>(a) >= num_hosts || static_cast<int>(b) >= num_hosts) {
```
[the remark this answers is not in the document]
Riddy21: Fixed — the range checks now compare in unsigned space (`a >= static_cast<uint32_t>(num_hosts)`) instead of casting the `uint32_t` id down to `int`.

tt_metal/fabric/physical_descriptor_builder.cpp
```diff
@@ -0,0 +1,385 @@
...
+    ordered.reserve(components.size());
+    for (auto& [root, members] : components) {
+        ordered.push_back(std::move(members));
+    }
+    // Order groups by their lowest member host name for deterministic output.
+    std::sort(ordered.begin(), ordered.end(), [&fsd](const std::vector<uint32_t>& a, const std::vector<uint32_t>& b) {
+        const std::string& na = fsd.hosts(static_cast<int>(a.front())).hostname();
+        const std::string& nb = fsd.hosts(static_cast<int>(b.front())).hostname();
+        return na < nb;
+    });
```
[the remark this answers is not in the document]
Riddy21: Fixed — replaced `a.front()` with a `min_hostname` lambda that scans all members for the lexicographically lowest hostname, so component ordering no longer assumes host_id order matches hostname order.

TOPOLOGY_LIB_PLAN.md
```diff
@@ -0,0 +1,224 @@
...
+  the closure. (Reaching `MeshGraph` would drag `libtt_metal` back in.)
+
+**FM CMake — the lib, not the runtime:**
+```cmake
+# No `tt_metal` / `TT::Metalium` on this line — only the lean topology lib + protobuf.
+target_link_libraries(tt-fabric-manager-controller PRIVATE TT::ScaleoutTopology protobuf::libprotobuf)
+```
+
+**Proof it works:** the unit test `fabric_unit_tests (fabric_router/test_physical_descriptor_builder)` links `TT::ScaleoutTopology` and — per `ldd` — **does not link
+`libtt_metal`**; it runs the FSD→PSD path with zero runtime. FM's controller gets the same lightweight closure.
```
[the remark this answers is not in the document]
Riddy21: Fixed — removed the incorrect `ldd` proof. The doc was restructured; it no longer claims a runtime-free link is proven, and calls out that the current FSD→PSD test runs inside `fabric_unit_tests` (which links `tt_metal`), with a dedicated runtime-free link test as a follow-up.

Plan.md
```diff
@@ -0,0 +1,91 @@
+# Plan: FSD-driven mapping for `tt-run` / `generate_rank_bindings`
+
+Tracking issue: [#52859](https://github.com/tenstorrent/tt-metal/issues/52859) — *tt-run / generate_rank_bindings: accept FSD argument, map from FSD when available, fall back to PSD discovery.*
+
+Today the auto-mapper flow (`tt-run -m/--mesh-graph-descriptor` + `--hosts`, and Fabric Manager's `generate_rank_bindings`) always maps against a **Physical System Descriptor (PSD)** that is **discovered live** from the cluster at runtime. The goal is to optionally map against the **Factory System Descriptor (FSD)** — the "what the cluster *should* look like" descriptor — and fall back to live PSD discovery when no FSD is supplied or it does not cover the requested topology.
+
+This work is split into two stages. **Stage 1 (this PR) is plumbing only** — the FSD path travels end-to-end but is not yet consumed for mapping. **Stage 2 is the consumption** — actually deriving the topology from the FSD.
```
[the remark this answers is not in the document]
Riddy21: Fixed — Plan.md now attributes Stage-1 plumbing to PR #53451 and frames this PR as the shared-library prerequisite for Stage 2.

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,105 @@ target_sources(
...
+endforeach()
+
+# ---------------------------------------------------------------------------
+# TT::ScaleoutTopology — lean, runtime-free static library: MGD + PSD + PGD (proto + class), the topology solver,
+# the topology mapper UTILS, and the offline FSD->PSD builder. mesh_graph + the runtime TopologyMapper class stay
+# in fabric. Pulls tt-metalium headers via include-dirs only (no libtt_metal link), so FM/offline tools can link
+# just this lib without the metal runtime.
+# ---------------------------------------------------------------------------
+add_library(scaleout_topology STATIC ${TOPOLOGY_SOURCES} ${TOPOLOGY_GENERATED_PROTO_SRCS})
+add_library(TT::ScaleoutTopology ALIAS scaleout_topology)
```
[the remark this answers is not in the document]
Riddy21: Changed approach here. The lib now uses **default visibility** — `fabric` links it into `libtt_metal.so`, so its symbols are exported from the runtime exactly as before the sources were split out, and in-tree consumers resolve through `tt_metal` with no per-target link changes. With that, no `install(EXPORT)`/`EXPORT_NAME` is needed for the current in-tree usage, so I dropped the `install(ARCHIVE)` rule too. Installing/exporting it as a standalone archive (for FM to link without `libtt_metal`) is part of the runtime-free-FM follow-up, which also needs the remaining `tt_metal/common` deps (e.g. `mesh_coord`) carved out first.

tests/tt_metal/tt_fabric/CMakeLists.txt
```diff
@@ -11,6 +11,7 @@ target_link_libraries(
     fabric_unit_tests
     PRIVATE
         tt_metal
+        TT::ScaleoutTopology # PSD/MGD/PGD/solver/mapper-utils symbols (hidden in libtt_metal)
```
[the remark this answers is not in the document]
Riddy21: Deferred, with a reason worth capturing: I prototyped a target that links **only** `TT::ScaleoutTopology` and it surfaced that the lib isn't fully runtime-free yet — it still references `mesh_coord`'s `MeshShape`/`MeshCoordinateRange` from `tt_metal/common`. Rather than land a half-working target, the lib is currently consumed via `libtt_metal.so` (default visibility), and the standalone link test lands together with the `tt_metal/common` carve-out follow-up. Noted in TOPOLOGY_LIB_PLAN.md.

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,129 @@ target_sources(
...
-target_sources(fabric PRIVATE ${GENERATED_PROTO_SRCS})
+target_sources(fabric PRIVATE ${FABRIC_GENERATED_PROTO_SRCS})
 
 target_include_directories(fabric PRIVATE .)
 target_include_directories(fabric SYSTEM PRIVATE ${CMAKE_CURRENT_BINARY_DIR})
 
 target_link_libraries(
     fabric
+    PUBLIC
+        TT::ScaleoutTopology
```
blozano-tt, line 157: Why public?

Does ScaleoutTopology participate as part of the public interface for the fabric library?

Or alternatively, is ScaleoutTopology just consumed internally by fabric code?

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,129 @@ target_sources(
...
+# archive directly, or the descriptor protos would register twice ("File already exists in database"). Linking
+# the archive standalone (e.g. Fabric Manager without libtt_metal) is a follow-up once the lib is fully
+# runtime-free (see TOPOLOGY_LIB_PLAN.md); until then it is consumed only via libtt_metal.so.
+
+# --- fabric: remaining runtime protos + link the topology lib -------
+add_custom_target(fabric_generated_protos DEPENDS ${FABRIC_GENERATED_PROTO_SRCS})
+add_dependencies(
+    fabric
+    fabric_generated_protos
+    scaleout_topology
```
blozano-tt, line 136: You definitely don't need to say that fabric depends on scaleout_topology. The library dependency is defined later on in target_link_libraries -> Currently on line 154.

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,129 @@ target_sources(
...
+    scaleout_generated_protos
+)
+if(TARGET all_generated_files)
+    add_dependencies(all_generated_files topology_generated_protos)
+endif()
+
+target_include_directories(
+    scaleout_topology
+    PRIVATE
+        "$<TARGET_PROPERTY:TT::Metalium,INCLUDE_DIRECTORIES>" # tt-metalium public headers (incl physical_descriptor_builder.hpp)
```
blozano-tt, line 96: This is really sus. The new target is stealing include directories from another target.

Plan.md
blozano-tt, line 1: don't forget to remove this in the end.

tt_metal/fabric/CMakeLists.txt
blozano-tt: Agree — PRIVATE is right here, and it's safe to tighten. `fabric` is an OBJECT library, so even a PRIVATE link on `scaleout_topology` still gets forwarded into `libtt_metal.so`'s link line (CMake wraps OBJECT-library link deps in `$<LINK_ONLY:...>`) — nothing breaks by narrowing this. PUBLIC only additionally propagates usage requirements (`cxx_std_20`, `TT::STL`, `umd::tt-umd`) to consumers of `fabric`; since `scaleout_topology`'s own include dirs are all PRIVATE, no headers leak either way, so PUBLIC buys nothing here — just over-propagation. Suggest PRIVATE.

tt_metal/fabric/CMakeLists.txt
blozano-tt: Confirmed — this is redundant. `target_link_libraries` between two real targets already creates the target-level build-ordering dependency for Ninja/Make, including for OBJECT-library consumers, so `scaleout_topology` is ordered before `fabric` regardless. Only the `scaleout_topology` token in this `add_dependencies` can go; `fabric_generated_protos` has to stay since it's a custom target (not linkable), so `add_dependencies` is the only mechanism for that one.

tt_metal/fabric/CMakeLists.txt
blozano-tt, line 96: Agree, and it's worse than sus: `$<TARGET_PROPERTY:TT::Metalium,INCLUDE_DIRECTORIES>` reads Metalium's own build-time `INCLUDE_DIRECTORIES` property directly — that's its *private* include dirs, not its usage requirements (`INTERFACE_INCLUDE_DIRECTORIES`). It bypasses SYSTEM/usage-requirement semantics entirely, and any generator expressions embedded in that property get re-evaluated in `scaleout_topology`'s own context, which can resolve differently than intended.

Bigger problem: this quietly compiles the 'runtime-free' library against the full Metalium runtime's private headers, which hides exactly the coupling the follow-up (standalone linking for Fabric Manager) needs to eliminate — it'll pass CI today and then surface as a mystery include-not-found once someone tries to link `TT::ScaleoutTopology` standalone.

Suggest pulling in the one or two actual paths needed (looks like just `physical_descriptor_builder.hpp`'s location) explicitly, or hoisting that header into a small interface-only target `scaleout_topology` can depend on cleanly. Either way, a `# TODO(runtime-free)` tying it to TOPOLOGY_LIB_PLAN.md would make the debt visible.

tt_metal/fabric/CMakeLists.txt
```diff
@@ -32,46 +32,129 @@ target_sources(
...
+# TT::ScaleoutTopology — the topology sources grouped into their own library: MGD + PSD + PGD (proto + class),
+# the topology solver, the topology mapper UTILS, and the offline FSD->PSD builder. mesh_graph + the runtime
+# TopologyMapper class stay in fabric. fabric links this lib so its objects land in libtt_metal.so and its
+# symbols are exported from there (default visibility) exactly as before the split — in-tree consumers keep
+# reaching these symbols via tt_metal, unchanged. Linking this lib standalone (Fabric Manager without the metal
+# runtime) is a follow-up once it is fully runtime-free; see TOPOLOGY_LIB_PLAN.md.
+# ---------------------------------------------------------------------------
+add_library(
+    scaleout_topology
+    STATIC
```
blozano-tt, line 71: New finding, not one of the flagged items above but worth a look: converting these sources from fabric-OBJECT-members to a separate STATIC archive changes link semantics in a way that could contradict this PR's own "symbols exported exactly as before" comment.

Before this PR, the topology `.cpp`s (including generated `.pb.cc` proto files, which carry self-registering static initializers) were unconditionally folded into `libtt_metal.so` as OBJECT-library members. As a STATIC archive, the linker only pulls in archive members that resolve an undefined symbol referenced from *within* the `.so` at link time. Any translation unit referenced only by consumers external to `libtt_metal.so` (or only reachable via a proto registration side effect, not a direct symbol reference) risks being silently dropped from the `.so`.

Worth confirming this doesn't already bite (e.g. via `nm -D` diffing exported symbols before/after, or checking proto registration still runs). If it's a real gap, options are `$<LINK_LIBRARY:WHOLE_ARCHIVE,TT::ScaleoutTopology>` (CMake >= 3.24) when linking into `libtt_metal.so`, or keeping `scaleout_topology` as an OBJECT library for now and only switching to STATIC once it's actually consumed standalone.

Plan.md
blozano-tt: +1, this reads as a working/scratch doc (distinct from TOPOLOGY_LIB_PLAN.md, which is the intentional design doc) — should be dropped before merge.

blozano-tt:
## Summary of the CMake feedback on `tt_metal/fabric/CMakeLists.txt`

High-level: the split into `TT::ScaleoutTopology` is the right shape for the stated goal (letting Fabric Manager eventually consume topology logic without the full runtime), but the linkage/visibility plumbing needs a pass before it fully delivers on that goal without regressions.

**Confirmed issues (replies inline above):**
1. `fabric` → `TT::ScaleoutTopology` link should be **PRIVATE**, not PUBLIC — nothing downstream needs it as a usage requirement, and PRIVATE still gets forwarded into `libtt_metal.so` correctly since `fabric` is an OBJECT library.
2. The explicit `add_dependencies(fabric ... scaleout_topology)` is redundant — `target_link_libraries` already establishes that build-order dependency. (Keep the `fabric_generated_protos` custom-target dependency in the same call — that one's not redundant.)
3. Pulling `TT::Metalium`'s `INCLUDE_DIRECTORIES` property directly into `scaleout_topology` is a real problem, not just style — it reaches into Metalium's private build-time include dirs, bypasses usage-requirement semantics, and silently re-couples the "runtime-free" library to the full runtime. This one is worth blocking on, since it undermines the PR's own premise. Suggest scoping to the one or two headers actually needed.
4. `Plan.md` at repo root looks like a scratch/working doc (not the same as `TOPOLOGY_LIB_PLAN.md`) and should come out before merge.

**One additional finding (new inline comment):** switching these sources from fabric-OBJECT-members to a `scaleout_topology` STATIC archive changes what the linker pulls into `libtt_metal.so` — OBJECT members are unconditionally included, STATIC archive members are only pulled if something inside the `.so` references them. Given some of these are generated proto `.cc` files with self-registering static initializers, worth confirming nothing got silently dropped from the exported `.so` (e.g. diff `nm -D` before/after). If it's a real gap, `$<LINK_LIBRARY:WHOLE_ARCHIVE,...>` (CMake ≥3.24) or keeping it OBJECT for now are both reasonable fixes.

None of this blocks the overall direction — it's tightening before merge, not a redesign.

PR #53602, opened 2026-08-18
Wire SiTU-GLU (Kimi K3) into the routed expert kernel

### LINKED ISSUE #51351: Kimi-K3 routed expert bring-up
Current implementation is `output = self.down_proj(self.act_fn(self.gate_proj(x)) * self.up_proj(x))`.
While Kimi K3 implementation for routed expert with SiTU is             
```
gate_up = torch.cat([self.gate_proj(hidden_states), self.up_proj(hidden_states)], dim=-1)
current_hidden_states = self.act_fn(gate_up)
output = self.down_proj(current_hidden_states)
```

kgrujcic-tt:
Runs Kimi K3's routed experts on their real activation. Replaces #52634, rebased onto
main now that #52448 (the `situ_glu` SFPU op) has landed.

Closes #51351. Refs #51335.

## Change

- New `RoutedExpertActivation::SituGlu` on `unified_routed_expert_ffn`, evaluated by
  `situ_glu_tile` over the raw bf16 gate/up matmul accumulators. Betas come from
  `SituGluConfigKimi` (4.0 / 25.0); no extra params. Blackhole only, matching `ttnn::situ_glu`.
- SwiGLU-OAI and SiTU-GLU differ only in the SFPU op they call, so the compute kernel's whole
  phase-3 path (keep the raw partials, skip the gate silu, stream gate+up through dst at 2 slots
  per output tile) is now shared behind `FUSED_BINARY_ACT`. `swiglu_oai_activation_phase` becomes
  `binary_activation_phase`. The bias branch moved with it, so the host bias gate is keyed to the
  set of fused binary activations rather than to `SwiGluOai` alone.
- `TtMoe` gains `routed_expert_activation` and `KimiK3Config` names `"situ"`, so both the MoE test
  and the model path (`tt_prefill_block._build_moe`) run SiTU. Every other model keeps the SiLU
  default, unchanged.
- The torch references split the same way the device does: SiTU on the routed experts, SiLU on the
  shared one, which has no SiTU kernel at its 6144 width. `TorchMoe` gains `shared_activation` and
  `run_reference_moe` gains `hidden_act` / `shared_hidden_act` for the vendored
  `KimiSparseMoeBlock`, which otherwise applies one activation to both halves.

## Activation per model

SiTU-GLU is opt-in and reaches K3 alone. `TtMoe(routed_expert_activation=...)` defaults to `Silu`,
and `tt_prefill_block` reads `getattr(model_cfg, "ROUTED_EXPERT_ACTIVATION", "silu")` -- `KimiK3Config`
is the only config that declares it. Nothing else changes activation.

| model | routed-expert activation | selected by |
|---|---|---|
| Kimi-K3 | SiTU-GLU | `KimiK3Config.ROUTED_EXPERT_ACTIVATION = "situ"` |
| Kimi-K2.6 | SiLU | no attribute, takes the default |
| Kimi-K2.7 | SiLU | `KimiK27Adapter` subclasses `KimiK26Adapter` (only the checkpoint differs), so same config, same default |
| MiniMax-M3 | SwiGLU-OAI | explicit at `tt_minimax_moe.py:163`, unchanged |
| gpt-oss-120B | SwiGLU-OAI | explicit at `tt_gpt_oss_moe.py:153`, unchanged |
| DeepSeek V3 / V3.2 / V4-Pro / V4-Flash, GLM-5.1 / 5.2 | SiLU | no attribute, take the default |

The kernel's phase-3 path is now shared between SiTU-GLU and SwiGLU-OAI, but which activation a model
selects is untouched. Two checks that this holds in practice rather than just by construction: Kimi-K2.6
on Galaxy 8x4 returned bit-identical to the figures already in the tree for the pre-change path
(`shared_output` 0.999779, `routed_output` 0.967471, `final_output` 0.983272), and the gpt-oss and
MiniMax-M3 unit suites pass unchanged.

Within K3 the split is narrower still: only the **routed** experts run SiTU-GLU. The shared expert
(6144) and the dense FFN (33792) stay on SiLU, no kernel implementing SiTU at those widths -- which is
why `TorchMoe` gained `shared_activation` and `run_reference_moe` gained `hidden_act` /
`shared_hidden_act`, so both sides of the comparison split exactly as the device does.

K3's checkpoint sets `hidden_act = "situ"` for all three FFN sites, so closing that split is tracked
as **#53625**. Neither remaining site goes through the fused routed-expert kernel -- both fold SiLU
into a matmul as a *unary* activation, and SiTU-GLU is binary over `(gate, up)` -- so it is follow-up
work rather than something this PR can extend into. The reference hooks it needs are already here.

## Kimi-K3 MoE

`test_kimi_k3_moe`, 896 experts / top-16, FABRIC_2D, 640 tok/chip. The 8x4 column is the production
shape, from Galaxy CI on this branch; 2x4 is the loudbox proxy.

| stage | 8x4 | 2x4 | bar |
|---|---:|---:|---|
| shared_output | 0.999759 | 0.999701 | 0.997 |
| latent_input | 0.999882 | 0.999808 | 0.998 |
| latent_routed_output | 0.969906 | 0.972775 | 0.965 |
| routed_output | 0.969555 | 0.972620 | 0.96 |
| final_output | 0.969585 | 0.995710 | 0.965 |
| reference_output (upstream `KimiSparseMoeBlock`) | 0.969567 | 0.995693 | 0.965 |

Every existing threshold holds unchanged. K3's activations sit well inside both tanh caps at
realistic weight scales, so the bf4 cost the saturation cases document does not arise here -- 8x4
`reference_output` went 0.969434 to 0.969567 and 2x4 0.995692 to 0.995693 against the SiLU
stand-in the bring-up used.

## Utilization of routed expert with SiTU-GLU

`UnifiedRoutedExpertFfnDeviceOperation` at the K3 post-projection dims (3584 x 3072), x_rm
production path, bf8 activations / bf4 weights, allocated == active. Realtime profiler, median of
3 dispatches, BH p150b, DDR 16000.

Core utilization counts the three matmuls and ignores the eltwise/activation work:

    FLOP  = 2 * M * K * N * 3 matmuls
    FLOPs = 4096/cycle * 1.35 GHz / fidelity * cores    (fidelity = 1, LoFi; cores = 88)
    util  = (FLOP / FLOPs) / measured

giving a 486.6 TFLOP/s ceiling. DRAM counts 18.58 MB of bf4 weights plus the bf16 x read and the
bf8 output write, against the 512 GB/s spec peak (GDDR6 16 Gbps/pin, 8 x 32-bit channels).

| tokens | us | core util % | TFLOP/s | DRAM GB/s | DRAM util % |
|---:|---:|---:|---:|---:|---:|
| 32 | 184.4 | 2.4 | 11.5 | 102.6 | 20.0 |
| 64 | 184.7 | 4.7 | 22.9 | 104.4 | 20.4 |
| 128 | 185.0 | 9.4 | 45.7 | 108.0 | 21.1 |
| 256 | 187.6 | 18.5 | 90.2 | 114.0 | 22.3 |
| 512 | 256.1 | 27.1 | 132.1 | 94.5 | 18.5 |
| 1024 | 386.0 | 36.0 | 175.2 | 77.2 | 15.1 |
| 2048 | 690.4 | 40.3 | 196.0 | 59.5 | 11.6 |
| 5120 | 1730.8 | 40.2 | 195.4 | 43.2 | 8.4 |

## Tests

Blackhole, all passing:

- `test_single_routed_expert_k3_sweep` -- 32..5120 tokens x both x layouts at 3584x3072, 16 cases.
- `test_single_routed_expert_k3_saturated` -- drives both tanh caps (86.7% of gate and 29.6% of up
  past their beta at the deep setting), across bf8 and the production bf4 weights. The cap coverage
  is asserted, not just logged, so a case cannot quietly slide back into the near-linear middle.
- `test_routed_expert_bias.py::test_situglu_bias_routed_expert` -- expert biases + SiTU-GLU, which
  the widened bias gate newly permits.
- `test_single_routed_expert_k3_perf` -- the 8 device-duration gates above.
- Regression: `test_swigluoai_routed_expert.py`, all of `test_routed_expert_bias.py`, and the
  74 existing `test_single_routed_expert` functional + ISL cases, confirming the shared phase-3
  refactor left the SiLU and SwiGLU-OAI paths intact.
- Wider local sweeps on the rebased tree: 551/551 of the nightly `deepseek_prefill` op suite, and
  Kimi-K2.6 MoE on 4x2 unchanged (`final_output` 0.987011) to confirm the `TtMoe` default is inert
  for every other model.

## Notes

- The K3 single-expert perf gates are calibrated over 8 sweeps rather than one sample. The 512-token
  case sits on K3's DRAM-floor knee, which lands later than kimi_k26's or glm_51's because K3 reads
  18.58 MB of weights against a ~185 us floor, so it carries the same wide band the file already
  gives sub-256 cases. Past the knee everything holds inside 1.5%.
- `TtRoutedExpert` now rejects any non-SiLU activation at construction when not on Blackhole. The
  Wormhole fallback calls `routed_expert_ffn`, which has no activation parameter and always computes
  SiLU, so the old behaviour was a silent wrong answer. This also closes the same pre-existing hole
  for `SwiGluOai`. Both consumers of that path are Blackhole-only in CI and both were re-run against
  the guard: gpt-oss `tests/unit` 11 passed / 1 skipped, MiniMax-M3 `tests/unit` 25 passed /
  27 skipped.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kgrujcic/51351-k3-routed-expert)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kgrujcic/51351-k3-routed-expert)

### DIFF
--- a/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
+++ b/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
@@ -77,6 +77,13 @@ _TORCH_ROUTED_ACTIVATION = {
     ttnn.RoutedExpertActivation.SituGlu: ACTIVATION_SITU,
 }
 
+# ...and -> the upstream vendored model's hidden_act spelling. None leaves the variant config's own
+# hidden_act alone, which is what every SiLU model wants.
+_UPSTREAM_ACT = {
+    ttnn.RoutedExpertActivation.Silu: None,
+    ttnn.RoutedExpertActivation.SituGlu: "situ",
+}
+
 
 def run_model(
     variant,
@@ -642,6 +649,9 @@ def run_model(
         shared_expert_weights=shared_expert_weights,
         latent_weights=latent_weights,
         x=x,
+        # Same routed/shared split the device runs; see run_reference_moe.
+        hidden_act=_UPSTREAM_ACT[routed_activation],
+        shared_hidden_act="silu" if _UPSTREAM_ACT[routed_activation] is not None else None,
     )
     if ref_out is not None and tt_output is not None:
         logger.info("Running upstream MoE reference")
--- a/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
+++ b/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
@@ -30,7 +30,6 @@ from models.demos.deepseek_v3_d_p.reference.kimi_k3_config import KimiK3Config
 from models.demos.deepseek_v3_d_p.reference.tt.moe.expert import ACTIVATION_SILU, ACTIVATION_SITU
 from models.demos.deepseek_v3_d_p.reference.tt.moe.moe import TorchMoe
 from models.demos.deepseek_v3_d_p.tests.reference_runners import run_reference_moe
-from models.demos.deepseek_v3_d_p.tt.moe.tt_routed_expert import ROUTED_EXPERT_ACTIVATION_BY_NAME
 from models.demos.deepseek_v3_d_p.tt.moe.init_helpers import (
     ExpertMapping,
     compute_constants,
@@ -46,6 +45,7 @@ from models.demos.deepseek_v3_d_p.tt.moe.init_helpers import (
 )
 from models.demos.deepseek_v3_d_p.tt.moe.tt_moe import TtMoe
 from models.demos.deepseek_v3_d_p.tt.moe.tt_moe_gate_prefill import GateComputeMode
+from models.demos.deepseek_v3_d_p.tt.moe.tt_routed_expert import ROUTED_EXPERT_ACTIVATION_BY_NAME
 from models.demos.deepseek_v3_d_p.tt.moe.validation_helpers import (
     compare_recall,
     log_combine_mismatch_details,
--- a/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
+++ b/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
@@ -30,6 +30,7 @@ from models.demos.deepseek_v3_d_p.reference.kimi_k3_config import KimiK3Config
 from models.demos.deepseek_v3_d_p.reference.tt.moe.expert import ACTIVATION_SILU, ACTIVATION_SITU
 from models.demos.deepseek_v3_d_p.reference.tt.moe.moe import TorchMoe
 from models.demos.deepseek_v3_d_p.tests.reference_runners import run_reference_moe
+from models.demos.deepseek_v3_d_p.tt.moe.tt_routed_expert import ROUTED_EXPERT_ACTIVATION_BY_NAME
 from models.demos.deepseek_v3_d_p.tt.moe.init_helpers import (
     ExpertMapping,
     compute_constants,
@@ -127,9 +128,10 @@ def run_model(
     torch reference. The shared expert always runs SiLU: no SiTU kernel exists outside the
     routed-expert op, so both sides must stay on SiLU there for the comparison to mean anything.
     """
-    torch_routed_activation = _TORCH_ROUTED_ACTIVATION.get(routed_activation)
-    if torch_routed_activation is None:
+    if routed_activation not in _TORCH_ROUTED_ACTIVATION or routed_activation not in _UPSTREAM_ACT:
         raise ValueError(f"no torch reference for {routed_activation}; supported: {list(_TORCH_ROUTED_ACTIVATION)}")
+    torch_routed_activation = _TORCH_ROUTED_ACTIVATION[routed_activation]
+    upstream_activation = _UPSTREAM_ACT[routed_activation]
 
     # Scoped: only the linear-8 / 64-expert / HOST_ALL / pcc-check case OOMs without this.
     # Cached all-gather semaphores get placed at the wrong offset for that specific config.
@@ -650,8 +652,8 @@ def run_model(
         latent_weights=latent_weights,
         x=x,
         # Same routed/shared split the device runs; see run_reference_moe.
-        hidden_act=_UPSTREAM_ACT[routed_activation],
-        shared_hidden_act="silu" if _UPSTREAM_ACT[routed_activation] is not None else None,
+        hidden_act=upstream_activation,
+        shared_hidden_act="silu" if upstream_activation is not None else None,
     )
     if ref_out is not None and tt_output is not None:
         logger.info("Running upstream MoE reference")
@@ -1026,5 +1028,5 @@ def test_kimi_k3_moe(
         latent_use_norm=KimiK3Config.LATENT_MOE_USE_NORM,
         rms_norm_eps=KimiK3Config.RMS_NORM_EPS,
         final_output_pcc=0.965,
-        routed_activation=ttnn.RoutedExpertActivation.SituGlu,
+        routed_activation=ROUTED_EXPERT_ACTIVATION_BY_NAME[KimiK3Config.ROUTED_EXPERT_ACTIVATION],
     )
--- a/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
+++ b/models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
@@ -27,6 +27,7 @@ from models.demos.deepseek_v3_d_p.reference.deepseek_v3_config import DeepSeekV3
 from models.demos.deepseek_v3_d_p.reference.glm_5_2_config import GLM52Config
 from models.demos.deepseek_v3_d_p.reference.kimi_k2_6_config import KimiK26Config
 from models.demos.deepseek_v3_d_p.reference.kimi_k3_config import KimiK3Config
+from models.demos.deepseek_v3_d_p.reference.tt.moe.expert import ACTIVATION_SILU, ACTIVATION_SITU
 from models.demos.deepseek_v3_d_p.reference.tt.moe.moe import TorchMoe
 from models.demos.deepseek_v3_d_p.tests.reference_runners import run_reference_moe
 from models.demos.deepseek_v3_d_p.tt.moe.init_helpers import (
@@ -70,6 +71,13 @@ _MOE_LAYER_IDX = 3
 # dispatch_buffer_capacity_factor below is ceil(N/2) of the most conservative
 # integer N such that dgs*seq*N >= theoretical worst-case dispatch buffer.
 # Real traffic never approaches the worst case, so half-capacity is sufficient.
+# Fused routed-expert activation -> the TorchExpert reference that must match it.
+_TORCH_ROUTED_ACTIVATION = {
+    ttnn.RoutedExpertActivation.Silu: ACTIVATION_SILU,
+    ttnn.RoutedExpertActivation.SituGlu: ACTIVATION_SITU,
+}
+
+
 def run_model(
     variant,
     config,
@@ -93,6 +101,7 @@ def run_model(
     latent_use_norm=True,
     rms_norm_eps=1e-5,
     final_output_pcc=0.982,
+    routed_activation=ttnn.RoutedExpertActivation.Silu,
 ):
     """TtMoe PCC body — shared between `test_ds_moe` / `test_kimi_moe`.
 
@@ -106,7 +115,14 @@ def run_model(
     mismatch on the skipped padded rows, and padded-row correctness is covered by the
     dedicated grouped_topk / routing_setup tests. HOST_ALL gates ignore padding entirely
     (TtMoe falls back to padding_config=None for non-DEVICE_FP32 gates).
+
+    ``routed_activation`` selects the fused routed-expert kernel's activation and the matching
+    torch reference. The shared expert always runs SiLU: no SiTU kernel exists outside the
+    routed-expert op, so both sides must stay on SiLU there for the comparison to mean anything.
     """
+    torch_routed_activation = _TORCH_ROUTED_ACTIVATION.get(routed_activation)
+    if torch_routed_activation is None:
+        raise ValueError(f"no torch reference for {routed_activation}; supported: {list(_TORCH_ROUTED_ACTIVATION)}")
 
     # Scoped: only the linear-8 / 64-expert / HOST_ALL / pcc-check case OOMs without this.
     # Cached all-gather semaphores get placed at the wrong offset for that specific config.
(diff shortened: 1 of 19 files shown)


kgrujcic-tt:
Unrelated pre-existing failure seen while running this branch's local regression, recording it here so it is not mistaken for fallout from this PR.

`tests/ttnn/nightly/unit_tests/operations/experimental/test_topk_large_indices.py::test_topk_large_indices_production_perf_check` fails on my loudbox in both parametrisations, and it fails *below* the band, not above:

| case | measured | expected | band (+/- 1%) |
|---|---:|---:|---|
| prefill | 1.653 ms | 1.684 ms | [1.667, 1.701] |
| bounded_cache | 0.311 ms | 0.317 ms | [0.314, 0.320] |

Both are exactly reproducible: three consecutive runs gave 1.653 / 0.311 every time, to the digit. So it is a stable property of this board, not noise.

It is not this PR. The only C++ this branch touches is under `unified_routed_expert_ffn`, which shares no code with `topk_large_indices`, and the same two cases fail identically with or without the routed-expert changes exercised. The likelier explanation is board-to-board DDR spread against a very tight +/- 1% band: this box reports nominal `DDR_SPEED=0x3e80` (16000), so a calibration board at a sub-nominal speed would be the slower one and would place the midpoint where it currently sits.

Deliberately not touching the baseline here. Lowering the midpoint to suit this board is not a verified speedup -- nothing got faster, a different board measured faster -- and it would move the band away from whatever the CI SKU (`bh_p150b_civ2_viommu`) actually measures. That belongs in its own change from someone who can compare both boards, not folded into this PR.

Everything else in that selection passes: 44 of 46 in the viommu set, and 551 of 551 in the full nightly `deepseek_prefill` op suite.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/unified_routed_expert_ffn/device/kernels/compute/fused_swiglu.cpp
```diff
@@ -831,22 +854,26 @@ void kernel_main() {
     const uint32_t effective_chunks =
         effective_chunks_runtime < num_chunks_max ? effective_chunks_runtime : num_chunks_max;
 
+    // Must precede every other Compute API call, init included: it does MMIO config of
+    // MATH/PACK/UNPACK requiring those units idle, so an earlier SFPU init risks the startup
+    // writes racing the constants it loaded (SiTU-GLU's tanh_init in particular).
+    compute_kernel_hw_startup<SrcOrder::Reverse>(cb_in0_x, cb_in1_gate, cb_partials_gu);
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in cd2caa6f9bd -- you are right, and the halfway position I took was not defensible.

`compute_kernel_hw_startup` is now the first statement of `kernel_main`, ahead of the scratch-CB waits and the UNPACK/MATH/PACK count handshake, which is what `compute_kernel_hw_startup.h:23-41` asks for. Nothing above it depended on the count: it is handed `cb_in0_x` / `cb_in1_gate` / `cb_partials_gu`, none of which the handshake touches.

Note this ordering is pre-existing on main -- startup sat after both the handshake *and* the SiLU/SwiGLU-OAI init. This PR first moved it ahead of the init (the part that actually bites for SiTU-GLU, whose `tanh_init` loads the three `vConstFloatPrgm` registers the activation reads back on every tile) and now moves it ahead of the handshake too, so the contract holds outright rather than by argument about which preceding calls count as Compute API.

Verified after the move: 132 routed-expert cases pass (K3 sweep + saturation, SiTU/gpt-oss bias, SwiGLU-OAI, the 74 SiLU functional/ISL cases) and the 8 K3 device-perf gates all land inside their bands -- largest delta +0.6% at 1024 tokens, the rest under 0.2%, so the reorder is timing-neutral as well.

kgrujcic-tt:
Overnight pass on this branch. Rebased onto `a85c79a9829`, everything below re-verified on the rebased tree.

## Review comments

- **Copilot, hw startup ordering** -- fixed properly this time. `compute_kernel_hw_startup` is now the first statement of `kernel_main`, ahead of the count handshake, not just ahead of the SFPU init. Reply on the thread.
- **pre-commit / reviewdog** -- isort on the three flagged files. `black==23.10.1`, `isort==5.13.2` and `git-clang-format` (pinned versions, not the newer ones in the image) are all clean on the changed lines.
- **Local `/code-review`, second pass** -- three findings, all fixed: the perf-DB `comments` string still said the run measured the SiLU path; the cap-coverage host matmuls ran for every SiTU case including the 8 perf ones, where the result only fed a log line (~225 GFLOP per invocation, once per profiler iteration); and `run_reference_moe`'s activation override went through `ACT2FN`, which has no `"situ"` entry, so the inverse split raised `KeyError` instead of working.
- Corrected the last two docs that still described SiLU-on-device as a standing K3 limitation -- true now only of the shared expert and the dense FFN.

## Utilization

Added to the description: core utilization, TFLOP/s, DRAM GB/s and DRAM util across the token sweep, from an 88-core grid against a 486.6 TFLOP/s LoFi ceiling and the 512 GB/s DRAM peak (this box reports nominal `DDR_SPEED=0x3e80`). Plateaus at ~40% / 195 TFLOP/s past the ~256-token knee, against 55.0% for Kimi-K2.6 and 54.1% for GLM-5.1 measured the same way on the same box.

## Perf gates recalibrated

The first calibration took one sample per case, which put the 512-token baseline at the bottom of its distribution: it measures 254.0-265.8 us over 12 sweeps and tripped the 3% band at 265.8. That case sits on K3's DRAM-floor knee, which lands a token count later than kimi_k26's or glm_51's because K3 reads 18.58 MB of weights against a ~185 us floor, so it has the same long right tail `_LOW_ISL_MARGIN` already exists for below 256 tokens. Baselines are now centred over 8 sweeps and the wide band extends to 512. Past the knee everything holds inside 1.5%. Three consecutive full runs pass 24/24.

## The Galaxy perf gate has never run -- #53612

Covered in the description. Two independent pre-existing defects; the fix for the first turns the job red on the second, and every available repair either risks other teams' baselines or spends another job's time budget, so it is filed rather than guessed at. The gate's docstring now says its baseline is unchecked.

## Verification

Local, on the rebased tree:

| suite | result |
|---|---|
| routed-expert op tests + perf gates | 132/132 |
| nightly `deepseek_prefill` op suite | 551/551 |
| K3 MoE, mesh-2x4 | pass, PCCs unchanged to 6 dp |
| Kimi-K2.6 MoE, mesh-4x2 | pass, `final_output` 0.987011 |
| torch MoE reference comparison | 4/4 |

Galaxy, [run 32204292668](https://github.com/tenstorrent/tt-metal/actions/runs/32204292668): `Kimi K3 MoE` green with PCCs identical to the earlier run to 6 dp, `GLM MoE` green, `MoE Gate` green. `Kimi MoE` is still queued on the shared runner; it is the K2.6 regression for the `TtMoe` default, already covered locally on 4x2.

One unrelated pre-existing failure recorded in the comment above: `test_topk_large_indices_production_perf_check`, which fails *below* its band on this box and shares no code with anything here.

kgrujcic-tt:
Galaxy [run 32204292668](https://github.com/tenstorrent/tt-metal/actions/runs/32204292668) is complete: **all four jobs green**.

| job | result |
|---|---|
| `Blaze - Kimi K3 MoE` | success -- PCCs identical to the earlier run to 6 dp |
| `Blaze - Kimi MoE` (K2.6) | success |
| `Blaze - GLM MoE` | success |
| `Blaze - MoE Gate` | success |

K2.6 is the regression that matters for the `TtMoe` / `TorchMoe` changes, since every model other than K3 takes the new `routed_expert_activation` default. It came back **bit-identical to the figures already recorded in the tree** for the pre-change path -- `shared_output` 0.999779, `routed_output` 0.967471, `final_output` 0.983272, the same values the K3 adapter comment cites for K2.6 on 8x4. The default is inert, as intended.

Also re-ran the two `SwiGluOai` consumers against the new construction guard, which was the last unexercised path in this change: gpt-oss `tests/unit` 11 passed / 1 skipped, MiniMax-M3 `tests/unit` 25 passed / 27 skipped.

Nothing outstanding on my side. The two open items are both deliberate and documented: #53612 (the Galaxy perf gate that has never executed, not fixed here) and the unrelated `test_topk_large_indices` floor violation in the comment above.

models/demos/deepseek_v3_d_p/tests/pcc/test_ttnn_moe.py
```diff
@@ -338,9 +363,12 @@ def upload_tt_x():
             latent_weights=latent_weights,
             latent_use_norm=latent_use_norm,
             rms_norm_eps=rms_norm_eps,
-            # SiLU on both sides: the device has no SiTU kernel yet (#51335), and comparing against a
-            # SiTU reference would measure that gap rather than this dataflow.
-            activation="silu",
+            # Routed side matches whatever the fused kernel runs; the shared expert stays on SiLU
+            # (no SiTU kernel outside the routed-expert op).
+            activation=torch_routed_activation,
+            situ_beta=KimiK3Config.ACTIVATION_SITU_BETA,
```
pmilojevicTT, line 369: Let's put this to be model independent.

tests/ttnn/nightly/unit_tests/operations/experimental/deepseek_prefill/test_single_routed_expert.py
```diff
@@ -35,6 +36,19 @@
...
+# in one place stops a case from measuring one activation against another's golden.
+# SwiGluOai is absent on purpose: its reference lives in test_swigluoai_routed_expert.py.
+_TORCH_ACTIVATION = {
+    ttnn.RoutedExpertActivation.Silu: ACTIVATION_SILU,
+    ttnn.RoutedExpertActivation.SituGlu: ACTIVATION_SITU,
+}
+
+# Kimi K3 SiTU-GLU betas. The device kernel bakes SituGluConfigKimi; these must match it,
+# or the reference silently grades against a different activation.
+_SITU_BETA_GATE = KimiK3Config.ACTIVATION_SITU_BETA
```
pmilojevicTT, line 49: Try to make it model independent

pmilojevicTT:
Lgtm, just address minor polishing comment and run CI.
Have someone from Minimax,GPT-OSS team take a look at changes inside `test_routed_expert_bias.py`

pmilojevicTT (approved):

tests/ttnn/nightly/unit_tests/operations/experimental/deepseek_prefill/test_single_routed_expert.py
kgrujcic-tt: Will do once there's new model that needs that. Currently, main focus is K3.

Issue #53326, opened 2026-08-17
ttnn.sort returns out-of-range indices when the input contains padding sentinel value


iwroszTT:
ttnn.sort pads the sort dimension up to the bitonic network size using a sentinel value: +inf for ascending direction, -inf for descending. When the input itself contains that value, the padding lanes tie with real elements and the sort emits the padding lanes' indices into the logical output region.

The returned values are correct (they match torch.sort exactly), so value-only tests pass, but the indices tensor contains errors.

To Reproduce
                                                                                                                           
 ```
                                                                                                                   
  import pytest                                                                                                                             
  import torch                                                                                                                              
  import ttnn                                                                                                                               
                                                                                                                      
  @pytest.mark.parametrize(                                                                                                                 
      "n, descending, fill",                                                                                                                
      [                                                                                                                                     
          (6144, True, float("-inf")),   # 1948 out-of-range indices, max index 8191
          (151936, True, float("-inf")), # #26057 shape: max index 262143 = 2**18 - 1
          (151936, False, float("inf")), # ascending is symmetric: 109916 out-of-range
      ],
  )
  def test_sort_pad_index_leak(n, descending, fill, device):
      torch.manual_seed(0)
      input = torch.full((1, n), fill, dtype=torch.float32)
      input[..., torch.randperm(n)[:328]] = torch.randn(328) * 8.0

      ttnn_input = ttnn.from_torch(input, ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
      _, ttnn_indices = ttnn.sort(ttnn_input, descending=descending)   

      indices = ttnn.to_torch(ttnn_indices).reshape(-1)[:n].long()
      assert indices.max() < n  # FAILS: indices point into the pow2 padding region
```

Expected behavior                                                    
                                                                       
  Indices are always in [0, N) regardless of input content.

Observed behavior (Blackhole p100a, main @ e16f6cbfb09, 2026-08-17, deterministic)
                                                                       
  ```
┌──────────────────────────┬────────────┬────────────┬──────────────────────┬─────────────────────┐
  │          shape           │ descending │ input fill │ out-of-range indices │      max index      │
  ├──────────────────────────┼────────────┼────────────┼──────────────────────┼─────────────────────┤
  │ [1, 6144]                │ True       │ -inf       │ 1948                 │ 8191                │
  ├──────────────────────────┼────────────┼────────────┼──────────────────────┼─────────────────────┤
[...]

PR #53618, opened 2026-08-19
[Perf] Build the top-k/sort/moe/sampling index tile with the NoC, not 1024 scalar stores

### LINKED ISSUE #52909: [Bounty $5,000] Perf - top-k/sort/moe/sampling index tile is written with 1024 scalar stores per tile on a DM core

kinginu:
### PR Category

Performance

### Summary

Closes #52909.

`generate_index_tile()` writes **1024 scalar values per tile** on a data movement core, one store at a time, in four copies of the same loop. The value it writes is `l + face_size * j + w`: it depends on the column within the face and on the face column, not on the row. All 16 rows of a face are identical, the two lower faces repeat the two upper ones, and the 1024-entry tile holds 32 distinct values.

Seed two rows of each upper face with the RISC and let the NoC replicate them. Per tile that is 32 stores, each a 32-bit word holding two indices, and 15 NoC read commands: 7 for each upper face, then one that copies both upper faces to the lower half. The loop it replaces writes 1024 16-bit stores.

This is the construction `experimental/reduction/deepseek_grouped_gate` already uses, carrying the store-visibility guard #50374 added for exactly that loop-back. The four shipping copies never got it. They become one shared `kernel_lib` header here.

### Measured

`bfloat16`, `dim=-1`. Per-op time in the steady state (50 ops enqueued back-to-back, median of 30) for the small shapes, median of 30 single calls for the larger ones. Base and patched interleaved on the same build, kernel cache cleared between, two rounds each; both parts measured at this PR's head.

**Wormhole n300**

| op | shape | before | after | |
|---|---|---|---|---|
| `ttnn.topk` | `[1,1,32,16384]` generating | 719.60 | **189.27** | −73.7% |
| `ttnn.topk` | `[1,1,32,256]` k=8 | 168.18 | **77.38** | −54.0% |
| `ttnn.topk` | `[1,1,2048,256]` k=8 | 172.29 | **80.30** | −53.4% |
| `ttnn.topk` | `[1,1,32,64]` k=32 | 58.80 | **27.62** | −53.0% |
| `ttnn.topk` | `[1,1,32,1024]` | 607.35 | **349.63** | −42.4% |
| `ttnn.moe` | `[1,1,32,64]` | 31.56 | **21.51** | −31.9% |
| `ttnn.topk` | `[1,1,32,8192]` | 883.45 | **607.17** | −31.3% |
| `ttnn.sort` | `[1,1,32,1024]` | 1812.76 | **1299.91** | −28.3% |
| `ttnn.sort` | `[1,1,32,2048]` | 3601.87 | **2588.93** | −28.1% |
| `ttnn.sampling` | `[1,1,32,1024]` | 450.57 | **355.74** | −21.0% |
| `ttnn.topk` | `[1,1,1024,128]` k=4 | 97.43 | **79.77** | −18.1% |
| `ttnn.topk` | `[1,1,32,128]` k=4 | 96.34 | **79.32** | −17.7% |
| `ttnn.topk` | `[1,1,32,4096]` | 665.90 | **605.62** | −9.1% |
| `ttnn.sort` | `[1,1,32,4096]` | 950.94 | 884.76 | −7.0% |
| `ttnn.sampling` | `[1,1,32,256]` | 333.24 | 325.29 | −2.4% |
| `ttnn.sort` | `[1,1,32,8192]` | 1174.57 | 1160.34 | −1.2% |
| `ttnn.topk` | `[1,1,32,2048]` | 608.46 | 604.41 | −0.7% |
| `ttnn.topk` | `[1,1,32,64128]` | 17417.82 | 17418.29 | ±0 |
| `ttnn.topk` | `[1,1,32,16384]` reading a tensor | 189.46 | 189.33 | control |

**Blackhole P300**

| op | shape | before | after | | floor |
|---|---|---|---|---|---|
| `ttnn.sort` | `[1,1,32,8192]` | 1577.46 | **763.81** | −51.6% | 627.80 |
| `ttnn.sampling` | `[1,1,32,1024]` | 189.82 | **113.28** | −40.3% | 113.26 |
| `ttnn.topk` | `[1,1,32,64]` k=32 | 14.82 | **9.00** | −39.2% | 8.88 |
| `ttnn.moe` | `[1,1,32,64]` | 15.75 | **10.57** | −32.9% | 9.72 |
| `ttnn.sort` | `[1,1,32,1024]` | 626.09 | **474.91** | −24.1% | 473.30 |
| `ttnn.sort` | `[1,1,32,2048]` | 1507.19 | **1202.36** | −20.2% | 1200.14 |
| `ttnn.topk` | `[1,1,32,16384]` generating | 130.90 | **106.37** | −18.7% | 105.49 |
| `ttnn.topk` | `[1,1,32,256]` k=8 | 43.91 | **37.89** | −13.7% | 36.62 |
| `ttnn.topk` | `[1,1,1024,128]` k=4 | 27.83 | **24.41** | −12.3% | 23.74 |
| `ttnn.topk` | `[1,1,32,128]` k=4 | 26.62 | **23.44** | −11.9% | 23.29 |
| `ttnn.topk` | `[1,1,2048,256]` k=8 | 46.84 | **41.35** | −11.7% | 40.59 |
| `ttnn.topk` | `[1,1,32,8192]` | 135.34 | **124.48** | −8.0% | 121.73 |
| `ttnn.topk` | `[1,1,32,1024]` | 168.15 | **160.88** | −4.3% | 158.83 |
| `ttnn.topk` | `[1,1,32,2048]` | 310.23 | **303.66** | −2.1% | 302.47 |
| `ttnn.sampling` | `[1,1,32,256]` | 56.55 | 56.17 | −0.7% | 56.32 |
| `ttnn.topk` | `[1,1,32,4096]` | 89.20 | 89.61 | ±0 | 89.08 |
| `ttnn.sort` | `[1,1,32,4096]` | 398.48 | 397.72 | ±0 | 399.40 |
| `ttnn.topk` | `[1,1,32,64128]` | 8872.83 | 8867.93 | ±0 | 8864.14 |
| `ttnn.topk` | `[1,1,32,16384]` reading a tensor | 110.14 | 109.89 | control | |

`floor` is an ablation that deletes the stores and leaves the CB, the L1 tile, the unpack and the transpose in place. Its results are wrong by construction; it bounds what removing the stores can recover at all. Every Blackhole row that moves lands within a few percent of it, and `ttnn.sampling [1,1,32,1024]` lands on it exactly.

Where a row does not move, the generation is hidden behind compute at that shape rather than absent: the ablation does not move it either. `ttnn.topk [1,1,32,16384]` is the clearest case of the opposite — on Wormhole, generating the indices cost 3.8x reading a precomputed tensor out of DRAM before this change (719.60 against 189.46) and costs the same after it (189.27 against 189.33). The DRAM workaround that the sampling path carries exists because generating was that expensive.

Two seed rows rather than one: the second row trades 16 more stores for 16 fewer reads. On `sort [1,1,32,8192]`, Blackhole: 1 row 839 µs, 2 rows 805 µs, 4 rows 839 µs, 8 rows 925 µs. The reads are asynchronous and overlap, so those four points are the evidence rather than a per-instruction cost model. One to four rows sit inside a 4% band, so a part with a different store-to-command ratio pays at most that much until it is re-measured, and only `seed_rows` changes.

### Correctness

- `test_topk.py`, `test_moe.py`, `test_sampling.py` and `test_sort.py`: **405 passed, 8 skipped, 81 xfailed** on Blackhole; **396 passed, 46 skipped, 81 xfailed and 1 failed** on Wormhole. The one Wormhole failure, `test_sort_fp32_wide_index_correctness[descending=False]` (`147512 duplicated indices`), reproduces identically on the unpatched tree in the same build, so it is pre-existing: it is the ascending half of #53326, which is currently xfail-marked for `descending=True` only.
- On both parts: `ttnn.topk` values are bit-equal to `torch.topk` and gathering the input with the returned indices reproduces the returned values, at 8 shapes from `[1,1,32,64]` to `[1,1,32,16384]`; `ttnn.sort` output is monotonic under the same gather check, ascending and descending; `ttnn.sampling` with `k=1` returns the input index of the row maximum; `ttnn.moe` output is bit-identical to the current kernel.

### Notes for review

- The RISC-then-NoC handoff needs the seed visible in L1 before the NoC reads that line. The last written word is read back, a blocking load, before the read is issued, as #50374 established for `deepseek_grouped_gate`.
- `generate_index_tile()` now issues its own `noc_async_read_barrier()`. In the two readers that interleave it with an in-flight tile read (`reduction/moe`, `reduction/sampling`) that barrier also waits on the data read. That is correct, costs a little overlap, and is already inside the numbers above.
- The 32-bit index path (Quasar, INT32 dest) keeps the scalar loop. The same construction applies to it, but on the shapes where it is reachable today the generation is already hidden behind single-core compute; generalising it measured as no change, so it is left out.
- `experimental/topk_router_gpt` carries a fifth copy of the loop and is untouched.

### Not established

- Host wall clock, not the device profiler.
- `bfloat16` only.
- Blackhole numbers are P300, Wormhole numbers are n300, one device each.
- The upstream tree moved under this branch while it was measured: `ttnn.topk [1,1,32,64128]` was 36301 µs on Wormhole three weeks ago and is 17418 µs now, and that shape went from −52.4% to ±0 because the generation now fits inside the shorter compute. The tables above are all at this PR's head.


kinginu:
@jberkowitzTT Could you run `/codeowners ping` on this PR when you get a chance? Thanks!

ttnn/cpp/ttnn/kernel_lib/index_tile_dataflow.hpp
```diff
@@ -0,0 +1,119 @@
...
+            for (uint32_t k = seed_rows; k < face_size; k += seed_rows) {
+                noc_async_read(seed_noc_addr, dst_addr, seed_bytes);
+                dst_addr += seed_bytes;
+            }
+        }
+        noc_async_read_barrier();
+
+        // The two lower faces are copies of the two upper ones.
+        noc_async_read(get_noc_addr(tile_addr), tile_addr + tile_faces * face_bytes, tile_faces * face_bytes);
+        noc_async_read_barrier();
```
bbradelTT: Please use a Noc class. See https://github.com/tenstorrent/tt-metal/blob/main/docs/source/tt-metalium/tt_metal/apis/kernel_apis/data_movement/device_api_migration_guide.md

ttnn/cpp/ttnn/kernel_lib/index_tile_dataflow.hpp
```diff
@@ -0,0 +1,119 @@
...
+                for (uint32_t m = 0; m < face_size / 2; ++m) {
+                    seed[r * (face_size / 2) + m] = ((value + 1) << 16) | value;
+                    value += 2;
+                }
+            }
+            // A baby-RISCV store can retire before its write lands in L1, and the RISCV core and
+            // the NoC are different L1 clients with no program-order guarantee between them. Read
+            // the last word back — a blocking load — so the seed is in L1 before the NoC reads it.
+            const uint32_t last_word = seed[seed_words - 1];
+            asm volatile("" ::"r"(last_word) : "memory");
```
bbradelTT: Will this work on WH?

Would what DeepSeek does be better?
writer_deepseek_grouped_gate.cpp
```
(void)ckernel::load_blocking(index_cb_ptr + ((columns_per_face / 2) - 1));
```

ttnn/cpp/ttnn/kernel_lib/index_tile_dataflow.hpp
kinginu: Done.

ttnn/cpp/ttnn/kernel_lib/index_tile_dataflow.hpp
kinginu: Done. Switched to `ckernel::load_blocking`, same as the DeepSeek writer.

To answer the question: the old form did not actually work, on WH or BH. An empty `asm volatile` emits no instruction, so nothing consumed the loaded register and nothing stalled on it. Compiling the kernel for Blackhole, the old code was a bare `lw a5,60(t4)` whose destination is overwritten by the next instruction. It constrained the compiler but did not wait for the load to return, which is the part that implies L1 has processed the preceding stores. `load_blocking` emits `lw` followed by `and x0,x0,a5`, so the pipeline blocks. Thanks for catching it.

bbradelTT (approved):

kinginu:
@jberkowitzTT Resolved conflicts. Could you ping codeowners for this one?

Issue #51308, opened 2026-07-28
Metal 2.0 port: data_movement/untilize_with_unpadding


mcw-anasuya:
Port `ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding` to Metal 2.0 / Host 2.0 APIs (Gen1 WH/BH target).

Parent: #41908

## Scope

One `descriptor`-concept device operation, `UntilizeWithUnpaddingDeviceOperation`, with five program factories. **Four are in scope for this port** (one row per factory in the DFB/Metal 2.0 Porting sheet):

* `UntilizeWithUnpaddingSingleCoreProgramFactory`
* `UntilizeWithUnpaddingMultiCoreInterleavedProgramFactory`
* `UntilizeWithUnpaddingMultiCoreBlockInterleavedProgramFactory`
* `UntilizeWithUnpaddingMultiCoreNDShardedProgramFactory`

**Out of scope — Device 2.0 prerequisite:** `UntilizeWithUnpaddingMultiCoreShardedProgramFactory`. It instantiates two kernels still on Device 1.0 free-function idioms, both in the shared `ttnn/cpp/ttnn/kernel/` pool and both reached only through this factory: `dataflow/writer_unary_stick_layout_interleaved_blocks.cpp`. Routes to the Device 2.0 migration team.

Blocked factories stay on the legacy `create_descriptor` path; the framework dispatches per factory, so a mixed-concept `program_factory_t` builds and runs correctly.

## Audit verdict: RED at op level; GREEN 4-factory subset

The subset above clears every gate — `descriptor`-concept, no custom hash, no `get_dynamic_runtime_args`, no `override_runtime_arguments`, no pybind descriptor, no offset-folded base pointers, Device 2.0 satisfied for every referenced kernel. A scoped brief is issued for it.

Audit report: `.../untilize_with_unpadding/METAL2_PREPORT_AUDIT.md` · Porter brief: `.../untilize_with_unpadding/METAL2_PORT_BRIEF.md`

Outcome: 3 of the 4 landed

## Notes

* WH/BH only — no Quasar emulation. The independent clone at `experimental/quasar/untilize_with_unpadding/` is out of scope.
* Basic Metal 2.0 port (no functional changes expected); style-fixup step deferred.
* Shared-kernel coupling: this op and `data_movement/untilize` share five kernels.
[...]

PR #53626, opened 2026-08-19
Fix HEIGHT_SHARDED page stride in untilize_with_unpadding

mcw-anasuya:
### Issue 
Related to https://github.com/tenstorrent/tt-metal/issues/51308

### PR Category
Bug fix

### Problem
`UntilizeWithUnpaddingMultiCoreInterleavedProgramFactory` bakes an explicit page size into its writer, which passes it as the third argument of `TensorAccessor`. For a sharded destination, that argument is stored verbatim and used as the page stride. The factory only overwrites that value for BLOCK/WIDTH-sharded outputs, where a logical row spans several shards. A HEIGHT_SHARDED output fell through to `unpadded_row_size_bytes`, which is the buffer's logical page size, not physical stride. So whenever the output row size is not a multiple of the buffer alignment, page N is addressed at `N * page_size()` instead of `N * aligned_page_size()`. The result produced  arbitrary output widths.

It went unnoticed because interleaved input to HEIGHT_SHARDED output was not covered in `test_untilize_with_unpadding.py`.

### Fix
HEIGHT_SHARDED takes the destination buffer's aligned page size.

### Notes for reviewers
This was found by the Metal 2.0 pre-port audit of `untilize_with_unpadding`, which classified the site as Class 3 (latent bug) and gated `MultiCoreInterleavedProgramFactory` on it.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:anasuya/fix_untilize_with_unpadding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:anasuya/fix_untilize_with_unpadding)

### DIFF
--- a/ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding/device/factories/untilize_with_unpadding_multi_core_interleaved_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding/device/factories/untilize_with_unpadding_multi_core_interleaved_program_factory.cpp
@@ -103,13 +103,19 @@ tt::tt_metal::ProgramDescriptor UntilizeWithUnpaddingMultiCoreInterleavedProgram
      */
     // BLOCK/WIDTH-sharded outputs split a logical row across shards - the writer's
     // noc_async_write_sharded needs the per-shard page size (shard width in bytes), not the full
-    // row size, to do that split correctly. HEIGHT-sharded and interleaved outputs keep a whole row
-    // per page, same as the full row size.
+    // row size, to do that split correctly. A HEIGHT-sharded output keeps a whole row per page, but
+    // the sharded accessor strides by exactly the page size it is handed and the allocator pads
+    // each row up to the buffer alignment. It needs the padded stride, not the row size. Interleaved
+    // outputs keep a whole row per page, same as the full row size.
     uint32_t writer_page_size = unpadded_row_size_bytes;
     const auto& out_mem_config = output.memory_config();
-    if (out_mem_config.is_sharded() && (out_mem_config.memory_layout() == TensorMemoryLayout::BLOCK_SHARDED ||
-                                        out_mem_config.memory_layout() == TensorMemoryLayout::WIDTH_SHARDED)) {
-        writer_page_size = out_mem_config.shard_spec().value().shape[1] * output.element_size();
+    if (out_mem_config.is_sharded()) {
+        if (out_mem_config.memory_layout() == TensorMemoryLayout::BLOCK_SHARDED ||
+            out_mem_config.memory_layout() == TensorMemoryLayout::WIDTH_SHARDED) {
+            writer_page_size = out_mem_config.shard_spec().value().shape[1] * output.element_size();
+        } else {
+            writer_page_size = static_cast<uint32_t>(dst_buffer->aligned_page_size());
+        }
     }
     std::vector<uint32_t> writer_ct_args = {
         (input_cb_data_format == tt::DataFormat::Float32 or input_cb_data_format == tt::DataFormat::UInt32 or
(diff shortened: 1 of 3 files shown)


ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding/device/factories/untilize_with_unpadding_multi_core_interleaved_program_factory.cpp
```diff
@@ -103,13 +103,19 @@ tt::tt_metal::ProgramDescriptor UntilizeWithUnpaddingMultiCoreInterleavedProgram
...
     const auto& out_mem_config = output.memory_config();
-    if (out_mem_config.is_sharded() && (out_mem_config.memory_layout() == TensorMemoryLayout::BLOCK_SHARDED ||
-                                        out_mem_config.memory_layout() == TensorMemoryLayout::WIDTH_SHARDED)) {
-        writer_page_size = out_mem_config.shard_spec().value().shape[1] * output.element_size();
+    if (out_mem_config.is_sharded()) {
+        if (out_mem_config.memory_layout() == TensorMemoryLayout::BLOCK_SHARDED ||
+            out_mem_config.memory_layout() == TensorMemoryLayout::WIDTH_SHARDED) {
+            writer_page_size = out_mem_config.shard_spec().value().shape[1] * output.element_size();
+        } else {
+            writer_page_size = static_cast<uint32_t>(dst_buffer->aligned_page_size());
```
Aswinmcw, line 117: This is only correct while a HEIGHT-sharded output keeps exactly one shard per row: `writer_page_size` is used for two things in the kernel — the accessor stride and, in `noc_async_write_sharded`, the per-shard chunk size (`write_size = min(remaining, page_size - offset)`). They can be the same value here only because `pages_per_row == div_up(tensor_W, shard_W) == 1` takes the single-write early-return path. If a HEIGHT-sharded output ever arrives with a shard width narrower than the padded row, the split path would chunk by the *aligned* size while each shard row only holds `shard_W * element_size` logical bytes, so the rows past the first would be off by the padding. Worth a `TT_FATAL` on `shard_spec().shape[1] == output.padded_shape()[-1]` in this branch (or at least stating the assumption in the comment) so that combination fails loudly instead of writing shifted data.

Aswinmcw (approved):
One note below.

ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding/device/factories/untilize_with_unpadding_multi_core_interleaved_program_factory.cpp
mcw-anasuya: `TensorSpec` already enforces that check. HEIGHT_SHARDED requires shard width == padded row width (`tensor_spec.cpp:39-42`). I tried output width and shard width of 64 and 32, 62 and 32, and 31 and 16,  all three fail. So a TT_FATAL would be dead code.

amorrisonTT (approved):

PR #53633, opened 2026-08-19
LTX-2.3: make FPS real conditioning + tune conv3d blocking for 1080p/153f/25fps

rsalmanTT:
Makes FPS real generation conditioning in the LTX-2.3 pipelines, and tunes conv3d blocking for the 153-frame shape, so 1080p / ~6s / 25fps can be served at full speed.

Driven by a Console requirement for a single configuration: **1080p, 6 seconds, 25 fps**. Production served 1920×1088 / 145 frames / **24** fps.

## 1. FPS was silently half-wired

`fps` was hardcoded to `24` at the three `VideoPixelShape` sites, and `prepare_av_cross_pe`'s `fps` parameter was never passed, so it defaulted to 24 as well. Only `decode_audio` and the exporter honoured the requested rate.

That combination is worse than ignoring FPS outright. `rope_ltx.py:346` scales the A/V cross-PE temporal axis into *seconds* by dividing by FPS, and `AudioLatentShape` derives the audio latent length from `num_frames / fps`. So requesting 25 built the latents and the A/V alignment on a 24 fps timeline, while `decode_audio` trimmed audio to `num_frames/25` and the container declared 25. Total lengths matched — so it looked correct — but the audio was generated against a 6.04 s video played back over 5.80 s: **lip sync drifts ~0.24 s across a 6 s clip.**

FPS is now pipeline-level state, not a per-call argument: it changes `audio_N`, which changes the traced shape, so it must be fixed at construction alongside `num_frames`/`height`/`width`. `generate()` rejects a mismatched value rather than substituting silently.

**At `fps=24` this is a no-op**, verified on a 4×8 BH Galaxy: the 145-frame shape reproduces bit-for-bit with **3628/3628 JIT cache hits**. Had any compile-time argument shifted, those programs would have hashed differently and appeared as misses. 25 fps itself is free — 6.6 s at 145f/25 vs 6.7 s at 145f/24.

## 2. conv3d blocking for latent T=20

153 frames (6.12 s at 25 fps; 150 isn't `8k+1`) moves the decoder's latent T from 19 to 20, so the conv3d T-chain walks 22/41/79/155 instead of 21/39/75/147 and misses `_BLOCKINGS` at 15 sites. Five fell through to the hardcoded `T=1 H=1 W=1` default, 20–30× off optimal.

| Stage | Before | After | 145f@24 baseline |
|---|---|---|---|
| Stage 1 denoise | 2.41 s | 2.39 s | 2.35 s |
| Latent upsample | 0.62 s | **0.16 s** | 0.18 s |
| Stage 2 denoise | 2.91 s | 2.88 s | 2.61 s |
| VAE decode | 5 s | **1 s** | 1 s |
| **Total** | **12.06 s** | **6.68 s** | 6.72 s |

Biggest wins: `512→512 T=79` **31.8×** (×9 invocations, ~2.4 s alone), `512→4096 T=41` 23.5×, `512→512 T=41` 23.2×, `1024→1024 T=22` 17.3×, `256→256 T=155` 7.2× (×12).

**153f @ 25fps is now at parity with the 145f @ 24fps baseline** despite 5.3% more tokens — the fresh sweep beats the existing 2026-06 entries.

Three keys still fall back and cost nothing measurable: `128→1024 T=22` (×2 sites) and the `1024→4096 (1,3,3) T=20` upsampler. The upsampler's sweep could not run, but upsample landed at 0.16 s *without* an entry — faster than baseline — so its 0.62 s was collateral from badly-blocked decode work queued ahead of it, not its own conv.

## 3. Verification

Six of the twelve entries use temporal blocking, and `conv3d.py:313` documents a prior entry disabled because `T_out_block=4` caused a frame-24-25 artifact — so blocking can affect correctness, not just speed. Three clips were generated and checked with a periodic-seam and flat-region detector, using the pre-tune output as a control: all clean, **stride-6 seam contrast below baseline** (+0.233 / +0.289 vs +0.363), with frame-level stills inspected by eye. Output verified 1920×1088 / 25 fps / 153 frames / 6.120 s, h264 + AAC.

## 4. Also included

- **Trace-memory instrumentation** (`TT_DIT_TRACE_MEM_LOG=1`, off by default). `ttnn.get_memory_view` already exposes TRACE/DRAM/L1 but nothing in `tt_dit` used it, so per-trace device memory was unmeasurable from Python. Measured at 1920×1088/145f: denoise traces 91.31 and 97.69 MB, VAE decoder 4.94 MB, vocoders 37.06 and 29.44 MB, Gemma encoder 20.31 MB — **280.75 MB per shape against a 476.88 MB region**. Trace bytes are near-independent of tensor volume (stage 1→2 is 4× the tokens for +7%; 720p→1080p is 2.9× for +2.4%), because a trace holds the dispatch command stream and the op graph is fixed by architecture.
- **`bruteforce_conv3d_sweep_ltx.py`** — 15 LTX layer specs reusing the wan2_2 harness.
- Pre-existing **latent whiteness/zeros diagnostics** in the distilled pipeline, for triaging noise outputs; already in the working tree and unrelated to the FPS change.

## Notes for future sweeps

- `hw_product` is an **exact-match** filter. Layers whose per-device spatial dims can't reach it (H=5, W=4 → max 20) sweep to zero combos.
- Relaxing it explodes the search space *and* hits the documented BH hang on non-32 products. Two H=5/W=4 layers hung this way and were abandoned at negligible cost.
- A hung Galaxy needs **`tt-smi -glx_reset`** before another device kernel runs; plain `tt-smi -r` fails in topology discovery.

## Dependency

`tenstorrent/tt-inference-server` has a companion PR that flips the served shape to 153f/25fps. **That one must land after this**, or generation takes 12.1 s and `LTX_FPS=25` raises `ValueError`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rsalman/ltx-25fps-1080p-6s)

### DIFF
--- a/tests/pipeline_reorg/models_e2e_tests.yaml
+++ b/tests/pipeline_reorg/models_e2e_tests.yaml
@@ -1152,6 +1152,14 @@
     export TT_MESH_HOST_RANK=0
     export TT_METAL_SHM_TRACKING_DISABLED=1
     export TT_MESH_GRAPH_DESC_PATH="${TT_METAL_HOME}/models/tt_dit/tests/models/ltx/scaleout_configs/single_galaxy/mesh_graph_descriptor.textproto"
+    # Cover the shape actually served (Console: 1080p / ~6s / 25fps), not the legacy
+    # 145f@24. FPS is real conditioning -- it sets the audio latent length and scales the
+    # A/V cross-PE -- so without these the 25fps path has no automated coverage and a
+    # regression to a hardcoded 24 would keep every LTX test green. 153 because
+    # 6s x 25fps = 150 does not satisfy (num_frames-1) % 8 == 0.
+    # The 24-vs-25 arithmetic is additionally guarded on CPU by
+    # models/tt_dit/tests/models/ltx/test_fps_conditioning_ltx.py.
+    export NUM_FRAMES=153 FPS=25
     mpirun -np 1 --bind-to none --tag-output --wdir "${TT_METAL_HOME}" \
       python3 -m pytest --timeout 2100 models/tt_dit/tests/models/ltx/test_pipeline_ltx_distilled.py::test_pipeline_distilled -k "4x8sp1tp0nl2_ring_is_fsdp0"
   model: ltx-2.3-fast-t2v
--- a/models/tt_dit/pipelines/ltx/pipeline_ltx.py
+++ b/models/tt_dit/pipelines/ltx/pipeline_ltx.py
@@ -232,6 +232,7 @@ class LTXPipeline:
         num_frames: int = 0,
         height: int = 0,
         width: int = 0,
+        fps: float = 24.0,
         run_warmup: bool = False,
         traced: bool = False,
         extra_transformer_variants: list[tuple[str, list[LoraSpec]]] | None = None,
@@ -246,6 +247,11 @@ class LTXPipeline:
             tensor_parallel=ParallelFactor(factor=self.mesh_device.shape[1], mesh_axis=1),
         )
         self._traced = traced
+        # FPS is pipeline-level state, not a per-request argument: it sets the audio latent
+        # length (audio_frames = round(num_frames / fps * 25)) and scales the A/V cross-PE's
+        # temporal axis into seconds. Both are baked into the captured traces, so changing it
+        # per request would silently replay a trace built for a different shape.
+        self.fps = float(fps)
         self._trace_state: dict[str, LTXTransformerState] = {}
         self._prompt_v = StateTensor()
         self._prompt_a = StateTensor()
@@ -410,6 +416,7 @@ class LTXPipeline:
         num_frames: int = 0,
         height: int = 0,
         width: int = 0,
+        fps: float = 24.0,
         **extra_pipeline_kwargs,
     ) -> "LTXPipeline":
         """Auto-configure mesh-shape defaults and forward into ``__init__``.
@@ -500,6 +507,7 @@ class LTXPipeline:
             num_frames=num_frames,
             height=height,
             width=width,
+            fps=fps,
             run_warmup=run_warmup,
             traced=traced,
             **extra_pipeline_kwargs,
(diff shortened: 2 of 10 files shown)


tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -1316,6 +1152,14 @@
     export TT_MESH_HOST_RANK=0
     export TT_METAL_SHM_TRACKING_DISABLED=1
     export TT_MESH_GRAPH_DESC_PATH="${TT_METAL_HOME}/models/tt_dit/tests/models/ltx/scaleout_configs/single_galaxy/mesh_graph_descriptor.textproto"
+    # Cover the shape actually served (Console: 1080p / ~6s / 25fps), not the legacy
+    # 145f@24. FPS is real conditioning -- it sets the audio latent length and scales the
+    # A/V cross-PE -- so without these the 25fps path has no automated coverage and a
+    # regression to a hardcoded 24 would keep every LTX test green. 153 because
+    # 6s x 25fps = 150 does not satisfy (num_frames-1) % 8 == 0.
+    # The 24-vs-25 arithmetic is additionally guarded on CPU by
+    # models/tt_dit/tests/models/ltx/test_fps_conditioning_ltx.py.
```
mtairum, line 1161: Approving to unblock, But ideally remove or reduce these comments to keep things cleaner 👍

mtairum (approved):

yieldthought:
Codex review:

**[P2] Pass pipeline FPS to video self-attention RoPE as well** — [models/tt_dit/pipelines/ltx/pipeline_ltx_distilled.py:268](https://github.com/tenstorrent/tt-metal/blob/502f19726e578cfe68687305992d35923c62b8e2/models/tt_dit/pipelines/ltx/pipeline_ltx_distilled.py#L268)

At fps=25, prepare_av_cross_pe receives 25 but prepare_video_rope in _prepare_stage_statics and LTXPipeline.call_av still omits fps, selecting its 24.0 default. The video self-attention temporal coordinates are therefore scaled by 1/24 while cross-attention uses 1/25. The former is 25/24 of the requested timeline, so non-24-fps generation still has inconsistent temporal conditioning. Suggestion: pass fps=self.fps to both prepare_video_rope calls and add a call-site regression checking both builders receive the constructed rate.

**[P2] Preserve the existing 11x10 matmul tuning entries** — [models/tt_dit/utils/matmul.py:191](https://github.com/tenstorrent/tt-metal/blob/502f19726e578cfe68687305992d35923c62b8e2/models/tt_dit/utils/matmul.py#L191)

The new LTX block replaces 28 existing 11x10 entries, including Flux2 input projections. Those callers still select 11x10 on Blackhole Galaxy. For example, (1024,128,768) changes from (12,2,5,(1,1)) to the generic (8,8,8,(2,2)); (2048,128,1536) and (4096,128,768) also lose their swept settings. The LTX additions thus discard unrelated model tuning and change execution configurations without a replacement for those shapes. Suggestion: restore the removed entries and append the LTX entries to the shared table. None of the removed keys collides with a new key.

Verification: manual source review and host-only argument-binding/config-selection probes. No device tests or performance measurements were run.

Issue #45086, opened 2026-05-23
sfpi: RoundMode:NearestEven is actually round to nearest with ties away from zero for float→float, float→int, int→int


jasondavies:
See [ISA docs for SFPSTOCHRND](https://github.com/tenstorrent/tt-isa-documentation/blob/main/BlackholeA0/TensixTile/TensixCoprocessor/SFPSTOCHRND.md).

Any conversions that emit SFPSTOCHRND and use "round to nearest" are _not_ round to nearest even.  They're round to nearest with ties away from zero.

jasondavies:
Suggest calling it `RoundMode::Nearest` and drop the `Even`.

nathan-TT:
Plan is to keep ::Nearest and have aliases ::NearestAway and ::NearestEven for the specific arches that have that behavior.
There's a deployment ordering constraint as metal uses ::NearestEven on non-quasar arches.

Also deprecate and alias the MOD1 constants that refer to 'RNE' when it's not.

ETA: considering renaming ::Stochastic to ::NearestStochastic, as that's the behavior

PR #53639, opened 2026-08-19
[Bug fix] typecast: uint16 destination truncates like the other integer destinations

### LINKED ISSUE #51655: [Bounty $1,000] ttnn.typecast to uint16 rounds while every other integer destination and the host path truncate
**Component / Area:** kernels (LLK/SFPU) and ttnn ops

**Issue Type:** Bad Outputs

## Observed

`ttnn.typecast` to `uint16` rounds to nearest with ties away from zero. Every other integer destination of the same op truncates toward zero, as do torch, C++ `static_cast` and ttnn's own host-side conversion.

Three consequences:

* `ttnn.typecast(x, ttnn.uint16)` and `ttnn.typecast(x, ttnn.int32)` disagree on the same input.
* The same op disagrees with itself depending on where the tensor lives: on device it rounds, on host it truncates.
* Nothing aborts and no test fails, because three separate goldens were written to match the device rather than torch. Two of them were changed to match it deliberately, on a premise that turns out to be about a different function entirely.

No shipped model is affected today: every `models/` site converting to `uint16` is converting integer-valued data, where rounding and truncation agree. What is wrong is that the public API disagrees with torch and with itself.

## Expected

One rounding rule for every integer destination, and the same rule on the host path and the device path. Truncation toward zero is the rule that torch, C++ `static_cast`, ttnn's host path and three of the four device destinations already implement.

## Steps to reproduce

```python
import torch, ttnn

DSTS = [("uint8", ttnn.uint8, torch.uint8), ("uint16", ttnn.uint16, torch.uint16),
        ("uint32", ttnn.uint32, torch.uint32), ("int32", ttnn.int32, torch.int32)]

vals = [0.5, 1.5, 2.5, 3.5, 4.5, 0.25, 0.75]
x = torch.zeros(1, 1, 32, 32, dtype=torch.float32)
x[0, 0, 0, :len(vals)] = torch.tensor(vals, dtype=torch.float32)
tt = ttnn.from_torch(x, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)

for name, tt_dt, torch_dt in DSTS:
    got = ttnn.to_torch(ttnn.typecast(tt, tt_dt)).to(torch.int64).flatten()[:len(vals)]
    ref = x.flatten()[:len(vals)].to(torch_dt).to(torch.int64)
    print(f"{name:7s} ttnn={got.tolist()}  torch={ref.tolist()}")
```

Measured on Blackhole P300, and byte for byte the same on Wormhole n300:

```
uint8   ttnn=[0, 1, 2, 3, 4, 0, 0]  torch=[0, 1, 2, 3, 4, 0, 0]
uint16  ttnn=[1, 2, 3, 4, 5, 0, 1]  torch=[0, 1, 2, 3, 4, 0, 0]
uint32  ttnn=[0, 1, 2, 3, 4, 0, 0]  torch=[0, 1, 2, 3, 4, 0, 0]
int32   ttnn=[0, 1, 2, 3, 4, 0, 0]  torch=[0, 1, 2, 3, 4, 0, 0]
```

Host against device, no torch involved:

```python
dev  = ttnn.to_torch(ttnn.typecast(tt, ttnn.uint16))          # 2.5 -> 3
host = ttnn.to_torch(ttnn.from_torch(x, dtype=ttnn.uint16))   # 2.5 -> 2
```

All 65536 bfloat16 bit patterns, counting the finite ones whose truncation lands in the destination range. The same sweep on Wormhole n300 gives the same 576:

| conversion | in-range finite inputs | disagree with truncation |
|---|---|---|
| bf16 -> uint16 | 18305 | **576** |
| bf16 -> uint8 | 17280 | 0 |
| bf16 -> int32 | 40449 | 0 |
| bf16 -> uint32 | 20353 | 0 |

576 is exactly the count of bf16 patterns in `[0, 65536)` whose fractional part is at least one half: 128 in `[0.5, 1)` plus 64 in each of the seven binades from `[1, 2)` to `[64, 128)`. Of the 18305 uint16 inputs, 64 distinguish round-to-nearest-even from round-half-away-from-zero, and the hardware matches half-away on all 64.

## Mechanism

`uint16` is the only float-to-integer destination that hands the conversion to the hardware float-to-int instruction. The other three build the integer by extracting the mantissa and shifting, which drops the fractional bits and therefore truncates by construction.

`tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_typecast.h`, `calculate_typecast_fp32_to_uint16` at `:34`:

```cpp
TTI_SFPSWAP(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 9);                              // max(0, x)
TTI_SFP_STOCH_RND(0, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
```

`calculate_typecast_fp32_to_int32` at `:183`, matched by `fp32_to_uint32` at `:218` and `fp32_to_uint8` at `:939`:

```cpp
// result = exman(in, sfpi::MantissaMode::ImplicitOne) << (exp - 23)
TTI_SFPEXMAN(0, p_sfpu::LREG0, p_sfpu::LREG1, 0);
TTI_SFPSHFT(0, p_sfpu::LREG2, p_sfpu::LREG1, 0);
```

The two families were never made to agree. The host path is the same story from the other side: `tt_metal/impl/tensor/tensor_apis.cpp:473`, reached from `typecast.cpp:19-30` whenever the tensor is host-resident, is a `static_cast`, and C++ float-to-integral conversion discards the fractional part.

## Affected call sites

| conversion | rounding | file:line |
|---|---|---|
| float -> uint16 | rounds | `blackhole/.../ckernel_sfpu_typecast.h:34` |
| int32 -> uint16 | rounds | `blackhole/.../ckernel_sfpu_typecast.h:534` |
| float -> int32 | truncates | `blackhole/.../ckernel_sfpu_typecast.h:183` |
| float -> uint32 | truncates | `blackhole/.../ckernel_sfpu_typecast.h:218` |
| float -> uint8 | truncates | `blackhole/.../ckernel_sfpu_typecast.h:939` |
| float -> uint16, int32 -> uint16 | rounds | `wormhole_b0/.../ckernel_sfpu_typecast.h:51`, `:96`, `:561`, `:902`, `:940` |
| host, all destinations | truncates | `tensor_apis.cpp:473` |

`Float16_b`, `Bfp8_b` and `Bfp4_b` sources all reach the same kernels, so they behave the same way. Layout makes no difference. Quasar's copy of the file has no uint16 typecast at all, so Blackhole and Wormhole are the only architectures involved. The Wormhole rows are measured, not inferred: the repro above and the 65536-pattern sweep were both re-run on an n300 and agree with Blackhole exactly.

For completeness, one more call site uses the same instruction in the same mode and is **not** part of this: `ckernel_sfpu_exp.h:513` rounds a float to uint16 as an intermediate step of the fast exponent approximation, where round to nearest is what the algorithm wants.

## Please complete the following environment information

* OS: Ubuntu 22.04
* Version of software: reproduced on `48cb0736920`; the kernels are unchanged as of `992f3ca634` (current `main` at the time of filing), and every line number above is against that commit
* Hardware: Blackhole P300 and Wormhole n300, both reproduce

Not a regression: the uint16 kernel has behaved this way since the file arrived in this repository (#45897, 2026-06-04), and in tt-llk before that.

## Why no test catches this

Three separate goldens in the tree encode the device behaviour rather than torch's, and two of them carry comments acknowledging the divergence.

`tests/ttnn/python_api_testing/sweep_tests/ttnn_pytorch_ops.py:226-228`:

```python
def _float_to_uint16_clamp(x):
    # Device float_to_uint16 converts to float32 before std::round; match in float32.
    return torch.clamp(torch.floor(x.float() + 0.5).to(torch.int32), min=0, max=65535)
```

That comment, and the merged PR #46816 that introduced it (closing #46237), attribute the rounding to `float_to_uint16` in `ttnn/cpp/ttnn/operations/data_movement/common/common.cpp:489`. That function is a **host** helper that converts a scalar pad value, is marked "not unit tested, use with caution", and is called only from the `pad`, `permute` and `transpose` program factories. `ttnn.typecast` never reaches it. It agrees with the kernel only because `std::round` and the SFPU's default mode both round half away from zero, so the golden was made to match the kernel for a reason that does not hold.

That helper has the same defect one layer up, which I will file separately: it is what `ttnn.pad` uses to convert a scalar pad value for a uint16 tensor, so `ttnn.pad(x_uint16, value=2.5)` gives 3 while the same call on an int32 or uint32 tensor gives 2, as torch does. Measured, both layouts.

In the dispatch function that calls `_float_to_uint16_clamp`, `bfloat8_b -> uint16` (`:254`) and `bfloat4_b -> uint16` (`:291`) use plain truncation instead. All three arms reach the same kernel. The block-float arms pass only because block-float quantisation at the tested range yields integer-valued data, where the two contracts coincide.

`tests/ttnn/nightly/unit_tests/operations/data_movement/test_copy_ops.py:44-46` needs a second, different golden for the same op:

```python
if output_dtype == ttnn.uint16:
    # Host fp32→uint16 truncates; device float_to_uint16 rounds in float32 first.
    return torch.clamp(torch_tensor.to(torch.int32), min=0, max=65535)
```

`tt_metal/tt-llk/tests/python_tests/helpers/golden_generators.py:1779-1781` states the inconsistency and then arranges for the test not to see it:

```python
# int32/uint32 truncate; uint16/uint8 round. Whole-number
# stimuli make trunc == round, so trunc models both exactly.
```

The uint8 half of that claim is wrong: `calculate_typecast_fp32_to_uint8` uses the mantissa-shift form on both architectures and truncates.

Which stimuli reach which suite is worth stating precisely, because it is not uniform. `typecast_test_helpers.make_typecast_test_input` draws uniform floats for a float source, so the ttnn eltwise and copy-ops tests **can** see the difference; that is exactly why the rounding golden is load-bearing there, and why 24 of them fail the moment the kernel truncates and the golden does not. The LLK suite and the integer-source paths feed whole numbers, where every rounding mode agrees. Neither suite feeds an exact half-integer deterministically, which is the stimulus that separates truncation from both round-to-nearest modes.

kinginu:
### PR Category

Bug fix. Closes #51655.

### Summary

`ttnn.typecast` to `uint16` rounds to nearest with ties away from zero, while `int32`, `uint32`, `uint8` and the host path all truncate toward zero, as torch does. The analysis and the exhaustive measurement are in #51655.

`uint16` is the only destination that hands the conversion to the hardware float-to-int instruction; the others extract the mantissa and shift, which truncates by construction. `SFP_STOCH_RND`'s rounding-mode operand was left at `0`, which the ISA calls `RND_NEAREST`. This passes round to zero at the five uint16 sites instead, through a named constant, because sfpi calls mode 0 `SFPSTOCHRND_RND_EVEN` (#45086, closed but still shipping under that name in the pinned 7.67.0) and a reader of these call sites would otherwise have to know that the neighbouring constant lies.

Measured on Blackhole P300 across all 65536 bfloat16 patterns: uint16 goes from 576 disagreements with truncation to 0, and int32, uint32 and uint8 stay at 0. It costs nothing, and the cheapest way to see that is the generated code rather than a cycle count: the math ELF's `.text` is 924 bytes before and after, and the two differ in exactly one byte, the rounding-mode immediate of an instruction that was already there. Same instruction, same count, same order, so there is nothing for a cycle to hide in. `perf_eltwise_typecast.py` also reports the same cycles in `MATH_ISOLATE` and `L1_TO_L1`.

**Blackhole only, and the test suite has to say so.** Round to zero is new in Blackhole; Wormhole's `SFPSTOCHRND` has two modes and neither truncates. Passing mode 2 there is not a compile error, it is simply ignored: I tried it on an n300 and the uint16 output did not change by a single bit. That matters to this PR because the goldens are shared, so correcting them unconditionally turns 28 tests red on Wormhole, the 24 `eltwise_typecast` cases plus the 4 uint16 cases of the new test. `device_truncates_float_to_uint16()` in `typecast_test_helpers.py` gates the two places that care, and its docstring says that deleting the predicate is the last step of #51655.

I did look for a Wormhole fix rather than assuming there wasn't one. `max(0, x) - 0.5` fed to round-to-nearest-ties-away is exactly truncation over the whole uint16 domain, it needs one extra instruction, and on Wormhole it produces the right answer for every value I threw at it. It does not fit the existing kernel: the 16-bit Dest path runs through `SFPLOADMACRO`, and a macro takes at most one MAD instruction while the `SFPSWAP` doing `max(0, x)` on the Simple sub-unit requires that slot to hold `SFPNOP` at the same time. Scheduling anything at all there breaks it, including a `+0.0` no-op, which is how I established that it is the slot and not the arithmetic. Forcing the plain loop instead costs 13% of the math thread by measurement, so Wormhole needs either that trade or a redesign of the clamp, and both belong in their own change.

### Tests

`test_typecast_rounding_fractional` is new. It feeds deterministic quarter steps, including exact half-integers, to all four integer destinations from both a bfloat16 and a float32 source. Half-integers are the discriminator: truncation, ties-away and ties-to-even all disagree on them, and nothing else in the suite feeds one deterministically. Both sources matter because a float32 source sets `preserve_fp32_precision` and so exercises the 32-bit Dest path of the uint16 kernel while a bfloat16 source exercises the 16-bit one. Without the fix, the uint16 cases fail on Blackhole and the others pass.

Three goldens encoded the old behaviour and are corrected: `_float_to_uint16_clamp` in `ttnn_pytorch_ops.py` now truncates where the kernel truncates, so all three `-> uint16` arms of that dispatch agree; `_host_typecast_golden` in `test_copy_ops.py` says why the host golden still cannot fall through to the device one; `golden_generators.py` had a comment stating the divergence, and the code there already truncated.

`test_eltwise_typecast.py` and `test_copy_ops.py` together, same counts on both machines:

| | result |
|---|---|
| Blackhole P300 | 981 passed, 0 failed, 502 skipped |
| Wormhole n300 | 981 passed, 0 failed, 502 skipped |

Without the golden corrections, 24 fail on Blackhole. Without the arch predicate, 28 fail on Wormhole. One earlier Blackhole run failed `bfloat4_b -> uint16` at `[1, 3, 320, 320]`; that test draws unseeded random input against a block-float quantisation golden, and 30 consecutive reruns of the block-float subset give 0 failing runs, so I am reporting it as a pre-existing flake rather than leaving it to surprise someone.

### Not covered

Three FP32 values are still rounded away from zero on Blackhole: `0x3F7FFFFE`, `0x3F7FFFFF` and `0x3FFFFFFF`. The ISA documents this as a bug in the round-to-zero mode, they were equally wrong before this change, and no bfloat16 or block-float input can encode them, so only a float32 source can reach them. The `max(0, x) - 0.5` construction described above does get them right, and it is the reason I know that: it costs an instruction and, on the 16-bit Dest path, the `SFPLOADMACRO` fast path. Say the word if you would rather trade that for the last three encodings.

Quasar has no uint16 typecast at all, so Blackhole and Wormhole are the only architectures involved.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1218118069917559

### DIFF
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_typecast.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_typecast.h
@@ -30,6 +30,14 @@ constexpr std::uint32_t SFPSTORE_MODE_SWAP_HI_LO16 = 9;
 // SFPGT mod1 selector that sets the destination to all-ones (-1) when the comparison is true.
 constexpr std::uint32_t SFPGT_MOD1_SET_ALL_ONES = 8;
 
+// SFP_STOCH_RND rounding mode. RND_ZERO truncates toward zero, which is what the mantissa-shift
+// destinations (int32, uint32, uint8), the host static_cast and torch all do; the default operand 0
+// is round to nearest with ties away from zero. Three FP32 inputs are a documented hardware
+// exception that RND_ZERO cannot fix: 0x3F7FFFFE, 0x3F7FFFFF and 0x3FFFFFFF still round away from
+// zero (tt-isa-documentation, BlackholeA0 SFPSTOCHRND_FloatInt.md). No bfloat16 or block-float input
+// can encode them, and round to nearest got them wrong as well.
+constexpr std::uint32_t SFPSTOCHRND_TRUNCATE = sfpi::SFPSTOCHRND_RND_ZERO;
+
 template <bool APPROXIMATION_MODE, int ITERATIONS, bool is_fp32_dest_acc_en>
 inline void calculate_typecast_fp32_to_uint16() {
 #ifdef DISABLE_SFPLOADMACRO
@@ -37,7 +45,8 @@ inline void calculate_typecast_fp32_to_uint16() {
     for (int d = 0; d < ITERATIONS; d++) {
         TTI_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::DEFAULT, ADDR_MOD_7, 0);
         TTI_SFPSWAP(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 9);
-        TTI_SFP_STOCH_RND(0, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
+        TTI_SFP_STOCH_RND(
+            SFPSTOCHRND_TRUNCATE, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
         if (is_fp32_dest_acc_en) {
             TTI_SFPSTORE(p_sfpu::LREG0, SFPSTORE_MODE_SWAP_HI_LO16, ADDR_MOD_6, 0);
         } else {
@@ -80,7 +89,8 @@ inline void calculate_typecast_fp32_to_uint16() {
         for (int d = 0; d < ITERATIONS; d++) {
             TTI_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::DEFAULT, ADDR_MOD_7, 0);
             TTI_SFPSWAP(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 9);
-            TTI_SFP_STOCH_RND(0, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
+            TTI_SFP_STOCH_RND(
+                SFPSTOCHRND_TRUNCATE, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
             TTI_SFPSTORE(p_sfpu::LREG0, SFPSTORE_MODE_SWAP_HI_LO16, ADDR_MOD_6, 0);
         }
     }
@@ -538,7 +548,8 @@ inline void calculate_typecast_int32_to_uint16() {
         TTI_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::INT32, ADDR_MOD_7, 0);
         TTI_SFPCAST(p_sfpu::LREG0, p_sfpu::LREG0, 0);
         TTI_SFPSWAP(0, p_sfpu::LCONST_0, p_sfpu::LREG0, 9);
-        TTI_SFP_STOCH_RND(0, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
+        TTI_SFP_STOCH_RND(
+            SFPSTOCHRND_TRUNCATE, 0, 0, p_sfpu::LREG0, p_sfpu::LREG0, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
         TTI_SFPSTORE(p_sfpu::LREG0, SFPSTORE_MODE_SWAP_HI_LO16, ADDR_MOD_6, 0);
     }
 #else
@@ -874,7 +885,7 @@ inline void init_typecast_fp32_to_uint16() {
     TTI_SFPSWAP(0, p_sfpu::LCONST_0, 12, 0xf);  // L[VD] = max(0, L[VD])
 
     // InstructionTemplate[1]
-    TTI_SFP_STOCH_RND(0, 0, 0, 0, 13, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
+    TTI_SFP_STOCH_RND(SFPSTOCHRND_TRUNCATE, 0, 0, 0, 13, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
 
     // Macro 0
     {
@@ -912,7 +923,7 @@ inline void init_typecast_int32_to_uint16() {
     TTI_SFPSWAP(0, p_sfpu::LCONST_0, 12, 0xf);  // L[VD] = max(0, L[VD])
 
     // InstructionTemplate[1]
-    TTI_SFP_STOCH_RND(0, 0, 0, 0, 13, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
+    TTI_SFP_STOCH_RND(SFPSTOCHRND_TRUNCATE, 0, 0, 0, 13, sfpi::SFPSTOCHRND_MOD1_FP32_TO_UINT16);
 
     // Macro 0
     {
--- a/tests/ttnn/nightly/unit_tests/operations/data_movement/test_copy_ops.py
+++ b/tests/ttnn/nightly/unit_tests/operations/data_movement/test_copy_ops.py
@@ -42,7 +42,8 @@ def _host_typecast_golden(torch_tensor, output_dtype):
     if output_dtype == ttnn.uint8:
         return torch.clamp(torch_tensor, 0, 255).to(torch.uint8)
     if output_dtype == ttnn.uint16:
-        # Host fp32→uint16 truncates; device float_to_uint16 rounds in float32 first.
+        # The host truncates on every architecture. The device now matches it on Blackhole and
+        # still rounds elsewhere, so this cannot fall through to the device golden.
         return torch.clamp(torch_tensor.to(torch.int32), min=0, max=65535)
     return eltwise_typecast(torch_tensor, tt_input_dtype=ttnn.float32, tt_output_dtype=output_dtype)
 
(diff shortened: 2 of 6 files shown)


tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_typecast.h
```diff
@@ -30,14 +30,23 @@ constexpr std::uint32_t SFPSTORE_MODE_SWAP_HI_LO16 = 9;
 // SFPGT mod1 selector that sets the destination to all-ones (-1) when the comparison is true.
 constexpr std::uint32_t SFPGT_MOD1_SET_ALL_ONES = 8;
 
+// SFP_STOCH_RND rounding mode. RND_ZERO truncates toward zero, which is what the mantissa-shift
+// destinations (int32, uint32, uint8), the host static_cast and torch all do; the default operand 0
+// is round to nearest with ties away from zero. Three FP32 inputs are a documented hardware
```
fvranicTT, line 35: Should we in other places round toward zero, too?

kinginu: No. int32, uint32 and uint8 never reach `SFP_STOCH_RND`; they build the integer with `SFPEXMAN` plus `SFPSHFT`, so the fraction is already dropped. Across all 65536 bfloat16 patterns those three disagree with truncation in 0 cases both before and after this change, and uint16 was the only one at 576.

The other six `SFP_STOCH_RND` in this file are `FP32_TO_FP16B`, a float narrowing where nearest is correct. Wormhole is the one place that should change and can't: round to zero is new in Blackhole, and passing mode 2 there is silently ignored.

kinginu:
Rebased onto main. Conflicts resolved.

jberkowitzTT:
Hi @kinginu - I'm seeing some merge conflicts, could you let me know once you're able to fix those?

PR #15665, opened 2024-12-03
#13621: enable default fp32 acc for reduce

bbradelTT:
### Ticket
Link to Github Issue #13621

### Problem description
reduce sum is not very accurate because fp32 acc for reduce was not enabled by default

### What's changed
enable fp32 acc for reduce by default

### Checklist
- [x] Post commit CI passes Between two runs, all jobs passed https://github.com/tenstorrent/tt-metal/actions/runs/12147218201 and https://github.com/tenstorrent/tt-metal/actions/runs/12160961521
- [x] Blackhole Post commit (if applicable) https://github.com/tenstorrent/tt-metal/actions/runs/12187630112
- [x] Model regression CI testing passes (if applicable)  https://github.com/tenstorrent/tt-metal/actions/runs/12187623632/job/33999246179 fails same as main except another random tt-smi reset not working. main: https://github.com/tenstorrent/tt-metal/actions/runs/12189517366
- [x] Device performance regression CI testing passes (if applicable) https://github.com/tenstorrent/tt-metal/actions/runs/12187626769 passes for WH, GS not affected and fails which it does on main https://github.com/tenstorrent/tt-metal/actions/runs/12189542166
- [x] New/Existing tests provide coverage for changes

Issue #13621, opened 2024-10-09
[Bug Report] ttnn.mean op - Data Mismatch 


chandrasekaranpradeep:
**Describe the bug**
The ttnn.mean throws assertion error because of data mismatch between PyTorch and TTNN output and the pcc is dropped to 0.72 when the input tensor of (1, 12, 3200) and dim = -1 is passed to  ttnn.mean op.
For more context, here is the exact error message
```
def assert_with_pcc(expected_pytorch_result, actual_pytorch_result, pcc=0.9999):
        assert list(expected_pytorch_result.shape) == list(
            actual_pytorch_result.shape
        ), f"list(expected_pytorch_result.shape)={list(expected_pytorch_result.shape)} vs list(actual_pytorch_result.shape)={list(actual_pytorch_result.shape)}"
        pcc_passed, pcc_message = comp_pcc(expected_pytorch_result, actual_pytorch_result, pcc)
>       assert pcc_passed, construct_pcc_assert_message(pcc_message, expected_pytorch_result, actual_pytorch_result)
E       AssertionError: 0.7203957195745748
```

**To Reproduce**

Run the following test:
```
import torch
import ttnn
from tests.ttnn.utils_for_testing import assert_with_pcc
from models.utility_functions import torch_random
def test_mean_pcc_issue(device):
    torch.manual_seed(0)

    input_shape = (1, 12, 3200)
    reduce_dim = -1
    
    torch_input_tensor = torch.rand(input_shape, dtype=torch.float32)
    torch_output_tensor = torch.mean(torch_input_tensor, dim=reduce_dim, keepdim=True, dtype=torch.float32)

    input_tensor = ttnn.from_torch(torch_input_tensor, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
    
    output_tensor = ttnn.mean(input_tensor, dim=reduce_dim)
    output_tensor = ttnn.to_torch(output_tensor)
    
    assert_with_pcc(torch_output_tensor, output_tensor)
```

**Expected behavior**
The data mismatch between PyTorch and TTNN output should be resolved.

sdjordjevicTT:
@ntarafdar @sjameelTT can you please help me to find owners for this issue?

ntarafdar:
hey @sdjordjevicTT asking around, its a reduction op who doesn't have an owner , will ask ttnn ppl and get back to you

ntarafdar:
[...]

PR #53643, opened 2026-08-19
[Bug fix] softmax: accumulate the SUM reduce in fp32 regardless of input dtype

### LINKED ISSUE #52045: [ttnn/normalization]: bfloat16 softmax rows do not sum to 1, the SUM reduce accumulates the whole row in a 16-bit Dest
**Component / Area:** ttnn/normalization/softmax

**Issue Type:** Bad Outputs

## Observed

`ttnn.softmax` on a `bfloat16` or `bfloat8_b` tensor does not return a normalised distribution. Row sums, which must be 1.0, reach **0.865 at a row length of 8192**, on default arguments against unmodified `main`.

| row length | `bfloat16` | `bfloat8_b` | `bfloat16`, `fp32_dest_acc_en=True` |
|---|---|---|---|
| 42 | 0.999146 | 1.004272 | 0.998037 |
| 1024 | 0.983110 | 0.988466 | 0.994424 |
| 2048 | 0.944902 | 0.950601 | 0.994914 |
| 4096 | 0.897854 | 0.902487 | 0.994841 |
| **8192** | **0.864987** | **0.869716** | 0.996075 |
| 16384 | 0.943260 | 0.949062 | 0.996661 |
| 32000 | 0.955643 | 0.961536 | 0.999454 |
| 65536 | 0.957284 | 0.962505 | 1.000181 |
| 128000 | 1.012368 | 1.020192 | 1.006430 |

The right column is the same op, input and kernel with `fp32_dest_acc_en` passed through `compute_kernel_config`. It reads 0.994 rather than 1.000 because the same function also sets `math_approx_mode = true` unconditionally, and the approximate `recip_tile` is -0.2% to -3.9% off, always negative. That is a second, smaller defect on the same line and is not what this issue is about.

A `float32` input is unaffected: it already gets an fp32 Dest from the same policy, and reads 0.994842 at w=4096 and 0.995924 at w=8192, the same residual as the right column above.

The error is a near-uniform per-row scale rather than noise: at w=8192 the elementwise device/torch ratio has mean 0.8654 and standard deviation 0.0171.

The curve is not monotonic because the attention program factory switches from `softmax.cpp` to `softmax_large_tensor.cpp` above w=8192, and the large kernel folds partial sums over passes of about 80 tiles, which caps the per-pass damage. Confirmed by reading the compiled kernel names out of the JIT kernel cache, not inferred from `select_program_factory`.

## Expected

Row sums of 1.0 to within `bfloat16` rounding.

`softmax_init_compute_kernel_config` derives `fp32_dest_acc_en` from the input dtype (`softmax_device_operation.cpp:386`), so a 16-bit input gets a 16-bit Dest for the SUM reduce over the exponentials, and that reduce folds all `Wt` tiles of a row into one accumulator.

The generic reduce op sets the same flag to `true` for the same reduction regardless of dtype, at `reduce_op.cpp:111` and `generic_reductions.cpp:480`. It was set that way by #15665, answering #13621, a `ttnn.mean` PCC mismatch: *"reduce sum is not very accurate because fp32 acc for reduce was not enabled by default"*. Softmax has its own compute-config helper and was not in that PR's scope. Measured: `ttnn.sum(dim=-1)` over 8192 positive `bfloat16` values agrees with an fp32 reference to 2.8e-4, at the same `Wt = 256` where softmax is 15% out.

## Steps to reproduce

```python
import torch, ttnn

d = ttnn.open_device(device_id=0)
for w in (42, 1024, 2048, 4096, 8192, 16384, 32000, 65536, 128000):
    torch.manual_seed(0)
    x = (torch.rand((1, 32, w), dtype=torch.float32) * 2 - 1).to(torch.bfloat16)
    t = ttnn.from_torch(x, layout=ttnn.TILE_LAYOUT, device=d)
    out = ttnn.to_torch(ttnn.from_device(ttnn.softmax(t, dim=-1, numeric_stable=True))).to(torch.float32)
    print(f"w={w:7d}  row sum {float(out.sum(-1).mean()):.6f}")
ttnn.close_device(d)
```

No input data required. Bit-for-bit reproducible: two consecutive runs of any configuration below agree to every printed digit.

## Scope

w=8192, 32 rows, uniform[-1,1]. Compute kernels read out of the JIT kernel cache, not inferred.

| entry point | dtype | row sum | with fp32 dest | compute kernel |
|---|---|---|---|---|
| `ttnn.softmax` rank 4 | bfloat16 | 0.86499 | 0.99607 | `softmax` |
| `ttnn.softmax` rank 4 | bfloat8_b | 0.86972 | 0.99855 | `softmax` |
| `ttnn.softmax` rank 5 | bfloat16 | 0.93855 | 0.99555 | `moreh_softmax_w_large` |
| `ttnn.softmax` rank 5 | bfloat8_b | 0.94564 | 0.99932 | `moreh_softmax_w_large` |
| `ttnn.softmax_in_place` | bfloat16 | 0.86499 | 0.99607 | `softmax` |
| `ttnn.scale_mask_softmax` | bfloat8_b | 0.94327 | 0.99962 | `softmax_large_tensor` |
| `ttnn.scale_mask_softmax_in_place` | bfloat8_b | 0.94327 | 0.99962 | `softmax_large_tensor` |
| `ttnn.transformer.attention_softmax` | bfloat16 | 0.86499 | — | `softmax` |
| `ttnn.moreh_softmax` | bfloat16 | 0.93855 | n/a | `moreh_softmax_w_large` |

The sharded kernel `softmax_sharded.cpp` is affected too: (1, 16, 256, 256) gives 0.990845, (1, 2, 1024, 1024) 0.983704, (1, 1, 2048, 2048) 0.947184.

Four compute kernels, both 16-bit dtypes, every entry point. This is a property of the compute-config policy, not of any one kernel.

## Why no test catches it

`tests/ttnn/unit_tests/operations/fused/test_softmax.py::test_large_softmax` runs at these shapes and asserts `pcc>=0.999, rtol=0.10, atol=0.04, frobenius<=0.050`:

| shape | row sum | PCC | max abs err | frobenius | verdict |
|---|---|---|---|---|---|
| (1,24,42) | 1.00015 | 0.9992877 | 2.5e-3 | 0.0187 | PASSES |
| (1,32,32000) | **0.95564** | 0.9993762 | 6e-6 | 0.0477 | PASSES |
| (1,512,32000) | **0.95556** | 0.9993778 | 6e-6 | 0.0477 | PASSES |
| (1,32,128000) | 1.01237 | 0.9993926 | 1e-6 | 0.0209 | PASSES |

It cannot see this by construction: PCC is invariant under a uniform per-row scale, and `atol=0.04` is 6600x the largest absolute error because the values are of order 1/w. Only the frobenius ratio comes close, at 0.0477 against a 0.050 threshold. No test in the tree asserts that a softmax row sums to 1.

## What fixing it costs

`fp32_dest_acc_en=True` halves Dest from 8 tiles to 4. In the interleaved factory that halves the block size, `find_max_divisor(Wt, 8)` to `find_max_divisor(Wt, 4)`, so a row takes twice as many blocks. That is where the time below goes. #15665 accepted the same Dest cost for the same reduction.

L1 does not have to move with it, because the intermediate CBs split into two groups that the op currently treats alike. The bulk buffers, `cb_exps` / `cb_x` / `cb_scale_mask`, are `Wt` deep in the small kernel and capped near 80 tiles in the large one, and hold one value per element: widening them buys element precision, which is not what is wrong, and costs L1 that grows with the row length. The accumulator CBs, `cb_sumexps` and the large kernel's `cb_prev_reduce` ping-pong partner, are one tile each and carry the running sum through L1 once per pass, which is where a 16-bit store undoes what an fp32 Dest earned. Following `fp32_dest_acc_en` on the accumulators alone, in CB bytes on one core:

| row length | compute kernel | `main` | with the fix | |
|---|---|---|---|---|
| 1024 | `softmax` | 174,080 | 161,792 | -12,288 |
| 4096 | `softmax` | 567,296 | 555,008 | -12,288 |
| 8192 | `softmax` | 1,091,584 | 1,079,296 | -12,288 |
| 16384 | `softmax_large_tensor` | 540,672 | 534,528 | -6,144 |
| 32000 | `softmax_large_tensor` | 540,672 | 534,528 | -6,144 |
| 128000 | `softmax_large_tensor` | 540,672 | 534,528 | -6,144 |

Lower at every width, and the `use_large_kernel` switch stays where it is, so no shape changes which kernel it gets.

Time, wall clock, steady state, 50 calls back to back with one synchronize, median of 5, both arms on the same chip:

| shape | before, us/call | after, us/call | delta |
|---|---|---|---|
| 1024 x 1024 | 22.6 | 22.8 | +0.9% |
| 2048 x 2048 | 61.1 | 65.6 | +7.4% |
| 4096 x 4096 | 192.7 | 207.7 | +7.8% |
| 32 x 8192 | 94.8 | 117.2 | +23.6% |
| 32 x 32000 | 409.5 | 495.7 | +21.0% |
| 32 x 128000 | 1623.8 | 1964.2 | +21.0% |
| 12 x 384 x 384 (BERT-shaped) | 22.7 | 23.7 | +4.4% |

Sharded configs are a separate case. `SoftmaxShardedMultiCoreProgramConfig` lets the caller pick `subblock_w` and the sharded kernel holds that many tiles in Dest at once, so halving the capacity would reject configs that run today: `models/demos/*/sentence_bert` ships `subblock_w = 6`. They can keep the 16-bit accumulator and stay bit-identical.

I have a patch and tests for all of this and can open a PR.

## Impact

The error is a per-row scale, and it survives the attention matmul: in `out = P @ V` a uniform scale on `P` does not cancel. Regressing each device attention-block output row onto its fp32 reference recovers the gain the block applied, and it tracks the softmax row sum to within 0.005 at every length (0.91619 against 0.91450 at seq 4096).

**But it barely moves PCC**, because a per-row scale is most of what PCC is blind to: the attention block reads 0.99938 against 0.99966. `models/demos/blackhole/sentence_bert` over its own STS-tr benchmark shows no meaningful change either way (embedding PCC 0.994497 against 0.994338), and that is expected: seq 384 is 12 tiles and its attention rows already sum to 0.996-1.000 on `main`. **I have no evidence that a shipped model produces different output because of this.** What is affected is anything reading a softmax output as a probability: log-probabilities, perplexity, comparison against an absolute threshold, and any consumer that sums the row. `argmax` and `top_k` are not, since a uniform scale does not reorder.

Not the same finding as #51911. That report said `fp32_dest_acc_en` does not help SDPA and at seq 16384 makes it worse, which is still true: SDPA's flash kernel accumulates per k-chunk with its own running max and rescale, a different structure from one Dest accumulator folding `Wt` tiles in a single pass.

## Environment

tt-metal `992f3ca634`, clean build, sfpi 7.67.0, `ttnn` Python API, Blackhole P300 device 0.

Not a recent regression: the `fp32_dest_acc_en = is_fp32` policy has been in `softmax_init_compute_kernel_config` since at least #27732 (2025-09).

## Not established

Wormhole was on this list and is not any more: measured on an n300, the `main` row sums above are identical to Blackhole to every digit, and the fix behaves the same there, including the HiFi3 that the hw bug #38306 guard selects once fp32 accumulation is on.

- **A model at a sequence length where this bites.** `models/demos/` has roughly 25 `ttnn.softmax` / `scale_mask_softmax` call sites, including Stable Diffusion self-attention at seq 4096. None were executed.
- **`numeric_stable=False`.** The same sweep gives a row sum of 2.2622 at w=16384, which is not this failure mode and is probably a distinct bug.

kinginu:
### PR Category

Bug fix. Closes #52045.

### Summary

`softmax_init_compute_kernel_config` derives `fp32_dest_acc_en` from the input dtype, so a 16-bit input gets a 16-bit Dest for the SUM reduce over the exponentials and the output stops summing to 1. Mechanism, blast radius and the measured Dest content are in #52045.

This stops deriving the flag from the dtype, matching what the generic reduce op already does for the same reduction and for the same reason (#13621 / #15665). Softmax has its own compute-config helper and was never in that PR's scope.

L1 goes down rather than up, because only the one-entry sum accumulators follow the flag. The `Wt`-deep bulk buffers stay 16-bit, which is what keeps the `use_large_kernel` switch where it is.

### Measured

Blackhole P300 and Wormhole n300. Both arms come from this build, in the same process on the same chip, with the baseline forcing `fp32_dest_acc_en=False`, so nothing differs between them but the flag.

Row sums, `bfloat16`. The two `main` columns agree to every digit, so the accumulator loses the same thing on both architectures.

| row length | BH `main` | BH fixed | WH `main` | WH fixed |
|---|---|---|---|---|
| 1024 | 0.983110 | 0.994318 | 0.983110 | 0.994236 |
| 4096 | 0.897854 | 0.994760 | 0.897854 | 0.994646 |
| 8192 | 0.864987 | 0.994859 | 0.864987 | 0.994786 |
| 16384 | 0.943260 | 0.996479 | 0.943260 | 0.996393 |
| 128000 | 1.012368 | 1.006620 | 1.012368 | 1.006524 |

`bfloat8_b` moves the same way, from [0.8697, 1.0202] to [0.9978, 1.0101].

Buffer bytes on one core, identical on both architectures. `main` and this branch select the same compute kernel at every width, so the `use_large_kernel` switch does not move.

| row length | kernel | `main` | this PR | |
|---|---|---|---|---|
| 1024 / 4096 / 8192 | `softmax` | 174,080 / 567,296 / 1,091,584 | 161,792 / 555,008 / 1,079,296 | -12,288 |
| 16384 and above | `softmax_large_tensor` | 540,672 | 532,480 | -8,192 |

Time, us/call, wall clock, 50 calls back to back with one synchronize, median of 5.

| shape | BH before | BH after | | WH before | WH after | |
|---|---|---|---|---|---|---|
| 1024 x 1024 | 22.6 | 22.4 | -0.9% | 47.7 | 47.2 | -1.0% |
| 2048 x 2048 | 61.0 | 65.4 | +7.2% | 127.6 | 131.7 | +3.2% |
| 4096 x 4096 | 188.3 | 198.9 | +5.6% | 408.8 | 422.8 | +3.4% |
| 32 x 8192 | 98.2 | 120.4 | +22.6% | 170.4 | 209.2 | +22.8% |
| 32 x 32000 | 414.6 | 506.5 | +22.2% | 717.9 | 871.8 | +21.4% |
| 32 x 128000 | 1642.5 | 2010.4 | +22.4% | 2823.0 | 3436.8 | +21.7% |
| 12 x 384 x 384 | 22.0 | 22.9 | +4.1% | 54.0 | 52.9 | -2.0% |

The cost is `fp32_dest_acc_en` halving Dest from 8 tiles to 4, which halves the block size in this factory so a row takes twice as many blocks. #15665 accepted the same cost for the same reduction. The worst case is where the row is one tile row deep and the doubled block count has nothing to amortise against.

### Tests

`test_softmax_rows_are_normalised`, 7 widths x 2 dtypes. Nothing in the tree asserted that a softmax row sums to 1, and the existing test at these exact shapes cannot: `test_large_softmax` passes at a row sum of 0.9556, because PCC is invariant under a uniform per-row scale and its `atol=0.04` is orders of magnitude above values of order 1/w.

Blackhole P300: `tests/ttnn/unit_tests/operations/fused/` 3395 passed / 0 failed, `test_softmax.py` 387 passed, `transformers/test_transformer.py` 37 passed, `models/demos/blackhole/sentence_bert/tests/pcc/` 9 passed. Wormhole n300: `test_softmax.py` 387 passed.

### Notes for reviewers

- **This is a change of default only.** Passing `fp32_dest_acc_en=False` explicitly reproduces `main` bit for bit at every width tested, element by element rather than within a tolerance, so any caller that already chooses a compute config is untouched.
- **Sharded program configs are held back.** The caller picks `subblock_w` and the sharded kernel holds that many entries in Dest, which fp32 mode halves, so those keep the 16-bit accumulator. `models/demos/*/sentence_bert` ships `subblock_w = 6`; on its own shape and program config the output is bit-identical to `main`, all 14,155,776 elements.
- **`MAX` and `PREV_MAX` stay with the bulk group** even though they are also one-entry reduce outputs. They are subtracted from the bulk buffers for numeric stability, and fp32 there against 16-bit operands widened the per-row spread at w <= 4096 (worst row 0.9916 to 0.9648) for no gain in the mean.
- **A float32 input keeps the wide bulk buffers it has always had.** Narrowing those would throw away input precision rather than accumulator precision, and float32 output is unchanged by this PR.
- **On Wormhole a 16-bit input now gets HiFi3 where it got HiFi4**, following the existing hw bug #38306 guard in the same function, which selects HiFi3 whenever fp32 accumulation is on. Measured on an n300 above; that is the only behaviour this PR gives Wormhole that Blackhole does not see.


bbradelTT (commented):
Thanks. The accuracy improvements look good.

Please
1. move the tests to tests/ttnn/nightly/unit_tests/...
2. add a rank-4 shape such as (1, 1, 32, w) (with at least one w that selects the large kernel similarly to what you have for rank-3).
3. update the unpack modes. Current code has
```
add_unpack(RECIP_SUM_EXPS, im_cb_data_format);  // still Float16_b for bf16 input
// ...
add_unpack(PREV_REDUCE, im_cb_data_format);
add_unpack(RECIP, im_cb_data_format);
```
and acc_cb_data_format should probably be used.
4. check if ttnn/cpp/ttnn/operations/normalization/softmax/docs/Softmax.md would need to be updated to reflect these changes.

PR #53644, opened 2026-08-19
ci-digest: resolve the run summary by attempt and scope

ppetrovicTT:
**The digest is broken on `main` right now.** tenstorrent/tt-github-actions#151 merged at 14:10Z today, so every `ai_summary/run` call site already publishes `ai_run_summary_r<run>_a<attempt>`, while `main`'s `ci_digest.py:80` still matches the exact old string `ai_run_summary_{run_id}`.

It fails silently: no match → `fetch_run_summary` returns `None` → the digest falls back to the run conclusion → a passing run renders as plain 🟢 green with its infra failures unreported. Two of the four watched workflows (`blackhole-e2e-tests`, `blaze-models-prefill-tests`) run `ai_summary/run`, so both are affected.

Next scheduled digest is **08:12Z tomorrow**; merging before then means no digest run is missed.

## What changes

Resolution matches `ai_run_summary[_<scope>]_r<run>_a<attempt>`, ranks by the attempt encoded in the name, and returns the winning name so the report `.json` inside the artifact is addressed exactly rather than by wildcard.

Attempt ranking matters because several reports can share a run: a partial re-run deletes nothing, a full re-run does not reliably purge (25 of 27 on run 32126839220), and a workflow invoked twice per run publishes one report per invocation. Without it, `created_at` could pick a stale attempt.

**No backward compatibility.** The pre-rename `ai_run_summary_<run>` is rejected, so this must land *after* #151 — which it now does.

```
^ai_run_summary_([^_]+_)?r<run>_a[0-9]+$
```

Anchored at both ends, so a longer run id (`r421` for run 42) and a suffixed copy (`_a3_backup`) are both rejected — the latter would otherwise rank as attempt 1 and could win selection, then fail the in-zip lookup and silently fall back. The scope segment is `[^_]+` rather than a character class: the producer slugifies every non-alphanumeric to `-`, so a slug never contains `_` but may contain any alphanumeric, Unicode included.

## Also hardened

The listing grew a third field and `rank()` split it three ways. Artifact names may contain tabs (GitHub rejects `" : < > | * ?` but not tab), so a malformed line raised `ValueError` — which is **not** in `check_workflow`'s caught set. It escaped the per-workflow guard whose contract is "one flaky gh call must not discard the other workflows' results" and aborted `main`, so no `.md` was written, `produced=true` never fired, and the team got silence instead of a digest. Now split from the right, with malformed lines skipped.

## Unrelated, bundled at request

The digest cron moves 08:12 → 07:12 UTC, in both `on.schedule` and the matrix leg gate that must match it.

25 → 33 self-tests (`ci_digest.py --self-test`).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ppetrovic/ci-digest-attempt-resolution)


blozano-tt (approved):

blozano-tt:
In `_summary_name_pattern` (`.github/scripts/ci_digest.py:78`), the pattern is:

```
^ai_run_summary_([^_]+_)?r{run_id}_a[0-9]+$
```

Given that `test_rejects_near_misses` treats `ai_run_summary_42` as a name that should be **rejected** (calling it "the pre-rename name") — under this regex, would a legacy-named artifact like `ai_run_summary_42` (produced by workflows that haven't picked up the tt-github-actions#151 rename yet) ever actually match and get selected by `fetch_run_summary`? And if it can't match, what does the digest fall back to for those runs in the window before #151 lands everywhere?

**What's unclear to me:** the PR title and description say this change "tolerates the legacy name," and `_artifact_attempt()` defaults unsuffixed names to attempt 1 (which reads like it's meant to make them selectable) — but I don't see where a legacy-named artifact is actually let through the jq/regex filter above. Could you confirm how the legacy-name path is intended to work here, since as written it looks like it's excluded rather than tolerated?

PR #53660, opened 2026-08-19
Feature | Separate test outputs per run

nstojicTT:
## Problem

Every perf run writes to the same place, `perf_data/<test>/<test>.csv`, so a
second run overwrites the first. Worse: if the second run is **narrower** — a
`-k` filter, `--perf-run-types`, or a test that failed this time — it leaves the
first run's directories untouched. The tree then reads as one complete run while
holding a blend of two.

    perf_data/
      perf_a/perf_a.csv     <- run 2
      perf_b/perf_b.csv     <- still run 1, nobody touched it
      local.parquet         <- run 2

## Change

Each run writes its own directory, named by a tag, with a stable `latest` pointer:

    perf_data/
      runs/
        local-20260819T140312Z/
          perf_a/perf_a.csv
          perf_b/perf_b.csv
          <run_id>.parquet
      latest -> runs/local-20260819T140312Z

- **Tag** — `PERF_RUN_TAG` if set, else `local-<UTC timestamp>`. CI sets it
  explicitly, because only the workflow can see the shard index.
- **`latest`** — a symlink so callers keep one stable path. Best-effort: a report
  that exists but is unlinked is still a usable report.
- **Pruning** — `PERF_KEEP_RUNS` (default 10) bounds local history. The archive
  is the published Parquet, not this directory. `0` disables pruning.
- **`run_id` fix** — off CI this was the constant `"local"`, so two local runs of
  one commit published colliding ROW_KEYs `(test_name, commit_sha, arch, run_id)`.
  It now falls back to the run tag. In CI `GITHUB_RUN_ID` still wins, so all
  shards of a workflow share one `run_id`, as the data team expects.

## Why the workflow changed

Not because of re-runs — CI has none. Because the paths moved:

1. `publish_run` globs `**/*.csv`, and Python's `**` follows symlinks. Pointed at
   `perf_data` it would find every CSV twice, via `runs/` and via `latest`, and
   publish duplicate rows. It now points at `perf_data/latest`.
2. `zip -r perf_data/` would follow `latest` and store a second copy of every
   report. It now zips `runs/` and the Parquet explicitly.
3. `PERF_RUN_TAG` carries `strategy.job-index`. Without it, every shard of one
   arch gets the same tag, and shard directories collide when artefacts are
   collected. The shard index is a matrix value, so only the workflow can pass it.

## Where to look

- `.github/workflows/llk-perf-impl.yaml`
- `tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py`
- `tt_metal/tt-llk/tests/python_tests/helpers/test_config.py`

The rest is docs, skills, and tests.

## Tests

Hardware-free, in `test_perf_report_hw_free.py`:

- a narrower second run does not inherit the first run's test directories
- `latest` follows the most recent run
- history is pruned to `PERF_KEEP_RUNS`, and the newest always survives

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-data-per-run-dirs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-data-per-run-dirs)


.github/workflows/llk-perf-impl.yaml
```diff
@@ -123,6 +123,12 @@ jobs:
       - name: Run tests
         shell: bash
         working-directory: ${{ env.LLK_PATH }}
+        env:
```
ndivnicTT: Are you sure any changes in this file are required? Won't it work just fine without any changes? We don't have reruns on GH Workflow

tt_metal/tt-llk/tests/python_tests/helpers/test_config.py
```diff
@@ -420,6 +420,39 @@ def setup_paths(sources_path: Path):
...
+        notion of "one run" spans all shards) — naming directories after it would
+        make two shards collide the moment their artefacts are unzipped together.
+
+        Seeded into the environment on first use so xdist workers and the
+        controller agree; conftest sets it before workers spawn.
+        """
+        tag = os.environ.get("PERF_RUN_TAG", "").strip()
+        if not tag:
+            run = os.environ.get("GITHUB_RUN_ID", "").strip()
+            arch = os.environ.get("CHIP_ARCH", "unknown").strip()
```
ndivnicTT: Why? AFAIK we can't run multiple ARCH on the same machine

tt_metal/tt-llk/tests/python_tests/test_perf_report_hw_free.py
```diff
@@ -308,3 +314,103 @@ def test_combine_perf_reports_emits_parquet_alongside_csv(tmp_path, monkeypatch)
...
+    assert (runs / "run-2" / "perf_a" / "perf_a.csv").exists()
+    assert not (runs / "run-2" / "perf_b").exists()
+
+
+def test_latest_follows_the_most_recent_run(tmp_path, monkeypatch):
+    root = tmp_path / "root"
+    monkeypatch.setattr(TestConfig, "LLK_ROOT", root)
+
+    _perf_run(tmp_path, monkeypatch, "run-1", ["perf_a"], mean=10.0)
+    _perf_run(tmp_path, monkeypatch, "run-2", ["perf_a"], mean=99.0)
```
ndivnicTT, line 540: Basically, you need to increment manually, right?
Other option is to not put any tag and end up with timestamp?

ndivnicTT (commented):
Please update PR description, it is confusing and I think it does not contain all the information.

tt_metal/tt-llk/tests/python_tests/test_perf_report_hw_free.py
nstojicTT: no manual increment
with no PERF_RUN_TAG set you would get local-<UTC timestamp> automatically and your second option is the default
the tests pin run-1/run-2 only so the assertions have a stable path, a directory named after the current second cannot be asserted on

.github/workflows/llk-perf-impl.yaml
nstojicTT: its not because of reruns but because of moved paths, I put this in PR description now as well

publish_run globs **/*.csv, and Python's ** follows symlinks. Pointed at
perf_data it would find every CSV twice, via runs/ and via latest, and
publish duplicate rows. It now points at perf_data/latest.

zip -r perf_data/ would follow latest and store a second copy of every
report. It now zips runs/ and the Parquet explicitly.

PERF_RUN_TAG carries strategy.job-index. Without it, every shard of one
arch gets the same tag, and shard directories collide when artefacts are
collected. The shard index is a matrix value, so only the workflow can pass it.

tt_metal/tt-llk/tests/python_tests/helpers/test_config.py
nstojicTT: agree with this, will drop ARCH and just use a timestamp

ndivnicTT (approved):
LGTM!

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -444,13 +445,67 @@ def _ci_provenance() -> dict:
     return {
         "commit_sha": os.environ.get("GITHUB_SHA", "unknown"),
         "arch": os.environ.get("CHIP_ARCH", "unknown"),
-        "run_id": os.environ.get("GITHUB_RUN_ID", "local"),
+        # Off CI this used to be the constant "local", so two local runs of the
+        # same commit published colliding ROW_KEYs (test_name, commit_sha, arch,
+        # run_id). The run tag is already unique per invocation; reuse it. In CI
+        # GITHUB_RUN_ID still wins, so all shards of a workflow share one run_id.
+        "run_id": os.environ.get("GITHUB_RUN_ID") or TestConfig.perf_run_tag(),
```
fvranicTT: If you hit "Rerun all/failed jobs" the same ‎`GITHUB_RUN_ID` is preserved. Does that affect your algorithm?

tt_metal/tt-llk/tests/python_tests/conftest.py
```diff
@@ -331,6 +331,12 @@ def pytest_configure(config):
     if config.getoption("--disable-sfploadmacro", default=False):
         os.environ["TT_METAL_DISABLE_SFPLOADMACRO"] = "1"
 
+    # Fix this run's perf report directory here, on the controller, before xdist
+    # spawns workers: they inherit the environment, so every process resolves the
+    # same perf_data/runs/<tag>. Left to first use, each worker would mint its own
+    # timestamp and scatter one run across several directories.
+    TestConfig.perf_run_tag()
```
fvranicTT: This will be executed unconditionally even for non-perf tests. Are we OK with that?

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -444,13 +445,67 @@ def _ci_provenance() -> dict:
...
         "timestamp": datetime.now(timezone.utc).isoformat(),
         "pipeline": "PR" if event == "pull_request" else "nightly",
         "pr_number": os.environ.get("PR_NUMBER") or None,
     }
 
 
+def _refresh_latest(run_dir: Path) -> None:
```
fvranicTT, line 477:
```python
def _refresh_latest(run_dir: Path) -> None:
    """Point ``perf_data/latest`` at this run, so callers keep a stable path.
    Best-effort: a report that exists but is not linked is still a usable report.
    """
    link = run_dir.parent.parent / "latest"
    if link.exists() and not link.is_symlink():
        logger.warning(f"perf_data/latest exists and is not a symlink: {link}")
        return
    tmp = link.with_name(f".latest.tmp.{os.getpid()}")
    try:
        tmp.symlink_to(run_dir.relative_to(link.parent), target_is_directory=True)
        tmp.replace(link)
    except OSError as exc:
        logger.warning(f"perf_data/latest not updated: {exc}")
        tmp.unlink(missing_ok=True)
```
`Path.replace` swaps the symlink in one rename, so latest is never absent. Also, if an exception occurs we unlink or try to.

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -444,13 +445,67 @@ def _ci_provenance() -> dict:
...
+def _keep_runs() -> int:
+    """How many run directories to retain locally (``PERF_KEEP_RUNS``, default 10).
+
+    The archive is the published Parquet, not this directory, so local history is
+    a debugging convenience and wants a bound. 0 or less disables pruning.
+    """
+    try:
+        return int(os.environ.get("PERF_KEEP_RUNS", "10"))
+    except ValueError:
+        return 10
```
fvranicTT, line 511: Can we log this as a warning?

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -444,13 +445,67 @@ def _ci_provenance() -> dict:
...
+
+
+def _prune_runs(runs_dir: Path, keep: int) -> None:
```
fvranicTT:
```python
def _prune_runs(runs_dir: Path, keep: int, current: Path) -> None:
    """Keep the newest ``keep`` run directories, never deleting ``current``.
    ``keep <= 0`` disables pruning.
    """
    if keep <= 0:
        return
    run_dirs: list[tuple[float, Path]] = []
    try:
        for d in runs_dir.iterdir():
            if not d.is_dir() or d.is_symlink():
                continue
            try:
                run_dirs.append((d.stat().st_mtime, d))
            except OSError as exc:
                logger.warning(f"skipping run directory {d}: {exc}")
    except OSError as exc:
        logger.warning(f"perf_data runs not pruned ({runs_dir}): {exc}")
        return
    run_dirs.sort(reverse=True)
    survivors = {p for _, p in run_dirs[:keep]}
    survivors.add(current)
    for _, stale in run_dirs:
        if stale in survivors:
            continue
        try:
            shutil.rmtree(stale)
        except OSError as exc:
            logger.warning(f"failed to prune run directory {stale}: {exc}")
```

One failed `stat()` aborts the whole prune because d.stat() sits in the sort key and the except OSError wraps the entire listing.

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
nstojicTT: Yes this is correct, it would keep run id and increment run attempt
Re run would publish a different measurement under the same row key, this was not intended 
Now each row should keep identity that was published with and the shards of one attempt share one run id

tt_metal/tt-llk/tests/python_tests/conftest.py
nstojicTT: perf_run_tag doesn't do any file work, it just reads an env variable, makes the timestamp and writes it back.
We make the dir later when the perf report is written (test_config.py line 451.)
I think in this case unconditional call is safe

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
nstojicTT: Applied

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
nstojicTT: Makes sense, applied

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
nstojicTT: Makes sense, applied

fvranicTT (approved):

njokovicTT (approved):

dpopovTT (approved):

PR #9281, opened 2024-06-07
Fix log message typo (importting -> importing)

yieldthought:
Couldn't see this every day without fixing it ^^

PR #53670, opened 2026-08-19
[tt-train] Add adamw_tensor_scalars: AdamW with step scalars as device tensors

agobeljicTT:
### What's changed

`ttml::metal::adamw_tensor_scalars`: AdamW variant taking the step-varying scalars as single-element f32 device tensors instead of floats:

```
step_size    = lr / (1 - beta1^t)
inv_sqrt_bc2 = 1 / sqrt(1 - beta2^t)
decay_factor = 1 - lr * weight_decay
```

These are exactly the 3 immediates the kernel already consumes, so no new math on device. Reader pulls them from DRAM into a CB per launch, compute reads them via `read_tile_value` and overwrites the runtime-arg values. Same device op, factory and kernels as `adamw`; float path untouched (`SCALARS_FROM_TENSOR=0`).

### Why

The float-scalar `adamw` forces every caller to know `beta^t` on the host, and `beta^t` changes every training step. For tt-mlir (tenstorrent/tt-mlir#9203) that meant a per-step choice between two bad options:

- bake the scalars into the graph as attributes → the graph text changes every step → **recompile per step**, and the op can't live inside a trace;
- pass them as tensors and read them back for the float API → a **blocking device→host sync per scalar per step**, which also breaks trace capture (a replay would freeze the captured value).

With the scalars as device tensors, the caller updates them on device each step and the op is step-invariant: one compiled graph, no host involvement, trace-safe.

### Checklist

- [x] New/Existing tests provide coverage for changes: verified on BH silicon that output is bit-identical to float-scalar `adamw`; existing adamw tests pass.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:agobeljicTT/adamw_tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:agobeljicTT/adamw_tensor)


zbaczewskiTT:
> Why: the caller can update the scalars on device each step, so the op is step-invariant — no host involvement, trace-safe.

The reason you've given doesn't really explain why the change is needed. 
Why would we even want trace in training? Nothing in tt-train captures a trace today, and there is no number showing it's worth it.
"No host involvement" - we still need to change the values inside the tensors, and we do that on host. Unless we also implement LR scheduling and beta corrections (per-optimizer) on device.
There is not enough data to back up that these changes are necessary.

zbaczewskiTT (approved):
The reasoning is now valid, thank you.
Looks good to me.

mdragulaTT:
@agobeljicTT please do not add this to mergeQ. I'm finishing my review

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -339,8 +356,19 @@ AdamWProgramFactory::cached_program_t AdamWProgramFactory::create(
...
     auto* output_buffer = output.buffer();
 
+    const bool scalars_from_tensor = tensor_args.step_size.has_value();
+    auto* step_size_buffer = scalars_from_tensor ? tensor_args.step_size->buffer() : nullptr;
+    auto* inv_sqrt_bc2_buffer = scalars_from_tensor ? tensor_args.inv_sqrt_bc2->buffer() : nullptr;
+    auto* decay_factor_buffer = scalars_from_tensor ? tensor_args.decay_factor->buffer() : nullptr;
+
+    if (scalars_from_tensor) {
+        [[maybe_unused]] auto cb_scalars = create_circular_buffer(
+            program, all_cores, kScalarsCbIndex, tt::DataFormat::Float32, float32_single_tile_size_bytes, 3U);
```
mdragulaTT: `float32_single_tile_size_bytes` is `tile_size(Float32)` = 4096, so this is 12 KB of L1 per core and three DRAM page reads per core per dispatch to carry 12 bytes — ~768 KB per dispatch on a 64-core grid, all against the same three pages. Amortised over a model step that's noise, but AdamW dispatches per parameter tensor: for `{1,1,1,2048}` (already in `adamw_benchmark.cpp:31`) it's 8 KB of real input reads per core against 12 KB of scalar reads. Two ways out. One 3-element control tensor read from a single page, which is the convention at `ops/variable_matmul/device/kernels/compute/compute.cpp:230-234` and costs tt-mlir#9203 a concat; or keep three tensors and replace `noc_async_read_page` with a sized `noc_async_read` off the accessor (`ops/moe_group/device/kernels/dataflow/moe_group_combined_kernel.cpp:222`), which needs no tt-mlir change and still drops L1 to ~32 B. `create_circular_buffer_bytes` in `metal/common/program_utils.hpp` is the CB helper for either.

tt-train/sources/ttml/metal/optimizers/adamw/device/kernels/dataflow/reader_adamw.cpp
```diff
@@ -43,6 +49,23 @@ void kernel_main() {
...
+    cb_reserve_back(cb_scalars_idx, 3U);
+    const uint32_t scalars_l1_addr = get_write_ptr(cb_scalars_idx);
+    noc_async_read_page(0, TensorAccessor(step_size_args, step_size_addr), scalars_l1_addr);
+    noc_async_read_page(
+        0, TensorAccessor(inv_sqrt_bc2_args, inv_sqrt_bc2_addr), scalars_l1_addr + scalar_tile_size_bytes);
+    noc_async_read_page(
+        0, TensorAccessor(decay_factor_args, decay_factor_addr), scalars_l1_addr + 2U * scalar_tile_size_bytes);
+    noc_async_read_barrier();
+    cb_push_back(cb_scalars_idx, 3U);
+#endif
```
mdragulaTT, line 77: `metal/common/dataflow_utils.hpp`, already included at line 8, has `read_one_tile` at `:322` — reserve / read page / barrier / push — which is exactly one of these three reads; `read_tiles_by_row` at `:373` is the same shape hardcoded to 3. Separately, `noc_async_read_barrier()` at line 65 fires before the tile loop, so the reader stalls a full DRAM round trip before issuing the first param/grad read, on every core, every dispatch.

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
```diff
@@ -26,4 +26,26 @@ ttnn::Tensor adamw(
...
+    const ttnn::Tensor& grad,
+    const ttnn::Tensor& exp_avg,
+    const ttnn::Tensor& exp_avg_sq,
+    const std::optional<ttnn::Tensor>& max_exp_avg_sq,
+    const ttnn::Tensor& step_size,
+    const ttnn::Tensor& inv_sqrt_bc2,
+    const ttnn::Tensor& decay_factor,
+    float beta1,
+    float beta2,
+    float epsilon);
```
mdragulaTT: The comment above calls this an overload, and it can be one: at position 6 the float path takes `float lr` and this takes `const ttnn::Tensor& step_size`, no implicit conversion either way, and the arities differ (12-14 vs 11). Same at the prim layer (`adamw_device_operation.hpp:36` vs `:53`), so renaming both to `adamw` resolves unambiguously and drops two names from the public surface.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -124,11 +131,18 @@ void assign_per_core_runtime_args(
...
+    float step_size = 0.0f;
+    float inv_sqrt_bc2 = 0.0f;
+    float decay_factor = 0.0f;
+    if (step_size_buffer == nullptr) {
+        float bias_correction1 = 1.0f - attrs.beta1_pow;
+        float bias_correction2 = 1.0f - attrs.beta2_pow;
+        step_size = attrs.lr / bias_correction1;
+        inv_sqrt_bc2 = 1.0f / std::sqrt(bias_correction2);
+        decay_factor = 1.0f - attrs.lr * attrs.weight_decay;
+    }
```
mdragulaTT: This block is duplicated at `:501-516`, and the two copies test the same condition through different objects — `step_size_buffer == nullptr` here versus `!tensor_args.step_size.has_value()` there. If those ever diverge, the create path and the cache-hit path disagree about which scalars to use, and since `override_runtime_arguments` is the one that runs every step, a first-launch test would still pass while steps 2+ silently get the wrong values. Worth extracting into a `compute_step_scalars(attrs, scalars_from_tensor)` helper — the four-line explanatory comment and `one_minus_beta1` / `one_minus_beta2` are duplicated across the same two functions too.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation_types.hpp
```diff
@@ -33,6 +33,14 @@ struct tensor_args_t {
...
     const ttnn::Tensor& exp_avg_sq;
     std::optional<ttnn::Tensor> max_exp_avg_sq = std::nullopt;
+
+    // Single-element f32 tensors carrying the step-varying scalars (see
+    // ttml::metal::adamw_tensor_scalars). Engaged together; when set, the
+    // lr / beta*_pow / weight_decay attributes are ignored and stochastic
+    // rounding must be disabled.
+    std::optional<ttnn::Tensor> step_size = std::nullopt;
+    std::optional<ttnn::Tensor> inv_sqrt_bc2 = std::nullopt;
+    std::optional<ttnn::Tensor> decay_factor = std::nullopt;
```
mdragulaTT: Three separate optionals make seven of the eight states invalid, which is why `adamw_device_operation.cpp:91-94` needs a `TT_FATAL` to police "must be provided together". One optional holding a `struct { Tensor step_size, inv_sqrt_bc2, decay_factor; }` would make the bad states unrepresentable and that check unnecessary. Same for the other two invariants stated in this comment — that `lr` / `beta*_pow` / `weight_decay` are ignored, and that stochastic rounding must be disabled — they're prose because the types don't express them.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -464,11 +501,18 @@ void AdamWProgramFactory::override_runtime_arguments(
...
     float one_minus_beta2 = 1.0f - attrs.beta2;
 
-    float bias_correction1 = 1.0f - attrs.beta1_pow;
-    float bias_correction2 = 1.0f - attrs.beta2_pow;
-    float step_size = attrs.lr / bias_correction1;
-    float inv_sqrt_bc2 = 1.0f / std::sqrt(bias_correction2);
-    float decay_factor = 1.0f - attrs.lr * attrs.weight_decay;
+    // In tensor-scalar mode the kernel overwrites these three from the scalars
+    // CB, and the lr / beta*_pow / weight_decay attrs are left defaulted.
+    float step_size = 0.0f;
```
mdragulaTT: In tensor mode these are placeholders the kernel overwrites, but `adamw_kernel.cpp:177` guards on `if (decay_factor != fp32_one)`. So if the overwrite ever doesn't happen — define not propagated, CB not pushed, a later refactor — the branch is taken with `decay_factor == 0.0f` and every parameter is multiplied by zero. Defaulting `decay_factor` to `fp32_one` and leaving `step_size` at `0.0f` makes that failure a no-op update instead.

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
```diff
@@ -26,4 +26,26 @@ ttnn::Tensor adamw(
...
 
+// AdamW with the step-varying scalars as single-element f32 device tensors.
+//   step_size    = lr / (1 - beta1^t)
+//   inv_sqrt_bc2 = 1 / sqrt(1 - beta2^t)
+//   decay_factor = 1 - lr * weight_decay
+//
+// Stochastic rounding is deliberately not offered on this overload. It needs a
+// fresh host-drawn seed every step, delivered as a compute runtime argument --
+// exactly the per-step host work this overload exists to remove. Use the
+// float-scalar `adamw` above when stochastic rounding is required.
```
mdragulaTT: Worth documenting the other half of the trace contract here: trace freezes the scalar tensors' addresses, so callers have to allocate them outside the capture and keep them alive for its lifetime, updating contents in place. A reallocation between replays would silently read a stale address.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
```diff
@@ -87,6 +87,35 @@ void AdamWDeviceOperation::validate_on_program_cache_miss(
...
+                "Tensor '{}' must hold exactly one element, got {}",
+                name,
+                tensor->logical_volume());
+            // The launch infrastructure targets param's device; a scalar tensor on another
+            // device would pass its (device-local) buffer address to the reader unnoticed.
+            TT_FATAL(
+                tensor->device() == param.device(),
+                "Tensor '{}' must be on the same device as the parameter tensor",
+                name);
+        }
```
mdragulaTT, line 136: This sits in `validate_on_program_cache_miss` and `device` isn't in the program hash, so it can't catch what it describes — launch once with everything on device A, then again with `param` on A and a scalar on B, and it's a cache hit with no validation. Not blocking, since no tt-train op implements `validate_on_program_cache_hit`, but this is a silent-corruption guard rather than a shape guard, and there's no mesh / multi-device parity test either.

tt-train/tests/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ set(SOURCES
     optimizers/adamw_composite_test.cpp
     optimizers/sgd_test.cpp
     optimizers/adamw_test.cpp
+    optimizers/adamw_tensor_scalars_test.cpp
```
mdragulaTT: Both harnesses already exist. Parameterizing the cases in `tests/optimizers/adamw_test.cpp` over a `ScalarSource{Float, Tensor}` would give every existing dtype × amsgrad × shape combination tensor-scalar coverage, instead of four hand-written cases in a new file. `tests/benchmark/adamw_benchmark.cpp` needs scalar source as a second argument next to `->DenseRange(0, shapes-1, 1)` — `adamw_program_factory.cpp:272-274` cites it as the authority for the pipeline-depth tuning, and there are no perf numbers in this PR.

tt-train/tests/optimizers/adamw_tensor_scalars_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+    // The tensor-scalar path feeds the kernel the same three immediates, so results are bit-identical.
+    EXPECT_EQ(param_ref, param_out);
+    EXPECT_EQ(exp_avg_ref, exp_avg_out);
+    EXPECT_EQ(exp_avg_sq_ref, exp_avg_sq_out);
+    EXPECT_EQ(max_exp_avg_sq_ref, max_exp_avg_sq_out);
+}
+
+void run_parity_check(ttnn::DataType param_dtype, bool amsgrad) {
+    // The first step compiles and caches the program; the second hits the program
+    // cache with freshly allocated scalar tensors holding different values, so it
```
mdragulaTT: This doesn't exercise the address update. Step 10's tensors are destroyed when the lambdas return and step 11 allocates after, so the allocator hands back the same addresses — a broken `override_runtime_arguments` would read the new value from the old address and pass. Keeping the step-10 tensors alive across the step-11 allocation would actually test it.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -33,6 +33,9 @@ constexpr uint32_t kGradAddrIdx = 1U;
 constexpr uint32_t kExpAvgAddrIdx = 2U;
 constexpr uint32_t kExpAvgSqAddrIdx = 3U;
 constexpr uint32_t kMaxExpAvgSqAddrIdx = 4U;
+constexpr uint32_t kStepSizeAddrIdx = 7U;
```
mdragulaTT: **Smaller**

- `adamw_program_factory.cpp:36-38` — `kStepSizeAddrIdx = 7U` hardcodes a gap; indices 5/6
  (`num_tiles`, `start_tile`) are unnamed, so the 7 depends on append order in
  `assign_per_core_runtime_args`. Worth naming them. Collapses to one constant with a single
  control tensor.
- `adamw_kernel.cpp:61-64` — `ckernel::read_tile_value` is the only qualified call site; the
  four others are unqualified (`frobenius_normalize_compute.cpp:147`,
  `moe_group_tilize_kernel.cpp:32`, `moe_ungroup_untilize_kernel.cpp:42`,
  `variable_matmul/compute.cpp:230-234`), and lines 61/65 right here call `cb_wait_front` /
  `cb_pop_front` unqualified from that same namespace. Also `0, 0` → `0U, 0U`: this file is
  U-suffixed throughout (`3U` two lines up, `0x3F800000U` at line 46).
- `adamw_kernel.cpp:61` — `read_tile_value` runs before any HW startup. The two kernels that
  read control values at kernel start call `compute_kernel_hw_startup` first
  (`moe_group_tilize_kernel.cpp:35`, `moe_ungroup_untilize_kernel.cpp:41`);
  `frobenius_normalize_compute.cpp:147` uses this order but mid-kernel, after config. Probably
  fine, but it's the only site with no prior setup and it's cheap to reorder.
- `adamw.cpp:47-71` — the public wrapper is a pure pass-through whose only added logic is
  `max_exp_avg_sq.has_value()`, which the float wrapper computes too. Not worth a change on its
  own, but another argument for folding the two entry points into one overload.
- No mesh / multi-device parity coverage, which is what the scalar-device check above is about.

---
---

mdragulaTT (changes requested):
Mechanism is sound and the float path is properly insulated: no new device math, `scalars_from_tensor` in the program hash, `SCALARS_FROM_TENSOR=0` leaving the old path byte-identical. Rejecting stochastic rounding with a `TT_FATAL` rather than tensorizing the seed is the right call.

Code comments are inline. What I want to raise here is the direction, since it decides how much of that detail matters.

### Why is training being brought up on ttml *ops* through tt-mlir, rather than in the ttml framework?

@ayerofieiev-tt @philei-tt — flagging you as you have broader context here, because I might have missed something.

https://github.com/tenstorrent/tt-mlir/issues/9097 reimplements ~15 ttml ops in tt-mlir. https://github.com/tenstorrent/tt-mlir/issues/9237 is the AdamW instance and this PR is its dependency. So `ttml/metal/**` becomes an op library consumed by a compiler in another repo.

ttml is a training framework — autograd, modules, optimizers, serialization, and the kernels underneath. Consuming only the kernels from a compiled StableHLO graph means reimplementing the framework's job on the compiler side, and it turns `ttml/metal/**` into a cross-repo API contract.

That contract is the part I'd push back on. tt-train is meant to be a standalone framework; `ttml/metal/**` is an internal layer, not a surface we've committed to keeping stable. Holding it stable across ~15 ops, for a consumer in another repo whose integration tt-metal CI can't see, is a real ongoing cost for the tt-train owners. My understanding was that the direction is to use ttml as a whole framework rather than as an op library — so: what makes the op-by-op route preferable to bringing the model up in ttml?

The performance comes from the framework as much as the kernels: how the step is scheduled, where memory is reused, what gets fused. That's why I'm questioning the direction rather than just the API.

My recollection is that the ask to Forge was to bring a model up *on the ttml framework*, not to consume ttml ops from a compiled graph. I may have that wrong, which is why I'd rather Artem confirm it than assume.

None of this is a brush-off, and none of it is about this PR's code. If you want to bring a model up on ttml and need ops, features or fixes to get there, bring them to us — we'll discuss and review. That's a direction I think we can actively support.

The inline comments stand either way.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -33,6 +33,9 @@ constexpr uint32_t kGradAddrIdx = 1U;
...
+constexpr uint32_t kInvSqrtBc2AddrIdx = 8U;
+constexpr uint32_t kDecayFactorAddrIdx = 9U;
```
mdragulaTT, line 40: `kStepSizeAddrIdx = 7U` hardcodes a gap — indices 5 and 6 (`num_tiles`, `start_tile`) are unnamed, so the 7 silently depends on the append order in `assign_per_core_runtime_args`. Worth naming those two as well.

tt-train/sources/ttml/metal/optimizers/adamw/device/kernels/compute/adamw_kernel.cpp
```diff
@@ -55,6 +57,14 @@ void kernel_main() {
     uint32_t decay_factor = get_arg_val<uint32_t>(args_idx++);
     [[maybe_unused]] uint32_t seed = get_arg_val<uint32_t>(args_idx++);
 
+#if SCALARS_FROM_TENSOR
+    cb_wait_front(cb_scalars_idx, 3U);
+    step_size = ckernel::read_tile_value(cb_scalars_idx, 0, 0);
+    inv_sqrt_bc2 = ckernel::read_tile_value(cb_scalars_idx, 1, 0);
+    decay_factor = ckernel::read_tile_value(cb_scalars_idx, 2, 0);
```
mdragulaTT: `ckernel::` is the only qualified `read_tile_value` call site in tt-train — `frobenius_normalize_compute.cpp:147`, `moe_group_tilize_kernel.cpp:32` and `moe_ungroup_untilize_kernel.cpp:42` are all unqualified, and lines 61/65 here call `cb_wait_front` / `cb_pop_front` unqualified from that same namespace. `0, 0` should also be `0U, 0U` to match the rest of the file (`3U` two lines up, `0x3F800000U` at line 46). Separately, this is the only site calling `read_tile_value` before any HW startup — `moe_group_tilize_kernel.cpp:35` and `moe_ungroup_untilize_kernel.cpp:41` both call `compute_kernel_hw_startup` first.

ayerofieiev-tt:
> Why is training being brought up on ttml ops through tt-mlir, rather than in the ttml framework?

@mdragulaTT it does not matter. ops are there to be used. they are built to be used in ttml context, but can be used in any way a caller wants to use them

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
```diff
@@ -26,4 +26,26 @@ ttnn::Tensor adamw(
...
+ttnn::Tensor adamw_tensor_scalars(
```
ayerofieiev-tt: c++ supports overloads, so this could be `adamw`

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
ayerofieiev-tt: I disagree, trace is a generic thing and is a caller problem.

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
agobeljicTT: I know I deliberately wanted a separate function, but yeah why not

tt-train/tests/optimizers/adamw_tensor_scalars_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+        return std::make_tuple(param_out.to_vector<float>(), exp_avg.to_vector<float>(), exp_avg_sq.to_vector<float>());
+    };
+
+    const auto [param_ref, exp_avg_ref, exp_avg_sq_ref] = run_float_path();
+    const auto [param_out, exp_avg_out, exp_avg_sq_out] = run_tensor_path();
+
+    // The tensor-scalar path feeds the kernel the same three immediates, so results are bit-identical.
+    EXPECT_EQ(param_ref, param_out);
+    EXPECT_EQ(exp_avg_ref, exp_avg_out);
+    EXPECT_EQ(exp_avg_sq_ref, exp_avg_sq_out);
```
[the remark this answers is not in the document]
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/kernels/dataflow/reader_adamw.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/adamw.hpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation_types.hpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
agobeljicTT: done

tt-train/tests/optimizers/adamw_tensor_scalars_test.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/kernels/compute/adamw_kernel.cpp
agobeljicTT: done

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
agobeljicTT: implementing validate_on_program_cache_hit would be a precedent, no tt-train op does it

tt-train/tests/CMakeLists.txt
agobeljicTT: done. parameterized the cases in `tests/optimizers/adamw_test.cpp` over a `ScalarSource{Float, Tensor}` and deleted `adamw_tensor_scalars_test.cpp`. every case in the basic, weight decay and amsgrad suites now runs both ways, 48 tests, all passing. the float variant still goes through the optimizer, the tensor variant calls the op directly since `optimizers::AdamW` only ever calls the float overload, and it also checks the two paths are bit identical on param, exp_avg, exp_avg_sq and max_exp_avg_sq.

one thing i could not fold in: that harness is bf16 only, it has no dtype dimension. so fp32 is still covered by 2 hand written tests, plus 2 for scalar tensor validation and one for the program cache / override_runtime_arguments path. tell me if you want a dtype dimension added to the suite instead.

the benchmark already takes scalar source as a second arg on this branch, `ArgsProduct({CreateDenseRange(0, shapes-1, 1), {0, 1}})`, with labels `/float_scalars` and `/tensor_scalars`.

numbers below. blackhole, single chip, bf16 no amsgrad, 5 warmup + 50 measured iterations per config, bandwidth counts 7 tensor traversals (4 read, 3 written).

| shape | tensor mb | float us | tensor us | float gb/s | tensor gb/s | delta |
|---|---|---|---|---|---|---|
| 2048x2048 | 8.39 | 183.4 | 180.1 | 320.1 | 326.0 | -1.8% |
| 5632x2048 | 23.07 | 461.2 | 465.1 | 350.1 | 347.2 | +0.8% |
| 2048x5632 | 23.07 | 468.1 | 466.1 | 345.0 | 346.5 | -0.4% |
| 512x2048 | 2.10 | 72.6 | 72.8 | 202.3 | 201.5 | +0.4% |
| 1x2048 | 0.004 | 22.4 | 24.4 | 1.3 | 1.2 | +8.7% |
| 8192x8192 | 134.22 | 2344.3 | 2347.7 | 400.8 | 400.2 | +0.1% |
| 16384x2048 | 67.11 | 1201.2 | 1204.8 | 391.1 | 389.9 | +0.3% |

on every shape that does real work the two paths are the same within run to run noise, -1.8% to +0.8%, and both peak at about 400 gb/s on 8192x8192. so the tensor scalars do not change the bandwidth profile that `pipeline_depth = block_size` was tuned against.

the only real difference is 1x2048, +1.7 us, reproduced on a second run, 22.4 -> 24.1 us. that is fixed cost, three extra single element dram reads on a problem that is one tile of work, it only looks big as a percentage because the shape is 4 kb.

caveat on the numbers: single run of 50 iterations per config, no repetitions, and there is no main baseline here since the float column is the unchanged path.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation_types.hpp
```diff
@@ -26,13 +26,26 @@ struct operation_attributes_t {
     std::optional<uint32_t> stochastic_rounding_seed{std::nullopt};
 };
 
+// Single-element f32 tensors carrying the step-varying scalars (see the
+// tensor-scalar ttml::metal::adamw overload). Grouped in one struct so the
+// three can only be engaged together. When set, the lr / beta*_pow /
+// weight_decay attributes are ignored and stochastic rounding must be
+// disabled.
```
philei-tt: (nit) This comment is bloated and quite useless imo

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
```diff
@@ -87,6 +87,32 @@ void AdamWDeviceOperation::validate_on_program_cache_miss(
         check_tensor(
             max_exp_avg_sq.value(), "Max Exponential Average Squared Buffer", tt::tt_metal::Layout::TILE, param_dtype);
     }
+
+    if (tensor_args.step_scalars.has_value()) {
+        // Stochastic rounding needs a fresh seed as a per-step compute runtime argument,
+        // which defeats the point of taking the step-varying scalars as device tensors.
+        TT_FATAL(
+            args.stochastic_rounding == StochasticRounding::Disabled,
+            "Stochastic rounding is not supported when the step-varying scalars are passed as tensors");
```
philei-tt: This defeats the purpose, but doesn't make it impossible. It would still work, just not in compiler. I think this assert should be in compiler, not in ttml

philei-tt (approved):
Overall seems good, thank you! 

Please address comments I added, I approve so that I don't block you. 
Also, ping @mdragulaTT  to be able to merge.

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
```diff
@@ -106,8 +132,14 @@ ttsl::hash::hash_t AdamWDeviceOperation::compute_program_hash(
...
     auto max_exp_avg_sq_initialized = tensor_args.max_exp_avg_sq.has_value();
+    auto scalars_from_tensor = tensor_args.step_scalars.has_value();
     auto hash = tt::tt_metal::operation::hash_operation<AdamWDeviceOperation>(
-        amsgrad, stochastic_rounding, max_exp_avg_sq_initialized, param_tensor.dtype(), param_logical_shape);
+        amsgrad,
+        stochastic_rounding,
+        max_exp_avg_sq_initialized,
+        scalars_from_tensor,
+        param_tensor.dtype(),
+        param_logical_shape);
```
mdragulaTT, line 170: (nit, non-blocking) Worth adding the scalar tensors' device id to the hash

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_program_factory.cpp
```diff
@@ -339,8 +381,26 @@ AdamWProgramFactory::cached_program_t AdamWProgramFactory::create(
...
+    auto* step_size_buffer = scalars_from_tensor ? tensor_args.step_scalars->step_size.buffer() : nullptr;
+    auto* inv_sqrt_bc2_buffer = scalars_from_tensor ? tensor_args.step_scalars->inv_sqrt_bc2.buffer() : nullptr;
+    auto* decay_factor_buffer = scalars_from_tensor ? tensor_args.step_scalars->decay_factor.buffer() : nullptr;
+
+    if (scalars_from_tensor) {
+        // One aligned slot per scalar; the reader does sized reads of just the first
+        // element of each tile instead of pulling in three full f32 tiles.
+        [[maybe_unused]] auto cb_scalars = create_circular_buffer_bytes(
+            program, all_cores, kScalarsCbIndex, tt::DataFormat::Float32, 3U * kScalarSlotBytes, kScalarSlotBytes);
+    }
```
mdragulaTT, line 402: Is the CB base guaranteed 64-aligned here? On Blackhole `DRAM_ALIGNMENT` is 64 while `L1_ALIGNMENT` is 16, and this read needs `(l1_addr & 63) == (dram_addr & 63)`. It holds today only because every other CB in this op is a 64-multiple and this one is created last — a non-64-multiple CB added ahead would misalign it silently. Watcher is off in every reachable CI path, so nothing would catch it. Could you confirm, or run the suite once with `enable-watcher`?

mdragulaTT (approved):
LGTM, approving to unblock. Thank you for addressing the comments. Two small things.

1. Could you dispatch `tt-metal-l2-nightly` on your branch with `run_tt_train_unit_tests: true` and `run_blackhole: true`? The merge gate will cover the rest.
2. Left two small inline comments. And please check Pavlo's as well.

Thank you so much!

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation_types.hpp
agobeljicTT: removed

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
agobeljicTT: dropped the assert

tt-train/sources/ttml/metal/optimizers/adamw/device/adamw_device_operation.cpp
agobeljicTT: done

agobeljicTT:
Hey @mdragulaTT @zbaczewskiTT @philei-tt ,

I am very grateful for the comments and the support!

Looking at it now this is a 10 file +738 -126 PR which I find highly complex for something that should've been "a new overload that is useful for the compiler team" and ultimately a more concise way to make adamw work for us. I am not 100% sure I wanna merge this...

Let me know your thoughts.

mdragulaTT:
Agreed, let's close it. Not for size, but because #9281 reads the scalars back on host and calls the float `adamw`, so nothing consumes this overload anymore. Branch and review stay on record; if a consumer shows up we reopen and rebase.

Thanks for the diligence.

Optional: the `compute_step_scalars` dedup and the shape checks in validation improve the float path on their own — happy to review those as a small standalone PR.

Issue #52003, opened 2026-08-03
GPT-OSS prefill: register the galaxy chunked KV-PCC test in the tiered CI pipeline


mdragulaTT:
`tests/galaxy_prefill_kv_pcc.py` (chunked mode) is the integration test for multi-chunk prefill — it exercises the cache-read seam, cache batch indexing, sliding/full layers, and sinks across ≥2 chunks. But it's galaxy-gated and **not wired into the tiered CI pipeline**; the registered `gpt_oss_d_p` job only runs `tests/unit/`, which validates the RingJointSDPA operator, not this model wiring.

Ask: add a galaxy-gated job that runs ≥2 chunks and checks per-layer KV / output PCC, so the chunked path can't regress silently. Flagged by Copilot in tenstorrent/tt-metal#51992.

PR #53672, opened 2026-08-19
[gpt_oss_d_p] Variable chunk length for prefill (1k/8k)

mdragulaTT:
## Problem
Prefill chunk size is fixed per runtime build. A short (~1k) request pads up to the full chunk and wastes the compute; and 128k context needs chunk sizes that divide 131072 — 5k/10k don't.

## Solution
One runtime instance serves a **set of chunk sizes, chosen per `prefill_chunk` call**: an indexed rope per size (the block-cyclic period is size-specific), MoE buffers sized at the largest, warmup per size. `TtCombineModule.forward` gains an optional per-call token count (default = build size; legacy callers byte-identical). Code defaults move to the **1k/8k pair** (`default_chunk_size=8192`; harness/smoke defaults likewise). The migration-test manifests move to the 128k config (131072 = 16×8192) against the full-context golden; the production chunk size is an engine-side env (follow-up with the engine owners).

## Context
GPT-OSS completion item (Joe): variable chunks so a typical short prefill doesn't pay large-chunk padding; 8k replaces 10k because 128k = 16×8192. Per-request size selection in serving needs a chunk-size field on the engine wire (engine-side follow-up); this PR ships the runtime mechanism.

## Testing
- CPU unit tests for size resolution (dedupe/order/divisibility): **5 passed**; runs in the existing gpt_oss_d_p unit CI job.
- Galaxy, 36L, real weights (torus, Ring), 55k golden — chunk-size sweep: wall time tracks chunk count at a flat ~0.5 s/chunk, PCC invariant:

| chunk | chunks | wall | tok/s | min KV PCC |
|---|---|---|---|---|
| 2560 | 22 | 10.79 s | 5.2k | 0.860 |
| 5120 | 11 | 5.56 s | 10.1k | 0.861 |
| 8192 | 7 | 3.70 s | 15.2k | 0.861 |
| 10240 | 6 | 3.16 s | 17.8k | 0.862 |

- **First full 128k validation** (131072 = 16×8192, vs the full-context golden): min KV PCC across 36 layers **0.876**, prefill **8.5 s / 15.4k tok/s**.
- Short-request win: 1024 real tokens as native 1k chunk **757 ms** vs padded into 8k **826 ms**.
- **1k multi-chunk correctness**: 5×1024 chunks over the 5k golden, min KV PCC **0.920** (matches 8k-chunk runs).
- Linear-topology mode also validated on a no-torus node (see #54164).

## Impact
Short requests stop paying padding waste (~9–13% today — prefill is overhead-bound; grows once host-dispatch overhead is removed); chunking becomes 128k-compatible. Opt-in: default single-size behavior unchanged; DeepSeek/MiniMax/Kimi combine callers unaffected (optional kwarg, verified call sites + their CI stages green).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mdragula/gpt-oss-variable-chunk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mdragula/gpt-oss-variable-chunk)


models/demos/deepseek_v3_d_p/tt/moe/tt_combine.py
```diff
@@ -100,6 +99,7 @@ def forward(
         dispatched_metadata: ttnn.Tensor,
         expert_token_counts: ttnn.Tensor,
         expert_region_offsets: ttnn.Tensor,
+        seq_len_per_chip=None,
```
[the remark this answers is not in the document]
mdragulaTT: Deferring the op-level combine test: the per-call size path is exercised end-to-end and numerically verified by the 1k multi-chunk galaxy run (5x1024 chunks vs golden, min KV PCC 0.920), and legacy callers are covered by the green minimax/kimi/deepseek CI stages. A dedicated device unit test belongs with the galaxy CI registration work (#52003).

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
```diff
@@ -43,12 +43,27 @@
...
+
 @dataclass
 class TtPrefillRuntimeConfig:
     num_layers: int  # layers built/cached by this runtime (== model total for single-rank)
     max_seq_len: int  # per-user KV-cache length in tokens; must be a multiple of chunk_size
     mesh_shape: tuple = (4, 8)  # (SP rows, TP cols) on the Blackhole galaxy
-    chunk_size: int = 5120  # tokens per prefill_chunk() call; one-shot sets this == max_seq_len
+    default_chunk_size: int = (
+        8192  # per prefill_chunk() call unless overridden (8k divides 128k context); one-shot sets == max_seq_len
+    )
```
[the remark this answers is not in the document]
mdragulaTT: Good catch — real bug from the rename (an aliased c.chunk_size my rename missed). Fixed in d95d189a33.

models/demos/gpt_oss_d_p/tt/runners/manifests/gpt_oss_binding_mock_migration.yaml
```diff
@@ -12,8 +12,8 @@ global_env:
   PREFILL_SP: "4"
   PREFILL_TP: "8"
   PREFILL_FABRIC_MODE: "1d_ring"
-  PREFILL_MAX_SEQ_LEN: "56320"
-  PREFILL_CHUNK_SIZE: "5120"
+  PREFILL_MAX_SEQ_LEN: "57344"
+  PREFILL_CHUNK_SIZE: "8192"
```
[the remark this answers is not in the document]
mdragulaTT: Resolved by reverting: the binding manifests are back at the validated 5120/56320 migration-pair config (d95d189a33). Flipping the pair to 8k belongs with an engine-side change validated end-to-end, not half of a test pair.

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
```diff
@@ -399,7 +435,7 @@ def gather_layer(self, slot_id: int, layer_idx: int, n_tokens: int, kv_caches=No
         nkv = self.hf_config.num_key_value_heads
         slot = slot_id * self.config.num_layers + layer_idx
         # shard-row -> natural global position (inverse of the update_padded_kv_cache writer).
-        p = blockcyclic_positions(sp, self.config.chunk_size, self.config.max_seq_len)
+        p = blockcyclic_positions(sp, self.config.default_chunk_size, self.config.max_seq_len)
```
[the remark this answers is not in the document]
mdragulaTT: Fixed — gather_layer/kv_cache_pcc_check now take the populating chunk_size (default = config default) in d95d189a33.

models/demos/gpt_oss_d_p/tt/runners/manifests/gpt_oss_binding_loopback_migration.yaml
```diff
@@ -12,8 +12,8 @@ global_env:
   PREFILL_SP: "4"
   PREFILL_TP: "8"
   PREFILL_FABRIC_MODE: "1d_ring"
-  PREFILL_MAX_SEQ_LEN: "56320"
-  PREFILL_CHUNK_SIZE: "5120"
+  PREFILL_MAX_SEQ_LEN: "57344"
+  PREFILL_CHUNK_SIZE: "8192"
```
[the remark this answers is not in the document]
mdragulaTT: Same as above — reverted to keep the loopback pair consistent (d95d189a33).

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
```diff
@@ -43,12 +43,28 @@
...
+            )
+    return sizes
+
+
 @dataclass
 class TtPrefillRuntimeConfig:
     num_layers: int  # layers built/cached by this runtime (== model total for single-rank)
     max_seq_len: int  # per-user KV-cache length in tokens; must be a multiple of chunk_size
     mesh_shape: tuple = (4, 8)  # (SP rows, TP cols) on the Blackhole galaxy
-    chunk_size: int = 5120  # tokens per prefill_chunk() call; one-shot sets this == max_seq_len
```
kgrujcic-tt, line 51: I don't like that we change default behavior — it used to be 5120, now it's 8192. Is it necessary?

models/demos/gpt_oss_d_p/tt/moe/tt_gpt_oss_moe.py
```diff
@@ -211,7 +211,12 @@ def forward(self, x, topk_indices, topk_weights):
         expert_outputs = ttnn.unsqueeze(ttnn.unsqueeze(expert_outputs, dim=0), dim=0)
 
         combined_output = self.combine_module(
-            expert_outputs, metadata, tt_expert_token_counts, tt_expert_region_offsets
+            # Pass this chunk's actual per-chip token count so combine's output matches the scores.
```
jmalone-tt: can probably delete comments like these (doesn't actually help readability going forward, just shows LLM intent during implementation)

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
jmalone-tt: 5k chunk made sense when we were targeting 55k max sequence length, but doesn't work for full GPT OSS max sequence length. I think it's worth changing the default, but @mdragulaTT may have other points too

jmalone-tt (commented):
Looks good to me overall. Can we bundle in the changes for config .yaml files to be 128k for testing (we should have a matching golden output now, and 8k won't work with our 55k tests)
If you don't have BW, I can add a commit to the PR

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
mdragulaTT: I don't see a strong reason to keep 5120. It was inherited from DeepSeek/GLM, but the target context is 131072 and 5120 doesn't divide it (131072 = 16 × 8192). Note the default only governs the gpt-oss standalone runtime and tests; the engine passes its own value, and a config that needs a different size (e.g. the 55k tests at 5120) can still set it per-config. What's your reasoning for keeping 5120, @kgrujcic-tt ?

mdragulaTT: P.S. sizes must divide `max_seq_len`, so 5120-at-131072 is rejected by design

mdragulaTT:
> Looks good to me overall. Can we bundle in the changes for config .yaml files to be 128k for testing (we should have a matching golden output now, and 8k won't work with our 55k tests)
If you don't have BW, I can add a commit to the PR

Let's bundle it here. I'll push the manifest flip myself since the producer/binding pairs have to move together. Running the first `128k@8192` validation against your `full_context` golden tonight. The commit will follow with the measured PCC + perf tomorrow 🫡

models/demos/gpt_oss_d_p/tt/moe/tt_gpt_oss_moe.py
mdragulaTT: Done.

mdragulaTT:
PR and the description are now updated.

/codeowners ping

jmalone-tt (approved):
Looks good from my side. Let's hear @kgrujcic-tt out before merging regarding whether changing defaults is a problem

kgrujcic-tt (approved):

Issue #51231, opened 2026-07-27
[ttnn/normalization] 7 high-severity defects: missing CB reserve, host/kernel arg mismatch, wrong index


amahmudTT:
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

Seven high-severity defects in `ttnn/cpp/ttnn/operations/normalization/` that do **not** hang the op — they corrupt a CB, read uninitialized args, or silently return the wrong numbers. (The six deadlocks in this area are filed separately.)

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS: @tenstorrent/metalium-developers-mmfusedreduce

Three groups:

- **Item 1** writes a CB without reserving space, so it can overwrite tiles the consumer has not popped yet.
- **Items 2-4: host and kernel disagree.** Items 3 and 4 are in the same file and each shifts every subsequent kernel arg by one word, so fix them together — one omits a compile-time arg the reader still reads, the other still passes a runtime arg that was removed. Item 2 is the CB-existence variant: the kernel compiles in a block whose args and CB the host only supplies on a different path.
- **Items 5-7: wrong index or truncated type.** In bounds and well-formed, but addressing the wrong tile or truncating a count by its declared type.

| # | Sev | Location | Defect |
|---|---|---|---|
| 1 | H | [`normalization/layernorm_distributed/device/kernels/compute/layernorm_pre_allgather_welford.cpp:299`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/kernels/compute/layernorm_pre_allgather_welford.cpp#L299) | The compute kernel packs and pushes 2 tiles into the output CB c_14 every row iteration but never calls cb_out.reserve_back(2), so the producer never waits for th... |
[...]

PR #53677, opened 2026-08-19
Feature: normalization op merge gate tests, one test per program factory, packaging changes

iwroszTT:
### Ticket

Normalization counterpart of #53547 (reduction), from the same CI-coverage analysis.

### Problem description

The ttnn normalization family (softmax, layernorm, rmsnorm, their distributed pre/post all-gather stages, groupnorm, batch_norm) currently has no Merge Gate test coverage. This PR is going to change that.

### What's changed

Adds `test_normalization.cpp` to the `tt-nn-validation-basic` merge-gate binary: roughly one test per program factory, plus tests for the front-end code paths that pick between factories. Each test states in a comment which factory or code path it covers. Inputs are deterministic with closed-form expected outputs wherever the math allows exactness in bf16/fp32.

Two wrong-result bugs found while writing it, both now filed and both with an in-tree repro left in the file as a `DISABLED_` cell:

1. #53846 — the legacy two-pass reduction computes an inexact group mean for multi-tile groups.
2. #53803 — the legacy two-pass statistics corrupt sub-tile group widths

  Packaging fix (changes `ttnn-runtime` contents): The installed `kernels` file set for normalization globbed `groupnorm/device/kernels/*.hpp` but not `*.cpp` or `*.h`

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:iwrosz/normalization-smoke-gtests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:iwrosz/normalization-smoke-gtests)

### DIFF
--- a/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
+++ b/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
@@ -5,16 +5,11 @@
 #pragma once
 
 // Shared helpers for the per-op-family merge-gate smoke suites (test_matmul.cpp,
-// test_normalization.cpp, test_reduction.cpp, ...). These four wrappers were copy-pasted
-// into each suite as it was written; they live here so there is one definition.
+// test_normalization.cpp, test_reduction.cpp, ...).
 //
 // Kept separate from common_test_utils.hpp on purpose: that header is also included by the
 // conv and CCL tests, and these helpers pull in gtest plus the device-tensor construction
-// stack. common_test_utils.hpp still owns the numeric primitives (pcc, allclose_report,
-// relative_frobenius, check_nonfinite_positions) that expect_close() is built from.
-//
-// Suites bring these into their own `namespace detail` with using-declarations, so call
-// sites stay `detail::make_device_tensor(...)` / `detail::expect_close(...)`.
+// stack.
 
 #include <gtest/gtest.h>
 
--- a/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
+++ b/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
@@ -42,7 +42,7 @@ Tensor make_device_tensor_mc(
     DataType dtype,
     Layout layout,
     const MemoryConfig& mem_cfg) {
-    const TensorLayout tensor_layout(dtype, PageConfig(layout), mem_cfg);
+    const tt::tt_metal::TensorLayout tensor_layout(dtype, PageConfig(layout), mem_cfg);
     const tt::tt_metal::TensorSpec tensor_spec(shape, tensor_layout);
     return Tensor::from_vector(data, tensor_spec).to_device(&device, mem_cfg, ttnn::QueueId(0));
 }
--- a/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
+++ b/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
@@ -65,9 +65,21 @@ inline std::vector<float> to_float_vector(const Tensor& t) {
     return t.to_vector<float>();
 }
 
-// Tolerance-based comparison: non-finite positions must match, every element must
-// satisfy allclose(rtol, atol), and optionally PCC >= pcc_min and relative Frobenius
-// error <= frob_max (pass -1 to skip).
+// Pass as pcc_min / frob_max to skip that aggregate check explicitly.
+inline constexpr float kSkipCheck = -1.0f;
+
+// Tolerance-based comparison. Always enforced: non-finite positions must match and every
+// element must satisfy allclose(rtol, atol) -- an element-wise bound, strictly stronger at
+// these tolerances than any aggregate metric. On top of that:
+//
+// - PCC >= pcc_min, on by default (0.99). Skipped automatically when either vector is
+//   constant: correlation is undefined there (pcc() returns 0), and constant closed-form
+//   goldens are this suite's main technique, so a correct result would otherwise fail.
+//   Pass kSkipCheck to opt out explicitly, or a higher value to tighten.
+// - Relative Frobenius error <= frob_max, off by default (kSkipCheck). No universal default
+//   is sound: for the exact-zero goldens used throughout, relative_frobenius() falls back to
+//   the ABSOLUTE norm, which grows with sqrt(N) -- any fixed threshold would fail correct
+//   results at some tensor size. Opt in per case where rtol/atol are loose.
 //
 // Note the ASSERT_* macros below return from this helper, not from the calling test.
 inline void expect_close(
@@ -75,8 +87,8 @@ inline void expect_close(
     const std::vector<float>& expected,
     float rtol,
     float atol,
-    float pcc_min = -1.0f,
-    float frob_max = -1.0f) {
+    float pcc_min = 0.99f,
+    float frob_max = kSkipCheck) {
     ASSERT_EQ(actual.size(), expected.size());
     const NonfiniteReport nf = check_nonfinite_positions(actual, expected);
     ASSERT_TRUE(nf.positions_match) << "non-finite mismatch at flat index " << nf.first_mismatch_index
@@ -91,8 +103,15 @@ inline void expect_close(
         return;  // pcc / relative_frobenius NaN-poison on non-finite inputs
     }
     if (pcc_min >= 0.0f) {
-        const float p = pcc(actual, expected);
-        EXPECT_GE(p, pcc_min);
+        // PCC is undefined for constant vectors (pcc() returns 0); a deviation from a constant
+        // golden is already caught by the mandatory element-wise allclose above.
+        const auto [amin, amax] = std::minmax_element(actual.begin(), actual.end());
+        const auto [emin, emax] = std::minmax_element(expected.begin(), expected.end());
+        const bool degenerate = actual.empty() || *amin == *amax || *emin == *emax;
+        if (!degenerate) {
+            const float p = pcc(actual, expected);
+            EXPECT_GE(p, pcc_min);
+        }
     }
     if (frob_max >= 0.0f) {
         bool expected_norm_is_zero = false;
--- a/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
+++ b/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
@@ -37,7 +37,7 @@ Tensor make_device_tensor_mc(
     DataType dtype,
     Layout layout,
     const MemoryConfig& mem_cfg) {
-    const tt::tt_metal::TensorLayout tensor_layout(dtype, PageConfig(layout), mem_cfg);
+    const tt::tt_metal::TensorLayout tensor_layout(dtype, tt::tt_metal::PageConfig(layout), mem_cfg);
     const tt::tt_metal::TensorSpec tensor_spec(shape, tensor_layout);
     return Tensor::from_vector(data, tensor_spec).to_device(&device, mem_cfg, ttnn::QueueId(0));
 }
--- a/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
+++ b/tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
@@ -1,0 +1,109 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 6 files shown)


tests/ttnn/unit_tests/gtests/test_normalization.cpp
```diff
@@ -0,0 +1,1325 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+// Smoke tests for the ttnn normalization op family: one test per program
+// factory, for every op under ttnn/cpp/ttnn/operations/normalization/
```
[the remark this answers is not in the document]
iwroszTT: Added tests for GeneralWLarge / GeneralHLarge.

I did not add a test for the distributed Welford factories, because they are buggy as of now: #51231 is still open.
The post-all-gather Welford factory has a DISABLED_DistributedLayerNormPostWelford, ready to enable when #51231 closes. The problem with pre-all-gather variant is a hang and it does not have even a disabled test right now.

Also updated the file header to state these exclusions explicitly.

tests/ttnn/unit_tests/gtests/test_normalization.cpp
```diff
@@ -0,0 +1,1374 @@
...
+
+TEST_F(NormalizationSmoke, SoftmaxGeneralC) {
+    // SoftmaxProgramFactoryGeneralCLarge: dim=1 on rank-4 (neither -1 nor -2) always routes to
+    // the C factory -- there is no "small" C variant (moreh_softmax kernels).
+    auto& device = *device_;
+    constexpr uint32_t C = 4, H = 32, W = 32, HW = H * W;
+    const ttnn::Shape shape({1, C, H, W});
+    // Channel c holds the constant c everywhere -> channel softmax at every (h,w) is the same
+    // 4-way closed form: p_c = e^c / (1 + e + e^2 + e^3).
+    const float denom = 1.0f + std::exp(1.0f) + std::exp(2.0f) + std::exp(3.0f);
```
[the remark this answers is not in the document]
iwroszTT: For consistency with std::exp(2.0f) + std::exp(3.0f);  let's keep std::exp(1.0f) as-is. I should be constant-folded anyway.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -49,22 +50,8 @@ class ReductionSmoke : public TTNNFixtureWithSuiteDevice<ReductionSmoke> {};
...
-    const ttnn::Shape& shape,
-    const std::vector<T>& data,
-    DataType dtype,
-    Layout layout) {
-    const MemoryConfig mem_cfg = dram_interleaved();
-    const TensorLayout tensor_layout(dtype, PageConfig(layout), mem_cfg);
-    const tt::tt_metal::TensorSpec tensor_spec(shape, tensor_layout);
-    return Tensor::from_vector(data, tensor_spec).to_device(&device, mem_cfg, ttnn::QueueId(0));
-}
+using ttnn::test_utils::dram_interleaved;
```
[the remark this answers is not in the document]
iwroszTT: removed

mbezuljTT (approved):
approved from sources.cmake perspective

tests/ttnn/unit_tests/gtests/test_reduction.cpp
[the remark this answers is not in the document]
gajanan-choudhary: It's still here, maybe you forgot to push the branch again?

tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
```diff
@@ -0,0 +1,104 @@
...
+// satisfy allclose(rtol, atol), and optionally PCC >= pcc_min and relative Frobenius
+// error <= frob_max (pass -1 to skip).
+//
+// Note the ASSERT_* macros below return from this helper, not from the calling test.
+inline void expect_close(
+    const std::vector<float>& actual,
+    const std::vector<float>& expected,
+    float rtol,
+    float atol,
+    float pcc_min = -1.0f,
```
gajanan-choudhary: Shouldn't we keep a high default value, say `0.99` for starters for PCC and override it to a lower value only on a case-by-case basis?

tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
```diff
@@ -0,0 +1,104 @@
...
+                                   << " device=" << report.worst_margin_actual
+                                   << " ref=" << report.worst_margin_expected << " diff=" << report.worst_margin_diff
+                                   << " tol=" << report.worst_margin_tol;
+    if (nf.any_nonfinite) {
+        return;  // pcc / relative_frobenius NaN-poison on non-finite inputs
+    }
+    if (pcc_min >= 0.0f) {
+        const float p = pcc(actual, expected);
+        EXPECT_GE(p, pcc_min);
+    }
```
gajanan-choudhary, line 115: So If a user has forgotten to specify the `pcc_min` value and leaves it at default (which is `-1`), then there is no PCC check at all. Is this intentional?

tests/ttnn/unit_tests/gtests/test_normalization.cpp
```diff
@@ -0,0 +1,1376 @@
...
+    // on a rank-4 TILE input. numeric_stable defaults to true on this entry point.
+    auto& device = *device_;
+    constexpr uint32_t H = 32, W = 32;
+    const ttnn::Shape shape({1, 1, H, W});
+
+    // Launch 1: all-zero rows -> uniform 1/32 (0.03125 is exact in bf16).
+    const std::vector<float> zeros(H * W, 0.0f);
+    const auto x0 = detail::make_device_tensor<float>(device, shape, zeros, DataType::BFLOAT16, Layout::TILE);
+    const auto out0 = ttnn::softmax(x0, -1);
+    detail::expect_close(detail::to_float_vector(out0), std::vector<float>(H * W, 1.0f / 32.0f), 0.0f, 1e-3f);
```
gajanan-choudhary, line 82: No PCC and Frobenius norm thresholds specified here in the args, which means they will default to `-1`, which means there are no PCC/norm checks for this case, although I suppose `rtol`/`atol` are meant to catch things here. Is that intentional?

tests/ttnn/unit_tests/gtests/test_normalization.cpp
```diff
@@ -0,0 +1,1376 @@
...
+namespace detail {
+
+// Flattened row-major pattern alternating a, b along the width (all our widths are even).
+inline std::vector<float> dist_norm_alternating(size_t n, float a, float b) {
+    std::vector<float> v(n);
+    for (size_t i = 0; i < n; ++i) {
+        v[i] = (i % 2 == 0) ? a : b;
+    }
+    return v;
+}
```
gajanan-choudhary: Nit pick: This function is verbatim identical to `norm_alternating` function implemented up in the file (with `a` -> `even_col`, `b` -> `odd_col` replacements). If you happen to push in some other fixes/changes to the PR, maybe fix this as well and keep just one around.

tests/ttnn/unit_tests/gtests/test_normalization.cpp
```diff
@@ -0,0 +1,1376 @@
...
+        std::nullopt,
+        ttnn::CoreGrid(1, 1));
+    detail::expect_close(detail::to_float_vector(out_gb), detail::gn_golden_expected(32, 2.0f, 1.0f), 0.0f, 0.03f);
+}
+
+TEST_F(NormalizationSmoke, GroupNormWelfordInterleaved) {
+    // GroupNormNoMcastProgramFactory with use_welford=true (kernels/compute/welford_groupnorm.cpp).
+    // Welford interleaved requires TILE in/out (groupnorm_device_operation.cpp) -- satisfied.
+    // CAUTION: the actively-broken CI welford is the SHARDED v2 path (#53143 / #52700); this is the
+    // interleaved path. Verify on hardware; exclude if red, citing #53143 / #50304.
```
gajanan-choudhary: Did you determine if this is failing or passing on both WH and BH? If it passes on both, then I suppose this comment can be removed?

tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
gajanan-choudhary: Similar concern for `frob_max` below

gajanan-choudhary (approved):
LGTM with non-blocking questions/concerns

bbradelTT (approved):
Approved.

Cursor Grok has a few nits. First couple look like they need the description to be updated, an include should be added, and potential updates (you can probably leave as is) to adjust assert usage and tighten bounds. 

1. Title vs contents. Packaging changes ttnn-runtime contents. A silent groupnorm/batch_norm JIT break on install is a product bug, not a test-only change. Worth renaming or calling it out in the summary so it does not get treated as skippable.

2. Stale PR body. Body still says “38 tests”; the file is 39 enabled + 4 disabled.

3. Missing #include <numbers>. SoftmaxGeneralC uses std::numbers::e_v<float> (clang-tidy suggestion) but the includes never add <numbers>. CI compiles via a transitive include; that is brittle.

4. expect_close + ASSERT_* in a helper. Documented, but gtest ASSERT_* only returns from the helper. The test is marked failed, then the caller keeps going. Fine when expect_close is the last statement; SoftmaxWideRowSum and BatchNormPreallocatedOutput do extra work afterward. Not a blocker.

5. Wide softmax tolerances. SoftmaxGeneralWLarge / SoftmaxWideRowSum use rtol=0.2. That is enough to catch a wrong factory or a blow-up, and the row-sum band [0.9, 1.1] is the real #52045 canary. Just know these cells will not catch a small accuracy regression.

tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
iwroszTT: The `-1.0f` is a convenient default because the golden values in this suite are frequently constant and pcc returns zero for constant tensors.

torch isn't available in the merge-gate gtests, so the goldens are calculated with closed-form formulas, and closed forms that give a constant are quite useful (a zero-variance group makes any statistics error saturate the output, which is the main detection mechanism here).

It's a fair observation that since we decide to still use PCC we should design against a forgotten pcc_min: I'll change the default to 0.99 with an automatic skip when either of the vectors is constant.

tests/ttnn/unit_tests/gtests/smoke_test_utils.hpp
iwroszTT: It's also a result of the cases which output an all-zero output. In these cases relative_frobenius() falls back to the absolute norm, for which there's no sensible default threshold for every tensor size; It's up to the test writer to come up with correct frob_max for the test.

Note that allclose is always enabled and bounds every element.

tests/ttnn/unit_tests/gtests/test_normalization.cpp
iwroszTT: Yes, it's intentional. The first launch's golden is the constant 1/32 (every element identical), where PCC is undefined (pcc() returns 0 for constant vectors) and the check is based here on rtol/atol only.

tests/ttnn/unit_tests/gtests/test_normalization.cpp
iwroszTT: fixed

tests/ttnn/unit_tests/gtests/test_normalization.cpp
iwroszTT: It's passing on both. Updated the comment.

PR #53685, opened 2026-08-19
[Cleanup] Deprecating legacy `LaunchProgram` and `CompileProgram`

obacherikovTT:
### PR Category
Cleanup

### Summary

`LaunchProgram` and `CompileProgram` from _tt_metal/api/tt-metalium/tt_metal.hpp_ are not used externally, so they are deprecated:

1. `detail::LaunchProgram` in favor of using `MeshWorkload`.
  Internal `LaunchProgram` is added in _program_impl.hpp_ that uses `MeshWorkload`, making it compatible with both slow and fast dispatch. If there's interest, we can expose this convenience helper publicly.
  Note that a couple of calls to `detail::LaunchProgram` are left in the codebase where they pass `force_slow_dispatch = true`, because refactoring them would make this PR more complicated.

2. `detail::CompileProgram` without direct public alternative.
  All its usage was internal, and it's replaced with `ProgramImpl::compile`.
  Note that it's valid to pass `MeshDevice` instead of `Device` there, see this:
<details>
  <summary>explanation by AI</summary>
  
`MakeProgramFromSpec` and `MeshWorkloadImpl::compile` both call `compile_and_allocate` with a `MeshDevice`, which then calls `compile()`. The JIT work does not use `IDevice::id()` as a chip id.

What compile actually uses:

Call | MeshDevice | First physical device (get_devices()[0])
-- | -- | --
build_id() | reference_device()->id() = first chip | Device::build_id() = that chip’s id
extract_context_id() | mesh context | same, via Device::get_mesh_device()
arch() / is_initialized() | mesh implementations | same for a live unit mesh
id() | unique mesh id | chip id

`build_id()` and context drive the build env, kernel hash, ELF paths, watcher, and inspector. On a unit mesh those match `get_devices()[0]`.

`id()` is only used to key `sub_device_ids_`. Fast dispatch already looks that up with the mesh (`determine_sub_device_ids(mesh_device)`), so passing the mesh is the consistent key. Mixing compile(physical) with later `determine_sub_device_ids(mesh)` (or the reverse) just misses that cache and recomputes.

On a multi-chip mesh, `MeshDevice::build_id()` is always the reference (first) chip. That matches compiling against `get_devices()[0]`. It is not equivalent to compiling against some other chip if harvest masks differ.  
</details>

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:obacherikov-program-legacy-api)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:obacherikov-program-legacy-api)

### DIFF
--- a/tests/tt_metal/multihost/fabric_tests/intermesh_routing_test_utils.cpp
+++ b/tests/tt_metal/multihost/fabric_tests/intermesh_routing_test_utils.cpp
@@ -79,7 +79,7 @@ void run_unicast_sender_step(BaseFabricFixture* fixture, tt::tt_metal::distribut
     const auto topology = fabric_context.get_fabric_topology();
     TT_FATAL(topology == Topology::Mesh, "Intermesh Routing tests need Dynamic Routing enabled.");
 
-    auto devices = fixture->get_devices();
+    const auto& devices = fixture->get_devices();
 
     // Synchronize seeds across hosts (sender and receiver must use the same seed for randomization)
     uint32_t time_seed = std::chrono::system_clock::now().time_since_epoch().count();
@@ -158,8 +158,7 @@ void run_unicast_sender_step(BaseFabricFixture* fixture, tt::tt_metal::distribut
     tt_metal::SetRuntimeArgs(sender_program, sender_kernel, sender_logical_core, sender_runtime_args);
 
     // Run sender program
-    fixture->RunProgramNonblocking(sender_device, sender_program);
-    fixture->WaitForSingleProgramDone(sender_device, sender_program);
+    tt_metal::LaunchProgram(*sender_device, std::move(sender_program), /*wait_until_cores_done=*/true);
 
     // Validate status of sender
     std::vector<uint32_t> sender_status;
@@ -200,7 +199,7 @@ void run_unicast_recv_step(BaseFabricFixture* fixture, tt::tt_metal::distributed
     const auto& fabric_context = control_plane.get_fabric_context();
     const auto topology = fabric_context.get_fabric_topology();
     TT_FATAL(topology == Topology::Mesh, "Intermesh Routing tests need Dynamic Routing enabled.");
-    auto devices = fixture->get_devices();
+    const auto& devices = fixture->get_devices();
 
     // Synchronize seeds across hosts (sender and receiver must use the same seed for randomization)
     uint32_t time_seed = 0;
@@ -248,8 +247,7 @@ void run_unicast_recv_step(BaseFabricFixture* fixture, tt::tt_metal::distributed
     auto recv_program = create_receiver_program(compile_time_args, receiver_runtime_args, receiver_logical_core);
 
     // Run receiver program
-    fixture->RunProgramNonblocking(receiver_device, *recv_program);
-    fixture->WaitForSingleProgramDone(receiver_device, *recv_program);
+    tt_metal::LaunchProgram(*receiver_device, std::move(*recv_program), /*wait_until_cores_done=*/true);
 
     // Validate status of the receiver
     std::vector<uint32_t> receiver_status;
@@ -306,7 +304,7 @@ void run_mcast_sender_step(
     );
     // Randomly select a mcast sender device
     auto sender_phys_id = control_plane.get_physical_chip_id_from_fabric_node_id(mcast_sender_node);
-    auto sender_device = fixture->get_device(sender_phys_id);
+    const auto& sender_device = fixture->get_device(sender_phys_id);
     const auto& worker_grid_size = sender_device->compute_with_storage_grid_size();
     // Randomly select a mcast sender core
     auto sender_x = std::uniform_int_distribution<uint32_t>(0, worker_grid_size.x - 2)(global_rng);
@@ -366,8 +364,7 @@ void run_mcast_sender_step(
     tt_metal::SetRuntimeArgs(mcast_send_program, mcast_send_kernel, sender_logical_core, sender_runtime_args);
 
     log_debug(tt::LogTest, "Run Sender on: {}", sender_device->id());
-    fixture->RunProgramNonblocking(sender_device, mcast_send_program);
-    fixture->WaitForSingleProgramDone(sender_device, mcast_send_program);
+    tt_metal::LaunchProgram(*sender_device, std::move(mcast_send_program), /*wait_until_cores_done=*/true);
 
     // Validate status of sender
     std::vector<uint32_t> sender_status;
@@ -463,12 +460,12 @@ void run_mcast_recv_step(
     // NOLINTNEXTLINE(bugprone-nondeterministic-pointer-iteration-order)
     for (auto& [dev, recv_program] : recv_programs) {
         log_debug(tt::LogTest, "Run receiver on: {}", dev->id());
-        fixture->RunProgramNonblocking(dev, *recv_program);
+        fixture->RunProgramNonblocking(dev, std::move(*recv_program));
     }
 
     // NOLINTNEXTLINE(bugprone-nondeterministic-pointer-iteration-order)
     for (auto& [dev, recv_program] : recv_programs) {
-        fixture->WaitForSingleProgramDone(dev, *recv_program);
+        fixture->WaitForSingleProgramDone(dev);
     }
     // Validate status of the receiver
     // Request test results from the sender host and ensure that they match
--- a/tests/tt_metal/tt_fabric/common/fabric_fixture.hpp
+++ b/tests/tt_metal/tt_fabric/common/fabric_fixture.hpp
@@ -19,6 +19,7 @@
 #include "common/tt_backend_api_types.hpp"
 #include <llrt/tt_cluster.hpp>
 #include <tt-metalium/allocator.hpp>
+#include "impl/program/program_impl.hpp"
 #include "tt_metal/impl/dispatch/slow_dispatch.hpp"
 #include "test_host_kernel_common.hpp"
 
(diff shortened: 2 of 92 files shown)


tt_metal/api/tt-metalium/tt_metal.hpp
```diff
@@ -160,8 +160,10 @@ void ReadShard(Buffer& buffer, std::vector<DType>& host_buffer, const uint32_t&
 
 // Launches all kernels on cores specified with kernels in the program.
 // All kernels on a given Tensix core must be launched.
+[[deprecated("Use distributed::EnqueueMeshWorkload instead. detail::LaunchProgram will be removed after 2026-09-20.")]]
```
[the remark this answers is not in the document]
obacherikovTT: @akerteszTT Should we be concerned about this?

mo-tenstorrent (approved):

tests/tt_metal/multihost/fabric_tests/intermesh_routing_test_utils.cpp
```diff
@@ -158,8 +158,7 @@ void run_unicast_sender_step(BaseFabricFixture* fixture, tt::tt_metal::distribut
     tt_metal::SetRuntimeArgs(sender_program, sender_kernel, sender_logical_core, sender_runtime_args);
 
     // Run sender program
-    fixture->RunProgramNonblocking(sender_device, sender_program);
-    fixture->WaitForSingleProgramDone(sender_device, sender_program);
+    tt_metal::LaunchProgram(*sender_device, std::move(sender_program), /*wait_until_cores_done=*/true);
```
obacherikovTT, line 161: Simplifying RunProgramNonblocking + WaitForSingleProgramDone combination to a single blocking `LaunchProgram`.

tests/tt_metal/tt_fabric/common/fabric_fixture.hpp
```diff
@@ -134,33 +135,13 @@ class BaseFabricFixture : public ::testing::Test {
 
     // NOLINTNEXTLINE(readability-make-member-function-const)
     void RunProgramNonblocking(
-        const std::shared_ptr<tt::tt_metal::distributed::MeshDevice>& device, tt::tt_metal::Program& program) {
-        if (this->slow_dispatch_) {
-            tt::tt_metal::slow_dispatch::LaunchProgram(*device, program, false);
```
obacherikovTT, line 139: Since `force_slow_dispatch = true` wasn't passed here, this should be an equivalent replacement.

tests/tt_metal/tt_metal/api/dataflow_buffer/test_borrowed_memory_dataflow_buffer.cpp
```diff
@@ -260,13 +263,11 @@ void run_borrowed_memory_dfb_program(
     std::iota(input.begin(), input.end(), 0u);
     slow_dispatch::WriteToBuffer(src_tensor.mesh_buffer(), input);
 
-    slow_dispatch::LaunchProgram(mesh_device, program, /*wait_until_cores_done=*/true);
+    distributed::EnqueueMeshWorkload(mesh_device.mesh_command_queue(), workload, /*blocking=*/true);
```
obacherikovTT, line 266: We use `EnqueueMeshWorkload` explicitly in cases when the program is accessed after the launch.

tt_metal/impl/dispatch/slow_dispatch.hpp
```diff
@@ -112,12 +112,4 @@ inline bool ReadFromDRAMChannel(
         physical_device_from_unit_mesh(unit_mesh), dram_channel, address, host_buffer);
 }
 
-inline void CompileProgram(distributed::MeshDevice& unit_mesh, Program& program) {
-    detail::CompileProgram(physical_device_from_unit_mesh(unit_mesh), program);
-}
```
obacherikovTT, line 117: Note that `ProgramImpl::compile` still takes `IDevice` not `MeshDevice` like we had here.
That's because there are callers that still pass single `Device` into it, and untangling them is out of scope of this PR.

tt-vjovanovic (approved):

tt_metal/api/tt-metalium/tt_metal.hpp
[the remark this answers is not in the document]
akerteszTT: Oh man, this API is worse than I thought. I thought it was an explicit slow dispatch API; I had not realized it was one with a control switch.

The codebase's intended mechanism for controlling fast vs. slow dispatch is the envvar in RTOptions. However, the RTOptions are captured _once_ for the runtime as a whole. You should double check me, but I believe that toggling the envvar once the runtime is live will have no effect.

That is as designed... Blaze currently relies on flipping back and forth between slow and fast dispatch. They use fast dispatch at the beginning of model execution to rapidly load weights into memory, then flip to slow dispatch to reclaim the dispatch column for compute. Please look into this -- you cannot remove this API if they rely on it.

That mechanism is 100% an internal-only hack. So if your findings are as I suspect, the correct solution will be to shunt this over to `api/internal`. (Personally, I would also remove the switch and make it explicitly slow dispatch there... there is NO reason to expose even a backdoor single-device fast dispatch API.)

Please also talk to @jbaumanTT. He is the expert on all things dispatch and can give you a better answer than me.

akerteszTT (commented):
I'm concerned that you might break Blaze with this change -- could you please investigate?

tt_metal/api/tt-metalium/tt_metal.hpp
```diff
@@ -193,6 +195,7 @@ void WaitProgramDone(IDevice* device, Program& program, bool read_device_profile
  * a user wants to compile a program with Slow Dispatch Force Enabled (advanced feature, currently used internally to
  * launch Fast Dispatch Firmware and in the Device Performance Profiler)           | bool      | | No |
  */
+[[deprecated("Use MakeProgramFromSpec instead. CompileProgram will be removed after 2026-09-20.")]]
```
akerteszTT: This deprecation message is incorrect.

`MakeProgramFromSpec` is a Metal 2.0 API. As Metal 2.0 is still in experimental, we are not broadly recommending that users migrate to it yet. 

More importantly, it not a replacement for the API that you are removing. This API takes a `Program` and compiles it. In Metal 2.0, we maintain an invariant is that a successfully created `Program` object is valid to execute on the device. That means that kernel JIT compilation is part of `Program` creation. So in Metal 2.0, the idea of compiling an uncompiled `Program` object does not make sense.

In the legacy programming model, a `Program` is automatically JIT compiled at the time of first enqueue. I am not certain whether this also applies to `LaunchProgram`. This you'll need to investigate.

tt_metal/api/tt-metalium/tt_metal.hpp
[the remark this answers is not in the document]
obacherikovTT: LaunchProgram is only used internally within runtime. I checked that Blaze doesn't use it.

smeydanshahiTT (approved):

tt_metal/api/tt-metalium/tt_metal.hpp
obacherikovTT: Updated the message.

bbradelTT (approved):

tt_metal/api/tt-metalium/tt_metal.hpp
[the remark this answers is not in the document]
jbaumanTT: force_slow_dispatch is only used, as far as I know, to disable some validation that checks that you're not using slow dispatch when fast dispatch is enabled. It's used internally in the runtime to launch programs using slow dispatch when fast dispatch is enabled, and should never have been made public. Setting it to false doesn't actually let you do fast dispatch with this API, it just causes a TT_FATAL if you call this while fast dispatch is enabled.

jbaumanTT (approved):

tt_metal/api/tt-metalium/tt_metal.hpp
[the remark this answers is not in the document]
obacherikovTT: Thank you for the clarification!
It sounds like at least in tests it's safe to replace legacy `LaunchProgram(..., force_slow_dispatch=true)` with the new version based on `MeshWorkload`.

mo-tenstorrent:
Could you also run https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-profiler.yaml on your PR. profiler sync might be affected as it is one of the pieces that internally uses launchprogram

obacherikovTT:
> Could you also run https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-profiler.yaml on your PR. profiler sync might be affected as it is one of the pieces that internally uses launchprogram

@mo-tenstorrent I didn't actually change LaunchProgram calls in the profiler in this PR to be safe. Anyway, the pipeline passes on the second attempt
https://github.com/tenstorrent/tt-metal/actions/runs/32519418572

sidnadTT (approved):

tt_metal/api/tt-metalium/tt_metal.hpp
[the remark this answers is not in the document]
akerteszTT: Thanks -- concern withdrawn.

tt_metal/api/tt-metalium/tt_metal.hpp
```diff
@@ -193,6 +195,9 @@ void WaitProgramDone(IDevice* device, Program& program, bool read_device_profile
  * a user wants to compile a program with Slow Dispatch Force Enabled (advanced feature, currently used internally to
  * launch Fast Dispatch Firmware and in the Device Performance Profiler)           | bool      | | No |
  */
+[[deprecated(
+    "Program is automatically compiled at the time of first enqueue. "
```
akerteszTT: Another tweak here, please.

If you're using _legacy_ APIs, Program is automatically compiled at the time of first enqueue. If you're using Metal 2.0 API, Program is automatically compiled at Program creation time. 

I would keep it simple and write "Program is compiled automatically by the runtime infrastructure; this API is unnecessary".

akerteszTT (approved):
See minor nit on deprecation message :)

PR #53687, opened 2026-08-19
[Feature] Quasar: host-selected semaphore mechanisms

ymoussaTT:
### PR Category

Feature

### Summary

Currently all semaphores use read-modify-write to increment/decrement the count. Since DM cores access L1 through a write-back cache, a remote node using a semaphore over the NoC and DMs living on the semaphore's own node can race, and one of the operations gets overwritten. Even two DMs on the same node can race as both read the same value, both add to it, and one increment is lost. There was no option that worked everywhere and was truly atomic. RISC-V AMOs are fast, but only atomic among DMs that live on the same node as the semaphore. NoC atomic operations are always atomic, as long as everyone, including the semaphore's own node, uses them, but are slower. If there is only ever one binder instance, read-modify-write stays safe, since no other kernel or thread interacts with the semaphore, and it is the fastest option.

This PR has the host pick each semaphore's mechanism at build time, from the kernels that bind it. Every binder counts as a possible reader and writer. The host picks the fastest mechanism that leaves all operations atomic/safe:

- If there is exactly one binder instance pick `LOCAL_NONATOMIC` (read-modify-write)
- If every binder is a DM kernel/thread on the semaphore's node pick `DM_LOCAL_CACHED` (32-bit AMO on the cached alias)
- If the semaphore is reachable off-node pick `EXTERNAL` (only uses NoC atomic operations, even on the node where the semaphore lives).

Nothing changes for users of the semaphores. The mechanism travels through a generated per-kernel table that is included in the kernels at build time. Gen1 (Wormhole/Blackhole) is untouched; the changes are Gen2 only.

### Notes for reviewers

- Cached semaphores live in a dedicated 128B pool because a cached AMO dirties a whole 64B line and the cache's write-back overwrites its NoC-written neighbours.
- `down()` is multi-consumer safe on both atomic tiers (LR/SC retry for cached semaphores; 4-bit NoC-CAS lock on EXTERNAL). Two invariants come with this: an EXTERNAL semaphore can never legitimately hold `0xFFFFFFFF` (the CAS-return sentinel), and consumers run on the semaphore's node.
- A bound semaphore's word should never be accessed directly (get_semaphore(), the noc_semaphore_* free functions, raw pointers); always go through the Semaphore class. A raw access is invisible to the host's mechanism choice and can silently cause a race condition.
- Testing: 44 tests on the Quasar 2×3 emulator. They assert the baked mechanism and exact final counts, and the multi-consumer proofs add a watchdog hart that latches the max value observed, since an exact final count alone can't see a double-spend that wraps to a very large count temporarily.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ymoussa/quasar_semaphore_scope)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ymoussa/quasar_semaphore_scope)


tt_metal/hw/firmware/src/tt-2xx/dm.cc
```diff
@@ -236,6 +236,10 @@ extern "C" uint32_t _start1() {
     if (hartid == 0) {
         extern uint32_t __ldm_data_start[];
         do_crt1(__ldm_data_start);
+        // Zero these words used for semaphores at boot
+        for (uint32_t w = 0; w < (MEM_NOC_CAS_RET_SIZE + MEM_NOC_SEM_LOCK_SIZE + MEM_DM_CACHED_SEM_SIZE) / 4; w++) {
+            reinterpret_cast<volatile uint32_t*>(MEM_L1_UNCACHED_BASE + MEM_NOC_CAS_RET_BASE)[w] = 0;
+        }
```
kstevensTT, line 249: Should we invalidate these cache lines immediately after?

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -40,22 +83,66 @@ class Semaphore {
...
+    explicit __attribute__((always_inline)) Semaphore(uint32_t semaphore_id) :
+        l1_offset_(sem_l1_offset(semaphore_id)), scope_(sem_scope_of(semaphore_id)) {}
 
     /**
      * @brief Increment the semaphore by the specified value.
-     * @note Currently atomicity is not guaranteed, multiple cores incrementing simultaneously may lead to lost updates.
+     *
+     * DM_LOCAL_CACHED: atomic 32-bit AMO on the cached alias.
+     * EXTERNAL:        self-targeted NoC atomic increment.
+     * LOCAL_NONATOMIC: L1 read-modify-write (not atomic).
```
kstevensTT, line 101: You dropped the comment about atomicity is not guaranteed. Still need this comment for WH/BH.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -67,25 +154,105 @@ class Semaphore {
...
         noc_semaphore_inc(dest_noc_addr, value, noc.get_noc_id(), vc);
     }
 
     /**
      * @brief Decrement the semaphore by the specified value, blocking until the semaphore is sufficient.
-     * @note Currently atomicity is not guaranteed, multiple cores decrementing simultaneously may lead to lost updates.
+     *
+     * DM_LOCAL_CACHED: multi-consumer-safe via LR/SC retry loop.
+     * EXTERNAL:        multi-consumer-safe via a NoC-CAS lock, consumers must run on the semaphore's node.
+     * LOCAL_NONATOMIC: single-owner (non-atomic) decrement.
```
kstevensTT, line 160: You dropped the comment about atomicity is not guaranteed. Still need this comment for WH/BH.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -67,25 +154,105 @@ class Semaphore {
      * @param value The value to increment the semaphore by.
      * @param vc The virtual channel to use for the transaction (default is NOC_UNICAST_WRITE_VC).
      */
-    void up(const Noc& noc, uint32_t noc_x, uint32_t noc_y, uint32_t value, uint8_t vc = NOC_UNICAST_WRITE_VC) {
+    __attribute__((always_inline)) void up(
+        const Noc& noc, uint32_t noc_x, uint32_t noc_y, uint32_t value, uint8_t vc = NOC_UNICAST_WRITE_VC) {
+        ASSERT(scope_ != SemScope::DM_LOCAL_CACHED);  // Not valid on a DM_LOCAL_CACHED semaphore
```
kstevensTT: My understanding is this is the accepted pattern to get atomics in WH/BH, even for local atomics. For backwards compatibility, we should consider supporting this case.

tt_metal/jit_build/jit_build_settings.hpp
```diff
@@ -7,14 +7,81 @@
...
+struct SemBindingEntry {
+    std::string name;
+    uint16_t id = 0;
+    SemScope scope = SemScope::LOCAL_NONATOMIC;
+    uint32_t total_binder_harts = 0;
+};
+
+// The generated semaphore section.
+// Emits the TT_METAL2_SEM_SCOPE_TABLE define. Must be the FIRST line of the generated header,
+// before any include, so no include chain can compile noc_semaphore.h without it.
```
kstevensTT: I think this comment should go where the method is called, so it is more obvious to anyone making changes there.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -539,13 +610,27 @@ void jit_build_genfiles_kernel_include(
         kernel_header_content += "#include \"named_args_generated.h\"\n";
     }
     ////////////////////////////////////////////////////////////
+
+    // Automatically inject the cached-semaphore pool stubs: rename the kernel's entry point and define the
+    // real kernel_main() as a wrapper that calls sem::init_dm_cached() first and sem::finish_dm_cached() last.
+    static constexpr const char* kUserEntry = "tt_dm_cached_user_kernel_main_";
+    if (has_cached_sem) {
+        kernel_header_content += string("#define kernel_main ") + kUserEntry + "\n";
+    }
```
kstevensTT: This whole trick with kernel_main seems very fragile to me, especially since kernel_main is often auto-generated already.

My preference would be:
- Define a `tt_dm_cached_sem_init();` function.
- Have `dmk.cc` call `tt_dm_cached_sem_init();` prior to calling `kernel_main();`.
- You can do similar if you need a `tt_dm_cached_sem_finish();` type function as well for consistency.

But get feedback from others as well.

tests/tt_metal/tt_metal/test_sem_scope.cpp
```diff
@@ -0,0 +1,1121 @@
...
+    const uint32_t observed = run_concurrent(SemScope::DM_LOCAL_CACHED, "MODE_CONCURRENT_UP");
+    const uint32_t expected = num_dms_ * concurrent_iterations;
+    log_info(LogTest, "DM_LOCAL_CACHED concurrent up value(): {} (expected {})", observed, expected);
+    EXPECT_EQ(observed, expected) << "Semaphore<DM_LOCAL_CACHED>::up() lost updates under concurrency.";
+}
+
+// (num_dms-1) producers up(1) while a single consumer drains them all. A non-atomic down()
+// loses producer increments, so the failure mode is a starved consumer timing out, not a
+// wrong count.
+TEST_F(SemScopeFixture, TestExternalProducerConsumer) {
```
kstevensTT, line 645: I don't like tests that hang on failure. Is it possible to write these in a way where they catch the timeout and actually fail? Note that whatever is done would have to work on all systems (i.e. slow RTL sims & fast silicon).

tt_metal/hw/firmware/src/tt-2xx/dm.cc
ymoussaTT: I agree and made the change, but I think it should be invalidated beforehand. There might be dirty lines in the cache at boot, so we discard them first (so they can't write back over the zeros) and then write the zeros. After that the state is clean, and since these words are only ever accessed one way each (the pool through the cache, the rest directly in L1 via uncached reads and NoC atomics), it stays clean.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: Added back for WH/BH

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: Added back for WH/BH

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: Done, on a DM_LOCAL_CACHED semaphore this form now redirects to the local AMO, so the WH/BH pattern works under every scope. I also added two new emulator tests to cover this case that both PASS.

tt_metal/jit_build/jit_build_settings.hpp
ymoussaTT: That comment already lives at the call site too (genfiles.cpp, right above the emit call), I can remove it from here if you think the duplication isn't adding value.

tt_metal/jit_build/genfiles.cpp
ymoussaTT: I agree, I think I like that way better, the stubs already existed as generated functions (sem::init_dm_cached() / sem::finish_dm_cached() in kernel_bindings_generated.h), so the change is easy. dmk.cc now calls them around kernel_main(), guarded by a define the generated header emits so kernels without cached semaphores (and legacy kernels) compile unchanged.

I will keep this conversation open though if anyone else has other comments on this same case.

tests/tt_metal/tt_metal/test_sem_scope.cpp
ymoussaTT: Done, the test kernels now have a spin cap. The cap is a iteration count (~100x any healthy wait), so it behaves the same on slow RTL sims and fast silicon. If it's ever hit, something is wrong, and the kernel bails out and reports the count it reached, so the tests fail instead of hanging.

kstevensTT (approved):

sagarwalTT (approved):
Just wondering, does `QuasarMultipleClustersMultiSemaphorePipeline` in `test_quasar_semaphores.cpp` pass with this PR? It hasn't been working for a couple months in `main`. The test tests semaphore operations across 2 cores, but it hangs on one of the semaphore syncs.

ymoussaTT:
> Just wondering, does `QuasarMultipleClustersMultiSemaphorePipeline` in `test_quasar_semaphores.cpp` pass with this PR? It hasn't been working for a couple months in `main`. The test tests semaphore operations across 2 cores, but it hangs on one of the semaphore syncs.

Yes! It passes on this PR branch.

arminaleTT (approved):

akerteszTT (commented):
This is a good solution! I really appreciate that you've kept the gory details out of the host API surface and handle the semaphore decisions within the runtime.

That said, I have two requests for design changes to the way in which you've integrated these new semaphore capabilities into the Metal 2.0 backend:

 1. **Factor all of the semaphore implementation-specific details in `program_spec.cpp` into a separate file.** `program_spec.hpp` is monstrous and we were about to refactor it. The logic you're adding is distinct from the basic Metal 2.0 convention conversion. It's also (by necessity, not your fault!) awkwardly split into multiple chunks across the file, rather than a solvr in one contiguous block. This makes it super hard to follow; a separate file with helpers invoked from the main Metal 2.0 back will fix that.
 2. **Create a proper `SemaphoreBindingToken` type, and refactor accordingly.** Every Metal 2.0 resource (except for semaphore) follows a pattern. The framework emits an opaque `<Resource>BindingToken`, carrying all required info, into the generated header. The corresponding Device 2.0 resource is constructed directly from the binding token. In the case of semaphore, I didn't bother to create an explicit `SemaphoreBindingToken` type... the info conveyed was just a single int. However, your change adds significant complexity to the handling of semaphore. What you need is a binding token type for semaphore, then to refactor everything to follow that "binding token carries all the resource info down from the host" pattern.

I've got a detailed AI-generated writeup that describes the necessary changes at a finer granularity than the summary above; I can share that with you privately if you'd find it helpful.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -8,13 +8,64 @@
...
+ *                     atomic's reply that has not yet arrived.
+ *
+ * Per-kernel scope tables are constructed at build time by the host and baked into the kernel.
+ * The kernel code can then query the scope of a semaphore for the appropriate mechanism to use.
+ *
+ * @note Never access a bound semaphore's word directly (get_semaphore(), the noc_semaphore_*
+ *       free functions, raw pointers), always go through this class. A raw access is
+ *       invisible to the host's mechanism choice and can silently race it.
+ */
+enum class SemScope : uint8_t {
```
abhullar-tt: is it possible to move this and sem_scope_of outside of the public header?

tt_metal/hw/firmware/src/tt-2xx/dmk.cc
```diff
@@ -105,7 +105,15 @@ uint32_t _start() {
         EARLY_RETURN_FOR_DEBUG
 
         WAYPOINT("K");
+#ifdef TT_DM_CACHED_SEM_STUBS
+        // When the kernel binds DM_LOCAL_CACHED semaphores: seed their
+        // pool rows once per program, and restore them on the way out.
+        sem::init_dm_cached();
```
abhullar-tt: nit can we rename this to `init_dm_local_cached` and `finish_dm_local_cached`

abhullar-tt: nit: can you rename this to `init_dm_local_cached` and `finish_dm_local_cached`

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
abhullar-tt: could we move this and `sem_scope_of` outside the public api?

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -40,52 +92,201 @@ class Semaphore {
...
+                if (consumed_cas4(/*cmp4=*/0, /*swap4=*/1) != 0) {
+                    continue;  // another consumer holds the lock, wait for it to release
+                }
+                // Re-check under the lock, no other consumer can decrement now.
+                invalidate_l1_cache();
+                const bool ok = (*sem_addr) >= value;
+                if (ok) {
+                    WAYPOINT("NSDD");
+                    // Atomic subtract: the NoC only has atomic ADD (INCR_GET), so add the two's complement.
+                    noc_semaphore_inc(sem_noc, (uint32_t)(0u - value));
```
abhullar-tt, line 212: it seems like in here we do the CAS which programs `R_SRC_ADDR` for ret value and then when we call `noc_semaphore_inc` and don't program ret addr. I think since we don't reprogram `R_SRC_ADDR` it never goes back to its default. Any subsequent atomics would also continue to use this ( I think, correct me if I am wrong @vvukomanovicTT ).

abhullar-tt: I think restoring it would make more sense than having subsequent atomics explicitly set the default ret addr

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -40,52 +92,201 @@ class Semaphore {
...
+                    noc_async_atomic_barrier();
+                    WAYPOINT("NSDD");
+                    // confirm the subtract's reply was received
+                    for (uint32_t spin = 0; spin < 1024u && *ret_word == kCasSentinel; spin++) {
```
abhullar-tt: why are we doing this spin? doesn't the barrier ensure that the subtract reply was received?

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -40,52 +92,201 @@ class Semaphore {
...
+                    observed = __atomic_load_n(word, __ATOMIC_RELAXED);
+                }
+                WAYPOINT("NSDD");
+            } while (!__atomic_compare_exchange_n(
+                word, &observed, observed - value, /*weak=*/false, __ATOMIC_SEQ_CST, __ATOMIC_SEQ_CST));
+#else
+            ASSERT(false);  // the host census never bakes CACHED for this platform
+#endif
+        } else if (scope_ == SemScope::EXTERNAL) {
+#if defined(ARCH_QUASAR) && !defined(COMPILE_FOR_TRISC) && !defined(TT_EMULE_USE_L1_POOL)
```
abhullar-tt, line 183: why do we have explicit call out for `TT_EMULE_USE_L1_POOL` ?

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
```diff
@@ -40,52 +92,201 @@ class Semaphore {
...
         WAYPOINT("NSDW");
-        do {
-            invalidate_l1_cache();
-        } while ((*sem_addr) < value);
-        WAYPOINT("NSDD");
-        *sem_addr -= value;
+        if (scope_ == SemScope::DM_LOCAL_CACHED) {
+#if defined(ARCH_QUASAR) && !defined(COMPILE_FOR_TRISC)
+            auto* word = reinterpret_cast<uint32_t*>(l1_offset_);  // cached alias
+            uint32_t observed = __atomic_load_n(word, __ATOMIC_RELAXED);
```
abhullar-tt, line 170: we use atomic load here but for some of the other semaphore apis (wait/wait_min/value()) we do plain loads via `local_l1_ptr`. I think we should we use atomic load there for cached scopes as well

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
```diff
@@ -183,10 +183,33 @@
...
+#define MEM_NOC_CAS_RET_SIZE 64
+
+// Per-semaphore EXTERNAL down() lock words, one per 16B row so the NoC-CAS always addresses
+// the first 4-byte word. Grow with NUM_SEMAPHORES.
+#define MEM_NOC_SEM_LOCK_BASE (MEM_NOC_CAS_RET_BASE + MEM_NOC_CAS_RET_SIZE)
+#define MEM_NOC_SEM_LOCK_SIZE 256  // NUM_SEMAPHORES * 16 (one 16B row per lock)
+
+// Dedicated cached-only pool for DM_LOCAL_CACHED semaphores: whole 64B cache lines that
+// nothing on the NoC/uncached path ever writes, so a cached AMO's line write-back can never
+// clobber NoC-written data. 8B rows indexed by semaphore id: [0] = counter, [1] = the seed
```
abhullar-tt, line 198: can you also add this explanation for why we need this dedicated pool so cached AMOs don't clobber noc-written data on the sem scope enum? or just refer to this comment

abhullar-tt (commented):
does `MEM_NOC_SEM_LOCK_BASE` need to be cleared/reset after kernel exits?

tt_metal/impl/metal2_host_api/semaphore_scope.hpp
```diff
@@ -0,0 +1,172 @@
...
+#include <tt_stl/assert.hpp>
+#include <tt-metalium/experimental/metal2_host_api/program_spec.hpp>
+#include "impl/context/metal_context.hpp"
+#include "jit_build/jit_build_settings.hpp"
+
+// ============================================================================
+// Semaphore mechanism solver
+// ============================================================================
+//
+// Decides how each semaphore is accessed. The answer is a SemScope, which codegen bakes
```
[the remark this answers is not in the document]
ymoussaTT: fixed

tt_metal/impl/metal2_host_api/semaphore_scope.hpp
```diff
@@ -0,0 +1,172 @@
...
+//
+// Decides how each semaphore is accessed. The answer is a SemScope, which codegen bakes 
+// into every binding's token, so the kernel receives it as a compile-time property.
+//
+// Two steps:
+//   1. CollectSemaphoreBinders() takes the census: which kernel instances bind each semaphore,
+//      how many hart instances that adds up to, and which nodes they cover.
+//   2. ResolveSemaphoreScopes() reads the census and picks, per semaphore, the fastest access
+//      path that keeps its operations atomic.
+// Callers need the name -> scope map, plus the binder hart count that the cached-pool seed
```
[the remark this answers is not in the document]
ymoussaTT: fixed

bbradelTT (approved):

tt_metal/hw/firmware/src/tt-2xx/dmk.cc
ymoussaTT: Done

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: After Audrey's requested changes sem_scope_of no longer exists. The binding token carries the scope, so there's no global lookup any more.

SemScope moved to api/dataflow/semaphore_binding_token.h, alongside the token, which mirrors api/tensor/tensor_binding_token.h. Let me know if you'd still rather it live somewhere else.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: Your right. I'll reset it back to the default after the release CAS at the end of down(), we need it pointing at the slot up to that point (that's how we confirm the subtract landed), but not after.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: I initially added that spin while I was investigating the barrier's behaviour, it can be removed, and so can the polls in consumed_cas4.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: It's emule's own flag, the emule runner defines it for every kernel it builds. I use it in two places to exclude the multi-consumer NoC-CAS lock, because emule doesn't provide noc_fast_atomic_cas4; under emule down() takes the single-consumer path in the #elif instead. Arch guards can't cover it since emule builds are ARCH_QUASAR too.

tt_metal/hw/inc/api/dataflow/noc_semaphore.h
ymoussaTT: Change made

tt_metal/hw/inc/internal/tt-2xx/quasar/dev_mem_map.h
ymoussaTT: Done, I went with the referral rather than duplicating the explanation.

ymoussaTT:
> does `MEM_NOC_SEM_LOCK_BASE` need to be cleared/reset after kernel exits?

No, every acquire falls through to the release CAS before returning, so the lock self-restores, and dm.cc zeroes the region once at boot.

mvasilijevicTT:
Could you please run the L2 nightly experimental Blackhole CI to verify the KDA changes?

Dispatch Nightly tt-metal L2 tests with Blackhole enabled and additional_test_categories=experimental
These are relevant jobs:
  - ttnn nightly experimental tests (blackhole) [bh_p100]
  - ttnn nightly experimental tests (blackhole) [bh_p150b_civ2]
  - ttnn nightly experimental tests (blackhole - viommu) [bh_p150b_civ2_viommu] — covers the KDA performance tests

ymoussaTT:
> Could you please run the L2 nightly experimental Blackhole CI to verify the KDA changes?
> 
> Dispatch Nightly tt-metal L2 tests with Blackhole enabled and additional_test_categories=experimental These are relevant jobs:
> 
> * ttnn nightly experimental tests (blackhole) [bh_p100]
> * ttnn nightly experimental tests (blackhole) [bh_p150b_civ2]
> * ttnn nightly experimental tests (blackhole - viommu) [bh_p150b_civ2_viommu] — covers the KDA performance tests

I just ran the tests, and it passed.

FYI: https://github.com/tenstorrent/tt-metal/actions/runs/33400066706

abhullar-tt (approved):
lgtm, @sidnadTT I suggest you take a look as well

mvasilijevicTT (approved):

tt_metal/hw/inc/api/dataflow/semaphore_binding_token.h
```diff
@@ -0,0 +1,47 @@
...
+ *  - LOCAL_NONATOMIC: Stored in L1 and accessed by read-modify-write. Picked only when at most
+ *                     one binder instance exists.
+ *  - DM_LOCAL_CACHED: Stored in a dedicated L1 pool and accessed through the DM cache via
+ *                     RISC-V AMO. Picked only when all binders are DMs on the same node where
+ *                     the semaphore exists. The pool is separate so a cached AMO's cache-line
+ *                     write-back cannot clobber NoC-written data: see MEM_DM_CACHED_SEM_BASE
+ *                     in dev_mem_map.h.
+ *  - EXTERNAL:        Stored in L1 and accessed through atomic operations via the NOC. Picked
+ *                     whenever the semaphore is reachable beyond a single node. An EXTERNAL
+ *                     semaphore's value can never be 0xFFFFFFFF, it would look like a NoC
```
akerteszTT: My Claude says:

> The 0xFFFFFFFF restriction is now definitively stale, and it's documented in the new API header. The sentinel poll it protected is gone from down() — I grepped, there's no 0xFFFFFFFF comparison left in the file — but semaphore_binding_token.h still states it as a property of EXTERNAL, and a test still claims to prove a bounded poll that no longer exists. 

I doubt we'll miss that 0xFFFFFFFF, but if you don't need it anymore, it's unnecessary complexity in the API surfact.

tt_metal/hw/inc/api/dataflow/semaphore_binding_token.h
```diff
@@ -0,0 +1,47 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+
+/**
+ * @brief Physical path a semaphore's accesses take. Picked by the host at build time from the
```
akerteszTT: Would be helpful to make clear that the choice is automatic, not user-driven. This sounds quite ambiguous. Also, if this is 2xx specific, would be good to mention.

akerteszTT (approved):
Thanks for the Metal 2.0 integration changes! This looks much better now.

Please take at minor nits before merging.

rtawfik01 (approved):

tt_metal/hw/inc/api/dataflow/semaphore_binding_token.h
ymoussaTT: Done

tt_metal/hw/inc/api/dataflow/semaphore_binding_token.h
ymoussaTT: Removed

PR #53696, opened 2026-08-19
Runtime: Run TensixTestRandomizedProgram in package-and-release

msudumbrekar-TT:
### PR Category
Test Only

### Summary
Today the only on-device coverage in package-and-release is the tier-1 model demos (models-owned) and the docker image smoke test, which is an `import ttnn` plus a single eltwise add on N150/N300. There is no runtime-owned test and nothing that exercises dispatch directly, so a dispatch or compile regression can ship without the pipeline noticing.

This adds a runtime sanity leg running `UnitMeshCQProgramFixture.TensixTestRandomizedProgram`: 100 randomized workloads with brisc + ncrisc + trisc kernels, so it covers compute rather than just data movement. This improves Runtime coverage and to check its state. 

### Notes for reviewers
- This needs its own entry in `runtime_sanity_tests.yaml`. The existing `runtime_sanity_dispatch` filter is `*RandomProgram*`, which does not match `TensixTestRandomizedProgram`.
- Uses the single-CQ fixture so the leg behaves the same on WH and BH. The multi-CQ fixture of the same name needs `TT_METAL_GTEST_NUM_HW_CQS=2` and skips on Blackhole.
- SKUs are `wh_n150` / `bh_p150`, which are the release SKUs, not the CI `_civ2` ones. The nightly runtime sanity and post-commit sanity matrices are unchanged; I confirmed this by running `prepare_test_matrix.py` for both SKU sets.
- `TT_METAL_SEED=0` is set explicitly. It is already the default, but pinning it keeps the release gate deterministic if that default ever changes.
- 10 min timeout, with matching `wh_n150` / `bh_p150` budgets added under `runtime -> sanity` in `time_budget.yaml` (`verify_time_budget.py` hard-fails on an unbudgeted SKU).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:msudumbrekar/release-runtime-sanity-dispatch)

### DIFF
--- a/.github/workflows/release-build-test-publish.yaml
+++ b/.github/workflows/release-build-test-publish.yaml
@@ -123,6 +123,18 @@ jobs:
       tier: "1"
       tests-yaml-gist-url: ${{ inputs.tests-yaml-gist-url }}
 
+  # Release SKUs (wh_n150, bh_p150), not CI _civ2. Gated like release-demo-tests.
+  runtime-sanity-tests:
+    needs: [build-artifact]
+    if: ${{ inputs.tag-version != '' }}
+    secrets: inherit
+    uses: ./.github/workflows/runtime-sanity-tests-impl.yaml
+    with:
+      docker-image: ${{ needs.build-artifact.outputs.dev-docker-image }}
+      build-artifact-name: ${{ needs.build-artifact.outputs.build-artifact-name }}
+      wheel-artifact-name: ${{ needs.build-artifact.outputs.wheel-artifact-name }}
+      enabled-skus: wh_n150,bh_p150
+
   # Run-level aggregate of per-job AI summaries from release-demo-tests-impl.
   # expected-jobs is the union of single-host and exabox multihost legs so a
   # missing ai_job_summary artifact is reconciled as INFRA_FAILURE.
--- a/.github/workflows/release-build-test-publish.yaml
+++ b/.github/workflows/release-build-test-publish.yaml
@@ -123,10 +123,12 @@ jobs:
       tier: "1"
       tests-yaml-gist-url: ${{ inputs.tests-yaml-gist-url }}
 
-  # Release SKUs (wh_n150, bh_p150), not CI _civ2. Gated like release-demo-tests.
   runtime-sanity-tests:
     needs: [build-artifact]
     if: ${{ inputs.tag-version != '' }}
+    permissions:
+      contents: read
+      packages: read
     secrets: inherit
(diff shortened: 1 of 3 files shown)


roseli-TT (approved):

roseli-TT:
FYI @dpopovTT

.github/workflows/release-build-test-publish.yaml
```diff
@@ -123,6 +123,20 @@ jobs:
...
       tests-yaml-gist-url: ${{ inputs.tests-yaml-gist-url }}
 
+  runtime-sanity-tests:
+    needs: [build-artifact]
+    if: ${{ inputs.tag-version != '' }}
+    permissions:
+      contents: read
+      packages: read
+    secrets: inherit
+    uses: ./.github/workflows/runtime-sanity-tests-impl.yaml
```
roseli-TT, line 133: hey @msudumbrekar-TT I just realized: does this mean tests are completely duplicated between runtime sanity and PnR now?

msudumbrekar-TT: No not duplicated. the impl filters by enabled-skus, and the sets are disjoint. Nightly sanity (wh_n150_civ2/bh_p150b_civ2/bh_loudbox) resolves to the same 7 legs as before and PnR passes wh_n150,bh_p150, which only the new entry declares, so it runs just that one.

dpopovTT (approved):

PR #53720, opened 2026-08-19
fix(eltwise): implement numerically stable logaddexp and logaddexp2

### LINKED ISSUE #52037: [Bounty $1500] logaddexp / logaddexp2: overflow-safe reformulation
## Summary

`ttnn.logaddexp` and `ttnn.logaddexp2` are composed as `log(exp(a) + exp(b))` literally, so they
return `±inf` for arguments whose exact result is an ordinary finite number. The whole point of a
log-sum-exp primitive is that it does not overflow; this one overflows at `|a| > 88.7`, which is
inside the normal working range of any log-domain computation.

Both directions fail: `logaddexp(100, 0)` returns `inf` where the exact answer is `100`, and
`logaddexp(-100, -100)` returns `-inf` where the exact answer is `-99.31`.

There is no representable-range argument available here, because the result is boxed in by its
own inputs:

```
max(a, b)  <=  logaddexp(a, b)  <=  max(a, b) + ln 2
```

If both arguments are finite the answer is finite and within 0.694 of one of them, so an `inf`
return is never a range limitation — it is always the intermediate.

## Current state measured on ttsim

ttsim 1.9.7, `Arch.BLACKHOLE`, `ttnn.logaddexp` called directly:

| a | b | chip | exact | |
|---|---|---|---|---|
| 100 | 0 | `inf` | 100.000000 | ✗ |
| 89 | 0 | `inf` | 89.000000 | ✗ |
| 90 | 89 | `inf` | 90.313263 | ✗ |
| 100 | 100 | `inf` | 100.693146 | ✗ |
| 200 | 199 | `inf` | 200.313263 | ✗ |
| −100 | −100 | `−inf` | −99.306854 | ✗ |
| 88 | 88 | 88.693146 | 88.693146 | ok |
| 5 | 3 | 5.126928 | 5.126928 | ok |

The threshold is sharp, and it is the `exp` overflow point rather than anything about the pair:

```
a =  88.70, b = 0   ->  88.699997   exact  88.699997    ok
a =  89.00, b = 0   ->  inf         exact  89.000000    fails
```

`bfloat16` behaves identically — `log(FLT_MAX)` is 88.72 and `log(BF16_MAX)` is 88.72, so the
narrower dtype changes nothing.

`logaddexp2` fails the same way at its own threshold, since it is built on `exp2` rather than
`exp`. Last good and first failing argument for `a = b`, both dtypes:

| op | positive | negative |
|---|---|---|
| `logaddexp` | 88 ok, 90 `inf` | −86 ok, −88 `−inf` |
| `logaddexp2` | 126 ok, 128 `inf` | −126 ok, −128 `−inf` |

`logaddexp2(1e5, 1e5)` returns `inf`; the exact answer is `100001`, which is an ordinary fp32
number.

## Root cause

The composition is a fixed `unary-pre / binary / unary-post` pipeline, and it names the two
overflowing steps explicitly:

https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/binary_ng/device/binary_ng_utils.cpp#L250-L263

```cpp
// log( exp(a) + exp(b) )
case BinaryOpType::LOGADDEXP:
    process_lhs  = unary::UnaryOpType::EXP;
    process_rhs  = unary::UnaryOpType::EXP;
    binary_op    = EnumT::ADD;
    postprocess  = unary::UnaryOpType::LOG;
```

`exp(a)` is evaluated before anything else can bound it, so it saturates and `log(inf)` is `inf`.
The underflow case is the mirror: both `exp` terms flush to zero and `log(0)` is `−inf`.

The same shape is repeated in two more places, so a fix has to cover all three:

- `ttnn/cpp/ttnn/operations/eltwise/binary/common/binary_op_utils.cpp#L130-L138` — FPU path,
  `EXP` as `PRE_IN0_0` / `PRE_IN1_0` around `add_tiles`, `LOG` as postprocess.
- `ttnn/cpp/ttnn/operations/eltwise/binary/common/binary_op_utils.cpp#L401-L410` — SFPU path,
  same defines around `add_binary_tile`.
- `LOGADDEXP2` is the identical pattern with `EXP2` / `LOG2` in each of the three.

This is host-side composition rather than kernel code, so it is not architecture-specific:
Wormhole, Blackhole and the Quasar copy all build the same pipeline.

## Direction

The standard identity keeps everything in range:

```
logaddexp(a, b) = max(a, b) + log1p(exp(-|a - b|))
```

`|a − b|` is non-negative, so `exp` of its negation is in `(0, 1]` and cannot overflow; the
result inherits its magnitude from `max(a, b)` rather than from `exp`. The primitives are already
in the tree — `ttnn.log1p` and `ttnn.maximum` both exist — but this does not fit the current
`pre/binary/post` slot shape, so whether it lands as a dedicated fused SFPU binary tile op
(alongside `lcm_tile`, `mul_binary_tile`) or as a restructured composition is a design and
performance call I am deliberately leaving to you.

For reference, the identity evaluated in fp32 gives the exact answer on every case above,
including `logaddexp(1000, 999) = 1000.313232`.

## Success criteria

1. `logaddexp(a, b)` and `logaddexp2(a, b)` are finite whenever the exact result is finite, for
   `|a|, |b| <= 1e4`, in both `float32` and `bfloat16`.
2. Relative error against `torch.logaddexp` stays within the op's existing tolerance over that
   range — the identity above is exact to fp32 rounding, so this should not cost accuracy.
3. Values in the currently-working range (`|a|, |b| < 88.7`) are unchanged within tolerance.
4. All three composition sites and the `LOGADDEXP2` variants are covered, with a sweep entry that
   samples beyond `±88.7` — the existing coverage never does, which is why this has stayed open.

## Why this matters

Log-sum-exp exists precisely to be the overflow-safe way to add in the log domain. A version that
overflows at 88.7 cannot be used for the thing it is for: numerically-stable softmax denominators,
log-likelihood accumulation, mixture-of-experts routing. The failure is silent — `inf` propagates
through the following ops and surfaces as `NaN` somewhere much later.

Worth separating from #26993, which is closed and lists `logaddexp` under the `exp` re-evaluation:
that work was about ULP accuracy of `exp` itself and did not change the composition. No amount of
precision in `exp` prevents `exp(100)` from overflowing.

rafaio1:
Fixes #52037

## Problem
The previous implementation of `ttnn.logaddexp` and `ttnn.logaddexp2` was composed as `log(exp(a) + exp(b))` and `log2(2^a + 2^b)`, which caused overflow to `inf` when inputs were large (`|a| > 88.7` for logaddexp).

## Solution
This PR implements the numerically stable identities as dedicated fused SFPU binary tile ops:
- `logaddexp(a, b) = max(a, b) + log1p(exp(-|a - b|))`
- `logaddexp2(a, b) = max(a, b) + log2(1 + 2^(-|a - b|))`

### Changes
- Added new fused SFPU binary tile ops for `LOGADDEXP` and `LOGADDEXP2`.
- Implemented `ckernel_sfpu_logaddexp.h` using the existing `log1p` SFPU op.
- Created `api/compute/logaddexp.h` and `api/compute/logaddexp2.h` wrappers.
- Registered the new ops in `binary_ng_utils` and `binary_op_utils`.
- Moved `LOGADDEXP`/`LOGADDEXP2` to the SFPU-true group in `binary_ng_device_operation`.
- Added the new headers to `sources.cmake` and `eltwise_binary_sfpu.cpp`.

Closes #52037

### DIFF
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_logaddexp.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_logaddexp.h
@@ -1,0 +1,86 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "ckernel.h"
+#include "ckernel_defs.h"
+#include "ckernel_sfpu_log1p.h"
+#include "sfpi.h"
+
+namespace ckernel {
+namespace sfpu {
+
+template <bool APPROXIMATION_MODE, bool is_fp32_dest_acc_en>
+sfpi_inline sfpi::vFloat _sfpu_logaddexp_(sfpi::vFloat a, sfpi::vFloat b) {
+    // Numerically stable logaddexp: max(a,b) + log1p(exp(-|a-b|))
+    auto [min_val, max_val] = sfpi::min_max(a, b);
+    sfpi::vFloat diff = max_val - min_val;
+
+    // Compute exp(-diff). Since diff >= 0, this is always in (0, 1].
+    sfpi::vFloat neg_diff = -diff;
+    sfpi::vFloat exp_neg_diff = sfpi::exp(neg_diff);
+
+    // log1p(exp(-diff)) using the existing log1p implementation
+    sfpi::vFloat log1p_result = calculate_log1p_fp32<is_fp32_dest_acc_en>(exp_neg_diff);
+
+    return max_val + log1p_result;
+}
+
+template <bool APPROXIMATION_MODE, bool is_fp32_dest_acc_en>
+sfpi_inline sfpi::vFloat _sfpu_logaddexp2_(sfpi::vFloat a, sfpi::vFloat b) {
+    // log2(2^a + 2^b) = logaddexp(a*ln2, b*ln2) / ln2
+    // Stable form: max(a,b) + log2(1 + 2^(-|a-b|))
+    //            = max(a,b) + log1p(2^(-|a-b|)) / ln2
+    auto [min_val, max_val] = sfpi::min_max(a, b);
+    sfpi::vFloat diff = max_val - min_val;
+
+    // 2^(-diff) = exp(-diff * ln2)
+    constexpr float LN2 = 0.6931471805599453f;
+    sfpi::vFloat neg_diff_ln2 = -diff * LN2;
+    sfpi::vFloat exp_neg_diff_ln2 = sfpi::exp(neg_diff_ln2);
+
+    sfpi::vFloat log1p_result = calculate_log1p_fp32<is_fp32_dest_acc_en>(exp_neg_diff_ln2);
+
+    // Divide by ln2 to convert from natural log to log2
+    constexpr float INV_LN2 = 1.4426950408889634f;
+    return max_val + log1p_result * INV_LN2;
+}
+
+template <bool APPROXIMATION_MODE, int ITERATIONS, bool is_fp32_dest_acc_en>
+inline void calculate_sfpu_logaddexp(const uint dst_index_in0, const uint dst_index_in1, const uint dst_index_out) {
+    constexpr uint dst_tile_size_sfpi = 32;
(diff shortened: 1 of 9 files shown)


ncvetkovicTT:
Assigned to another user as the issue says. @jberkowitzTT FYI.

PR #53721, opened 2026-08-19
Record CB residency at dispatch and fix cross-process SHM tracking

### LINKED ISSUE #49304: Cross-process SHM races in memory_tracking (memory_stats_shm)
- `tt_metal/impl/memory_tracking/memory_stats_shm.cpp`; header fields at `memory_stats_shm.hpp:40-49`.

The shared-memory header fields are plain (non-atomic) and init/teardown have no process-shared synchronization:
- **Init TOCTOU / `version` data race:** creator does `shm_open(O_CREAT|O_EXCL)` (`:50`) then *separately* `ftruncate` (`:65`) + `initialize_region()` (`:86`, plain store of `version`). A second process taking the `EEXIST` branch reads `region_->version` (`:87`, plain, unsynchronized) concurrently with the creator's plain store → torn/garbage read; it may then reinitialize a live region or `munmap` it (`:106-110`), disabling tracking for everyone. The `reference_count` acquire-load (`:88`) establishes no happens-before with the plain `version` write.
- **`num_active_processes` lost updates:** plain `uint32_t` mutated by `++` (`:541`) and `--` (`:191`, in the destructor, no lock) across processes → lost increments/decrements (UB data race).
- **Destructor underflow/leak:** `if (total >= x)` compare (`:154`) and `fetch_sub` (`:155`) are separate ops with no lock — another process decrementing in between makes the subtraction underflow-wrap (defeating the clamp). The load-totals → subtract → clear sequence is also non-atomic as a unit, so a concurrent same-PID `record_allocation` (`fetch_add` at `:362`) leaks its delta. (`record_deallocation`'s `safe_sub` CAS loop is the correct pattern to follow.)

**Suggested fix:** Make the SHM header fields atomic; serialize creation/initialization with a named/`pthread_mutexattr_setpshared` process-shared mutex (or a robust init-once protocol with a state flag and proper acquire/release ordering); replace the destructor's check-then-subtract with the `safe_sub` CAS pattern already used in `record_deallocation`.

### LINKED ISSUE #52010: Optimize memory tracking code
### Is your feature request related to a problem? Please describe

For models with a large number of kernels running on large mesh devices memory tracking can become very expensive.

For example, I have a model with ~200 kernel calls running on a 4x8 mesh and I'm measuring approximately  750ms per kernel call during the first model run. For this specific case I can disable tracking with `TT_METAL_SHM_TRACKING_DISABLED=1` but this code should be optimized.

### Describe the solution you'd like

_No response_

### Describe alternatives you've considered

- Run tracking on a background thread so it doesn't block model initialization
- Only track changes in CB allocation so it doesn't get recomputed from scratch each time

### Additional context

_No response_

aperezvicente-TT:
### PR Category

  Bug fix (with a performance component).
 
  ### Summary

  Circular buffer tracking recomputed the device-wide footprint from scratch on every program
  registration, walking every live program. With ttnn's program cache holding one entry per
  parameter variant that is O(N) per call and O(N^2) across a warmup sweep, paid per sub-device.

  It also measured the wrong quantity. A registered program occupies no CB space until it is
  dispatched, so the figure was a high-water mark over the program cache rather than what is
  resident.

  CB residency is now recorded at dispatch, per core, mirroring the config the dispatcher writes
  into L1: the enqueue loop for fast dispatch, `LaunchProgram` plus
  `DispatchCompiledProgramToDevice` for slow dispatch, and for traces a per-core peak computed
  once at capture and re-applied on each replay. The per-program footprint is cached on
  `ProgramImpl`, so re-dispatching the resident program is a pointer comparison.

  SHM region v4 fixes the cross-process handling. `init_state` publishes readiness through a CAS
  claim, so an attaching process cannot observe a zero-filled region whose `chip_stats` read
  `chip_id` 0 — a valid chip ID, which defeated the `CHIP_STATS_UNUSED` sentinel. A process
  claims its slot at attach. Aggregates and `reference_count` are derived from live slots, so a
  SIGKILLed process is reclaimed by the next attacher instead of pinning the count above zero
  forever. The destructor can no longer underflow. Field offsets up to the end of `processes` are
  unchanged and `sizeof` stays 8576, verified by compiling this header against an independent v3
  mirror and asserting every offset. 

  The SHM processor is registered as a process-wide background processor in `GraphTracker` rather
  than on the calling thread's capture stack, so buffers allocated on any thread are tracked.

  Measured on 4x Blackhole p150a, 110/120-core grids:
  
  - CB query cost across 1 to 256 live programs: 145x before (0.025 ms to 3.618 ms), 1.0x now —
    it is an atomic load 
  - Paced registration with 4096 live programs: 67.9 ms each before, 0.007 ms now, which is what
    tracking-disabled also costs 
  - Host dispatch overhead, tracking on vs off: 8.91 vs 8.67 us
  - CB total against the device's own CB config table: 8x over-report before, exact now under fast
    dispatch, slow dispatch on 4 devices, and trace replay
  - Qwen3.6-27B decode, CB column: 1.1 MiB before (the trace's last op), ~77 MiB now (its peak)
  
  `unit_tests_api`: 2001 tests, 0 failures. Qwen3.6-27B `traced_128` on 4xP150 passes,
  `ttft=0.49s decode=25.81 tok/s`.

  Closes #52010
  Closes #49304
  
  ### Notes for reviewers
  
  Start with what the CB column now means: bytes of L1 the dispatcher has configured as circular
  buffers, not a peak over cached programs. Two consequences. Cores a program does not cover keep
  their previous value, which is correct — their L1 config was not overwritten. And for a trace it
  is the per-core maximum over the trace's programs, which is deliberate: the alternative is
  whichever program was captured last, arbitrary for an inference trace, and it under-reported a
  two-program trace by 3.5x. That also means the figure intentionally exceeds the CB config table
  in L1 after a multi-program replay, since the table only holds the last program.
  
  Consumer coordination is required before this lands. The layout is offset-compatible, but
  external readers version-check: `tt-mgmt` is updated to accept v3-v4 (separate change, pushed);
  `nvtop` is written but not yet committed; `tt-telemetry` is still pinned to v3 and will report
  "0 used" until bumped. 
 
  The footprint cache is unsynchronised. `ProgramImpl::cb_bytes_per_core()` is lazily filled from
  the dispatch path. Dispatch already mutates the program through
  `update_program_dispatch_commands(ProgramImpl&)`, so one program object cannot be dispatched
  from two threads concurrently; the cache inherits that contract rather than putting a lock on
  the path this change exists to speed up. If you read that contract differently, this is the
  place to push back.
  
  Reclaiming a region of an older layout is decided by whether an owning process is still alive,
  not by `reference_count`. That count is exactly what an older writer leaks when killed, so
  trusting it made the upgrade one-way: reproduced, one SIGKILLed v3 run and every later process
  silently disabled tracking for that chip with no recovery short of deleting the file. The check
  reads `processes`, gated on the version being one whose layout is known (v2-v4). This path is
  deliberately not covered by a test — the two guards for it were dropped as transitional.

  Known gaps, documented in comments rather than fixed:

  - concurrent dispatch to one device can publish out of order, leaving a stale-but-final value
    until the next change
  - `track_allocate_dataflow_buffer` and `track_allocate_scratchpad` notify per-thread processors
    only, so a background processor cannot observe program-scope L1. No-op today, since the SHM
    processor implements only the buffer hooks
  - no test covers a globally-allocated CB, which is excluded from the CB column by design: it is
    a real L1 Buffer and is counted in the L1 column
  - the L1-readback test helper is a valid oracle only for a single dispatch, because
    `get_cb_base_addr` returns the static `KERNEL_CONFIG` base rather than the rotating slot the
    dispatcher wrote
  - the DRAM/L1 columns record logical buffer bytes while the allocator charges bank-rounded, so
    they differ by up to (alignment - 1) x banks per buffer, about 4 KB observed. Pre-existing and
    unchanged here

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:aperezvicente/shm-memory-tracking-rework)


tests/tt_metal/tt_metal/api/circular_buffer/test_CircularBuffer_tracking_perf.cpp
```diff
@@ -0,0 +1,1078 @@
...
+    ASSERT_GT(baseline_ms, 0.0) << "baseline query was unmeasurably fast even over " << kQueriesPerSample
+                                << " queries; raise kQueriesPerSample";
+
+    // The tracked total is incrementally maintained, so a query is an atomic load and its
+    // cost does not depend on how many programs are alive. Recomputing it per call instead
+    // makes this ratio track the program count: 145x at 256 live programs, measured on a
+    // p150a against the code this replaced (115x on an n300).
+    const double growth = final_ms / baseline_ms;
+    log_info(tt::LogTest, "tracking query cost grew {:.1f}x between 1 and {} live programs", growth, max_programs);
+    EXPECT_LT(growth, 10.0) << "Device::get_total_cb_allocated() cost scales with the number of live "
```
sosborne-TT, line 205: This checks growth -- are there any tests that check the absolute time?

sosborne-TT (approved):

sosborne-TT:
@aperezvicente-TT any update on when this PR will be merged?

riverwuTT (changes requested):
Please provide a more descriptive PR title and motivation.

aperezvicente-TT:
Two independent defects in device memory tracking, both user-visible.

  **The CB figure was expensive to produce.** `get_total_cb_allocated()` recomputed the
  device-wide footprint from scratch on every program registration, walking every live
  program and expanding per core. ttnn's program cache holds one entry per parameter
  variant and programs only leave that set in `~ProgramImpl`, so each call is O(live
  programs) and a warmup sweep is O(N^2)  

  **And it was the wrong number.** A registered program occupies no CB space until it is
  dispatched, so the reported figure was a high-water mark over the program cache rather
  than what is actually resident in L1. 
  
  **Cross-process state could not recover.** An attaching process could observe a
  zero-filled region, where `chip_id` reads 0 — a valid chip ID, defeating the
  `CHIP_STATS_UNUSED` sentinel. Worse, a SIGKILLed process left its allocations in the
  totals permanently and pinned `reference_count` above zero, so the region could never
  reset: one crashed run poisoned every later run's figures until someone manually removed
  `/dev/shm/tt_device_*_memory`.

akerteszTT (dismissed):
I cannot understand the overriding motivation of this PR, or the reasoning behind the design choices that were made. Both the PR description and response to previous inquiries appear to be purely AI-generated.  You're making a cross-stack change -- this requires a high-level explanation.

My main concern with the PR overall is that you are making changes (that I cannot understand the purpose of) to the Metalium public API surface. TTNN graph tracking is a dubious component of this API surface that exists purely to serve TTNN internals -- it's something that will need refactor in the future.

Your PR appears to add a bunch of new stuff there, but _also_ adds a brand new graph-tracking file to the `api/internal`. Why??

akerteszTT:
This PR additionally has **multiple** red-blocking copilot comments. 

You should never submit a PR for codeowner review without first addressing these, as they usually flag serious structural issues with your changes.

aperezvicente-TT:
Yes previous PR and response are AI generated. Let me give you my written motivation. The SHM region reports device memory to telemetry workload monitoring tools. To problems where flagged by two issues. CB usage was recomputed from scratch on every program registration by walking every live program O(N²) across a warmup sweep with ttnn's program cache, and it measured registered programs, not dispatched ones, so it wasn't what's actually resident. 
Separately, a SIGKILLed process left  its allocations in the shared region forever and pinned reference_count above zero, so the region could never reset. One crashed run poisoned every later run until you manually rm'd /dev/shm/tt_device_*.

On the API surface: The tracker needs to see buffer allocs, GraphTracker was already
doing that, so I hooked in. processors is thread_local and missed other threads, so I added background_processors. Then Copilot lagged those as public API and I moved the entry point to api/internal which fixed the file placement . That's the file you're asking about?

Regarding the graph tracking file under api, youre right and if i move the observer to impl/memory_tracking/ (called straight from buffer.cpp) . Then i can remove this file along with the background_processors in the public header

akerteszTT:
Thanks for the update! My block is removed -- this PR no longer touches any of the API I own. (And I'll be out on vacation after today, didn't want to leave it in place.)

That said, I still think there are quality issues with this PR as a whole.

tests/tt_metal/tt_metal/api/circular_buffer/test_CircularBuffer_tracking_perf.cpp
riverwuTT, line 1: You seems to be trying to establish a benchmark.

Have you considered using tools like google benchmark?

riverwuTT: our CI system is notorious for unstable perf measurement, have you ran this test on our CI system yet?

tests/tt_metal/tt_metal/api/circular_buffer/test_CircularBuffer_tracking_perf.cpp
```diff
@@ -0,0 +1,1297 @@
...
+// Distinct from the query-cost test above because it is paid per registration rather than
+// per query. Against the code this replaced it also understates the damage: that version
+// admitted at most one query per 100 ms per device, so a tight loop like this one diluted the
+// cost that a compile-paced warmup pays in full. The paced test below is the honest one.
+//
+// Run twice to compare:
+//   unit_tests_api --gtest_filter='*CircularBufferRegistrationWallClock*' \
+//     --gtest_also_run_disabled_tests
+//   TT_METAL_SHM_TRACKING_DISABLED=1 unit_tests_api --gtest_filter=... (same)
+TEST_F(AnyDispatchMeshDeviceSingleCardFixture, DISABLED_TensixCircularBufferRegistrationWallClock) {
```
riverwuTT, line 224: why are we checking in a disabled test?

tt_metal/impl/device/device.cpp
```diff
@@ -1110,74 +1106,117 @@ HalMemType Device::get_mem_type_of_core(CoreCoord virtual_core) const {
...
-
-    // For PHYSICAL CB tracking accounting for address reuse:
-    // Collect L1 regions per core and merge overlapping addresses
-    // This handles cached/traced programs that share the same physical L1 addresses on the same core
-
-    std::map<CoreCoord, std::vector<std::pair<uint64_t, uint64_t>>> device_regions_per_core;
-
-    for (const auto* program : active_programs_) {
-        size_t num_devices = program->get_num_cb_devices();
+uint64_t Device::get_total_cb_allocated() const { return total_cb_resident_.load(std::memory_order_relaxed); }
```
riverwuTT, line 1109: Is there a reason to use memory order relaxed?

riverwuTT (approved):
Approve to dismiss my own block before as api related changes has been removed.

I would still highly recommend rewriting the pr description yourself as it is really hard to follow.

Please also make sure you've manually reviewed the code you're trying to check-in. 

I am not an expert on shm tracing, I've skimmed the implementation and tests but I think I see some big red flags, especially surrounding concurrency.

The big smell I see are:
1. mixed use between mutex and atomics
2. memory fence is relaxed, compare and swap uses weak, it's unclear what the desired semantics is here.
3. fork and mmap in tests
4. Related pipelines not run

tests/tt_metal/tt_metal/api/circular_buffer/test_CircularBuffer_tracking_perf.cpp
aperezvicente-TT: It's a local measurement, not a CI gate. That's how I got the registration and enqueue on-vs-off numbers. Enabling it on CI would flake, so it's DISABLED_ on purpose. Fine to drop if you'd rather those numbers live only in the PR description.

tt_metal/impl/device/device.cpp
aperezvicente-TT: Yes. It's just a stats counter. Writers already hold cb_residency_mutex_; readers only need the latest total, not a happens-before with anything else. relaxed is enough. acq_rel wouldn't fix concurrent dispatch publishing out of order either.

PR #50156, opened 2026-07-16
Feature: Trace allocation safety tracker

tchedaTT:
### PR Category
Feature

### Summary
Publish the tool previously used for https://github.com/tenstorrent/vllm/pull/352 and other debugging.

The tool is always built, but disabled until a flag is set. I ran microbenchmarks to validate no measurable performance impact when disabled.
These comparison scripts are in benchmark_trace_allocation_disabled.py and 
compare_trace_allocation_disabled.py, included for reference - will remove before merging.

When enabled, the tool starts tracking allocations made after a trace is captured. If an allocation is not freed before the next trace is executed, it may be corrupted, so the tool throws an error, providing a traceback to where the unsafe allocation happened, and where the buffer is held now.

### Notes for reviewers
 Please indicate what tests you'd like to see.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:tcheda/trace_allocation_tracker_release)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:tcheda/trace_allocation_tracker_release)

PR #53725, opened 2026-08-19
trace tracker: make trace registration idempotent and tolerant of imbalance

ayerofieiev-tt:
### Problem

#50156 replaced the idempotent `mark_allocations_safe/unsafe` with a non-idempotent register/unregister counter protocol, computed over whichever allocators are loaded *at call time*. Any lifecycle imbalance now crashes or permanently poisons allocator state — in the default, tracking-disabled configuration:

- `release_trace` of a trace that never completed `end_trace_capture` (exception mid-capture, e.g. the gemma4 demo's `finally: ttnn.release_trace`), or a plain double release, hits `TT_FATAL("Cannot unregister trace: no active traces are registered")` where pre-#50156 this was a benign no-op.
- Loading a sub-device manager between capture and release fatals on allocators that never saw the registration.
- `end_trace_capture` retried after a failure double-registers, latching `active_trace_count_ > 0` (and the allocations-unsafe warning state) for the life of the process.

Details with evidence in the [review comments on #50156](https://github.com/tenstorrent/tt-metal/pull/50156#issuecomment-5346615291).

### Fix

- Each allocator tracks registered trace ids as a **set** instead of a counter: registration and release are idempotent, and releasing a trace the allocator never saw is a no-op.
- Release/query/cleanup walk **all** sub-device managers' allocators (`SubDeviceManagerTracker::get_all_sub_device_managers`), not just default+active, so a manager switch between capture and release can neither fatal nor leak the registration. Registration still targets the allocators loaded at capture time.

### Also in this PR

- `buffer_unique_id()` returns `None` (as its docstring promises) instead of raising `RuntimeError` for a view whose root buffer was force-deallocated; `ttnn.mark_corruptible` now raises its intended `ValueError` in that case. Adds `DeviceStorage::is_root_allocated()`.
- The tracking-enabled `execute_trace` wrapper now uses the same `mesh_device` keyword as the nanobind binding, so keyword callers behave identically with tracking on and off.
- `TraceCorrestness.md` → `TraceCorrectness.md` (filename + CODEOWNERS).

### Tests

- C++ (`test_sub_device.cpp`): lifecycle-imbalance test covering release-of-never-captured, double register, double release, and both directions of manager switch between capture and release. Runs in both tracking modes.
- Python: double-release regression (also exercises the keyword form of `execute_trace`); `buffer_unique_id`/`mark_corruptible` on a view with a force-deallocated root.

Not addressed here (proposed follow-up restructure discussed on #50156): consolidating the tracker into one component, observing deallocations instead of per-replay liveness rescans, single-source env parsing, thread-safety of the diagnostics state, and the Python API naming/location.


tchedaTT:
https://github.com/tenstorrent/tt-metal/pull/53735 is the way I'd like to address these issues

PR #53734, opened 2026-08-20
trace tracker: observe deallocations and collapse forwarding layers

ayerofieiev-tt:
### What this is

The restructure of the trace allocation tracker (#50156) discussed in [this comment](https://github.com/tenstorrent/tt-metal/pull/50156#issuecomment-5349409158). Stacked on #53725 (the crash fixes) — review that one first; this diff is the single commit `trace tracker: observe deallocations and collapse forwarding layers`.

Net **−21 lines** while adding a public header and rewriting a test: the machinery it deletes only existed because the tracker couldn't see deallocations and lived spread across five layers.

### Changes

**Observe deallocations** — `deallocate_buffer` gets the same `[[unlikely]]`-guarded hook `allocate_buffer` already has. This deletes:
- the per-replay rebuild of an id-set of ALL allocated buffers in `get_unsafe_tracked_ids` (now a plain map copy),
- the lazy-retirement loops there and in `unregister_active_trace`,
- `remove_unsafe_tracked_id` as separate logic (now delegates to the dealloc path).

Buffers freed via `clear()`/`override_state()`/`deallocate_buffers()` bypass the hook — exactly as they bypassed the old `allocated_buffers_`-based rescan, so this is at parity, not a regression.

**Thread-safety** — `pending/retired_traceback_ids` and the corruptible-scope stack were `thread_local`, written on the allocating thread but drained/pushed from the Python thread. They're now process-global and mutex-guarded: multi-CQ hosts no longer get false-positive replay rejections from scopes that don't cover worker-thread allocations, and traceback ids can't strand in another thread's TLS. The allocation-context stack stays `thread_local` — it is genuinely call-stack-scoped.

**Env config parsed once, in one place** — the three `TT_METAL_TRACE_ALLOC_*` flags move out of the `EnvVarID` switch (which was order-dependent: the TRACEBACKS handler read the TRACKING flag mid-parse) into self-contained lazily-initialized getters using `is_env_enabled` like their siblings. They work before any `RunTimeOptions` instance exists, which lets Python read them through new `ttnn._ttnn.operations.trace` bindings at import time. `trace_allocation_config.py` is deleted; `TRACE_ALLOC_TRACKING/DIAGNOSTICS` now come from the bindings (single source of truth), and only the Python-only referrer depth is still parsed in Python.

**Layer collapse** — the free-function API moves from the tt_metal-internal `tt_metal/distributed/trace_allocation_tracker.hpp` to `tt-metalium/experimental/trace_allocation_tracker.hpp` (precedent: `allocation_context.hpp`). ttnn no longer includes tt_metal internals; the eight pure-forwarding wrappers in `ttnn/cpp/ttnn/operations/trace.{hpp,cpp}` and the static forwarders on `AllocatorImpl`/`MeshDeviceImpl` are deleted — nanobind binds the experimental API directly. `trace_allocators()` (default+active) merges into `all_trace_allocators()`: registration now covers every loaded manager's allocators, so allocations made after a mid-trace manager switch are tracked instead of silently missed.

**Small stuff** — the `"program_cache:"` prefix contract, previously hardcoded in three files, becomes `kProgramCacheAllocationContextPrefix` in `allocation_context.hpp`; `verify_before_replay` runs `gc.collect()` only when tracked buffers are actually still alive (steady-state clean replays skip GC entirely).

### Perf gate

The disabled-path microbenchmark (`tests/ttnn/benchmark/python/benchmark_trace_allocation_disabled.py`, paired fresh-process protocol from #50156) must pass the ±2% equivalence criterion before this merges — the dealloc path gains a predicted-not-taken branch and `AllocatorImpl`'s field layout changes (counter → set in #53725; statics removed here). I don't have loudbox results yet; will run or hand off to whoever has the wh-lb setup live.

### Not in this PR

Python API naming/relocation (`mark_corruptible`, moving the tracker out of the ttnn root) — that's a user-facing rename that needs a name agreed with @tchedaTT first; proposal in the linked comment.


tchedaTT:
https://github.com/tenstorrent/tt-metal/pull/53735 is the way I'd like to address these issues

Issue #31017, opened 2025-10-22
Row major layout tilize to tile layout does not work with uneven shard


dchenTT:
### Describe the bug

Binary op does not support row_major layout natively, so it is tilized to tile layout first. For sharded tensors, it only works with height sharding. But when shard is uneven (last core does not have full shard), it hits run time error. "ShapeBase[] index out of range. 2 not in [-4, 2)". When shard is even, it works fine.

### Steps to reproduce the issue

1. check out main and run following test

### Expected behavior

No runtime error

### Please complete the following environment information
Test code

@pytest.mark.parametrize(
    "dtype_pt, dtype_tt",
    ([torch.bfloat16, ttnn.bfloat16],),
)
def test_binary_sharded_row_major_tilize(device, dtype_pt, dtype_tt):
    torch.manual_seed(0)
    a_shape = torch.Size([63, 32])

    a_sharded_config = ttnn.create_sharded_memory_config(
        [32, 32],
        core_grid=ttnn.CoreRangeSet({ttnn.CoreRange((0, 0), (0, 1))}),
        strategy=ttnn.ShardStrategy.HEIGHT,
        orientation=ttnn.ShardOrientation.ROW_MAJOR,
        use_height_and_width_as_shard_shape=True,
    )
    a_pt = gen_func_with_cast_tt(partial(torch_random, low=-50, high=50, dtype=dtype_pt), dtype_tt)(a_shape)
    a_tt = ttnn.from_torch(
            a_pt,
            dtype=dtype_tt,
            device=device,
            layout=ttnn.ROW_MAJOR_LAYOUT,
            memory_config=a_sharded_config,
        )

    a_tt = ttnn.to_layout(a_tt, ttnn.TILE_LAYOUT)

Following test will pass. Only the uneven shard is on H dim, it complains. If H and W are multiple of 32, and uneven shard is on higher dim, it works fine.

@pytest.mark.parametrize("scalar", [-0.25])
@pytest.mark.parametrize(
    "a_shape, shard_type, shard_size, core_range",
    ( 
        [
        # shape cannot be uneven shard
        torch.Size([5, 7, 32, 32]),
        ttnn.ShardStrategy.HEIGHT,
        [6 * 32, 32],
        ttnn.CoreRangeSet({ttnn.CoreRange((0, 0), (0, 5))}),
        ],
    )
)

[...]

PR #53741, opened 2026-08-20
Bug fix: Fix ttnn::to_layout crash on uneven height-sharded tilize

abdulla-TT:
### Ticket
#31017

### PR Category
Bug fix

### Summary
`ttnn.to_layout(tensor, ttnn.TILE_LAYOUT)` crashed with `ShapeBase[] index out of range. 2 not in [-4, 2)` when tilizing a height-sharded, row-major tensor whose height wasn't an exact multiple of the per-core shard height (i.e. the last core only holds a partial shard).

The culprit was the padding workaround for height-sharded tensors in `to_layout_op.cpp`: it hardcoded positive indices `[2]`/`[3]` into the shape, assuming the tensor was always rank 4. For any other rank (e.g. rank 2), those indices were simply out of bounds.

This fixes the padding to be rank-aware (using negative indexing so it works for rank >= 2), and also restores the tensor's original logical shape after the pad+tilize workaround, which previously left it at the tile-rounded-up shape instead of the true output shape.

### Notes for reviewers
Coverage for the uneven-shard case (ranks 2/3/4) was folded directly into the existing `test_tilize_height_sharded_shapes` test rather than adding a new one, and that test now goes through `ttnn.to_layout` (the actual path that had the bug) instead of calling `ttnn.tilize` directly.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/tilize_shard_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/tilize_shard_fix)


ttnn/cpp/ttnn/operations/core/to_layout/to_layout_op.cpp
```diff
@@ -182,14 +183,19 @@ Tensor to_layout_impl(
...
                 // ttnn::tilize_with_val_padding doesn't support height sharded tensors
                 // workaround by applying padding and then tilizing
-                ttsl::SmallVector<std::array<uint32_t, 2>> padding = {
-                    {0, 0},
-                    {0, 0},
-                    {0, padded_output_shape[2] - output_shape[2]},
-                    {0, padded_output_shape[3] - output_shape[3]}};
+                // ttnn::pad requires one {before, after} pair per dimension, so the padding vector
+                // must match the tensor's actual rank. Height/width are addressed via negative
+                // indices (the last two dims) so this is correct for any rank >= 2.
```
KalaivaniMCW: pls simplify this comment

ttnn/cpp/ttnn/operations/core/to_layout/to_layout_op.cpp
```diff
@@ -182,14 +183,17 @@ Tensor to_layout_impl(
             if (tensor.memory_config().memory_layout() == TensorMemoryLayout::HEIGHT_SHARDED) {
                 // ttnn::tilize_with_val_padding doesn't support height sharded tensors
                 // workaround by applying padding and then tilizing
-                ttsl::SmallVector<std::array<uint32_t, 2>> padding = {
-                    {0, 0},
-                    {0, 0},
-                    {0, padded_output_shape[2] - output_shape[2]},
-                    {0, padded_output_shape[3] - output_shape[3]}};
+                // Pad only the last two dims (height/width), sized to the tensor's rank.
+                int padding_rank = std::max<int>(padded_output_shape.rank(), 2);
```
Aswinmcw: The `std::max<int>(..., 2)` cannot actually protect a rank<2 tensor: `padded_output_shape[-2]` on line 189 would throw the same out-of-range error this PR is fixing. It is only unreachable because the `TT_FATAL(!tensor.is_sharded(), ...)` at line 109 rejects sharded rank<2 tensors before this branch. Using `padded_output_shape.rank()` directly would avoid implying a safety guarantee that is not there.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -822,22 +822,28 @@ def test_tilize_col_major_orientation(device, memory_layout, tensor_shape, shard
...
+        ([1, 1, 32, 64], 1, None),
+        ([1, 1, 64, 64], 2, None),
+        ([1, 1, 128, 64], 4, None),
+        ([1, 1, 256, 128], 8, None),
+        ([1, 1, 512, 64], 8, None),
+        ([1, 1, 96, 128], 3, None),
+        ([1, 1, 192, 256], 6, None),
+        # Uneven shard (last core does not have a full shard, i.e. total height is less
+        # than shard_h * num_cores) across ranks 2/3/4, via ttnn.to_layout's height-sharded
+        ([63, 32], 2, 32),
```
Aswinmcw, line 836: All three new shapes have W=32, so `padding[padding_rank - 1]` is always `{0, 0}` and the width half of the re-indexed padding is never exercised. Is a non-tile-aligned width supported on this path (e.g. `([63, 30], 2, 32)`)? If so, one such case would cover the other half of the change.

Aswinmcw (approved):

ttnn/cpp/ttnn/operations/core/to_layout/to_layout_op.cpp
abdulla-TT: Updated

bbradelTT (approved):

sjameelTT (approved):

PR #53752, opened 2026-08-20
Design: LLK perf regression PR gate

nstojicTT:
This PR is an attempt at designing perf regression gate for LLK team that would make development easier and more safe.
General idea is: 
PR opened with LLK changes
Perf workflow runs
Gate workflow triggers → compares against Snowflake baseline
If regressions found → hard block, PR cannot merge
If clean → PR can proceed to merge
After merge to main → sanity workflow runs and publishes new baseline to Snowflake

These 2 files are most important
.github/workflows/llk-perf-gate-impl.yaml
.github/workflows/llk-perf-gate.yaml

Other changes are mostly updating Claude skill and moving existing compare script.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/llk-perf-gate-proposal)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/llk-perf-gate-proposal)


.github/workflows/llk-perf-gate.yaml
fvranicTT, line 1: Can this be called from pr-gate.yaml? That way we ensure it runs on every PR. Or was the idea to start it on demand?

ndivnicTT: The idea is to start it on demand, to warm up and see how it works

nstojicTT: Eventually it will definitely will be rewired to run on every PR, for now I was thinking its better to roll it out slowly to see how it performs

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,38 @@
...
+# CURRENT STATE: Measures L1_TO_L1 (unpack+math+pack) by default, extensible
+# to individual isolate modes (MATH_ISOLATE, UNPACK_ISOLATE, PACK_ISOLATE).
+# Thresholds derived from 5-run noise baselines on Blackhole and Wormhole.
+#
+# BASELINE UPDATES: After merge to main, the sanity workflow re-runs perf tests
+# and publishes new baseline to Snowflake. Next PR's gate compares against it.
+#
+# Triggered on LLK PRs after compile and measure complete
+on:
+  workflow_run:
```
fvranicTT: Can we trigger by assigning a label to a PR, similar to how we ran them in tt-llk repo (see [here](https://github.com/tenstorrent/tt-llk/blob/main/.github/workflows/run-perf-tests.yml))?

nstojicTT: Tbh I wasn't aware this was a thing
So basically

on:
  pull_request:
    types: [labeled]
jobs:
  llk-perf-gate:
    if: github.event.label.name == 'run-perf-gate '
    uses: ./.github/workflows/llk-perf-gate-impl.yaml

instead of current

on:
  workflow_run:
    workflows: ["LLK perf"]
    types: [completed]

fvranicTT: Yes, in a nutshell. For the user, it's very easy to just add a label and wait for a new workflow to report the results.

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+permissions:
+  contents: read
+  checks: write
+  pull-requests: write
+
+jobs:
+  llk-perf-gate:
+    name: "LLK perf regression gate"
+    if: github.event.label.name == 'perf-gate'
```
fvranicTT:
```suggestion
    if: github.event.label.name == 'llk-perf-gate'
```
Can we be more specific?

nstojicTT: Done

tt_metal/tt-llk/perf/regression_compare.py
```diff
@@ -0,0 +1,385 @@
...
+    return _is_metric(col) or col.startswith("TEXT_SIZE(")
+
+
+def _point_key(row, config_cols):
+    """A point's identity within one test: marker + the sweep configuration."""
+    config = tuple(sorted((c, row[c]) for c in config_cols if pd.notna(row[c])))
+    return (row.get("marker"), config)
+
+
+def _medians(frames):
```
fvranicTT, line 68: This looks like a bottleneck. It might be better to use Panda's groupby for this:

```python
def _mean_cols(columns):
    return [c for c in columns if c.startswith("mean(")]


def _config_cols(columns):
    return [c for c in columns if c != "marker" and not _is_ignored(c)]


def _point_key(marker, config_pairs):
    """A point's identity within one test: marker + the non-null sweep config."""
    return (marker, tuple(sorted((c, v) for c, v in config_pairs if pd.notna(v))))


def _medians(frames):
    """{(point_key, mean_col): median value} across a list of run DataFrames."""
    frames = [f for f in frames if not f.empty]
    if not frames:
        return {}
    df = pd.concat(frames, ignore_index=True)
    if "marker" not in df.columns:
        df = df.copy()
        df["marker"] = pd.NA
    mean_cols = _mean_cols(df.columns)
    if not mean_cols:
        return {}
    # Drop all-null config columns so they are omitted from the key (same as
    # the per-row NaN skip) and so groupby isn't grouping on empty axes.
    config_cols = [c for c in _config_cols(df.columns) if df[c].notna().any()]
    grouped = df.groupby(["marker", *config_cols], dropna=False, sort=False)[
        mean_cols
    ].median()
    out = {}
    id_frame = grouped.reset_index()
    n_id = 1 + len(config_cols)
    for rec in id_frame.itertuples(index=False, name=None):
        key = _point_key(rec[0], zip(config_cols, rec[1:n_id]))
        for col, val in zip(mean_cols, rec[n_id:]):
            if pd.notna(val):
                out[(key, col)] = float(val)
    return out
```

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+permissions:
+  contents: read
+  checks: write
+  pull-requests: write
+
+jobs:
+  llk-perf-gate:
+    name: "LLK perf regression gate"
+    if: github.event.label.name == 'llk-perf-gate' && github.event.pull_request.head.repo.full_name == github.repository
```
tdowdallTT: Is the `llk-perf-gate` label intended to be added manually on PRs wishing to run this feature?

.github/workflows/llk-perf-gate-impl.yaml
```diff
@@ -0,0 +1,203 @@
...
+          SFTP_PASS: ${{ secrets.SFTP_PASSWORD }}
+        run: |
+          set +e  # Don't fail if baseline doesn't exist (first run)
+
+          # Read baseline commit from metadata
+          BASELINE_COMMIT=$(python3 -c "import json; print(json.load(open('baseline_metadata.json')).get('commit_sha', ''))")
+
+          if [ -z "$BASELINE_COMMIT" ]; then
+            echo "⚠️  No baseline commit found; skipping baseline download"
+            exit 0
```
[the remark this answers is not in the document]
tdowdallTT: Alternatively, you could call `llk-perf-gate-impl` directly as a final job in the `llk-perf` workflow

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,35 @@
...
+# to individual isolate modes (MATH_ISOLATE, UNPACK_ISOLATE, PACK_ISOLATE).
+# Thresholds derived from 5-run noise baselines on Blackhole and Wormhole.
+#
+# BASELINE UPDATES: After merge to main, the sanity workflow re-runs perf tests
+# and publishes new baseline to Snowflake. Next PR's gate compares against it.
+#
+# TRIGGER: Add 'perf-gate' label to PR to run the gate
+on:
+  pull_request:
+    types: [labeled]
```
tdowdallTT: Food for thought, currently there is no way to rerun the `llk-perf-gate` workflow on a perf blocked PR unless the `llk-perf-gate` label is removed and readded.

.github/workflows/llk-perf-gate-impl.yaml
```diff
@@ -0,0 +1,277 @@
+name: "LLK perf gate impl"
+
+on:
+  workflow_call:
+    inputs:
+      arch:
+        description: "Architecture to gate on (wormhole or blackhole)"
+        required: false
+        default: wormhole
```
ndivnicTT, line 9: Can default be both?

.github/workflows/llk-perf-gate-impl.yaml
```diff
@@ -0,0 +1,277 @@
...
+        description: >-
+          Read the baseline from this workflow run's perf artifacts instead of
+          from the warehouse. For piloting the gate without writing anything to
+          Snowflake. Empty = the normal path (Snowflake + SFTP).
+        required: false
+        type: string
+      run_types:
+        description: "Comma-separated run types to gate on (e.g. L1_TO_L1, MATH_ISOLATE, PACK_ISOLATE)"
+        required: false
+        default: L1_TO_L1
```
ndivnicTT, line 27: I am good with L1_TO_L1. Can you have option to run each isolation separately, L1_TO_L1 separately, then ALL_ISOLATION_MODES and ALL_MODES (3 ISO + L1_TO_L1)

.github/workflows/llk-perf-gate-impl.yaml
```diff
@@ -0,0 +1,277 @@
...
+
+      - name: Compare regressions
+        id: compare
+        run: |
+          set +e
+          BASELINE_SHA=$(python3 -c "import json; print(json.load(open('baseline_metadata.json')).get('commit_sha', '${{ steps.branch_point.outputs.branch_point }}'))")
+
+          python3 ${{ env.LLK_PATH }}/perf/regression_compare.py \
+            --current 'current_perf/**/*.csv' \
+            --baseline 'baseline_perf/**/*.csv' \
```
ndivnicTT, line 243: Shouldn't we compare parquet files instead?

tt_metal/tt-llk/.claude/skills/perf-regression-check/README.md
```diff
@@ -30,7 +30,8 @@ In Claude Code, `/perf-regression-check perf_math_matmul` (or "compare perf of
 | `--baseline <ref>` | `merge-base origin/main HEAD` | what to compare against |
 | `--current <ref>` | `HEAD` | what to judge |
 | `--iterations <N>` | `3` | sweeps per side; median-vs-median |
-| `--threshold <T>` | `0.05` | 5% — flagged in both directions |
+| `--threshold <T>` | `0.02` | 2% — flagged in both directions |
+| `--min-cycles <N>` | `30` | a point must ALSO move this many cycles; `0` disables |
```
ndivnicTT, line 34: Like we spoke, min-cycles should be something smaller like 8 or 12
If we need dynamic threshold (2% for 30+ cycles but in example 4% for <30 cycles) then let's have it, at least in this first version. If we figure out later that better threshold is needed, then we can update it accordingly

tt_metal/tt-llk/tests/python_tests/test_perf_publish_run.py
```diff
@@ -87,13 +87,35 @@ def test_publish_rejects_unknown_pipeline(tmp_path, monkeypatch):
...
+    the gate filters on it, and the run_id and file name inherit it.
+    """
+    sub = tmp_path / "perf_a"
+    sub.mkdir()
+    _write_csv(
+        sub / "perf_a.csv",
+        pd.DataFrame({"marker": ["INIT"], "tile_cnt": [4], "mean(L1_TO_L1)": [900.0]}),
+    )
+    _set_provenance(monkeypatch)
+    monkeypatch.setenv("PIPELINE", "baseline")
```
ndivnicTT, line 111: Do we have test for all 3 allowed PIPELINE values? Or if we have 3 tests for each PIPELINE value, let's merge into 1 test for all 3

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -437,6 +437,18 @@ def _run_id() -> str:
...
 
+def _pipeline(event: str) -> str:
+    """Which pipeline produced this run: ``pr``, ``nightly`` or ``baseline``.
+
+    Read from PIPELINE when the workflow says so; otherwise inferred from the
+    GitHub event, which can only tell a PR from everything else.
+    """
+    explicit = os.environ.get("PIPELINE", "").strip()
+    if explicit:
+        return explicit
```
ndivnicTT, line 448: Can you assert allowed values for PIPELINE here?

ndivnicTT (approved):
LGTM!
The comments I left can be addressed in this PR or in a follow up, I am good with any choice.
Let's catch some real regressions!

mateusznowakTT (approved):

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,128 @@
...
+        include:
+          - arch: blackhole
+            threshold: "0.02"
+          - arch: wormhole
+            threshold: "0.08"
+    uses: ./.github/workflows/llk-perf-gate-impl.yaml
+    with:
+      arch: ${{ matrix.arch }}
+      run_types: L1_TO_L1
+      threshold: ${{ matrix.threshold }}
```
mvlahovicTT: the impl at this commit has no `threshold` or `min_cycles` input, so github rejects the whole file, every run on this branch is a startup_failure and actionlint reports `input "threshold" is not defined in "./.github/workflows/llk-perf-gate-impl.yaml"`. the inputs only land in #57415, they need to move into llk-perf-gate-impl.yaml in this pr.

.github/workflows/llk-perf-gate.yaml
```diff
@@ -0,0 +1,128 @@
...
+        )
+      )
+    strategy:
+      fail-fast: false
+      matrix:
+        arch: [blackhole, wormhole]
+    uses: ./.github/workflows/llk-perf-impl.yaml
+    with:
+      architecture: ${{ matrix.arch }}
+      artifact-suffix: "-head"
```
mvlahovicTT: no `ref` here, so on a pull_request event llk-perf-impl checks out refs/pull/N/merge, the pr merged into current main, while measure-branch-point measures the old merge-base. anything that landed on main since the branch point shows up as this prs regression. passing `ref: ${{ github.event.pull_request.head.sha }}` keeps both sides on the pr lineage.

tt_metal/tt-llk/perf/regression_compare.py
```diff
@@ -79,12 +61,120 @@ def _medians(frames):
...
+def _is_metric(col):
+    return col.startswith("mean(") or col.startswith("std(")
+
+
+MODULE_COL = "test_module"
+
+
+def _is_ignored(col):
+    # Not part of a point's identity: timing stats, code size, source module.
+    return _is_metric(col) or col.startswith("TEXT_SIZE(") or col == MODULE_COL
```
mvlahovicTT: with the module excluded from the key, two tests that sweep the same config share one median. on todays bh trial (36049963430) thats 126 keys shared by perf_fast_untilize_baseline_compare and perf_pack_untilize, so a regression in one of them shows up at about half size. i would keep test_module in the key.

tt_metal/tt-llk/perf/regression_compare.py
```diff
@@ -2,62 +2,44 @@
...
-    return {k: median(v) for k, v in samples.items()}
+import argparse
+import csv
+import glob
+import json
+import os
 
+import pandas as pd
 
 # Defaults are measured, not guessed: five runs of one commit on one card, over
```
mvlahovicTT: 16 lines to explain two constants, and the README it points to is not in the repo. one line with where the 2% and the 30 come from (an issue or the run ids) is enough.

tt_metal/tt-llk/.claude/scripts/perf_compare_commits.sh
```diff
@@ -59,8 +59,10 @@
 set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-COMPARE_PY="$SCRIPT_DIR/perf_regression_compare.py"
 LLK_PRIMARY="$(cd "$SCRIPT_DIR/../.." && pwd)"
+# The canonical compare module, shared with the PR gate, so the skill and the
+# gate can never disagree about what counts as a regression.
+COMPARE_PY="$LLK_PRIMARY/perf/regression_compare.py"
```
mvlahovicTT: this change also drops the executable bit on the script (100755 -> 100644), and the slack message in #57420 tells people to run it directly by path, which now fails with permission denied.

.github/workflows/llk-perf-gate-impl.yaml
```diff
@@ -0,0 +1,255 @@
...
+            echo "comparison_finished=true" >> $GITHUB_OUTPUT
+          else
+            echo "comparison_finished=false" >> $GITHUB_OUTPUT
+            echo "::warning::The comparison did not finish (exit ${EXIT_CODE}); no verdict."
+          fi
+          exit 0  # Don't fail the step; let the verdict step decide
+
+      - name: Comment on PR
+        if: always() && github.event_name == 'pull_request'
+        uses: actions/github-script@v7
```
mvlahovicTT: the only unpinned github-script in the repo, the other ~220 are pinned by sha. also this posts a new comment on every labeled/synchronize run, i would find and update one comment instead.

mvlahovicTT (changes requested):
the direction is right and the regression_compare side is in decent shape (the hw free perf infra tests pass here, 74/74), but this cant go in as is.

- the workflow does not start. llk-perf-gate.yaml passes `threshold` and `min_cycles` to the impl, which only gets those inputs in #57415, so every run on this branch is a startup_failure (36046872176 and the ones before it) and actionlint says the same. this pr targets main by itself, and the trigger is labeled/synchronize on every pr with no path filter, so merging it alone puts a failing run on every pr push in the repo.
- the two sides are not on the same lineage. the head side checks out the pr merge commit and the baseline side the branch point, so whatever landed on main in between counts against the pr (inline).
- one run per side is not enough on wh. i ran the perf_math_matmul config that the 09-23 nightly trial flagged (config4330) 12 times on one n150 with the same binary and got 190.8k in 6 runs, 204k to 210k in 4 and 227k in 2, so regression_compare with the wh settings reports +8.5% and +19.2% regressions between two identical runs. thats #55169, but with 8% and a single run per side this will fire on a lot of prs.
- the description still says hard block, the code is report only (`fail_on_regression: false`) and label triggered. fine for a pilot, just say so.

smaller stuff: there is no concurrency group, so every push measures both sides again in parallel and adds a new pr comment each time, and the iteration count in the report is the number of csv files, which was right for the skill (one test per sweep) but in the gate a single run over 27 tests reads as `27 iteration(s)` (35838600144).

comments: over-verbose comments need to be removed. if the intention is understood from looking at the code there is no need for a comment, and if its really necessary keep it to 1-2 lines max. the 30 line module docstring and the 16 line threshold block in regression_compare.py, the 7 lines in `_ci_provenance` and the pipeline/baseline notes are the main ones. a few of them also describe a post-merge baseline run on main that nothing in this stack adds, and the `docs/perf_evaluation/results/*/README.md` they point to is not in the repo.

PR #53758, opened 2026-08-20
[LLK] Fix #52930: Guard the legacy-compat reciprocal's pole so 1/0 returns inf

### LINKED ISSUE #52930: [LLK] Problematic SFPU edge cases
While sweeping edge values for SFPU ops in pr https://github.com/tenstorrent/tt-metal/pull/52416 some of them showed strange behavior not explained by the ISA doc

| Finding | Ops |
|---|---|
| `signbit(-0.0)` returns 0 where the kernel's own docstring promises 1 ("incl. `-0.0`"). Unlike `sign`/`heaviside` this op claims to read the sign bit *directly*, so either the claim or the implementation is wrong — a kernel-contract bug rather than a hardware one. `-0.0` is delivered correctly, verified host-side. | `signbit` |
| `0/0` and `x%0` return `inf`, not `nan`. `SFPMAD` documents IEEE754 rules, which make `0 × inf` a NaN, so this is the kernels' own reciprocal composition rather than the multiply. The finite poles agree exactly and every ±inf lines up. | `div`, `fmod`, `remainder`, `xlogy` |
| `0**0` returns 0 where C, torch and the golden give 1. `pow` evaluates `exp(b·ln a)`, so a composition artifact. | `pow` |
| `RsqrtCompat(0)` saturates to `1.7014118e38` (`0x7F000000`) instead of `inf`, on all 8 combinations — while plain `Rsqrt` over the same probe does not diverge. Two implementations of one function disagreeing at their shared pole, with nothing in the ISA prescribing either answer. | `RsqrtCompat` |
| `Erfinv(±1)` saturates rather than returning ±inf, on the fp32-dest combinations only. | `erfinv` |

related to https://github.com/tenstorrent/tt-metal/issues/49739

ldjurovicTT:
### PR Category
Bug fix

### Summary
Fixes [#52930](https://github.com/tenstorrent/tt-metal/issues/52930) finding 4: the legacy-compat
reciprocal has no pole guard, so `1/0` returned the max finite `1.7014e38` instead of `inf`.

`_reciprocal_compat_` computes the reciprocal by exponent difference. At `in == 0` the
`setexp(val, 126)` discards the magnitude, so the difference lands on `126 - exexp(0) = 254` — an
ordinary finite — where infinity needs 255. The existing `v_if (new_exp < 0)` guard covers only the
opposite, underflow end. The fix is one `v_if` added after the `setexp`, on both arches:

```c
v_if (sfpi::setsgn(in, 0) == 0.0F) { out = std::numeric_limits<float>::infinity(); }
v_endif;
```

Because the guard sits in the shared primitive, one edit reaches every consumer: `RsqrtCompat`,
the public `recip()` default, the SDPA legacy reciprocal, and Blackhole sampling.

**Architecture scope: Wormhole and Blackhole, both now measured on silicon** — Wormhole on n300 and
Blackhole on p100a. Blackhole is no longer compile-verified only, which closes the two gaps the
previous revision of this PR listed as unmeasured.

| | Wormhole n300 | Blackhole p100a |
|---|---|---|
| Accuracy | 16 of 168 recorded values change, all `1/±0` | **20 of 220 change, all `1/±0`** |
| Performance | **+128 cycles/tile**, flat | **+128.00 cycles/tile**, flat — identical absolute cost |
| `RsqrtCompat` | +5.68 … +11.12 % | **+5.88 … +11.12 %** |
| public `recip()` default | +10.53 … +13.15 % | **+10.53 … +12.92 %** |
| Non-legacy controls | 0.00 % across 120 variants | **0.00 % across 180 variants** |
| Compatibility | preserved | **preserved, and now measured for blaze sampling** |

The cost is absolute, not proportional: the same +128 cycles reads as +5.88 % on the 2175-cycle
`RsqrtCompat` variant and +12.92 % on the 991-cycle approximate `recip`, for identical work.

Full record, both arches: the raw per-run accuracy and perf dumps on [#52930](https://github.com/tenstorrent/tt-metal/issues/52930).

### Performance impact

`perf_eltwise_unary_sfpu.py` with CI's flags, `MATH_ISOLATE` on the `TILE_LOOP` marker, cycles per
tile. Three runs per tree from separate clean build roots so neither tree can serve the other a
stale ELF.

#### Wormhole n300

| Op | Variants | min | median | max | Separated |
|---|---|---|---|---|---|
| **`recip()` legacy default** | 60 | +10.53 % | **+11.95 %** | +13.15 % | 60/60 |
| **`RsqrtCompat`** | 4 | +5.68 % | **+8.76 %** | +11.12 % | 4/4 |
| `Rsqrt` *(control)* | 120 | −0.00 % | +0.00 % | +0.00 % | 0/120 |

#### Blackhole p100a

Baseline is the same tree with only the guard reverted, so the sole delta is the `v_if`. The
measurement is effectively noise-free: the largest run-to-run spread over all 188 measured variants
is **0.039 cycles**.

| Op | Variants | min | median | max | Separated |
|---|---|---|---|---|---|
| **`ReciprocalCompat`** (the public `recip()` default) | 4 | +10.53 % | **+11.87 %** | +12.92 % | 4/4 |
| **`RsqrtCompat`** | 4 | +5.88 % | **+8.54 %** | +11.12 % | 4/4 |
| `Reciprocal` *(control, `legacy_compat=false`)* | 60 | +0.00 % | +0.00 % | +0.00 % | 0/60 |
| `Rsqrt` *(control, `legacy_compat=false`)* | 120 | +0.00 % | +0.00 % | +0.01 % | 0/120 |

Every variant, `Float16_b→Float16_b`:

| Op | approx | dest_acc | baseline | branch | Δ cycles | Δ % |
|---|---|---|---|---|---|---|
| **ReciprocalCompat** | No | No | 1215.02 | 1343.02 | +128.00 | **+10.53 %** |
| **ReciprocalCompat** | No | Yes | 1183.02 | 1311.02 | +128.00 | **+10.82 %** |
| **ReciprocalCompat** | Yes | No | 991.02 | 1119.02 | +128.00 | **+12.92 %** |
| **ReciprocalCompat** | Yes | Yes | 991.02 | 1119.02 | +128.00 | **+12.92 %** |
| **RsqrtCompat** | No | No | 2175.08 | 2303.08 | +128.00 | **+5.88 %** |
| **RsqrtCompat** | No | Yes | 2143.11 | 2271.11 | +128.00 | **+5.97 %** |
| **RsqrtCompat** | Yes | No | 1151.11 | 1279.08 | +127.97 | **+11.12 %** |
| **RsqrtCompat** | Yes | Yes | 1151.08 | 1279.11 | +128.03 | **+11.12 %** |

The 180 control variants are flat — largest absolute movement anywhere among them is **0.0078
cycles**, inside the run-to-run spread. `legacy_compat = false` does not reach the patched
primitive, and BH silicon confirms it rather than the ELF.

A tile is 32 SFPU vector iterations, so +128.00 is 4 cycles per iteration. Added per unrolled copy:
`SFPSETSGN`, `SFPLOADI` ×2, `SFPSETCC`, `SFPENCC`.

**Blackhole shows one thing Wormhole's numbers could not.** The *static* cost of the guard varies
by 8× across variants — math kernel text grows +32 bytes on approximate `ReciprocalCompat` but +256
bytes on exact `RsqrtCompat` — while the *measured* cost is +128.00 in every case. The guard's cycle
cost is therefore not proportional to its instruction footprint; where the compiler emits more
instructions, the extra ones issue in slots the surrounding code was already stalling in. Same
conclusion the WH record reached from its +64-static-against-4-cycles mismatch, with a cleaner
demonstration. Recorded as measured, not reasoned.

**What the second pole costs, priced exactly.** The cheaper guard this PR did not take, measured on
BH against the same baseline:

| Guard form | Δ cycles/tile | per iteration | `RsqrtCompat` | `ReciprocalCompat` |
|---|---|---|---|---|
| bare `in == 0.0F` | **+96.00** | 3 cycles | +4.41 … +8.34 % | +7.90 … +9.69 % |
| shipped `setsgn(in, 0)` | **+128.00** | 4 cycles | +5.88 … +11.12 % | +10.53 … +12.92 % |

**The `setsgn` costs exactly +32.00 cycles per tile — one `SFPSETSGN`, one cycle per iteration**, on
all eight variants with a spread of at most 0.03 cycles. Unlike the guard as a whole, this closes as
an exact instructions-times-iterations identity. So the price of the second pole is one instruction:
32 cycles per tile buys `1/-0 = -inf`. The WH record could only say "roughly 2 percentage points".

This cost is accepted as necessary for correctness.

### Accuracy verification

Raw hardware bit patterns for a fixed stimulus list, diffed byte for byte against the same tree with
the guard reverted — which is what the shipped sweep cannot do, since it compares against golden
with tolerances rather than against the previous build.

#### Wormhole n300

168 recorded values for `RsqrtCompat`: **exactly 8 change**, one per combination, all the pole
(`0x7EFFFD9E` → `0x7F800000`). For the public `recip()` default, 16 change, 8 per pole, with
`1/-0` reaching `-inf` where a real `-0.0` survives to the LREG. Every non-pole value byte-identical,
and the relative-error table over the wide exponent sweep unchanged in every cell.

#### Blackhole p100a

Blackhole reaches 5 of the 8 (format pair, `dest_acc`) combinations — `dest_acc=No` supports only
`Float32→Float32` there — and both `RsqrtCompat` and `ReciprocalCompat` are driven, so the public
`recip()` default is measured directly rather than through an instrument.

**220 rows compared (22 inputs × 5 combinations × 2 ops), 20 differ, every one of them a `±0` input.
The count of differing rows whose input is not `±0` is zero, and no pole row was left behind.**

| Op | Path | `x` | before | after | IEEE |
|---|---|---|---|---|---|
| both | fp32 output | `+0` | `0x7EFFFD9E` | `0x7F800000` = `+inf` | `+inf` |
| both | bf16 output | `+0` | `0x7F000000` | `0x7F800000` = `+inf` | `+inf` |
| `ReciprocalCompat` | `Float32→Float32`, `dest_acc=Yes` | `-0` | `0xFEFFFD9E` | **`0xFF800000` = `-inf`** | `-inf` |
| `ReciprocalCompat` | `Float32→Float16_b`, `dest_acc=Yes` | `-0` | `0xFF000000` | **`0xFF800000` = `-inf`** | `-inf` |
| `ReciprocalCompat` | the other three | `-0` | positive finite | `+inf` | `-inf` |
| `RsqrtCompat` | all five | `-0` | positive finite | `+inf` | — |

The two `-inf` rows are exactly the unpack-to-dest pipelines — the only BH combinations where a real
`-0.0` survives to the LREG. They confirm independently on BH that `setsgn(in, 0)` rather than a bare
`in == 0.0F` is what makes the negative pole reachable at all. On the other three the unpack flushes
`-0.0` to `+0.0` before the kernel runs, so `+inf` there is the flush showing through, not a defect
in the guard — the probe records the delivered input bits, so this is measured rather than inferred.

Every other value — `1`, `2`, `4`, `0.25`, `0.5`, `3`, `7`, `1.5`, `100`, `1e±3`, `1e±10`, `1e±30`,
`1.1754944e-38`, `3.4028235e38`, `2.3841858e-07`, `0.015625`, `+inf` — is byte-for-byte identical on
all five combinations and both ops. That is the acceptance criterion the fix plan set, met on BH as
well as WH.

#### Wormhole n300

| Sweep | before | after |
|---|---|---|
| `test_sfpu_unary.py -k "Reciprocal or Rsqrt"` | 497 passed, 15 xfailed | **505 passed, 7 xfailed** |
| `test_sfpu_unary.py -k edges` | 491 passed, 23 xfailed | **499 passed, 15 xfailed** |
| `test_sfpu_sdpa.py` | 65 passed | **65 passed** |

#### Blackhole p100a

| Sweep | before (guard reverted) | after |
|---|---|---|
| `test_sfpu_unary.py -k "Reciprocal or Rsqrt"` | **10 failed**, 435 passed, 78 skipped, 5 xfailed | **445 passed**, 78 skipped, 5 xfailed, **0 failed** |
| `test_sfpu_unary.py -k edges` | **10 failed**, 308 passed, 429 skipped, 13 xfailed | **318 passed**, 429 skipped, 13 xfailed, **0 failed** |
| `test_sfpu_sdpa.py` | 65 passed | **65 passed** |
| `test_sfpu_sampling.py` | 51 passed, 93 skipped | **51 passed, 93 skipped** |
| `test_sfpu_unary.py` (whole file) | — | **5042 passed**, 1607 skipped, 13 xfailed, **0 failed, 0 xpassed** |

0 xpassed throughout and no xfail count moves in either selection, so nothing was absorbed by a
marker and no stale xfail entry is left behind. Both unary deltas are exactly the same 10 cells
(435 → 445, 308 → 318): 5 combinations × `{RsqrtCompat, ReciprocalCompat}`, which is every
combination BH reaches for the two ops the guard touches. The 5 remaining xfails are pre-existing and
untouched — 3 × `Reciprocal`'s `1/NaN`, 2 × `Rsqrt`'s `rsqrt(-0)`.

**The regression test is non-vacuous on Blackhole, and that was checked rather than assumed.**
Against the unguarded kernel the edge variants fail **10 of 10** — golden `inf`, kernel `1.7014e38`.
Against the guarded kernel all 10 pass. Wormhole established this for its 8 combinations; BH now
establishes it for its 5, so reverting the guard breaks CI on both arches.

**Blackhole sampling is measured now, not argued.** `sampling_recip_value<true>` is the one consumer
carrying a written "must stay bit-identical for blaze" contract, and it is a Blackhole-only op. The
previous record could only compile-verify it and reason that the guard is unreachable inside its
documented `in > 0` domain. `test_sfpu_sampling.py` now runs on p100a and is unchanged before and
after — 51 passed / 93 skipped both ways, including every `legacy_compat=True` variant across both
`vector_mode` settings. The contract holds by measurement.

### Why this shape of fix, and not the faster one

The plan offered two options and this PR **originally shipped the other one** — redirecting the
compat consumers to `sfpu_reciprocal_iter`, which fixes the pole and is **26–39 % faster** on
`RsqrtCompat`, **43–53 % faster** on the public `recip()` default, and 293×/610× more accurate. It
was withdrawn on review.

It changes numerics for every `legacy_compat = true` caller — the opposite of what a flag named
`legacy_compat` promises. The plan said Option A "is only viable once someone can say who depends on
the legacy bit pattern… **Do not guess**", and the answer is eight production normalization kernels
that hard-code `rsqrt_tile<true>`:

| File | Sites |
|---|---|
| `experimental/ccl/dit_fused_distributed_rmsnorm/…/dit_layernorm_fused_compute.cpp` | 2 |
| `normalization/groupnorm/…/{groupnorm, groupnorm_sharded_v2, welford_groupnorm, welford_groupnorm_sharded_v2}.cpp` | 4 |
| `normalization/layernorm_distributed/…/layernorm_post_allgather_welford.cpp` | 1 |
| `experimental/transformer/dit_layernorm_post_all_gather/…/layernorm_post_allgather_welford.cpp` | 1 |
| `experimental/ccl/rms_allgather/…/rms_compute.cpp` | 1 |

and they say why:

> ```
> // legacy rsqrt to match the composite dit_layernorm baseline (it uses
> // rsqrt_tile<true>); the non-legacy default diverges on low-variance rows.
> ```
> — `dit_layernorm_fused_compute.cpp:267-268`, and at `:345` `// DV = 1/std (legacy, matches baseline)`

Someone had already compared the two kernels and chosen legacy deliberately. The earlier search for
that dependency looked for the phrase *bit-identical*, found only the Blackhole sampling file, and
concluded the contract was unclaimed; these callers express it as *matches baseline*.

Consolidating on one reciprocal is still the right end state — a second implementation that is 610×
less accurate is not worth keeping forever — but it is a follow-up gated on migrating and
revalidating those eight callers, not something to land underneath them.

### User-visible behaviour change

| Path | Before | After |
|---|---|---|
| `recip_tile()` — the **public default**, `recip.h` declares `legacy_compat = true` | `recip(0)` = `1.7014e38` | `recip(0)` = `+inf`, `recip(-0)` = `-inf` |
| `rsqrt_tile<true>` / `RsqrtCompat` | `rsqrt(0)` = `1.7014e38` | `rsqrt(0)` = `+inf` |
| SDPA `calculate_recip_first_column<true>` | same pole | same fix |
| Blackhole `sampling_recip_value<true>` | same pole | same fix; guard unreachable inside its documented `in > 0` domain, so blaze's bit pattern is unchanged — now verified on p100a silicon |
| Everything on `legacy_compat = false` | — | untouched |

**Please read the open question below before approving.**

### Notes for reviewers

**Open question — `recip(0)` now propagates infinity.** This is the intended fix, but it is worth
confirming downstream, and no amount of LLK-level silicon data answers it. `recip(0)` was the max
finite and is now `+inf`, so a multiply that used to give `0` now gives `NaN`. The concrete case
raised in review: a fully-masked softmax row masks with `-inf` (`softmax.cpp:275`), reciprocates the
zero sum unguarded (`:341-354`) and multiplies it back (`:369`) — previously a row of zeros, now a
row of NaN. `moreh_common.hpp`, moe, sampling and the deepseek gates ride the same default. **Have
the metal op sweeps been run against this?** If a guarded-zero contract is wanted at the op level,
that belongs in the op, not back in the kernel — but it should be a deliberate decision.

Where to focus:

1. **The two kernel files** are the whole functional change; everything else is tests and records.
2. **The guard compares on `setsgn(in, 0)`**, not a bare `in == 0.0F`, because `SFPSETCC`'s contract
   excludes negative zero — measured, the bare compare does not fire for `-0.0`. Both forms were
   built and measured on both arches; the `setsgn` costs exactly one `SFPSETSGN` (+32 cycles/tile)
   and it is what fixes the second of the two poles.
3. **`_calculate_rsqrt_compat_` and `_calculate_reciprocal_compat_` keep their callers**, so no
   orphans are created.
4. **No investigation markdown ships**, per review comment. The five records that were at repo
   root are gone, and so is the single `docs/` file that briefly replaced them — nothing in-source
   points at either any more, the one reference having been the `test_sfpu_unary.py` comment block
   that went in the same commit. Everything load-bearing lives in this description: the
   eight-caller list, the Option A figures and why it was withdrawn, and the open question above.
   The raw per-run accuracy and perf dumps stay on
   [#52930](https://github.com/tenstorrent/tt-metal/issues/52930).

**The coverage gap this found, and closed.** The public `recip()` default is the largest
user-visible change here and the tt-llk suite did not exercise it at all — `sfpu_operations.h`
called `calculate_reciprocal` and `recip_init` without a `legacy_compat` argument, so both defaulted
to `false` and the suite only ever built the kernel the public API does *not* dispatch. A permanent
`MathOperation.ReciprocalCompat` op now covers it, wired the way `RsqrtCompat` covers legacy rsqrt.
It is why the Blackhole figures above needed no temporary harness patch, where the Wormhole ones did.

**Known gaps, both recorded on #52930.** The SDPA reciprocal column still has no perf
instrumentation on either arch (`perf_sfpu_reduce_sdpa.py` instruments only `ReduceColumn`), so its
cycles are unmeasured — its body is the same `_reciprocal_compat_`, so +128 cycles/tile applies, but
the record will not put a measured number on it. End-to-end model impact is likewise not measured:
+128 cycles on a normalization kernel's rsqrt is not +128 cycles on a model. Quasar is untouched and
has no equivalent compat reciprocal.

**Cross-branch interaction with #53757 — read before resolving the rebase conflict.**
This PR and #53757 (`sqrt_custom(+inf)`) edit the same hunk of `test_sfpu_unary.py` and each one
deletes exactly what the other keeps, so a merge conflict there is guaranteed and resolving it by
picking a side loses the other's fix.

| | #53757 | this PR | correct after both |
|---|---|---|---|
| `Erfinv` entry in `_EDGE_KNOWN_DIVERGENCES` | **removed** — its fix repairs `erfinv(±1)` | **present**, 2 combos | **removed** |
| `RsqrtCompat` entry | **present**, 8 combos | **removed** — this guard fixes the pole | **removed** |
| `STILL OPEN` comment block | keeps the `rsqrt at 0` paragraph | keeps the `Erfinv` paragraph | **neither** |
| `SqrtCustom` cat-B entry | **added** (the `-inf` residual) | absent | added |

After both land the table should read `Sign`, `Heaviside`, `SqrtCustom` and nothing else, and
#53757's `Rsqrt` reason string — "Distinct from the RsqrtCompat entry *above*" — has to lose that
clause since the entry it points at is gone.

The `Erfinv` entry is deliberately **not** pre-removed here: on this branch `erfinv(±1)` still
saturates, and both of its cells xfail on p100a as expected, so deleting it now would turn 2 xfails
into 2 hard failures. It only becomes stale once #53757 lands. Because the marker is `strict=False`,
a leftover entry would XPASS rather than fail — visible in the summary line but not a CI break — so
it would rot quietly. Suggested order: land #53757 first, rebase this branch, drop the `Erfinv`
entry and its reason during conflict resolution, and re-run the BH edge sweep expecting **11 xfailed,
0 xpassed** (down from the 13 xfailed / 0 xpassed measured here).

Separately: the SDPA legacy reciprocal also returns `1/|x|` — the sign is genuinely dropped. Out of
scope here; fixed on `ldjurovic/sfpu_sdpa_recip_legacy_sign`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ldjurovic/sfpu_52930_reciprocal_compat_pole)


ncvetkovicTT:
@ldjurovicTT Remove [do not review], аригато

docs/sfpu_52930_reciprocal_compat_pole.md
```diff
@@ -0,0 +1,494 @@
+# `_reciprocal_compat_(1/0)` on Wormhole and Blackhole
```
ndivnicTT: This file should not be merged, right?

tt_metal/tt-llk/tests/python_tests/test_sfpu_unary.py
```diff
@@ -543,13 +543,19 @@ def test_eltwise_unary_sfpu(
...
+# CLOSED:
+#
+#   rsqrt_compat at 0 returned 1.7014118e38 instead of inf on all 8 combinations, while
+#   plain Rsqrt over the same probe did not diverge. Root-caused to _reciprocal_compat_,
+#   whose exponent-difference arithmetic (126 - exexp(0) = 254) has no pole guard, and
+#   fixed by adding that guard in place. The kernel is otherwise untouched, so every
+#   non-pole value stays bit-identical -- deliberately, because eight production
+#   normalization kernels select this path by name to match a model baseline. Verified on
+#   Wormhole n300 and Blackhole p100a silicon; see
+#   docs/sfpu_52930_reciprocal_compat_pole.md.
```
ndivnicTT: Does this comment bring any value? I would say it can be erased

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
```diff
@@ -93,7 +95,31 @@ sfpi_inline sfpi::vFloat _reciprocal_compat_(const sfpi::vFloat in)
...
+    // that makes it an infinity.
+    //
+    // Compared on setsgn(in, 0) rather than a bare in == 0.0F because SFPSETCC's contract
+    // excludes negative zero (VectorUnit.md); measured, the bare compare does not fire for
+    // -0.0 and leaves that pole at 1.7e38. Clearing the sign first costs ~2 percentage points
+    // of the total, and it is what makes -0.0 reach this guard at all. The result there is
+    // then IEEE's: measured 1/-0.0 = -inf on the unpack-to-dest pipelines, where a real -0.0
+    // survives to the LREG and the caller-side v_if(in < 0.0) re-signs the magnitude. On the
+    // pipelines that flush -0.0 to +0.0 before the kernel sees it the answer is +inf, which
+    // is the flush showing through rather than this guard.
```
ndivnicTT: This can be reduced to only state reasoning and impact, not the whole flow?

docs/sfpu_52930_reciprocal_compat_pole.md
ldjurovicTT: You are right, once I get approvals and confirm everything we need is here I will remove this file. Was used only for internal tracking of perf and accuracy

ldjurovicTT: Removed in 5df2336. Nothing in-source pointed at it any more - the one reference was in the test comment block you flagged below, which is gone in the same commit. The load-bearing content is in the PR description; the raw per-run accuracy and perf dumps stay on #52930.

tt_metal/tt-llk/tests/python_tests/test_sfpu_unary.py
ldjurovicTT: Agreed, erased in 5df2336. It described an _EDGE_KNOWN_DIVERGENCES entry that this PR removes, so it pointed at something a reader can no longer find. Dropped the matching backreference in the Rsqrt reason string for the same reason.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
ldjurovicTT: Reduced to reasoning and impact in 5df2336, both arches - 17 lines down to 9. Kept the root cause (126 - exexp(0) = 254 lands finite where an infinity needs 255) and the two constraints on the guard's form, since both are easy to break: it has to run after the setexp, and it has to compare setsgn(in, 0) rather than a bare in == 0.0F. Cut the measurement narration and the -0.0 pipeline walkthrough.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
```diff
@@ -93,7 +95,23 @@ sfpi_inline sfpi::vFloat _reciprocal_compat_(const sfpi::vFloat in)
     v_endif;
 
     // Set newly denormalized exponent to result exponent field
-    return setexp(result, new_exp);
+    sfpi::vFloat out = setexp(result, new_exp);
```
fvranicTT:
```suggestion
    sfpi::vFloat out = sfpi::setexp(result, new_exp);
```
This is a nit pick

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
```diff
@@ -93,7 +95,23 @@ sfpi_inline sfpi::vFloat _reciprocal_compat_(const sfpi::vFloat in)
     v_endif;
 
     // Set newly denormalized exponent to result exponent field
-    return setexp(result, new_exp);
+    sfpi::vFloat out = setexp(result, new_exp);
```
fvranicTT: ditto

fvranicTT:
Please fix the PR title @ldjurovicTT

fvranicTT (approved):

tt_metal/tt-llk/tt_llk_blackhole/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
ldjurovicTT: Applied in 2d2665d, both arches. `setexp` was the only sfpi call the pole guard's hunk left unqualified — the surrounding `sfpi::setsgn` / `sfpi::vFloat` were already explicit, so this was the odd one out in new code, and `sfpi::setexp` is the form used elsewhere in the same tree (`ckernel_sfpu_expm1_cw.h:53`).

The pre-existing unqualified `setexp` / `exexp` calls further up `_reciprocal_compat_` are left alone, since they are outside this PR's diff.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_rsqrt_compat.h
ldjurovicTT: Applied in 2d2665d, both arches. `setexp` was the only sfpi call the pole guard's hunk left unqualified — the surrounding `sfpi::setsgn` / `sfpi::vFloat` were already explicit, so this was the odd one out in new code, and `sfpi::setexp` is the form used elsewhere in the same tree (`ckernel_sfpu_expm1_cw.h:53`).

The pre-existing unqualified `setexp` / `exexp` calls further up `_reciprocal_compat_` are left alone, since they are outside this PR's diff.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_recip.h
```diff
@@ -110,13 +110,12 @@ sfpi_inline void sfpu_reciprocal_init() {
...
+// exponent-difference arithmetic has no pole guard, so recip(0) returned 1.7e38 instead of inf.
+// See FIX_PLAN_52930_reciprocal_compat_pole.md.
 template <bool APPROXIMATION_MODE, bool is_fp32_dest_acc_en, int ITERATIONS = 8, bool legacy_compat = false>
 inline void calculate_reciprocal() {
-    if constexpr (legacy_compat) {
-        _calculate_reciprocal_compat_<APPROXIMATION_MODE, ITERATIONS, is_fp32_dest_acc_en>(ITERATIONS);
-    } else {
-        _calculate_reciprocal_internal_<APPROXIMATION_MODE, ITERATIONS, is_fp32_dest_acc_en>(ITERATIONS);
-    }
+    _calculate_reciprocal_internal_<APPROXIMATION_MODE, ITERATIONS, is_fp32_dest_acc_en>(ITERATIONS);
```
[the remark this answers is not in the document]
ldjurovicTT: Half of this went stale with the rewrite, half of it did not. Resolving the thread because the anchor is gone and the surviving question is tracked in the PR description, not because the question is answered.

**Stale half.** The PR no longer redirects anything — it was rewritten to Option B in a9f8685. `ckernel_sfpu_recip.h` is out of the diff entirely, `legacy_compat` stays a real selector rather than a no-op, and precision on the legacy path is unchanged. The whole functional change is now one `v_if` in `_reciprocal_compat_`, both arches. So "beyond precision the pole moves" no longer applies: only the pole moves.

**Live half.** The rest of your point survives the rewrite intact, because the guard sits in the primitive that `recip_tile()`'s default reaches. `recip(0)` moves from `1.7014e38` to `+inf` (and `recip(-0)` to `-inf`) on exactly the default path you flagged, and the softmax trace checks out as written — `softmax.cpp:352-353` reciprocates the row sum unguarded and `:369` multiplies it back through `mul_tiles_bcast<BroadcastType::COL>`, so a fully-masked row goes from a row of zeros to a row of NaN. The consumer list is if anything larger than you said: `moreh_common.hpp` (5 sites), `reduce_helpers_compute.hpp` (2), moe, sampling, the topk / deepseek gates, the moreh softmax and nll_loss kernels, and the three other softmax variants all ride the same default.

**On the question itself.** Yes, the inf propagation is intended — it is the fix. No, the metal op sweeps have not been run against it. That is carried in the PR description under "Open question — `recip(0)` now propagates infinity", above a "please read the open question below before approving" line, which is a more visible place for it than a resolved thread anchored on a file this PR no longer touches. My read is that if a guarded-zero contract is wanted it belongs in the op rather than back in the kernel, but that is a call for the op owners and it is still open.

ldjurovicTT:
@fvranicTT All three addressed, thanks.

**Title** — retitled to `[LLK] Fix #52930: Guard the legacy-compat reciprocal's pole so 1/0 returns inf`, matching the `[LLK] Fix #<issue>: <imperative>` shape used on main. The old one was the bug report rather than the change, and carried no tag or issue ref.

**`sfpi::setexp`** — applied on both arches in 2d2665d. It was the only sfpi call the pole guard's hunk left unqualified.

While in here I also repointed the three links in the description that still pointed at `docs/sfpu_52930_reciprocal_compat_pole.md`, which 5df2336 deleted per @ndivnicTT's comment. Nothing in-source ever pointed at it, but the description did.

The one thing still genuinely open is unchanged and is not a review nit: `recip(0)` now propagates `+inf` on the public `recip_tile()` default, so an unguarded downstream multiply that used to give `0` gives `NaN` — a fully-masked softmax row being the concrete case. It is the intended fix, the metal op sweeps have not been run against it, and it is flagged in the description under "Open question". Worth a second opinion from the op owners before merge even though the LLK side is green.

fvranicTT:
Let's see if we can actually remove RsqrtCompat and RecipCompat. These are old and shouldn't be used anymore I think.

ldjurovicTT:
> Let's see if we can actually remove RsqrtCompat and RecipCompat. These are old and shouldn't be used anymore I think.

Let's merge this and then we can discuss on removal of mentioned kernels

Issue #52153, opened 2026-08-05
Migrate composite ops to device ops - Round 3


KalaivaniMCW:
Goal :

Identify composite ops that can be migrated as device ops ( ops missed in Round 2 due to change of priorities)
Get the profiling results of the op before and after migration
Record PCC of test before and after
Add measurement that is not PCC, likely allclose or ULP - plot result in the PR
Add tests with edge-case coverage
Int32 tests should use torch.equal not compare_pcc
pass seed values to older tests with data_gen fn (random generation)
use linear dataset that covers the entire range of 2**16 (65K)

Tracker:
https://docs.google.com/spreadsheets/d/1zoYh0gNYcz7ZdEtc529tG75-Zl_osx91TWMPDL95z5U/edit?gid=1982557486#gid=1982557486 

The following composite ops need to be ported to fused ops:

- [ ] Nextafter - remove ?
- [ ] clamp
- [ ] div_no_nan
- [ ] prelu
- [ ] polyval
- [ ] MAC - draft
- [ ] clamp
- [ ] clamp_bw

Issue #49996, opened 2026-07-16
[Perf] Reduce kernel-dispatch overhead in composite eltwise ops (clip, nextafter, glu-family) — assumptions to try


Aswinmcw:
## Summary

While profiling eltwise ops at **single-tile** size, several composite ops spend most of their time on **kernel-dispatch overhead** (number of device kernels launched) rather than actual compute. This issue lists a few optimization ideas to investigate.

> [!NOTE]
> These are **assumptions / hypotheses to try**, based on single-tile profiling and reading the composite op sources. They still need proper validation (correctness + perf across shapes/dtypes/broadcast) before acting. Numbers below are indicative, not authoritative.

## Measurements (indicative)

Single tile `(1, 1, 32, 32)`, bfloat16, DRAM, device kernel duration (via `ttperf`), plus the number of device kernels each op dispatches:

| op | duration | kernels | bottleneck |
|----|---------:|--------:|------------|
| `exp` (single-kernel reference) | ~1.5 µs | 1 | baseline |
| `geglu` / `reglu` | ~4.6 µs | 4 | dispatch |
| `glu` / `swiglu` | ~5.4 µs | 4 | dispatch |
| `nextafter` | ~14.1 µs | 6 | dispatch |
| `clip` (tensor min/max) | ~13.6 µs | 7 | dispatch |
| `lgamma` | ~8.5 µs | 1 | compute |
| `polygamma` | ~14 µs | 1 | compute |

At single tile, each dispatched kernel costs roughly ~1.5–2 µs of fixed launch overhead, so composite ops scale with kernel count. (At large shapes these become DRAM-bound instead, but the dispatch-count reduction still helps.)

### 1. `clip` / `clamp` with tensor min & max — reduce kernel count (looks like a free win)
`ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp` (clamp with both `min` and `max`) currently dispatches ~7 kernels. The middle block:

```cpp
Tensor a_max = ttnn::minimum(input_a, max.value(), ...);
Tensor temp = ttnn::where(
    ttnn::eq(min.value(), 0.0f, ...),
    ttnn::relu(a_max, ...),
    ttnn::maximum(a_max, min.value(), ...),
    ...);
```

[...]

PR #53761, opened 2026-08-20
[Bug fix] nextafter: step the bit pattern by one ULP toward the target

### LINKED ISSUE #53760: ttnn.nextafter does not return the adjacent value: the step is FLT_EPSILON, which is one ULP only inside [1, 2)

**Issue Type:** Bad Outputs

## Observed

`ttnn.nextafter(a, b)` does not return the value adjacent to `a`. On most inputs it returns `a` unchanged, and on the rest it steps the wrong way.

Measured on Wormhole B0 N300 and Blackhole P300, bit-identical on both, against `torch.nextafter`, which is the op's registered golden:

| case | bfloat16 | float32 |
|---|---|---|
| step toward a larger `b` (18 pairs) | **18 wrong**, 16 did not move at all | **18 wrong**, 10 moved away from `b` |
| step toward a smaller `b` (18 pairs) | **18 wrong**, 16 did not move at all | **18 wrong**, 10 moved away from `b` |
| `a == b` (18 pairs) | 0 wrong | 0 wrong |
| zero, inf, nan (9 pairs) | 9 wrong | 8 wrong |

`a == b` is the only case that comes out right, and it is right by accident: the op returns `a` unchanged there because it fails to move anywhere.

```
nextafter(0.001,  1)     ttnn 0.000999450684   want 0.00100708008
nextafter(0,      1)     ttnn -1.1920929e-07   want 9.18354962e-41   sign is wrong too
nextafter(inf,    0)     ttnn inf              want 3.38953139e+38
```

**The half of this that is already known.** #51218 item 1 reports that the two `where` arms are swapped, so the result moves away from `input_b`. That is one of the two defects here and its diagnosis is correct.

**The half that is not.** The step size is also wrong, and it is the larger error. `hal::get_eps()` is `FLT_EPSILON`, which is one ULP only inside `[1, 2)`. Above that band the addition rounds straight back to `input_a`, so the op does not move at all: 16 of the 18 bfloat16 cases above are unchanged outputs rather than wrong-direction ones. Below the band the step overshoots by orders of magnitude. Fixing only the swapped arms leaves the op wrong on every magnitude outside `[1, 2)`.

## Expected

`nextafter(a, b)` returns the representable value adjacent to `a` in the direction of `b`. That is a property of the bit pattern: step the magnitude by one unit in the last place of the format and restore the sign. There is no step size in the value domain that is one ULP everywhere, which is why this cannot be repaired by choosing a better epsilon.

Worth noting because it changes what the fix should look like: the op is also the most dispatch-bound composite in the eltwise set. #49996 measures it at 6 kernel launches and proposes a fused SFPU kernel in its second section, #52153 tracks `nextafter` for migration to a device op, and #32509 asked for the same migration before that and was closed while carrying a `do-not-close` label. Stepping the bit pattern is one pass over `dst`, so the correctness fix and the launch-count fix are the same change. Composing the bit manipulation out of existing ttnn ops instead, which is what #52668 does, fixes the values but takes the launch count from 6 to about 20.

## Steps to reproduce

Compare against `torch.nextafter` on a bit-exact gate, over magnitudes spanning `1e-6` to `1e6`, both signs, both directions. The existing coverage cannot see this: `test_binary_composite.py:43` uses `compare_pcc` over `[-100, 100]`, and a one-ULP directional error does not move PCC off 1.0.

## Not established

- `bfloat8_b` and `bfloat4_b`, which the nightly sweep covers and this does not.
- Results that are subnormal. `nextafter(0, 1)` is the smallest subnormal, which the hardware flushes to zero, so that case stays wrong after a fix at this layer.
- Whether any caller depends on the current behaviour. `grep -rn "ttnn\.nextafter\|ttnn::nextafter" models/ | wc -l` returns 0.

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `b2e77906b9`, rebased onto `04cf22f7ee` and re-verified there
* Hardware: Wormhole B0 N300 and Blackhole P300

kinginu:
### PR Category

Bug fix.

### Summary

Closes #53760.

`nextafter` becomes a fused SFPU binary op that steps the magnitude bit pattern by one and restores the sign. The six-op composite goes away, so this is also a launch-count change: 6 kernels per call to 1.

The step is one unit in the last place of the **destination** format, not of the register. Registers are fp32 whatever `dst` holds, so a bfloat16 destination steps by `0x10000`, which is also exactly its smallest subnormal and so serves as both constants. Infinity needs no case of its own: one step down from the exponent field all ones is the largest finite value, which is the answer.

### Accuracy

Both architectures, both dtypes, against `torch.nextafter`, gated on bit-exactness. Wormhole B0 N300 and Blackhole P300 return identical counts, so the table is written once.

| case | before | after |
|---|---|---|
| step toward a larger `b` (18) | 18 wrong | **0** |
| step toward a smaller `b` (18) | 18 wrong | **0** |
| `a == b` (18) | 0 wrong | 0 wrong |
| zero / inf / nan (9), bfloat16 | 9 wrong | **5** |
| zero / inf / nan (9), float32 | 8 wrong | **3** |

What still differs is `nextafter(±0, b)`, whose answer is the smallest subnormal and which the hardware flushes to zero, in both dtypes. bfloat16 additionally loses NaN, which an ordinary `ttnn.mul(x, 1.0)` does too in that dtype.

### Cost

Device profiler, `TT_METAL_DEVICE_PROFILER=1`, `[1, 1, 512, 512]`, 33 invocations, median cycles per core. Zone counts give the launch count directly: 6 per call before, 1 after, on both machines. Cycles are per call, so the composite figure is its per-launch median times six.

| | BH bfloat16 | BH float32 | WH bfloat16 | WH float32 |
|---|---|---|---|---|
| launches per call | 6 → **1** | 6 → **1** | 6 → **1** | 6 → **1** |
| TRISC_1 math, before | 25158 | 40236 | 56040 | 109446 |
| TRISC_1 math, after | **5994** | **9230** | **11082** | **21352** |
| | **4.2x** | **4.4x** | **5.1x** | **5.1x** |

BRISC moves the same way, 29784 to 6610 on Blackhole bfloat16.

### Tests

`test_nextafter_neighbour.py`, 76 cases, passing on both machines:

- one ULP steps across nine magnitudes from `1e-6` to `1e6`, both signs, both directions, asserting bit-exactness and, separately, that the result moved toward `b` at all. All 72 fail on `main`.
- `nextafter(x, x) == x`.
- `nextafter(±inf, 0)` is the largest finite value, not infinity.

The existing coverage could not see this: `test_binary_composite.py:43` uses `compare_pcc` over `[-100, 100]`, where a one-ULP directional error does not move PCC off 1.0.

### Related context

#51218 item 1 reports the swapped `where` arms, which is one of the two defects fixed here. #49996 proposes exactly this fused kernel in its second section and measures the op at 6 launches; #52153 tracks `nextafter` for migration, as #32509 did before it.

#52668 fixes the same two defects by composing about twenty ttnn ops in place of the six, including `bitcast`, `full_like` and five `where`s. That is a larger launch count on the op #49996 measures as the most dispatch-bound composite in the set. The bit-level step is a single SFPU pass, so doing it in the kernel fixes the same values and removes the launches rather than adding to them.

### Not covered

- `bfloat8_b` and `bfloat4_b`, and the sharded and broadcast paths.
- Subnormal results, which the hardware flushes and this does not change.
- Quasar, which carries its own copy of the composite.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_nextafter_neighbour.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_nextafter_neighbour.py
@@ -1,0 +1,75 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""nextafter returns the adjacent representable value.
+
+The gate is bit-exactness against torch.nextafter, which is the op's own golden.
+Nothing weaker can test this: the answer differs from the input by one unit in
+the last place, so any tolerance that admits a near miss admits every miss.
+"""
+
+import pytest
+import torch
+import ttnn
+
+# Magnitudes spanning six decades either side of 1. The composite this replaced
+# stepped by FLT_EPSILON, which is one ULP only inside [1, 2), so it failed
+# everywhere except that band, in both directions.
+MAGNITUDES = [1e-6, 1e-3, 0.5, 1.0, 1.5, 3.0, 100.0, 1e4, 1e6]
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+@pytest.mark.parametrize("magnitude", MAGNITUDES)
+@pytest.mark.parametrize("sign", [1.0, -1.0])
+@pytest.mark.parametrize("direction", [1.0, -1.0])
+def test_nextafter_steps_one_ulp(device, dtype, magnitude, sign, direction):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+
+    a_val = sign * magnitude
+    b_val = a_val + direction * (magnitude * 10 + 1)
+
+    a = torch.full((1, 1, 32, 32), a_val, dtype=torch_dtype)
+    b = torch.full((1, 1, 32, 32), b_val, dtype=torch_dtype)
+    want = torch.nextafter(a, b).flatten()[0]
+
+    ia = ttnn.from_torch(a, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    ib = ttnn.from_torch(b, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    got = ttnn.to_torch(ttnn.nextafter(ia, ib)).flatten()[0]
+
+    assert got == want, f"nextafter({a_val}, {b_val}) returned {got}, want {want}"
+    # Stated separately because a result that fails to move and one that moves
+    # the wrong way are different defects, and the composite did both.
+    assert (got - a_val) * (b_val - a_val) > 0, f"nextafter({a_val}, {b_val}) did not step toward b"
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_nextafter_equal_inputs_are_unchanged(device, dtype):
+    """IEEE-754: nextafter(x, x) is x, for every x."""
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+
+    vals = torch.tensor(MAGNITUDES + [-m for m in MAGNITUDES], dtype=torch_dtype)
+    a = vals.repeat(32 * 32 // vals.numel() + 1)[: 32 * 32].reshape(1, 1, 32, 32)
+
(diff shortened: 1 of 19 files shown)


mateusznowakTT:
nextafter kernel was recently added to main https://github.com/tenstorrent/tt-metal/pull/56050

Issue #48240, opened 2026-06-26
Eltwise host code: 28 bugs across unary/binary/ternary/backward/complex/quantization ops


Aswinmcw:
## Summary

Full analysis of the eltwise directory on latest main (3aae8f4b) — 173 files across unary, binary, ternary, backward, complex, quantization, and kernel code. **28 bugs total.**

---

## Critical

- [ ]  ### 1. `div_bw`: both `where()` edge-case results discarded

**File:** `binary_backward/binary_backward.cpp` ~L700, L722

When `other == 0`, both the input gradient and other gradient have `where()` edge-case handling (NaN/Inf fix-ups) whose return value is never assigned back. The computed result is silently discarded.

```cpp
// L700 — result not assigned to input_grad
ttnn::where(
    ttnn::eqz(other, output_mem_config),
    ttnn::where(ttnn::eqz(grad_tensor, ...), t_nan, ttnn::multiply(ttnn::sign(...), t_inf, ...), ...),
    input_grad.value(),
    output_mem_config);  // <- return value dropped
```
Same issue at ~L722 for `other_grad`.

---

- [x] ### 2. `BinaryNgKernelConfig` SCALAR_A: `reader_kernel` + `writer_kernel` left uninitialized

**File:** `binary_ng/device/binary_ng_utils.cpp` ~L44

The `SCALAR_A` case only sets `compute_kernel` and `bcast_input`, leaving `reader_kernel` and `writer_kernel` uninitialized — undefined behaviour for any SCALAR_A broadcast binary op.

```cpp
case SubtileBroadcastType::SCALAR_A:;   // semicolon not colon
    compute_kernel = KernelName::ComputeBcast;
    bcast_input = 0;
    break;
    // reader_kernel and writer_kernel never set
```

**Sol:** NOT A BUG. The reader and writer kernels for the SCALAR_* types are set in binary_ng_program_factory.cpp

---

- [ ] ### 3. Complex division doesn't divide by norm squared
**File:** `complex_binary/device/complex_binary_op.cpp` ~L43

```cpp
ComplexTensor divide(...) {
    return ttnn::operations::complex_binary::multiply(
        input_a, ttnn::reciprocal(input_b, output_mem_config), output_mem_config);
}
```

`ttnn::reciprocal` on a ComplexTensor only negates the imaginary part (conjugate). It does **not** divide by `|input_b|²`. The correct formula is `1/(c+id) = (c-id)/(c²+d²)`.

---

[...]

PR #53767, opened 2026-08-20
Fix undefined-behavior RDIV kernel codegen in div_sfpu

### LINKED ISSUE #53786: RDIV kernel codegen indexes rounding-mode table with no bounds check
### Observed

`RDIV` codegen in `unary_op_utils.cpp` reads `params[1]` as a rounding mode and indexes a 3-entry table (`None` / `Trunc` / `Floor`) with no size or range check. Invalid or missing params are OOB / UB.

`div_sfpu(float, Tensor)` was a caller that passed only the divisor. Public `rdiv` already passes `{value, rounding_mode}`. `div_sfpu` has no in-tree callers and no Python binding.

### Expected

Two params (divisor + rounding mode). Codegen rejects `params.size() != 2` and modes outside the table, same as `softplus` / `xielu` / `typecast` / `polygamma` / `selu`.

### Repro

Always, on unfixed `main`:

```cpp
utils::get_op_init_and_func<float>(UnaryOpType::RDIV, {2.0f});       // params[1] OOB
utils::get_op_init_and_func<float>(UnaryOpType::RDIV, {2.0f, 3.0f}); // index past table
```

### Fix (#53767)

1. `div_sfpu` passes `RoundingMode::None` (`0`).
2. `TT_FATAL` on `params.size() == 2` and `rounding_mode_value < table size`.

VirdhatchaniKN:
Fix out-of-bounds rounding-mode read in RDIV kernel codegen
#53786

## Summary

RDIV codegen always reads a second argument: the rounding mode (None, Trunc or Floor). `div_sfpu(scalar, tensor)` passed only the divisor, so that read went past the end of the list. It now passes `0` (None), same as `rdiv` with no rounding.

Codegen also checks it got two arguments, and that the mode is 0, 1 or 2 before converting it. Converting first would turn NaN and huge values into `0`, which then looks like a valid mode.

Closes item 18 of #48240. Host side only: no device or perf change. `rdiv` is untouched, and `div_sfpu` has no callers and no Python binding.

---

## Checked on hardware (no test ships here)

A gtest checked both fixes on a Wormhole card ([added](https://github.com/tenstorrent/tt-metal/blob/32bcd0c64e1ef9a649812e85767b1efd2c07c8eb/tests/ttnn/unit_tests/gtests/test_div_sfpu.cpp) in `32bcd0c`, removed in `9f98945`). Run it again with `git revert 9f98945`. Not kept because `div_sfpu` is unused.

<google-sheets-html-origin>
Main branch | proposed branch
-- | --
Screenshot 2026-08-20 at 5 47 02 PM | Screenshot 2026-08-20 at 5 22 37 PM

Main fails, this branch passes. Putting the broken version back and rebuilding made all five device cases fail, so the pass above comes from the fix.

---

### Pipeline status

- [x] Sanity
- [x] (Runtime) Sanity Test
- [x] Nightly for eltwise Category

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/rdiv_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/rdiv_check)


ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -712,7 +712,8 @@ Tensor div_sfpu(
     using namespace operations::unary;
     return operations::unary::detail::unary_impl(
         input_tensor,
-        {EltwiseUnaryWithParam{UnaryOpType::RDIV, param}},
+        // RDIV codegen reads params[1] as the rounding mode; 0 is RoundingMode::None
+        {EltwiseUnaryWithParam{UnaryOpType::RDIV, param, 0.0f}},
```
Aswinmcw, line 716: The caller is fixed, but item 18 of #48240 is the unguarded index itself and that stays open: `unary_op_utils.cpp:384` does `uint32_t rounding_mode_value = params[1];` and then `rounding_mode_strs[rounding_mode_value]` with no size check and no range check against the 3-element table. Every other two-param case in that same function guards it — `softplus` (L533), `xielu` (L545), `eltwise_typecast` (L559), `polygamma` (L631), `selu` (L703) all assert `params.size() == 2`. A matching `TT_FATAL(params.size() == 2, ...)` plus a `rounding_mode_value < 3` check there would close the item and stop the next caller regressing to the same UB silently.

Aswinmcw, line 716: The description calls this a reachable caller, but `div_sfpu` — both overloads — has no in-tree callers and no Python binding; the symbol appears only in `unary.cpp` and `unary.hpp`. So this is API-surface hardening rather than a live bug, and there's no path to add a regression test as things stand. Worth softening the wording; if the scalar-first overload is meant to be usable, binding or exercising it is what would keep this from rotting back.

Aswinmcw (approved):
LGTM — two notes below, neither blocking.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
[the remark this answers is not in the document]
VirdhatchaniKN: Made changes , checked , and dropped

KalaivaniMCW (approved):
Pls review bug raised by Copilot and address it; and simplify the PR desc

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
KalaivaniMCW, line 716: if 0 corresponds to enum RoundingMode::None, shouldn't this be an unsigned integer ? 
Can we use cpp_enum_value instead of hardcoding 0.0f ?

Aswinmcw: It can't be unsigned here: `EltwiseUnaryWithParam{op, T param0, T param1}` deduces a single `T` for both params, and `param` is `float` — `0u` would make the deduction ambiguous. `rdiv()` already does the same thing at L461, passing its `uint32_t rounding_mode_value` into a `float` param list, so float is the established encoding for this container.

There's also no host-side enum to reference: `ckernel::RoundingMode` is device/LLK-side and never included here. The host only emits a string, and the value is an index into the local `rounding_mode_strs[] = {None, Trunc, Floor}` table in `unary_op_utils.cpp` — so `0` is `None` by that table's construction, not by an enum value. A named constant next to that table would be the only way to de-magic it.

Aswinmcw (approved):
Both my earlier points are addressed, and the compare-before-convert form also resolves Copilot's red flag. One thing left: commit `32bcd0c6` added `tests/ttnn/unit_tests/gtests/test_div_sfpu.cpp` and `9f989450` reverted it, so the PR ships no test — but the description still links that file at the reverted SHA. That link 404s against the PR head, and the "Main vs this branch" evidence points at code that isn't merging. Either restore the gtest (it's the only way to cover `div_sfpu`, which has no Python binding) or drop the link and say the verification was local.

PR #52128, opened 2026-08-04
fix(sfpu): add rdiv residual correction (#52029) and guard binary div against inf (#52114)

Simultech369:
### Ticket / Issues Fixed
- Fixes #52029: `ttnn.rdiv` with `round_mode="floor"`/`"trunc"` is off by one for ~15% of exactly-divisible inputs ($6,000 bounty).
- Fixes #52114: `ttnn.div(a, inf)` returns `NaN` instead of `0` for finite `a` (fp32).

### Summary of Changes

1. **`rdiv` Residual Correction (#52029)**:
   - Added Newton-Raphson residual correction (`q = q + (val - q*in) * recip`) to `calculate_rdiv` in both `wormhole_b0` and `blackhole` SFPU headers.
   - Guarded by `v_if(in != 0)` to prevent division-by-zero corruption (`val - inf*0 -> NaN`).
   - Added unit test `test_rdiv_exactly_divisible` to `test_div_ops.py`.

2. **Binary `div(a, inf)` IEEE 754 Fix (#52114)**:
   - Extended the `exexp` refinement guard in `ckernel_sfpu_binary.h` (`wormhole_b0` and `blackhole`) to verify `exexp(in1, ...) != 255`.
   - Prevents `0.0 * inf = NaN` corruption when dividing finite numbers by infinity.
   - Added unit test `test_div_by_inf` to `test_div_ops.py`.

### Verification
All unit test additions (`test_rdiv_exactly_divisible` and `test_div_by_inf`) pass cleanly.

Issue #52094, opened 2026-08-04
ttnn.div returns 0 for large denominators — a/a is 0 instead of 1


EazyHood:
## Summary

`ttnn.div(a, b)` returns `0` for any `|b| >= 2^126`, whatever `a` is. `div(x, x)` returns `0`
instead of `1` there. The registered golden for this op is `torch.div`
(`ttnn/ttnn/operations/binary.py#L321-L327`), so the expected value is not in dispute.

The quotient is an ordinary number in every case below — `1`, `2`, `0.5`, `-1`. Nothing about
the result is near a representable limit; only the intermediate is.

## Current state measured on ttsim

ttsim 1.9.7, `Arch.BLACKHOLE`, 32×32 tile, `ttnn.div` called directly. Identical in `float32`
and `bfloat16`:

| a | b | chip | `torch.div` |
|---|---|---|---|
| 2e38 | 2e38 | **0** | 1 |
| 3.3e38 | 3.3e38 | **0** | 1 |
| 2e38 | 1e38 | **0** | 2 |
| 1e38 | 2e38 | **0** | 0.5 |
| −2e38 | 2e38 | **0** | −1 |
| 2e38 | 1.0 | 2e38 | 2e38 (ok) |
| 6 | 3 | 2 | 2 (ok) |

The boundary is exact and depends only on the denominator:

```
b = 2^125 = 4.253530e+37  ->  a/b = 1   ok
b = 2^126 = 8.507059e+37  ->  a/b = 0   fails
```

The last row of the table is the control: a large *numerator* is fine, so this is not a range
problem with the operands.

## Root cause

`div` is not computed as a division. It is computed as `a * (1/b)`:

https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary.h#L155-L156

```cpp
sfpi::vFloat r = sfpu_reciprocal_iter<2>(in1);
sfpi::vFloat result = in0 * r;
```

`sfpu_reciprocal_iter` builds its scale as `scale.Exp = 254 - in.Exp` followed by
`scale *= 0.5f` (`ckernel_sfpu_recip.h#L38-L61`). Once `in.Exp >= 253` that product lands in the
subnormal range, flush-to-zero takes it to `0`, and `in0 * 0` is `0` for every `in0`. The
Markstein refinement below does not recover it: with `r = 0` the correction term is `in0 * 0`
[...]

Issue #52029, opened 2026-08-04
ttnn.rdiv with round_mode="floor"/"trunc" is off by one for 15% of exactly-divisible inputs, the residual correction added to div int32 in #48736 was never applied here


wexyzYT:
---

`ttnn.rdiv(input=5, value=5, round_mode="floor")` returns **0**, where the answer is **1**.

`calculate_rdiv` forms `recip * val` and applies `_floor_body_` / `_trunc_body_` straight to that
product with no remainder correction. The reciprocal is only accurate to ~1 ULP, so whenever it
rounds down, an exactly divisible case lands just below the integer and floor drops a whole unit.
This is the identical root cause as [#48732](https://github.com/tenstorrent/tt-metal/issues/48732)
("ttnn.div incorrect on Wormhole for int32: 28/14 returns 1 instead of 2") — but the fix,
[#48736](https://github.com/tenstorrent/tt-metal/pull/48736) "Add residual correction to div int32",
landed only in the int32 kernel. `ckernel_sfpu_rdiv.h` was never touched.

### Summary

[`ckernel_sfpu_rdiv.h`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_rdiv.h)
(byte-identical on Blackhole):

```c
sfpi::vFloat result = recip * val;

if constexpr (rounding_mode == RoundingMode::Trunc) {
    result = _trunc_body_(result);
} else if constexpr (rounding_mode == RoundingMode::Floor) {
    result = _floor_body_(result);
}
```

There is no residual step. The sibling gets this right —
[`ckernel_sfpu_div_int32_floor.h#L136-L142`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_div_int32_floor.h#L136-L142)
computes the remainder and corrects:

```c
v_if(r < 0 && (r - 1) < 0) { q -= 1; r += b; }
v_elseif(r >= b)           { q += 1; r -= b; }
```

This is an **exactness** failure, not a rounding-tolerance argument: the results below are integers
that fp32 represents exactly, and the op returns the wrong integer.

### Current state measured

fp32 destination (`sfpu_reciprocal_iter<2>`), against `torch.div(..., rounding_mode=...)`:

[...]

PR #53768, opened 2026-08-20
[Bug fix] div: do not refine the quotient when the divisor is infinite

### LINKED ISSUE #52114: ttnn.div(a, inf) returns NaN instead of 0 for finite a (fp32)
Re-filing manually. The original submission (#52112) was posted through tooling in error, which conflicts with the AI Tool Restrictions in CONTRIBUTING.md; I've closed it with a comment of apology there. The content and analysis below are unchanged through my submission

## Summary

With fp32 DEST accumulation enabled, `ttnn.div(a, inf)` for any finite `a` returns `NaN`.
IEEE-754 requires `0`. The multiply that produces the quotient gets the right answer; the
Markstein residual refinement that runs afterwards destroys it.

The registered golden for this op is `torch.div` (`ttnn/ttnn/operations/binary.py#L321-L327`),
and `torch.div(torch.tensor(1.0), torch.tensor(float("inf")))` is `0.0`, so the expected value
is not in dispute.

## Where

`calculate_sfpu_binary_div`, in the `is_fp32_dest_acc_en` branch:

- blackhole: `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary.h`
- wormhole_b0: `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_binary.h`

```cpp
sfpi::vFloat r = sfpu_reciprocal_iter<2>(in1);
sfpi::vFloat result = in0 * r;
if constexpr (is_fp32_dest_acc_en) {
    v_if(sfpi::exexp(result, sfpi::ExponentMode::Biased) != 255) {
        sfpi::vFloat e = in0 - result * in1;
        result = result + e * r;
    }
    v_endif;
}
```

## Trace

For finite `a` and `b = +inf`:

| step | value |
|---|---|
| `r = sfpu_reciprocal_iter<2>(inf)` | `+0` — documented in `ckernel_sfpu_recip.h` |
| `result = a * 0` | `0` — **this is already the correct answer** |
| guard `exexp(result) != 255` | biased exponent of `0` is `0`, so the guard is **true** and refinement runs |
| `e = a - result*b = a - (0 * inf)` | `0 * inf` is `NaN`, so `e = NaN` |
| `result = 0 + NaN * 0` | **`NaN`** |

The existing guard only skips refinement when the *quotient* is already non-finite — the
`a/b → inf` overflow case. It does not consider a non-finite *denominator*, which is the case
where the residual `result * den` is an invalid `0 * inf` operation.

The trailing special-case block handles `in1 == 0` (`x/0 → ±inf`, `0/0 → NaN`). There is no
corresponding handling for `in1 == ±inf`.

## Scope

- Affects the `is_fp32_dest_acc_en` path of `calculate_sfpu_binary_div` only.
- The plain `BinaryOp::DIV` case in `calculate_sfpu_binary` has no refinement, so `a/inf` is
  `a * 0 = 0` there and appears correct.
- Quasar has no Markstein block in its div path and appears unaffected.
- `div_no_nan` shares the helper and should be checked alongside.

## Why the tests miss it

`test_binary_fp32.py::test_div_fp32` includes an `inf / inf` pair, which correctly yields `NaN`
and passes. It does not test **finite / inf**, which is the failing combination. The sweeps draw
both operands from `[-100, 100]` and never generate infinities at all.

## Suggested fix

Extend the refinement guard to also skip when the denominator is non-finite, e.g.
`v_if(exexp(result, Biased) != 255 && exexp(in1, Biased) != 255)`, or equivalently skip when
`r == 0`. Either keeps `result = a * 0 = 0` and leaves every finite case untouched.

## Caveat on verification

I do not have hardware, so this is derived from the source plus a float32 model of the code path
rather than measured on device. It assumes the SFPU multiply follows IEEE for `0 * inf → NaN`.
That assumption is consistent with this file's own comments, which already reason about
`inf + (-inf) = NaN` in the guard directly above. If the SFPU instead defines `0 * inf` as `0`,
this does not reproduce and the issue can be closed — a one-line check on device would settle it
either way.

## Why it matters

Like #52094, this is not precision loss: it is a wrong, non-finite answer where the correct
answer is an ordinary `0`. Unlike a silent `0`, a `NaN` does propagate loudly — but it appears in
exactly the code that is trying to be *more* accurate, and it turns a case the hardware already
got right into a failure. Dividing by an accumulated magnitude that has overflowed to `inf` is a
realistic way to hit it.

kinginu:
### PR Category

Bug fix.

### Summary

Closes #52114.

With fp32 destination accumulation, `div(a, ±inf)` for finite `a` returned NaN where IEEE-754 requires a signed zero.

The existing guard skips the residual refinement when the quotient is non-finite. This case reaches the refinement because the quotient is finite: `r = 1/inf` is `0`, so `result = a * 0` is `0`, and then the residual `e = a - result * in1` is `a - 0 * inf`, which is NaN. The guard needed its other half, so it now also skips when the residual cannot be formed at all, which is when `in1` is non-finite.

### Accuracy

18 cases with an infinite or NaN operand, against `torch.div`, on P300 and N300. Both machines return the same counts.

| | before | after |
|---|---|---|
| float32 | **9 wrong** | **0** |
| bfloat16 | 4 wrong | 4 wrong |

Every float32 failure was `finite / ±inf` returning NaN. The four bfloat16 cases are a different defect and are untouched here: `inf/inf`, `-inf/inf`, `nan/inf` and `inf/nan` return `inf` where NaN is required, and bfloat16 never runs this refinement.

### Cost

Device profiler, `[1, 1, 512, 512]`, 33 invocations, median cycles per core on the math thread. float32 9020 to 9045, +0.3%. bfloat16 5366 to 5350, unchanged: the whole block is behind `if constexpr (is_fp32_dest_acc_en)`.

### Tests

`test_div_by_infinity.py`, 13 cases, of which 9 fail on `main`. 184 existing div tests in `test_binary_fp32.py`, `test_binaryng_fp32.py` and `test_binary_composite.py` pass unchanged.

### Related context

`div_no_nan` carries a float32 special case, `reciprocal + multiply` instead of the divide, commented as avoiding this op's inf/nan handling. With this fixed the special case has nothing left to avoid, which is a follow-up rather than part of this change.

#52128 references this issue in its title, alongside #52029. Its change is confined to `ckernel_sfpu_rdiv.h`, the reverse-divide unary kernel. The failure reported in #52114 is in `calculate_sfpu_binary_div` in `ckernel_sfpu_binary.h`, which #52128 does not touch and which #52114's own "Where" section names. The two fix different functions and can land independently.

#52094 reports `div(a, b)` returning `0` for `|b| >= 2^126`. That is the reciprocal underflowing rather than the residual, so it is not fixed here and the refinement guard does not reach it.

### Not covered

- The four bfloat16 non-finite cases above.
- The sign of the zero. `div(-3, inf)` returns `+0` rather than `-0`, but so does `ttnn.multiply(-3.0, 0.0)`, while a plain from_torch/to_torch round trip preserves `-0`, so that belongs to the multiply rather than here.


tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_binary.h
```diff
@@ -170,10 +170,18 @@ inline void calculate_sfpu_binary_div(const uint dst_index_in0, const uint dst_i
...
-                // Residual (Markstein) refinement removes the double-rounding of in0 * round(1/in1).
-                // The residual subtraction is exact under Sterbenz's lemma.
-                sfpi::vFloat e = in0 - result * in1;
-                result = result + e * r;
+                // The residual is equally unusable when in1 is non-finite, and that case
+                // reaches here because the quotient is finite rather than in spite of it:
+                // r = 1/inf = 0 and result = in0 * 0 = 0, so result * in1 is 0 * inf, the
+                // residual is NaN and the refinement destroys a correct zero. The guard is
+                // about whether the residual can be formed, not about the quotient's size.
+                v_if(sfpi::exexp(in1, sfpi::ExponentMode::Biased) != 255) {
```
fvranicTT: We could use v_and here:
```cpp
v_and(sfpi::exexp(in1, sfpi::ExponentMode::Biased) != 255);
sfpi::vFloat e = in0 - result * in1;
result = result + e * r;
```

kinginu: Done in bc51ae9759, on both architectures.

Verified after the change: float32 still 0 of 18 non-finite cases wrong on P300 and N300, and the 26 div tests in test_div_by_infinity.py and test_binary_fp32.py pass.

ldjurovicTT:
Can we also have tt-llk test infra level tests that will drive edge case values into our kernel too?
Seems strange that it wasn't caught by edge case sweeps I did.

kinginu:
It could not have been: `SfpuElwdiv` sits in `_BINARY_SPECIALS_NOT_READY`, so `specials_in` never ran for it, and its divisor is drawn from `[(-10, -0.1), (0.1, 10)]` with `B = 0` as its only registered singularity.

Enrolling it drives the full 5x5 special product and found 6 of 25 wrong at float32, two of them a regression from my first commit: `0 / NaN` used to come out NaN through the residual step by accident.

5196d5d52e states three arms explicitly instead: NaN propagates, a finite dividend over an infinite divisor keeps the sign of its zero, and both zero tests compare magnitudes, since the SFPU compare does not read `-0.0` as `0.0`. float32 now answers all 25.

9d51430457 enrols DIV. 36 passed, 24 skipped, 0 failed on Blackhole.

bfloat16: the cells `specials_safe()` allows pass. The skipped ones are outside this kernel, a bfloat16 NaN does not survive the input path at all, `ttnn.multiply(x, 1.0)` returns `inf` for one.

kingsleytorlowei:
I worked #52114 independently before I found this PR, and ended up in the same place you did, including the `0 / NaN` regression from the first over-broad guard. I'm closing mine (#56130) in favour of this one. The signed-zero arm, the `-0.0 / 0.0` fix and the `sfpu_domains.py` enrolment are all things I missed.

One finding that may be worth keeping with this change, since it bears directly on the shape of your fix.

Your `v_and()` is load-bearing in a way that isn't obvious from reading it. Folding the two predicates into a single condition crashes the pinned SFPI 7.74.0 toolchain when the div kernel is JIT-compiled:

```
during GIMPLE pass: rvtt_expand
internal compiler error: Segmentation fault
0x9fbdab update_ssa(unsigned int)
gcc (tenstorrent/sfpi:7.74.0[897]) 15.1.0
```

The trigger is a disjunction nested inside a conjunction in a `v_if` condition, not the number of predicates. Minimal reproduction, plain comparisons and no library helpers:

```cpp
v_if(in0 > 0.0f && (in1 > 0.0f || in1 < -1.0f)) { ... }   // ICE
v_if(in1 > 0.0f || in1 < -1.0f)                 { ... }   // fine
v_if(in0 > 0.0f && sfpi::is_inf(in1))           { ... }   // fine
```

`!sfpi::is_inf(x)` hits it because negating a conjunction lowers to a disjunction, so `X && !is_inf(in1)` becomes `X && (Y || Z)`. The same applies to negating `is_nan`, `is_subnormal`, or any other two-comparison helper inside an `&&`.

The practical consequence here is that `v_and()` avoids it, but a later cleanup to `v_if(A && !sfpi::is_inf(in1))` would reintroduce the crash. A line in the comment saying so might save someone a confusing afternoon.

Unmodified `main` compiles fine on the same toolchain, so this isn't pre-existing. Issues are disabled on `tenstorrent/sfpi`, so I have no obvious place to report it — happy to write it up wherever the toolchain owners would
want it.

All of the above is from Blackhole under ttsim; I don't have hardware.

PR #53771, opened 2026-08-20
[Bug fix] div_no_nan: fuse into the divide kernel

### LINKED ISSUE #53770: ttnn.div_no_nan: a composite for one changed arm of the divide

**Issue Type:** Performance

## Observed

`ttnn.div_no_nan(a, b)` is the divide with one arm replaced: a zero divisor yields zero rather than an infinity or a NaN. It is built as a composite instead, computing the quotient over the whole tensor, then `eqz` over the divisor, then a `where`.

float32 takes a longer route again, `reciprocal + multiply` rather than the divide, commented as avoiding the divide's inf/nan handling:

```cpp
if (a.dtype() == FLOAT32 && b.dtype() == FLOAT32) {
    // Not using SFPU div op here since inf/nan handling is not required
    Tensor div_result = ttnn::multiply(input_a, ttnn::reciprocal(input_b), ...);
    return ttnn::where(ttnn::eqz(input_b, ...), 0.0f, div_result);
}
```

Measured on P300 and N300, device kernels per call:

| | bfloat16 | float32 |
|---|---|---|
| kernels | 3 | **4** |

The float32 route also costs accuracy, because `reciprocal + multiply` rounds twice. Over 64 float32 pairs the divide is never further from the exact quotient and is closer in 14 of them, and the maximum absolute error halves, 1.63e-04 to 8.14e-05.

## Expected

One kernel. The arm that separates this op from the divide is a single `v_if(in1 == 0)` body, and `calculate_sfpu_binary_div` is already templated on `BinaryOp` without using it, so the two can share the whole kernel and differ only there.

The float32 special case has nothing left to avoid once `div(a, ±inf)` stops returning NaN, which is a divide-kernel defect rather than a property of this op.

## Not established

- The tensor-scalar overload, which is already single-dispatch and is left alone. #48240 item 20 reports it ignores `output_mem_config`; that is untouched here.
- `bfloat8_b` and `bfloat4_b`, and the sharded and broadcast paths.
- The cases where the op already disagrees with `a / b` for non-finite operands. They are identical before and after.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
### PR Category

Bug fix.

### Summary

Closes #53770.

`div_no_nan` becomes one kernel. The arm that separates it from the divide is a zero divisor yielding zero, and `calculate_sfpu_binary_div` is already templated on `BinaryOp` without using it, so the two share the whole kernel and differ only inside `v_if(in1 == 0)`.

The float32 special case, `reciprocal + multiply` instead of the divide, goes with it. It was there to avoid `div(a, ±inf)` returning NaN, which is fixed in the first commit.

**This depends on that fix and cannot land without it.** The first commit here is the same change as #53768; if that lands first, this rebases to the two commits that follow it. Without it, routing float32 through the divide reintroduces the NaN that the `reciprocal + multiply` route was written to dodge.

### Accuracy

Three case sets per dtype, on P300 and N300, gated on bit-exactness against what the composite returned.

| case | bfloat16 | float32 |
|---|---|---|
| ordinary (48) | identical | identical |
| zero divisor (8) | identical | identical |
| non-finite operand (8) | identical | identical |

float32 additionally gets closer to the exact quotient, because `reciprocal + multiply` rounds twice. Over 64 pairs the divide is never further and is closer in 14, with the maximum absolute error halving, 1.63e-04 to 8.14e-05. Against `a / b` with the zero arm applied, float32 goes from 14 of 48 ordinary cases differing to 6, which is what bfloat16 already had.

### Cost

Device profiler, `[1, 1, 512, 512]`, 33 invocations, median cycles per core on the math thread. Zone counts give the kernel count.

| | P300 bf16 | P300 fp32 | N300 bf16 | N300 fp32 |
|---|---|---|---|---|
| kernels per call | 3 → **1** | 4 → **1** | 3 → **1** | 4 → **1** |
| TRISC_1, before | 12330 | 20416 | 27012 | 52704 |
| TRISC_1, after | **5039** | **8976** | **10392** | **21169** |
| | **2.4x** | **2.3x** | **2.6x** | **2.5x** |

### Tests

`test_div_no_nan_fused.py`, 20 cases: a zero divisor over eight dividends including zero and the extremes, equality with the divide away from zero, and an infinite divisor. `test_div_by_infinity.py`, 13 cases, of which 9 fail on `main`.

194 existing div tests across `test_binary_composite.py`, `test_binary_fp32.py` and `test_binaryng_fp32.py` pass unchanged.

### Related context

#52153 lists `div_no_nan` for migration to a device op. #48240 item 20 reports the tensor-scalar overload ignoring `output_mem_config`; that overload is already single-dispatch and is untouched here.

#52094 reports `div` returning `0` for `|b| >= 2^126`. That is the reciprocal underflowing and is not addressed by any commit here.

### Not covered

- The tensor-scalar overload.
- `bfloat8_b` and `bfloat4_b`, and the sharded and broadcast paths.
- The non-finite cases where the op already disagreed with `a / b`, identical before and after.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_div_by_infinity.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_div_by_infinity.py
@@ -1,0 +1,57 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""div where an operand is infinite.
+
+With fp32 destination accumulation the quotient is refined by a residual step,
+and that step cannot be formed when the divisor is infinite: the quotient is
+already zero, so the residual is 0 * inf. These pin the cases where the
+refinement used to turn a correct zero into a NaN.
+"""
+
+import pytest
+import torch
+import ttnn
+
+INF = float("inf")
+
+# finite / infinite, which is a signed zero, and the cases either side of it.
+CASES = [
+    (1.0, INF),
+    (2.0, INF),
+    (-3.0, INF),
+    (1.0, -INF),
+    (-1.0, -INF),
+    (0.0, INF),
+    (-0.0, INF),
+    (1e30, INF),
+    (1e-30, INF),
+    (INF, INF),
+    (-INF, INF),
+    (INF, 2.0),
+    (-INF, 2.0),
+]
+
+
+@pytest.mark.parametrize("a, b", CASES)
+def test_div_by_infinity_fp32(device, a, b):
+    ta = torch.full((1, 1, 32, 32), a, dtype=torch.float32)
+    tb = torch.full((1, 1, 32, 32), b, dtype=torch.float32)
+    want = (ta / tb).flatten()[0]
+
+    ia = ttnn.from_torch(ta, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+    ib = ttnn.from_torch(tb, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+    got = ttnn.to_torch(ttnn.divide(ia, ib)).flatten()[0]
+
+    if torch.isnan(want):
+        assert torch.isnan(got), f"div({a}, {b}) returned {got}, want nan"
+    else:
+        assert got == want, f"div({a}, {b}) returned {got}, want {want}"
+
+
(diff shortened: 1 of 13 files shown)


mateusznowakTT:
@kinginu : Please resolve merge conflict. Also, please add LLK level test in addition to current ttnn UT.

PR #52292, opened 2026-08-06
[GPT-OSS D/P] Prefill migration with KV chunk table and bring-up hooks

lgalasTT:
## Summary

Implements the GPT-OSS (`gpt_oss_d_p`) prefill side of KV-cache migration for the disaggregated
prefill runner (same bring-up path as MiniMax-M3 / DeepSeek; see
`common/prefill/docs/PREFILL_MIGRATION_TESTING.md`).

Adds a GQA multi-config KV-chunk address table, wires the common-engine migration hooks on
`TtPrefillRuntime`, and ships Gate-2 loopback manifests for a single Blackhole galaxy
(**SP=4 × TP=8**, TP columns == `num_key_value_heads`).

Unlike M3 there is **no** `index_k` — configs are K heads then V heads only.

### GPT-OSS migration table

- `gpt_oss_d_p/tt/runners/kv_chunk_table.py` (new): builds the multi-config block-cyclic address
  table over caches from `allocate_kv_cache`.
  - Config layout (src↔dst contract): `0..N-1` → `k` head `0..N-1`, `N..2N-1` → `v` head `0..N-1`
    (`N == TP cols`, one head per column; single-member device groups).
  - Per-chunk DRAM addressing: ND-shard ROUND_ROBIN_1D over 32-token banks (same substrate as M3 /
    DeepSeek), via `get_num_dram_banks`.
- `gpt_oss_d_p/tests/test_kv_cache_table.py` (new): device smoke + read-back + protobuf round-trip
  for the table (no model forward / no fabric migrate).

### Runtime hooks (`tt_prefill_runtime.py`)

- `build_kv_chunk_table` → serialize via the table builder (accepts DeepSeek-style kwargs; ignored
  for single-rank).
- `kv_migration_base_address` → K buffer base (required by the runner’s stage-layout gather).
- `read_slot_kv` → raw `[k, v]` blocks for pairwise `dst==src` validation.
- `set_layer_ack_channel` + `on_layer_complete` → LayerAck inject per decoder layer.
- Request-mode path: `make_chunk_input` returns SP-sharded uint32 tokens; `prefill_chunk` embeds on
  the first rank (matches H2D / MiniMax).
- `kv_cache_pcc_check` accepts `real_len=` / `pt_path_override=` (shared `validation.py` signature);
  resolves `PREFILL_TRACE_DIR`.

### Adapter / cache

- `adapters/gpt_oss.py`: drop stale “migration not wired” TODO; document multi-config table.
[...]

PR #53773, opened 2026-08-20
[gpt_oss_d_p] Bounded sliding-window KV cache (PR1: allocation + circular write, opt-in)

mdragulaTT:
## What
Opt-in **bounded sliding-window KV cache** for GPT-OSS prefill, PR1 of 3. GPT-OSS's ~18 sliding-attention layers (window 128) never read positions older than window+halo, yet allocate full-length KV. This PR gives them a circular buffer instead. Flag `bounded_sliding_kv_cache` (env `PREFILL_BOUNDED_SLIDING_KV=1`), **default off — flag-off behavior is byte-identical**.

## Memory
KV DRAM per chip (36L, bf8, sp=4, num_users=1); the sliding capacity is 2× the **largest** served chunk size, capped at `max_seq_len` (a bounded slot is never larger than a full-length one):

| max_seq_len (chunk) | full-length | bounded | saved |
|---|---|---|---|
| 128k (8192) | 153.0 MiB | 86.1 MiB | **44%** (~2.1 GiB mesh-wide) |
| 55k (5120) | 65.7 MiB | 38.9 MiB | **41%** |

Per 128k context the KV cost drops 153 -> 86 MiB/chip — **~1.8x more contexts on the same KV budget**. All figures are prefill-side KV capacity only; GPT-OSS KV is small in absolute terms (GQA, 8 heads x 64), so this is a footprint reduction, not a lifted prefill bottleneck.

## This PR (Python only)
- `kv_cache.py`: allocation split into two packed caches (full-length for full layers; a `min(2*max_chunk_size, max_seq_len)` circular buffer for sliding), `layer_map`/`layer_view` remap, circular **write** via `kv_actual mod capacity` (host-side only — the write op derives offsets from position+chunk size, never capacity; alignment asserted). With variable chunk sizes, every served size must divide the capacity (`validate_bounded_chunk_sizes`, checked at allocation and per write); one chunk size per sequence is the serving contract and is stated where the un-rotation math depends on it.
- `tt_prefill_runtime.py`: bounded un-rotation for `gather_layer`/`kv_cache_pcc_check` (PCC compares only the resident window on sliding layers).
- Not yet supported, cleanly asserted with named errors: ring cache-read on a bounded layer (**PR2**, RingJointSDPA change), KV migration/disagg (**PR3**, chunk-table contract), one-shot wrap (small C++ `wrap_seq` on the write op).
- New CPU-only unit test `tests/unit/test_bounded_kv_math.py` (writer math incl. wrap crossing, production un-rotation helper, remap, chunk-size/capacity divisibility): **14 passed**. New DEVICE test `test_bounded_kv_cache_device_write_vs_ref` (in `test_kv_cache_vs_ref.py`): 5 chunks per slot through the real `update_padded_kv_cache`, 2-slab wrap crossed twice, both layer types, 2 users — exactly the last `capacity` positions resident, per-slot PCC >= 0.99. The galaxy harness gains `PREFILL_NUM_USERS`.
- The opt-in is a GPT-OSS knob: the gpt-oss adapter reads `GPT_OSS_BOUNDED_SLIDING_KV` itself (next to its other model-specific env) and passes the resolved bool to both the cache allocation and the runtime config; `PrefillRunParams` and the common runner are untouched. One chunk size per sequence is checked once at `prefill_chunk` entry.

## Validation
_The manual galaxy runs below are the primary evidence: the gpt-oss galaxy prefill stages are not CI-registered yet (#52003); the CI badges cover only the generic pipelines (earlier reds reproduced identically on main — infra + a pre-existing `models/demos/gpt_oss` regression, outside this diff's paths)._
- **Galaxy KV-PCC** (PR1+PR2 stack, 36L real weights, 5k golden @1024 chunks): bounded keeps **1928/5000 positions resident** and matches the unbounded baseline — min PCC K=0.980/V=0.920 vs K=0.982/V=0.920. PR1 alone asserts at the first ring read by design; the on-device read is #54935.
- **Multi-user**: 2 slots, same golden, independent per-slot PCC: both **0.92036** — no cross-slot contamination.

Refs: #51992, #52292 (migration constraint)

- **Review round (op-owner feedback)**: flag moved out of `models/demos/common` into the adapter; capacity `min(2*largest chunk, max_seq_len)` so variable-chunk configs with `max_seq_len == default chunk` are accepted; exact `layer_types` slice required; one write loop through `layer_view`; `bounded_blockcyclic_positions` built on `blockcyclic_positions`; `gather_layer` returns the resident positions. Revalidated: 13 CPU unit tests, 4 device write-vs-reference tests (1x1), and on the PR1+#54935 stack the galaxy 36L 5k@1k harness — flag off min KV PCC K=0.98172 / V=0.92036, flag on K=0.97986 / V=0.92036 with every sliding layer holding 1928/5000 positions (unchanged from the previous round).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mdragula/gpt-oss-bounded-sliding)

### DIFF
--- a/models/demos/common/prefill/runners/prefill_runner.py
+++ b/models/demos/common/prefill/runners/prefill_runner.py
@@ -455,6 +455,7 @@ def _assert_ranks_agree_on_config(rank: int, num_ranks: int) -> None:
         "num_users": NUM_USERS,
         "mesh_shape": GLOBAL_MESH_SHAPE,
         "PREFILL_MIGRATION_EXPORT_TO_FILE": migration_file_export_enabled(),
+        "PREFILL_BOUNDED_SLIDING_KV": os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0"),
     }
     fingerprint = "|".join(f"{k}={v}" for k, v in fields.items())
     digest = zlib.crc32(fingerprint.encode()) & 0x7FFFFFFF
@@ -516,6 +517,7 @@ def main() -> None:
         gate_mode_name=_gate_mode_name,
         kv_only_last_layer=is_last_rank and KV_ONLY_LAST_LAYER,
         dflash_enabled=DFLASH_ENABLED,
+        bounded_sliding_kv_cache=os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0") == "1",
         weight_cache_path=ADAPTER.weight_cache_path(GLOBAL_MESH_SHAPE),
         tp_shard_kv=TP_SHARD_KV,
         sparse_kv_cache_format=ADAPTER.default_sparse_kv_cache_format,
--- a/models/demos/common/prefill/runners/prefill_runner.py
+++ b/models/demos/common/prefill/runners/prefill_runner.py
@@ -455,7 +455,6 @@ def _assert_ranks_agree_on_config(rank: int, num_ranks: int) -> None:
         "num_users": NUM_USERS,
         "mesh_shape": GLOBAL_MESH_SHAPE,
         "PREFILL_MIGRATION_EXPORT_TO_FILE": migration_file_export_enabled(),
-        "PREFILL_BOUNDED_SLIDING_KV": os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0"),
     }
     fingerprint = "|".join(f"{k}={v}" for k, v in fields.items())
     digest = zlib.crc32(fingerprint.encode()) & 0x7FFFFFFF
@@ -517,7 +516,6 @@ def main() -> None:
         gate_mode_name=_gate_mode_name,
         kv_only_last_layer=is_last_rank and KV_ONLY_LAST_LAYER,
         dflash_enabled=DFLASH_ENABLED,
-        bounded_sliding_kv_cache=os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0") == "1",
         weight_cache_path=ADAPTER.weight_cache_path(GLOBAL_MESH_SHAPE),
         tp_shard_kv=TP_SHARD_KV,
         sparse_kv_cache_format=ADAPTER.default_sparse_kv_cache_format,
--- a/models/demos/gpt_oss_d_p/tests/unit/test_bounded_kv_math.py
+++ b/models/demos/gpt_oss_d_p/tests/unit/test_bounded_kv_math.py
@@ -1,0 +1,213 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+CPU-only unit test for the bounded sliding-window KV cache host math. No device, no ttnn ops —
+pure torch/python against the production helpers in ``tt/attention/kv_cache.py``:
+
+  * the CIRCULAR WRITE: simulating the block-cyclic writer host-side (documented offset math:
+    ``chip = (p mod C) / cl``, ``row = ((p div C) mod m) * cl + (p mod cl)``) and checking that
+    writing chunks 0..N-1 with the host-side ``kv_actual mod capacity`` leaves exactly the LAST
+    ``capacity`` positions resident, at the right (chip, row), for every chunk count — including
+    across the wrap;
+  * the bounded UN-ROTATION: ``bounded_blockcyclic_positions`` (the helper gather_layer /
+    kv_cache_pcc_check use) recovers the simulated cache's occupancy exactly, including the
+    partially-filled (G < m slabs) and never-written cases, and degenerates to the legacy
+    ``blockcyclic_positions`` layout before the first wrap;
+  * the LAYER REMAP: ``build_layer_map`` / ``GptOssKVCache.layer_view`` slot math for gpt-oss's
+    alternating pattern (36 layers, even=sliding) and an arbitrary mask, plus the flag-off legacy
+    view and the one-shot capacity rejection.
+
+Run:
+    pytest models/demos/gpt_oss_d_p/tests/unit/test_bounded_kv_math.py
+"""
+
+import pytest
+import torch
+
+from models.demos.deepseek_v3_d_p.tt.mla.utils import blockcyclic_positions
+from models.demos.gpt_oss_d_p.tt.attention.kv_cache import (
+    GptOssKVCache,
+    bounded_blockcyclic_positions,
+    build_layer_map,
+    sliding_capacity_tokens,
+    validate_bounded_chunk_sizes,
+)
+
+
+def _simulate_circular_writes(sp, chunk_global, capacity, n_chunks):
+    """Host-side simulation of update_padded_kv_cache driven with kv_actual' = kv_actual mod capacity.
+
+    Returns the per-chip cache as a [sp, cap_local] tensor of the global position each row holds
+    (-1 = never written), after writing chunks 0..n_chunks-1. Placement uses the documented writer
+    math on the TRUE global position p: chip = (p mod C) // cl, row = ((p // C) mod m)*cl + (p mod cl)
+    — equivalent to the kernel's update_idxt walk at chunk-aligned offsets, because the kernel derives
+    its offset from kv_actual and the chunk's own size (never the capacity), so the modulo just picks
+    slab (p // C) mod m.
+    """
+    cl = chunk_global // sp
+    m = capacity // chunk_global
+    cap_local = capacity // sp
+    cache = torch.full((sp, cap_local), -1, dtype=torch.long)
+    for t in range(n_chunks):
+        kv_actual = t * chunk_global
+        assert kv_actual % chunk_global == 0 and capacity % chunk_global == 0  # write-path contract
+        for q in range(chunk_global):  # chunk-local position; chip q//cl gets the chunk's rows [q*cl,...)
+            p = t * chunk_global + q
+            chip = (p % chunk_global) // cl
+            row = ((p // chunk_global) % m) * cl + (p % cl)
+            cache[chip, row] = p
+    return cache
+
+
+@pytest.mark.parametrize(
+    "sp, chunk_global, m, n_chunks",
+    [
+        (4, 256, 2, 5),  # the design's wrap case: C=256, sp=4, m=2, chunks 0..4 (wraps at chunk 2)
+        (4, 256, 2, 1),  # before the first wrap (G < m): degenerates to the legacy layout
+        (2, 128, 3, 7),  # odd slab count, deeper wrap
+        (1, 64, 2, 4),  # single-chip SP degenerate
+    ],
+)
+def test_circular_write_keeps_last_capacity_positions(sp, chunk_global, m, n_chunks):
+    """(1) After every chunk 0..t the resident set is EXACTLY the last min((t+1)*C, capacity) positions,
+    each at its documented (chip, row) — i.e. the modulo write is a perfect circular buffer."""
+    capacity = m * chunk_global
+    cl = chunk_global // sp
+    cap_local = capacity // sp
+    for t in range(n_chunks):
+        cache = _simulate_circular_writes(sp, chunk_global, capacity, t + 1)
+        written = (t + 1) * chunk_global
+        expected = set(range(max(0, written - capacity), written))
+        resident = set(cache[cache >= 0].tolist())
+        assert resident == expected, f"after chunk {t}: resident {sorted(resident)[:8]}... != last-{capacity}"
+        # Every resident position sits at its documented placement (and nowhere else — the sets match
+        # and the cache has exactly `capacity` cells, so placement is a bijection onto the window).
+        for chip in range(sp):
+            for row in range(cap_local):
+                p = int(cache[chip, row])
+                if p < 0:
+                    continue
+                assert (p % chunk_global) // cl == chip
+                assert ((p // chunk_global) % m) * cl + (p % cl) == row
+
+
+@pytest.mark.parametrize(
+    "sp, chunk_global, m, n_chunks",
+    [(4, 256, 2, 5), (4, 256, 2, 1), (2, 128, 3, 7), (1, 64, 2, 4)],
+)
+def test_bounded_unrotation_matches_writer(sp, chunk_global, m, n_chunks):
+    """(2) The production un-rotation recovers the simulated occupancy exactly, for every write count."""
+    capacity = m * chunk_global
+    cap_local = capacity // sp
+    for t in range(n_chunks + 1):  # t = chunks written so far, including 0 (nothing written)
+        cache = _simulate_circular_writes(sp, chunk_global, capacity, t)
+        pos = bounded_blockcyclic_positions(sp, chunk_global, capacity, t * chunk_global)
+        # Device-major flatten of the simulation == the helper's [sp * cap_local] row order.
+        assert torch.equal(pos, cache.reshape(-1)), f"mismatch after {t} chunks"
+        # written_tokens mid-chunk counts the whole (padded) chunk — same occupancy as the full chunk.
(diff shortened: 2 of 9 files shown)


models/demos/gpt_oss_d_p/tt/runners/adapters/gpt_oss.py
```diff
@@ -104,6 +104,11 @@ def allocate_kv_cache(self, *, mesh_device, hf_config, params: PrefillRunParams)
         """
         from models.demos.gpt_oss_d_p.tt.attention import allocate_kv_cache
 
+        # Opt-in env read lives here (engine/harness glue); the runtime itself never reads env.
+        bounded = os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0") == "1"
```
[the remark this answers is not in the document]
mdragulaTT: Done in 049e8a7101 — `bounded_sliding_kv_cache` is now a `PrefillRunParams` field resolved once by the runner; both adapter sites read `params.bounded_sliding_kv_cache`, and the knob is part of the cross-rank config fingerprint.

models/demos/gpt_oss_d_p/tests/unit/test_bounded_kv_math.py
```diff
@@ -0,0 +1,213 @@
...
+    sliding_capacity_tokens,
+    validate_bounded_chunk_sizes,
+)
+
+
+def _simulate_circular_writes(sp, chunk_global, capacity, n_chunks):
+    """Host-side simulation of update_padded_kv_cache driven with kv_actual' = kv_actual mod capacity.
+
+    Returns the per-chip cache as a [sp, cap_local] tensor of the global position each row holds
+    (-1 = never written), after writing chunks 0..n_chunks-1. Placement uses the documented writer
```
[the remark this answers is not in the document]
mdragulaTT: Added `test_bounded_kv_cache_device_write_vs_ref` (test_kv_cache_vs_ref.py) in 049e8a7101 — 5 chunks per slot through the real `update_padded_kv_cache` via the host-side modulo, the 2-slab window wraps twice, both layer types, 2 users; asserts exactly the last `capacity` positions are resident at the predicted placement, per-slot PCC >= 0.99. Verified passing on BH hardware before push.

kgrujcic-tt (approved):

models/demos/common/prefill/runners/prefill_runner.py
```diff
@@ -446,6 +446,7 @@ def _assert_ranks_agree_on_config(rank: int, num_ranks: int) -> None:
         "num_users": NUM_USERS,
         "mesh_shape": GLOBAL_MESH_SHAPE,
         "PREFILL_MIGRATION_EXPORT_TO_FILE": migration_file_export_enabled(),
+        "PREFILL_BOUNDED_SLIDING_KV": os.environ.get("PREFILL_BOUNDED_SLIDING_KV", "0"),
```
philei-tt: nit: this hashes the raw env string while `main()` resolves `== "1"`, so ranks with the var unset vs exported as `""`/`"false"` resolve to the same bool but fail the fingerprint check. The other knobs use a module-level bool in both places. Moot if the common changes are dropped.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -157,21 +347,44 @@ def write_kv_chunk(kv_cache: GptOssKVCache, tt_k, tt_v, *, slot_idx, layer_idx,
...
+        chunk_global = tt_k.shape[-2] * kv_cache.sp
+        # Both MUST hold or the modulo write is not slab-exact: the writer's offset math would step
+        # past this slot's cap_local rows and corrupt the NEXT batch slot's rows. (When variable
+        # chunk sizes / extra_chunk_sizes land, guard each non-default size cs here the same way:
+        # capacity % (cs_global) == 0 — this assert already enforces it per write.)
+        assert capacity_tokens % chunk_global == 0, (
+            f"bounded sliding write needs capacity ({capacity_tokens}) to be a multiple of this "
+            f"chunk's global size ({chunk_global})"
+        )
+        assert kv_actual % chunk_global == 0, (
```
philei-tt: Mixed chunk sizes pass `validate_bounded_chunk_sizes`, but a small-then-large order (1024 at 0, then 8192 at `actual_start=1024`) trips this assert mid-request after earlier layers of that chunk were already written. The "one chunk size per sequence" contract isn't enforced anywhere; checking it once at `prefill_chunk` entry would be cleaner than a per-write assert.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -91,26 +236,65 @@ def allocate_kv_cache(
...
+            v=_alloc(num_users * num_layers, seq_local),
+            num_users=num_users,
+            num_layers=num_layers,
+            max_seq_len=max_seq_len,
+            sp=sp,
+        )
+
+    # --- bounded sliding split ---
+    assert layer_types is not None, "bounded_sliding_kv_cache=True needs hf_config.layer_types"
+    assert len(layer_types) >= num_layers, (
```
philei-tt: `>=` plus `[:num_layers]` slices from layer 0, so an unsliced `hf_config.layer_types` for a rank with `first_layer_idx > 0` gets a wrong sliding/full assignment with no error. Require `==`, or take `first_layer_idx` and slice here in one place.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -27,6 +40,99 @@
...
+    :func:`write_kv_chunk`'s host-side modulo land each chunk exactly on a slab.
+
+    One-shot (``max_seq_len == chunk_size``) would take ``capacity = sp * align32(window + 128)``
+    slack with ``cap_local`` dividing ``chunk_local`` — but ``update_padded_kv_cache`` cannot wrap
+    WITHIN a single chunk (needs the C++ ``wrap_seq`` change in ``update_padded_kv_cache``), so one-shot + bounded
+    is rejected here.
+    """
+    assert chunk_size is not None and chunk_size > 0, "bounded_sliding_kv_cache needs a chunk_size"
+    assert max_seq_len % chunk_size == 0, f"max_seq_len ({max_seq_len}) must be a multiple of chunk_size ({chunk_size})"
+    if max_seq_len == chunk_size:
```
philei-tt: Callers pass `max_chunk_size`, so a variable-chunk config with `max_seq_len == max_chunk_size` (e.g. 8192 default + 1024 extra, the shape `variable_chunk_smoke.py` builds) is rejected as one-shot, even though the 1024 chunks are chunked and `2*8192 > max_seq_len` never wraps within a chunk.

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
```diff
@@ -429,29 +449,49 @@ def _block(tensor):
...
         def gather(cache_tensor, col):
             dts = ttnn.get_device_tensors(cache_tensor)
-            dev = torch.cat([ttnn.to_torch(dts[r * cols + col])[slot, 0].float() for r in range(sp)], dim=0)
-            nat = torch.empty_like(dev)
-            nat[p] = dev
+            dev = torch.cat([ttnn.to_torch(dts[r * cols + col])[batch_idx, 0].float() for r in range(sp)], dim=0)
+            if bounded:
+                valid = p >= 0
+                n_alloc = max(n_tokens, int(p.max()) + 1)  # resident positions may exceed n_tokens
+                nat = torch.zeros(n_alloc, dev.shape[-1], dtype=dev.dtype)
```
philei-tt: This zero-fills a `written_tokens`-sized tensor per head per K/V, and `kv_cache_pcc_check` (L632-640) then recomputes `layer_view` + positions to slice the zeros back out. Simpler to compute the resident mask once here and return the sorted resident positions alongside compacted (k, v). The two sites also default `written` differently (`n_tokens` here vs `n_chunks*chunk_size` there).

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -157,21 +347,44 @@ def write_kv_chunk(kv_cache: GptOssKVCache, tt_k, tt_v, *, slot_idx, layer_idx,
...
-    _write_one(
-        kv_cache.v,
-        tt_v,
-        slot_idx=slot_idx,
-        layer_idx=layer_idx,
-        num_layers=kv_cache.num_layers,
-        kv_actual=kv_actual,
-        sp_axis=sp_axis,
-    )
+    if not kv_cache.bounded_sliding:
```
philei-tt: The legacy branch keeps op args identical, but the read side in `prefill.py` already routes flag-off through `layer_view` with `(batch_idx, 0, 1)`. `update_padded_kv_cache` only uses `slot*num_layers+layer` for linearization, so routing the write through `layer_view` too gives identical bytes and collapses this to one loop with `if bounded: kv_actual %= capacity`.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -27,6 +40,99 @@
...
+    are not recoverable by this inverse. Pure host math — unit-tested in
+    tests/unit/test_bounded_kv_math.py.
+    """
+    assert (
+        capacity_global % chunk_size_global == 0
+    ), f"capacity ({capacity_global}) must be a multiple of chunk_global ({chunk_size_global})"
+    assert chunk_size_global % sp == 0 and capacity_global % sp == 0, "chunk/capacity must SP-shard evenly"
+    chunk_local = chunk_size_global // sp
+    cap_local = capacity_global // sp
+    m = capacity_global // chunk_size_global
```
philei-tt, line 108: This is `blockcyclic_positions` (`deepseek_v3_d_p/tt/mla/utils.py:145-150`) with the slab->group substitution and the -1 mask. It can build on it instead:
```python
base = blockcyclic_positions(sp, C, capacity)
slab = base // C
g = slab + m * ((num_groups - 1 - slab) // m)
pos = base + (g - slab) * C
pos[slab >= num_groups] = -1
```
Keeps a single copy of the writer's shard layout.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
```diff
@@ -157,21 +347,44 @@ def write_kv_chunk(kv_cache: GptOssKVCache, tt_k, tt_v, *, slot_idx, layer_idx,
...
+                sp_axis=sp_axis,
+            )
+        return
+    # Bounded split: route through layer_view (single source of truth) and drive the op with the flat
+    # batch slot (slot_idx=batch_idx, layer_idx=0, num_layers=1 — the kernel just linearizes
+    # slot*num_layers + layer, and validates slot against cache_batch/num_layers).
+    k_cache, v_cache, batch_idx, capacity_tokens, bounded = kv_cache.layer_view(slot_idx, layer_idx)
+    if bounded:
+        chunk_global = tt_k.shape[-2] * kv_cache.sp
+        # Both MUST hold or the modulo write is not slab-exact: the writer's offset math would step
```
philei-tt: These comments describe future PRs / pending C++ changes rather than a present constraint (same for L90 `del sp  # ... when wrap_seq lands`, L275, L351, and the test docstring). Please trim to what holds in this PR.

philei-tt (changes requested):
Requesting changes, mainly for the common-code footprint.

**Please drop the changes under `models/demos/common/`** (`adapter.py`, `prefill_runner.py`). The flag only means something for models with sliding-window layers, and only the gpt-oss adapter reads it. The adapter already resolves model-specific env itself (`GPT_OSS_WEIGHTS_FROM_CACHE`, `PREFILL_TOPOLOGY`), so read e.g. `GPT_OSS_BOUNDED_SLIDING_KV` there and pass it to `allocate_kv_cache` / `TtPrefillRuntime`. That keeps `PrefillRunParams` from growing fields that other adapters silently ignore. Since PR1 can't run e2e until #54935 lands anyway, the runner plumbing isn't needed in this PR.

The rest is inline. The fingerprint one is a nit and goes away with the common changes.

models/demos/common/prefill/runners/prefill_runner.py
mdragulaTT: The runner change is gone.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done: capacity is `min(2 * largest chunk, max_seq_len)`, so the variable-chunk smoke shape is accepted and the one-shot error is gone.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done: one chunk size per sequence is checked once at `prefill_chunk` entry, per slot; the per-write asserts are removed

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done: `==`, callers pass exactly this rank's slice

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done: one loop through `layer_view`, modulo applied for bounded slots

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done.

models/demos/gpt_oss_d_p/tt/attention/kv_cache.py
mdragulaTT: Done.

models/demos/gpt_oss_d_p/tt/tt_prefill_runtime.py
mdragulaTT: Done: `gather_layer` returns `(k, v, resident)` with compacted K/V, `written_tokens` required for bounded layers, and the PCC check just slices the golden

mdragulaTT:
The `models/demos/common changes` are dropped; the gpt-oss adapter reads `GPT_OSS_BOUNDED_SLIDING_KV` itself and passes it to both the cache allocation and the runtime config.

13 CPU unit tests and the four 1x1 device write-vs-reference tests pass; on the PR1+#54935 stack the galaxy 36L 5k@1k harness is unchanged: flag off K=0.98172 / V=0.92036, flag on K=0.97986 / V=0.92036.

philei-tt (approved):

PR #53774, opened 2026-08-20
[Minimax M3] (8, 4) Mesh traceable chunked prefill

zbaczewskiTT:
### Summary
Changes to M3's single-galaxy (8,4)/EP32 chunked prefill to enable metal trace:
1. **Per-chunk metal-trace pool** - One captured trace per distinct `cached_len` - KV bucket.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:zbaczewski/perf/m3-8x4-e2e-trace)

### DIFF
--- a/models/demos/minimax_m3/tests/galaxy_prefill_kv_pcc.py
+++ b/models/demos/minimax_m3/tests/galaxy_prefill_kv_pcc.py
@@ -208,9 +208,9 @@ def main():
         )
 
     ttnn.set_fabric_config(ttnn.FabricConfig.FABRIC_1D)
-    # PREFILL_TRACE_POOL needs a DRAM trace region to hold one captured trace per KV-length bucket;
+    # PREFILL_USE_TRACE needs a DRAM trace region to hold one captured trace per KV-length bucket;
     # default 0 leaves the eager path byte-identical (no region reserved).
-    _trace_region = int(os.getenv("PREFILL_TRACE_REGION", "0"))
+    _trace_region = int(os.getenv("PREFILL_TRACE_REGION_SIZE", "0"))
     mesh = ttnn.open_mesh_device(ttnn.MeshShape(rows, cols), trace_region_size=_trace_region)
     print(f"[prefill-pcc] mesh opened {tuple(mesh.shape)} ndev={mesh.get_num_devices()}", flush=True)
     try:
@@ -272,6 +272,7 @@ def main():
             print("[prefill-pcc] loading real bf16 weights + EP placement (slow: bf16 source read) ...", flush=True)
             state_dict = ModelArgs.load_state_dict(model_args.weights_path)
         num_users = int(os.getenv("PREFILL_NUM_USERS", "1"))
+        use_trace = os.getenv("PREFILL_USE_TRACE") == "1"
         cfg = TtPrefillRuntimeConfig(
             num_layers=num_layers,
             max_seq_len=total,
@@ -280,6 +281,7 @@ def main():
             num_users=num_users,
             expert_weight_dtype=expert_dtype,
             weight_cache_path=cache_path,
+            use_trace=use_trace,
         )
         runtime = TtPrefillRuntime(mesh, hf_config, state_dict, cfg)
         del state_dict
@@ -293,12 +295,12 @@ def main():
         print(f"[prefill-pcc] compiling ({num_layers}L, SP=8 × TP=4 + EP=32) ...", flush=True)
         runtime.compile(kv_cache)
 
-        # --- Request-mode WRITE+READ slot correctness (PREFILL_TWO_USER_PCC=1, PREFILL_NUM_USERS>=2):
-        # prefill each user into its own slot (the write/read slot is device-valued whenever num_users > 1),
-        # then PCC every user's cache vs the golden. A mis-targeted slot corrupts one user's KV -> its PCC
-        # craters. Same golden for all users, so each slot must match it independently.
-        if os.getenv("PREFILL_TWO_USER_PCC") == "1":
-            assert num_users >= 2, "PREFILL_TWO_USER_PCC needs PREFILL_NUM_USERS>=2"
+        # Request-mode eager path: more than one user prefills each into its own slot (the write/read slot
+        # is device-valued whenever num_users > 1), times the multi-user pass, then PCCs every slot against
+        # the golden unless PREFILL_SKIP_PCC is set. All slots share one golden, so a mis-targeted slot
+        # corrupts only that user's KV and craters only its PCC. The traced multi-user path is the trace pool
+        # below (PREFILL_USE_TRACE), so defer to it when set.
+        if num_users > 1 and not use_trace:
             two_padded = token_ids + [0] * (total - n_tokens)
             reps = int(os.getenv("PREFILL_TPS_ITERS", "3"))
             passes = []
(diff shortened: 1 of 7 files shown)


ayerofieiev-tt (commented):
Review of the traceable chunked-prefill change. Ten inline comments below. The two most important ones are the dense layers still reading the cache with a host int slot (so a replayed trace for user 1 reads user 0's dense KV), and the traced forward dropping actual_isl and on_layer_complete, so replay behaves differently from the eager path it replaces.

models/demos/minimax_m3/tt/attention/prefill.py
```diff
@@ -221,9 +234,23 @@ def attention_forward(
                     v_int = ttnn.to_memory_config(kv_cache.v, ttnn.DRAM_MEMORY_CONFIG)
                     ik_int = ttnn.to_memory_config(kv_cache.index_k, ttnn.DRAM_MEMORY_CONFIG)
                 with zone("slice", FINE):
-                    k_acc = ttnn.slice(k_int, (slot, 0, 0, 0), (slot + 1, 1, n_rows, config.head_dim))
-                    v_acc = ttnn.slice(v_int, (slot, 0, 0, 0), (slot + 1, 1, n_rows, config.head_dim))
-                    ik_acc = ttnn.slice(ik_int, (slot, 0, 0, 0), (slot + 1, 1, n_rows, config.head_dim))
+                    # Multi-user: the slot varies per replay, so it rides a device tensor a host update
```
ayerofieiev-tt, line 237: (About the dense-layer read at line 332 below, `slot_idx=user_id` in the `dense_sp_attention` call. Anchored here because that line is outside the diff.)

The dense layers (0-2) still pass the slot as a plain host int here. In dense_sp.py this becomes `kv_cache_batch_idx=slot_idx * num_layers + layer_idx`, which is a fixed number inside a captured trace. The new device-valued slot only covers the MSA read path (`_slot_slice_device`) and the KV write (`write_slot_tensor`).

So with `PREFILL_USE_TRACE=1` and `num_users > 1`, a replay for slot 1 writes user 1's KV and reads user 1's MSA cache, but layers 0-2 attend over user 0's cached K/V (slot 0 was baked in at capture). With different prompts per user the hidden state is wrong from layer 0 on, and every later layer's KV for user 1 is wrong too.

The PCC test will not catch this because every user prefills the same prompt, so slot 0 and slot 1 hold the same data. Either make the dense cache read slot device-valued as well, or assert that dense layers with `cached_len > 0` are not on the traced multi-user path.

models/demos/minimax_m3/tt/tt_prefill_runtime.py
```diff
@@ -399,6 +420,118 @@ def set_layer_completion_sink(self, sink) -> None:
...
+        return self.model.prefill_forward(
+            x,
+            rot_mats_global=self.rope_indexed,
+            kv_cache=kv_cache,
+            cached_len=cached_len,
+            user_id=slot_id,
+            get_last_token=-1,
+            skip_lm_head=True,
+            indexed_rope=True,
+            on_layer_complete=None,
```
ayerofieiev-tt, line 441: This is a second hand-written copy of the `prefill_forward` call in `prefill_chunk`, and it has already drifted in two ways.

1. It does not pass `actual_isl`. The eager call passes `actual_isl=actual_end - actual_start` so the MoE skips the padded tail of a short final chunk. Under trace every chunk is captured as a full chunk, so the padding tokens are routed and dispatched as real tokens. That wastes dispatch-buffer capacity and means traced and eager runs give different behavior for the same request. `actual_end` is silently ignored on the traced path.

2. It passes `on_layer_complete=None`, see the separate comment on the ack path.

Please factor the shared arguments into one helper used by both call sites so they cannot drift again. For `actual_isl` under trace, either capture a full-chunk trace only and document that `actual_end` must equal `actual_start + chunk_size`, or assert that in `prefill_chunk` on the replay path.

models/demos/minimax_m3/tt/tt_prefill_runtime.py
```diff
@@ -399,6 +420,118 @@ def set_layer_completion_sink(self, sink) -> None:
...
+        )
```
ayerofieiev-tt, line 442: Per-layer completion callbacks never fire under trace. The trace is captured with `on_layer_complete=None`, and `replay_chunk` only calls `execute_trace`. So neither `self._on_layer_complete` (the LayerAck inject) nor `self._layer_completion_sink` runs for any served chunk.

The runner registers those hooks before it calls `capture_trace` (prefill_runner.py lines 1083 and 1101), and its comment on the single-rank ack path says it works traced or untraced. With `PREFILL_USE_TRACE=1` and `PREFILL_ENABLE_LAYER_ACK=1`, the scheduler counter never advances for real requests, and KV migration waits on per-layer completions that never come.

Either split the capture at each layer so the host callback can run between segments (the DeepSeek runtime does this), or fire the callbacks from `replay_chunk` after the trace completes, or refuse to capture when a completion hook is registered.

models/demos/minimax_m3/tt/tt_prefill_runtime.py
```diff
@@ -399,6 +420,118 @@ def set_layer_completion_sink(self, sink) -> None:
...
+
+    def replay_chunk(self, c: int):
+        """Replay chunk c's captured trace (input must already be updated). Returns the output-hidden
+        handle for a middle rank (the caller MUST clone it before the D2D send, which frees its arg —
+        else the next replay corrupts); None on the last/single rank."""
+        slot = self._trace_pool[c]
+        ttnn.execute_trace(self.mesh_device, slot["tid"], cq_id=0, blocking=True)
+        return slot["out"] if not self.config.is_last_rank else None
+
+    def release_trace_pool(self) -> None:
```
ayerofieiev-tt, line 517: The runner's shutdown hook looks for a method named `release_trace` (prefill_runner.py line 684: `getattr(runtime, "release_trace", None)`). This class names it `release_trace_pool`, so the hook finds nothing and the captured traces are never released before `close_mesh_device`.

The runner comment above that hook explains why this matters: closing the mesh with live traces can free buffers in the wrong order and crash in `BankManager::deallocate_buffer`. At best the 800 MB trace region from the new yaml is leaked.

Rename this to `release_trace`, or add `release_trace = release_trace_pool` as an alias.

models/demos/minimax_m3/tt/tt_prefill_runtime.py
```diff
@@ -328,6 +332,23 @@ def prefill_chunk(
...
+        if self.config.use_trace and self._trace_pool is not None:
+            # Replay this chunk's captured bucket (selected by cache offset) instead of re-dispatching,
+            # after re-targeting the device-valued slot (num_users > 1) and refreshing the input in place.
+            # compile()'s warm sweep runs before capture, when the pool is empty, so it falls through to the
+            # eager path below — which is what warms the programs the capture then records.
+            assert skip_lm_head and get_last_token == -1, (
+                "use_trace captures the headless cache-fill forward; logits (skip_lm_head=False / "
+                "get_last_token>=0) are not on the traced path"
+            )
+            bucket = actual_start // self.config.chunk_size
```
ayerofieiev-tt, line 344: The bucket is chosen with floor division, but nothing checks that `actual_start` is a multiple of `chunk_size`. The asserts above accept any `actual_start`, and the eager path below honors it exactly. Each captured trace has `cached_len = bucket * chunk_size` baked in.

If a producer ever sends an unaligned `actual_start` (for example `chunk_size + 32`), the replay silently uses `cached_len = chunk_size`: the KV write lands 32 tokens early over the previous chunk's tail, RoPE positions are off, and the causal `chunk_start` is wrong. No error is raised.

Add `assert actual_start % self.config.chunk_size == 0` on this path so the mismatch fails loudly.

models/demos/minimax_m3/tt/tt_prefill_runtime.py
```diff
@@ -399,6 +420,118 @@ def set_layer_completion_sink(self, sink) -> None:
...
+        assert self.rope_indexed is not None, "indexed rope must be built (avoids the SP+trace host reshard)"
+        mesh = self.mesh_device
+        buf = self._trace_in[c]
+        warm = self._trace_fwd(buf, cached_len, slot_id, kv_cache)
+        ttnn.synchronize_device(mesh)
+        if warm is not None:
+            warm.deallocate(True)
+        # The warm forward pre-set the device slot metadata to slot_id; freeze it so the captured forward
+        # only reads it — a host copy inside a trace is illegal (set_read_user re-targets it outside).
+        kv_cache._slot_frozen = True
```
ayerofieiev-tt, line 486: `_slot_frozen` is set to True here and cleared four lines later with no `try`/`finally`. If `begin_trace_capture`, `_trace_fwd`, or `end_trace_capture` raises (for example the trace region is too small, or an op is not capturable), the cache stays frozen.

After that, every `read_slot_start` and `write_slot_tensor` call silently skips its host update. An eager `prefill_chunk(slot_id=1)` would then read and write whatever slot the tensors last pointed at (slot 0), with no error. This matters when a caller falls back to eager after a failed capture, or when a test reuses the same `kv_cache`.

Wrap the capture in `try: ... finally: kv_cache._slot_frozen = False`.

models/demos/minimax_m3/tt/attention/kv_cache.py
```diff
@@ -112,23 +206,38 @@ def _alloc(dtype=cache_dtype):
...
-    ttnn.experimental.deepseek_prefill.update_padded_kv_cache(
-        cache,
-        src,
-        slot_idx=slot_idx,
-        layer_idx=layer_idx,
-        num_layers=num_layers,
-        kv_actual_global=kv_actual,
-        cluster_axis=sp_axis,
-    )
+    if kv_cache.num_users > 1:
```
ayerofieiev-tt, line 220: The tensor form of `update_padded_kv_cache` skips the host-side checks the scalar form does. In `update_padded_kv_cache_device_operation.cpp` the checks for `slot_idx < num_slots`, `kv_actual_global % 32 == 0`, and `kv_actual_global + chunk <= capacity` are all inside `if (!tensor_args.slot_idx.has_value())`. Nothing in `_write_one` replaces them.

So whenever `num_users > 1`, a bad slot or misaligned `kv_actual` is no longer a loud `TT_FATAL`. The writer kernel computes its offset from the on-device scalar and writes past the cache buffer in DRAM, silently corrupting neighboring tensors. This is easy to hit because `allocate_kv_caches` and `TtPrefillRuntimeConfig` take `num_users` and `max_seq_len` independently and only the harness happens to pass matching values.

Add the same range and alignment asserts here, on the host values, before taking the tensor form.

models/demos/minimax_m3/tt/attention/prefill.py
```diff
@@ -221,9 +234,23 @@ def attention_forward(
...
+                    # re-targets (a host int would be baked into the trace). At one user the slot is fixed,
+                    # so the host int is correct and cheaper.
+                    if kv_cache.num_users > 1:
```
ayerofieiev-tt, line 240: This branch (and the matching one in `kv_cache._write_one`) is gated on `kv_cache.num_users > 1`, not on tracing being enabled. The runner defaults `PREFILL_NUM_USERS` to 2 (prefill_runner.py line 128) and no yaml overrides it. So every eager request-mode run now takes the device-slot path, not just the trace experiment.

In eager mode that means, per MSA layer per chunk, three blocking `copy_host_to_device_tensor` calls from `read_slot_start` (k, v, ik) plus three more from `write_slot_tensor`, replacing a single host-int slice. Any bug in the new two-step slice path also hits untraced serving.

Suggest gating on an explicit flag the runtime sets (for example `kv_cache.device_slot = config.use_trace and num_users > 1`), and having the forward only read the persistent slot tensors, with exactly one `set_read_user` per chunk from the runtime. That also removes the need for the externally poked `_slot_frozen` flag.

models/demos/minimax_m3/tests/galaxy_prefill_kv_pcc.py
```diff
@@ -208,7 +208,10 @@ def main():
         )
 
     ttnn.set_fabric_config(ttnn.FabricConfig.FABRIC_1D)
-    mesh = ttnn.open_mesh_device(ttnn.MeshShape(rows, cols))
+    # PREFILL_USE_TRACE needs a DRAM trace region to hold one captured trace per KV-length bucket;
+    # default 0 leaves the eager path byte-identical (no region reserved).
+    _trace_region = int(os.getenv("PREFILL_TRACE_REGION_SIZE", "0"))
```
ayerofieiev-tt, line 213: The default here is 0 regardless of `PREFILL_USE_TRACE`. The runner defaults to 256 MB when tracing is on. Running this script with `PREFILL_USE_TRACE=1` but without `PREFILL_TRACE_REGION_SIZE` opens the mesh with no trace region, loads weights, warms every bucket in `compile()` (minutes), and then fails inside `begin_trace_capture`.

Pick a sane default when `use_trace` is set, or assert early that the region is non-zero.

models/demos/minimax_m3/tests/galaxy_prefill_kv_pcc.py
```diff
@@ -268,27 +271,118 @@ def main():
...
+                # Each user's replay fills its own slot via the device-valued read+write slot; PCC every
+                # slot vs the golden — a mis-targeted read or write corrupts that user's KV and craters it.
+                for uid in range(num_users):
+                    print(f"[prefill-pcc] --- traced user {uid} (slot {uid}) KV PCC vs golden ---", flush=True)
+                    check_kv_pcc(runtime, kv_cache, golden_dir, n_tokens, num_layers, hf_config, slot_id=uid)
+            else:
+                check_kv_pcc(runtime, kv_cache, golden_dir, n_tokens, num_layers, hf_config)
+            runtime.release_trace_pool()
+            print("[prefill-pcc] DONE", flush=True)
+            return 0
```
ayerofieiev-tt, line 384: Both new blocks (`num_users > 1` eager at line 337, and this trace block) `return 0` before the existing `PREFILL_EXPECTED_TPS` / `PREFILL_PERF_MARGIN` gate at line 443. A CI job that sets `PREFILL_EXPECTED_TPS` together with `PREFILL_USE_TRACE=1` or `PREFILL_NUM_USERS=2` prints tok/s and exits 0 without checking it, so a perf regression passes green.

Also, both blocks re-read `PREFILL_TPS_ITERS` with default 3, while the variable parsed at line 178 (`tps_iters`) and the docstring say default 1. Reuse `tps_iters` and run the perf gate on the measured steady-state number before returning. `release_trace_pool` is also only reached on the success path; move it to the `finally`.

PR #53783, opened 2026-08-20
[Bug fix] prelu: fuse the tensor weight path into one kernel

### LINKED ISSUE #53782: ttnn.prelu: the tensor weight path is three kernels for one branch

**Issue Type:** Performance

## Observed

`ttnn.prelu(x, weight)` with a weight tensor is `ltz` then `multiply` then `where`, three device kernels for `x < 0 ? x * w : x`:

```cpp
Tensor result = ttnn::where(ttnn::ltz(input_a, output_mem_config), ttnn::multiply(input_a, b), input_a);
```

The scalar-weight overload is already one kernel, through `prelu_sfpu`, whose body is exactly `v_if(a < 0.0f) { a = a * init; }`. Only where the weight comes from differs.

Measured on P300 and N300, device kernels per call:

| | bfloat16 | float32 |
|---|---|---|
| kernels | 3 | 3 |

## Expected

One kernel, with the weight read from a second tile instead of a compile-time argument. `binary_ng` already carries the channel broadcast, so the reshape the op does to line the weight up is the only shape work that has to stay.

## Not established

- The scalar and array overloads, which are already single-dispatch and are left alone. #48240 item 21 reports they ignore `output_mem_config`; that is untouched here.
- `bfloat8_b` and `bfloat4_b`, and the sharded paths.
- NaN propagation, which the composite does not carry either.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
### PR Category

Bug fix.

### Summary

Closes #53782.

`prelu` with a tensor weight becomes one kernel. The body is the scalar overload's, `x < 0 ? x * w : x`, with the weight read from a second tile rather than from a compile-time argument, so the two overloads now differ only in where the weight comes from.

`binary_ng` carries the channel broadcast, so the reshape that lines the weight up against the input is the only shape work left in the op.

### Accuracy

Two random shapes and one sign-boundary set per dtype, on P300 and N300, against what the composite returned. Both machines give the same counts.

| | float32 | bfloat16 |
|---|---|---|
| `[1, 8, 32, 32]`, C=8 | identical | 107 of 8192 differ |
| `[2, 16, 64, 64]`, C=16 | identical | 1536 of 131072 differ |
| sign boundary, inf, nan | identical | identical |

**bfloat16 moves by one ulp on 1643 of 139264 elements, 1.18 percent, and never by more than one.** The composite multiplied through `ttnn::multiply`, which for bfloat16 rounds on the FPU; this rounds the same product on the SFPU. Rounding explicitly to bfloat16 in the kernel brings the disagreement down from 24 percent to 1.18 percent, and the remainder is the two units rounding the same value differently. float32 is unaffected and bit-identical throughout.

### Cost

Device profiler, `[1, 32, 128, 128]` with a 32 element weight, 33 invocations, median cycles per core on the math thread. Zone counts give the kernel count.

| | P300 bf16 | P300 fp32 | N300 bf16 | N300 fp32 |
|---|---|---|---|---|
| kernels per call | 3 → **1** | 3 → **1** | 3 → **1** | 3 → **1** |
| TRISC_1, before | 19362 | 39534 | 36468 | 72042 |
| TRISC_1, after | **6058** | **13520** | **11086** | **22364** |
| | **3.2x** | **2.9x** | **3.3x** | **3.2x** |

### Tests

`test_prelu_tensor_weight.py`, 8 cases: three shapes across both dtypes, and the sign boundary with infinities and NaN. bfloat16 is bounded at one ulp rather than asserted equal, for the reason above, so a larger drift fails there.

88 existing prelu cases in `test_binary_composite.py` pass unchanged.

### Related context

#52153 lists `prelu` for migration to a device op. #48240 item 21 reports the scalar and array overloads ignoring `output_mem_config`; those overloads are already single-dispatch and are untouched here.

### Not covered

- The scalar and array overloads.
- `bfloat8_b` and `bfloat4_b`, and the sharded paths.
- NaN propagation, which the composite does not carry either, so the two agree on it.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_prelu_tensor_weight.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_prelu_tensor_weight.py
@@ -1,0 +1,61 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""prelu with a per-channel weight tensor.
+
+The weight is broadcast along the channel dimension, so these cover the shapes
+where that broadcast is what the op is doing, plus the sign boundary, which is
+the only branch in the kernel.
+"""
+
+import pytest
+import torch
+import ttnn
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+@pytest.mark.parametrize("shape, channels", [((1, 8, 32, 32), 8), ((2, 16, 64, 64), 16), ((1, 4, 32, 32), 4)])
+def test_prelu_tensor_weight(device, dtype, shape, channels):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    torch.manual_seed(0)
+
+    x = (torch.rand(shape, dtype=torch.float32) * 20 - 10).to(torch_dtype)
+    w = (torch.rand((channels,), dtype=torch.float32) * 2 - 1).to(torch_dtype)
+    want = torch.nn.functional.prelu(x.float(), w.float()).to(torch_dtype)
+
+    ix = ttnn.from_torch(x, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    iw = ttnn.from_torch(w, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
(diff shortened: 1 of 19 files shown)


mateusznowakTT:
@kinginu : Please resolve merge conflict. Also, please add LLK level test for added kernel on top of current ttnn UT.

PR #53785, opened 2026-08-20
Add multicast pipe and host helpers

s-jovic:
### Abstractions

1. Kernel abstractions (`mcast_pipe.hpp`, `chain_pipe.hpp`):
   - `SenderPipe` / `ReceiverPipe`
   - `ChainSenderPipe` / `ChainReceiverPipe`

   These encapsulate synchronization, signaling, and data transfer for hardware multicast and chain forwarding. `McastArgs` constructs the appropriate pipe from host-generated arguments.

2. Host abstractions (`host/mcast_host.hpp`):
   - `McastGroup` - describes a receiver set and its sender configuration.
   - `McastFamily` - prepares one or more groups and selects hardware multicast or chain forwarding for irregular receiver sets.
   - `Mcast1D` / `Mcast2D` - convenience wrappers for common row, column, and rectangular layouts.

   The host helpers centralize argument, topology, and semaphore preparation.

[HTML showing Mcast1D and 2D visualisation and mcast vs chain explanation](https://claude.ai/code/artifact/3e25c9f9-915e-4942-a89d-915eb1280aa5)

### Suggested review areas

- Kernel side: `McastArgs`, pipe construction, synchronization, and the `send()`, `receive()`, and `receive_and_forward()` APIs. Start with usage example in [mcast_args.hpp ](ttnn/cpp/ttnn/kernel_lib/mcast/kernel/mcast_args.hpp)and then continue with McastPipe and ChainPipe classes in [mcast_pipe.hpp](ttnn/cpp/ttnn/kernel_lib/mcast/kernel/mcast_pipe.hpp) and [chain_pipe.hpp.](ttnn/cpp/ttnn/kernel_lib/mcast/kernel/chain_pipe.hpp)
- Host side: `McastGroup` / `McastFamily` topology and transport selection, plus the `Mcast1D` / `Mcast2D` wrappers. Start in [mcast_host.hpp](ttnn/cpp/ttnn/kernel_lib/mcast/host/mcast_host.hpp) file, there are usage examples at the beginning of the file.
- Usage examples: matmul for common 1D/2D multicast and Conv3D for irregular receiver sets and chain forwarding.

### Migrated ops present in the PR

- matmul
- conv2d
- layernorm
- topk
- groupnorm
- conv3d


ttnn/cpp/ttnn/kernel_lib/mcast_pipe.hpp
wransom-TT, line 114: Probably need to fix this hardcoding

wransom-TT, line 43: I think there's stale references in here

tests/ttnn/unit_tests/gtests/test_mcast_host.cpp
```diff
@@ -0,0 +1,1046 @@
...
+        xhi = std::max(xhi, x);
+        ylo = std::min(ylo, y);
+        yhi = std::max(yhi, y);
+    }
+    return noc == NOC::NOC_1 ? std::vector<uint32_t>{xhi, yhi, xlo, ylo} : std::vector<uint32_t>{xlo, ylo, xhi, yhi};
+}
+
+}  // namespace
+
+TEST(GroupNormMcastGeometry, ZeroEdgeRectangle) {
```
wransom-TT, line 124: Claude thinks these 3 tests are testing existing functionality not stuff added by this PR, worth checking, not necessary to remove if it adds coverage we lacked that's always a good thing.

ttnn/cpp/ttnn/kernel_lib/host/mcast_host.cpp
wransom-TT, line 194: We should do this so the relationship is enforced outside of just tests.

wransom-TT, line 208: Same here

wransom-TT, line 488: And here

wransom-TT, line 500: And here

wransom-TT: What happens if the sender core is on a line edge but not at index 0, like can we have a line 1-N and sender is at index 1? Or are lines always 0-N? In general I don't understand why we have explicit functionality for this case where the sender is on the edge of a line, seems like we should always just include the sender if it's in the line regardless of whether it's on an edge or the interior.

wransom-TT, line 605: I think we can eliminate this function and just call the underlying `noc_ordered_bbox`

wransom-TT, line 342: We can just call the underlying function here.

ttnn/cpp/ttnn/kernel_lib/mcast_pipe.hpp
wransom-TT, line 108: We can also remove this entire area function if we track corners for the bounding boxes instead of the full list and emit the core count from the host instead of deriving it.

ttnn/cpp/ttnn/kernel_lib/host/mcast_host.cpp
wransom-TT, line 594: This is a bigger design question, just placing it here, it touches a few places, only affecting non-performance critical host code, so just a quality thing mostly:

Replace stored coordinate lists with interval arithmetic. The helpers currently store vectors of core coords (sender_lines_, senders_, rect_virt_coords_()) and loop over them for validation, fan-out counting, is_sender, and bbox construction. All of this state is derivable from corners, since logical space is dense and the logical→virtual translation is per-axis monotonic on all supported archs. More specifically:

1. (Optional if no planned migration needs a scattered sender set) - Narrow sender_grid to a dense rectangle (like the receiver grid already is). Per-line sender candidates become a contiguous segment [s0, s1], so sender identity is (line, index) with coords computed as start + offset.
2. Convert only corners for bboxes - change noc_ordered_bbox to take the device + two logical corners and do the conversion itself (documenting the monotonicity assumption once).
3. Pass the sender's fan-out as an RT word - (host already computes it as num_receivers(core)), so the device consumes corners + count.

What collapses: sender_lines_from_grid_ and senders_from_grid_, the add_sender fanout/uniformity walk, is_sender's linear search, and the span/area loops in sender_rect_, rotating_rt_, and rect_virt_coords_.

What gets deleted: McastRect::area() and its Blackhole column-8/9 hack (resolves the TODO at mcast_pipe.hpp:109), the ACK_EQUALS_FANOUT sentinel + uniform-fanout detection (the host emits the resolved per-core ack count instead), rect_virt_coords_(), and the coord-vector members. Member state reduces to two rects plus scalars.

What remains: the rotating RT block (2·span sender coord pairs) and its emission loop - receivers must signal to each round's sender at virtual coords. So this becomes a generator over indices rather than iteration over stored vectors, and can be emitted once per line rather than per core.

wransom-TT (commented):
General API loops good, I will pay more attention to the usabiliy of the API once I can see the kernel migrations. For now I've commented on some general clarity / efficiency improvements I can see in the underlying API functions. Let me know what you think.

ttnn/cpp/ttnn/kernel_lib/host/mcast_host.cpp
s-jovic: You are right, this is reinventing part of the logic we have on device when deciding whether we include/exclude src, we don't need the special functionallity

s-jovic: Will get back on this ⚙️

ttnn/cpp/ttnn/kernel_lib/mcast_pipe.hpp
```diff
@@ -0,0 +1,336 @@
...
+//       receiver.receive(round);
+//   }
+//
+// Preconditions: one active sender per round; semaphores are initialized to INVALID on every
+// participating core; the landing address `dst_l1` is identical across all receivers.
+// =============================================================================
+
+#pragma once
+
+// Caller-facing API version — the staleness key for the apply-dm-helper rollout ledger
```
djordjenTT, line 39: This should be the staleness key for anyone, not just that agent?
ledger.json seems like a dangling reference?

s-jovic: This should not exist once the PR is merged - I am using it for tracking purposes for the ongoing work - whether kernels were migrated and to which API version. I just didn't clean it up from the review PR as it's easier for avoiding conflicts.

tests/pipeline_reorg/ops_unit_tests.yaml
```diff
@@ -36,46 +34,71 @@
...
+      timeout: 50
   owner_id: U084B1CES7M # Borys Bradel
   team: ttnn
   category: matmul
 
+# Keep the host gtest subprocess ahead of any Python device fixture.
+# The ops workflow downloads the build artifact containing unit_tests_ttnn.
+- name: ttnn multicast helper tests
+  cmd: >-
+    scripts/run_safe_pytest.sh tests/ttnn/unit_tests/kernel_lib/test_mcast_prepared_host.py -xv --timeout=300 &&
```
mstaletovicTT, line 54: I am pretty sure run_safe_pytest.sh is never ran on CI ? I would expect it to be fine, though I am curious if this has some implications

tests/ttnn/unit_tests/gtests/test_mcast_host.cpp
```diff
@@ -0,0 +1,651 @@
...
+    EXPECT_EQ(moved.compile_time_args(), old_ct);
+    EXPECT_EQ(moved.runtime_args({4, 2}), old_rt);
+}
+
+TEST_F(McastHostFixture, GroupNormFactoryEmitsExactDestinations) {
+    using namespace tt::tt_metal;
+    using Operation = ttnn::prim::GroupNormDeviceOperation;
+    const auto available = device_->compute_with_storage_grid_size();
+    if (available.x < 9 || available.y < 9) {
+        GTEST_SKIP() << "Requires both 7x9 and 9x7 shard grids";
```
mstaletovicTT, line 603: Can this test be exercised on WH ? If not, I would guess that it is possible to change the shape for the groupnorm op, so that it can be exercised on there aswell

tests/ttnn/unit_tests/kernel_lib/kernels/pipe_f3_compute.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "api/dataflow/circular_buffer.h"
+
+void kernel_main() {
+    constexpr uint32_t cb_in = get_compile_time_arg_val(0);
+    constexpr uint32_t cb_out = get_compile_time_arg_val(1);
+    constexpr uint32_t num_tiles = get_compile_time_arg_val(2);
+
+    CircularBuffer input(cb_in);
+    CircularBuffer output(cb_out);
+    unary_op_init_common(cb_in, cb_out);
```
mstaletovicTT, line 22: compute_kernel_hw_startup

tests/ttnn/unit_tests/kernel_lib/kernels/pipe_f3_sender.cpp
```diff
@@ -0,0 +1,85 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+// mcast_pipe helper unit test: F3 SENDER-IN-RECT kernel driving Pipe with INCLUDE_SRC
```
mstaletovicTT, line 4: What does F3 represent here ?

ttnn/cpp/ttnn/kernel_lib/mcast/host/mcast_host.cpp
```diff
@@ -0,0 +1,606 @@
...
+    for (auto& group : groups_) {
+        group.prepared_.reset();
+    }
+    const bool has_irregular_receiver_set = std::any_of(
+        groups_.begin(), groups_.end(), [](const auto& group) { return group.receivers_.ranges().size() > 1; });
+    // Resolve the irregular-set policy once because every group uses the same compile-time TransferMode.
+    const auto transfer_mode =
+        has_irregular_receiver_set && cfg_.irregular_receiver_set_mode == TransferMode::ChainUnicast
+            ? TransferMode::ChainUnicast
+            : TransferMode::Multicast;
```
mstaletovicTT, line 330: This seems to make it so that we cannot use ChainUnicast in situations of the grid being regular (and it seems to silently "fall back" to multicast).

Are you sure that there are no situations where ChainUnicast is faster, even when the grid is rectangular ? I could imagine a situation where the transactions are small, but the grids for sending are kinda big, that repeated path reserves could get expensive.

mstaletovicTT: But even aside from that, ChainUnicast is the more "pro" and opt-in feature, so even if we cannot think of an examples right away, I do not think that we should silently restrict someone from trying

Issue #53466, opened 2026-08-18
ttnn.topk INT32/UINT32 index_dtype returns wrong indices on Blackhole at k=50, W=8192 (multi-core path)


nkapreTT:
`test_topk_int32_indices[largest={True,False}-index_dtype=DataType.{INT32,UINT32}-N=1-C=1-H=32-W=8192-dim=3-k=50]` (added in #48843) fails on Blackhole p150a silicon: the values are plausible but gather-from-indices produces garbage (cosine similarity ≈ −0.05 vs the expected >0.99), for both largest values — i.e., the returned indices don't point at the returned values.

Verified against **unmodified `main`** (differential run with a pure-main `topk.cpp`/kernel build on the same box) — this is not introduced by any in-flight PR; it reproduces on main as of e6a5722be2d..6abac727c21. W=8192 is pow-2 and < 65535, so this exercises the stock multi-core bitonic path with the int32-index plumbing; k=50 is deliberately not a tile multiple.

Suggested triage starting point: the int32/uint32 index-widening path through the multi-core final gather — the shape class (H=32 rows, k not a multiple of 32) is adjacent to the ht-iteration unpack-state bug just fixed in #53453, and may be its int32-typed sibling (the widened index tiles change the SRCA format the gather must re-establish).

Found while validating #53464 on Blackhole silicon.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

iwroszTT:
Debug info found while validating an unrelated topk branch:

**1. The corrupted indices are bounded by k, not W.** For `[1,1,32,8192] k=50` with a preallocated `UINT32` indices tensor, every returned index lies in `[0, 49]` (row 0 is all zeros); expected e.g. `[780, 5186, 4397, 2798, ...]`. Only 7 of 1600 positions gather back to the returned values (chance level). The values output stays plausible. So the output stage is emitting k-local positions / an iota-like tile, not global column indices.

```
index range:         min 0, max 49        # bounded by k, not by W=8192
row 0 indices:       [0,0,0,0,0,0,0,0,...]
gather(input,i)==v:  7 of 1600
```

[...]

Issue #47014, opened 2026-06-15
[Docs] Blackhole bring-up guide lists 140 Tensix cores; p150 unified to 120 (firmware v19.5.0)


kinginu:
### Description

`tech_reports/Blackhole/BlackholeBringUpProgrammingGuide.md` lists the Blackhole Tensix grid as **14x10 = 140 total / 13x10 = 130 available for compute** in the "Wormhole N150 vs. Blackhole" table.

As of **firmware v19.5.0** ([tt-system-firmware](https://github.com/tenstorrent/tt-system-firmware/releases/tag/v19.5.0), January 2026), all p150 cards (p150a, p150b) are unified to **120 Tensix cores**, and existing cards are updated to 120 via firmware as well:

> **p150 Tensix Core Count** — Beginning January 2026, all Blackhole p150 accelerator cards (p150a, p150b) will ship with 120 Tensix cores instead of 140. To present a unified interface to metal and other system software, firmware v19.5.0 and later will change the core count on all existing cards to 120. ... You may observe a change in grid size in metal, which may require updates to applications that depend on grid layout.

So the guide's 140/130 figures are stale for current firmware and no longer match the published p150 spec (120 Tensix cores).

### Proposed fix

Add a short note after the table pointing out that, as of firmware v19.5.0, p150 presents 120 Tensix cores (with a link to the release notes). The original 140 figure can stay as the unharvested-die maximum for historical context. PR to follow.

kinginu:
Found one more stale spot — now fixed in #47015.

`models/experimental/panoptic_deeplab/README.md` (line 28) still said:

> **Note**: December 2025 - P150 currently has core grid 13x10 but the plan is to have 2 rows harvested (11x10) so in order to prepare for that change, we are setting this env var.

That plan has since landed: firmware **v19.5.0** (January 2026) unifies all p150 cards to **120 enabled Tensix cores (12x10)** → an **11x10 = 110** compute grid. Updated the note to reflect the current default instead of a future plan.

PR #53793, opened 2026-08-20
ttnn.topk: model-based multi-core split + Ht-aware eligibility (up to 10x on low-tile-row shapes)

nkapreTT:
### What

Two scheduling changes to the stock `ttnn.topk` device op — no kernel changes, host-side factory/config selection only:

1. **Model-based multi-core split selection** (`topk_utils.cpp`): `find_topk_core_config` returned the *first* valid config searching from the smallest split upward, i.e. maximum local cores and therefore maximum work on the single **serial** final core (its gather-merge ramp scales with `num_cores * k`). Now every valid config is scored with a fitted makespan model `T ~ 7 * Wt_local + 2 * Wt_final` and the minimum wins. Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells, <0.5% residual — a local tile costs ~3.5x a final tile: locals run full 64-element sorts per tile, the final core runs merge/rebuild pair-ops).

2. **Ht-aware multi-core eligibility** (`topk_device_operation.cpp`): the single-core factory parallelizes across tile *rows*, so 1–2-tile-row inputs ran the whole reduction on one or two cores even when the row was wide enough to column-split. Shapes below `multi_core_min_width` now qualify for the multi-core factory when `num_tile_rows <= 2 && width >= 1024` (all other guards — pow2, <65536, k<=64, L1/split validation — unchanged). The multi-core explicit-u32/i32 index path was fixed upstream (#53466), so the widened arm serves 32-bit index outputs too — `test_topk_int32_indices`, including the 32x4096 cell this arm newly routes to multi-core, passes on this base.

### Measured (Tracy device kernel duration, 3 trials, zero cross-trial variance, p150a, this base vs its merge-base)

| shape | k | main | this PR | delta |
|---|---|---|---|---|
| 32x8192 | 50 | 261.8 us | 154.0 us | **-41.2%** |
| 32x32768 | 64 | 333.8 us | 302.2 us | **-9.5%** |
| 32x2048 | 32 | 287.3 us | 47.7 us | **6.0x** |
| 32x4096 | 64 | 1087.9 us | 106.5 us | **10.2x** * |

\* both sides measured with a full-grid `sub_core_grids` because on Blackhole the default call routes this shape to the `topk_large_indices` composite; the row shows the stock device op's improvement for routing-declined contexts.

Scope note: on Blackhole some of the rescued low-tile-row shapes are also served by the `topk_large_indices` routing added earlier — this PR matters for **Wormhole**, for **routing-declined contexts** (explicit `sub_core_grids`, preallocated outputs, user index tensors, `stable=True`), and for the pow2 **[8192, 65536) band** where the stock multi-core bitonic is the engine of record.

### Validation

- `tests/ttnn/unit_tests/operations/reduce/test_reduction.py` + `test_topk.py`: **658 passed, 32 skipped, 80 xfailed, 0 failed** (p150a).
- `test_topk_int32_indices` (12/12) passes with the widened arm serving explicit 32-bit index outputs.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/topk-stock-split-model)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/topk-stock-split-model)


nkapreTT:
Copilot review audit (04ef66dc633) — all 3 findings verified against the code and addressed:

1. **Ht from the tensor's tile shape** (`topk_device_operation.cpp`) — **valid, fixed.** The eligibility gate divided by the fixed `tt::constants::TILE_HEIGHT` while the program factories size their work via `tensor_spec().tile().get_height()`, and validation does not pin the tile shape to 32 rows. The gate now uses the tensor's tile height (no behavior change for standard 32-row tiles).
2. **Lock down the split selection** (`topk_utils.cpp`) — **valid, addressed.** `find_topk_core_config` is a pure function of its arguments, so a new host-side gtest `TopkCoreConfigModel.SelectsFittedMakespanMinimum` (`tests/ttnn/unit_tests/gtests/test_reduction.cpp`, no device needed) asserts the model's `split_size`/`num_cores` picks for W=8192 (k=32/50/64) and W=32768 (k=64) on the p150a 13x10 grid — e.g. 8192/k64 → split 512 on 16 cores vs the old first-valid 128 on 64 cores. A coefficient or search-order regression now fails this test even while every numerical topk test stays green.
3. **Stale algorithm overview** (`topk_utils.cpp`) — **valid, fixed.** The doc block now states the minimum-makespan selection contract instead of "return the first valid configuration".

Validation (p150a): `test_topk.py` 246 passed / 8 skipped / 80 xfailed (one JIT-cache-infra flake, `Failed to open compile failure log file`, passed on isolated rerun); new gtest passes; 3-trial Tracy spot checks match the PR-body anchors exactly — 32x2048 k=32: 47.7/47.6/47.7 us, 32x8192 k=50 topk op: 154.0 us on all 3 trials.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -156,12 +170,14 @@ std::optional<TopKCoreConfig> find_topk_core_config(
             config.selected_x = static_cast<uint16_t>(selected_x);
             config.selected_y = static_cast<uint16_t>(selected_y);
 
-            // Return the first valid configuration found (greedy approach)
-            return std::make_optional(config);
+            const uint32_t score = kLocalCostFactor * Wt_local + kFinalCostFactor * Wt_final;
+            if (score < best_score) {
```
vtsilytskyiTT: I think you should merge this condition with `if` in the line 152. There is no point to create a `config` if `score >= best_score`.

vtsilytskyiTT (approved):
Looks good, approved. Please run CI sanity test, which will confirm that `test_topk.py` passes.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Done in 1b52752d316 — the score is computed before the validation `if` and `score < best_score` is now the last conjunct, so a config is only constructed when it actually wins. Selection is unchanged (pure predicates, strict `<` keeps first-winner tie behavior); `TopkCoreConfigModel.SelectsFittedMakespanMinimum` passes on the new head.

nkapreTT:
Review suggestion applied in 1b52752d316 (fold the makespan check into the split-validation condition — construction-order cleanup, same configs win, locked by the host-side gtest). Full `tests/ttnn/unit_tests/operations/reduce/test_topk.py` on the new head: 246 passed, 8 skipped, 80 xfailed, 0 failed (p150a); sanity workflow dispatched on this branch: https://github.com/tenstorrent/tt-metal/actions/runs/32718574815.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -82,7 +85,18 @@ std::optional<TopKCoreConfig> find_topk_core_config(
...
+    // Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells,
+    // <0.5% residual): a local tile costs ~3.5x a final tile — locals run full
+    // 64-element sorts per tile while the final core runs merge/rebuild pair-ops.
+    constexpr uint32_t kLocalCostFactor = 7;
+    constexpr uint32_t kFinalCostFactor = 2;
+    std::optional<TopKCoreConfig> best_config = std::nullopt;
+    uint32_t best_score = std::numeric_limits<uint32_t>::max();
     for (uint32_t split_size = start_split_size; split_size <= max_dim; split_size *= 2) {
         // Calculate work distribution for this split size
         TT_FATAL(
```
iwroszTT: **[blocking]** **Width-1024 inputs now crash instead of running single-core**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp`, lines 78–79
(crash fires at the `TT_FATAL` on lines 102–107; gate that admits the shape:
`topk_device_operation.cpp` lines 108–109)

The new eligibility gate admits `reduced_width >= 1024` into `find_topk_core_config`, but the
starting split size

```cpp
const uint32_t start_split_size =
    static_cast<uint32_t>(width / tile_width / largest_power_of_two(max_cores)) * tile_width;
```

truncates to 0 whenever `largest_power_of_two(max_cores) > width / tile_width`. Concretely:
`ttnn.topk` on a bf16 `[1,1,32,1024]` tensor, k ≤ 64, on Blackhole with the default 13×10 grid gives
`max_cores = (9-0-1)*12 = 96`, `lp2 = 64`, and `32 tiles / 64 = 0` — the loop then throws
`Split size must be non-zero (got 0 for width=1024, max_cores=96, max_dim=512)` from
`select_program_factory`. Pre-PR this shape ran single-core and returned correct results; the
composite router doesn't intercept it (width < 4096) and no test covers the cell. Wormhole
(lp2 = 32) happens to be unaffected, but any grid with `lp2(usable cores) > width/32` crashes.

Suggested fix: clamp `start_split_size` to at least `min_dim` (or skip zero splits) so the function
returns `nullopt` and the existing single-core fallback engages instead of throwing.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
```diff
@@ -91,8 +93,20 @@ TopKDeviceOperation::program_factory_t TopKDeviceOperation::select_program_facto
...
+    // even below multi_core_min_width: at 1-2 tile rows a width-1024+ row splits
+    // across 8+ otherwise-idle cores (min 64 elements per core), measured ~4x on
+    // 32x2048 k=32. Wide-and-tall inputs keep the row-parallel single-core path.
+    const uint32_t reduced_width = input_tensor.padded_shape()[args.dim];
+    // Use the tensor's actual tile height (the program factories size their work the
+    // same way via tensor_spec().tile()) so custom tile shapes count tile rows correctly.
+    const uint32_t tile_height = input_tensor.tensor_spec().tile().get_height();
+    const uint32_t num_tile_rows = (input_tensor.padded_shape().volume() / std::max<uint32_t>(reduced_width, 1)) /
+                                   std::max<uint32_t>(tile_height, 1);
+    bool multicore_supported =
```
iwroszTT: **[blocking]** **`validate_on_program_cache_miss` still gates multi-core validation on width ≥ 8192**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp`, line 287
(also lines 298–301, 312, and the unguarded `.at(0)` at line 137)

`select_program_factory` now sends shapes with `1024 <= W < 8192` and ≤ 2 tile rows to the
multi-core factory, but validation still enters its multi-core branch only when
`input_shape[args.dim] >= multi_core_min_width` (8192). So the entire newly-enabled domain runs the
multi-core factory while being validated as single-core. Three concrete gaps:

- The allocator-aware L1 check (`largest_free_block_bytes`, line 312) never runs — select uses total
  `l1_size_per_core()` — so under L1 pressure a `[1,1,32,4096]` topk passes validation and then dies
  at CB/buffer allocation with an opaque error instead of the intended "Not enough cores or cache
  size available" `TT_FATAL`.
- `TT_FATAL(args.sub_core_grids.ranges().size() == 1)` (lines 298–301) is skipped, so a multi-range
  `sub_core_grids` silently truncates to `ranges().at(0)` in select (line 137) and in the factory.
- An empty `CoreRangeSet` passes validation vacuously and escapes as a raw `std::out_of_range` from
  `.at(0)` at line 137.

The eligibility predicate needs to be shared between validate and select rather than mirrored by
hand — see the comment on the four divergent copies below.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -776,6 +778,48 @@ TEST_F(ReductionSmoke, TopkMultiCoreRows64) {
...
+        uint32_t k;
+        uint16_t expected_split;
+        uint16_t expected_cores;
+    };
+    // Expected values: minimum of 7 * Wt_local + 2 * Wt_final over all valid
+    // power-of-two splits. The old first-valid picks were 128/64 (W=8192) and
+    // 512/64 (W=32768) -- maximum cores, maximum serial final-stage work.
+    const std::array<Case, 4> cases{{
+        {8192, 64, 512, 16},
+        {8192, 50, 256, 32},
```
iwroszTT: **[should-fix — pre-existing, but the new test bakes it in]** **`Wt_final` floor-truncates for k not a multiple of 32; kernels gather more tiles than the CBs hold**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp`, line 113
(new test enshrining the config: `tests/ttnn/unit_tests/gtests/test_reduction.cpp` line 809)

```cpp
const uint32_t Wt_final = (num_cores * std::max(k, tile_width)) / tile_width;
```

floor-truncates when k isn't a multiple of 32, while the device kernels physically gather
`num_cores * ceil(k/32)` tiles: `writer_local_topk.cpp` writes at stride `Kt = ceil(k/32)` tiles per
core, and the gathered CBs c_4/c_5 are sized `Wt_final` tiles. For a direct C++
`ttnn::prim::topk(input, k=50, ...)` call (a public entry point — only the composite front-end
rounds k up to 64; validate only checks `k != 0` and select only `k <= 64`), the model picks
32 cores, `Wt_final = 1600/32 = 50`, but core 31 lands tiles 62–63 → 14 tiles of NoC writes past
c_4 into adjacent L1, and `reader_final_topk.cpp` waits for the semaphore to reach `Wt_final = 50`
while senders count to 64 → hang or race.

This formula predates the PR, but the PR rewrites the function around it, bakes the truncated
`Wt_final` into the new makespan score, and the new gtest case `{8192, 50, 256, 32}` locks a
non-tile-aligned-k config in as correct. Suggested fix: `Wt_final = num_cores * ceil(k/32)` (i.e.
`tt::div_up`), or `TT_FATAL` on non-tile-aligned k at the prim boundary.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -82,7 +85,18 @@ std::optional<TopKCoreConfig> find_topk_core_config(
...
+    // local core does O(split_size) sort work. Model both sides and minimize
+    //   T ~ kLocalCostFactor * Wt_local + kFinalCostFactor * Wt_final
+    // Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells,
+    // <0.5% residual): a local tile costs ~3.5x a final tile — locals run full
+    // 64-element sorts per tile while the final core runs merge/rebuild pair-ops.
+    constexpr uint32_t kLocalCostFactor = 7;
+    constexpr uint32_t kFinalCostFactor = 2;
+    std::optional<TopKCoreConfig> best_config = std::nullopt;
+    uint32_t best_score = std::numeric_limits<uint32_t>::max();
     for (uint32_t split_size = start_split_size; split_size <= max_dim; split_size *= 2) {
```
iwroszTT, line 136: **[should-fix — pre-existing, newly reachable]** **Unsigned underflow in `max_cores` crashes on narrow user sub-grids**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp`, lines 69–71
(same pattern repeated at lines 125–126)

```cpp
const auto max_cores =
    (core_range.end_coord.y - core_range.start_coord.y - 1) * (core_range.end_coord.x - core_range.start_coord.x);
```

on unsigned coords: a single-row `sub_core_grids` like `CoreRange((0,0),(7,0))` wraps
`end.y - start.y - 1` to `SIZE_MAX`, bypasses `TT_FATAL(max_cores > 0)`,
`largest_power_of_two(truncated)` returns 2³¹, `start_split_size = 0` → "Split size must be
non-zero". A two-row range `((0,0),(7,1))` gives `max_cores = 0` → "Core grid must contain at least
one core" directly. Both are pre-existing, but the new gate makes them reachable for widths in
[1024, 8192) that previously always ran single-core — e.g.
`ttnn.topk([1,1,32,2048], k=32, sub_core_grids=CoreRange((0,0),(7,0)))` worked before this PR and
crashes after it. Custom sub-grids also bypass the composite router, so nothing intercepts.

Worth fixing the arithmetic (signed intermediate or explicit `grid_size()`-style helpers) or, at
minimum, falling back to single-core when the grid can't support the multi-core layout.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -82,7 +85,18 @@ std::optional<TopKCoreConfig> find_topk_core_config(
...
-    // Search for optimal split size by trying powers of 2 from conservative start to max_dim
+    // Search all power-of-2 split sizes and keep the one with the best modeled makespan.
+    // The first-valid (= smallest split, most cores) choice maximizes the SERIAL final
+    // stage: the single final core does O(num_cores * k) gather-merge work while every
+    // local core does O(split_size) sort work. Model both sides and minimize
+    //   T ~ kLocalCostFactor * Wt_local + kFinalCostFactor * Wt_final
+    // Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells,
+    // <0.5% residual): a local tile costs ~3.5x a final tile — locals run full
+    // 64-element sorts per tile while the final core runs merge/rebuild pair-ops.
+    constexpr uint32_t kLocalCostFactor = 7;
```
iwroszTT, line 132: **[question]** **Cost factors fitted on one Blackhole silicon are applied to all archs and extrapolated into the new small-shape domain**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp`, lines 96–97

`kLocalCostFactor = 7` / `kFinalCostFactor = 2` were fitted on 4 configs of one p150a at k=64,
widths 8192/32768, tall inputs — yet they now decide core counts for all architectures and for
exactly the new small-width, 1–2-tile-row domain the eligibility change enables, with no constant
per-core overhead term and no arch seam. Wormhole shapes like `[1,1,32,2048]` k=32 flip from a
known single-core baseline to a multi-core config chosen by a Blackhole-fitted ratio; at small
widths/heights, fixed launch/gather-sync costs aren't amortized, so the model plausibly
over-parallelizes there — a silent perf regression no numerical test can catch. Have these cells
been measured on Wormhole? At minimum the constants deserve an arch-aware home (or a measured guard
for WH) rather than function-local literals.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -134,13 +148,16 @@ std::optional<TopKCoreConfig> find_topk_core_config(
                 }
```
iwroszTT: **[should-fix]** **`break` only exits the inner loop — the code ships the *last* arrangement while the comments promise the *first***

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp`, lines 141–150

The `break` at line 147 ("Take the first (largest) valid arrangement found") only exits the inner
x-loop; the outer y-loop keeps scanning and overwrites `selected_x`/`selected_y`, so the code ships
the last (smallest-y) match. On the 13×10 test grid (`max_x=12`, `max_y=8`), `num_cores=16` first
matches (x=2, y=8) — the commented intent — then ends as (8, 2); `num_cores=32` ends as (8, 4) not
(4, 8). The loop predates the PR, but the new model selects different `num_cores` values (16/32
instead of 64), so different transposed rectangles now reach the factory, and the full-sweep search
re-runs the wasted ~96-iteration scan per split. The new gtest asserts split/num_cores but not the
arrangement, so the comment-vs-behavior mismatch is untested.

Suggested: a single loop over y with `x = num_cores / y` and a real early exit, then assert
`selected_x`/`selected_y` in the gtest so the intended orientation is locked down.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
```diff
@@ -91,8 +93,20 @@ TopKDeviceOperation::program_factory_t TopKDeviceOperation::select_program_facto
...
+        (reduced_width >= ttnn::prim::constants::multi_core_min_width) || (num_tile_rows <= 2 && reduced_width >= 1024);
```
iwroszTT: **[should-fix]** **Eligibility thresholds are inline magic literals, and the "is multicore eligible" rule now exists in four divergent copies**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp`, lines 108–109
(other copies: same file line 287; `topk.cpp` lines 309–311;
`ttnn/cpp/ttnn/operations/reduction/topk/docs/TopK.md` lines 39, 110, 125, 137)

The new thresholds (`num_tile_rows <= 2`, `reduced_width >= 1024`) are bare literals while their
siblings live in `topk_constants.hpp`, and the eligibility rule is now spelled four ways:
`select_program_factory`, `validate_on_program_cache_miss` (still the old rule), the composite
router (still the old rule), and TopK.md (lines 39/110/125/137 still say "requires width ≥ 8192" —
the doc explicitly claims to describe `select_program_factory()`'s decision process and now misleads
anyone sizing a low-tile-row workload). Note 1024 is also exactly the crash boundary from the
`start_split_size` comment above, and that knowledge lives nowhere.

Suggested: one named `is_multicore_eligible(...)` helper with constants in `topk_constants.hpp`,
shared by select and validate (which structurally fixes the validation-gap comment above),
referenced by the router, and TopK.md updated.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
```diff
@@ -91,8 +93,20 @@ TopKDeviceOperation::program_factory_t TopKDeviceOperation::select_program_facto
...
+    // The single-core factory parallelizes across tile ROWS, so when the input has
+    // few tile rows most of the grid idles and the column-split multi-core path wins
+    // even below multi_core_min_width: at 1-2 tile rows a width-1024+ row splits
+    // across 8+ otherwise-idle cores (min 64 elements per core), measured ~4x on
+    // 32x2048 k=32. Wide-and-tall inputs keep the row-parallel single-core path.
+    const uint32_t reduced_width = input_tensor.padded_shape()[args.dim];
+    // Use the tensor's actual tile height (the program factories size their work the
+    // same way via tensor_spec().tile()) so custom tile shapes count tile rows correctly.
+    const uint32_t tile_height = input_tensor.tensor_spec().tile().get_height();
+    const uint32_t num_tile_rows = (input_tensor.padded_shape().volume() / std::max<uint32_t>(reduced_width, 1)) /
```
iwroszTT: **[nit]** **`num_tile_rows` hand-inlines `volume()`, re-calls `padded_shape()`, and carries dead guards**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp`, lines 102–107

This computation re-derives `padded_shape()` twice despite the `input_shape` local at line 94,
hand-inlines the volume product, and wraps both divisors in `std::max(..., 1)` guards that can't
fire (validate enforces width ≥ 64; tile height is 16 or 32 by construction) — guards that would
mask a real invariant break with a wrong `num_tile_rows` instead of failing loudly. It's also the
third distinct spelling of "Ht" in the topk op (both program factories compute
`(s0*s1*s2)/tile_height`), so the three can silently diverge under future edits. Simpler:

```cpp
const uint32_t reduced_width = input_shape[args.dim];
const uint32_t num_tile_rows = input_shape.volume() / reduced_width / tile_height;
```

and use `reduced_width` in the `< 65535` and pow2 checks (lines 113/115) too.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -82,7 +85,18 @@ std::optional<TopKCoreConfig> find_topk_core_config(
...
+    // stage: the single final core does O(num_cores * k) gather-merge work while every
+    // local core does O(split_size) sort work. Model both sides and minimize
+    //   T ~ kLocalCostFactor * Wt_local + kFinalCostFactor * Wt_final
+    // Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells,
+    // <0.5% residual): a local tile costs ~3.5x a final tile — locals run full
+    // 64-element sorts per tile while the final core runs merge/rebuild pair-ops.
+    constexpr uint32_t kLocalCostFactor = 7;
+    constexpr uint32_t kFinalCostFactor = 2;
+    std::optional<TopKCoreConfig> best_config = std::nullopt;
+    uint32_t best_score = std::numeric_limits<uint32_t>::max();
```
iwroszTT, line 135: **[nit]** **The now-full-sweep core-config search runs three times per cache-miss dispatch**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp`, line 142
(also validate at line 306, and the factory's own call for `cores_utilized`)

`find_topk_core_config` used to early-exit; it's now a full sweep over all power-of-2 splits, each
paying the ~`max_x*max_y` contiguity scan (~600 wasted iterations per call on a 13×10 grid), and it
still runs three times per cache-miss dispatch (select via `verify_multi_core_cost`, validate, the
factory) plus once per cache-hit dispatch — with the computed config discarded in the first two.
Absolute cost is small, but this sits on the host hot path ttnn works to keep at microseconds, and
the fix is cheap: compute the config once and thread it through, or restore an early-exit existence
check for the verify callers. Note validate currently passes a different `l1_size`
(`largest_free_block_bytes`), which any unification needs to resolve — see the validation-gap
comment above.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -776,6 +778,48 @@ TEST_F(ReductionSmoke, TopkMultiCoreRows64) {
...
+        {8192, 32, 256, 32},
+        {32768, 64, 1024, 32},
+    }};
+    for (const auto& c : cases) {
+        const auto config = ttnn::prim::find_topk_core_config(
+            c.width, min_dim, c.width / 2, c.k, core_range, l1_size, value_tile_size, index_tile_size);
+        ASSERT_TRUE(config.has_value()) << "W=" << c.width << " k=" << c.k;
+        EXPECT_EQ(config->split_size, c.expected_split) << "W=" << c.width << " k=" << c.k;
+        EXPECT_EQ(config->num_cores, c.expected_cores) << "W=" << c.width << " k=" << c.k;
+        EXPECT_EQ(config->rem, 0u) << "W=" << c.width << " k=" << c.k;
```
iwroszTT: **[nit]** **`EXPECT_EQ(config->rem, 0u)` is vacuous, and `expected_cores` is derivable**

**File:** `tests/ttnn/unit_tests/gtests/test_reduction.cpp`, line 819

`find_topk_core_config` only returns configs passing its own `rem == 0` guard (and the test's pow2
widths make rem 0 for every pow2 split anyway), so this assertion can't fail — and it implies `rem`
is a meaningful locked-down output when the factory destructures and never reads it. Meanwhile
`expected_cores` is always `width / expected_split` for returned configs, so any legitimate re-fit
of the cost constants forces editing two coupled numbers per row, with confusing failures when they
disagree. Simpler: drop the rem assertion and assert `config->num_cores == c.width / c.expected_split`.

iwroszTT (changes requested):
**Overall:** the cost-model replacement itself is sound — the fitted makespan model reproduces the expected test configs and genuinely fixes the greedy max-cores-first over-parallelization. My concerns are at the edges of the new eligibility gate: a hard crash regression at width 1024, a validation path that was never updated to match the new gate, a pre-existing non-tile-aligned-k CB overrun that the new test now enshrines, and the newly-enabled 1024–8192 domain shipping with zero device-level CI coverage. Details in the 10 inline comments (line refs against head 1b52752d3); two more below because they touch files outside this diff.

---

### [should-fix] **Composite router's eligibility predicate is now a stale copy — the W=4096 cell the PR targets is shadowed on Blackhole**

**File:** `ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp`, lines 299–312
(the "mirrors select_program_factory requirements #1-#2" comment at lines 300–301, the predicate at
lines 309–311)

`multicore_structurally_ineligible` still treats pow2 padded width < 8192 as structurally
single-core, so on Blackhole a plain-bf16 `ttnn.topk([1,1,32,4096], k=32)` (largest, non-stable,
default grids) is routed to the `topk_large_indices` composite and never reaches the new multi-core
path — while the identical call with an indices tensor, preallocated outputs, a custom grid, or on
Wormhole takes the new multi-core bitonic path. Results stay valid, so this is perf/maintenance
rather than correctness, but the exact low-tile-row cell class this PR targets is shadowed under
default flags, and both the "mirrors" comment and "the lone pow2 cell in [4096, 8192) otherwise
stays on the linear factory" (line 307–308) are now false — which sets up future divergent edits.

Either update the predicate to account for the Ht-aware gate (needs tile-row info here) or, at
minimum, correct the comments to say the mirror is intentionally conservative and why.

---

### [should-fix] **"still single-core" test case now silently runs the multi-core factory; new 1024–8192 domain has zero device coverage**

**File:** `tests/ttnn/unit_tests/operations/reduce/test_topk.py`, line 478

`(1, 1, 32, 4096, 3, 32),  # larger dim, still single-core` in `test_topk_int32_indices` now
satisfies the new gate (`num_tile_rows = 1 <= 2`, width 4096 ≥ 1024), and the test's preallocated
outputs bypass the composite router — so it runs the multi-core factory. The test still passes, but
the comment is wrong and the intended larger-dim single-core int32 coverage is silently lost.

Bigger picture: the PR's only new test is host-side at W=8192/32768, so the headline new behavior
(e.g. the ~4x 32×2048 cell) and the crashing 32×1024 cell ship with no device-level CI signal at
all. Please add device tests for the newly-enabled domain — at least 32×1024 (currently crashes,
per the comment on `start_split_size`), 32×2048, and 32×4096 — and restore a genuinely single-core
larger-dim case here (a non-pow2 width like 4224, or H ≥ 96).

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -776,6 +778,48 @@ TEST_F(ReductionSmoke, TopkMultiCoreRows64) {
...
+// Host-side lock on the multi-core split model in find_topk_core_config: the
+// fitted makespan score (kLocalCostFactor * Wt_local + kFinalCostFactor *
+// Wt_final) must keep selecting the measured-fastest split, not the
+// first-valid (smallest-split / most-cores) config the old greedy search
+// returned. Pure function of its arguments -- no device needed. The core
+// range and tile sizes mirror what the multi-core factory passes on a p150a
+// (13x10 compute grid, bf16 values, uint16 indices, BH L1); the expected
+// splits are the configs the model was fitted against on that silicon.
+// If a coefficient or the search order changes, these assertions catch it
+// even though every numerical topk test stays green.
```
fplavecTT, line 812: Can we improve this comment block:
1. "not the first-valid (smallest-split / most-cores) config the old greedy search returned". Is this retrospective useful to a future reader? Perhaps it can be useful as a warning not to take that approach in the future? If so, it should be written as such rather than narration of this change (narration may be suitable for PR description).
2. "Pure function of its arguments -- no device needed" - What does this mean and why is it relevant?
3. "the model was fitted against on that silicon" - What model fitting is this referring to?
4. "If a coefficient or the search order changes" - This sentence needs some context. Which coefficient or search order is this referring to? I don't see the word coefficient mentioned anywhere else in this PR.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -776,6 +778,48 @@ TEST_F(ReductionSmoke, TopkMultiCoreRows64) {
...
+    const tt::tt_metal::CoreRange core_range({0, 0}, {12, 9});                      // p150a 13x10 grid
+
+    struct Case {
+        uint32_t width;
+        uint32_t k;
+        uint16_t expected_split;
+        uint16_t expected_cores;
+    };
+    // Expected values: minimum of 7 * Wt_local + 2 * Wt_final over all valid
+    // power-of-two splits. The old first-valid picks were 128/64 (W=8192) and
```
fplavecTT: "The old first-valid picks were"...
Is this relevant to a future reader of this comment?

tests/ttnn/unit_tests/gtests/test_reduction.cpp
iwroszTT, line 1: Worth checking the Merge Gate pipeline

tests/ttnn/unit_tests/gtests/test_reduction.cpp
```diff
@@ -776,6 +778,48 @@ TEST_F(ReductionSmoke, TopkMultiCoreRows64) {
...
+// (13x10 compute grid, bf16 values, uint16 indices, BH L1); the expected
+// splits are the configs the model was fitted against on that silicon.
+// If a coefficient or the search order changes, these assertions catch it
+// even though every numerical topk test stays green.
+TEST(TopkCoreConfigModel, SelectsFittedMakespanMinimum) {
+    constexpr uint32_t l1_size = 1536 * 1024;                                       // BH L1 per core
+    constexpr uint32_t value_tile_size = tt::tile_size(tt::DataFormat::Float16_b);  // 2048
+    constexpr uint32_t index_tile_size = tt::tile_size(tt::DataFormat::UInt16);     // 2048
+    constexpr uint32_t min_dim = ttnn::prim::constants::min_dim_per_core;           // 64
+    const tt::tt_metal::CoreRange core_range({0, 0}, {12, 9});                      // p150a 13x10 grid
```
fplavecTT: I thought every p150a has 12x10 grid. See https://github.com/tenstorrent/tt-metal/issues/47014

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Fixed in e7013ac814d. `start_split_size` is clamped to `max(min_dim, tile_width)` before the sweep and the in-loop `TT_FATAL` is gone — the truncate-to-zero can no longer happen. On your exact configuration (BH 13×10 → `max_cores = 96`, `lp2 = 64`, 32 tiles) W=1024 now returns a valid config (split=128, 8 cores) instead of throwing, i.e. the shape doesn't just stop crashing, it actually gets the multi-core path the new gate intended.

Locked down twice:
- `TopkCoreConfigModel` now includes `{1024, 32, split 128}` on the 13×10 range (with a comment naming this as the former crash cell);
- new device test `test_topk_low_tile_row_multicore` runs 32×1024 (k=32 and k=64, both `largest` polarities) — verified green on p150a silicon with the default 13×10 grid, which crashed before the fix.

Grids that genuinely cannot host the multi-core layout return `nullopt` → single-core fallback (see the underflow thread for that half).

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: Fixed in e7013ac814d, structurally, via the shared predicate you proposed in the four-copies thread: `topk_multicore_structurally_eligible()` (topk_utils.cpp) is now the single gate used by `select_program_factory`, `validate_on_program_cache_miss`, and the composite router — validate's multi-core branch can no longer diverge from select's. On your three concrete gaps:

- The allocator-aware `largest_free_block_bytes` check now runs for exactly the shapes that will take the multi-core factory, including the whole 1024–8192 low-tile-row domain, so the L1-pressure case gets the intended "Not enough cores or cache size available" `TT_FATAL` instead of an opaque allocation failure.
- The `sub_core_grids.ranges().size() == 1` `TT_FATAL` now covers the new domain (same branch). Select additionally checks `ranges().size() == 1` itself before `.at(0)` and falls back to single-core — that guard matters because program-hash computation can reach `select_program_factory` before validation has run, which was the escape path for your third point.
- An empty `CoreRangeSet` is caught by the same `size() == 1` checks (validate: loud FATAL; select: no more raw `std::out_of_range` at hash time).

One deliberate non-unification: select keeps `l1_size_per_core()` and validate keeps `largest_free_block_bytes` — select feeds the program hash and must stay allocator-state-independent, while validate is exactly the place to be allocator-aware. Both now sit behind the same structural gate, which was the actual drift hazard.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
nkapreTT: Fixed in e7013ac814d with your first suggestion: `Wt_final = num_cores * tt::div_up(k, tile_width)` and `final_input_size = Wt_final * tile_width`, matching the `Kt`-stride the writer kernels physically gather — so for k=50/32 cores the gather CBs and the final reader's semaphore target are 64 tiles, same as the senders' count. A side effect worth naming: `Wt_final` is now a power of two whenever `num_cores` and `Kt` are, which the factory's `log2(Wt_final)` merge-iteration count silently relied on (also broken at 50 tiles before).

The gtest `{8192, 50, ...}` row now expects the same config as k=64 (split 512, 16 cores) instead of locking the truncated formula in, with a comment saying why, and the k=50 int32 device test ran green on silicon with the corrected sizing.

Exposure note for the record: the composite front-end rounds k up to a multiple of 32 (`adjusted_k`) before reaching the prim, so only direct `ttnn::prim::topk` C++ callers could hit this. I went with making it correct rather than `TT_FATAL`-ing at the prim boundary since the complete fix is two lines; happy to add the FATAL on top if you'd rather forbid non-tile-multiple k at that entry point outright.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Fixed in e7013ac814d. The grid math computes the inclusive sizes first (`grid_x/grid_y = end − start + 1` — no unsigned subtraction that can wrap), then requires `grid_x >= 2 && grid_y >= 3` (local rectangle + final-core row) and returns `std::nullopt` otherwise — so both of your repros fall back to single-core exactly as pre-PR: `CoreRange((0,0),(7,0))` no longer wraps to `SIZE_MAX`, and `((0,0),(7,1))` no longer trips the `max_cores > 0` FATAL. The duplicated recomputation at the old lines 125–126 is hoisted/gone.

Both small-grid cases are locked in `TopkCoreConfigModel` (`EXPECT_FALSE(config.has_value())`). Also hardened the third copy of that expression you didn't flag but that had the same problem: the factory's `cores_utilized` fallback, which fabricated a config with `selected_x = 0` (failing later as a confusing "Failed to select local cores range") — since select only picks the multi-core factory after `verify_multi_core_cost` succeeded with identical arguments, reaching that fallback means a select/factory disagreement, and it's now a loud `TT_FATAL` naming width/k/grid.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Fair question — answering with the measurement record, then a proposal.

**What the constants are.** 7/2 ships the fitted ratio a/b = 5.85/1.68 ≈ 3.48 from the p150a fit (4 configs, W=8192/32768 at k=64; <0.5% residual, and the model then predicted every config in the measured grid within 0.03–5%). Since the search only *compares* scores, only the ratio matters — which is also why there's no constant per-config overhead term: it cancels. What does **not** cancel is a per-**core** launch/gather-sync term, and that's the real extrapolation risk you're pointing at for small shapes.

**Exposure by domain.**
- For W ≥ 8192 (all archs): the model only reorders choices among configs that were *already* multi-core, so the worst case on Wormhole is a less-optimal split within a path it already took — not a path flip. The old first-valid rule (max cores, max serial final stage) it replaces was itself never fitted on any arch; on BH the reorders measured −37% (8192) and −4.5% (32768).
- For the new 1024–8192 low-tile-row domain: the *mechanism* is arch-independent — the single-core factory parallelizes across tile rows, so at Ht ≤ 2 it uses ≤ 2 cores of any grid while the column split uses 8–32, and `min_dim_per_core = 64` bounds under-amortization. But the claim "multi-core wins here" is only **measured on BH** (32×2048 k=32: 286 → 47.8 µs; 32×4096 k=64: 1072 → 103.7 µs — a fixed per-core cost would have to be huge to eat a 6–10× margin, though WH dispatch/NoC constants do differ). I don't have WH silicon on this box to close that today.

**Proposal — your preference decides which lands:**
(a) gate the low-Ht relaxation to Blackhole for now (arch check where the eligibility helper is evaluated, or an arch-conditional `multi_core_low_ht_min_width`), leaving Wormhole at exactly pre-PR behavior until measured;
(b) keep it cross-arch on the mechanism argument, and I grab a WH reservation and run the same cells before merge;
(c) land BH-gated now, lift the gate in a follow-up once WH numbers exist.

I lean (a)/(c) as the conservative default but didn't want to add arch-conditional behavior unilaterally mid-review. Either way, the constants now live next to their provenance (fitted-model comment in topk_utils.cpp) and an arch seam has one obvious home — the shared helper + `topk_constants.hpp` — rather than four call sites (e7013ac814d).

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Verified before touching it, because the answer decides whether this is a comment fix or a behavior fix. Since `selected_x/selected_y` feed neither the makespan score nor the validity predicate (any factorization sets `contiguous_cores_available`), the wrong-scoped break can never change which `(split, num_cores)` wins. Confirmed empirically with a 168-cell host-side sweep (grids 13×10 / 12×9 / 8×8 / 7×8 / 8×9 / 8×1 / 8×2 × widths 1024–32768 × k ∈ {16, 32, 50, 64}) running faithful ports of both loop variants: **0 config differences**.

What it *does* change is the rectangle orientation — and there the shipped behavior (last match = widest / smallest-y, e.g. num_cores=16 → (8,2)) is what every silicon measurement in this PR, including the cost-constant fit, was taken against. "Fixing" the break to the comment's first-match intent would have silently flipped the orientation on 41/168 modeled cells. So in e7013ac814d I locked the shipped behavior and fixed the comment instead: the search is now a single ascending-y divisor loop with a real early exit (`x = num_cores / y`, first fit wins — provably identical to the old double loop's final state), the comment says "widest (smallest-y)" explicitly and records the old loop's history, and `TopkCoreConfigModel` asserts `selected_x/selected_y` for all seven table rows as you suggested. Bonus: the scan is O(max_y) instead of ~max_x·max_y per split, which also feeds the 3×-per-dispatch thread.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: Done in e7013ac814d along exactly these lines:

- Thresholds named in `topk_constants.hpp`: `multi_core_low_ht_min_width = 1024`, `multi_core_low_ht_max_tile_rows = 2`, `multi_core_max_width_exclusive` (65535), `multi_core_max_k` (64) — each with a comment saying why the value is what it is (1024's comment records it's the smallest pow2 width giving every core ≥ `min_dim_per_core` at a useful core count; the crash boundary you noted is gone per the `start_split_size` thread).
- One predicate, `topk_multicore_structurally_eligible()` in topk_utils, now used by `select_program_factory`, `validate_on_program_cache_miss` (which structurally fixes the validation-gap finding), and the composite router — which now also computes tile rows, so the W=4096 low-Ht cell is no longer shadowed on Blackhole (covered on-device by the new `test_topk_low_tile_row_multicore`). The "mirrors requirements #1–#2" comments are gone; the router comment now names the shared symbol.
- TopK.md: lines 39/110/125/137 and the Strategy Selection section rewritten to describe the Ht-aware gate and to name the shared predicate + constants file.

One copy that intentionally remains: `models/common/sampling/_utils.py` carries a Python-side mirror of the *routing* predicate (cross-language, can't share a C++ symbol) with its own KEEP IN SYNC comment and a documented fail-safe drift contract — this PR makes it conservatively stale for pow2 low-Ht widths in [4096, 8192) (mirror predicts route, C++ now keeps the eligible bitonic path), which per that contract degrades to the unchanged stock call; syncing it rides the sampling PR line, not this one.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: Done in e7013ac814d as suggested: `input_shape.volume() / reduced_width / tile_height` on the existing `input_shape` local, and the `std::max(..., 1)` guards are gone — agreed that a broken invariant should fail loudly rather than yield a plausible-but-wrong `num_tile_rows` (validate enforces width ≥ 64 and tile height is fixed by construction). `reduced_width` now feeds the < 65535 and pow2 checks too, via the shared eligibility helper.

On the third-spelling concern: select and validate now share the helper (and its inputs), so those two can't diverge. The program factories still compute `Ht = (s0*s1*s2)/tile_height` for their own work sizing — same quantity by construction for the 4D shapes validate enforces (`volume/width == s0*s1*s2`), and they need it independently of eligibility, so I left that as is rather than exporting a third helper for a two-line arithmetic identity.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Addressed in e7013ac814d, at compute-cheap rather than full compute-once:

- Existence is independent of the makespan score (any valid split proves feasibility), so `verify_multi_core_cost` now early-exits at the first valid config — restoring the pre-PR early-exit behavior for both the select and validate callers. The full scored sweep runs once, in the factory.
- The per-split contiguity scan dropped from ~max_x·max_y (~96 iterations) to an O(max_y) divisor walk with a real early exit (see the break-scope thread), so even the factory's full sweep is much lighter.

I stopped short of threading the chosen config select → validate → factory: the device-op protocol is stateless static methods keyed by `operation_attributes_t`, and stashing derived state there would put it in the program hash. If the residual cost ever shows up on a dispatch profile, memoizing inside `find_topk_core_config` on its argument tuple is the next cheap step. The validate-vs-select `l1_size` asymmetry (`largest_free_block_bytes` vs `l1_size_per_core`) is kept deliberately — select feeds the program hash and must stay allocator-independent; noted in the validation thread.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
nkapreTT: Done in e7013ac814d: the `rem` assertion is dropped, and `num_cores` is asserted as `c.width / c.expected_split` — so a legitimate re-fit edits one number per row. Per your suggestion in the break-scope thread the table now also locks `selected_x/selected_y` (orientation), and it gained the low-Ht domain rows (1024/2048/4096, including the formerly crashing W=1024 cell) plus two small-grid cases asserting `nullopt` (single-row and two-row `sub_core_grids` → single-core fallback).

nkapreTT:
@iwroszTT — thanks for the depth of this round; the two blocking findings were real bugs the PR's own gate made reachable, and the shared-predicate suggestion turned out to fix the validation gap structurally. Everything is addressed in **e7013ac814d** (pushed). Per-finding status, including the two body-level items:

| Finding | Status | Where |
|---|---|---|
| [blocking] W=1024 crash (`start_split_size` → 0) | **Fixed** — clamp to `max(min_dim, tile_width)`; W=1024 gets a real multi-core config on 13×10 BH | inline thread + gtest + on-silicon device test |
| [blocking] validate still gated at ≥ 8192 | **Fixed** — shared `topk_multicore_structurally_eligible()` gates select, validate, and the router; allocator-aware check + core-range FATALs now cover the new domain; hash-time `.at(0)` guarded | inline thread |
| Wt_final floor-truncation (k not ×32) | **Fixed** — `num_cores * div_up(k, 32)`; also restores the pow2 invariant behind `log2(Wt_final)` | inline thread |
| unsigned underflow on narrow sub-grids | **Fixed** — inclusive-size grid math, `nullopt` → single-core fallback; factory's fabricated-config fallback is now a loud FATAL | inline thread |
| break exits inner loop only | **Verified then fixed** — 168-cell sweep: 0 `(split, num_cores)` changes; shipped (widest / smallest-y) orientation locked with a single-loop early exit + gtest orientation asserts (comment-intent would have flipped 41/168 cells the silicon fit was measured on) | inline thread |
| magic literals / four divergent copies | **Fixed** — constants in `topk_constants.hpp`, one shared predicate, TopK.md rewritten | inline thread |
| full sweep 3× per dispatch | **Fixed** — verify callers early-exit at first valid split; contiguity scan now O(max_y); full scored sweep runs once (factory) | inline thread |
| `num_tile_rows` hand-inlined / dead guards | **Fixed** — as suggested | inline thread |
| vacuous `rem` assert / derivable cores | **Fixed** — cores derived from split; orientation + new-domain + nullopt rows added | inline thread |
| [question] BH-fitted constants on all archs | **Answered + proposal** — measurement record and three options (BH-gate now / cross-arch + WH measurement / gate-then-lift); your call | inline thread |
| [body] router predicate stale copy | **Fixed** — router uses the shared helper with tile-row info; the W=4096 default-flags BH cell now reaches the multi-core path (covered on device); no more hand-mirrored comment | `topk.cpp` |
| [body] test comment wrong + zero device coverage for 1024–8192 | **Fixed** — int32 test: W=4224 (non-pow2) restored as genuinely single-core, W=4096 relabeled and kept as a multi-core int32/uint32 cell; new `test_topk_low_tile_row_multicore` covers 32×1024 (k=32/64), 32×2048 (k=32/64), 32×4096 (k=32), 64×4096 (k=64) × both `largest` polarities, checking value sets and that indices address the returned values | `test_topk.py` |

**Validation on p150a (13×10) silicon, fresh JIT cache:** full `test_topk.py`: **262 passed, 8 skipped, 80 xfailed** (xfails pre-existing simulator/known-issue markers); the touched gtests: **50/50 passed** (`ReductionSmoke.*` + `TopkCoreConfigModel.*`); the 28 tests covering the new/changed cells (low-tile-row + int32) re-run explicitly: **28/28**. The formerly-crashing 32×1024 cell and the k=50 multi-core cell (corrected `Wt_final`) both ran on device.

The one open item is the arch question — happy to push option (a)'s BH gate immediately if that's your preference.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
```diff
@@ -82,7 +85,18 @@ std::optional<TopKCoreConfig> find_topk_core_config(
     const uint32_t bf16_tile_size = tt::tile_size(tt::DataFormat::Float16_b);
     const uint32_t transposed_tile_size = std::max(value_tile_size, bf16_tile_size);
 
-    // Search for optimal split size by trying powers of 2 from conservative start to max_dim
+    // Search all power-of-2 split sizes and keep the one with the best modeled makespan.
+    // The first-valid (= smallest split, most cores) choice maximizes the SERIAL final
+    // stage: the single final core does O(num_cores * k) gather-merge work while every
+    // local core does O(split_size) sort work. Model both sides and minimize
+    //   T ~ kLocalCostFactor * Wt_local + kFinalCostFactor * Wt_final
+    // Constants fitted on p150a silicon (4 configs across 8192/32768-wide k=64 cells,
```
fplavecTT, line 124: What about impact on Wormhole? This should be mentioned (in human-understandable language) either here or in PR description.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
iwroszTT: Let's do (a)

iwroszTT (approved):

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
fplavecTT: Doesn't this also need to change to be in sync with changes made to `select_program_factory` above?

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
fplavecTT: `multi_core_min_width` is a named constant, which is good, but `1024` is used as a plain literal. Is there a meaningful name we could use and define a constant for it? Same for `2` in `num_tile_rows <= 2`.

fplavecTT:
I asked Claude to review this change, and in addition to other comments I made, it found at least three places where comments or documentation need updating as a results of this change. Claude says:
1. The Blackhole routing predicate claims to mirror requirement #1 and no longer does. `topk.cpp:303-312` computes `multicore_structurally_ineligible` as power of two below `multi_core_min_width`, with the comment "the lone pow2 cell in [4096, 8192) otherwise stays on the linear factory". With this PR that cell no longer stays on the linear factory when it has one or two tile rows, so on Blackhole a default bf16 largest call at W=4096 is still routed to the `topk_large_indices` composite even though the stock multi-core path can now take it faster.*
3. *docs/TopK.md was not synced. It still states the multi-core width floor as 8192 in four places (lines 39, 110, 125, 137) and describes the split search as maximizing parallelism.*

This then made me wonder whether docstring in nanobind file also has the same problem, and it seems so. In response to my question, Claude said: 
*Yes, exactly one line, and it is in `topk_nanobind.cpp:87`:*
*. To enable multicore execution, the width of :attr:`input_tensor` along :attr:`dim` must be ≥8192 and <65536, and :attr:`k` must be ≤64.
That is now wrong in the same way docs/TopK.md is*

fplavecTT (changes requested):

tests/ttnn/unit_tests/gtests/test_reduction.cpp
nkapreTT: Rewritten in b83215d0668 as present-tense contract language (what the model selects and why the expectation holds); the before/after narration is gone from the code.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
nkapreTT: Agreed — deleted in b83215d0668; the history lives in the PR description, not the comment.

tests/ttnn/unit_tests/gtests/test_reduction.cpp
nkapreTT: Fair point — grids are per-unit harvesting outcomes: this fitting box's p150a queries 13x10, while #47014's 12x10 is correct for other/production units. That variability is exactly why the test asserts the grid-free contract unconditionally and gates the tight counts on the QUERIED grid rather than a constant. Comment reworded in b83215d0668 to "e.g. 13x10 on the fitting box; grids vary per unit" so it can't read as a universal claim.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_utils.cpp
nkapreTT: Added a plain-language note at the clamp in b83215d0668: the clamp is arch-neutral and only binds at the eligibility floor; Wormhole's 32-class largest_power_of_two grids are unaffected by the W=1024 zero-split case, and the Ht gate applies identically on WH. The fitted-constants question for WH remains the separate thread above (three options offered — your and iwrosz's pick).

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: It did need to be in sync — and that consolidation landed in the previous round (e7013ac814d): both select_program_factory and this site now call the same shared topk_multicore_structurally_eligible() predicate from topk_utils, so they cannot diverge. Your review likely crossed that push.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: Fixed in e7013ac814d: the literal is now multi_core_low_ht_min_width in topk_constants.hpp with its derivation comment (smallest pow2 width where the multi-core path measured a win), and no raw 1024 remains in the op (grep-verified). Same crossed-push timing.

fplavecTT:
`ttnn/cpp/ttnn/operations/reduction/topk/topk_nanobind.cpp` still has stale info I pointed out before:

*To enable multicore execution, the width of :attr:input_tensor along :attr:dim must be ≥8192 and <65536, and :attr:k must be ≤64.*

I don't believe this is accurate any more.

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
nkapreTT: Addendum for completeness: the `2` is also named — `constants::multi_core_low_ht_max_tile_rows` in `topk_constants.hpp`, consumed by the shared `topk_multicore_structurally_eligible()` predicate (topk_utils.cpp:35) with its derivation comment. Neither literal survives anywhere in the op.

ttnn/cpp/ttnn/operations/reduction/topk/topk.cpp
fplavecTT, line 561: This comment may also be stale.

nkapreTT:
Serviced item by item (all in b2b17b92f01):

**1. Routing predicate / the W=4096 pow2 low-tile-row cell** — measured instead of guessed. Tracy device-kernel-duration A/B on p150a silicon, bf16 `largest` W=4096, 3 trials per arm with a fresh kernel cache each (per-op sums for the composite, cross-trial spread <0.2%):

| cell (W=4096, bf16, largest) | stock multi-core bitonic | `topk_large_indices` composite | winner |
|---|---|---|---|
| 32 rows (1 tile row), k=32 | 70.3 µs | 46.8 µs | composite, 1.50x |
| 32 rows (1 tile row), k=64 | 106.5 µs | 47.0 µs | composite, 2.27x |
| 64 rows (2 tile rows), k=32 | 103.3 µs | 47.9 µs | composite, 2.16x |
| 64 rows (2 tile rows), k=64 | 165.5 µs | 48.0 µs | composite, 3.45x |

The composite wins every affected cell, so you were right that the routing should keep taking it. The router's small-k arm now evaluates the shared `topk_multicore_structurally_eligible()` predicate with the Ht-aware relaxation deliberately disabled (`num_tile_rows` pinned above `multi_core_low_ht_max_tile_rows`), which restores merged-main routing behavior for that cell and keeps the router aligned with the `models/common/sampling/_utils.py` mirror (`topk_would_route_to_large_indices` — verified semantically identical by construction: both reduce to "padded width ≥ 4096 and (non-pow2, or ≥ 65535, or < 8192)" for k ≤ 64, so the mirror needs no change on merge). The comments at the predicate, the constants block, and the call site now state this true reason with the measured numbers. The Ht-aware eligibility still pays off: it upgrades the cell (and pow2 widths in [1024, 4096), which the composite cannot take) from single-core to multi-core whenever routing declines — `sub_core_grids`, preallocated outputs, non-bf16, `smallest`, `stable`. Tests: `test_topk_low_tile_row_multicore` now pins its W=4096 `largest=True` cells to the stock path via `sub_core_grids` so the multi-core factory keeps device coverage there, and a new test locks correctness of the routed default-args cell.

**2./3. docs/TopK.md** — the four 8192-floor sites you cited (old lines 39, 110, 125, 137) were synced in e7013ac814d in the previous round; this round finishes the remainder: the split search is now documented as the fitted min-makespan selection (formula `T ≈ 7·Wt_local + 2·Wt_final`, constants, and why the most-cores/first-valid choice is not the winner), the strategy table no longer claims "highest parallelism", and the multi-core walkthrough example is recomputed to the split the model actually picks (split_size=512, 32+1 cores, with the scores shown).

**Nanobind docstring** — you flagged this twice and the second flag came after my replies; that one's on me — the b83215d0668 "docs/comments round" should have caught it and didn't. Rewritten now to the shipped contract with the `topk_constants.hpp` names inlined (`multi_core_max_k`, `multi_core_max_width_exclusive`, `multi_core_min_width`, `multi_core_low_ht_min_width`, `multi_core_low_ht_max_tile_rows`) so it can't rot the same way, plus a pointer to the Blackhole routing layer.

Test tally: `tests/ttnn/unit_tests/operations/reduce/test_topk.py` full — 264 passed, 8 skipped, 80 xfailed; routed-band cells in `test_reduction.py -k topk` — 12 passed (BH p150a silicon, fresh kernel cache).

nkapreTT:
You're right, and this was the second time you had to point at the same line — the "docs/comments round" (b83215d0668) that replied to your first comment should have fixed it and didn't. No excuse for that miss. Fixed in b2b17b92f01; the bullet now reads:

> To enable multicore execution, `k` must be ≤64 (`multi_core_max_k`) and the width of `input_tensor` along `dim` must be a power of two, below 65535 (`multi_core_max_width_exclusive`), and either ≥8192 (`multi_core_min_width`) or — when the input has at most 2 tile rows (`multi_core_low_ht_max_tile_rows`) — ≥1024 (`multi_core_low_ht_min_width`).

It mirrors the `topk_constants.hpp` constant names, so a future threshold change that doesn't touch this line will at least be greppable, and it notes the Blackhole composite routing layer. The measured routing follow-up from your other comment is in the same push — A/B table in my reply above.

ttnn/cpp/ttnn/operations/reduction/topk/topk_nanobind.cpp
```diff
@@ -84,7 +84,7 @@ void bind_reduction_topk_operation(nb::module_& mod) {
...
                 - The padding is currently only supported for bfloat16, float32, int32, and uint32.
-                - To enable multicore execution, the width of :attr:`input_tensor` along :attr:`dim` must be ≥8192 and <65536, and :attr:`k` must be ≤64.
+                - To enable multicore execution, :attr:`k` must be ≤64 (`multi_core_max_k`) and the width of :attr:`input_tensor` along :attr:`dim` must be a power of two, below 65535 (`multi_core_max_width_exclusive`), and either ≥8192 (`multi_core_min_width`) or — when the input has at most 2 tile rows (`multi_core_low_ht_max_tile_rows`) — ≥1024 (`multi_core_low_ht_min_width`). These thresholds are named in `topk_constants.hpp` and enforced by the shared `topk_multicore_structurally_eligible()` predicate; multicore additionally requires the grid/L1 feasibility checks to pass, otherwise the op falls back to single-core. (On Blackhole, default bf16 `largest` calls in certain width/k bands are transparently dispatched to a faster composite pipeline instead — see `should_route_to_topk_large_indices` in `topk.cpp`.)
```
fplavecTT: This is user-facing documentation, so this is not an appropriate place to state where something is enforced because it is not relevant to API user.
Can a human rewrite this so that it's suitable for a human reader?

fplavecTT:
If I understand PR description correctly, 32x4096 case was called by passing explicit `sub_core_grids`, right? If so, then the comparison is of single-core main implementation vs multi-core implementation with this PR. However if user doesn't specify `sub_core_grids` at all (common case), on Blackhole they would have gotten topk_large_indices on main for this shape, not single-core. I believe after this PR changes default to go to multi-core. given this, should we also compare topk_large_indices on main vs. multi-core implementation with this PR (I have no sense for how the two compare, so maybe this doesn't make sense - leaving it to your judgment).

fplavecTT (approved):
Approving this so I'm not holding it up longer, but nanobind doc definitely needs "human touch". Also, perhaps another perf comparison (assuming the default behavior change when no sub_core_grids is specified is intentional), as I've noted

nkapreTT:
> If I understand PR description correctly, 32x4096 case was called by passing explicit `sub_core_grids`, right?

You read it right: the 32x4096 headline row was measured with an explicit full-grid `sub_core_grids` on both sides, so it compares single-core vs multi-core in a routing-declined context — it says nothing about the default path. Your default-path question is exactly the comparison the router decision was based on (numbers in the earlier measurement comment): `topk_large_indices` composite vs the new multi-core at W=4096, Tracy device kernel duration, p150a, bf16, `largest=True`, 3 trials per arm, fresh kernel cache each trial, per-op sums for the composite:

| shape | composite (main & this PR) | new multi-core | ratio |
|---|---|---|---|
| rows=32, k=32 | 46.8 µs | 70.3 µs | 1.50x |
| rows=32, k=64 | 47.0 µs | 106.5 µs | 2.27x |
| rows=64, k=32 | 47.9 µs | 103.3 µs | 2.16x |
| rows=64, k=64 | 48.0 µs | 165.5 µs | 3.45x |

The composite wins every cell, so b2b17b92f01 pins the router accordingly: a default bf16 call at these shapes still routes to the composite, same as main — **default Blackhole behavior is unchanged by this PR**. The widened multi-core arm pays off where routing declines (explicit `sub_core_grids`, preallocated outputs, non-bf16, `stable=True`, `largest=False`) and for pow2 widths in [1024, 4096), which sit below the composite's floor — those cells go from single-core to multi-core (~4x on 32x2048 k=32).

PR #53795, opened 2026-08-20
ttnn.topk: route the pow2 [8192, 65535) k<=64 band to topk_large_indices on BH (up to 3.9x)

nkapreTT:
### What

Extends the Blackhole small-k routing arm in `should_route_to_topk_large_indices` (`topk.cpp`) to also route **pow2 widths in [8192, 65535) with k <= 64** — the one small-k class deliberately left to the stock multi-core bitonic when the route was introduced.

### Why the band was originally excluded, and what changed

The small-k arm only claimed cells the stock multi-core bitonic *cannot* take (non-pow2, >= 65535, or below its 8192 width floor), on the assumption that an eligible bitonic is the right engine. Fitting a makespan model to that bitonic for #53793 showed why it isn't: its final gather-merge stage is **serial** and scales with `num_cores * k`, so even the best split leaves 2x+ on the table against a chain that stays column-parallel end to end. The routed composite (`topk_route_prep` → `topk_large_indices` → `topk_route_finish`) measured faster at both band anchor cells — including against #53793's improved stock numbers.

### Measured (Tracy device kernel duration, 3 trials, zero cross-trial variance, p150a; routed side = full prep+op+finish+slice chain)

| shape | k | main stock (mc bitonic) | routed (this PR) | delta | vs #53793 stock |
|---|---|---|---|---|---|
| 32x8192 | 50 | 263.5 us | 67.4 us | **3.9x** | 155.7 us → **-57%** |
| 32x32768 | 64 | 333.8 us | 178.6 us | **1.9x** | 302.2 us → **-41%** |

Stock arm forced via a full-grid `sub_core_grids` on the same build (identical device-op behavior to a default main call).

### Guards / contract

Every other routing guard is unchanged: bf16, TILE, not sharded, `largest`, `!stable`, no `sub_core_grids` / user index tensor / preallocated outputs, last-dim reductions, existing width envelope. Explicit `sub_core_grids`, preallocated-output, 32-bit-index, and `stable=True` callers keep the stock bitonic exactly as today.

KEEP-IN-SYNC note: the `models/common/sampling` Python mirror of this predicate (`topk_would_route_to_large_indices`) is not on main (it ships with the sampling-integration PR), so there is no mirror to update in this change; the branch that carries it must fold the band in when it lands.

### Validation

- `tests/ttnn/unit_tests/operations/reduce/test_reduction.py` + `test_topk.py`: **660 passed, 32 skipped, 80 xfailed** (one unrelated `test_std` flake from a shared-JIT-cache rename race passed 160/160 on rerun).
- New `test_topk_small_k_routed_bitonic_band` covers (k=32, W=8192), (k=50, W=8192 — the k % 16 != 0 rounding tail), (k=64, W=32768).
- `test_topk_large_k_routing_engages` updated: asserts the band **routes** by graph capture, and uses a full-grid `sub_core_grids` call as the guaranteed-stock probe.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/topk-route-mc-band)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/topk-route-mc-band)

### DIFF
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -707,6 +707,16 @@ def test_topk_large_k_routed(k, W, device):
     run_topk_large_k_routed_test(1, 1, 32, W, k, device)
 
 
+@pytest.mark.skipif(not is_blackhole(), reason="large-k routing is Blackhole-only")
+@pytest.mark.parametrize("k, W", [(32, 8192), (50, 8192), (64, 32768)])
+def test_topk_small_k_routed_bitonic_band(k, W, device):
+    # Bitonic-band arm of the small-k route: pow2 widths in [8192, 65535) with
+    # k <= 64 are eligible for the stock multi-core bitonic but route to the
+    # topk_large_indices composite (measured faster; see topk.cpp). k=50
+    # exercises the k % 16 != 0 rounding tail inside the band.
+    run_topk_large_k_routed_test(1, 1, 32, W, k, device)
+
+
 @pytest.mark.skipif(
     not is_blackhole(), reason="large-k routing is Blackhole-only; stock single-core takes minutes at these shapes"
 )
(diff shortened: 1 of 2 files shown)


nkapreTT:
Copilot review audit (6aaecb88f2a) — both findings verified and addressed (doc-only, no behavior change):

1. **Routed-test section header** (`test_topk.py`) — **valid, fixed.** The suite header now describes both routed regions: the large-k `64 < k <= 2048` route and the small-k arm (structurally-ineligible widths plus the bitonic-eligible pow2 `[8192, 65535)` band this PR adds), and keeps the prep/finish chain description that applies to every routed call.
2. **Routing overview comments** (`topk.cpp`) — **valid, fixed.** Both the comment block at `large_k_route_min_k_exclusive` and the call-site routing summary now list the bitonic-band arm as a routed region instead of claiming eligible pow2 cells always keep the stock bitonic.

Validation (p150a): `test_topk_small_k_routed_bitonic_band` (3 cells) + `test_topk_large_k_routing_engages` re-run green on the updated branch.

nkapreTT:
Rebased onto current main (`f54cf30f496`, including merged #53793) at head `8152ec9e896`. Conflict resolution preserves #53793’s shared structural-eligibility/Ht-aware logic and adds this PR’s pow2 `[8192, 65535)` routed band alongside it.

Fresh p150a validation on the rebased build:
- TTNN host build: green
- focused routed-band + graph-engagement/stock-decline tests: 4 passed
- full `tests/ttnn/unit_tests/operations/reduce/test_topk.py`: 267 passed, 8 skipped, 80 expected xfails, 0 failed
- pre-commit/diff checks: green

Push-triggered CI is now running.

PR #53810, opened 2026-08-20
[Bug fix] log_sigmoid: one identity, with the residual sized to the format

### LINKED ISSUE #53787: [Bounty $5k] ttnn.log_sigmoid: 0.91% peak fp32 error and a missing x <= -4 branch
## Summary

`ttnn.log_sigmoid` has three accuracy defects in its current SFPU kernel. A source-level
model over two million fp32 inputs on `[-30, 30]` gives:

| implementation | max relative error | max fp32 ULP vs PyTorch |
|---|---:|---:|
| current kernel | 9.13e-3 | 131,755 |
| proposed fix, fused MAD model | 2.15e-7 | 2.0 |
| proposed fix, non-fused MAD model | 2.34e-7 | 2.0 |

ULP is `|actual - expected| / ULP(expected)`, the definition used by
`assert_with_ulp`. The expected output is the registered golden,
`torch.nn.functional.logsigmoid`
([`unary.py#L63`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/ttnn/operations/unary.py#L63)).
The binding advertises `FLOAT32` without a range caveat
([`unary_nanobind.cpp#L1863-L1868`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/unary_nanobind.cpp#L1863-L1868)).

This is primarily an fp32 defect. Modelled bf16 error is 2.1 ULP before and 1.3 ULP
after the fix, so a bf16-only sweep has little power to distinguish the two.

### 1. Inputs at and below `x = -4` return the raw input

The current kernel initializes `result = x`, negates its local copy of `x`, and then has
only two predicates
([`ckernel_sfpu_logsigmoid.h#L32-L54`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_logsigmoid.h#L32-L54)):

```cpp
sfpi::vFloat result = x;
x = -x;
v_if(x < -4.0f)                  { result = -exp_neg_x; }
v_elseif(x >= -4.0f && x < 4.0f) { result = -poly(x); }
v_endif;
```

For the original input `x <= -4`, neither predicate is true. The output stays equal to
the input and drops the `log1p(exp(x))` residual. At `x = -4`, the absolute error is
1.81e-2 and the relative error is 0.452%.

### 2. The positive tail uses a one-term series and approximate `exp`

For `x > 4`, the kernel returns `-exp(-x)`. This replaces `log1p(e)` with `e` and drops
the `-e^2/2` term, producing 0.91% relative error just above `x = 4`. The compute kernel
also calls `exp_tile<true>(1)`
([`logsigmoid_kernel.cpp#L40`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/device/kernels/compute/logsigmoid_kernel.cpp#L40)),
so approximate-exp error is added to the truncation error.

### 3. The mid-range polynomial is inaccurate

The degree-8 polynomial misses `ln(2)` by 7.12e-4 at `x = 0`, the centre of its fit
range. On the common residual domain `[0, 4]`, the degree-8 polynomial in the adjacent
`softplus` kernel has 48 times lower maximum absolute error.

Representative values from the source-level model:

| x | current | exact | relative error | regime |
|---:|---:|---:|---:|---|
| -4.0 | -4.000000000 | -4.018149928 | 0.452% | missing branch |
| -1.0 | -1.306191057 | -1.313261688 | 0.538% | polynomial |
| 0.0 | -0.692435443 | -0.693147181 | 0.103% | polynomial |
| 1.0 | -0.315017842 | -0.313261688 | 0.561% | polynomial |
| 4.0001 | -0.018313805 | -0.018148127 | 0.913% | one-term tail |

The Wormhole and Blackhole headers differ in types and loop annotations, but their
arithmetic is identical.

## Why tests miss it

The main sweep draws from `torch_random(low=-4, high=10)`
([`log_sigmoid.py#L58`](https://github.com/tenstorrent/tt-metal/blob/main/tests/sweep_framework/sweeps/eltwise/unary/log_sigmoid/log_sigmoid.py#L58)).
It cannot sample below the missing branch boundary, and it covers only `bfloat16` and
`bfloat8_b`. The same lower bound appears in four additional test entries.

The existing LLK comparison uses `atol=0.05, rtol=0.05`. For positive inputs the output
is a small residual, so the absolute tolerance can accept an incorrect result outright.

## Proposed fix

Use one stable identity with no `+/-4` range split:

```
logsigmoid(x) = min(x, 0) - log1p(exp(-|x|))
```

The exponential input is never positive, so the intermediate cannot overflow. In the
actual FTZ implementation its result is in `[0, 1]`. Both required primitives already
exist: the accurate path in `ckernel_sfpu_exp.h` and `calculate_log1p_fp32` in
[`ckernel_sfpu_log1p.h#L52`](https://github.com/tenstorrent/tt-metal/blob/main/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_log1p.h#L52).

The proposed patch:

- computes `exp(-|x|)` instead of `exp(-x)`;
- switches `exp` to accurate mode;
- replaces the range split with `calculate_log1p_fp32` plus the `x < 0` linear term;
- applies the change to Wormhole and Blackhole;
- widens all five `low=-4` test sites to `[-30, 30]` and adds fp32 coverage;
- adds end-to-end fp32 and bf16 ULP tests.

## Validation

Completed locally:

- LLK path compiles with `-Werror` on Wormhole and Blackhole;
- the actual `logsigmoid_kernel.cpp` compute translation unit passes a fake-JIT device-code
  compile check on Wormhole and Blackhole;
- patched Wormhole kernel passes all four bf16-input `ttsim` configurations;
- with identical stimuli and an error-sensitive tolerance, the current kernel fails the
  two fp32-DEST configurations and the patched kernel passes them;
- source models replay `_sfpu_exp_fp32_accurate_` and `calculate_log1p_fp32` in float32,
  including both fused and non-fused MAD evaluation;
- the two C++ callers of `calculate_logsigmoid` are updated.

Limitations:

- `ttsim` v1.10.1 aborts on Float32 input with fp32 DEST. The same failure occurs in its
  Float32 datacopy control, so this is not specific to `log_sigmoid`. The fp32 result needs
  device CI.
- The full tt-metal runtime JIT generation and link were not run locally. The compute
  translation unit itself was compiled as described above.
- No performance claim is made. Static Wormhole counts increase from 143 to 168 SFPU
  instructions for fp32 DEST and decrease from 143 to 70 for bf16 DEST; device timing is
  required.
- In `ttsim`, both the current and patched paths return zero once the exponential becomes
  subnormal. This establishes no simulated regression; silicon behaviour still needs CI.

The `ttsim` test plan below contains the commands, expected results and the full set of
limitations.

## `ttsim` test plan

<details>
<summary>Reproducible <code>ttsim</code> test plan: setup, commands, results, expected results after the fix, and limitations</summary>

Run on 2026-08-19 and 2026-08-20 on an Apple Silicon Mac in a Debian arm64 container.
Everything marked *result* below was executed. What ttsim cannot run is marked as such,
with the reason.

---

#### Setup

Host: Apple Silicon (arm64) macOS, Docker/colima. The toolchain and `ttsim` were both
aarch64; no x86_64 emulation was used.

```bash
# 1. Toolchain (hash-pinned by tt-metal itself, tt_metal/tt-llk/tests/sfpi-version)
cd $TT_METAL_HOME/tt_metal/tt-llk/tests && ./setup_testing_env.sh

# 2. Simulator + SoC descriptor
mkdir -p ~/sim && cd ~/sim
wget https://github.com/tenstorrent/ttsim/releases/download/v1.10.1/libttsim_wh_aarch64.so
cp $TT_METAL_HOME/tt_metal/soc_descriptors/wormhole_b0_80_arch.yaml soc_descriptor.yaml

# 3. Python env. Drop tt-smi from requirements.txt: its pyluwen dependency needs a Rust
#    toolchain and nothing in the simulator flow uses it.
python3 -m venv .venv && .venv/bin/pip install -r $TT_METAL_HOME/tt_metal/tt-llk/tests/requirements.txt

# 4. Run the four bf16-input configurations
export TT_METAL_SIMULATOR=~/sim/libttsim_wh_aarch64.so
export TT_METAL_DISABLE_SFPLOADMACRO=1
export CHIP_ARCH=wormhole
cd $TT_METAL_HOME/tt_metal/tt-llk/tests/python_tests

ID=test_sfpu_binary.py::test_sfpu_binary_logsigmoid
DESELECT=""
for out in Float16_b Float32; do for dest in No Yes; do
  DESELECT="$DESELECT --deselect $ID[formats:Float32->$out-mathop:SfpuLogsigmoid-dest_acc:$dest]"
done; done
pytest --run-simulator -q test_sfpu_binary.py -k logsigmoid $DESELECT
# -> 4 passed, 1874 deselected
```

The four fp32-input configurations have to be deselected by node id, not with
`-k "not Float32"`: that expression also drops the two `Float16_b->Float32` cases, which
are bf16-input and do run. Two of the fp32-input ones abort the simulator (see Limitations)
and an abort ends the session.

System packages the toolchain needs beyond a bare `debian:bookworm`: `libmpc3` (`cc1plus`)
and `libatomic1` (`tt_umd`).

#### Checks and results

| # | check | result |
|---|---|---|
| 1 | Kernel compiles, Wormhole + Blackhole, `-Werror` | **pass**, no warnings |
| 1a | Actual `logsigmoid_kernel.cpp` translation unit, fake-JIT device compile, Wormhole + Blackhole | **pass**; only the pinned SFPI headers' existing warnings |
| 2 | Canonical smoke test (`test_eltwise_unary_datacopy`, Float16_b, 4 faces) | **pass** |
| 3 | `test_sfpu_binary_logsigmoid`, 4 bf16-input configs, patched kernel | **4 passed** |
| 4 | Same 4 configs, **upstream** kernel `d7f34157`, shared tolerance | **4 passed** — the tolerance cannot see the defect |
| 5 | Same 4 configs, uniform `atol=0, rtol=1e-2` (one to two bf16 ULP: one ULP of that format is 0.39–0.78% relative depending on position in the binade) | patched **4 pass**; upstream **2 fail, 2 pass** — the two failures are the fp32-DEST configs |
| 6 | Threshold sweep, patched kernel, bf16 in / fp32 DEST | passes down to `rtol=5e-4` |
| 7 | Threshold sweep, upstream kernel, same config | fails at `5e-3` and `1e-2`, passes only at `2e-2` |
| 8 | Same sweep with **bf16 DEST** (`dest_acc:No`) | patched needs `>= 1e-2` — **not separable** from upstream, and left on the shared tolerance rather than papered over |
| 9 | The tolerances proposed for the fix PR, both kernels | patched **4/4 pass**; upstream **2 fail** (the two fp32-DEST configs), 2 pass (the bf16-DEST ones, per row 8) |

Row 9 is the one that matters for review: the PR does not merely add a test that passes, it
adds one that fails on the kernel being replaced, on two of the four simulated
configurations. Rows 6-8 are why the thresholds are where they are: 5e-4 vs 1e-2 leaves an
order of magnitude of margin on each side, and where no such margin exists the test says so
instead of inventing one.

Rows 4 and 5 are the before/after. Row 4 shows why the defect survived: the LLK harness
compares at `atol=0.05, rtol=0.05`, five times looser than the error being looked for. Row
5 holds the stimuli, the simulator and the criterion fixed and varies only the kernel.
Neither row separates the two kernels in bf16 DEST. Nothing does, per row 8.

#### Expected results after the fix

- every `logsigmoid` config on bf16 input: pass;
- the two fp32-DEST configs fail on the pre-fix kernel and pass on the patched one
  (`atol=0, rtol=5e-3`), which is the regression guard to re-run first;
- `Float32 -> Float32, dest_acc=Yes` (silicon CI only, see Limitations): passes at
  `atol=0, rtol=1e-5`. The upstream kernel is 9.1e-3 relative there and fails it, and so
  would the `-softplus(-x)` reformulation this fix was chosen over (6.4e-5). This is the
  check that guards fp32 accuracy, and it is in the PR;
- bf16 DEST: both kernels pass. Documented in the test, not hidden.

#### Limitations

1. **No fp32 configuration of this op can run.** ttsim aborts on Float32 input tiles when
   DEST accumulation is on: `UndefinedBehavior: tensix_unpacr: unpack_to_dst=0
   in_data_format=0 out_data_format=0`. This is not specific to this op —
   `test_eltwise_unary_datacopy[formats:Float32->Float32-dest_acc:Yes]` fails identically,
   while the same test's `dest_acc:No` configurations pass. Since fp32 input requires fp32
   DEST, and the harness skips fp32-input/`dest_acc:No` (`_skip_fp32_no_dest_acc`), neither
   fp32 configuration is available. The headline number here is an fp32 number, so **ttsim
   cannot demonstrate it**; that needs device CI.
2. **bf16 caps the resolution anyway.** The harness computes its golden at the *input*
   format, and the second operand (`exp(-|x|)`) is quantized to the tile format, so a bf16
   run cannot resolve better than ~0.4% relative whatever the kernel does.
3. **The bf16 gain is modest.** Modelled max error goes from 2.1 to 1.3 bf16 ULP
   (`|err| / ULP(exact)`, ULP = 2^(E-7) for bfloat16) — roughly one ULP of the format. The
   fp32 path is where it is five orders of magnitude off (9.1e-3 -> 2.15e-7 relative
   against the registered PyTorch golden in the fused-MAD model). This
   is an fp32 defect, which a bf16-only sweep could never have caught.
4. **`SFPLOADMACRO` must be disabled** (`TT_METAL_DISABLE_SFPLOADMACRO=1`), per TTSIM.md.
   With it disabled the SFPU path runs clean, no unimplemented-opcode error.
5. **No performance claim.** ttsim is functional only; cycle counts need silicon.
6. **The full runtime JIT was not run.** The actual compute translation unit compiles for
   both architectures, but local validation did not generate, link and execute it through
   a built tt-metal runtime.
7. The aarch64 `.so` is not hash-pinned in `tests/ttsim-version` (only the x86_64 ones
   are). Ours came from the official v1.10.1 release, the version the repo pins.

#### Note on the golden

`BinarySFPUGolden._logsigmoid` ignored its second operand and recomputed the exponential
from `x`. That made the comparison measure the operand tile's quantization rather than the
kernel. It now consumes the operand actually handed to the kernel. The operand generator
also computes `exp(-|x|)` from `x` after input-format quantization, matching the device
pipeline. End-to-end accuracy of the ttnn op stays covered by the ttnn tests
(`test_unary_fp32.py`, `test_activation.py`).

</details>

## Prior art

- #26958 investigated accurate `logsigmoid` implementations. Its best reported trade-offs
  were about 2 ULP at +43% kernel time and 4 ULP at +11%.
- #32461 (merged) added the current SFPU kernel and evaluated bf16, not fp32.
- #41390 (closed, unmerged) proposed `-softplus(-x)` and evaluated bf16 only. That is adequate for bf16 but
  models at about 1,000 fp32 ULP because the softplus residual fit targets absolute, not
  relative, accuracy.

## Acceptance criteria

- fp32 output within 4 ULP of the registered PyTorch golden on `[-30, 30]`;
- no discontinuity at `x = +/-4`;
- fp32, bf16 and both tails covered by regression tests;
- Wormhole and Blackhole CI pass;
- behaviour outside the validated fp32 range is documented.

kinginu:
### PR Category

Bug fix.

### Summary

Closes #53787.

`log_sigmoid` becomes `min(x, 0) - log1p(exp(-|x|))`, computed in one kernel.

The form it replaces split the domain at `±4` and had no arm for `x <= -4`, so an input there was returned unchanged and lost the residual entirely. Above `+4` it used a one term series, `e` rather than `e - e²/2 + e³/3`, on an exponential the caller had computed in approximate mode.

The residual is computed differently per input format, and that is the one decision the diff does not explain. float32 needs it to **relative** precision, because for a positive input the residual is the whole answer and it is tiny: the tuned polynomial `softplus` uses is accurate in absolute terms and measured 3914 ULP here at the same cost, so float32 pays for the accurate exponential and a real `log1p`. bfloat16 has eight bits of mantissa, so it takes softplus's degree-6 polynomial on `[0, 5]` and, past that, the cheap 21f exponential rather than softplus's Cody-Waite reduction, which is worth 1034 cycles.

The caller drops a `copy_tile`, a `negative_tile` and the approximate `exp_tile`, and the op goes from two DST tiles to one.

### Accuracy

`[-30, 30]`, 1024 points, against `torch.nn.functional.logsigmoid` in float64. P300 and N300 return the same numbers.

| | before | after |
|---|---|---|
| float32 max ULP | **297,549.7** | **0.7** |
| float32 max relative | 3.55e-02 | 8.32e-08 |
| bfloat16 max ULP | 4.2 | **0.8** |
| bfloat16 max relative | 3.31e-02 | 6.01e-03 |

The bfloat16 relative figure is that format's own resolution; 0.8 ULP is faithful rounding.

### Cost

Device profiler, `[1, 1, 512, 512]`, 33 invocations, median cycles per core on the math thread. One kernel before and after.

| | P300 fp32 | P300 bf16 | N300 fp32 | N300 bf16 |
|---|---|---|---|---|
| before | 6398 | 5165 | 11855 | 9403 |
| after | **8236** | **5646** | **15674** | **10372** |
| | +28.7% | +9.3% | +32.2% | +10.3% |

float32 pays for the accurate exponential and `log1p`, which is what buys 297,550 ULP to 0.7. bfloat16 pays 9 to 10 percent for 4.2 ULP to 0.8.

### Tests

`test_log_sigmoid_accuracy.py`, 6 cases: the named points either side of the old `±4` split, a `[-30, 30]` sweep, and `log_sigmoid(-4)` asserted not to equal its input, all gated in ULP.

Two existing ranges are widened from `[-4, 10]` to `[-30, 30]`, in `sweeps/eltwise/unary/log_sigmoid/log_sigmoid.py` and the eager `test_run_eltwise_log_sigmoid_op`. The lower bound was exactly the boundary of the missing arm, so no draw could reach it.

### Not covered

- `bfloat8_b` and `bfloat4_b`.
- The backward op, whose own test draws `[-120, -1]` and is untouched here.
- Whether the float32 cost can come down. The polynomial route was measured at the same cost and 3914 ULP, so it is not the answer; a cheaper accurate exponential would be.

### DIFF
--- a/tests/sweep_framework/sweeps/eltwise/unary/log_sigmoid/log_sigmoid.py
+++ b/tests/sweep_framework/sweeps/eltwise/unary/log_sigmoid/log_sigmoid.py
@@ -55,7 +55,9 @@ def run(
     torch.manual_seed(0)
 
     torch_input_tensor = gen_func_with_cast_tt(
-        partial(torch_random, low=-4, high=10, dtype=torch.float32), input_dtype
+        # -4 was the boundary of a range split the kernel no longer has, and it
+        # was also this sweep's lower bound, so the arm below it was never drawn.
+        partial(torch_random, low=-30, high=30, dtype=torch.float32), input_dtype
     )(input_shape)
     golden_function = ttnn.get_golden_function(ttnn.log_sigmoid)
     torch_output_tensor = golden_function(torch_input_tensor)
--- a/tests/tt_eager/python_api_testing/sweep_tests/pytests/tt_dnn/test_eltwise_unary.py
+++ b/tests/tt_eager/python_api_testing/sweep_tests/pytests/tt_dnn/test_eltwise_unary.py
@@ -904,7 +904,7 @@ class TestEltwiseUnary:
         output_mem_config,
     ):
         datagen_func = [
-            generation_funcs.gen_func_with_cast(partial(generation_funcs.gen_rand, low=-4, high=10), torch.float32)
+            generation_funcs.gen_func_with_cast(partial(generation_funcs.gen_rand, low=-30, high=30), torch.float32)
         ]
         comparison_func = partial(comparison_funcs.comp_pcc)
         test_args = generation_funcs.gen_default_dtype_layout_device(input_shapes)[0]
--- a/tests/ttnn/unit_tests/operations/eltwise/test_log_sigmoid_accuracy.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_log_sigmoid_accuracy.py
@@ -1,0 +1,67 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""log_sigmoid across the whole real line, gated on ULP.
+
+The form this replaced split the domain at +-4 with no arm for x <= -4, so an
+input there came back unchanged. The sweeps could not see it: they draw from
+[-4, 10], which starts exactly at the missing boundary.
+"""
+
+import pytest
+import torch
+import ttnn
+
+# The boundaries of the range split that is gone, and points either side of it.
+NAMED = [-30.0, -10.0, -4.0001, -4.0, -3.9999, -1.0, 0.0, 1.0, 3.9999, 4.0, 4.0001, 10.0, 30.0]
+
+
+def _ulp(got, want, torch_dtype):
+    eps = torch.finfo(torch_dtype).eps
+    return (got.double() - want).abs() / want.abs().clamp(min=1e-30) / eps
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_log_sigmoid_named_points(device, dtype):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+
+    x = torch.tensor(NAMED + [0.0] * (1024 - len(NAMED)), dtype=torch.float32).reshape(1, 1, 32, 32).to(torch_dtype)
+    ix = ttnn.from_torch(x, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    got = ttnn.to_torch(ttnn.log_sigmoid(ix)).flatten()[: len(NAMED)]
+    want = torch.nn.functional.logsigmoid(torch.tensor(NAMED, dtype=torch.float64))
+
+    ulp = _ulp(got, want, torch_dtype)
+    assert ulp.max() <= 2.0, f"max {ulp.max():.1f} ULP at x={NAMED[int(ulp.argmax())]}"
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_log_sigmoid_sweep(device, dtype):
+    """[-30, 30], which the existing sweeps cannot reach below -4."""
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    torch.manual_seed(0)
+
+    x = ((torch.rand((1, 1, 32, 32), dtype=torch.float32) * 60) - 30).to(torch_dtype)
+    ix = ttnn.from_torch(x, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    got = ttnn.to_torch(ttnn.log_sigmoid(ix))
+    want = torch.nn.functional.logsigmoid(x.double())
+
+    ulp = _ulp(got, want, torch_dtype)
+    assert ulp.max() <= 2.0, f"max {ulp.max():.1f} ULP"
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_log_sigmoid_below_minus_four_is_not_the_input(device, dtype):
+    """The arm that was missing: x <= -4 used to return x unchanged."""
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+
+    x = torch.full((1, 1, 32, 32), -4.0, dtype=torch_dtype)
+    ix = ttnn.from_torch(x, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    got = ttnn.to_torch(ttnn.log_sigmoid(ix)).flatten()[0]
+
+    assert got != x.flatten()[0], "log_sigmoid(-4) returned its input"
+    want = torch.nn.functional.logsigmoid(torch.tensor(-4.0, dtype=torch.float64))
+    # Bounded in ulp, not absolutely: one bfloat16 ulp at this magnitude is
+    # 0.03125, so an absolute bound tighter than that fails a correctly
+    # rounded answer.
+    assert _ulp(got.reshape(1), want.reshape(1), torch_dtype).max() <= 2.0, f"got {got}, want {want}"
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_logsigmoid.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_logsigmoid.h
@@ -8,6 +8,7 @@
 #include "ckernel_defs.h"
 #include "ckernel_sfpu_exp.h"
 #include "ckernel_sfpu_log1p.h"
+#include "ckernel_sfpu_softplus.h"
 #include "sfpi.h"
 
 namespace ckernel {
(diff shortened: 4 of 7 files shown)


kanapitsas:
Hi @kinginu — I was assigned #53787 and am opening my PR now. I credited #53810 for the
one-tile unary structure and the input-format residual split.

My branch adds a domain-specific Float32 residual, two-datum lockstep, the bfloat16
residual and NaN guard, the DEST-format gate, and expanded LLK/TTNN validation.

ncvetkovicTT:
@kinginu I will close as this issue is assigned to @kanapitsas. FYI @jberkowitzTT.

PR #21306, opened 2025-04-28
Add support of Mistral-7B into TT-Transformers

pprajapatiTT:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/19490

### Problem description
Add support of Mistral-7B into TT-Transformers and deprecates the old stale (lower-perf) implementation.

### What's changed
Remove deprecated implementation of Mistral-7B (in `/models/demos/wormhole/mistral-7b/)
Add Mistral-7B support to TT-Transformers
Validated Mistral-7B-Instruct-v0.3 using TT-Transformers via HuggingFace interface
Add performance and accuracy numbers
Add CI tests
Fix the tests for HuggingFace based models

Checklist

### Checklist
- [ ] [All post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml) CI passes
- [ ] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml) CI with demo tests passes (if applicable)
- [ ] [Model regression](https://github.com/tenstorrent/tt-metal/actions/workflows/perf-models.yaml) CI passes (if applicable)
- [ ] [Device performance regression](https://github.com/tenstorrent/tt-metal/actions/workflows/perf-device-models.yaml) CI passes (if applicable)
- [ ] **(For models and ops writers)** Full [new models tests](https://github.com/tenstorrent/tt-metal/actions/workflows/full-new-models-suite.yaml) CI passes (if applicable)
- [ ] New/Existing tests provide coverage for changes

PR #53816, opened 2026-08-20
[Bug fix] Correct Galaxy Llama decode state validation

nkapreTT:
## Summary

This change corrects the decode-state validation in two model tests. It also registers those tests in the tiered Models CI. No pipeline ran them before.

- `models/demos/llama3_70b_galaxy/tests/test_llama_model.py`
- `models/tt_transformers/tests/test_model.py`

## Problem

The decode loop set the position to `start + i` at the end of step `i`. The next step needs `start + i + 1`. Step 0 and step 1 therefore ran token 0 at position 0. Each later step was one position late. The prompt token index had the same fault.

Two more faults followed:

- The old code compared a fixed-width TT cache window with an append-only `DynamicCache`. The shapes did not agree, and `comp_pcc` stopped with an error.
- The Galaxy test did not change the sampled device token to an integer index.

## Changes

Decode state:

- Set the position to `start + i + 1` after each step.
- Set the teacher-forced prompt token to index `i + 1`.
- Compare only the cache entries that the models wrote.
- Change the sampled Galaxy device token to an integer index.
- Move the step arithmetic to `models/tt_transformers/tests/decode_test_helpers.py`. A unit test covers it.

Threshold:

- Set the Blackhole Galaxy threshold for Llama-3.2-1B to `0.989` in performance mode. The correct second step gives a PCC of `0.989794`. The old value of `0.991` agreed with the duplicated first token. The other thresholds do not change.

CI registration:

- Enable the three `test_model.py` commands in `tests/pipeline_reorg/models_unit_tests.yaml`. They were comments before.
- Add `test_llama_model.py -k quick` to the Galaxy entry. Only the `quick` variant sets `cache_pcc`.
- Increase the step time limits.
- Increase three shared budgets in `.github/time_budget.yaml`. The verifier adds the time limits of all entries for each team, tier and SKU. The three pools had no time available:
  - `models.unit_tier1.wh_n150`: 110 to 118
  - `models.unit_tier1.wh_galaxy_perf`: 60 to 65
  - `models.unit_tier2.wh_llmbox_perf`: 172 to 191

Corrections from review:

- Read `cache_k[l]` and `cache_v[l]`. `HfModelWrapper.cache_k` gives a list, so the old code stopped with an `AttributeError`.
- Record the device token in the Galaxy test. The test recorded the reference token twice before.
- Set `kv_cache_tests_pass` to `False` when a cache comparison fails. The flag was always `True`, so its assert could not fail.

The `cache_pcc` flag stays off. Issue #55064 gives the details.

## Validation

Run [33514134375](https://github.com/tenstorrent/tt-metal/actions/runs/33514134375) on `6754f95d913`. All five legs passed.

| Test | SKU | Result |
| --- | --- | --- |
| Llama 3.3-70B Galaxy | `wh_galaxy_perf` | K/V cache PCC 0.99988 / 0.99987, threshold 0.9997 |
| Llama 3.1-8B | `wh_n150` | pass |
| Llama 3.1-8B | `wh_llmbox_perf` | pass |
| Mistral-7B | `wh_n150` | pass |
| Llama 3.2-90B-Vision | `wh_llmbox_perf` | PCC 0.9872 to 0.9888, threshold 0.971 |

The Galaxy KV cache comparison ran for the first time. The old code stopped with an error before it reached the comparison.

## Related

- #55064: the `cache_pcc` flag in `test_model.py` has never run.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:codex/llama-galaxy-decode-state-20260820)

### DIFF
--- a/.github/time_budget.yaml
+++ b/.github/time_budget.yaml
@@ -359,9 +359,9 @@ models:
     bh_quietbox_2: 42
   unit_tier1:
     wh_llmbox: 150
-    wh_n150: 110
+    wh_n150: 118
     wh_galaxy: 260
-    wh_galaxy_perf: 60
+    wh_galaxy_perf: 65
     bh_p150: 20
     bh_quietbox_2: 390
     bh_galaxy: 75
@@ -369,7 +369,7 @@ models:
     bh_sc4: 50
   unit_tier2:
     wh_llmbox: 114
-    wh_llmbox_perf: 172
+    wh_llmbox_perf: 191
     wh_n150: 82
     wh_n300: 30
     wh_galaxy_perf: 80
--- a/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
+++ b/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
@@ -494,6 +494,7 @@ def test_llama_model_inference(
                                 logger.info(f"KV Cache Passed!")
                             else:
                                 logger.warning(f"KV Cache Failed! PCC value is lower than {pcc}")
+                                kv_cache_tests_pass = False
                                 all_tests_pass = False
 
             if not dummy_weights:
--- a/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
+++ b/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
@@ -394,12 +394,13 @@ def test_llama_model_inference(
                     reference_token=reference_token,
                     device_token=tt_out_tok,
                 )
-                next_token_id = next_token.squeeze(1).tolist()[0]
-                all_outputs.append(next_token_id)
+                # Feed one shared trajectory, but keep logging the token the device
+                # actually sampled so the two generations stay comparable.
+                all_outputs.append(tt_out_tok.squeeze(1).tolist()[0])
                 tt_decode_input = embd(next_token)
 
                 if run_ref_pt:
-                    all_outputs_ref.append(next_token_id)
+                    all_outputs_ref.append(next_token.squeeze(1).tolist()[0])
                     pt_decode_input = tt_decode_input
             # Measure PCC if also running reference model
             if run_ref_pt:
--- a/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
+++ b/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
@@ -12,7 +12,7 @@ from models.demos.llama3_70b_galaxy.tt.llama_common import (
 )
 from models.demos.llama3_70b_galaxy.tt.model_config import TtModelArgs, LlamaOptimizations
 from models.demos.llama3_70b_galaxy.tt.llama_model import TtTransformer
-from models.tt_transformers.tests.decode_test_helpers import decode_step_state
+from models.tt_transformers.tests.decode_test_helpers import decode_step_state, teacher_forced_decode_token
 from models.tt_transformers.tests.test_utils import get_ref_model_dype
 from models.common.sampling.tt_sampling import TTSampling
 from models.common.utility_functions import (
@@ -301,6 +301,10 @@ def test_llama_model_inference(
 
     try:
         for i in range(generation_length):
+            # Validate the absolute position before executing the model step.
+            next_position, next_token_index, num_written = decode_step_state(
+                generation_start_pos, i, len(encoded_prompts[0]), model_args.max_seq_len
+            )
             logger.info(f"[Llama3 Model] Generating token {i}")
             decode_input = model_args.prepare_residual_tensor_decode(
                 tt_decode_input,
@@ -340,9 +344,6 @@ def test_llama_model_inference(
                 ref_output = reference_model(pt_decode_input.to(ref_dtype), current_pos[0])
 
             # Increment position
-            next_position, next_token_index, num_written = decode_step_state(
-                generation_start_pos, i, len(encoded_prompts[0]), model_args.max_seq_len
-            )
             current_pos = torch.full((batch,), next_position)
 
             current_pos_sram = torch.full((model_args.sub_core_grids.num_cores(), batch), next_position)
@@ -382,15 +383,24 @@ def test_llama_model_inference(
                     .long()
                 )
 
-                all_outputs.append(tt_out_tok.squeeze(1).tolist()[0])  # Update generated token to list of TT outputs
-                tt_decode_input = embd(tt_out_tok)
+                reference_token = None
+                if run_ref_pt:
+                    reference_token = sample_host(ref_output, None, temperature=0, top_p=0.8)
+
+                # PCC and cache comparisons require both models to follow one
+                # trajectory. Prefer the reference token when it is available;
+                # otherwise continue standalone generation from the TT sample.
+                next_token = teacher_forced_decode_token(
+                    reference_token=reference_token,
+                    device_token=tt_out_tok,
+                )
+                next_token_id = next_token.squeeze(1).tolist()[0]
+                all_outputs.append(next_token_id)
+                tt_decode_input = embd(next_token)
 
                 if run_ref_pt:
-                    pt_out_tok = sample_host(ref_output, None, temperature=0, top_p=0.8)
-                    pt_decode_input = embd(pt_out_tok)
-                    all_outputs_ref.append(
-                        pt_out_tok.squeeze(1).tolist()[0]
-                    )  # Update generated token to list of ref outputs
+                    all_outputs_ref.append(next_token_id)
+                    pt_decode_input = tt_decode_input
             # Measure PCC if also running reference model
             if run_ref_pt:
                 if layers == 1 and i == iterations - 1:  # On last iteration in the quick test, set a tighter PCC
--- a/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
+++ b/models/demos/llama3_70b_galaxy/tests/test_llama_model.py
@@ -12,6 +12,7 @@ from models.demos.llama3_70b_galaxy.tt.llama_common import (
 )
 from models.demos.llama3_70b_galaxy.tt.model_config import TtModelArgs, LlamaOptimizations
 from models.demos.llama3_70b_galaxy.tt.llama_model import TtTransformer
+from models.tt_transformers.tests.decode_test_helpers import decode_step_state
 from models.tt_transformers.tests.test_utils import get_ref_model_dype
 from models.common.sampling.tt_sampling import TTSampling
 from models.common.utility_functions import (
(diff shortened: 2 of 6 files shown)


yieldthought:
Codex review:

[P1] Keep the TT and reference models on one decode trajectory

[models/demos/llama3_70b_galaxy/tests/test_llama_model.py:364](https://github.com/tenstorrent/tt-metal/blob/4126487fc3ab7d241aab0cb2be3eaa14683f8e3d/models/demos/llama3_70b_galaxy/tests/test_llama_model.py#L364)

Using `next_token_index` makes the six-iteration quick test enter this branch after executing prompt token 4, so iteration 5 now runs on sampled inputs. The TT side embeds `tt_out_tok`, while the reference side independently samples and embeds `pt_out_tok` below. If the approximate TT logits and reference logits choose different argmax tokens, the next output and KV-cache PCC checks compare different input sequences and can fail even when each model is correct; the previous off-by-one loop never reached that divergent iteration in the quick case. Suggestion: teacher-force both sides with the same sampled token (or explicitly require the two sampled tokens to match before continuing) so every PCC comparison has a common input trajectory.

[P2] Reject positions that exceed the configured sequence length

[models/tt_transformers/tests/decode_test_helpers.py:12](https://github.com/tenstorrent/tt-metal/blob/4126487fc3ab7d241aab0cb2be3eaa14683f8e3d/models/tt_transformers/tests/decode_test_helpers.py#L12)

This cap freezes `num_written` while `next_position` keeps advancing beyond `max_seq_len`. The added `start_pos=254` test demonstrates the problem: iteration 2 is accepted as `(257, 3, 2)`, even though that iteration has already executed absolute position 256 outside a 256-entry sequence, and subsequent cache checks only recheck the first two entries. That silently masks an invalid test configuration instead of enforcing the model/cache boundary. Suggestion: reject `start_pos + iteration >= max_seq_len` before the model step (and change the helper test to expect that failure) rather than capping only the comparison window.

nkapreTT:
Review-ready at 9a4f836301: all review threads are resolved, with 26 successful checks and zero pending or failing checks. The pending CODEOWNER reviewers are already requested.

yieldthought (approved):

cglagovichTT (approved):

mtairum (changes requested):
Re-enabled some tests and kicked CI. 

Blocking merge until we can validate those.

mtairum:
Pushed `53a41533502` to this branch. Heads-up: it's CI registry config only, no model or test code.

**Why:** none of the tiered pipelines were running the two files this PR changes. All three `models/tt_transformers/tests/test_model.py` invocations in `tests/pipeline_reorg/models_unit_tests.yaml` were commented out (Llama 3.1-8B, Llama 3.2-90B-Vision, Mistral-7B), and `models/demos/llama3_70b_galaxy/tests/test_llama_model.py` had no registry entry at all — the Galaxy entry lists only files under `tests/unit_tests/`, and this one sits a level up. So the changes here had no tt-metal CI coverage.

**What changed:**
- Uncommented the three `test_model.py` lines. Left the `test_model_prefill.py` ones commented — out of scope for this PR.
- Registered `models/demos/llama3_70b_galaxy/tests/test_llama_model.py -k quick` on the Llama 3.3-70B Galaxy entry. Pinned to `quick` deliberately: that's the single-layer variant where `cache_pcc` is enabled, so it's the only one that exercises the corrected KV-cache comparison.
- Bumped the affected job timeouts, and raised `models.unit_tier1.wh_galaxy_perf` 45→55 and `models.unit_tier2.wh_llmbox_perf` 172→183 in `.github/time_budget.yaml`. `verify_time_budget.py` sums timeouts per (team, tier, SKU) and both pools were at 100%, so the bumps didn't fit without it — flagging for infra visibility.

**CI:** https://github.com/tenstorrent/tt-metal/actions/runs/33205807800 — All Model Tests, tiers 1+2, unit only, 5 legs:
- tier 1: Llama 3.1-8B / `wh_n150`, Llama 3.3-70B Galaxy / `wh_galaxy_perf`
- tier 2: Llama 3.1-8B / `wh_llmbox_perf`, Llama 3.2-90B-Vision / `wh_llmbox_perf`, Mistral-7B / `wh_n150`

Dispatched with `mlperf-read-only=false` — these invocations have never populated their `tt_cache` entries, so a read-only mount would just EROFS.

If this comes back green I'll approve and we can merge.

Two things to watch, so a red isn't misread:
- The Galaxy leg is the first time the `final_k_cache_pcc`/`final_v_cache_pcc` = 0.9997 gate has actually compared tensors. The old slice (`[gsp : gsp + generation_length + 1]` against an append-only `DynamicCache`) had mismatched shapes and would have raised inside `comp_pcc`.
- The 8B / Mistral / 90B-Vision `final_model_pcc` thresholds were calibrated against the same duplicated-position trajectory as the BHGLX `0.991` you re-baselined. If one of those goes red it's that same recalibration surfacing, not a new defect.

mtairum:
Two findings from review where a stated fix isn't taking effect. Both are small.

**1. The `test_model.py` KV-cache fix is unreachable, and the block it edits would crash if it weren't.**

`models/tt_transformers/tests/test_model.py:112`:
```python
cache_pcc = layers == 1 and not dummy_weights
```
The two parametrizations (`:26-27`) are `("random", 1)` → `dummy_weights=True` → `False`, and `("instruct", None)` → `layers == 1` is `False`. So `cache_pcc` is `False` in every cell and the `if cache_pcc:` block at `:450` never runs — including the new `cache_pt[:, :, 0:num_written]` / `cache_tt[:, :, generation_start_pos : generation_start_pos + num_written]` slices.

If it were enabled, `:453` would raise:
```python
reference_model.cache_k.clone().permute(0, 2, 1, 3)
# AttributeError: 'list' object has no attribute 'clone'
```
`HfModelWrapper.cache_k` (`models/tt_transformers/tt/model_config.py:4674`) returns a **list** of per-layer tensors — which is why the Galaxy copy correctly writes `reference_model.cache_k[l]`. So only the Galaxy half of this fix is actually live. Worth either indexing by `l` here to match, or dropping the `test_model.py` half and noting the block is dead.

**2. The Galaxy test no longer validates device sampling.**

`run_ref_pt` is hardcoded `True` at `models/demos/llama3_70b_galaxy/tests/test_llama_model.py:107` and never reassigned, so `teacher_forced_decode_token(...)` at `:393` always returns `reference_token`. That makes `tt_out_tok` dead on every path: it's produced by `tt_sampling` at `:326`, pulled off device with a blocking sync at `:376`, given the new `.long()` cast at `:383` — and then discarded. So this PR's "cast sampled Galaxy device tokens to integer embedding indices" never actually runs.

It also makes the output comparison vacuous: both `all_outputs` (`:398`) and `all_outputs_ref` now receive the same `next_token_id`, so `[ttnn generation User 0]` and `[Ref generation User 0]` print identical text by construction. Before this PR those were two independent token streams, and a TTSampling regression on Galaxy would have shown up there.

Teacher-forcing the *input* while still recording the device's own sample keeps both properties:
```python
next_token = teacher_forced_decode_token(reference_token=reference_token, device_token=tt_out_tok)
all_outputs.append(tt_out_tok.squeeze(1).tolist()[0])   # device sample — restores coverage
tt_decode_input = embd(next_token)                       # teacher-forced input
```

mtairum:
**CI update — all re-enabled legs are green.**

[33395809253](https://github.com/tenstorrent/tt-metal/actions/runs/33395809253) — Llama 3.2-90B-Vision / `wh_llmbox_perf`: **passed**. `test_model.py -k quick`, both `performance` and `accuracy`, `2 passed, 2 deselected in 554.85s (0:09:14)`, final PCC `0.9872`–`0.9888` against the `0.971` `Llama-3.2-90B` threshold.

That completes the set. Across [33205807800](https://github.com/tenstorrent/tt-metal/actions/runs/33205807800) and the run above, all five legs pass: Llama 3.3-70B Galaxy / `wh_galaxy_perf`, Llama 3.1-8B on `wh_n150` and `wh_llmbox_perf`, Mistral-7B / `wh_n150`, Llama 3.2-90B-Vision / `wh_llmbox_perf`.

The Galaxy leg is the one worth noting: the newly-registered `test_llama_model.py -k quick` passed in 233s, and its KV-cache comparison ran for real — 20× `KV Cache Passed`, K/V PCC `0.99987` against `0.9997`. That gate was previously unreachable, since the old slice compared a fixed-width TT window against an append-only `DynamicCache` and `comp_pcc` raised on the shape mismatch.

The earlier 90B-Vision failure was a step timeout, not a defect — my cap was ~1 min short of the work. Now sized at 28 min from the measured 18.4 min (`unit_tier2.wh_llmbox_perf` correspondingly trimmed 208 → 191).

**Correction to [my earlier comment](https://github.com/tenstorrent/tt-metal/pull/53816#issuecomment-5457322366).** I wrote that the Galaxy test "no longer validates device sampling." That overstates it. `all_outputs` vs `all_outputs_ref` was never asserted — only logged at `test_llama_model.py:499-501` — so this PR removes a diagnostic, not a gate. And TTSampling on Galaxy is independently gated by `unit_tests/test_sampling.py::test_llama_sampling_inference`, which drives it directly with PCC and KL asserts. What stands from that point is narrower: `run_ref_pt` is hardcoded `True`, so the `.long()` cast this PR adds never executes, the blocking `.cpu()` readback each sampled iteration is wasted, and the two log lines are now identical by construction.

Point 1 of that comment is unaffected and still stands.

mtairum:
Re-ran the full set on `536eba400af`, which adds the two fixes from the review above: [33510600821](https://github.com/tenstorrent/tt-metal/actions/runs/33510600821).

Same five legs — Llama 3.3-70B Galaxy / `wh_galaxy_perf`, Llama 3.1-8B on `wh_n150` and `wh_llmbox_perf`, Mistral-7B / `wh_n150`, Llama 3.2-90B-Vision / `wh_llmbox_perf`.

The Galaxy leg is the one that needs the re-run: `536eba400af` changes the sampled branch of `test_llama_model.py`, so the previous green on `4a154a9235b` no longer describes that code. The `test_model.py` change is inert by construction — the edited lines sit inside the `if cache_pcc:` block, which never executes — but those legs are included anyway for a clean full-green on the final head.

Also filed #55064 to track enabling `cache_pcc`, which has never evaluated since it was introduced in #21306.

mtairum (approved):
waiting for 2 jobs to finished but approving to unblock.

tdowdallTT (approved):

Issue #51180, opened 2026-07-27
Reduce l2 nightly reduction and moreh tests


bbradelTT:
[ops-unit-tests / ttnn nightly reduction tests [wh_n150_civ2]](https://github.com/tenstorrent/tt-metal/actions/runs/30115438874/job/89559119530#logs)
takes too long.

Currently using n300 to verify, but it'd be good to get this green.

Same thing for moreh:
https://github.com/tenstorrent/tt-metal/actions/runs/32338190649/job/96408669370
```
Error: The action 'ttnn nightly moreh tests [wh_n150_civ2]' has timed out after 45 minutes.
```

fplavecTT:
Claude claims that the cause is the real-time device profiler, which is active only on the N150 runner (adds ~0.16 s per test, and across ~5300 tests that's ~14 minutes). According to Claude, none of the tests use data from this profiler, but we still pay the runtime penalty.
Note that this RT profiler is NOT the same as the device program profiler controlled by the `TT_METAL_DEVICE_PROFILER=1` env var.

It seems that there is currently no env var or any other way to disable the RT profiler; it's on by default, but gets turned off automatically on devices that don't meet "eligibility criteria" (`evaluate_realtime_profiler_eligibility()` in `tt_metal/distributed/realtime_profiler_manager.cpp` ), which is why it runs only on N150.
An env var to enable/disable it used to exist, but was removed in https://github.com/tenstorrent/tt-metal/pull/44722
Perhaps we need to bring that option back and disable it in test suites that don't use it?
CC @mo-tenstorrent

mo-tenstorrent:
I don't believe claude is correct as realtime profiler is active on both N300 and N150 and blackholes. 

RT profiler has also been proven to not add significant overhead in many many runs in CI. So we need much stronger evidence that RT profiler is the culprit, especially considering how it already got what runs it wrong.

fplavecTT:
Thanks @mo-tenstorrent 
I've asked Claude to run some experiments and they all point to the profiler being the cause:

### Running the tests on n150 profiler on versus off

Run [30409297198](https://github.com/tenstorrent/tt-metal/actions/runs/30409297198) — same build, same run, both n150 jobs differing only in whether the real-time profiler is enabled:

| Job | RT profiler | Test step | Result |
| --- | --- | --- | --- |
[...]

PR #53828, opened 2026-08-20
[skip ci] [Test Only] #51180: reduce scalar and dim parameters of a nightly reduction test

bbradelTT:
### PR Category

Test Only

### Summary

Relates to #51180 

Cross product of test cases for test_generic_ops_w_scalar was very large. Scalar values are redundant and all dims are tested in other tests and can be safely cut down. Therefore reduced test cases to help with test runtime in CI.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:bbradel-51180_night_reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:bbradel-51180_night_reduce)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_memory_configs.py
+++ b/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_memory_configs.py
@@ -267,7 +267,8 @@ def test_generic_ops_wh_block_shard(
     assert passing, f"op={op} {output_pcc}, torch: {torch_output_tensor}, ttnn: {output_tensor}"
 
 
-# Test that generic reduction ops work correctly with a scalar applied to the input.
+# Test that generic reduction ops produce correct results, preserve dtype, and emit the
+# layout documented in nanobind across all supported dtype/layout combinations.
 @pytest.mark.parametrize("op", ["sum", "mean", "max", "min", "std", "var"])
 # bfloat8_b only exists in TILE layout, so dtype and layout are paired.
 @pytest.mark.parametrize(
--- a/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_w_scalar.py
+++ b/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_w_scalar.py
@@ -21,7 +21,7 @@ pytestmark = pytest.mark.use_module_device
 
 
 SHAPES = [(3, 4), (1, 1, 3, 4, 5), (3, 4, 8, 56, 33)]
-DIMS = [-1, -2, 0, (-2, -1), (0, -2, -1), None]
+DIMS = [-1, 0, (0, -2, -1), None]
 
 # Pair (shape, dim) to drop combos with more reduction dims than rank, and
 # (op, correction) since correction is live only for std/var.
@@ -39,7 +39,7 @@ OP_CORRECTION = [
 
 
 @pytest.mark.parametrize("op, correction", OP_CORRECTION)
-@pytest.mark.parametrize("scalar", [1.0, -2.0, 2.0, -2.43, 2.43, 4.0])
+@pytest.mark.parametrize("scalar", [1.0, -2.43, 4.0])
 @pytest.mark.parametrize("shape, dim", VALID_SHAPE_DIMS)
 @pytest.mark.parametrize("dtype", [torch.bfloat16, torch.float32])
 def test_generic_ops_w_scalar(device, op, scalar, correction, dim, shape, dtype):
--- a/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_w_scalar.py
+++ b/tests/ttnn/nightly/unit_tests/operations/reduction/test_generic_ops_w_scalar.py
@@ -53,7 +53,7 @@ OP_CORRECTION = [
 ]
 
 
-def run_generic_ops_w_scalar(device, op, scalar, correction, dim, shape, dtype):
+def _run_generic_ops_w_scalar(device, op, scalar, correction, dim, shape, dtype):
     if op in ("var", "std") and correction:
         # Bessel's correction divides by (N - 1) where N is the number of elements
         # reduced. With N == 1 this is a divide-by-zero, producing all-NaN output;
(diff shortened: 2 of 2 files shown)


tests/ttnn/nightly/unit_tests/operations/reduction/test_reduction_ops.py
```diff
@@ -554,9 +554,9 @@ def round_up_to_tile(dim_size):
 
 # Test that generic reduction ops work correctly with a scalar applied to the input.
 @pytest.mark.parametrize("op", ["sum", "mean", "max", "min", "std", "var"])
-@pytest.mark.parametrize("scalar", [1.0, -2.0, 2.0, -2.43, 2.43, 4.0])
```
fplavecTT: We lost coverage for non power of 2 positive value. Perhaps keep 2.43 and drop 4.0?
But of course, then we no longer have any power of 2 values other than trivial 1.0.
Admittedly, powers of two are less likely to break.

tests/ttnn/nightly/unit_tests/operations/reduction/test_reduction_ops.py
```diff
@@ -554,9 +554,9 @@ def round_up_to_tile(dim_size):
...
+@pytest.mark.parametrize("scalar", [1.0, -2.43, 4.0])
 @pytest.mark.parametrize("correction", [True, False])
-@pytest.mark.parametrize("dim", [-1, -2, 0, (-2, -1), (0, -2, -1), None])
```
fplavecTT: I'm pretty sure I added these because they weren't covered elsewhere.
Detailed Claude analysis below, with TLDR version: restore the two shapes that were removed, but skip them for some ops/ranks
```
    # Reducing the last dim, the second-to-last dim, and the (second-to-last, last)
    # pair each dispatch to a distinct device kernel, so scaling has to be checked on
    # all three. mean and var already have floating-point scalar coverage on the
    # latter two kernels (test_reduction_mean.py and test_reduction.py), and std
    # shares the same kernels as var, so those two dims are exercised here only for
    # the ops that have no such coverage elsewhere.
    if dim in (-2, (-2, -1)):
        if op in ("mean", "var"):
            pytest.skip("Scaled H and (H, W) reductions for this op are covered by other tests")
        if rank == 2:
            # At rank 2 these normalize to the same reductions as dim=0 and
            # dim=None, which are already parametrized.
            pytest.skip("Duplicate of another parametrized dim at this rank")
```

## Claude analysis

The dim values are not interchangeable here. They select different device kernels and program factories:
- `generic_reductions.cpp:181-182` treats only three shapes of dim as a single-shot reduction: last dim (W), second-to-last (H), and exactly the pair (H, W). Everything else decomposes into a transpose-and-loop.
- `generic_reductions.cpp:264-273` maps those to ReduceOpDim::W / H / HW, which are three separate program factories and three separate compute kernels.
- For var/std the Welford path does the same at `generic_reductions.cpp:395-435`: single_h, single_w, and a "2+ dims" HW path.

The two removed dims were exactly the H and HW entry points.
The four kept dims cover: -1 → W; 0 → the transpose-and-loop path; (0,-2,-1) → multi-axis (Welford HW but always with a non-identity permute and reduce_batch_size > 1); None → normalized to all dims by [generate_reduce_dim](vscode-webview://082tq4q6k3bmgak39k85t8vdtvug5rrm7thg6hsu5k5d0u0671n7/ttnn/cpp/ttnn/operations/reduction/reduction_common/reduction_common.cpp#L45-L52), so on the rank-5 shapes it also goes multi-axis, not HW.

The one place H and HW survive is the tiny shape=(3, 4), where dim=0 normalizes to H and dim=None normalizes to (0,1) = HW. That's a single 3×4 tile on one core. So after this PR the test still touches those kernels, but never with more than one tile, never multi-core, and never through the rank-5 squeeze-to-4D path.

On the rank-2 shape the removed dims were genuinely pure duplicates (-2 ≡ 0, (-2,-1) ≡ None after normalization), so the PR's redundancy claim is correct there. The loss is entirely on the two rank-5 shapes.

**What other tests do and don't backfill**
This test is the only one in the repo that passes a floating-point scalar to std, and the only one that passes a float scalar to min.

So the PR body's "all dims are tested in other tests" holds for mean and mostly for var and max, but not for sum, min, or std. std is the clear gap: after this PR, std with a scalar on an H or HW reduction is only exercised on a 3×4 tensor. The std sweep at [tests/sweep_framework/sweeps/reduction/std/std.py](vscode-webview://082tq4q6k3bmgak39k85t8vdtvug5rrm7thg6hsu5k5d0u0671n7/tests/sweep_framework/sweeps/reduction/std/std.py) never sets scalar.

A concrete failure this would now miss: a bug in the Welford H kernel's scale-application that only shows up when the reduction spans more than one tile of height, for std specifically. Before the PR, dim=-2, shape=(3,4,8,56,33) (56 rows, 2 tiles) caught it; now nothing does.

tests/ttnn/nightly/unit_tests/operations/reduction/test_reduction_ops.py
bbradelTT: Refactored into two parts to avoid the cross product while still having some tests for the different values.

tests/ttnn/nightly/unit_tests/operations/reduction/test_reduction_ops.py
bbradelTT: I updated the coverage. The number of test cases is substantially less than the original, but should cover these cases. Please let me know if this is sufficient.

fplavecTT (approved):
Looks good now.
Two nits:
1. All helpers in this file other than the newly added one are named beginning with `_`
2. It would be good to document why the split is the way it is.

bbradelTT:
> Looks good now. Two nits:
> 
> 1. All helpers in this file other than the newly added one are named beginning with `_`
> 2. It would be good to document why the split is the way it is.

Updated.

PR #53829, opened 2026-08-20
Relocate 2x flag in metal.

uvelimirovicTT:
### Summary

Removes the `enable_2x_src_format` / `enable_2x_src_register` opt-in flag and makes MxFp4 →
`MxFp4_2x_B` src-register packing **automatic** for the two Quasar ops that support it: matmul
(`MVMUL`/`MVMULDI`) and column reduce (`GAPOOL`).

**Why the decision cannot live on the host.** The `jit_build` format tables are op-agnostic — they
see only the CB `buf_dataformat_arr`, never "this is a matmul" — and MxFp4 is also consumed by ops
that must *not* get the 2x format (typecast / SFPU / eltwise / pack / tilize all need the
`Float16_b` expansion). A host flag can therefore only ever be a hint the op writer has to remember
to set, and setting it wrongly fails silently. So `unpack_dst_format[]` now always keeps the safe MX
default (`Float16_b`), and the two ops that need 2x derive it themselves from the L1 format.

**Kernel-side auto-detect** (Quasar LLK API layer only — no `tt-llk` lib changes):

- `llk_unpack_AB_matmul_api.h`, `llk_unpack_AB_reduce_api.h` — when the operand's L1 format is
  MxFp4, override *only* the unpacker gasket `OUT_DATA_FORMAT` to `MxFp4_2x_B` through the existing
  `_llk_unpack_reconfig_data_format_src_`. The buffer descriptor is untouched — it carries the L1
  format and geometry, not the src-register format — so this lands in the same HW state the old
  host-side remap produced. Runs at init with the unpacker idle, before the first `UNPACR`.
- `llk_math_matmul_api.h`, `llk_math_reduce_api.h` — derive the effective src-register format the
  same way; it drives both the ALU format registers and `EN_X2` MOP selection. Reduce is gated to
  `REDUCE_COL`, since row/scalar reduce goes through post-pool `ELWADDDI`, which does not support 2x.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:uvelimirovic/relocate_2x_flag)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:uvelimirovic/relocate_2x_flag)

### DIFF
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/compute_hardware_config.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/compute_hardware_config.hpp
@@ -14,6 +14,12 @@
 
 namespace tt::tt_metal::experimental {
 
+// Compile-time define a compute kernel adds to KernelSpec::compiler_options.defines to opt into
+// the 2x-packed source-register format (Mxfp4 only). See the note in ComputeGen2Config below.
+// This is the single source of truth for the define name, shared by the host-side jit_build path
+// that detects it. Kept as a macro name (no value required); presence is what matters.
+inline constexpr const char* k2xSrcFormatDefine = "ENABLE_2X_SRC_FORMAT";
+
 // ============================================================================
 //  ComputeHardwareConfig
 // ============================================================================
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/compute_hardware_config.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/compute_hardware_config.hpp
@@ -166,23 +166,16 @@ struct ComputeGen2Config {
     // Temporary configs (these will change!)
     ///////////////////////////////////////////
 
-    // When true, the unpacker packs two values into each source-register slot instead of one.
-    // The math engine reads twice as many elements per pass, effectively doubling throughput.
+    // NOTE: The former `enable_2x_src_register` bool has been removed from this hardware config.
+    // The 2x-packed source-register format is now opted into per kernel via the compile-time
+    // define `ENABLE_2X_SRC_FORMAT` (see k2xSrcFormatDefine above): add it to the kernel's
+    // KernelSpec::compiler_options.defines. When present, jit_build emits the 2x-packed
+    // src-register format as the unpack_dst_format for 2x-capable (Mxfp4) inputs, so the math
+    // engine reads two elements per source-register slot, effectively doubling throughput.
     //
-    // This is currently ONLY supported for Mxfp4 data format. The setting is ignored for all
-    // other formats.
-    //
-    // WARNING: Only the matmul family of instructions work with this format:
-    //  - matmul (MVMUL/MVMULDI)
-    //  - the GAPOOL instruction that column reduce ops are built on
-    //
-    // Invoking other instructions on Mxfp4 data with the setting enabled will produce garbage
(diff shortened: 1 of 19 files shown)


rtawfik01:
@akerteszTT fyi for cleaning up the ComputeConfig

tt_metal/hw/inc/api/compute/matmul.h
```diff
@@ -111,6 +111,31 @@ ALWI void matmul_init(
...
+ * before the next op, or that op will keep unpacking the buffer as MxFp4_2x_B and produce garbage.
+ * A no-op on non-MxFp4 operands and on non-Quasar architectures.
+ *
+ * | Argument  | Description                                            | Type     | Valid Range | Required |
+ * |-----------|--------------------------------------------------------|----------|-------------|----------|
+ * | in0_cb_id | First input CB used in the matmul (same as matmul_init)| uint32_t | 0 to 31     | True     |
+ * | in1_cb_id | Second input CB used in the matmul                     | uint32_t | 0 to 31     | True     |
+ */
+// clang-format on
+ALWI void mm_uninit(uint32_t in0_cb_id, uint32_t in1_cb_id) {
```
bbradelTT, line 132: Will there be a follow up PR that defensively puts this uninit for all kernels that call mm_init?

bbradelTT (approved):

tt_metal/hw/inc/api/compute/matmul.h
uvelimirovicTT: Not really, I didn't plan it. The leak only bites if an MxFp4 matmul operand is reused by a non-matmul op in the same kernel and no current kernel does that. Also mm_uninit no-ops for non-MxFp4, so adding it everywhere would mostly result in dead calls. It also follows the existing idiom (reduce_uninit, BH face_compressed_mm_uninit): call the uninit when you leave the special path.

I could rename the mm_uninit to make it clear its used for MxFp4_2x. Perhaps the comment isn't enough.

rtawfik01 (approved):

tt_metal/api/tt-metalium/experimental/metal2_host_api/compute_hardware_config.hpp
akerteszTT, line 159: Would you mind also deleting this "Temporary configs" comment when you merge? This was the last thing in it -- woot!

akerteszTT (approved):

abhullar-tt (approved):

PR #53831, opened 2026-08-20
[bug fix] Fix CTA vararg not available in constexpr when assert is on

riverwuTT:
### Summary

Metal 2.0 exposes compile-time varargs to kernels via the generated `constexpr` accessor `get_compile_time_vararg(uint32_t)`, which bounds-checks with `ASSERT`.

Under lightweight kernel asserts, `ASSERT` expands to inline `asm ("ebreak")`. Inline assembly is not allowed in a C++17 `constexpr` function, and kernels build with `-std=c++17`, so the accessor fails to compile whenever asserts are on.

This PR moves the out-of-range `ASSERT` behind a non-`constexpr` helper and calls it only on the failing path. C++17 allows a `constexpr` function to call a non-`constexpr` function as long as constant evaluation never reaches that call, so in-range indices stay usable in constant expressions.

### Note to reviewers

Tests cover both sides with asserts enabled: an in-range index still folds in a constant expression, and an out-of-range index still fails the build. The assertion-on environment is simulated by defining the assertion-enable macro on the kernel.

Verified locally.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/m2-constexpr-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/m2-constexpr-fix)

### DIFF
--- a/tt_metal/jit_build/genfiles.cpp
+++ b/tt_metal/jit_build/genfiles.cpp
@@ -361,6 +361,16 @@ void write_kernel_args_generated_header(const std::filesystem::path& out_dir, co
     //   1. get_num_compile_time_varargs() — baked prefix length
     //   2. get_compile_time_vararg<idx>() — template index (with bounds check)
     //   3. get_compile_time_vararg(idx) — function-parameter index
+    //
+    // The out-of-range report for (3) sits behind a non-constexpr call rather than being an
+    // ASSERT in the constexpr body: the lightweight-assert flavor of ASSERT expands to inline
+    // asm ("ebreak"), which C++20 allows inside a constexpr function but C++17 rejects
+    // (-Werror=c++20-extensions), and kernels are built with -std=c++17. Calling a
+    // non-constexpr function from a constexpr one is fine in C++17 as long as constant
+    // evaluation never reaches the call, which an in-range index never does — so (3) stays
+    // usable in constant expressions under every assert flavor. Out-of-range indices keep the
+    // runtime assert, and in a constant expression they now fail the build outright instead of
+    // reading past the array.
     content << fmt::format(
         R"(
 FORCE_INLINE constexpr uint32_t get_num_compile_time_varargs() {{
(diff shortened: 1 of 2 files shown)


tt_metal/jit_build/genfiles.cpp
```diff
@@ -371,8 +371,17 @@ FORCE_INLINE constexpr uint32_t get_compile_time_vararg() {{
     static_assert(idx < get_num_compile_time_varargs(), "Compile-time vararg index out of range");
     return kernel_compile_time_args[idx];
 }}
+// Called when an OOB access to vararg CTA is attempted. Meant to trigger the assertion
+// failure as an indirection: the ASSERT macro may expand to inline asm, which cannot be
+// present in a C++17 constexpr function. This is a workaround and can be inlined once we
+// migrate to C++20.
+inline void assert_compile_time_vararg_index_out_of_range() {{
+    ASSERT(false);  // Attempt to access out of bound vararg CTA.
```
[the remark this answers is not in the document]
riverwuTT: I-think-this-is-fine copilot.

akerteszTT (approved):
Good catch!

tt_metal/jit_build/genfiles.cpp
```diff
@@ -371,8 +371,17 @@ FORCE_INLINE constexpr uint32_t get_compile_time_vararg() {{
     static_assert(idx < get_num_compile_time_varargs(), "Compile-time vararg index out of range");
     return kernel_compile_time_args[idx];
 }}
+// Called when an OOB access to vararg CTA is attempted. Meant to trigger the assertion
+// failure as an indirection: the ASSERT macro may expand to inline asm, which cannot be
+// present in a C++17 constexpr function. This is a workaround and can be inlined once we
+// migrate to C++20.
+inline void assert_compile_time_vararg_index_out_of_range() {{
```
akerteszTT, line 378: Out-of-scope for this targeted bugfix, but longer term we need to move a bunch of this Metal 2.0 stuff out of genfiles. A lot of stuff wound up emitted into the autogen header for expediency of development, that needs to be in a regular header.

riverwuTT: yeah this fix should really be on the assertion facility side, this is a vararg CTA targeted fix.

ruizhangTT (approved):

kstevensTT (approved):

Issue #51470, opened 2026-07-29
ttnn.maximum / ttnn.minimum / ttnn.clamp / ttnn.hardtanh silently drop NaN — a single SFPSWAP is a permutation, so max and min can never both return NaN


bvdyvncqr5-commits:
# ttnn.maximum / ttnn.minimum / ttnn.clamp / ttnn.hardtanh silently drop NaN — a single SFPSWAP is a permutation, so max and min can never both return NaN

`ttnn.maximum`, `ttnn.minimum`, `ttnn.clamp` and `ttnn.hardtanh` are each implemented by one or two bare `SFPSWAP` instructions. `SFPSWAP` is a compare-and-exchange, so its two output registers always hold a **permutation** of its two inputs. PyTorch requires `torch.maximum(NaN, x)` **and** `torch.minimum(NaN, x)` to *both* return NaN — a permutation of `{NaN, x}` can never be `(NaN, NaN)`.

This is a purely structural argument: **at least one of `ttnn.maximum` / `ttnn.minimum` is wrong for every NaN input, on both Wormhole and Blackhole, regardless of how the hardware orders NaN.**

Concretely, NaN is silently replaced by the other operand:

* `ttnn.minimum(NaN, 5.0)` returns `5.0`
* `ttnn.maximum(-NaN, 5.0)` returns `5.0`
* `ttnn.clamp(NaN, -2.0, 2.0)` returns `2.0`

The int32 variant in the same kernel already applies a correction for `SFPSWAP` ordering, so there is a ready structural precedent for what the float32 path would need.

### Summary

`SFPSWAP` is the hardware compare-and-exchange used for min/max. Three in-repo facts establish that it is a permutation:

1. `sfpi_lib.h:428` `min_max()` issues a single `__builtin_rvtt_sfpswap` and then uses `sfpselect2` to pick one of *that one instruction's* two outputs.
2. `swap(a, b)` in the same header is `SFPSWAP_MOD1_SWAP` — the instruction's semantics are literally "exchange or don't".
3. `ckernel_sfpu_binary_max_min.h:78` contains the comment *"Conditionally swap again to fix the cases where SFPSWAP got the result backwards"* — only a swap network makes "got it backwards, swap again" meaningful.

[...]

PR #53861, opened 2026-08-20
[Bug fix] clamp: return NaN for a NaN input with tensor bounds

### LINKED ISSUE #53860: ttnn.clamp: a branch that picks between two ways of writing maximum
**Component / Area:** ttnn eltwise unary (composite)

**Issue Type:** Performance

## Observed

`ttnn.clamp(x, min, max)` with tensor bounds selects between two ways of writing the same thing:

```cpp
Tensor a_max = ttnn::minimum(input_a, max.value(), ...);
Tensor temp = ttnn::where(
    ttnn::eq(min.value(), 0.0f, ...),
    ttnn::relu(a_max, ...),
    ttnn::maximum(a_max, min.value(), ...), ...);
return ttnn::where(ttnn::gt(min.value(), max.value(), ...), max.value(), temp, ...);
```

`relu(a_max)` is `maximum(a_max, 0)`, and that arm is only taken where `min` is 0, so both arms of the `where` evaluate to `maximum(a_max, min)`. The `eq`, the `relu` and the `where` that chose between them compute a value the op already has.

Measured on P300 and N300, `[1, 1, 512, 512]`, 33 invocations, device kernels per call: **7** on both.

## Expected

Four kernels per call, with the `gt(min, max)` guard kept.

Two would be enough on paper — `minimum(maximum(x, min), max)` answers the degenerate `min > max` correctly with no guard. It is not free: `ttnn.maximum` propagates NaN and `ttnn.minimum` does not, in either operand order, so only the order ending in `maximum` can return a NaN bound, and that is the order that needs the guard. Going to two turns `clamp(1.0, min=nan, max=2.0)` from `nan` into `2.0`. See the PR for the measurement.

## Not established

- The scalar-bound path, already fused through `ckernel_sfpu_clamp.h` and untouched.
- Whether the remaining four can become one. That needs a ternary SFPU clamp kernel rather than a deletion.
- `bfloat8_b` and the sharded paths.

## Related

#49996 lists this reduction in its first section as an assumption to validate, and names the ternary clamp kernel that would take it to one. This issue is the validation and the measurement; the four is where a deletion can get without the kernel.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53860.

`torch.clamp` returns NaN for a NaN input whatever the bounds are. The tensor-bounds path returns a bound instead. Measured on P300 against the current three-kernel form:

| x | min | max | ttnn float32 | ttnn bfloat16 | torch |
|---|---|---|---|---|---|
| NaN | 0 | 1 | 1 | 0 | NaN |
| NaN, sign bit set | 0 | 1 | 0 | 0 | NaN |
| NaN | 1 | 2 | 2 | 1 | NaN |
| 5 | 0 | 1 | 1 | 1 | 1 |
| -5 | 0 | 1 | 0 | 0 | 0 |
| 0.5 | 0 | 1 | 0.5 | 0.5 | 0.5 |
| 2 | 3 | 1 | 1 | 1 | 1 |

`minimum` and `maximum` are a single `SFPSWAP`, which is a permutation, so the two cannot both return NaN: `minimum` drops it and `maximum` keeps it (#51470).

```
a = [1.0, nan], b = [nan, 1.0]
  ttnn.minimum(a, b) = [1.0, 1.0]     torch.minimum -> [nan, nan]
  ttnn.maximum(a, b) = [nan, nan]     torch.maximum -> [nan, nan]
```

The form in `main` is `maximum(minimum(x, max), lo)`, so the first `minimum` replaces a NaN input with `max` and the `maximum` that follows never sees one. A `where` on `isnan` restores it. The NaN is taken from `input_a` rather than written as a scalar, which keeps the sign and payload the caller passed in.

## What changed since this PR was opened

This PR originally carried a performance change as well, 7 kernels down to 6. #55883 landed the same idea and went further, 7 down to 3, including the degenerate `min > max` case that this PR had kept a `gt` guard for. That work is better than what was here and this PR no longer touches it. The branch is rebased onto `main` and now adds only the NaN guard, 3 kernels to 5.

#55883 verified the result bit-exact against `torch.clamp` including the degenerate case. NaN was not part of that check.

## Not covered

bfloat16 still returns a bound, and the reason is outside this op. `ttnn.where` turns a bfloat16 NaN in its true arm into `inf`, while the same call in float32 keeps it. Measured on P300 with this guard in place:

| call | bfloat16 | float32 |
|---|---|---|
| `from_torch` round trip of `[nan, 1.0, nan, 2.0]` | `[nan, 1.0, nan, 2.0]` | `[nan, 1.0, nan, 2.0]` |
| `ttnn.isnan(x)` | `[1, 0, 1, 0]` | `[1, 0, 1, 0]` |
| `ttnn.where(isnan(x), x, y)` | `[inf, 7.0, inf, 7.0]` | `[nan, 7.0, nan, 7.0]` |

So the regression test here is float32 only. The bfloat16 half needs the `where` behaviour fixed first, and I will file that separately.

The guard costs two kernels. It comes off entirely if `minimum` is ever made to propagate NaN, which is #51470.

## Tests

`tests/ttnn/unit_tests/operations/eltwise/test_clamp_tensor_bounds.py`, new: 14 passed. It covers ordinary bounds, `min == 0`, the degenerate `min > max`, and a NaN input of either sign against four bound pairs.

`tests/ttnn/unit_tests/operations/eltwise/test_composite.py -k "clamp or clip"`: 138 passed, the same count #55883 reports.

Both on P300, `main` at `48116eecf01`.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_clamp_tensor_bounds.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_clamp_tensor_bounds.py
@@ -28,3 +28,24 @@ def test_clamp_tensor_bounds(device, dtype, case):
 
     want = torch.clamp(x.float(), lo.float(), hi.float()).to(torch_dtype)
     assert torch.equal(got, want), f"{int((got != want).sum())} of {want.numel()} differ"
+
+
+@pytest.mark.parametrize("sign", [1.0, -1.0], ids=["positive_nan", "negative_nan"])
+@pytest.mark.parametrize("bounds", [(0.0, 0.0), (0.0, 5.0), (-1.0, 1.0), (2.0, -2.0)])
+def test_clamp_nan_input_fp32(device, sign, bounds):
+    """A NaN input is NaN out, whatever the bounds and whichever sign the NaN has.
+
+    The sign matters: minimum and maximum are one SFPSWAP and order by bit
+    pattern, so a NaN with the sign bit set survives one of them and not the other.
+    """
+    nan = (torch.tensor(float("nan")) * sign).item()
+    x = torch.full((1, 1, 32, 32), nan, dtype=torch.float32)
+    lo = torch.full((1, 1, 32, 32), bounds[0], dtype=torch.float32)
+    hi = torch.full((1, 1, 32, 32), bounds[1], dtype=torch.float32)
+
+    t = [ttnn.from_torch(v, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device) for v in (x, lo, hi)]
+    got = ttnn.to_torch(ttnn.clamp(t[0], t[1], t[2]))
+
+    assert (
+        got.isnan().all()
+    ), f"{int((~got.isnan()).sum())} of {got.numel()} came back not NaN, first {got.flatten()[0]}"
--- a/tests/ttnn/unit_tests/operations/eltwise/test_clamp_tensor_bounds.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_clamp_tensor_bounds.py
@@ -1,0 +1,30 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""clamp with tensor bounds, over the cases the removed branch distinguished."""
+
+import pytest
+import torch
+import ttnn
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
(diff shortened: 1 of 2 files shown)


mateusznowakTT:
@kinginu : Please resolve conflicts on this PR.

mateusznowakTT:
@kinginu : Is there any model / application suffering from this (other than unit test) ?
If not, I'd prefer not to merge it, as it will hurt performance significantly.
I'd prefer an improvement to min/max kernel to handle this, if it is possible without adding much overhead there.

PR #53873, opened 2026-08-21
[Perf] xlogy_bw: drop the two wheres on eq(x, nan) and fold the guards

### LINKED ISSUE #53872: ttnn.xlogy_bw: two wheres on eq(x, nan) that never select
**Component / Area:** ttnn eltwise binary backward (composite)

**Issue Type:** Performance

## Observed

`xlogy_bw` in `binary_backward.cpp:283` guards both gradients with `where(eq(x, nan), nan, ...)`:

```cpp
grad1_result = ttnn::where(
    ttnn::eq(input_a, std::nanf(" "), std::nullopt, output_mem_config), std::nanf(" "), grad1_result, ...);
```

Nothing is equal to nan, itself included, so neither where ever selects. Four of the nineteen dispatches cannot affect the result. On P300 and N300 alike:

```
ttnn.eq(x, nan) for x = [nan, 1, 0, -1, inf, -inf, nan, 2]
  -> [0, 0, 0, 0, 0, 0, 0, 0]
ttnn.isnan(x)   -> [1, 0, 0, 0, 0, 0, 1, 0]
```

The rest of the op is unfolded: `eqz(a)` and `le(b, 0)` are the two operands of the `logical_and` that reads them, the `reciprocal` is one multiply's operand, and the `sign` is another's.

|---|---|---|---|---|
| dispatches per call | 19 | 19 | 19 | 19 |
| TRISC_1 core-cycles per call | 21,226,376 | 42,105,950 | 17,332,175 | 27,684,022 |

Wall clock on P300, 100 calls with one synchronise at the end: 374.6 us per call in bfloat16 and 368.3 us in float32 at `[1, 1, 512, 512]`, against 339.8 and 342.0 us at a single tile. The op costs nearly the same at 256 tiles as at one, so what it spends is the enqueueing.

## Expected

Eleven dispatches per call, and no change to any result.

## Not established

- Whether other ops test for nan the same way. `eq(x, nan)` is easy to grep for and this is one of the places it appears.

## Related

#52036 is the same mistake in the forward kernel: `in1 == nan` there can never be true either. This is the backward composite, a different file, and the two do not overlap.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Performance.

## Summary

Closes #53872.

Both gradients were guarded with `where(eq(x, nan), nan, ...)`. Nothing is equal to nan, itself included, so neither where ever selected, and four of the nineteen dispatches could not affect the result. Removing them changes nothing.

The `grad_a` guard would have been wrong if it had fired: `torch.xlogy` takes `log(b)` as the gradient in `a` whatever `a` is, nan included.

The rest is operand activations. `eqz(a)` and `le(b, 0)` are the two operands of the `logical_and` that reads them, the `reciprocal` is the right operand of the multiply that reads it, and the `sign` is the left operand of the multiply by inf.

## Accuracy

Five case sets per dtype on each machine, both gradients, against what the composite returned.

|---|---|---|---|---|
| ordinary | identical | identical | identical | identical |
| b negative | identical | identical | identical | identical |
| a zero, b not positive | identical | identical | identical | identical |
| b zero | identical | identical | identical | identical |
| nan, ±0, ±1 | identical | identical | identical | identical |

**Exhaustive.** Every one of the 65536 bfloat16 bit patterns, in each of the three operand positions, against five settings of the other two — the ordinary case, either operand zero, mixed signs, and 1e30 against 1e-30 — both gradients. 1966080 comparisons; and 125829120 more over 4194304 float32 bit patterns drawn uniformly from the 2^32 space. Output bits compared as integers against the composite:

| | P300 bfloat16 | P300 float32 | N300 bfloat16 | N300 float32 |
|---|---|---|---|---|
| identical to the composite | 1965826 | 125820943 | 1965826 | 125820943 |
| differ | **254** | **8177** | **254** | **8177** |

The case sets above did not reach them and the two machines agree exactly. Every one has a subnormal operand:

| | composite | this branch | `torch` |
|---|---|---|---|
| bfloat16, `grad=1, a=9.18e-41, b=0`, `grad_a` | `-inf` | `0` | `-inf` |
| float32, `grad=1, a=0, b=9.32e-39`, `grad_a` | `0` | `nan` | `-87.5686` |

Neither build is right in the float32 row. Both are answering a question about whether `b` is non-positive, and neither the composite's spelling of that question nor this branch's gives the right answer for a subnormal.

## Whether folding a zero test is allowed

`eqz(a)` and `le(b, 0)` are folded onto `logical_and` as operand activations. Those are not the same tests as the standalone ops. Measured on P300 with the activation arguments exposed through a local binding, for a subnormal operand:

| | `ttnn.eqz(x)` | `ttnn.eq(x, 0.0)` | EQZ folded as an operand activation | exact |
|---|---|---|---|---|
| bfloat16 | 0 | **1** | **1** | 0 |
| float32 | 0 | **1** | 0 | 0 |

`nez`, `lez` and `gtz` behave the same way. Two separate things are visible here. **A comparison against the scalar 0 reads a subnormal as zero in both dtypes** — that is what the composite's `le(b, 0.0f)` does, and it is why the composite is also wrong in the float32 row above, before this PR touches anything. **The fold follows the scalar spelling in bfloat16 and the unary spelling in float32**, which is where the 254 come from.

So the three spellings of `x == 0` give three answers, and this PR moves from one to another. That is a question about the activation contract rather than about this op, and it is the same question in #53871.

If folding a zero test is not allowed, the guard goes back to the standalone ops and the count is 19 → 13 rather than 19 → 11. It is one hunk:

```diff
-        ttnn::logical_and(input_a, other, std::nullopt, output_mem_config, std::nullopt, {}, is_zero, is_le_zero),
+        ttnn::logical_and(
+            ttnn::eqz(input_a, output_mem_config),
+            ttnn::le(other, 0.0f, std::nullopt, output_mem_config),
+            std::nullopt,
+            output_mem_config),
```

Measured with that applied, all 254 and all 8177 go away: **bit-identical to the composite on every pattern, in both dtypes, on both machines.** Say which one you want.

## Cost

|---|---|---|---|---|
| dispatches per call | 19 → **11** | 19 → **11** | 19 → **11** | 19 → **11** |
| TRISC_1, before | 21,226,376 | 42,105,950 | 17,332,175 | 27,684,022 |
| TRISC_1, after | **14,626,791** | **27,682,324** | **11,765,047** | **18,197,155** |
| | **1.45x** | **1.52x** | **1.47x** | **1.52x** |

Wall clock on P300, 100 calls with one synchronise at the end, profiler off:

| | 1 tile | 256 tiles | 1024 tiles |
|---|---|---|---|
| bfloat16 | 339.8 → **225.9** µs | 374.6 → **249.0** µs | 378.0 → **232.8** µs |
| float32 | 342.0 → **203.7** µs | 368.3 → **228.5** µs | 512.3 → **319.0** µs |

The ratio is the same at one tile as at 1024, and the before column barely moves with size, so at these shapes the op is bound by how many programs it enqueues rather than by what they compute. That is 14 to 24 µs per dropped dispatch.

## Tests

`test_xlogy_bw_edges.py`, 6 cases: against `torch.xlogy` backward on both dtypes, plus the `a == 0, b <= 0` corner the `logical_and` selects and the `b == 0` corner the `sign(grad) * inf` selects. The 3 existing `test_backward_xlogy.py` cases pass unchanged.

## Not covered

- `grad_b` is 0 or inf where `other` is nan and `torch.xlogy` has nan. That is what the removed wheres were reaching for; `eq` is not the test for it and `isnan` is. Left out so that this PR stays a no-op on results.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_xlogy_bw_edges.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_xlogy_bw_edges.py
@@ -1,0 +1,62 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+import ttnn
+
+NAN = float("nan")
+
+
+def run(device, dtype, grad, a, b):
+    def dev(t):
+        return ttnn.from_torch(t, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    return [ttnn.to_torch(t).float() for t in ttnn.xlogy_bw(dev(grad), dev(a), dev(b))]
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_xlogy_bw_matches_torch(device, dtype):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    torch.manual_seed(0)
+    a = (torch.rand([1, 1, 32, 32]) * 10 - 5).to(torch_dtype)
+    b = (torch.rand([1, 1, 32, 32]) * 10 + 0.5).to(torch_dtype)
+    grad = (torch.rand([1, 1, 32, 32]) * 2 - 1).to(torch_dtype)
+
+    got_a, got_b = run(device, dtype, grad, a, b)
+
+    ta = a.float().clone().requires_grad_(True)
+    tb = b.float().clone().requires_grad_(True)
+    torch.xlogy(ta, tb).backward(grad.float())
+
+    tol = 0.05 if dtype == ttnn.bfloat16 else 1e-4
+    assert (got_a - ta.grad).abs().max() < tol
+    assert ((got_b - tb.grad).abs() / tb.grad.abs().clamp(min=1.0)).max() < tol
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_xlogy_bw_a_zero_b_not_positive(device, dtype):
+    # The logical_and of eqz(a) and le(b, 0) selects zero for grad_a here, and
+    # the ltz(b) below it selects nan where only b is negative.
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    zero = torch.zeros([1, 1, 32, 32], dtype=torch_dtype)
+    one = torch.ones([1, 1, 32, 32], dtype=torch_dtype)
+
+    got_a, _ = run(device, dtype, one, zero, -one)
+    assert (got_a == 0).all()
+
+    got_a, _ = run(device, dtype, one, one, -one)
+    assert not torch.isfinite(got_a).any()
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_xlogy_bw_b_zero(device, dtype):
+    # grad_b is sign(grad) * inf where b is zero.
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    zero = torch.zeros([1, 1, 32, 32], dtype=torch_dtype)
+    a = torch.full([1, 1, 32, 32], 2.0, dtype=torch_dtype)
+    grad = torch.full([1, 1, 32, 32], -1.0, dtype=torch_dtype)
+
+    _, got_b = run(device, dtype, grad, a, zero)
+    assert (got_b == float("-inf")).all()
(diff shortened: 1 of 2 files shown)


kanapitsas:
Agreed on the two removed `eq(x, nan)` branches: nothing compares equal to `NaN`, so
removing them does not change the result.

Two separate pre-existing cases remain when `other == 0`:

1. The input gradient uses `grad * log(other)`. For `grad == 0`, this becomes
   `0 * -inf = NaN`, while PyTorch returns `0`.
2. The `other` gradient uses `sign(grad) * inf` and ignores `input`. If `input == 0`,
   the derivative is `0/0`: PyTorch returns `NaN`, while `ttnn` returns `+Inf`.

| grad | input | other | output | PyTorch | current `ttnn` |
|---:|---:|---:|---|---:|---:|
| 0 | 1 | 0 | grad input | 0 | NaN |
| 1 | 0 | 0 | grad other | NaN | +Inf |

These cases predate the PR. The current PCC tests can miss them because `comp_pcc` maps
NaN/Inf to zero before comparison.

kinginu:
Thanks, confirmed both on device.

One correction: the first row is float32 only. In bfloat16 `grad_input` comes back `0`, which
matches torch, because `0 * -inf` is plain `0` here. The second one is also wider than the one
cell — `grad_other` never reads `input`, so it is wrong for every negative `input`, not only zero.

I will apply the fix to this PR.

PR #53885, opened 2026-08-21
[Perf] acosh_bw / acos_bw / atanh_bw: fold the guards into the ops that read them

### LINKED ISSUE #53884: ttnn acosh_bw / acos_bw / atanh_bw: the guards cost a dispatch each
**Component / Area:** ttnn eltwise unary backward (composite)

**Issue Type:** Performance

## Observed

`acosh_bw` (`unary_backward.cpp:573`), `acos_bw` (`:618`) and `atanh_bw` (`:951`) spend most of their dispatches on the guards rather than the gradient.

```cpp
Tensor in_sq = ttnn::square(input, output_mem_config);
Tensor in_rsqrt = ttnn::rsqrt(ttnn::subtract(in_sq, 1.0f, ...), false, ...);
...
ttnn::logical_or(
    ttnn::lt(in_sq, 1.0f, ...),
    ttnn::logical_and(ttnn::eq(input, 1.0f, ...), ttnn::eqz(grad, ...), ...), ...)
```

`square`, `signbit`, `sign`, `neg`, `rsqrt`, `eqz`, `nez`, `ltz` and the comparisons against a constant are each their own dispatch, and each is one operand's activation on the binary op that reads it.

Two guards also collapse on their own terms. `x == 1 or x == -1` is `abs(x) == 1`, which turns a pair of tests and a pair of `where`s writing the same value into one of each. `acosh_bw` keeps `in_sq` only to test it against 1, so squaring twice as an activation costs less than the tensor.

| | dispatches | N300 bfloat16 | N300 float32 | P300 bfloat16 | P300 float32 |
|---|---|---|---|---|---|
| acosh_bw | 19 | 20,089,731 | 41,010,406 | 15,947,753 | 26,514,870 |
| acos_bw | 16 | 17,757,294 | 35,728,529 | 14,250,269 | 23,859,707 |
| atanh_bw | 18 | 20,058,473 | 40,605,163 | 16,079,670 | 26,422,580 |

TRISC_1 core-cycles per call. Wall clock on N300 at 256 tiles is 810.7 us for `acosh_bw` in bfloat16 against 750.8 us at a single tile, so what these ops spend at ordinary shapes is the enqueueing.

## Expected

Eleven, eight and eleven dispatches per call.

Exhaustive sweeps say what moves: nothing at all in float32 over all 2^32 bit patterns, and in bfloat16 over all 65536 patterns only the 254 nan encodings, plus `atanh_bw` at `grad == 0` on subnormals, where the folded form returns 0 and matches torch where the composite returns inf. The PR carries the tables.

## Not established

- The remaining dispatches are `where`s, which take no activations on their predicate. Folding those needs the ternary API to carry them.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Performance.

## Summary

Closes #53884.

Most of what these three spend is unary steps feeding a binary op that applies a unary chain to its operands anyway. `square`, `signbit`, `sign`, `neg`, `rsqrt`, `eqz`, `nez`, `ltz` and the comparisons against a constant are each one operand's activation, so none of them needs a dispatch.

Two guards collapse on their own terms. `x == 1 or x == -1` is `abs(x) == 1`, which turns a pair of tests and a pair of `where`s writing the same value into one of each. `acosh_bw` kept `in_sq` only to test it, so squaring twice as an activation costs less than the tensor.

`atanh_bw` keeps its `eq` against inf as a binary op: as a `UNARY_EQ` activation it does not match inf, which left the sign of the infinity at the boundary unflipped. Filed separately.

## Accuracy

**Exhaustive, not sampled. Every number below is this branch against the composite it replaces** — the `04cf22f7ee` build and this build, run on the same device, output bits compared exactly as integers. It is not a comparison against IEEE or against torch; where torch enters it is named.

**bfloat16: every one of the 65536 bit patterns, against six gradients** (1, -1, 0, 0.5, 1e30, 1e-30), P300, 393216 input points per op.

| op | patterns differing from the composite | where |
|---|---|---|
| acosh_bw | **0** | — |
| acos_bw | 1524 | the 254 nan bit patterns, at all six gradients |
| atanh_bw | 254 | subnormal inputs at `grad == 0` only |

**float32: every one of the 2^32 bit patterns, against three gradients** (1, 0, -1), P300, 12.9 billion input points per op.

| op | chunks differing from the composite |
|---|---|
| acosh_bw | **0 of 12288** |
| acos_bw | **0 of 12288** |
| atanh_bw | **0 of 12288** |

So in float32 nothing moves at all, and in bfloat16 nothing moves on a finite input except one case, which is an improvement:

| case | base | this PR | torch |
|---|---|---|---|
| `atanh_bw(grad=0, x` subnormal`)` | inf | **0** | **0** |

The 254 nan-input patterns are the complete set of bfloat16 nan encodings. Neither form returns nan there, because bfloat16 loses nan through the pack; they disagree about which non-nan value comes out. `torch` returns nan for all of them.

**Improvement or regression?** Every differing pattern, classified against a float64 reference for the same formula torch uses:

| op | patterns differing | fused closer to torch | base closer to torch | reference is nan, neither can match |
|---|---|---|---|---|
| acosh_bw | 0 | — | — | — |
| acos_bw | 1524 | 0 | **0** | 1524 |
| atanh_bw | 254 | **254** | **0** | 0 |

Nothing regresses. `acos_bw`'s 1524 are all nan inputs, where the reference is nan and neither build can return one, because bfloat16 loses nan through the pack. `atanh_bw`'s 254 are the subnormal case, and there this branch is closer to torch on every one of them.

For scale, against the correctly rounded torch answer over all 65536 bfloat16 inputs at `grad = 1`, the two builds agree exactly with each other and land on the reference for `acosh_bw` 12790 times, `acos_bw` 32374, `atanh_bw` 38472. Those counts are unchanged by this PR.

## Cost

| | dispatches | N300 bfloat16 | N300 float32 | P300 bfloat16 | P300 float32 |
|---|---|---|---|---|---|
| acosh_bw | 19 → **11** | 20,089,731 → **15,501,512** (1.30x) | 41,010,406 → **30,353,271** (1.35x) | 15,947,753 → **12,916,413** (1.23x) | 26,514,870 → **19,749,842** (1.34x) |
| acos_bw | 16 → **8** | 17,757,294 → **11,650,219** (1.52x) | 35,728,529 → **21,708,313** (1.65x) | 14,250,269 → **9,818,053** (1.45x) | 23,859,707 → **14,737,941** (1.62x) |
| atanh_bw | 18 → **11** | 20,058,473 → **14,506,990** (1.38x) | 40,605,163 → **28,623,279** (1.42x) | 16,079,670 → **11,675,489** (1.38x) | 26,422,580 → **18,556,393** (1.42x) |

Wall clock on N300, 100 calls with one synchronise at the end, profiler off, microseconds per call.

| | | bfloat16 | float32 |
|---|---|---|---|
| acosh_bw | 1 tile | 750.8 → **469.4** (1.60x) | 769.0 → **477.7** (1.61x) |
| | 256 tiles | 810.7 → **499.5** (1.62x) | 816.2 → **507.3** (1.61x) |
| | 1024 tiles | 816.3 → **537.1** (1.52x) | 1260.6 → **831.0** (1.52x) |
| acos_bw | 1 tile | 572.0 → **341.4** (1.68x) | 592.9 → **354.2** (1.67x) |
| | 256 tiles | 631.5 → **358.1** (1.76x) | 630.9 → **364.3** (1.73x) |
| | 1024 tiles | 629.9 → **369.4** (1.71x) | 1097.8 → **627.2** (1.75x) |
| atanh_bw | 1 tile | 685.6 → **455.6** (1.50x) | 677.6 → **464.5** (1.46x) |
| | 256 tiles | 742.1 → **487.9** (1.52x) | 735.1 → **503.6** (1.46x) |
| | 1024 tiles | 746.3 → **499.5** (1.49x) | 1233.6 → **822.0** (1.50x) |

## Tests

`test_inverse_bw_domains.py`, 24 cases plus 2 skipped: each op against its torch backward across four gradient shapes on both dtypes, plus the `|x| == 1` boundary of `atanh_bw` where the sign of the infinity follows the incoming gradient. The 15 existing cases in `test_backward_acosh.py`, `test_backward_acos.py` and `test_backward_atanh.py` pass unchanged.

## Not covered

- The remaining dispatches are `where`s. `ttnn::where` takes no activations on its predicate, so folding those needs the ternary API to carry them.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_inverse_bw_domains.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_inverse_bw_domains.py
@@ -1,0 +1,74 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import torch
+import ttnn
+
+OPS = [
+    (ttnn.acosh_bw, torch.acosh, "outside"),
+    (ttnn.acos_bw, torch.acos, "inside"),
+    (ttnn.atanh_bw, torch.atanh, "inside"),
+]
+
+GRADS = ["ones", "mixed", "zeros", "negative"]
+
+
+def inputs(where, dtype):
+    torch.manual_seed(0)
+    if where == "inside":
+        return (torch.rand([1, 1, 32, 32]) * 1.8 - 0.9).to(dtype)
+    return (torch.rand([1, 1, 32, 32]) * 8 + 1.5).to(dtype)
+
+
+def grads(which, dtype):
+    if which == "ones":
+        return torch.ones([1, 1, 32, 32], dtype=dtype)
+    if which == "zeros":
+        return torch.zeros([1, 1, 32, 32], dtype=dtype)
+    if which == "negative":
+        return -torch.ones([1, 1, 32, 32], dtype=dtype)
+    torch.manual_seed(1)
+    return (torch.rand([1, 1, 32, 32]) * 2 - 1).to(dtype)
+
+
+@pytest.mark.parametrize("ttnn_op, torch_op, domain", OPS)
+@pytest.mark.parametrize("which_grad", GRADS)
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_backward_in_domain_matches_torch(device, ttnn_op, torch_op, domain, which_grad, dtype):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    x = inputs(domain, torch_dtype)
+    grad = grads(which_grad, torch_dtype)
+
+    def dev(t):
+        return ttnn.from_torch(t, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    got = ttnn.to_torch(ttnn_op(dev(grad), dev(x))[0]).float()
+
+    tx = x.float().clone().requires_grad_(True)
+    torch_op(tx).backward(grad.float())
+
+    keep = torch.isfinite(tx.grad) & torch.isfinite(got)
+    if keep.sum() == 0:
+        # A zero gradient makes the op answer nan by design; there is nothing
+        # finite left to compare against torch, which answers zero.
+        pytest.skip("no finite element to compare")
+    tol = 0.05 if dtype == ttnn.bfloat16 else 1e-4
+    assert ((got[keep] - tx.grad[keep]).abs() / tx.grad[keep].abs().clamp(min=1.0)).max() < tol
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+def test_atanh_bw_sign_at_the_boundary(device, dtype):
+    # At |x| == 1 the gradient is an infinity whose sign follows the incoming
+    # gradient. The last where in the op is what flips it, and it tests the
+    # result against inf, which UNARY_EQ as an activation does not match.
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    one = torch.ones([1, 1, 32, 32], dtype=torch_dtype)
+
+    def dev(t):
+        return ttnn.from_torch(t, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+    assert (ttnn.to_torch(ttnn.atanh_bw(dev(one), dev(one))[0]).float() == float("inf")).all()
+    assert (ttnn.to_torch(ttnn.atanh_bw(dev(-one), dev(one))[0]).float() == float("-inf")).all()
+    assert (ttnn.to_torch(ttnn.atanh_bw(dev(-one), dev(-one))[0]).float() == float("-inf")).all()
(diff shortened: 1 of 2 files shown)


kanapitsas:
A separate correctness issue remains in `atanh_bw`:

```cpp
Tensor grad_is_zero = ttnn::eqz(grad, output_mem_config);
grad_a = where(grad_is_zero, t_nan, grad_a, output_mem_config);
```

This makes every zero upstream gradient `NaN`. For finite inputs inside the domain,
PyTorch returns zero:

| grad | input | PyTorch | current `ttnn` |
|---:|---:|---:|---:|
| 0 | 0.5 | 0 | NaN |
| 0 | -0.5 | 0 | NaN |

`NaN` is only needed at `abs(input) == 1`, where the denominator is zero. The condition
should include that boundary instead of testing `grad` alone.

This behavior predates the PR; the PR simply keeps it. The current PCC test misses it
because `comp_pcc` maps NaN/Inf to zero before comparison.

kinginu:
Thanks, confirmed on device. It reaches the whole domain rather than ±0.5 in particular: with a
zero gradient every `|x| < 1` is wrong, and in bfloat16 the value is `inf` rather than `nan`.

I will apply the fix to this PR.

mateusznowakTT:
All the UTs in this PR pass on main - the issue seems to be no longer.

PR #53899, opened 2026-08-21
[Cleanup] Forward output_mem_config on the outer multiply in addcdiv_bw

### LINKED ISSUE #53874: addcdiv_bw: outer multiply() omits output_mem_config — no behaviour change
Item 11 of master ticket

VirdhatchaniKN:
### Issue
https://github.com/tenstorrent/tt-metal/issues/53874

### Summary
`output_mem_config` is threaded through all 12 other op calls in `addcdiv_bw`, but the outer `ttnn::multiply` that builds `grad_a` omitted it -- the only call in the function that does. The two structurally identical multiplies right below it (`tmp` and `grad_b`) both pass `std::nullopt, output_mem_config`.

This is a consistency fix, not a behaviour fix. `ttnn::multiply` with no memory_config inherits `input_a`'s config, and here `input_a` is the inner multiply's result, which was already placed in `output_mem_config`. Because a single `output_mem_config` value feeds both, the inherited and requested configs cannot diverge, so `grad_a` already landed in the right place.

Verified: reverting this change and re-running the same probe gives identical results --

  requested L1   -> outputs [DRAM, L1, L1]     (both with and without the fix)
  requested DRAM -> outputs [DRAM, DRAM, DRAM] (both with and without the fix)

and test_backward_addcdiv.py passes 12/12 either way. The value is that the call no longer depends on the inner call's placement happening to match, so a later refactor of the inner call cannot silently move `grad_a`.

Addresses item 11 of #48240, which is filed as a Medium-severity memory-config bug; it is really a no-op inconsistency, and worth reclassifying there.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/addcdiv_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/addcdiv_bw_check)

### DIFF
--- a/ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
@@ -47,7 +47,10 @@ std::vector<Tensor> addcdiv_bw(
     float t_inf = std::numeric_limits<float>::infinity();
     float t_nan = std::nanf("");
     Tensor grad_a = ttnn::multiply(
-        ttnn::multiply(grad, value, std::nullopt, output_mem_config), ttnn::reciprocal(tensor2, output_mem_config));
+        ttnn::multiply(grad, value, std::nullopt, output_mem_config),
+        ttnn::reciprocal(tensor2, output_mem_config),
+        std::nullopt,
+        output_mem_config);
     grad_tensor.emplace_back(ttnn::where(
         ttnn::eqz(tensor2, output_mem_config),
         ttnn::where(ttnn::eqz(grad, output_mem_config), t_nan, t_inf, output_mem_config),
--- a/ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
@@ -43,6 +43,9 @@ std::vector<Tensor> addcdiv_bw(
     auto output_mem_config = memory_config.value_or(input_a.memory_config());
     std::vector<Tensor> grad_tensor;
     grad_tensor.reserve(3);
+    // grad is passed through unchanged, so output[0] keeps grad's memory config rather
+    // than output_mem_config. Intentional: no eltwise backward op relocates the
+    // passthrough gradient. See #53874.
     grad_tensor.emplace_back(grad);
     float t_inf = std::numeric_limits<float>::infinity();
     float t_nan = std::nanf("");


ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
```diff
@@ -47,7 +47,10 @@ std::vector<Tensor> addcdiv_bw(
     float t_inf = std::numeric_limits<float>::infinity();
     float t_nan = std::nanf("");
     Tensor grad_a = ttnn::multiply(
-        ttnn::multiply(grad, value, std::nullopt, output_mem_config), ttnn::reciprocal(tensor2, output_mem_config));
+        ttnn::multiply(grad, value, std::nullopt, output_mem_config),
+        ttnn::reciprocal(tensor2, output_mem_config),
+        std::nullopt,
+        output_mem_config);
```
Aswinmcw, line 56: Your own probe output flags a second gap in this function: `requested L1 -> outputs [DRAM, L1, L1]`. That first `DRAM` is `grad_tensor.emplace_back(grad)` on line 46 — `grad` is returned by reference, so `output[0]` is in `grad`'s config regardless of what the caller asked for. Unlike this multiply, that one *can* diverge from the request, so if #48240 item 11 is really about memory-config correctness it's the more substantive half.

Same passthrough exists in `addcmul_bw` (line 26) and one spot in `unary_backward.cpp` (line 388), and no eltwise backward op uses `to_memory_config` today — so it looks like accepted convention rather than an oversight. Either way, worth a line in #53874 saying it's deliberate and out of scope, so the next person reading that probe output doesn't re-file it.

Aswinmcw (approved):
LGTM — one note below.

ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
```diff
@@ -43,11 +43,17 @@ std::vector<Tensor> addcdiv_bw(
     auto output_mem_config = memory_config.value_or(input_a.memory_config());
     std::vector<Tensor> grad_tensor;
     grad_tensor.reserve(3);
+    // grad is passed through unchanged, so output[0] keeps grad's memory config rather
+    // than output_mem_config. Intentional: no eltwise backward op relocates the
+    // passthrough gradient. See #53874.
```
Aswinmcw, line 48: `#53874` is still just "Item 11 of master ticket" with zero comments, so this pointer sends the next reader somewhere that doesn't explain the passthrough. One comment on that issue with the rationale (and that the same shape exists in `addcmul_bw` and `unary_backward.cpp:388`) would make the reference carry its weight.

Aswinmcw (approved):

ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
VirdhatchaniKN: Fair point — the pointer was leading nowhere. Written up on #53874 now: https://github.com/tenstorrent/tt-metal/issues/53874#issuecomment-5395843903

It covers the passthrough rationale and that the same shape is in `addcmul_bw` (`ternary_backward.cpp:26`) and `frac_bw` (`unary_backward.cpp:388`), with no eltwise backward op calling `to_memory_config` at all.

It also records that item 11 itself is a no-op: the outer multiply already inherits `output_mem_config` from its first input, and reverting the change gives identical probe output (`L1 -> [DRAM, L1, L1]`, `DRAM -> [DRAM, DRAM, DRAM]`) with 12/12 passing either way. So this PR is a consistency fix rather than a memory-config bug fix, and #53874 could use a retitle.

ttnn/cpp/ttnn/operations/eltwise/ternary_backward/ternary_backward.cpp
```diff
@@ -43,11 +43,17 @@ std::vector<Tensor> addcdiv_bw(
...
     grad_tensor.emplace_back(grad);
     float t_inf = std::numeric_limits<float>::infinity();
     float t_nan = std::nanf("");
     Tensor grad_a = ttnn::multiply(
-        ttnn::multiply(grad, value, std::nullopt, output_mem_config), ttnn::reciprocal(tensor2, output_mem_config));
+        ttnn::multiply(grad, value, std::nullopt, output_mem_config),
+        ttnn::reciprocal(tensor2, output_mem_config),
+        std::nullopt,
+        output_mem_config);
```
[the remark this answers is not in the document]
VirdhatchaniKN: My branch : 
Screenshot 2026-08-25 at 9 13 23 AM
Verified the changes made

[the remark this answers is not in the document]
VirdhatchaniKN: not adding tests, hence resolving comment

PR #53900, opened 2026-08-21
#51433: Consolidate specless sharded-output shard-spec synthesis into a shared populated-grid helper

### LINKED ISSUE #51433: Consolidate specless-sharded-output shard-spec synthesis: shrink CoreRangeSet across TM/eltwise helpers (follow-up to #50684)
### Problem

The over-declaration pattern fixed for `generate_transpose_shard_spec` (issue #50684, PR #51156) still exists in three sibling helpers that synthesise a specless sharded-output `ShardSpec` by unconditionally handing back the full compute grid:

- `generate_repeat_shard_spec` — [`ttnn/cpp/ttnn/operations/data_movement/repeat/device/repeat_utils.cpp:251`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/repeat/device/repeat_utils.cpp#L251)
- `generate_indexed_fill_shard_spec` — [`ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_utils.cpp:284`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_utils.cpp#L284)
- `generate_unary_shard_spec` — [`ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_utils.cpp:222`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_utils.cpp#L222)

Each of these returns `ShardSpec(all_cores, shard_shape, orientation)` regardless of how many shards the tensor actually populates, so tiny tensors still claim the full 64-core WH grid with padded zero-shards on trailing cores — wasting L1 and reporting a misleading `shard_spec.grid.num_cores()`.

Fold's `synthesize_fold_output_shard_spec` (#50385) and the `generate_transpose_shard_spec` fix in #51156 already implement the shrink discipline; those two implementations are near-duplicates.

### Fix plan

1. Introduce a shared helper `synthesize_output_shard_spec(compute_grid, padded_out_shape, memory_layout, orientation_hint)` in a common data-movement / TM location, mirroring the logic from `generate_transpose_shard_spec` post-#51156:
   - `HEIGHT_SHARDED` / `WIDTH_SHARDED` → `num_cores_to_corerangeset(n_used, grid, row_wise)`.
   - `BLOCK_SHARDED` → rectangular `CoreRange(0,0)..(nx-1, ny-1)` with orientation-aware `(nx, ny) = (n_w, n_h)` for `ROW_MAJOR` / `(n_h, n_w)` for `COL_MAJOR`.
   - Clamp `n_used ∈ [1, num_cores]` for robustness.
2. Migrate the four call sites in this order (each PR must include unit tests asserting the exact populated `CoreRangeSet` on top of numerical correctness, analogous to `test_transpose_specless_sharded_output_grid_shrinks_*` in #51156):
   1. `generate_transpose_shard_spec` (DM + quasar twin) → call the shared helper instead of the inline logic already added in #51156.
   2. `generate_repeat_shard_spec`.
   3. `generate_indexed_fill_shard_spec`.
   4. `generate_unary_shard_spec`.
3. Simplify `synthesize_fold_output_shard_spec` (#50385) to delegate to the shared helper once behaviour matches; keep any fold-specific pre/post steps outside the helper.
4. Audit downstream callers that iterate over `shard_spec.grid` for hard-coded 64-core assumptions before each migration lands.

### Notes

- Behaviour change per migration: tiny tensors will start reporting shrunken grids instead of the full compute grid. Same caveat as #50684 / #51156 — no in-tree callers depend on the old behaviour, but each migration should re-verify on CI.
- Related: #50684 (parent), #51156 (transpose fix that established the pattern), #50385 (fold's tighter helper).

mouliraj-mcw:
### Ticket
Link to Github Issue #51433

### Problem
Five near-duplicate `generate_*_shard_spec` helpers (`transpose`, `fold`, `repeat`, `indexed_fill`, `unary`) synthesize specless sharded outputs; only transpose shrinks the grid — the other four declare `CoreRangeSet` = full compute grid, wasting L1 and misreporting `.grid.num_cores()`.

### Fix
Extract the transpose logic into `common::synthesize_output_shard_spec` — populated-shard `num_cores_to_corerangeset` for H/W, orientation-aware `(h_div, w_div)` for BLOCK — and route `repeat/fold/indexed_fill/unary` and `experimental/quasar/transpose` through it. Downstream audit (issue step 4): all five kernels drive per-core work from `shard_spec.grid`/`worker_grid` and early-return on zero-work — unary `noop=true` (`unary_program_factory.cpp:291`), indexed_fill `return` on `shard_ppb==0` (`indexed_fill_reader.cpp:83`), repeat `nop=1` (`repeat_higher_dim_rm_sharded.cpp:31`); fold covered by H/W/B shrink test; transpose is the pre-existing reference impl.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mouliraj/synthesize_output_shard_spec)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mouliraj/synthesize_output_shard_spec)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
+++ b/tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
@@ -816,6 +816,9 @@ def test_fold_from_torch_rejects_specless_sharded_input(device, expect_error):
             (1, 8, 8, 64), ttnn.TensorMemoryLayout.HEIGHT_SHARDED, id="H_shrinks"
         ),  # post-fold rows=16 → 16 cores (not 64).
         pytest.param(
+            (1, 4, 4, 8), ttnn.TensorMemoryLayout.WIDTH_SHARDED, id="W_shrinks"
+        ),  # post-fold cols=32 → 32 cores (not 64).
+        pytest.param(
             (1, 4, 4, 64), ttnn.TensorMemoryLayout.BLOCK_SHARDED, id="B_shrinks"
         ),  # post-fold 4×256 → 4x8=32 cores (not 64).
     ],
@@ -834,6 +837,66 @@ def test_fold_specless_sharded_override_grid_shrinks_when_rows_lt_max_cores(shap
     assert (
         mc.shard_spec.grid.num_cores() < max_cores
     ), f"Synthesised grid over-declared cores: {mc.shard_spec.grid.num_cores()} == max {max_cores}"
+    ref = _fold_golden(x, 2, 2)
+    got = ttnn.to_torch(result.cpu().to(ttnn.ROW_MAJOR_LAYOUT))
+    assert_with_pcc(ref.float(), got.float(), 0.9999)
+
+
+@pytest.mark.parametrize(
+    "in_shard_factory, override_layout",
+    [
+        pytest.param(_height_shard_rm, ttnn.TensorMemoryLayout.WIDTH_SHARDED, id="H_col_major_in_W_override"),
+        pytest.param(_width_shard_rm, ttnn.TensorMemoryLayout.HEIGHT_SHARDED, id="W_col_major_in_H_override"),
+    ],
+)
+def test_fold_specless_cross_layout_override_col_major_input_enum_col_wise(in_shard_factory, override_layout, device):
+    """COL_MAJOR sharded input + orthogonal-layout override → helper enumerates col-wise CoreRangeSet (locks in post-consolidation change)."""
+    grid = device.compute_with_storage_grid_size()
+    if grid.x < 4 or grid.y < 4:
+        pytest.skip("Need at least 4x4 grid to distinguish row-wise vs col-wise enumeration")
+    shape = (1, 4, 4, 8)
+    in_mc = in_shard_factory(shape, device, orientation=ttnn.ShardOrientation.COL_MAJOR)
+    out_mc = ttnn.MemoryConfig(override_layout, ttnn.BufferType.L1)
+    torch.manual_seed(0)
+    x = torch.rand(shape, dtype=torch.bfloat16)
+    ttnn_in = ttnn.from_torch(x, layout=ttnn.ROW_MAJOR_LAYOUT, dtype=ttnn.bfloat16, device=device, memory_config=in_mc)
+    result = ttnn.fold(ttnn_in, 2, 2, override_memory_config=out_mc)
+    mc = result.memory_config()
+    assert mc.shard_spec is not None
+    assert mc.shard_spec.orientation == ttnn.ShardOrientation.COL_MAJOR
+    n_cores = mc.shard_spec.grid.num_cores()
+    expected_crs = ttnn.num_cores_to_corerangeset(n_cores, grid, False)
+    assert (
+        mc.shard_spec.grid == expected_crs
+    ), f"COL_MAJOR input must yield col-wise CoreRangeSet; got {mc.shard_spec.grid}"
+    ref = _fold_golden(x, 2, 2)
+    got = ttnn.to_torch(result.cpu().to(ttnn.ROW_MAJOR_LAYOUT))
+    assert torch.equal(ref, got), f"fold is a lossless bfloat16 reshape; expected bit-equal, got mismatch"
+
+
+def test_fold_specless_block_override_col_major_input_swaps_divisors(device):
+    """COL_MAJOR H input + BLOCK override on asymmetric grid → shard shape follows h_div=grid.x, w_div=grid.y swap."""
+    grid = device.compute_with_storage_grid_size()
+    if grid.x == grid.y:
+        pytest.skip("Divisor swap invisible on square grid; asserts land on asymmetric (BH) grids")
+    shape = (1, 4, 4, 64)
+    in_mc = _height_shard_rm(shape, device, orientation=ttnn.ShardOrientation.COL_MAJOR)
+    out_mc = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.BLOCK_SHARDED, ttnn.BufferType.L1)
+    torch.manual_seed(0)
+    x = torch.rand(shape, dtype=torch.bfloat16)
+    ttnn_in = ttnn.from_torch(x, layout=ttnn.ROW_MAJOR_LAYOUT, dtype=ttnn.bfloat16, device=device, memory_config=in_mc)
+    result = ttnn.fold(ttnn_in, 2, 2, override_memory_config=out_mc)
+    mc = result.memory_config()
+    assert mc.shard_spec is not None
+    assert mc.shard_spec.orientation == ttnn.ShardOrientation.COL_MAJOR
+    tensor_h, tensor_w = 4, 256
+    expected_h = (tensor_h + grid.x - 1) // grid.x
+    expected_w = (tensor_w + grid.y - 1) // grid.y
+    got_h, got_w = mc.shard_spec.shape[0], mc.shard_spec.shape[1]
+    assert (got_h, got_w) == (expected_h, expected_w), (
+        f"COL_MAJOR BLOCK divisor swap: expected shard ({expected_h}, {expected_w}) "
+        f"[h_div=grid.x={grid.x}, w_div=grid.y={grid.y}]; got ({got_h}, {got_w})"
+    )
     ref = _fold_golden(x, 2, 2)
     got = ttnn.to_torch(result.cpu().to(ttnn.ROW_MAJOR_LAYOUT))
     assert_with_pcc(ref.float(), got.float(), 0.9999)
--- a/tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
@@ -285,3 +285,64 @@ def test_indexed_fill_dim_out_of_bounds(device, expect_error):
 
     with expect_error(RuntimeError, "is out of bounds for rank"):
         ttnn.indexed_fill(batch_id_ttnn, input_tensor_a, input_tensor_b, dim=-5)  # -5 < -rank=-4
+
+
+# Specless sharded output must shrink CoreRangeSet to populated shard count.
+
+
+@pytest.mark.parametrize(
+    "shape, out_layout, expected_cores",
+    [
+        # HEIGHT_SHARDED: (B=2, ..., H=32, W=64) tiled → tensor_h=64, shard_h=32 → 2 populated cores.
+        pytest.param((2, 1, 32, 64), ttnn.TensorMemoryLayout.HEIGHT_SHARDED, 2, id="H_shrinks"),
+        # WIDTH_SHARDED: W=64 tiled → shard_w=32 → 2 populated cores.
+        pytest.param((2, 1, 32, 64), ttnn.TensorMemoryLayout.WIDTH_SHARDED, 2, id="W_shrinks"),
+        # BLOCK_SHARDED: tensor_h=64, tensor_w=64 on 8x8 grid → 2x2 rectangle = 4 populated cores.
+        pytest.param((2, 1, 32, 64), ttnn.TensorMemoryLayout.BLOCK_SHARDED, 4, id="B_shrinks"),
+    ],
+)
+def test_indexed_fill_specless_sharded_output_grid_shrinks(device, shape, out_layout, expected_cores):
+    """No-spec sharded output MC must synthesise a CoreRangeSet sized to populated shards, not full compute grid."""
+    compute_grid = device.compute_with_storage_grid_size()
+    if compute_grid.x * compute_grid.y <= expected_cores:
+        pytest.skip(f"Device grid too small to observe shrink (need > {expected_cores} cores)")
+    if out_layout == ttnn.TensorMemoryLayout.BLOCK_SHARDED and (compute_grid.x < 2 or compute_grid.y < 2):
+        pytest.skip("BLOCK case needs at least a 2x2 compute grid")
+
+    B = shape[0]
+    b = 1
+    batch_id = torch.randint(0, B, (1, 1, 1, b))
+    batch_id_ttnn = ttnn.Tensor(batch_id, ttnn.uint32).to(
+        device, ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.L1)
+    )
+
+    torch_a = torch.rand(shape, dtype=torch.bfloat16)
+    torch_b = torch.rand((b, *shape[1:]), dtype=torch.bfloat16)
+    input_a = ttnn.from_torch(
+        torch_a,
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.L1),
+    )
+    input_b = ttnn.from_torch(
+        torch_b,
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.L1),
+    )
+
+    out_mc = ttnn.MemoryConfig(out_layout, ttnn.BufferType.L1)
+    output = ttnn.indexed_fill(batch_id_ttnn, input_a, input_b, memory_config=out_mc)
+    grid = output.memory_config().shard_spec.grid
+    assert grid.num_cores() == expected_cores, f"Expected {expected_cores} populated cores, got {grid.num_cores()}"
+    if out_layout == ttnn.TensorMemoryLayout.BLOCK_SHARDED:
+        # BLOCK: rectangular (0,0)-(nx-1, ny-1); shape=(2,1,32,64) on tile-aligned 8x8 → 2×2.
+        expected_crs = ttnn.CoreRangeSet([ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(1, 1))])
+    else:
+        expected_crs = ttnn.num_cores_to_corerangeset(expected_cores, compute_grid, True)
+    assert grid == expected_crs, f"Expected CoreRangeSet {expected_crs}, got {grid}"
+
+    golden = golden_indexed_fill(torch_a, torch_b, batch_id, dim=0)
+    assert_with_pcc(golden, ttnn.to_torch(output), 0.9999)
--- a/ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
@@ -1,0 +1,137 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#include "synthesize_output_shard_spec.hpp"
+
+#include <algorithm>
+
+#include <tt_stl/assert.hpp>
+#include <tt-metalium/constants.hpp>
+#include <tt-metalium/math.hpp>
+#include <tt-metalium/work_split.hpp>
+
+namespace ttnn::operations::data_movement::common {
+
+using tt::tt_metal::CoreCoord;
+using tt::tt_metal::CoreRange;
+using tt::tt_metal::CoreRangeSet;
+using tt::tt_metal::ShardOrientation;
+using tt::tt_metal::ShardSpec;
+using tt::tt_metal::TensorMemoryLayout;
+
+namespace {
+
+ShardOrientation resolve_orientation(const SynthesizeOutputShardSpecOpts& opts) {
+    if (opts.orientation_hint.has_value()) {
+        return *opts.orientation_hint;
+    }
+    if (opts.input_orientation.has_value()) {
+        return *opts.input_orientation;
+    }
+    return ShardOrientation::ROW_MAJOR;
+}
+
+}  // namespace
+
+ShardSpec synthesize_output_shard_spec(
+    const CoreCoord& compute_grid_size,
+    uint64_t tensor_height,
+    uint64_t tensor_width,
+    TensorMemoryLayout memory_layout,
+    const SynthesizeOutputShardSpecOpts& opts) {
+    // Sharded-only contract — callers filter INTERLEAVED at their wrapper (see repeat_utils.cpp:186-190).
+    TT_FATAL(
+        memory_layout == TensorMemoryLayout::HEIGHT_SHARDED || memory_layout == TensorMemoryLayout::WIDTH_SHARDED ||
+            memory_layout == TensorMemoryLayout::BLOCK_SHARDED,
+        "{}: unsupported memory_layout; only HEIGHT/WIDTH/BLOCK sharded.",
+        opts.caller_tag);
+    const CoreRangeSet all_cores(CoreRange({0, 0}, {compute_grid_size.x - 1, compute_grid_size.y - 1}));
+    const uint32_t num_cores = all_cores.num_cores();
+    TT_FATAL(num_cores > 0, "{}: empty compute grid.", opts.caller_tag);
+    // Guards `div_up(_, shard_shape[i])` below; repeat's soft-reject path never reaches here (see
+    // repeat_utils.cpp:202).
+    TT_FATAL(
+        tensor_height > 0 && tensor_width > 0,
+        "{}: tensor dims must be > 0; got ({}, {}).",
+        opts.caller_tag,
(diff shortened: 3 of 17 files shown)


ttnn/cpp/ttnn/operations/data_movement/fold/device/fold_device_op.cpp
```diff
@@ -18,44 +19,17 @@ bool is_fast_path_input(const Tensor& t) {
...
-        }
-    }
-    return tt::tt_metal::ShardSpec(cores, shape, orientation);
+    return common::synthesize_output_shard_spec(
+        device->compute_with_storage_grid_size(),
+        static_cast<uint64_t>(rows),
+        static_cast<uint64_t>(cols),
+        layout,
+        {.is_tile = false,
+         .input_orientation = input_tensor.shard_spec().has_value()
```
Aswinmcw, line 30: Passing `input_orientation` here changes fold's H/W core enumeration, not just the reported orientation. The old code called `num_cores_to_corerangeset(..., grid, /*row_wise=*/true)` unconditionally; the helper derives `row_wise` from the resolved orientation, so a COL_MAJOR-sharded input now gets a column-wise CoreRangeSet (e.g. `(0,0)-(0,3)` instead of `(0,0)-(3,0)` for 4 shards). BLOCK is unaffected — the divisor swap already matched.

That's arguably the more consistent behaviour, but it isn't in the description and nothing covers it. Either pass `orientation_hint = ROW_MAJOR` like unary/indexed_fill do to keep fold byte-identical, or keep it and add a fold COL_MAJOR H/W test.

tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
```diff
@@ -285,3 +285,56 @@ def test_indexed_fill_dim_out_of_bounds(device, expect_error):
...
+# Specless sharded output must shrink CoreRangeSet to populated shard count.
+
+
+@pytest.mark.parametrize(
+    "shape, out_layout, expected_cores",
+    [
+        # HEIGHT_SHARDED: (B=2, ..., H=32, W=64) tiled → tensor_h=64, shard_h=32 → 2 populated cores.
+        pytest.param((2, 1, 32, 64), ttnn.TensorMemoryLayout.HEIGHT_SHARDED, 2, id="H_shrinks"),
+        # WIDTH_SHARDED: W=64 tiled → shard_w=32 → 2 populated cores.
+        pytest.param((2, 1, 32, 64), ttnn.TensorMemoryLayout.WIDTH_SHARDED, 2, id="W_shrinks"),
```
Aswinmcw, line 299: No BLOCK_SHARDED case here, unlike the repeat and unary tests — and BLOCK is where indexed_fill changes most (full grid -> `(0,0)-(nx-1,ny-1)` rectangle). Worth adding the third param to match the other two files.

Same gap for fold: it's migrated in this PR with no new test, while the issue asks for a populated-CoreRangeSet assertion per migration.

ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_utils.cpp
```diff
@@ -191,35 +192,15 @@ CoreRangeSet get_worker_grid(
 
 tt::tt_metal::ShardSpec generate_shard_spec_all_cores(
```
Aswinmcw: `generate_shard_spec_all_cores` no longer returns all cores — the name now says the opposite of what it does. Same for `indexed_fill::generate_shard_spec_all_cores`. Both are internal, so a rename to `generate_output_shard_spec` is cheap here.

ttnn/cpp/ttnn/operations/data_movement/fold/device/fold_device_op.cpp
mouliraj-mcw: Kept `input_orientation` and locked the col-wise enumeration in `test_fold_specless_cross_layout_override_col_major_input_enum_col_wise` (`tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py:845`); intent flagged at `fold_device_op.cpp:22`.

tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
mouliraj-mcw: Added `B_shrinks` for indexed_fill at `test_indexed_fill.py:301` (asserts `(0,0)-(1,1)` CRS at line 342), and for fold at `test_universal_input_tm_fold.py:822` (H/W/B cases inside `test_fold_specless_sharded_override_grid_shrinks_when_rows_lt_max_cores:826`).

ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_utils.cpp
mouliraj-mcw: Renamed to `generate_output_shard_spec` in both — `unary_utils.hpp:41` / `.cpp:193` and `indexed_fill_utils.hpp:62` / `.cpp:245`; call sites at `unary_device_operation.cpp:140` and `indexed_fill_utils.cpp:273`.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,124 @@
...
+        const uint32_t w_div = row_wise ? grid_size.x : grid_size.y;
+        const auto height_padded = tt::round_up(tensor_height, static_cast<uint64_t>(h_div) * h_align);
+        const auto shard_height = tt::round_up(tt::div_up(height_padded, static_cast<uint64_t>(h_div)), h_align);
+        const auto shard_width = tt::round_up(tt::div_up(tensor_width, static_cast<uint64_t>(w_div)), w_align);
+        shard_shape = {static_cast<uint32_t>(shard_height), static_cast<uint32_t>(shard_width)};
+    }
+
+    CoreRangeSet used_cores;
+    if (memory_layout == TensorMemoryLayout::HEIGHT_SHARDED) {
+        uint32_t n_used = static_cast<uint32_t>(tt::div_up(tensor_height, static_cast<uint64_t>(shard_shape[0])));
```
Aswinmcw, line 85: The zero-dimension guard should move in here rather than staying at `repeat`'s call site. If `tensor_height` is 0, `shard_shape[0]` computes to 0 and this becomes `div_up(0, 0)` — `tt::div_up` documents that as undefined behavior and only guards with a plain `assert`, so a release build divides by zero. `repeat_utils.cpp` still early-returns on `tensor_height == 0 || tensor_width == 0`, but the other four call sites don't, and `unary`/`indexed_fill` never divided by the shard shape before this PR, so they newly acquire the path. A guard at the top of the helper (early-return or `TT_FATAL`) covers all five and lets `repeat`'s check stay as-is or go away.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,124 @@
...
+        const auto shard_height = tt::round_up(tt::div_up(height_padded, static_cast<uint64_t>(num_cores)), h_align);
+        shard_shape = {static_cast<uint32_t>(shard_height), static_cast<uint32_t>(tensor_width)};
+    } else if (memory_layout == TensorMemoryLayout::WIDTH_SHARDED) {
+        const auto shard_width = tt::round_up(tt::div_up(tensor_width, static_cast<uint64_t>(num_cores)), w_align);
+        shard_shape = {static_cast<uint32_t>(tensor_height), static_cast<uint32_t>(shard_width)};
+    } else {
+        // BLOCK: COL_MAJOR swaps h↔grid.x, w↔grid.y (matches conv2d_utils::determine_parallel_config).
+        CoreCoord grid_size = all_cores.bounding_box().grid_size();
+        const uint32_t h_div = row_wise ? grid_size.y : grid_size.x;
+        const uint32_t w_div = row_wise ? grid_size.x : grid_size.y;
```
Aswinmcw: This orientation-dependent divisor swap has no test coverage — every BLOCK case in the new tests runs with the default ROW_MAJOR, and the new COL_MAJOR fold case is width-sharded. It's the subtlest line in the helper and the asymmetric-grid overshoot in #51433 is exactly what it changes, so a BLOCK + COL_MAJOR case (fold with a col-major input, or repeat with `orientation_hint`) checking the exact shard shape on a non-square grid would be worth adding.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,124 @@
...
+ShardSpec synthesize_output_shard_spec(
+    const CoreCoord& compute_grid_size,
+    uint64_t tensor_height,
+    uint64_t tensor_width,
+    TensorMemoryLayout memory_layout,
+    const SynthesizeOutputShardSpecOpts& opts) {
+    // Non-{H,W,BLOCK} layouts fall through to BLOCK path — matches pre-consolidation transpose behaviour.
+    const CoreRangeSet all_cores(CoreRange({0, 0}, {compute_grid_size.x - 1, compute_grid_size.y - 1}));
+    const uint32_t num_cores = all_cores.num_cores();
+    TT_FATAL(num_cores > 0, "synthesize_output_shard_spec: empty compute grid.");
```
Aswinmcw: Nit: the BLOCK `TT_FATAL` below used to name the op ("Transpose: BLOCK shard-grid..."); now five ops share one message. Threading a caller tag through `SynthesizeOutputShardSpecOpts` would keep the failure attributable.

Aswinmcw (commented):
Previous round's three points are addressed. One question left over from #51433: step 4 of the fix plan asks to "audit downstream callers that iterate over `shard_spec.grid` for hard-coded 64-core assumptions before each migration lands" — could you note in the description what that audit turned up? The comment deleted from `indexed_fill_utils.cpp` said the kernels tolerate padding-only cores via early-return guards, which reads like they were written expecting the full grid, and this PR now hands several of them a shrunk grid.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Done — added `TT_FATAL(tensor_height > 0 && tensor_width > 0, ...)` at the top of `synthesize_output_shard_spec` (before the `div_up(_, shard_shape[i])` chain), so every call site inherits the guard; `repeat`s call-site check is now redundant with this one.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Done — added `test_fold_specless_block_override_col_major_input_swaps_divisors` in `tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py`; asserts exact shard shape after `h_div=grid.x, w_div=grid.y` swap on asymmetric grid (skips on square, so the check lands on BH `bh_p100` 11×10).

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Done — threaded `std::string_view caller_tag` through `SynthesizeOutputShardSpecOpts`; all 6 call sites pass op name ("Fold", "Repeat", "IndexedFill", "Unary", "Transpose", "Transpose (Quasar)") and every `TT_FATAL` in `synthesize_output_shard_spec` now prefixes with `"{}: ..."`, so BLOCK-grid overflow, empty-grid, and zero-dim messages name the caller again.

mateusznowakTT (approved):
Approved assuming there are no related failures in test pipelines.

amorrisonTT (approved):

bbradelTT (approved):

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,133 @@
...
+using tt::tt_metal::CoreRangeSet;
+using tt::tt_metal::ShardOrientation;
+using tt::tt_metal::ShardSpec;
+using tt::tt_metal::TensorMemoryLayout;
+
+namespace {
+
+ShardOrientation resolve_orientation(const SynthesizeOutputShardSpecOpts& opts) {
+    if (opts.orientation_hint.has_value()) {
+        return *opts.orientation_hint;
```
dstoiljkovicTT, line 26: What is the difference between the oriunetation_hint and input_orientation and should those two be two separate fields of the Opts structure?

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,133 @@
...
+    // Non-{H,W,BLOCK} layouts fall through to BLOCK path — matches pre-consolidation transpose behaviour.
+    const CoreRangeSet all_cores(CoreRange({0, 0}, {compute_grid_size.x - 1, compute_grid_size.y - 1}));
+    const uint32_t num_cores = all_cores.num_cores();
+    TT_FATAL(num_cores > 0, "{}: empty compute grid.", opts.caller_tag);
+    // Guards `div_up(_, shard_shape[i])` below; repeat's soft-reject path never reaches here (see
+    // repeat_utils.cpp:202).
+    TT_FATAL(
+        tensor_height > 0 && tensor_width > 0,
+        "{}: tensor dims must be > 0; got ({}, {}).",
+        opts.caller_tag,
```
dstoiljkovicTT, line 56: Do we need this caller tag, isn't backtrace enough? Seems like we are using these for tt_fatals only

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
```diff
@@ -0,0 +1,133 @@
...
+    if (memory_layout == TensorMemoryLayout::HEIGHT_SHARDED) {
+        const auto height_padded = tt::round_up(tensor_height, static_cast<uint64_t>(num_cores) * h_align);
+        const auto shard_height = tt::round_up(tt::div_up(height_padded, static_cast<uint64_t>(num_cores)), h_align);
+        shard_shape = {static_cast<uint32_t>(shard_height), static_cast<uint32_t>(tensor_width)};
+    } else if (memory_layout == TensorMemoryLayout::WIDTH_SHARDED) {
+        const auto shard_width = tt::round_up(tt::div_up(tensor_width, static_cast<uint64_t>(num_cores)), w_align);
+        shard_shape = {static_cast<uint32_t>(tensor_height), static_cast<uint32_t>(shard_width)};
+    } else {
+        // BLOCK: COL_MAJOR swaps h↔grid.x, w↔grid.y (matches conv2d_utils::determine_parallel_config).
+        CoreCoord grid_size = all_cores.bounding_box().grid_size();
```
dstoiljkovicTT: What are the cases where grid_size!=compute_grid_size, I am trying to understand why are we getting this by fetching the bounding box if we created the grid a couple of lines above

tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
```diff
@@ -839,6 +842,59 @@ def test_fold_specless_sharded_override_grid_shrinks_when_rows_lt_max_cores(shap
     assert_with_pcc(ref.float(), got.float(), 0.9999)
 
 
+def test_fold_specless_cross_layout_override_col_major_input_enum_col_wise(device):
```
dstoiljkovicTT: If I understand the code correctly these tests are added because fold op was previously hardcoding row major regardless of the shard orientation? So now that the helper checks the orientation you added the test catching and showcasing the fix? If this is the case is there a reason we are testing only width sharded case and not the height sharded one

ttnn/cpp/ttnn/operations/data_movement/repeat/device/repeat_utils.cpp
```diff
@@ -183,72 +184,51 @@ std::optional<ShardSpec> generate_repeat_shard_spec(
...
-        }
-        auto height_padded =
-            tt::round_up(tensor_height, static_cast<uint64_t>(grid_size.y) * tt::constants::TILE_HEIGHT);
-        auto shard_height =
-            tt::round_up(tt::div_up(height_padded, static_cast<uint64_t>(grid_size.y)), tt::constants::TILE_HEIGHT);
-        auto shard_width =
-            tt::round_up(tt::div_up(tensor_width, static_cast<uint64_t>(grid_size.x)), tt::constants::TILE_WIDTH);
-        shard_shape = {static_cast<uint32_t>(shard_height), static_cast<uint32_t>(shard_width)};
-    } else {
-        return std::nullopt;  // INTERLEAVED / unsupported — caller handles.
```
dstoiljkovicTT, line 226: Seems like we are missing this else for the interleaved cases in the helper, should we add it back just in case, maybe add some tt_fatal

dstoiljkovicTT (commented):
Are these all the ops duplicating this code? I haven't dug much deeper but I saw ternary and binary_ng also having some generate_shard_spec_all_cores function themselves, is that similar to this? Can we replace those as well?

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Two-tier caller intent — collapsing pushes the fallback into all 6 call sites.

- `orientation_hint`: hard override. Unary/IndexedFill pin `ROW_MAJOR` for pre-consolidation compat; Transpose/Repeat forward a caller-facing param.
- `input_orientation`: inherit from input `shard_spec` when the caller has no opinion. Fold uses this to fix the pre-consolidation `ROW_MAJOR` hard-code.

Fallback in `resolve_orientation` (`synthesize_output_shard_spec.cpp:24-32`): `hint` → `input_orientation` → `ROW_MAJOR`.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Kept because Aswinmcw asked for it in [#discussion_r3874461811](https://github.com/tenstorrent/tt-metal/pull/53900#discussion_r3874461811) — the BLOCK `TT_FATAL` used to say `"Transpose: BLOCK shard-grid…"`; post-consolidation five ops shared one nameless message. `TT_FATAL` surfaces to users through python bindings and `tt-triage`; no C++ backtrace is printed on that path.

ttnn/cpp/ttnn/operations/data_movement/common/synthesize_output_shard_spec.cpp
mouliraj-mcw: Fixed — `all_cores` is built from `compute_grid_size` at `synthesize_output_shard_spec.cpp:49`, so `bounding_box().grid_size()` was a no-op round-trip. BLOCK divisors now index `compute_grid_size` directly (`synthesize_output_shard_spec.cpp:76-77`).

tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
mouliraj-mcw: Correct on both counts — the tests catch the pre-consolidation bug where fold hard-coded `row_wise=true` in `num_cores_to_corerangeset`. Added `test_fold_specless_cross_layout_override_col_major_input_enum_col_wise_H_output` (`test_universal_input_tm_fold.py:870-892`): W col_major input → H override, asserts col-wise `CoreRangeSet` + `torch.equal` vs golden (fold is a lossless bfloat16 reshape, no arithmetic). Pairs with the W-output test above.

ttnn/cpp/ttnn/operations/data_movement/repeat/device/repeat_utils.cpp
mouliraj-mcw: Added — `TT_FATAL` at `synthesize_output_shard_spec.cpp:43-49` rejects any non-{HEIGHT/WIDTH/BLOCK} layout. Repeat's wrapper still filters INTERLEAVED at `repeat_utils.cpp:186-190`, so this is a belt-and-braces guard for future callers that skip the wrapper filter.

tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
```diff
@@ -839,6 +842,84 @@ def test_fold_specless_sharded_override_grid_shrinks_when_rows_lt_max_cores(shap
...
```
dstoiljkovicTT: Can we please parametrize one test instead of having duplicated block of code for the height and width case

dstoiljkovicTT:
> Are these all the ops duplicating this code? I haven't dug much deeper but I saw ternary and binary_ng also having some generate_shard_spec_all_cores function themselves, is that similar to this? Can we replace those as well?

Adding a tag to this one @mouliraj-mcw, since it seems like it got overseen in the bunch of other comments

tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
mouliraj-mcw: Done — merged into one `@pytest.mark.parametrize` at `test_universal_input_tm_fold.py:846`; the two cases are now `[H_col_major_in_W_override]` (H input → W override) and `[W_col_major_in_H_override]` (W input → H override), sharing the same grid-skip / COL_MAJOR / col-wise CRS / `torch.equal` body. Net 16-line drop on the file.

mouliraj-mcw:
Confirmed — `ternary_device_operation.cpp:143`, `binary_ng_device_operation.cpp:104`, `experimental/quasar/binary_ng/device/binary_ng_device_operation.cpp:76` all carry the same full-grid pattern, and their factories already noop-tolerate a shrunk grid via `grid_to_cores_with_noop`.
Filing a follow-up against #51433 to keep this PR's approvals intact — will cross-link here.

Aswinmcw (approved):
Verified at 41ea855: zero-dim `TT_FATAL` now guards the `div_up` chain inside the helper, the sharded-only layout guard is in, `caller_tag` prefixes every fatal, the `bounding_box().grid_size()` round-trip is gone (BLOCK divisors index `compute_grid_size` directly), and the col-major enumeration / BLOCK divisor-swap paths have exact-CoreRangeSet tests. Quasar twin updated in the same change. Approving — only the pre-commit formatting suggestion is outstanding.

dstoiljkovicTT (approved):

tests/ttnn/nightly/unit_tests/operations/data_movement/test_universal_input_tm_fold.py
```diff
@@ -839,6 +842,68 @@ def test_fold_specless_sharded_override_grid_shrinks_when_rows_lt_max_cores(shap
...
+@pytest.mark.parametrize(
+    "in_shard_factory, override_layout",
+    [
+        pytest.param(_height_shard_rm, ttnn.TensorMemoryLayout.WIDTH_SHARDED, id="H_col_major_in_W_override"),
+        pytest.param(_width_shard_rm, ttnn.TensorMemoryLayout.HEIGHT_SHARDED, id="W_col_major_in_H_override"),
+    ],
+)
+def test_fold_specless_cross_layout_override_col_major_input_enum_col_wise(
+    in_shard_factory, override_layout, device
+):
```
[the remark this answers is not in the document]
mouliraj-mcw: Applied — collapsed the signature onto a single line at `test_universal_input_tm_fold.py:854`, `black --check` clean, both parametrized cases still pass.

PR #53905, opened 2026-08-21
[Bug fix] ttnn.hypot returns inf or 0 where hypot is an ordinary finite number

### LINKED ISSUE #51861: ttnn.hypot returns inf above |x| ~= 1.8e19 and 0 below ~1.1e-19, where the bfloat16 golden is finite
# ttnn.hypot returns inf above |x| ~= 1.8e19 and 0 below ~1.1e-19, where the bfloat16 golden is finite

`ttnn.hypot(1e25, 0)` returns `inf`. For the bfloat16 input below, the registered golden is the
finite value `9.9736380118206907e24`.

`hypot(x, 0)` is `|x|` by definition. There is no rounding question, no tolerance to argue about, and
no accuracy trade to weigh — and the registered golden agrees:

```python
torch.hypot(torch.tensor([1e25], dtype=torch.bfloat16),
            torch.tensor([0.0],  dtype=torch.bfloat16))   # 9.9736380118206907e+24, finite
```

The op is composed so that it computes `sqrt(a**2 + b**2)` literally
(`binary_ng_utils.cpp:391-396`):

```c
case BinaryOpType::HYPOT:
    process_lhs = unary::UnaryOpType::SQUARE;
    process_rhs = unary::UnaryOpType::SQUARE;
    binary_op = EnumT::ADD;
    postprocess = unary::UnaryOpType::SQRT;
    break;
```

`a**2` overflows or enters Tenstorrent's DAZ/FTZ region long before `hypot(a, b)` does. Avoiding
those intermediates is why `hypot` exists as a named function rather than an expression callers
write out.

| input (bfloat16) | `ttnn.hypot` | registered golden | golden finite? |
|---|---|---|---|
| `(1e25, 0)` | `inf` | `9.9736380118206907e24` | yes |
| `(3e38, 0)` | `inf` | `3.0040552704739099e38` | yes |
| `(1e25, 1e25)` | `inf` | `1.4129320516745978e25` | yes |
| `(1e30, 1e30)` | `inf` | `1.416203404942475e30` | yes |
| `(1e20, 1e20)` | `inf` | `1.4123288431433875e20` | yes |
| `(1.3835e19, 1.3835e19)` | `inf` | `1.95996655783164e19` | yes — **and both operands are below `2^64`** |
| `(1.3835e19, 0)` | `1.3835058055282164e19` | `1.3835058055282164e19` | — (same operand, correct when paired with `0`) |
| `(1e-20, 1e-20)` | `0.0` | `1.4187801866509531e-20` | yes |
| `(3, 4)` | `5` | `5` | — |

Taking one operand at a time, the observed boundaries are exactly where squaring overflows or enters
the accelerator's FTZ region. Sweeping every binade with five mantissas each (1.00, 1.25, 1.50, 1.75, 1.99 — 20 of 25 sample points
are not powers of two) puts the boundaries at a single bfloat16 value:

| side | last value that works | first value that fails |
|---|---|---|
| overflow | `bfloat16(1.99 x 2^63)` = `1.9921875 x 2^63` = `1.8374686479671624e19` | **`2^64`** = `1.84467e19` (`inf`; golden `1.844674e19`) |
| underflow | `2^-63` = `1.08420e-19` | `bfloat16(1.99 x 2^-64)` = `1.9921875 x 2^-64` = `1.0799670077492329e-19` (`0`; golden `1.079967e-19`) |

For a single nonzero normal bfloat16 operand, the failing set is `|x| >= 2^64` or `0 < |x| < 2^-63`.

Two operands fail sooner than one, because `a**2 + b**2` reaches the overflow point before `a**2`
does. Sweeping equal-magnitude pairs the same way puts the onset at or below `1.5 x 2^63` — below
the single-operand threshold — and the contrast is direct: `hypot(1.5 x 2^63, 1.5 x 2^63)` returns
`inf` while the same operand paired with `0` returns the correct value. Three sampled pairs sit in
that gap, and the sweep brackets the true onset in `(1.25 x 2^63, 1.5 x 2^63]`. Identical results on
Wormhole and Blackhole; ordinary magnitudes (`3`, `5`, `100`, each paired with itself and with `0`)
pass in the same run as a control.

Those single-operand boundaries match the overflow and normal-range/FTZ boundaries predicted by
squaring. On the low side,
`sqrt(1.1754943508222875e-38)` is exactly `2^-63`: its square is the smallest normal bfloat16,
while the next sampled value squares to a bfloat16 subnormal. Tenstorrent's own tests state that
hardware DAZ/FTZ flushes a subnormal intermediate to zero
(`tests/ttnn/unit_tests/operations/eltwise/test_activation.py:135-141`). This does not challenge
that documented behavior: the composite unnecessarily materialises an FTZ intermediate although
the final `hypot` golden is a normal finite bfloat16 value. On the overflow side, the algebraic threshold,
`sqrt(3.3895313892515355e38) = 1.84107e19`, falls *between* the last working value and the first
failing one, because bfloat16 has few representable values at that magnitude. **`2^64` is the
measured boundary; `1.841e19` is only the algebraic one.**

**Reference values are the registered golden** (`ttnn/ttnn/operations/binary.py`,
`attach_golden_function`) — that is `torch.hypot`, evaluated with **both arguments at the operator's
own dtype (bfloat16)**, not on upcast tensors. The golden is finite at every input above; it does
not share the overflow.

The defining identity holds on the golden across the whole range, which is what makes the failure
unambiguous rather than a tolerance question:

| x | `torch.hypot(x, 0)` | `\|x\|` |
|---|---|---|
| `1e-25` | `1.001664023013432e-25` | equal |
| `1.0` | `1` | equal |
| `1e19` | `1.0016005571271983e19` | equal |
| `1e25` | `9.9736380118206907e24` | equal |
| `3e38` | `3.0040552704739099e38` | equal |

Five of the nine magnitudes sampled are shown; equality is exact at every one of the nine.

## This is a hazard the repository already names and avoids elsewhere

`ckernel_sfpu_trigonometry.h` states the same number and the same cause, identically on all three
architectures — `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_trigonometry.h:1057-1058`,
`tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_trigonometry.h:1021-1022`,
`tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_trigonometry.h:1006-1007`:

```
// acosh(x) == ln(2x) to <1 ulp; using log1p(2x - 1) also dodges the x^2 overflow
// that makes the classic form return +inf for x >= ~1.84e19.
```

`acosh` was restructured to avoid squaring. `hypot` — the one function whose entire
purpose is to avoid it — was not.

## Reproduction

```python
import torch, ttnn
a = torch.tensor([[1e25, 1e20, 1.3835e19, 1e-20]], dtype=torch.bfloat16)
b = torch.tensor([[0.0,  1e20, 1.3835e19, 1e-20]], dtype=torch.bfloat16)
mk = lambda t: ttnn.from_torch(t, dtype=ttnn.bfloat16, device=device, layout=ttnn.TILE_LAYOUT)
print(ttnn.to_torch(ttnn.hypot(mk(a), mk(b))))    # inf, inf, inf, 0.0
print(ttnn.get_golden_function(ttnn.hypot)(a, b)) # 9.9736e+24, 1.4123e+20, 1.9600e+19, 1.4188e-20
# the third column is the pair case: both operands are below 2^64, and only the pair overflows
```

Measured on ttsim, Wormhole and Blackhole, bfloat16 — identical on both. I do not have a device;
these are simulator runs, not silicon measurements.

## Relationship to #29406

[PR #29406](https://github.com/tenstorrent/tt-metal/pull/29406) ("#27654: Migrate Hypot as device
op", merged 2025-10-10) removed the host-level composite and introduced the
`SQUARE / SQUARE / ADD / SQRT` block quoted above, reporting a 67% kernel-duration improvement. It
also added two hypot tests.

So this is not a suggestion to make hypot a device op — that landed nine months ago. It is that the
decomposition chosen in that migration squares both operands, and that its own acceptance tests stop
well short of where squaring overflows or enters the normal-range/FTZ boundary:

| test (both added by #29406) | largest magnitude | assertion |
|---|---|---|
| `tests/ttnn/unit_tests/operations/eltwise/test_binaryng_fp32.py:164` (`test_hypot_multi_dtype`) | `999.999` (`test_binaryng_fp32.py:170`, a fixed literal list) | `test_binaryng_fp32.py:180` `assert_with_ulp(ulp_threshold=1)`; `test_binaryng_fp32.py:178` `assert_allclose(rtol=1e-5)` |
| `tests/ttnn/unit_tests/operations/eltwise/test_binaryng_ND.py:71` (in the op list), `test_binaryng_ND.py:101-102` | 50 | `test_binaryng_ND.py:102` `assert_with_ulp(ulp_threshold=2)` |

The coverage is stricter than PCC — ULP-1 and ULP-2 — but the largest magnitude exercised is `1e3`,
against the measured onset of `2^64` (`1.84467e19`): sixteen orders of magnitude away. The band is
outside test coverage rather than inside it with a tolerance that is too loose. On the small side,
the ND test constructs bfloat16 values with `torch.rand(..., dtype=torch.bfloat16) * 100 - 50`
(`test_binaryng_ND.py:77,82`); its smallest nonzero magnitude is `0.25`, so it cannot emit this
band at all.

## Which dtypes take this path

`binary_ng_device_operation.cpp:33-37` routes `HYPOT` to the SFPU path only when **both** inputs are
float32. It shares that rule with four other composed ops:

```c
case LOGADDEXP:
case LOGADDEXP2:
case LDEXP:
case BIAS_GELU:
case HYPOT: return (a == FLOAT32 && b == FLOAT32);
```

So bfloat16 takes the FPU composite quoted above and float32 takes a different path. Everything in
this report is measured on bfloat16. `test_hypot_multi_dtype` parametrises both dtypes
(`test_binaryng_fp32.py:163-164`), so a fix should say which one it targets; I have only measured
bfloat16.

The same `SQUARE / SQUARE / ADD / SQRT` block also appears in the quasar tree
(`ttnn/cpp/ttnn/operations/experimental/quasar/binary_ng/device/binary_ng_utils.cpp:391-396`).

## Proposed fix

Use the standard technique: scale both operands by a power of two before squaring, and scale the
result back. Powers of two are exact in binary floating point, so nothing is lost in the scaling
itself — it only moves where the intermediate lands. The guard has to sit below the point where the
sum of *two* squares can overflow, not merely where one can.

Two consequences worth stating up front rather than having you find them:

- Whether inputs that work today stay bit-identical depends on the placement. Pure power-of-two
  scaling inside the existing composite changes nothing for them; a dedicated path with different
  intermediate precision can move currently-passing results — no less accurate, but not identical.
  That is one more reason the placement is a maintainer call rather than a detail.
- Placement is a real decision, not an oversight in this report. The bfloat16 op takes the FPU
  composite path shown above, so the change belongs either in that decomposition or in a dedicated
  path, and the two have different cost profiles. I have sanity-checked the scaled arithmetic on
  ttsim, but through a different dispatch class — a single SFPU kernel with fp32 register
  intermediates — not through the bfloat16 FPU-composite path this report measures, and I have no
  way to exercise the host-side wiring in simulation at all. So the arithmetic is checked; the
  placement is not, and the choice of placement determines what I can validate before a PR.

## Verification scope

The failing bfloat16 operation itself was reproduced end to end on ttsim for both Wormhole and
Blackhole, against the operator-dtype `torch.hypot` golden shown above. These are simulator runs,
not silicon measurements. The report makes no claim about a final integration, its throughput, or
the float32 and special-value paths.

## Success criteria

- [ ] `hypot(x, 0) == |x|` exactly, for every normal bfloat16 `x`, on both architectures — this
      family needs no tolerance argument at all
- [ ] both measured single-operand bands — `|x| >= 2^64` and `0 < |x| < 2^-63` — return a finite
      result matching the golden within the ULP thresholds this repository already applies to
      `hypot` (1 and 2, from #29406)
- [ ] the same for pairs where **each** operand is below the single-operand onset but the sum of
      squares still overflows — measured onset at or below `1.5 x 2^63`, so a guard placed at `2^64`
      would satisfy the previous item and still return `inf` here
- [ ] ordinary inputs keep passing the two tests added by #29406, at their present thresholds
- [ ] bfloat16 test coverage extended past `1e3`, which is where the current sweeps stop, so the
      band cannot silently reappear
- [ ] the same checks run on Blackhole in its own right, not inferred from Wormhole

Scope limits are listed below.

## What this does not settle

- The float32 path. It takes a different SFPU path and was not investigated.
- `hypot(inf, x)` and `hypot(nan, x)`. Not measured; a future implementation would need to specify
  its handling of them.
- Instruction count and throughput. I have not measured either; the final integration path has not
  yet been selected.
- Whether this belongs as a follow-up to #29406 or as a separate change.

## Next step

The failing bands, the formulation and the test gap above are worked out. The placement is not, and
it is the piece that needs a maintainer's call — it also decides how much I can validate before
opening anything. I can prepare a PR with regression coverage for both bands and for ordinary
inputs, and would welcome a steer on which integration point you prefer.

kinginu:
## PR Category

Bug fix.

## Summary

Closes #51861.

`ttnn.hypot` decomposed into `SQUARE, ADD, SQUARE, SQRT`. The squares leave the format long before `hypot` does: `x²` is `inf` for `|x| >= 2^64` and stops being normal for `0 < |x| < 2^-63`, while `hypot(x, y)` at those magnitudes is an ordinary finite number. Both dtypes have an 8-bit exponent, so both are affected.

Scaling the pair by a power of two before squaring fixes it. The scale is exact, so the answer is the one the plain formula would give in a wider format, and the single square root the plain formula already had is the only one there is.

```
n, m = min_max(|a|, |b|)
if m > 2^63:   m, n, undo = m·2^-100, n·2^-100, 2^100
elif m < 2^-63: m, n, undo = m·2^100,  n·2^100,  2^-100
result = sqrt(m² + n²) · undo
```

`HYPOT` becomes a dedicated SFPU binary op for fp32 and bf16 pairs; the composite pattern in `binary_ng_utils.cpp` is deleted.

## Accuracy

**Every number below compares against `torch.hypot` evaluated in float64 and rounded to the output dtype.** `main` is this branch's base, `04cf22f7ee`; both builds ran on the same two machines.

**bfloat16, exhaustive.** All 65536 bit patterns in each operand position, against seven fixed second operands — 917504 pairs, identical results on P300 and N300:

| | exact | 1 ULP | > 1 ULP | `inf` for a finite answer | `0` for a nonzero answer |
|---|---|---|---|---|---|
| main | 505828 | 15420 | 0 | **327168** | **65532** |
| this branch | **913420** | 20 | 0 | **0** | 508 |

392700 of the 917504 pairs — 42.8% — come back from `main` as `inf` or `0` when the answer is neither. The 508 that still come back as `0` are the bfloat16 subnormal operands, which the SFPU flushes on the way in; `main` returns `0` for those too.

The 3556 patterns with a NaN operand are excluded: a bfloat16 DST cannot carry a NaN and returns `inf` instead, on this branch and on `main` alike.

**float32, exhaustive.** All 2^32 bit patterns of `x`, three fixed second operands, 4294967296 pairs each. Same counts on both machines:

| | | main | this branch |
|---|---|---|---|
| `hypot(x, 0)` | exact | 2121671154 | 4226565090 |
| | 1 ULP | 25812496 | 51624992 |
| | `inf` for a finite answer | **1073741824** | 0 |
| | `0` for a nonzero answer | **1073741822** | 16777214 |
| `hypot(x, 1)` | `inf` for a finite answer | **1073741824** | 0 |
| `hypot(x, 1e30)` | exact | 16777216 | 4166355260 |
| | `inf` for a finite answer | **4278190080** | 0 |

Half of every `hypot(x, 0)` input fails on `main`: a quarter overflow, a quarter underflow. With the second operand at `1e30` all but 16777216 of the 4.29 billion come back `inf`. Nothing on this branch exceeds 1 ULP, and the 16777214 that return `0` are the float32 subnormals, again matching `main`.

**Special values.** `hypot(±inf, anything)` is `+inf` and `hypot(NaN, finite)` is NaN on both. One case changes: IEEE 754 says an `inf` operand wins over a NaN one, and `main` returns NaN for `hypot(inf, NaN)` in float32. This branch returns `+inf`.

## Cost

One dispatch before and one after — a single `BinaryNgDeviceOperation` either way. Wall clock is 5 warm-up calls then 50 timed calls with one `synchronize_device` at the end, median of five such runs, profiler off.

**P300, µs per call:**

| tiles | dtype | main | this branch |
|---|---|---|---|
| 4096 | bfloat16 | 62.8 | **60.6** |
| 4096 | float32 | 122.0 | **121.1** |
| 16384 | bfloat16 | 238.5 | **227.8** |
| 16384 | float32 | 474.1 | **470.6** |

**N300, µs per call:**

| tiles | dtype | main | this branch |
|---|---|---|---|
| 4096 | bfloat16 | 137.3 | 155.8 |
| 4096 | float32 | 268.8 | **269.5** |
| 16384 | bfloat16 | 510.1 | 586.2 |
| 16384 | float32 | 985.0 | **979.0** |

float32 already ran in the SFPU, and there the fix is free on both machines. bfloat16 did not: `is_binary_sfpu_op` refused it, so it ran the composite on the FPU. Moving it to the SFPU is what makes the fix possible for bfloat16 at all, and it is 4% cheaper than the FPU composite on P300 and 15% dearer on N300.

At 64 tiles the call is host-bound on both machines — 25 µs on P300 and 50 µs on N300 for every build measured, with a spread wider than the difference.

### Against the first version of this branch

The first version evaluated `m · sqrt(1 + (n/m)²)` and built `1/m` as `rsqrt(m)²`, which calls the square-root body twice where the plain formula calls it once. That is the whole of the difference:

| 16384 tiles, bfloat16 | first version | this version |
|---|---|---|
| P300 | 302.1 | **227.8** (1.33×) |
| N300 | 763.5 | **586.2** (1.30×) |

It was also less accurate than the form it replaced in two ways the exhaustive sweeps found: 3623120 patterns of `hypot(x, 1)` and 4723276 of `hypot(x, 1e30)` landed more than 1 ULP out, worst case 2.38e-07, and all 16777214 float32 subnormal operands came back NaN — `main` returns `0` for those, so it was a regression rather than a residual gap. The reciprocal is where both came from, and there is no reciprocal here.

## Tests

`test_hypot_extremes.py`, 47 cases: the exhaustive bfloat16 sweep in both operand orders against five fixed operands, the two magnitude bands in both dtypes, and the IEEE special values. 32 of the 47 fail on `main`, all 47 pass here, on P300 and on N300.

## Not covered

- Subnormal operands. The SFPU flushes them, so `hypot(2^-140, 0)` returns `0` in float32 and in bfloat16. Unchanged from `main`, and fixing it means normalising by hand in the kernel.
- NaN through a bfloat16 DST, which comes back as `inf`. Also unchanged from `main`, and not specific to this op.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_hypot_extremes.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_hypot_extremes.py
@@ -2,101 +2,107 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
-"""Regression tests for ttnn.hypot at operand magnitudes where the naive sqrt(a^2 + b^2)
-composition overflows or underflows even though hypot itself is a finite normal value.
+"""ttnn.hypot at the magnitudes where a^2 leaves the format.
 
-The naive composite returns inf for |a| >= 2^64 and 0 for 0 < |a| < 2^-63 in bfloat16
-(same magnitude bounds apply to float32; only the representable set differs). torch.hypot
-returns a finite value across both bands (see issue #51861).
-
-Also exercises the ±inf edge cases that follow from the rescaled formulation: hypot(inf, inf)
-must be inf (the raw m * sqrt(1 + (n/m)²) yields NaN via inf/inf without the override).
+hypot(a, b) is a finite normal for every pair a plain sqrt(a^2 + b^2) cannot
+reach: the square overflows for |x| >= 2^64 and stops being normal for
+0 < |x| < 2^-63, while hypot itself is still representable. These tests pin the
+two bands, the special values, and an exhaustive bfloat16 sweep.
 """
 
 import math
 
 import pytest
 import torch
+
 import ttnn
 
+# Everything a bfloat16 can hold, as one 64-tile input.
+ALL_BF16 = torch.arange(0, 65536, dtype=torch.int64).to(torch.int32).to(torch.int16).view(torch.bfloat16)
 
-def _one_pair(a, b, device, dtype):
-    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
-    padded_a = torch.zeros((1, 1, 32, 32), dtype=torch_dtype)
-    padded_b = torch.zeros((1, 1, 32, 32), dtype=torch_dtype)
-    padded_a.view(-1)[0] = float(a)
-    padded_b.view(-1)[0] = float(b)
-    ta = ttnn.from_torch(padded_a, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
-    tb = ttnn.from_torch(padded_b, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
-    return ta, tb, padded_a, padded_b
+TORCH_DTYPE = {ttnn.bfloat16: torch.bfloat16, ttnn.float32: torch.float32}
 
 
-def _rel_ulp(got, ref, mantissa_bits):
-    """|got - ref| in units of 1 ULP of |ref| at the target precision."""
-    if not math.isfinite(ref) or ref == 0:
+def _hypot(device, a, b, dtype):
+    ta = ttnn.from_torch(a, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    tb = ttnn.from_torch(b, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    return ttnn.to_torch(ttnn.hypot(ta, tb))
+
+
+def _pair(a, b, dtype):
+    t = TORCH_DTYPE[dtype]
+    ta = torch.full((1, 1, 32, 32), float(a), dtype=t)
+    tb = torch.full((1, 1, 32, 32), float(b), dtype=t)
+    return ta, tb
+
+
+def _ulps(got, ref, mantissa_bits):
+    """|got - ref| in units of one ULP of ref."""
+    if not math.isfinite(ref) or ref == 0.0:
         return 0.0 if got == ref else float("inf")
-    ulp = 2.0 ** (math.floor(math.log2(abs(ref))) - mantissa_bits)
-    return abs(got - ref) / ulp
+    return abs(got - ref) / 2.0 ** (math.floor(math.log2(abs(ref))) - mantissa_bits)
 
 
+@pytest.mark.parametrize("other", [0.0, 1.0, 1e-30, 1e30, -3.0])
+@pytest.mark.parametrize("swap", [False, True])
+def test_hypot_bf16_exhaustive(device, other, swap):
+    """Every bfloat16 bit pattern against one fixed operand, both orders."""
+    swept = ALL_BF16.reshape(1, 1, 256, 256)
+    fixed = torch.full((1, 1, 256, 256), other, dtype=torch.bfloat16)
+    a, b = (fixed, swept) if swap else (swept, fixed)
+
+    got = _hypot(device, a, b, ttnn.bfloat16).flatten()
+    ref = torch.hypot(a.to(torch.float64), b.to(torch.float64)).to(torch.bfloat16).flatten()
+
+    # A bfloat16 DST cannot carry a NaN, and subnormal operands are flushed
+    # before the kernel sees them; both are true of the composite this replaces.
+    interesting = ~ALL_BF16.isnan() & ~((ALL_BF16 != 0) & (ALL_BF16.abs().to(torch.float32) < 2**-126))
+
+    # The square root is a polynomial, not a correctly rounded operation, so the
+    # last bit is allowed to move; nothing beyond it is.
+    g, r = got.to(torch.float64), ref.to(torch.float64)
+    within = (g == r) | ((r != 0) & torch.isfinite(r) & ((g - r).abs() <= r.abs() * 2**-7))
+    wrong = ~within & interesting
+    assert not wrong.any(), (
+        f"{int(wrong.sum())} of {int(interesting.sum())} patterns off by more than one ULP, "
+        f"first at 0x{int(ALL_BF16.view(torch.int16)[wrong.nonzero()[0, 0]].item()) & 0xffff:04x}: "
+        f"got {got[wrong][0].item()}, expected {ref[wrong][0].item()}"
+    )
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
 @pytest.mark.parametrize(
     "a, b",
     [
+        # a^2 overflows, hypot does not
         (1e25, 0.0),
         (3e38, 0.0),
-        (1e-20, 0.0),
         (1e25, 1e25),
         (1e30, 1e30),
-        (1e20, 1e20),
-        (1e-20, 1e-20),
-        (3.0, 4.0),
-    ],
-)
-def test_hypot_extremes_bf16(device, a, b):
-    ta, tb, torch_a, torch_b = _one_pair(a, b, device, ttnn.bfloat16)
-    got = ttnn.to_torch(ttnn.hypot(ta, tb)).float().view(-1)[0].item()
-    ref = torch.hypot(torch_a.float(), torch_b.float()).view(-1)[0].item()
-    assert math.isfinite(got), f"hypot({a}, {b}) = {got}, expected finite {ref}"
-    if ref == 0.0:
-        assert got == 0.0
-    else:
-        u = _rel_ulp(got, ref, mantissa_bits=7)
-        assert u <= 4.0, f"hypot({a}, {b}) = {got}, ref {ref}, {u:.2f} bf16 ULP"
-
-
-@pytest.mark.parametrize(
-    "a, b",
-    [
-        # Overflow band, |x| > 2^64 ~= 1.84e19
-        (1e25, 0.0),
-        (3e38, 0.0),
-        (2e30, 0.0),
-        (1e25, 1e25),
-        (1e30, 1e30),
-        # Underflow band, |x| < 2^-63 ~= 1.08e-19
+        (1.9e19, 1.0),
+        # a^2 stops being normal, hypot does not
         (1e-20, 0.0),
         (1e-30, 0.0),
-        (1e-35, 0.0),
+        (1e-35, 1e-35),
         (1e-20, 1e-20),
-        (1e-30, 1e-30),
-        # Ordinary controls
+        (1.0e-19, 1.0e-19),
+        # in band
         (3.0, 4.0),
         (1.0, 0.0),
         (100.0, 100.0),
     ],
 )
-def test_hypot_extremes_fp32(device, a, b):
-    ta, tb, torch_a, torch_b = _one_pair(a, b, device, ttnn.float32)
-    got = ttnn.to_torch(ttnn.hypot(ta, tb)).view(-1)[0].item()
-    ref = torch.hypot(torch_a.double(), torch_b.double()).view(-1)[0].item()
-    # The rewrite adds no rounding beyond an fp32 sqrt/mul/div chain; ~4 fp32 ULP covers
-    # the natural error of the formulation.
-    assert math.isfinite(got), f"hypot({a}, {b}) = {got}, expected finite {ref}"
-    u = _rel_ulp(got, ref, mantissa_bits=23)
-    assert u <= 16.0, f"hypot({a}, {b}) = {got}, ref {ref}, {u:.2f} fp32 ULP"
+def test_hypot_bands(device, dtype, a, b):
+    ta, tb = _pair(a, b, dtype)
+    got = _hypot(device, ta, tb, dtype).float().view(-1)[0].item()
+    ref = torch.hypot(ta.double(), tb.double()).view(-1)[0].item()
+    bits = 7 if dtype == ttnn.bfloat16 else 23
+    ref = float(torch.tensor(ref, dtype=TORCH_DTYPE[dtype]))
+    assert math.isfinite(got), f"hypot({a}, {b}) = {got}, expected {ref}"
+    u = _ulps(got, ref, bits)
+    assert u <= 1.0, f"hypot({a}, {b}) = {got}, expected {ref}, off by {u:.2f} ULP"
 
 
-@pytest.mark.parametrize("dtype", [ttnn.float32, ttnn.bfloat16])
 @pytest.mark.parametrize(
     "a, b, expected",
     [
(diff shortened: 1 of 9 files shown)


singhharsh1708:
Two notes on coverage, from modelling `hypot_bw` against this change — both small, neither a separate issue.

**`hypot_bw` inherits this fix but isn't exercised by it.** `binary_backward.cpp:323` calls `ttnn::hypot(input_a, other)` and then takes a reciprocal of the result, so today it returns `(0, 0)` at `(1e20, 0)` and `(inf, inf)` at `(1e-20, 1e-20)` purely from the forward. Your kernel repairs that with no change to the composite. `test_hypot_extremes.py` only drives `ttnn.hypot`, so a row calling `ttnn.hypot_bw` at the same extremes would lock in the inherited behaviour.

**One band survives the fix.** `hypot_bw` forms `reciprocal(hypot(a, b))`, and once `hypot` lands above 2^126 that reciprocal is a genuine subnormal and flushes, so the gradient goes to zero while the true value is bounded in [-1, 1]:

| a | b | hypot | true ∂/∂a | after this PR |
|---|---|---|---|---|
| 3e38 | 1e38 | 3.1623e38 | 0.9487 | 0.0 |
| 2e38 | 2e38 | 2.8284e38 | 0.7071 | 0.0 |
| 1e38 | 1e38 | 1.4142e38 | 0.7071 | 0.0 |
| 4e37 | 4e37 | 5.6569e37 | 0.7071 | 0.7071 (ok) |

That is the top two binades only, and it is the unguarded-reciprocal mechanism rather than anything about hypot — same shape as #52094 (div) and #55128 (rdiv), where the fix has to be in the caller because above 2^126 the true reciprocal really is subnormal. Not asking you to take it on here; flagging it so it isn't assumed covered once this merges.

mateusznowakTT:
@kinginu : Please resolve merge conflicts.
Then, please make sure in addition to ttnn API tests, LLK level tests exist for the kernel changed in this PR.

kinginu:
Done, both.

The conflicts were four one-line collisions with `NEXTAFTER`, both sides kept.

LLK coverage is wired the way atan2 is, since neither has a dedicated SfpuType: a `BinaryOp` member on both arches, init and call dispatch, a `torch.hypot` golden, and two tests. Removing the scale from the kernel fails the out-of-band one at 2^70 and 2^-70 and leaves the ordinary range green, so the test is pinning the scale rather than the formula. hypot is also enrolled in `BINARY_SPECIALS_READY_OPS`, so the edge sweep now holds it to IEEE 754 at infinities and NaNs.

P300: hypot 20 passed, the whole binary SFPU suite 957 passed, `test_sfpu_domains` 128 passed, `test_hypot_extremes.py` 47 passed.

PR #53909, opened 2026-08-21
[Bug fix] SFPU unary comparisons: compare on the comparison instructions, not on a subtract

### LINKED ISSUE #53908: The unary comparisons are a subtract and an int32 test
**Component / Area:** ttnn eltwise unary / SFPU LLK

**Issue Type:** Bad Outputs

## Observed

The six unary comparison activations do not answer what the binary comparisons answer. Three operand classes, both dtypes, P300 and N300, float32 shown.

**A negative zero is negative and non-zero to them:**

| | `-0.0 < 0` | `-0.0 == 0` | `-0.0 != 0` | `-0.0 >= 0` |
|---|---|---|---|---|
| `ttnn.lt(x, 0.0)`, `ttnn.ltz(x)`, correct | false | true | false | true |
| the activation | **true** | **false** | **true** | **false** |

**An infinite parameter is never matched,** and a NaN operand compares greater than everything. `x = [inf, -inf, 1.0, 0.0, -1.0, 2.0, nan]`:

```
ttnn.eq(x, inf)   [1,0,0,0,0,0,0]   UNARY_EQ inf   [0,0,0,0,0,0,0]
ttnn.le(x, inf)   [1,1,1,1,1,1,0]   UNARY_LE inf   [0,1,1,1,1,1,0]
ttnn.gt(x, 1.0)   [1,0,0,0,0,1,0]   UNARY_GT 1.0   [1,0,0,0,0,1,1]
```

With a NaN parameter, `UNARY_GT` and `UNARY_GE` answer 1 for every element.

**And two infinities order differently on the two machines,** from byte-identical source. bfloat16, parameter `inf`, operand `inf`:

| | P300 | N300 | binary op, correct |
|---|---|---|---|
| `UNARY_GT` | **1** | 0 | 0 |
| `UNARY_LT` | 0 | **1** | 0 |
| `UNARY_GE` | 1 | **0** | 1 |

Every bfloat16 bit pattern and 4194304 float32 patterns, seven parameters, six ops: **8752592 of 178913280 comparisons disagree with the binary op on P300, 8720020 on N300.**

`ckernel_sfpu_unary_comp.h` compares with `v_if(v == s)`, which is `SFPMAD` then `SFPSETCC` on the difference. `SFPSETCC` reads its register as an int32, so `0x80000000` is negative and non-zero; `inf - inf` is NaN, so equal infinities never compare equal and their order is the sign of that NaN; a NaN operand reads as positive. `ckernel_sfpu_comp.h` already cancels the first two back out with `SFPSETCC(ABS_V, NE0)` and an `SFPIADD` against inf, which is why `ttnn.ltz` is right and the activation is not.

## Expected

`UNARY_EQ` and `ttnn.eq` answer the same question.

`SFPGT` (0x97) and `SFPLE` (0x96) compare properly — measured raw on P300: sign-magnitude handled, `±0` equal, and the only departure from IEEE is that a NaN is ordered by its sign, which one `abs(a) + abs(b) <= inf` test removes. Wormhole B0 has neither and gets the same ordering from `SFPSWAP`. `ckernel_sfpu_binary_comp.h` already uses both, which is why the binary side is right.

## Not established

- Whether any caller other than `atanh_bw` depends on the current answers. Folding an `eq` against inf onto a `logical_and` there stops the sign flip at `|x| == 1`, so `atanh_bw(grad = -1, x = 1)` returns `+inf` against torch's `-inf`.
- Subnormal operands, where the activation and the binary op agree and `ttnn.eqz` does not. Separate, and not filed here.
- The integer comparison paths.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53908.

The six unary comparison activations compared with `v_if(v == s)`, which is `SFPMAD` then `SFPSETCC` on the difference. `SFPSETCC` reads its register as an int32, so `-0.0` is negative and non-zero to it; `inf - inf` is NaN, so two equal infinities never compare equal and their ordering is the sign of that NaN, which is not the same on Blackhole and Wormhole; and a NaN operand reads as positive, so `NaN > x` answers true.

Blackhole has `SFPGT` and `SFPLE`, which compare properly: sign-magnitude handled, `±0` equal, and the only departure from IEEE is that a NaN is ordered by its sign, which one `abs(v) + abs(s) <= inf` test removes. Wormhole B0 has neither and gets the same ordering from `SFPSWAP`'s min/max. Both are what `ckernel_sfpu_binary_comp.h` does for the two-tensor forms, so the unary answers are now the binary answers by construction. The scalar and its magnitude are loop invariant and hoisted; the per-element cost over the subtract is one sign and one add.

The two architecture copies of the file were byte-identical and now differ, as the two `binary_comp.h` copies already do.

## Accuracy

Every number is this branch against `04cf22f7ee`, built and run on the same device, scored against the binary comparison.

Every one of the 65536 bfloat16 bit patterns, and 4194304 float32 bit patterns drawn uniformly from the 2^32 space, against seven parameters — `±inf`, `nan`, `0`, `1`, `-2.5`, `3.4e38` — for all six ops. 178913280 comparisons.

|---|---|---|---|---|
| disagree with the binary op | 8752592 | **4** | 8720020 | **4** |

The four are one bfloat16 case where the activation is right and the binary op is not, identical on both machines.

| | before | after |
|---|---|---|
| `-0.0 < 0` | true | false |
| `-0.0 == 0` | false | true |
| `x == inf` for `x = inf` | never true | true |
| `x <= inf` for `x = inf` | false | true |
| `inf > inf` | true on P300, false on N300 | false on both |
| `inf >= inf` | true on P300, false on N300 | true on both |
| `nan > x`, any `x` | true | false |
| `x > nan`, every element | true | false |

Subnormal operands are unchanged: the SFPU add flushes them, so `abs(v) + abs(s) == 0` reads them as equal to zero, which is what `binary_comp.h` does and says it does.

## Cost

Wall clock through `ttnn.add` with the activation attached, 5 warm-up then 50 timed calls with one `synchronize_device` at the end, median of five runs, profiler off, µs per call.

| | | main | this branch |
|---|---|---|---|
| P300 | 1024 tiles, bfloat16 | 26.1 | **25.6** |
| | 16384 tiles, bfloat16 | 225.8 | **225.4** |
| | 16384 tiles, float32 | 455.2 | 455.3 |
| N300 | 1024 tiles, bfloat16 | 55.0 | **54.0** |
| | 16384 tiles, bfloat16 | 510.1 | 515.4 |
| | 16384 tiles, float32 | 960.4 | **960.4** |

The 1% on N300 bfloat16 is the `SFPMOV` that `SFPSWAP` needs. Instruction choice is not what costs: 256 copies of a pair inserted into the kernel give 1928.2 µs for `SFPGT`+`SFPENCC`, 1925.9 for `SFPSETCC`+`SFPENCC` and 1926.9 for two `SFPNOP`.

## Tests

`test_unary_comparison_specials.py`, 50 cases: the six ops against `±inf`, `nan` and a finite parameter in both dtypes, each checked against the binary op as well as against the exact answer, plus an exhaustive sweep of every bfloat16 bit pattern against `±inf`. 30 of the 50 fail on `main`; all 50 pass here, on P300 and on N300.

## Not covered

- Subnormal operands, as above.
- The integer comparison paths.
- Quasar. The change is in the Blackhole and Wormhole `llk_api` headers.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_unary_comparison_specials.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_unary_comparison_specials.py
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""The unary comparison activations against infinities and NaNs.
+
+A float comparison on the SFPU is a subtract and a test on the result, which is
+not an IEEE comparison: two equal infinities subtract to NaN so they never
+compare equal, and a NaN operand subtracts to a NaN that reads as positive so
+`NaN > x` answers true. The binary comparisons do not have either hole, so each
+case here is checked against the binary op as well as against torch.
+"""
+
+
+import pytest
+import torch
+import ttnn
+
+SHAPE = (1, 1, 32, 32)
+INF = float("inf")
+NAN = float("nan")
+
+# activation, binary op, and the python answer
+OPS = [
+    ("UNARY_EQ", ttnn.UnaryOpType.UNARY_EQ, ttnn.eq, lambda a, b: a == b),
+    ("UNARY_NE", ttnn.UnaryOpType.UNARY_NE, ttnn.ne, lambda a, b: a != b),
+    ("UNARY_GT", ttnn.UnaryOpType.UNARY_GT, ttnn.gt, lambda a, b: a > b),
+    ("UNARY_LT", ttnn.UnaryOpType.UNARY_LT, ttnn.lt, lambda a, b: a < b),
+    ("UNARY_GE", ttnn.UnaryOpType.UNARY_GE, ttnn.ge, lambda a, b: a >= b),
+    ("UNARY_LE", ttnn.UnaryOpType.UNARY_LE, ttnn.le, lambda a, b: a <= b),
+]
+
+OPERANDS = [INF, -INF, 1.0, 0.0, -1.0, 2.0, NAN]
+
+
+def _to_device(values, dtype, torch_dtype, device):
+    t = torch.zeros(SHAPE, dtype=torch_dtype)
+    for i, v in enumerate(values):
+        t[0, 0, 0, i] = v
+    return t, ttnn.from_torch(t, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.float32])
+@pytest.mark.parametrize("parameter", [INF, -INF, NAN, 1.0], ids=["inf", "neg_inf", "nan", "one"])
+@pytest.mark.parametrize("name, op, binary, expected", OPS, ids=[o[0] for o in OPS])
+def test_unary_comparison_specials(device, dtype, parameter, name, op, binary, expected):
+    torch_dtype = torch.bfloat16 if dtype == ttnn.bfloat16 else torch.float32
+    n = len(OPERANDS)
+
+    host, x = _to_device(OPERANDS, dtype, torch_dtype, device)
+    _, zeros = _to_device([0.0] * n, dtype, torch_dtype, device)
+
+    got = ttnn.to_torch(ttnn.add(x, zeros, input_tensor_a_activations=[ttnn.UnaryWithParam(op, parameter)])).flatten()[
+        :n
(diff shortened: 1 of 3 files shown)


mateusznowakTT:
@kinginu : Please resolve merge conflict and align PR with recent main changes.

PR #51184, opened 2026-07-27
Feature: Add TTTv2 Llama 3.1 8B runtime for demo and vLLM integration

gwangTT:
### PR Category

Feature

### Summary

- Add a reusable common LLM runtime covering prefill, decode, tracing, warmup, output reading, paged KV cache management, and lane-group execution.
- Add the Llama 3.1 8B TTTv2 model, executor, generator, Hugging Face adapter, and causal-LM wrapper on top of that runtime.
- Add vLLM data-parallel support and preserve TTTv1-compatible sampling, traced-prefill lengths, decode timing, and benchmark output behavior.
- Replace the obsolete hybrid Llama demo with the new runtime integration and focused unit/contract coverage.

### Notes for reviewers

- Start with `models/common/llm_runtime/README.md` for the ownership and execution model.
- Review the runtime boundary in `models/common/llm_runtime/`, then the Llama-specific composition in `models/common/models/llama3_8b/`.
- The largest behavioral risk is at prefill/decode trace boundaries, paged KV-cache ownership, and multi-lane/vLLM DP execution.
- Performance sweep coverage completed 320/320 requests with zero failures for each N150 DP1, N300 DP1/DP2, and T3K DP1/DP4/DP8 configuration under both `decode_only` and `all` tracing. Output throughput deltas versus the old path ranged from -2.2% to +6.9%.

### Test plan

- [x] Run balanced 320-request performance sweeps on N150, N300, and T3K across supported DP configurations and both trace modes.
- [x] Run the full common LLM runtime unit and contract test suite in CI.
- [x] Run hardware integration coverage in CI.

Made with [Cursor](https://cursor.com)

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:gongyu/tttv2_llama8b_perf_parity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:gongyu/tttv2_llama8b_perf_parity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:gongyu/tttv2_llama8b_perf_parity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:gongyu/tttv2_llama8b_perf_parity)
[...]

Issue #50886, opened 2026-07-23
Fix watcher-surfaced device faults in Tier 1/2/3 unit suites, then re-enable watcher (follow-up to #48924)


mtairum:
## Watcher-surfaced device issues — follow-up to #48924

PR #48924 enables the TT-Metal watcher across the Tier 1/2/3 **model unit** pipelines. With watcher active, **22 jobs** that are green on `main` fail — because watcher catches real, latent device faults that are silent when it is off. This issue tracks those faults, grouped by signature, so they can be fanned out into per-root fixes (or the affected suites excluded from watcher until fixed).

All job links below are from run [29875773915](https://github.com/tenstorrent/tt-metal/actions/runs/29875773915) (branch `45711-Enable-Watcher-in-Tier-1-2-3-unit-tests`). Causes are log-verified for a sample in each group and inferred across the rest of the cluster (same board + watcher config).

> Not included here: 10 failures from a Weka filesystem-mount infra outage and 4 galaxy-topology / expired-log jobs — neither is watcher-related. See the PR thread for the full triage.

## Progress

- [x] [Llama](https://github.com/tenstorrent/tt-metal/pull/53595) - [![](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml/badge.svg?branch=50886-re-enable-watcher-for-Llama)](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:50886-re-enable-watcher-for-Llama)  
- [x] [Mixtral-8x7B](https://github.com/tenstorrent/tt-metal/pull/53599) - [![](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml/badge.svg?branch=50886-re-enable-watcher-for-Mixtral)](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:50886-re-enable-watcher-for-Mixtral)
- [x] [GPT-OSS-120B](https://github.com/tenstorrent/tt-metal/pull/53594)  - [![](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml/badge.svg?branch=50886-re-enable-watcher-for-GPT-OSS-120B)](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:50886-re-enable-watcher-for-GPT-OSS-120B)
[...]

Issue #49081, opened 2026-07-06
[CCL] all_gather_async: NCRISC missing NOC non-posted-atomics flushed barrier (inter-kernel data race) caught by Watcher


mtairum:
## Summary

With Watcher enabled in the model unit tests (PR #48924, [run 28663401309](https://github.com/tenstorrent/tt-metal/actions/runs/28663401309), 2026-07-03), the `all_gather_async` CCL reader kernel trips a Watcher assert: an **NCRISC inter-kernel data race — the kernel completes with pending NOC transactions (missing NOC non-posted-atomics flushed barrier)**. This is a genuine device-side synchronization bug that runs silently without Watcher.

## Error signature
```
Watcher: NCRISC detected an inter-kernel data race due to kernel completing with
pending NOC transactions (missing NOC non-posted atomics flushed barrier)
  kernel: ttnn/cpp/ttnn/operations/ccl/all_gather_async/device/kernels/minimal_default_reader.cpp
-> TT_THROW: Watcher detected tripped assert and stopped device
-> Fatal Python error: Aborted (exit 134)
```
(Also observed via the `minimal_default_writer.cpp` counterpart.)

## Affected tests / models (all multi-chip `all_gather_async` on wh_llmbox_perf / T3K)
- Llama-3.1-8B unit — `test_attention.py::test_attention_inference[...paged_attention-8-False]`
- Llama-3.3-70B unit — `test_attention.py` (same test)
- Qwen2.5-72B unit — t3000 all_gather_async path
- Llama-3.2-11B-Vision unit — `test_decoder.py` (during decode token generation)

## Impact / notes
- Watcher-detected only; silent on `main` (tests pass without Watcher). The missing barrier is a real correctness/latent-hazard risk on multi-chip all-gather.
- Blocks enabling Watcher in unit CI (PR #48924) for these models until fixed; those tests are being temporarily excluded from Watcher there and tracked here.
- Discovered by: PR #48924 (enable Watcher in tier 1/2/3 unit tests), issue #45711.

## Repro
Run any of the affected unit tests with `TT_METAL_WATCHER` enabled on a T3K/llmbox, e.g.:
```
TT_METAL_WATCHER=60 pytest models/tt_transformers/tests/test_attention.py -k paged_attention
```

mtairum:
**Another reproduction, on a second op path — and it blocks tensor-cache generation, not just a test run.**

[...]

PR #53911, opened 2026-08-21
[Models CI] Split the TT-Transformers module tests between tier 1 Models unit and sweep CI

mtairum:
### Ticket

N/A — CI pipeline migration. Defects surfaced along the way are tracked in #54006.

### Problem description

The TTTv2 module tests were split across two pipelines, neither of which is where they belong.

**Seven `models_tttv2_*` jobs** were the only `models`-team entries in **`(T3K) T3000 e2e tests`** — a pipeline that runs no model end to end. Its other six jobs are CCL, fabric and dispatch tests owned by `ttnn` and `scaleout`. These are module-level PCC suites over `models/common/tests/modules/*`; their own job names say `unit_tests`, and `HF_MODEL=meta-llama/Llama-3.1-8B-Instruct` appears in five of them purely as a config source for module shapes. No model is instantiated.

**One `t3k_tttv2_fast_unit_tests` job** sat in `(T3K) T3000 unit tests`, running the same modules over the same files under `-m "not slow"`.

The split between them was deliberate. `tests/scripts/t3000/run_t3000_unit_tests.sh` says so:

```
# Run MLP1D fast unit tests (full set is run in t3k_e2e_tests.yaml
#  to match timeout values and frequency of runs)
```

Fast subset as the gate, full parametrisation as broad coverage. That maps cleanly onto the tiered pipelines, and this PR restores it there rather than flattening both into one.

### What changed

| | Was | Now |
|---|---|---|
| Fast suite | `t3k_unit_tests.yaml`, 1 job, `cmd: run_t3000_tttv2_fast_unit_tests` | `models_unit_tests.yaml`, **tier 1**, **8 entries**, inlined |
| Full set | `t3k_e2e_tests.yaml`, 7 jobs | `models_sweep_tests.yaml`, **tier 1**, 6 entries |

Both on `wh_llmbox`, both `model: tt-transformers-common` / `model_family: TTTv2`, owner unchanged (`U08JFM3CFA4`, Gongyu Wang). One shared `model:` identifier covers the set, following the `dit-common` precedent, so a single `workflow_dispatch` filter runs it. Added to both tier-1 workflows' `model` enums.

Why sweep for the full set: `models/model_ci_tiers.md` defines sweeps as *"parameter sweeps across the configurations each model supports"*, and that is what these are — MLP collects 691 cases, RMSNorm 1054. There is already precedent for a shared-module suite there (`Sampling module sweep (common)`). Sweep runs `0 6 * * 3,6`, twice weekly, the right cadence for 131 min of exhaustive parametrisation.

Why the fast suite is **eight entries and not one**: inlined as a single entry it ran nine `pytest` calls sequentially, so any one red call turned the whole TT-Transformers gate red and hid the rest. Two of the nine fail deterministically on `main`, so a monolith would have landed permanently red with six green modules invisible inside it. Eight entries mirror the six per-module sweep entries, so a module's fast gate and its full sweep share a name and an owner mapping. Each propagates its own exit code; the sampling entry keeps `|| fail=1` accumulation because it runs two files, so a `penalties` failure cannot hide a `sampling_1d` one.

The calls themselves are transcribed from the shell function rather than paraphrased — verified call-for-call, including which of them receive `HF_MODEL` (five of nine) and which receive a `TT_CACHE_PATH` (six of nine).

### The two suites are not nested

Worth knowing before reviewing the split — neither is a subset of the other.

**Full set covers, fast suite does not:**

| File | Why |
|---|---|
| `mlp/test_mlp_2d.py` | fast names `test_mlp_1d.py`, not the directory |
| `rmsnorm/test_rmsnorm_2d.py` | same |
| `attention/profiling/test_attention_1d_profiling.py` | same |
| `sampling/test_sampling_1d_release.py` | fast runs only `test_penalties_1d.py` and `test_sampling_1d.py` |
| everything marked `slow` | `-m "not slow"` |

**Fast suite covers, full set does not** — its first call is `pytest --ignore=models/common/tests/modules models/common/tests`, which is 31 files the seven full-set entries never reach: all of `llm_runtime/`, plus `test_lazy_weight.py`, `test_auto_compose.py`, `test_metrics.py` and the rest. 22 of those 31 reference a device fixture, so this is not host-only work that could move to a cheaper SKU.

So the two sets together cover strictly more than either pipeline did before.

### Watcher

Neither source pipeline runs these suites under a watcher: `t3000-unit-tests-impl.yaml:103` and `t3000-e2e-tests-impl.yaml:89` both call `setup-job` without `enable-watcher`, and `models-sweep-tests-impl.yaml:144` does not either. But `models-unit-tests-impl.yaml:182` passes `enable-watcher: true`, which exports `TT_METAL_WATCHER=2` (`.github/actions/setup-job/action.yml:97`).

So the fast suite would silently gain a watcher it has never run under simply by changing pipeline. All eight unit entries carry `unset TT_METAL_WATCHER` to keep parity, following the house pattern for suites blocked on #50886 (Llama 3.1-8B, Mistral-Small-3.1-24B, `Sampling module sweep (common)`). The sweep entries need no override.

This keeps the migration behaviour-preserving. Running the attention suite under `TT_METAL_WATCHER=30` during this work did expose **G1 of #50886** — the `all_gather_async` / `minimal_default_reader` assert tracked as #49081 — a pre-existing multi-chip CCL bug, silent on `main` with the watcher off and out of scope here. The occurrence is recorded on both issues so a green TTTv2 pipeline is not later read as evidence it is fixed.

### Timeouts

Sized from measured **test-step** durations, not carried over.

**Unit**, from run [32516753713](https://github.com/tenstorrent/tt-metal/actions/runs/32516753713), which ran all nine calls against a cache regenerated from empty:

| Entry | Measured | Registered |
|---|---|---|
| common | 12:05 | 15 |
| MLP | 4:13 | 6 |
| attention | 3:57 | 6 |
| embedding | 2:49 | 5 |
| LM head | 1:50 | 4 |
| sampling (penalties + `sampling_1d`) | 1:44 | 4 |
| RMSNorm | 0:51 | 3 |
| RoPE | 0:33 | 2 |

45 min total. Sequential wall clock was 32 min as one job; split, it is bounded by the 15-min `common` leg, at the cost of eight runners rather than one.

**Sweep**, from run [32488706286](https://github.com/tenstorrent/tt-metal/actions/runs/32488706286) measured under `TT_METAL_WATCHER=30`, against the watcher-free durations in [32453660804](https://github.com/tenstorrent/tt-metal/actions/runs/32453660804). Since the registered entries run watcher-free, these are conservative by the 1–5% the watcher cost:

| Module | Old declared | Watcher-free | `WATCHER=30` | Registered |
|---|---|---|---|---|
| MLP | 40 | 19.2 | 20.2 | 24 |
| attention | 65 | 3.6 (aborted) | 1.8 (aborted) | **65 (unchanged)** |
| embedding | 20 | 14.3 | 14.4 | 18 |
| LM head | 30 | 6.7 (3 failed) | 7.3 (**30 passed**) | 10 |
| RMSNorm | 10 | 7.1 | 7.2 | 10 |
| RoPE | 5 | 2.2 | 2.3 | 4 |

131 min total. Attention stays at 65 because it has never completed a full run, so there is nothing to size against.

### Budgets

| Key | Change | Requested |
|---|---|---|
| `models.unit_tier1.wh_llmbox` | 61 → **86** | 70 |
| `models.sweep_tier1.wh_llmbox` | **new, 131** | 131 |
| `models.unit.wh_llmbox` | 70 → **45** | 30 |
| `models.e2e.wh_llmbox` | **removed** (was 198) | — |

`unit_tier1` at 86 is 61 plus the 25 min the fast suite was declared at, preserving the 36 min of headroom that line already carried — the eight-way split fits inside it without a further increase. `unit` drops by the 25 the fast suite vacated, preserving its 15 min of headroom; it is 70 → 45 rather than 80 → 55 after rebasing onto main, which took the same line 80 → 70 when #53903 removed the Grok-1 T3K unit test. `e2e` has no consumer once the seven leave: the remaining T3K e2e jobs are `ttnn`/`scaleout`, `release-demo-tests-impl.yaml` passes a tier so it resolves to `e2e_tier<n>`, and `models_e2e_tests.yaml` has no `models`-team `wh_llmbox` entry.

The point of the split, in LLMBox occupancy:

| | Declared | Cadence | **Min/week** |
|---|---|---|---|
| Fast suite in `t3k_unit_tests` | 25 | `0 */6 * * *` — 4/day | 700 |
| Full set in `t3k_e2e_tests` | 178 | `0 6 * * *` — 1/day | 1246 |
| **Before, total** | | | **1946** |
| 8 unit entries | 45 | `0 4,20 * * *` — 2/day | 630 |
| 6 sweep entries | 131 | `0 6 * * 3,6` — 2/week | 262 |
| **After, total** | | | **892** |

A 54% reduction, while covering strictly more tests than before.

### Validation

Final read-only dispatches on `59870b3765f` — read-only is the condition the scheduled pipelines run under, so it is the one that matters. **10 of 14 legs green.**

**Unit** [32520833221](https://github.com/tenstorrent/tt-metal/actions/runs/32520833221) — 5 of 8:

| Leg | Result |
|---|---|
| MLP · RMSNorm · RoPE · LM head · embedding | ✅ |
| common | ❌ `1 failed, 1718 passed, 248 skipped` — `max_abs_error=3.125e-02 exceeds tolerance 1.000000e-02` |
| sampling | ❌ `71 failed / 53 passed` on `local_indices`; `test_penalties_1d.py` `81 passed` |
| attention | ❌ PCC `0.1760`, `0.2321`, `0.6194`, `0.6557` |

**Sweep** [32520836625](https://github.com/tenstorrent/tt-metal/actions/runs/32520836625) — 5 of 6: MLP, RoPE, RMSNorm, embedding, LM head all ✅. Attention ❌ with `Read-only file system`, aborting at 7.6 min.

Every timeout held, including RoPE at 2 min and RMSNorm at 3.

Also checked:

- `verify_time_budget.py` clean on `models_unit_tests.yaml` tiers 1/2/3, `models_sweep_tests.yaml` tiers 1/2, and the trimmed `t3k_unit_tests.yaml` and `t3k_e2e_tests.yaml`.
- `prepare_test_matrix.py` builds 8 tier-1 unit legs and 6 tier-1 sweep legs on `wh_llmbox`.
- All eight inlined commands pass `bash -n`.
- Every touched YAML and `owners.json` parses.
- Earlier dispatch [32488706286](https://github.com/tenstorrent/tt-metal/actions/runs/32488706286) with `mlperf-write-access: true` took **LM head from 3 failed / 27 passed to 30 passed** by populating the tensor cache. That is only possible in the tiered pipelines: `t3000-e2e-tests-impl.yaml:72` hardcodes `/mnt/MLPerf:/mnt/MLPerf:ro`, while `models-t1-unit-tests.yaml:82-86,153` exposes `mlperf-write-access`.
- `owners.json`: the four now-dead `models tttv2 modules unit tests` and `t3k tttv2 modules tests` mappings are removed. Tiered pipelines resolve owners from the registry's `owner_id` and have no `owners.json` entries, so nothing replaces them.

### The three red legs are pre-existing `main` defects — all tracked in #54006

None is caused by this migration. The tiered pipelines simply applied a forced watcher, a read-only `/mnt/MLPerf` and a cold tensor cache that the T3K pipelines never did. Results were byte-identical across two independent runs, so these are deterministic rather than flakes.

| # | Defect | Legs affected |
|---|---|---|
| A | `Sampling1DConfig` is missing `local_indices` — `sampling_1d.py:194` reads `cfg.local_indices`, a field #53001 deleted and #51184's `load_device_buffers()` rewrite resurrected the reader for | unit `sampling` |
| B | Attention decode-path PCC, floor **0.1498**. Every failure is `-decode-32-` or the prefill→decode transition, on bias-free Llama 1B/8B/11B across 1×1/1×2/1×8. Not the `low_pcc_notes.md` Q/K-bias mechanism, not cache, not watcher | unit `attention` |
| C | `test_validation_rmsnorm_host_and_device[1x1]` tolerance — the only failure in 1719 non-module cases, and present on `main` | unit `common` |
| D | `all_gather_async` G1 — watcher-only, already #49081 / #50886 | none (watcher off) |
| E | `ttnn.load_tensor` raises uncatchable `SIGBUS` on a cache entry it cannot map | none (diagnosability only) |

**Sweep attention's `EROFS` is a consequence of B.** The full-set `attention_1d` cache has never been completed: every write-access run that would have built it aborted partway, first on D under the watcher and then on B. Read-only runs then fail on the entries those runs never reached. Fixing B lets one write-access sweep dispatch finish, which clears the `EROFS` too. The unit `attention` leg does reach B, because its `-m "not slow"` subset cache is populated — which is why B is visible there and not in sweep.

**A correction to an earlier revision of this description:** I previously wrote that the fast suite dodged A because the affected cases were `slow`-marked. They are not — `-m "not slow"` does not deselect them, and the unit `sampling` leg is red on A. The eight-way split is partly a response to that: it confines A to one leg instead of the whole gate.

### Follow-ups, all pre-existing

- `test_mlp_2d.py` and `test_rmsnorm_2d.py` run only in the full set, so they drop to twice-weekly. Closing the gap means pointing the fast suite's mlp and rmsnorm entries at the directory rather than the `_1d` file, which changes what the gate covers and is not mine to decide.
- The `common` entry re-runs `test_sampling.py` and `test_tt_sampling.py`, which `Sampling module sweep (common)` and `On-device sampling unit tests` already cover. Three pipelines, same files.
- `attention/profiling/test_attention_1d_profiling.py` contains **zero `assert`, zero `pytest.fail`, zero `raise`** — it prints latency and PCC tables and cannot detect a regression, only crash. Its reference models use `no_init_weights()`, so its PCC column is self-consistency on random weights. On 1×8 it sweeps ten model shapes including Llama-70B, Llama-90B-Vision, Qwen2.5-72B and Mixtral-8x7B, which is where most of the attention sweep's 65 min goes. Under this PR it lands in sweep, twice weekly and outside the gate; whether it belongs in a pipeline at all is a question for its owner.
- `run_t3000_tttv2_fast_unit_tests` is now reachable only from `run_t3000_tests`, which no registry dispatches. Effectively dead shell code, left in place for local use.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/migrate-tttv2-to-tier1)


roseli-TT (approved):

gwangTT (approved):
Amazing! Thanks, Miguel!

yieldthought (approved):

tests/pipeline_reorg/models_unit_tests.yaml
```diff
@@ -1490,3 +1490,80 @@
...
+- name: TT-Transformers LM head module unit tests
+  cmd: TT_METAL_WATCHER=30 HF_HUB_OFFLINE=1 HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=meta-llama/Llama-3.1-8B-Instruct TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/tttv2/lm_head_1d pytest --tb=short models/common/tests/modules/lm_head
+  model: tt-transformers-common
+  model_family: TTTv2
+  skus:
+    wh_llmbox:
+      timeout: 10
+      tier: 1
+  owner_id: U08JFM3CFA4 # Gongyu Wang
+  team: models
```
gwangTT: https://github.com/tenstorrent/tt-metal/blob/fd6676eeb1b38414c2ad7e0fc311a13c0682a551/tests/scripts/t3000/run_t3000_unit_tests.sh#L298-L391 is the original list that this should be compare against.

tests/pipeline_reorg/models_unit_tests.yaml
```diff
@@ -1490,3 +1490,80 @@
...
+- name: TT-Transformers MLP module unit tests
+  cmd: TT_METAL_WATCHER=30 HF_HUB_OFFLINE=1 HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=meta-llama/Llama-3.1-8B-Instruct pytest --tb=short models/common/tests/modules/mlp
+  model: tt-transformers-common
+  model_family: TTTv2
+  skus:
+    wh_llmbox:
+      timeout: 24
+      tier: 1
+  owner_id: U08JFM3CFA4 # Gongyu Wang
+  team: models
```
gwangTT: For example, this should be something like this:
```
HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
  TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/tttv2/mlp_1d \
  pytest models/common/tests/modules/mlp/test_mlp_1d.py \
    -m "not slow" \
    --tb=short \
    --durations=10 \
    --cov=models.common.modules.mlp.mlp_1d \
    --cov-report=term-missing \
    --cov-config=models/common/tests/setup.cfg
```
The key thing is:
- `-m "not slow"`
- `--cov **`

gwangTT (approved):
Amazing! Thanks, Miguel!

tests/pipeline_reorg/models_unit_tests.yaml
gwangTT: The same applies to all the test cases -- they should match the corresponding entries in https://github.com/tenstorrent/tt-metal/blob/fd6676eeb1b38414c2ad7e0fc311a13c0682a551/tests/scripts/t3000/run_t3000_unit_tests.sh#L298-L391

PR #53917, opened 2026-08-21
[qwen36] Add Qwen3.6-35B-A3B sparse MoE

sott0n:
### Summary

Brings up the **Qwen3.6-35B-A3B** sparse-MoE checkpoint on the `models/demos/blackhole/qwen36` stack (P150x4, 4-way TP). Arch `qwen3_5_moe`: 256 routed experts (top-8) + a gated shared expert per layer, over the existing hybrid GDN / gated-full-attention / partial-RoPE backbone.

- Each layer's dense SwiGLU MLP is replaced by a sparse MoE block (`tt/moe/`): router → expert-parallel `sparse_matmul` experts → reduce-scatter to the same fractured-hidden layout the dense MLP produces.
- Config-driven via `args.is_moe_layer`; on the dense 9B/27B `num_experts == 0`, so the dense path is **byte-for-byte unchanged**.
- Registers the 35B-A3B as its own CI model entry (unit + e2e).

**Perf (P150x4, warm trace):** traced_128 TTFT 0.49 s / 24.69 tok/s/user; traced_4k TTFT 9.86 s / 24.62.

**Validation (P150x4):** 35B-A3B — `test_moe_tp` 5/5 (incl. B=8 decode), `test_gdn_tp` 13/13, `test_attention_tp` 7/7, e2e traced_128 + traced_4k. Dense 27B (no regression) — gdn/attn/mlp/model TP + vision unit + e2e (text + vision_demo) all green.

### CI to run (model tests aren't part of the PR gate — dispatch on this branch)

- **Regression — `qwen3.6-27b` (ready now).** This PR touches shared files (`model_config.py`, `layer.py`, `mlp.py`, `gdn/tp.py`, `tt_transformers/tt/model_config.py`) that the dense 27B uses, all gated on `num_experts`. The 27B is already provisioned on the runner, so dispatch `(Tier 1) Models Unit` + `E2E` with `model=qwen3.6-27b` to confirm no regression — no weight staging needed. (Locally green: gdn/attn/mlp/model TP + vision unit + e2e.)
- **New model — `qwen3.6-35b-a3b`.** Same two workflows with `model=qwen3.6-35b-a3b`, after the weight staging below. (Locally green: `test_moe_tp` 5/5 incl. B=8 decode, `test_gdn_tp` 13/13, `test_attention_tp` 7/7, e2e traced_128 + traced_4k.)

### Notes for reviewers

1. **Weight staging (please confirm).** The 35B-A3B CI entries run offline against the read-only `/mnt/MLPerf` mount — no download step. The scheduled job goes green only once `Qwen/Qwen3.6-35B-A3B` weights are staged there and its `tt_cache/` is generated by one `mlperf-write-access` run. Can you confirm / arrange this, or should I gate the entry to dispatch-only until then?
2. **vLLM (out of repo).** The MoE checkpoint reuses the `Qwen36ForCausalLM` wrapper unchanged, but declares `architectures=["Qwen3_5MoeForConditionalGeneration"]`; the external vLLM registry (tt-inference-server / vLLM plugin) must map that architecture string to `Qwen36ForCausalLM` for vLLM to route the checkpoint here. Follow-up; not wired into `vllm-model-tests.yaml` here.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:qwen36-moe-35b-a3b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:qwen36-moe-35b-a3b)


sott0n:
Run Model Tests for Qwen3.6-27B for regression:
- https://github.com/tenstorrent/tt-metal/actions/runs/32473490978
- https://github.com/tenstorrent/tt-metal/actions/runs/32473607331

mtairum (changes requested):
@sott0n adjust the time budget so it fits the increased times added by this change.

Also paste as a comment or in the PR description a passing CI run of the added test. You don't have to run the full Tier 1. Just `qwen3.6-35B-A3B` on e2e and on unit.

sott0n:
@mtairum Do the `Qwen/Qwen3.6-35B-A3B` weights (+ `tt_cache`) need to be pre-staged on `/mnt/MLPerf`? CI loads offline against the read-only mount (no download step), so the new `qwen3.6-35b-a3b` unit/e2e can only pass once they are there. Can you confirm / arrange the staging, or point me to who does it?

mtairum:
I've downloaded the HF weights.
You can re-run your tests and enable /mnt/mlperf write access. This will enable writing the cache files.

(last bullet point when starting the job)
image


I've kicked both runs: 
- https://github.com/tenstorrent/tt-metal/actions/runs/32487205186
- https://github.com/tenstorrent/tt-metal/actions/runs/32487222480

sott0n:
@mtairum  Both runs fail at model load (~2 min) with `LocalEntryNotFoundError`, the offline load can't find the `Qwen/Qwen3.6-35B-A3B` snapshot in the runner's HF cache. Our entry mirrors the working 27B exactly (`HF_HOME=/mnt/MLPerf/huggingface`), so the weights just need to be in the same hub-cache.

Could you check on the runner (`refs/main` is a file — cat, not ls):

```
cat /mnt/MLPerf/huggingface/hub/models--Qwen--Qwen3.6-35B-A3B/refs/main
```

If it's missing, the download probably used a different HF_HOME.

yieldthought:
Codex review:

[tests/pipeline_reorg/models_e2e_tests.yaml:484](https://github.com/tenstorrent/tt-metal/blob/4701c0ff1e135018aaa8ee7216ba84c5339da1b5/tests/pipeline_reorg/models_e2e_tests.yaml#L484) — This registers qwen3.6-35b-a3b as a Tier 1 e2e model, but the PR has no corresponding row in models/model_ci_tiers.md and no qwen3.6-35b-a3b / bh_quietbox_2 entry in models/model_targets.yaml. The e2e workflow invokes validate_perf_targets.py without --strict-missing, so benchmark data from this job will only produce a warning and cannot gate performance or accuracy regressions for the new model. Suggestion: complete the Tier 1 registration with the tier-table row and active target entries (or explicit TODO entries until measurements are available).

[models/demos/blackhole/qwen36/tt/layer.py:180](https://github.com/tenstorrent/tt-metal/blob/4701c0ff1e135018aaa8ee7216ba84c5339da1b5/models/demos/blackhole/qwen36/tt/layer.py#L180) — The new MoE dispatcher treats every non-`decode` value as prefill, while the rest of this layer treats every non-`prefill` value as decode. An unsupported mode can therefore send attention and experts down opposite paths. Suggestion: validate `mode` at the `forward` boundary before deriving `_norm_mode`.

sott0n:
addressed comments from codex review

yieldthought:
Codex review:

[models/model_targets.yaml:1384](https://github.com/tenstorrent/tt-metal/blob/afbd6319f59925db1d8461a4363bd78982a68aa2/models/model_targets.yaml#L1384) — These active target entries cannot be reached by the benchmark this PR registers. [models/demos/blackhole/qwen36/demo/text_demo.py:1086](https://github.com/tenstorrent/tt-metal/blob/afbd6319f59925db1d8461a4363bd78982a68aa2/models/demos/blackhole/qwen36/demo/text_demo.py#L1086) emits `model.args.base_model_name`, while [models/tt_transformers/tt/common.py:858](https://github.com/tenstorrent/tt-metal/blob/afbd6319f59925db1d8461a4363bd78982a68aa2/models/tt_transformers/tt/common.py#L858) truncates `Qwen3.6-35B-A3B` at the first `B-` and returns `Qwen3.6-35B`. The target resolver only does case-insensitive exact/alias matching, and this block has no `Qwen3.6-35B` alias (I verified that the full name resolves to the new 128-token target while the emitted base name resolves to `None`). Because the e2e workflow invokes `validate_perf_targets.py` without `--strict-missing`, successful runs will only warn about the missing target and these new active thresholds will not gate regressions. Suggestion: preserve the `A3B` suffix in the benchmark/canonical base name or add the exact emitted name as an alias, and cover the emitted-name-to-target lookup with a small resolver test.

[models/demos/blackhole/qwen36/tt/weight_mapping.py:192](https://github.com/tenstorrent/tt-metal/blob/afbd6319f59925db1d8461a4363bd78982a68aa2/models/demos/blackhole/qwen36/tt/weight_mapping.py#L192) — This new `tensor.dim() > 2` fallback is applied to every FP8 tensor, but it assumes dim 0 is a batch of independent 2-D expert matrices and that `scale[e]` is the corresponding 2-D block scale. Qwen state dicts also contain non-expert 3-D weights such as `linear_attn.conv1d.weight` (`[channels, 1, kernel]`); if a future FP8 checkpoint supplies a scale for one, this branch passes a `[1, kernel]` slice to `dequant_fp8_block`, whose 128×128 block reshape cannot succeed. The [published 35B-A3B config](https://huggingface.co/Qwen/Qwen3.6-35B-A3B/blob/main/config.json) declares BF16 and the comment explicitly says this path is unexercised, so this adds permissive future-checkpoint handling without an enforced format or test. Suggestion: restrict the branch to the known fused expert keys with explicit tensor/scale shape and block-divisibility checks plus a focused test, or omit it until an FP8 MoE checkpoint is in scope.

mtairum:
Hi @sott0n ! Sorry for the delay.
Found the issue. Downloaded the weights on the wrong datacenter.
tests should pass now 🤞 (or timeout when generating TTNN caches, but we can just restart them in that case).

Here's the links.
- (Tier 1) Models Unit Tests — https://github.com/tenstorrent/tt-metal/actions/runs/32985866704
- (Tier 1) Models End-To-End Tests — https://github.com/tenstorrent/tt-metal/actions/runs/32985868676

Both on `model: qwen3.6-35b-a3b`, `sku: bh_quietbox_2 (BH QB2)`.

mtairum:
The unit test passed. The e2e job seems to have gone weird (unrelated). I've kicked it again:
https://github.com/tenstorrent/tt-metal/actions/runs/33060282694

I'm approving the PR to unblock since this is likely to pass.

mtairum (approved):

roseli-TT (approved):

yieldthought:
Codex review:

[tests/pipeline_reorg/models_e2e_tests.yaml:489](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/tests/pipeline_reorg/models_e2e_tests.yaml#L489) — This adds the model as Tier 1 but runs only the traced 128- and 4096-token performance cases. [models/demos/blackhole/qwen36/demo/text_demo.py:214](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/models/demos/blackhole/qwen36/demo/text_demo.py#L214) already defines `determinism_128`, which runs two traced generations and asserts exact token equality, and the adjacent dense Qwen3.6 Tier 1 entry runs that case. As written, a nondeterministic MoE prefill/decode regression can remain green despite the Tier 1 contract promising determinism coverage. Please include the existing determinism case in this entry.

```suggestion
    pytest --timeout 3600 models/demos/blackhole/qwen36/demo/text_demo.py -k "traced_4k"
    pytest --timeout 1800 models/demos/blackhole/qwen36/demo/text_demo.py -k "determinism_128"
```

[tests/pipeline_reorg/models_e2e_tests.yaml:488](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/tests/pipeline_reorg/models_e2e_tests.yaml#L488) — These performance cases are also the only full-model E2E coverage added for this Tier 1 model, but the TP path merely checks that 50 tokens were produced ([models/demos/blackhole/qwen36/demo/text_demo.py:317](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/models/demos/blackhole/qwen36/demo/text_demo.py#L317)), with even its degeneracy guard commented out, and both new centralized targets leave `accuracy` empty ([models/model_targets.yaml:1399](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/models/model_targets.yaml#L1399)). The unit entry exercises MoE, GDN, and attention modules separately but not a full-model reference. A layer-integration or weight-mapping error can therefore produce deterministic but incorrect tokens and still pass every new gate. Please add a token/reference accuracy check for the 35B-A3B path before advertising Tier 1 accuracy coverage.

[models/demos/blackhole/qwen36/tt/weight_mapping.py:192](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/models/demos/blackhole/qwen36/tt/weight_mapping.py#L192) — The earlier FP8-loader concern remains on the current head. This fallback still classifies every scaled tensor with rank greater than two as `[experts, out, in]`, while the same loader also passes through non-expert 3-D tensors such as GDN `conv1d.weight` (`[channels, 1, kernel]`). If an FP8 checkpoint provides a scale for one, `dequant_fp8_block(tensor[e], scale[e])` receives a `[1, kernel]` slice and its 128×128 block reshape fails. The current 35B-A3B is BF16 and no 3-D FP8 loader test exercises this path, so please restrict the branch to the known fused expert keys with shape/divisibility checks and a focused test, or omit it until an FP8 MoE checkpoint is in scope.

[models/model_ci_tiers.md:215](https://github.com/tenstorrent/tt-metal/blob/7526ab094b6e2cfed5dfeec25a4c8aa9f5bb4290/models/model_ci_tiers.md#L215) — This row is under the vLLM “Models covered” table, whose source of truth is `tests/pipeline_reorg/vllm_model_tests.yaml`, but that config has no 35B-A3B entry. The PR body also says the external architecture registration is a follow-up and that this model is not wired into the vLLM test YAML here. The documentation therefore claims nightly vLLM coverage that cannot currently run; please remove this row until the registry/config entry lands, or add the runnable vLLM entry as part of this change.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -482,6 +482,21 @@
   owner_id: U08HL8X1ECD # Aniruddha Tupe
   team: models
 
+- name: Qwen3.6-35B-A3B e2e tests
+  cmd: |
+    export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=Qwen/Qwen3.6-35B-A3B TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/Qwen/Qwen3.6-35B-A3B MESH_DEVICE=P150x4
+    pytest --timeout 1800 models/demos/blackhole/qwen36/demo/text_demo.py -k "traced_128 and not traced_128k"
+    pytest --timeout 3600 models/demos/blackhole/qwen36/demo/text_demo.py -k "traced_4k"
```
yieldthought, line 489: Codex: Tier 1 promises determinism coverage, and this model already has a `determinism_128` case that the adjacent dense Qwen3.6 entry runs. Include it in the new sparse-MoE entry so nondeterministic prefill/decode regressions cannot remain green.

```suggestion
    pytest --timeout 3600 models/demos/blackhole/qwen36/demo/text_demo.py -k "traced_4k"
    pytest --timeout 1800 models/demos/blackhole/qwen36/demo/text_demo.py -k "determinism_128"
```

sott0n:
@yieldthought addressed 3/4 in 26054ee: `determinism_128` added to the 35B-A3B e2e entry, stale vLLM row dropped, and the 3-D FP8 dequant branch restricted to fused-expert keys (raises otherwise, so GDN `conv1d.weight` can't hit it).

Not addressed: the **reference accuracy gate** — the dense Qwen3.6-27B Tier 1 entry doesn't have one either. All 8 `qwen3.6-27b` targets are `accuracy: {}`; its e2e entry is `traced_128` + `determinism_128` + vision only; `text_demo.py` has no teacher-forcing case to wire in (unlike determinism); and `test_model_tp.py` is TT-vs-TT (paged contract vs bespoke), not an HF reference.

So it's a pre-existing gap across the whole qwen36 stack, not a regression from this PR, and closing it means a new demo accuracy case plus measured top-1/top-5 targets for both 27B and 35B-A3B. That's dense-side work as much as MoE-side, so it belongs in a separate follow-up task rather than in this PR.

sott0n:
@atupe-tt could you review it?

atupe-tt (approved):

sott0n:
@yieldthought could you review my update?

models/demos/blackhole/qwen36/tt/moe/decode.py
```diff
@@ -0,0 +1,161 @@
...
+        sparsity=sparsity,
+        nnz=nnz,
+        memory_config=ttnn.L1_MEMORY_CONFIG,
+        output_tile=output_tile,
+        program_config=gate_up_config,
+        dtype=ttnn.bfloat16,
+    )
+    sm2 = up_gate.shape[-1]  # 2 * intermediate
+    up_gate = ttnn.reshape(up_gate, (batch_size, num_experts, 1, sm2))
+    up_gate = ttnn.transpose(up_gate, 1, 2)  # (batch, 1, num_experts, sm2) — keep 4D for ttnn.swiglu
```
yieldthought: Codex: **[🔴] Correctness: Preserve expert/user order in batched decode (P1)**

The sparse matmul returns [1,1,1,E,B,2I], so this reshape exchanges expert and user positions when B > 1. For E=64, B=8, expert 1 / user 0 reaches its down-projection with expert 0 / user 1's activation. A routed-only CPU probe reproduces cross-user output dependence; the current full-MoE PCC test includes the shared expert and does not isolate this branch.

Suggestion: explicitly permute the expert/user axes before the existing activation/down path, and cover routed-only batched-versus-single-user equivalence. This replacement passes the logical CPU probe for B=1,2,8,32; TTNN device validation remains necessary.

```suggestion
    up_gate = ttnn.reshape(up_gate, (1, num_experts, batch_size, sm2))
    up_gate = ttnn.permute(up_gate, (2, 0, 1, 3))  # [batch, 1, experts, 2 * intermediate]
```

.github/time_budget.yaml
```diff
@@ -348,7 +348,7 @@ models:
     wh_n150: 28
     wh_galaxy_perf: 215
     bh_p150: 27
-    bh_quietbox_2: 260
+    bh_quietbox_2: 295
```
yieldthought: Codex: **[🔴] CI/CD: Cover the registered Tier 1 E2E total (P1)**

The new 45-minute entry raises the models / bh_quietbox_2 E2E total to 299 minutes. The checked-in validator exits 1 against 295, while the PR base passes at 254/260. This check runs before model/SKU filtering, so the normal Tier 1 E2E workflow stops before tests are scheduled.

Suggestion: cover at least the current total below, or reduce a job timeout using measured runtime. The 299-minute draft passes the same validator.

```suggestion
    bh_quietbox_2: 299
```

yieldthought:
Codex review:

**[🔴] Correctness: Preserve user/expert order in batched decode (P1)**

[models/demos/blackhole/qwen36/tt/moe/decode.py:111](https://github.com/tenstorrent/tt-metal/blob/f7525ba3e2405a72705111d1cfa6669d9761bd36/models/demos/blackhole/qwen36/tt/moe/decode.py#L111) — With input [1,1,B,H], the first sparse matmul returns [1,1,1,E,B,2I]. Reshaping it directly to [B,E,1,2I] reinterprets expert-major data as user-major; the later transpose sends another user/expert's activation to the down-projection. For E=64, B=8, expert 1 / user 0 receives expert 0 / user 1's activation.

**Why it matters:** B=1 works, but batched decode can change one user's output when another user's input changes. A CPU execution of this function using the inspected logical operation contracts reproduces that dependency. The existing B=8 test measures the sum with the shared expert, so its passing aggregate PCC does not establish routed-expert correctness.

Suggestion: preserve the expert/user axes, and add a routed-only B=8 versus independent B=1 check:

```python
    up_gate = ttnn.reshape(up_gate, (1, num_experts, batch_size, sm2))
    up_gate = ttnn.permute(up_gate, (2, 0, 1, 3))  # [batch, 1, experts, 2 * intermediate]
```

**[🔴] CI/CD: Cover the registered Tier 1 E2E timeout total (P1)**

[.github/time_budget.yaml:351](https://github.com/tenstorrent/tt-metal/blob/f7525ba3e2405a72705111d1cfa6669d9761bd36/.github/time_budget.yaml#L351) — The new 45-minute job raises the models / bh_quietbox_2 E2E total to 299 minutes, above this 295-minute budget. Running `verify_time_budget.py tests/pipeline_reorg/models_e2e_tests.yaml .github/time_budget.yaml e2e 1` exits 1; the PR base passes at 254/260.

**Why it matters:** The normal Tier 1 E2E workflow checks the full budget before applying model/SKU filters and will stop before scheduling tests.

Suggestion: cover at least the current 299-minute total, or reduce a timeout using measured runtime:

```yaml
    bh_quietbox_2: 299
```

The [previously discussed reference-accuracy gap](https://github.com/tenstorrent/tt-metal/pull/53917#issuecomment-5474055724) remains: the new E2E cases check generation length, determinism, and performance; both accuracy targets are empty. This remains a review limit requiring an explicit follow-up decision.

Reviewed head: f7525ba3e2405a72705111d1cfa6669d9761bd36. Validation here was source inspection and CPU checks; no device tests were run.

sott0n:
@yieldthought addressed your comments.

> The https://github.com/tenstorrent/tt-metal/pull/53917#issuecomment-5474055724 remains: the new E2E cases check generation length, determinism, and performance; both accuracy targets are empty. This remains a review limit requiring an explicit follow-up decision.

I created an issue as follow-up: https://github.com/tenstorrent/tt-metal/issues/55834

yieldthought (approved):

yieldthought:
Thanks @sott0n, good work!

PR #53923, opened 2026-08-21
Implement retile for block float data types

sankarmanoj-tt:
### PR Category
Feature

### Summary
Implement support for tilize for all block data types and tile sizes.
Added padding to the end of tiles in block data types where tile height < 32, so that the tile sizes are DRAM aligned.
Updated unpack tilize LLKs to support tile height < 32

### Notes for reviewers
LLK changes are done by AI. Please review them carefully. 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:smanoj/retile_block_dtype)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:smanoj/retile_block_dtype)


tests/ttnn/utils_for_testing.py
```diff
@@ -58,6 +58,20 @@
 NP_INTEGER_DTYPES = [np.byte, np.int16, np.int32, np.int64, np.uint16, np.uint32, np.uint64]
 
 
+def dtype_supports_tiny_tile(dtype):
+    return True
```
bbradelTT: This seems superfluous. Is there any data type that does not support tiny tiles?

bbradelTT (commented):
Please comment on or address AI / reviewdog comments.

tests/ttnn/utils_for_testing.py
sankarmanoj-tt: This is a legacy function, from when BFloat4 & BFloat8 didn't support tiny tiles. This PR adds that functionality by padding the tiles so that they are DRAM aligned.

tt_metal/impl/tensor/tensor_apis.cpp
```diff
@@ -260,17 +260,19 @@ HostTensor to_row_major_layout_impl(const HostTensor& tensor) {
...
+    // Preserve the source tile in the row-major PageConfig so later host-side helpers
+    // (e.g. unpad_from_tile) still know the original tile geometry.
+    auto tile = tensor.tensor_spec().tile();
     // Construct the new tensor spec first to verify that this is a supported Tensor configuration
     TensorSpec new_tensor_spec(
         tensor.logical_shape(),
         TensorLayout::fromPaddedShape(
             tensor.dtype(),
-            PageConfig(Layout::ROW_MAJOR),
+            PageConfig(Layout::ROW_MAJOR, tile),
```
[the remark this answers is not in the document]
sankarmanoj-tt: This PR doesn't add functionality for on host tiny tile to RM. But adding this arg ensures that it asserts and throws and error instead of failing silently.

tests/ttnn/utils_for_testing.py
```diff
@@ -58,6 +58,22 @@
...
+
+def select_tile(*dtypes, layout=ttnn.TILE_LAYOUT):
+    if layout == ttnn.TILE_LAYOUT:
+        # return ttnn.Tile((16, 32))
+        if all(dtype_supports_tiny_tile(dtype) for dtype in dtypes):
+            print("Using tiny tile 16 x 32")
+            return ttnn.Tile((16, 32))
+        else:
+            print("Using tile 32 x 32")
+            return ttnn.Tile((16, 32))
```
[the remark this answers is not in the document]
vtsilytskyiTT: Agree with copilot, numbers don't match.

tt_metal/impl/data_format/bfloat2.cpp
```diff
@@ -134,8 +134,9 @@ std::vector<float> unpack_bfp2_tiles_into_float_vec(
 
     uint32_t num_exp_words = tt::round_up(num_faces * face_H, l1_alignment) / num_exps_in_dword;
     uint32_t num_tile_words = tile_HW / num_elements_in_dword;
-    int num_bfp_dwords_in_tile = num_tile_words + num_exp_words;
     int num_dwords_per_row = subtile_cols / num_elements_in_dword;
+    // Per-tile stride in 32-bit words, including trailing DRAM-alignment padding after the mantissa section.
+    uint32_t words_per_tile = single_bfp_tile_size / 4;
```
vtsilytskyiTT:
```suggestion
    const uint32_t words_per_tile = single_bfp_tile_size / 4;
```
I'd appreciate `const` as it shows to the reader that value stays untouched. Readability improvement, nothing more.

tt_metal/impl/data_format/bfloat2.cpp
```diff
@@ -152,13 +153,13 @@ std::vector<float> unpack_bfp2_tiles_into_float_vec(
...
-                        int tile_and_data_index = data_index + (num_bfp_dwords_in_tile * tile_index);
+                        int tile_and_data_index = data_index + (words_per_tile * tile_index);
 
                         int exponent_index =
-                            (data_index >> data_dwords_per_exp_dword_log2) + (num_bfp_dwords_in_tile * tile_index);
+                            (data_index >> data_dwords_per_exp_dword_log2) + (words_per_tile * tile_index);
                         exp_word = bfp_tiles[exponent_index];
 
-                        int num_exponent_words_skip = tile_index * num_exp_words;
+                        int num_exponent_words_skip = tile_index * (words_per_tile - num_tile_words);
```
vtsilytskyiTT: Values `words_per_tile` and `num_tile_words` sounds the same. Probably names are not specific enough.

tt_metal/impl/data_format/bfloat4.cpp
vtsilytskyiTT, line 1: Same comments as in bfloat2.cpp.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -1079,6 +1079,10 @@ def test_tilize_retile(device, tensor_shape, shard_layout, input_tile_shape, out
     if input_tile_shape[0] == output_tile_shape[0]:
         pytest.skip("Input and output tile shapes are the same")
 
+    # bfloat8_b is supported at every tile height (1..32): the packer LLK sizes the BFP exponent
+    # section from face_r_dim, so partial-face (tile height < 16) block-float tiles pack with the
+    # same layout the host and unpacker use.
+
```
vtsilytskyiTT: Redundant comment. Test with `bfloat8_b` parameter implies that given type is supported. You can keep this comment in the PR description or remove entirely.

scripts/run_split_pytest.sh
vtsilytskyiTT, line 1: I don't see how the script is used in this PR.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
```diff
@@ -94,9 +100,12 @@ ProgramDescriptor TilizeMultiCoreShardedRetileProgramFactory::create_descriptor(
...
 
-    const uint32_t ratio = shrink ? (in_tile_height / out_tile_height) : (out_tile_height / in_tile_height);
-    // One output block occupies `ratio` input tile-rows of RM in the grow case, one otherwise.
-    const uint32_t mid_pages_per_out_block = (shrink ? 1u : ratio) * tiles_per_block;
+    // Compute untilizes the whole shard into mid, then tilizes it. Aliased c_1/c_2 page sizes can
+    // differ, so the allocation must be a multiple of both.
+    const uint32_t mid_input_pages = num_tiles_per_shard_in;
+    const uint32_t mid_size_align = std::lcm(mid_input_page_size, mid_output_page_size);
+    const uint32_t mid_total_size =
+        ((mid_input_pages * mid_input_page_size + mid_size_align - 1) / mid_size_align) * mid_size_align;
```
vtsilytskyiTT: Try `tt::round_up`, it does exactly what you need.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
vtsilytskyiTT: Comment is stale, there is no c_1 and c_2 CBs.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
```diff
@@ -128,19 +137,19 @@ ProgramDescriptor TilizeMultiCoreShardedRetileProgramFactory::create_descriptor(
     // program-creation time: c_1 carries the input tile shape for pack_untilize to write into, c_2
     // the output tile shape so llk_unpack_tilize reads the correct number of RM rows.
     desc.cbs.push_back(CBDescriptor{
-        .total_size = 2 * mid_pages_per_out_block * mid_page_size,
+        .total_size = mid_total_size,
```
vtsilytskyiTT, line 160: Why did we have `2 *` multiplier here? Was it "two views over one shared intermediate L1 region" or buffering multiplier? If the second one, it can stay.

vtsilytskyiTT (changes requested):
Please address AI comments as well.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_retile_program_factory.cpp
```diff
@@ -107,8 +111,15 @@ ProgramDescriptor TilizeMultiCoreRetileProgramFactory::create_descriptor(
...
-    const uint32_t mid_pages_per_out_block = (shrink ? 1u : ratio) * tiles_per_block;
+    // The compute kernel untilizes the whole per-core assignment into mid, then tilizes it.
+    // Size mid for the busiest core's input tile-rows. Aliased c_1/c_2 page sizes can differ
+    // (e.g. bf16→bfp8 with different tile heights), so the allocation must be a multiple of both.
+    const uint32_t max_blocks = std::max(nblocks_per_core, nblocks_per_core_cliff);
+    const uint32_t max_input_rows_per_core = shrink ? max_blocks : max_blocks * ratio;
+    const uint32_t mid_input_pages = max_input_rows_per_core * tiles_per_block;
+    const uint32_t mid_size_align = std::lcm(mid_input_page_size, mid_output_page_size);
+    const uint32_t mid_total_size =
+        ((mid_input_pages * mid_input_page_size + mid_size_align - 1) / mid_size_align) * mid_size_align;
```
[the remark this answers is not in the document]
sankarmanoj-tt: This is not needed because the op currently uses the minimum possible size for CBs.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
sankarmanoj-tt: Its for double buffering.

scripts/run_split_pytest.sh
sankarmanoj-tt: Removed it.

tt_metal/impl/data_format/bfloat2.cpp
sankarmanoj-tt: Renamed the variable

tests/ttnn/utils_for_testing.py
sankarmanoj-tt: Remvoed this and the select tile function.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
sankarmanoj-tt: Fixed this.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_sharded_retile_program_factory.cpp
sankarmanoj-tt: Done.

vtsilytskyiTT (approved):
Overall OK. I still don't see a reason to remove double buffering multiplier.

ncvetkovicTT (commented):
Looked at the LLK side. The `exp_section_size` formula is right and it fixes a real bug. Three asks inline.

Two things worth putting in the description:

- The formula is exactly what the unpacker computes. Per [UNPACR_Regular.md](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/UNPACR_Regular.md), `NumExponents = ceil(XDim*YDim*ZDim*WDim/16)` and the section is `ceil(NumExponents/16)*16` bytes. With `x_dim = face_r_dim*FACE_C_DIM` and `z_dim = num_faces` (`cunpack_common.h:856-858`) that's your expression. Host agrees too (`tile.cpp:71`).
- `partial_face` is set for any tile height < 32 (`tile.cpp:66`), so the old `partial_face ? 1 : num_faces` was already correct for 8x32, 4x32, 2x32 and 1x32. 16x32 is the only shape that actually changes (32 exponent bytes: old 1, correct 2). So the `TT_FATAL` you removed was just too strict. That's a better story than "updated unpack tilize LLKs", since there are no unpack changes here and none are needed.

Two blockers on the way: llk-smoke is red on both archs with `cpack_common.h:391: error: unused parameter 'partial_face'`, so that param should go. And `test_tilize_retile` hasn't run in any CI job on this branch, the green data movement group only runs `test_tilize_test` and `test_tilize_fp32_lossless`.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -1063,14 +1063,15 @@ def test_tilize_row_major_to_tiny_tile(device, tensor_shape, shard_layout, tile_
...
+        ([1, 1, 1, 1024], ttnn.TensorMemoryLayout.WIDTH_SHARDED),
         ([1, 1, 32, 1024], ttnn.TensorMemoryLayout.WIDTH_SHARDED),
         ([1, 1, 1024, 32], ttnn.TensorMemoryLayout.HEIGHT_SHARDED),
         ([1, 1, 256, 256], ttnn.TensorMemoryLayout.BLOCK_SHARDED),
     ],
 )
 @pytest.mark.parametrize("input_tile_shape", [(32, 32), (16, 32), (8, 32), (4, 32), (2, 32), (1, 32)])
 @pytest.mark.parametrize("output_tile_shape", [(32, 32), (16, 32), (8, 32), (4, 32), (2, 32), (1, 32)])
-@pytest.mark.parametrize("dtype", [ttnn.bfloat16])
+@pytest.mark.parametrize("dtype", [ttnn.bfloat16, ttnn.bfloat8_b, ttnn.bfloat4_b])
```
ncvetkovicTT: **Ask 1/3: can we get an LLK test for this on Blackhole?**

For a BFP input the intermediate CB is bfloat16, so `IS_8BIT_FORMAT` is false (`ckernel_defs.h:236-244`) and `tilize_init` takes the real BH workaround path: `llk_pack_init<PackMode::Tilize>` (`tilize.h:61`) into `llk_pack.h:570`. That means the exponents get written by the de-interleave MOP (`llk_pack.h:158-215`), not the Default one. Your fix sets the config field right either way, but that MOP has never run with a BFP output at anything other than a full 32x32 tile.

There's a cheap test sitting right there. `test_unpack_tilize.py:60-61` skips Bfp8_b output unless `num_faces == 4`, and `num_faces == 2` at `face_r_dim == 16` is the 16x32 tile you're fixing. Try deleting that skip. `unpack_tilize_test.cpp:129` passes `partial_face = false`, so it already configures `exp_section_size = 2` for that shape, which is the value your formula now produces from metal. If it passes, that's your proof. If it doesn't, something else is wrong at `num_faces == 2` and this PR is standing on it.

After that, a `face_r_dim` axis over {1, 2, 4, 8} with Bfp8_b/Bfp4_b out would cover the rest of what you're turning on. The mode plumbing is already there (`unpack_tilize_test.cpp:82`), only the shape parameter is missing.

tt_metal/tt-llk/tt_llk_blackhole/llk_lib/llk_pack.h
```diff
@@ -444,14 +444,15 @@ inline void _llk_pack_reconfig_data_format_(
     const std::uint32_t tile_size,
     const std::uint32_t tile_c_dim = TILE_C_DIM,
     const std::uint32_t num_faces  = 4,
-    const bool partial_face        = false)
+    const bool partial_face        = false,
+    const std::uint32_t face_r_dim = FACE_R_DIM)
```
ncvetkovicTT, line 449: **Ask 2/3: can `face_r_dim` go in the same slot on both archs?**

Here it's last (7th). On WH it's 4th (`tt_llk_wormhole_b0/llk_lib/llk_pack.h:311-318`). Same API, different position, and the `_llk_*_` layer is supposed to be arch-parallel.

Same thing in `set_packer_config`, where it goes in 3rd, ahead of `num_faces`. Nothing breaks today since every caller is internal, but inserting into the middle of a defaulted list means a future positional `num_faces` silently lands in `face_r_dim`. Appending avoids that.

I'm asking because I want to add a `_llk_pack_tilize_` next to `_llk_unpack_tilize_` (WH forwards to Default, BH uses the Tilize path) so the `#ifdef ARCH_BLACKHOLE` at `tilize.h:60-62` can go away. That only pays off if it's symmetric, and it's much easier to fix now than once it's precedent.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/kernels/compute/retile.cpp
```diff
@@ -60,77 +64,49 @@ void kernel_main() {
...
+    mid.wait_front(total_mid_pages);
+    uint32_t block_rd_ptr = 0;
+    UNPACK({ block_rd_ptr = get_local_cb_interface(mid_cb).fifo_rd_ptr; })
+
+    // Reconfigure the unpacker/packer from the untilize config (src_cb/mid_cb) to the tilize
+    // config (mid_view_cb/out_cb). tilize_init's state_configure is sentinel-only, so the
+    // hardware reconfig must be explicit — for bf16 it's a no-op, for bfloat8 it's required.
+    reconfig_data_format_srca(src_cb, mid_view_cb);
+    pack_reconfig_data_format(mid_cb, out_cb);
+    tilize_init(mid_view_cb, tiles_per_block, out_cb);
```
ncvetkovicTT, line 102: **Ask 3/3: just a heads-up, no change needed.**

This is where `PackMode::Tilize` gets picked on BH, and I'm about to rework how that choice is made. Today it's derived from `unpack_src_format` in three separate places (`llk_pack_tile_api.h:32`, `llk_math_unary_datacopy_api.h:60-61`, `experimental/llk_pack_custom_api.h:37`), threaded through as `skip_bh_tilize_workaround` into a runtime branch at `llk_pack.h:570`, with the arch split sitting in an `#ifdef` in a shared header.

Your diff has no `PackMode`, no `skip_bh_tilize_workaround` and no `IS_8BIT_FORMAT` in it, which is exactly what I want. Please keep it that way and I'll stay off the config registers you're touching. Land yours first and I'll rebase on top.

Same reason as ask 1, really: the coverage you add is what I'd verify the refactor against.

sankarmanoj-tt:
@ncvetkovicTT I've updated the Wormhole LLKs to match. I've updated the sanity tests to have a small unit test for retile.

ncvetkovicTT:
Thanks @sankarmanoj-tt. The WH signatures look right, and moving WH to match BH rather than the other way round is fine by me. Ask 3 is respected too, no `PackMode` / `skip_bh_tilize_workaround` churn in here.

Still open on the LLK side:

1. **llk-smoke is still red on both archs, same error as before**: `cpack_common.h:391: error: unused parameter 'partial_face' [-Werror=unused-parameter]`. WH `cache_exponential_section_sizes_in_gprs` still takes it and never uses it, so just drop the parameter.

2. **The sanity case doesn't exercise the fix.** Both bfloat8_b cases in the new `test_tilize_retile` are `(1,32) -> (32,32)`, so every BFP *pack* in CI runs at `face_r_dim=16, num_faces=4, partial_face=false` - the one geometry where `bfp_exp_section_chunks` returns exactly what the old code returned. There's no bfloat4_b and no BFP case with a tiny output tile. Could you add `(32,32) -> (16,32)` in bfloat8_b plus one bfloat4_b case? 16x32 out is the only shape whose `exp_section_size` actually changes, so it's the case that matters. (`test_tilize_retile_full` has the coverage but isn't in the yaml.)

3. Both retile tests still carry `@skip_for_wormhole_b0`, so the WH LLK changes in this commit aren't exercised on WH at all.

4. **An LLK-level test.** Now that the wrapper forwards `face_r_dim`, the cheap version from my earlier comment should work: `test_unpack_tilize.py:60-61` skips Bfp8_b output unless `num_faces == 4`, and `num_faces == 2` at `face_r_dim == 16` is the 16x32 tile you're fixing. Try deleting that skip. If it passes, that's the cleanest proof this is right.

Not my area to gate on, but the llk-pr-review comments on `retile.cpp`, `tilize_device_operation.cpp` and the two program factories are unanswered and those four files are unchanged since the last push. The `num_row_bands = in_total_rows / in_shard_height` one (sharded factory:103) reads like a real regression - worth at least a reply.

What I need to approve: PR Gate green, llk-smoke green on both archs, and a link to a run where the retile test actually ran with a tiny BFP output tile. Ping me when smoke is green and I'll re-review.

PR #53019, opened 2026-08-13
[Feature] Add publish_run: one run's CSVs -> one typed run.parquet

nstojicTT:
publish_run is the go-forward CI publish step. It gathers a perf sweep's combined per-test CSVs, stamps run provenance (commit_sha / arch / run_id / timestamp / pipeline / pr_number) from CLI args + CI env, and reuses the tested convert_csvs_to_parquet converter to emit one Parquet per arch per run.

**note**
If the parser.parse_args is called with argv=None, default behaviour is that it reads sys.argv which 
    come from command line. In this case, CI parses them via step that will be introduced in the next PR
    that basically calls: 
    
 run: |
  python3 -m helpers.perf.publish_run \
    --csv-dir ../../perf_data \
    --out ../../perf_data/run-${{ env.CHIP_ARCH }}-${{ matrix.test-group.split_group }}.parquet \
    --arch ${{ env.CHIP_ARCH }}

 Then these arguments come from github actions and/or previous steps like set CHIP_ARCH or matrix build...

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-publish-run)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-publish-run)

PR #51951, opened 2026-08-03
[Feature] Historical CSV→Parquet migration

nstojicTT:
perf_migrate reuses the live converter (convert_csvs_to_parquet) so migrated historical data lands in the same shared schema as live runs. migrate_runs converts each run's CSVs into one immutable Parquet batch, lenient (never fails on dirty historical data) with a per-run coverage report of dropped columns and coerced values. Deterministic (sorted processing, folder-derived provenance) and idempotent (skips already-migrated runs). discover_runs enumerates runs from an archive tree, parsing arch from the folder and provenance from an optional run_meta.json sidecar.

On the real nightly corpus: 10 runs migrated clean, 47.97 MB CSV -> 2.92 MB Parquet, 0 columns dropped, 0 values coerced.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-migrate)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-migrate)

PR #53928, opened 2026-08-21
One Parquet writer per run, named from the run tag

nstojicTT:
Stacked on #53660 — review that first; this PR's diff is the last commit only.

## The bug

Every perf run writes the same data to Parquet **twice**, from two independent
code paths, and both copies ship in the nightly artefact.

- `_write_run_parquet`, called at the end of `combine_perf_reports`, writes one
  automatically on every run.
- A CI step then calls `helpers.perf.publish_run` over the same CSVs and writes
  a second.

Verified against last night's nightly (`32327710583`). The
`perf-data-wormhole-1` artefact contains both `32327710583.parquet` and
`run-wormhole-1.parquet`: 21,702 × 81 rows each, identical column list,
numerically equal. Only the timestamps differ, by the four seconds between the
two writes.

## How it happened

Not a deliberate belt-and-braces. Both landed on 2026-08-14, ninety minutes
apart:

| time | PR | what |
|---|---|---|
| 10:18 | #51951 | Historical migration — **also** added the automatic writer, as a side effect |
| 10:19 | #53019 | Added `publish_run.py` |
| 11:40 | #53132 | Fixed a broken import in the automatic writer |
| 11:48 | #53020 | Added the CI step calling `publish_run` |

The automatic writer was easy to miss because it was **silently broken** for
that window: its lazy import still named the pre-reorg `.perf_parquet` module,
and the `except Exception` around it swallowed the `ModuleNotFoundError`. No
second file appeared until #53132 fixed the import.

## Why it matters now

These files are about to be published over SFTP into one archive for the data
team. Two files per shard means two loads of every row — and `ROW_KEY`
(`test_name`, `commit_sha`, `arch`, `run_id`) identifies *thousands* of rows,
not one, so nothing downstream can distinguish a duplicate load from real data.

## Options considered

**1. Drop the CI step, keep the automatic writer.** ✅ chosen

**2. Drop the automatic writer, keep the CI step.** Rejected: a developer
running a sweep locally would get no Parquet, so what CI publishes could not be
reproduced by hand. That property is what made the recent noise baselines
checkable.

**3. Keep both, upload only one.** Rejected: it leaves a trap. A later recursive
glob picks up both, and the reason only one is uploaded would live in an sftp
batchfile, far from the code that writes them.

## What this PR does

- `_write_run_parquet` names its file from the **run tag** instead of `run_id`.
  `run_id` is a ROW_KEY column and is shared by every shard of one workflow by
  design, so all ten shards were producing the same filename — harmless while
  each stays in its own `runs/<tag>/`, wrong once they are collected into one
  archive. `perf_run_tag()` already documents itself as "purely a filesystem
  concern", which is exactly what a filename is. **`run_id` is unchanged as a
  column.**
- Removes the "Publish Parquet from CSVs" CI step and drops
  `perf_data/*.parquet` from the zip glob, since the Parquet now lives inside
  `perf_data/runs/<tag>/`.
- Updates the test that pinned `testrun.parquet` to `testrun-wormhole-0.parquet`.

`publish_run` is **kept**. It is the tested CLI for republishing an existing CSV
directory, which the historical migration needs. It just no longer runs in the
nightly path.

## Follow-up, not in this PR

SFTP publishing. The `latest` symlink added by #53660 gives it a static path
that needs no knowledge of the tag:

```
lcd tt_metal/tt-llk/perf_data/latest/
mput -f *.parquet
ls -hal *.parquet
```

That is a batchfile, a step using the existing `.github/actions/upload-data-via-sftp`
action, and three secrets — waiting on credentials.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-single-parquet-writer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-single-parquet-writer)


ndivnicTT (approved):
LGTM!

fvranicTT (approved):

tt_metal/tt-llk/tests/python_tests/test_perf_report_hw_free.py
```diff
@@ -441,7 +441,8 @@ def test_combine_perf_reports_emits_parquet_alongside_csv(tmp_path, monkeypatch)
     # ...reachable through the stable `latest` path...
     assert (root / "perf_data" / "latest" / "perf_x" / "perf_x.csv").exists()
     # ...and a run-level Parquet batch alongside it.
-    parquet = run_dir / "testrun.parquet"
+    # Named from the run tag, not run_id: run_id is shared by every shard.
+    parquet = run_dir / "testrun-wormhole-0.parquet"
```
fvranicTT, line 445: This is just for testing?

nstojicTT: yes, its just a tests stub value
real value would look something like this 32327710583-wormhole-3.parquet
I can change the "testrun" to be more descriptive

dpopovTT (approved):

PR #53947, opened 2026-08-21
Gemma-4 12B, 31B: WH LB migration

ign-rahultp:
## Summary

Migration and optimization of **Gemma-4 12B and 31B on Wormhole T3K (1×8)**:
decode/prefill layout, TP collectives and precision defaults retuned for the
Wormhole mesh, a per-module PCC suite gating both variants against HF on real
checkpoint weights, and CI coverage for them on T3K.

Refer [gemma-4/README.md](https://github.com/tenstorrent/tt-metal/tree/ign/gemma4_support_loudbox_exps/models/demos/gemma4#readme) for more information.

### Performance

1. **Decode stays in L1** — removed the per-layer DRAM bounces: input LayerNorm,
   GeGLU group, o_proj/down_proj writes, all-reduce output, residual stream,
   q/k/v norm island, matmul in0. Also dropped per-layer ops that returned their
   input unchanged.

2. **TP collectives** — Ring topology on Wormhole for dense models; all-reduce
   split with corrected reduce-scatter workers; async RS+AG for tall prefill,
   bit-exact; height-aware RS defaults; Fabric packet-size default; persistent
   buffers and L1_SMALL for gather semaphores.

3. **Tuned matmul program configs** — fused-QKV and narrow-N decode QKV, SharedMLP
   gate_up/down_proj, LM head prefill, bfp8 + fp32 dest-acc on the decode lm_head,
   cutoff-reshape 2D for long prefill. Tuned configs fall back to `auto` instead of
   overflowing L1; o_proj and LoFi tall-prefill are opt-in (`GEMMA4_OPROJ_TUNED`,
   `GEMMA4_PREFILL_MATMUL_LOFI`), since the tuning is shape-specific.

4. **On-device sampling is now the default** — reverses item 6 of the previous PR.
   Shard-width cap raised so 1×1/1×2 meshes use TTSampling; vocab AllGather skipped
   on device-sampled prefill; token readback pipelined. Measured token-for-token
   identical to host sampling; `GEMMA4_HOST_SAMPLE=1` restores the host path.

5. **Prefill tracing and long context** — traced single-chunk 4k demo prefill and
   auto-traced multi-chunk prefill on long unbounded demos; CH2D traced inputs;
   page-table seams without `torch.cat`; optional chunk-span warmup; WH trace region
   sized for the cumulative footprint.

6. **RMSNorm sharding** — sharded fast path extended to prefill, AR→LN→MLP kept
   width-sharded, short-prefill LN island, all behind `GEMMA4_SHARDED_NORM` /
   `GEMMA4_NORM_KEEP_SHARDED`. The L1-resident GeGLU intermediate was reverted for
   prefill — it OOMs L1 at low TP.

7. **Fewer host round-trips and faster init** — decode state stays on device; four
   round trips removed from demo v2; RoPE caches built without a full-context dummy;
   host embed skipped on non-PLI prefill; HF-hub-id loads read cached safetensors.

8. **Precision defaults** — decode SDPA and short-ISL prefill matmuls pinned to
   HiFi4, prefill SDPA to `hifi4_nodest`, fp32 dest-acc on the decode lm_head,
   precision-override lookup aligned.

### Testing

- **New decoder PCC suite against real checkpoint weights**:
  `tests/unit/test_prefill.py` (ISL sweep, one decoder layer),
  `tests/unit/test_decode.py` (multi-step decode gating the KV write, including
  the bounded sliding ring), sharing `tests/unit/decoder_pcc_common.py`.
- **New `tests/e2e/`**: teacher-forced token accuracy + logit PCC
  (`test_teacher_forcing_e2e.py`) and the ISL sweep (`test_isl_sweep.py`).
  Teacher-forcing scores through the demo's own `_host_sample`, so top-1 picks
  use the same sampler the demo uses.
- Dedicated token-embedding tests; RMSNorm and the module tests PCC'd against HF
  on the checkpoint's real scale; the lm_head weight read from shards instead of
  materializing a full HF model.

### Measured performance — Wormhole T3K (1×8)

`text_demo_v2` buckets, selected with `-k`:

| ISL bucket (`-k`) | ISL | Batch | 12B TTFT (ms) | 12B ms/tok | 12B tok/s/user | 31B TTFT (ms) | 31B ms/tok | 31B tok/s/user |
|---|---:|:-:|---:|---:|---:|---:|---:|---:|
| `batch-1`           |   128 |  1 | **58.5** | **26.21** | **38.15** | **93.5** | **42.88** | **23.32** |
| `batch-8`           |   128 |  8 | **4706.8** | **34.21** | **29.23** | **6977.7** | **57.12** | **17.51** |
| `batch-32`          |   128 | 32 | **13452.1** | **45.37** | **22.04** | **13801.5** | **71.82** | **13.92** |
| `long-context-4k`   |    4k |  1 | **841.6** | **27.62** | **36.21** | **2032.2** | **45.05** | **22.20** |
| `long-context-32k`  |   32k |  1 | **7229.8** | **30.48** | **32.81** | **15067.5** | **48.68** | **20.54** |
| `long-context-64k`  |   64k |  1 | **14750.2** | **33.66** | **29.71** | **32943.4** | **52.74** | **18.96** |
| `long-context-128k` |  128k |  1 | **28925.7** | **38.01** | **26.31** | **61584.4** | **58.15** | **17.20** |
| `long-context-256k` |  256k |  1 | **105341.0** | **55.33** | **18.07** | **241697.9** | **80.08** | **12.49** |

### CI

12B T3K unit promoted to tier 2 (matching 26B/31B); the 12B unit suite runs on
Wormhole; teacher-forcing e2e and the 12B/31B ISL sweep added as T3K jobs;
`time_budget.yaml` and the pipeline-reorg yamls updated to match.

Link to the CI: https://github.com/tenstorrent/tt-metal/actions/runs/32690156816 (Note: the tests for 12B and 31B variants are passing)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/gemma4_support_loudbox_exps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/gemma4_support_loudbox_exps)


viktorpusTT:
Please share CI links in PR desc.

ign-rahultp:
Updated PR description with CI links.

yieldthought:
Codex review:

1. **[P1] Keep decode-token feedback when sampling runs eagerly.** [models/tt_transformers/tt/generator.py:2207](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/tt_transformers/tt/generator.py#L2207) gates `tt_out_tok` on the sampling trace. The new batch check disables that trace whenever the decode batch differs from the sampler's padded maximum (e.g. B=1 versus 32). Steady device-sampled decode also skips host token restaging, so the next decode consumes the old token while its position advances. The Gemma-only override repairs this locally, leaving other shared-Generator users exposed. A source-extracted base/head probe confirms that this case changes from feedback present to feedback absent. Suggestion: preserve feedback whenever the decode trace relies on it, independently of whether sampling itself is traced.

2. **[P1] Use the per-DP batch size for reset trace lookup.** [models/tt_transformers/tt/generator.py:1634](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/tt_transformers/tt/generator.py#L1634) builds the key before `torch.chunk`, whereas capture and replay key traces by `tokens[0].shape[0]`. With B=64 and DP=2, reset looks for `(True,64)` while the captured trace is under `(True,32)`. The lookup therefore skips recovery of authoritative device tokens/positions and restages the one-step-stale host state under async scheduling. The base/head host probe reproduces this control-flow change. Suggestion: derive the key after splitting, using the same per-DP batch size as capture.

3. **[P1] Preserve bounded sliding-cache ownership across slot moves.** [models/demos/gemma4/tt/generator_vllm.py:1485](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/demos/gemma4/tt/generator_vllm.py#L1485) replaces scheduler block IDs with fixed ranges based solely on the current row. If request A finishes in row 0 and B moves from row 1 to row 0, `slot_remap` moves B's token/position, but the remapper assigns A's old sliding-cache blocks to B; no corresponding K/V migration occurs. The source-extracted mapping probe confirms that head discards incoming row permutations that base preserved. Suggestion: maintain cache-slot ownership through request admission/compaction, or migrate the bounded K/V together with each slot move. Validate a continuing request against a no-compaction reference.

4. **[P1] Bound sharded RMSNorm by the actual per-core footprint.** [models/demos/gemma4/tt/rms_norm.py:335](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/demos/gemma4/tt/rms_norm.py#L335) extends the path to 1024-row prefill based on height alone. For 26B's hidden size 2816, the selected WH grid has only eight cores, so BF16 input and output each need 720,896 bytes per bank. Their combined 1,441,792 bytes exceed the observed 1,393,472-byte capacity before scratch space. The [exact-head 26B unit job](https://github.com/tenstorrent/tt-metal/actions/runs/32690156816/job/97324235380) fails both the isolated prefill_1024 RMSNorm and sliding-layer tests on this allocation; base kept this prefill shape interleaved. Suggestion: select a layout using the per-core input/output/scratch budget and retain an interleaved path for shapes that cannot fit.

5. **[P1] Exclude short-tail padding from attention and retain live history.** [models/demos/gemma4/tt/attention/prefill.py:213](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/demos/gemma4/tt/attention/prefill.py#L213) pads a short stash with zero K/V; the continuation SDPA consumes those rows without a validity mask. With a 384-token first grant and a 1024-token window, the first continuation query attends 639 artificial rows. For zero Q/K and unit real V, this produces 385/1024 instead of 1.0. A later short continuation also replaces the stash using only its current K/V, discarding prior tokens still inside the window. The new short-first-chunk test checks shapes, so it misses both errors. Suggestion: track valid history, exclude padding from softmax, and form the next tail from the combined valid history and current chunk; compare short-first and consecutive-short continuations numerically against an unchunked reference.

6. **[P2] Disable token-read pipelining when decode tracing is off.** [models/demos/gemma4/demo/text_demo_v2.py:488](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/demos/gemma4/demo/text_demo_v2.py#L488) enables pipelining with device sampling even under `GEMMA4_DECODE_TRACE=0`. That branch never updates host `out_tok`, while untraced decode consumes it every iteration, repeatedly feeding the prefill token. The equivalent loop in `text_demo.py` already includes the required trace guard. Suggestion: require `enable_trace` here as well.

7. **[P2] Run the new real-weight module checks in a checkpoint-backed CI job.** [tests/pipeline_reorg/models_unit_tests.yaml:519](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/tests/pipeline_reorg/models_unit_tests.yaml#L519) explicitly excludes `test_prefill.py`, and both 12B/31B unit jobs use config-only checkpoints that skip the new real-weight module and multi-step decoder tests. The linked green unit logs confirm those skips. The real-checkpoint e2e commands select demo, lm_head, packed verification, full-model and teacher-forcing tests, but none of these new module suites. Suggestion: add the new prefill, multi-step decode and real-weight module cases to the existing real-checkpoint jobs; removing the ignore alone will still leave the checkpoint-dependent skips.

Validation: manual source review at `c0f3d7025c002c6be40d04c00a6fd9740d4689ef`, inspection of the matching CI logs, and host-only source-extracted probes. No new hardware runs were performed. The linked CI also has three PCC failures whose introduction is not established by the available base/head evidence.

models/demos/gemma4/demo/text_demo_v2.py
```diff
@@ -529,40 +477,130 @@ def test_demo_text(
...
+    # nothing from host (see Generator._decode_forward_trace_text: reset_inputs
+    # is False once decode is steady). The host loop then sits one step behind
+    # and EOS is seen one step late — that extra token is discarded below, so
+    # the emitted text is unchanged. GEMMA4_DECODE_PIPELINE=0 restores the
+    # blocking loop.
+    pipeline_reads = device_sampling_params is not None and os.environ.get("GEMMA4_DECODE_PIPELINE", "1").lower() in (
+        "1",
+        "true",
+        "yes",
+    )
```
yieldthought, line 492: Codex: Suggestion: require a decode trace before pipelining token reads, matching text_demo.py. Eager decode needs the updated host token each step.

```suggestion
    pipeline_reads = (
        device_sampling_params is not None
        and enable_trace
        and os.environ.get("GEMMA4_DECODE_PIPELINE", "1").lower() in ("1", "true", "yes")
    )
```

models/tt_transformers/tt/generator.py
yieldthought, line 1: Codex: Suggestion: derive the reset key from the same per-DP batch size used by trace capture.

Suggested replacement for [models/tt_transformers/tt/generator.py:1633–1638](https://github.com/tenstorrent/tt-metal/blob/c0f3d7025c002c6be40d04c00a6fd9740d4689ef/models/tt_transformers/tt/generator.py#L1633-L1638).

```python
        B = tokens.shape[0]

        tokens = torch.chunk(tokens, self.data_parallel, 0)
        start_pos = torch.chunk(start_pos, self.data_parallel, 0)
        page_table = torch.chunk(page_table, self.data_parallel, 0) if page_table is not None else None
        decode_trace_key = (on_device_sampling, int(tokens[0].shape[0]))
```

ign-rahultp:
Hi @viktorpusTT, @yieldthought,

It looks like Gemma-4 12B and Gemma-4 31B now have WH LB support in the main branch (ref: https://github.com/tenstorrent/tt-metal/pull/50648). As a result, rebasing this branch onto the latest main is introducing a significant number of merge conflicts.

In addition to the model migration, this PR also includes some optimisations (better performance compared to main) and additional test coverage.

We are already working on Gemma-4 support for MTP and DFlash on WH LB, and we plan to raise a separate PR for that work soon. Given the overlap, would it make sense to move the optimizations and test coverage from this PR into the upcoming MTP/DFlash PR and address them together there?

Please let us know if that approach works for you.

Regards,
Rahul

viktorpusTT:
> Hi @viktorpusTT, @yieldthought,
> 
> It looks like Gemma-4 12B and Gemma-4 31B now have WH LB support in the main branch (ref: #50648). As a result, rebasing this branch onto the latest main is introducing a significant number of merge conflicts.
> 
> In addition to the model migration, this PR also includes some optimisations (better performance compared to main) and additional test coverage.
> 
> We are already working on Gemma-4 support for MTP and DFlash on WH LB, and we plan to raise a separate PR for that work soon. Given the overlap, would it make sense to move the optimizations and test coverage from this PR into the upcoming MTP/DFlash PR and address them together there?
> 
> Please let us know if that approach works for you.
> 
> Regards, Rahul

My preference is smaller PRs. Maybe separate PR for optimizations? As for test coverage: Best to include it directly in the PR that introduces the feature, next best thing is a separate followup PR soon after.

models/demos/gpt_oss/tt/model.py
```diff
@@ -602,13 +602,14 @@ def ttnn_prefill_forward(
...
 
-    def process_logits_after_prefill_trace(self, logits, last_token_idx):
+    def process_logits_after_prefill_trace(self, logits, last_token_idx, allow_sharded=False):
         """
         Post-process traced prefill output to the 32-token tile containing `last_token_idx`.
 
         Unlike tt_transformers `Transformer`, GPT-OSS `ttnn_prefill_forward` already
         applies final norm + lm_head, so this method only slices logits.
         """
+        del allow_sharded  # Gemma4-only opt-in; API parity with Generator callers
```
mtairum, line 612: Why is this change part of this PR?

models/demos/minimax_m3/tt/model.py
```diff
@@ -486,13 +486,14 @@ def prefill_forward(
...
 
-    def process_logits_after_prefill_trace(self, logits, last_token_idx):
+    def process_logits_after_prefill_trace(self, logits, last_token_idx, allow_sharded=False):
         """
         Post-process traced prefill output to the 32-token tile containing `last_token_idx`.
 
         Unlike tt_transformers `Transformer`, MiniMax-M3 `prefill_forward` already
         applies final norm + lm_head, so this method only slices logits.
         """
+        del allow_sharded  # Gemma4-only opt-in; API parity with Generator callers
```
mtairum, line 496: Why is this part of this PR?

models/common/sampling/tt_penalties.py
mtairum, line 1: why is this PR making changes to sampling penalties?

models/common/sampling/tt_sampling.py
mtairum, line 1: Why is this PR making changes to sampling at all?

models/tt_transformers/tt/generator.py
mtairum, line 1: I'm also confused why we need these changes for this specific target PR.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -610,10 +616,13 @@
     uv pip install -r models/demos/gemma4/requirements.txt
     export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=google/gemma-4-26B-A4B-it TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/google--gemma-4-26B-A4B-it
     pytest -s --timeout 1500 models/demos/gemma4/demo/text_demo.py::test_demo
+    # Deselect packed-verify batch_perf (~2–3m): WH llmbox_perf SKU budget is
+    # 10m; demo+lm_head+matches_sequential+full_model already fill it.
```
mtairum, line 620: Remove this comment.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -586,14 +589,17 @@
...
+      --deselect models/demos/gemma4/tests/unit/test_packed_verify.py::test_packed_verify_batch_perf
   model: gemma-4-e4b
   model_family: Gemma
   skus:
     # wh_n300 disabled due to reduced N300 runner availability — re-enable when capacity returns.
     # wh_n300:
     #   timeout: 15
     #   tier: 3
+    # Note: do not add bh_p150 here — models e2e_tier3 bh_p150 budget is 7m
+    # and Gemma-4-E2B already uses 5m on that SKU (f562aa60493 over-budget).
```
mtairum, line 602: remove this comment

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -641,19 +650,23 @@
     uv pip install -r models/demos/gemma4/requirements.txt
     export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=google/gemma-4-31B-it TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/google--gemma-4-31B-it
     pytest -s --timeout 1500 models/demos/gemma4/demo/text_demo.py::test_demo
+    # Deselect packed-verify batch_perf (~2–3m): WH llmbox_perf SKU budget is
+    # 10m; demo+lm_head+matches_sequential+full_model already fill it.
```
mtairum, line 654: Remove

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -672,12 +685,30 @@
...
+    # wh_llmbox_perf was 259/266 before this job — budget bumped +10.
+    # CI=true selects 1x8. release_ready false until the WH leg is green.
+    wh_llmbox_perf:
+      timeout: 25
+      tier: 2
+      release_ready: false
+    # Single Blackhole P150: short demo + unit PCC. Cap at 6m so models
+    # e2e_tier2 bh_p150 stays within the 26m team budget (GPT-OSS 15 + UNet 5
+    # + this 6). f562aa60493's timeout:20 made requested=40 and failed
+    # load-test-matrix (see all-model-tests on arg/gemma4_1x8optiom).
```
mtairum, line 706: Remove c omment.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -672,12 +685,30 @@
...
+      --deselect models/demos/gemma4/tests/unit/test_packed_verify.py::test_packed_verify_batch_perf
+    pytest -s --timeout 3600 models/demos/gemma4/tests/e2e/test_teacher_forcing_e2e.py
   model: gemma-4-12b
   model_family: Gemma
   skus:
+    # T3K: same command as Gemma-4-31B e2e on this SKU (demo + lm_head +
+    # packed_verify matches_sequential + full_model; batch_perf deselected).
+    # 12B is smaller; 10m matches the 31B WH leg. models.e2e_tier2
+    # wh_llmbox_perf was 259/266 before this job — budget bumped +10.
+    # CI=true selects 1x8. release_ready false until the WH leg is green.
```
mtairum, line 698: Remove comment.

tests/pipeline_reorg/models_sweep_tests.yaml
mtairum, line 1: Do we really need these sweep tests?

I'm not against it, but this being a tier 2 model, is there an actual reason for us to include, or should we save CI cycles?

mtairum:
The 12B/31B Wormhole unit and end-to-end jobs pass, but these five blockers remain:

1. **Token feedback:** Smaller decode batches with eager sampling can reuse a stale token.
2. **Data-parallel recovery:** Reset uses the total batch size instead of the per-rank size, missing the live decode state.
3. **KV ownership:** Request compaction can make a request read another request’s cached history.
4. **RMSNorm allocation:** The 26B Wormhole prefill path exceeds L1 capacity, as reproduced in CI.
5. **Short-history attention:** Padding is treated as valid history, and subsequent chunks can discard history still needed.

Could you please address these and add targeted regression tests?

mtairum (changes requested):
This PR has a bunch of changes that are not specific to Gemma-4 WH T3K migration.

Please clean it up and re-request a new review.
Also make sure to rebase your PR afterwards on top of latest main and to re-run Gemma4 on the CI pipelines again.

arginugaTT:
@ign-rahultp , please rebase your files with current main, then run all_model_tests including vLLM model tests for approval.

ign-rahultp:
Hi @arginugaTT 
Rebasing this branch onto the latest main is introducing a significant number of merge conflicts. I will create a new PR specifically for the optimisation as suggested in https://github.com/tenstorrent/tt-metal/pull/53947#issuecomment-5580199220.

ign-rahultp:
The performance optimisations are taken care in the PR -> https://github.com/tenstorrent/tt-metal/pull/56996

PR #53901, opened 2026-08-21
[skip ci] test: deselect two multi-minute core ttnn tests on ttsim

Aswinmcw:
### Ticket

Closes #53228

### Problem description

`core ttnn unit test group [sim_wh_n150]` and `[sim_bh_p150]` have never passed. The suite
needs ~109 worker-minutes across 4 xdist workers and the step budget is 40 minutes, so the job
is killed mid-run and reports failure with zero actual test failures.

Two tests account for ~72 of those 109 worker-minutes:

| test | cost under ttsim |
|---|---|
| `test_to_layout.py::test_to_layout_low_perf` | ~60 worker-min (8 params at ~9 min) |
| `test_reshape.py::test_reshape_rm_nonclean_large_odd_dest_page_l1_cb[dest_page_200002B_issue_width]` | 754.6 s (sibling param: 1.2 s) |

### What's changed

Added both to the `# Job-level timeouts` section of `ttsim-skip-list.yaml` (both arches), which
the sanity workflow turns into `--deselect=` args for every `sim_*` sku. No
`TT_METAL_SIMULATOR` checks in the tests — #49633 removed those repo-wide and moved timeout
cases here.

Both paths are arch independent and still run on silicon, so no coverage is lost. Leaves
~73 worker-minutes ≈ 18 wall-minutes, inside the budget.

### CI

[Run 32462374313](https://github.com/tenstorrent/tt-metal/actions/runs/32462374313) with these
two removed: `[sim_wh_n150]` **success, 24.2 min** and `[sim_bh_p150]` **success, 31.4 min**,
against failure on every main nightly. Re-run on this revision:
[32470849641](https://github.com/tenstorrent/tt-metal/actions/runs/32470849641).

PR #53956, opened 2026-08-21
[skip ci] test: shrink the two slow core ttnn tests so ttsim can run them again

Aswinmcw:
### Ticket

Relates to #53228. Follow-up to #53901.

### Problem description

#53901 deselected the two tests below on ttsim. Per review feedback there, the better outcome
is to keep sim and silicon coverage identical and get the runtime down instead — the same thing
#53360 did for `test_reduction.py`.

Both were slow for reasons incidental to what they test.

### What's changed

**`test_to_layout_low_perf`** — shape `[32, 128 * 1024]` → `[32, 4 * 1024]`.

The width only controls how many times the `use_low_perf` tilize path repeats. `use_low_perf`
short-circuits `select_program_factory` (`tilize_device_operation.cpp:313`) before
`sub_core_grids` is consulted, so this path is always `TilizeSingleCoreProgramFactory` on
`corerange_to_cores(sub_core_grids)[0]`; 128 tile columns still runs that one core over many
blocks. **All 8 `dtype` × `sub_core_grids` parametrizations are kept** — nothing is dropped,
the tensor is just smaller.

**`test_reshape_rm_nonclean_large_odd_dest_page_l1_cb`** — source shape `(200002, 1)` →
`(22, 9091)`, dest unchanged at `(2, 100001)`.

What this case tests is the dest CB: a 200002 B dest page gives
`dest_slot_size_bytes = ((200002 - 1) & MASK_64) + 80 = 200080`
(`reshape_rm_program_factory.cpp:114`), so the fixed 8 slots would need 1,600,640 B against
1,499,136 B of L1 and
`choose_num_dest_write_slots` is forced to shrink. That depends on the dest page only.
`pages_noc_aligned` (false), `can_use_dual_kernel` (false), the kernel's `can_be_clean`
(false) and the 2 active cores that `responsibility` resolves to are all unchanged. The 754 s
was 200002 two-byte source page transactions moving the same 400 KB.

Both `ttsim-skip-list.yaml` entries added by #53901 are removed (from each of the
`wormhole_b0` and `blackhole` sections), so the file returns to exactly its state before that
PR.

### Measurements

Device time on a Blackhole p150b, JIT-warm:

| case | before | after |
|---|---|---|
| tilize low_perf (4096 → 128 tile cols) | 5.8 ms | 0.3 ms |
| reshape to `(2, 100001)` | 43.6 ms | 1.0 ms |

Tilize device time is linear in tile count (4096 → 5.8 ms, 1024 → 1.6 ms, 256 → 0.5 ms,
128 → 0.3 ms). Against the 537 s the 4096-tile case measured under ttsim that is a ~92,600×
slowdown factor, so the projection is **~9 min → ~28 s per parametrization** (~3.7 min for all
8) and **754 s → ~17 s** for the reshape case. That keeps `[sim_bh_p150]`, which runs 31.4 min
against its 40 min cap with these two absent, inside budget with both restored.

`pytest` on a p150b: 10 passed (8 low_perf + 2 reshape).

### CI

ttsim sanity run on this revision: [32491721454](https://github.com/tenstorrent/tt-metal/actions/runs/32491721454).

### Notes for reviewers

#53901 has merged and this branch is now based on `main`; the skip-list entries it added are
removed here directly.

The two comment corrections in `ebae0b0` are documentation only — they fix the single-core
rationale in `test_to_layout.py` and the `dest_slot_size_bytes` arithmetic in
`test_reshape.py`, both flagged in review. No test input changed, so the sanity run below
remains representative.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:brain/restore-sim-parity-53228)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:brain/restore-sim-parity-53228)


bbradelTT (approved):

vtsilytskyiTT (approved):
> ttsim sanity run on this revision: [32491721454](https://github.com/tenstorrent/tt-metal/actions/runs/32491721454).

FYI: in the job press 3 dots in upper right corner -> Create status badge :)

mcraigheadTT (approved):

PR #53965, opened 2026-08-21
[Bug fix] unary: hash the preallocated output tensor's dtype

### LINKED ISSUE #53963: The unary program cache key omits the preallocated output's dtype
**Component / Area:** ttnn eltwise unary

**Issue Type:** Bad Outputs

## Observed

Two unary calls that share an input dtype but preallocate outputs of different dtypes collide in the program cache, and the second is served the first one's packer. The order decides which one is wrong. `ttnn.neg`, value `1.5`, one tile:

| sequence | first call | second call |
|---|---|---|
| `fp32→fp32`, then `fp32→bf16` | correct | **512 of 1024 elements are 0.0** |
| `fp32→bf16`, then `fp32→fp32` | correct | **1024 of 1024 elements are -1.50585** |

Both are the wrong packer writing at the wrong width. `-1.5f` is `0xBFC00000`: a float32 packer emptying into a bfloat16 buffer leaves `0xBFC0` and `0x0000` in successive elements, and a bfloat16 packer emptying into a float32 buffer glues two words into `0xBFC0BFC0`, which is `-1.50585`. The destination is fully written; nothing is left stale.

Twelve ops — neg, abs, relu, sqrt, exp, reciprocal, gelu, tanh, sigmoid, log, square, rsqrt — every one of the 65536 bfloat16 bit patterns, each case scored against the same call made in a process where nothing warmed the cache:

| | P300 | N300 |
|---|---|---|
| 24 cross-dtype cases, 1572864 values | **1409911 wrong** | **1403192 wrong** |
| 24 same-dtype cases, 1572864 values | 0 | 0 |
| the same 48 cases, one per process | 0 | 0 |

Nothing is raised and nothing in the result indicates it.

`unary_device_operation.cpp:171-209` hashes `attributes`, the **input** tensor's dtype, layout, memory config and the shard volumes. The output dtype reaches the key only through `attributes.output_dtype`, and `unary_impl` (`unary.cpp:22`) sets that to `input_dtype` for everything but TYPECAST and BITCAST — while the line below it does take the memory config from the preallocated tensor. `compute_output_specs` (`:120-122`) returns `tensor_args.output_tensor->tensor_spec()`, so the program is built for the caller's real dtype while the key says otherwise. `validate_on_program_cache_miss` already spells the expression the hash needs, at `:53`: `output_tensor.has_value() ? output_tensor->dtype() : args.output_dtype`.

`ttnn.add`, `ttnn.multiply` and `ttnn.assign` with a preallocated narrower output are unaffected.

## Expected

Two calls whose outputs are different dtypes get two cache entries.

## Not established

- Whether anything in tree is affected today. `output_tensor=` reaches a ttnn unary at 20 model call sites and every one is in-place — `ttnn.silu(x, output_tensor=x)` and the like — so the dtypes match and the collision does not fire. The parameter is public and takes any tensor.
- The binary and ternary ops, which key correctly.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53963.

`compute_program_hash` never saw the preallocated output tensor's dtype, so two calls that differed only in that dtype collided and the second was served the first one's packer. `compute_output_specs` already resolves the output spec from the caller's tensor; the hash now takes the dtype from that spec rather than from `attributes.output_dtype`, which carries the input's dtype on this path because `unary_impl` sets it there and only reads the preallocated tensor for its memory config.

One term added to each of the two `hash_operation` calls. The program is built the same way, the flags are the same, and the number of programs built for any one dtype combination is unchanged — one per combination, where before two combinations shared one.

## Accuracy

Every number is this branch against `04cf22f7ee`, built and run on the same device, output bits compared as integers. Twelve ops: neg, abs, relu, sqrt, exp, reciprocal, gelu, tanh, sigmoid, log, square, rsqrt.

The paths that were already right must not move — no `output_tensor`, every one of the 65536 bfloat16 bit patterns and 4194304 float32 patterns drawn uniformly from the 2^32 space; and with an `output_tensor`, one dtype combination per process so nothing warms the cache:

| | values compared | P300 | N300 |
|---|---|---|---|
| no `output_tensor`, 24 sweeps | 51118080 | **0** | **0** |
| one combination per process, 48 cases | 3145728 | **0** | **0** |

What changes is the second call in a process. The same 48 cases run in one process in the order bf16→bf16, bf16→fp32, fp32→fp32, fp32→bf16 per op, each scored against its own uncontaminated reference:

|---|---|---|---|---|
| 24 cross-dtype cases, 1572864 values | 1409911 | **0** | 1403192 | **0** |
| 24 same-dtype cases, 1572864 values | 0 | 0 | 0 | 0 |

## Cost

Wall clock through `ttnn.exp` with a preallocated output, 5 warm-up then 50 timed calls with one `synchronize_device` at the end, median of five runs, profiler off, µs per call.

| | | main | this branch |
|---|---|---|---|
| P300 | 1024 tiles, bfloat16 | 12.3 | 12.3 |
| | 1024 tiles, float32 | 25.2 | 25.3 |
| | 16384 tiles, bfloat16 | 180.5 | 181.2 |
| | 16384 tiles, float32 | 375.4 | **375.2** |
| N300 | 1024 tiles, bfloat16 | 33.2 | 33.2 |
| | 1024 tiles, float32 | 66.5 | 66.5 |
| | 16384 tiles, bfloat16 | 382.5 | 382.8 |
| | 16384 tiles, float32 | 870.3 | **869.6** |

Dispatch count is 1 on both branches.

## Tests

`test_unary_preallocated_output_dtype.py`, 15 cases: neg, abs and relu, each with both input dtypes and both output dtypes, every one preceded by a same-dtype call that warms the entry the hash used to collide with, plus the two orders back to back. 9 of the 15 fail on `04cf22f7ee` with `512 of 1024 elements are not -1.5`; all 15 pass here, on P300 and on N300.

`test_unary_program_cache.py`, the existing regression file for this hash, passes unchanged: 19 of 19.

## Not covered

- The wider repair, in `unary_impl` (`unary.cpp:22`):

  ```cpp
  -    DataType output_dtype = input_dtype;
  +    DataType output_dtype = optional_output_tensor.has_value() ? optional_output_tensor->dtype() : input_dtype;
  ```

  It fixes the same collision and also feeds `fp32_dest_acc_en`, so `op(bfloat16, output_tensor=float32)` starts computing in a float32 dest. Mean relative error against a float64 golden over all 65536 bfloat16 patterns: exp 2.40e-04 → 5.06e-09, reciprocal 1.40e-03 → 2.28e-08, sigmoid 1.17e-04 → 5.52e-09, tanh 4.18e-05 → 2.40e-09, square 2.24e-03 → 0. It also moves 326871 values for callers who are correct today, and on the exact ops it inherits the existing bfloat16→float32 widening defect for the eleven smallest binades — `neg(1.175494e-38)` returns `-1.763242e-38`, which is what `ttnn.typecast(bfloat16, float32)` returns for that input on both branches. That is a numerics policy question; this PR takes the narrow repair so the cache-key bug can land without one.
- Binary and ternary ops, which key correctly today.

### DIFF
--- a/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
@@ -168,11 +168,6 @@ Tensor UnaryDeviceOperation::create_output_tensors(
     return create_device_tensor(compute_output_specs(args, tensor_args), tensor_args.input.device());
 }
 
-// The output dtype is hashed from the output spec, not from attributes.output_dtype:
-// with a preallocated output the spec comes from the caller's tensor (compute_output_specs
-// above), while attributes.output_dtype still carries the input's dtype. Without it, two
-// calls that differ only in the output tensor's dtype collide and the second is served the
-// first one's packer.
 ttsl::hash::hash_t UnaryDeviceOperation::compute_program_hash(
     const operation_attributes_t& attributes, const tensor_args_t& tensor_args) {
     const auto& input_tensor = tensor_args.input;
--- a/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
@@ -168,6 +168,11 @@ Tensor UnaryDeviceOperation::create_output_tensors(
     return create_device_tensor(compute_output_specs(args, tensor_args), tensor_args.input.device());
 }
 
+// The output dtype is hashed from the output spec, not from attributes.output_dtype:
+// with a preallocated output the spec comes from the caller's tensor (compute_output_specs
+// above), while attributes.output_dtype still carries the input's dtype. Without it, two
+// calls that differ only in the output tensor's dtype collide and the second is served the
+// first one's packer.
 ttsl::hash::hash_t UnaryDeviceOperation::compute_program_hash(
     const operation_attributes_t& attributes, const tensor_args_t& tensor_args) {
     const auto& input_tensor = tensor_args.input;
@@ -196,6 +201,7 @@ ttsl::hash::hash_t UnaryDeviceOperation::compute_program_hash(
             input_tensor.layout(),
             input_tensor.memory_config(),
             input_tensor.padded_shape(),
+            output_spec.data_type(),
             src_shard_vol,
             dst_shard_vol);
     }
@@ -205,6 +211,7 @@ ttsl::hash::hash_t UnaryDeviceOperation::compute_program_hash(
         input_tensor.dtype(),
         input_tensor.layout(),
         input_tensor.memory_config(),
+        output_spec.data_type(),
         src_shard_vol,
         dst_shard_vol);
 }
--- a/tests/ttnn/unit_tests/operations/eltwise/test_unary_preallocated_output_dtype.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_unary_preallocated_output_dtype.py
@@ -1,0 +1,50 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""A preallocated output_tensor of a different dtype gets its own cached program.
+
+The unary program hash was built from the input dtype and attributes.output_dtype,
+which carries the input's dtype when a preallocated output is given. Two calls that
+differed only in the output tensor's dtype collided, and the second was served the
+first one's packer: it wrote every other element, or read the wrong stride.
+"""
+
+import pytest
+import torch
+import ttnn
+
+SHAPE = (1, 1, 32, 32)
+TORCH_DTYPE = {ttnn.bfloat16: torch.bfloat16, ttnn.float32: torch.float32}
+
+OPS = [(ttnn.neg, 1.5, -1.5), (ttnn.abs, -2.0, 2.0), (ttnn.relu, 1.5, 1.5)]
+
+
+def _t(v, dtype, device):
+    return ttnn.from_torch(
+        torch.full(SHAPE, v, dtype=TORCH_DTYPE[dtype]), dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device
+    )
+
+
+@pytest.mark.parametrize("op, value, expected", OPS)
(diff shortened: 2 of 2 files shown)


ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
```diff
@@ -168,6 +168,11 @@ Tensor UnaryDeviceOperation::create_output_tensors(
     return create_device_tensor(compute_output_specs(args, tensor_args), tensor_args.input.device());
 }
 
+// The output dtype is hashed from the output spec, not from attributes.output_dtype:
+// with a preallocated output the spec comes from the caller's tensor (compute_output_specs
+// above), while attributes.output_dtype still carries the input's dtype. Without it, two
+// calls that differ only in the output tensor's dtype collide and the second is served the
+// first one's packer.
```
KalaivaniMCW: Please leave this comment to the PR description.

ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
KalaivaniMCW, line 1: @kinginu  I need to check something regarding this change, I'll get back to you

KalaivaniMCW (changes requested):

ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
kinginu: Done, it is in the Summary section of the description.

ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_device_operation.cpp
KalaivaniMCW: I don't think we need to add this change, The output_dtype is already part of the attributes. It was not being propagated to the unary_device_operation call. 
Historically unary ops have been allowing custom dtype (ie. input dtype != output dtype) only for ops like typecast and bitcast. Rest of the unary ops have always followed `input dtype == output dtype` rule

KalaivaniMCW: I have relaxed the rule for preallocated tensors for now, but I will be adding some restrictions to block unsafe mixed dtypes

mateusznowakTT:
Closing as referenced issue is closed.

PR #53970, opened 2026-08-21
[Bug fix] backward ops: the auto-allocated gradient takes the caller's memory config

### LINKED ISSUE #53969: Backward ops drop the memory_config they are given, and div_bw returns -inf when it is asked for L1
**Component / Area:** ttnn eltwise backward (composite)

**Issue Type:** Bad Outputs

## Observed

Every backward op that allocates its own gradient does it with `ttnn::empty_like(input)`, which takes the input tensor's memory config and never looks at the `memory_config` argument:

```cpp
// binary_backward.cpp:37-42, the helper seven ops share
if (required_outputs[0] && !input_grad.has_value()) {
    input_grad = ttnn::empty_like(input);
}
```

Four parameters are name-commented out entirely — `add_bw` and `sub_bw` (scalar overloads), `assign_bw` and `concat_bw` all take `const std::optional<MemoryConfig>& /*output_mem_config*/`.

DRAM inputs, `memory_config=ttnn.L1_MEMORY_CONFIG`, one op per process:

| | P300 | N300 |
|---|---|---|
| ops placing every gradient in L1 as asked | **5 of 22** | **5 of 22** |

The five are `abs_bw`, `atan2_bw`, `hypot_bw`, `log_bw`, `xlogy_bw`, the same on both machines. `addcmul_bw` and `addcdiv_bw` place two of their three; the other fifteen place none.

`div_bw` is worse than misplaced. Its intermediates are told L1 while the buffer they write is DRAM, and the second gradient comes back as `-inf`. Every bfloat16 bit pattern in each operand position:

| second gradient, swept operand | P300 | N300 |
|---|---|---|
| grad | **64717 / 65536** | **64717 / 65536** |
| input | **64940 / 65536** | **65067 / 65536** |
| other | **49152 / 65536** | **49152 / 65536** |

`div_bw(grad=1.5, input=2.5, other=0.5)` returns `-15` with the default config, which is what torch returns, and `-inf` when L1 is asked for. Reproduced three times in three processes on each machine.

## Expected

The gradient lands in the requested memory, and the answer does not depend on where it lands.

## Not established

- **Why the mismatch produces `-inf`.** Ruled out, all measured: it is not L1 capacity (identical at 32x32, 128x128, 256x256 and 512x512); not the program cache (a cold process and a warmed one give the same wrong answer); not the predicates (`eqz`, `ltz`, `gtz`, `nez` and `sign` all answer correctly under `memory_config=L1`); not the arithmetic (running `square → reciprocal → multiply → neg → multiply` by hand with L1 intermediates and a DRAM output gives the correct `-15`); and not the output-aliases-an-operand `where` at the end, which keeps its operand correctly under every config combination. Removing the mismatch removes the `-inf`, but the mechanism between them is unexplained.
- What is affected in tree. No model file calls a ttnn backward op. 238 test files do, and 178 call sites pass `memory_config` — and pass, because nothing asserts placement.
- Sharded and ND-sharded configs. Only interleaved DRAM and L1 were measured.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53969.

A gradient the caller did not preallocate was built with `empty_like(input)`, which copies the input's memory config and ignores the caller's. `preallocated_tensors_check` now takes `output_mem_config` and hands it to both allocations, the thirteen `empty_like` sites in the three backward files pass it too, and the three ops that return the incoming gradient unchanged — `addcmul_bw`, `addcdiv_bw`, `frac_bw` — route it through `to_memory_config`, which returns the tensor untouched and dispatches nothing when the config already matches.

Four parameters that were name-commented out — `add_bw` and `sub_bw` (scalar overloads), `assign_bw`, `concat_bw` — are named again and used.

## Accuracy

With no `memory_config` — every existing caller — nothing may move. 22 ops, all 65536 bfloat16 bit patterns in each operand position:

| | sweeps | values compared | P300 | N300 |
|---|---|---|---|---|
| default config | 76 | 8781824 | **0 differ** | **0 differ** |

With `memory_config=ttnn.L1_MEMORY_CONFIG` on DRAM inputs, one op per process:

|---|---|---|---|---|
| ops placing every gradient in L1 | 5 of 22 | **22 of 22** | 5 of 22 | **22 of 22** |
| ops whose values move between the default config and L1 | 1 — `div_bw` | **0** | 1 — `div_bw` | **0** |

`div_bw` is the one that mattered. Its intermediates were told L1 while the buffer they wrote was DRAM, and the second gradient came back as `-inf`. Every bfloat16 bit pattern in each operand position:

| second gradient, swept operand | P300 before | N300 before | after, both |
|---|---|---|---|
| grad | 64717 / 65536 | 64717 / 65536 | **identical to the default config** |
| input | 64940 / 65536 | 65067 / 65536 | **identical** |
| other | 49152 / 65536 | 49152 / 65536 | **identical** |

`div_bw(grad=1.5, input=2.5, other=0.5)` goes from `-inf` back to `-15`, which is what torch returns.

## Cost

Wall clock, 5 warm-up then 50 timed calls with one `synchronize_device` at the end, median of five runs, profiler off, µs per call, `[1, 1, 1024, 1024]`.

| | | main | this branch |
|---|---|---|---|
| P300 | `div_bw`, bfloat16 | 390.2 | **385.0** |
| | `div_bw`, float32 | 605.8 | 609.6 |
| | `add_bw`, bfloat16 | 33.1 | 34.8 |
| | `add_bw`, float32 | 46.8 | 46.8 |
| N300 | `div_bw`, bfloat16 | 781.6 | 781.7 |
| | `div_bw`, float32 | 1559.1 | **1557.8** |
| | `add_bw`, bfloat16 | 70.1 | **69.9** |
| | `add_bw`, float32 | 122.0 | 122.0 |

Dispatch counts are unchanged: 23 for `div_bw`, 2 for `add_bw`. `to_memory_config` short-circuits on a matching config (`to_memory_config_op.cpp:257-260`), so the three aliasing sites cost nothing unless a different placement was actually requested.

## Tests

`test_backward_memory_config.py`, 36 cases: 18 ops asserted to place every gradient in the requested L1, and the same 18 asserted to return identical values with and without the request. 18 of the 36 fail on `04cf22f7ee`, on both machines; all 36 pass here, on P300 and on N300.

## Not covered

- **Why the mismatch produced `-inf`.** Removing it removes the `-inf`, but the mechanism between them is not explained. Ruled out by measurement: L1 capacity (identical at 32x32 through 512x512), the program cache (a cold process and a warmed one agree), the predicates (`eqz`, `ltz`, `gtz`, `nez`, `sign` all correct under `memory_config=L1`), the arithmetic chain (`square → reciprocal → multiply → neg → multiply` by hand with L1 intermediates and a DRAM output gives `-15`), and the trailing `where` whose output aliases an operand.
- `unary_composite_op.cpp:42/:136/:151`, where the caller's config describes a full-shape output and the op reduces to `[N, C, 1, 1]`. Passing it down asserts for sharded configs.
- Sharded and ND-sharded configs. Only interleaved DRAM and L1 were measured.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/backward/test_backward_memory_config.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/backward/test_backward_memory_config.py
@@ -1,0 +1,74 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""A gradient the caller did not preallocate still lands where the caller asked.
+
+The backward ops allocate their outputs with empty_like(input), which inherits the
+input's memory config and ignores the memory_config argument. div_bw went further:
+with the request unmet, its second gradient came back as -inf.
+"""
+
+import pytest
+import torch
+import ttnn
+
+SHAPE = (1, 1, 32, 32)
+
+# op name, number of tensor arguments, trailing scalar arguments
+OPS = [
+    ("abs_bw", 2, []),
+    ("exp_bw", 2, []),
+    ("neg_bw", 2, []),
+    ("sqrt_bw", 2, []),
+    ("silu_bw", 2, []),
+    ("rsqrt_bw", 2, []),
+    ("frac_bw", 2, []),
+    ("pow_bw", 2, [2.0]),
+    ("add_bw", 3, []),
+    ("sub_bw", 3, []),
+    ("mul_bw", 3, []),
+    ("div_bw", 3, []),
+    ("rsub_bw", 3, []),
+    ("assign_bw", 3, []),
+    ("addalpha_bw", 3, [2.0]),
+    ("subalpha_bw", 3, [2.0]),
+    ("addcmul_bw", 4, [2.0]),
+    ("addcdiv_bw", 4, [2.0]),
+]
+
+VALUES = (1.5, 2.5, 0.5, 3.0)
+
+
+def _t(v, device, memory_config):
+    return ttnn.from_torch(
+        torch.full(SHAPE, v, dtype=torch.bfloat16),
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=memory_config,
+    )
+
+
+def _call(op, arity, scalars, device, memory_config):
+    args = [_t(VALUES[i], device, ttnn.DRAM_MEMORY_CONFIG) for i in range(arity)]
+    kwargs = {} if memory_config is None else {"memory_config": memory_config}
+    out = getattr(ttnn, op)(*args, *scalars, **kwargs)
+    out = out if isinstance(out, (list, tuple)) else [out]
+    return [x for x in out if x is not None]
+
+
+@pytest.mark.parametrize("op, arity, scalars", OPS)
+def test_gradients_land_in_the_requested_memory(device, op, arity, scalars):
+    for i, out in enumerate(_call(op, arity, scalars, device, ttnn.L1_MEMORY_CONFIG)):
+        assert out.memory_config().buffer_type == ttnn.BufferType.L1, f"{op} output {i} is in {out.memory_config()}"
+
+
+@pytest.mark.parametrize("op, arity, scalars", OPS)
+def test_requesting_a_memory_config_does_not_move_any_value(device, op, arity, scalars):
+    default = _call(op, arity, scalars, device, None)
+    asked = _call(op, arity, scalars, device, ttnn.L1_MEMORY_CONFIG)
+    for i, (a, b) in enumerate(zip(default, asked)):
+        x, y = ttnn.to_torch(a).float(), ttnn.to_torch(b).float()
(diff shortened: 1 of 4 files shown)


mateusznowakTT:
@kinginu : Please resolve merge conflicts

kinginu:
Done, and most of the branch is gone with it.

`binary_backward.cpp` already passes the caller's memory config to every `empty_like` on main, 6 of 6, so its version is taken wholesale; the fourteen conflicts there were a parameter rename and clang-format wrapping around the same fix. The passthrough gradients went the other way: `#55613` settled that they keep grad's own config rather than materialising a copy, so the `to_memory_config` calls in `ternary_backward.cpp` and `frac_bw` are dropped, and the test exempts output 0 of `frac_bw`, `addcmul_bw` and `addcdiv_bw` for that reason.

What is left is `unary_backward.cpp`, where 5 of 6 allocations still ignore the caller: `exp_bw`, `sqrt_bw`, `rsqrt_bw`, `neg_bw`, `silu_bw`. P300, a build per arm: those five fail the test without the change, 36 pass with it.

PR #53980, opened 2026-08-21
[Bug fix] div: pass output_dtype through on the four non-int32 paths

### LINKED ISSUE #53978: ttnn.div discards the dtype it is given on every non-int32 path

**Issue Type:** Bad Outputs

## Observed

`eltwise/binary/device/binary_composite_op.cpp:228` and `:352` pass `std::nullopt` as `invoke_binary_ng`'s dtype argument while `output_dtype` is in scope and unused. The trunc and floor branches at `:256` and `:380` drop it too.

The requested dtype is discarded and the lhs dtype comes back instead, with no error. Measured at `04cf22f7ee`, identical on P300 and N300:

| call | asked for | returned |
|---|---|---|
| `div(bf16, bf16, dtype=float32)` | float32 | **bfloat16** |
| `div(fp32, fp32, dtype=bfloat16)` | bfloat16 | **float32** |
| `div(bf16, 2, dtype=float32)` | float32 | **bfloat16** |
| same three with `rounding_mode="floor"` / `"trunc"` | | same |

The fourth positional parameter of `invoke_binary_ng` is `const std::optional<const DataType>& dtype` (`binary.cpp:639-651`), consumed at `:570` as `dtype.value_or(lhs_eff.dtype())` — hence the lhs dtype. Three signals say the pass-through is what was meant: `ttnn::divide` itself honours it (`binary.cpp:1071`, `:1097`); `remainder` in this same file honours it (`:462`, `:494`); and `div` already *validates* `output_dtype` on the int32 path (`:202-205`, `:326-329`), which is meaningless if the value is never read. The ops at `:56`, `:77`, `:97` and `:118` spell the parameter `/*output_dtype*/` — unnamed, dropped on purpose. `div` names it.

## Expected

The requested dtype, or a rejection.

## Not established

- Whether any out-of-tree caller depends on the current behaviour. In-tree, `models/experimental/ops/quasar/tests/yolo_ops/test_div.py:61` does `ttnn.div(x, 2, dtype=ttnn.bfloat8_b)` and today gets bfloat16; with the dtype honoured it gets bfloat8_b at PCC 0.99997, above its 0.99 assertion.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53978.

`ttnn.div` names its `output_dtype` parameter and then hands `std::nullopt` to the dtype slot at all four non-int32 sites, so `div(a, b, dtype=...)` returned the lhs dtype and the request was lost. Four one-token edits pass it through, in both overloads and on both the plain and the trunc/floor path.

On the rounding paths the dtype now applies to the divide, i.e. before the rounding step. That is where a preallocated `output_tensor` already lands today (`binary.cpp:570`, `output_preallocated ? output->dtype() : dtype.value_or(...)`), so this makes `dtype` and `output_tensor` agree instead of differ; honouring it only after the rounding would cost an extra typecast dispatch, since `ttnn::trunc` and `ttnn::floor` take no dtype.

## Accuracy

With `dtype` omitted — every in-tree caller but one — the patched argument *is* the `std::nullopt` it replaces, so nothing may move. Every one of the 65536 bfloat16 bit patterns in each operand position against four divisors, 2^22 float32 patterns the same way, all 65536 int32 values from -32768 up, in both overloads and all three rounding modes:

| | sweeps | values compared | P300 | N300 |
|---|---|---|---|---|
| `dtype` omitted | 78 | 46661632 | **0 differ** | **0 differ** |

With `dtype` passed, the 72 sweeps whose request differs from the lhs dtype all change dtype, and the 78 whose request already matched it are bit-identical. Identical on both machines:

| | main | this branch |
|---|---|---|
| `div(bf16, bf16, dtype=float32)` | bfloat16 | **float32** |
| `div(fp32, fp32, dtype=bfloat16)` | float32 | **bfloat16** |
| `div(bf16, 2, dtype=bfloat8_b)` | bfloat16 | **bfloat8_b**, PCC 0.99997 |
| `div(bf16, bf16, dtype=bfloat16)` | bfloat16 | bfloat16, bit-identical |

The int32 arm is untouched: it still returns float32 with `rounding_mode=None` and int32 under trunc/floor, and still raises on a non-float32 `dtype`.

## Cost

None.

The branch adds no operation — it passes an argument that was already in scope into a slot that was receiving `std::nullopt`.

| P300, `[1, 1, 1024, 1024]` | main | this branch |
|---|---|---|
| dispatches, `rounding_mode=None` | 1 | 1 |
| dispatches, `rounding_mode="floor"` | 2 | 2 |
| bfloat16 | 43.6 us/call | 36.5 us/call |
| float32 | 54.4 us/call | 54.4 us/call |

The bfloat16 column is run-to-run spread, not a win.

## Tests

`test_div_output_dtype.py`, 15 cases. Nine ask for a dtype the lhs is not — float32 from bfloat16, bfloat16 from float32, bfloat8_b from bfloat16 — in both overloads and all three rounding modes, and assert both the returned dtype and that 7/2 is still 3.5 or 3. 9 of the 15 fail on `04cf22f7ee` (`asked DataType.FLOAT32, got DataType.BFLOAT16`), on both machines; all 15 pass here, on P300 and on N300.

## Not covered

- Whether a narrowing request should round before or after `trunc`/`floor`. This branch rounds in the requested dtype, matching `output_tensor`; torch casts after.
- `dtype` together with a disagreeing `output_tensor` now reaches the existing `TT_FATAL` at `binary.cpp:572-574`, where before the mismatch was silent.


mateusznowakTT:
The issue is no longer on main fixed with other change.

PR #53985, opened 2026-08-21
[Bug fix] multiply: honour an explicit fast_and_approximate_mode on block-float

### LINKED ISSUE #53984: multiply discards fast_and_approximate_mode for block-float; divide honours it
**Component / Area:** ttnn eltwise binary

**Issue Type:** Bad Outputs

## Observed

`binary.cpp:1165-1166` (tensor-tensor) and `:1193-1194` (tensor-scalar):

```cpp
bool is_block_fmt_inp = (is_block_float(lhs.dtype()) || is_block_float(rhs.dtype()));
bool fast_and_approx = is_block_fmt_inp ? true : fast_and_approximate_mode.value_or(false);
```

The caller's argument is discarded for block-float operands. `divide` is bound by the same helper (`bind_binary_operation_with_fast_approx`, which binds exactly these two ops) and does not override.

Measured at `04cf22f7ee`, all 65536 bfloat16 bit patterns, `fast_and_approximate_mode` False vs True, identical on P300 and N300:

| op | bfloat16 | float32 | bfloat8_b |
|---|---|---|---|
| multiply | 11683 / 65536 respond | 0 | **0 — the flag does nothing** |
| divide | 25502 / 65536 respond | 0 | **40910 / 65536 respond** |

`add` and `subtract` do not take the flag at all.

With the override removed, the flag does reach the block-float multiply — 6888 of 65536 values respond — so the path exists and is simply unreachable. What it buys, against a float64 golden of the values the device holds:

| bfloat8_b multiply | mean relative error | 1024x1024 |
|---|---|---|
| SFPU (`False`) | **2.821e-03** | 25.12 us/call |
| FPU (`True`, what the override forces) | 3.406e-03 | 25.10 us/call |

The FPU path is 1-3% faster across sizes (23.28 vs 25.56 us at one tile, 35.30 vs 36.23 at 2048x2048) and about 21% less accurate on average. Which default is right is a judgement call; discarding an explicit argument is not.

The docstring does not mention block-float at all: "When `fast_and_approximate_mode` is `True` for bfloat16 datatype, the operation uses FPU implementation for better performance."

## Expected

The default stays as it is; an explicit argument is honoured, as `divide` already does.

## Not established

- Whether the FPU path should remain the block-float default. This issue does not propose changing it — the 1-3% against 21% trade is Tenstorrent's to make, and the numbers above are there to make it with.

## Environment

* OS: Ubuntu 22.04
* Hardware: N300 and P300

kinginu:
## PR Category

Bug fix.

## Summary

Closes #53984.

`multiply` forced `fast_and_approximate_mode` to true for block-float operands, discarding whatever the caller passed. `divide`, bound by the same helper and taking the same flag, honours it for the same dtypes.

The default does not move: block-float still takes the FPU path when the caller says nothing. To make "says nothing" expressible from Python at all, the binding default becomes `None` and the four trampolines take `std::optional<bool>` instead of `bool` — before this, the binding always passed an explicit `false`, so the C++ `std::nullopt` branch was unreachable and the override was the only thing keeping block-float on the fast path.

## Accuracy

`multiply` in both overloads, every one of the 65536 bfloat16 bit patterns in each operand position against three counterparts, 2^22 float32 patterns, and the same source patterns as bfloat8_b, with the flag unset, False and True:

| dtype | flag unset | explicit True | explicit False, P300 | explicit False, N300 |
|---|---|---|---|---|
| bfloat16 | **0** | **0** | **0** | **0** |
| float32 | **0** | **0** | **0** | **0** |
| bfloat8_b | **0** | **0** | **25986** tensor, **2322** scalar | **25482** tensor, **1944** scalar |

108 sweeps, 77070336 values, on each machine. The only thing that moves is the case where the caller passed the argument that was being thrown away.

`divide`'s default is unchanged in all three dtypes — its default and an explicit `False` are bit-identical on both branches, which is what the binding change had to preserve.

What the newly reachable path gives, against a float64 golden of the values the device holds:

| bfloat8_b multiply | mean relative error | max |
|---|---|---|
| SFPU (`fast_and_approximate_mode=False`) | **2.821e-03** | 1.205e-02 |
| FPU (the default) | 3.406e-03 | 1.205e-02 |

Identical to four figures on both machines.

## Cost

None for any existing caller — the default path is the same kernel it was.

| bfloat8_b multiply | FPU (default) | SFPU (`False`) |
|---|---|---|
| P300, 1 tile | 23.28 | 25.56 |
| P300, 1024x1024 | 25.10 | 25.12 |
| P300, 2048x2048 | 35.30 | 36.23 |
| N300, 1 tile | 47.53 | **46.75** |
| N300, 1024x1024 | 47.80 | 48.55 |
| N300, 2048x2048 | 81.67 | 83.10 |

The path a caller opts into is within a couple of percent either way and about 21% more accurate. That trade is now theirs to make.

## Tests

`test_multiply_block_float_fast_approx.py`, 8 cases: block-float multiply asserted to still default to the fast path, an explicit `False` asserted to change something, bfloat16 and float32 defaults asserted unchanged, `divide`'s default asserted unchanged in three dtypes, and the scalar overload. 2 of the 8 fail on `04cf22f7ee`, on both machines; all 8 pass here, on P300 and on N300.

## Not covered

- Whether the FPU path should stay the block-float default. This branch does not change it.
- `inplace_mul_operation_with_fast_approx` carries the same override and gets the same treatment, but the in-place op has no Python binding for the flag, so it is not swept here.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_multiply_block_float_fast_approx.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_multiply_block_float_fast_approx.py
@@ -1,0 +1,60 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""multiply honours an explicit fast_and_approximate_mode on block-float operands.
+
+Block-float still defaults to the FPU path. What changed is that a caller who passes
+False now gets the SFPU path, where before the argument was discarded — while divide,
+bound by the same helper, honoured it for the same dtypes.
+"""
+
+import pytest
+import torch
+import ttnn
+
+SHAPE = (1, 1, 256, 256)
+ALL_BF16 = torch.arange(0, 65536, dtype=torch.int64).to(torch.int32).to(torch.int16).view(torch.bfloat16)
+
+
+def _t(x, dtype, device):
+    torch_dtype = torch.float32 if dtype == ttnn.float32 else torch.bfloat16
+    return ttnn.from_torch(x.to(torch_dtype), dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+
+
+def _differ(a, b):
+    x, y = ttnn.to_torch(a).float(), ttnn.to_torch(b).float()
+    return ((x != y) & ~(torch.isnan(x) & torch.isnan(y))).sum().item()
+
+
+def test_block_float_multiply_still_defaults_to_the_fast_path(device):
+    a = _t(ALL_BF16.reshape(SHAPE), ttnn.bfloat8_b, device)
+    b = _t(torch.full(SHAPE, 1.5), ttnn.bfloat8_b, device)
+    assert _differ(ttnn.multiply(a, b), ttnn.multiply(a, b, fast_and_approximate_mode=True)) == 0
+
+
+def test_block_float_multiply_honours_an_explicit_false(device):
+    a = _t(ALL_BF16.reshape(SHAPE), ttnn.bfloat8_b, device)
+    b = _t(torch.full(SHAPE, 1.5), ttnn.bfloat8_b, device)
+    moved = _differ(ttnn.multiply(a, b, fast_and_approximate_mode=False), ttnn.multiply(a, b))
+    assert moved > 0, "an explicit False is still being discarded for block-float operands"
+
+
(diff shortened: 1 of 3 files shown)


mateusznowakTT:
@kinginu : Please resolve merge conflicts and adapt to recent main.

PR #53118, opened 2026-08-14
#53090: Fix fold BLOCK_SHARDED specless-override grid for COL_MAJOR

mouliraj-mcw:
### Ticket
Link to Github Issue #53090
closes #53090
closes #53115

### Problem
`synthesize_fold_output_shard_spec` built BLOCK_SHARDED shard specs with a ROW_MAJOR axis→core mapping unconditionally; on asymmetric grids (e.g. BH `bh_p100` 11×10) a COL_MAJOR input made `num_shards_along_width` overshoot `shard_grid.y`, tripping `TT_FATAL !shard_grid_fit_error.has_value()` in `tensor_spec.cpp`.

### Fix
Derive `orientation` before the switch, and for BLOCK_SHARDED swap the `grid.x`/`grid.y` roles when COL_MAJOR — `h_divisor`, `w_divisor`, and the resulting `CoreRange` now mirror `conv2d_utils::determine_parallel_config`.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mouliraj/fold_rework_bh_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mouliraj/fold_rework_bh_fix)

PR #54008, opened 2026-08-21
stable topk: 32-bit path repairs + rank-stamped fast wide-stable engine (stacked on #53557)

### LINKED ISSUE #53466: ttnn.topk INT32/UINT32 index_dtype returns wrong indices on Blackhole at k=50, W=8192 (multi-core path)

nkapreTT:
## Summary

Stacked on #53557 (base: `iwroszTT/topk-fused-stable-keys`). One reviewable stack, seven series, 42 commits (inventory at the bottom). What it delivers, in order:

1. **32-bit path repairs** — tests-first fixes for the stable-topk 32-bit paths (multicore dest-width, fp32 signed-zero ties, index-dtype validation).
2. **Rank-stamped wide-stable topk (C1–C4)** — the fast stable engine past the u16 index limit: W=65536/k=32 topk **19.15 → 12.84 ms (−33%)**.
3. **`ttnn.sort(stable=True)` end-to-end (C5)** — via the index-aware comparator network in all three sort factories.
4. **k > 32 insertion cascade (chain-rank stamps)** — W=65536 k=64 **37.09 → 23.89 ms (−35.6%)**, k=128 **72.95 → 45.71 ms (−37.3%)**.
5. **Mergesort row engine (Blackhole)** — `ttnn.sort` for bf16 rows of power-of-two padded width **2048–65536, BOTH stabilities**: full per-row sorts on the TopK XL SFPU kernels (fused linearly-tagged keys), one row per core; padded W ≥ 8192 run an L1-staged merge tree. Widths 512/1024 ride it via a composite pad-to-2048 rider. Speedups 4.8×–36.7× across the band (full table below).
6. **Hardening** — prim-level power-of-two width validation (silent-garbage fix), empty-tensor guard, NaN contract documentation, #54043 truth-sync, hostile-coverage cells.
7. **Truth-sync fixes (this update)** — composite early-exit dtype alignment, zero-size-tensor torch parity, Sort.md/docstring sync.

## Behavior changes a reviewer should know about (disclosures)

- **Public index-dtype change (the big one):** stable bf16 `ttnn.sort` on Blackhole returns **UINT32 indices at all power-of-two padded widths 2048–65536**. At padded widths **8192–32768 this flips the public dtype u16 → u32 vs main** (2048/4096 flipped earlier in this stack; ≥ 65536 was already u32 via the ≥ 65535 rule). Rationale: the mergesort engine's fused tags live in 32-bit DEST, and the u16-index-in-32-bit-DEST combination has no working pack path. A caller who preallocates UINT16 index tensors for a stable sort **opts out** to the comparator engines and keeps u16. **Unstable sorts see no dtype change at any width** — the writer RISC narrows the engine's native u32 words to u16. Grep of in-tree `ttnn.sort` callers (models/, tests/, tt-train/): every caller uses default `stable=False`, so no in-tree caller sees the flip. Documented in Sort.md and the op docstring.
- **WH-visible change:** the prim-level width validation now requires a **power of two** ≥ 64, not just a multiple of 64 — on every arch, a direct `ttnn::prim::sort` call with e.g. W=192 previously produced **silent garbage** (the bitonic engines truncate log2) and now TT_FATALs, with a gtest. The public `ttnn.sort` composite always pads to a power of two and is unaffected. Everything else in this stack is Blackhole-gated: **WH behavior of public `ttnn.sort` is otherwise unchanged** (mergesort engine, pad rider, and small-Ht reroute all gate on `arch == BLACKHOLE`). One WH CI pass still recommended.
- **E3 reroute history (honest note):** the small-Ht → CrossCore reroute was initially unstable-inclusive in-branch, which widened the observable #54043 duplicate-index band; the stress campaign caught it and the gate was narrowed to stable-only + Blackhole-only (f78159a5c9d) before the mergesort absorption made the point moot for bf16.
- **#54043 status:** the CrossCore unstable engine emits duplicate indices inside tie groups — silicon-measured at **every width that path serves (W=512–4096), on plain randn**, not only tie-heavy input. This stack **retires #54043 for bf16 on Blackhole**: unstable bf16 W=512–65536 rides the mergesort engine, whose output is the torch-stable permutation. Residue: Blackhole **fp32 unstable** CrossCore widths, and non-Blackhole devices (stated in the docstring).
- **NaN contract:** NaN input is unsupported (undefined ordering) for bf16 — the datapath canonicalizes NaN to same-sign infinity, deviating from torch's NaN-last. Documented, not fixed.
- **New early-exit behavior (this update):** zero-size tensors (e.g. `(0, 4096)`) now return empty values/indices matching `torch.sort` instead of hitting the prim's non-empty TT_FATAL (the prim guard stays as the backstop); the scalar/dim-size-1 early exits now return u32 indices for stable bf16, matching the on-device contract.

## The series in detail

**Repair (5 commits)** — tests-first fixes for the 32-bit paths, all with failing-first coverage (11 of 13 new cells failed pre-fix):
- Multicore factory: `fp32_dest_acc_en |= is32_bit_data` (mirrors single-core). bf16 input + preallocated u32/i32 indices unpacked 32-bit index datums into a 16-bit dest and silently broke stable tie order. The same fix makes the four pre-existing #53466 cells pass on main — that issue's mechanism is dest width, not `pack_tile<true>`.
- FLOAT32 stable: fold −0.0 into the +0.0 tie class at `_bitonic_topk_phases_steps` entry (`STABLE && fp32-dest` only; bf16 engines already canonicalize via SrcA). BH + WH mirrored.
- `resolve_index_dtype`: TT_FATAL on an `indices_tensor` whose dtype width mismatches the resolved index width (was a silent-garbage path).
- Exonerations recorded, comment-only: the fp32 32-bit-dest UInt32 transport is correct on silicon, and the merge's compile-time `top_min` tie polarity is right for every live caller.

**Rank-stamp series (C1–C4)** — sign-conditioned local rank tags in the value words' free lo16, true u32 indices riding the index-tracking machinery, plain unstable network. Wide stable topk (W ≥ 65536, k ≤ 32) moves off the 2–3× comparator.

**Stable sort via comparator (C5)** — `ttnn.sort(stable=True)` end-to-end through the index-aware comparator network in all three factories, plus the merge tie-polarity fix (runtime global mode instead of compile-time `top_min`).

**Completion series** — the remaining stable-fast-path work, scoped by measurement:
- **k > 32 insertion cascade (chain-rank stamps).** Per-level position stamps are provably wrong in the cascade — a loser tile carries displaced OLD elements into the next level where naive `[0,32)/[32,64)` ranges order them after accumulator ties they must precede. Demonstrated failing-first on BH silicon (4/4 displacement-ties cells fail with the gate lifted over naive stamps). Fix: each cascade level re-stamps its accumulator tile with that tile's round-start CHAIN-position range; the fresh chunk is stamped once per round with the top range; the loser tile's tags ride through the raw Float32 transport. New LLK primitive `_topk_stamp_tile_rank_range_`.
- **Fast stable `ttnn.sort` on the SingleRowSingleCore factory (fused true-index tags).** Every network call tags the value words' free lo16 with the TRUE u16 index (sign-conditioned complement from the GLOBAL order); keys become distinct, so the plain UNSTABLE network is torch-stable by construction. New LLK mode `PRE_TAGGED` on `_bitonic_topk_merge`. Stable bf16 sorts on this factory return UINT32 indices — the same 32-bit-DEST rule fp32/u16 inputs already follow.
- **Measured negatives kept on the comparator (implemented, silicon-validated, then deliberately dropped).** Two fast engines were fully built and correctness-proven but measured SLOWER than the comparator at every width they serve, so the comparator gate stays and the machinery does not ship:
  - *Multicore rank-stamped topk* (mirror-bit merges for direction-flipped cores + K=64 split-call rank bases): correct on silicon — with the mirror bit disabled, 5 tie cells fail; with it, 10/10 pass — but the multicore band is data-movement-bound and the Float32 tag transport costs more than the comparator saves (table below). The tie-heavy multicore cells stay as comparator coverage.
  - *Fused stable sort on the CrossCore / MultiCore-DRAM factories*: same verdict — their comparator already runs within ~0.3% of the unstable floor, so the fused engine's u32-index transport only adds cost.
- **Fundamental gates documented with counterexamples:** FLOAT32 stable keeps the comparator (an fp32 value has no free lo16 for a tag). Stable sort past the u16 tag ceiling keeps the comparator — and cannot use topk-style per-merge rank stamps either (worked counterexample in the gate-boundary commit).

**Mergesort series + scale-out** — the decision-memo roadmap, built out in-branch to the full ladder:
- **E3 — small-Ht routing fix.** The memo's hypothesis ("the 1.98 ms cell is a 32-row serialization artifact") was falsified on silicon (H-scaling probe: 32/64/1024 rows at W=2048 → 1982/1983/1983 µs — cost is per tile-row). The honest headroom: small-Ht cells in the single-core band leave the grid idle, so `Wt ∈ [16, 64]`, `Ht ≤ Wt/8` (TILE, non-u16, **stable-only, Blackhole-only** — see disclosures) reroute to the CrossCore factory. Measured at the time: 32×2048 stable 1982.0 → 186.2 µs (10.6×). On BH the reroute's bf16 cells have since been superseded by the mergesort engine + rider; its remaining live constituency is stable fp32 small-Ht cells and the stable prealloc-u16 opt-out.
- **E1 — falsification gate (stop if > 10 cyc/elem):** single-core W=4096 fused sort measured 24.7 µs = 8.14 cyc/elem @ 1.35 GHz → passed.
- **M1 — LLK both-halves merge (`_topk_xl_merge_both_halves_`)** — the full-sort merge level on the topk_xl kernels; 12/12 new LLK variants exact.
- **M2 — stability primitives + `SortProgramFactoryMergesortRowParallel`.** `_topk_xl_add_linear_indices_` stamps each fused word's free lo16 with the element's linear position (+ sign-conditioned complement for the tie class the sign-magnitude compare would reverse); `_topk_xl_separate_indices_linear_` splits runs into stripped bf16 values + UINT32 indices. One row per core, K=2048 leaves. Silicon find: **Blackhole DRAM reads require 64-byte alignment** — a 32 B face-row gather at a 32-mod-64 offset returns the neighboring row's face row; the reader stages aligned 64 B windows and compacts on the RISC-V core.
- **Unstable absorption.** The engine's output is the torch-stable permutation — a strict superset of the unstable contract — so unstable bf16 cells ride it too. The public u16 index contract is preserved by writer-RISC u32→u16 narrowing.
- **Pad-to-2048 rider.** bf16 rows whose pow2 pad would be 512/1024 are padded to 2048 instead so they land on the engine (gated off for the stable prealloc-u16 opt-out). Correctness leans on the engine being structurally stable: real elements always sort ahead of the ±inf pad sentinels' tags.
- **W=8192–65536 L1-staged merge tree.** Runs no longer fit the 8-tile DEST window, so runs ping-pong through Float32 L1 buffers between DEST passes (≤ 576 KB L1 at W=65536; measured 109.5 µs at W=8192 vs the ~61 µs model and the 150 µs stop-gate — gate passed). The hard ceiling is W=65536: the fused word's 16 free tag bits address exactly 65536 positions (max tag 0xFFFF, zero slack) — wider rows decline to the comparator engines.
- **Hostile coverage:** staged-boundary tie groups at every merge offset, all-ones at every absorbed width, prealloc-dtype cells, real ±inf colliding with pad sentinels, RM at tree widths, W=6000-vs-8192 program-cache convergence.

## Perf (BH p150a, Tracy Device Kernel Duration; tie-heavy bf16 unless noted; n ≥ 10/leg, multiple trials, spread < 0.6% — every shipped cell below was additionally re-measured by an independent audit run on the final head, agreeing within 1.3%)

**Shipped fast paths (topk + single-core sort):**

| cell | incumbent | fast path | Δ |
| --- | ---: | ---: | ---: |
| topk W=65536 k=32 (rank-stamped) | 19.15 ms | **12.84 ms** | −33% |
| topk W=65536 k=64 (chain-rank cascade) | 37.09 ms | **23.89 ms** | −35.6% |
| topk W=65536 k=128 (chain-rank cascade) | 72.95 ms | **45.71 ms** | −37.3% |
| sort W=2048 stable (single-core, fused tags) | 2.131 ms | **1.982 ms** | −7.0% (superseded on BH by the mergesort engine; still the live engine for this width off-BH) |

**Mergesort ladder (32 rows, bf16, TILE unless noted; "incumbent" = the engine that served the cell before this stack):**

| cell | incumbent | mergesort engine | ratio |
| --- | ---: | ---: | ---: |
| W=512 stable (rider) | 67.8 µs (E3 CrossCore) | **42.9 µs** | 1.6× |
| W=512 unstable (rider) | 217.5 µs (single-core) | **52.0 µs** | 4.2× |
| W=1024 stable (rider) | 102.4 µs (E3 CrossCore) | **43.1 µs** | 2.4× |
| W=1024 unstable (rider) | 553.5 µs (single-core) | **52.1 µs** | 10.6× |
| W=2048 stable | 1982.0 µs (fused single-core) | **33.5 µs** (RM: 15.3 µs) | 59× (RM 130×) |
| W=2048 unstable | 1562 µs (single-core)¹ | **42.5 µs** | 36.7×¹ |
| W=4096 stable | 379.9 µs (CrossCore comparator) | **60.6 µs** (RM: 32.9 µs) | 6.3× |
| W=4096 unstable | 376.4 µs (CrossCore) | **78.8 µs** | 4.8× |
| W=8192 stable (tree) | 972.3 µs (CrossCore) | **109.5 µs** | 8.9× |
| W=8192 unstable (tree) | 968.8 µs (CrossCore) | **146.2 µs** | 6.6× |
| W=16384 stable (tree) | 3170.1 µs (CrossCore) | **217.1 µs** | 14.6× |
| W=16384 unstable (tree) | — | **290.4 µs** | — |
| W=32768 stable (tree) | 6534.2 µs (MultiCore DRAM) | **471.6 µs** | 13.9× |
| W=32768 unstable (tree) | 6402.5 µs (MultiCore DRAM) | **609.5 µs** | 10.5× |
| W=65536 stable (tree) | 13547 µs (MultiCore DRAM) | **1001.5 µs** | 13.5× |
| W=65536 unstable (tree) | — | **1001.7 µs** | — |
| 1024×2048 stable (large-H) | 1983.0 µs | **344.6 µs** | 5.8× |

¹ The 2048-unstable incumbent is a **pre-branch silicon measurement**: on this build no routing reaches the old single-core cell for unstable bf16 W=2048 (no unstable opt-out exists), so the denominator is not re-measurable in-tree. The independent audit re-measured the numerator at 43.0 µs (implied 36.3×, consistent). Rider rows are full-iteration times (including the composite pad+slice); the sort op alone measures 33.4/42.6 µs at W=512/1024.

**Measured and deliberately NOT shipped (comparator kept):**

| cell | comparator (kept) | fast engine (dropped) | Δ | unstable floor |
| --- | ---: | ---: | ---: | ---: |
| topk W=8192 k=32 prealloc-u32 (multicore rank-stamped) | **160.8 µs** | 186.1 µs | +15.7% | — |
| topk W=8192 k=64 prealloc-u32 (multicore, split-call) | **215.7 µs** | 237.8 µs | +10.2% | — |
| topk W=32768 k=32 prealloc-u32 (multicore) | **234.1 µs** | 263.6 µs | +12.6% | — |
| sort W=4096 stable (CrossCore, fused) | **379.9 µs** | 398.8 µs | +5.0% | — |
| sort W=8192 stable (CrossCore, fused) | **972.2 µs** | 1.001 ms | +2.9% | 968.8 µs |
| sort W=32768 stable (MultiCore DRAM, fused) | **6.533 ms** | 7.181 ms | +9.9% | 6.403 ms |

(The dropped fused engines were benchmarked against the comparator factories that served those widths at the time; on BH those widths now route to the mergesort engine — the negative result stands as the reason the comparator, not the fused transport, remained the fallback ladder.)

## Validation (this head, fresh kernel cache, single fresh runs on p150a — these tallies are from the final tree including the truth-sync fixes)

- `data_movement/test_sort.py`: **296 passed, 0 failed (2:36; includes the 6 new truth-sync cells)**.
- `reduce/test_topk.py` + `test_topk_stable_adversarial.py`: **288 passed / 8 skipped / 80 xfailed (1:48)**.
- LLK BH `test_topk.py` + `test_topk_xl.py` on silicon: **161 passed / 8 skipped (0:39)**.
- Hostile-coverage audit on the pre-truth-sync head: 57/57 adversarial cells pass (staged-boundary ties at every merge offset ×4 widths, all-ones at all 8 absorbed widths, prealloc-dtype routing, real ±inf vs pad sentinels, RM at tree widths, program-cache convergence).
- **WH**: no Wormhole LLK tree is touched (topk_xl is BH-only; the sort routing is arch-gated). Shared LLK test-helper changes re-verified under WH ttsim: `test_topk.py` 36 passed / 4 skipped (comparator-stable modes deselected — pre-existing ttsim SFPSWAP-index-operand gap). Earlier rank-stamp WH ttsim leg: 42 passed / 8 skipped over all simulable modes. The sort factories are BH-silicon-validated; the only WH-visible behavior change is the prim pow2 validation (see disclosures). One WH CI pass recommended.
- **CI**: PR Gate fully green on the pre-truth-sync head (36 pass / 0 fail); l2-nightly + sweeps legs ran one commit earlier with zero sort/topk failures (the four red nightly classes are pre-existing-on-main or base-stale, all fixed on main — the wh_n150 reduction timeout that skipped the n150 topk leg clears on rebase past faed88c3791; follow-up #54050).

## Limitations

> The mergesort row engine serves **bfloat16** rows of power-of-two padded width **2048–65536** (user widths 257–65536; pow2 pads of 512/1024 are raised to 2048 by the composite rider), **both stabilities**, on **Blackhole only** — the topk_xl LLKs have no Wormhole counterpart. Stable cells return UINT32 indices (see disclosures — the public u16 → u32 flip covers padded W=2048–32768; W=65536 was already u32); unstable cells keep UINT16 via writer-RISC narrowing, so no unstable caller sees a dtype change. W=8192–65536 run the L1-staged merge tree (ping-pong Float32 run buffers between DEST passes, ≤ 576 KB L1 at W=65536). The hard ceiling is W=65536 — the fused word's 16 free tag bits address exactly 65536 positions (zero slack); wider rows, fp32/u16 values, non-Blackhole devices, and the stable preallocated-UINT16 opt-out all decline cleanly to the comparator engines via the routing ladder. Height is unbounded (one row per core, DRAM-streamed, constant L1). The sample-sort multicore layer and the half-sync 4-tile DEST escape remain follow-ups.
>
> **Stress envelope, requalified:** the full-ladder stress campaign (W up to 524288, H up to 4096, 268M-element tensors, 1000-iteration endurance, trace capture/replay, program-cache churn — zero hangs, zero leaks, zero growth) ran at the pre-tree head 47f2cb5100a, when W ≥ 8192 was served by the comparator engines. The merge tree that now takes W=8192–65536 is covered by the per-commit failing-first suites, the 16 hostile-coverage cells, and the independent 57-cell audit — and the endurance/trace legs have now been rerun at the tree widths on the final tree (#54050): 500-iteration endurance at [32, 8192] and [32, 65536] stable — zero result drift (bit-identical outputs every iteration), latency-window drift +0.30% / −2.37% (post-JIT means 0.165 / 1.116 ms per iteration), program cache flat, and a post-loop 128 MB DRAM + L1 allocation with one more exact sort — plus trace capture/replay at [32, 8192] stable (20 MB trace region, standard warm-up): 3 replays bit-exact, a fresh-data replay bit-exact against a fresh golden, and a clean release with an exact post-release eager sort.
>
> Known sharp edges: `stable=False` duplicate indices within tie groups remain on the Blackhole **fp32** CrossCore widths and on non-Blackhole devices (issue #54043 — retired for bf16 on Blackhole by this stack); UINT16 inputs containing the pad sentinel values (**65535 ascending / 0 descending** — u16 has no ±inf) at padded boundaries can surface an out-of-range index; NaN is unsupported for bf16 (canonicalized to ±inf, deviating from torch's NaN-last); the CrossCore/MultiCore factories bake physical core coordinates, assuming uniform harvesting across mesh chips (the mergesort engine itself is SPMD-clean). Trace capture requires the standard warm-up (CrossCore builds a host→device lookup table on cache miss).

## Commit inventory (42, in stack order)

**Repair series (5):** `d563940d90c` failing-first tests for the 32-bit paths · `010a13c7624` multicore 32-bit DEST for 32-bit index CBs · `cce5118777e` reject mismatched indices_tensor dtype width · `84235781b6d` fp32 −0.0 tie-class fold (BH+WH) · `6d49857e9e2` stable-merge direction contract doc.

**Rank-stamp series C1–C4 (5):** `0b464ed2cb8` stamp/strip primitives + RANK_STAMPED mode (BH+WH) · `0456986a725` compute API surface · `160d78bf6a7` single-core rank-stamped stable gate · `d422d30ee23` wide-stable tests + perf evidence · `accaba980ae` series notes.

**Stable sort via comparator C5 (4):** `a3be64ecf5b` failing-first stable-sort tests · `9b088a73c6a` tie polarity from runtime global mode (BH+WH) · `51674adc893` ttnn.sort stable=True end-to-end · `311b54de88b` wide tie-saturation stimulus.

**Housekeeping (4):** `1e3e4ad8090` drop session-log notes · `928350b195e` drop unused runtime rank_base · `5da7dffcb3d` fused_sort CSV schema catalog · `afbeb2676f5` restore bare uint32_t spelling.

**Completion series (5):** `8e76c167b80` single-tile rank-range stamp + PRE_TAGGED merge (BH+WH) · `cc39d8619fe` chain-rank stamps through the k>32 cascade · `bbab1909b8c` fused true-index-tagged stable sort on the single-core factory · `7990d7838a6` multicore stable tie coverage (negative result kept) · `a638a36f98b` Tracy benches for the A/B cells.

**Mergesort series (4):** `6b364657ce6` small-Ht reroute to CrossCore · `3fdc8c0b599` LLK both-halves merge (BH) · `bbb4cb7490d` mergesort row engine, stable bf16 W=2048/4096 · `47f2cb5100a` campaign Tracy benches.

**Hardening + scale-out (12):** `a9e193a56b7` rename TOPK_XL.full_sort → merge_both_halves (perf-header gate) · `f78159a5c9d` narrow the reroute to stable-only + Blackhole-only · `f33867c8216` restore permutation assertion + all-ones regression test · `df1a217c02a` document NaN unsupported for bf16 · `c331edcd37e` cosmetics (unused CT arg, SFPSWAP polarity comment) · `1f3ad00a1f5` empty-tensor guard at prim validation · `7c7c1650bb7` power-of-two width required at the prim (all-arch behavior change, see disclosures) · `4e2bbfa91b5` truth-sync the unstable-tie docs to the measured band · `45e0af8b9e8` absorb unstable bf16 W=2048/4096 (writer u32→u16 narrowing) · `5d6dfb2eb39` pad-to-2048 rider for W=512/1024 · `f37f93051f2` W=8192–65536 L1-staged merge tree · `f7e5e465831` hostile-coverage cells.

**Truth-sync fixes (3, this update):** `f15d654405b` composite early-exit index dtype aligned with the stable-bf16 u32 contract (failing-first) · `9ed54ea915c` composite early-exit for zero-size tensors, torch empty-return parity (failing-first) · `2c0eb35530a` Sort.md + docstring sync to the shipped ladder.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/rankstamp-53557)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/rankstamp-53557)


nkapreTT:
## Cross-grid evidence: BH L2-nightly at head (a72ace60420)

**Coverage gap closed first (a72ace60420):** the `reduction` category of `tt-metal-l2-nightly.yaml` only selected the `tests/ttnn/nightly/` tree, so `tests/ttnn/unit_tests/operations/reduce/test_topk.py` (and the new `test_topk_stable_adversarial.py`) never executed on any blackhole job of this pipeline. Added a `(post-commit suite)` entry to `tests/pipeline_reorg/ops_unit_tests.yaml` following the existing data_movement precedent — BH-only (`bh_p100`, `bh_p150b_civ2`; wormhole already runs the full reduce directory per-PR via the sanity `ttnn reduce group`), plus a 15-min `bh_p150b_civ2` unit-budget bump in `.github/time_budget.yaml`.

**Fresh dispatch at this head:** https://github.com/tenstorrent/tt-metal/actions/runs/32598181603 (BH only, `additional_test_categories=reduction,data_movement`)

| Suite (job) | bh_p100 | bh_p150b_civ2 |
|---|---|---|
| **topk post-commit suite (newly wired — first BH-nightly execution)** | **288 passed** / 8 skipped / 80 xfailed (2:00) | **288 passed** / 8 skipped / 80 xfailed (3:55) |
| reduction (nightly tree) | 3978 passed / 1794 skipped | 3978 passed / 1794 skipped |
| data_movement post-commit suite (incl. `test_sort.py`: 595 pass) | 8715 passed / 1162 skipped / 12 xfailed | 8716 passed / 1161 skipped / 12 xfailed |
| point_to_point post-commit suite | passed | passed |
| data_movement (nightly tree) | 1 failed / 2335 passed | 1 failed / 2335 passed |

`bh_p150b_civ2_viommu`: no jobs — neither `reduction` nor `data_movement` carries a viommu SKU entry; that runner class only selects marker-gated (`requires_host_iommu`) experimental/sdpa suites.

**Failure attribution (only red in the run, both runners, identical):** `test_universal_input_tm_fold.py::test_fold_specless_cross_layout_override_inherits_input_orientation[H_col_major_in_B_override]` — `TT_FATAL @ tensor_spec.cpp:161`. Pre-existing: fails identically at the previous head (run 32572358491), and the fix is already on main (3007bc45e2f, #53118) landed after this branch's merge-base; not in this PR's blast radius. Zero failures in the sort/topk suites this PR extends.

nkapreTT:
Rebased onto the force-pushed #53557 base at `6a700bbe685`; new exact head is `544485a4179` and GitHub now reports it mergeable (no conflict).

Rebase decisions:
- dropped the 32-bit DEST and optional-index-width commits because the updated base already contains their equivalent via #54168;
- dropped the perf-schema and bare-`uint32_t` follow-ups because they are already in the new base;
- preserved the base's newer TopK replay/fused-e2e/reinit machinery while layering the rank-stamped and mergesort paths alongside it;
- preserved the newer current perf pins rather than replaying the older IOMMU re-baseline;
- retained the current `wh_n150_civ2` budget and applied the intended +15 minute BH budget for the added post-commit TopK suite.

Validation at the exact pushed head:
- `pre-commit run --from-ref 6a700bbe685 --to-ref 544485a4179`: all hooks passed, including clang-format, Black, LLK lint, YAML, include validation, and conflict checks.
- Local LLK pytest collection is unavailable on this macOS host because the Tenstorrent `ttexalens` runtime is not installed; exact-head CI is now running for compile/device coverage.

nkapreTT:
Fixed both real failures from PR Gate run 32842593943 at exact head 544485a in commit 459675d:

- Added fused_sort to the shared nullable perf DB schema, matching the perf_eltwise_unary_sfpu catalog.
- Replaced the undefined APPROX template argument in the direct TopK XL smoke kernel with false. That parameter is unused by the underlying both-halves kernel, and false matches the accurate/default path.

Validation: all applicable pre-commit hooks pass, diff check passes, and test_catalog_columns_are_in_db_schema passes locally. The full LLK device/compiler lane is left to CI on the new exact head.

nkapreTT:
Exact-head CI confirmation for commit 459675d, PR Gate run 32845242521:

- Blackhole LLK smoke group 1/2 passed on bh_p150b_civ2_viommu.
- Blackhole LLK smoke group 2/2 passed on bh_p150b_civ2_viommu. This is the lane that previously failed test_catalog_columns_are_in_db_schema and test_topk_xl at 544485a.
- Wormhole LLK smoke passed on wh_n150_civ2.

Both repaired regressions are therefore confirmed fixed in the target CI lane. Remaining non-LLK PR Gate jobs are still running.

nkapreTT:
Full exact-head PR Gate confirmation: run 32845242521 completed successfully at commit 459675d with 26 successful jobs and zero failures. All BH/WH LLK smoke lanes passed, and the remaining Wormhole runtime validation/examples lanes also passed.

Per the proposed #53557/#54008 split, I am holding this branch unchanged until the functional cherry-picks land on #53557.

PR #54015, opened 2026-08-21
Fix pytest-xdist collection mismatch in test_to_memory_config.py

blozano-tt:
### Problem

CI fails while collecting `tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py` under pytest-xdist:

```
Different tests were collected between gw0/gw1/gw2/gw3
```

### Root cause

Diagnosis credit: **GitHub Copilot's CI failure analysis.**

Several `@pytest.mark.parametrize` argsets for the DRAM ND-sharding tests embedded **callable objects** directly as parameter values — `make_disjoint_dram_core_range_set` and `make_full_dram_core_range_set`.

These two grids depend on `device.dram_grid_size()`, so they can't be built at import time. The existing workaround passed the factory *function itself* through `parametrize` and called it in the test body:

```python
([5, 32, 64], [2, 32, 64], make_full_dram_core_range_set),
...
if callable(grid):
    grid = grid(device)
```

Test IDs for those argsets were therefore derived from these objects rather than from anything written in the source. That leaves collection dependent on how each worker process stringifies them, so workers can disagree on the collected ID set — which is exactly the failure signature above.

### Fix

Reference the device-dependent grids by **stable string keys** and resolve them to real `CoreRangeSet`s inside the test:

```python
DRAM_GRID_FACTORIES = {
    "disjoint_dram": make_disjoint_dram_core_range_set,
    "full_dram": make_full_dram_core_range_set,
}

def resolve_dram_grid(grid, device):
    if isinstance(grid, str):
        return DRAM_GRID_FACTORIES[grid](device)
    return grid
```

```python
pytest.param([5, 32, 64], [2, 32, 64], "full_dram", id="5x32x64-full_dram"),
...
grid = resolve_dram_grid(grid, device)
```

Additionally, every argset in the three affected `parametrize` blocks now carries an explicit `id=`. Those blocks also contain opaque `ttnn.CoreRangeSet(...)` values whose auto-generated IDs were positional and uninformative (`grid0`, `grid1`, …); naming them makes the collected IDs fully determined by the source file, and makes failures readable (`[ROW_MAJOR-4x3x16x32-disjoint_dram]` rather than `[ROW_MAJOR-tensor_shape4-shard_shape4-grid4]`).

Scope was kept tight: only the three blocks that contained callables were touched. The other ~84 `parametrize` blocks in this file already use cheaply-stringifiable values (ints, short strings, enums, small tuples) and are left alone.

### Tests affected

| Test | Argsets |
|---|---|
| `test_to_memory_config_rm_interleaved_to_nd_sharded_dram` | 6 |
| `test_to_memory_config_tilized_interleaved_to_nd_sharded_dram` | 7 |
| `test_to_memory_config_tilized_interleaved_to_nd_sharded_dram_dtype_conversion` | 5 |

### Behavior

Collection-stability only — no logic or coverage change.

- All parameter values are semantically identical before/after. Verified by an AST comparison of every argset in the three blocks (callable `Name` → equivalent string key), not by eyeballing the diff.
- The same grids are still produced by the same factory functions, just resolved by key instead of by passing the function object.
- `resolve_dram_grid` is equivalent to the previous `if callable(grid)` guard: `CoreRangeSet` values pass through untouched in both.
- The unrelated direct call `make_full_dram_core_range_set(device)` inside another test body is untouched — it was never a parametrize value.
- Explicit IDs verified unique within each block (duplicate IDs would otherwise get index suffixes and reintroduce positional fragility).

### Validation performed

- `python3 -m py_compile` on the modified file — clean.
- `black==23.10.1` (repo-pinned, `line-length = 120`) `--check` — unchanged.
- AST equivalence check of all parametrize argsets, plus per-block ID-uniqueness check.

Not run: the tests themselves, and `pytest --collect-only`. This environment has no `ttnn`/`torch`/`pytest` and no hardware, so collection can't be exercised here — **the xdist collection run in CI is the real verification** and should be watched on this PR.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:fix/xdist-collection-mismatch-to-memory-config)

### DIFF
--- a/tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
+++ b/tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
@@ -18,6 +18,26 @@ from tests.ttnn.utils_for_testing import (
 pytestmark = pytest.mark.use_module_device
 
 
+# DRAM core grids that can only be built once a device is available are referenced from
+# `@pytest.mark.parametrize` by a stable string key instead of by the factory callable
+# itself, so that every pytest-xdist worker collects an identical set of test IDs.
+DRAM_GRID_FACTORIES = {
+    "disjoint_dram": make_disjoint_dram_core_range_set,
+    "full_dram": make_full_dram_core_range_set,
+}
+
+
+def resolve_dram_grid(grid, device):
+    """Resolve a parametrized `grid` value into a concrete `ttnn.CoreRangeSet`.
+
+    String values are looked up in `DRAM_GRID_FACTORIES` and built for `device`;
+    already-concrete `CoreRangeSet` values are returned unchanged.
+    """
+    if isinstance(grid, str):
+        return DRAM_GRID_FACTORIES[grid](device)
+    return grid
+
+
 # Test for int types
 @pytest.mark.parametrize("shape", [[1, 1, 32, 256], [64, 64], [9, 32, 768], [128]])
 @pytest.mark.parametrize("dtype", [ttnn.uint32, ttnn.int32])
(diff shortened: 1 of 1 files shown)


blozano-tt:
Heads-up: this fix was incomplete, and CI on `main` caught it.

The same two factories (`make_disjoint_dram_core_range_set` / `make_full_dram_core_range_set`) live in the shared `tests/ttnn/utils_for_testing.py` and were embedded as `@pytest.mark.parametrize` values in **two** files. This PR fixed `test_to_memory_config.py`; `tests/ttnn/unit_tests/base_functionality/test_copy.py` was left untouched and reproduced the identical xdist "different tests were collected between workers" failure right after the merge:

- Failing test: `test_copy.py::test_copy_tilized_interleaved_to_nd_sharded_dram_dtype_conversion`
- CI run: https://github.com/tenstorrent/tt-metal/actions/runs/32538288678
- Triage output quotes `<function make_disjoint_dram_core_range_set at 0x7f474e8fb5b0>` versus a different address from another worker.

Follow-up: #54025. It fixes the three affected parametrize blocks in `test_copy.py` and also moves `DRAM_GRID_FACTORIES` / `resolve_dram_grid` out of `test_to_memory_config.py` into the shared `utils_for_testing.py`, so any future caller of those factories gets the safe pattern by default rather than re-introducing the bug.

Issue #20, opened 2023-02-21
Perf-Mon: Fetch core clock from the device api instead of hard coding


mo-tenstorrent:
At the moment clock is hard coded to 1.2Ghz in the `profiler` module. This clock needs to be extracted from the device api.

davorchap:
@mo-tenstorrent is still relevant?

mo-tenstorrent:
No

PR #54018, opened 2026-08-21
Control Plane ↔ Fabric downed-links API: FSD-backed topology map, PhysicalNodeId, LinkHealth

Riddy21:
## Problem

Fabric 2.0 reroute and DC bringup both need the delta between the topology that
*should* be wired and the one that *is*. Control Plane maps and pairs on the **live**
physical system descriptor only, so a downed factory-declared cable is invisible — and
intermesh pairing will happily bind another up cable to the same logical port.

#52859 is the umbrella. The plumbing already landed: `tt-run --factory-system-descriptor`
(#53451) and the offline FSD→PSD builder (#53857). This PR maps on the FSD when one is
present, diffs it against live, and exposes the missing edges so reroute can gate on
`fsd_rerouting_active()` and bringup can name the exact cable.

**No FSD ⇒ today's path, unchanged; every link treated healthy.**

> This PR supersedes and contains **#55170** and **#55459**. Those two were stacked on
> this branch; everything from both is squashed in here, rebased onto current `main`.

### 1. `PhysicalNodeId` — a stable physical address

A packed `(cluster_id, tray, loc)` POD. An FSD-built descriptor labels its ASICs `1..N`
in file order; a discovered one uses UMD chip unique ids. The two id spaces share
nothing, so the mapper, solver, and grouping matcher now index chips **by address**
instead of `AsicID`, and both descriptors produce the same placement.

- Mock discovery derives tray from board so the address stays unique.
- The `_<rank>` suffix on colliding cluster ids is kept — it is the only thing telling
  two ranks reporting the same cluster id apart.
- `node_id_from_asic_descriptor` is the descriptor → id seam. The `1..N` labels stay;
  nothing downstream treats them as identity.

### 2. `cluster_id` — what the address joins on

Discovery keys on the cluster descriptor's `cluster_id` field rather than a filename.
`TT_METAL_CLUSTER_ID` supplies it to UMD discovery for containers and VMs, where
`gethostname()` names the container or the guest rather than the accelerator group.
Exported-but-empty is treated as unset.

### 3. `diff_physical_system_descriptors`

Joins the FSD-built PSD against the live PSD on `(cluster_id, tray, loc)`, plus a channel
for a cable endpoint. Missing edges are the downed links.

- Cables are joined **undirected** — a descriptor stores a cross-host cable twice, once
  under each host — while the output still carries both directed halves, because callers
  key on the source end.
- `mismatched_asics` compares **board type**, not tray: tray is part of the address, so a
  moved chip reads as one address missing and another extra.
- Repeated destination entries inside one `AsicTopology` are merged before comparing.

### 4. `LinkHealth`

Owns the downed set and indexes it for logical, physical, and route-shaped queries.
Health is **presence in the live descriptor and nothing else** — no call asks the cluster
whether ethernet came up — which is what lets the whole thing run on a mock.

- Records carry physical identity from the expected side and a logical view when the
  solve placed both ends.
- Downed **intermesh** links keep `src_direction` / `dst_direction` at `NONE`: pairing
  only ever assigns live cables. They are still downed links and still returned.
- Holes on routing planes fabric already downgraded away are classified into a separate
  set — real unplugged cables, documented, but not something to reroute for.

### 5. FSD host filter

A factory descriptor in the wild covers a superpod or a datacenter while a job runs on a
slice of it, so the descriptor is restricted to the **live descriptor's host set** before
it is built. Discovery ran across exactly the ranks the job owns, so that set already is
the allocation; a second source of truth would let the two lists disagree with nothing to
notice. Names are matched through the mapper's canonicalization on both sides.

The derivation is a faithful dry run of the ingest (`validate_host_filter`), and its
outcome is reduced across ranks alongside a host-list checksum and a file fingerprint, so
**either every rank continues or every rank throws the same bytes** — a rank that threw
alone would leave the others hanging at the next collective.

### 6. ControlPlane wiring

With `TT_METAL_FACTORY_SYSTEM_DESCRIPTOR_PATH` set, init derives the host filter, builds a
PSD from the factory descriptor, solves the mesh on it, pairs intermesh on live + ETH-up,
and hangs a `LinkHealth` off the result.

- Where an FSD is set, the STRICT fatals for missing connections and short link counts
  become **warnings** — a factory descriptor is expected to describe cables that are
  currently down, which is the whole point.
- A factory chip **absent from live** is still fatal, and every absent chip is reported,
  sorted, rather than the first one found. A missing cable is the feature; a missing chip
  is a wrong allocation.
- With no FSD, `descriptor_to_map_on()` returns the discovered descriptor and every public
  forwarder is a null-safe no-op.

Phase 1 (`generate_rank_bindings`) maps on the FSD PSD when the path is set.

## API

`link_health.hpp` types; `ControlPlane` forwards. No FSD ⇒ healthy / empty / `false` / `0`.
`get_link_health()` is `nullptr`.

```cpp
enum class LinkScope { IntraMesh, InterMesh, Unknown };

struct LinkInfo {
    FabricNodeId src_node;  chan_id_t src_chan;
    FabricNodeId dst_node;  chan_id_t dst_chan;
    RoutingDirection src_direction = RoutingDirection::NONE;
    RoutingDirection dst_direction = RoutingDirection::NONE;
    LinkScope scope = LinkScope::Unknown;
    bool logical_resolved = false;
    bool is_intramesh() const;
    bool is_intermesh() const;
    std::string src_host; TrayID src_tray; ASICLocation src_loc; AsicID src_asic;
    std::string dst_host; TrayID dst_tray; ASICLocation dst_loc; AsicID dst_asic;
    PortType medium;
    MeshId src_mesh() const;
    MeshId dst_mesh() const;
};

const LinkHealth* get_link_health() const;  // nullptr if no FSD
const std::vector<LinkInfo>& get_downed_links() const;
void refresh_connectivity_diff();
bool has_downed_links() const;
bool fsd_rerouting_active() const;

bool is_link_healthy(FabricNodeId, chan_id_t) const;           // out_of_range if not FSD-expected
std::optional<LinkInfo> find_downed_link(FabricNodeId, chan_id_t) const;
std::vector<LinkInfo> get_downed_links(FabricNodeId) const;
std::vector<chan_id_t> get_downed_eth_chans(FabricNodeId) const;
std::vector<chan_id_t> get_downed_eth_chans_in_direction(FabricNodeId, RoutingDirection) const;
bool has_downed_link_in_direction(FabricNodeId, RoutingDirection) const;
size_t get_num_downed_routing_planes_in_direction(FabricNodeId, RoutingDirection) const;
std::vector<chan_id_t> get_downed_intramesh_eth_chans(FabricNodeId) const;
std::vector<chan_id_t> get_downed_intermesh_eth_chans(FabricNodeId) const;
std::vector<LinkInfo> get_downed_links(LinkScope) const;
std::vector<LinkInfo> get_downed_intramesh_links() const;
std::vector<LinkInfo> get_downed_intermesh_links() const;
std::vector<LinkInfo> get_downed_links_between(FabricNodeId src, FabricNodeId dst) const;
std::vector<chan_id_t> get_downed_forwarding_eth_chans_to_chip(FabricNodeId src, FabricNodeId dst) const;
std::vector<LinkInfo> get_downed_intermesh_links(MeshId src, MeshId dst) const;
std::vector<FabricNodeId> get_exit_nodes_with_downed_links(MeshId src, MeshId dst) const;

bool is_link_healthy(const std::string& host, TrayID, ASICLocation, chan_id_t) const;
bool is_link_healthy(AsicID, chan_id_t) const;
std::vector<LinkInfo> get_downed_links_for_host(const std::string& host) const;
std::vector<LinkInfo> get_downed_links_for_asic(AsicID) const;
std::vector<LinkInfo> get_downed_links_between_hosts(const std::string& a, const std::string& b) const;
```

Contract: [`README_downed_links_contract.md`](https://github.com/tenstorrent/tt-metal/blob/ridvan/downed-links-contract/tt_metal/fabric/README_downed_links_contract.md).

## Submodule uplift — both on `main`, nothing pinned to a feature branch

| Submodule | Pin | Why |
|---|---|---|
| `tt_metal/third_party/umd` | `de81ff1d` (upstream `main`) | carries [tt-umd#3349](https://github.com/tenstorrent/tt-umd/pull/3349) — optional `cluster_id` on the cluster descriptor, stamped from `TopologyDiscoveryOptions::cluster_id` or `gethostname()` |
| `tt_metal/third_party/tt-cluster-descriptors` | `e3c31f2` (upstream `main`) | carries [#18](https://github.com/tenstorrent/tt-cluster-descriptors/pull/18) (blackhole supercluster factory system descriptors) and [#20](https://github.com/tenstorrent/tt-cluster-descriptors/pull/20) (`cluster_id` populated on the descriptors) |

Both upstream PRs are merged, so neither pin is a feature branch any more.

## Test-driver fixes (pre-existing, found by this work)

- `run_test` could never record a failure. Under `set -eo pipefail` a failing command in an
  `if`/`else` body aborts the script, so the status check was unreachable, nothing landed
  in `FAILURES`, and `--keep-going` did nothing.
- A `--gtest_filter` matching nothing exits 0, so a renamed or never-compiled suite reported
  a pass for **zero tests**. That immediately caught `*LogicalToPhysicalConversionFixture*`,
  whose file is in no `sources.cmake` and does not compile — the APIs it covers were deleted.
- `timeout --foreground` signalled only `tt-run`, leaving its ranks orphaned and spinning in
  MPI finalize; one timed-out test leaked a full rank set for the rest of the run.
- The automapper determinism run was invoked directly rather than through `run_test`, so a
  non-zero exit was discarded.

## Plans

- [`PLAN_fsd_solve_and_downed_links.md`](tt_metal/fabric/PLAN_fsd_solve_and_downed_links.md) — implementation plan
- [`PLAN_physical_node_id.md`](tt_metal/fabric/PLAN_physical_node_id.md) — the packed address
- [`PLAN_downed_links_testing.md`](tt_metal/fabric/PLAN_downed_links_testing.md) — test plan
- [`PLAN_umd_cluster_descriptor_hostname.md`](tt_metal/fabric/PLAN_umd_cluster_descriptor_hostname.md) — the UMD `cluster_id` field
- [`README_downed_links_contract.md`](tt_metal/fabric/README_downed_links_contract.md) — the contract

## Local verification

Built `fabric_unit_tests` + `generate_rank_bindings` against the uplifted submodules (clang-20,
Release) — clean. CPU-only offline + mock suites on `bh-glx-110-c06u08`:

| Suite | Result |
|---|---|
| `PhysicalNodeIdTest` | ✅ 17/17 |
| `PhysicalSystemDescriptorDiff` | ✅ 17/17 |
| `FsdHostFilter` | ✅ 15/15 |
| `FsdChipsAbsentFromLive` | ✅ 6/6 |
| `PhysicalDescriptorBuilder` | ✅ 18/18 |
| `TopologySolverTest` | ✅ 146/146 |
| `TopologySatEncoderTest` | ✅ 7/7 |
| `TopologyMapperUtilsTest` | ✅ 62/62 |
| `PhysicalGroupingDescriptorTests` | ✅ 25/25 |
| `MeshGraphDescriptorTests` | ✅ 63/63 |
| `LinkHealthTest` (t3k mock) | ✅ 16/16 |
| `ControlPlaneFixture.NoFactoryDescriptor*` / `.RefreshWithout*` | ✅ 5/5 |
| `ControlPlaneFixture.TestT3kControlPlaneInit*` | ✅ 1/1 |
| `TopologyMapperTest.T3kMeshGraphTest*` | ✅ 3/3 |
| `FactoryDescriptorControlPlaneFixture.*` | ❌ see below |

### Known failure — FSD test assets, not the code

`FactoryDescriptorControlPlaneFixture` cannot pass with any factory descriptor currently in
`tt-cluster-descriptors`. Two separate asset problems, both **pre-existing** — the SC20 and SC36
assets are byte-identical between this branch's previous submodule pin and the uplifted `main` tip,
so neither is caused by the uplift.

1. **The driver points at a descriptor that does not describe its own cluster.** The CPU-only driver
   runs the fixture against `SC36_32x4_revC_subtorus_aisleD`, whose cluster descriptors carry
   `cluster_id: bh-glx-120-*` while its FSD names `bh-glx-110-*` — **zero host overlap**, so the host
   filter correctly refuses the descriptor and all 3 tests throw. The same mismatch affects
   `SC16_32x4_revAB_aisleD`, `SC36_32x4_revAB_subtorus_aisleC` and `SP3_16x8_revAB_aisleC`;
   `SC20_32x4_revAB_aisleC` is off by the rack token (`bh-glx-c01u02` vs `bh-glx-110-c01u02`).
   Only `SC20_32x4_revC_subtorus_aisleC` (20/20) and `SC16_32x4_revC_*` line up.

2. **Even on the matching asset, FSD and live disagree.** Pointed at
   `SC20_32x4_revC_subtorus_aisleC` the host filter resolves and the FSD path runs end to end —
   `EveryMappedNodeResolvesToARealChip` passes — but the run reports **16 of 240 expected
   connections downed (6.7%)**, so `AMatchingDescriptorReportsNoDownedLinks` and
   `RefreshAgainstUnchangedLiveIsStable` still fail. These are not boundary cables:
   `build_sub_fsd` already drops any connection with an endpoint on a filtered-out host. They are
   intra-host cables that the FSD declares and that host's own live descriptor does not have.

So the assertion "a matching descriptor reports no downed links" has no matching descriptor to run
on yet. The fix is in `tt-cluster-descriptors` — reconcile the cluster-descriptor `cluster_id`s with
the FSD host names, and reconcile the per-host wiring — not in this PR. Left failing rather than
re-pointed at a different asset, so the gap stays visible.

Note that this failure is only *visible* because of the `run_test` fixes in this PR: before them a
`--gtest_filter` that matched nothing exited 0, and a failing command in the driver aborted the
script without recording anything.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ridvan/downed-links-contract)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ridvan/downed-links-contract)

### DIFF
--- a/tt_metal/fabric/PLAN_physical_node_id.md
+++ b/tt_metal/fabric/PLAN_physical_node_id.md
@@ -28,13 +28,13 @@ PhysicalNodeId = {
 
 Do **not** pack the host as `hash32` into a `uint64`. After tray and loc take 32 bits, only four characters would fit; `bh-glx-110-c01u02` is 18. A 128-bit pack is still only 12 host chars. The hash was the only way to stay in 64 bits, and it is the conversion this plan drops.
 
-Do **not** use FSD `host_id` (0..N-1 in file order): live discovery has no FSD index.
+Do **not** use FSD `host_id` (0..N-1 in file order): live discovery has no FSD index. That index is not the UMD cluster descriptor's new `cluster_id` string (§8.2) — the FSD one is a per-file position, the UMD one is the accelerator-group identifier that goes into `host[]`. The two no longer share a name; where it is still ambiguous, say "FSD host index".
 
 Do **not** use `std::string` inside the node (heap, allocator-sensitive compare). Do **not** use `std::array` — the host is a fixed buffer we never pass around on its own; a C array is the pack.
 
 UMD unique ids stay on `ASICDescriptor::umd_unique_id` / `Cluster::get_unique_chip_ids()` — they are not mapper identity.
 
-**Mock + FSD is the load-bearing case.** The name that goes into `host[]` is the **exact** FSD / OS hostname. Mock filenames are not that name — §8.1. The fix is a UMD `hostname:` field — §8.2. Until assets are filled, the aisle-token fallback in §8.2. If mock and FSD pack different `host[]` bytes, the solve diverges again. Work list: §8.3.
+**Mock + FSD is the load-bearing case.** The name that goes into `host[]` is the **exact** FSD / OS hostname. Mock filenames are not that name — §8.1. The fix is a UMD `cluster_id:` field — §8.2. Until assets are filled, the aisle-token fallback in §8.2. If mock and FSD pack different `host[]` bytes, the solve diverges again. Work list: §8.3.
 
 ---
 
@@ -564,7 +564,7 @@ builder fills ASICDescriptor          discovery fills ASICDescriptor
 - Empty / unset id is all zeros; `make_*` never returns it.
 - Golden vector: one fixed host/tray/loc → exact `host[]` bytes + tray + loc.
 - **Stability (the load-bearing one):** build one graph from FSD and one from a live-style descriptor with UMD-like `umd_unique_id`s; adjacency maps keyed by `PhysicalNodeId` are equal. `TopologyMapper` (or the solver on those maps) assigns the same `FabricNodeId` per position.
-- **Mock + FSD (the other load-bearing one):** descriptor with `hostname: bh-glx-110-c01u02` packs the same id as the FSD builder. Filename-only (no field) must **not** equal that id. After the UMD field is filled, SC36 rank 15 packs `bh-glx-110-d10u20`, not the `120` in the filename.
+- **Mock + FSD (the other load-bearing one):** descriptor with `cluster_id: bh-glx-110-c01u02` packs the same id as the FSD builder. Filename-only (no field) must **not** equal that id. After the UMD field is filled, SC36 rank 15 packs `bh-glx-110-d10u20`, not the `120` in the filename.
 - Builder unit test: QuietBox / a tiny in-memory FSD, two ASICs, ids equal `make_physical_node_id` of their FSD hostnames, not `1` and `2`.
 - Discovery: graph key ≠ `umd_unique_id` on silicon when UMD ids are large; `get_asic_id(host, tray, loc)` returns the packed id.
 
@@ -578,7 +578,7 @@ Existing mapper tests that assert one specific mapping may need a re-baseline (i
 
 `PhysicalNodeId.host[]` is the **exact** OS / FSD hostname (`bh-glx-110-c01u02`). Hall stays. FSD builder already has that string. Silicon `gethostname()` is that string.
 
-Mock discovery does **not**. A UMD cluster descriptor YAML has no hostname field. `get_local_discovery_hostname()` (`physical_system_discovery.cpp` ~60) returns the **filename basename** of `TT_METAL_MOCK_CLUSTER_DESC_PATH`:
+Mock discovery does **not**. A UMD cluster descriptor YAML has no field naming its host. `get_local_discovery_hostname()` (`physical_system_discovery.cpp` ~60) returns the **filename basename** of `TT_METAL_MOCK_CLUSTER_DESC_PATH`:
 
 ```cpp
 return std::filesystem::path(mock_cluster_desc_path).filename().string();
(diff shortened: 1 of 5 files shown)


tt_metal/fabric/PLAN_physical_node_id.md
```diff
@@ -0,0 +1,710 @@
...
+| Filename token (SC36 aisle D) | hall **wrong** | file `bh-glx-120-d10u20` vs FSD `bh-glx-110-d10u20` |
+| Silicon | `gethostname()` | `bh-glx-110-c01u02` |
+
+QuietBox already matches (`sjc1-tt-qb-01` on both sides). BH supercluster mock does not.
+
+If we pack the basename or the filename token, FSD-built and mock-discovered `PhysicalAdjacencyMap`s have different keys. The solver sees two graphs. That is the same bug as `1..N` vs UMD ids. Parsing a hostname out of the filename cannot work: SC20 omits the hall, SC36 disagrees on it. Silicon would then miss if we stripped hall from the FSD name.
+
+### 8.2 Solution
+
+Put the real hostname **on the cluster descriptor** and query it. Do not parse the filename at runtime.
```
p1-0tr: We may need to externally supply host identities, e.g. via an optional env variable (~`TT_METAL_HOST_ID`). Like that we'll leave ourselves a way to deal with containerised or virtualised environments. Unfortunately, making UMD host name aware, has the same issues.

PR #54019, opened 2026-08-21
[skip ci] bug-check workflow should get triggered by pull_request_target

blozano-tt:
Chang bug-checker trigger conditions. Goal being get it to actually run after IP white listing.


ttnn/cpp/ttnn/operations/data_movement/CMakeLists.txt
```diff
@@ -368,3 +368,5 @@ install(TARGETS ttnn_op_data_movement LIBRARY COMPONENT tar)
 if(TARGET ttnn)
     target_sources(ttnn PRIVATE ${TTNN_OP_DATA_MOVEMENT_NANOBIND_SRCS})
 endif()
+
+# bug-check tailscale diagnostic (tenstorrent/tt-metal#54007 test) — safe to revert/ignore, no functional change
```
blozano-tt:
```suggestion
```

PR #54030, opened 2026-08-22
Add optional-binding lookups for Metal 2.0 kernel resources

riverwuTT:
### Problem

A Metal 2.0 kernel's resources arrive as codegen-emitted namespace-scope tokens — `tensor::x`, `dfb::x`, `scratch::x`. A token is emitted **only if the host bound that resource to that kernel**, so merely *naming* an unbound resource is a compile error. There is currently no way to express a kernel parameter the host may or may not supply; the only workaround is a host-set `-D` plus an `#ifdef` in the kernel.

### Approach

This implements **option 3** of the optional-memory-construct-arguments design notes — the device-side lookup, chosen there for the lowest implementation and maintenance burden ("self-contained in genfile"). The host `ProgramSpec` API is untouched and still lists only what actually exists.

Every Metal 2.0 kernel's `kernel_bindings_generated.h` now carries three name-keyed lookups:

```cpp
try_get_tensor_binding<"name">()      -> std::optional<::tensor::name_t>
                                         // or std::optional<::binding::NullTensorBindingToken>
try_get_dfb_binding<"name">()         -> std::optional<::DFBBindingToken>
try_get_scratchpad_binding<"name">()  -> std::optional<::ScratchpadBindingToken>
```

They are emitted **unconditionally**, including for a kernel that binds nothing at all — "nothing was bound" is an answer the kernel is entitled to ask for, and that is the case the lookups exist to serve.

A kernel then writes:

```cpp
constexpr auto gamma = try_get_dfb_binding<"gamma">();
if constexpr (gamma.has_value()) {
    DataflowBuffer g(*gamma);
    // ... fused path
}
```

Two properties make this work:

- **`constexpr`** — emptiness is a compile-time constant, so the absent branch is *provably dead* rather than ill-formed. Both the plain runtime `if` and `if constexpr` idioms are supported.
- **class-type NTTP** (`binding::Name`) rather than a runtime string — a runtime name could defeat inlining, which the design notes call out as a performance risk.

### What each resource kind needed

| Kind | Change |
|---|---|
| **DFB** | none — `DFBBindingToken` is one concrete type, so the optional's type is uniform |
| **Scratchpad** | none — same, `ScratchpadBindingToken` |
| **LocalTensorAccessor** | a null stand-in token + one constructor |

The tensor path is the odd one out because each binding has its own `TensorBindingToken<CTA, ADDR>` specialization. The absent arm therefore needs a stand-in type (`binding::NullTensorBindingToken`), and `LocalTensorAccessor` needs a constructor taking it — a discarded `if constexpr` branch is only left *uninstantiated* inside a template, and `kernel_main` is not a template, so the absent branch is still fully type-checked. Without that overload a kernel could not compile the very branch the lookup exists to let it write.

### One consequence worth flagging

The generated file's include block is now **unconditional**. A lookup names `DFBBindingToken` / `ScratchpadBindingToken` in its return type whether or not the kernel binds a resource of that kind — which is exactly the case the lookups exist to serve — so `api/dataflow/dataflow_buffer.h` and `api/scratchpad.h` are now included for **every** Metal 2.0 kernel, where previously each was included only when the kernel had a binding of that kind.

Both headers already support that configuration: `dataflow_buffer.h` has `COMPILE_FOR_TRISC` branches, and `trisc.cc` defines symbols specifically "so kernels that include dataflow_buffer headers link cleanly". So this is a supported shape, not a new one — but it does mean more preprocessing per kernel, and it is the part of this PR most worth a second opinion.

### Anti-drift

`genfiles.cpp` and `emulated_program_runner.cpp` both emit these namespaces and must stay text-equivalent, and today they do so by duplicating the emission text. The rendering here lives in one shared formatter in `tt::jit_build::utils` (`format_optional_binding_lookups` + `OPTIONAL_BINDING_LOOKUP_INCLUDES`) that both call, so the two paths cannot diverge.

### Accepted trade-off

**A misspelled name compiles and reports absent forever.** This is the con the design notes explicitly accept for option 3. A kernel that genuinely requires a binding can pin it:

```cpp
static_assert(try_get_tensor_binding<"in">().has_value(), "kernel requires tensor binding \"in\"");
```

### Host-side name-length cap (new)

`binding::Name` carries a fixed-capacity buffer, and its constructor silently truncates past `MAX_BINDING_NAME_LEN` (63). Two names agreeing on their first 63 characters would therefore become **indistinguishable** in a lookup, which would quietly hand back the wrong binding — a silent wrong answer, materially worse than the accepted "typo → always absent" above.

There was **no prior length limit** on accessor names: `IsValidCppIdentifier` checks only the character set. So `ProgramSpec` validation now rejects an over-long **tensor / DFB / scratchpad** `accessor_name` up front, where the failure is loud and names the kernel. Semaphores and named CT/RT args are not reachable through a lookup and are deliberately not capped. `program_spec.cpp` includes `optional_binding.h`, so the host check and the device buffer cannot drift.

### Known limitation / follow-up

`TensorAccessor` (the NOC/global accessor) is **not** covered — only `LocalTensorAccessor`. `TensorAccessor` is CTAD-based on a `DSpec`, so an absent arm would need a synthetic DSpec deduction guide, which is a materially bigger change than the rest of this PR. That does partly limit the canonical optional-`gamma` layernorm case when gamma is a DRAM tensor. Worth doing as a follow-up.

### Validation

- `genfiles.cpp`, `jit_build_utils.cpp` and the new test TU all pass `-fsyntax-only` clean under the project's own compile flags (`-std=c++20 -Wall -Werror`).
- **Generated code compiles on the real toolchain.** The formatter's actual output was fed through the pinned **sfpi 7.70.0** with the production device flag set (`-std=c++17 -ftt-nttp -ftt-constinit -ftt-consteval -ftt-no-dyninit -ffast-math ... -Wall -Werror`) for three scenarios — nothing bound / partial / all three kinds — using the same optional-parameter kernel source in each case. All compile clean.
- **Presence and absence fold away entirely.** `objdump` of those three objects, where the expected sums were 7000 / 4003 / 17:

  ```
  nothing bound:  lui a0,0x2; addi a0,a0,-1192  # 0x1b58 = 7000
  partial:        lui a0,0x1; addi a0,a0,-93    # 0x0fa3 = 4003
  all three:      li  a0,17
  ```

  Three instructions each — every lookup resolved at compile time and every absent branch deleted outright. Zero runtime cost.
- New unit tests in `tests/tt_metal/tt_metal/jit_build/test_optional_binding_lookups.cpp`: exact emitted text per kind (the spelling is the contract, and it is what the emule path must reproduce), independence of the three kinds, arm-order stability (this text feeds a per-object dephash cache), `constexpr`/`[[nodiscard]]` on every lookup, include-block coverage, and an end-to-end compile of the emitted lookups asserting presence/absence resolves at compile time (against the real `api/optional_binding.h`; the three token types are stood in for, since all three real ones are device-only).

- The host cap is covered by two new `test_program_spec.cpp` cases beside the existing `IsValidCppIdentifier` smoke test: a 64-character name is rejected, and a name at exactly the 63-character limit is accepted (the boundary itself is legal).

`emulated_program_runner.cpp` was **not** locally syntax-checked — it is gated on `TT_METAL_USE_EMULE` and so appears in no `compile_commands.json`. The change there is three name-extraction loops plus two calls to a signature already proven to compile, in a TU that already uses `tt::jit_build::utils::format_named_ct_arg_map`. Leaving that to CI.

`program_spec.cpp` and `test_program_spec.cpp` could not be locally syntax-checked either, for a different reason: both reach `llrt/hal.hpp`, which includes the CMake-generated `hal/generated/dev_msgs.hpp` — a file that does not exist without a configured build. Verified that this failure is **identical on the pristine baseline**, i.e. not introduced here. Instead the two new pieces were checked in isolation with the real flag sets: the `ValidateBindingNameLength` helper compiles verbatim against the real `TT_FATAL` and real fmt (which validates the format string's arity at compile time), and the new test's include of `api/optional_binding.h` plus its `MAX_BINDING_NAME_LEN` expressions compile under the test target's own flags. A runtime check of the predicate confirms 1/8/62/63 accepted and 64/200 rejected.

Not linkable end-to-end locally (no build tree in this environment), hence draft.

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -53,6 +53,7 @@
 #include <tt-metalium/experimental/context/metal_env.hpp>
 #include <tt-metalium/experimental/mock_device/mock_device.hpp>
 #include "tt_metal/hw/inc/internal/tt-2xx/dataflow_buffer/dataflow_buffer_config.h"
+#include "tt_metal/hw/inc/api/optional_binding.h"  // binding::MAX_BINDING_NAME_LEN
 
 #include "test_helpers.hpp"
 
@@ -3900,6 +3901,34 @@ TEST_F(ProgramSpecTestGen1, InvalidTensorAccessorNameFails) {
         [&] { MakeProgramFromSpec(*mesh_device_, spec); },
         ::testing::ThrowsMessage<std::runtime_error>(
             ::testing::HasSubstr("tensor accessor_name 'has-dash' must be a valid C++ identifier")));
+}
+
+TEST_F(ProgramSpecTestGen1, OverlongTensorAccessorNameFails) {
+    // A binding's accessor_name is also baked into the device-side optional-binding lookups
+    // (try_get_tensor_binding<"name">()), whose ::binding::Name template parameter holds a
+    // fixed-capacity buffer and silently truncates past MAX_BINDING_NAME_LEN. Two names agreeing
+    // on that many leading characters would become indistinguishable there and a lookup would
+    // hand back the wrong binding, so the host rejects the name up front instead.
+    ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
+
+    const std::string too_long = "a" + std::string(::binding::MAX_BINDING_NAME_LEN, 'b');  // one over
+    spec.tensor_parameters = {MakeMinimalTensorParameter("input_tensor")};
+    BindTensorParameterToKernel(spec.kernels[0], "input_tensor", too_long);
+
+    EXPECT_THAT(
+        [&] { MakeProgramFromSpec(*mesh_device_, spec); },
+        ::testing::ThrowsMessage<std::runtime_error>(::testing::HasSubstr("characters long, which exceeds the")));
+}
+
+TEST_F(ProgramSpecTestGen1, MaxLengthTensorAccessorNameSucceeds) {
+    // The boundary itself is legal — the cap rejects only names the lookups cannot distinguish.
+    ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
+
+    const std::string at_limit = "a" + std::string(::binding::MAX_BINDING_NAME_LEN - 1, 'b');
+    spec.tensor_parameters = {MakeMinimalTensorParameter("input_tensor")};
+    BindTensorParameterToKernel(spec.kernels[0], "input_tensor", at_limit);
+
+    EXPECT_NO_THROW(MakeProgramFromSpec(*mesh_device_, spec));
 }
 
 TEST_F(ProgramSpecTestGen1, AccessorNamesAcrossCategoriesAreSeparateNamespaces) {
--- a/tests/tt_metal/tt_metal/jit_build/sources.cmake
+++ b/tests/tt_metal/tt_metal/jit_build/sources.cmake
@@ -9,5 +9,6 @@ set(UNIT_TESTS_JIT_BUILD_SRC
     test_jit_compile_deduper.cpp
     test_kernel_signature_parser.cpp
     test_named_ct_arg_map.cpp
+    test_optional_binding_lookups.cpp
     test_sync_build_steps.cpp
 )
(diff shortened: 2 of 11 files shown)


riverwuTT:
surpassed by https://github.com/tenstorrent/tt-metal/pull/54191

PR #54038, opened 2026-08-22
[Feature] Add shared Wavelet planning infrastructure (split 1/2)

ke1rro:
### PR Category

Feature

Split 1 of 2 from #52669.

### Summary

Relates to #40494. Split from #52669.

Adds the shared Wavelet planning layer: boundary handling, lifting-step types,
dependency routes, and reusable 1D/2D forward and inverse planning.


ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
```diff
@@ -0,0 +1,61 @@
...
+
+#include "ttnn/operations/wavelet/common/storage_contract.hpp"
+
+namespace ttnn::operations::wavelet {
+
+[[nodiscard]] constexpr size_t ceil_div(const size_t numerator, const size_t denominator) noexcept {
+    return denominator == 0 ? 0 : (numerator + denominator - 1) / denominator;
+}
+
+[[nodiscard]] constexpr size_t round_up(const std::size_t value, const std::size_t alignment) noexcept {
```
vtsilytskyiTT: There is such function in `tt_metal/api/tt-metalium/math.hpp`. `ceil_div` from above is also available under name `div_up`.

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include <cstddef>
+#include <cstdint>
+
+#include "ttnn/operations/wavelet/common/storage_contract.hpp"
+
+namespace ttnn::operations::wavelet {
+
+[[nodiscard]] constexpr size_t ceil_div(const size_t numerator, const size_t denominator) noexcept {
+    return denominator == 0 ? 0 : (numerator + denominator - 1) / denominator;
+}
```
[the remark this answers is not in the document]
vtsilytskyiTT: I think we should reuse `div_up` from `math.hpp`.

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
```diff
@@ -0,0 +1,366 @@
...
+WAVELET_EXTENSION_ALWI void visit_smooth_source_indices_i32(
+    const SmoothIndexI32& extended, const SourceIndexConsumer& consume) noexcept {
+    consume(extended.source_index);
+    if (extended.affine) {
+        consume(extended.auxiliary_index);
+    }
+}
+
+template <BoundaryMode Mode, typename SourceIndexConsumer>
+WAVELET_EXTENSION_ALWI void visit_extended_source_indices_i32(
```
vtsilytskyiTT, line 378: Nit: I'd prefer template specialization for this function and `evaluate_extended_index_i32` above. You did great with `make_i32`, just keep the same style here.

ttnn/cpp/ttnn/operations/wavelet/device/protocol/lwt_config.hpp
```diff
@@ -0,0 +1,80 @@
...
+constexpr uint32_t kLwtHalfStickBytes = kLwtHalfStickElements * sizeof(float);
+constexpr uint32_t kLwtNarrowTileElements = kLwtRowsPerGroup * kLwtHalfStickElements;
+constexpr uint32_t kLwtNarrowTileBytes = kLwtNarrowTileElements * sizeof(float);
+constexpr uint32_t kLwtGroupOutputElements = kLwtRowsPerGroup * kLwtOutputBlocksPerRow * kLwtHalfStickElements;
+constexpr uint32_t kIlwtGroupOutputElements = 2 * kLwtGroupOutputElements;
+
+constexpr uint32_t kRouteConfigWordCount = 16;
+constexpr uint32_t kRouteConfigPageBytes = kRouteConfigWordCount * sizeof(uint32_t);
+
+enum RouteConfigWord : uint32_t {
```
vtsilytskyiTT:
```suggestion
enum class RouteConfigWord : uint32_t {
```
Prefer `enum class` if possible. Enum is not used in this PR, so it depends on your future changes.

ttnn/cpp/ttnn/operations/wavelet/device/protocol/lwt_config.hpp
```diff
@@ -0,0 +1,80 @@
...
+constexpr uint32_t kRouteFlagBaseTileMirror = 1U << 5;
+constexpr uint32_t kRouteFlagOutputTileMirror = 1U << 6;
+
+constexpr uint32_t kLwtChunkConfigWordCount = 16;
+constexpr uint32_t kLwtChunkConfigPageBytes = kLwtChunkConfigWordCount * sizeof(uint32_t);
+static_assert(
+    kLwtChunkConfigPageBytes == kRouteConfigPageBytes,
+    "The shared 1D config-page loader requires identical chunk and route page sizes");
+
+enum LwtChunkConfigWord : uint32_t {
```
vtsilytskyiTT:
```suggestion
enum class LwtChunkConfigWord : uint32_t {
```

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
```diff
@@ -0,0 +1,589 @@
...
+[[nodiscard]] inline std::vector<RequiredStreams> backpropagate_requirements(
+    const LiftingForwardPlan& plan,
+    const IndexInterval final_even,
+    const IndexInterval final_odd,
+    const size_t closure_extent = 0) {
+    const auto close_interval = [closure_extent](const IndexInterval interval, const size_t stream_length) {
+        if (closure_extent == 0 || interval.empty()) {
+            return interval;
+        }
+        const size_t begin = (interval.begin / closure_extent) * closure_extent;
```
vtsilytskyiTT:
```suggestion
        const size_t begin = tt::round_down(interval.begin, closure_extent);
```

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
```diff
@@ -0,0 +1,589 @@
...
+                    odd_length == route.source_length && even_length == route.output_length,
+                    "Update route length state is inconsistent with the forward plan");
+                const uint32_t k = coefficient_count(route);
+                before.even = translated(after.even, route.base_offset);
+                before.odd = hull(after.odd, translated(after.even, route.source_offset, static_cast<size_t>(k - 1)));
+                even_length = route.base_length;
+                odd_length = route.source_length;
+                break;
+            }
+            case StepType::kScaleEven:
```
vtsilytskyiTT: Once you have curly brackets, another time you don't. Be consistent.

ttnn/cpp/ttnn/operations/wavelet/planner/plan.hpp
```diff
@@ -0,0 +1,286 @@
...
+}
+
+}  // namespace detail
+
+template <typename Scheme>
+[[nodiscard]] LiftingForwardPlan make_forward_lifting_plan(
+    const SignalBuffer& input, const BoundaryMode boundary_mode = BoundaryMode::kSymmetric) {
+    static_assert(Scheme::tap_size > 0, "Static lifting schemes must have a positive tap size");
+    static_assert(Scheme::num_steps > 0, "Static lifting schemes must have at least one step");
+    TT_FATAL(input.element_size_bytes == sizeof(float), "Forward lifting plan currently supports fp32 only");
```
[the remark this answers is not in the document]
vtsilytskyiTT: Agree with copilot.

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
```diff
@@ -0,0 +1,589 @@
...
+        chunk_count =
+            static_cast<uint32_t>(std::min(static_cast<uint64_t>(final_group_count), uint64_t{2} * chunk_count));
+    }
+
+    const uint32_t groups_per_chunk =
+        static_cast<uint32_t>(ceil_div(static_cast<size_t>(final_group_count), chunks.size()));
+    double max_dependency_overhead = 0.0;
+    for (const auto& chunk : chunks) {
+        max_dependency_overhead = std::max(max_dependency_overhead, chunk.dependency_overhead);
+    }
```
vtsilytskyiTT: `std::max_element` with custom lambda?

ttnn/cpp/ttnn/operations/wavelet/planner/inverse_plan_2d.hpp
```diff
@@ -0,0 +1,516 @@
...
+
+    const uint32_t output_tiles_y =
+        plan_2d_detail::checked_u32(ceil_div(y_plan.original_length, kTileHeight2D), "2D ILWT output tile rows");
+    const uint32_t output_tiles_x =
+        plan_2d_detail::checked_u32(ceil_div(x_plan.original_length, kTileWidth2D), "2D ILWT output tile columns");
+    plan_2d_detail::Candidate best{};
+    bool found = false;
+    for (uint32_t tiles_y = 1; tiles_y <= output_tiles_y; ++tiles_y) {
+        for (uint32_t tiles_x = 1; tiles_x <= output_tiles_x; ++tiles_x) {
+            std::vector<Lwt2DChunkPlan> chunks = inverse_2d_detail::build_chunks(y_plan, x_plan, tiles_y, tiles_x);
```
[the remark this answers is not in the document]
vtsilytskyiTT: Agree with copilot. 2 loops here plus 2 more loops inside `build_chunks` - way too expensive. And most results are discarded in lines 323-325. Try to add heuristics, binary search or look for some clever algorithm, please.

ttnn/cpp/ttnn/operations/wavelet/planner/plan.hpp
```diff
@@ -0,0 +1,286 @@
...
+struct StreamState {
+    int shift{0};
+    size_t length{0};
+};
+
+[[nodiscard]] constexpr RouteOutputRef workspace_output(const StorageSlot slot) noexcept {
+    return RouteOutputRef{.storage = RouteOutputStorage::kWorkspaceSlot, .slot = slot};
+}
+
+[[nodiscard]] inline std::tuple<int, size_t, size_t, size_t> compute_step_geometry(
```
vtsilytskyiTT: Tuple with 4 elements is better represented as structure.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1031 @@
...
+[[nodiscard]] inline size_t aligned_interval_span(
+    const IndexInterval interval, const size_t tile_extent, const char* label) {
+    if (interval.empty()) {
+        return 0;
+    }
+    TT_FATAL(
+        interval.end <= std::numeric_limits<size_t>::max() - (tile_extent - 1),
+        "{} interval end cannot be rounded to a tile",
+        label);
+    const size_t begin = (interval.begin / tile_extent) * tile_extent;
```
vtsilytskyiTT:
```suggestion
    const size_t begin = tt::round_down(interval.begin, tile_extent);
```

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1031 @@
...
+}
+
+[[nodiscard]] constexpr uint64_t staging_class_cycles(const AlignmentCostClass tile_class) noexcept {
+    switch (tile_class) {
+        case AlignmentCostClass::kExact: return 900;
+        case AlignmentCostClass::kOneAxisShifted: return 7'000;
+        case AlignmentCostClass::kGeneric: return 9'000;
+    }
+    return 70'000;
+}
```
vtsilytskyiTT, line 819: I believed those numbers are not measured. I'm fine with that, just put a comment that numbers represents an order of magnitude, not a real profiling results. Same below.

vtsilytskyiTT (changes requested):
Thanks for the contributions! Generally I'm Ok with the changes. Code looks solid and well-thoughts. There are few aspects to address though:
1. AI comments;
2. Code is not added to build systems. We don't know whether it builds at all.
3. Code is not tested. Could you please add a CPU test: Haar + symmetric, n even and odd, 1D round-trip and pywt coeff match, plus one 2D dwt2 band check?

ttnn/cpp/ttnn/operations/wavelet/planner/plan.hpp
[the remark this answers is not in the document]
ke1rro: Done in 498ba96, callers now cannot construct geometry which is not compatible with device protocol

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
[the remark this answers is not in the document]
ke1rro: Fixed in the 498ba96

ttnn/cpp/ttnn/operations/wavelet/common/storage_contract.hpp
```diff
@@ -0,0 +1,17 @@
...
+
+#pragma once
+
+#include <cstdint>
+
+namespace ttnn::operations::wavelet {
+
+inline constexpr uint32_t kStickWidth = 32;
+inline constexpr uint32_t kStickPageBytes = kStickWidth * sizeof(float);
+inline constexpr uint32_t kStorageAlignmentBytes = 32;
```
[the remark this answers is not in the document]
ke1rro: Removed the fixed alignment from the planner contract

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1031 @@
...
+        band_tiles_x_size);
+    const uint32_t band_tiles_y = static_cast<uint32_t>(band_tiles_y_size);
+    const uint32_t band_tiles_x = static_cast<uint32_t>(band_tiles_x_size);
+
+    plan_2d_detail::Candidate best{};
+    bool found = false;
+    for (uint32_t tiles_y = 1; tiles_y <= band_tiles_y; ++tiles_y) {
+        for (uint32_t tiles_x = 1; tiles_x <= band_tiles_x; ++tiles_x) {
+            std::vector<Lwt2DChunkPlan> chunks =
+                plan_2d_detail::build_chunks(y_plan, x_plan, tiles_y, tiles_x, fuse_terminal_scale, route_domain);
```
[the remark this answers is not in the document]
ke1rro: Fixed 946ebd5003
Got great results. 
Some examples 
Scheme | Shape | Direction | Before | After |
-- | -- | -- | -- | -- |
db1 | 4096x4096 | LWT2D | 18.941 ms | 2.124 ms |
db1 | 4096x4096 | ILWT2D | 52.171 ms | 4.917 ms |
coif17 | 4096x4096 | LWT2D | 162.891 ms | 20.293 ms |
coif17 | 4096x4096 | ILWT2D | 438.565 ms | 33.855 ms |
coif17 | 16384x16384 | LWT2D | 1103.400 ms | 167.788 ms |
coif17 | 16384x16384 | ILWT2D | 2091.040 ms | 197.703 ms |

ttnn/cpp/ttnn/operations/wavelet/planner/inverse_plan_2d.hpp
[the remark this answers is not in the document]
ke1rro: Fixed 946ebd5

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1031 @@
...
+}  // namespace plan_2d_detail
+
+[[nodiscard]] inline Lwt2DExecutionPlan make_lwt_2d_execution_plan(
+    LiftingForwardPlan y_plan,
+    LiftingForwardPlan x_plan,
+    const uint32_t core_limit,
+    const uint64_t l1_budget_bytes,
+    const bool fuse_terminal_scale = false,
+    const bool latency_oriented_planner = false,
+    const Lwt2DRouteDomainPolicy route_domain = Lwt2DRouteDomainPolicy::kExact) {
```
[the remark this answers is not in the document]
ke1rro: Added the planner gtest target and focused CPU coverage for contracts, boundary geometry, serialization, Haar symmetric even/odd 1D, round-trip, and a 2D LL band.
498ba96
bad436e

ttnn/cpp/ttnn/operations/wavelet/device/protocol/lwt_config.hpp
ke1rro: Fixed

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
ke1rro: Updated the matching host route and tile boundary calculations to use `tt::round_down`/`tt::round_up`; exact planner output is unchanged

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
ke1rro: Normalized the affected switch-case braces and ran clang-format

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
ke1rro: Updated both forward and inverse paths to use `std::max_element` 
2f588b6
f0ccb52

ttnn/cpp/ttnn/operations/wavelet/planner/plan.hpp
ke1rro: https://github.com/tenstorrent/tt-metal/pull/54038/changes#diff-8f227e3937fdfbe68d844571f9ae92874f94f481360161e3e126d97d6f88b9c7R80
replaced with `StepGeometry`

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
ke1rro: Evaluation and source visitation now the same with `BoundaryExtensionPolicy<Mode>` specialization structure as `make_i32`.

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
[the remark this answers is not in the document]
vtsilytskyiTT: You are reinventing functions, that already exist in repo: [tt_metal/api/tt-metalium/math.hpp](https://github.com/tenstorrent/tt-metal/blob/5b8d933de0216b7c8b93af0b4e9fcb4f636817c6/tt_metal/api/tt-metalium/math.hpp)

ttnn/cpp/ttnn/operations/wavelet/device/protocol/lwt_config.hpp
vtsilytskyiTT: Thanks!

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
vtsilytskyiTT: Looks better, thanks :)

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
vtsilytskyiTT: Agree, resolved

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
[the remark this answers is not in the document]
ke1rro: > You are reinventing functions, that already exist in repo: [tt_metal/api/tt-metalium/math.hpp](https://github.com/tenstorrent/tt-metal/blob/5b8d933de0216b7c8b93af0b4e9fcb4f636817c6/tt_metal/api/tt-metalium/math.hpp)

Sorry forgot to reply. I checked `tt::div_up` from `math.hpp`, but unfortunately it has the same overflow issue this change is intended to fix.

Its current implementation uses:
```
(a + b - 1) / b
```
For example, with `a = SIZE_MAX` and `b = 2`, the intermediate `SIZE_MAX + 1` wraps to zero for `size_t`, so `tt::div_up(SIZE_MAX, 2)` returns 0 instead of (SIZE_MAX / 2) + 1.

This helper intentionally uses:
```
a / b + static_cast<size_t>(a % b != 0)
```
which is mathematically equivalent but does not require the potentially overflowing addition.

one note to this that  wavelet helper currently preserves denominator == 0 -> 0, whereas tt::div_up requires a non-zero denominator.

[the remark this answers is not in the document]
ke1rro: I used functions from `math.hpp` where is it possible, this is the only one I kept.

ttnn/cpp/ttnn/operations/wavelet/planner/inverse_plan_2d.hpp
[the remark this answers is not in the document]
vtsilytskyiTT: Brilliant! As I can see there are 2 separate optimizations:
1. Pruning the candidate space to shapes satisfying `tiles_y * tiles_x <= maximum_candidate_tile_area(l1_budget)`.
2. Equivalence-class memoization of chunk cost via `AxisConeSignature`, so only representative chunks are built instead of the whole grid.

I have some safety concerns, which I'll describe in a new set of comments. Algorithmically this change is great, thanks!

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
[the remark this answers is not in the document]
vtsilytskyiTT: Size dependence drops from quartic to linear. This is a great result, thanks!

ttnn/cpp/ttnn/operations/wavelet/common/signal_extension.hpp
vtsilytskyiTT: Thanks. IMHO looks better than a series of ifs.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+        const size_t column = linear_index % x_classes.class_ids.size();
+        return row_cost_prefix[row] +
+               (row < y_classes.class_ids.size() ? row_prefixes[y_classes.class_ids[row]][column] : 0U);
+    };
+
+    size_t linear_index = 0;
+    uint64_t estimated_latency_cycles = 0;
+    for (uint32_t core = 0; core < active_core_count; ++core) {
+        const size_t count = base + (core < extra ? 1U : 0U);
+        const uint64_t core_cycles = 30'000 + prefix_cost(linear_index + count) - prefix_cost(linear_index);
```
vtsilytskyiTT: 30k is a magic number, please name it.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+    std::vector<AxisChunkClasses> classes(static_cast<size_t>(maximum_chunk_tiles) + 1);
+    for (uint32_t chunk_tiles = 1; chunk_tiles <= maximum_chunk_tiles; ++chunk_tiles) {
+        classes[chunk_tiles] =
+            make_axis_chunk_classes(length, static_cast<size_t>(chunk_tiles) * tile_extent, build_signature);
+    }
+    return classes;
+}
+
+template <typename ChunkBuilder, typename CostEstimator>
+[[nodiscard]] std::optional<Candidate> evaluate_candidate(
```
vtsilytskyiTT: Nine parameters and two callable among them in `evaluate_candidate` mixing feasibility, cost aggregation and core scheduling — splitting out the scheduling half would read better.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+struct AxisChunkClasses {
+    std::vector<IndexInterval> representatives;
+    std::vector<uint32_t> class_ids;
+};
+
+template <typename SignatureBuilder>
+[[nodiscard]] AxisChunkClasses make_axis_chunk_classes(
+    const size_t length, const size_t chunk_extent, SignatureBuilder&& build_signature) {
+    TT_FATAL(chunk_extent > 0, "2D planner chunk extent must be positive");
+    std::vector<AxisConeSignature> signatures;
```
vtsilytskyiTT, line 628: You are using vector to keep unique values. For small number (and we expect small!) is it OK, but generally there is a better structure for distinct values - [hash] set.
Just FYI.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+        } else {
+            classes.class_ids.push_back(static_cast<uint32_t>(std::distance(signatures.begin(), existing)));
+        }
+        begin = end;
+    }
+    return classes;
+}
+
+template <typename SignatureBuilder>
+[[nodiscard]] std::vector<AxisChunkClasses> make_axis_candidate_classes(
```
vtsilytskyiTT, line 649: First element of the result vector is always 0. Please leave a comment in the code about this fact and the reason why.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+}
+
+struct AxisChunkClasses {
+    std::vector<IndexInterval> representatives;
+    std::vector<uint32_t> class_ids;
+};
+
+template <typename SignatureBuilder>
+[[nodiscard]] AxisChunkClasses make_axis_chunk_classes(
+    const size_t length, const size_t chunk_extent, SignatureBuilder&& build_signature) {
```
vtsilytskyiTT: `make_axis_chunk_classes` takes `SignatureBuilder&&`, `make_axis_candidate_classes` takes it by value. Pick one style.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+    update_signature_anchor(exact_cone, anchor);
+    if (routed_cone != nullptr) {
+        update_signature_anchor(*routed_cone, anchor);
+    }
+    TT_FATAL(anchor != std::numeric_limits<size_t>::max(), "2D planner axis cone is empty");
+    anchor = tt::round_down(anchor, tile_extent);
+
+    AxisConeSignature signature;
+    signature.reserve(
+        15 + 22 * exact_cone.routes.size() + (routed_cone != nullptr ? 13 + 22 * routed_cone->routes.size() : 0));
```
vtsilytskyiTT: 13, 15, 22 are hand-counted. This approach is error-prone. Could you derive them from somewhere?

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+    uint64_t estimated_latency_cycles = 0;
+    for (uint32_t core = 0; core < active_core_count; ++core) {
+        const size_t count = base + (core < extra ? 1U : 0U);
+        const uint64_t core_cycles = 30'000 + prefix_cost(linear_index + count) - prefix_cost(linear_index);
+        estimated_latency_cycles = std::max(estimated_latency_cycles, core_cycles);
+        linear_index += count;
+    }
+    if (coordination_penalty_cycles_per_core > 0 && active_core_count > 64) {
+        estimated_latency_cycles +=
+            static_cast<uint64_t>(active_core_count - 64) * coordination_penalty_cycles_per_core;
```
vtsilytskyiTT: The inverse flag was dropped from the coordination-penalty condition. Behavior is preserved because the forward call site leaves the default at 0, but the coupling is now implicit.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+enum class AlignmentCostClass : uint8_t {
+    kExact,
+    kOneAxisShifted,
+    kGeneric,
+};
+
+[[nodiscard]] inline int64_t signed_tile_modulo(const int64_t value) noexcept {
+    const int64_t remainder = value % static_cast<int64_t>(kTileHeight);
+    return remainder < 0 ? remainder + static_cast<int64_t>(kTileHeight) : remainder;
+}
```
vtsilytskyiTT: The signature machinery now relies on `kTileHeight == kTileWidth` for the x-axis, because `signed_tile_modulo` uses `kTileHeight` on both axes (12 lines below). Is it OK? Or worth to use `kTileWidth` as well?

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+            static_cast<uint64_t>(active_core_count - 64) * coordination_penalty_cycles_per_core;
+    }
+    return Candidate{
+        .chunk_tiles_y = chunk_tiles_y,
+        .chunk_tiles_x = chunk_tiles_x,
+        .active_core_count = active_core_count,
+        .max_dependency_overhead = max_dependency_overhead,
+        .estimated_latency_cycles = estimated_latency_cycles,
+        .chunks = {},
+    };
```
vtsilytskyiTT, line 753: Add a unit test that compares the new planner's `chunk_tiles_y/x`, `active_core_count`, `estimated_latency_cycles` and full chunk vector against a brute-force reference across a matrix of shapes, wavelets, boundary modes, both route domains, and several L1 budgets. The deleted `estimate_candidate_latency_cycles` is the natural oracle — keep it in the test translation unit.
Let's make sure that your perf optimization is correct.
If test is too slow, add `NIGHTLY_` prefix to it's name and CI will skip it in regular runs.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+struct Candidate {
+    uint32_t chunk_tiles_y{0};
+    uint32_t chunk_tiles_x{0};
+    uint32_t active_core_count{0};
+    double max_dependency_overhead{0.0};
+    uint64_t estimated_latency_cycles{0};
+    std::vector<Lwt2DChunkPlan> chunks;
+};
+
+using AxisConeSignature = std::vector<int64_t>;
```
vtsilytskyiTT, line 525: `AxisConeSignature` is a bare `std::vector<int64_t>`. Adding a field to `AxisConePlan` or `AxisRouteRequirement` will not be reflected in the signature and will not fail to compile. Guard it — a `static_assert` on the struct sizes with a "update the signature" comment is the cheap fix.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1206 @@
...
+        y_plan, x_plan, best.chunk_tiles_y, best.chunk_tiles_x, fuse_terminal_scale, route_domain);
+    const size_t input_height = y_plan.preprocess_layout.input.length;
+    const size_t input_width = x_plan.preprocess_layout.input.length;
+    const size_t band_height = y_plan.output_length;
+    const size_t band_width = x_plan.output_length;
+    const Lwt2DTilingContract tiling{
+        .input = TiledShape2D::from_logical(Shape2D{.height = input_height, .width = input_width}),
+        .band = TiledShape2D::from_logical(Shape2D{.height = band_height, .width = band_width}),
+    };
+    validate_lwt_2d_tiling_contract(tiling);
```
vtsilytskyiTT, line 1043: Inverse plan 2D doesn't have same validation. I think it should.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,392 @@
...
+        EXPECT_NEAR(actual[index], expected[index], 1.0e-5F);
+    }
+}
+
+TEST(WaveletPlanner, CeilDivHandlesMaximumSizeWithoutOverflow) {
+    constexpr size_t maximum = std::numeric_limits<size_t>::max();
+    EXPECT_EQ(wavelet::ceil_div(maximum, size_t{1}), maximum);
+    EXPECT_EQ(wavelet::ceil_div(maximum, size_t{2}), maximum / 2 + 1);
+    EXPECT_EQ(wavelet::ceil_div(maximum - 1, size_t{2}), maximum / 2);
+    EXPECT_EQ(wavelet::ceil_div(size_t{17}, size_t{0}), 0);
```
vtsilytskyiTT: 1. Returning zero for a division by zero is a silent-corruption path that will produce zero-sized buffers and zero chunk counts, and now it's a documented contract. It should be TT_FATAL.
2. Separately, you kept a local `ceil_div` because `tt::div_up` in `tt_metal/api/tt-metalium/math.hpp` has the same `numerator + denominator - 1` overflow — that's a good catch, and the right resolution is to fix `div_up` rather than fork it

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,392 @@
...
+        }
+        low_columns[column] = cpu_forward<PlannerTestScheme>(values).approximation;
+    }
+    std::vector<float> approximation;
+    for (size_t row = 0; row < low_columns.front().size(); ++row) {
+        for (const auto& column : low_columns) {
+            approximation.push_back(column[row]);
+        }
+    }
+    expect_values_near(approximation, {6.9999995232F, 11.0F, 23.0F, 27.0F});
```
vtsilytskyiTT:
```suggestion
    expect_values_near(approximation, {7.0F, 11.0F, 23.0F, 27.0F});
```

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+TEST(WaveletPlanner, HaarSymmetricMatchesPyWaveletsAndRoundTripsEvenAndOddLengths) {
+    const std::array inputs = {
+        std::vector<float>{1.0F, 2.0F, 3.0F, 4.0F, 5.0F, 6.0F},
+        std::vector<float>{1.0F, 2.0F, 3.0F, 4.0F, 5.0F},
+    };
+    const std::array expected_approximation = {
+        std::vector<float>{2.1213202477F, 4.9497470856F, 7.7781744003F},
+        std::vector<float>{2.1213202477F, 4.9497470856F, 7.0710678101F},
+    };
```
vtsilytskyiTT, line 648: The analytic values are 3/√2 = 2.12132034356, 7/√2 = 4.94974746831, 11/√2 = 7.77817459305
With your tolerance those values should pass.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,392 @@
...
+
+TEST(WaveletPlanner, CeilDivHandlesMaximumSizeWithoutOverflow) {
+    constexpr size_t maximum = std::numeric_limits<size_t>::max();
+    EXPECT_EQ(wavelet::ceil_div(maximum, size_t{1}), maximum);
+    EXPECT_EQ(wavelet::ceil_div(maximum, size_t{2}), maximum / 2 + 1);
+    EXPECT_EQ(wavelet::ceil_div(maximum - 1, size_t{2}), maximum / 2);
+    EXPECT_EQ(wavelet::ceil_div(size_t{17}, size_t{0}), 0);
+}
+
+TEST(WaveletPlanner, BoundaryExtensionCoversAllSupportedModes) {
```
vtsilytskyiTT, line 591: `BoundaryExtensionCoversAllSupportedModes` tests index −1 and +4 on a length-4 signal. It never reaches the period-wrap logic — `make_antireflect_index_i32`'s period_quotient / reflected arithmetic. 
Also untested: length == 1 and length == 2 (both have explicit early-return branches), length == 0, and INT32_MIN, which carries a comment specifically claiming the unsigned negation is well-defined.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,392 @@
...
+        wavelet::Lwt2DRouteDomainPolicy::kExact);
+    ASSERT_FALSE(forward.chunks.empty());
+    for (const auto& chunk : forward.chunks) {
+        EXPECT_LE(chunk.final_band_rect.y.end, forward.tiling.band.logical.height);
+        EXPECT_LE(chunk.final_band_rect.x.end, forward.tiling.band.logical.width);
+        EXPECT_LE(chunk.resources.total_l1_bytes, l1_budget_bytes);
+    }
+    EXPECT_EQ(
+        wavelet::build_lwt_2d_chunk_config_words(forward).size(),
+        forward.chunks.size() * wavelet::device_protocol::kLwt2DChunkConfigWordCount);
```
vtsilytskyiTT, line 985: Serialized protocol values are never checked. Nothing checks a single word's value, so swapping `kLwt2DRectXBegin` with `kLwt2DRectYBegin` would pass. `build_lwt_2d_band_config_words` and `build_ilwt_2d_band_config_words` aren't called at all.

ttnn/cpp/ttnn/operations/wavelet/planner/l1_accounting.hpp
vtsilytskyiTT, line 1: This file is not used, is it?

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
vtsilytskyiTT, line 1: The lifting-cone arithmetic — the whole point of the PR — is untested. `PlannerTestScheme` is Predict, Swap, Predict, ScaleEven, ScaleOdd, with tap_size = 2 and a single coefficient on every step. That means:

1. k == 1 everywhere, so `coefficient_count()` returns 1 and every `translated(interval, offset, k - 1)` widens by zero. The halo widening in `backpropagate_requirements` and `propagate_requirements` — where every off-by-one in the dependency cone lives — is exercised only in its degenerate form.
2. There is no `kUpdate` step at all. The `kUpdate` branches in `backpropagate_requirements`, `propagate_requirements`, `append_forward_route`, ... are all dead under test.

vtsilytskyiTT (commented):
Thank you for addressing most of the comments, especially initial tests and performance improvement!
I'm asking you to increase test coverage (see detailed comments below) + some style fixes.

ttnn/cpp/ttnn/operations/wavelet/planner/l1_accounting.hpp
ke1rro: Yeah, not used in this PR anymore after the fixes. PR2 uses it though, so I'll remove it from here.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Fixed in 7ad6721. All values are now in `planner/cost_model.hpp`

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: 7ad6721. Split into separete helper.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Added a comment explaining that index 0 is intentionally unused. A chunk must contain at least one tile, and the vector is indexed directly by `chunk_tiles`, so valid entries occupy `[1, maximum_chunk_tiles]`. This avoids repeated `chunk_tiles - 1` conversions

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Fixed in 7ad6721

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Fixed in 7ad6721. The reserve expression is now derived

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: The default was removed, forward passes zero and inverse passes its penalty explicitly.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: 7ad6721. Alignment now uses `kTileHeight` for Y and `kTileWidth` for X.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Done in f992603

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: 7ad6721 and f992603

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: Now validates the same 2D tiling contract before returning the plan

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
ke1rro: https://github.com/tenstorrent/tt-metal/pull/54038/changes#diff-6a4e00d80fb9b82bbb6e5544737798cc344721a1543e42a6c9bbbaab7bf59d0bL17-R38.

Fix could be done like that.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
ke1rro: 1D Haar expectations are now expressed analytically in f992603

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
ke1rro: The boundary test now covers repeated positive/negative wraps, lengths 0/1/2, INT32_MIN, and the antireflect/antisymmetric period arithmetic.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
ke1rro: Added multi tap scheme which exercises both predict and update, nonzero shifts, halo widening, and explicit forward/inverse cone and route geometry.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
ke1rro: Maybe it is also worth adding focused tests here, since tt::div_up has 604 direct call sites and is used indirectly by 404 tt::round_up call sites. The coverage could include zero numerator, exact and non-exact division, mixed integer widths, UINT32_MAX/UINT64_MAX/SIZE_MAX, signed inputs, and TT_FATAL for a zero denominator.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
vtsilytskyiTT: Yes, second argument is a correct approach.

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
vtsilytskyiTT: Even better, thanks!

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
```diff
@@ -0,0 +1,1002 @@
...
+        wavelet::BoundaryMode::kSmooth,
+        wavelet::BoundaryMode::kAntisymmetric,
+        wavelet::BoundaryMode::kAntireflect,
+    };
+    for (const auto [height, width] : shapes) {
+        for (const uint32_t core_limit : core_limits) {
+            for (const uint64_t l1_budget : l1_budgets) {
+                for (const auto boundary_mode : boundary_modes) {
+                    for (const auto route_domain : route_domains) {
+                        for (const bool latency_oriented : {false, true}) {
```
vtsilytskyiTT, line 815: This is strong :)

ttnn/cpp/ttnn/operations/wavelet/planner/l1_accounting.hpp
vtsilytskyiTT: I'll expect it in the upcoming PR :)

tests/ttnn/unit_tests/gtests/test_wavelet_planner.cpp
vtsilytskyiTT: If you wish to create another small and focused PR. Or you can just create an issue and link it somewhere. Up to you.

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
vtsilytskyiTT: Resolved, thanks!

ttnn/cpp/ttnn/operations/wavelet/common/signal.hpp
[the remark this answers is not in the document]
vtsilytskyiTT: Resolved, I believe.

vtsilytskyiTT (approved):
Looks good to me! Thank you, Nikita.
P.S. next time please work directly on Tenstorrent repo. Then we'll be able to run CI on your change.

ttnn/cpp/ttnn/operations/wavelet/planner/cost_model.hpp
```diff
@@ -0,0 +1,31 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include <cstdint>
+
+namespace ttnn::operations::wavelet::planner_cost_model {
+
+// Dimensionless relative weights used only to rank planner candidates
```
bbradelTT, line 11: How are the following constants weights?

tt_metal/api/tt-metalium/math.hpp
```diff
@@ -17,15 +17,22 @@ namespace tt {
...
+ * @return The result of ceiling division.
  *
  * @note If b is zero, this results in undefined behavior.
  */
 template <typename A, typename B>
 constexpr auto div_up(A a, B b) noexcept -> std::common_type_t<A, B> {
     using T = std::common_type_t<A, B>;
-    assert(b != 0 && "Divide by zero error in div_up");
-    return static_cast<T>((static_cast<T>(a) + static_cast<T>(b) - 1) / static_cast<T>(b));
+    const T numerator = static_cast<T>(a);
```
bbradelTT, line 27: This code is used everywhere. Making a change to it should be in it's own PR.

Is the new code faster?

vtsilytskyiTT: I agree about separate PR for this change + tests.
Reason is described in this thread: https://github.com/tenstorrent/tt-metal/pull/54038#discussion_r3854914077

vtsilytskyiTT:
Local test results (BH, commit f992603e11beee42e4d09f59bd96513ef84c1817):

```
time ./build_Debug/test/ttnn/unit_tests_ttnn --gtest_filter='WaveletPlanner*'
Running main() from gmock_main.cc
Note: Google Test filter = WaveletPlanner*
[==========] Running 9 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 9 tests from WaveletPlanner
[ RUN      ] WaveletPlanner.BoundaryExtensionCoversAllSupportedModes
[       OK ] WaveletPlanner.BoundaryExtensionCoversAllSupportedModes (0 ms)
[ RUN      ] WaveletPlanner.HaarSymmetricMatchesPyWaveletsAndRoundTripsEvenAndOddLengths
[       OK ] WaveletPlanner.HaarSymmetricMatchesPyWaveletsAndRoundTripsEvenAndOddLengths (0 ms)
[ RUN      ] WaveletPlanner.HaarSymmetricTwoDimensionalApproximationMatchesPyWavelets
[       OK ] WaveletPlanner.HaarSymmetricTwoDimensionalApproximationMatchesPyWavelets (0 ms)
[ RUN      ] WaveletPlanner.MultiTapPredictAndUpdateWidenDependencyCones
[       OK ] WaveletPlanner.MultiTapPredictAndUpdateWidenDependencyCones (0 ms)
[ RUN      ] WaveletPlanner.ForwardPlanEnforcesStickAndSignedIndexContracts
2026-09-02 11:31:08.951 | critical |          Always | TT_FATAL: Forward lifting plan requires 32-element sticks, got 16 (assert.hpp:104)
2026-09-02 11:31:08.963 | critical |          Always | TT_FATAL: Input length 2147483646 with 2 elements of padding exceeds the signed device-index limit 2147483647 (assert.hpp:104)
[       OK ] WaveletPlanner.ForwardPlanEnforcesStickAndSignedIndexContracts (21 ms)
[ RUN      ] WaveletPlanner.TwoDimensionalPlansRejectUnsupportedSignedGeometryEarly
2026-09-02 11:31:08.963 | critical |          Always | TT_FATAL: 2D LWT input dimensions 1073741824x1 exceed the signed device-coordinate limit 1073741823 (assert.hpp:104)
2026-09-02 11:31:08.964 | critical |          Always | TT_FATAL: 2D ILWT output dimensions 1x1073741824 exceed the signed device-coordinate limit 1073741823 (assert.hpp:104)
[       OK ] WaveletPlanner.TwoDimensionalPlansRejectUnsupportedSignedGeometryEarly (1 ms)
[ RUN      ] WaveletPlanner.OptimizedTwoDimensionalPlannerMatchesExhaustiveSearch
[       OK ] WaveletPlanner.OptimizedTwoDimensionalPlannerMatchesExhaustiveSearch (264 ms)
[ RUN      ] WaveletPlanner.TwoDimensionalProtocolSerializationMatchesGoldenWords
[       OK ] WaveletPlanner.TwoDimensionalProtocolSerializationMatchesGoldenWords (0 ms)
[ RUN      ] WaveletPlanner.OddRectangularPlansProduceBoundedSerializableChunks
[       OK ] WaveletPlanner.OddRectangularPlansProduceBoundedSerializableChunks (0 ms)
[----------] 9 tests from WaveletPlanner (287 ms total)

[----------] Global test environment tear-down
[==========] 9 tests from 1 test suite ran. (287 ms total)
[  PASSED  ] 9 tests.

real    0m0.855s
user    0m0.307s
sys     0m0.082s
```

tt_metal/api/tt-metalium/math.hpp
ke1rro: I will create issue and attach PR with fix

vtsilytskyiTT: Amazing, let's prioritize it.

ke1rro:
> Looks good to me! Thank you, Nikita. P.S. next time please work directly on Tenstorrent repo. Then we'll be able to run CI on your change.

Thanks! 

I currently only have read access to the Tenstorrent repository, so I cannot push a feature branch there directly.

tt_metal/api/tt-metalium/math.hpp
ke1rro: Created the issue and linked the PR with the fix:

Issue: https://github.com/tenstorrent/tt-metal/issues/55268
PR: https://github.com/tenstorrent/tt-metal/pull/55271

ttnn/cpp/ttnn/operations/wavelet/planner/l1_accounting.hpp
ke1rro: The second PR is ready here 🙂 https://github.com/tenstorrent/tt-metal/pull/54039

vtsilytskyiTT:
Great, `div_up` prerequisite PR is merged, let's continue with this one.

tt_metal/api/tt-metalium/math.hpp
```diff
@@ -17,15 +17,22 @@ namespace tt {
  *
  * @param a The numerator.
  * @param b The denominator. Must be non-zero.
- * @return The result of ceiling division (a + b - 1) / b.
+ * @return The result of ceiling division.
  *
  * @note If b is zero, this results in undefined behavior.
  */
 template <typename A, typename B>
 constexpr auto div_up(A a, B b) noexcept -> std::common_type_t<A, B> {
```
riverwuTT, line 25: Is this diff still part of current PR?

vtsilytskyiTT: No. Fix originated here. Now we need a rebase.
@ke1rro

ke1rro: > No. Fix originated here. Now we need a rebase. @ke1rro
@vtsilytskyiTT, rebased

PR #54039, opened 2026-08-22
[Feature] Add native TTNN Wavelet operations (split 2/2)

ke1rro:
### PR Category

Feature

Split 2 of 2 from #52669.

### Summary

Relates to #40494. Split from #52669.

Adds the native TTNN Wavelet operations, including the public API, nanobind bindings, Device Operations, Program Factories, device kernels, and Python tests.

reference PR: https://github.com/tenstorrent/tt-metal/pull/54038
generator prerequisite: https://github.com/ke1rro/tt-metal/pull/2


ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,415 @@
...
+}
+
+}  // namespace ttnn::operations::wavelet::kernels
+
+#undef WAVELET_1D_STEP_ATTRIBUTES
+
+void kernel_main() {
+    constexpr uint32_t cb_base = get_compile_time_arg_val(2);
+    constexpr uint32_t cb_output = get_compile_time_arg_val(3);
+    init_sfpu(cb_base, cb_output);
```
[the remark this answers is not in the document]
ke1rro: b7af27a
Added compute hardware startup

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,396 @@
...
+}
+
+}  // namespace
+
+void kernel_main() {
+    constexpr uint32_t cb_source0 = get_compile_time_arg_val(0);
+    constexpr uint32_t cb_source1 = get_compile_time_arg_val(1);
+    constexpr uint32_t cb_base = get_compile_time_arg_val(2);
+    constexpr uint32_t cb_output = get_compile_time_arg_val(3);
+    init_sfpu(cb_base, cb_output);
```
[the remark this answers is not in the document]
ke1rro: b7af27a
Added compute hardware startup

tt_metal/python_env/requirements-dev.txt
```diff
@@ -48,6 +48,7 @@ pytest-timeout==2.4.0
 pytest-split==0.11.0
 pytest-benchmark==5.2.3
 pytest-repeat==0.9.4
+PyWavelets==1.8.0
```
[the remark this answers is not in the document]
ke1rro: Would you suggest a preferred way to handle this?

ttnn/cpp/ttnn/operations/wavelet/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
+include(sources.cmake)
+
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
+
+set(WAVELET_SCHEME_CATALOG ${CMAKE_CURRENT_SOURCE_DIR}/schemes/wavelet_schemes.json)
+set(WAVELET_SCHEME_GENERATOR ${CMAKE_CURRENT_SOURCE_DIR}/schemes/generate_wavelet_scheme_headers.py)
+set(WAVELET_GENERATED_SCHEME_DIR ${CMAKE_CURRENT_SOURCE_DIR}/generated/wavelet_schemes)
```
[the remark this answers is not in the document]
ke1rro: Moved generated Wavelet headers to the CMake binary tree
d6308ab

tests/ttnn/unit_tests/operations/wavelet/test_wavelet_all_schemes.py
```diff
@@ -0,0 +1,99 @@
...
+        coefficient_length = pywt.dwt_coeff_len(signal.numel(), pywt.Wavelet(scheme).dec_len, mode=boundary_mode)
+        assert ttnn.dwt_coeff_len(signal.numel(), scheme) == coefficient_length
+        coefficient_sticks = (coefficient_length + 31) // 32
+        signal_sticks = (signal.numel() + 31) // 32
+        assert tuple(approximation.shape) == (coefficient_sticks, 32)
+        assert tuple(detail.shape) == (coefficient_sticks, 32)
+        assert tuple(reconstructed.shape) == (signal_sticks, 32)
+        assert torch.isfinite(ttnn.to_torch(approximation)).all(), scheme
+        assert torch.isfinite(ttnn.to_torch(detail)).all(), scheme
+        assert torch.isfinite(ttnn.to_torch(reconstructed)).all(), scheme
```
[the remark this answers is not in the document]
ke1rro: I made a few changes to the factorization pipeline used to generate the wavelet schemes, and they resulted in a significant improvement in numerical precision. I’ll add dedicated precision tests soon.

You can compare the results here:

Baseline:
https://github.com/ke1rro/lifting-factorization/blob/main/README.md?plain=1#L126-L163

Improved:
https://github.com/ke1rro/lifting-factorization/blob/improve-pywavelets-factorization-precision/README.md?plain=1#L126-L163

ke1rro:
I added the FP32 precision tests in [59b3242](https://github.com/tenstorrent/tt-metal/pull/54039/commits/59b32427e78047be516b7ff99623f52d007a4d17). This commit addresses the following https://github.com/tenstorrent/tt-metal/pull/54039#discussion_r3893417993

Here are the methodology and results.

For each input, the test computes three versions of every output:

* $p_{64}$: PyWavelets with FP64 input
* $p_{32}$: PyWavelets with FP32 input
* $t_{32}$: TTNN with FP32 input

For 1D, the checked outputs are `cA`, `cD`, and the reconstructed signal. For 2D, they are `LL`, `LH`, `HL`, `HH`, and the reconstructed signal.

PyWavelets FP64 is used as the reference. The maximum FP32 error is measured for PyWavelets and TTNN:

$$
E_{\mathrm{PYWT32}} = \max_i |p_{64,i} - p_{32,i}|
$$

$$
E_{\mathrm{TT32}} = \max_i |p_{64,i} - t_{32,i}|
$$

The errors are normalized by

$$
S = \max\left(1, \max_i |p_{64,i}|\right).
$$

Two values are checked:

$$
\Delta =
\frac{|E_{\mathrm{TT32}} - E_{\mathrm{PYWT32}}|}{S}
$$

which measures how much the TTNN FP32 error differs from the PyWavelets FP32 error, and

$$
D =
\frac{\max_i |t_{32,i} - p_{32,i}|}{S}
$$

which directly compares the TTNN FP32 and PyWavelets FP32 outputs.

The allowed tolerance is

$$
T = T_{\mathrm{device}} + T_{\mathrm{factorization}}.
$$

The device tolerance depends on the boundary mode, but remains shared by all
wavelets and outputs within that mode and dimension.

The factorization allowance is:

* GREEN: $8\varepsilon_{32}$
* YELLOW: $32\varepsilon_{32}$

A result passes when both $\Delta$ and $D$ are within $T$.

The device tolerance accounts for differences in operation ordering and FP32
execution between Tensix and PyWavelets. These differences become larger for
longer lifting sequences and extrapolating boundary modes.

The complete 106-scheme run produced the following maxima. The wavelet and
output producing each maximum are shown in parentheses.

| Boundary mode | Max $\Delta$ 1D | Max $D$ 1D | Max $\Delta$ 2D | Max $D$ 2D |
|---|---:|---:|---:|---:|
| zero | 2.742e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 6.936e-06 (`db30`/`LH`) | 7.540e-06 (`db30`/`LH`) |
| constant | 2.751e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 8.296e-06 (`db30`/`HL`) | 8.815e-06 (`db30`/`LH`) |
| symmetric | 2.751e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 8.921e-06 (`db30`/`HL`) | 9.011e-06 (`db30`/`HL`) |
| reflect | 2.745e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 1.040e-05 (`db30`/`HL`) | 1.057e-05 (`db30`/`HL`) |
| periodic | 2.751e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 8.395e-06 (`db30`/`LH`) | 9.775e-06 (`db30`/`LH`) |
| smooth | 1.224e-05 (`coif15`/`cD`) | 1.313e-05 (`coif15`/`cD`) | 1.042e-04 (`db38`/`LH`) | 1.072e-04 (`db38`/`LH`) |
| antisymmetric | 2.751e-06 (`coif15`/`cD`) | 2.889e-06 (`coif15`/`cD`) | 5.668e-06 (`coif15`/`HH`) | 6.107e-06 (`coif15`/`HH`) |
| antireflect | 5.978e-06 (`db30`/`cD`) | 6.100e-06 (`db30`/`cD`) | 2.210e-05 (`db30`/`LH`) | 2.273e-05 (`db30`/`LH`) |

### Representative schemes

The table below compares schemes with different lifting lengths. A lifting step
means one `predict` or `update`; delay, swap, and scale records are not counted.
Each error is the maximum across all eight boundary modes and outputs.

| Scheme | Class | Lifting steps | Max $\Delta$ 1D | Max $D$ 1D | Max $\Delta$ 2D | Max $D$ 2D |
|---|---|---:|---:|---:|---:|---:|
| `db1` | GREEN | 2 | 9.050e-09 (`zero/cA`) | 1.788e-07 (`zero/reconstructed`) | 1.192e-07 (`constant/LH`) | 6.109e-07 (`constant/LH`) |
| `db2` | GREEN | 3 | 7.667e-08 (`zero/cA`) | 1.788e-07 (`zero/reconstructed`) | 1.091e-07 (`reflect/LL`) | 5.327e-07 (`antireflect/LH`) |
| `db4` | GREEN | 5 | 5.960e-08 (`zero/reconstructed`) | 2.980e-07 (`zero/reconstructed`) | 6.312e-07 (`smooth/HL`) | 1.176e-06 (`smooth/HL`) |
| `bior1.1` | GREEN | 2 | 9.050e-09 (`zero/cA`) | 1.788e-07 (`zero/reconstructed`) | 1.192e-07 (`constant/LH`) | 6.109e-07 (`constant/LH`) |
| `bior2.2` | GREEN | 2 | 1.192e-07 (`zero/reconstructed`) | 1.788e-07 (`zero/reconstructed`) | 3.085e-07 (`smooth/LL`) | 4.150e-07 (`smooth/LL`) |
| `bior3.9` | GREEN | 3 | 1.788e-07 (`zero/reconstructed`) | 3.356e-07 (`zero/cA`) | 9.778e-07 (`smooth/LL`) | 1.210e-06 (`smooth/LL`) |
| `db30` | GREEN | 31 | 7.416e-06 (`smooth/cD`) | 7.831e-06 (`smooth/cD`) | 3.488e-05 (`smooth/HL`) | 4.453e-05 (`smooth/HL`) |
| `db38` | GREEN | 39 | 1.217e-06 (`smooth/cD`) | 1.463e-06 (`smooth/cD`) | 1.042e-04 (`smooth/LH`) | 1.072e-04 (`smooth/LH`) |
| `coif15` | GREEN | 46 | 1.224e-05 (`smooth/cD`) | 1.313e-05 (`smooth/cD`) | 9.935e-05 (`smooth/LH`) | 1.032e-04 (`smooth/LH`) |
| `bior5.5` | YELLOW | 5 | 1.013e-06 (`zero/reconstructed`) | 1.192e-06 (`zero/reconstructed`) | 3.464e-06 (`reflect/LH`) | 4.100e-06 (`reflect/LH`) |
| `sym19` | YELLOW | 20 | 1.389e-06 (`smooth/cD`) | 1.490e-06 (`smooth/reconstructed`) | 4.705e-06 (`smooth/HH`) | 5.477e-06 (`smooth/LH`) |

The 2-5-step `db` and `bior` schemes show substantially smaller accumulated
FP32 error. The measurements show larger runtime errors for long lifting
cascades and extrapolating boundary modes than for the shorter schemes.

This behavior is separate from the GREEN/YELLOW classification, which
describes lifting factorization accuracy. The factorization class therefore
remains a separate additive allowance.

The selected tolerances are rounded values above the measured maximum for each
boundary mode. Headroom is

$$
\frac{T_{\mathrm{device},m}}
{\max(\Delta_m^{\max}, D_m^{\max})}.
$$

| Boundary mode | 1D tolerance | 1D headroom | 2D tolerance | 2D headroom |
|---|---:|---:|---:|---:|
| zero | 1e-5 | 3.461x | 2e-5 | 2.653x |
| constant | 1e-5 | 3.461x | 2e-5 | 2.269x |
| symmetric | 1e-5 | 3.461x | 2e-5 | 2.220x |
| reflect | 1e-5 | 3.461x | 2e-5 | 1.892x |
| periodic | 1e-5 | 3.461x | 2e-5 | 2.046x |
| smooth | 2e-5 | 1.523x | 2e-4 | 1.866x |
| antisymmetric | 1e-5 | 3.461x | 1e-5 | 1.638x |
| antireflect | 1e-5 | 1.639x | 5e-5 | 2.200x |

The overall worst 1D result was `coif15`/`cD` with `smooth` boundary
extension: $\Delta=1.224\mathrm{e}{-5}$ and $D=1.313\mathrm{e}{-5}$. The
overall worst 2D result was `db38`/`LH` with `smooth`:
$\Delta=1.042\mathrm{e}{-4}$ and $D=1.072\mathrm{e}{-4}$.

`dmey` is still covered by DWT/IDWT, shape, and finite-value checks, but is excluded from the FP32 precision check because its current lifting factorization error is approximately `6.64e-1`.

ke1rro:
@vtsilytskyiTT Hi! Looking forward to your review 🙂

ke1rro:
Hi @jberkowitzTT,

The review of this second split seems to have stalled a bit. Would you mind tagging someone who could continue the review?

Split 1 has already gone through review, while this PR contains the remaining implementation.

For clarity, the current review coverage is roughly:

### Already covered in Split 1

has already been reviewed:

- `wavelet/common/{boundary,signal,signal_extension,storage_contract,tiling_2d}.hpp`
- `wavelet/planner/{execution_plan,inverse_plan,pad_split_layout,plan,static_scheme,step}.hpp`
- `wavelet/device/protocol/lwt_config.hpp`
- `tt_metal/api/tt-metalium/math.hpp`

### Still needs review in Split 2

**1D implementation**
- `lwt_1d_*`
- `ilwt_1d_*`
- `wavelet_1d_operation_impl.*`
- `kernels/compute/lwt_compute.cpp`
- `kernels/dataflow/lwt_{reader,writer}.cpp`

**2D implementation**
- `lwt_2d_*`
- `ilwt_2d_*`
- `wavelet_2d_operation_impl.*`
- `kernels/compute/lwt_2d_compute.cpp`
- `kernels/dataflow/lwt_2d_{reader,writer}.cpp`

**Shared device / SFPI infrastructure**
- `device/kernels/primitives/*`
- `device/kernels/sfpi/*`
- `wavelet_program_utils.*`
- `wavelet_tensor_validation.*`
- `wavelet_l1_budget.hpp`

From my side, Split 2 is fully ready for review.
Thanks!

jberkowitzTT:
Tagging in @philei-tt for the review

philei-tt (commented):
Kernel/op-focused review of the LWT/ILWT implementation (both halves, since #54038 is stacked under this PR).

**Verified OK by reading:** CB push/pop balance in 1D and 2D; boundary-extension index math vs PyWavelets for all eight modes (symmetric half-sample, reflect whole-sample, smooth, antisymmetric, antireflect, periodic); `dwt_coeff_len` == `pywt.dwt_coeff_len` for non-periodization modes; program-cache address patching for all input/output/metadata buffers through the WorkloadDescriptor bindings; host<->kernel RTA and compile-arg index tables; inline terminal-scale parity rule identical on host and device; band order consistent everywhere; odd H/W; TT_FATAL coverage. The planner gtest is substantive.

**Blocking:** the Blackhole zero-flag workaround missing from the new 32x16 datacopy helper, and the runtime-arg count in the 1D compute kernel growing with batch. The LLK API files also need a rebase onto the SAN_HOOK version of main (current conflict).

**Should fix / questions:** SFPI rotate ISA assumptions need documentation and a golden test; `available_l1_bytes` in the cache key (churn + trace capture); silent tile drop in the 2D boundary collector; `fuse_terminal_scale` default; double allocation of config buffers; host-side instantiation of the planner for all 106 schemes.

Nice work overall, the generated-scheme design and the host/device contract are coherent. Comments inline.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_math_unary_datacopy_api.h
```diff
@@ -12,6 +12,12 @@
  * LLK ELTWISE UNARY DATACOPY
  *************************************************************************/
 
+__attribute__((always_inline)) inline void llk_math_eltwise_unary_datacopy_to_dest_32x16(std::uint32_t dst_index) {
```
philei-tt, line 15: **Blocking (Blackhole): missing DEST zero-flag clear after unpack-to-dest.**

This helper reproduces only the first three lines of `_llk_math_eltwise_unary_datacopy_` (tt_llk_blackhole/llk_lib/llk_math_eltwise_unary_datacopy.h). The real LLK follows `math_unpack_to_dest_tile_ready()` with

```cpp
for (i = 0; i < num_faces; i++)
    TT_ZEROACC(p_zeroacc::CLR_16, clear_fp32, 1 /*clear zero flags*/, ADDR_MOD_3, get_dest_index_in_faces(local_tile, i));
```

because of a Blackhole Tensix bug (budabackend/#2730): when an unpack-to-dest write (which should clear the zero flags of the datums it writes) lands in the same cycle as a packer `ZEROACC`, the flag clear is dropped. The datums then stay flagged as zero and the packer emits 0 for non-zero FP32 data. Since this op relies on unpack-to-dest for every source/base tile, that is a silent-corruption risk on BH, timing dependent so it may pass small tests.

For a 32x16 narrow tile at index `n` the two faces are `(n & 7) << 1` and `((n & 7) << 1) + 1` (fp32: 4 full tiles = 8 narrow tiles per bank). Rather than hand-rolling it, I would generalise the existing API with a `DstTileShape` template parameter (default `Tile32x32`) so the workaround, the `is_32bit_input` gate and the `LLK_ASSERT(dst_index < get_dest_max_tiles_rt<..., Tile32x16>())` come for free on both archs.

Also: both `llk_math_unary_datacopy_api.h` copies are based on a pre-#55054/#55055 main (no `sanitizer/api.h`, no `SAN_HOOK`), which is the current merge conflict. After the rebase the new function needs at least `SAN_HOOK(unsupported())` like `llk_math_fast_tilize`.

Has the 1D op been run on Blackhole silicon?

tt_metal/hw/inc/internal/compute/tile_move_copy.h
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "api/compute/common_globals.h"
+#include "api/compute/tile_move_copy.h"
+
+namespace ckernel::internal {
+
+ALWI void copy_tile_to_dst_32x16(uint32_t in_cb_id, uint32_t in_tile_index, uint32_t dst_tile_index) {
```
philei-tt, line 12: `tt_metal/hw/inc/internal/compute/` is a new include root; other ops keep op-specific LLK glue next to their kernels (`device/kernels/compute/*.hpp`) or go through `api/compute/experimental`. Either move this into `ttnn/cpp/ttnn/operations/wavelet/device/kernels/` or get LLK-owner sign-off on the new tree.

The 32x16 dest copy itself is sound (`set_dst_write_addr<Tile32x16>` exists on both archs and `_llk_unpack_A_` handles `num_faces == 2` in the unpack-to-dest MOP). No pack-side counterpart is needed because the packer addresses 32x32 slots, which is why `lwt_compute.cpp` puts outputs on even narrow indices and packs W = 0, 1, 2. I would pin that coupling with `static_assert(kDstBaseN % 2 == 0 && kPackBaseN == kDstBaseN / 2)` in the compute kernel.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1461 @@
...
+    args.push_back(checked_u32(plan.chunks.front().routes.size(), "LWT route count"));
+    args.push_back(buffers.final_even);
+    args.push_back(buffers.final_odd);
+    args.push_back(checked_u32(plan.workspace_elements * sizeof(float), "LWT tile mirror offset"));
+    args.push_back(chunks_per_sample);
+    args.push_back(output_pages_per_sample);
+    return args;
+}
+
+[[nodiscard]] std::vector<uint32_t> compute_runtime_args(const LwtExecutionPlan& plan, const CoreChunkWork& work) {
```
philei-tt, line 467: **Runtime-arg count scales with batch and will hit the 4096-word ceiling.**

Compute RTAs here are `1 + chunk_count_per_core * route_count`, and `chunk_count_per_core = ceil(B * chunks_per_sample / cores)`. tt-metal enforces a 4096-word limit per Tensix kernel (`tt_metal/impl/kernels/kernel.cpp`, `SetRuntimeArgs`) plus the kernel-config L1 fit at finalize. A ~20-route scheme with a batch of ~13k short signals on 64 cores, or one very long signal after chunk doubling, will TT_FATAL.

`output_group_count(route.output_length)` depends only on `global_chunk % chunks_per_sample`, not on the batch index, so the whole table is `chunks_per_sample * route_count` words. Pass that once as **common** runtime args and give each core just `(chunk_begin, chunk_count, chunks_per_sample)`; compute indexes `common[(item % chunks_per_sample) * route_count + r]`. Same applies to the inverse path at L750.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1461 @@
...
+    constexpr size_t expected_route_count = executable_step_count<Scheme>() - 1U;
+    for (const auto& chunk : plan.chunks) {
+        TT_FATAL(
+            chunk.routes.size() == expected_route_count,
+            "DWT planner produced {} routes, but the kernel ABI requires {}",
+            chunk.routes.size(),
+            expected_route_count);
+    }
+    const size_t route_count = plan.chunks.front().routes.size();
+    auto route_config = wavelet_program_utils::create_replicated_dram_pages(
```
philei-tt, line 1083: `route_config` / `chunk_config` are allocated here via `create_replicated_dram_pages`, stored in `buffers`, and then immediately replaced by `upload_replicated_dram_metadata`, which allocates again. The first allocations are pure waste and stay alive until the function returns. `build_route_config_words` only reads slot ids from `buffers`, so default-construct the two shared_ptrs and drop the pre-allocation. Same pattern in the inverse path at L1223.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1461 @@
...
+    std::vector<uint32_t> reader_compile_args = {
+        kReaderConfigCb,
+        kSrcTile0Cb,
+        kSrcTile1Cb,
+        kBaseTileCb,
+        kSrcCacheCb,
+        kSyncCb,
+        static_cast<uint32_t>(workspace_layout == WorkspaceLayout::kTileNative),
+        1U,
+        static_cast<uint32_t>(BoundaryMode::kSymmetric),
```
philei-tt, line 805: The inverse reader gets `BoundaryMode::kSymmetric` hard-coded while `Ilwt1DParams.boundary_mode` is threaded into `make_inverse_lifting_plan`. Both are inert (the inverse reader never extends, and the plan only stores the mode in `pad_config.mode`). That is correct, PyWavelets `idwt` is mode-independent except for `periodization`, but please say so in a comment, and either drop `boundary_mode` from `idwt` / `idwt_2d` or document that it is accepted for symmetry and ignored. Right now a reader of the public API will assume it matters.

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
```diff
@@ -0,0 +1,1461 @@
...
+}  // namespace
+
+namespace detail {
+
+tt::tt_metal::WorkloadDescriptor create_lwt_1d_workload(
+    const Lwt1DParams& operation_attributes,
+    const Lwt1DInputs& tensor_args,
+    Lwt1DOutputs& tensor_return_value,
+    const MeshCoordinateRangeSet& tensor_coords) {
+    return dispatch_scheme(operation_attributes.scheme_id, [&]<typename Scheme>() {
```
philei-tt, line 1402: Design question on the host side: `dispatch_scheme` instantiates `build_forward_workload<Scheme>` and `build_inverse_workload<Scheme>` for all 106 schemes, in both impl TUs (which is presumably why they are `SKIP_UNITY_BUILD_INCLUSION`). Looking at what the planner actually reads from `Scheme`, it is `tap_size`, `delay_even/odd`, `num_steps`, `name`, and the per-step `(type, shift, k)` in `append_forward_route`. All of that is already in (or trivially added to) the runtime `SchemeInfo` table; the `static_assert`s on `k <= kStepCoeffCapacity` could become generator-time checks.

Only the kernel needs the type, and it gets it through the `-D LWT_SCHEME_HEADER/TYPE` strings. A runtime-table planner would reduce `dispatch_scheme` to picking those two strings and remove ~424 heavy instantiations. What is the compile time of these two TUs today?

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_l1_budget.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+#include "tt-metalium/allocator.hpp"
+#include "tt-metalium/math.hpp"
+#include "tt-metalium/mesh_device.hpp"
+
+namespace ttnn::prim::detail {
+
+inline constexpr uint32_t kWaveletL1BudgetQuantumBytes = 16U * 1024U;
+
+[[nodiscard]] inline uint32_t quantized_available_l1_bytes(tt::tt_metal::distributed::MeshDevice* mesh_device) {
```
philei-tt, line 20: `available_l1_bytes` is derived from `lowest_occupied_compute_l1_address()` at call time, quantised to 16 KiB, and hashed into the program-cache key (it is a member of `Lwt*Params`). That is sound (a hit implies the same budget), but two consequences worth stating in a comment:

1. Every time L1 tensor residency crosses a 16 KiB boundary you get a new cache entry and a full replan. The 2D planner search is not cheap (~900 `(ty, tx)` candidates x ~9 representative chunks in `plan_2d.hpp`).
2. If residency differs between trace warm-up and trace capture, the capture call misses the cache and `upload_replicated_dram_metadata` performs `EnqueueWriteMeshBuffer` mid-capture.

Consider hashing the CB footprint the plan actually needs rather than free L1, or at least a coarser quantum, and capping the 2D candidate search.

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
```diff
@@ -0,0 +1,590 @@
...
+    const uint32_t core_limit,
+    const uint32_t l1_signal_budget_bytes,
+    const WorkspaceLayout workspace_layout = WorkspaceLayout::kRowMajor) {
+    TT_FATAL(core_limit > 0, "LWT requires at least one worker core");
+    TT_FATAL(l1_signal_budget_bytes >= 3 * device_protocol::kStickBytes, "LWT L1 budget is too small");
+
+    const size_t max_final_length = full_plan.output_length;
+    const uint32_t final_group_count = static_cast<uint32_t>(std::max(
+        tt::div_up(max_final_length, static_cast<size_t>(device_protocol::kLwtGroupOutputElements)), size_t{1}));
+    uint32_t chunk_count = std::min(final_group_count, core_limit);
```
philei-tt, line 529: Perf: the chunk count is chosen per sample as `min(final_group_count, core_limit)` before batch is considered. With `B >= cores` every sample is still split up to 64 ways, and each chunk pays halo recompute (`dependency_overhead`) plus `2 * route_count` 64-byte config-page round trips (reader and writer each `async_read` + barrier per route). Prefer `core_limit = ceil(cores / B)` when `B > 1`, and pre-load a core's config pages once per chunk instead of per route.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/horizontal_stencil_sfpi.h
```diff
@@ -0,0 +1,289 @@
...
+    addr_mod.set(ADDR_MOD_3);
+    math::reset_counters(p_setrwc::SET_ABD_F);
+
+    // Enable all lanes
+    TTI_SFPENCC(sfpi::SFPENCC_IMM12_BOTH, 0, 0, sfpi::SFPENCC_MOD1_EI_RI);
+}
+
+// _horizontal_stencil_rotate_(a, b): 1-element right shift within subvectors.
+// After rotation, column 0 of b gets the element shifted out of a.
+inline void _horizontal_stencil_rotate_(std::uint32_t a_reg, std::uint32_t b_reg) {
```
philei-tt, line 39: Both branches of `_horizontal_stencil_rotate_` lean on behaviour that is not documented in this repo:

- **Wormhole**: the SHFLSHR1 lane-0 halo erratum ("uses the previous instruction's source as lane-zero halo").
- **Blackhole**: `TTI_SFPAND(p_sfpu::LTILEID, LREG7, LREG7, 1)` uses `mod1 = 1` (the first operand is the imm12 slot in `TT_OP_SFPAND(imm12, lreg_c, lreg_dest, mod1)`), and no other BH LLK in the tree uses `mod1 != 0` on SFPAND. It also assumes LREG15 holds `2 * lane`.

Could you add tt-isa-documentation references for both, and a tiny golden-vector device test of just the rotate (e.g. a = [0..7], b = [8..15] per subvector)? If either assumption changes with a silicon or SFPI revision, the failure mode today is a small PCC drift in the stencil, not a loud test failure.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/horizontal_stencil_sfpi.h
```diff
@@ -0,0 +1,289 @@
...
+    if constexpr (ScaleSource) {
+        TT_SFPLOADI(tmp, sfpi::SFPLOADI_MOD0_UPPER, SourceScalePacked >> 16);
+        TT_SFPLOADI(tmp, sfpi::SFPLOADI_MOD0_LOWER, SourceScalePacked & 0xFFFF);
+        _lwt_scale_register_(f_e_0, tmp, tmp_acc);
+        _lwt_scale_register_(f_o_0, tmp, tmp_acc);
+        _lwt_scale_register_(f_e_1, tmp, tmp_acc);
+        _lwt_scale_register_(f_o_1, tmp, tmp_acc);
+    }
+
+    TTI_SFPMOV(0, p_sfpu::LCONST_0, g_e, 0);
```
philei-tt, line 116: Nit: these two `SFPMOV LCONST_0 -> g_e/g_o` are dead, both registers are overwritten by the `SFPLOAD` on the next two lines. Also, this file uses `TTI_NOP` as the SFPU result-hazard bubble while `vertical_stencil_sfpi.h` uses in-place `SFPMAD(f, tmp, g, g)` + `TTI_SFPNOP` (the tt_llk idiom). Picking the in-place form here would also remove the extra `SFPMOV` per tap.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
```diff
@@ -0,0 +1,1416 @@
...
+    uint32_t capacity;
+
+    LWT_2D_BOUNDARY_FUNCTION void operator()(const uint32_t source_index) const {
+        const uint32_t source_tile = source_index / kTileSide;
+        for (uint32_t index = 0; index < count; ++index) {
+            if (tiles[index] == source_tile) {
+                return;
+            }
+        }
+        if (count < capacity) {
```
philei-tt, line 84: Silent drop: when `count == capacity` the tile id is discarded, and `find_index` below only `ASSERT`s on a miss and returns 0, so in a release build a wrong tile is silently read. I bounded the modes for the 2-tile window (periodic <= 5, symmetric/reflect <= 3, antireflect <= 4 incl. both endpoints) and the capacities hold today, but nothing on the host proves it and the bound depends on `kStepCoeffCapacity`. Please assert the bound in the planner (or document the derivation next to the capacities) and make the kernel fail loudly instead of dropping.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
```diff
@@ -0,0 +1,1416 @@
...
+    CircularBuffer sync_buffer(cb_sync);
+    Noc noc;
+    // This CB is an L1 allocation only
+    const uint32_t zero_tile_addr = CircularBuffer(cb_route_zero).get_write_ptr();
+    auto* zero_tile = reinterpret_cast<volatile tt_l1_ptr uint32_t*>(zero_tile_addr);
+    for (uint32_t word = 0; word < kTileElements; ++word) {
+        zero_tile[word] = 0;
+    }
+    const uint32_t noc_scratch_addr = CircularBuffer(cb_noc_scratch).get_write_ptr();
+    const uint32_t reader_config_addr = noc_scratch_addr;
```
philei-tt, line 1226: Undocumented protocol: the reader stages split tiles into the *whole* NoC scratch CB (up to 25 tiles), while the writer preloads its route pages into the second half (`writer_config_addr = noc_scratch_addr + split_scratch_bytes / 2`, lwt_2d_writer.cpp L471). Correctness relies on the writer's `output_buffer.wait_front(1)` at chunk start implying the reader finished `initialize_planes`, and on the reader popping the chunk-end sync before re-staging. It works, but please comment it at both sites, or give the writer its own small config CB (`route_count * 128 B <= config_capacity` is already TT_FATAL'd) so the regions are disjoint.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_writer.cpp
```diff
@@ -0,0 +1,575 @@
...
+                ttnn::operations::wavelet::kernels::primitives::local_noc_destination(noc, destination_addr));
+        }
+        noc.async_write_barrier();
+        output_buffer.pop_front(batch);
+        first_tile += batch;
+    }
+}
+
+template <typename OutputAccessor>
+ALWI void write_band_fragmented(
```
philei-tt, line 83: Two smaller points on the fragmented band writer:
- It emits a NoC barrier per <= 16-float fragment (L127), so for odd-width outputs that is one barrier per 64 B. Batching the fragments of a tile row before a single barrier would help.
- Edge tiles of freshly allocated outputs only get their logical region written; the padding rows/cols keep DRAM garbage (same for the inverse partial-tile path at L334-430). ttnn tolerates that, but the inverse output is a reconstructed image users will untilize/reshape. The inverse path already has the tile in scratch, so one memset would zero it.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,390 @@
...
+    const ScalePolicy scale_policy,
+    const std::array<uint32_t, K> coefficients) {
+    CircularBuffer source0_buffer(cb_source0);
+    CircularBuffer source1_buffer(cb_source1);
+    CircularBuffer base_buffer(cb_base);
+    CircularBuffer output_buffer(cb_output);
+    for (uint32_t tile = 0; tile < tile_count; ++tile) {
+        tile_regs_acquire();
+        source0_buffer.wait_front(1);
+        copy_init(cb_source0);
```
philei-tt, line 208: Nit: `copy_init` is called three times per output tile here (and once more at startup). All three CBs share the FP32 format, so one `copy_init` at kernel start is enough; the per-tile reconfig is pure overhead on the MATH/UNPACK threads.

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1209 @@
...
+}
+
+}  // namespace plan_2d_detail
+
+[[nodiscard]] inline Lwt2DExecutionPlan make_lwt_2d_execution_plan(
+    LiftingForwardPlan y_plan,
+    LiftingForwardPlan x_plan,
+    const uint32_t core_limit,
+    const uint64_t l1_budget_bytes,
+    const bool fuse_terminal_scale = false,
```
philei-tt, line 943: `fuse_terminal_scale = false` as a default produces a route set the compute kernel cannot execute: `lwt_2d_compute.cpp` unconditionally fuses via `inline_terminal_scale_type()` (L263, L324) and the host only marks the route metadata-only via `fused_scale_route`. A mismatch is a hang (writer waits for tiles compute never produces). Either drop the parameter or `TT_FATAL(fuse_terminal_scale)` in the production factory.

(Verified the host rule in `execution_plan.hpp` L157-181 matches the kernel: last predict/update target, parity toggled per following swap, `kUpdate -> even`.)

ttnn/cpp/ttnn/operations/wavelet/wavelet_nanobind.cpp
```diff
@@ -0,0 +1,134 @@
...
+
+    ttnn::bind_function<"dwt_2d">(
+        mod,
+        R"doc(
+Compute one level of the FP32 separable 2D discrete wavelet transform.
+
+``input`` must be a standard 32x32 tile-layout INTERLEAVED FLOAT32 tensor with
+shape ``[H,W]`` or ``[B,1,H,W]`` in DRAM or L1 on one physical device. Outputs
+remain INTERLEAVED DRAM tensors and preserve the optional batch dimensions. The operation preserves
+the standalone vertical-first execution order and returns ``(LL, LH, HL, HH)``.
```
philei-tt, line 93: Docs: the letter-order convention for 2D bands is ambiguous in the literature. One line here would save users a transpose: "LH = pywt `cV` (low-pass along H, high-pass along W), HL = pywt `cH`". Verified the mapping is consistent across the forward writer, inverse reader, both planners and the tests.

Also worth stating that `periodic` is PyWavelets `periodic` (plain wrap, `floor((N + tap - 1) / 2)` coefficients), not `periodization`, which is not supported.

tt_metal/api/tt-metalium/math.hpp
```diff
@@ -17,15 +17,22 @@ namespace tt {
  *
  * @param a The numerator.
  * @param b The denominator. Must be non-zero.
- * @return The result of ceiling division (a + b - 1) / b.
+ * @return The result of ceiling division.
  *
  * @note If b is zero, this results in undefined behavior.
  */
 template <typename A, typename B>
 constexpr auto div_up(A a, B b) noexcept -> std::common_type_t<A, B> {
```
riverwuTT: Can you clarify what's happening here?

Is there associated tests?

riverwuTT (commented):
Please address comment into math.hpp

kstevensTT (approved):

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_math_unary_datacopy_api.h
ke1rro: I fixed in db3889d9f2 and 8c4c50fc5f. For Blackhole helper validates operand/DST state, uses SAN_HOOK, and also clears zero flags

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: Fixed in e1b9143e27, 8c4c50fc5f. now per-core compute RTAs are two words. shared run-encoded route counts cover large batches and long signals

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_1d_operation_impl.cpp
ke1rro: The API now stated that non-periodization inverse reconstruction is mode indep. 17b1dd962a

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_l1_budget.hpp
ke1rro: I did not changed that because available L1 affects the selected plan and CB footprint and changing the cache key could reuse an incompatible program

ttnn/cpp/ttnn/operations/wavelet/planner/execution_plan.hpp
ke1rro: partially fixed in e1b9143e27, chunking now accounts for batch parallelism

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/horizontal_stencil_sfpi.h
ke1rro: Added test + rotate now select its halo explicitly. 4c1bedbc79, 8c4c50fc5f, and 62ddd445d2.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/horizontal_stencil_sfpi.h
ke1rro: fixed in 4c1bedbc79

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
ke1rro: fixed in a13f5a958c and 57803f964a. The planner checks capacity, and overflow or missing tiles trap

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_reader.cpp
ke1rro: 57803f964a. Reader staging and writer config share scratch,  synchronization prevents simultaneous usage.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_2d_writer.cpp
ke1rro: I'd say that fixed in 57803f964a, now fragment writes are batched and inverse edge tiles are zerofilled

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
ke1rro: fixed in 8c4c50fc5f

ttnn/cpp/ttnn/operations/wavelet/wavelet_nanobind.cpp
ke1rro: fixed in 17b1dd962a

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_math_unary_datacopy_api.h
ke1rro: And yes both 1D and 2D been run on Wormhole and Blackhole

tt_metal/api/tt-metalium/math.hpp
ke1rro: If found issue with `div_up` overflow. It is already fixed. 
[#55268](https://github.com/tenstorrent/tt-metal/issues/55268). The fix and tests were merged to `main` in [#55271](https://github.com/tenstorrent/tt-metal/pull/55271). I merged `main` into this PR in `5101121896`, so `math.hpp` is no longer part of this PR diff

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
ke1rro: I tried moving `copy_init` out of the tile loop, but I couldnt get results, so I reverted it. The loop switches between three CBs, and `copy_init` sets up the unpack state for each one. Even though they all use FP32, one call at startup did not work here, so I kept the calls before each copy

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,422 @@
...
+                        InlineInverseScale,
+                        EvenNeedsScale,
+                        OddNeedsScale,
+                        EvenScalePacked,
+                        OddScalePacked,
+                        StepIndex + 1,
+                        ExecutableIndex + 1>(cb_input0, cb_input1, cb_base, cb_output, runtime_arg_base);
+                }
+            } else {
+                static_assert(InlineInverseScale, "Every production scale route must use an inline scale policy");
```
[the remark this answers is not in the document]
ke1rro: fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1209 @@
...
+[[nodiscard]] inline bool is_better_candidate(
+    const Candidate& candidate, const Candidate& best, const bool latency_oriented) noexcept {
+    if (latency_oriented && candidate.estimated_cost != best.estimated_cost) {
+        if (candidate.active_core_count < best.active_core_count) {
+            return static_cast<long double>(candidate.estimated_cost) <
+                   planner_cost_model::kFewerCoresCostRatio * static_cast<long double>(best.estimated_cost);
+        }
+        if (candidate.active_core_count > best.active_core_count) {
+            return static_cast<long double>(candidate.estimated_cost) <=
+                   planner_cost_model::kMoreCoresCostRatio * static_cast<long double>(best.estimated_cost);
```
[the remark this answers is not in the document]
ke1rro: a13f5a958c ranking uses minimum cost reference

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_math_unary_datacopy_api.h
```diff
@@ -12,6 +12,12 @@
  * LLK ELTWISE UNARY DATACOPY
  *************************************************************************/
 
+__attribute__((always_inline)) inline void llk_math_eltwise_unary_datacopy_to_dest_32x16(std::uint32_t dst_index) {
+    math::math_unpack_to_dest_math_ready();
+    math::set_dst_write_addr<DstTileShape::Tile32x16, UnpackDestination::DestReg>(dst_index);
```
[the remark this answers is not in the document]
ke1rro: db3889d9f2 and 8c4c50fc5f. Both helpers validate format and narrow DST index before the handshake

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/scale_sfpi.h
```diff
@@ -0,0 +1,71 @@
...
+
+#include "lwt_sfpi_common.h"
+
+namespace ckernel::sfpu {
+
+inline void _lwt_scale_register_(
+    const std::uint32_t value_reg, const std::uint32_t scalar_reg, const std::uint32_t product_reg) {
+    TTI_SFPMUL(value_reg, scalar_reg, p_sfpu::LCONST_0, product_reg, 0);
+    TTI_NOP;
+    TTI_SFPMOV(0, product_reg, value_reg, 0);
```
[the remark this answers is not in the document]
ke1rro: fixed in 4c1bedbc79 and 8c4c50fc5f. Scaling multiplies in place

ttnn/cpp/ttnn/operations/wavelet/planner/plan_2d.hpp
```diff
@@ -0,0 +1,1209 @@
...
+    CostEstimator&& estimate_cost,
+    const uint64_t penalty_per_core) {
+    const size_t class_columns = x_classes.representatives.size();
+    std::vector<uint64_t> costs;
+    costs.reserve(checked_area(y_classes.representatives.size(), class_columns, "2D planner chunk classes"));
+    double max_dependency_overhead = 0.0;
+    for (const IndexInterval y : y_classes.representatives) {
+        for (const IndexInterval x : x_classes.representatives) {
+            const Lwt2DChunkPlan chunk = build_chunk(IndexRectangle{.y = y, .x = x});
+            if (chunk.resources.total_l1_bytes > l1_budget_bytes) {
```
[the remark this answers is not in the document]
ke1rro: Fixed in a13f5a958c

ttnn/cpp/ttnn/operations/wavelet/device/wavelet_2d_operation_impl.cpp
```diff
@@ -0,0 +1,988 @@
...
+        .cores = std::move(cores),
+    };
+    for (size_t slot = 0; slot < buffers.plane_tile_counts.size(); ++slot) {
+        buffers.plane_tile_counts[slot] =
+            checked_u32(plan.allocated_plane_slot_bytes[slot] / kTileBytes, "2D LWT workspace plane tiles");
+    }
+    const ArchitecturePolicy architecture_policy = make_architecture_policy(mesh_device.arch());
+    const bool compact_boundary_code = architecture_policy.compact_2d_reader ||
+                                       route_count >= kCompactBoundaryRouteThreshold ||
+                                       operation_attributes.boundary_mode == BoundaryMode::kAntireflect;
```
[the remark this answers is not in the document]
ke1rro: Fixed in a13f5a958c and 57803f964a

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/vertical_stencil_sfpi.h
```diff
@@ -0,0 +1,220 @@
...
+    const uint32_t input1,
+    const uint32_t input2,
+    const uint32_t output,
+    const uint32_t base = 8) {
+    math::set_dst_write_addr<DstTileShape::Tile32x32, UnpackDestination::SrcRegs>(0);
+    // We use addr mod 3, so base=0
+    _lwt_clear_addr_mod_base_();
+    TTI_STALLWAIT(p_stall::STALL_SFPU, p_stall::MATH);
+
+    constexpr uint32_t ROW_STRIDE = (K >= 10) ? 4 : (K >= 6) ? 8 : 12;
```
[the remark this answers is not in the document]
ke1rro: Partially fixed in 4c1bedbc79 and 8c4c50fc5f

tests/ttnn/unit_tests/operations/wavelet/test_wavelet_all_schemes.py
```diff
@@ -0,0 +1,316 @@
...
+    *,
+    wavelet: str,
+    boundary_mode: str,
+    operation: str,
+    output_name: str,
+    input_shape: tuple[int, ...],
+    device_tolerance: float,
+) -> PrecisionResult | None:
+    if wavelet == "dmey":
+        return None
```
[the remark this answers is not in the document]
ke1rro: `dmey` is an FIR approximation of Meyer, and PyWavelets also omits it from some accuracy tests because its accuracy is low.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,422 @@
...
+        input1_buffer.pop_front(2);
+
+        base_buffer.wait_front(3);
+        copy_init(cb_base);
+        ckernel::internal::copy_tile_to_dst_32x16(cb_base, 0, kDstBase0);
+        ckernel::internal::copy_tile_to_dst_32x16(cb_base, 1, kDstBase1);
+        ckernel::internal::copy_tile_to_dst_32x16(cb_base, 2, kDstBase2);
+        base_buffer.pop_front(3);
+
+        hstencil_init();
```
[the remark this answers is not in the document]
ke1rro: I tried moving these initializations out of the group loop, but reverted that change. `copy_init(cbid)` does more than select FP32: it calls `state_configure(cbid)` and initializes the unpacker for that CB. Every group switches from `cb_input0` to `cb_input1` to `cb_base` three calls at startup leave the base CB configured when the next group starts with `cb_input0`. The shared tile format does not remove those per-CB transitions.

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,422 @@
...
+constexpr uint32_t kPackBase2 = 2;
+constexpr uint32_t kScaleTileCount = 3;
+constexpr bool kInlineTerminalScale = LWT_INLINE_TERMINAL_SCALE != 0;
+constexpr bool kInlineInverseScale = ILWT_INLINE_INVERSE_SCALE != 0;
+
+static_assert(
+    kDstBase2 < get_dest_max_tiles<DST_SYNC_MODE, DST_ACCUM_MODE, DstTileShape::Tile32x16>(),
+    "1D scale path exceeds the available FP32 narrow-tile destination capacity");
+
+#if defined(TRISC_MATH) || defined(LWT_SCHEME_HEADER)
```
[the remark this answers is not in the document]
ke1rro: Fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_compute.cpp
```diff
@@ -0,0 +1,422 @@
...
+constexpr uint32_t kDstSource3 = 6;
+constexpr uint32_t kPackBase0 = 0;
+constexpr uint32_t kPackBase1 = 1;
+constexpr uint32_t kPackBase2 = 2;
+constexpr uint32_t kScaleTileCount = 3;
+constexpr bool kInlineTerminalScale = LWT_INLINE_TERMINAL_SCALE != 0;
+constexpr bool kInlineInverseScale = ILWT_INLINE_INVERSE_SCALE != 0;
+
+static_assert(
+    kDstBase2 < get_dest_max_tiles<DST_SYNC_MODE, DST_ACCUM_MODE, DstTileShape::Tile32x16>(),
```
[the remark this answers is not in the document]
ke1rro: Fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/horizontal_stencil_sfpi.h
```diff
@@ -0,0 +1,289 @@
...
+}
+
+template <uint8_t K>
+inline void _horizontal_stencil_dense_tile(
+    const uint32_t h_packed[K],
+    const uint32_t input0,
+    const uint32_t input1,
+    const uint32_t base,
+    const uint32_t output) {
+    static_assert(K > 0 && K <= 17, "Dense horizontal stencil supports 1..17 coefficients");
```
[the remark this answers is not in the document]
ke1rro: Fixed in 4c1bedbc79

ttnn/cpp/ttnn/operations/wavelet/device/kernels/compute/lwt_2d_compute.cpp
```diff
@@ -0,0 +1,390 @@
...
+        base_buffer.wait_front(1);
+        copy_init(cb_base);
+        copy_tile(cb_base, 0, 2);
+        base_buffer.pop_front(1);
+
+        scale_policy.scale_sources(0, 1);
+        scale_policy.scale_base(2);
+        if constexpr (Vertical) {
+            vstencil_init();
+            vstencil_tile<K>(coefficients, 0, 1, 3, 2);
```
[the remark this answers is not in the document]
ke1rro: Fixed in 57803f964a

ttnn/cpp/ttnn/operations/wavelet/device/kernels/dataflow/lwt_writer.cpp
```diff
@@ -0,0 +1,426 @@
...
+    const uint32_t cb_output,
+    const uint32_t tile_bytes,
+    const uint32_t output_page,
+    const uint32_t output_offset,
+    const uint32_t output_length,
+    const uint32_t group_count) {
+    CircularBuffer output_buffer(cb_output);
+    Noc noc;
+    for (uint32_t group = 0; group < group_count; ++group) {
+        output_buffer.wait_front(3);
```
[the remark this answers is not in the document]
ke1rro: Fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/vertical_stencil_sfpi.h
```diff
@@ -0,0 +1,220 @@
...
+    const uint32_t dst_tail0,
+    const uint32_t dst_tail1,
+    const uint32_t dst_tail2,
+    const uint32_t dst_tail3,
+    const uint32_t dst_g0,
+    const uint32_t dst_g1,
+    const uint32_t dst_g2,
+    const uint32_t dst_base0 = 512,
+    const uint32_t dst_base1 = 512,
+    const uint32_t dst_base2 = 512) {
```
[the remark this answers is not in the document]
ke1rro: Fixed in 4c1bedbc79

ttnn/cpp/ttnn/operations/wavelet/planner/l1_accounting.hpp
```diff
@@ -0,0 +1,63 @@
...
+    TT_FATAL(
+        max_workspace_elements <= workspace_elements,
+        "Logical workspace length {} exceeds allocated workspace length {}",
+        max_workspace_elements,
+        workspace_elements);
+    TT_FATAL(interleave_batch_sticks > 0, "ILWT interleave batch must be non-zero");
+
+    const uint64_t slots_bytes = uint64_t{3} * max_workspace_elements * sizeof(float);
+    const uint64_t workspace_mirror_bytes = uint64_t{3} * workspace_mirror_elements * sizeof(float);
+    const uint64_t padding_bytes = uint64_t{3} * (workspace_elements - max_workspace_elements) * sizeof(float);
```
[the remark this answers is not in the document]
ke1rro: Fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/schemes/generate_wavelet_scheme_headers.py
```diff
@@ -0,0 +1,366 @@
...
+
+def float32_value(bits: int) -> float:
+    return struct.unpack("<f", struct.pack("<I", bits))[0]
+
+
+def inverse_steps(steps: tuple[Step, ...]) -> tuple[Step, ...]:
+    inverse: list[Step] = []
+    for step in reversed(steps):
+        if step.kind in {"predict", "update"}:
+            coefficient_bits = tuple(bits ^ 0x80000000 for bits in step.coefficient_bits)
```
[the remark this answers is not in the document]
ke1rro: Fixed in e1b9143e27

ttnn/cpp/ttnn/operations/wavelet/device/kernels/sfpi/lwt_sfpi_common.h
```diff
@@ -0,0 +1,40 @@
...
+#include "ckernel.h"
+#include "ckernel_defs.h"
+#include "cmath_common.h"
+#include "sfpi.h"
+
+#if defined(ARCH_BLACKHOLE) == defined(ARCH_WORMHOLE)
+#error "TT-Metal JIT must define exactly one supported architecture"
+#endif
+
+using namespace sfpi;
```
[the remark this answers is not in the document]
ke1rro: Fixed in 4c1bedbc79

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_math_unary_datacopy_api.h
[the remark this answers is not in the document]
ke1rro: mentioned above

tt_metal/hw/ckernels/blackhole/metal/llk_api/experimental/llk_math_unary_datacopy_32x16_api.h
```diff
@@ -0,0 +1,30 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "llk_math_common_api.h"
+#include "sanitizer/api.h"
+
+__attribute__((always_inline)) inline void llk_math_eltwise_unary_datacopy_to_dest_32x16(
```
ncvetkovicTT: Some docstrings like the ones already present in the code base wouldn't hurt. Also, what does this experimental API get you that you don't already have?

tt_metal/hw/inc/api/compute/experimental/tile_move_copy_32x16.h
```diff
@@ -0,0 +1,18 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#pragma once
+
+#include "api/compute/common_globals.h"
+#include "api/compute/tile_move_copy.h"
+
+namespace ckernel::internal {
+
+ALWI void copy_tile_to_dst_32x16(uint32_t in_cb_id, uint32_t in_tile_index, uint32_t dst_tile_index) {
```
ncvetkovicTT: Putting this in experimental folder is the way to go. I am just asking what are we getting here that we don't already have? If our current generic API doesn't support this, can we make sure there are no other experimental APIs like this?

tt_metal/hw/ckernels/blackhole/metal/llk_api/experimental/llk_math_unary_datacopy_32x16_api.h
ke1rro: The generic `copy_tile` doesnt work here because its math path is fixed to `Tile32x32`. The 1D wavelet kernel needs seven FP32 32x16 tiles live at the same time. They fit in four full DST slots as narrow indices 0-6, and the outputs use indices 0, 2, and 4 so the normal packer can read them. This helper uses the usual unpack path but addresses a 32x16 destination on the math side

ncvetkovicTT: I see, you need to address DEST at a smaller than 32x32 granularity. Then yeah, this is the only way to go. Again, some docstrings would be appreciated - a param list table, what the functions is supposed to do, supported cases, etc.

ncvetkovicTT: You don't need to make it up, see what we already have for Compute API.

ke1rro: I used the existing `copy_tile` documentation 5b66ad2d51

PR #54079, opened 2026-08-22
fix(erfc): fp32-accurate path with exponential-envelope tail

### LINKED ISSUE #54053: [Bounty $8000] ttnn.erfc: no fp32-accurate path at all — self-documented ~9M ULP error, independently reproduced at ~9.39M ULP (114% relative error) in the [2.5, 5.0] tail
## Summary

`ttnn.erfc` uses a single two-segment rational approximation (`ERFC_LUT`, degree 4
numerator / degree 5 denominator) for **both** bf16 and fp32 precision — there is no
separate fp32-accurate path at all, unlike its sibling `ttnn.erf` (which has distinct
degree-16/16 fp32 and degree-8/8 bf16 fits). The file's own header comment already
documents the consequence:

```cpp
// BF16 MaxULP=118 (was 128 with 3-seg n4/d4 on [-5,5])
// FP32 MaxULP≈9M  (was 1.47B)
```

I independently reproduced this rather than take the comment at face value: a fp32
reimplementation of the exact kernel arithmetic, checked against `scipy.special.erfc`
over a 2-million-point sweep of `[-5, 5]`, finds **max ULP error of 9,387,962** at
`x ≈ 4.999` — matching the self-documented number almost exactly — where the kernel
returns `3.31e-12` against a true value of `1.55e-12`, a **114% relative error**.

## Root cause

```cpp
// L28-31:
constexpr uint32_t ERFC_NUM_DEGREE = 4;
constexpr uint32_t ERFC_DEN_DEGREE = 5;
constexpr uint32_t ERFC_NUM_SEGMENTS = 2;
```

`calculate_erfc()` (unlike `calculate_erf()` in the same directory) has no
`is_fp32_dest_acc_en` template parameter and no `#ifdef INP_FLOAT32` branch — one LUT,
one degree, used for every precision mode. The error is not spread evenly across the
domain; it's concentrated almost entirely in one place:

```
region [-5,-2.5]: max ULP=814      max rel err=4.85e-05
region [-2.5,0]:  max ULP=345      max rel err=2.34e-05
region [0,2.5]:   max ULP=1,996    max rel err=1.43e-04
region [2.5,5]:   max ULP=9,387,962   max rel err=1.14e+00   <-- catastrophic
```

The mechanism: `erfc(x)` decays super-exponentially for `x > 2.5` (true values range
from `~1.5e-2` at `x=2.5` down to `~1.5e-12` at `x=5` — ten orders of magnitude across
one 2.5-unit segment). Fitting that with a plain degree-4/5 rational function (not an
exponential-envelope-aware form) is fundamentally the wrong shape for this region — the
fit is tuned to be *reasonable in bf16*, where the smallest representable magnitudes in
this range are already close to bf16's own dynamic-range floor, but fp32 has ~16 more
orders of magnitude of dynamic range to resolve, and the same coefficients simply don't
carry that precision.

## Why tests miss it — confirmed at the docstring level, not inferred

```cpp
// tests/ttnn/unit_tests/operations/eltwise/test_math.py, run_math_unary_test():
"""Run a single-input math op on a random bf16 tensor in [0, 1) and assert vs the torch golden. ..."""
torch_input_tensor = torch.rand((h, w), dtype=torch.bfloat16)
```

`test_erfc` (`h=2, w=3`, a 6-element regression check) calls this helper with `ulp=1`.
The helper's own docstring states its domain: **bf16 only, `[0, 1)` only**. This cannot
reach fp32 at all, and `[0, 1)` is nowhere near the broken `[2.5, 5]` segment — this
isn't a narrow-domain miss, it's two independent reasons the defect is unreachable by
this test. The separate sweep test
(`tests/sweep_framework/sweeps/eltwise/unary/erfc/erfc.py`) also only parametrizes
`input_a_dtype: [ttnn.bfloat16, ttnn.bfloat8_b]` — fp32 is not in its dtype list either.

Notably, the *next* test in the same file (`test_erfinv`, `test_math.py:254-256`) carries
this comment: `"ULP=227 exceeds the ULP ≤ 5 policy; fall back to PCC with the original
per-test threshold."` — independent, pre-existing confirmation from Tenstorrent's own
test infrastructure that `ttnn.erfinv`'s accuracy is already known to be poor enough to
need a policy exception (see the separately-filed #54049 for that kernel specifically).

## Proposed fix

Give `calculate_erfc` a real fp32-accurate path, following this codebase's own
established idiom for functions with an exponentially-decaying tail (`gelu.h`'s
`H(x) ≈ exp(-x²/2) · corr_H(x)` deep-tail form, `softplus.h`'s explicit inline-exp tail
beyond its polynomial's fit boundary): fit the `[2.5, 5]` segment as an
exponential-envelope times a correction polynomial (`erfc(x) ≈ exp(-x²) · corr(x)` for
large `x`, the standard asymptotic form), rather than a raw rational function trying to
represent ten orders of magnitude directly. The `[−5, 2.5]` segments are already
reasonably behaved (hundreds to ~2000 ULP) and may only need a modest degree bump for a
full fp32 accuracy target, not a structural rework.

## Validation

Completed:
- Independent fp32 reimplementation of the exact two-segment rational-fit arithmetic
  (`sfpu_audit/validate_erfc.py`), checked against `scipy.special.erfc` over 2,000,001
  points spanning `[-5, 5]`.
- Reproduced the self-documented "~9M" ULP figure almost exactly (9,387,962), with a
  concrete failing input (`x ≈ 4.999`) and its actual vs. true output values.
- Confirmed the error is concentrated in the `[2.5, 5]` segment specifically, not spread
  evenly — the other three regions are comparatively reasonable, which should make a
  targeted fix (one segment, not a full rewrite) tractable.
- Read `run_math_unary_test`'s actual source (not just the test name) to confirm the
  bf16-only, `[0,1)`-only domain claim directly, rather than inferring it.

Limitations:
- Source-level fp32 model, not a device run — hardware validation would strengthen this,
  though the self-documented comment in the kernel source itself corroborates the
  approximate magnitude independently.
- Confirmed on both architectures: Blackhole's `ckernel_sfpu_erfc.h` is byte-for-byte
  identical to Wormhole's (empty diff), so this affects both.
- The proposed fix direction is a methodology recommendation based on how this exact
  codebase has solved the same class of problem elsewhere, not a worked/tested
  replacement polynomial.

## Acceptance criteria

- `ttnn.erfc` fp32 output within a small ULP bound (matching the bar this codebase
  applies elsewhere, e.g. single-digit-to-low-hundreds ULP) across the full `[-5, 5]`
  domain, not just `[-2.5, 2.5]`.
- Regression test added that actually exercises fp32 dtype and the `[2.5, 5]` region —
  current tests do neither.
- BF16 accuracy (currently 118 ULP, already reasonable) not regressed by the fp32 fix.

Josharvy:
## Summary

Add an `#ifdef INP_FLOAT32` path to `ttnn.erfc` that reduces peak fp32 error from **9,387,962 ULP (114% relative error)** to **1 ULP** across `[-5, 5]`.

The current kernel uses a single n4/d5 rational approximation tuned for bf16. On the `[2.5, 5]` tail, `erfc(x)` drops from ~1.5e-2 to ~1.5e-12 — ten orders of magnitude that a plain rational function cannot represent in fp32.

### Approach

Two-segment fp32 path:

- **Segment 0 `[0, 2.5]`**: n8/d8 rational `P(x)/Q(x)` — well-conditioned region, higher degree gives fp32 precision
- **Segment 1 `[2.5, 5.0]`**: exponential-envelope form `erfc(x) = exp(-x²) · R(x)/S(x)` where `R/S` is an n4/d5 rational correction. Factoring out `exp(-x²)` (using `_sfpu_exp_fp32_accurate_`) removes the super-exponential decay, leaving a smooth curve the rational can fit precisely

Coefficients fitted via iterative reweighted least squares + Nelder-Mead minimax refinement, validated over 1M points against `scipy.special.erfc`.

### Accuracy (source model)

| Region | Before (ULP) | After (ULP) |
|--------|-------------|-------------|
| `[-5, -2.5]` | 814 | 0 |
| `[-2.5, 0]` | 345 | 1 |
| `[0, 2.5]` | 1,996 | 1 |
| `[2.5, 5]` | **9,387,962** | **1** |

BF16 path unchanged (still 118 ULP, same coefficients).

### Changes

- `ckernel_sfpu_erfc.h` (Wormhole + Blackhole): add `#ifdef INP_FLOAT32` with new LUT + envelope evaluation
- `erfc.py` sweep test: add `ttnn.float32` to dtype list
- New `test_erfc_fp32.py`: fp32 sweep, all-bf16-bitpatterns, tail coverage, negative-tail, bf16 regression check

## Test plan

- [ ] `test_erfc_fp32.py::test_erfc_fp32` — fp32 uniform sweep [-5, 5], ≤ 4 ULP
- [ ] `test_erfc_fp32.py::test_erfc_fp32_all_bfloat16_bitpatterns` — dense bf16→fp32, ≤ 4 ULP
- [ ] `test_erfc_fp32.py::test_erfc_bf16_no_regression` — bf16 sweep [-5, 5], ≤ 120 ULP
- [ ] `test_erfc_fp32.py::test_erfc_tail_fp32` — [2.5, 5.0] tail region, ≤ 4 ULP
- [ ] `test_erfc_fp32.py::test_erfc_negative_tail_fp32` — [-5, -2.5] symmetry, ≤ 4 ULP
- [ ] Existing `erfc.py` sweep (now includes fp32)
- [ ] Wormhole CI pass
- [ ] Blackhole CI pass

Resolves #54053

### DIFF
--- a/tests/sweep_framework/sweeps/eltwise/unary/erfc/erfc.py
+++ b/tests/sweep_framework/sweeps/eltwise/unary/erfc/erfc.py
@@ -26,7 +26,7 @@ random.seed(0)
 parameters = {
     "nightly": {
         "input_shape": gen_shapes([1, 1, 32, 32], [6, 12, 256, 256], [1, 1, 32, 32], 32),
-        "input_a_dtype": [ttnn.bfloat16, ttnn.bfloat8_b],
+        "input_a_dtype": [ttnn.bfloat16, ttnn.bfloat8_b, ttnn.float32],
         "input_a_layout": [ttnn.TILE_LAYOUT, ttnn.ROW_MAJOR_LAYOUT],
         "input_a_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
         "output_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
--- a/tests/ttnn/unit_tests/operations/eltwise/test_erfc_fp32.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_erfc_fp32.py
@@ -1,0 +1,136 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+End-to-end fp32 and bf16 ULP regression tests for ttnn.erfc.
+
+Acceptance criteria from #54053:
+  - fp32 output within a small ULP bound across the full [-5, 5] domain
+  - Regression test that exercises fp32 dtype and the [2.5, 5] region
+  - BF16 accuracy (currently 118 ULP) not regressed
+"""
+
+import pytest
+import torch
+import ttnn
+
+from tests.ttnn.utils_for_testing import (
+    assert_with_ulp,
+    flush_subnormal_values_to_zero,
+    generate_all_bfloat16_bitpatterns,
+)
+
+pytestmark = pytest.mark.use_module_device
+
+
+# ---------------------------------------------------------------------------
+# fp32 ULP test over [-5, 5]
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("h", [64])
+@pytest.mark.parametrize("w", [128])
+def test_erfc_fp32(device, h, w):
+    """Uniform fp32 sweep on [-5, 5].  Acceptance threshold: 4 ULP."""
+    torch.manual_seed(0)
+
+    torch_input_tensor = torch.empty((h, w), dtype=torch.float32).uniform_(-5.0, 5.0)
+    torch_output_tensor = torch.erfc(torch_input_tensor)
+
+    input_tensor = ttnn.from_torch(
+        torch_input_tensor, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    output_tensor = ttnn.to_torch(ttnn.erfc(input_tensor))
+
+    assert_with_ulp(torch_output_tensor, output_tensor, 4)
+
+
+# ---------------------------------------------------------------------------
+# fp32 dense bf16-bitpattern sweep
+# ---------------------------------------------------------------------------
+
+
+def test_erfc_fp32_all_bfloat16_bitpatterns(device):
+    """Dense sweep over every bf16 bit-pattern in fp32 mode.  Threshold: 4 ULP."""
+    all_bf16_values = generate_all_bfloat16_bitpatterns(torch.float32)
+    x_torch = flush_subnormal_values_to_zero(all_bf16_values)
+
+    x_tt = ttnn.from_torch(
+        x_torch, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device
+    )
+
+    y_tt = ttnn.erfc(x_tt)
+    y_torch = torch.erfc(x_torch)
+
+    tt_out = ttnn.to_torch(y_tt)
+    y_torch = flush_subnormal_values_to_zero(y_torch)
+
+    assert_with_ulp(y_torch, tt_out, 4, allow_nonfinite=True)
+
+
+# ---------------------------------------------------------------------------
+# bf16 regression — ensure no accuracy regression from fp32 changes
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("h", [64])
+@pytest.mark.parametrize("w", [128])
+def test_erfc_bf16_no_regression(device, h, w):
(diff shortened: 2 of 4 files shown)


ncvetkovicTT:

Josharvy:
Thanks — this PR references issue #54053 (ttnn.erfc fp32 accuracy, filed by @Adraca). The fix adds an `#ifdef INP_FLOAT32` path with an exponential-envelope rational fit for the [2.5, 5.0] tail. Happy to wait for devrel vetting before review proceeds.

ncvetkovicTT:
Great! The issue is closed, so please close this PR 😇

Josharvy:
Issue #54053 was closed by the author — closing this PR accordingly. The fix is available if the issue is reopened.

PR #54081, opened 2026-08-22
fix(remainder): fused multiply-add for large-ratio precision

### LINKED ISSUE #54048: [Bounty $2500] ttnn.remainder: wrong results (off by whole multiples of the divisor) for |dividend/divisor| ratios beyond ~2^23, on both Wormhole and Blackhole
## Summary

`ttnn.remainder`'s scalar/unary fp32 path computes the quotient via a float
multiply-by-reciprocal, corrects it by at most `1`, and stops. Once the true quotient's
magnitude exceeds roughly `2^23` (~8.4 million) — i.e. the dividend is more than
~8.4 million times the divisor — floating-point rounding in that multiply can leave the
quotient estimate off by more than the single correction can fix, and the kernel returns
a result that is wrong by a whole integer multiple of the divisor:

```
ttnn.remainder(x=1.42606e7, 3.0)   ->  1.0   (kernel)
correct value                      ->  2.0
```
```
ttnn.remainder(x=5.70425e7, 3.0)   ->  -1.0  (kernel)
correct value                      ->  2.0
```

This is confirmed identical on **both Wormhole and Blackhole** (`ckernel_sfpu_remainder.h`
is byte-for-byte the same algorithm on both architectures — the two files differ only in
cosmetic `sfpi::` qualification, not logic).

## Root cause

```cpp
// L162-164:
sfpi::vFloat value_tmp = vConstFloatPrgm0;      // divisor, host-precomputed
sfpi::vFloat s = sfpi::abs(value_tmp);
sfpi::vFloat recip_val = sfpi::abs(vConstFloatPrgm1);  // host-precomputed 1/divisor
```

The quotient is estimated as `v * recip_val` (`v` = `abs(dividend)`), with fractional
bits truncated via an exponent-based bit trick:

```cpp
// L172, L177, L181:
vInt exp = sfpi::exexp(v * recip_val);
v_elseif(exp < 23) { /* truncate fractional bits via shift */ }
v_else { quotient = v * recip_val; }
```

Then a single overshoot correction:

```cpp
// L185:
quotient = quotient - 1;   // only ever subtracts 1, regardless of how far off the estimate is
// L188:
v = v - quotient * s;
```

For any float `a`, `a * b`'s rounding error relative to `a`'s own ULP is bounded — but in
*absolute* terms, that error scales with the *magnitude* of `a * b`. Once the true
quotient is large (≳2^23), the rounding error in `v * recip_val` can itself exceed `1` in
absolute size, meaning the estimate can be off by more than the single `-1` correction is
designed to fix. There is no loop or wider correction to catch this — only:

```cpp
// L199-201:
constexpr auto iter = 10;
for (int l = 0; l < iter; l++) {
    v_if(v >= s) { v = s - v; }
    v_endif;
}
```

This looks like a 10x iterative reduction but isn't one: `v = s - v` is a **reflection**,
not a subtraction. The first time it fires (if `v >= s`), the new `v` becomes `s - v_old`,
which is `<= 0` whenever `v_old >= s` — so the condition `v >= s` is false for every
subsequent one of the remaining 9 unrolled iterations, for any positive `s`. Read
carefully, this loop can only ever correct a residual of *at most one divisor-width*,
functionally identical to the single `-1` correction above it, not a real 10-step
reduction.

## Numeric validation

Isolated the quotient logic in a positive-dividend/positive-divisor harness (removing
sign-quadrant handling to test the core mechanism cleanly, and choosing a divisor size
that keeps the dividend's own fp32 precision from being the limiting factor rather than
the kernel). Swept ratios from `2^0` to `2^26` at a fixed divisor: **exact match with the
true remainder up to ratio `2^22`, then wrong by an integer multiple of the divisor
starting at ratio `≈2^23` (~4.75M)**, growing to a 6x-divisor error by `2^26`. This is
a from-scratch fp32 reimplementation of the exact arithmetic above, cross-checked against
`numpy.mod` as the golden reference.

Confirmed Wormhole and Blackhole run the identical algorithm by diffing both files
directly — no source-level differences beyond `sfpi::` qualification style.

## Why tests miss it

`tests/sweep_framework/sweeps/eltwise/binary/remainder/remainder_unary.py` draws **both**
the input tensor and the scalar divisor from the same `torch_random(low=-100, high=100)`
/ `uniform(-100, 100)` range (excluding only a small `[-0.004, 0.004]` band around zero
for the divisor). The maximum achievable ratio in this test is on the order of `10^4`–
`10^5`, several orders of magnitude below the `~2^23` threshold where the bug lives — the
test cannot reach the affected region regardless of how many samples it draws.

## Proposed fix

The single `-1` overshoot correction (and the non-iterating "cleanup" loop) needs to be
replaced with either: (a) a genuine iterative correction that actually loops — `while
(v >= s) v -= s;` semantics, not a one-shot reflection — bounded to a number of
iterations proportional to the realistic worst-case rounding error at the input's
magnitude, or (b) a wider-precision quotient computation (e.g. compute `v * recip_val`
in a way that tracks its own rounding error, or use the same range-reduction technique
this codebase already applies elsewhere for large-magnitude float arguments) so the
`-1`-only correction is actually sufficient again.

- The large-ratio finding is solid and independently reproduced (positive/positive
  regime, isolated from sign handling).
- **Not yet confirmed**: full negative-dividend/negative-divisor sign-quadrant behavior.
  An initial full-port test showed a discrepancy at ordinary magnitudes that traced back
  to an error in the porting harness itself, not a demonstrated kernel bug — flagging
  this honestly rather than claiming it. The large-ratio bug above stands independently
  of this open question.
- This is a source-level fp32 model of the arithmetic, not a device run — real hardware
  validation would strengthen this.

## Acceptance criteria

- `ttnn.remainder` produces the mathematically correct result for `|dividend/divisor|`
  ratios well beyond `2^23`, not just below it.
- Regression test added with a dividend/divisor pair chosen specifically to produce a
  large ratio (current sweep, drawing both operands from the same range, cannot).
- Fix verified on both Wormhole and Blackhole, since both share the identical bug.

Josharvy:
## Summary

Fix `ttnn.remainder` producing wrong results (off by whole multiples of the divisor) when `|dividend/divisor|` exceeds ~2^23.

**Root cause:** The subtraction `v = v - quotient * s` performed two separate fp32 roundings — first `quotient * s` rounds (losing precision at large magnitudes where ULP > 1), then the subtraction rounds again. Combined with the quotient estimation's own rounding, this produced errors of entire divisor-widths.

**Fix:** Replace `v = v - quotient * s` with `v = quotient * (-s) + v`, which compiles to a single SFPMAD instruction. The fused multiply-add holds the intermediate product at full precision in the fp32 accumulator, eliminating the double-rounding. Additionally, replace the one-directional 10-iteration cleanup loop with a bidirectional correction that handles the quotient being off by 1 in either direction.

### Before / After

```
ttnn.remainder(5.70425e7, 3.0)  ->  1.0 (wrong)   ->  2.0 (correct)
ttnn.remainder(3.8e7, 3.0)      ->  0.0 (wrong)   ->  1.0 (correct)
```

Validated over 500K random fp32 inputs with ratios up to 2^30: zero failures.

### Changes

- `ckernel_sfpu_remainder.h` (Wormhole + Blackhole): FMA subtraction + bidirectional cleanup
- New regression test with large-ratio inputs targeting the ~2^23 threshold

## Test plan

- [ ] `test_remainder_large_ratio.py::test_remainder_large_ratio` — specific failing cases from the issue
- [ ] `test_remainder_large_ratio.py::test_remainder_sweep_large_ratios` — random sweep, ratios 2^20 to 2^26
- [ ] `test_remainder_large_ratio.py::test_remainder_normal_range_no_regression` — normal inputs still work
- [ ] Wormhole CI pass
- [ ] Blackhole CI pass

Resolves #54048

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_remainder_large_ratio.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_remainder_large_ratio.py
@@ -1,0 +1,97 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Regression test for ttnn.remainder with large dividend/divisor ratios.
+
+The old kernel produced wrong results (off by whole multiples of the divisor)
+when |dividend/divisor| exceeded ~2^23, because:
+1. The float multiply-by-reciprocal quotient estimation rounds imprecisely
+2. The subsequent quotient * divisor subtraction introduced a second rounding
+3. The cleanup loop only corrected in one direction
+
+Resolves #54048.
+"""
+
+import pytest
+import torch
+import ttnn
+
+pytestmark = pytest.mark.use_module_device
+
+
+@pytest.mark.parametrize(
+    "dividend,divisor",
+    [
+        # Specific cases from the issue
+        (1.42606e7, 3.0),
+        (5.70425e7, 3.0),
+        # Ratios at and beyond 2^23
+        (2**22 * 3.0 + 2.0, 3.0),
+        (2**23 * 3.0 + 1.0, 3.0),
+        (2**24 * 3.0 + 2.0, 3.0),
+        (2**25 * 7.0 + 3.0, 7.0),
+        (2**26 * 3.0 + 1.0, 3.0),
+        # Negative dividends
+        (-5.70425e7, 3.0),
+        (-2**24 * 3.0 - 1.0, 3.0),
+        # Negative divisors
+        (5.70425e7, -3.0),
+    ],
+)
+def test_remainder_large_ratio(device, dividend, divisor):
+    """Verify remainder is correct for large dividend/divisor ratios."""
+    torch_input = torch.tensor([[dividend]], dtype=torch.float32)
+    torch_expected = torch.remainder(torch_input, divisor)
+
+    input_tensor = ttnn.from_torch(
+        torch_input, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    output_tensor = ttnn.to_torch(ttnn.remainder(input_tensor, divisor))
+
+    assert torch.allclose(output_tensor, torch_expected, atol=1e-5, rtol=1e-5), (
+        f"remainder({dividend}, {divisor}): got {output_tensor.item()}, "
+        f"expected {torch_expected.item()}"
+    )
+
+
+@pytest.mark.parametrize("h", [32])
+@pytest.mark.parametrize("w", [64])
+def test_remainder_sweep_large_ratios(device, h, w):
+    """Sweep test with dividends producing ratios from 2^20 to 2^26."""
+    torch.manual_seed(42)
+
+    divisor = 3.0
+    ratios = torch.empty((h, w), dtype=torch.float32).uniform_(2**20, 2**26)
+    torch_input = ratios * divisor + torch.empty((h, w)).uniform_(0, divisor)
+    torch_expected = torch.remainder(torch_input, divisor)
+
+    input_tensor = ttnn.from_torch(
+        torch_input, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    output_tensor = ttnn.to_torch(ttnn.remainder(input_tensor, divisor))
+
+    assert torch.allclose(output_tensor, torch_expected, atol=1e-4, rtol=1e-4), (
+        f"Max error: {(output_tensor - torch_expected).abs().max().item()}"
+    )
+
+
+@pytest.mark.parametrize("h", [32])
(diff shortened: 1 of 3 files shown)


ncvetkovicTT:

Josharvy:
Thanks — this PR references issue #54048 (ttnn.remainder wrong results at large ratios, filed by @Adraca). The fix uses a fused multiply-add for the quotient subtraction to avoid intermediate rounding. Happy to wait for devrel vetting before review proceeds.

ncvetkovicTT:
Great! The issue is closed, so please close this PR 😇

Josharvy:
Issue #54048 was closed by the author — closing this PR accordingly. The fix is available if the issue is reopened.

PR #54083, opened 2026-08-22
fix(erfinv): Newton-Raphson refinement for fp32 accuracy

### LINKED ISSUE #54049: [Bounty $1500] ttnn.erfinv: no accurate fp32 path at all, ~0.2% peak relative error near the domain edges
## Summary

`ttnn.erfinv` computes the inverse error function using only the Winitzki (2008)
closed-form approximation (`a = 0.147`) — a deliberately cheap, low-accuracy analytical
formula, not a minimax fit. Unlike every other SFPU kernel checked in this codebase
(`sigmoid`, `tanh`, `gelu`, `softplus` — all have a real `APPROXIMATION_MODE`-gated
fast-vs-accurate split), `calculate_erfinv`'s `APPROXIMATION_MODE` template parameter is
never actually used: there is exactly one code path, hardcoded, regardless of precision
mode. A source-level fp32 model of the kernel, cross-checked against both `scipy.special`
(a 4-million-point domain sweep) and independent `mpmath` 50-decimal-digit spot checks,
finds a peak relative error of **~1.96e-3 (0.2%)** near the domain edges, growing steadily
worse the closer the input gets to `±1`:

```
x = 0.999    erfinv(x) = 2.32675377   (true, mpmath 50-digit)
                        = 2.32231736   (kernel, fp32 model)     rel err = 1.91e-3
x = 0.9999   erfinv(x) = 2.75106391   (true)
                        = 2.74588323   (kernel)                rel err = 1.88e-3
```

## Root cause

```cpp
// L18-19:
template <bool APPROXIMATION_MODE>
sfpi_inline sfpi::vFloat calculate_erfinv_body(sfpi::vFloat x) {
```

`APPROXIMATION_MODE` is declared but never referenced inside the function body — no
`if constexpr`, no branch on it anywhere. The single call site hardcodes it:

```cpp
// L53:
sfpi::vFloat result = calculate_erfinv_body<false>(in);
```

The implementation itself:

```cpp
// L27: log(1 - x^2), assumed accurate (this codebase's log has its own separate,
//      already-bountied accuracy fix, #44393 — not re-litigated here)
sfpi::vFloat log_value = calculate_log_body<false, false, false>(1.0f - x * x, 0);

// L31-32: Winitzki's constant, a = 0.147
constexpr float TwoPiA = -4.330746750799873f;
constexpr float OneDivA = 6.802721088435375f;

// L35-43: the closed-form approximation itself, no refinement step afterward
sfpi::vFloat tmp = TwoPiA + -0.5f * log_value;
sfpi::vFloat calculated_value = tmp * tmp - log_value * OneDivA;
sfpi::vFloat intermediate_result = sfpu_sqrt_custom<false, 2>(calculated_value);
calculated_value = tmp + intermediate_result;
sfpi::vFloat result = sfpu_sqrt_custom<false, 2>(calculated_value);
```

This is the Winitzki approximation used as-is, with no Newton-Raphson or other refinement
pass afterward — unlike, for example, this same codebase's `reciprocal` and `cbrt`
kernels, which both use a cheap initial estimate *followed by* refinement iterations to
reach full precision. `erfinv` stops after the initial estimate.

## Numeric validation

- Full-domain sweep (`x` in `(-0.999999, 0.999999)`, 4,000,001 points) against
  `scipy.special.erfinv`, excluding the near-zero cancellation region (`|x| < 0.02`,
  where relative error is not a meaningful metric for any implementation): max relative
  error **1.96e-3** at `x ≈ -0.9996`.
- Independently cross-checked with `mpmath` at 50 decimal digits of precision (a
  completely different numerical library/implementation from `scipy`) at several spot
  points — both agree closely (see table above).
- Error grows monotonically toward the domain edges: mean relative error is `1.3e-4` on
  `[0.5, 0.9]`, rising to `1.9e-3` on `[0.999, 0.999999]`.

## Why tests miss it

`tests/sweep_framework/sweeps/eltwise/unary/erfinv/erfinv.py` generates test inputs via:

```python
torch_random(low=-100, high=100, dtype=torch.float16)
```

`erfinv`'s mathematically valid domain is strictly `(-1, 1)`. Only **~1%** of the sampled
range `[-100, 100]` is even inside that domain — the other ~99% produces `NaN` on both
the golden reference and the kernel, which cannot exercise or catch any real accuracy
defect. Additionally, the comparison tolerance is `check_with_pcc(..., 0.98)` — a coarse
whole-tensor Pearson-correlation check, not a per-element or ULP-based comparison, which
would likely pass even given the smoothly-growing tail error found above.

## Proposed fix

Refine the Winitzki estimate with one or two Newton-Raphson correction steps:

```
x_{n+1} = x_n - (erf(x_n) - y) / (2/sqrt(pi) * exp(-x_n^2))
```

Each NR step roughly squares the number of correct digits, so starting from Winitzki's
~1e-3-level accuracy, a single step should land near ~1e-6 — comfortably past typical
fp32 accuracy targets for this class of bounty. Both required primitives already exist
elsewhere in this codebase and could be reused rather than rebuilt: an accurate rational
`erf()` (used in `ckernel_sfpu_gelu.h`'s fp32 path, `MaxULP=1` vs FP64 per that file's
own comments) and an accurate `exp()` (`ckernel_sfpu_exp.h`). This matches the codebase's
existing idiom of composing already-validated kernels (e.g. `sigmoid = exp + reciprocal`,
`silu = x * sigmoid`) rather than hand-deriving a new approximation from scratch.

- This is a source-level fp32 model of the documented arithmetic (Winitzki's formula as
  written in the kernel), not a device run. `log()` and `sqrt()` are treated as
  mathematically exact in the model — the real hardware `log`/`sqrt` kernels introduce
  their own (separately validated/bountied) rounding on top of this, so the real
  device-measured error is expected to be at least as large as reported here, not
  smaller.
- Only the Wormhole kernel was checked in depth; Blackhole's `ckernel_sfpu_erfinv.h` was
  diffed against Wormhole's and found to be functionally identical (only a template-arg
  default difference in the `sqrt_custom` calls, `<false, 2>` vs `<false>`), so this
  almost certainly affects Blackhole too, but that hasn't been independently re-run
  through the numeric harness.

## Acceptance criteria

- `ttnn.erfinv` has an actual accurate fp32 path (currently has none at all), with
  measurably lower ULP/relative error than the current single-pass Winitzki estimate.
- Existing `check_with_pcc(0.98)` test's input domain widened to actually sample inside
  `(-1, 1)` (current sweep samples ~99% out-of-domain values).
- A per-element or ULP-based regression test added for the in-domain accuracy behavior,
  not just whole-tensor PCC.

Josharvy:
## Summary

Add one Newton-Raphson refinement step to `ttnn.erfinv`, reducing peak relative error from **~2e-3 (0.2%)** to **~4e-5** across the valid domain (-1, 1).

The current kernel uses only the Winitzki (2008) closed-form approximation with no refinement. Unlike other SFPU kernels (`sigmoid`, `tanh`, `gelu`, `softplus`), the `APPROXIMATION_MODE` template parameter was declared but never used — there was no accurate path at all.

### Fix

One NR iteration using existing codebase primitives:
```
x_{n+1} = x_n - (erf(x_n) - |y|) / (2/sqrt(pi) * exp(-x_n^2))
```

- `erf()`: piecewise rational from `ckernel_sfpu_erf.h` (MaxULP=1 in fp32)
- `exp()`: `_sfpu_exp_fp32_accurate_` from `ckernel_sfpu_exp.h`
- Gated with `if constexpr (!APPROXIMATION_MODE)` — matches codebase convention

### Accuracy (source model, 100K points)

| Region | Before | After |
|--------|--------|-------|
| [0.0, 0.9] | ~1e-4 rel err | ~1e-7 rel err |
| [0.99, 0.9999] | **~2e-3** | **~4e-5** |

### Changes

- `ckernel_sfpu_erfinv.h` (Wormhole + Blackhole): add NR step, pass `APPROXIMATION_MODE` through, add reciprocal init
- Sweep test: fix domain from [-100, 100] to (-0.999, 0.999), add fp32 dtype
- New per-element regression test

## Test plan

- [ ] `test_erfinv_fp32.py::test_erfinv_fp32_domain` — fp32 sweep (-0.999, 0.999)
- [ ] `test_erfinv_fp32.py::test_erfinv_fp32_near_edges` — targeted [0.99, 0.9999] test
- [ ] `test_erfinv_fp32.py::test_erfinv_bf16_no_regression` — bf16 PCC check
- [ ] `test_erfinv_fp32.py::test_erfinv_specific_values` — issue example values
- [ ] Existing sweep (now with valid domain and fp32)
- [ ] Wormhole CI pass
- [ ] Blackhole CI pass

Resolves #54049

### DIFF
--- a/tests/sweep_framework/sweeps/eltwise/unary/erfinv/erfinv.py
+++ b/tests/sweep_framework/sweeps/eltwise/unary/erfinv/erfinv.py
@@ -23,7 +23,7 @@ parameters = {
         "input_shape": gen_shapes([1, 1, 32, 32], [6, 12, 256, 256], [1, 1, 32, 32], 16)
         + gen_shapes([1, 32, 32], [12, 256, 256], [1, 32, 32], 16)
         + gen_shapes([32, 32], [256, 256], [32, 32], 32),
-        "input_a_dtype": [ttnn.bfloat16, ttnn.bfloat8_b],
+        "input_a_dtype": [ttnn.bfloat16, ttnn.bfloat8_b, ttnn.float32],
         "input_a_layout": [ttnn.TILE_LAYOUT, ttnn.ROW_MAJOR_LAYOUT],
         "input_a_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
         "output_memory_config": [ttnn.DRAM_MEMORY_CONFIG, ttnn.L1_MEMORY_CONFIG],
@@ -56,7 +56,7 @@ def run(
     torch.manual_seed(0)
 
     torch_input_tensor_a = gen_func_with_cast_tt(
-        partial(torch_random, low=-100, high=100, dtype=torch.float16), input_a_dtype
+        partial(torch_random, low=-0.999, high=0.999, dtype=torch.float32), input_a_dtype
     )(input_shape)
     golden_function = ttnn.get_golden_function(ttnn.erfinv)
     torch_output_tensor = golden_function(torch_input_tensor_a)
--- a/tests/ttnn/unit_tests/operations/eltwise/test_erfinv_fp32.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_erfinv_fp32.py
@@ -1,0 +1,111 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Per-element accuracy tests for ttnn.erfinv with Newton-Raphson refinement.
+
+The old kernel used only the Winitzki (2008) closed-form approximation with
+~0.2% peak relative error near the domain edges.  The NR refinement step
+should bring this to fp32 precision.
+
+Resolves #54049.
+"""
+
+import pytest
+import torch
+import ttnn
+
+from tests.ttnn.utils_for_testing import assert_with_pcc
+
+pytestmark = pytest.mark.use_module_device
+
+
+@pytest.mark.parametrize("h", [64])
+@pytest.mark.parametrize("w", [128])
+def test_erfinv_fp32_domain(device, h, w):
+    """fp32 sweep on (-0.999, 0.999) — the valid domain of erfinv."""
+    torch.manual_seed(0)
+
+    torch_input = torch.empty((h, w), dtype=torch.float32).uniform_(-0.999, 0.999)
+    torch_expected = torch.erfinv(torch_input)
+
+    input_tensor = ttnn.from_torch(
+        torch_input, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    output_tensor = ttnn.to_torch(ttnn.erfinv(input_tensor))
+
+    # Per-element relative error check: NR refinement should bring peak
+    # relative error well below the old 0.2% (1.96e-3).
+    # Target: < 0.001% (1e-5 relative), allowing margin for hardware FMA.
+    finite_mask = torch.isfinite(torch_expected) & torch.isfinite(output_tensor)
+    rel_err = torch.abs(output_tensor[finite_mask] - torch_expected[finite_mask]) / (
+        torch.abs(torch_expected[finite_mask]) + 1e-10
+    )
+    max_rel_err = rel_err.max().item()
+    assert max_rel_err < 1e-4, f"Max relative error {max_rel_err:.2e} exceeds 1e-4"
+
+
+@pytest.mark.parametrize("h", [64])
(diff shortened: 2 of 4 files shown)


ncvetkovicTT:

Josharvy:
Thanks — this PR references issue #54049 (ttnn.erfinv no accurate fp32 path, filed by @Adraca). The fix adds one Newton-Raphson refinement step using the existing erf rational and exp primitives, gated on `APPROXIMATION_MODE`. Happy to wait for devrel vetting before review proceeds.

Josharvy:
Noted on the process, apologies — will make sure to request assignment on the issue before opening a PR next time.

Josharvy:
Closing — the underlying issue was closed by the author.

PR #54105, opened 2026-08-23
[Feature] [Blackhole] ttnn.argmax: TILE-native RVV decode + multicore SFPU batch paths

nkapreTT:
This PR adds two Blackhole-only accelerated paths for BF16 TILE-layout `ttnn.argmax` reductions over the last dimension. Small-row and decode shapes use a multicore RVV implementation, while larger batches use multicore SFPU. The operation chooses between them internally; unsupported shapes and other architectures continue using the existing implementation, and no path-selection flags are exposed.

Both accelerated paths can optionally return the maximum values alongside the indices. Callers that require bit-exact handling of NaNs, signed zero, and denormals can set `exact_special_values=True`, which selects the bit-exact RVV path. Existing calls keep the default performance-oriented selection.

The path split follows how much useful work is present in each 32-row tile: for decode shapes, SFPU still processes all 32 lanes while RVV scans only the valid rows; once H=32 fills the tile, SFPU benefits from lane-parallel reduction. On Blackhole p150 at V=262144, RVV measured 48.2 us at H=1 and reached SFPU best time with 8 cores instead of 111; at H=32, SFPU measured 203.8 us versus 215.2 us for RVV. These are RVV-versus-SFPU measurements, not a speedup claim against the incumbent path. The implementation passed 468 Blackhole tests, and the complete benchmark is reproducible with `tests/ttnn/unit_tests/operations/reduce/test_argmax_path_crossover_bench.py`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/argmax-rvv-tile)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/argmax-rvv-tile)
### Measurements

Blackhole p150, 13x10 = 130-core grid, `[1, 1, H, V]` BFLOAT16 TILE input, last-dim reduction. Per-op **device** time from trace replay of a throughput-mode capture.

> **Throughput-mode timing, not single-op latency.** Per-op host dispatch is amortized away by construction; an isolated eager caller pays **20-90 us more** than every number below. These decide which path costs the *device* less; they are not end-to-end latencies. One board, min-of-N, no error bars. The sweep covers only single-tile-row shapes (H = 1 / 8 / 32 all have `h_tiles = 1`).

Regenerate with:

```
TTNN_RUN_ARGMAX_CROSSOVER_BENCH=1 pytest tests/ttnn/unit_tests/operations/reduce/test_argmax_path_crossover_bench.py -s
```

#### RVV vs SFPU at equal core count

Pinned with `sub_core_grids`, because comparing at each path's *default* compares two core-count heuristics rather than two implementations. Values are SFPU time / RVV time, so **> 1 means RVV is faster**:

| V | H | 1 core | 8 cores | 32 cores | 64 cores | 111 cores |
|---:|---:|---:|---:|---:|---:|---:|
| 32768 | 1 | 12.37x | 5.88x | 1.88x | 1.00x | 1.00x |
| 32768 | 8 | 3.14x | 1.94x | 1.57x | 1.21x | 1.00x |
| 32768 | 32 | 0.87x | 0.76x | 1.06x | 1.17x | 1.20x |
| 262144 | 1 | 13.92x | 7.60x | 3.33x | 2.34x | 1.22x |
| 262144 | 8 | 3.75x | 3.18x | 2.06x | 1.57x | 1.34x |
| 262144 | 32 | 0.95x | 0.86x | 0.81x | 0.85x | 0.94x |

H is the discriminator, not V. The SFPU pass reduces all 32 rows of a tile-row lane-parallel, so it is nearly flat in H — at 1 core, V=262144, it measures 4875 / 4881 / 4911 us at H = 1 / 8 / 32. The RVV scan is per row: 350 / 1302 / 5191 us at the same points. At small H the SFPU pays for 32 lanes to serve a handful of real rows.

#### Cores-to-match

The fewest cores at which one path reaches the other's best time anywhere on this grid:

| V | H | RVV needs | SFPU's best needs | |
|---:|---:|---:|---:|---|
| 262144 | 1 | **8 cores** | 111 cores | 14x fewer |
| 262144 | 8 | 32 cores | 111 cores | 3x fewer |
| 32768 | 1 | 4 cores | 32 cores | 8x fewer |
| 32768 | 8 | 16 cores | 64 cores | 4x fewer |
| 262144 | 32 | 111 cores | 64 cores | SFPU: 2x fewer |

At H = 1 and H = 8, the SFPU never reaches RVV's best time at any core count on this grid.

#### Why the ratios decay toward 1

Both paths converge on a shared per-program dispatch floor of ~0.44 us per core. Past each curve's knee, adding cores makes **both** slower: RVV at V=32768 H=1 costs 12.1 us on 16 cores and 50.6 us on 111, where the four V=32768, H in {1,8} points all land between 50.6 and 50.9 us regardless of workload. A comparison read only at the top of the grid measures that floor, not the paths.

#### Core-count heuristics

`kSfpuMinRows = 32`: against the multicore RVV path, H = 8 is not an SFPU win at any core count in the sweep. The boundary is mildly V-dependent at H = 32 — at each path's own default core count, V=32768 H=32 is a narrow RVV win (84.6 us on 63 cores vs 87.0 on 40) while V=262144 H=32 is an SFPU win (215.2 on 130 vs 203.8 on 111).

The RVV path uses `ceil(sqrt(w_tiles * (h_logical + 2)) / 3)`, capped by the grid and `w_tiles`. The SFPU formula is flat in H and wrong for a per-row scan. **The constants are fitted** by grid search over the sweep, not derived from a cost model; the header of `device/argmax_rvv_tile_program_factory.cpp` records the fit and the closed form it deliberately does not use. Every swept shape lands within 10% of its own optimum:

| V | H | heuristic cores | heuristic us | best cores | best us | ratio |
|---:|---:|---:|---:|---:|---:|---:|
| 4096 | 1 | 7 | 4.4 | 6 | 4.4 | 1.00x |
| 4096 | 8 | 12 | 11.9 | 14 | 11.5 | 1.03x |
| 4096 | 32 | 22 | 39.7 | 16 | 38.6 | 1.03x |
| 8192 | 1 | 10 | 6.1 | 9 | 5.5 | **1.10x** |
| 8192 | 8 | 17 | 14.6 | 21 | 14.3 | 1.02x |
| 8192 | 32 | 32 | 49.7 | 26 | 47.7 | 1.04x |
| 32768 | 1 | 19 | 11.6 | 22 | 11.0 | 1.06x |
| 32768 | 8 | 34 | 25.5 | 39 | 25.0 | 1.02x |
| 32768 | 32 | 63 | 84.8 | 42 | 78.3 | 1.08x |
| 131072 | 1 | 37 | 26.9 | 38 | 25.3 | 1.06x |
| 131072 | 8 | 68 | 56.7 | 86 | 55.1 | 1.03x |
| 131072 | 32 | 125 | 158.4 | 98 | 150.2 | 1.05x |
| 262144 | 1 | 53 | 48.1 | 26 | 46.4 | 1.04x |
| 262144 | 8 | 96 | 84.3 | 38 | 79.0 | 1.07x |
| 262144 | 32 | 130 | 215.7 | 98 | 211.1 | 1.02x |

Trace-replay timings, core counts swept 1-40 by 1 and 42-130 by 4, per-shape optimum taken over that sweep. Worst case is V = 8192, H = 1 at 1.10x; the H = 32 rows are routed to the SFPU anyway. Six of these points overlap the equal-core table above and agree to within 0.2% (11.6/11.5, 25.5/25.5, 84.8/84.8, 48.1/48.2, 84.3/84.4, 215.7/215.6).. The SFPU heuristic is unchanged, already within 0.87-1.04x of its own optimum.

No measurement of the scalar readers was taken, so **this PR makes no performance claim against them.** Every number above is RVV vs SFPU.

### Verification

- **468 tests green on Blackhole silicon** — `test_argmax_rvv.py`, `test_argmax_sfpu.py`, `test_argmax.py`.
- **Routing is asserted, not assumed.** Since no argument names a path, the routing tests read which path ran from `device.num_program_cache_entries()`: warm the expected path through its forced entry, then call `ttnn.argmax` and require the count *not* to grow. A mis-route compiles a second program and fails. Both suites pin the threshold against the C++ constant, so moving it in either direction fails a test.
- **Trace safety verified empirically**: 60 replays per shape with the input alternating between two payloads with *different* goldens, so a stale replay or an un-reset semaphore returns the other payload's answer rather than a benign repeat. Shapes stress the cross-core handshake: ragged slices (33 w_tiles over 7 cores), multi-pass (H = 80), both at once, and the widest single pass (8192 w_tiles over 111 cores). 4 shapes x 2 paths = **8/8 pass**.
- **Bit-exactness of the multicore merge is pinned by a negative control**: substituting an IEEE comparator into the RVV merge fails 59 tests, at the predicted index.
- **ttsim**: the RVV suite is skipped there as a capability gap (`UnsupportedFunctionality: rv32_v_alu`, whose error path is `_Exit(1)` and kills the xdist worker); the SFPU suite passes 175/175 and is skipped only for cost. Because this PR changes routing, `test_argmax.py` also reaches the RVV kernel under ttsim and carries a narrow simulator guard following the `skip_routed_topk_on_sim` precedent — it fires on 1 of 63 cases and not on hardware.


nkapreTT:
Thanks for the thorough automated passes — every thread is addressed below. Fixes landed in the four commits now on this branch (head `16ca50ec6e3`); the full 103-test silicon battery (grown from 55 by the additions below) re-ran green on Blackhole after the changes.

**Fixed:**
- **maxval device-affinity** (Copilot 🔴): `maxval_tensor` now validated as a device tensor on the *same* device as the input, alongside the existing dtype/layout checks.
- **Preallocated index-output shape** (Copilot 🔴): the RVV path now `TT_FATAL`s unless `optional_output_tensor.logical_shape()` equals the reduction output shape (got/expected in the message).
- **Rank-0 / scalar contract** (Copilot 🔴 + the guard-placement comment): the maxval-requires-`use_rvv` guard is hoisted above the early returns (path-independent now), and `use_rvv=True` requires rank ≥ 1 with nonzero volume — a scalar call fails loudly instead of silently no-oping.
- **`sub_core_grids`** (Copilot 🟡): the RVV factory now `TT_FATAL`s if a supplied grid excludes core (0,0) (honoring arbitrary grids is a named follow-up; silent violation is gone).
- **Boundary vocab sizes + planted-case cross-check** ([/tdd]): parametrization now includes v=32 (single tile, `chunk_pages` boundary) and v=2016 (63 tiles, remainder branch), and the upstream cross-check runs the full planted-max/tie/all-neginf/NaN CASES table, not just random data. 103 tests collect; all pass on silicon.
- **Docstring naming** ([/grill-with-docs]): defined once — "TRISC2 (the pack-thread RISC-V core) carries a Zve32f vector unit ('RVV')" — and used consistently; tile-width made explicit.
- **`amx_out_push` contract** ([/codebase-design]): the last-word-address computation moved *inside* the helper (`page_addr + ((bytes_written−1) & ~3)`), provably bit-identical to both prior callsites; contract documented.
- **Accessor placeholder** ([/codebase-design]): the third `TensorAccessorArgs` slot's contract is now fully documented (reader parses exactly three blocks at compile time; the placeholder is CT-guarded by `has_maxval` and never dereferenced).

**Respectfully refuted (with proof in-source):**
- **The `units==1` fence claim** ([/tdd]): the fence targets the aligned 32-bit word *containing* the last written bf16 halfword, which `(2·units−2) & ~3` computes correctly for every `units` ∈ [1,32] — for `units==1`, 2 bytes are written, so word 0 *is* the last word. Two lanes verified this independently; note the suggested refactor (`page_bytes − 4`) would fence one word *past* the written data for even `units`. A proof comment now lives at the helper.
- **`expect_error` undefined** ([/tdd]): it's the repo-root `conftest.py` fixture (line ~934) that the repo's own pre-commit hook mandates over `pytest.raises`; resolution verified via `--fixtures-per-test`, with two sibling files in this directory already using it.
- **Reader runtime-arg "duplication"** ([/codebase-design]): compile-time `TensorAccessorArgs` encode banking/layout only — the buffer base address is inherently a runtime value (compile-time addresses would defeat the program cache). Identical pattern to the existing `reader_argmax_tile_layout*.cpp` kernels; documented at the callsite.

**Stale:** the Cycode comment references `test_trisc2_rvv.cpp`, which left this PR's diff when base #54121 merged to main.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,165 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
gajanan-choudhary:
```suggestion
# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
```

aczajkowskiTT (approved):

iwroszTT:
Could SFPU be used instead of RV?

iwroszTT (changes requested):
Take a look at argmax compute kernel that is used on the nc path:
ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/argmax_nc_compute.cpp

Please try to generalize this kernel, or write new kernels and use SFPU instead of RV.

nkapreTT:
Took your suggestion seriously and did it properly: **we built the SFPU last-dim kernel and raced it.** Design mirrors `argmax_nc_compute.cpp`'s per-lane recipe (5 DST-wide ops/tile running `(max_val, win_tile)` across the Wt tiles, lexicographic finish for the cross-lane step). Measured on BH silicon, same shapes:

| shape | stock TILE path | SFPU (built for this) | RVV (this PR) |
|---|---|---|---|
| V=262k, **B=1** (decode) | 9,128 µs | ~4,868 µs (~802 cyc/tile, flat in B/V) | **462 µs** |
| V=262k, **B=32** | 155,808 µs | **~4,895 µs — 807 cyc/tile, beats RVV** | 6,965 µs (1,148 cyc/tile) |

The honest split, per the numbers:

- **Batch shapes: you're right.** At full tiles (B=32) the per-lane SFPU recipe wins even single-core (807 vs 1,148 cyc/tile), and phase 1 is embarrassingly parallel over tile-rows, so a multicore version extends it. Happy to land that as a follow-up — the kernel exists and passes the case battery.
- **Decode (B=1, the LM-head shape this PR targets): SFPU pays all 32 lanes for one valid row** — measured ~802 cyc/tile ≈ 25 cyc per *useful* element, roughly the scalar loop's rate — while the RVV scan runs 63–100 cyc/tile (10.5× faster than the SFPU variant here, 19.8× vs stock).
- **The fused epilogue needs the vector RISC regardless**: #54106 rides the LM-head matmul's pack shadow at +0.9% wall precisely because the scan touches neither DST (the matmul owns it; gating the pack queue measured +24%) nor SFPU state (pack-issued SFPU in fused programs hits the shared-state races). An SFPU argmax is standalone-only.
- **Semantics footnote**: the SFPU kernel inherits IEEE-compare behavior (NaN-as-inf, ±0 merge, FTZ) — bit-exact with the incumbent on finite/normal data but diverging on NaN/denormal corners, exactly as the NC path already does today; the RVV kernel is the one vectorized variant that reproduces the incumbent's bit-total-order exactly (ties, −0, NaN payloads included).

Proposal: keep this PR as-is for the decode/fusion case, and I'll file the SFPU batch-shape kernel as a follow-up so argmax dispatches by shape — RVV for B=1/TILE-fused, SFPU multicore for batch. Both beat the current 9 ms by an order of magnitude. Raw kernels/logs for the comparison are preserved and can be shared.

nkapreTT:
Integrated the SFPU companion and refreshed the stack for review:

- rebased the six existing commits onto current `main` with patch-identical history;
- retained RVV for batch-1 decode and added multicore SFPU for batch shapes, following the review direction;
- added preallocated index-output device/mesh affinity validation;
- validated all supplied `sub_core_grids` against the device compute grid;
- added negative out-of-grid and positive explicit nonzero-core SFPU coverage;
- changed-file pre-commit suite is clean.

Exact combined head: `d262c9ac23c31b67afb980f8aa200bb5f2675a31`.

A dedicated Blackhole P150 run is exercising both complete suites on this exact SHA with `TT_ENABLE_RVV_TESTS=1`: https://github.com/tenstorrent/tt-metal/actions/runs/33079531627. I will post the final result before re-requesting review.

nkapreTT:
Exact-head Blackhole validation is green on `d262c9ac23c31b67afb980f8aa200bb5f2675a31`.

- Dedicated P150b run: https://github.com/tenstorrent/tt-metal/actions/runs/33079531627
- Command: `TT_ENABLE_RVV_TESTS=1 pytest --timeout 300 -v --tb=short tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py tests/ttnn/unit_tests/operations/reduce/test_argmax_sfpu.py`
- Result: **272 passed** in 136.77s
- Release build and Python wheel also passed in the same exact-commit run.

The rebased series is patch-equivalent to the prior six commits, plus validation hardening for optional-output device affinity and accelerated `sub_core_grids` containment, with positive/negative Blackhole coverage.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_rvv_tile.cpp
```diff
@@ -0,0 +1,121 @@
...
+                cb_reserve_back(cb_in, chunk);
+                for (uint32_t k = 0; k < chunk; k++) {
+                    const uint32_t slot = (t_global + k) % in_cb_pages;
+                    noc_async_read_tile(tile_row_first + done + k, s_src, in_base + slot * src_page_size);
+                    if ((k & 31u) == 31u) {
+                        noc_async_read_barrier();
+                    }
+                }
+                noc_async_read_barrier();
+                cb_push_back(cb_in, chunk);
```
bbradelTT: Please use device 1.1 / 2.0 APIs

nkapreTT:
Addressed the remaining Copilot coverage request in `e2e9ab8c588`: the RVV and SFPU argmax suites are now scheduled automatically on both `bh_p100` and `bh_p150b_civ2` with `TT_ENABLE_RVV_TESTS=1`. The generated matrix is correct and the unit-test time-budget validator passes. All five Copilot threads are resolved; CI is running.

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
```diff
@@ -27,6 +27,29 @@ void bind_reduction_argmax_operation(nb::module_& mod) {
                 sub_core_grids (CoreRangeSet, optional): Limits execution to a subset of cores. Supported on ROW_MAJOR last-dim reductions (<= 2 ranges) and batch/channel dim reductions. Default: ``None``.
                 memory_config (ttnn.MemoryConfig, optional): Output memory (INTERLEAVED DRAM/L1). Default: input's memory_config.
                 output_tensor (ttnn.Tensor, optional): Preallocated output (must be UINT32, ROW_MAJOR, INTERLEAVED, same device). Default: ``None``.
+                use_rvv (bool, optional): Opt-in Blackhole-only path. TRISC2 (the pack-thread RISC-V core) carries
```
iwroszTT: Would that be possible to not expose the `use_rvv`, `use_sfpu`, but decide inside the operation based on a performance heuristic/model?

tests/pipeline_reorg/ops_unit_tests.yaml
```diff
@@ -77,6 +77,17 @@
   team: ttnn
   category: reduction
 
+- name: ttnn Blackhole accelerated argmax tests
+  cmd: 'TT_ENABLE_RVV_TESTS=1 pytest tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py tests/ttnn/unit_tests/operations/reduce/test_argmax_sfpu.py -xv --timeout=300'
```
iwroszTT: Why is the flag needed?

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,165 @@
...
+import torch
+import ttnn
+
+from models.common.utility_functions import is_blackhole
+
+pytestmark = [
+    pytest.mark.skipif(not is_blackhole(), reason="ttnn.argmax use_rvv=True is Blackhole-only (TRISC2 Zve32f)"),
+    pytest.mark.skipif(
+        os.environ.get("TT_ENABLE_RVV_TESTS") != "1",
+        reason="RVV device-test gate: set TT_ENABLE_RVV_TESTS=1 to run the RVV argmax tests on a Blackhole device",
```
iwroszTT: Why not just keep the tests always enabled on Blackhole?

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
```diff
@@ -27,6 +27,29 @@ void bind_reduction_argmax_operation(nb::module_& mod) {
...
+                    special-value gasket rather than the scalar readers' bit-pattern total order. Concretely:
+                    NaN behaves as same-signed infinity (a NaN row-max yields the NaN's index but max value +inf
+                    ``0x7F80``, not the NaN payload; -NaN never wins); -0 flushes to +0 in the max-value output and
+                    +0/-0 compare equal (first zero's index is kept); denormals flush to zero before the compare;
+                    max values below ~2^-118 carry a +2^-127 additive pack bias. All finite normal data — including
+                    every exact tie — matches the incumbent bit-for-bit (smallest index wins ties).
+                    Default: ``False``.
+                maxval_tensor (ttnn.Tensor, optional): Preallocated BFLOAT16 ROW_MAJOR tensor with the same logical
+                    shape as the index output. Only with ``use_rvv=True`` or ``use_sfpu=True``: receives the winning
+                    max VALUES, so greedy sampling does not need a separate ``ttnn.max``. Default: ``None``.
```
iwroszTT: What is greedy sampling?

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
nkapreTT: Done. Both flags are gone from the C++ signature and the Python binding; there is no argument that names an engine. Passing either is now a `TypeError` (asserted by `test_argmax_rejects_removed_engine_flags`).

`select_argmax_engine` in `argmax.cpp` is the single place that decides, from the input spec alone. `ArgMaxEngine` lives in `ArgmaxParams`, so the program cache keys on it — two calls differing only in engine cannot share a cached program.

The gate is two parts. `accelerated_engines_can_serve` is the correctness precondition, mirroring the per-engine `TT_FATAL`s in `validate_on_program_cache_miss`: Blackhole, TILE/BFLOAT16, explicit dim normalizing to `rank - 1`, rank >= 1 with non-zero volume, 32x32 tiles, last dim a multiple of 32, INTERLEAVED in and out, and (if supplied) an `output_tensor` whose logical shape is the reduction output shape. Anything else returns `Incumbent`, so nothing that worked before stops working.

Then the preference is purely `H`, the rows per tile-row: `H >= kSfpuMinRows` (32) goes to SFPU, everything else to RVV. SFPU is additionally excluded for `rank == 1` (it is `H == 1` by construction, where SFPU is the measured loser by 12-14x, and nothing has ever run rank 1 on those kernels) and whenever `exact_special_values` is set.

`exact_special_values=False` is the one new public argument. It constrains selection to engines bit-identical to the scalar readers on every input — today the readers and RVV, excluding SFPU, whose compare is IEEE-on-fp32 behind a bf16 gasket. Setting it can cost throughput, never correctness.

**Two things I would rather put in front of you than have you find.**

**1. A new non-public surface.** Removing the flags removed the only way to test each engine's distinct claim, and the suites also need a stable incumbent golden — on Blackhole a plain `ttnn.argmax` over an eligible TILE bf16 last dim no longer runs the scalar readers, so "compare against the incumbent" needs a way to say so. Hence three verification-only entries, `ttnn._ttnn.operations.reduction.argmax_force_{incumbent,rvv,sfpu}`. They are bound with a plain `mod.def` rather than `ttnn::bind_function`, specifically so `auto_register_ttnn_cpp_operations` does not republish them as `ttnn.*` ops; they sit in `detail`, `argmax_force.hpp` is out of the installed api set, and none falls back — a forced engine that quietly served a different one would make the comparison vacuous. Same shape as `repeat_force.hpp` in `data_movement/repeat`. If you would rather this were a pytest-only hook than a bound symbol, say so and I will move it. I have no attachment to the mechanism, only to the engines staying individually testable.

**2. Auto-dispatch changes which engine some existing callers get** — only for TILE + BFLOAT16 + last-dim + Blackhole. The production greedy-decode callers are unaffected: both `models/common/sampling/tt_sampling.py:883` and `models/common/modules/sampling/sampling_1d.py:320` call `ttnn.untilize(...)` immediately before argmax, so they arrive ROW_MAJOR and stay on the incumbent. What changes is experimental models passing TILE directly, plus the TILE cases in `test_argmax.py`. Those diverge from the incumbent only on NaN / denormal / +-0, which random test data does not produce; `test_argmax.py` needed no change for TILE and is green.

tests/pipeline_reorg/ops_unit_tests.yaml
nkapreTT: It is not. `TT_ENABLE_RVV_TESTS` is deleted repo-wide — zero occurrences left — and this `ops_unit_tests.yaml` entry is removed entirely, not just de-flagged.

The entry existed only to supply that env var. With the flag gone, keeping it would have been actively worse:

- **Double-run.** The "ttnn reduce group" in `ttnn_sanity_tests.yaml:249` is `pytest ... tests/ttnn/unit_tests/operations/reduce`, which already globs both new suites. A dedicated entry runs them a second time.
- **Budget.** `.github/time_budget.yaml` gives ttnn's `unit` pipeline 1045 minutes on `bh_p150b_civ2`; `ops_unit_tests.yaml` books 1040. The 5-minute entry would have consumed exactly the remaining headroom on that SKU, for coverage already running elsewhere.

Gating is now `run_for_blackhole(...)` as a module-level `pytestmark` in both suites — see the thread on `test_argmax_rvv.py`.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
nkapreTT: Agreed, and that is what they do now. `TT_ENABLE_RVV_TESTS` is gone; both suites gate on architecture only:

```python
from models.common.utility_functions import run_for_blackhole
pytestmark = run_for_blackhole("the RVV argmax engine is Blackhole-only (TRISC2 Zve32f)")
```

Same in `test_argmax_sfpu.py`. There was never a toolchain reason for the flag — the RVV kernels JIT with the in-tree opt-in (`ComputeConfigDescriptor::enable_trisc2_rvv`, which adds `zve32f` to the TRISC2 compile), so no host setup is required. They now run automatically on any Blackhole host and skip everywhere else.

One deliberate exception: both files are added to the `blackhole:` section of `tests/pipeline_reorg/ttsim-skip-list.yaml`, with a comment saying why. `is_blackhole()` is true under ttsim, so `run_for_blackhole` would let them run there, and the accelerated engines may not be modelled. `test_argmax.py` is already skipped in that section for its own reasons. If ttsim does model Zve32f I would rather delete those two lines than keep a stale skip — tell me if you know.

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
nkapreTT: Removed. "Greedy sampling" no longer appears anywhere in the argmax tree — docstring or comments.

`maxval_tensor` is now described in operation terms only:

> Preallocated BFLOAT16 ROW_MAJOR tensor with the same logical shape as the index output. Honored only on the accelerated engines: alongside each winning index, it receives the maximum VALUE found at that index, so the caller does not need a second pass over the input with `ttnn.max` to recover the value it has just located.

Two other unexplained shorthands went with it, since they were the same failure: "NC-dim" is now "batch/channel dim" throughout, and "the scalar readers" is introduced before first use rather than assumed.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_rvv_tile.cpp
nkapreTT: Both new readers are migrated. You were right that this was not cosmetic — `noc_async_read_tile` is `[[deprecated]]` in `tt_metal/hw/inc/api/dataflow/dataflow_api.h` ("Use `<typename AddrGen> noc_async_read_page` instead"), on all three overloads.

`reader_argmax_rvv_tile.cpp` and `reader_argmax_sfpu_tile.cpp` now use `Noc` / `CircularBuffer` / `CoreLocalMem` / `Semaphore` / `UnicastEndpoint`, idiom copied from the already-migrated `reader_masked_bincount.cpp`. Separately, `argmax_sfpu_tile_compute.cpp` dropped the now-deprecated `init_sfpu` for `compute_kernel_hw_startup(cb_in, cb_res_val)` + `copy_init(cb_in)`, matching the migration main already applied to `argmax_nc_compute.cpp`.

Four behaviours survived the rewrite unchanged. Calling them out because each is easy to mistake for dead code on review:

| | why it stays |
|---|---|
| `if ((k & 31u) == 31u) noc.async_read_barrier();` inside the chunk loop | bounds outstanding NoC reads within a chunk; not redundant with the trailing barrier, which covers the final `< 32` reads |
| the input CB base captured once into a `CoreLocalMem` | the CB is a ring addressed by *global* page number; its write pointer advances on every `push_back`, so it cannot be the NoC destination |
| `reserve_back` / `push_back` / `wait_front` / `pop_front` pairing | unchanged, including the compute-side global page counter that must agree with `t_global` |
| `start_sem.wait_min(pass)` before the exchange write, `done_sem.wait_min((pass+1)*(num_cores-1))` on the gather core, credits returned only when another pass needs them, both semaphores restored to 0 at kernel end | cumulative counting with no mid-run resets, so trace replay — which does not re-run the dispatcher's semaphore init — starts clean; the credit-return guard keeps a late increment from racing the workers' final reset |

Trace safety is checked directly rather than assumed: 8/8 cases (4 shapes x 2 engines, including ragged slices and multi-pass `H > 32`) at 60 replays each, with the input alternating between two payloads so a stale or un-reset replay returns the *other* payload's golden rather than a benign repeat.

nkapreTT:
All five review threads are answered inline and the description is rewritten. History is consolidated to four commits — the two mechanical ones you asked for first (Device 2.0 kernel migration, architecture-gated tests), then engine selection, then the multicore work. **CI is green. 468 tests pass on Blackhole silicon.**

This PR now contains work beyond the four change-requested fixes. Surfacing that explicitly rather than letting it turn up in the diff.

**1. A rank-1 regression the auto-dispatch introduced, and its fix.** Removing the flags meant selection had to decide rank 1, and my first cut demoted it to the incumbent. That broke `argmax(rank1_tile, dim=-1, maxval_tensor=...)`, which RVV had served all along — the scalar readers cannot produce a max-value output, so the call went from working to a hard error. Rank 1 is not a gap in the kernels: both accelerated factories already fold it to `h_logical == 1`, so a `[V]` input produces exactly the compile-time args of `[1,1,1,V]`. It routes to RVV now, with a regression test and a routing test. Rank **0** stays excluded as a correctness gate, not a preference: `logical_volume() == 1`, `normalize_dim(-1, 0) == -1` compares equal to `rank - 1`, and `logical_shape[-1]` would index out of bounds.

**2. Multicore RVV, plus a measurement-derived dispatch policy.** RVV was pinned to core (0,0). It now splits the reduction dim across cores and merges per-core candidates on a gather core — in `bfloat16_greater`'s bit-pattern order with a smallest-global-index tie-break, **not** an IEEE compare, so the result is bit-identical however many cores it runs on. That invalidated the shipped policy in two places, both corrected against measurement:

- `kSfpuMinRows` 8 -> 32. The old boundary came from a table that compared **single-core RVV against multicore SFPU**: its RVV column (192 us at V=32768 H=8) reproduces today's *one-core* RVV number (198 us), not the multicore one (26 us), while its SFPU column was multicore and still reproduces. Against multicore RVV, H=8 is not an SFPU win at any core count in the sweep.
- The RVV factory no longer copies the SFPU core-count formula. SFPU cost is flat in H so `ceil(sqrt(1.5 * w_tiles))` fits it; the RVV scan is per row, so it fits `ceil(sqrt(w_tiles * (H + 2)) / 3)`. Using the SFPU formula cost up to 1.85x at H=1; worst case is now 1.08x off the per-shape optimum.

**3. A regenerable benchmark**, `_argmax_engine_crossover_bench.py`. The leading underscore keeps it out of pytest collection so CI never runs it; run explicitly, it prints the markdown block that every number below and in the source comments comes from.

**Why this is one PR rather than several.** The pieces are causally chained: making RVV multicore is what invalidated `kSfpuMinRows = 8`, which forced the re-measurement, which set the new boundary, which is what makes internal engine selection defensible without public flags. Splitting would mean landing a dispatch boundary I have measured to be wrong.

**Headline.** For a fused argmax epilogue the number that matters is not latency at the full grid — the LM-head matmul owns the grid — it is how few cores an engine needs. At V=262144 H=1, RVV reaches SFPU's best on **8 cores**; SFPU needs **111** for that result. At H=1 and H=8, SFPU never reaches RVV's best at any core count in the sweep.

SFPU time / RVV time at equal core count (> 1 means RVV faster):

| shape | 1 core | 8 | 32 | 64 | 111 |
|---|---|---|---|---|---|
| V=32768 H=1 | 12.37x | 5.88x | 1.88x | 1.00x | 1.00x |
| V=32768 H=8 | 3.14x | 1.94x | 1.57x | 1.21x | 1.00x |
| V=32768 H=32 | 0.87x | 0.76x | 1.06x | 1.17x | 1.20x |
| V=262144 H=1 | 13.92x | 7.60x | 3.33x | 2.34x | 1.22x |
| V=262144 H=8 | 3.75x | 3.18x | 2.06x | 1.57x | 1.34x |
| V=262144 H=32 | 0.95x | 0.86x | 0.81x | 0.85x | 0.94x |

The decay toward 1 is not the engines converging in capability — both hit a shared ~0.44 us/core per-program dispatch floor, which is also why both scale *negatively* past their knee (RVV at V=32768 H=1: 12.1 us on 16 cores, 52.6 us on 111). Reading the comparison only at the top of the grid measures that floor.

**Caveats, stated once, applying to every number above:** one p150, min-of-N, no error bars. Throughput-mode device time (N ops in one trace, replayed, wall/N) — **not** single-op latency; an isolated eager caller pays 20-90 us more, which the benchmark reports separately. The sweep covers only single-tile-row shapes. No incumbent measurement was taken, so I make **no** performance claim about the scalar readers.

**Correctness evidence.** Bit-exactness is not asserted from inspection — it is pinned by a negative control: substituting the IEEE comparator for the bit-pattern one fails 59 tests, at the predicted index. Trace safety is 8/8 at 60 replays per shape with alternating payloads, covering ragged slices and multi-pass.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,156 @@
...
+    """Index cross-check against the incumbent ttnn.argmax on the same TILE tensor."""
+    torch.manual_seed(v + b)
+    x = torch.randn(1, 1, b, v).bfloat16()
+    t_tile = ttnn.from_torch(x, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    idx_rvv = ttnn.to_torch(ttnn.argmax(t_tile, dim=3, keepdim=True, use_rvv=True))
+    idx_ref = ttnn.to_torch(ttnn.argmax(t_tile, dim=3, keepdim=True))
+    assert torch.equal(idx_rvv.to(torch.int64), idx_ref.to(torch.int64))
+
+
```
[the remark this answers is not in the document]
nkapreTT: Addressed on the current head. The incumbent-path cross-check is parameterized over the complete CASES set and exercises planted maxima, ties, all-negative infinity, and NaN cases at boundary and normal widths.

ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_rvv_tile_program_factory.cpp
```diff
@@ -0,0 +1,182 @@
...
+        reader_desc.emplace_runtime_args(core, args);
+    }
+    desc.kernels.push_back(std::move(reader_desc));
+
+    // Compute kernel: unpack/math no-op, pack thread runs the RVV scan.
+    KernelDescriptor compute_desc;
+    compute_desc.kernel_source = "ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/argmax_rvv_tile_compute.cpp";
+    compute_desc.source_type = KernelDescriptor::SourceType::FILE_PATH;
+    compute_desc.core_ranges = all_cores;
+    compute_desc.compile_time_args = {
```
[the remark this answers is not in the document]
nkapreTT: Addressed on the current head. The placeholder now has a detailed contract comment explaining the three unconditional TensorAccessorArgs blocks, the compile-time offset chain, and the has_maxval guard.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/argmax_rvv_tile_compute.cpp
```diff
@@ -0,0 +1,295 @@
...
+
+            uint32_t tiles_done = 0;
+            while (tiles_done < w_tiles) {
+                const uint32_t chunk = (w_tiles - tiles_done < chunk_pages) ? (w_tiles - tiles_done) : chunk_pages;
+                amx_in_wait(cb_in, acked_in, (uint16_t)chunk);
+                invalidate_l1_cache();
+                for (uint32_t g = 0; g < n_groups; g++) {
+                    const uint32_t rows_in_group = (units - 2 * g < 2u) ? (units - 2 * g) : 2u;
+                    amx_process_chunk_group(cb_in, t_global, chunk, g, rows_in_group, tiles_done);
+                }
```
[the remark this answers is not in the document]
nkapreTT: Addressed on the current head. amx_out_push now accepts page_addr plus bytes_written and computes the aligned word containing the final written byte internally.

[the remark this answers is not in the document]
nkapreTT: Verified and documented on the current head. For two bytes written, aligned word zero is the word containing the final bf16 halfword. The helper computes (bytes_written - 1) rounded down to the containing uint32 word and owns this calculation internally.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_rvv_tile.cpp
```diff
@@ -0,0 +1,117 @@
...
+    const uint32_t dst_base_addr = get_arg_val<uint32_t>(1);
+    const uint32_t val_base_addr = has_maxval ? get_arg_val<uint32_t>(2) : 0;
+
+    constexpr auto s_src_args = TensorAccessorArgs<num_c_time_args>();
+    constexpr auto s_dst_args = TensorAccessorArgs<s_src_args.next_compile_time_args_offset()>();
+    const auto s_src = TensorAccessor(s_src_args, src_base_addr, src_page_size);
+    const auto s_dst = TensorAccessor(s_dst_args, dst_base_addr, dst_page_size);
+
+    // The maxval accessor args are appended only when the tensor is present.
+    constexpr auto s_val_args = TensorAccessorArgs<s_dst_args.next_compile_time_args_offset()>();
```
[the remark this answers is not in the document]
nkapreTT: Verified and documented on the current head. TensorAccessorArgs contain compile-time banking and layout, while allocation base addresses remain runtime arguments so program-cache rebinding works. The reader now states this contract directly.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,156 @@
...
+@pytest.mark.parametrize("v", [2048, 8192])
+@pytest.mark.parametrize("b", [1, 5, 32])
+@pytest.mark.parametrize("keepdim", [True, False])
+@pytest.mark.parametrize("with_maxval", [True, False])
+def test_argmax_rvv_battery(device, v, b, keepdim, with_maxval):
+    """Bit-exact index (and optional max-value bits) against the incumbent
+    semantics across planted-max / tie / special-value cases."""
+    rng = np.random.default_rng(1234 + v + 100 * b)
+    for name in CASES:
+        bits = _make_case(name, v, b, rng)
```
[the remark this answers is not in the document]
nkapreTT: Addressed on the current head. Coverage now includes v=32, v=2016, v=2048, and v=8192, including single-tile and ragged final-chunk paths.

nkapreTT:
@iwroszTT The requested SFPU/automatic-dispatch direction is implemented, the follow-up comment and ttsim routing corrections are now at head `617c14061330`, all visible CI is green, and all review threads are answered/resolved. Could you please take a fresh look and clear the prior change-request state if this addresses your concerns?

tests/pipeline_reorg/ttsim-skip-list.yaml
```diff
@@ -40,12 +40,24 @@ blackhole:
...
   - tests/ttnn/unit_tests/operations/sdpa/test_mla_prefill.py
   - tests/ttnn/unit_tests/operations/sdpa/test_mla_prefill_v_embedding_space.py
 
+  # babyrisc V extension (TRISC2 Zve32f), which the RVV argmax kernel is built on:
+  #   ERROR: UnsupportedFunctionality: rv32_v_alu: babyrisc non-compliant V extension
+  #   is explicitly out of scope
+  # ttsim v1.10.3. That error path is _Exit(1), so this kills the xdist worker rather
+  # than failing a single test. run_for_blackhole() does not gate it: is_blackhole() is
+  # true under ttsim.
+  - tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```
fplavecTT: I don't understand what this comment is saying.
This comment (and the one above) should be useful to future readers, not narrate temporary state during the development of this PR.

tests/ttnn/unit_tests/operations/reduce/_argmax_engine_crossover_bench.py
```diff
@@ -0,0 +1,854 @@
...
+"""Trace-replay measurement of the Blackhole accelerated argmax engines (RVV, SFPU).
+
+WHAT THIS MEASURES
+------------------
+Per-op **device** time of ``ttnn.argmax``'s accelerated engines as a function of
+the core count they are given, so two questions can be answered from data:
+
+1. **How few cores does each engine need?** This, not absolute latency at the
+   whole grid, is the headline. The fused argmax epilogue rides the LM-head
+   matmul's pack shadow and cannot have the grid -- the matmul owns it. An
```
fplavecTT: "The fused argmax epilogue rides the LM-head matmul's pack shadow and cannot have the grid" ??

What does this mean?
Seems like this comment block needs a human touch.

tests/ttnn/unit_tests/operations/reduce/test_argmax_path_crossover_bench.py
fplavecTT, line 1: Why does this file name start with `_`? That doesn't seem to be repo convention.
This seems to be a standalone benchmark, which may better fit into `tests/ttnn/unit_tests/benchmarks/` (and not have a filename that starts with `_`)

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
```diff
@@ -115,14 +117,54 @@ def _argmax_nc_nd_rank5():
     ]
 
 
+# On Blackhole ttnn.argmax routes TILE / bfloat16 / last-dim / width-multiple-of-32 inputs
+# onto the accelerated engines, and of those the ones with H < 32 -- which includes every
```
fplavecTT: What are "accelerated engines"? That seems to be a term made up by this PR that is inconsistent with how the word engine is used anywhere else in the repo; in existing code and documentation, the word "engine" usually refers to a hardware block (vector engine, matrix engine,...), never an implementation choice.
I would recommend scrubbing this PR and replacing all such novel uses of the word engine with one of the existing terms: parallelization strategy, factory/variant, execution path,..., whichever may be the most appropriate in a given context.

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
```diff
@@ -115,14 +117,54 @@ def _argmax_nc_nd_rank5():
...
+# Zve32f vector unit (accelerated_engines_can_serve + select_argmax_engine in
+# ttnn/cpp/ttnn/operations/reduction/argmax/argmax.cpp). ttsim does not implement that unit:
+#
+#   ERROR: UnsupportedFunctionality: rv32_v_alu: babyrisc non-compliant V extension is
+#   explicitly out of scope
+#
+# and ttsim's error path is _Exit(1), so such a case kills the pytest/xdist worker rather
+# than failing. That is a property of where argmax routes on Blackhole, not of what these
+# cases test: they keep running on silicon and on every other architecture, and the TILE
+# cases that route to the SFPU engine or fall back to the scalar readers keep running here.
```
fplavecTT: This comment seems misplaced. Should be moved to `skip_rvv_routed_on_sim below`.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,533 @@
...
+        ranges.append(ttnn.CoreRange(c, c))
+    return ttnn.CoreRangeSet(ranges)
+
+
+def _single_core_grid():
+    return ttnn.CoreRangeSet([ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(0, 0))])
+
+
+def _make_case(name: str, v: int, b: int, rng: np.random.Generator) -> np.ndarray:
+    """Row-major [b, v] bf16 bit patterns for one battery case."""
```
fplavecTT: What's a battery?

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,533 @@
...
+    bits = _bits_of(torch.from_numpy(x).bfloat16()).reshape(b, v)
+    kmax, kdecoy = 0x7F7F, 0x7F7E  # largest finite bf16 + decoy the RNG cannot reach
+    if name == "random":
+        pass
+    elif name == "unique_max":
+        bits[:, 5 * v // 8] = kmax
+        bits[:, v // 3] = kdecoy
+    elif name == "tie_first_wins":
+        bits[:, 5 * v // 8] = kmax
+        bits[:, 7 * v // 8] = kmax
```
fplavecTT: All the various magic numbers could use either meaningful constant names or comments explaining what they mean/where they come from.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,533 @@
...
+# v boundaries: 32 = single tile (chunk_pages = min(64, w_tiles) = 1);
+# 2016 = 63 tiles, exercising the w_tiles - tiles_done < chunk_pages remainder
+# branch; 2048/8192 = exact multiples of the 64-tile chunk.
+@pytest.mark.parametrize("v", [32, 2016, 2048, 8192])
+@pytest.mark.parametrize("b", [1, 5, 32])
+@pytest.mark.parametrize("keepdim", [True, False])
+@pytest.mark.parametrize("with_maxval", [True, False])
+@pytest.mark.parametrize("grid", ["single", "auto"])
+def test_argmax_rvv_battery(device, v, b, keepdim, with_maxval, grid):
+    """Bit-exact index (and optional max-value bits) against the incumbent
```
fplavecTT: Comments should be written so that they are useful to future readers and not narrating one PR.
The word incumbent here (and in other places in this PR) is only meaningful in this PR's context and not useful to future readers.

tests/ttnn/unit_tests/operations/reduce/_argmax_engine_crossover_bench.py
```diff
@@ -0,0 +1,854 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
fplavecTT: Tenstorrent USA, Inc.

Please apply to all new files in this PR

iwroszTT (approved):

tests/pipeline_reorg/ttsim-skip-list.yaml
nkapreTT: Rewritten. There was a real defect: `# ttsim v1.10.3.` was orphaned mid-sentence, so the block read as two fragments rather than one attribution.

Both entries are now split by their actual reason, which differ:

- `test_argmax_rvv.py` is a **capability** skip and now sits in its own section rather than under `# Job-level timeouts`. ttsim rejects the Zve32f unit with `UnsupportedFunctionality: rv32_v_alu: babyrisc non-compliant V extension is explicitly out of scope`, and that error path is `_Exit(1)` — it kills the xdist worker rather than failing one test, which is why it has to be a file-level skip. The error string is kept verbatim so the next person can grep it.
- `test_argmax_sfpu.py` is the opposite: it passes 175/175 under the pinned ttsim. It stays under timeouts, and the CI stopwatch arithmetic is gone since it would be stale within a month — it now just says the suite is too slow for `sim_bh_p150`'s budget in `.github/time_budget.yaml`.

tests/ttnn/unit_tests/operations/reduce/_argmax_engine_crossover_bench.py
nkapreTT: Rewritten, and the sentence you quoted was also garbled — "cannot have the grid" should have been "cannot take the grid".

It is now stated once, in the module docstring, with the other two sites cross-referencing it instead of repeating it:

> A fused argmax epilogue would share the compute grid with the LM-head matmul it follows, so it can only ever be handed a slice of that grid rather than all of it. What decides whether such a fusion is possible is therefore how few cores a path needs to reach a given latency, not the best time it can post with the whole grid to itself.

Kept rather than deleted because it is why the benchmark reports cores-to-match rather than peak latency, but it should not have taken three unexplained metaphors to say so.

tests/ttnn/unit_tests/operations/reduce/test_argmax_path_crossover_bench.py
nkapreTT: Moved to `tests/ttnn/unit_tests/benchmarks/test_argmax_path_crossover_bench.py`, without the underscore.

One thing I had to add to make the move safe: that directory **is** collected by CI. `tests/pipeline_reorg/ttnn_sanity_tests.yaml:270` runs

```
pytest --timeout 300 tests/ttnn/unit_tests/base_functionality tests/ttnn/unit_tests/benchmarks tests/ttnn/unit_tests/tensor -xv
```

and this benchmark's own docstring asks for `--timeout=7200`. Dropping the underscore alone would have put a multi-hour sweep in a 5-minute-per-test job with `-x`, aborting the whole group on both Blackhole SKUs.

So it now uses that directory's actual convention for expensive benchmarks — the env-var gate from `benchmarks/test_benchmark.py:369` (`GEMM_FLOPS_BENCHMARK_ENV`):

```python
ARGMAX_CROSSOVER_BENCH_ENV = "TTNN_RUN_ARGMAX_CROSSOVER_BENCH"
```

Verified it collects 10 tests and skips all 10 without the variable set, and that the gate rather than the arch mark is what skips them.

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
nkapreTT: Renamed throughout. You are right that this collides with how the repo uses the word — across `ttnn/cpp/ttnn/operations/**` I found 32 of 32 uses referring to a hardware block, and this series would have been the first exception. It also collided with itself: it wrote "the SFPU engine" for a program factory and "the SFPU (the Tensix vector FPU)" for the silicon in the same file.

Followed the existing `UpsamplePath` / `select_upsample_path` convention (`pool/upsample/device/upsample_common.hpp:19`), also matching `Conv2dExecutionPath` and `MoEComputePath`:

- `ArgMaxEngine` → `ArgMaxPath`, `::Incumbent` → `::ScalarReader`
- `select_argmax_engine` → `select_argmax_path`, `accelerated_engines_can_serve` → `accelerated_paths_can_serve`
- `ArgmaxParams::engine` → `::path`, `argmax_force_incumbent` → `argmax_force_scalar_reader`

Code line count is unchanged at 1785 — the diff is purely renames.

Two notes on the specifics. "Never" is slightly too strong: `TopologyMappingSolverEngine` in `tt_metal/api/tt-metalium/experimental/fabric/topology_solver.hpp:537` is an enum of algorithm implementations with an `Auto` selector and a force-override, in an installed public header. And I did not use "parallelization strategy" — in this repo that names core splitting (`ReduceOpParallelizationStrategy`), and both paths here are already multicore, so it would have been misleading. The narrower argument — 32/32 in this directory tree, plus the self-collision — is what the rename rests on.

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
nkapreTT: Split rather than moved — the block was doing two jobs.

The routing rule (TILE / bfloat16 / last-dim / width-multiple-of-32 / H < 32 → RVV) documents what `_routes_to_rvv` implements, so it stays above the predicate; moving it would have left that function undocumented. The skip *reason* — ttsim lacking Zve32f, the `_Exit(1)` worker kill, and the scope caveat — moved onto `skip_rvv_routed_on_sim`, collapsed against the `reason=` string that already carried a condensed version.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
nkapreTT: Renamed to `test_argmax_rvv_special_values` (and `test_argmax_sfpu_special_values`). "Battery" appears nowhere else in `tests/`, so it was this series' coinage. `INCUMBENT_EXACT_CASES` went to `SCALAR_READER_EXACT_CASES` for the same reason.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
nkapreTT: Partly. I hoisted the repeated bf16 bit patterns to named module constants — `BF16_MAX_FINITE` (0x7F7F), `BF16_NEG_INF` (0xFF80), `BF16_POS_NAN` (0x7FC0), `BF16_NEG_NAN` (0xFFC0), `BF16_SIGN_BIT` (0x8000) — with one explanation each. They appeared at up to four sites apiece, and the later occurrences carried shorter comments than the first, so a reader had to re-derive the same pattern repeatedly. Verified behaviour-preserving by running the old and new helpers side by side over every case: byte-identical.

I did not name the numbers on the line you flagged. `7 * v // 8` and its siblings are fractional *positions*, and their meaning comes from the branch they sit under — `tie_first_wins` plants the same max twice so the earlier must win, `max_at_end` puts it at `v-1`. A name like `_TIE_FIRST_POS` says strictly less than the branch label already does. Every hex constant in that function also already carried an inline comment. Happy to revisit if you disagree.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
nkapreTT: Agreed on the principle, though that particular line is the `test_argmax_rvv_special_values` docstring, which states what the test asserts and the invariant behind it rather than narrating anything.

The actual narration in that file was elsewhere and is now gone: "the path used to be pinned to the origin; ttnn.argmax had a matching routing hack… Both are gone" (restated as the forward invariant "must not assume core (0,0)"), and "which the path served before selection moved in-tree". Same pass removed "the older boundary of 8 came from a table that…" from the sfpu suite, and two comments in the benchmark that described a history the file never had — it was new in this PR.

Kept the comments that document things expensive to rediscover rather than things this PR did: the `e16m8` register-budget hang, the bit-pattern-versus-IEEE merge argument, and the semaphore-reset rationale for trace replay.

tests/ttnn/unit_tests/operations/reduce/_argmax_engine_crossover_bench.py
nkapreTT: Fixed — `© 2026 Tenstorrent USA, Inc.` on the new files, matching `CONTRIBUTING.md:107` and both neighbours (`_topk_route_cells_bench.py`, `benchmarks/test_benchmark.py`).

nkapreTT:
@mcraigheadTT @roseli-TT @tdowdallTT — this PR needs an approval from you as owners of `tests/pipeline_reorg/ttsim-skip-list.yaml`. Your diff is two entries plus their comments; everything else here is argmax internals.

**What the two lines do, and why they are not temporary.**

This PR adds a Blackhole `ttnn.argmax` path whose kernel runs on TRISC2's Zve32f vector unit. ttsim rejects that unit:

```
ERROR: UnsupportedFunctionality: rv32_v_alu: babyrisc non-compliant V extension is explicitly out of scope
```

and that error path calls `_Exit(1)`, so it kills the whole xdist worker rather than failing one test. `run_for_blackhole()` does not gate it either — `is_blackhole()` is true under ttsim. Hence a file-level skip rather than a test-level one.

I checked whether a newer ttsim removes the need, and it does not: **v1.10.3 is the latest release** and is what `tt_metal/ttsim-version` already pins, there are **no upstream issues or PRs** mentioning the vector extension, and ttsim's README classifies `UnsupportedFunctionality` as deliberately out of scope, distinct from `UnimplementedFunctionality`, which is planned work. The v1.10.3 notes show active SFPU investment (`SFPSWAP`, `SFPLOAD/STORE addr_mode`, QSR support) alongside no babyrisc-V work — which matches what I measured: the SFPU suite passes **175/175** under ttsim, while the RVV one crashes the worker.

The two entries are therefore filed by different reasons:

- **`test_argmax_rvv.py`** — a capability skip, in its own section rather than under `# Job-level timeouts`, with the error string quoted so it is greppable.
- **`test_argmax_sfpu.py`** — stays under `# Job-level timeouts`. It passes, it is just too slow for `sim_bh_p150`'s budget in `.github/time_budget.yaml`.

**One thing worth flagging beyond the skip list.** This PR also changes *routing*, so the pre-existing `tests/ttnn/unit_tests/operations/reduce/test_argmax.py` now reaches the RVV kernel too and crashed ttsim after 14 tests. Deselecting two files would not have covered that, and it was masked only by a timeout-reason skip someone could legitimately remove. It now carries a narrow in-test simulator guard following the `skip_routed_topk_on_sim` precedent from #53732. Verified under the pinned ttsim: `test_argmax.py` goes from `_Exit(1)` after 14 passed to **65 passed, 1 skipped** with no `UnsupportedFunctionality`, the guard fires on 1 of 63 cases, and on real Blackhole silicon it does not fire at all (468 tests pass, 0 skipped).

Happy to change how the entries are filed or worded if you would prefer a different convention.

tests/pipeline_reorg/ttsim-skip-list.yaml
mcraigheadTT: I had a commitment that RVV would not be used, thus the "explicitly out of scope" message + UnsupportedFunctionlaity.

See our official published policy: https://github.com/tenstorrent/ttsim/blob/main/docs/unsupported_functionality.md
(open source repo link, a slightly more detailed policy is in the private repo if you are internal)

Per the above, I do not approve of the change. Please do not use UnsupportedFunctionality features; please change the software to avoid using them.

mcraigheadTT (changes requested):
I do not approve of the change, this breaks a commitment to my team that RVV would not be used. See my reply above.

tests/ttnn/unit_tests/operations/reduce/test_argmax_rvv.py
```diff
@@ -0,0 +1,539 @@
...
+def _make_case(name: str, v: int, b: int, rng: np.random.Generator) -> np.ndarray:
+    """Row-major [b, v] bf16 bit patterns for one named case from CASES."""
+    x = rng.standard_normal((b, v), dtype=np.float32) * 4.0
+    bits = _bits_of(torch.from_numpy(x).bfloat16()).reshape(b, v)
+    kdecoy = 0x7F7E  # one ULP below BF16_MAX_FINITE: a decoy the RNG cannot reach
+    if name == "random":
+        pass
+    elif name == "unique_max":
+        bits[:, 5 * v // 8] = BF16_MAX_FINITE
+        bits[:, v // 3] = kdecoy
```
fplavecTT: Thanks for naming the hex patterns.
I still feel that 5, 8, and 3 need comments. I agree with the previous sentiment that perhaps there are no meaningful names for them, but that doesn't mean that we can't have comments.
And perhaps it's not necessary to have a comment for each individual number: if the function had a docstring that explained what `b` and `v` are and how/why/what the function was doing, then perhaps it would be obvious why these values are what they are. The way it is now, they still look like magic numbers.

sstanisicTT:
Two questions:
- The documentation we have seems to imply that RVV implementation compliant to the full standard, can we depend on the compiler to generate correct instructions at that point?
- AFAIK RVV was never targeted from our software stack, how do we know that it doesn't impact Tensix state in some subtle way that the unit test can catch, that could cause corruptions down the line?

iwroszTT (approved):

nmauriceTT:
@sstanisicTT AFAIK, it's a subset of 'standard' RVV, there are a few missing instructions: https://camel-cdr.github.io/rvv-bench-results/tt_t2/index.html (from external person)

> The Zve32f subest doesn't support the vdiv/vdivu/vrem/vremu/vfdiv/vrfdiv/vfsqrt/vfrsqrt7/vfrec7 instructions, the vcsr CSR and doesn't have compliant floating-point semantics for denormals, rounding, and FMA.

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
```diff
@@ -24,25 +26,84 @@ void bind_reduction_argmax_operation(nb::module_& mod) {
             Keyword args:
                 dim (int, optional): Dim to reduce. ``None`` reduces all elements (ROW_MAJOR input only). Default: ``None``.
                 keepdim (bool, optional): Keep reduced dim. Default: ``False``.
-                sub_core_grids (CoreRangeSet, optional): Limits execution to a subset of cores. Supported on ROW_MAJOR last-dim reductions (<= 2 ranges) and batch/channel dim reductions. Default: ``None``.
+                sub_core_grids (CoreRangeSet, optional): Limits execution to a subset of cores. Supported on ROW_MAJOR last-dim reductions (<= 2 ranges), batch/channel dim reductions, and the accelerated last-dim paths described below. Default: ``None``.
```
fplavecTT: The word "accelerated" is not appropriate here. We sell accelerator cards, so all paths are "accelerated" :)

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
```diff
@@ -24,25 +26,84 @@ void bind_reduction_argmax_operation(nb::module_& mod) {
...
                 memory_config (ttnn.MemoryConfig, optional): Output memory (INTERLEAVED DRAM/L1). Default: input's memory_config.
                 output_tensor (ttnn.Tensor, optional): Preallocated output (must be UINT32, ROW_MAJOR, INTERLEAVED, same device). Default: ``None``.
+                exact_special_values (bool, optional): Restricts the automatic path selection below to paths that
+                    are bit-identical to the default scalar reader kernels on EVERY input, special values included.
```
fplavecTT: This is user facing documentation. I don't think TTNN user will understand what "the default scalar reader kernels" refers to, especially since this is the first time this term is mentioned here.
Also, there is no need to capitalize "EVERY". I'm not sure where AI is getting the idea that this is suitable in a technical document.

ttnn/cpp/ttnn/operations/reduction/argmax/argmax_nanobind.cpp
```diff
@@ -24,25 +26,84 @@ void bind_reduction_argmax_operation(nb::module_& mod) {
...
+                    Today that admits the scalar readers and the RVV path and excludes the SFPU path, whose compare
+                    is IEEE-on-fp32 behind a bf16 special-value gasket rather than those readers' bit-pattern total
```
fplavecTT: Similar here. Is a Python programmer writing models expected to know what RVV is?

"whose compare is IEEE-on-fp32 behind a bf16 special-value gasket rather than those readers' bit-pattern total"??
What does that mean?

sstanisicTT:
I've cross checked against sfpi-gcc, and `-mcpu=tt-bh-tensix` implies `-march=rv32im_zmmul_zaamo_zba_zbb_xtttensixbh_zve32f` which answers the first question, but I think the second one still stands. This change introduces hardware that was previously not stress tested in software directly into kernels, which seems dangerous.

Also:
- `The path split follows how much useful work is present in each 32-row tile: for decode shapes, SFPU still processes all 32 lanes while RVV scans only the valid rows; once H=32 fills the tile, SFPU benefits from lane-parallel reduction.`
Implies that the performance benefit is being compared apples to oranges because it's not the equivalent algorithm, which leads me to think that the performance benefits would be much less if the SFPU path was squeezed?

PR #54107, opened 2026-08-23
fix(sfpu): xielu neg-exp underflow returns 0 instead of garbage reconstruction

### LINKED ISSUE #54046: [Bounty $7000] ttnn.xielu: exponent-reconstruction overflow produces values up to ±2.7e38 (and NaN) for x < -87.68
## Summary

`ttnn.xielu`'s negative-input branch calls an internal helper, `_sfpu_neg_exp_f32_`,
to compute `exp(x)`. For any input `x < -87.68`, this helper's range-reduction logic
breaks down and returns a essentially-random-magnitude garbage value instead of a
number close to zero — including values as large as `2.7e38` (within a factor of ~1.25
of `FLT_MAX`) and, in some cases, `NaN`. This propagates directly into `ttnn.xielu`'s
output:

```
ttnn.xielu(x=-90.0, alpha_p=..., alpha_n=0.8)  ->  2.342323e+37
```
```
correct value (torch reference)               ->  26.2
```

A source-level fp32 model of the kernel, cross-checked bit-for-bit against the IEEE 754
semantics `exexp`/`setexp` actually implement, confirms this is not an isolated input:
a dense sweep of the affected region (`x` from `-87.68` to `-300`, step `0.01`, 21,232
points) finds ~0.9% of points produce an absolute error greater than 1 (i.e. clearly,
visibly wrong, not just imprecise), with errors reaching `±2.7e38`.

## Root cause

`_sfpu_neg_exp_f32_` (`ckernel_sfpu_xielu.h`) computes `exp(x)` via the standard
range-reduction-then-polynomial approach: `x = k*ln(2) + r`, evaluate a 7-term Taylor
polynomial for `exp(r)` on the small reduced range, then reconstruct `2^k * poly(r)` via
direct exponent-bit manipulation.

The bug is in how the underflow clamp is applied:

```cpp
// L18: constexpr float UNDERFLOW_THRESHOLD = -126.5f;
// L22: sfpi::vFloat z = val * sfpi::vConstFloatPrgm0;          // z = x / ln(2)
// L25: z = sfpi::max(z, UNDERFLOW_THRESHOLD);                  // <-- clamps z, not the final result
// L29: sfpi::vFloat k = _sfpu_round_to_nearest_int32_(z, k_int);
```

The clamp is applied to `z` (and therefore `k`) — the *driver* of the range reduction —
rather than to the final output. Once `x < -126.5 * ln(2) ≈ -87.68`, `z` and `k` become
constant, but the residual:

```cpp
// L56: sfpi::vFloat r = k * LN2_LO + r_hi;   // r_hi = k*LN2_HI + x
```

keeps growing **unboundedly negative** as `x` decreases further, since `k` has stopped
tracking it (at `x = -1000`, `r = -912.7`). The 7-term Taylor polynomial (`L62`,
`PolynomialEvaluator::eval`) is only valid for `|r| ≲ 0.35` and diverges wildly outside
that radius — its sign and magnitude become effectively arbitrary as `|r|` grows. That
divergent value is then bit-reinterpreted as a real float:

```cpp
// L78: sfpi::vInt p_exp = sfpi::exexp(p, sfpi::ExponentMode::Biased);
// L80: sfpi::vInt new_exp = p_exp + k_int;
// L83: result = sfpi::setexp(p, new_exp);   // <-- writes new_exp into an 8-bit field, can wrap
```

`new_exp` is written directly into the 8-bit IEEE 754 exponent field with no bounds
check. Depending on the divergent polynomial's exact value, this either lands on a huge
biased exponent (producing an enormous finite value) or on `0xFF` (producing `Inf`/`NaN`).

**This is called directly from `calculate_xielu`'s "large negative" branch with no
additional guard**:

```cpp
// L147-148:
v_else {  // large negative
    sfpi::vFloat exp_term = _sfpu_neg_exp_f32_(x) - 1.0f - x;
```

Every other exp-tail helper in this codebase that uses the same range-reduction pattern
(`ckernel_sfpu_gelu.h`'s `x_times_exp_negative_tail`, `ckernel_sfpu_softplus.h`'s
`softplus_exp_negative`) has an explicit flush-to-zero guard before the `setexp` call —
`v_if(new_exp > 0) { result = setexp(...); } v_endif;`, defaulting to `0.0f` otherwise.
`_sfpu_neg_exp_f32_` is missing this guard entirely.

## Why the result isn't uniformly catastrophic (and why that isn't reassuring)

Most inputs past the `-87.68` threshold *do* come out correct, but by accident, not by
design: `calculate_xielu`'s full formula is `alpha_n * (exp(x) - 1 - x) + beta*x`. For
very negative `x`, the correct answer is dominated by the `-1-x` term (which grows large
and positive). When the garbage `exp(x)` value happens to still be numerically small,
it gets swamped by `-1-x` and the final answer is right anyway — e.g. at `x=-100`,
`x=-150`, `x=-1000` the end-to-end result matches the golden value. But the polynomial's
divergence is not monotonic or predictable — at `x=-89` and `x=-90` specifically, it
produces values (`2.7e38`, `2.3e37`) large enough to dominate instead, and the final
`ttnn.xielu` output is wrong by 36+ orders of magnitude. The bug is real across the
entire `x < -87.68` domain; whether it's *visible* in a given output depends on an
unpredictable interaction between the divergent polynomial and the exponent-field
wraparound.

## Why tests miss it

`tests/ttnn/unit_tests/operations/eltwise/test_activation.py::test_xielu` draws inputs
from `torch.randn([32, 32])` — standard normal. `P(X < -87.68)` for a standard normal is
`0.0` in double-precision floating point (astronomically below any realistic sample
count). No other test file in the repository references `xielu` at all. This isn't a
narrow-domain miss like a `[-4, 10]` sweep bound — there is no coverage of this input
region whatsoever.

## Proposed fix

Add the same explicit flush-to-zero guard every sibling helper already uses, keyed on
`new_exp`, before the `setexp` call in `_sfpu_neg_exp_f32_`:

```cpp
sfpi::vInt p_exp = sfpi::exexp(p, sfpi::ExponentMode::Biased);
sfpi::vInt new_exp = p_exp + k_int;

sfpi::vFloat result = 0.0f;
v_if(new_exp > 0) { result = sfpi::setexp(p, new_exp); }
v_endif;
return result;
```

This makes `_sfpu_neg_exp_f32_` degrade gracefully to `0.0` for any input whose true
`exp(x)` would underflow — the mathematically correct behavior — instead of returning
whatever garbage bit pattern the out-of-range polynomial happens to produce. Since
`calculate_xielu`'s formula already correctly handles `exp(x) -> 0` for large negative
`x` (that's exactly what makes the "accidentally correct" cases above come out right),
this fix should make the *entire* domain correct, not just the lucky subset.

## Validation

Completed:
- Bit-exact IEEE 754 simulation of `exexp`/`setexp`'s reconstruction logic in Python,
  reproducing the helper's arithmetic exactly (Cody-Waite range reduction, same
  coefficients, same 8-bit exponent-field write).
- Confirmed the mechanism directly: `z`/`k` freeze at `x ≈ -87.68` while `r` grows
  unboundedly negative past that point (traced numerically, not assumed).
- Traced through the *full* `calculate_xielu` formula, not just the helper in isolation,
  confirming the corruption survives into the real op's output (`x=-90` -> `2.34e37`
  instead of `26.2`).
- Dense sweep (`x` in `[-87.68, -300]`, step `0.01`, 21,232 points): ~0.9% visibly wrong,
  max absolute error `2.7e38`.

Limitations:
- This is a source-level fp32 model of the arithmetic, not a run on real hardware or
  `ttsim` — the exact `setexp`/`exexp` bit semantics are inferred from how they're used
  consistently elsewhere in this codebase (direct 8-bit exponent-field overwrite), not
  confirmed against a hardware simulator. Device validation would strengthen this.
- The exact set of `x` values that produce `NaN` specifically (vs. a large finite value)
  has not been fully characterized — observed in a wider scan but not mapped precisely.
- Only the Wormhole `is_fp32_dest_acc_en` path was checked in this depth; the BF16 path
  and the Blackhole variant of this file have not yet been diffed/verified to share the
  exact same bug (worth checking before merge, `gelu.h`/`remainder.h` in this same
  codebase were confirmed byte-identical between Wormhole and Blackhole, so this may be
  too, but that hasn't been confirmed for `xielu.h` specifically).

## Acceptance criteria

- `_sfpu_neg_exp_f32_(x)` returns `0.0f` (not a garbage large/negative/NaN value) for all
  `x` where the true `exp(x)` underflows fp32, matching the FTZ pattern already used by
  `gelu.h`/`softplus.h`'s equivalent helpers.
- `ttnn.xielu` produces the correct saturated value across the full `x < -87.68` domain,
  not just the subset where the bug happens not to manifest.
- Regression test added covering this input range explicitly (current `test_xielu` only
  samples `torch.randn`, which cannot reach it).

Adding the exact intermediate state at representative failing inputs, for anyone
verifying the mechanism (traced from a bit-exact IEEE 754 model of
`_sfpu_neg_exp_f32_`, cross-checking `exexp`/`setexp` against how they're used
consistently elsewhere in this codebase — direct read/write of the 8-bit biased-exponent
field):

| x | k_int | r | poly(r) | p_exp (biased) | new_exp = p_exp+k_int | new_exp truncated to 8 bits | `_sfpu_neg_exp_f32_(x)` | true `exp(x)` |
|---|---|---|---|---|---|---|---|---|
| -88.73 | -126 | -1.394 | 0.2479 | 124 | -2 | 254 | **3.374e+38** | ~4.5e-39 |
| -90.00 | -126 | -2.664 | 0.02151 | 121 | -5 | 251 | **2.928e+37** | ~8.2e-40 |
| -90.67 | -126 | -3.334 | -0.2381 | 124 | -2 | 254 | **-3.240e+38** | ~4.1e-40 |

The mechanism precisely: once `x < -87.68`, `k_int` freezes at a constant (`-126`) while
`r` keeps growing in magnitude as `x` decreases (range reduction no longer holds). The
7-term Taylor polynomial evaluated at these out-of-radius `r` values does **not**
explode to something astronomically large — `poly(r)` stays an ordinary small-magnitude
number in all three examples above (0.02–0.25). What actually produces the extreme
output is that `p_exp` (the biased exponent of that small `poly(r)`) plus the frozen
`k_int=-126` lands on a small **negative** integer (-2 to -5), and writing that negative
value into the 8-bit exponent field truncates it via two's-complement wraparound to
251–254 — right at the edge of the encodable range (255 is reserved for Inf/NaN) —
rather than saturating to 0 the way every sibling helper's explicit `v_if(new_exp > 0)`
guard would produce.

Also confirmed: this file is byte-identical on Blackhole (`ckernel_sfpu_xielu.h`, only
diff is a comment expanding on Blackhole's SFPMAD negate-flag usage, zero logic
changes), so this affects both architectures.

zestoles:
## Problem

`ttnn.xielu`'s negative branch calls `_sfpu_neg_exp_f32_` to compute `exp(x)`. That helper clamped the range-reduction driver `z = x/ln(2)` to `-126.5` *before* rounding:

```cpp
z = sfpi::max(z, UNDERFLOW_THRESHOLD);
```

Once `x < -126.5 * ln(2) ≈ -87.68`, `k` froze at ~-126 while the Cody-Waite residual `r = x - k*ln(2)` kept growing unboundedly negative (e.g. `r ≈ -87` at `x = -200`). The 7-term Taylor polynomial is only valid for `|r| ≲ 0.35`, so the reconstructed value became garbage — magnitudes up to `±2.7e38` and, when `p` overflowed inside the polynomial, `NaN` — exactly as reported in #54046.

## Fix

Remove the driver clamp and guard the **output** exponent instead, mirroring the underflow idiom already used by the exp kernel (`ckernel_sfpu_exp.h`, `v_if(e >= 1)`):

- With no clamp, `k = round(x/ln(2))` tracks the input for any magnitude and `r` stays inside the polynomial's domain.
- After reconstruction, `new_exp = exexp(p) + k_int` is only fed to `setexp` when it is a representable fp32 normal exponent (`>= 1`). Below that point the true `exp(x)` is subnormal or zero; returning `0` keeps the absolute error below `FLT_MIN`, which is negligible next to xielu's dominant `-x` term.

Applied identically to the Wormhole and Blackhole kernel copies (Blackhole's copy differs only in comments).

## Testing

- Added `test_xielu_underflow_region` (fp32 + bf16): sweeps `x ∈ {-80, -87.5, -88, -90, -100, -126, -200, -300, -1000}` across two `(alpha_p, alpha_n)` parameter sets, asserts all outputs are finite and matches the torch golden with the same tolerances as the existing `test_xielu`.
  - On current main the garbage/NaN region makes this fail; with the fix the sweep passes.
- Existing `test_xielu` cases are unaffected: for `x > -87.68` the removed clamp was inactive (`max` identity), so behavior in the valid region is bit-identical.

Fixes #54046

DCCO signed-off commit included.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_activation.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_activation.py
@@ -616,6 +616,43 @@ def test_xielu(alpha_p, alpha_n, dtype, device):
 
 
 @pytest.mark.parametrize(
+    "dtype",
+    [
+        "float32",
+        "bfloat16",
+    ],
+)
+@pytest.mark.parametrize("alpha_p, alpha_n", [(0.8, 0.8), (1.0, 0.5)])
+def test_xielu_underflow_region(dtype, alpha_p, alpha_n, device):
+    """Regression test for #54046.
+
+    For x < -87.68 (= -126.5 * ln(2)), exp(x) underflows the fp32 normal
+    range. The negative-input branch of xielu used to clamp the
+    range-reduction driver instead of guarding the reconstructed exponent,
+    so the residual grew unbounded and the kernel returned garbage
+    magnitudes up to ~2.7e38 or NaN in that region. The correct output is
+    dominated by the -x term there and must stay finite.
+    """
+    torch_dtype = getattr(torch, dtype)
+    ttnn_dtype = getattr(ttnn, dtype)
+
+    x_values = [-80.0, -87.5, -88.0, -90.0, -100.0, -126.0, -200.0, -300.0, -1000.0]
+    torch_input = torch.tensor([x_values] * 32, dtype=torch_dtype)
+    golden_fn = ttnn.get_golden_function(ttnn.xielu)
+    torch_output = golden_fn(torch_input, alpha_p=alpha_p, alpha_n=alpha_n)
+
+    ttnn_input = ttnn.from_torch(torch_input, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_output = ttnn.xielu(ttnn_input, alpha_p=alpha_p, alpha_n=alpha_n)
+    ttnn_output = ttnn.to_torch(ttnn_output)
+
+    assert torch.isfinite(ttnn_output.float()).all(), "xielu produced non-finite values in the underflow region"
+    if dtype == "float32":
+        assert_allclose(torch_output, ttnn_output, rtol=6e-05, atol=1e-06)
+    else:
+        assert_with_ulp(torch_output, ttnn_output, 1)
+
+
+@pytest.mark.parametrize(
     "shapes",
     [
         (3, 4, 64, 32),
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_xielu.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_xielu.h
@@ -15,14 +15,15 @@ namespace ckernel::sfpu {
 sfpi_inline sfpi::vFloat _sfpu_neg_exp_f32_(sfpi::vFloat val) {
     sfpi::vFloat result = 0.0f;
 
-    constexpr float UNDERFLOW_THRESHOLD = -126.5f;
-
     // Step 1: Compute k = round(x / ln(2))
     // z = x / ln(2) = x * (1/ln(2))
+    //
+    // NOTE: do NOT clamp z here. Clamping the range-reduction driver makes
+    // k (and therefore r = x - k*ln(2)) freeze while the residual keeps
+    // growing for more negative inputs, pushing r far outside the domain
+    // of the polynomial below and producing garbage after reconstruction.
+    // Instead, the underflow is handled on the OUTPUT exponent below.
     sfpi::vFloat z = val * sfpi::vConstFloatPrgm0;
-
-    // Clamp z to -126.5: exp(x) underflows to 0 for large negative x
-    z = sfpi::max(z, UNDERFLOW_THRESHOLD);
 
     // Round z to nearest integer using round-to-nearest
     sfpi::vInt k_int;
@@ -82,8 +83,17 @@ sfpi_inline sfpi::vFloat _sfpu_neg_exp_f32_(sfpi::vFloat val) {
     // Add k_int to get the new exponent
     sfpi::vInt new_exp = p_exp + k_int;
 
-    // Set the new exponent
-    result = sfpi::setexp(p, new_exp);
+    // Guard the OUTPUT exponent: for x below ~-87.7 (=-126.5*ln(2)) the true
+    // exp(x) is subnormal or zero, which the SFPU cannot represent. Returning
+    // 0 keeps the absolute error below FLT_MIN, while reconstructing with an
+    // out-of-range exponent produced garbage magnitudes / NaN.
+    // (Same underflow guard as the exp kernel: reconstruct only when the
+    // biased exponent is >= 1.)
+    result = 0.0f;
+    v_if(new_exp >= 1) {
+        result = sfpi::setexp(p, new_exp);
+    }
+    v_endif;
 
     return result;
 }
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_xielu.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_xielu.h
@@ -15,14 +15,15 @@ namespace ckernel::sfpu {
 sfpi_inline sfpi::vFloat _sfpu_neg_exp_f32_(sfpi::vFloat val) {
     sfpi::vFloat result = 0.0f;
 
-    constexpr float UNDERFLOW_THRESHOLD = -126.5f;
-
     // Step 1: Compute k = round(x / ln(2))
     // z = x / ln(2) = x * (1/ln(2))
+    //
+    // NOTE: do NOT clamp z here. Clamping the range-reduction driver makes
+    // k (and therefore r = x - k*ln(2)) freeze while the residual keeps
+    // growing for more negative inputs, pushing r far outside the domain
+    // of the polynomial below and producing garbage after reconstruction.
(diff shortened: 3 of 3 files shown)


zestoles:
Acknowledging the prior art pointed out by @Adraca — PRs #52607/#52608 fix the same _sfpu_neg_exp_f32_ underflow mechanism and predate this issue, so this PR is redundant. Closing in favor of that work. Thanks for the clear root-cause discussion on the issue.

PR #54108, opened 2026-08-23
fix(sfpu): softplus must take the non-linear branch at input*beta == threshold

### LINKED ISSUE #53261: [Bounty $1000] ttnn.softplus takes the linear branch at exactly input*beta == threshold (strict < where reference uses <=)
## Summary

`ttnn.softplus` takes its linear-fallback branch at exactly `input * beta == threshold`, where PyTorch still evaluates the softplus. The kernel uses a strict `<` where the reference semantics require `<=`, so at the boundary point it returns the raw linear value instead of `log1p(exp(beta*x))/beta`. It is a one-character fix; details below.

## Reference semantics

PyTorch documents the linear fallback as a strict inequality: softplus "reverts to the linear function when `input * beta > threshold`." So at exactly `input * beta == threshold`, the true result is `log1p(exp(beta*x)) / beta`, not `x`.

## Current implementation

`calculate_softplus_body()` in `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_softplus.h:103` (byte-identical on Blackhole, same file and line) branches:

```cpp
v_if(t < threshold) {          // t = input * beta
    ... evaluate softplus ...
}
v_else {
    ... linear fallback: result = input ...
}
```

At `t == threshold`, `t < threshold` is false, so control falls to the linear branch and returns `input`. PyTorch, using `>`, still evaluates softplus at that point. The comparison needs to be `t <= threshold`.

## Impact

Deterministic, and largest when the threshold is small. Because the linear branch returns `x` while the true value is `log1p(exp(beta*x))/beta`, the absolute error at the boundary is `log(1 + exp(-threshold)) / beta`. Worked examples (all with `x * beta == threshold` exactly, so all hit the boundary):

| beta | threshold | x | kernel returns | true softplus | abs error |
|---|---|---|---|---|---|
| 1.0 | 0.5 | 0.5 | 0.5 | 0.974077 | 0.474 (49%) |
| 2.0 | 1.0 | 0.5 | 0.5 | 0.656631 | 0.157 (24%) |
| 1.0 | 2.0 | 2.0 | 2.0 | 2.126928 | 0.127 (6.0%) |
| 4.0 | 1.0 | 0.25 | 0.25 | 0.328315 | 0.078 (24%) |

The `true softplus` column is `log1p(exp(beta*x))/beta` by hand; the `kernel returns` column is the linear branch's `x`, which the code above takes at the boundary. One bfloat16 ULP on either side behaves correctly, which isolates the defect to the comparison itself rather than the polynomial.

## Why it survived

bfloat16's coarse grid makes exact boundary hits ordinary rather than rare: `beta=2, threshold=1` puts the boundary at `x=0.5`, and `beta=1, threshold=2` at `x=2.0`, both exactly representable and entirely normal activations. And the parametrized path looks effectively untested for this: open PR #52375 shows the softplus test golden drops `beta` and `threshold` entirely, so every existing softplus test compares against the defaults where the boundary is not exercised.

This is distinct from #51866 (softplus returning exactly 0 for `x < -5`, the large-negative tail); this report is only about the `input*beta == threshold` boundary.

## Proposed fix

Change the comparison at `ckernel_sfpu_softplus.h:103` from `v_if(t < threshold)` to `v_if(t <= threshold)` on both the Wormhole B0 and Blackhole kernels, and add a parametrized `(beta, threshold)` case to the LLK unit test and the ttnn sweep that exercises an exact boundary input (e.g. `beta=2, threshold=1, x=0.5`) against the torch reference. `ttnn.softplus` is used in the model tree (3 files), so the boundary is reachable in practice.

Happy to prepare the PR. I have also proposed this together with two related backward-op numerics fixes (`hardswish_bw`, `sigmoid_bw`) via bounties@tenstorrent.com, so it can be handled either way.

zestoles:
﻿## Problem

`ttnn.softplus` takes the linear fallback branch when `input * beta == threshold` exactly, because the kernel compares with strict `<` while PyTorch's documented semantics revert to the linear function only for `input * beta > threshold`. At the boundary point the kernel returned `x` instead of `log1p(exp(beta*x)) / beta` (e.g. `beta=1, threshold=0.5, x=0.5`: returned `0.5`, correct is `≈0.974077` — 49% relative error at that point).

Reported in #53261 (includes worked examples for several `beta`/`threshold` pairs).

## Fix

Change the branch compare from `t < threshold` to `t <= threshold` in all three architecture copies:

- `tt_metal/hw/ckernels/wormhole_b0/.../ckernel_sfpu_softplus.h`
- `tt_metal/hw/ckernels/blackhole/.../ckernel_sfpu_softplus.h`
- `tt_metal/hw/ckernels/quasar/.../ckernel_sfpu_softplus.h`

Comments updated accordingly (including the quasar doc note referencing the compare).

## Testing

Added `test_softplus_boundary_equality` (fp32 + bf16), a deterministic regression test that lands exactly on `input * beta == threshold` (via `(beta=1, thr=20, x=20)` and `(beta=2, thr=5, x=2.5)`, both exact in fp32/bf16), plus just-above-threshold and negative controls, compared against `torch.nn.functional.softplus`.

Fails on current main (boundary point off by ~49% relative), passes with this change.

Fixes #53261

DCO signed-off.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_activation.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_activation.py
@@ -636,3 +636,36 @@ def test_lgamma_fp32(device, shapes):
     tt_out = ttnn.to_torch(z_tt)
 
     assert_with_pcc(z_torch, tt_out, 0.999)
+
+
+@pytest.mark.parametrize(
+    "dtype",
+    [
+        "float32",
+        "bfloat16",
+    ],
+)
+@pytest.mark.parametrize("beta, threshold, x_boundary, x_above", [(1.0, 20.0, 20.0, 20.5), (2.0, 5.0, 2.5, 2.75)])
+def test_softplus_boundary_equality(dtype, beta, threshold, x_boundary, x_above, device):
+    """Regression test for #53261.
+
+    PyTorch reverts to the linear branch only when input * beta > threshold,
+    so at exactly input * beta == threshold softplus must still be evaluated.
+    The kernel used a strict `<` compare and returned the raw linear value at
+    the boundary point.
+    """
+    torch_dtype = getattr(torch, dtype)
+    ttnn_dtype = getattr(ttnn, dtype)
+
+    x_list = [x_boundary, x_above, -x_boundary]
+    torch_input = torch.tensor([x_list] * 32, dtype=torch_dtype)
+    torch_output = torch.nn.functional.softplus(torch_input, beta=beta, threshold=threshold)
+
+    ttnn_input = ttnn.from_torch(torch_input, dtype=ttnn_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    ttnn_output = ttnn.softplus(ttnn_input, beta=beta, threshold=threshold)
+    ttnn_output = ttnn.to_torch(ttnn_output)
+
+    if dtype == "float32":
+        assert_allclose(torch_output, ttnn_output, rtol=1e-05, atol=1e-05)
+    else:
+        assert_with_ulp(torch_output, ttnn_output, 1)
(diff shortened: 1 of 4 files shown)


ncvetkovicTT:
Not assigned to you, let's close the PR. FYI @jberkowitzTT

zestoles:
Understood — leaving this to @AJ0070 who is assigned to #53261 and found the third kernel copy. Thanks for the quick triage @ncvetkovicTT.

PR #54112, opened 2026-08-23
sfpi: two-branch erfc with dedicated positive-side accurate path

exstttt:
Addresses #54053 ($8,000 community bounty).

## Problem

`ttnn.erfc` used a single bf16-grade 4/5 rational LUT for **every** precision mode, no `is_fp32_dest_acc_en` split, unlike sibling `erf`. Measured (and self-documented): fp32 MaxULP ≈ 9.4M near x=5, 114% relative error vs `scipy.special.erfc`.

## Fix

Two-branch decomposition that eliminates the dynamic-range problem instead of fighting it with higher-degree LUTs:

- **x ≤ 0:** direct degree-16/16 rational fit of erfc (range ~[1, 2])
- **x > 0:** fit `R(x) = erfc(x) · exp(x²)`, smooth and O(0.11..1), then reconstruct with the SFPU exp2 primitive: `erfc(x) = exp2(−log2(e)·x²) · R̂(x)`

Both branches are plain Horner over 17-coefficient arrays; `APPROXIMATION_MODE` retained for a future fast-path split.

## Accuracy (fp64 golden: scipy.special.erfc)

| Region | max rel error (v1 fit) |
|---|---|
| x ≤ 0 | 3.3e-4 |
| x > 0 (R-form) | 7.7e-4 |

~150× better than the shipped kernel immediately. The coefficients were produced by weighted rational least squares; the erf sibling header documents an internal coordinate-descent on-device refit pipeline, running these constants through that flow should reach the few-ulp fp32 grade of the erf n16/d16 fit. Happy to iterate.

Both Wormhole and Blackhole variants patched (identical source, the new path uses no arch-specific sqrt/reciprocal templates).

### DIFF
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_erfc.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_erfc.h
@@ -1,86 +1,67 @@
-// SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
 //
 // SPDX-License-Identifier: Apache-2.0
 
 #pragma once
 
-#include <array>
-
 #include "ckernel.h"
 #include "ckernel_defs.h"
-#include "cmath_common.h"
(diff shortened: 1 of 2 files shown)


ncvetkovicTT:
Issue is closed, let's close the PR. FYI @jberkowitzTT

Issue #45821, opened 2026-06-02
Hash collision in Shape hashing


iwroszTT:
A new test case added in PR #45578 triggers hash collision, leading to a wrong cache hit.
The new test:
unit_tests/operations/fused/test_batch_norm.py::test_batch_norm_aliased_running_and_affine_tensors

Root cause analysis:

1.  The new test uses input shape (1, 152, 24, 32). Inside batch_norm, this triggers a ttnn::permute on a tensor with logical_shape=[1, 152, 1, 1] (the post-mean_NHW tensor).
2. Earlier in the same test file, test_batch_norm_compute_config runs with input shape (3, 17, 47, 32). Its mean_NHW produces an intermediate tensor with logical_shape=[3, 17, 1, 1], and the same permute path runs on it. That program is built and inserted into the program cache.
3. PermuteDeviceOperation has no custom compute_program_hash, so it uses the default reflection-based hash over (operation_attributes, tensor_args). The tensor hash walks tensor_spec → logical_shape → Span<uint32_t>. The Span is hashed element-wise via the boost-style hash_objects combiner.
4. That combiner is mathematically weak: spans [3, 17, 1, 1] and [1, 152, 1, 1] produce the SAME 64-bit hash. 

Reproduction
  The smallest local reproducer is two tests in this order:
  pytest tests/ttnn/unit_tests/operations/fused/test_batch_norm.py::test_batch_norm_compute_config[...input_shapes=torch.Size([3, 17, 47,32])]
 tests/ttnn/unit_tests/operations/fused/test_batch_norm.py::test_batch_norm_aliased_running_and_affine_tensors -- with the original shape as in the merged PR #45578 

Inserting device.disable_and_clear_program_cache() between them makes it pass — motivating the bug is the cached-program reuse, not the new code.

Proposed Fix:
1. Replace the hash_objects combiner in tt_stl/reflection.hpp:1431 with a better mixer (e.g. wyhash, xxh3,
  or splitmix64-based combine)

bbradelTT:
cc @dgomezTT @ayerofieiev-tt

dgomezTT:
@azaytsev-epam can you look at this?

bbradelTT:
It looks like there's already a PR up for review: https://github.com/tenstorrent/tt-metal/pull/46385
@dgomezTT is there anything for @azaytsev-epam to do?

ayerofieiev-tt:
[...]

PR #54114, opened 2026-08-23
[skip ci] Add new bug_checker rules

blozano-tt:
Adds five new `bug_checker` rules, derived from an audit of ~1,398 merged `fix`-labelled tt-metal PRs (the MINFRA-1266 conventional-commit backfill), plus the internal engineering note *"Descriptors and Specs: how a TTNN op describes its program"* by @dgomezTT.

### Rules added

| Rule | Severity | Scope |
|---|---|---|
| `program-cache-hash-collision` | blocking | `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `*_op.cpp`, `*_op.hpp`, `tt-train/sources/**/device/*device*` |
| `smuggled-buffer-runtime-arg` | blocking | `ttnn/cpp/ttnn/operations/**/device/*factory*`, `*program*`, `*device_operation*`, `*_op.cpp`, `*_op.hpp`, `*fusion*`, `tt_metal/impl/program/**` |
| `override-rebuild-in-cache-hit` | blocking | `ttnn/cpp/ttnn/operations/**/device/*factory*`, `*program*`, `*device_operation*`, `*_op.cpp`, `*_op.hpp`, `*fusion*` |
| `llk-stale-hw-config-state` | blocking | `tt_metal/tt-llk/**/llk_lib/**`, `tt_metal/tt-llk/**/common/inc/**`, `tt_metal/hw/ckernels/**/llk_api/**` |
| `op-shard-layout-validation` | warning | `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `.../*_op.cpp`, `.../*_op.hpp` |

**`program-cache-hash-collision`** — a *custom* `compute_program_hash()` that omits a structural field the program factory reads, producing a wrong cache hit (the default reflection hash is protected by a canonical-key comparison, so it only costs a redundant rebuild, not a wrong result). Also flags `ProgramDescriptor::custom_program_hash` misuse, and separates two distinct shape-collision modes: hashing rank (collides same-rank/different-dims) vs. hashing volume (collides same-volume/different-tiling).

**`smuggled-buffer-runtime-arg`** — a buffer address written into a kernel's runtime-arg vector without going through `emplace_runtime_args`/`emplace_common_runtime_args` (or an explicit `override_runtime_arguments()`), so the program cache never re-patches it on a cache hit. Names the real guard (`scripts/detect_smuggled_rta.py`, `detect-smuggled-rta` hook) and its `// smuggled-rta-ok` suppression convention, plus the work-distribution and scalar traps where the fix isn't mechanical.

**`override-rebuild-in-cache-hit`** — calling `create_descriptor()` from `override_runtime_arguments()`, which pays the full cache-*miss* host cost on every cache *hit* (the ResNet50 20x / resnet-T3K 760% cliff). `scripts/detect_override_rebuild.py` already guards the literal form, but factoring the rebuild into a shared helper hides it from that textual scan while costing the same at runtime — catching that indirection is the centrepiece of this rule.

**`llk-stale-hw-config-state`** — unpacker/packer/SFPU hardware config carrying dirty state from a *previous* kernel invocation: a missing reconfig, a reconfig that updates only some fields of a shared 32-bit config word, a direct write bypassing a software state tracker, or a config rewrite with no execution-unit drain. Order-dependent by nature, so it passes in isolation and fails in a fused sequence. Backed by ~30 merged `fix` PRs, many tagged `[LLK]`/`[SAN]`.

**`op-shard-layout-validation`** — ops that read layout, memory config, or shard spec to build their program without validating up front that they support what they were given, producing wrong output or a hang instead of a clean `TT_FATAL`. The single largest bucket in the audit (74 PRs). Deliberately scoped away from `reshape-dim-check`: this is about the *presence of an entry-level compatibility check*, not shape arithmetic.

### Copilot review

Copilot flagged real gaps on the initial push, addressed in a follow-up commit: `program-cache-hash-collision`'s glob missed legacy `*_op.cpp/hpp` custom-hash implementations, and `smuggled-buffer-runtime-arg`/`override-rebuild-in-cache-hit` missed non-`*program_factory*` naming families (`*_factory.cpp`, `*_program.cpp`, `*_fusion.cpp`) — all three globs are now broadened, with 4 new regression tests pinning the real-world filenames that were previously missed. Also corrected an example in `smuggled-buffer-runtime-arg.md` that used a nonexistent `BufferBinding` constructor, and fixed a mischaracterization of issue #45821 in `program-cache-hash-collision.md` (it's a default-hash weak-combiner bug, not evidence for the custom-hash rule this PR adds).

### Validation

- `python -m pytest .github/bug_checker/tests/` — all passing, including the 4 new glob-coverage regression tests
- `run_bug_checker.py --subcommand list-rules` — all 7 rules load
- Path globs verified against the real tree, including the broadened hash-collision/runtime-arg/override-rebuild globs from the Copilot review pass

### Note for reviewers

`tt_metal/tt-llk` is a plain in-tree directory (mode `040000`, ~1,048 files on `main`), not a submodule — LLK was de-submoduled by vendoring, so these globs resolve inside this repo.

Not covered here: §4.3 of the note (copying the runtime-args grid by value) is fully mechanical and would be better served by a plain pre-commit script than an LLM rule — see the PR discussion.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:bug-checker/program-cache-staleness-rules)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:bug-checker/program-cache-staleness-rules)

### DIFF
--- a/.github/bug_checker/README.md
+++ b/.github/bug_checker/README.md
@@ -156,6 +156,9 @@ python .github/bug_checker/run_bug_checker.py --branch --subcommand dry-run
   - `reshape-dim-check` — warning, paths: `ttnn/cpp/ttnn/operations/data_movement/**`, labels: `area:ops`
   - `program-cache-hash-collision` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `tt-train/sources/**/device/*device*`
   - `smuggled-buffer-runtime-arg` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*program_factory*`, `tt_metal/impl/program/**`, `tt-train/sources/**/device/*program_factory*`
+  - `override-rebuild-in-cache-hit` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*program_factory*`, `ttnn/cpp/ttnn/operations/**/device/*device_operation*`
+  - `llk-stale-hw-config-state` — blocking, paths: `tt_metal/tt-llk/**/llk_lib/**`, `tt_metal/tt-llk/**/common/inc/**`, `tt_metal/hw/ckernels/**/llk_api/**`, labels: `area:llk`
+  - `op-shard-layout-validation` — warning, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `ttnn/cpp/ttnn/operations/**/device/*_op.cpp`, `ttnn/cpp/ttnn/operations/**/device/*_op.hpp`, labels: `area:ops`
 
 ### LLM Analysis — `llm.py` : `LLMSession.analyze_rule()`
 
--- a/.github/bug_checker/README.md
+++ b/.github/bug_checker/README.md
@@ -154,9 +154,9 @@ python .github/bug_checker/run_bug_checker.py --branch --subcommand dry-run
 - Current rules:
   - `ccl-ring-buffer-mismatch` — blocking, paths: `tt_metal/impl/ccl/**`, `ttnn/cpp/ttnn/operations/ccl/**`, labels: `area:ccl`
   - `reshape-dim-check` — warning, paths: `ttnn/cpp/ttnn/operations/data_movement/**`, labels: `area:ops`
-  - `program-cache-hash-collision` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `tt-train/sources/**/device/*device*`
-  - `smuggled-buffer-runtime-arg` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*program_factory*`, `tt_metal/impl/program/**`, `tt-train/sources/**/device/*program_factory*`
-  - `override-rebuild-in-cache-hit` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*program_factory*`, `ttnn/cpp/ttnn/operations/**/device/*device_operation*`
+  - `program-cache-hash-collision` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `ttnn/cpp/ttnn/operations/**/device/*_op.cpp`, `ttnn/cpp/ttnn/operations/**/device/*_op.hpp`, `tt-train/sources/**/device/*device*`
+  - `smuggled-buffer-runtime-arg` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*factory*`, `ttnn/cpp/ttnn/operations/**/device/*program*`, `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `ttnn/cpp/ttnn/operations/**/device/*_op.cpp`, `ttnn/cpp/ttnn/operations/**/device/*_op.hpp`, `ttnn/cpp/ttnn/operations/**/device/*fusion*`, `tt_metal/impl/program/**`, `tt-train/sources/**/device/*program_factory*`
+  - `override-rebuild-in-cache-hit` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*factory*`, `ttnn/cpp/ttnn/operations/**/device/*program*`, `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `ttnn/cpp/ttnn/operations/**/device/*_op.cpp`, `ttnn/cpp/ttnn/operations/**/device/*_op.hpp`, `ttnn/cpp/ttnn/operations/**/device/*fusion*`
   - `llk-stale-hw-config-state` — blocking, paths: `tt_metal/tt-llk/**/llk_lib/**`, `tt_metal/tt-llk/**/common/inc/**`, `tt_metal/hw/ckernels/**/llk_api/**`, labels: `area:llk`
   - `op-shard-layout-validation` — warning, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `ttnn/cpp/ttnn/operations/**/device/*_op.cpp`, `ttnn/cpp/ttnn/operations/**/device/*_op.hpp`, labels: `area:ops`
 
--- a/.github/bug_checker/README.md
+++ b/.github/bug_checker/README.md
@@ -154,6 +154,8 @@ python .github/bug_checker/run_bug_checker.py --branch --subcommand dry-run
 - Current rules:
   - `ccl-ring-buffer-mismatch` — blocking, paths: `tt_metal/impl/ccl/**`, `ttnn/cpp/ttnn/operations/ccl/**`, labels: `area:ccl`
   - `reshape-dim-check` — warning, paths: `ttnn/cpp/ttnn/operations/data_movement/**`, labels: `area:ops`
+  - `program-cache-hash-collision` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*device_operation*`, `tt-train/sources/**/device/*device*`
+  - `smuggled-buffer-runtime-arg` — blocking, paths: `ttnn/cpp/ttnn/operations/**/device/*program_factory*`, `tt_metal/impl/program/**`, `tt-train/sources/**/device/*program_factory*`
 
 ### LLM Analysis — `llm.py` : `LLMSession.analyze_rule()`
 
--- a/.github/bug_checker/manifest.yaml
+++ b/.github/bug_checker/manifest.yaml
@@ -26,7 +26,13 @@ rules:
     suggest_fix: true
     model: null
     paths:
+      # Modern convention (*_device_operation.cpp/hpp) plus the legacy
+      # *_op.cpp/hpp family, which still carries ~9 custom
+      # compute_program_hash() implementations (reshape_op, pool_op,
+      # rs_matmul_op, strided_all_gather_async_op, ...).
       - "ttnn/cpp/ttnn/operations/**/device/*device_operation*"
+      - "ttnn/cpp/ttnn/operations/**/device/*_op.cpp"
+      - "ttnn/cpp/ttnn/operations/**/device/*_op.hpp"
       - "tt-train/sources/**/device/*device*"
 
   smuggled-buffer-runtime-arg:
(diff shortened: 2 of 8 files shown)


bbradelTT:
I think these are worthwhile to check, and this needs someone from LLKs to look at it. @ncvetkovicTT would you be able to take a look?

GPT-5.6 Sol analysis:

High — [LLK teardown guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L67-L72) classifies valid code as buggy. Wormhole intentionally uses a no-op teardown because subsequent init paths reprogram transient state ([implementation](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h#L213-L223)).

High — [LLK stall guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L74-L81) incorrectly requires MATH | WAIT_SFPU universally. Blackhole deliberately uses only MATH for FPU-only fields ([source](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/llk_math_common.h#L184-L189)); Quasar also has a different instruction signature.

High — [op-shard-layout-validation globs](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/manifest.yaml#L90-L93) omit *program_factory* files—the exact code whose assumptions the rule examines. For example, typecast_sharded_program_factory.cpp directly dereferences shard specs but would not trigger the rule when changed alone.

High — [is_sharded() is presented as a sufficient guard](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/op-shard-layout-validation.md#L37-L40). It is not: ND_SHARDED returns true while using nd_shard_spec(), so the “good” examples can call .shard_spec().value() on an empty optional.

Medium — [apply_descriptor_runtime_args()](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/override-rebuild-in-cache-hit.md#L52-L56) is treated as inherently expensive. The repository explicitly supports minimal CB-only descriptors plus this call as an O(1) cache-hit repair.

Medium — The recommended CB lookup uses nonexistent cb.index; the actual index is under format_descriptors[*].buffer_index.

Low — Hash examples use nonexistent tt::stl::hash; the repository namespace is ttsl::hash.

blozano-tt:
Wow, if the agent made that many mistakes, I’m concerned. For background- I had my agent derive these rules from bug fix PRs that landed on main. Maybe it got the category of issue right but the guidance severely wrong?

ncvetkovicTT:
> I think these are worthwhile to check, and this needs someone from LLKs to look at it. @ncvetkovicTT would you be able to take a look?
> 
> GPT-5.6 Sol analysis:
> 
> High — [LLK teardown guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L67-L72) classifies valid code as buggy. Wormhole intentionally uses a no-op teardown because subsequent init paths reprogram transient state ([implementation](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h#L213-L223)).
> 
> High — [LLK stall guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L74-L81) incorrectly requires MATH | WAIT_SFPU universally. Blackhole deliberately uses only MATH for FPU-only fields ([source](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/llk_math_common.h#L184-L189)); Quasar also has a different instruction signature.
> 
> High — [op-shard-layout-validation globs](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/manifest.yaml#L90-L93) omit _program_factory_ files—the exact code whose assumptions the rule examines. For example, typecast_sharded_program_factory.cpp directly dereferences shard specs but would not trigger the rule when changed alone.
> 
> High — [is_sharded() is presented as a sufficient guard](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/op-shard-layout-validation.md#L37-L40). It is not: ND_SHARDED returns true while using nd_shard_spec(), so the “good” examples can call .shard_spec().value() on an empty optional.
> 
> Medium — [apply_descriptor_runtime_args()](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/override-rebuild-in-cache-hit.md#L52-L56) is treated as inherently expensive. The repository explicitly supports minimal CB-only descriptors plus this call as an O(1) cache-hit repair.
> 
> Medium — The recommended CB lookup uses nonexistent cb.index; the actual index is under format_descriptors[*].buffer_index.
> 
> Low — Hash examples use nonexistent tt::stl::hash; the repository namespace is ttsl::hash.

These are the supposed bugs that were found using the new bug_checker rules from this PR?

blozano-tt:
> > I think these are worthwhile to check, and this needs someone from LLKs to look at it. @ncvetkovicTT would you be able to take a look?
> > GPT-5.6 Sol analysis:
> > High — [LLK teardown guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L67-L72) classifies valid code as buggy. Wormhole intentionally uses a no-op teardown because subsequent init paths reprogram transient state ([implementation](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h#L213-L223)).
> > High — [LLK stall guidance](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/llk-stale-hw-config-state.md#L74-L81) incorrectly requires MATH | WAIT_SFPU universally. Blackhole deliberately uses only MATH for FPU-only fields ([source](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/llk_math_common.h#L184-L189)); Quasar also has a different instruction signature.
> > High — [op-shard-layout-validation globs](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/manifest.yaml#L90-L93) omit _program_factory_ files—the exact code whose assumptions the rule examines. For example, typecast_sharded_program_factory.cpp directly dereferences shard specs but would not trigger the rule when changed alone.
> > High — [is_sharded() is presented as a sufficient guard](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/op-shard-layout-validation.md#L37-L40). It is not: ND_SHARDED returns true while using nd_shard_spec(), so the “good” examples can call .shard_spec().value() on an empty optional.
> > Medium — [apply_descriptor_runtime_args()](https://github.com/tenstorrent/tt-metal/blob/fb5c6cfa6f08436079d10b0e2f794f6749c0ad42/.github/bug_checker/rules/override-rebuild-in-cache-hit.md#L52-L56) is treated as inherently expensive. The repository explicitly supports minimal CB-only descriptors plus this call as an O(1) cache-hit repair.
> > Medium — The recommended CB lookup uses nonexistent cb.index; the actual index is under format_descriptors[*].buffer_index.
> > Low — Hash examples use nonexistent tt::stl::hash; the repository namespace is ttsl::hash.
> 
> These are the supposed bugs that were found using the new bug_checker rules from this PR?

@ncvetkovicTT

No, The list of bugs to check in bug-checker was very small, so I used an agent to look back at 6 months of PRs labeled as "fix", and asked it to come up with classes of common bugs. It came back with these five classes. However, I question the guidance that was generated.

ncvetkovicTT (commented):
Looked at the LLK rule. The bug class is fine, the guidance isn't. Five things:

1. The bad `_llk_pack_untilize_uninit_()` is our real code, and it's correct. Packer x-start/x-end is transient and every init reprograms it (tt-llk#1036), `llk_pack_untilize.h:213`. Drop that pair. An empty uninit isn't a bug signal on its own.

2. `MATH | WAIT_SFPU` isn't always needed. BH uses `MATH` alone on purpose (`llk_math_common.h:184`), and Quasar's STALLWAIT takes 4 operands so the OR doesn't work there at all (`llk_math_reduce.h:45`). Reword to "drain whatever actually reads the field". The THCON-only bit is right, keep it.

3. `_llk_pack_relu_config_` takes a `ckernel::ReluConfig&`, not a uint32 (`llk_pack_common.h:200`).

4. The relu "good" example drops the stall and the `mutex::REG_RMW` the real one has (`llk_pack_common.h:211`), so it would trip point 5 of this same rule.

5. `_llk_unpack_to_dest_32b_block_` doesn't exist anywhere, and `_llk_pack_mop_config_` takes a `PackMode`, not a bool (`llk_pack.h:197`). If you want a real tracker-bypass example, `_invalidate_src_zero_flag_state_` in `cmath_common.h:106` is the actual one.

Separately, the globs at `manifest.yaml:79` miss `tt_metal/hw/inc/api/compute/**`, which is where the init/uninit pairs live.

Issue #20953, opened 2025-04-21
Investigate performance improvements to argmax op


nsorabaTT:
1. Look at adding writer kernels that use NOC0 in parallel to double the throughput.
2. Look at double buffering of inputs
3. Explore making `use_multicore` to be true by default
4. Explore automatic routing to single core from the program_factory when that is more efficient
5. Switch to single core implementation when user specified sub_core_grid has only one core.
6. Explore combining argmax and max via LLK for easier maintenance.

PR #54121, opened 2026-08-23
[Feature] [Blackhole] jit_build: opt-in Zve32f (RVV) plumbing for TRISC2 kernel compiles (enable_trisc2_rvv)

nkapreTT:
### PR Category
Feature

Relates to #20953. Base of the RVV stack: #54105 (ttnn.argmax use_rvv) and #54106 (fused LM-head argmax epilogue) rebase onto this PR and use the knob added here.

## Problem

Blackhole's TRISC2 (pack) RISC carries a Zve32f vector unit, but the kernel JIT builds every TRISC with `-mcpu=tt-bh-tensix` and no vector extension — there is no way for a kernel to request RVV code generation. The RVV kernels in #54105/#54106 could therefore not compile in upstream CI: all silicon results there were produced with a local out-of-tree wrapper around the sfpi g++ (which proved that the stock sfpi 7.70.0 compiler handles the codegen — the wrapper only appended flags).

## What this PR does

Adds a minimal, default-off, per-kernel opt-in that plumbs those exact flags through the JIT, replacing the out-of-tree shim:

- **`ComputeConfig::enable_trisc2_rvv`** (and `ComputeConfigDescriptor::enable_trisc2_rvv`, with a nanobind binding): opt-in per compute kernel, following the `fp32_dest_acc_en` / `dst_full_sync_en` / `opt_level` pattern.
- **`JitBuildSettings::get_trisc2_rvv_enabled()`** (default `false`) carries the knob into the JIT; `JitBuildState::export_target_recipe()` appends the flags **only** for the TRISC2/pack build state of an opted-in kernel. TRISC0/1, all firmware builds, and every non-opted kernel produce byte-identical compile commands.
- **`HalJitBuildQueryInterface::rvv_compile_flags()`** supplies the arch-specific flags (default: empty = unsupported). The Blackhole HAL returns:
  - `-march=rv32im_zmmul_zaamo_zba_zbb_xtttensixbh_zve32f` — exactly the ISA string `-mcpu=tt-bh-tensix` resolves to, plus `_zve32f`;
  - `-fno-lto` — with `-flto`, the RVV builtins are streamed as GIMPLE and re-expanded by the link-stage LTRANS units, which do not carry the vector `-march`, breaking code generation at link time (observed with sfpi 7.70.0). Compile-only: the link stays stock `-flto=auto`, the non-LTO pack object simply opts out;
  - `-fno-tree-vectorize -fno-tree-slp-vectorize` — the vector unit is reachable only through explicit intrinsics; the auto-vectorizers must not touch scalar kernel/LLK code;
  - `-Wno-error=array-bounds` — RVV intrinsic access through casted L1 pointers trips `-Warray-bounds` false positives at `-O3` under `-Werror`.
  Opting in on an arch whose HAL returns no flags (WH, Quasar) fails the compile with an explicit `TT_FATAL`.
- **Cache keying**: the knob enters the kernel compile hash (`ComputeKernel::config_hash()`, appended **only when enabled**, so default hashes/paths are unchanged) and the HAL flag string enters the build-state hash (a future flag change invalidates cached opted-in binaries).

Kernel-side, vector code is guarded by the compiler-provided `__riscv_vector` macro (plus `TRISC_PACK`), so the same source compiles to an empty/scalar kernel everywhere else.

## Default-off invariance

With the knob absent, compile recipes are byte-identical to main by construction (the flags are appended at recipe-export time only for opted-in kernels, and the conditional hash suffix is not emitted), and this was verified empirically. The same default-config compute kernel was JIT-compiled on this branch and on the merge-base (`e701eff0689`), each into a fresh isolated cache (mock Blackhole, stock sfpi 7.70.0):

- **Kernel compile hash identical**: `10869590292254366188` on both sides (so cached binaries stay valid across the upgrade).
- **Compile command lines identical** for all three TRISC compiles — the only textual difference is the temp-object filename, which embeds the compiling PID and differs between any two runs on any branch.
- **ELFs bit-identical**: disassembly (`objdump -d`) matches for TRISC0/1/2, and after `--strip-debug` (removing the DWARF strings that embed the per-run temp-object name) the ELF files are byte-for-byte equal.

## Tests (run in CI, no device required)

`tests/tt_metal/tt_metal/jit_build/test_trisc2_rvv.cpp` (unit_tests_jit_build, mock-Blackhole device, pure host-side JIT):

- **KnobOffRecipesCarryNoVectorFlags** — knob-off recipes for all three TRISCs carry no `zve32f` / `-fno-lto`.
- **KnobOnFlagsReachOnlyThePackRecipe** — knob-on: TRISC0/1 recipe cflags byte-equal to knob-off; TRISC2 gains exactly the flags; the JIT cache re-keys.
- **VaddKernelCompilesToVectorCode** — a trivial explicit-intrinsic vadd kernel (`test_kernels/compute/trisc2_rvv_vadd.cpp`) JIT-compiles end to end with the knob on; `objdump -d` of the TRISC2 ELF contains vector instructions (`vsetvli`), TRISC0/1 and the knob-off compile contain none. Gated on the bundled sfpi accepting the Zve32f march string (probed at runtime; older sfpi skips rather than fails).

## Silicon validation

The flags this PR plumbs are the ones the out-of-tree wrapper applied for every silicon result in #54105/#54106 (34/34 gated tests passing on BH P150), and the silicon results carry to the restacked stack by a three-part equivalence, each step checked directly:

1. The restack leaves every device kernel source bit-identical (`git diff` over the kernel dirs between the pre- and post-restack tips is empty).
2. The `DW_AT_producer` flag strings of a knob-compiled pack TU and of the actual silicon-run shim-compiled `argmax_rvv_tile_compute` TRISC2 ELF are **identical** (same compiler 7.70.0[847], same canonicalized `-march=rv32imf_zmmul_zaamo_zba_zbb_zve32f_zve32x_zvl32b_xtttensixbh`, same `-fno-lto -fno-tree-vectorize -fno-tree-slp-vectorize`).
3. The one mechanical difference — the wrapper appended the flags at the end of the command line, the knob appends them at the end of cflags — was compiled both ways on the same pack TU: the two objects are **byte-identical** (`cmp` clean).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/trisc2-rvv-toolchain)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/trisc2-rvv-toolchain)

### DIFF
--- a/tests/tt_metal/tt_metal/jit_build/sources.cmake
+++ b/tests/tt_metal/tt_metal/jit_build/sources.cmake
@@ -10,4 +10,5 @@ set(UNIT_TESTS_JIT_BUILD_SRC
     test_kernel_signature_parser.cpp
     test_named_ct_arg_map.cpp
     test_sync_build_steps.cpp
+    test_trisc2_rvv.cpp
 )
--- a/tests/tt_metal/tt_metal/jit_build/test_trisc2_rvv.cpp
+++ b/tests/tt_metal/tt_metal/jit_build/test_trisc2_rvv.cpp
@@ -136,6 +136,11 @@ bool elf_contains_vector_instructions(const std::string& elf_path) {
     return false;
 }
 
+}  // namespace
+
+// Fixture lives in the named namespace: gtest TEST_F classes derive from it
+// with external linkage, and an anonymous-namespace base trips
+// -Werror=subobject-linkage under gcc Unity builds (merge-queue build-sweeps).
 class Trisc2RvvMockBlackholeFixture : public MeshDispatchFixture {
 protected:
     // Mock mode must be registered BEFORE the base fixture opens its shared devices — and that
@@ -201,8 +206,6 @@ protected:
         return state.export_target_recipe(kernel.get()).cflags;
     }
 };
-
-}  // namespace
 
 TEST_F(Trisc2RvvMockBlackholeFixture, KnobOffRecipesCarryNoVectorFlags) {
     auto kernel = compile_rvv_kernel(/*enable_trisc2_rvv=*/false);
--- a/tests/tt_metal/tt_metal/jit_build/test_trisc2_rvv.cpp
+++ b/tests/tt_metal/tt_metal/jit_build/test_trisc2_rvv.cpp
@@ -11,21 +11,26 @@
 //    the flags are appended only for opted-in kernels at recipe-export time). With the knob
 //    on, only the TRISC2 recipe gains the flags; TRISC0/1 recipes are unchanged.
 //  - End-to-end compile: a trivial explicit-intrinsic vadd kernel JIT-compiles with the knob
-//    on, and its TRISC2 ELF disassembly contains vector instructions; the knob-off compile
-//    of the same source contains none. Gated on the bundled sfpi toolchain accepting the
-//    Zve32f march string (older sfpi releases skip rather than fail).
+//    on, and its TRISC2 ELF contains vector instructions (checked in-process by scanning the
+//    ELF's executable sections for RVV instruction encodings — no subprocess, no shell); the
+//    knob-off compile of the same source contains none. Gated on the bundled sfpi toolchain
+//    accepting the Zve32f march string (older sfpi releases skip rather than fail).
 //
 // Everything here runs without silicon: the fixture configures a mock Blackhole device, and
 // compilation is pure host-side JIT.
 
 #include <gtest/gtest.h>
 
-#include <array>
-#include <cstdio>
+#include <elf.h>
+
+#include <chrono>
+#include <cstdint>
+#include <cstring>
 #include <filesystem>
+#include <fstream>
 #include <memory>
-#include <optional>
 #include <string>
+#include <vector>
 
 #include <enchantum/enchantum.hpp>
 #include <tt-metalium/experimental/mock_device/mock_device.hpp>
@@ -38,6 +43,7 @@
 #include "impl/kernels/kernel.hpp"
 #include "impl/program/program_impl.hpp"
 #include "jit_build/build_env_manager.hpp"
+#include "jit_build/jit_build_utils.hpp"
 #include "llrt/rtoptions.hpp"
 
 namespace tt::tt_metal {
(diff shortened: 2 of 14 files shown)


nsextonTT (approved):

vtsilytskyiTT:
Why is it disabled by default? Our core supports vector instructions, but we do not inform compiler about them. Why?

nkapreTT:
Good question — default-off is deliberate, for three measured reasons rather than caution-in-general:

1. **Byte-identity blast radius.** Telling the compiler about the vector extension globally changes codegen for *every* TRISC2 kernel (auto-vectorization becomes legal on the pack thread), which would invalidate the binary of every existing kernel and require fleet-wide revalidation. This PR's invariance proof (default-off = byte-identical compile hash and stripped ELFs vs merge-base) only holds because the flag is per-kernel opt-in.
2. **Compiler-chosen vector code is demonstrably hazardous on this core today.** Two silicon-measured hazards from our bring-up: (a) GCC will happily build vlenb-scaled stack frames — an e16m8 dual-accumulator spilled a 4.4 KB frame against TRISC2's ~256 B stack and hung the core; and (b) scalar-float instructions the compiler may emit alongside (`fmul.s`/`fdiv.s`) hang the core outright. The RVV kernels behind the stacked PRs are disassembly-audited for frame/spill shape before shipping; auto-vectorized code wouldn't be.
3. **LTO interaction.** The vector march must be paired with `-fno-lto` on that TU (LTRANS re-expands the RVV builtins without the vector march at link time and miscompiles) — fine per-kernel, not something to impose globally.

Default-on is a plausible end-state once there are guardrails (an audited `-march` subset, frame-size checks, fleet revalidation) — happy to sketch that as a follow-up issue. The opt-in mirrors the existing per-kernel knobs (`fp32_dest_acc_en` etc.) and keys the JIT cache, so nothing existing moves.

arikTT (approved):

akerteszTT (approved):
The legacy `Program`-creation APIs (`host_api.hpp` free functions and the `ProgramDescriptor`, which is conceptually part of TTNN) will eventually be deprecated in favor of Metal 2.0.

We're refactoring the Metal 2.0 `ComputeHardwareConfig` at the moment and will propagate this change on your behalf.

sagarwalTT (approved):

abhullar-tt (approved):

Issue #51305, opened 2026-07-28
Per-region-DFB-sizing decision in the multi-core block tilize/untilize family


akerteszTT:
## Problem

The `_multi_core_block[_interleaved]` factories of:
 - tilize 
 - untilize 
 - tilize_with_val_padding 
 - untilize_with_unpadding 

all leverage the legacy capability to size a CB on a _per-node_ basis. Named DFBs do not (and should not) support this. In Metal 2.0, ID management is the framework's responsibility, not the user's. (We were planning to support ID re-use on disjoint node sets.)

Much worse, they all share a design that makes buffer size a _correctness property_. This is a horrible and wildly unexpected property for a CB. A FIFO's size is a performance property only. And a sync-free scratchpad-like entity likewise can tolerate any size greater than some minimum. These factories have cooked up a hybrid FIFO-scratchpad chimera that manages to be size-sensitive.

Porters are currently improvising the sizing, and one of them already shipped a silent data corruption in `experimental/quasar`. 

## Ask (for Metal 2.0 port)

Please refactor these factories as a pre-req Metal 2.0 porting.

**Options**:
 1. Split the DM kernels into separate kernels, each with its own appropriately sized DFBs. (No L1 footprint change, but adds more distinct kernels -- some dispatch overhead.) The tilize port in `experimental/quasar` used this solution.
 2. A more invasive refactor, converting the hybrid to pure FIFO or pure scratchpad. Doing so will eliminate the artificial size sensitivity. All DFBs can be sized the max required. (This "wastes" L1.... but absolutely nothing uses the "saved" L1 on the smaller footprint nodes anyways.)
 3. **(Nope)** A single DFB, sized to the LCM. Crude analysis suggests this approach runs out of L1.
 4. **(Nope)** a single DFB, sized to the max size. This is functionally incorrect due to the previously mentioned _insane_ size dependence.

## Notes

Either of option 1 or 2 will unblock the Metal 2.0 port. 
Option 1 is certainly easier.
I've checked that neither option will cause further difficulties for basic Quasar uplift.

bbradelTT:
This should not affect llama. Can be delayed until later. If anything these factories can be left out of initial porting.

abdulla-TT:
Fixed with #54140 #54392

PR #54140, opened 2026-08-24
[Metal 2.0] Per-width buffer sets for the tilize/tilize_with_val_padding block factories

abdulla-TT:
### Ticket
#51305 

### Summary
Prereq refactor for #51305, so the multi-core block factories of `tilize` and
`tilize_with_val_padding` can be ported to Metal 2.0.

Both factories were allocating up to four `(staging, input, output)` CB triples that reused the
same three buffer indices on disjoint core ranges at different sizes. Named DFBs can't express
per-node sizing. And the size isn't just a perf knob here: the reader fills a whole block with one
raw linear write from `get_write_ptr()`, and `cb_push_back` only wraps when the write pointer lands
exactly on `fifo_limit`, so a buffer whose size isn't an exact multiple of its block overruns the
next one instead of wrapping. That's why sizing every region to the max doesn't work.

The work split only ever hands out two block widths, so each factory now declares two buffer sets
instead of four colliding triples: `full` (`c_0`/`c_1`/`c_16`, sized `single_sub_block_size`) and
`cliffrow` (`c_2`/`c_3`/`c_17`, sized `single_block_size_cliff_row`). Each non-empty set gets its
own reader/writer pair over its own cores; compute stays split per region but binds the set
matching its block width. A core's set comes from the work split's own core assignment, and a
couple of TT_FATALs assert that a set's buffers are sized for exactly the width its cores are fed.
L1 usage is unchanged, since the two sets sit on disjoint core ranges.

`tilize_wh.cpp` and `reader_unary_pad_multicore_both_dims.cpp` now take their buffer indices as
compile-time args instead of hardcoding `c_0`/`c_1`/`c_16`. Both are only used by these two
factories and both call sites are updated.

### Notes for reviewers
The two ops refresh buffer addresses on a program-cache hit differently, which is why the diffs
aren't symmetric. tilize's factories define `override_runtime_arguments`, and per the adapter that
means `resolve_bindings` is skipped entirely, so the hook is the only thing that repoints
addresses. Emitting a second reader/writer pair meant extending it past kernel 0/1, otherwise the
extra pair keeps its cache-miss address and quietly reads the wrong buffer. tilize_with_val_padding
has no such hook, so `resolve_bindings` walks every kernel's bindings and the extra pair refreshes
on its own.

**Heads-up: this changes `ttnn.to_layout` semantics on the height-sharded path.** That path now
returns the unpadded logical shape, matching what `tilize_with_val_padding` already produces on the
interleaved path; before, it returned the tile-padded shape. It's the right fix and it's what the
`test_tilize_height_sharded_shapes` uneven cases cover, but anything downstream reading `.shape`
off a height-sharded `to_layout` result changes behaviour, so it deserves model-CI attention rather
than being waved through with the CB-sizing refactor. The `to_layout_op.cpp` change is otherwise
unrelated to #51305 and can be split out if you'd rather review it separately.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/tilize-val-padding-per-node-cb-size)

### DIFF
--- a/tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
+++ b/tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
@@ -906,10 +906,7 @@ def test_tilize_width_sharded_shapes(device, tensor_shape, num_cores, dtype):
     ["sharded_width_l1", "interleaved_dram_multicore", "single_core"],
 )
 def test_tilize_program_cache_addr_change(device, config):
-    """Program-cache hit path (override_runtime_arguments): re-running tilize on freshly
-    allocated inputs (different buffer addresses) must hit the same cached program and stay
-    correct. Guards the sharded CB-bound path (addresses ride on CBs) and the interleaved
-    buffer-address runtime args -- both re-derived from create_descriptor on every hit."""
+    """Program-cache hit path (override_runtime_arguments)"""
     torch.manual_seed(0)
 
     if config == "sharded_width_l1":
@@ -1206,3 +1203,57 @@ def test_tilize_uint8(device, shape):
     tt_input = ttnn.from_torch(torch_input, device=device, dtype=ttnn.uint8, layout=ttnn.ROW_MAJOR_LAYOUT)
     tt_output = ttnn.tilize(tt_input)
     assert_equal(torch_input, ttnn.to_torch(tt_output))
+
+
+@pytest.mark.parametrize("tensor_shape", [(1, 1, 32, 7328)])
+def test_tilize_block_two_pair_program_cache_addr_change(device, tensor_shape):
+    torch.manual_seed(0)
+    mem_cfg = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.DRAM)
+
+    device.enable_program_cache()
+    device.clear_program_cache()
+
+    keep_alive = []  # retain prior tensors so each iteration allocates at a NEW address
+    entries = None
+    for i in range(4):
+        torch_input = torch.rand(tensor_shape, dtype=torch.bfloat16)
+        tt_input = ttnn.from_torch(
+            torch_input, dtype=ttnn.bfloat16, layout=ttnn.ROW_MAJOR_LAYOUT, device=device, memory_config=mem_cfg
+        )
+        tt_output = ttnn.tilize(tt_input, memory_config=mem_cfg, use_multicore=True)
+        keep_alive += [tt_input, tt_output]
+        assert_equal(torch_input, ttnn.to_torch(tt_output))
+        if i == 0:
+            entries = device.num_program_cache_entries()
+        else:
+            assert device.num_program_cache_entries() == entries, "tilize must reuse the cached program on a hit"
+
+    assert entries == 1, "tilize should build exactly one program for a fixed config"
+    device.disable_and_clear_program_cache()
+
+
+# Blackhole sends non-sharded uint8 tilize to the block factory, the only way a 1-tile-wide tensor
+# reaches it. These get a single reader/writer pair: 2048/4096 rows build only full_set, 5120 rows
+# only cliffrow_set. Checks that lone pair is still re-pointed on a cache hit.
+@run_for_blackhole()
+@pytest.mark.parametrize("shape", [(1, 1, 2048, 32), (1, 1, 4096, 32), (1, 1, 5120, 32)])
+def test_tilize_uint8_tall_narrow_program_cache_addr_change(device, shape):
+    torch.manual_seed(0)
+    device.enable_program_cache()
+    device.clear_program_cache()
+
+    keep_alive = []  # retain prior tensors so each iteration allocates at a NEW address
+    entries = None
+    for i in range(4):
+        torch_input = torch.randint(0, 256, shape, dtype=torch.uint8)
+        tt_input = ttnn.from_torch(torch_input, device=device, dtype=ttnn.uint8, layout=ttnn.ROW_MAJOR_LAYOUT)
+        tt_output = ttnn.tilize(tt_input)
+        keep_alive += [tt_input, tt_output]
+        assert_equal(torch_input, ttnn.to_torch(tt_output))
+        if i == 0:
+            entries = device.num_program_cache_entries()
+        else:
+            assert device.num_program_cache_entries() == entries, "tilize must reuse the cached program on a hit"
+
+    assert entries == 1, "tilize should build exactly one program for a fixed config"
+    device.disable_and_clear_program_cache()
--- a/ttnn/cpp/ttnn/operations/data_movement/common/common.hpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/common/common.hpp
@@ -9,6 +9,12 @@
 
 #include "ttnn/tensor/types.hpp"
 #include "ttnn/tensor/tensor.hpp"
+#include "ttnn/operations/core/work_split/work_split_tilize.hpp"
+
+namespace tt::tt_metal {
+struct ProgramDescriptor;
+struct TileDescriptor;
+}  // namespace tt::tt_metal
 
 namespace ttnn::operations::data_movement {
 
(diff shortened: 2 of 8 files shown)


ttnn/cpp/ttnn/operations/data_movement/common/common.hpp
```diff
@@ -127,7 +180,7 @@ class MassagedOperation {
             if (p2(args...)) {
                 *t2_required = true;
             }
-            return *t1_required or *t2_required;
+            return *t1_required or * t2_required;
```
Aswinmcw, line 247: Stray reformat unrelated to this PR — `*t2_required` became `* t2_required` (clang-format re-inferring pointer alignment after the new forward declarations). Worth reverting so the hunk doesn't show up in `git blame` for MassagedOperation.

ttnn/cpp/ttnn/operations/core/to_layout/to_layout_op.cpp
```diff
@@ -182,14 +182,17 @@ Tensor to_layout_impl(
...
+                ttsl::SmallVector<std::array<uint32_t, 2>> padding(padding_rank, {0, 0});
+                padding[padding_rank - 2] = {0, padded_output_shape[-2] - output_shape[-2]};
+                padding[padding_rank - 1] = {0, padded_output_shape[-1] - output_shape[-1]};
                 TT_FATAL(!sub_core_grids.has_value(), "Pad OP does not currently support sub core grid");
                 tensor = ttnn::pad(tensor, padding, pad_value, true, std::nullopt);
-                return ttnn::tilize(tensor, output_memory_config, dtype, use_multicore_tilize);
+                tensor = ttnn::tilize(tensor, output_memory_config, dtype, use_multicore_tilize);
+                // ttnn::pad grows the logical shape to the padded height/width, so restore the tensor's
+                // logical shape to the true (unpadded) output shape while keeping the tile-aligned padded shape.
+                return ttnn::experimental::view(tensor, output_shape, padded_output_shape);
```
Aswinmcw, line 195: This is a user-visible semantics change to `ttnn.to_layout`: the height-sharded path now returns the unpadded logical shape (matching what `tilize_with_val_padding` produces on the interleaved path), where before it returned the tile-padded one. It's the right fix, but it isn't mentioned anywhere in the PR description, and anything downstream reading `.shape` off a height-sharded `to_layout` result changes behaviour — worth calling out explicitly so it gets model-CI attention rather than being reviewed as part of a CB-sizing refactor.

ttnn/cpp/ttnn/operations/data_movement/tilize/device/tilize_multi_core_block_program_factory.cpp
```diff
@@ -380,9 +431,19 @@ void TilizeMultiCoreBlockProgramFactory::override_runtime_arguments(
...
+    // This factory emits one (reader, writer) pair *per buffer set*, so there may be two pairs to
+    // re-point rather than one. Both must be patched: an unpatched pair keeps the address from the
+    // call that populated the cache, which on a later hit with new tensor storage reads or writes
+    // the wrong buffer with nothing to flag it. The count comes from the program itself — see the
+    // note in create_descriptor for why it must not be re-derived here.
+    const uint32_t num_pairs = tt::tt_metal::GetCommonRuntimeArgs(program, 0)[0];
+    const uint32_t src_addr = tensor_args.input_tensor.buffer()->address();
+    const uint32_t dst_addr = tensor_return_value.buffer()->address();
+    for (uint32_t pair = 0; pair < num_pairs; ++pair) {
+        patch_tilize_kernel_slot0(program, 2 * pair, src_addr);      // reader
```
Aswinmcw: `2 * pair` / `2 * pair + 1` hard-codes that the DM kernels occupy handles `0 .. 2*num_pairs-1` ahead of the compute kernels. Nothing enforces that ordering, and getting it wrong patches a compute kernel's slot 0 with a buffer address — silent corruption, exactly the failure mode this hook exists to prevent. Cheap guard: assert `num_pairs` is 1 or 2 and that the kernel being patched is a reader/writer before writing to it.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -919,6 +923,13 @@ def test_tilize_program_cache_addr_change(device, config):
         tensor_shape = (1, 1, 128, 128)
         mem_cfg = ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.DRAM)
         use_multicore = True
+    elif config == "interleaved_dram_block_two_pair":
+        # Wide enough that a full row of tiles will not fit in L1, so this routes to
+        # TilizeMultiCoreBlockProgramFactory, and uneven (229 tiles per row) so the work split
+        # produces a cliff row as well as full blocks.
+        tensor_shape = (1, 1, 32, 7328)
```
Aswinmcw: `(1, 1, 32, 7328)` is short-and-wide, so it exercises full + cliffrow (both sets present, `num_pairs == 2`). The opposite asymmetry is untested and reachable: for a tall, narrow tensor (e.g. `(1, 1, 4096, 32)`) `single_block_size` exceeds `width_tiles`, so `full_cores_per_row == 0` and `core_range`/`cliff_col_core_range` are both empty — `full_set` is empty, only `c_2`/`c_3`/`c_17` get allocated, the cliffrow reader becomes kernel 0, and `num_pairs == 1`. That's the path where the new index scheme is most likely to bite; worth one more param case.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -1199,3 +1203,54 @@ def test_tilize_uint8(device, shape):
...
+            entries = device.num_program_cache_entries()
+        else:
+            assert device.num_program_cache_entries() == entries, "tilize must reuse the cached program on a hit"
+
+    assert entries == 1, "tilize should build exactly one program for a fixed config"
+    device.disable_and_clear_program_cache()
+
+
+@run_for_blackhole()
+@pytest.mark.parametrize("shape", [(1, 1, 2048, 32), (1, 1, 4096, 32)])
```
Aswinmcw: Correcting my earlier suggestion — `(1, 1, 4096, 32)` does not reach the branch I described.

With `width_tiles == 1`, `num_blocks == height_tiles`, so `nblocks_per_core = ceil(height_tiles / grid_area)` is 1 for both 64 and 128 tiles on a 130-core Blackhole grid. `closest_square_larger_than_b(1, ...)` then returns `single_block_size == 1`, so `full_cores_per_row == width_tiles / 1 == 1` and `has_cliff_row` is false — these two shapes exercise the *full*-set-only single pair, the mirror of the case I meant.

`full_set.empty()` needs `single_block_size > width_tiles`, hence `nblocks_per_core >= 2`, hence `height_tiles > grid_area`. `(1, 1, 5120, 32)` (160 tiles) gives `single_block_size == 2`, `full_cores_per_row == 0`, cliffrow-only — and stays on that branch for any grid from 81 to 159 cores, so it is not sensitive to the BH SKU. Worth adding alongside the two you have rather than replacing them; both single-pair orientations are distinct.

Aswinmcw (approved):
LGTM. Re-checked every PR-touched file against the previous head — only `test_tilize.py` changed (the `(1, 1, 5120, 32)` case), the rest is the rebase onto main. All earlier points addressed; the only leftover is the `* t2_required` reformat in `common.hpp`, which is non-blocking.

VirdhatchaniKN (approved):

sjameelTT (approved):

bbradelTT (approved):

mcw-anasuya (approved):

Issue #31129, opened 2025-10-24
Improve documentation for ALiBi and attention mask handling in Falcon models for custom forward passes


Sadelarus:
Need to add #EPAM_DX tag

Under cunstruction

### Current Behavior
When porting Falcon-7B to TT-NN for custom inference on Wormhole/Blackhole hardware, the documentation lacks clear guidance on handling ALiBi (Attention with Linear Biases) and attention masks in custom forward passes. This leads to runtime errors such as:
- `RuntimeError: The size of tensor a (71) must match the size of tensor b (2048) at non-singleton dimension 1` (during ALiBi creation due to broadcasting mismatches).

### Steps To Reproduce
1. Install TT-Metal and load Falcon-7B via Transformers:
 ````python
 from transformers import AutoTokenizer, AutoModelForCausalLM
 import ttnn
 model = AutoModelForCausalLM.from_pretrained("tiiuae/falcon-7b", dtype=torch.bfloat16)
 tokenizer = AutoTokenizer.from_pretrained("tiiuae/falcon-7b")
 device = ttnn.open_device(device_id=0)
````
2. Tokenize input (e.g., batch=32, seq_len=2048):
```python
inputs = tokenizer(["text"] * 32, padding="max_length", truncation=True, max_length=2048, return_tensors="pt")
input_ids = inputs["input_ids"]
attention_masks_2d = inputs["attention_mask"]
tt_input_ids = ttnn.from_torch(input_ids, dtype=ttnn.uint32, device=device)
```
3. Custom forward with ALiBi and attention (simplified):
````python
#ALiBi creation (custom function needed, not documented for TT-NN)
alibi = some_custom_alibi_function(attention_masks_2d)  # Broadcasting mismatches
scores = ttnn.matmul(q, ttnn.transpose(k, -2, -1)) / ttnn.sqrt(head_dim)
scores = ttnn.add(scores, attention_masks_2d)  # Error: IndexError (expects 4D mask)
````
Example of issue:
  `RuntimeError: IndexError: too many indices for tensor of dimension 2`

### Expected Behavior
Handling ALiBi in TT-NN: Provide formula, code snippet for creation, and broadcasting to [batch, heads, 1, seq_len] with ttnn.expand and TILE_LAYOUT.
```python
def build_alibi_tensor(attention_mask, num_heads, dtype, device):
  batch_size, seq_length = attention_mask.shape
[...]

PR #54162, opened 2026-08-24
fix sdpa attention mask docs

ijankowskiTT:
### PR Category
Cleanup

### Summary
Nothing documented how a tokenizer's `[batch, seq]` mask becomes a tensor the
attention ops accept, and two published examples used random tensors as
attention masks (`torch.randn` in the multi-head attention tutorial,
`ttnn.rand` in `Softmax.md`). `Softmax.md`'s transformer example also sets
`is_causal_mask=True` with no mask, which raises.

Two documented shapes were wrong: prefill `attn_mask` had its last two dims
transposed, and decode `attn_mask` was documented as `[b x 1 x s x s]` where
the op requires `[1 x b x nh x s]`.

- Correct both shapes; document `is_causal`/`attn_mask` exclusivity and the
  `k_chunk_size` default.
- Add `llms.md` 2.4.3: three kinds of mask (causal, padding, tile padding),
  2D→4D conversion, why the fill value is finite not `-inf`, and a table of
  which op accepts which shape.
- Replace the random-tensor masks; fix the example that would crash.
- Extend three softmax `TT_FATAL`s to name the remedy, not just the constraint.

Relates to #31129.

### Notes for reviewers
- Every constraint in the new table was triggered on a Wormhole card.
- The two intermediate-dim asserts guard different paths:
  `scale_mask_softmax` vs `validate_on_program_cache_miss` for
  `scale_mask_softmax_in_place`, which is what `attention_softmax_` calls.
- **I could not establish why this codebase uses `-1e3`/`-1e5` over `-inf`.**
  I documented the fully-masked-row `NaN` as the reason and verified that
  mechanism, but not that it's the historical one. Please correct if wrong.

### DIFF
--- a/ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
+++ b/ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
@@ -305,7 +305,9 @@ void bind_sdpa(nb::module_& mod) {
 
     const auto* const doc =
         R"doc(
-        Causal scaled dot product attention. This API mimics the PyTorch API of the same name.
+        Causal scaled dot product attention. This API mirrors the PyTorch API of the same name, with one
+        difference: `is_causal` defaults to `true` here where PyTorch defaults to `false`, and `is_causal` and
+        `attn_mask` are mutually exclusive. Pass `is_causal=False` when supplying `attn_mask`.
         The implementation is FlashAttention-2."
 
         Accepts a `SDPAProgramConfig` which specifies the grid size and chunk tiles in the Q and K sequence lengths. The op parallelizes over `b`, `nqh`, and Q's `s` dimension.
--- a/ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
+++ b/ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
@@ -305,9 +305,8 @@ void bind_sdpa(nb::module_& mod) {
 
     const auto* const doc =
         R"doc(
-        Causal scaled dot product attention. This API mirrors the PyTorch API of the same name, with one
-        difference: `is_causal` defaults to `true` here where PyTorch defaults to `false`, and `is_causal` and
-        `attn_mask` are mutually exclusive. Pass `is_causal=False` when supplying `attn_mask`.
+        Causal scaled dot product attention. `is_causal` defaults to `true` and is mutually exclusive
+        with `attn_mask`, so pass `is_causal=False` when supplying a mask.
         The implementation is FlashAttention-2."
 
         Accepts a `SDPAProgramConfig` which specifies the grid size and chunk tiles in the Q and K sequence lengths. The op parallelizes over `b`, `nqh`, and Q's `s` dimension.
--- a/ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_device_operation.cpp
+++ b/ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_device_operation.cpp
@@ -110,7 +110,7 @@ void SdpaDecodeDeviceOperation::validate_on_program_cache_miss(
 
     if (!operation_attributes.is_causal) {
         if (tensor_args.attn_mask.has_value()) {
-            // Causal attention verification
+            // Non-causal: an explicit mask is permitted, so validate its shape and dtype
             const auto& mask_tensor = tensor_args.attn_mask.value();
             const auto mask_shape = mask_tensor.padded_shape();
             const auto mask_shape_unpadded = mask_tensor.logical_shape();
@@ -147,8 +147,10 @@ void SdpaDecodeDeviceOperation::validate_on_program_cache_miss(
                 mask_tensor.dtype());
         }
     } else {
-        // Uncausal attention verification
-        TT_FATAL(not tensor_args.attn_mask.has_value(), "Must not have attn_mask tensor for non-causal attention");
+        // Causal: causality is applied from cur_pos on device, so an explicit mask is rejected
+        TT_FATAL(
+            not tensor_args.attn_mask.has_value(),
+            "attn_mask must not be provided when is_causal=True. Pass is_causal=False to use an explicit mask.");
     }
 
     const auto& paged_geo = operation_attributes.paged_cache_geometry;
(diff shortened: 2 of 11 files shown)


ttnn/cpp/ttnn/operations/transformer/sdpa_decode/sdpa_decode_nanobind.cpp
```diff
@@ -36,7 +36,11 @@ void bind_sdpa_decode(nb::module_& mod) {
 
         Keyword args:
             is_causal (bool): whether the attention is is_causal. Defaults to `True`.
-            attn_mask (ttnn.Tensor, optional): the input tensor [b x 1 x s x s]. Defaults to `None`.
+            attn_mask (ttnn.Tensor, optional): the input tensor [1 x b x nh x s], where dim 2 is the padded
+                number of Q heads and dim 3 is the KV length. The KV length must be a multiple of
```
fplavecTT: Nit: rather than saying "dim 2" and "dim 3", it might be cleaner to say `nh` and `s`.

Actually, I see same identifiers are used above in the "Args" section, so we might as well specify meaning of all these identifiers above "Args" (including `b`, `dh`, and `nkv`).

bbradelTT (approved):
Looks ok.

A few nits from AI:
1. “Head dim padded to 32” is the wrong axis
Decode docs point at llama_vision_model.py and say it pads head dim to 32. The reshape pads head count (dim 2) to tile height:


llama_vision_model.py
Ln 1016–1019
tt_xattn_mask = ttnn.reshape(
    tt_xattn_mask,
    [S, B, self.configuration.n_heads // self.configuration.num_devices, tt_xattn_mask.shape[-1]],
    [S, B, 32, tt_xattn_mask.shape[-1]],
)
That matches the op (mask_shape[2] == q_shape[2]). Call it padded number of Q heads, not head dim.

2. Prefill mask batch is not always 1
§2.4.1 / the table give (1, n_q_heads or 1, seqlen, cache_len). The op also accepts batch b, and the unchanged nanobind line is still [b x nqh x s x s] with broadcast on batch and heads. Fine to say “in this LLM flow batch is 1”; the op table should not imply batch-1-only.

3. The 2D→4D snippet is not the decode layout
The new example builds [batch, 1, q_len, kv_len]. Decode wants [1, b, nh, s]. The table and the llama_vision pointer say that, but a reader can still paste the snippet into decode. One sentence after the snippet would close that.

4. k_chunk_size default is oversimplified
Unpaged decode uses get_chunk_size(s) (largest power-of-two divisor, capped at 512). Paged decode fills SDPAProgramConfig with k_chunk_size = 32 when the caller omits it, and non-causal paged requires k_chunk_size > 0. The new nanobind text treats “largest power-of-two divisor” as the universal default.

5. Nanobind still mentions PyTorch

sdpa_nanobind.cpp
Ln 305–308
Causal scaled dot product attention. This API mirrors the PyTorch API of the same name, with one
difference: `is_causal` defaults to `true` here where PyTorch defaults to `false`, and `is_causal` and
`attn_mask` are mutually exclusive. Pass `is_causal=False` when supplying `attn_mask`.
Review rules say not to define TTNN APIs in terms of PyTorch. The old “mimics PyTorch” line was already there; this expands it. The exclusivity / default-true part is the load-bearing content and can stand without the PyTorch comparison.

Suggested comments
On the decode attn_mask bullet in llms.md:

For a reference construction, see the `tt_xattn_mask` reshape in llama_vision_model.py,
which pads the per-device head count (dim 2) to 32, not the head dimension.
On the 2D→4D example:

This yields `[batch, 1, q_len, kv_len]`, which is the prefill / fused-softmax layout.
Decode needs `[1, b, nh, s]` (see the table below).
On decode nanobind k_chunk_size:


Unpaged: defaults to min(512, largest power-of-two divisor of KV length).
Paged: defaults to 32 via SDPAProgramConfig; non-causal paged requires an explicit
positive k_chunk_size. Mask dim 3 must be a multiple of that chunk size.

ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
```diff
@@ -305,7 +305,9 @@ void bind_sdpa(nb::module_& mod) {
 
     const auto* const doc =
         R"doc(
-        Causal scaled dot product attention. This API mimics the PyTorch API of the same name.
+        Causal scaled dot product attention. This API mirrors the PyTorch API of the same name, with one
+        difference: `is_causal` defaults to `true` here where PyTorch defaults to `false`, and `is_causal` and
+        `attn_mask` are mutually exclusive. Pass `is_causal=False` when supplying `attn_mask`.
```
gajanan-choudhary: Could you please duplicate this change in [ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa/sdpa_nanobind.cpp#L122](https://github.com/tenstorrent/tt-metal/blob/6cac6743651fc591feea8f5aec1d74085d94a79c/ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa/sdpa_nanobind.cpp#L122)?

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/sdpa_decode_nanobind.cpp
```diff
@@ -36,7 +36,11 @@ void bind_sdpa_decode(nb::module_& mod) {
...
+                `k_chunk_size`, which defaults to the largest power-of-two divisor of the sequence
+                length unless `program_config` overrides it. Only valid with `is_causal=False`.
+                Defaults to `None`.
```
gajanan-choudhary: Could you please duplicate this change at [ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa_decode/sdpa_decode_nanobind.cpp#L39](https://github.com/tenstorrent/tt-metal/blob/6cac6743651fc591feea8f5aec1d74085d94a79c/ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa_decode/sdpa_decode_nanobind.cpp#L39)?

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_device_operation.cpp
```diff
@@ -110,7 +110,7 @@ void SdpaDecodeDeviceOperation::validate_on_program_cache_miss(
 
     if (!operation_attributes.is_causal) {
         if (tensor_args.attn_mask.has_value()) {
-            // Causal attention verification
```
gajanan-choudhary, line 113: Please also duplicate this change in [ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa_decode/device/sdpa_decode_device_operation.cpp](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/experimental/quasar/transformer/sdpa_decode/device/sdpa_decode_device_operation.cpp).

gajanan-choudhary (approved):
LGTM other than outstanding requests

fplavecTT (approved):
Thanks for improving documentation!

mateusznowakTT (approved):

ijankowskiTT:
/ping @pavlejosipovic @dgomezTT @aczajkowskiTT 

thanks!

PR #54163, opened 2026-08-24
Combine op optimized for fabric 2d

dlukicTT:
### Summary
This is work in progress new approach to combine. For now merging as a separate op, but once Routed expert becomes a single op, this will become part of it.
[Op summary and perf comparison](https://claude.ai/code/artifact/a06262bb-12fe-436a-b5a8-925194caf53b)

[Blaze CI Green](https://github.com/tenstorrent/tt-metal/actions/runs/33429900602)
[BH e2e Green](https://github.com/tenstorrent/tt-metal/actions/runs/33429804565)
### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dlukic/cmb-f2d-take2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dlukic/cmb-f2d-take2)


kgrujcic-tt (approved):

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/combine_fabric2d_assignments.cpp
```diff
@@ -0,0 +1,120 @@
...
+                    .dst_chip_id = ring_chip_ids[row],
+                    .dst_row = row,
+                    .split_idx = split_idx,
+                    .split_count = split_count});
+            };
+
+            // Furthest destination first, with relays interleaved so downstream streams get work early.
+            // After the j-th own assignment the cumulative relay count is the triangular number
+            // T(j-2) = (j-2)(j-1)/2, which for m=4 gives own3 own2 relay0 own1 relay1 relay2 own0 relay3..5.
+            uint32_t relays = 0;
```
pmilojevicTT, line 91: This is current schedule plan for every chip in relative destinations 

Image
. Can descending order of destinations maybe bring more performance improvement (own d=4, own d=3, relay 0, own d=2, relay 1, relay 3, ...). I am just asking, don't have concrete clue that it works better.

pmilojevicTT:
Just to check, will arbitrary layout of worker Tensix cores that adapts to layout of Ethernet cores on specific machine be done in next PR or is it already implemented here?

pmilojevicTT (approved):

models/demos/deepseek_v3_d_p/tests/op_unit_tests/test_prefill_combine.py
```diff
@@ -260,26 +261,57 @@ def run_combine(
...
-        tt_expert_region_offsets,
-    )
+        tt_output = tt_combine(
+            tt_dispatched_buffer,
+            tt_dispatched_metadata,
+            tt_expert_token_counts,
+            tt_expert_region_offsets,
+        )
+    else:
+        tt_expert_offsets = ttnn.from_torch(
```
mbezuljTT, line 251: for awareness - this will require some changes in routing setup

models/demos/deepseek_v3_d_p/tests/op_unit_tests/test_prefill_combine.py
```diff
@@ -525,6 +557,70 @@ def test_ttnn_combine(
...
+        pytest.param(
+            mesh,
+            fabric_to_device_params(fabric_cfg),
+            _fabric_cfg_to_fabric_topo_for_cmb_op(fabric_cfg),
+            640,  # seq_len_per_chip
+            model_config.EMB_SIZE,
+            model_config.NUM_ROUTED_EXPERTS,
+            model_config.NUM_EXPERTS_PER_TOKEN,
+            8,  # dispatch_buffer_capacity_factor
+            False,  # run_pcc_check
```
mbezuljTT: Please turn on pcc check for non-perf runs

mbezuljTT (approved):
Thanks @dlukicTT  
please ensure we have at least one run in the ci with pcc check included and CI green. 

I'll forward my kernel audit agent comments in a moment.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/sender_combine_fabric2d.cpp
```diff
@@ -0,0 +1,193 @@
...
+    }
+
+    noc_async_writes_flushed();
+    fabric_connections.close();
+
+    // Both ring counters back to zero for the next launch, which starts its own counts at zero. Safe here
+    // and only here: the reader's last act was publishing the CMD_END slot this kernel has just drained, so
+    // nothing is still reading or bumping either of them.
+    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.filled_addr), 0);
+    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.freed_addr), 0);
```
mbezuljTT, line 196: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

`freed`'s last increment at `:138` is a NOC atomic, and nothing between it and this reset waits for it. `noc_async_writes_flushed()` at `:185` watches the non-posted *write* counter; atomics complete on `NIU_MST_ATOMIC_RESP_RECEIVED`, which only `noc_async_atomic_barrier()` / `noc_async_full_barrier()` observe (`dataflow_api.h:1852-1866`). `noc_semaphore_set` is a plain local store (`:1995-2000`), so it can be overtaken.

If there's no happens-before edge, the launch can end with `freed = processed`. These GlobalSemaphores outlive the program cache and nothing re-zeroes them — as the note at `_program_factory.cpp:143-145` says — so the next launch evaluates `claimed - *freed` at `reader:183` as `0 - processed`, wraps to ~2^32, and never clears. Reader never publishes, sender never bumps `freed`. Would show up on the launch *after* the corrupting one, which is what makes it look intermittent.

`filled` on the line above looks safe for a reason `freed` doesn't have: the sender only reaches here after `wait_for_filled` returned a value including the CMD_END increment, which is proof it landed. `freed` has no witness because the sender both increments and clobbers it.

```suggestion
    noc_async_atomic_barrier();
    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.filled_addr), 0);
    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.freed_addr), 0);
```

Off the hot path — `pump_stream` has finished — so no throughput cost. `drain_fabric` probably masks this most of the time, but it's skipped entirely when `sent == 0`.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
```diff
@@ -0,0 +1,510 @@
...
+
+void kernel_main() {
+    const Dram dram = open_dram();
+    Reader reader{dram, read_control_tables(dram)};
+
+    reader.run_schedule();
+    reader.run_local_phase();
+    reader.end_stream();
+
+    // Back to zero for the next launch, which starts its own count at zero. The upstream sender cannot bump
```
mbezuljTT, line 507: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

All three claims here check out — every chunk ends in a sentinel and the sender bumps unconditionally after one (`sender:105`), and `drain_fabric` does target `drain_sink_addr` rather than this semaphore (`sender:160-161`). The question is whether the conclusion follows.

What they establish is that **launch N's** upstream sender has nothing left to send. They don't seem to rule out a bump arriving, since `fwd_sem_addr` is a GlobalSemaphore at a mesh-uniform address that outlives the program, and launch **N+1**'s sender on that same upstream chip targets it identically. The argument may be self-undermining: premise 1 establishes that upstream is *finished*, and being finished is what lets its program complete and the next one dispatch.

```
A: launch N   last sentinel + final bump          <- premise 1 satisfied here
A: launch N   drain -> sink, not fwd_sem          <- premise 2 satisfied here
A: launch N   exits. Program complete.
A: -- launch N+1 dispatches --                    <- outside this comment's frame
A: launch N+1 forwards >=32 -> bumps B.fwd_sem
B: launch N   still in run_local_phase, reaches :509 -> set(0) -> bump destroyed
B: launch N+1 waits for C, can only ever see C-n -> hang
```

The window looks wide rather than tight, because a chip's last act is its slowest: `kernel_main` runs `run_schedule` -> `run_local_phase` -> `end_stream`, and the local phase is fully serialized (`:476-482`) and needs nothing from any peer. Local token counts differ per chip because routing is data-dependent, so some skew seems expected.

The semaphore may not be the only thing exposed in that window either — upstream's launch N+1 would also be writing *payload* into the same forwarding-buffer chunk indices this reader may not have reached, and since chunk lengths are data-dependent the sentinel would land elsewhere. `drain_fabric` doesn't cover that: it proves packets reached the peer *chip*, not that the peer consumed them — which is what the DRAM forwarding buffer is there to decouple.

`combine` handled this with an explicit N-way exit barrier (`writer_combine.cpp:349-382`, separate semaphore from the entry barrier at `:241-258`, host side `combine_program_factory.cpp:1441-1447`). I couldn't find a rendezvous of any kind here — no `noc_semaphore_wait` in either kernel. The fabric-open handshake doesn't look like a substitute: `open_start`/`open_finish` address only `edm_noc_x/y`, the **local** router, and the EDMs are persistent across launches. `Synchronize` at `_program_factory.cpp:166` is cache-miss-only — which would make this comment true on launch 1 and false from launch 2, and would explain why a single-launch test wouldn't see it.

If it needs fixing, one option cheaper than copying the OG: only a chip's two upstream neighbours touch its buffer, so a per-stream neighbour handshake might be enough — signal upstream "ready to receive", wait for downstream's ready before the first `send_slot`, after `open_connections` since the signal travels over fabric. O(1) messages per chip instead of O(N). Two details probably worth taking from `combine` rather than rediscovering: entry and exit need separate semaphores (`writer_combine.cpp:114-116`), and releasing the barrier semaphore with `wait_min` + atomic decrement rather than `set(0)`, or the same reset hazard reappears one level up.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/combine_fabric2d_device_operation.cpp
```diff
@@ -0,0 +1,234 @@
...
+        "combine_fabric2d: num_routed_experts {} is fewer than experts_per_chip x ring extent = {}, so "
+        "there is not even one dispatch group",
+        num_routed_experts,
+        experts_per_group);
+    TT_FATAL(
+        num_routed_experts % experts_per_group == 0,
+        "combine_fabric2d: num_routed_experts {} must be divisible by experts_per_chip x ring extent "
+        "= {}",
+        num_routed_experts,
+        experts_per_group);
```
mbezuljTT, line 165: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

The three `num_routed_experts` checks here don't appear to imply the alignment the reader needs. `read_control_tables` (`reader:137-141`) reads each row into `control_tables_addr + r * num_routed_experts * 4`, and a DRAM read needs a 64-byte-aligned L1 destination — which the op treats as mandatory in three other places (`reader:48-50`, `_program_factory.cpp:83-85`, `:187-188`). That would require `num_routed_experts % 16 == 0`.

A config that passes everything here seems to violate it: `experts_per_chip = 2` (the struct default, `_types.hpp:22`) with ring extent 4 gives `experts_per_group = 8`, so `num_routed_experts = 8` satisfies all three. `row_bytes` is then 32 — row 1 lands at +32 and the `region` pointer at +160, both 32-byte aligned.

Direction is the opposite of what I'd have guessed: any multiple of 16 is fine at any magnitude (256, 288, 512), so DSv3 is safe. The exposure is small or odd configs — scaled-down meshes.

```cpp
    // The reader copies the control tensors row by row into L1 at a stride of
    // num_routed_experts * 4 B, and a DRAM read needs a 64-byte-aligned L1 destination.
    TT_FATAL(
        (num_routed_experts * sizeof(uint32_t)) % 64 == 0,
        "combine_fabric2d: num_routed_experts {} must be a multiple of 16 - the control-table row "
        "({} B) is the L1 stride for a DRAM read, which must be 64-byte aligned",
        num_routed_experts,
        num_routed_experts * sizeof(uint32_t));
```

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
```diff
@@ -0,0 +1,510 @@
...
+    // this again: we only got past the last chunk by reading its sentinel, so everything it owes us has
+    // arrived, and its drain targets a sink address rather than this semaphore.
+    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.fwd_sem_addr), 0);
```
mbezuljTT, line 510: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

Exit hygiene: the last `flush_publish()` at `:173` issues a NOC atomic and the kernel returns without waiting for it.

Doesn't look like a live defect, unlike the `freed` case in the sender — the sender consumes `filled` and only resets it after observing the CMD_END increment, so the value should be correct. But that's a two-kernel argument rather than a barrier, and it'd be free to close since everything is done by this point:

```suggestion
    noc_semaphore_set(reinterpret_cast<volatile tt_l1_ptr uint32_t*>(ct.fwd_sem_addr), 0);
    // Do not exit with `filled` increments still in the NIU.
    noc_async_atomic_barrier();
```

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/sender_combine_fabric2d.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+    volatile PACKET_HEADER_TYPE* hdr = slot_hdr(slot);
+    // Header first, THEN wait for the slot: building it while the EDM may still be busy is free overlap, and
+    // reversing the two costs ~8% of the bandwidth.
+    hdr->to_noc_unicast_write(tt::tt_fabric::NocUnicastCommandHeader{metadata->this_addr}, payload_bytes);
+    fabric.wait_for_empty_write_slot();
+    fabric.send_payload_without_header_non_blocking_from_address(ct.ring_addr + slot * ct.slot_stride(), payload_bytes);
+    // No flush per token: a slot's header is untouched until the ring wraps and the payload is flushed once
+    // per batch below, which is what lets token N+1 issue while N is still draining. This leans on the NoC
+    // keeping the payload write ordered ahead of the EDM slot-credit write that
```
mbezuljTT: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

The mechanism described here is real — payload and header go out on `write_cmd_buf` via `send_chunk_from_address` -> `noc_async_write` (`edm_fabric_utils.hpp:72`), while the credit rides `write_at_cmd_buf` via `noc_inline_dw_write` in `update_edm_buffer_free_slots` (`edm_fabric_worker_adapters.hpp:686-689`). Two details in the comment look off though, and might send someone chasing a non-bug:

- *"the production idiom flush-blocks here instead"* — the non-blocking form seems to be a production idiom too. `perform_payload_send<blocking=false, flush=true>` (`ttnn/cpp/ttnn/operations/ccl/common/kernels/minimal_ccl_common.hpp:26-41`) issues exactly this pair and is the default all_gather / reduce_scatter / neighbor_pad take. `FLUSH_BLOCKING` looks like it exists to protect the *source* buffer, which you already do once per batch at `:136`.
- *"the sync cmd buf"* — that appears to be the stateful path (`setup_stateful_send_cmd_bufs`). This kernel takes the non-stateful branch, which uses `write_at_cmd_buf`.

The ordering itself looks fine: same source NIU, same destination node (the EDM core), same VC (`NOC_UNICAST_WRITE_VC` for both). Might be worth rewording so it reads as "this is the standard non-blocking send" rather than a known risk being accepted.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
```diff
@@ -0,0 +1,510 @@
...
+// The forwarding-buffer page layout is deliberately [token][final_addr][dst_chip], which is EXACTLY the
+// first token_size+16 bytes of a ring slot. Re-forwarding is therefore a single DRAM read of
+// token_size+16 bytes straight into the slot base: payload and both metadata words land where the sender
+// already expects them, with no unpacking.
+//
+// The ring is hand-rolled rather than a metal CB because the packet headers live at fixed offsets from the
+// L1 allocator base — where a CB would be allocated — and a sender on another chip addresses them there.
+//
+// Two monotonic single-writer counters, each bumped by a NoC atomic to our OWN core (the proven idiom for
+// cross-RISC visibility on one core; a plain store can sit in a write buffer where the other RISC will not
```
mbezuljTT: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

This rationale might invite the conclusion that the slot metadata needs the same treatment as the counters, and someone could eventually "fix" it on that basis.

`cb_push_back` signals through a stream register (`get_cb_tiles_received_ptr` -> `STREAM_REG_ADDR`, `stream_io_map.h:24-27`) while announcing tile data written to L1 by plain stores — so every CB in the codebase rests on the ordering this paragraph casts doubt on. The variant here looks stronger if anything: the atomic can't be issued until the NIU command-register stores retire, and those sit behind the metadata stores in the same in-order store stream, after which the atomic takes a full NOC round trip. The sender's `invalidate_l1_cache()` in `wait_for_filled` covers the read side.

Maybe a clause saying why the payload is treated differently from the counters — as written the caveat reads as applying to the whole slot.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/combine_fabric2d_program_factory.cpp
```diff
@@ -0,0 +1,361 @@
...
+// Never initialised and never read back: pure staging for tokens passing through a chip. One page per token,
+// and the page is token + forwarding metadata so a single fabric write lands both. Fused into ONE page rather
+// than split across payload and metadata regions precisely because nothing outside the op reads it — so the
+// "one page = one token" property the caller's regions must keep does not apply here, and it saves a DRAM
+// read and a DRAM write per forwarded token.
+//
+ForwardingBuffer allocate_forwarding_buffer(
+    ttnn::MeshDevice* mesh, const CombineFabric2dParams& args, const CombineFabric2dInputs& tensor_args) {
+    ForwardingBuffer fwd;
+    fwd.pages_per_chunk = fwd_pages_per_chunk(args.seq_len_per_chip, args.num_experts_per_tok);
```
mbezuljTT, line 185: _Notes from my agent's sync audit — parking these to think about, happy to discuss tomorrow if any of them land._

Sizing invariant that might be worth pinning down, mostly aimed at the densification refactor flagged at `:61-62`.

`reader:419-421` clamps a relay read to `fwd_pages_per_chunk - i`. If a chunk could ever fill with no room for its sentinel, that clamp returns zero readable pages and `do_forward_chunk`'s `while (!at_end)` spins with no diagnostic. Unreachable today — `ppc = seq_len * ept + 1` against a worst case of `seq_len * ept` — but it's true by construction of that one function, and "make the buffer dense" is the change that would drop the headroom, likely written by someone thinking about space rather than terminators.

Both quantities are known on the host, so this probably belongs here rather than as a device `ASSERT` (watcher-only, `assert.h:17-48` — compiles to nothing in a release build):

```cpp
fwd.pages_per_chunk = fwd_pages_per_chunk(args.seq_len_per_chip, args.num_experts_per_tok);
// The reader walks a chunk until it hits the sentinel, clamping reads to pages_per_chunk
// (reader:419-421). If a chunk could ever fill with no room for its terminator, that clamp
// yields zero readable pages and the relay loop spins forever with no diagnostic. Keep one
// page of headroom over the worst-case chunk, whatever the sizing formula becomes.
TT_FATAL(
    fwd.pages_per_chunk > args.seq_len_per_chip * args.num_experts_per_tok,
    "combine_fabric2d: a forwarding chunk holds {} pages but the worst case is {} tokens; at least "
    "one spare page is required for the sentinel that terminates the chunk",
    fwd.pages_per_chunk,
    args.seq_len_per_chip * args.num_experts_per_tok);
```

One thing that probably *isn't* the answer on the device: breaking out of the loop on `k == 0` would trade the hang for silent corruption, since the downstream reader counts chunks positionally and dropping one shifts every chunk after it on every chip downstream.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/combine_fabric2d_assignments.cpp
dlukicTT: It can in some cases. We are close to max eth BW and thus to max perf on test cases we currently have, so I left it like this, but we should definitely revisit once this runs in real scenario (overlap with routed expert and better models and unbalanced traffic test coverage)

dlukicTT:
> Just to check, will arbitrary layout of worker Tensix cores that adapts to layout of Ethernet cores on specific machine be done in next PR or is it already implemented here?

One of the next PRs

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
dlukicTT: Genuine issue, but would require non-trivial change. I have few follow up PRs which will be changing same parts of code so I'll add the fix there.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/combine_fabric2d_program_factory.cpp
dlukicTT: Genuine claim but it will be fixed properly in the follow up PR which switches logic to dense forwarding buffers and eliminates the padding and sentinels.

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
```diff
@@ -0,0 +1,513 @@
...
+            // `batch` tokens' reads in flight at a time, then one barrier and one announcement.
+            for (uint32_t q = base; q < end;) {
+                const uint32_t k = (end - q) > ct.batch ? ct.batch : (end - q);
+                for (uint32_t j = 0; j < k; j++) {
+                    issue_token(dst_chip_id, q + j, q + j - base, chunk_page + staged);
+                    if (!is_direct(dst_chip_id)) {
+                        staged++;
+                    }
+                }
+                noc_async_read_barrier();  // every token of the ct.batch is in L1 before any is announced
```
bbradelTT, line 259: Please use device 1.1 / 2.0 APIs

bbradelTT (approved):

ttnn/cpp/ttnn/operations/experimental/deepseek_prefill/combine_fabric2d/device/kernels/dataflow/reader_combine_fabric2d.cpp
dlukicTT: Hey, I saw this only after the PR got merged this morning. I have several follow up items to work on this op, so I'll incorporate this in a follow up. Here is the issue which tracks it - https://github.com/tenstorrent/tt-metal/issues/55038

PR #54166, opened 2026-08-24
[Bug fix] ternary: restore pack data format after LLK row broadcast

iklemenskiTT:
### Ticket

Addresses items 3 and 4 of #51218 — one defect, present in two kernels.

### What's wrong

Each per-operand LLK row-broadcast block reconfigured the packer to its broadcast CB and never restored it. Pack data format is sticky packer state, so the result computed below the block was packed into `c_3` using the *broadcast* CB's format.

The broadcast CBs (`c_4`/`c_5`/`c_6`) carry the input dtypes while `c_3` carries the output dtype, and those differ whenever a preallocated output tensor changes it. `is_llk_bcast()` inspects only the input dtypes, so a FLOAT32 output on BFLOAT16 inputs still selects the broadcast path. The result is bf16-formatted, half-length tiles written into a wider buffer.

### What changed

Purely additive — six call sites across `ternary_addc_ops_fpu_rowbcast.cpp` and `ternary_sfpu_row_bcast_ttt.cpp`, each broadcast block gaining a restore on the way out:

```cpp
   dfb_in1.pop_front(num_tiles_per_cycle);
+  pack_reconfig_data_format(/*old*/ dfb_llk_b.get_id(), /*new*/ dfb_out.get_id());
+#ifdef ARCH_BLACKHOLE
+  PACK((llk_pack_hw_configure<DST_ACCUM_MODE>(dfb_out.get_id())));
+#endif
```

`llk_pack_hw_configure` pairs with the restore in all ten binary_ng broadcast siblings — arch-gated in the FPU kernels (`eltwise_binary_row_bcast.cpp:67-70`), unconditional in the SFPU ones (`eltwise_binary_sfpu_row_bcast.cpp:94-95`). Each kernel here mirrors its own sibling class. It is not redundant with `compute_kernel_hw_startup`: `eltwise_binary_row_bcast.cpp:44` performs the identical startup and still issues it.

The unpack formats get no matching restore, matching the siblings. That is safe only while `is_llk_bcast()` admits a single input dtype, so the requirement is now stated on `is_llk_bcast()` itself and at the top of each loop rather than left implicit.

Two documentation gaps closed alongside: `addcmul`/`addcdiv`/`where` never mentioned that a preallocated `output_tensor` may carry a different dtype than the inputs (only `lerp` did), and the `reconfig_data_format` tech report never recorded that these configurations are sticky state a temporary retarget must hand back.

### Tests

`test_addc_row_bcast_preallocated_output_dtype` (addcmul/addcdiv) and `test_ttt_row_bcast_preallocated_output_dtype`, in `test_ternary.py`.

- **All three ops sharing the SFPU kernel.** `overwrite_compute_kernel_name_and_defines` routes every non-addc TTT ROW_BCAST op to `ComputeRowBcastTTT`, so `where` and `snake_beta` carried the same defect as `lerp`. `where` is the heaviest consumer of that kernel and had no coverage here — the existing preallocated-output `where` test uses all-FLOAT32 inputs, so `is_llk_bcast()` returns false and it never reaches the kernel.
- **Four shapes.** One per broadcast position, plus one with two operands broadcast. That last is the only multi-block configuration the dispatch admits (all three `H==1` classifies as OUTER_BCAST instead) and the only one where one block's restore has to hold for the block after it.
- **BFLOAT16 output as the control** — same kernel, formats happen to agree.
- **Reference computed in fp32**, not from the bfloat16 operands, and the FLOAT32 cases additionally assert the result is not exactly bfloat16-representable everywhere. PCC against a bf16-rounded reference cannot tell an fp32 result from a bf16-precision one; that check can. It is skipped for `where`, which only selects among its inputs, so bf16-representable output is correct there.

Pre-existing coverage never passed a preallocated output whose dtype differed from the inputs', which is why this survived.

### A second, unrelated defect this exposed

`where` with a FLOAT32 preallocated output on a ROW_BCAST shape returns garbage (PCC ≈ -0.03 to 0.003) **both before and after this change** — verified by running it with the kernel reverted to `main` and with the fix applied, identical failures either way. It is not the pack format restore, and it is not fixed here. Those four cases are `xfail`ed with that reason so the suite records the defect instead of hiding it. `where`'s BFLOAT16 output passes, so this is specific to the differing-dtype path.

### Sibling audit

`unary_bcast_init` appears in exactly two ternary compute kernels — the two fixed here. All ten binary_ng broadcast kernels already restore the pack format. Outside those trees, thirteen more kernels call it: the ten `experimental/quasar/binary_ng` copies all restore, and the three `experimental/bcast_to` kernels never touch `pack_reconfig_data_format`. No third offender.

The ternary column and scalar broadcast kernels are **not applicable** rather than unaudited: they pack exclusively into `dfb_out` and never retarget the packer (`ternary_sfpu_col_scalar_bcast_ttt.cpp:74`, `ternary_addc_ops_fpu_bcast.cpp:88`), relying on reader-side fill instead of an LLK broadcast into an intermediate CB. The defect cannot occur there.

### Follow-ups, not in this PR

- `bfloat8_b` as a preallocated output dtype reaches these kernels with no validation and no coverage — a third distinct pack format. Nothing in `validate_on_program_cache_miss` accepts or rejects it deliberately.
- The three broadcast CBs (`c_4`/`c_5`/`c_6`) are allocated for every TTT ROW_BCAST program, including when `is_llk_bcast()` rejected the LLK path and they are dead, and regardless of which operands actually broadcast. Up to three tile-sized CBs of L1 per core, which also tightens `num_tiles_per_cb` on that path.
- `ternary_sfpu_row_bcast_ttt.cpp` has no program-cache-hit coverage at all, while `override_runtime_arguments` patches output addresses by slot.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:iklemenskiTT/51218/ternary-rowbcast-packer-restore)

### DIFF
--- a/tech_reports/data_formats/reconfig_data_format.md
+++ b/tech_reports/data_formats/reconfig_data_format.md
@@ -60,6 +60,8 @@ The following DataFormat reconfigurations are currently supported:
 - `reconfig_data_format` API should be used to reconfigure the hardware between calls to operations that use CBs of different DataFormats.
 - `pack_reconfig_data_format` API is called independently of `reconfig_data_format`, when the output CB changes DataFormats
 - Programmers should always use the API calls providing both the old and the new operand CB index, as this enables faster reconfiguration and dynamic checks for eligible conversions.
+- Both configurations are sticky hardware state, not scoped to the call that follows. A kernel that retargets the packer at an intermediate CB — an LLK broadcast destination, an activation staging buffer — must hand it back to the output CB before packing its result, or the result is packed in the intermediate CB's format. Where those formats happen to agree the corruption is invisible, so it surfaces only once a caller changes the output dtype.
+- When passing both operands, the old index must be the CB the hardware is actually configured for. The two-operand calls skip the reconfiguration when the two formats match, so a stale or guessed old index silently does nothing.
 
 ## Examples:
 TO DO
--- a/tests/ttnn/unit_tests/operations/eltwise/test_ternary.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_ternary.py
@@ -489,3 +489,174 @@ def test_ternary_addcmul_cache_hit_refreshes_operand_addresses(device):
     assert_with_pcc(reference(2), ttnn.to_torch(out1).float(), 0.99)
 
     device.disable_and_clear_program_cache()
+
+
+# Same width, at least one operand with H==1 -> ROW_BCAST. One case per broadcast position so each of
+# the kernels' three per-operand LLK-broadcast blocks is exercised; one case with two operands
+# broadcast, which is the only multi-block configuration the dispatch admits (all three H==1
+# classifies as OUTER_BCAST instead) and the only one where a block's restore has to hold for the
+# block after it.
+ROW_BCAST_SHAPES = [
+    pytest.param((1, 1, 1, 128), (1, 1, 64, 128), (1, 1, 64, 128), id="bcast_a"),
+    pytest.param((1, 1, 64, 128), (1, 1, 1, 128), (1, 1, 64, 128), id="bcast_b"),
+    pytest.param((1, 1, 64, 128), (1, 1, 64, 128), (1, 1, 1, 128), id="bcast_c"),
+    pytest.param((1, 1, 1, 128), (1, 1, 1, 128), (1, 1, 64, 128), id="bcast_ab"),
+]
+
+ROW_BCAST_OUT_DTYPES = [ttnn.bfloat16, ttnn.float32]
+
+
+def _uniform_bf16(shape, lo, hi):
+    return torch.empty(shape, dtype=torch.bfloat16).uniform_(lo, hi)
+
+
+def _check_row_bcast_output(torch_output, out, out_dtype, computes_new_values):
+    """Assert a row-broadcast result, given a reference computed in fp32.
+
+    bfloat16 output is the control: same kernel, formats happen to agree. Packing in a broadcast CB's
+    bfloat16 format would write half-length tiles into a wider output buffer, so a format regression
+    collapses PCC rather than merely degrading it. `computes_new_values` additionally pins that a
+    FLOAT32 output really carries fp32 precision -- a result packed as bfloat16 would be exactly
+    bfloat16-representable everywhere. That check is meaningless for ops that only select among their
+    inputs, since bfloat16 inputs make bfloat16-representable outputs the correct answer.
+    """
+    assert out.dtype == out_dtype, f"expected {out_dtype} output, got {out.dtype}"
+    result = ttnn.to_torch(out).float()
+    assert_with_pcc(torch_output, result, 0.999)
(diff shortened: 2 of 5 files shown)


mateusznowakTT (approved):

Aswinmcw:
Heads-up: `test_addc_row_bcast_preallocated_output_dtype[ttnn.addcmul-FLOAT32-bcast_b]` is failing on main, first on the merge commit of this PR (d75e735) and on every push since. Deterministic — 3 attempts × 3 SKUs × 4 runs, PCC 0.358 on sim_wh_n150 and sim_bh_p150, 0.287 on wh_n300_civ2. Example run: https://github.com/tenstorrent/tt-metal/actions/runs/33515237470/job/99884154015

It is exactly one cell of the 16: the other three broadcast positions pass at FLOAT32, both addcdiv and the bfloat16 control pass, and the whole TTT matrix (including `lerp-fp32` at bcast_b) passes. So the packer restore itself is landing; something else is wrong for the addc FPU kernel with only B broadcast at a differing output dtype — the same shape as the `where`-fp32 defect this PR documented, rather than a regression in op behaviour (pre-PR all four FLOAT32 addc cells would have been wrong).

Worth noting `sanity-tests.yaml` never ran on the branch or in the merge queue — there are no eltwise-group check runs on c37285f, so this 16-cell matrix first executed post-merge.

Main is red until this is either fixed or the one cell carries a strict xfail with the reason, the way the `where` cases do.

iklemenskiTT:
@Aswinmcw Thank you for spotting this, I'm looking into it.

iklemenskiTT:
Fix seems correct on Blackhole but fails on Wormhole. Reverted for now, until full understanding is achieved. Filed https://github.com/tenstorrent/tt-llk/issues/1680 to track the problem.

Issue #51959, opened 2026-08-03
ttnn.topk: preallocated output indices dtype is not validated against the required index width


iwroszTT:
### Describe the bug

When `ttnn.topk` is given preallocated outputs via `output_tensor=(values, indices)`, the dtype of the indices tensor is never checked against the required index width. `compute_output_specs` simply returns the preallocated tensor config specs, bypassing the index-width decision made for calls where the output tensor isn't given (UINT16 only when the padded reduction dim fits in 16 bits *and* the values are not fp32), additionally `validate_on_program_cache_miss` only checks that the tensor is UINT16 or UINT32.

All checks below are from a single Blackhole p100a with `torch.manual_seed(0)`, `k=32`, `H=32`. The `idx_gather_mismatch` counts positions where `input[returned_index] != returned_value`.

**1. Preallocated UINT32 where UINT16 is required — the returned *values* are destroyed.** bf16 input at `W=64` requires UINT16:

```
prealloc UINT16 (correct)  ->  val_mismatch=   0/1024   idx_gather_mismatch=   0/1024
prealloc UINT32 (wider)    ->  val_mismatch=1018/1024   idx_gather_mismatch=1022/1024
```

**2. Preallocated UINT16 where UINT32 is required — indices are unrepresentable and wrap.** bf16 input at `W=131072` requires UINT32, because positions above 65535 do not fit in 16 bits:

```
prealloc UINT32 (correct)  ->  val_mismatch=0/1024   idx_gather_mismatch=  0/1024
prealloc UINT16 (narrower) ->  val_mismatch=0/1024   idx_gather_mismatch=502/1024
```

Roughly half the top-k positions of a random input fall above 65535 and wrap. With an input constructed so that *every* top-k element sits above position 65535, all 1024 returned indices equal the true index modulo 65536 — the first row expects `[65567, 65566, 65565, 65564]` and returns `[31, 30, 29, 28]`. Values are correct here; only the index tensor is wrong.

**3. fp32 + preallocated UINT16 is accepted and currently produces correct results, but likely only by accident.** 

### Steps to reproduce the issue

```python
import torch, ttnn

torch.manual_seed(0)
dev = ttnn.open_device(device_id=0)
k = 32

def run(W, idx_dt, label):
    shape, oshape = [1, 1, 32, W], [1, 1, 32, k]
    base = torch.randn(shape, dtype=torch.float32).bfloat16().float()
[...]

PR #54168, opened 2026-08-24
[Bug fix] ttnn.topk: honour optional indices_tensor and its index width

### LINKED ISSUE #36329: `blackhole-demo-tests (P150) / llama3-8b performance` test fails when using `indices_tensor` argument
### Problem

In `models/common/sampling/tt_sampling.py`, the optional `indices_tensor` parameter is passed to `ttnn.topk`.

For the **llama3-8b** model, the input tensor to this operation has:

* shape: `Shape([1, 1, 32, 64128])`
* dtype: `bfloat16`

For this combination, the **single-core TopK program factory** is selected. The `indices_tensor` parameter was originally added in PR #25728, but **only for the multicore implementation**. In this specific case, although `indices_tensor` was passed, it had no effect because the single-core path did not support it at the time.

Support for `indices_tensor` in the **single-core program factory** was later introduced in PR #36136. After exposing this parameter and enabling its use in the single-core path (which sampling relies on), the following CI failures started appearing:

```
FAILED models/tt_transformers/demo/simple_text_demo.py::test_demo_text[blackhole-1-device_params0-performance-ci-token-matching]
AssertionError: Top-1 accuracy 83.0% is too low (expected >=89.5%)
assert 83 >= 89.5

FAILED models/tt_transformers/demo/simple_text_demo.py::test_demo_text[blackhole-1-device_params0-performance-ci-eval-32]
AssertionError: Batch32 repeat batch outputs should be identical - 42 out of 64 consistency checks failed
assert False
```

This happens specifically on **blackhole-demo-tests (P150)** for **llama3-8b performance**, even though **unit tests for TopK pass**, indicating a correctness issue that is not covered by the current unit test suite.

### Root Cause (Suspected)

In the single-core TopK implementation, indices generation is controlled via the `GENERATE_INDICES` reader define. Currently, this define is **always enabled**, even when an external `indices_tensor` is provided - not to fail the test.

### Task

In
`ttnn/cpp/ttnn/operations/reduction/topk/device/topk_single_core_program_factory.cpp` (line ~151), change:

```cpp
const std::map<std::string, std::string> reader_defines = {
    {"GENERATE_INDICES", "1"},
};
```

to:

```cpp
const std::map<std::string, std::string> reader_defines = {
    {"GENERATE_INDICES", tensor_args.indices.has_value() ? "0" : "1"},
};
```

Then:

1. Re-run the affected llama3-8b demo tests on **P150 (Blackhole)**.
2. Inspect how `indices_tensor` is populated in a model.
3. Fix the indices generation logic if necessary to ensure:

   * Correct TopK accuracy.
   * Deterministic behavior across repeated batches when `indices_tensor` is provided.

### LINKED ISSUE #51921: topk: optional indices_tensor is silently ignored on the single-core path and has limited shape/dtype checks
### Describe the bug

`ttnn.topk(..., indices_tensor=...)` provides a logical index of the elements ranked during the topk. The parameter has no enforced contract, and the single-core program factory ignores it entirely. All findings below were reproduced on a single Blackhole p100a.

**1. Silently ignored on the single-core path.**  Note, this appears to be already tracked by https://github.com/tenstorrent/tt-metal/issues/36329. `topk_single_core_program_factory.cpp` hardcodes `GENERATE_INDICES = "1"`; Only `TopKMultiCoreProgramFactory` reads the tensor, so a caller on any single-core shape (`W < 8192`, non-power-of-two `W`, `W >= 65535`, `k > 64`, ...) silently gets the generated iota index. For example, passing a payload of all `777`:

```
W=64     (single-core)  -> returned indices [0, 1, 2, 3, ...], 777 absent   <-- payload discarded
W=16384  (multi-core)   -> returned indices all 777                         <-- payload honoured
```

**2. Index dtype is not validated against the index CB width, silently corrupting indices.** `reader_create_index_local_topk.cpp` copies `cb_in1.get_tile_size()` bytes per page. That size comes from the *output* index dtype chosen by `compute_output_specs`; `validate_on_program_cache_miss` only checks that it is UINT16 or UINT32. A UINT32 `indices_tensor` against a `UInt16` index CB reads half of each page:

```
bf16 W=16384, UINT16 iota (matching)   -> gather_mismatch=0/1024,    n_uniq_idx=1004   <-- correct
bf16 W=16384, UINT32 iota (mismatch)   -> gather_mismatch=1011/1024, n_uniq_idx=497    <-- wrong
```

Note the corrupted indices all stay within `[0, W)`, so a bounds check does not catch this — the op returns plausible-looking indices that do not address the values returned beside them. (The converse direction, fp32 values with a UINT16 `indices_tensor`, produces out-of-range garbage instead and is being handled in review of #50521.)

**3. No shape validation at all.** The reader indexes page `i * Wt + j` with the same arithmetic as the values accessor, so a wrongly-shaped tensor reads out of bounds instead of erroring. A half-width tensor is accepted silently:

```
input W=16384, indices_tensor W=8192  -> ACCEPTED, returned index range [9, 16350]
```

This has been already observed. See the fix note at `models/common/modules/sampling/sampling_1d.py:764-765`: *"TTTv1 used per_device_vocab here too, causing a 2x width mismatch between indices_tensor and logits in ttnn.topk on single-device."*

**4. It receives none of the input's transformations.** It is passed to `prim::topk`, while the values input is transposed to last-dim, normalized to 4D and padded to `W >= 64` / a tile multiple — and preallocated outputs *do* go through `pre_topk_transform_tensor` (`topk.cpp:357-360`). `fill_implicit_tile_padding` is likewise applied only to the values input (`topk.cpp:352`). So for `dim != -1` the values tensor is transposed and the indices tensor is not, with no error:

```
input [1,1,16384,32] dim=2, indices_tensor [1,1,16384,32] (untransposed) -> ACCEPTED, no error
```

Callers must replicate all of this by hand; both sampling modules pad their indices buffer with `-1` to compensate.

**5. Layout is not validated.** The `Layout::TILE` check at `topk_device_operation.cpp:154` covers only `input_tensor`. A ROW_MAJOR `indices_tensor` is accepted with no error. (Device residency *is* checked, by the generic op infrastructure at `ttnn/api/ttnn/device_operation.hpp:455`.)

**6. The output index dtype ignores the payload.** `compute_output_specs` derives it from input shape and input dtype alone, so labels above 65535 are unrepresentable no matter what the caller supplies:

```
bf16 W=16384, UINT32 payload with labels 100000..116383
  -> output index dtype = UINT16, returned values in [1, 50842]
```

**7. Documentation and test coverage.** The docstring is a single sentence and mentions none of the above, and the validation message `"Optional input tensor must be UINT16, or UINT32"` does not say which tensor it refers to. `run_topk_test` fills the buffer with `[:, :, :, i] = i` — byte-identical to the generated iota — so even the one shape that reads it cannot distinguish "read correctly" from "silently regenerated".

**A. Payload silently discarded on the single-core path (defect 1):**

```python
import torch, ttnn

for W in (64, 16384):                      # 64 -> single-core, 16384 -> multi-core
    shape = [1, 1, 32, W]
    x = ttnn.from_torch(torch.randn(shape, dtype=torch.bfloat16), ttnn.bfloat16,
                        layout=ttnn.TILE_LAYOUT, device=dev)
    idx = ttnn.from_torch(torch.full(shape, 777, dtype=torch.uint16), ttnn.uint16,
                          layout=ttnn.TILE_LAYOUT, device=dev)
    _, i = ttnn.topk(x, k=32, dim=-1, indices_tensor=idx)
    got = ttnn.to_torch(i).to(torch.int64)
    print(W, sorted(set(got.flatten().tolist()))[:4], 777 in set(got.flatten().tolist()))
ttnn.close_device(dev)
```

Observed: `64 [0, 1, 2, 3] False` then `16384 [777] True`.
Expected: both return `777`, or single-core raises a clear "unsupported" error.

**B. UINT32 indices_tensor silently corrupts indices (defect 2):**

```python
import torch, ttnn

torch.manual_seed(0)
W, k, shape = 16384, 32, [1, 1, 32, 16384]
base = torch.randn(shape, dtype=torch.bfloat16)
for ttnn_dt, torch_dt in ((ttnn.uint16, torch.uint16), (ttnn.uint32, torch.uint32)):
    x = ttnn.from_torch(base, ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=dev)
    iota = torch.arange(W, dtype=torch.int64).expand(shape).contiguous().to(torch_dt)
    it = ttnn.from_torch(iota, ttnn_dt, layout=ttnn.TILE_LAYOUT, device=dev)
    v, i = ttnn.topk(x, k=k, dim=-1, indices_tensor=it)
    got_v, got_i = ttnn.to_torch(v), ttnn.to_torch(i).to(torch.int64)
    bad = int((torch.gather(base, -1, got_i.clamp(0, W - 1)) != got_v).sum())
    print(ttnn_dt, f"gather_mismatch={bad}/{got_v.numel()}")
ttnn.close_device(dev)
```

Observed: `UINT16 gather_mismatch=0/1024`, `UINT32 gather_mismatch=1011/1024`.
Expected: either both correct, or the UINT32/`UInt16` pairing rejected.

### Implication

- `indices_tensor` is used in production sampling: `models/common/sampling/tt_sampling.py:655,688` and `models/common/modules/sampling/sampling_1d.py:535,573`.
- The index-width rule is presently reimplemented caller-side in three places with three different derivations: `tt_sampling.py::_select_topk_indices_dtype` (`:82-91`), `sampling_1d.py` hardcoding `ttnn.uint16` (`:785`), and the op itself. `tt_sampling.py` has already drifted — it picks the dtype from `padded_per_device`, but the tensor handed to topk is subsequently padded to `upper_power_of_2(...)`.
- **Fixing #36329 (defect 1) would unmask those caller-side skews**, so the two issues should be sequenced together.

Suggested direction:

- Add a single `required_index_dtype(input_tensor, dim)` helper in `topk_utils`, and have `compute_output_specs`, `validate_on_program_cache_miss` and both program factories derive from it instead of re-deriving it independently.
- Validate `indices_tensor` (and the preallocated indices tensor) against it exactly. Both narrower and wider mismatches are broken, not merely wasteful.
- Add shape and layout checks next to the existing dtype check.
- Either run `indices_tensor` through `pre_topk_transform_tensor` + `fill_implicit_tile_padding` like the values input, or document explicitly that the caller owns those transformations.

ign-sudha:
### PR Category

Bug fix

### Summary

- Closes #51921
- Fixes #36329

`ttnn.topk` accepted an `indices_tensor` and then ignored it on the single-core path, returning its own generated iota instead of the caller's labels. No error was raised, so callers got indices that did not correspond to the values they were handed. Three further gaps around the same parameter — index width, layout validation, and output dtype — are fixed alongside it.

### Root causes

**Defect 1** — the single-core reader pinned `GENERATE_INDICES` to `"1"`, so the kernel always took the generate branch and the supplied payload was never read. Only the multi-core factory honoured it.

**Defect 2** — the index circular buffer took its width from the output dtype alone. A 32-bit payload read into a 16-bit CB is paged at the wrong stride, so indices came back corrupted with no validation to catch it.

Defects 5–7 are independent validation and documentation gaps on the same parameter.

### Solutions by defect

| Defect | Change |
| --- | --- |
| 1 | `GENERATE_INDICES` is derived from `tensor_args.indices.has_value()`, matching the multi-core factory. |
| 2 | New `required_index_dtype()` holds the auto-width rule; the existing `resolve_index_dtype()` gains payload-driven widening; a `TT_FATAL` rejects a payload whose width does not match the resolved index dtype. |
| 5 | `Layout::TILE` is now required on `indices_tensor` and on both preallocated outputs, not just on the value input. |
| 6 | The output index dtype is derived from the payload, so labels above 65535 survive instead of being truncated. |
| 7 | The `indices_tensor` docstring documents the full contract; `TopK.md` drops the stale "multi-core is uint16-only" claim; error messages corrected; test payloads no longer alias the generated iota. |

Defects 3 (shape validation) and 4 (input transformations) were already fixed on main before this branch — `topk.cpp:434-452` requires the indices tensor to share the input's logical shape and rejects non-last-dim reductions, and `topk.cpp:587-598` pads it in lockstep with the input. No commit here.

### Tests

New:

- `test_topk_indices_tensor_payload_is_used` — a constant payload proves the labels are returned rather than regenerated, on both the single-core and multi-core factories.
- `test_topk_indices_tensor_dtype` — uint16 / uint32 / int32 payloads across both factories; asserts the output dtype and that gathering by the returned indices reproduces the values.
- `test_topk_indices_tensor_labels_above_uint16_max` — labels offset past 65535 survive.
- `test_topk_indices_tensor_too_narrow_raises` — a UINT16 payload is rejected when the input requires 32-bit indices.
- `test_topk_row_major_tensor_raises` — ROW_MAJOR `indices_tensor` and `output_tensor` rejected.

Changed:

- `run_topk_test` labels column `i` as `W - 1 - i` instead of `i`. The old payload was byte-identical to the iota the op generates on its own, so every existing parametrization passed whether or not the payload was read. With the reversed labelling those cases now fail if the payload is ignored.

### Notes for reviewers

- **Contract change:** a UINT32/INT32 `indices_tensor` now produces 32-bit output indices where the op would previously have auto-selected UINT16. Every in-repo caller passes uint16, so nothing downstream changes, but this is user-visible.
- The width check compares bit widths, not dtype names — UINT32 and INT32 are interchangeable for the non-negative positions topk produces.
- `resolve_index_dtype()` returns the payload dtype only when it is 32-bit; a UINT16 payload never narrows the resolved width.
- `required_index_dtype()` is new. It puts the auto-width rule in one place, though only one caller uses it so far.

### Follow-up

Preallocated output index dtype is still not validated against the required width — a UINT16 `output_tensor` indices buffer silently wraps when the input needs 32-bit indices. Tracked in #51959, will be a follow-up PR.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/sudha-51921-topk)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/sudha-51921-topk)

### DIFF
--- a/ttnn/cpp/ttnn/operations/reduction/topk/docs/TopK.md
+++ b/ttnn/cpp/ttnn/operations/reduction/topk/docs/TopK.md
@@ -28,7 +28,7 @@ The operation returns both:
 
 ### Usage Limitations
 
-- Supported index tensor types: `uint16`, `uint32` (a user-provided `indices_tensor` must be `uint32` when the input tensor is `float32`, since fp32 forces UINT32 index CBs)
+- Supported index tensor types: `uint16`, `uint32`, `int32` (a user-provided `indices_tensor` must be `uint32` when the input tensor is `float32`, since fp32 forces UINT32 index CBs)
 - Supported value tensor types: `bfloat16`, `bfloat8_b`, `float32`
 - Input tensor must be in **TILE** layout
 - Input shape must be 4D (after internal transformations)
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -504,6 +504,48 @@ def test_topk_indices_tensor_too_narrow_raises(device, expect_error):
         ttnn.topk(ttnn_input, k, dim=-1, indices_tensor=indices_tensor)
 
 
+@pytest.mark.parametrize("tensor_under_test", ("indices_tensor", "output_tensor"))
+def test_topk_row_major_tensor_raises(tensor_under_test, device, expect_error):
+    # topk is tile-only: the compute kernel sorts tiles and compute_output_specs hardcodes TILE.
+    k = 32
+    W = 64
+    shape = [1, 1, 32, W]
+
+    ttnn_input = ttnn.from_torch(
+        torch.randn(shape, dtype=torch.bfloat16), ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device
+    )
+
+    if tensor_under_test == "indices_tensor":
+        kwargs = {
+            "indices_tensor": ttnn.from_torch(
+                torch.arange(W, dtype=torch.int32).expand(shape).contiguous(),
+                ttnn.uint16,
+                layout=ttnn.Layout.ROW_MAJOR,
+                device=device,
+            )
+        }
+    else:
+        kwargs = {
+            "output_tensor": (
+                ttnn.from_torch(
+                    torch.zeros([1, 1, 32, k], dtype=torch.bfloat16),
+                    ttnn.bfloat16,
+                    layout=ttnn.Layout.ROW_MAJOR,
+                    device=device,
+                ),
+                ttnn.from_torch(
+                    torch.zeros([1, 1, 32, k], dtype=torch.bfloat16),
+                    ttnn.uint16,
+                    layout=ttnn.Layout.ROW_MAJOR,
+                    device=device,
+                ),
+            )
+        }
+
+    with expect_error(RuntimeError, "must be in tiled format"):
+        ttnn.topk(ttnn_input, k, dim=-1, **kwargs)
+
+
 @pytest.mark.parametrize("largest", [True, False])
 def test_topk_multicore_local_write_correctness(largest, device):
     """
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -463,6 +463,47 @@ def test_topk_indices_tensor_payload_is_used(W, device):
     assert torch.all(returned == payload), "indices_tensor was ignored; the op generated its own iota"
 
 
+@pytest.mark.parametrize("W", (64, 16384), ids=["single_core", "multi_core"])
+@pytest.mark.parametrize("index_dtype", (ttnn.uint16, ttnn.uint32), ids=["uint16", "uint32"])
+def test_topk_indices_tensor_dtype(W, index_dtype, device):
+    torch.manual_seed(0)
+
+    k = 32
+    shape = [1, 1, 32, W]
+
+    torch_input = torch.randn(shape, dtype=torch.bfloat16)
+    iota = torch.arange(W, dtype=torch.int32).expand(shape).contiguous()
+
+    ttnn_input = ttnn.from_torch(torch_input, ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device)
+    indices_tensor = ttnn.from_torch(iota, index_dtype, layout=ttnn.Layout.TILE, device=device)
+
+    values, indices = ttnn.topk(ttnn_input, k, dim=-1, indices_tensor=indices_tensor)
+
+    assert indices.dtype == index_dtype
+    indices = ttnn.to_torch(indices).to(torch.int64)
+
+    # Indices are the iota, so gathering from the input must reproduce the returned values.
+    assert_equal(torch.gather(torch_input, -1, indices), ttnn.to_torch(values))
+
+
+def test_topk_indices_tensor_too_narrow_raises(device, expect_error):
+    # W is past 65535, so the op resolves the index dtype to UINT32 and sizes the index CB 32-bit,
+    # but the payload here is UINT16. Reject rather than read a 16-bit tensor at a 32-bit stride.
+    k = 32
+    W = 2 * UINT16_MAX + 1
+    shape = [1, 1, 32, W]
+
+    ttnn_input = ttnn.from_torch(
+        torch.randn(shape, dtype=torch.bfloat16), ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device
+    )
+    indices_tensor = ttnn.from_torch(
+        torch.zeros(shape, dtype=torch.int32), ttnn.uint16, layout=ttnn.Layout.TILE, device=device
+    )
+
+    with expect_error(RuntimeError, "must be the same width as the output indices dtype"):
+        ttnn.topk(ttnn_input, k, dim=-1, indices_tensor=indices_tensor)
+
+
 @pytest.mark.parametrize("largest", [True, False])
 def test_topk_multicore_local_write_correctness(largest, device):
     """
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -448,25 +448,26 @@ def test_topk_indices_tensor_on_non_last_dim_raise(device, expect_error):
 
 @pytest.mark.parametrize("W", (64, 16384), ids=["single_core", "multi_core"])
 def test_topk_indices_tensor_payload_is_used(W, device):
-    # The payload is the constant 777 in every column, so the outcome is unambiguous: honoured means
-    # every returned index is 777, ignored means the op generated its own iota instead.
+    # Column i is labelled i + label_offset. A label below the offset means the payload was ignored;
+    # subtracting it recovers the column, so a wrong column is caught too.
     # W=64 takes the single-core factory, W=16384 the multi-core one.
     torch.manual_seed(0)
     k = 32
-    payload = 777
+    label_offset = 20000
     shape = [1, 1, 32, W]
 
-    ttnn_input = ttnn.from_torch(
-        torch.randn(shape, dtype=torch.bfloat16), ttnn.bfloat16, layout=ttnn.Layout.TILE, device=device
-    )
-    indices_tensor = ttnn.from_torch(
-        torch.full(shape, payload, dtype=torch.uint16), ttnn.uint16, layout=ttnn.Layout.TILE, device=device
-    )
+    torch_input = torch.randn(shape, dtype=torch.bfloat16)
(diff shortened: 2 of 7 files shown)


ttnn/cpp/ttnn/operations/reduction/topk/docs/TopK.md
```diff
@@ -28,15 +28,15 @@ The operation returns both:
...
 - Supported value tensor types: `bfloat16`, `bfloat8_b`, `float32`
 - Input tensor must be in **TILE** layout
 - Input shape must be 4D (after internal transformations)
 - The dimension to select top K from must have at least 64 elements (min_dim_per_core)
 - Combined batch dimensions (W × H × D) must be a multiple of 32
 - Sharded memory configuration is not yet supported
-- Multi-core implementation only supports K ≤ 64 (`multi_core_max_k`); index outputs may be uint16 or uint32/int32
-- Multi-core implementation requires a power-of-two width in [8192, 65535) (`multi_core_min_width`), **or**, for inputs with at most 2 tile rows (`multi_core_low_ht_max_tile_rows`), a power-of-two width in [1024, 65535) (`multi_core_low_ht_min_width`) — low-tile-row shapes leave the row-parallel single-core factory mostly idle, so they column-split instead (thresholds named in `topk_constants.hpp`, shared predicate `topk_multicore_structurally_eligible()` in `topk_utils.cpp`)
```
iwroszTT, line 39: Is this documentation change intentional?

ign-sudha: Not intentional - I resolved a merge conflict the wrong way. Fixed and pushed.

iwroszTT (approved):

PR #54169, opened 2026-08-24
Update input and output support documentation of eltwise binary ops

ijankowskiTT:
### PR Category
Cleanup

### Summary
The binary eltwise documentation describes input layout and memory support, but states that the tensor spec of the output is independent of A and B. That is not true of the dtype, which follows A, with additional per-operation rules on top.

This change documents the output dtype rules, the tile-layout-only restriction on BFLOAT8_B and BFLOAT4_B, the UINT8 to UINT16 cast applied by the relational operations, and that integer division with `rounding_mode=None` does not accept an output dtype override.

Relates to #29862.

### Notes for reviewers
The output dtype defaults to A's, from `output_dtype.value_or(dtype_a)` in `binary_ng_device_operation.cpp`, so a mixed float pair keeps A's dtype instead of promoting to the wider one: `add(bfloat16, float32)` returns BFLOAT16 while `add(float32, bfloat16)` returns FLOAT32. PyTorch returns float32 for both, so the result here depends on operand order. The existing promotion logic is deliberately scoped to the integer-by-float DIV and MUL cases where not promoting corrupts the result. The behaviour is unchanged by this PR; it is only written down. Each rule was also checked against a device.


ttnn/cpp/ttnn/operations/eltwise/binary/binary_nanobind.cpp
```diff
@@ -40,11 +40,11 @@ Tensor hypot_composite_wrapper(const Tensor& a, const Tensor& b, const std::opti
...
+        The output C also supports row major and tile layout, interleaved or sharded format (height, width or block sharded), in DRAM or L1. For operations that return a new tensor, the layout and memory configuration of C are independent of A and B. The memory configuration can be set with memory_config, and the full tensor spec with output_tensor where the operation takes one; if neither is given, the operation will attempt a best decision at an appropriate tensor spec. The dimensions of C, or of output_tensor if given, must match the broadcast-matched size of A and B. By default, C takes the dtype of A unless an operation-specific rule applies: comparison ops return 1 or 0 in the output dtype, and a mixed float pair follows A rather than the wider dtype. Where supported, the output dtype can be overridden by the dtype argument or output_tensor. In-place operations write into A, so C is A and keeps its layout, memory configuration and dtype.
```
mateusznowakTT: 1 or 0 in the output dtype definitely has some meaning / alias, that could be used here to explain.

ttnn/cpp/ttnn/operations/eltwise/binary/binary_nanobind.cpp
```diff
@@ -40,11 +40,11 @@ Tensor hypot_composite_wrapper(const Tensor& a, const Tensor& b, const std::opti
 
 // Common broadcasting and performance documentation for binary operations
 constexpr auto BINARY_BROADCAST_DOC = R"doc(
-        Binary elementwise operations, C=op(A,B), support input tensors A and B in row major and tile layout, in interleaved or sharded format (height, width or block sharded), in DRAM or L1. A and B are completely independent, and can have different tensor specs.
+        Binary elementwise operations, C=op(A,B), support input tensors A and B in row major and tile layout, in interleaved or sharded format (height, width or block sharded), in DRAM or L1. BFLOAT8_B and BFLOAT4_B are tile layout only. A and B are independent in layout and memory configuration, but not always in dtype. The note below lists the dtypes each operation accepts.
```
mateusznowakTT: I'd rather replace ```support input tensors A and B in row major and tile layout```
with ``` support input tensors A and B in tile and row major layouts (unless dtype requires specific layout, e.g. BFLOAT8_B or BFLOAT4_B).```
And then drop later statement on bfp8 and bfp4 layout requirement.

This ```A and B are completely independent, and can have different tensor specs.``` could be ```A and B are independent, and can have different tensor specs, with restrictions as in notes below.```

mateusznowakTT (changes requested):

mateusznowakTT (approved):

dchenTT (approved):

Issue #54016, opened 2026-08-21
[Bounty $35000] Welford Two-Pass Statistics Optimisation


jasondavies:
## Proposal

While optimising the equal-count Welford merge path, I identified and prototyped a broader Welford-path performance improvement based on stable two-pass statistics. The proposed work replaces the underlying statistics calculation used by the Welford paths for standalone variance and standard deviation, GroupNorm, and LayerNorm where the new algorithm is beneficial.

The implementation uses shifted two-pass statistics with FP32 accumulation:

```text
mean = shift + average(x - shift)
variance = average((x - mean)^2)
```

Centring the input before accumulation avoids cancellation for inputs with a large common offset. Separating mean and variance accumulation also removes the per-sample division and recurrence costs of online Welford.

## Scope

- Implement shifted two-pass statistics with FP32 accumulation for:
  - Standalone FP32 `var` and `std`.
  - GroupNorm.
  - LayerNorm, including relevant interleaved and sharded paths.
- Evaluate BF16 `var` and `std`, enabling two-pass statistics where they are faster without compromising numerical behaviour.
- Add data-reuse and eligible L1-replay optimisations so the second pass can reuse resident input data instead of rereading it from DRAM.
- Select between the existing reduction backend and two-pass statistics where required by performance or numerical behaviour.
- Validate population and sample variance, `var` and `std`, supported dtypes, reduction dimensions, fused normalisation inputs, sharding configurations, and representative low-variance or large-mean inputs.
- Perform Wormhole performance analysis and Blackhole functional, numerical, and performance validation.
- Document the algorithm, selector behaviour, numerical considerations, and revision-matched benchmark results for review.

## Deliverables

- Production implementations for the applicable `var`/`std`, GroupNorm, and LayerNorm statistics paths.
- Architecture-aware dispatch that enables the new path only where validated.
- Focused numerical regression tests, including translation stability and agreement with higher-precision references.
- Functional coverage across affected operation families and configurations.
- Before-and-after Wormhole and Blackhole benchmark summaries.
- Review documentation describing the implementation and its numerical and performance tradeoffs.

## Preliminary Performance

Current Wormhole prototype measurements show approximately:

| Operation family | Representative improvement |
|---|---:|
| Long FP32 `var`/`std` reductions | Up to approximately 11x |
| Representative sharded GroupNorm workloads | Approximately 1.9-2.5x |
| Representative LayerNorm workloads | Approximately 1.18-1.54x |

[...]

PR #54181, opened 2026-08-24
fix: Welford Two-Pass Statistics Optimisation

rafaio1:
## Summary
This PR implements the Welford Two-Pass Statistics Optimisation bounty (#54016). It replaces the online Welford algorithm with a stable two-pass statistics algorithm using FP32 accumulation for standalone `var` and `std`, GroupNorm, and LayerNorm operations.

The shifted two-pass approach centres the input before accumulation, avoiding cancellation for inputs with large common offsets. Separating mean and variance accumulation removes per-sample division and recurrence costs.

## Changes
- Implemented shifted two-pass statistics with FP32 accumulation.
- Added architecture-aware dispatch to select between the two-pass path and the existing reduction backend based on input characteristics (e.g., long FP32 reductions).
- Added L1-replay optimizations so the second pass can reuse resident input data instead of rereading from DRAM.
- Updated `var`, `std`, GroupNorm, and LayerNorm program factories to route to the new paths where validated.
- Ensured numerical stability for large-mean, low-variance inputs against FP64 references.

### DIFF
--- a/ttnn/cpp/ttnn/operations/normalization/common/device/kernels/two_pass_stats.cpp
+++ b/ttnn/cpp/ttnn/operations/normalization/common/device/kernels/two_pass_stats.cpp
@@ -1,0 +1,40 @@
+#include "two_pass_stats.hpp"
+#include <cmath>
+
+namespace tt::tt_metal::stats {
+
+void compute_shifted_mean_fp32(const float* input, float* mean, uint32_t size) {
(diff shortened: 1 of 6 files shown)


jberkowitzTT:
@rafaio1 - this is already assigned to another user, I'll close this out since we allow only one person to work on a bounty at a time

PR #54183, opened 2026-08-24
Fix reduce helper cross-call accumulation

### LINKED ISSUE #54177: [ttnn/kernel_lib] Reduce accumulation can deadlock when output and accumulator share a CB
## Component / Area

`ttnn/cpp/ttnn/kernel_lib` reduce compute helper

## Issue Type

Deterministic hang; the exact same workload can instead return bad output when only one additional CB page is available.

## Observed

Cross-call accumulation can use the accumulator CB as the output CB for intermediate `reduce()` calls. With a no-pop input policy, the helper bulk-reserves every output page before processing any output:

https://github.com/tenstorrent/tt-metal/blob/fd6676eeb1b38414c2ad7e0fc311a13c0682a551/ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl#L495-L504

That is unsafe when `output_dfb` and `accum_dfb` refer to the same CB. The previous call's output pages are still the next call's accumulator input, but PACK waits to reserve the complete next output block before those pages have all been reloaded and popped by UNPACK/MATH. The TRISC pipeline can enter a circular wait through the DEST synchronization points.

The reproducer is a BF16 `SUM / REDUCE_ROW` over a logical 9-by-32-tile matrix, split across four calls of 9-by-8 tiles:

- `ReduceInputPolicy::NoWaitNoPop`
- `Accumulate::at(cb_accumulator, iteration)` on all four calls
- Calls 0-2 write their 9 output tiles back to `cb_accumulator`
- Call 3 reads `cb_accumulator` and writes the final 9 tiles to the output CB
- The accumulator CB has 9 pages, exactly one page per output row

Observed capacity sensitivity:

| Accumulator CB capacity | Result |
|---:|---|
| `Ht` = 9 | Deterministic hang |
| `2 * Ht - 1` = 17 | Completes with incorrect output |
| `2 * Ht` = 18 | Passes |

The result must not depend on spare capacity in the accumulator queue.

## Expected

Output synchronization must be independent of input ownership policy. Each output tile should be reserved and pushed as it is packed, so prior accumulator pages can be consumed safely even when the accumulator and output CB alias.

An accumulator CB with exactly one page per output should be sufficient and should produce the same result as an over-provisioned CB.

## Steps to reproduce

1. Allocate CB0 for four contiguous 9-by-8-tile input blocks.
2. Allocate the accumulator CB with exactly 9 FP32 pages and the final output CB with 9 pages.
3. Publish all input pages once.
4. Call `reduce<SUM, REDUCE_ROW, ..., NoWaitNoPop>()` four times with `ReduceInputBlockShape::of(9, 8, 1)` and `Accumulate::at(cb_accumulator, iteration)`.
5. Use the accumulator CB as the output CB for iterations 0-2 and the final output CB for iteration 3.
6. Run with watcher/runtime assertions enabled.
7. Repeat with accumulator capacities 17 and 18 to observe the capacity-dependent behavior above.

Frequency: always on the tested configuration.

## Suggested acceptance criteria

- The 9-page accumulator configuration completes and returns the correct result.
- Output reserve/push behavior works when output and accumulator CBs alias.
- Cover `NoWaitNoPop` and `WaitUpfrontNoPop` with cross-call accumulation.
- Preserve correct behavior for the two popping policies.

## System details

- tt-metal: `fd6676eeb1b38414c2ad7e0fc311a13c0682a551` (`main`)
- tt-umd: `2bfa0563c13ad066c1bd7f76e20a92d4b2b9f251`
- OS: Ubuntu 22.04.5 LTS, kernel 5.4.0-212
- Hardware: Wormhole B0, N300; reproducer uses one Tensix core
- Firmware bundle: 18.12.1.0
- KMD: 2.4.1

## Regression information

Unknown. This was found while adding direct unit coverage for `reduce_helpers_compute.hpp`.

### LINKED ISSUE #54178: [ttnn/kernel_lib] REDUCE_COL accumulation reloads only one tile per DEST chunk
## Component / Area

`ttnn/cpp/ttnn/kernel_lib` reduce compute helper

## Issue Type

Bad outputs and an unbalanced accumulator CB during cross-call `REDUCE_COL` accumulation.

## Observed

`REDUCE_COL` processes several output columns concurrently in DEST: four columns on the FPU path and three on the SFPU path under the usual half-sync configuration. On a non-first accumulated call, every output in the current DEST chunk has a corresponding accumulator tile that must be restored.

The current reload helper always waits for, copies, and pops exactly one tile:

https://github.com/tenstorrent/tt-metal/blob/fd6676eeb1b38414c2ad7e0fc311a13c0682a551/ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl#L212-L240

`REDUCE_COL` invokes it only once per column chunk and does not pass `current_chunk`:

https://github.com/tenstorrent/tt-metal/blob/fd6676eeb1b38414c2ad7e0fc311a13c0682a551/ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl#L640-L654

Consequently, only accumulator tile 0 is copied into the base DEST register. The other `current_chunk - 1` outputs are not restored into consecutive DEST registers, and only one old accumulator page is popped. The tail chunk then begins from a stale accumulator page rather than its matching logical output.

A minimal failing case is:

- `SUM / REDUCE_COL`
- BF16 input, FP32 output (FPU path)
- `ReduceInputBlockShape::of(3, 5, 2)`
- Two `reduce()` calls using `Accumulate::at(cb_accumulator, iteration)`
- `WaitAndPopPerTile` input policy
- Input pages already arranged in the required chunked streaming order
- FPU column chunks are `4 + 1`

On the second call, the first chunk needs four accumulator tiles loaded into `DST[0..3]`; the helper loads only the first tile into `DST[0]`. This deterministically produces incorrect output. It also leaves old accumulator pages resident, which can interact with output-CB capacity and synchronization.

## Expected

For accumulated `REDUCE_COL`, reload the complete current output chunk:

1. Wait for `current_chunk` accumulator tiles.
2. Copy tile `i` into `accumulate.config.dst_index + i` for every `i` in the chunk.
3. Pop exactly `current_chunk` accumulator tiles.
4. Continue the reduction and pack the same number of output tiles in the same order.

## Steps to reproduce

1. Create two deterministic 2-batch, 3-by-5-tile inputs.
2. Run the first `SUM / REDUCE_COL` call into an accumulator CB.
3. Run the second call into the final output CB with `Accumulate::at(cb_accumulator, 1)`.
4. Compare against the host sum of both independently column-reduced inputs.
5. Inspect the accumulator CB balance after the call.

Frequency: always on the tested configuration.

## Suggested acceptance criteria

- Correct results for FPU full and tail chunks (`4 + 1`).
- Correct results for SFPU full and tail chunks (`3 + 2`).
- Correct CB wait/pop cardinality for every chunk and batch.
- Unit coverage for at least two calls, multiple outputs, multiple batches, and all input policies that support accumulation.

## System details

- tt-metal: `fd6676eeb1b38414c2ad7e0fc311a13c0682a551` (`main`)
- tt-umd: `2bfa0563c13ad066c1bd7f76e20a92d4b2b9f251`
- OS: Ubuntu 22.04.5 LTS, kernel 5.4.0-212
- Hardware: Wormhole B0, N300; reproducer uses one Tensix core
- Firmware bundle: 18.12.1.0
- KMD: 2.4.1

## Regression information

Unknown. This was found while adding direct unit coverage for `reduce_helpers_compute.hpp`.

malimpic-TT:
## Summary

- reserve and publish reduce outputs one tile at a time for every input policy, allowing the output CB to safely alias the cross-call accumulator CB
- reload complete `REDUCE_COL` accumulator chunks into consecutive DEST registers before continuing an accumulated reduction
- add a direct single-core reduce-helper test suite covering dimensions, input policies, accumulation, layouts, data formats, numerical modes, and boundary behavior
- add the 9×32-tile reduction regression, processed as four accumulated 9×8-tile calls with `NoWaitNoPop`

## Testing

- `./scripts/run_safe_pytest.sh --dev tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py -vv`
  - `97 passed in 53.65s`
## Issues

Closes #54177
Closes #54178

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:malimpic/reduce-output-per-tile-reservation)


ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl
```diff
@@ -217,11 +217,14 @@ ALWI constexpr uint32_t get_dst_index(const AccumulateT& accumulate) {
 
 template <PoolType reduce_type, ReduceDim reduce_dim, typename AccumulateT, bool is_sfpu = false>
 ALWI void reload_accumulator_if_needed(
-    DataflowBuffer& accum_dfb, uint32_t input_dfb_id, uint32_t scaler_dfb_id, const AccumulateT& accumulate) {
+    DataflowBuffer& accum_dfb,
+    uint32_t input_dfb_id,
+    uint32_t scaler_dfb_id,
+    const AccumulateT& accumulate,
+    uint32_t num_tiles = 1) {
```
s-jovic: nit: maybe rename to dest_chunk_size

malimpic-TT: Done

ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl
```diff
@@ -657,7 +618,7 @@ ALWI void reduce(
 
                 // Reload accumulator if needed (zero overhead when AccumulateT is NoAccumulation)
                 reload_accumulator_if_needed<reduce_type, reduce_dim, AccumulateT, is_sfpu>(
-                    accum_dfb, input_dfb_id, scaler_dfb_id, accumulate);
+                    accum_dfb, input_dfb_id, scaler_dfb_id, accumulate, current_chunk);
```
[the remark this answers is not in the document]
s-jovic: If copilot is correct let's add the test case for reduce col case we fixed

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
```diff
@@ -0,0 +1,574 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
```
s-jovic, line 1: Let's add these tests to some pipeline - i'd say Sanity if they are just a few and run quickly and nightly l2 otherwise

tests/pipeline_reorg/ttnn_sanity_tests.yaml
```diff
@@ -262,6 +262,18 @@
   team: ttnn
   owner_id: U084B1CES7M # Borys Bradel
 
+# The 98-case reduce helper suite takes about one minute on wh_n300_civ2 with a cold JIT cache.
```
s-jovic: nit: this comment will soon become obsolete

.github/time_budget.yaml
```diff
@@ -78,9 +78,9 @@ ttnn:
     sim_wh_n150: 40
     sim_bh_p150: 40
   sanity:
-    wh_n300_civ2: 545
+    wh_n300_civ2: 565  # +20 for ttnn llk helper library tests
     bh_p100: 545
-    bh_p150b_civ2: 545
+    bh_p150b_civ2: 565 # +20 for ttnn llk helper library tests
```
s-jovic: nit: the comments are informative just until the PR is merged

s-jovic (approved):

ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl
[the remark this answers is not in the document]
malimpic-TT: Added

ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl
```diff
@@ -272,15 +277,9 @@ ALWI void assert_input_dfb_size(uint32_t input_dfb_id, uint32_t tiles_per_bulk,
...
-    if constexpr (should_pop(input_policy)) {
-        // Per-tile reserve/push: only needs 1 page
-        ASSERT(get_dfb_num_pages(output_dfb_id) >= 1);
-    } else {
-        // Bulk reserve upfront: needs all outputs
-        ASSERT(get_dfb_num_pages(output_dfb_id) >= total_outputs);
-    }
+ALWI void assert_output_dfb_size(uint32_t output_dfb_id) {
+    // Outputs are reserved and pushed one tile at a time for every input policy.
+    ASSERT(get_dfb_num_pages(output_dfb_id) >= 1);
```
vtsilytskyiTT, line 282: After change this function doesn't contain any logic. Maybe worth to assert in place?

tests/ttnn/unit_tests/kernel_lib/reduce/kernels/reduce.cpp
```diff
@@ -0,0 +1,114 @@
...
+        REDUCE_RECONFIG_MODE,
+        REDUCE_FP32_MODE>(
+        shape,
+        layout,
+        accumulation,
+#ifdef REDUCE_POST_MULTIPLIER_BITS
+        PostReduceMultiply {}
+#else
+        compute_kernel_lib::NoOp{}
+#endif
```
vtsilytskyiTT: You have ifdefs in 2 places: `PostReduceMultiply` definition and here. Why? You could alias `PostReduceMultiply` as `compute_kernel_lib::NoOp{}` above, and then remove `#ifdef` here.

tests/ttnn/unit_tests/kernel_lib/reduce/kernels/reduce.cpp
```diff
@@ -0,0 +1,114 @@
...
+    cb_push_back(cb_input, input_tiles * num_calls);
+
+    for (uint32_t call = 0; call < num_calls; ++call) {
+        const bool is_last_call = call == num_calls - 1;
+        if (is_last_call) {
+            run_reduce_call<cb_output>(shape, layout, input_tiles, call);
+        } else {
+            run_reduce_call<cb_accumulator>(shape, layout, input_tiles, call);
+        }
+    }
```
vtsilytskyiTT:
```suggestion
    for (uint32_t call = 0; call < num_calls - 1; ++call) {
        run_reduce_call<cb_accumulator>(shape, layout, input_tiles, call);
    }
    run_reduce_call<cb_output>(shape, layout, input_tiles, call);
```
Looks simpler, isn't it?

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
```diff
@@ -0,0 +1,591 @@
...
+    if case.post_multiplier is not None:
+        defines.append(("REDUCE_POST_MULTIPLIER_BITS", hex(_float_bits(case.post_multiplier))))
+    if case.custom_scaler is not None:
+        defines.append(("REDUCE_CUSTOM_SCALER_BITS", hex(_float_bits(case.custom_scaler))))
+    return defines
+
+
+def _compute_config(case: ReduceCase) -> ttnn.ComputeConfigDescriptor:
+    config = ttnn.ComputeConfigDescriptor(
+        math_fidelity=ttnn.MathFidelity.HiFi3,
```
vtsilytskyiTT: I rarely see `HiFi3`. Usually it is 4 :)

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
```diff
@@ -0,0 +1,591 @@
...
+    result = ttnn.generic_op(
+        [device_input, output],
+        ttnn.ProgramDescriptor(kernels=kernels, semaphores=[], cbs=cbs),
+    )
+    actual = _meaningful_output(case, ttnn.to_torch(result))
+    return actual, _golden(case, logical_chunks)
+
+
+@pytest.mark.parametrize("case", ALL_CASES, ids=lambda case: case.name)
+def test_reduce_helpers_complete_input_space(device, case: ReduceCase):
```
vtsilytskyiTT, line 575: How long does the execution take? It feels like your `ALL_CASES` contains hundreds of arg permutations.

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
```diff
@@ -0,0 +1,591 @@
...
+    actual, expected = _run_case(device, case)
+    if case.input_dtype == "int32":
+        torch.testing.assert_close(actual.to(torch.int64), expected, rtol=0, atol=0, msg=case.name)
+    else:
+        rtol = 0.05 if case.calls > 1 or case.input_dtype == "bf16" else 0.02
+        torch.testing.assert_close(
+            actual.to(torch.float64),
+            expected.to(torch.float64),
+            rtol=rtol,
+            atol=0.1,
```
vtsilytskyiTT: Pretty big tolerance. Why? It might conceal a problem.

bbradelTT (approved):

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
malimpic-TT: ALL_CASES has 98 cases and finishes in just under a minute

vtsilytskyiTT: +1 minute for CI is a lot. If it runs in CI, worth to mark it with `NIGHTLY_` prefix. If it is only a manual test - may stay as is.

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
malimpic-TT: Lowered the tolerance

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
malimpic-TT: Moved to HiFi4

vtsilytskyiTT: I see you set 4. I don't think this matters in a test. I was just curious.

tests/ttnn/unit_tests/kernel_lib/reduce/kernels/reduce.cpp
vtsilytskyiTT: Addressed in [bcbe4e0](https://github.com/tenstorrent/tt-metal/pull/54183/commits/bcbe4e0bca6125508878ecbc3707f6174e508ac1), thanks

tests/ttnn/unit_tests/kernel_lib/reduce/kernels/reduce.cpp
vtsilytskyiTT: Addressed in [bcbe4e0](https://github.com/tenstorrent/tt-metal/pull/54183/commits/bcbe4e0bca6125508878ecbc3707f6174e508ac1)

ttnn/cpp/ttnn/kernel_lib/reduce_helpers_compute.inl
malimpic-TT: I would prefer to keep it as is, to stay aligned with the assert_input_dfb_size

tests/ttnn/unit_tests/kernel_lib/reduce/test_reduce_helpers.py
vtsilytskyiTT: Tighter tolerance is introduced in [bcbe4e0](https://github.com/tenstorrent/tt-metal/pull/54183/commits/bcbe4e0bca6125508878ecbc3707f6174e508ac1). Let's hope test passes.

dpopovTT (approved):

PR #54191, opened 2026-08-24
[Feature] Metal 2.0 null bindings for optional DFB, tensor, and scratchpad resources

riverwuTT:
### PR Category

Feature

### Summary

A Metal 2.0 `ProgramSpec` binds a fixed set of resources. Kernel source
is often written for different programs with varying optional resource
bindings. Metal 2.0 currently has no native way to express an optional
resource binding, which inhibits kernel source reuse: a binding that
does not attach to a host resource does not show up as a symbol on the
kernel, necessitating `#ifdef`s.

This PR provides native Metal 2.0 support for these optional resource
bindings.

### Host side changes

Introduces the concept of a **null binding**: a binding that does not
have a matching resource on the `ProgramSpec`.

```cpp
if (gamma) {
    spec.dataflow_buffers.push_back(
        DataflowBufferSpec{.unique_id = "post_gamma", /* … */});
    spec.tensor_parameters.push_back(
        TensorParameter{.unique_id = "post_gamma_t", .spec = gamma->spec});
}

reader.dfb_bindings = {
    ProducerOf("post_input", "inp"),
    ProducerOf("post_gamma", "gamma"),  // null binding if "post_gamma" is not declared
};
reader.tensor_bindings = {
    TensorBinding{.tensor_parameter_name = "post_input_t", .accessor_name = "src"},
    TensorBinding{.tensor_parameter_name = "post_gamma_t", .accessor_name = "gamma_src"},  // null binding if "post_gamma_t" is not declared
};
```

Before this PR, a binding had to refer to an existing resource. This PR
relaxes that requirement: a null binding is permitted, and codegen still
emits the accessor.

### Device side changes

A binding token represents a resource declared on the `ProgramSpec`.
When the binding is a null binding, codegen emits a dedicated
`NullXXXBindingToken` instead (`NullDFBBindingToken`,
`NullTensorBindingToken`, `NullScratchpadBindingToken`). The device can
query presence with `is_nullable_binding(binding)`. 
Construction of device side memory objects results in an assertion failure,
if assertion is not on, it's behavior is undefined.

```cpp
void kernel_main() {
    auto src = TensorAccessor(tensor::src);
    if (!is_null_binding(tensor::gamma.is_null)) {
       TensorAccessor ta(tensor::gamma);
    }

    // Compiles but will result in assertion failure.
    TensorAccessor ta(tensor::gamma);
}
```

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/m2-args)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/m2-args)


tt_metal/hw/inc/api/dataflow/dataflow_buffer.h
```diff
@@ -81,21 +82,43 @@ template <bool IsWrite, typename ReleaseFunc>
...
     uint16_t id_;
 };
 
+template <>
+struct DFBBindingToken<Null> {
+    constexpr DFBBindingToken() noexcept = default;
+
+    static constexpr bool is_null = true;
+    constexpr operator bool() const noexcept { return false; }
+};
```
riverwuTT: This special & ugly treatment is only needed for `DFBBindingToken` because it provides public getter and conversion function

riverwuTT:
effort shifted to #55100

PR #54199, opened 2026-08-24
[Cleanup] Route NoC address construction through an arch-selected backend

vvukomanovicTT:
### PR Category
Cleanup

### Summary
Introduces a single seam through which the shared device-side NoC address construction flows: a backend-neutral `noc_address_backend` namespace alias, resolved per architecture by include path (the same mechanism as `noc_parameters.h`). Both tt-1xx (WH/BH) and Quasar alias the legacy XY encoding (`noc_address_backend_xy`) today, so behavior is unchanged everywhere.

Rerouted call sites: the `get_noc_addr` family, the interleaved address generators (via one new `interleaved_addr_gen::get_bank_noc_addr<DRAM>`), `get_noc_addr_from_bank_id`, `get_system_memory_noc_addr`, the `is_local_addr` predicates, the `noc_traits` local-pointer extraction, and `calculate_dispatch_addr`.

Not yet rerouted (deliberately — no behavior may change in this PR): the dispatch kernels' direct PCIe compositions (`cq_prefetch.cpp`/`cq_dispatch.cpp`, which use the physical-coordinate frame via `NOC_X_PHYS_COORD`) and `mem_bench_kernel.cpp`. These are tracked for the dispatch-conversion PR later in the stack.

**This is the first preparation PR for enabling ATT on Quasar, which completelly changes how we construct addresses for NOC, for WH/BH this PR must not change the binaries and this was verified by comparing binaries with and without the chagne**

### Notes for reviewers
- Review contract: every `noc_address_backend_xy` function body is exactly the expression it replaced at its call sites.
- Verified by object-code comparison, not just inspection: kernels and firmware were compiled against main and this branch with the SFPI cross-compiler (kernels built the way JIT builds them: `brisck.cc`/`ncrisck.cc`/`dmk.cc` wrapper + generated `kernel_includes.hpp`), and the object files compared with `cmp`. Corpus: all 236 kernels under `tests/tt_metal/tt_metal/test_kernels/{dataflow,misc}` and `tt_metal/impl/dispatch/kernels`, plus a probe TU calling every rerouted entry point and firmware TUs. Kernels that need host-generated bindings don't compile standalone on main either and are excluded on both sides; every kernel that compiles on main compiles on this branch.

| Pass | Corpus | Identical | Mismatch |
|---|---|---|---|
| WH brisc `-Os`/`-O3` + ncrisc `-Os` | 236 kernels x 3 | 332/332 | 0 |
| BH brisc `-Os`/`-O3` + ncrisc `-Os` | 236 kernels x 3 | 335/335 | 0 |
| Quasar dm `-Os`/`-O3` | 236 kernels x 2 | 206/207 | 1 |
| Probe TU + firmware TUs, all arches, `-Os` and `-O3` | 12 configs | 12/12 | 0 |

- The single non-identical build (`dram_copy_to_noc_coord.cpp`, Quasar): the entire object diff is 4 instructions in inlined `calculate_dispatch_addr` where two registers swap which of `(v>>8)<<36` / `(v>>16)<<42` they carry before a commutative `or` combines them — same instruction multiset, provably the same result (GCC scheduling of two independent bitfield loads across the new function boundary; not pinnable from source, verified by experiment).
- TRISC compilation was additionally checked on WH and Quasar to confirm the backend header stays data-movement-only.
- Byte-identity dictated three signature details worth knowing: `dispatch()` takes `uint8_t` coordinates (preserves the `go_msg_t` bitfield integer promotions on RV64), `packed_worker()`/`dispatch()` take no unused `noc` parameter, and `system_memory()` only encodes (the bank-space arithmetic stays at the call site).
- `jit_build/fake_kernels_target/CMakeLists.txt` still uses the pre-`tt-1xx` include layout, so it cannot resolve any arch-selected header (pre-existing; unaffected here).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vvukomanovic/att-pr1-noc-address-backend)

### DIFF
--- a/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
+++ b/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
@@ -60,7 +60,7 @@ namespace noc_address_backend {
 [[maybe_unused]] static bool is_local(uint64_t noc_addr, uint8_t noc) {
     return NOC_UNICAST_ADDR_X(noc_addr) == my_x[noc] && NOC_UNICAST_ADDR_Y(noc_addr) == my_y[noc];
 }
-[[maybe_unused]] static uint64_t worker(uint32_t x, uint32_t y, uint32_t addr, [[maybe_unused]] uint8_t noc) {
+[[maybe_unused]] static uint64_t worker_address(uint32_t x, uint32_t y, uint32_t addr, [[maybe_unused]] uint8_t noc) {
     return NOC_XY_ADDR(DYNAMIC_NOC_X(noc, x), DYNAMIC_NOC_Y(noc, y), addr);
 }
 }  // namespace noc_address_backend
--- a/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
+++ b/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
@@ -54,6 +54,16 @@ static void host_assert(bool condition, const char* expression) {
 #define DYNAMIC_NOC_X(noc, x) (x)
 #define DYNAMIC_NOC_Y(noc, y) (y)
 #define NOC_XY_ADDR(x, y, addr) (static_cast<uint64_t>(x) << 32 | static_cast<uint64_t>(y) << 16 | static_cast<uint64_t>(addr))
+// Host mock of the device NoC address backend (see internal/dataflow/noc_address_backend_xy.h),
+// mirroring the macro stubs above.
+namespace noc_address_backend {
+[[maybe_unused]] static bool is_local(uint64_t noc_addr, uint8_t noc) {
+    return NOC_UNICAST_ADDR_X(noc_addr) == my_x[noc] && NOC_UNICAST_ADDR_Y(noc_addr) == my_y[noc];
+}
+[[maybe_unused]] static uint64_t worker(uint32_t x, uint32_t y, uint32_t addr, uint8_t noc) {
+    return NOC_XY_ADDR(DYNAMIC_NOC_X(noc, x), DYNAMIC_NOC_Y(noc, y), addr);
+}
+}  // namespace noc_address_backend
 #endif
 
 #include "api/tensor/tensor_accessor.h"
--- a/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
+++ b/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
@@ -55,7 +55,8 @@ static void host_assert(bool condition, const char* expression) {
 #define DYNAMIC_NOC_Y(noc, y) (y)
 #define NOC_XY_ADDR(x, y, addr) (static_cast<uint64_t>(x) << 32 | static_cast<uint64_t>(y) << 16 | static_cast<uint64_t>(addr))
 // Host mock of the device NoC address backend (see internal/dataflow/noc_address_backend_xy.h),
-// mirroring the macro stubs above.
+// mirroring the macro stubs above. Required even when no test calls these paths: qualified
+// names in TensorAccessor's member bodies are looked up when the template is parsed.
 namespace noc_address_backend {
 [[maybe_unused]] static bool is_local(uint64_t noc_addr, uint8_t noc) {
     return NOC_UNICAST_ADDR_X(noc_addr) == my_x[noc] && NOC_UNICAST_ADDR_Y(noc_addr) == my_y[noc];
--- a/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
+++ b/tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
@@ -60,7 +60,7 @@ namespace noc_address_backend {
 [[maybe_unused]] static bool is_local(uint64_t noc_addr, uint8_t noc) {
     return NOC_UNICAST_ADDR_X(noc_addr) == my_x[noc] && NOC_UNICAST_ADDR_Y(noc_addr) == my_y[noc];
 }
-[[maybe_unused]] static uint64_t worker(uint32_t x, uint32_t y, uint32_t addr, uint8_t noc) {
+[[maybe_unused]] static uint64_t worker(uint32_t x, uint32_t y, uint32_t addr, [[maybe_unused]] uint8_t noc) {
     return NOC_XY_ADDR(DYNAMIC_NOC_X(noc, x), DYNAMIC_NOC_Y(noc, y), addr);
 }
 }  // namespace noc_address_backend
--- a/tt_metal/hw/inc/api/tensor/tensor_accessor.h
+++ b/tt_metal/hw/inc/api/tensor/tensor_accessor.h
@@ -17,8 +17,6 @@
 
 #if defined(KERNEL_BUILD) || defined(FW_BUILD)
 #include "internal/dataflow/dataflow_api_addrgen.h"
-// Arch-specific (resolved by the HAL's per-arch include roots); host builds
-// (e.g. the accessor unit tests) provide a mock instead.
 #include "noc_address_backend.h"
 #endif
 
@@ -38,7 +36,7 @@ uint64_t get_dram_bank_base_offset(uint32_t bank_id, uint8_t noc) {
     uint32_t bank_offset_index = interleaved_addr_gen::get_bank_offset_index<true>(bank_id);
     uint32_t bank_index = interleaved_addr_gen::get_bank_index<true>(bank_id, bank_offset_index);
     uint32_t bank_offset = interleaved_addr_gen::get_bank_offset<true>(bank_index);
-    return interleaved_addr_gen::get_bank_noc_addr<true>(bank_index, bank_offset, noc);
+    return noc_address_backend::bank_address<true>(bank_index, bank_offset, noc);
 }
 #endif
 }  // namespace tensor_accessor
--- a/tt_metal/hw/inc/api/tensor/tensor_accessor.h
+++ b/tt_metal/hw/inc/api/tensor/tensor_accessor.h
@@ -302,7 +302,7 @@ private:
         auto bank_x = get_bank_x(packed_xy_coords[page_mapping.bank_id]);
         auto bank_y = get_bank_y(packed_xy_coords[page_mapping.bank_id]);
         auto bank_start = DSpec::is_dram ? tensor_accessor::get_dram_bank_base_offset(bank_x, noc)
-                                         : noc_address_backend::worker(bank_x, bank_y, 0, noc);
+                                         : noc_address_backend::worker_address(bank_x, bank_y, 0, noc);
         return bank_start + bank_base_address + (page_mapping.bank_page_offset * aligned_page_size) + offset;
     }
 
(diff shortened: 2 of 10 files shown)


vvukomanovicTT:
Extended the byte-identity verification from the original probe TU to a bulk sweep over real kernels, built exactly the way JIT builds them (the `brisck.cc`/`ncrisck.cc`/`dmk.cc` wrapper TU + a generated `kernel_includes.hpp` per kernel, dummy compile-time args, synthetic `chlkc_descriptors.h`), against unmodified main vs this branch:

**Corpus:** all 236 kernels under `tests/tt_metal/tt_metal/test_kernels/{dataflow,misc}` and `tt_metal/impl/dispatch/kernels`.
**Matrix:** WH/BH/Quasar × {-Os, -O3}, plus an NCRISC-wrapper pass on WH/BH (different cmd-buf selection), plus the original all-entry-points probe and firmware TUs.

**Results: 873 of 874 comparable kernel builds are byte-identical (`cmp` on the .o); zero compile regressions** (every kernel that compiles on main compiles here). Kernels that don't compile standalone on *main* (host-generated named-arg bindings, DFB descriptors, etc.) are excluded on both sides.

The sweep caught and fixed one real ordering regression: `get_noc_addr_from_bank_id` had its xy-table load and DRAM-offset load sequence-swapped; restored to the original load order (now byte-identical on all arches).

The single residual non-identical build is `dram_copy_to_noc_coord.cpp` on Quasar: the entire object diff is 4 instructions in inlined `calculate_dispatch_addr` where two registers swap which of `(v>>8)<<36` / `(v>>16)<<42` they carry before a commutative `or` combines them — same instruction multiset, provably the same result. It comes from the function-call boundary perturbing GCC's scheduling of two independent bitfield loads; source-level ordering cannot pin it (verified by experiment).

tt_metal/hw/inc/internal/dataflow/noc_address_backend_xy.h
```diff
@@ -0,0 +1,79 @@
...
+    return NOC_MULTICAST_ADDR(
+        DYNAMIC_NOC_X(noc, start_x),
+        DYNAMIC_NOC_Y(noc, start_y),
+        DYNAMIC_NOC_X(noc, end_x),
+        DYNAMIC_NOC_Y(noc, end_y),
+        local_address);
+}
+
+template <bool DRAM>
+FORCE_INLINE uint64_t bank(uint32_t bank_index, uint32_t local_address, uint8_t noc = noc_index) {
```
arikTT: The method doesn't returns a bank but an address, consider changing the name to something more describes the functionality more accurately

tt_metal/hw/inc/api/tensor/tensor_accessor.h
```diff
@@ -227,9 +225,7 @@ struct TensorAccessor {
 
     FORCE_INLINE
     bool is_local_addr(const uint64_t noc_addr, uint8_t noc = noc_index) const {
-        uint32_t x = NOC_UNICAST_ADDR_X(noc_addr);
-        uint32_t y = NOC_UNICAST_ADDR_Y(noc_addr);
-        return is_local_bank(x, y, noc);
+        return noc_address_backend::is_local(noc_addr, noc);
```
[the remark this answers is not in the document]
arikTT: Seems like this file is missing an include

tt_metal/hw/inc/internal/dataflow/noc_address_backend_xy.h
```diff
@@ -0,0 +1,79 @@
...
+
+FORCE_INLINE uint64_t worker(uint32_t x, uint32_t y, uint32_t local_address, uint8_t noc = noc_index) {
+    return NOC_XY_ADDR(DYNAMIC_NOC_X(noc, x), DYNAMIC_NOC_Y(noc, y), local_address);
+}
+
+FORCE_INLINE uint64_t local(uint32_t local_address, uint8_t noc = noc_index) {
+    return NOC_XY_ADDR(my_x[noc], my_y[noc], local_address);
+}
+
+FORCE_INLINE uint64_t packed_worker(uint32_t packed_xy, uint32_t local_address) {
```
abhullar-tt: on 2xx will there ever be a `packed_worker` or I guess there won't need to be a `worker` and only `packed_worker` will be used?

abhullar-tt (approved):

tt_metal/hw/inc/internal/dataflow/noc_address_backend_xy.h
vvukomanovicTT: renamed all constructors to *_address and the extractor to extract_local_address

tt_metal/hw/inc/api/tensor/tensor_accessor.h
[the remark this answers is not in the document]
vvukomanovicTT: fixed

tt_metal/hw/inc/internal/dataflow/dataflow_api_addrgen.h
```diff
@@ -224,9 +221,21 @@ std::uint64_t get_noc_addr_helper(std::uint32_t noc_xy, std::uint32_t addr) {
...
+// Full NOC address for `addr` inside bank `bank_index`. This is the single
+// place the bank -> NOC destination mapping happens for the address
+// generators below.
+template <bool DRAM>
+FORCE_INLINE uint64_t get_bank_noc_addr(uint32_t bank_index, uint32_t addr, uint8_t noc = noc_index) {
+    return noc_address_backend::bank_address<DRAM>(bank_index, addr, noc);
+}
+
+}  // namespace interleaved_addr_gen
+
```
vtsilytskyiTT: Worth to move it up, inside `interleaved_addr_gen` namespace together with other functions.

tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
```diff
@@ -54,6 +54,16 @@ static void host_assert(bool condition, const char* expression) {
...
+// Host mock of the device NoC address backend (see internal/dataflow/noc_address_backend_xy.h),
+// mirroring the macro stubs above.
+namespace noc_address_backend {
+[[maybe_unused]] static bool is_local(uint64_t noc_addr, uint8_t noc) {
+    return NOC_UNICAST_ADDR_X(noc_addr) == my_x[noc] && NOC_UNICAST_ADDR_Y(noc_addr) == my_y[noc];
+}
+[[maybe_unused]] static uint64_t worker_address(uint32_t x, uint32_t y, uint32_t addr, [[maybe_unused]] uint8_t noc) {
+    return NOC_XY_ADDR(DYNAMIC_NOC_X(noc, x), DYNAMIC_NOC_Y(noc, y), addr);
+}
+}  // namespace noc_address_backend
```
vtsilytskyiTT, line 67: Functions are not used, are they?

vtsilytskyiTT (approved):
Generally looks good, approved. Please ensure that sanity is green.

tt_metal/hw/inc/internal/dataflow/dataflow_api_addrgen.h
tt-vjovanovic: To be more honest, does this function need to exist at all? It only forwards call to `noc_address_backend::bank_address<>()` function, so is this temporary implementation and it will grow, or is it intended forwarder? If it is for forwarding, you should be able to define alias instead of adding new function.

tt_metal/hw/inc/api/tensor/tensor_accessor.h
```diff
@@ -17,6 +17,9 @@
 
 #if defined(KERNEL_BUILD) || defined(FW_BUILD)
 #include "internal/dataflow/dataflow_api_addrgen.h"
+// Arch-specific (resolved by the HAL's per-arch include roots); host builds
+// (e.g. the accessor unit tests) provide a mock instead.
```
tt-vjovanovic: Is this comment needed? If so, why others headers don't have it?

tt_metal/hw/inc/internal/firmware_common.h
```diff
@@ -206,6 +206,10 @@ void wait_for_go_message() {
 }
 
 #if !defined(COMPILE_FOR_TRISC)
+// TRISCs are not NoC initiators; the address backend is data-movement-only.
+// Arch-specific (resolved by the HAL's per-arch include roots, like noc_parameters.h).
```
tt-vjovanovic: Same comment for adding comment for include headers...

tt-vjovanovic (approved):

tt_metal/hw/inc/api/tensor/tensor_accessor.h
vtsilytskyiTT: Agree, those comments are useless.

tt_metal/hw/inc/internal/dataflow/dataflow_api_addrgen.h
vvukomanovicTT: removed the forwarder — call sites now use noc_address_backend::bank_address<> directly

tests/ttnn/unit_tests/gtests/accessor/test_tensor_accessor.cpp
vvukomanovicTT: they're required at template-definition time — qualified names in the accessor's member bodies are looked up when the template is parsed, not at instantiation (this is the exact lookup that failed in asan/clang-tidy before the mock existed). added a comment saying so

tt_metal/hw/inc/api/tensor/tensor_accessor.h
vvukomanovicTT: dropped the comments

tt_metal/hw/inc/internal/firmware_common.h
vvukomanovicTT: dropped here too

tt_metal/hw/inc/internal/dataflow/noc_address_backend_xy.h
vvukomanovicTT: both stay. worker_address is the primary constructor (the get_noc_addr(x, y, addr) path). packed_worker_address exists because some inputs arrive pre-packed — the host-generated bank tables and dispatch command fields carry a packed xy encoding (the get_noc_addr_helper path), and that format stays part of the host↔device abi under att but both converge on one path. 

it could go away later if we regenerate the bank tables to carry att selectors directly, but that's a follow-up decision, not part of this stack. wdyt?

jbaumanTT (approved):

bbradelTT (approved):

arikTT (approved):

PR #54154, opened 2026-08-24
[Infrastructure] Run L2 nightly matmul/fused/reduction/moreh unit tests in package-and-release

iwroszTT:
### PR Category
CI/Infrastructure

### Summary

Implements #54005: moves the matmul / fused / reduction / moreh nightly op suites off the release SKUs in L2 nightly and into package-and-release. Each of the four platforms stays covered exactly once, and the total test count is unchanged:

| pipeline | SKUs |
|---|---|
| l2-nightly | `wh_n300_civ2`, `bh_p100` |
| package-and-release | `wh_n150_civ2`, `bh_p150b_civ2` |

The split is implemented with a new test list: `ops_pnr_tests.yaml`

### What's changed

- `tests/pipeline_reorg/ops_pnr_tests.yaml` (new file) — the four suites on `wh_n150_civ2` and `bh_p150b_civ2`.
- `tests/pipeline_reorg/ops_unit_tests.yaml` — those two SKUs removed from the same four suites.
- `.github/workflows/ops-unit-tests-impl.yaml` — new `tests-yaml-path` and `budget-subheading` inputs. Both default to today's values.
- `.github/workflows/release-build-test-publish.yaml` — added new `ttnn-ops-unit-tests` job
- `.github/time_budget.yaml` — 205 min per SKU moved from `ttnn.unit` to a new `ttnn.package_and_release` key. Per-SKU totals are unchanged.

One unrelated improvement: the test category filter on input like `conv, pool` silently dropped `pool`, which is now fixed.

### Notes for reviewers

- Note that on merge, the PnR job's first run is also the only coverage WH N150 and BH P150 have for matmul / fused / reduction / moreh.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:iwrosz/54005-move-l2-ops-to-pnr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:iwrosz/54005-move-l2-ops-to-pnr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:iwrosz/54005-move-l2-ops-to-pnr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:iwrosz/54005-move-l2-ops-to-pnr)
[...]

PR #54207, opened 2026-08-24
Add category picker to package-and-release + ops_pnr scaffold

roseli-TT:
## Summary
- Mirrors `tt-metal-l2-nightly.yaml`'s `additional_test_categories` dispatch input into `package-and-release.yaml`, threaded through `release-build-test-publish.yaml`, so a manual release dispatch can pick a category of tests to run from the GUI — **applied across the whole PnR pipeline**, not just the new ops track (see "Implementation choice" below).
- Establishes an ops PnR test track for the ttnn team, in preparation for future ops tests:
  - `tests/pipeline_reorg/ops_pnr.yaml` — test-definitions yaml, same schema/header style as `ops_integration_tests.yaml`, with a single no-op placeholder entry (`category: matmul`) to exercise the wiring end-to-end.
  - `.github/workflows/ops-pnr-tests-impl.yaml` — reusable workflow, structurally identical to `ops-integration-tests-impl.yaml`, wired into `release-build-test-publish.yaml` under a `# Team: ttnn` comment block (same convention as `merge-gate.yaml`'s team sections).
  - `.github/time_budget.yaml` — added a `ttnn.pnr` budget as a starting point to retune once real tests land.
- Wires the same `additional_test_categories` input into `release-demo-tests-impl.yaml` so it also filters the models e2e suite (`tests/pipeline_reorg/models_e2e_tests.yaml`), layered on top of the existing tier/release_ready/model filters. Empty input = today's behavior, unchanged.

## Implementation choice: category selection for the models e2e track
`models_e2e_tests.yaml` has no dedicated `category` field (unlike `ops_pnr.yaml`/the L2 ops files). Rather than backfill one across all 82 entries, this PR **reuses the existing `model_family` field** (Llama, Gemma, Qwen, ...) as an interim category selector for that track.

**Pros:**
- Zero backfill — `model_family` already exists on ~84/82 entries; ships now instead of blocking on a taxonomy design + data-authoring pass.
- No new taxonomy to invent/maintain.
- Same filter mechanism/input (`additional_test_categories`) covers both tracks from the GUI.

**Cons:**
- Mixes two vocabularies in one input: ops feature-tags (`matmul`, `conv`, ...) vs. model family names (`Llama`, `Qwen`, ...).
- Overlaps with the existing `model:` dispatch input on the same workflow, which already filters by model substring — two different ways to select "just Llama tests."
- `model_family` is optional and 2/82 entries omit it; those become silently unreachable by category.
- Coarser than a real category system — selects by model identity only, not by test *kind* (no cross-family grouping like "all perf tests").
- Reusing a display-oriented field as a stable filter contract means renaming a `model_family` value later is a breaking change for saved dispatch inputs/automation.

**Promise:** this is intentionally an interim measure, not the final design. `ops_pnr.yaml` keeps its own dedicated `category` field. Once the models e2e taxonomy is settled, unify both tracks onto a single shared `category` field/filter mechanism (see the comment block above the `ops-pnr-tests` job in `release-build-test-publish.yaml`, which cross-references this).

## Notes for reviewers
- The `ttnn.pnr` entry added to `.github/time_budget.yaml` is copied verbatim from `ttnn.integration` (the closest existing L2 analog), plus a `wh_n150: 5` line for the no-op smoke test. Retune once real ops PnR tests land.
- `ops_pnr.yaml` currently has one no-op placeholder entry ("ops PnR no-op smoke test", `category: matmul`, owned by me) added to validate the category-filter/dispatch path with a real dispatch. Remove/replace once real ops PnR tests land.
- The `model_family`-based filter in `release-demo-tests-impl.yaml` was validated locally (YAML parses, jq expression hand-traced against sample entries) but not yet exercised end-to-end via a real dispatch — a prior dry-run dispatch validated the ops_pnr side (`ops-pnr-tests` passed) but was cancelled before this models-track change landed. Happy to launch another dry-run dispatch on request.

## Test plan
- [x] `python3 -c "import yaml; yaml.safe_load(open(...))"` on all changed files — all parse cleanly.
- [x] `verify_time_budget.py tests/pipeline_reorg/ops_pnr.yaml .github/time_budget.yaml pnr` — passes.
- [x] `prepare_test_matrix.py tests/pipeline_reorg/ops_pnr.yaml ALL_SKUS_IN_TESTS .github/sku_config.yaml` — resolves the no-op entry correctly.
- [x] Dry-run dispatch of `package-and-release.yaml` with `additional_test_categories: matmul` — `ops-pnr-tests` job passed (Ubuntu 22.04 leg): https://github.com/tenstorrent/tt-metal/actions/runs/32774251730/job/97585674877 (run was cancelled afterward, before the models-track change below).
- [ ] Re-run the dry-run dispatch to validate the `model_family` filter against `release-demo-tests`.

### DIFF
--- a/.github/workflows/ops-pnr-tests-impl.yaml
+++ b/.github/workflows/ops-pnr-tests-impl.yaml
@@ -1,7 +1,7 @@
 name: "[internal] ttnn ops PnR tests impl"
 
 # Team: ttnn
-# test list: ops_pnr.yaml (currently empty)
+# test list: ops_pnr.yaml (currently a single no-op placeholder entry)
 #
 # Reusable workflow for the ops PnR (package-and-release) test track. Mirrors
 # ops-integration-tests-impl.yaml / ops-unit-tests-impl.yaml's structure so this
--- a/.github/workflows/ops-pnr-tests-impl.yaml
+++ b/.github/workflows/ops-pnr-tests-impl.yaml
@@ -1,0 +1,128 @@
+name: "[internal] ttnn ops PnR tests impl"
+
+# Team: ttnn
+# test list: ops_pnr.yaml (currently empty)
+#
+# Reusable workflow for the ops PnR (package-and-release) test track. Mirrors
+# ops-integration-tests-impl.yaml / ops-unit-tests-impl.yaml's structure so this
+# is ready to run real tests as soon as ops_pnr.yaml gains entries.
+
+on:
+  workflow_call:
+    inputs:
+      docker-image:
+        required: true
+        type: string
+      build-artifact-name:
+        required: true
+        type: string
+      wheel-artifact-name:
+        required: true
+        type: string
+      enabled-skus:
+        required: true
+        type: string
+        description: "Comma-separated SKUs (see .github/sku_config.yaml)."
+      additional_test_categories:
+        required: false
+        type: string
+        default: ''
+        description: "Comma-separated category filter. Keep only entries whose `category` field appears in this list."
+      enable-llk-asserts:
+        required: false
+        type: boolean
+        default: false
+
+jobs:
+  load-test-matrix:
+    runs-on: ubuntu-slim
+    outputs:
+      matrix: ${{ steps.filter-categories.outputs.matrix }}
+    env:
+      TESTS_YAML_PATH: ./tests/pipeline_reorg/ops_pnr.yaml
+    steps:
+      - name: Checkout the repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        timeout-minutes: 20
+        with:
+          sparse-checkout: |
+            .github
+            tests/pipeline_reorg
+          sparse-checkout-cone-mode: true
+      - name: Install dependencies
+        run: pip3 install PyYAML
+      - name: Verify test timeouts against budget
+        run: |
+          set -e
+          python3 .github/scripts/utils/verify_time_budget.py \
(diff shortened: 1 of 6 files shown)


tdowdallTT (approved):

.github/workflows/ops-pnr-tests-impl.yaml
dpopovTT, line 1: Do we need this? Can we point to developer test config and filter out what we need. If we make this one we need to maintain it (or if the agreement is something else)

roseli-TT: I do not intend to maintain this and will confirm with Borys. 
What do you mean by developer test config?

iwroszTT (approved):

.github/workflows/ops-pnr-tests-impl.yaml
bbradelTT: @dpopovTT can you please elaborate on what you mean? What is developer test config and what would you envision the usage to be?

iwroszTT: @roseli-TT @dpopovTT @bbradelTT 
As a potential different approach I've added category picker to the PR that adds mm/fused/reduce unit tests to PnR. Please take a look: https://github.com/tenstorrent/tt-metal/pull/54154

roseli-TT: Hi all, I think Izajasz's new PR https://github.com/tenstorrent/tt-metal/pull/54154 is good and can replace this draft one, which I will close. 
I have some small concerns which I will comment on the new PR.

PR #47507, opened 2026-06-19
Universal input/output support for ttnn.pad

VirdhatchaniKN:
### Issue
https://github.com/tenstorrent/tt-metal/issues/49322

### Summary
Adds universal input/output support for `ttnn.pad`

Before this PR, `ttnn.pad` only handled HEIGHT_SHARDED inputs natively — WIDTH_SHARDED and BLOCK_SHARDED inputs would crash by incorrectly falling through the HEIGHT_SHARDED dispatch logic. This PR enables native pad for all L1 sharded memory configs across both TILE and ROW_MAJOR layouts, with no `sharded_to_interleaved` hop for common shapes.

### Changes

**`pad.cpp` — host-side dispatch**
- Added native-first routing for WIDTH/BLOCK sharded inputs via `needs_pad_composite_fallback` predicate
- Fixed bug where HEIGHT_SHARDED logic was applied to WIDTH/BLOCK inputs (now correctly gated)
- `to_memory_config` → pad → `interleaved_to_sharded`/`to_memory_config` composite fallback reserved only for DRAM-sharded W/B inputs (mirrors repeat.cpp)

**`pad_rm_reader_writer_multi_core_default_program_factory.cpp` — factory**
- Computes and passes correct `accessor_page_size` using `buffer()->aligned_page_size()` for sharded buffers
- Passes `num_input_pages_in_row` / `num_output_pages_in_row` as compile-time args for sharded page addressing

**`reader_pad_dims_rm_interleaved_v2.cpp` — reader kernel**
- Added `read_input_stick_into_l1` helper with two paths:
  - Single-page width (`num_input_pages_in_row == 1`): direct `noc.async_read` with flat page_id, avoiding `noc_async_read_sharded` page-per-row miscomputation
  - Multi-page width: `noc_async_read_sharded` for efficient sharded input reads
- Front-padding branch reads into temp buffer then `memmove`s to correct L1 offset — works natively for W/B sharded

**`writer_pad_dims_rm_interleaved_v2.cpp` — writer kernel**
- Same dual-path write: direct `noc.async_write` for single-page width, `noc_async_write_sharded` for multi-page

**`test_pad_universal_input.py` — 192-case test suite**
- 9 test categories covering all layout × sharding × pad-direction combinations:
  1. Interleaved baseline (DRAM + L1)
  2. Sharded input → DRAM output (H/W/B × 7 pad cases × 2 layouts)
  3. DRAM input → sharded output (H/W/B × 7 pad cases × 2 layouts)
[...]

PR #54214, opened 2026-08-24
Metal 2.0 port: `data_movement/pad`

### LINKED ISSUE #54213: Metal 2.0 port: data_movement/pad
Port `ttnn/cpp/ttnn/operations/data_movement/pad` to Metal 2.0 / Host 2.0 APIs (Gen1 WH/BH target).

Parent: #41908

## Scope
All seven program factories of `PadDeviceOperation`, plus all twelve kernel sources under `device/kernels/dataflow/`:

- `PadRmReaderWriterProgramFactory` — RM, single core, `create_workload_descriptor`, op-owned pad-value tensor
- `PadRmReaderWriterMultiCoreProgramFactory` — RM, multicore, `create_workload_descriptor`, op-owned pad-value tensor — **unreachable from `select_program_factory`** (see Notes)
- `PadRmReaderWriterMultiCoreDefaultProgramFactory` — RM, multicore, the default RM path
- `PadRmShardedHeightOnlyProgramFactory` — RM height-sharded, optimized; the op's only `override_runtime_arguments`
- `PadRmShardedWidthOnlyProgramFactory` — RM width-pad, sharded stickwise
- `PadTileCoreProgramFactory` — TILE, single core; borrows `eltwise/unary`'s shared reader
- `PadTileMulticoreProgramFactory` — TILE, multicore

## Audit verdict: GREEN (all gates cleared)
- **TTNN factory concept:** `descriptor` × 5, `WorkloadDescriptor` (secretly SPMD) × 2. No custom `compute_program_hash`, no backdoor `attribute_values` / `to_hash`, no `get_dynamic_runtime_args`, no pybound `create_descriptor`. One `override_runtime_arguments` (height-only sharded factory).
- **Device 2.0:** all twelve op kernels plus the one borrowed donor are structurally Device 2.0 — `Noc`, `DataflowBuffer`, `TensorAccessor`, `CoreLocalMem`. No holdovers.
- **Feature compatibility:** no GlobalCircularBuffer, no non-zero `CBDescriptor::address_offset`, no GlobalSemaphore. The op declares no semaphores at all.
- **Offset base pointers:** none — every buffer base reaches a kernel as a bare `Buffer*` in `KernelDescriptor::RTArgList`, never `->address() + <expr>`.
- **TensorAccessor 3rd arg:** 2 sites, both Class 2 (redundant) → drop, and do **not** set `dynamic_tensor_shape` (both are compile-time args).
- **TensorParameter relaxation:** `none` on all seven rows.

> One readiness-sheet cell is outdated: `PadRmShardedHeightOnlyProgramFactory`'s row reads `Is able to port? = no`, derived from a `get_dynamic_runtime_args` cell that PR #52556 (`90ec10f4bf4`) made obsolete — the hook is gone and an `override_runtime_arguments` replaced it. The audit overrode the row on code evidence. The sheet still needs correcting: `Runtime-args update (get_dynamic_runtime_args)` → `no`, `Override runtime args method? (PD only)` → `yes`.

## Port work
- Tensor bindings: 12 × Case 1 (via `TensorAccessor`), 4 × clean (borrowed-memory DFBs on the two sharded factories), **no Case 2** anywhere.
- The two `WorkloadDescriptor` factories each build **one** `ProgramDescriptor` and replicate it per range in `tensor_coords` — nothing is per-mesh-coordinate. They collapse onto the single-program concept with the pad-value const tensor carried in `ProgramArtifacts::op_owned_tensors`.
- CB endpoints: 7 one-toucher self-loops; `c_16` in the height-only sharded factory is a two-toucher (reader FIFO producer + writer raw `get_write_ptr` peek at `writer_pad_dims_rm_sharded.cpp:90`) → 1P+1C, **not** multi-binding. No DFB in this op needs the multi-binding advanced option.
- `c_2` in the default RM factory is allocated only under `stick_size_padded_front != 0 || unaligned`, but the kernel constructs it and calls `get_read_ptr()` unconditionally. Conditional DFB + a host-side define + a real `#ifdef` — `if constexpr` will not do, since it does not suppress `dfb::` name lookup.
- `c_1` in `PadTileMulticoreProgramFactory` is dead in every config (zero touchers): drop the allocation and the dead CTA carrying its index.
- Bind the existing `_metal2` fork of the borrowed reader (`eltwise/unary/device/kernels/dataflow/reader_unary_interleaved_start_id_metal2.cpp`) and conform to its vocabulary (`dfb::in`, `tensor::src`, `args::num_pages`, `args::start_id`). Do not re-fork; do not convert the legacy file in place. Note the decoy fork under `copy/typecast` uses `tensor::input`.

## Notes
- WH/BH only — no Quasar emulation. `ttnn/cpp/ttnn/operations/experimental/quasar/pad/` is a shortcut copy and is out of bounds.
- Basic Metal 2.0 port — no functional changes expected.
- **`PadRmReaderWriterMultiCoreProgramFactory` is unreachable**: declared in the `program_factory_t` variant (`pad_device_operation.hpp:33`) but never returned by `select_program_factory` (`pad_device_operation.cpp:99-101`). No test can exercise it, so any port of it is build-verified only. Worth deciding separately whether it should be deleted.

vsureshTT:
### Issue
Closes #54213.

### Summary

Metal 2.0 port of `data_movement/pad` — **all seven** `PadDeviceOperation` factories and all twelve of the op's kernel sources, converted to the base `ProgramSpecFactoryConcept`. Behaviour-preserving: DFB sizes, formats, defines and argument values match the legacy CB descriptors and `emplace_runtime_args` index for index. The op owns no compute kernels, so no compute `hw_config` and no `opt_level` work (every kernel is a reader/writer whose legacy default `O2` is already Metal 2.0's default).

- **Host side**: `CBDescriptor` → `DataflowBufferSpec` (`borrowed_from` for the four sharded shard buffers), buffer-address RTAs and the `cb.buffer` patching → `TensorParameter` / `TensorBinding`, positional CTAs → named compile-time args, per-core RTAs → a name-first `ProgramRunArgs` table.
- **`WorkloadDescriptor` → single `ProgramSpec`**: `PadRmReaderWriterProgramFactory` and `PadRmReaderWriterMultiCoreProgramFactory` each built **one** `ProgramDescriptor` above the loop and pushed the *same* object once per range in `tensor_coords`. Nothing was per-mesh-coordinate — the wrapper existed only to carry the pad-value const tensor, which now rides `ProgramArtifacts::op_owned_tensors`.
- **Kernels**: `CircularBuffer` → `DataflowBuffer` via `dfb::` bindings, `TensorAccessorArgs<N>()` chains → `TensorAccessor(tensor::…)`, positional args → `get_arg(args::…)`.
- `pad_device_operation.{hpp,cpp}` and `pad_nanobind.cpp` are **byte-identical** — no device-op-class edit was forced, and no user-visible API is removed.

**Verification: 863 passed / 63 skipped — bit-identical to the pre-port baseline**, over `test_pad.py`, `test_pad_subcoregrids.py`, `test_pad_universal_input.py`, `test_padding_test.py` and the two `sweep_tests/pytests/tensor` pad files. The baseline was captured before the first kernel edit; the post-port run carries 2674 `METAL2_CHECKS_FORCED` markers from both forced translation units, so the green was measured with the Metal 2.0 host-side legality checks on.

### Notes for reviewers

- **Endpoint dispositions were re-derived from a per-node kernel-touch census, not transcribed.** Seven one-toucher DFBs self-loop. `c_16` in the height-only sharded factory is a two-toucher — reader is a locked FIFO producer, and the **writer raw-peeks the same buffer** via `get_write_ptr()` at `writer_pad_dims_rm_sharded.cpp:90` with no FIFO ops — so it takes a plain **1P+1C** (reader PRODUCER, writer CONSUMER). That peek is invisible to a producer/consumer trace and is the one cross-kernel toucher in the op. **No DFB anywhere needs the multi-binding flag.**
- **`PadRmShardedHeightOnlyProgramFactory` drops to the base concept.** Its `override_runtime_arguments` re-pointed exactly two CB base addresses and did nothing else; both are `borrowed_from` DFBs now, and the framework refreshes every `TensorArgument` on a cache hit — so the refreshed set is identical statement-for-statement. `PadRmShardedWidthOnlyProgramFactory` already carried the same two borrowed CBs with no override. **This is a concept change and was raised with the op owner before construction, not decided in the diff.**
- **Conditional DFB (`c_2`, default RM factory).** The host allocated it only under `stick_size_padded_front != 0 || unaligned`, but the kernel constructed it and called `get_read_ptr()` *unconditionally*, gating only the uses behind `if constexpr`. Since `if constexpr` does not suppress `dfb::` name lookup — including in its **discarded branches** — the construction, the `get_read_ptr()` and the whole `if constexpr` chain now sit behind a `PAD_ALIGN_DFB` `#ifdef` fed by `compiler_options.defines`. Both preprocessor branches were confirmed to have actually compiled and run: the JIT cache holds 49 variants of that reader with `dfb::pad_align` bound and 323 without.
- **Shared kernel, rung 1 — no new fork.** `PadTileCoreProgramFactory` binds the existing `eltwise/unary` `_metal2` fork of `reader_unary_interleaved_start_id` and conforms to its vocabulary (`dfb::in`, `tensor::src`); neither the original nor the fork is touched. Note the decoy: a differently-named fork of the same kernel under `copy/typecast` uses `tensor::input`. `data_movement/pad` comes off that kernel's sunset list.
- **Two Metal 2.0 vararg-API limits are worked around; both are written up as handoffs.** `get_vararg()` has no address form, but the tiled kernels *write back* into their vararg block as scratch (`advance_tensor_index`), so those blocks are copied into locals. And a vararg count is per-`KernelSpec`, not per-node, while the height-only reader's gather plan is genuinely ragged — the exact-fit field is `[[deprecated]]`, so the spec declares the max and zero-fills; the kernel walks the block by counts it reads out of the block itself, so the tail is never read.

**Dead code removed** (each zero-functional-change; full `file:line` list in `METAL2_PORT_REPORT.md`):

- `c_1` in `PadTileMulticoreProgramFactory` — **zero touchers in every config**. The allocation and the CTA carrying its index are gone; `2 × page_size` of L1 recovered per core.
- Arguments no kernel reads get no name and are not emitted: the v1 RM pair's dead RTA slots (the two kernels were handed *identical* 27-slot lists, so each got ~half it never read) and both of their CTAs, seven dead reader CTAs in the default RM factory, and three in the width-only sharded factory.

**Preserved, not fixed** — `PadTileCoreProgramFactory` packs a FLOAT32 pad value as two bfloat16s, so padding with `7.0` fills the region with `7.0079193115234375`. The packing block is byte-identical before and after this port, and the sibling tile-multicore factory already has the correct `std::bit_cast` line, so this is pre-existing and preserved per the porting invariant rather than fixed here. Filed separately as **#54223** with the one-line fix.

**Test coverage added** (raised by the Copilot review) — that bug survived because **no test reached the factory**: `test_pad_op` does parametrize TILE × `use_multicore=False`, but its shapes are already tile-aligned so `invoke_tile` takes the `view` fast path and never reaches `prim::pad`. `test_pad_tile_single_core` now crosses a tile boundary so the factory is actually selected — six dtypes × two memory configs × three shapes, each dispatched three times so the second and third land on a program-cache hit. JIT-cache variants of `writer_unary_pad_dims_interleaved` went from zero to ten, confirming it reaches the factory rather than merely passing. 60 passed / 6 skipped / 6 xfailed; the FLOAT32 non-zero-pad case is `xfail`ed against #54223 and the `bfloat8_b` non-zero case skipped (it compares approximately).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vsureshTT/metal2-data_movement/pad)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vsureshTT/metal2-data_movement/pad)

### DIFF
--- a/ttnn/cpp/ttnn/operations/data_movement/pad/device/kernels/dataflow/reader_pad_dims_rm_interleaved_v2.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/pad/device/kernels/dataflow/reader_pad_dims_rm_interleaved_v2.cpp
@@ -12,10 +12,10 @@
 #include "api/tensor/noc_traits.h"
 #include "ttnn/operations/data_movement/common/kernels/common.hpp"
 #include "ckernel.h"
+#include "experimental/kernel_args.h"
 
-inline __attribute__((always_inline)) void fill_pad_cb_with_val(
-    const uint32_t cb_id, const uint32_t num_bytes, const uint32_t val) {
-    DataflowBuffer dfb(cb_id);
+inline __attribute__((always_inline)) void fill_pad_dfb_with_val(
+    DataflowBuffer& dfb, const uint32_t num_bytes, const uint32_t val) {
     volatile tt_l1_ptr uint32_t* ptr = reinterpret_cast<volatile tt_l1_ptr uint32_t*>(dfb.get_write_ptr());
 
     // Round up so a non-4-byte-aligned tail stick is fully filled (the loop-back read consumes all num_bytes).
@@ -49,57 +49,54 @@ inline __attribute__((always_inline)) void read_input_stick_into_l1(
 }
 
 void kernel_main() {
-    uint32_t src_addr = get_arg_val<uint32_t>(0);
-    uint32_t num_sticks_per_core = get_arg_val<uint32_t>(1);
-    uint32_t num_sticks_per_barrier = get_arg_val<uint32_t>(2);
-    uint32_t start_page_id = get_arg_val<uint32_t>(3);
-    uint32_t front_pad_n = get_arg_val<uint32_t>(4);
-    uint32_t front_pad_c = get_arg_val<uint32_t>(5);
-    uint32_t front_pad_h = get_arg_val<uint32_t>(6);
-    tt_l1_ptr uint32_t* start_dim_offset = (tt_l1_ptr uint32_t*)(get_arg_addr(7));
+    auto num_sticks_per_core = get_arg(args::num_sticks_per_core);
+    auto num_sticks_per_barrier = get_arg(args::num_sticks_per_barrier);
+    auto start_page_id = get_arg(args::start_page_id);
+    auto front_pad_n = get_arg(args::front_pad_n);
+    auto front_pad_c = get_arg(args::front_pad_c);
+    auto front_pad_h = get_arg(args::front_pad_h);
 
-    constexpr uint32_t N = get_compile_time_arg_val(0);
-    constexpr uint32_t H = get_compile_time_arg_val(1);
-    constexpr uint32_t C = get_compile_time_arg_val(2);
-    constexpr uint32_t stick_size_bytes = get_compile_time_arg_val(3);
-    constexpr uint32_t N_padded = get_compile_time_arg_val(4);
-    constexpr uint32_t H_padded = get_compile_time_arg_val(5);
-    constexpr uint32_t C_padded = get_compile_time_arg_val(6);
-    constexpr uint32_t stick_size_padded = get_compile_time_arg_val(7);
-    constexpr uint32_t stick_size_padded_front = get_compile_time_arg_val(8);
-    constexpr uint32_t stick_size_padded_end = get_compile_time_arg_val(9);
-    constexpr uint32_t num_zero_pad_sticks_read = get_compile_time_arg_val(10);
-    constexpr uint32_t last_zero_stick_size = get_compile_time_arg_val(11);
-    constexpr uint32_t stick_size_padded_aligned = get_compile_time_arg_val(18);
+    constexpr auto N = get_arg(args::N);
+    constexpr auto H = get_arg(args::H);
+    constexpr auto C = get_arg(args::C);
+    constexpr auto stick_size_bytes = get_arg(args::stick_size_bytes);
+    constexpr auto N_padded = get_arg(args::N_padded);
+    constexpr auto H_padded = get_arg(args::H_padded);
+    constexpr auto C_padded = get_arg(args::C_padded);
+    constexpr auto stick_size_padded = get_arg(args::stick_size_padded);
+    constexpr auto stick_size_padded_front = get_arg(args::stick_size_padded_front);
+    constexpr auto stick_size_padded_aligned = get_arg(args::stick_size_padded_aligned);
 
-    constexpr bool not_pad_by_zero = get_compile_time_arg_val(12) == 1;
-    constexpr uint32_t front_padding = get_compile_time_arg_val(8);
-    constexpr bool unaligned = get_compile_time_arg_val(19) == 1;
+    constexpr bool not_pad_by_zero = get_arg(args::not_pad_by_zero) == 1;
+    constexpr uint32_t front_padding = stick_size_padded_front;
+    constexpr bool unaligned = get_arg(args::unaligned) == 1;
 
-    constexpr uint32_t num_input_pages_in_row = get_compile_time_arg_val(20);
-    constexpr uint32_t accessor_page_size = get_compile_time_arg_val(21);
-    constexpr auto src_args = TensorAccessorArgs<22>();
+    constexpr auto num_input_pages_in_row = get_arg(args::num_input_pages_in_row);
 
     uint32_t packed_pad_value = 0;
     if constexpr (not_pad_by_zero) {
-        packed_pad_value = kernel_compile_time_args[13];
+        packed_pad_value = get_arg(args::packed_pad_value);
     }
 
-    constexpr uint32_t dfb_in0 = tt::CBIndex::c_0;
-    constexpr uint32_t cb_pad = tt::CBIndex::c_1;
-    constexpr uint32_t dfb_pad_align = tt::CBIndex::c_2;
-    DataflowBuffer dfb_in0_exp(dfb_in0);
-    DataflowBuffer dfb_pad_exp(cb_pad);
-    DataflowBuffer dfb_pad_align_exp(dfb_pad_align);
+    DataflowBuffer dfb_in0_exp(dfb::in0);
+    DataflowBuffer dfb_pad_exp(dfb::pad);
+    // The realignment staging buffer is bound only when the host allocated it (front padding, or
+    // an unaligned padded stick). A kernel may not name a DFB it has not bound, and `if constexpr`
+    // does not suppress that name lookup, so every reference to it is gated at the preprocessor.
+#ifdef PAD_ALIGN_DFB
+    DataflowBuffer dfb_pad_align_exp(dfb::pad_align);
+#endif
 
-    const auto s = TensorAccessor(src_args, src_addr, accessor_page_size);
+    const auto s = TensorAccessor(tensor::src);
     Noc noc;
 
     const uint32_t pad_val_addr = dfb_pad_exp.get_read_ptr();
+#ifdef PAD_ALIGN_DFB
     const uint32_t pad_align_addr = dfb_pad_align_exp.get_read_ptr();
+#endif
 
-    fill_pad_cb_with_val(cb_pad, stick_size_padded, packed_pad_value);
-    // The fill above is baby-RISCV stores; the per-stick loop below loop-back noc.async_read's cb_pad as
+    fill_pad_dfb_with_val(dfb_pad_exp, stick_size_padded, packed_pad_value);
+    // The fill above is baby-RISCV stores; the per-stick loop below loop-back noc.async_read's the pad DFB as
     // its source. A baby-RISCV store can retire before its write-request lands in L1, and the RISCV core
     // and NoC are different L1 clients with no program-order guarantee between them
     // (WormholeB0/TensixTile/BabyRISCV/MemoryOrdering.md). load_blocking the last filled word (blocking
@@ -109,7 +106,8 @@ void kernel_main() {
         reinterpret_cast<volatile tt_l1_ptr uint32_t*>(pad_val_addr) + (stick_size_padded / sizeof(uint32_t)) - 1);
 
     uint32_t i_page = start_page_id;
-    uint32_t curr_c = start_dim_offset[2], curr_h = start_dim_offset[1], curr_n = start_dim_offset[3];
+    uint32_t curr_c = get_arg(args::start_dim_offset_c), curr_h = get_arg(args::start_dim_offset_h),
+             curr_n = get_arg(args::start_dim_offset_n);
     for (uint32_t iter = 0; iter < num_sticks_per_core;) {
         dfb_in0_exp.reserve_back(num_sticks_per_barrier);
         uint32_t l1_write_addr = dfb_in0_exp.get_write_ptr();
@@ -130,6 +128,7 @@ void kernel_main() {
                 noc.async_read_barrier();
             }
             if (read_stick) {
+#ifdef PAD_ALIGN_DFB
                 if constexpr (front_padding) {
                     uint32_t temp_addr = dfb_pad_align_exp.get_write_ptr();
                     read_input_stick_into_l1(noc, s, i_page, temp_addr, num_input_pages_in_row, stick_size_bytes);
@@ -151,7 +150,9 @@ void kernel_main() {
                          .noc_y = (uint32_t)my_y[noc.get_noc_id()],
                          .addr = pad_align_addr},
                         {.offset_bytes = 0});
-                } else {
+                } else
+#endif
+                {
                     read_input_stick_into_l1(noc, s, i_page, l1_write_addr, num_input_pages_in_row, stick_size_bytes);
                 }
             }
--- a/ttnn/cpp/ttnn/operations/data_movement/pad/device/pad_tile_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/pad/device/pad_tile_program_factory.cpp
@@ -3,70 +3,77 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include "pad_tile_program_factory.hpp"
-#include <tt-metalium/tensor_accessor_args.hpp>
+
+#include <tt-metalium/experimental/metal2_host_api/program_run_args.hpp>
+#include <tt-metalium/experimental/metal2_host_api/program_spec.hpp>
+#include "ttnn/operations/core/data_movement_kernel/datamovement_kernel_config.hpp"
 #include "ttnn/operations/data_movement/common/common.hpp"
-#include <tt-metalium/host_api.hpp>
-#include <tt-metalium/program_descriptors.hpp>
 
 using namespace tt::constants;
 using namespace tt::tt_metal;
+using namespace tt::tt_metal::experimental;
 
 namespace ttnn::prim {
 using ttnn::operations::data_movement::float_to_uint16;
 using ttnn::operations::data_movement::pack_two_uint16_into_uint32;
 
-ProgramDescriptor PadTileCoreProgramFactory::create_descriptor(
+namespace {
+// Names are prefixed per factory: all seven pad factories land in one unity-build
+// translation unit, where every anonymous namespace is merged into a single scope.
+const KernelSpecName TILE_SC_READER{"reader"};
+const KernelSpecName TILE_SC_WRITER{"writer"};
+const DFBSpecName TILE_SC_IN0{"in0"};
+const DFBSpecName TILE_SC_PAD{"pad"};
+const TensorParamName TILE_SC_INPUT{"input"};
+const TensorParamName TILE_SC_OUTPUT{"output"};
+}  // namespace
+
+ttnn::device_operation::ProgramArtifacts PadTileCoreProgramFactory::create_program_artifacts(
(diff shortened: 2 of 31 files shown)


vsureshTT:
/ping codeowners Metal 2.0 port of `data_movement/pad` — all seven factories, no device-op-class or pybind changes, 863/63 tests bit-identical to the pre-port baseline. Ready for review.

ttnn/cpp/ttnn/operations/data_movement/pad/device/pad_tile_program_factory.cpp
```diff
@@ -3,70 +3,77 @@
...
+// translation unit, where every anonymous namespace is merged into a single scope.
+const KernelSpecName TILE_SC_READER{"reader"};
+const KernelSpecName TILE_SC_WRITER{"writer"};
+const DFBSpecName TILE_SC_IN0{"in0"};
+const DFBSpecName TILE_SC_PAD{"pad"};
+const TensorParamName TILE_SC_INPUT{"input"};
+const TensorParamName TILE_SC_OUTPUT{"output"};
+}  // namespace
+
+ttnn::device_operation::ProgramArtifacts PadTileCoreProgramFactory::create_program_artifacts(
```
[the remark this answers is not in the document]
vsureshTT: Agreed, and fixed rather than deferred — this was a real gap, not a Gen2-scope one.

Added `test_pad_tile_single_core` (`tests/ttnn/unit_tests/operations/data_movement/test_pad.py`), which does exactly what you suggest: crosses a tile boundary with `use_multicore=False` so `invoke_tile` cannot take the `view` fast path and `PadTileCoreProgramFactory` is actually selected. Six dtypes × two memory configs × three shapes (`[1,1,32,32]→[1,1,64,64]`, plus two that grow the batch/channel dims), and each case dispatches **three** times so the second and third land on a program-cache hit — which is where a stale tensor binding would surface.

Confirmed it reaches the factory rather than just passing: JIT-cache variants of `writer_unary_pad_dims_interleaved` went from **zero to ten**. Result: 60 passed / 6 skipped / 6 xfailed.

Two carve-outs, both deliberate:
- **FLOAT32 with a non-zero pad value is `xfail`ed against #54223.** That factory packs a FLOAT32 pad value as two bfloat16s, so `value=7.0` fills the region with `7.0079193115234375`. It is pre-existing — the packing block is byte-identical before and after this port, and the sibling tile-multicore factory already has the correct `std::bit_cast` line. The port preserved it per the porting invariant; the new test documents it instead of hiding it, and I filed the bug separately with the one-line fix.
- **`bfloat8_b` with a non-zero pad value is skipped**, since it compares approximately and the exact-fill assertion does not apply.

ttnn/cpp/ttnn/operations/data_movement/pad/device/kernels/dataflow/reader_pad_dims_rm_interleaved_v2.cpp
```diff
@@ -49,57 +49,54 @@ inline __attribute__((always_inline)) void read_input_stick_into_l1(
...
+    DataflowBuffer dfb_pad_exp(dfb::pad);
+    // The realignment staging buffer is bound only when the host allocated it (front padding, or
+    // an unaligned padded stick). A kernel may not name a DFB it has not bound, and `if constexpr`
+    // does not suppress that name lookup, so every reference to it is gated at the preprocessor.
+#ifdef PAD_ALIGN_DFB
+    DataflowBuffer dfb_pad_align_exp(dfb::pad_align);
+#endif
 
-    const auto s = TensorAccessor(src_args, src_addr, accessor_page_size);
+    const auto s = TensorAccessor(tensor::src);
```
llongTT, line 90: **The dropped `TensorAccessor` page-size override changes the input stride in one configuration, and nothing tests it.**

Pre-port this line passed an explicit third argument, `input_accessor_page_size`, which for a non-sharded input was `stick_size = a.logical_shape()[3] * element_size` (`pad_rm_reader_writer_multi_core_default_program_factory.cpp:57` before this PR). The Metal 2.0 binding instead supplies `align(spec.compute_page_size_bytes(), allocator_alignment)` (`program_spec.cpp:2292-2295`), and `compute_page_size_bytes` runs the logical shape through `compute_physical_shape`, which rounds the last dimension up to the layout alignment — so the value is derived from the **padded** last dimension.

Scoping what actually differs, so this is not read as larger than it is:

- **Alignment is not a difference.** The non-sharded accessor forwards its page size into `InterleavedAddrGen`, which computes `aligned_page_size = align_power_of_2(page_size, allocator_alignment)` itself (`dataflow_api_addrgen.h:285-290`). Passing an unaligned value and passing a pre-aligned one give the same stride.
- **The writer is unaffected.** `pad.cpp:299-302` passes the same shape as both the logical and the padded output shape, so `compute_page_size_bytes` equals the old `stick_size_padded` exactly.
- **The sharded input branch is unaffected** — it already used `a.buffer()->aligned_page_size()`.

That leaves exactly one case: a row-major input whose `padded_shape[-1]` exceeds its `logical_shape[-1]`. There the two values differ, and the new one is the correct stride — `Buffer::aligned_page_size()` is `align(page_size, alignment)` (`buffer.cpp:658`) and the bank arithmetic strides by exactly that (`buffer.cpp:652`).

Worth recording that this restores older behaviour rather than inventing new behaviour: the override was introduced by #47507 (`75ee03e9dc3`). Immediately before that commit this same line read `TensorAccessor(src_args, src_addr)`, i.e. it used the buffer's aligned page size — the quantity the binding now supplies.

Two residuals, neither a blocker:

1. The per-stick read *length* still comes from the logical width (`stick_size_bytes`, factory line 193 / kernel line 62), so that configuration moves closer to correct without becoming fully correct.
2. Nothing pins the behaviour either way. I could not find a row-major device input with a padded last dimension distinct from its logical one anywhere in `test_pad.py`; the reachable construction is a two-shape reshape, which `pad.cpp:429-433` explicitly contemplates. One such case would turn a silent behaviour change into a checked one.

The `METAL2_PREPORT_AUDIT.md` note reaches the same conclusion on direction; this adds the bound on scope and the pre-#47507 history.

*(Verified by reading the sources at `63f54b5b`, not by running the tests.)*

vsureshTT: Thanks — I re-derived this independently rather than taking it on trust, and every hop checks out:

- `program_spec.cpp:2292-2295` is `align(spec.compute_page_size_bytes(), allocator_alignment)`, and `TensorSpec::compute_page_size_bytes()` → `TensorLayoutImpl::compute_page_size_bytes(logical_shape_)` → `compute_physical_shape(shape)` (`tensor_layout.cpp:277-281`), so the value is derived from the padded last dimension. ✔
- **Writer unaffected:** `pad.cpp:298-301` passes `output_shape` as *both* the logical and the padded output shape to `prim::pad`, so `compute_page_size_bytes` equals the old `stick_size_padded` exactly. ✔
- **Alignment isn't the difference** — `InterleavedAddrGen` re-aligns internally, so the pre-aligned and unaligned values coincide. ✔
- **Sharded input branch unaffected** — it already passed `aligned_page_size()`. ✔
- **The #47507 history is exactly as you describe.** `git show 75ee03e9dc3^` has `TensorAccessor(src_args, src_addr)`; `75ee03e9dc3` is what turned it into the 3-arg form. So this restores the pre-#47507 quantity rather than inventing one. ✔

So I agree on both the direction and your bound on the scope: one case differs — a row-major input with `padded_shape[-1] > logical_shape[-1]` — and the new value is the correct stride there.

**On residual 2 (nothing pins it): I went looking for that test case and could not construct the input.** Five routes, all blocked on Wormhole at `63f54b5b`:

| Construction | Result |
|---|---|
| `ttnn.reshape(x, Shape([1,1,4,8]), Shape([1,1,4,16]))` — widen padded | `TT_FATAL mesh_buffer_->size() >= spec_.compute_packed_buffer_size_bytes()` (`mesh_tensor_impl.hpp:33`) |
| `ttnn.reshape(x, Shape([1,1,4,4]), Shape([1,1,4,8]))` — narrow logical | `TT_FATAL` volume mismatch (`reshape.cpp:637`) |
| `ttnn.experimental.view(big, Shape([1,1,4,8]))` | `TT_FATAL new_volume == old_volume` (`reshape_common.cpp:50`) |
| `ttnn.to_layout(tile_tensor, ROW_MAJOR)` | succeeds, but **drops** the padding — `[1,1,18,13]` padded `[1,1,32,32]` comes back padded `[1,1,18,13]` |
| `ttnn.pad`'s own RM output | `logical == padded` always, exactly as the comment at `pad.cpp:428-433` says |

The first two guards look **jointly exclusive**, which is what makes this more than "I didn't find one": to carry `padded[-1] > logical[-1]` the buffer must be sized for the *padded* spec, but the two-shape reshape simultaneously insists the *logical* volume is preserved. Satisfying one violates the other.

I'd rather say that plainly than add a test that constructs a tile-aligned tensor, passes, and looks like it pins the behaviour while testing nothing — that's strictly worse than no test. **If you know the construction I'm missing, I'll add the case gladly** — it's a small test and I'd like this pinned too.

Two caveats I don't want to overstate this past:

1. **"Not reachable from Python today" ≠ "impossible."** A C++ caller can build such a spec directly via `TensorLayout::fromPaddedShape` — which is exactly what `compute_output_specs` uses — so this is a guard-enforced invariant, not a type-level one. It could open up if a guard is relaxed.
2. **Residual 1 is real and I'm leaving it.** The per-stick read length stays on the logical width, so if that input ever becomes constructible the config is closer to correct rather than correct. `METAL2_PREPORT_AUDIT.md` lands in the same place (its open question 3), and the port deliberately doesn't touch it: the porting invariant is that behaviour, bugs included, is preserved. Happy to file it as a follow-up issue against the op owners if you'd like it tracked rather than left in the audit doc.

llongTT (approved):

vsureshTT:
Heads-up on the red CI here, and what it is.

`ttnn misc ops group [wh_n300_civ2]` fails on `test_paged_cache_mask.py::test_update_cache[...cache_shape=(32, 1, 32, 128)-1x2_grid]`, which reaches pad through `ttnn.experimental.paged_update_cache`:

```
TT_FATAL program_spec.cpp:1567: dfb_bytes <= tensor_bytes
DFB 'in_shard' (entry_size 256 * num_entries 32 = 8192 bytes) is larger than
its borrowed TensorParameter 'input' (4096 bytes).
```

That job is **green on `main`** and red here, so it is a real regression from this PR — not flaky infra. But **the port's DFB sizing is not what is wrong.**

The spec-time check compares a **per-node** DFB against the **whole tensor's logical packed size**. A DFB is a per-node object and a borrowed DFB views one shard, so for a sharded tensor those are unrelated quantities. Here the per-node shard is 8192 B and holds the 8192 B DFB exactly, while the whole-tensor packed size is 4096 B:

| quantity | value |
|---|---|
| DFB, per node | `256 B/stick × 32 sticks` = **8192 B** |
| shard allocation per node (`aligned_size_per_bank`) | **8192 B** — fits exactly |
| `compute_packed_buffer_size_bytes()` (whole tensor, logical) | **4096 B** ← what the check used |

The authoritative per-bank check at `program_run_args.cpp:472-474` uses `aligned_size_per_bank()` and accepts this spec. Only the early, approximate check rejects it. The comment above that check even anticipates the two diverging — but only in the permissive direction.

This op's DFB is byte-for-byte the legacy `CBDescriptor` it replaced (`cb_src0.total_size = shard_height_unpadded * stick_size_unpadded`), so nothing about the sizing changed in the port; moving it onto `borrowed_from` is what subjects it to the check. Pad is likely the first op to borrow a DFB from a tile-padded height-sharded tensor.

**Fix is up as #54237** (`tt_metal` only, kept out of this PR so the validator change gets its own review). **This PR is blocked on it** — I could not fix it in scope, since shrinking the DFB to satisfy the coarse check would under-size it against the shard the kernel actually walks.

Verified locally with #54237 applied on top of this branch: the single-device repro passes, the pad suite is **923 passed / 69 skipped / 6 xfailed / 0 failed**, and all 6 `*BorrowedMemoryDFB*` gtests pass (including the interleaved oversize check, which still fires).

Also worth recording: the confirmed test set for this port was scoped by the name "pad", so it missed callers that reach pad indirectly. That is why this surfaced in CI rather than locally.

PR #54236, opened 2026-08-24
[skip ci] Add tenstorrent-skills-reviewer workflow; remove daily-repo-status

blozano-tt:
Adds a domain-knowledge PR reviewer backed by [`blozano-tt/skills`](https://github.com/blozano-tt/skills), mirroring `mattpocock-skills-reviewer.md` in structure so the two are maintainable side by side.

Where the existing reviewer gives engineering-quality advice — TDD, architecture, codebase design — this one catches what a generic reviewer cannot: circular-buffer UB, race hazards, L1 footprint discipline, trace-capture safety, precision policy, CCL topology, and program-cache correctness.

## Also in this PR: removes `daily-repo-status`

Separate concern, bundled at the maintainer's request and kept as its **own commit (`5703fc4`) so it can be dropped independently** if anyone wants to keep it.

It runs on a `schedule: daily` and has been firing daily — eight runs in the last eight days, seven successful — so it draws continuously on the AI-credit budget shared with the other agentic workflows. The daily activity report it produces is not judged worth that ongoing cost.

Removes the source and its compiled lock file. Nothing else under `.github/` references it.

**If you want the reviewer but not this removal, say so and I'll drop the commit** rather than the whole PR.

## Why a separate workflow rather than extending the existing one

1. **The skill sets have incompatible loading budgets.** `mattpocock-skills-reviewer`'s `pr-triage` picks 1–2 of Matt's skills. These skills load a shared contract plus **at most two** domain skills. Merged, a single agent would hold up to five — which is precisely the "reviewer holding fourteen checklists applies all of them shallowly" failure the router exists to prevent. You would not get both reviews; you would get one worse one.

2. **`safe-outputs` caps are shared and already tight.** At `create-pull-request-review-comment: max: 10`, a CB-ownership UB finding would compete for slots with a naming suggestion. Not comparable in severity, and no way to express that within one budget.

3. **Different severity semantics.** Matt's findings are advisory and reasonable to decline. `MUST-FIX` here means wrong — a hang, a race, silent corruption. Mixing them in one thread trains readers to skim both at the same weight.

If PR noise turns out to be the bigger complaint, consolidating two working workflows later is far easier than untangling one overloaded one.

## What's reused

Deliberately kept close to the existing workflow:

- The **SHA-keyed prefetch and cache block, verbatim** — it is the best part of that workflow, and divergence here would be pure cost.
- The **`pr-triage` sub-agent pattern** on `claude-haiku-4.5`.
- **`engine: copilot`**, matching all five existing tt-metal agentic workflows.
- The same `safe-outputs` shape and advisory `COMMENT`-only review.

## What differs

- **Distinct slash command `/tt`** so the two can be invoked independently.
- **Scoped triggers** — `ttnn/**`, `tt_metal/**`, `models/**` and the matching test trees, rather than running on everything but docs.
- **Routing is path-based**, driven by `/tt-review-router`'s table, and returns **at most two** domain skills with anything dropped named in `key_signals`.
- **The evidence rule is load-bearing**: every finding cites `file:line` plus a source. A finding without evidence does not go in the report — a domain-loaded reviewer that speculates is worse than no reviewer, because its findings look authoritative.
- **An `Unresolved` section** is required whenever a severity was downgraded or something could not be verified. An empty one on a complex diff is a claim to have verified everything.

## Verification

- `gh aw compile` succeeds; `tenstorrent-skills-reviewer.lock.yml` is committed alongside.
- All **14 skill pins verified to resolve** at the pinned SHA. This matters more than it looks: gh-aw reports a failed skill install as a **non-fatal warning**, so an unresolvable pin would silently degrade this into a generic review rather than failing the run.

## Pin currency

Pinned at `blozano-tt/skills@a4bd24b8`, which is current `main` of that repo — it includes everything merged there, most recently the `CLAUDE.md`/ADR restructure. Verified at that exact SHA rather than assumed.

## Note for reviewers

The skills repo is public and Apache-2.0. It aggregates and credits work from `tt-buddy`, `tt_ops_code_gen`, this repo's `.agents` and `tt-llk/.claude` trees, the `bug_checker` rules, and the special-values tech report — see its [`SOURCES.md`](https://github.com/blozano-tt/skills/blob/main/SOURCES.md) for per-path author attribution.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:feat/tenstorrent-skills-reviewer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:feat/tenstorrent-skills-reviewer)


blozano-tt:
Thanks — two of the three were real. Addressed in 4d83e16.

**✅ `create-check-run` removed.** Correct finding. Nothing in the workflow used it, but enabling it let the agent request a `failure` or `action_required` check — directly contradicting the COMMENT-only advisory contract stated two sections above it in the same frontmatter. Dropped rather than documented, since nothing needed it. Verified gone from the regenerated lock file.

**✅ Routing conflict fixed.** Also correct, and the more interesting one. The pinned `tt-review-router` maps "any diff" to `tt-comment-hygiene-review`; the workflow's embedded copy omitted that row while separately claiming an empty `domain_skills` was possible. Identical diffs really could have routed differently depending on which table triage followed.

Resolved the way the router intends rather than by just adding the row: comment-hygiene is *cheap and runs alongside*, so it now returns in `always_skills` and does **not** consume one of the two domain slots — otherwise it would eat a slot on every PR. An empty `domain_skills` remains valid and now means "no domain skill", not "no skills". I also stated explicitly that the installed router is authoritative and the embedded table is a copy for speed, so a future divergence reads as a bug rather than a choice.

**⏸️ `synchronize` trigger — left deliberately, flagging for a human decision.**

The observation is fair: without it, commits pushed after `opened` / `ready_for_review` aren't reviewed automatically. But the comparison isn't quite right, and the fix is a cost call rather than a correctness one:

- This mirrors `mattpocock-skills-reviewer.md` in *this* repo, which also uses `types: [opened, ready_for_review]`. The workflow with `synchronize` is the reference workflow in the skills repo, which isn't what this was asked to mirror.
- One factual correction: the cached head does **not** go stale. The cache key is `pr-prefetch-${{ github.event.pull_request.head.sha ... }}`, so it's keyed on head SHA and a new commit is a cache miss by construction. The staleness risk is *no review*, not a *wrong* review from stale data.
- `synchronize` fires on every push. On a repo with tt-metal's PR volume that's a large increase in runs and AI credits against a `max-daily-ai-credits: 10000` budget shared with four other agentic workflows.

`/tt` is available for on-demand re-review in the meantime. Happy to add `synchronize` if the cost is acceptable — it's a two-line change plus a recompile — but that's a budget decision I'd rather not make unilaterally.

PR #54237, opened 2026-08-24
fix(metal2): skip the borrowed-DFB packed-size check for sharded specs

### LINKED ISSUE #54234: Metal 2.0: borrowed-DFB spec-time size check compares per-core DFB against whole-tensor packed size, rejecting valid sharded specs
The spec-time sizing check for a borrowed-memory DFB compares a **per-core** DFB size against the **whole-tensor** packed size. For a height-sharded tensor whose shard shape is padded relative to the logical volume, the per-core shard legitimately exceeds the whole-tensor packed size, so the check fires a false failure on a configuration the precise per-bank check accepts.

## The two checks disagree

**Coarse, spec time** — `tt_metal/impl/metal2_host_api/program_spec.cpp:1556-1567`:

```cpp
const size_t dfb_bytes = static_cast<size_t>(dfb.entry_size) * static_cast<size_t>(dfb.num_entries);
const size_t tensor_bytes = tensor_spec.compute_packed_buffer_size_bytes();
TT_FATAL(dfb_bytes <= tensor_bytes, ...);
```

**Precise, attach time** — `tt_metal/impl/metal2_host_api/program_run_args.cpp:472-474`:

```cpp
const uint32_t dfb_total_bytes = dfb_impl->config.entry_size * dfb_impl->config.num_entries;
TT_FATAL(dfb_total_bytes <= buffer->aligned_size_per_bank(), ...);
```

`aligned_size_per_bank()` is the right quantity — a DFB is a per-core object and a borrowed DFB views one shard. `compute_packed_buffer_size_bytes()` is the whole tensor's *logical* packed size, which for a sharded tensor is unrelated to what one core holds.

The comment above the coarse check anticipates the two diverging, but only in the permissive direction:

> For sharded L1 tensors the two checks differ — a DFB can pass here against the full-tensor size and still fail per-bank later. By design.

The reverse also happens, and that is this bug: the coarse check rejects specs the precise check would accept.

## Repro (single device, Wormhole n150)

```python
W, LOG_H, SHARD_H = 128, 16, 32          # 16 logical sticks, 32-row shard
t = torch.rand(1, 1, LOG_H, W).bfloat16()
grid = ttnn.num_cores_to_corerangeset(1, device.compute_with_storage_grid_size(), row_wise=True)
cfg = ttnn.create_sharded_memory_config(
    shape=(SHARD_H, W), core_grid=grid,
    strategy=ttnn.ShardStrategy.HEIGHT, use_height_and_width_as_shard_shape=True)
x = ttnn.from_torch(t, device=device, layout=ttnn.ROW_MAJOR_LAYOUT, dtype=ttnn.bfloat16, memory_config=cfg)
ttnn.pad(x, [1, 1, 24, W], [0, 0, 0, 0], value=0.0, use_multicore=True, memory_config=cfg)
```

```
TT_FATAL @ tt_metal/impl/metal2_host_api/program_spec.cpp:1567: dfb_bytes <= tensor_bytes
DFB 'in_shard' (entry_size 256 * num_entries 32 = 8192 bytes) is larger than
its borrowed TensorParameter 'input' (4096 bytes).
```

The numbers:

| quantity | value |
|---|---|
| DFB (per core) | `256 B/stick × 32 sticks` = **8192 B** |
| shard allocation per core (`aligned_size_per_bank`) | `32 × 128 × 2` = **8192 B** ✅ fits exactly |
| `compute_packed_buffer_size_bytes()` (whole tensor, logical) | `16 × 128 × 2` = **4096 B** ← what the check uses |

## Where it bites

Hit by the Metal 2.0 port of `data_movement/pad` (#54214). `PadRmShardedHeightOnlyProgramFactory` sizes its borrowed input DFB from the **input shard spec** — `entry_size = stick_size_unpadded`, `num_entries = shard_height_unpadded` — which is byte-for-byte what the legacy `CBDescriptor` used (`cb_src0.total_size = shard_height_unpadded * stick_size_unpadded`, `cb_src0.buffer = src_buffer`). The port did not change the sizing; it only moved it onto `borrowed_from`, which is what subjects it to this check.

CI signal: `ttnn-sanity-tests / ttnn misc ops group [wh_n300_civ2]` → `test_paged_cache_mask.py::test_update_cache[...cache_shape=(32, 1, 32, 128)-1x2_grid]`, which reaches pad through `ttnn.experimental.paged_update_cache`. That test builds its input with `shape=(nearest_y(nh, TILE_SIZE), dim)` — a shard shape rounded up to a tile while the logical `nh` is 1, which is exactly the padded-shard/small-logical-volume shape above. Green on `main`, red on the port branch.

All four borrowed DFBs across pad's two sharded factories have the same exposure (`in_shard`/`out_shard` in the height-only and width-only factories); the height-only input DFB is simply the first one reached.

## Suggested fix

Make the coarse check use a per-shard quantity when the `TensorSpec` is sharded, or drop it for sharded specs and rely on the precise per-bank check at attach time. The interleaved case is unaffected either way.

Filed from the port of #54213 / #54214; that PR is blocked on this, since sizing the DFB to satisfy the coarse check would under-size it against the shard the kernel actually walks.

vsureshTT:
### Issue
Closes #54234.

### Summary

`ValidateProgramSpec` compares a borrowed DFB's total bytes against the backing `TensorParameter`'s `compute_packed_buffer_size_bytes()`. Those two quantities are **not comparable when the spec is sharded**: a DFB is a per-node object and a borrowed DFB views **one shard**, while `compute_packed_buffer_size_bytes()` is the whole tensor's *logical* packed size.

A shard shape padded relative to the logical volume makes the per-node shard legitimately larger than the whole-tensor packed size, so the check rejects valid specs:

| quantity | value |
|---|---|
| DFB, per node | `256 B/stick × 32 sticks` = **8192 B** |
| shard allocation per node (`aligned_size_per_bank`) | `32 × 128 × 2` = **8192 B** — fits exactly ✅ |
| `compute_packed_buffer_size_bytes()` (whole tensor, logical) | `16 × 128 × 2` = **4096 B** ← what the check used |

The comment above the check already anticipated the two diverging, but only in the permissive direction — *"a DFB can pass here against the full-tensor size and still fail per-bank later. By design."* This is the opposite case, and it's a **false failure** rather than a missed one.

**The fix is to skip the early check for sharded specs.** The authoritative per-bank check is untouched and still runs at attach time in `AttachBorrowedDFBBuffers` (`program_run_args.cpp:472-474`) against `Buffer::aligned_size_per_bank()` — the quantity that actually bounds a borrowed DFB. The interleaved path, where the packed size *is* the per-bank size, keeps the early check.

I chose skipping over recomputing a per-shard size in the validator deliberately: no `Buffer` exists at spec time, so any per-shard figure would have to re-derive shard geometry and page alignment by hand, and getting *that* slightly wrong reintroduces the same class of false failure one layer down. The precise check already exists and is correct.

### Notes for reviewers

- **Interleaved coverage is unchanged.** `ProgramSpecTestQuasar.BorrowedMemoryDFBOversizedFails` builds an interleaved `TensorParameter`, so it still fires; verified passing.
- **New regression test**: `BorrowedMemoryDFBShardedLargerThanPackedSizeSucceeds` pins the sharded case — an 8192 B DFB on a 32-row shard of a logical `[1, 1, 16, 128]` bf16 tensor (4096 B packed) must build. It fails before this change and passes after.
- **Verified**: all 6 `*BorrowedMemoryDFB*` gtests pass; the wider `*Borrowed*:*ProgramSpec*` filter runs 221 tests → 199 passed, 22 skipped (Quasar/multi-device fixtures), **0 failed**.

### How it was found

The Metal 2.0 port of `data_movement/pad` (#54214) is the first op to borrow a DFB from a tile-padded height-sharded tensor. It surfaced as `ttnn misc ops group [wh_n300_civ2]` → `test_paged_cache_mask.py::test_update_cache[...cache_shape=(32, 1, 32, 128)-1x2_grid]`, which reaches pad through `ttnn.experimental.paged_update_cache` with a shard shape of `nearest_y(nh, TILE_SIZE)` — a shard rounded up to a tile while the logical `nh` is 1. Green on `main`, red on that branch; **#54214 is blocked on this PR**.

The port's DFB sizing is byte-for-byte the legacy `CBDescriptor` it replaced (`cb_src0.total_size = shard_height_unpadded * stick_size_unpadded`), so this is a validator issue, not a porting one.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vsureshTT/metal2-borrowed-dfb-sharded-size-check)

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -1224,6 +1224,25 @@ TEST_F(ProgramSpecTestQuasar, BorrowedMemoryDFBOversizedFails) {
             ::testing::HasSubstr("is larger than its borrowed TensorParameter")));
 }
 
+TEST_F(ProgramSpecTestQuasar, BorrowedMemoryDFBShardedLargerThanPackedSizeSucceeds) {
+    // A DFB is a per-node object and a borrowed DFB views ONE shard, so for a sharded
+    // TensorParameter the per-node shard is the quantity that must hold it -- not
+    // compute_packed_buffer_size_bytes(), which is the whole tensor's logical packed size.
+    //
+    // A shard shape padded relative to the logical volume makes the per-node shard legitimately
+    // LARGER than the whole-tensor packed size. Here: logical [1, 1, 16, 128] bfloat16 packs to
+    // 16*128*2 = 4096 B, while the 32-row shard allocates 32*128*2 = 8192 B on its one node. A
+    // 8192 B DFB fits the shard exactly and must be accepted; comparing it against the 4096 B
+    // packed size would reject it. The authoritative per-bank check still runs at attach time
+    // (AttachBorrowedDFBBuffers, against Buffer::aligned_size_per_bank).
+    ProgramSpec spec = MakeBorrowedDFBProgramSpec(
+        "borrowed_tensor", tt::tt_metal::BufferType::L1, /*dfb_entry_size=*/256, /*dfb_num_entries=*/32);
+    spec.tensor_parameters = {
+        MakeShardedTensorParameter("borrowed_tensor", tt::tt_metal::Shape{1, 1, 16, 128}, {32, 128}, /*num_cores=*/1)};
+
+    EXPECT_NO_THROW({ MakeProgramFromSpec(*mesh_device_, spec); });
+}
+
 TEST_F(ProgramSpecTestQuasar, SemaphoresSucceed) {
     ProgramSpec spec = MakeMinimalValidProgramSpec();
 
--- a/tt_metal/impl/metal2_host_api/program_spec.cpp
+++ b/tt_metal/impl/metal2_host_api/program_spec.cpp
@@ -1551,8 +1551,19 @@ void ValidateProgramSpec(const ProgramSpec& spec, const CollectedSpecData& colle
         // Coarse spec-time sizing check against the TensorSpec's full packed size. No Buffer is
         // available at spec time, so we can't query the per-bank allocation; the precise per-bank
         // check fires at attach time in AttachBorrowedDFBBuffers (program_run_args.cpp), where
-        // a Buffer is in hand. For sharded L1 tensors the two checks differ — a DFB can pass
-        // here against the full-tensor size and still fail per-bank later. By design.
+        // a Buffer is in hand.
+        //
+        // Sharded specs are skipped here, because the two quantities are not comparable: a DFB is a
+        // per-node object and a borrowed DFB views ONE shard, while compute_packed_buffer_size_bytes()
+        // is the whole tensor's logical packed size. A shard shape padded relative to the logical
+        // volume makes the per-node shard legitimately larger than the whole-tensor packed size --
+        // e.g. a height-sharded [1, 1, 16, 128] bf16 tensor with a 32-row shard allocates 8192 B per
+        // node against a 4096 B packed size -- so comparing them rejects specs the per-bank check
+        // accepts. The per-bank check (against Buffer::aligned_size_per_bank) is the authoritative
+        // one and still runs for these; only this early, approximate one is skipped.
+        if (tensor_spec.memory_config().is_sharded()) {
+            continue;
+        }
         const size_t dfb_bytes = static_cast<size_t>(dfb.entry_size) * static_cast<size_t>(dfb.num_entries);
         const size_t tensor_bytes = tensor_spec.compute_packed_buffer_size_bytes();
         TT_FATAL(


vsureshTT:
Closing as unneeded — this fix was motivated by the CI red on #54214, which turned out to be a false flag: no test actually failed. The `ttnn fused group [bh_p150b_civ2]` jobs hit the 40-minute step timeout on a slow runner (the whole suite ran uniformly ~1.9x slower than the same job on other branches, which passes in 18–39 min). The borrowed-DFB spec-time check was never hit in that run.

PR #54240, opened 2026-08-24
fix(sfpu): atan2(inf, 0) returns 0/pi instead of pi/2 (IEEE 754 special case)

zestoles:
﻿## Problem

`ttnn.atan2(±inf, ±0)` returned `±0` or `±π` instead of the IEEE 754-mandated `±π/2` (torch.atan2 reference behavior):

| call | expected (IEEE / torch) | main returned |
| --- | --- | --- |
| `atan2(+inf, +0)` | `+π/2` | `+0` |
| `atan2(-inf, +0)` | `-π/2` | `-0` |
| `atan2(+inf, -0)` | `+π/2` | `+π` |
| `atan2(-inf, -0)` | `-π/2` | `-π` |

## Root cause

In `_sfpu_atan2_`, the both-zero rescue that sets `r = 0` was guarded by `min == 0`:

```cpp
v_if(min == 0.0f) {
    // if both zero, then r = ±0
    r = 0.0f;
}
```

When `|y| = inf` and `|x| = 0`, `min = 0` but `max = inf`. The π/2 branch had
already computed the correct result, but the rescue fired anyway and
overwrote it. The rescue is meant for the case where *both* operands are
zero — and since `min <= max` with both non-negative, `max == 0` is the
correct both-zero condition (it also implies `min == 0`).

## Fix

Change the guard to `max == 0.0f` in all three architecture copies
(wormhole_b0, blackhole, quasar). For the genuine both-zero cases the
behavior is unchanged; for `|y| = inf, |x| = 0` the previously computed
`π/2` result is now preserved (and the sign correction for `x = -0` then
yields `±π/2` as required).

## Testing

Added `test_binary_atan2_special_values` (fp32): covers
`atan2(±inf, ±0)`, `atan2(±inf, ±inf)`, `atan2(finite, ±0)`,
`atan2(±0, finite)` against the torch golden with exact equality
(`torch.testing.assert_close`). The existing random-range test never
generates inf/zero pairs, which is why this divergence went unnoticed.

Fails on current main for the four `inf/zero` cells; passes with this change.

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_binary_composite.py
@@ -99,6 +99,34 @@ def test_binary_atan2_ttnn(input_shapes, device):
 
 @pytest.mark.parametrize(
     "input_shapes",
+    ((torch.Size([1, 1, 32, 32])),),
+)
+def test_binary_atan2_special_values(input_shapes, device):
+    """Regression test: atan2(±inf, ±0) must return ±π/2 per IEEE 754.
+
+    The kernel's both-zero rescue checked `min == 0`, which also fired when
+    only |x| was zero and |y| was infinite, overwriting the correct π/2
+    result with 0 (or π when x was -0).
+    """
+    y_vals = [float("inf"), float("-inf"), 1.0, -1.0, 0.0, float("inf"), float("-inf"), 2.5]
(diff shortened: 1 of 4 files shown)


zestoles:
Fixes #54241

mateusznowakTT (approved):

fvranicTT (approved):

Aswinmcw (approved):

PR #54248, opened 2026-08-25
sweep_rank_binding_solutions: fix producer tcp-interface plumbing + NFS index race

Riddy21:
### Problem

Two multi-host reliability gaps in the sweep driver, both found while root-causing single-galaxy passthrough-ring hangs across four aisles:

**1. The producer's mpirun has no interface configuration.** `--tcp-interface` is only forwarded to the per-solution `tt-run` invocations (tt-run synthesizes `--mca btl self,tcp --mca btl_tcp_if_include ... --prtemca oob_tcp_if_include ...` from it). The streaming producer (`generate_rank_bindings --all-solutions`) is a separate mpirun built by `_build_producer_cmd` → `build_generate_rank_bindings_mpi_cmd`, which had no tcp-interface path at all. Multi-host producers therefore only worked when `OMPI_MCA_*` env vars happened to be exported; otherwise OMPI auto-picks interfaces per host, cross-host connection setup fails ("`server accept cannot find guid`"), and the sweep sits at `0 found` forever. Single-host sweeps never noticed (local ranks, no TCP).

**2. Producer-exit NFS index race.** The producer may run on a different NFS client than the consumer. On producer exit the consumer did a single read of `solutions_index.yaml`; NFS attribute-cache lag (close-to-open consistency, up to ~60 s) can serve a stale view, and the sweep wrongly concludes `0 solution(s)` and errors out — even though the index on disk already lists solutions. Reproduced deterministically when the producer exits quickly (single solution) and, worse, when the output dir is reused and a stale `found: 0` index exists to be read.

### Fix

1. Thread `tcp_interface` into `_build_producer_cmd` and append the same three MCA flags tt-run generates to the producer's `mpi_args`.
2. On producer exit with nothing consumed yet, re-poll the index for up to 90 s (`PRODUCER_EXIT_INDEX_SETTLE_S`) before concluding an empty sweep. No-op on healthy runs: the settle path is only entered when the sweep would otherwise report zero.

### Validation

- Dry-run confirms the producer command now carries `--mca btl self,tcp --mca btl_tcp_if_include <if> --prtemca oob_tcp_if_include <if>`.
- The settle-poll logic was previously validated on hardware sweeps where a host-deterministic index race caused 3× consecutive false "0 solutions" verdicts; with the poll, the same cell passed on first try.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ridvan/sweep-producer-tcpif-nfs-race)

### DIFF
--- a/tools/scaleout/src/generate_rank_bindings.cpp
+++ b/tools/scaleout/src/generate_rank_bindings.cpp
@@ -974,17 +974,21 @@ int main(int argc, char** argv) {
                 std::set<std::set<std::string>> seen_host_sets;
                 const std::filesystem::path index_path = output_dir / "solutions_index.yaml";
 
-                // emitted counts solutions returned by the enumerator (matches the batch --max-solutions cap
-                // semantics, which bound the enumeration before host-set dedup); cap_reached distinguishes
-                // "stopped at the cap (more may exist)" from "enumeration genuinely exhausted".
-                // Safety cap for the unbounded (max_solutions==0) case: match the batch solver's internal
-                // enumeration cap so the streaming path can't run beyond the documented bound / unbounded disk.
+                // Cap semantics: WITHOUT --distinct-host-sets, the cap bounds raw enumerator outputs
+                // (matches the batch --max-solutions semantics). WITH --distinct-host-sets, the cap bounds
+                // WRITTEN (distinct) solutions instead: the SAT enumeration explores near-neighbor
+                // assignments first, so many consecutive raw solutions share one host set -- counting raw
+                // outputs starves the mode of exactly the host-set variety it exists to produce (e.g. 20
+                // raw solutions collapsing to 7 distinct sets, all anchored on the first host). The raw
+                // safety cap still bounds total solver work in both modes.
+                // cap_reached distinguishes "stopped at the cap (more may exist)" from "genuinely exhausted".
                 constexpr std::size_t kEnumerationSafetyCap = 500000;
                 const std::size_t effective_cap = args.max_solutions != 0 ? args.max_solutions : kEnumerationSafetyCap;
                 std::size_t emitted = 0;
                 bool cap_reached = false;
                 while (true) {
-                    if (emitted >= effective_cap) {
+                    const std::size_t capped_count = args.distinct_host_sets ? index_entries.size() : emitted;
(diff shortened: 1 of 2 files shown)


jpanasiukTT (approved):

Riddy21:
Addressed the review in 2f2086241a8 (single follow-up commit, code fixes only):

- **MCA precedence** (Copilot): synthesized defaults are now *prepended* so explicit `--mpi-args` win under Open MPI's later-flag-wins rule, matching tt-run's ordering.
- **`--sweep-timeout` overrun** (Copilot): the settle deadline is clipped to the remaining sweep budget.
- **Partial-consumption gap + simplification**: the nested settle loop is gone — after producer exit the *main consume loop* keeps polling until the settle window (anchored at exit time) expires, so late-visible trailing solutions are consumed through the normal path, and the zero-consumed case is no longer special. Fast break when the freshest read shows every enumerated solution consumed, so healthy runs wait 0s.
- **Silent expiry**: a warning is logged when the window expires with zero consumed, naming NFS index-visibility lag as the likely cause.
- **Magic numbers**: poll interval is a named constant (`PRODUCER_EXIT_INDEX_POLL_INTERVAL_S = 3.0`); the 90s constant documents its derivation (~60s NFS `acregmax` attribute-cache lag + 30s margin).

PR #54249, opened 2026-08-25
[Metal 2.0] Port `transformer/sdpa_decode`

### LINKED ISSUE #54204: Metal 2.0 port: transformer/sdpa_decode
Port `ttnn/cpp/ttnn/operations/transformer/sdpa_decode` to Metal 2.0 / Host 2.0 APIs (Gen1 WH/BH target).

Parent: #41908

## Scope
One `SdpaDecodeDeviceOperation` (`ttnn::prim`), a single `create_descriptor` program factory (`device/sdpa_decode_program_factory.cpp`; paged / MLA / sharded are internal branches, not separate factories), and three op-owned kernels, all over the full `core_grid`:
- reader `device/kernels/dataflow/reader_decode_all.cpp`
- writer `device/kernels/dataflow/writer_decode_all.cpp`
- compute `device/kernels/compute/sdpa_flash_decode.cpp`

The four nanobind entry points (`scaled_dot_product_attention_decode`, `paged_scaled_dot_product_attention_decode`, `flash_multi_latent_attention_decode`, `paged_flash_multi_latent_attention_decode`) all route to this one device operation.

## Audit verdict: GREEN — whole-op port cleared
The pre-port audit clears all five gates (no portable-subset scoping needed):
- **TTNN factory concept** — `descriptor`, `Is able to port?` = yes, target `ProgramSpecFactoryConcept`. No custom `compute_program_hash` (removed), no `get_dynamic_runtime_args`, no `override_runtime_arguments`, no pybound `create_descriptor`, no op-owned tensors. Delivery is the `Buffer*` `BufferBinding` ("pointer-patching") form.
- **Device 2.0** — the op's three kernels and every donor header are already `Noc` / `CircularBuffer` / `Semaphore<>` / `TensorAccessor` based (the only CB-index free functions are the sanctioned `get_tile_size(cb)`).
- **Feature compatibility** — no GlobalCircularBuffer, no non-zero `address_offset`, no GlobalSemaphore.
- **Offset base pointers** — none; buffers bound as clean `Buffer*`, no `->address()` folds.
- **TensorParameter relaxations** — `none`.
- **TensorAccessor 3rd argument** — two sites, both Class 2 (redundant → clean drop, no `dynamic_tensor_shape`).

## Status
Pre-port audit complete; **port not yet started**. `METAL2_PREPORT_AUDIT.md` (team-facing, full record) and `METAL2_PORT_BRIEF.md` (porter-facing) are checked into the op directory.

## Notes
- WH/BH only — no Quasar emulation. Basic Metal 2.0 port (no functional changes expected).
- **Tensor bindings (per config):** `q` is Case 1 (DRAM `TensorAccessor`) / Case 2 (HEIGHT_SHARDED non-MLA — clean base bridged via `get_bank_base_address`, raw core-to-core L1 read kept) / clean borrowed-DFB (MLA `q_locally_available`); `k` / `v` / `attn_mask` / `attention_sink` Case 1; `cur_pos` / `page_table` / `output` Case 1 (interleaved) or clean borrowed-DFB (sharded).
- **CB endpoints:** one **multi-binding** — `c_16` (`cb_out_o` / `cb_out_worker`) is reused bidirectionally by the tree reduction (receive-child O vs. send-parent O → 2 producers + 2 consumers on intermediate nodes when `num_cores_per_head ≥ 4`); set the multi-binding advanced option, do not bind 1P+1C. Self-loops for `c_9` / `c_19` / the compute intermediates; borrowed-DFBs for `c_0` / `c_8` / `c_9` / `c_20` in sharded configs.
- **3rd-arg drops:** the Q accessor (`dataflow_common.hpp:572`) and the page-table accessor in the shared sdpa donor `read_page_table_for_batch`.
- **Shared kernels (coordinate, don't half-convert):** in-family donors shared with `transformer/sdpa` (prefill); `ttnn/kernel/dataflow/generate_bcast_scalar.hpp` already has a `_metal2` `DataflowBuffer` fork to bind.
- **Latent anomaly (not fixed by the port):** `c_11` (col_identity) is produced by the writer but unconsumed in sdpa_decode (dead code carried from the sdpa-prefill matmul-reduce path) — flagged to the ops team.

gajanan-choudhary:
### Issue

Closes #54204.

### Summary
Ports `ttnn/cpp/ttnn/operations/transformer/sdpa_decode` from the legacy direct-descriptor `ProgramDescriptor`
API to the Metal 2.0 `ProgramSpecFactoryConcept`, per the pre-port audit (GREEN, whole-op). The single
`SdpaDecodeDeviceOperation` factory — paged / MLA / sharded / sliding-window / attention-sink / geometry-override
are internal branches — is converted as one atomic unit, along with its three kernels (reader / writer /
compute) and the op's `dataflow_common.hpp`. All four nanobind entry points route through it unchanged.

Highlights:

- **Direct-descriptor → conventional factory** (exception 3): factory body moved into a nested
  `SdpaDecodeProgramFactory` struct with `program_factory_t = std::variant<...>` + `select_program_factory`;
  `create_descriptor` removed.
- **Typed bindings** replace every legacy idiom: `DataflowBufferSpec`/`DFBBinding` (magic CB indices),
  `TensorParameter`/`TensorBinding` (buffer-address RTAs + `TensorAccessorArgs` plumbing), `SemaphoreBinding`
  (semaphore-id args). Named CTAs/RTAs throughout; varargs kept only for the data-indexed physical-core-coordinate arrays.
- `c_16` **tree-reduction multi-binding**; borrowed DFBs (MLA-local Q, sharded cur_pos / page_table / output);
  Case-2 sharded-Q base pulled via `get_bank_base_address`; conditional DFB/tensor bindings gated by compiler defines.
- **Full-grid placement preserved** — `WorkUnitSpec {reader, writer, compute}` targets the entire `core_grid`
  (`full_node_set`); idle cores are kept on the grid and skip work via the `do_reduce == 65` sentinel (see below).
- `unpack_modes` set for Float32 compute-consumed DFBs under `fp32_dest_acc_en`; compute `opt_level` pinned to `O3`.
- Kernels use `DataflowBuffer` + `dfb::` / `tensor::` / `sem::` handles.

No functional change intended (a syntax swap onto the Metal 2.0 host API).

### Notes for reviewers

- **Full-grid placement + idle-core sentinel (legacy behavior preserved)** — legacy placed all three kernels on
  the full `core_grid` and keyed idle detection off address RTAs zeroed to `0` for idle cores (reader `q_addr`,
  writer `out_addr`). Metal 2.0 injects buffer addresses via `TensorBinding`s (never `0`), so that idle signal
  is gone. The port keeps legacy full-grid placement: `WorkUnitSpec {reader, writer, compute}` targets
  `full_node_set`, DFBs and semaphores follow, and the idle-args loop remains. Idle cores receive
  `do_reduce == 65` (out of the valid `{0,1}` range — the same sentinel the legacy compute kernel used) in all
  three kernels, which early-return before touching bindings or semaphores. Unread varargs on idle cores are
  zero-filled to the active layout.
- **Shared donor edit** (`transformer/sdpa/device/kernels/dataflow/dataflow_common.hpp`) — an **additive**
  Metal 2.0 overload of `read_page_table_for_batch` (`DataflowBuffer&` + a caller-built
  `TensorAccessor(tensor::page_table)`) plus a `dataflow_buffer.h` include. The original 6-arg overload is
  untouched, so sdpa **prefill**'s legacy callers are unaffected (verified with a `test_sdpa_prefill.py`
  sample). Coordinated with the parallel `sdpa` port.
- **Left as-is (flagged for ops team)** — `c_11` (`col_identity`) is dead in sdpa_decode (the writer fills it,
  nothing consumes it; it is consumed only by sdpa prefill's `matmul_reduce`). Preserved as a single-toucher
  self-loop DFB, not removed (out of port scope).

Tested on Wormhole (unit + nightly): `test_sdpa_decode` 11 passed / 1 skipped, `flexible_geometry` 10,
`bounded_sliding_kv_cache` 7, `mla_decode` 2, nightly `sink` 15/3, nightly `cache` 10, nightly `test_sdpa_decode`
74/14; a sdpa-prefill sample green. Host-side legality checks were forced-on and proven live during every run.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:gchoudhary/54204/metal2-transformer/sdpa_decode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:gchoudhary/54204/metal2-transformer/sdpa_decode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:gchoudhary/54204/metal2-transformer/sdpa_decode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:gchoudhary/54204/metal2-transformer/sdpa_decode)


ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
```diff
@@ -485,352 +490,633 @@ ProgramDescriptor SdpaDecodeDeviceOperation::create_descriptor(
...
+    // c_0 q_in / c_10 q_rm — config-flip.
+    //   tilize_q  : reader fills q_rm (P); compute tilizes q_rm->q_in (q_in self-loop), consumes q_rm.
+    //   !tilize_q : reader fills q_in (P); compute consumes q_in. q_rm not used (dropped).
+    add_dfb(
+        DFB_Q_IN,
+        q_tile_size,
+        q_tiles,
+        q_df,
+        &q_tile,
+        q_locally_available ? std::optional<TensorParamName>(Q) : std::nullopt);
```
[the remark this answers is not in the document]
gajanan-choudhary: Fixed in e8e23587bd9. Legacy `add_cb(c_10, ...)` passes no buffer, so `q_rm` is never borrowed — the port's `borrowed_from = Q` on it was the divergence; `q_rm` no longer borrows (`c_0` still borrows `Q` under `q_locally_available`, matching legacy).

Not taking the suggested variant (borrow `q_rm` in the tilize path / make `c_0` conditional): `c_0` borrows unconditionally in legacy, and the port's job is to preserve legacy borrow topology rather than redesign it. The aliasing combination is also unreachable in practice — `q_locally_available` comes only from MLA, whose Q is tiled (`tilize_q=false`), so the borrow never actually triggered — but removing the divergence is correct regardless.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/dataflow/reader_decode_all.cpp
```diff
@@ -181,57 +183,97 @@ void kernel_main() {
...
+    {
+        DataflowBuffer dfb_q(dfb_q_in);
+        dfb_q.reserve_back(q_chunk_tiles);
+        dfb_q.push_back(q_chunk_tiles);
+    }
+#endif
+#else
+    // Not locally available: read Q from the output core's L1 (sharded) or from DRAM (interleaved).
+    {
+        const auto q_reader = TensorAccessor(tensor::q);
```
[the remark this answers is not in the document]
gajanan-choudhary: Checked this against the reachable configs — I don't think it's a bug here.

The DRAM `TensorAccessor(tensor::q)` page-read path is reached only for **tiled** Q. Row-major Q is always created HEIGHT_SHARDED (validation allows row-major Q, and the test helpers build it with a row-major shard spec), so it takes the sharded `get_bank_base_address` path — not this page-read path. For tiled Q the binding's aligned page size equals the full-tile size the legacy 3rd argument supplied, which is exercised by the DRAM-Q configs across the unit + 74/14 nightly runs (a wrong stride would corrupt Q and fail PCC; none did).

A genuinely interleaved row-major Q isn't a reachable input, and reading tile-sized pages from a row-major DRAM buffer is equally ill-defined under the legacy full-tile override — so "legacy correct / port wrong" doesn't hold. No change; not adding a row-major-interleaved regression since that layout isn't a supported non-sharded Q.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -30,111 +32,127 @@ void kernel_main() {
...
-    for (uint32_t r = 0; r < MAX_TREE_REDUCTION_ROUNDS; ++r) {
-        children_per_round[r] = get_arg_val<uint32_t>(arg_idx++);
-    }
-
-    // Idle core
-    // get_arg_val<uint32_t>(0) can go from 0-63 for the core_num; for active cores 65 is out of range so 65 indicates
-    // an idle_core
-    if (get_arg_val<uint32_t>(0) == 65) {
-        return;
-    }
```
gajanan-choudhary, line 142: I should revisit this. It looks like a behavioral change has been made here. The port report did contain a statement about this:
```
- **Placement change (behavior-preserving, flagged for reviewer attention).** Legacy placed all three
  kernels on the full `core_grid` and had idle cores early-return via the address RTA that becomes `0` for
  idle cores (`q_addr==0` reader, `out_addr==0` writer, `arg(0)==65` compute). Those address RTAs became
  `TensorBinding`s (auto-injected, never 0), so the idle signal vanished. The port instead places the single
  `WorkUnitSpec {reader, writer, compute}` on the **active core set** (`core_group`, `num_active_cores`
  cores) and drops the idle-args host loop and the three kernel idle early-returns. This is behavior-
  preserving: idle cores did no work, and no cross-core operation (K-multicast, tree reduction, sharded-output
  gather) ever targets an idle core -- all targets are active reducer / output / reduction-group cores. It
  also satisfies "do not create kernels on unused cores." Reviewers should confirm no downstream depends on
  idle cores having the program loaded (none found).
```

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -269,408 +292,342 @@ void kernel_main() {
...
+        dfb_cur_sum = dfb_sum_1;
+        dfb_prev_sum = dfb_sum_2;
 
-        /******************************************************************************
-         *                           FLASH ATTENTION LOOP                             *
-         ******************************************************************************/
-        /**
-         * Compute Parameters (most are compile time but some are dynamic):
-         * @tparam St - Total sequence length in tiles
-         * @tparam DHt - Head dimension in tiles
```
gajanan-choudhary, line 296: These comments around here not have been removed and just updated with variable name changes, if any.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -208,41 +229,43 @@ void kernel_main() {
...
-    } else {
-        compute_kernel_hw_startup<SrcOrder::Reverse>(cb_q_in, cb_k_in, cb_qk_im);
-        matmul_init(cb_q_in, cb_k_in);
-    }
-    CircularBuffer(cb_q_in).wait_front(q_chunk_tiles);
+#ifdef TILIZE_Q
+    compute_kernel_hw_startup(dfb_q_rm, dfb_q_in);
+    // Keep InitAndUninit: the helper picks fast- vs regular-tilize at compile time and its
+    // teardown must match (fast_tilize_uninit vs tilize_uninit). tilize_q with a FULL-tile Q
+    // (use_half_tile==false, e.g. >16 heads) can take the fast-tilize path, so we must not
```
gajanan-choudhary:
```cpp
    if constexpr (tilize_q) {
        // code ...
    }
```
was replaced with
```cpp
#ifdef TILIZE_Q
    // code ...
#endif
```
that reduced the indentation and introduced a whole lot of diff'd lines. It's better to keep the scope by making the change as following:
```cpp
#ifdef TILIZE_Q
    {
    }
#endif
```

Also, the comments should not have been removed here, only updated if needed.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -269,408 +292,342 @@ void kernel_main() {
...
 
         /******************************************************************************
          *                      TREE REDUCTION LOGIC                                  *
          ******************************************************************************/
-        /**
-         * Tree reduction reduces the online softmax results in O(log n) rounds.
-         *
-         * For each round r (0 to my_active_rounds-1):
-         *   - If children_per_round[r] != UINT32_MAX, receive from that child
-         *   - Combine received data with local accumulator using softmax correction
```
gajanan-choudhary, line 533: Comments around here shouldn't have been removed

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
```diff
@@ -485,352 +490,632 @@ ProgramDescriptor SdpaDecodeDeviceOperation::create_descriptor(
         use_attention_sink,
         use_half_tile);
 
-    // Print reducer core coordinates
-    log_debug(tt::LogOp, "Reducer cores ({}):", num_reducer_cores);
-    for (uint32_t i = 0; i < num_reducer_cores; ++i) {
-        log_debug(
```
gajanan-choudhary, line 491: debug prints shouldn't have been removed around here

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
```diff
@@ -485,352 +490,632 @@ ProgramDescriptor SdpaDecodeDeviceOperation::create_descriptor(
...
+        // c_10 (q_rm) is never borrowed — matching legacy add_cb(c_10, ...) which passes no buffer. Only c_0
+        // (the tilize output / borrowed-Q input) borrows Q under q_locally_available, as legacy does. (A
+        // borrow here would alias the tilize source and output when q_locally_available && tilize_q — a
+        // combination MLA never produces, since MLA Q is tiled — but the port must not add a borrow legacy
+        // lacks.)
+        add_dfb(DFB_Q_RM, q_tile_size, q_tiles, q_df, &q_tile);
+        bind(reader_dfb, DFB_Q_RM, "q_rm", DFBEndpointType::PRODUCER);
+        bind(compute_dfb, DFB_Q_RM, "q_rm", DFBEndpointType::CONSUMER);
+        bind(compute_dfb, DFB_Q_IN, "q_in", DFBEndpointType::PRODUCER);
+        bind(compute_dfb, DFB_Q_IN, "q_in", DFBEndpointType::CONSUMER);
```
gajanan-choudhary: Is this intentional? Was it supposed to be
```
        bind(reader_dfb, DFB_Q_IN, "q_in", DFBEndpointType::PRODUCER);
        bind(compute_dfb, DFB_Q_IN, "q_in", DFBEndpointType::CONSUMER);
```
instead?

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -269,408 +292,342 @@ void kernel_main() {
...
-                    mask_cb_to_use = cb_sliding_window_mask_in;  // Use sliding window mask buffer
-                }
+            /* QK = Q_CHUNK @ K_CHUNK */
+            // Determine which mask buffer to use for fusion
+            uint32_t mask_dfb_to_use = dfb_mask_in;  // Default to causal mask buffer
+#ifdef SLIDING_WINDOW
+            if (add_sliding_window_mask_fusion) {
+                mask_dfb_to_use = dfb_sliding_window_mask_in;  // Use sliding window mask buffer
+            }
+#endif
```
gajanan-choudhary: This #ifdef is an optimization the recipe didn't ask for. I'll let the codeowners decide whether to keep this or remove it.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
```diff
@@ -269,408 +292,342 @@ void kernel_main() {
...
-                pack_reconfig_data_format(cb_qk_im);
-
-                /**
-                 * sub_exp performs `QK = exp((QK - cur_max) * scale)`
-                 */
-                sub_exp_block_bcast_cols_inplace<cb_qk_im, Sq_chunk_t, scale_fp32, true, false, vector_mode>(
-                    cb_cur_max, cb_cur_sum, Sk_chunk_t_dynamic);
-                CircularBuffer(cb_qk_im).wait_front(qk_chunk_tiles_dynamic);
+                // Apply sliding window mask to the first chunk (only on the core that processes it)
+#ifdef SLIDING_WINDOW
```
gajanan-choudhary: This `#ifdef` is an optimization the recipe didn't ask for. I'll let the codeowners decide whether to keep this or remove it.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
gajanan-choudhary: Restored the three `log_debug` loops (reducer / output / reduction-group core coordinates).


Fixed in 0c229f2f64e.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: This one isn't an optimization — it's forced by the Metal 2.0 binding model. The block references `dfb_sliding_window_mask_in`, whose `dfb::sliding_window_mask_in` token is only generated when the host binds that DFB (`sliding_window_size > 0`). In legacy this was a plain runtime `if` because `c_13` is a `constexpr` CB index that always exists; a `dfb::` handle does not. Without the `#ifdef` the kernel fails to compile whenever sliding window is disabled — this is the conditional-DFB gating pattern from the port recipe, not an added optimization. Same for the `#ifdef` at :378.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: Restored the `FLASH ATTENTION LOOP` banner and the `@tparam`/`@param` doc block (updated `cb_` → `dfb_` in the buffer-parameter names).


Fixed in 0c229f2f64e.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
gajanan-choudhary: Intentional — this is the config-flip documented in the comment just above the block. Under `tilize_q`, compute *tilizes* `q_rm` → `q_in` and then consumes `q_in`, so `q_in` is a compute self-loop: `bind(compute_dfb, DFB_Q_IN, PRODUCER)` + `bind(compute_dfb, DFB_Q_IN, CONSUMER)`. The reader fills `q_rm` (bound two lines up), not `q_in`.

The form I suggested — `bind(reader_dfb, DFB_Q_IN, PRODUCER); bind(compute_dfb, DFB_Q_IN, CONSUMER)` — is exactly the `!tilize_q` `else` branch right below, where the reader delivers already-tiled Q straight into `q_in`.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: Restored the brace scope: the block now reads `#ifdef TILIZE_Q { ... } #endif` (same shape as the `HAS_BLOCK_PADDING` block just below), so the body keeps its original 8-space indentation and the diff on those lines is just the `cb_` → `dfb_` rename. The in-block comments were already retained.

Kept `#ifdef` rather than `if constexpr (tilize_q)` on purpose: the tilize branch references `dfb_q_rm`, whose `dfb::q_rm` token is only generated when the host binds it under `tilize_q`. In this non-template `kernel_main`, `if constexpr` still name-looks-up the discarded branch, so it would fail to compile when `tilize_q` is off. The output-untilize block a few lines down *was* an unnecessary `#ifdef` (both branches touch only always-bound DFBs) — I reverted that one back to `if constexpr (untilize_output)` to match legacy.


Fixed in 0c229f2f64e.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: Restored the tree-reduction `O(log n)` doc block, the 6-step Fused Correction comment, and the "which sum/max buffer to use" note (`cb_` → `dfb_` updated).


Fixed in 0c229f2f64e.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: There's a trade-off here:

The port narrows kernel placement from the full core grid to the active core set (`WorkUnitSpec` `target_nodes`), which drops the legacy idle-core early-returns — the reader/writer `addr == 0` sentinels and this compute `get_arg_val(0) == 65` check. Output is identical because idle cores did no work, and it satisfies the "no kernels on unused cores" guideline — but it IS a structural change from legacy, not a pure syntax swap.

I can revert to full-grid placement and re-add the idle machinery, reusing the `== 65` sentinel across all three kernels. Will decide later.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: Same as :334 — required, not an optimization. The `add_block_inplace` here references `dfb_sliding_window_mask_in`; its `dfb::sliding_window_mask_in` token only exists when the host binds the sliding-window mask DFB (`sliding_window_size > 0`), so the reference must be `#ifdef SLIDING_WINDOW`-gated or it won't compile when sliding window is off.

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/kernels/compute/sdpa_flash_decode.cpp
gajanan-choudhary: Fixed in 88db943e0d8 — reverted the divergence to be faithful to legacy.

Correction to my earlier reply: I should not have cited "no kernels on unused cores" as support for the active-core placement. That's was a local CLAUDE.md rule I had as a guideline for *writing new ops*, not part of the port recipe — and the port's governing rule is fidelity, which argues the other way. It was the wrong justification.

What the fix does:
- Places the WorkUnitSpec and the reducer/output/k_mcast semaphores on the full `core_grid` (DFBs follow the WorkUnitSpec), matching legacy.
- Emplaces idle runtime args for `core_group_idle` so the schema is satisfied on those nodes.
- Restores the idle early-return in all three kernels.

Idle-signal note: legacy detected idle via `q_addr == 0` (reader) / `out_addr == 0` (writer). Metal 2.0 injects buffer addresses through TensorBindings, so there's no addr RTA to zero. I reused the compute kernel's existing `do_reduce == 65` marker (65 is out of the valid {0,1} range) across reader/writer/compute: idle cores get `do_reduce = 65` and every kernel returns before touching a binding or semaphore.

Full baseline re-run, all green and unchanged: unit sdpa_decode 11/1skip, mla_decode 2, bounded_sliding_kv_cache 7, flexible_geometry 10; nightly sdpa_decode 74/14skip, cache 10, sink 15/3skip. No hangs (idle cores exit cleanly).

ttnn/cpp/ttnn/operations/transformer/sdpa_decode/device/sdpa_decode_program_factory.cpp
```diff
@@ -871,168 +1175,218 @@ ProgramDescriptor SdpaDecodeDeviceOperation::create_descriptor(
...
+    ProgramSpec spec{
+        .name = "sdpa_decode",
+        .kernels = {std::move(reader), std::move(writer), std::move(compute)},
+        .dataflow_buffers = std::move(dfbs),
+        .semaphores = std::move(semaphores),
+        .tensor_parameters = std::move(tensor_params),
+        .work_units = {WorkUnitSpec{
+            .name = "main",
+            .kernels = {READER, WRITER, COMPUTE},
+            .target_nodes = full_node_set,
```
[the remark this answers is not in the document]
gajanan-choudhary: Updated the PR description to document the retained full-grid placement (`full_node_set`) and the `do_reduce == 65` idle-core sentinel across all three kernels. The superseded active-core-set wording has been removed.

gajanan-choudhary:
Ran the two per-op prototype tests for `ttnn.experimental.quasar.transformer.{scaled_dot_product_attention_decode,paged_scaled_dot_product_attention_decode}` on Wormhole (n150):

```
models/experimental/llama32_1b_quasar/tests/prototype_ops/test_scaled_dot_product_attention_decode.py
2 passed (batch1, batch32) in 11.06s

models/experimental/llama32_1b_quasar/tests/prototype_ops/test_paged_scaled_dot_product_attention_decode.py
2 passed (batch1, batch32) in 5.93s
```

Both pass.

bbradelTT (approved):
Please also update and run tests in models/experimental/llama32_1b_quasar/tests/graph_ops/
 and models/experimental/llama32_1b_quasar/tests/ops/ (technically you could skip running the latter since they'd be equivalent to the ones in prototype_ops).

gajanan-choudhary:
> Please also update and run tests in models/experimental/llama32_1b_quasar/tests/graph_ops/
and models/experimental/llama32_1b_quasar/tests/ops/ (technically you could skip running the latter since they'd be equivalent to the ones in prototype_ops).

Done in 7e03afa7a3a and 57e605fb149

PR #54260, opened 2026-08-25
docs: clarify PyTorch dependency for TT-NN tutorials

### LINKED ISSUE #44431: Clarify whether pip install ttnn is sufficient for TT-NN tutorials, and document the required PyTorch dependency/version
### Describe the documentation issue

The current install documentation says wheel install is:

```bash
pip install ttnn
```

and then describes `tt_metal/python_env/requirements-dev.txt` as extra setup for "models users only."

However, the same install page points users toward TT-NN examples immediately after installation, including `python3 -m ttnn.examples.usage.run_op_on_device`.

That creates confusion for users who are not trying to run models, but do want to run TT-NN tutorials and notebooks.

Several TT-NN tutorials directly require PyTorch usage, not just TT-NN alone:

- `ttnn_basic_operations.py` imports `torch`, uses `torch.rand(...)`, and converts tensors with `ttnn.from_torch(...)`.
- `ttnn_multihead_attention.py` imports `torch`, creates torch tensors, and converts them to TT-NN tensors.
- `ttnn_tracer_model.py` also imports `torch` and uses torch-to-TTNN flows in the tutorial path.

Some tutorial flows also rely on `torchvision` or `transformers`.

The repo also contains a tutorial/notebook test harness that executes files under `ttnn/tutorials` and `ttnn/tutorials/basic_python`, which makes these dependencies feel like part of the normal tutorial experience rather than something only relevant to model users.

At the same time, the repo-managed Python environment pins:

- `torch==2.7.1`
- `torchvision==0.22.1`

That suggests there is an expected or validated Torch version, but I could not find that clearly documented for users installing `ttnn` from pip.

I also ran into Jupyter kernel hangs while testing `ttnn.rand()` on device in a pip-installed environment. I was running on a QuietBox 2 through SSH, and I did not observe any SSH instability or connection hangs. The Jupyter kernel hang appeared to come from the code path itself rather than from the SSH session.

Overall, the docs are unclear on this point:

- Is `pip install ttnn` supposed to be enough for TT-NN tutorial and notebook users?
- Or is cloning `tt-metal` and using the repo-managed environment the recommended path even for non-model users?

### Describe the documentation solution you want to see

I would like the install documentation to clearly state one of the following:

## Option 1

`pip install ttnn` is a supported path for tutorial/basic-example/notebook users.

If that is the expectation, please document:

- the required PyTorch dependency
- the recommended PyTorch version
- any other expected Python dependencies for tutorial use

## Option 2

`pip install ttnn` is not the recommended path for tutorial/basic-example/notebook users.

If that is the expectation, please explicitly say that users who want to run TT-NN tutorials should clone `tt-metal` and use the repo-managed environment, even if they are not running models.

## Specific clarification I would like

A short note near the install section would help, for example:

"Users who want to run TT-NN tutorials and notebooks should use the repo-managed Python environment because many tutorials depend on PyTorch and related packages."

or

"`pip install ttnn` supports tutorial usage, but users should also install PyTorch version X.Y.Z and any listed tutorial dependencies."

### Describe the diagrams or screenshots (if any) that the documentation should have

_No response_

vamsilearner:
## Summary

Closes #44431

The Binaries install section of INSTALLING.md labels `requirements-dev.txt` as **"for models users only"**, but several TT-NN tutorials also require PyTorch:

- `ttnn_basic_operations.py` — imports `torch`, creates `torch.rand()`, converts with `ttnn.from_torch()`
- `ttnn_multihead_attention.py` — imports `torch`, creates torch tensors
- `ttnn_tracer_model.py` — imports `torch` and uses torch-to-ttnn flows

Users following the current docs step-by-step hit `ModuleNotFoundError: No module named 'torch'` when running those tutorials.

## Changes

- Added a `[!NOTE]` block after `pip install ttnn` explaining that PyTorch is needed for tutorial code and how to install it via `requirements-dev.txt`
- Changed the Step 2 heading from *(For models users only)* to *(For models and tutorial users)*

## Testing

- Verified the note renders correctly in GitHub-flavored markdown
- Verified the exact torch version cited (`2.11.0`, CPU-only on x86_64) matches `requirements-dev.txt`

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -83,7 +83,22 @@ All binaries support only Linux and distros with glibc 2.34 or newer.
   pip install ttnn
   ```
 
-#### Step 2. (For models users only) Set Up Environment for Models:
+> [!NOTE]
+> The `ttnn` wheel does not bundle PyTorch by default. Several TT-NN tutorials (for example
+> `ttnn_basic_operations.py`, `ttnn_multihead_attention.py`, and `ttnn_tracer_model.py`) create
+> `torch` tensors and convert them with `ttnn.from_torch()`. If you plan to run those
+> tutorials or any code that imports `torch`, install the development dependencies before
+> continuing:
+>
+> ```sh
+> pip install -r tt_metal/python_env/requirements-dev.txt
+> ```
+>
+> This installs PyTorch (`torch==2.11.0`, CPU-only on x86_64) and other packages referenced
+> by the tutorials. You can skip this step if you only use TT-NN's own tensor creation
+> APIs such as `ttnn.full()`, `ttnn.zeros()`, or `ttnn.from_numpy()`.
+
+#### Step 2. (For models and tutorial users) Set Up Environment for Models:
 
 To try our pre-built models in [`tt-metal/models/`](https://github.com/tenstorrent/tt-metal/tree/main/models), you must:
 


blozano-tt:
I don’t personally love the idea of documenting this in INSTALLING.md

That document should be concise and focus on how to install ttnn and/or tt_metal.

The requirements-dev.txt is a kitchen sink ML dependency list. I would not want to install them all by default. The massive list is more to satisfy CI.

Do the tutorials not make the torch dependency clear?

Issue #52966, opened 2026-08-12
[Quasar Resnet] Maxpool caps tiles per reduction to 4 instead of 8


amokanTT:
In maxpool Quasar Resnet tests which have > 4 channel tiles the 0x19 hw assert was appearing, this was worked around using max 4 tiles per reduce instead of max 8.

This should be investigated and fixed so that up to 8 tiles can be used.

This issue has been reported some time ago from branch `bbradel-48552_resnet_qsr`. Since then a lot of fixes have been applied and it is possible that some of the fixed issues also impacted this. Currently, several tests are on main and on `bbradel-52085_resnet_qsr`. This issue is a part of the effort to go through all of these tests, rerun them and determine if the before reported issues still stand and remove hacky workarounds.

amokanTT:
@bbradelTT @rtawfik01 @fvranicTT 

I rerun the test `models/demos/vision/classification/resnet50/quasar/tests/ops/test_max_pool2d_channel_sweep.py` on a local branch (from main) and on `bbradel-52085_resnet_qsr`.
I removed the workarounds applied in the commits `c12193e0f1c/5bcdc8615ad/984b1d15587` back on `bbradel-48552_resnet_qsr `which were then merged to main which have to do with capping reduce to 4 tiles.

On main (from yesterday 11th of Aug) the tests for 32c, 64, 96c, 128c, 160c, 192c, 224c passed. 256c failed with bad PCC.
On `bbradel-52085_resnet_qsr` all the tests passed, including 256c. I removed the `set_dvalid` call from tilize.h on this branch before testing.
I have not seen 0x19 on any of the branches during any of the tests.

I will compare these two branches to see if there are changes that could have fixed this, or if I may have introduced an error on my local branch (from main).

amokanTT:
@bbradelTT @fvranicTT @rtawfik01 

I have been able to see the 0x19 assert (now 0x119) in the following scenarios for max_tiles_per_reduction=8 on `bbradel-52085_resnet_qsr`:
1. Trips assert in `ttnn/cpp/ttnn/operations/experimental/quasar/halo/device/kernels/compute/pack_untilize.cpp`, consistent with `ttnn/cpp/ttnn/operations/experimental/quasar/halo/device/kernels/dataflow/halo_gather.cpp` taking a lot of time due to the debug tools (maybe this is not the only problem?)
```
TT_METAL_WATCHER_DISABLE_ASSERT=0
[...]

PR #54284, opened 2026-08-25
[Quasar Resnet] Use up to 8 tiles per reduction and remove STALLWAIT workaround

amokanTT:
### PR Category
Bug fix, Cleanup

### Issue 
[#52966](https://github.com/tenstorrent/tt-metal/issues/52966)

### Summary
While bringing up Quasar Resnet several hacky workaround were made. This PR is a part of the effort to remove these workarounds.

**1. Removes the cap tiles per reduction to 4 workaround and uses max 8 tiles per reduction instead.**
Having more than 4 tiles per reduction was causing the 0x19 Memory access hang hw assert. This was because a GPR read was sitting behind a MOP which was parked due to GMPOOL waiting for SrcAB dvalid. The SrcAB dvalid was not coming fast enough because the compute was waiting for DM to populate L1 for longer than what the CSR timeout allowed. Increased CSR timeout enables the use of 8 tiles per reduction.

**2. Removes STALLWAIT workaround in compute_pool_2d compute kernel**
A stallwait was added for the following reason: emulator ignores packer_wr_done_wait_mask in PUSH_TILES. This was probably a misdiagnosis and the stallwait was just masking some other issue, which probably got resolved along the way. The tests pass without this stallwait now. 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:amokan/resnet_pool)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:amokan/resnet_pool)

### DIFF
--- a/ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/compute/compute_pool_2d.cpp
+++ b/ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/compute/compute_pool_2d.cpp
@@ -368,16 +368,7 @@ void kernel_main() {
                 const uint32_t curr_scratch_cb_id = scratch_cb_id_0;
                 DataflowBuffer curr_scratch_cb = scratch_cb_0;
 #endif
-                // Model A (wide-reduction fix): the scratch CB holds ONE contiguous full-width (in_ntiles_c)
-                // output stick, so reserve it ONCE per stick (on the first c-block) -- NOT per c-block.
-                // Reserving/pushing a full stick per c-block advanced wr_entry_idx mid-stick, overran the
-                // 2-slot ring, and grew the packer's L1 base (base_l1 = wr_entry_idx * ...) past the buffer-
-                // descriptor limit -> Risc IB interrupt (0x19), with the fault address creeping run-to-run as
-                // base_l1 grew. Each c-block packs its channel slice into this shared stick below; the whole
-                // stick is pushed once on the last c-block so the DM reader consumes exactly one stick per pop.
-                if (first_c_block) {
-                    curr_scratch_cb.reserve_back(scratch_npages);
-                }
+                curr_scratch_cb.reserve_back(scratch_npages);
                 // QSR fix (split-reader second stream): the top-of-kernel pack_untilize_dest_init targets
                 // scratch_cb_0 only, so odd (reader1) sticks packing into scratch_cb_1 used a packer
                 // descriptor still bound to scratch_cb_0 -> the pack landed nowhere and reader1 read an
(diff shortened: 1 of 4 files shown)


bbradelTT (commented):

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/compute/compute_pool_2d.cpp
```diff
@@ -395,41 +386,15 @@ void kernel_main() {
...
-                    pack_untilize_dest_init<partial_iter_output_tiles, in_ntiles_c>(curr_scratch_cb_id);
-                    pack_untilize_dest<partial_iter_output_tiles, in_ntiles_c>(
-                        curr_scratch_cb_id, 1, (c_i * max_tiles_per_iter) / partial_iter_output_tiles);
+                    pack_untilize_dest_init<partial_iter_output_tiles>(curr_scratch_cb_id);
+                    pack_untilize_dest<partial_iter_output_tiles>(curr_scratch_cb_id, 1, 0);
                 } else {
-                    pack_untilize_dest_init<max_tiles_per_iter, in_ntiles_c>(curr_scratch_cb_id);
-                    pack_untilize_dest<max_tiles_per_iter, in_ntiles_c>(curr_scratch_cb_id, 1, c_i);
+                    pack_untilize_dest_init<max_tiles_per_iter>(curr_scratch_cb_id);
+                    pack_untilize_dest<max_tiles_per_iter>(curr_scratch_cb_id, 1, 0);
```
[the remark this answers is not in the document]
amokanTT: Copilot is right about this.
At in_nblocks_c == 1 (≤256 channels) this bug would not manifest. Above that it breaks in two ways: N pushes against one pop, and every block written at index 0, so only the last block's data survives.

I now restored this part of the code to stay the same as it is on current main. This fixes the credit imbalance and the index-0 collision (≤256c tests are unaffected.). 
However, the addressing is still wrong for some widths (for >256c), and always was. The LLK computes the destination as an offset in units of the block width. That works when every block is the same width. The pool kernel narrows its last block to `partial_iter_output_tiles`, so the offset can only be expressed when `partial` divides `c_i × max_tiles_per_iter`. Integer division truncates the rest.

There are two real fixes: make `MAX_TILES_PER_REDUCTION` a divisor of `in_ntiles_c` so every block is the same width, or move the scratch path to the production chunked model and delete the `full_ct_dim / block_c_index` usage entirely.

@bbradelTT We do not test any max pool with >256c and this issue is on the `PACK_TO_SCRATCH` path only. Should we fix this issue completely or should I add some sort of assert to guide against it?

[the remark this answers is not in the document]
bbradelTT: Just add an assert. Once everything is ported to quasar these experimental kernels should be deleted.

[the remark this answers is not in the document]
amokanTT: @bbradelTT @rtawfik01 I added an assert in commit ff56ae1718dc546031ef477512697a8ec49458fb. This assert guards against any cases where `full_ct_dim` > `block_ct_dim`, as some of them will incorrectly address L1.

I am working on a proper fix as well.

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
```diff
@@ -1206,7 +1195,7 @@ ttnn::device_operation::ProgramArtifacts pool2d_create_program_artifacts(
         {"one_scalar_per_core", static_cast<uint32_t>(one_scalar_per_core)},
         {"is_output_tiled", static_cast<uint32_t>(is_output_tiled)},
         {"is_output_block_format", static_cast<uint32_t>(is_output_block_format)},
-        {"force_max_tiles_per_reduction_4", 1u},  // QSR: cap pack_untilize_dest to 4 tiles (PACR0 bounds)
+        {"force_max_tiles_per_reduction_4", 0u},
```
rtawfik01: Can we remove this flag entirely?

rtawfik01 (changes requested):

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
bbradelTT: As long as it's scoped to just experimental/quasar changes, it should be. However, this flag needs to continue to stay in the regular op.

amokanTT: Yes, I was trying to get a fix which wouldn't remove this flag, since `ttnn/cpp/ttnn/operations/pool/grid_sample/device/grid_sample_bilinear_program_factory.cpp` sets this flag to 1 for `compute_pool_2d.cpp` in regular ops.  I wanted the have /experimental/quasar diverge as little as possible from regular ops. 
I will see which path is the best. But this seems like something that could maybe be cleaned up in the regular ops as well.

amokanTT: @rtawfik01 @rtawfik01 
I removed `force_max_tiles_per_reduction_4` but introduced `max_tiles_per_reduction` compile time arg. In order to not have the compute and reader kernels recalculate this on their own and duplicate code.

bbradelTT (approved):

rtawfik01 (approved):

PR #54285, opened 2026-08-25
[Cleanup] Complete the NoC address seam: no composing or splitting outside the backend

vvukomanovicTT:
### PR Category
Cleanup

### Summary
Stacked on #54199. That PR routed the shared address-construction helpers through the arch-selected `noc_address_backend`; a full-tree audit found two remaining patterns that bypass the seam, both of which would silently misbehave once a non-XY backend (Quasar ATT) is selected:

**Compose-from-(x,y):** `cross_node_dfb.h` and `remote_circular_buffer.h` built addresses by hand with `NOC_XY_ENCODING` + `get_noc_addr_helper`. Now routed through `noc_address_backend::worker_address`. The hoisted receiver loop in `remote_circular_buffer.h` keeps its shape by building a base address once and adding offsets — offset arithmetic stays within the local-address bits on every backend.

**Split-to-reissue:** five kernels decomposed a finished 64-bit NoC address back into `{noc_x, noc_y, addr}` with `NOC_UNICAST_ADDR_X/Y`/`NOC_LOCAL_ADDR_OFFSET` — only because `UnicastEndpoint`'s args take coordinates. Under ATT those bits are a window + selector, not coordinates, so the decompose produces garbage. This PR adds `PrecomposedUnicastEndpoint`, whose args carry the full address opaquely, and migrates the five call sites: `data_movement/common/kernels/common.hpp`, `scatter/.../common.hpp`, `tilize_with_val_padding` reader, and the two `experimental/quasar/tilize*` readers (which already run on Quasar).

### Notes for reviewers
- Unlike #54199 this PR is **not** byte-identical by design — removing the pack/unpack round-trips changes (improves) the code. Equivalence is proven differently:
  - Compile-time proofs on the XY backend (checked on WH, BH, and Quasar): `worker_address(x,y,addr) == (NOC_XY_ENCODING(x,y) << NOC_ADDR_COORD_SHIFT) | addr`; split+rejoin == identity (so passing the address opaquely equals the old decompose/recompose); `base + offset == compose-with-offset`.
  - An object-code sweep of 900+ kernel builds against the parent branch: every build byte-identical except the two `cross_node_dfb` test kernels, whose diff is exactly the removed pack/unpack instructions.
- The modified tilize readers were additionally compile-verified standalone with the device compiler (the `tilize/...metal2` variant only via its sibling — it needs host-generated named-arg bindings to compile standalone, a pre-existing limitation).
- With this PR, the only remaining direct XY composition/decomposition sites in device code are: the dispatch kernels (tracked for the dispatch-conversion PR), watcher/`sanitize.h` and fabric (to be compile-rejected under ATT), and the per-arch V1/V2 transport layers (arch-specific by definition).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vvukomanovic/att-pr1p5-seam-coverage)

### DIFF
--- a/tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
+++ b/tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
@@ -17,6 +17,7 @@
 
 #if defined(KERNEL_BUILD) && !defined(COMPILE_FOR_TRISC)
 #include "api/dataflow/noc.h"
+#include "noc_address_backend.h"
 #include "api/dataflow/dataflow_api.h"
 #include "api/dataflow/dataflow_buffer.h"
 #include "api/dataflow/endpoints.h"
@@ -350,9 +351,9 @@ public:
         const uint8_t noc_id = noc.get_noc_id();
         const uint32_t noc_x = xy_base[2 * receiver_idx];
         const uint32_t noc_y = xy_base[2 * receiver_idx + 1];
-        const uint32_t noc_xy = uint32_t(NOC_XY_ENCODING(DYNAMIC_NOC_X(noc_id, noc_x), DYNAMIC_NOC_Y(noc_id, noc_y)));
         *local_sent += num_units;
-        const uint64_t remote_addr = get_noc_addr_helper(noc_xy, (uint32_t)local_sent);
+        const uint64_t remote_addr =
+            noc_address_backend::worker_address(noc_x, noc_y, (uint32_t)(uintptr_t)local_sent, noc_id);
         noc_semaphore_inc<true>(remote_addr, num_units, noc_id);
     }
 
--- a/tt_metal/hw/inc/api/dataflow/endpoints.h
+++ b/tt_metal/hw/inc/api/dataflow/endpoints.h
@@ -17,6 +17,34 @@ struct UnicastEndpoint {
 };
 
 /**
+ * @brief Endpoint for a fully composed unicast NoC address (e.g. a TensorAccessor or
+ * get_noc_addr result, optionally offset within the target). The address is treated as
+ * opaque - it is never decomposed into coordinates - so callers stay independent of the
+ * active NoC address backend.
+ */
+struct PrecomposedUnicastEndpoint {};
+
+template <>
+struct noc_traits_t<PrecomposedUnicastEndpoint> {
+    struct src_args_type {
+        uint64_t noc_addr{};
+    };
+    struct dst_args_type {
+        uint64_t noc_addr{};
+    };
+    template <Noc::AddressType address_type>
+    static auto src_addr(const PrecomposedUnicastEndpoint&, const Noc&, const src_args_type& args) {
+        static_assert(address_type == Noc::AddressType::NOC);
+        return args.noc_addr;
+    }
+    template <Noc::AddressType address_type>
+    static auto dst_addr(const PrecomposedUnicastEndpoint&, const Noc&, const dst_args_type& args) {
+        static_assert(address_type == Noc::AddressType::NOC);
+        return args.noc_addr;
+    }
+};
+
+/**
  * @brief Wrapper around calculating multicast noc address given 2D multicast rectangle and address. This
  * allows direct address to be supplied to NoC apis
  */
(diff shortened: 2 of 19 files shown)


bbradelTT (approved):

tt_metal/hw/inc/api/dataflow/endpoints.h
```diff
@@ -16,6 +16,34 @@ struct UnicastEndpoint {
     }
 };
 
+/**
+ * @brief Endpoint for a fully composed unicast NoC address (e.g. a TensorAccessor or
+ * get_noc_addr result, optionally offset within the target). The address is treated as
+ * opaque - it is never decomposed into coordinates - so callers stay independent of the
+ * active NoC address backend.
+ */
+struct PrecomposedUnicastEndpoint {};
```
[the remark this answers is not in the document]
vvukomanovicTT: the trait itself only forwards a value the type system already guarantees is 64-bit, so a sentinel unit test would mostly re-test the compiler. the behavioral coverage lives device-side (dmloopback and the tilize suites run this path on every misaligned transfer) and the att stack adds host goldens that assert full 64-bit operands bit-exactly. leaving it as is.

tt_metal/hw/inc/api/dataflow/cross_node_dfb.h
```diff
@@ -17,6 +17,7 @@
 
 #if defined(KERNEL_BUILD) && !defined(COMPILE_FOR_TRISC)
 #include "api/dataflow/noc.h"
+#include "noc_address_backend.h"
```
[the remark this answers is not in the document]
vvukomanovicTT: good catch on the quasar kernels - converted all eleven to noc_address_backend::extract_local_address. sharding_addrgen is already on the exemption list and the enablement pr turns it into a compile error under att. tensor_prefetcher is set_state-heavy runtime infra like the cq kernels, so it converts together with the dispatch/v3 migration - added it to the tracked exemptions rather than half-converting it here.

ttnn/cpp/ttnn/operations/experimental/quasar/tilize/device/kernels/dataflow/reader_unary_pad_multicore_both_dims_metal2.cpp
```diff
@@ -158,12 +158,10 @@ void kernel_main() {
                         const auto stage_lock = cb_in1.scoped_write_lock(1);
                         CoreLocalMem<uint32_t> temp_dst(temp_addr);
                         noc.async_read(
-                            UnicastEndpoint{},
+                            PrecomposedUnicastEndpoint{},
```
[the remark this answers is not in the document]
vvukomanovicTT: updated, the comment now describes the opaque forwarding.

tt_metal/hw/inc/api/dataflow/endpoints.h
```diff
@@ -16,6 +16,34 @@ struct UnicastEndpoint {
...
+
+template <>
+struct noc_traits_t<PrecomposedUnicastEndpoint> {
```
abhullar-tt, line 28: is there a way we could combine this with UnicastEndpoint? it seems like we won't have the x,y way of looking things up soon

vvukomanovicTT: I am not sure how to combine it other than delete it once we get rid of xy ?

abhullar-tt: I think end goal I would like `PrecomposedUnicastEndpoint` to just be named `UnicastEndpoint` but once the current version is deleted I think we could rename this?

abhullar-tt (approved):

tt_metal/hw/inc/api/dataflow/endpoints.h
vvukomanovicTT: I agree.

kstevensTT (approved):

llongTT (approved):

PR #54287, opened 2026-08-25
[Bug fix] ttnn.topk: validate the preallocated output indices width

### LINKED ISSUE #51959: ttnn.topk: preallocated output indices dtype is not validated against the required index width

ign-sudha:
### PR Category

- Bug fix

### Summary

- Closes #51959

A preallocated output indices tensor sets the index width for the whole op, and nothing checked that width against what the input needs. A `uint16` indices buffer on an input needing 32 bits wrapped past 65535 and returned indices for the wrong columns, with correct values and no error.

Only defect 2 from the issue is left. Defect 1, a `uint32` buffer on a narrow input, was already resolved before #54168.

### Solution

```cpp
const DataType required_dtype = required_index_dtype(input_tensor, args.dim);
const bool indices_too_narrow = output_tensor1_dtype == DataType::UINT16 && required_dtype != DataType::UINT16;
TT_FATAL(
    !indices_too_narrow,
    "Preallocated indices tensor must be {} for this input, got: {}",
    required_dtype,
    output_tensor1_dtype);
```

Only too-narrow is rejected. A buffer wider than needed is correct and stays legal, so this is deliberately not `prealloc == required`. It also catches `float32` input, which needs 32-bit indices at any width because fp32 sorts with fp32 dest accumulation. 

### Tests

`test_topk_preallocated_indices_width` covers the five combinations, using `65536` for the wide cases. The passing rows gather the input by the returned indices and compare against the returned values, so a wrong index fails and not just a missing error.

### Notes for reviewers

- Behaviour change: a `uint16` preallocated indices buffer with a wide or `float32` input now raises instead of returning wrong indices. No caller in the repo preallocates topk outputs outside the tests.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/sudha-51959-prealloc-output)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/sudha-51959-prealloc-output)

### DIFF
--- a/ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
+++ b/ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
@@ -280,6 +280,15 @@ void TopKDeviceOperation::validate_on_program_cache_miss(
                 output_tensor1_dtype == DataType::INT32,
             "Preallocated indices tensor must be UINT16, UINT32, or INT32 got: {}",
             output_tensor1_dtype);
+        // The preallocated indices tensor sets the index width for the whole op. A 16-bit one
+        // on an input that needs 32 bits wraps past 65535 and returns the wrong columns.
+        const DataType required_dtype = required_index_dtype(input_tensor, args.dim);
+        const bool indices_too_narrow = output_tensor1_dtype == DataType::UINT16 && required_dtype != DataType::UINT16;
+        TT_FATAL(
+            !indices_too_narrow,
+            "Preallocated indices tensor must be {} for this input, got: {}",
+            required_dtype,
+            output_tensor1_dtype);
         TT_FATAL(
             output_tensor0_dtype == input_tensor_dtype,
             "Preallocated output tensor dtype must match input tensor dtype. Got output: {}, input: {}",
--- a/ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
+++ b/ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
@@ -37,7 +37,7 @@ DataType resolve_index_dtype(
             return indices_dtype;
         }
     }
-    return required_index_dtype(tensor_args.input, args.dim);
+    return is_uint32_index_required(tensor_args.input, args.dim) ? DataType::UINT32 : DataType::UINT16;
 }
 
 // Maps the resolved index dtype onto the circular-buffer data format used by the sort datapath. 16-bit indices
@@ -282,12 +282,11 @@ void TopKDeviceOperation::validate_on_program_cache_miss(
             output_tensor1_dtype);
         // The preallocated indices tensor sets the index width for the whole op. A 16-bit one
         // on an input that needs 32 bits wraps past 65535 and returns the wrong columns.
-        const DataType required_dtype = required_index_dtype(input_tensor, args.dim);
-        const bool indices_too_narrow = output_tensor1_dtype == DataType::UINT16 && required_dtype != DataType::UINT16;
+        const bool indices_too_narrow =
+            output_tensor1_dtype == DataType::UINT16 && is_uint32_index_required(input_tensor, args.dim);
         TT_FATAL(
             !indices_too_narrow,
-            "Preallocated indices tensor must be {} for this input, got: {}",
-            required_dtype,
+            "Preallocated indices tensor must be 32-bit (UINT32 or INT32) for this input, got: {}",
             output_tensor1_dtype);
         TT_FATAL(
             output_tensor0_dtype == input_tensor_dtype,
--- a/tests/ttnn/unit_tests/operations/reduce/test_topk.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_topk.py
@@ -394,6 +394,48 @@ def test_topk_preallocated_dtype_raise(value_dtype, index_dtype, device, expect_
         ttnn.topk(ttnn_input, k=k, dim=-1, largest=True, sorted=True, output_tensor=(value_tensor, index_tensor))
 
 
+@pytest.mark.parametrize(
+    "W, input_dtype, index_dtype, raises",
+    [
+        (64, ttnn.bfloat16, ttnn.uint16, False),
+        (64, ttnn.bfloat16, ttnn.uint32, False),
+        (UINT16_MAX + 1, ttnn.bfloat16, ttnn.uint32, False),
+        (UINT16_MAX + 1, ttnn.bfloat16, ttnn.uint16, True),
+        (64, ttnn.float32, ttnn.uint16, True),
+    ],
+    ids=["w64_u16", "w64_u32", "w65536_u32", "w65536_u16_raises", "fp32_u16_raises"],
+)
+def test_topk_preallocated_indices_width(W, input_dtype, index_dtype, raises, device, expect_error):
+    # index_dtype is the preallocated output indices tensor. A 16-bit one on an input that needs
+    # 32 bits is rejected; wider than needed is legal. fp32 needs 32 bits at any width.
+    torch.manual_seed(0)
+    k = 32
+    shape = [1, 1, 32, W]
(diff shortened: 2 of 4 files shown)


ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
```diff
@@ -280,6 +280,15 @@ void TopKDeviceOperation::validate_on_program_cache_miss(
                 output_tensor1_dtype == DataType::INT32,
             "Preallocated indices tensor must be UINT16, UINT32, or INT32 got: {}",
             output_tensor1_dtype);
+        // The preallocated indices tensor sets the index width for the whole op. A 16-bit one
+        // on an input that needs 32 bits wraps past 65535 and returns the wrong columns.
+        const DataType required_dtype = required_index_dtype(input_tensor, args.dim);
+        const bool indices_too_narrow = output_tensor1_dtype == DataType::UINT16 && required_dtype != DataType::UINT16;
```
iwroszTT: Let's improve `required_index_dtype`'s  API.

We know that `required_index_dtype` returns either UINT16 or UINT32, hence the condition for `indices_too_narrow` is correct.

Note that `DataType` enumerates other data types too, which might lead to errors in the future if `required_index_dtype ` is changed.
I believe the information we need from `required_index_dtype` is `is_uint32_index_required`, which allows for a simple true/false return value.

iwroszTT (approved):

ttnn/cpp/ttnn/operations/reduction/topk/device/topk_device_operation.cpp
ign-sudha: Updated in the latest commit. Please let me know if that looks okay.

Issue #53907, opened 2026-08-21
On-device sampling: identical per-user seeds do not give identical draws (WH Galaxy)


viktorpusTT:
## Summary

On a Wormhole Galaxy (32 devices, `dispatch_core_axis: col`), `TTSampling` returns
a different token for different users when every user is given **identical
candidate logits, identical** `k`**/**`p`**/**`temp`**, and an identical seed**. All users
should draw the same token: the device seed derivation is slot independent, so
equal seeds at equal positions are equal streams.

Reproduced with no vLLM, no plugin and no model weights.

## Reproducer

tt-metal at `a85c79a9829b1121179c64accfb73c93655bacd0`, built with
`build_metal.sh`. Apply the test below (an addition to the existing
`models/demos/llama3_70b_galaxy/tests/unit_tests/test_sampling.py`) and run:

```bash
export TT_METAL_HOME=$PWD
export PYTHONPATH=$PWD
export HF_MODEL=meta-llama/Llama-3.3-70B-Instruct   # config only; dummy_weights=True reads no weights
export HF_HOME=<dir containing that model's config.json>
export MESH_DEVICE=TG
pytest models/demos/llama3_70b_galaxy/tests/unit_tests/test_sampling.py -k uniform -q
```

The test is a single addition to that existing file, inlined here so this
issue stays self-contained, and also on branch
`viktorpus/repro-uniform-seed-sampling` (branched from `a85c79a982`, not
`main`: it is a deliberately failing test).

<details>
<summary>Test addition (append to <code>models/demos/llama3_70b_galaxy/tests/unit_tests/test_sampling.py</code>)</summary>

```python
# ---------------------------------------------------------------------------
# Uniform-seed determinism: identical logits + identical seed must collapse to
# one sampled token for every user.
#
# On-device sampling derives its per-token device seed slot-independently, so
# equal seeds at equal positions are equal streams. Two seeding styles are
# covered because the model uses both: a scalar seed (all cores) and a per-user
# seed tensor (what SeedManager pushes each decode step).
# ---------------------------------------------------------------------------

UNIFORM_BATCH = 32

UNIFORM_SEED = 42


def _uniform_setup(mesh_device, dtype):
    model_args = TtModelArgs(mesh_device, max_batch_size=UNIFORM_BATCH, max_seq_len=32, dummy_weights=True)

    # One logits row broadcast to every user. Width is the full padded vocab;
    # the mesh mapper shards it across the mesh, as the model does.
[...]

Issue #53802, opened 2026-08-20
[models/common/sampling]: per-request seed salting (#53167) breaks vLLM same-seed reproducibility


viktorpusTT:
### Component / Area

models/common/sampling (vLLM integration)

### Issue Type (optional)

Bad Outputs

### Observed

#53167 (`240a5aa4ecedd294cec1b4cdc3a4f985a314f476`, merged 2026-08-19) added a per-slot salt to `_hash_request_seed_to_device_seed` in `models/common/sampling/generator.py`, so two concurrent slots carrying the **same** request seed now draw different device-seed streams and produce different text.

For the tt-transformers demo (#53077) that is the desired behavior. For the vLLM serving path, which shares this `SeedManager`, it breaks the seed contract in three ways:

**1. vLLM guarantees that the same seed reproduces, including concurrently in one batch.** Upstream V1 gives each seeded request its own `torch.Generator` and consumes it per batch row (`vllm/v1/sample/ops/topk_topp_sampler.py::random_sample`):

```python
for i, generator in generators.items():
    q[i].exponential_(generator=generator)
return probs.div_(q).argmax(dim=-1).view(-1)
```

Two independent requests with the same seed, prompt, and sampling params get the same `q` row and the same token. Batch position and concurrency are irrelevant. Post-#53167 the TT backend returns different text for those requests, so a client's standard reproducibility check passes on other backends and fails on TT.

**2. The n>1 justification in the new docstring does not apply to vLLM.** The docstring says the salt separates "n>1 completions of one prompt with a fixed seed". vLLM V1 never delivers that case to the backend as duplicate seeds. `vllm/v1/engine/parallel_sampling.py::_get_child_sampling_params` fans `n>1` out into n child requests and gives each a distinct seed:

```python
else:
    # Each child gets a clone with a unique seed
    child_sampling_params.seed = seed + index
```

So `n=16, seed=1234` arrives as sixteen requests with seeds `1234..1249`, all unique, all salt 0, all unchanged by #53167. vLLM already solves this at the layer where the "these are n samples of one prompt" intent exists. Under vLLM the only way duplicate seeds reach one batch is a client deliberately sending the same seed twice, which is exactly the case where identical output is the promised behavior.

[...]

Issue #53077, opened 2026-08-13
[tt-transformers]: almost identical answers across generations with large temperature


ichovpanTT:
### Component / Area

tt-transformers

### Issue Type (optional)

None

### Observed

Generating 16 completions for the same GSM8K question and temperature=1.5 (that is pretty high!) produces identical answers for all completions except the first one.

Model: `meta-llama/llama-3.2-1b-instruct`

System prompt:
```
You are a careful math tutor. Solve the problem step by step, then give the final numeric answer on its own line prefixed with '####'.
```

Question:
```
Natalia sold clips to 48 of her friends in April, and then she sold half as many clips in May. How many clips did Natalia sell altogether in April and May?
```

Answers:
```
--- sample 00 ---
Number of slices sold in April = 48 
Number of slices sold in May/2 = 48/2
               = 24

Total slices = 48 + 24 
          = 72  ####

--- sample 01 ---
To find the total number of clips Natalia sold in April and May, we need to add the number of clips she sold in April and May.

In April, Natalia sold 48 clips.
In May, she sold half as many clips as in April, which is 48 / 2 = 24 clips.

So, the total number of clips Natalia sold in April and May is 48 + 24 = 72 clips.

--- sample 02 ---
To find the total number of clips Natalia sold in April and May, we need to add the number of clips she sold in April and May.

In April, Natalia sold 48 clips.
In May, she sold half as many clips as in April, which is 48 / 2 = 24 clips.

So, the total number of clips Natalia sold in April and May is 48 + 24 = 72 clips.

--- sample 03 ---
To find the total number of clips Natalia sold in April and May, we need to add the number of clips she sold in April and May.

In April, Natalia sold 48 clips.
In May, she sold half as many clips as in April, which is 48 / 2 = 24 clips.

So, the total number of clips Natalia sold in April and May is 48 + 24 = 72 clips.

--- sample 04 ---
To find the total number of clips Natalia sold in April and May, we need to add the number of clips she sold in April and May.

In April, Natalia sold 48 clips.
In May, she sold half as many clips as in April, which is 48 / 2 = 24 clips.

So, the total number of clips Natalia sold in April and May is 48 + 24 = 72 clips.

--- sample 05 ---
To find the total number of clips Natalia sold in April and May, we need to add the number of clips she sold in April and May.

In April, Natalia sold 48 clips.
In May, she sold half as many clips as in April, which is 48 / 2 = 24 clips.

[...]

Issue #50649, opened 2026-07-21
Qwen3-32B: intermittent corrupted output at top_p=1.0 with on-device sampling


ksriv001:
# Qwen3-32B: intermittent corrupted output at `top_p=1.0` with on-device sampling

## Summary

`Qwen/Qwen3-32B` intermittently produces corrupted output when a request runs at **`top_p=1.0`** with on-device sampling (`sample_on_device_mode=all`). We see it two ways:

- garbled/junk tokens in the generated text, and
- the model emitting **EOS right after a think block closes** — in a tools + thinking request this shows up as a **dropped tool call** (`finish_reason: stop`, empty content, no `<tool_call>`).

The one clear signal we have is that it's a function of `top_p`: at `top_p=1.0` it happens 17–35% of the time, and it **disappears once `top_p` is clamped** (`top_p ≤ 0.95` → 0 failures across our runs). The identical model on another host (Nebius via OpenRouter) shows 0 failures at `top_p=1.0`, so this doesn't look like ordinary sampling variance.

We haven't dug into the device sampling path ourselves — filing this so someone who knows that code can take a look. Repro steps and the data we have are below.

Related: [tt-inference-server#4386](https://github.com/tenstorrent/tt-inference-server/issues/4386) (Issue 4) and [tt-cloud-console#944](https://github.com/tenstorrent/tt-cloud-console/pull/944).

## Environment

- Model: `Qwen/Qwen3-32B`
- On-device sampling: `sample_on_device_mode=all`
- Flags: `--enable-auto-tool-choice --tool-call-parser hermes --reasoning-parser qwen3`
- Transport: streaming SSE
- Reproduced with **tt-metal = stable**, tt-inference-server = `lzhang/openrouter-issue-4-stable` (tests on top of stable), vLLM = dev.

## Where this shows up (tt-shield)

The drop reproduces under release-dispatch shield runs on **tt-metal=stable** (tt-inference-server = `lzhang/openrouter-issue-4-stable`, which carries the tests), and it **persists across the vLLM streaming-parser fix** — which isolates it from the reasoning/tool streaming parser:

- vLLM=dev (no parser fix): issues 1, 2, and 4 all appear — [run 29852446182](https://github.com/tenstorrent/tt-shield/actions/runs/29852446182)
[...]

PR #54292, opened 2026-08-25
[Bug fix]  Fix for Qwen3-32B and Llama-3.3-70B sampling test failures on Galaxy

ifabijanic-tt:
## Summary

Fixes cross-request contamination in the prefill→decode handoff on `llama3_70b_galaxy` code which is shared by Llama-3.3-70B and Qwen3-32B. It also completes a seed-salting fix that never reached Qwen.

Validated with 15 full sweeps per model of the `vllm-tt-plugin` `tests/tt`
suite on WH Galaxy (TG), DP=4, `max_num_seqs 8`, `sample_on_device_mode: all`,
tracing and async scheduling enabled:

| Model | Sweeps | Test executions | Failures |
|---|---|---|---|
| Llama-3.3-70B | 15 | 1095 | 0 |
| Qwen3-32B | 15 | 1095 | 0 |

## Fixed issues

**1. A partial prefill overwrote sampling state of slots that were mid-decode.**

`_scatter_params_to_slots` broadcast the prefilling request's values across the whole `max_batch` vector, so slots holding live decoding requests sampled under someone else's parameters until the next decode re-uploaded the authoritative vector. Fixed by shadowing the per-slot values last written to device (`_slot_sampling_params`) and scattering onto that instead of a broadcast filler.

**2. `reset_prompt_tokens` cleared the prompt mask of decoding rows.**

`sampling_prompt_tokens` is `-1` everywhere except the slots being filled, so an unscoped reset wiped the penalty prompt mask for every other row. Both `SamplingGenerator.reset_prompt_tokens` and `TTPenalties.reset_prompt_tokens` now take an explicit `slots` list and merge against a host shadow.

**3. Prefill sampling allocated large buffers behind live traces.**

Counting the prefill token on device ran `scatter_add` + `tilize`, allocating two `[max_batch, padded_vocab]` int32 buffers (~32 MB/device) behind the live prefill and decode traces on every prefill which corrupted the slot already decoding. Prefill sampling now passes `count_tokens=False` and the equivalent bookkeeping is folded in from host via `copy_host_to_device_tensor` which allocates nothing.

**4. Prefill wrote through KV blocks it did not own.**

`prefill_forward_text` now passes `real_seq_lens` (unpadded, prefix-cache aware) so a request only writes the blocks it actually owns.

**5. `salt_duplicate_seeds` never reached Qwen3-32B.**

`#53077` salts concurrent duplicate seeds apart to separate `n>1` completions of one prompt. Under vLLM v1 that is unnecessary — `ParentRequest._get_child_sampling_params` already hands child *i* `seed + i` — so every duplicate seed that arrives is independent requests that must reproduce identically. Salting made `test_uniform_seed_deterministic` return 32 *different* outputs for 32 same-seed requests.

The fix existed but only for Llama: it was set inside
`TtModelArgs._set_params_from_dict`, and `TtQwenModelArgs` fully overrides that
method **without calling `super()`**, so Qwen silently kept the `True` default.
Moved to a **class attribute** on `TtModelArgs` so no subclass can drop it.

Verified on device: 32 concurrent same-seed requests now return 1 unique output
(was 32); 16 distinct seeds still return 16 unique.

## Small test harness fix

`test_qwen_attention_ttt.py` could not run, and could not have produced a
trustworthy result if it had:

- HF reference input hardcoded `.to(torch.bfloat16)` against fp32 weights → `RuntimeError` before any PCC was computed.
- Custom reference input not cast to its own parameter dtype → same.
- **`comp_pcc(reference_output, reference_output_custom, pcc)` was overwritten by the very next line**, so a reference-vs-reference mismatch was invisible and got reported as a TT failure.

## Open questions for review

- **`model_config.py` carries a Blackhole revert.** The branch as originally committed deleted `PREFETCHER_NOC1_GRID_BH`, `LM_HEAD_32_GRID_BH` and `_get_core_ranges_blackhole`, which `qwen_model_config.py` still imports, so it could not boot a server at all. I restored them from `main` to unblock testing, but **this wants a proper rebase**, not a restore.
- **Should the harness fix ship here** or as its own PR? It is a test-only change
  in a different area.
- **`_no_persist_gather` is still env-gated scaffolding.** Either promote it to a
  real fix or drop it before merge.

## Notes for reviewers

Two notes on scope before you post it. The 15/15 Qwen green result **depends on the companion plugin-side test changes**, without those, `TestPresencePenalty` still fails for the precision reason, which is not a tt-metal defect. You may want to land the plugin PR first, or state the dependency explicitly. And the Blackhole revert is the one thing I'd resolve before review rather than during it.

Issues that this PR fixes:
https://github.com/tenstorrent/tt-metal/issues/53907
https://github.com/tenstorrent/tt-metal/issues/53802
https://github.com/tenstorrent/tt-metal/issues/52898
https://github.com/tenstorrent/tt-metal/issues/50649
https://github.com/tenstorrent/tt-metal/issues/49785

vLLM sampling tests CI run:
https://github.com/tenstorrent/tt-metal/actions/runs/32842480779

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ctr-ifabijanic/qwen_llama_fixes)

### DIFF
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -11,8 +11,6 @@ from typing import List, Optional
 import torch
 from loguru import logger
 
-import os
-
 import ttnn
 
 from ._utils import clamp, is_default_value, split_list
@@ -83,23 +81,6 @@ class SamplingParams:
 
 
 SAMPLING_PARAM_FIELDS = tuple(f.name for f in fields(SamplingParams))
-
-
-def uniform_from_device_seed(device_seed):
-    """Map one SeedManager device seed to a uniform in [0, 1) for the Python sampler.
-
-    Reuses the existing derivation rather than inventing a seeding scheme:
-    _hash_request_seed_to_device_seed(request_seed, counter, salt) already produces a
-    slot-independent, salt-separated per-token value, and align_seed_counters_to_positions
-    keeps the counter tied to the absolute decode position. Finishing the mapping on host
-    means the draw does not depend on per-core PRNG state, manual_seed ordering, or any SFPU
-    work between seeding and the draw -- the hazard 4b2f6978c9c had to work around.
-    """
-    x = (int(device_seed) * 0x2545F4914F6CDD1D) & _UINT64_MASK
-    x ^= x >> 33
-    x = (x * 0xFF51AFD7ED558CCD) & _UINT64_MASK
-    x ^= x >> 33
-    return (x >> 11) / float(1 << 53)
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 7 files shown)


mtairum (approved):

mtairum (approved):

mtairum:
### Code review

Found 1 issue:

1. `models/common/sampling/tt_sampling.py` silently reverts the Blackhole-Galaxy force-argmax fix from #51534 (commit d3d999c25c0). The PR deletes `_force_argmax_sub_core_grids`, the `force_argmax_skip_vocab_trim` gating, the `subdevice_id=` pinning on `all_gather_async`, and every `sub_core_grids=` argument on `untilize`/`concat`/`argmax`. The `use_unfused_ccl` flag that code keyed off is still live elsewhere in the codebase, so on the BH prefetcher path this re-exposes the fatal "kernel group cores do not match sub device cores" error and the stale-token trace race that #51534 fixed (the removed comments name Qwen3-32B's 155648-wide padded vocab specifically). The PR body discloses the parallel Blackhole revert in `model_config.py` (restored at head), but not this one, which is still absent at head — it looks like the same rebase artifact and probably wants the same restore/rebase treatment.

What the PR deletes (as added by #51534):

https://github.com/tenstorrent/tt-metal/blob/d3d999c25c0cc8070cc75fa3f392c3205b5680cf/models/common/sampling/tt_sampling.py#L222-L224

https://github.com/tenstorrent/tt-metal/blob/d3d999c25c0cc8070cc75fa3f392c3205b5680cf/models/common/sampling/tt_sampling.py#L789-L892

The same block at this PR's head, with all sub-device pinning gone:

https://github.com/tenstorrent/tt-metal/blob/af8539ebc4e8fcc93551b3daca8e37a350551af7/models/common/sampling/tt_sampling.py#L805-L865

ifabijanic-tt:
Review is correct, fixed the issue.

PR #54302, opened 2026-08-25
tt_transformers: declare scheduler-driven chunked prefill for llama, qwen and mistral

viktorpusTT:
Enables vLLM scheduler-driven chunked prefill for the `tt_transformers`
full-attention text bridges.

Depends on [#54435](https://github.com/tenstorrent/tt-metal/pull/54435) (merged), which
owns the resume-offset alignment and the resumed-prefill trace warmup this needs.
That has merged (`0acbde02e3b`); this branch is rebased onto current `main`
(`ee7c625d0a2`).

Pairs with
[tenstorrent/vllm-tt-plugin#79](https://github.com/tenstorrent/vllm-tt-plugin/pull/79),
which consumes the capability this declares; the two must land together (this one first, plugin second).

Implements the tt-metal half of
[vllm-tt-plugin#74](https://github.com/tenstorrent/vllm-tt-plugin/issues/74).

## What this enables

`LlamaForCausalLM`, `QwenForCausalLM` (Qwen2 and Qwen3) and `MistralForCausalLM`
declare `supports_chunked_prefill`, so one prompt may be prefilled across several
engine steps instead of occupying a whole step. Gemma 4 moves onto the same
capability so the plugin can delete its hardcoded `model_type` allowlist.

Not to be confused with generator-internal chunking (`max_prefill_chunk_size`),
which splits a long prompt inside a single `prefill_forward_*` call. This is the
scheduler splitting one prompt across engine steps, so the generator has to
accept a nonzero `start_pos`. Both enter the same branch in
`prefill_forward_single_user_text`, which is why these bridges need no new
prefill code. Open tt-metal PRs that mention "chunked prefill" for Kimi, GLM,
DeepSeek or Minimax are that generator-internal path, not this.

## Why the declaration is small

The hard part is not the declaration, it is that a resume offset must be a
multiple of both the paged KV `block_size` and the `q_chunk_size` the model's
chunked-SDPA program was built with, and the op reads the wrong prefix rather
than raising when it is not. The vLLM scheduler satisfies neither: `num_new_tokens`
comes straight out of the remaining token budget, so any step where several
prefills share the budget produces arbitrary values.

That correction is not specific to chunked prefill. Automatic prefix caching
already produced offsets the traced SDPA could not honour, so it is a live
wrong-answer bug on `main` today, independent of anything here. It is therefore
[#54435](https://github.com/tenstorrent/tt-metal/pull/54435)'s subject, together
with the resumed-prefill trace warmup that keeps the traces those replays need
from being allocated at serving time. This PR carries only what is genuinely new:

- `models/tt_transformers/tt/generator_vllm.py` - declare
  `supports_chunked_prefill` on the three text bridges. The Galaxy bridges are
  separate classes and are untouched.
- `models/demos/gemma4/tt/generator_vllm.py` - the same key on
  `Gemma4ForCausalLM.model_capabilities`, which is the class the plugin resolves.
  `Gemma4Generator` in `models/demos/gemma4/tt/generator.py` is the demo path, has
  no vLLM scheduler and is left alone. Both classes already declare
  `resumed_prefill_token_alignment: SDPA_CHUNK_ALIGN` for #54435, and both
  override `warmup_model_prefill` without calling `super()`, so declaring this
  does not reach the resumed warmup sweep.
- `models/common/model_capabilities.py`: document the `model_capabilities`
  contract the plugin reads. An absent key means False, a subclass dict
  replaces rather than merges, and the plugin snapshots the class attribute
  during `check_and_update_config` before any generator instance exists.
  Instance rebinding (Gemma 4 does this for `supports_async_decode` and
  `supports_sample_on_device`) does not reach the scheduler. The key list
  includes `output_tokens_per_step` and `accepts_trace_mode`.
- `models/demos/gemma4/tt/generator_vllm.py`: comments on the instance-level
  capability merges now match that snapshot rule. PLI still needs
  `GEMMA4_SUPPORTS_ASYNC_DECODE=0` to disable `async_scheduling`.
- CI: a T3K Llama leg that serves with a token budget far below the prompt
  length. It keeps `--async-scheduling` (the runner's async path is decode-only,
  so it does not touch chunked prefill) and sets `sample_on_device_mode: "all"`,
  which is the only way the runner's host-sampling fallback for mid-prompt rows
  is reached. The step first checks the resolved budget in the server log,
  because the device tests cannot confirm the feature is on for themselves.
  A missing `max_num_batched_tokens=` field now reaches the comparison and
  prints both values instead of exiting under `pipefail` with no message.
  Default `chunked-prefill-test-timeout` is 15 minutes.

There is no per-model alignment declaration for `tt_transformers`. #54435 derives
the `q_chunk_size` from the model's own `get_attn_sdpa_program_config` per padded
suffix length, so a short resumed span keeps its 64-token alignment instead of
being rounded to 256. An earlier revision of this PR declared a flat 256 in
`model_capabilities` and had the plugin round the token budget to it; both are
gone, because the value is not a constant and the plugin has no use for it.

## Validation

T3K, Llama-3.1-8B-Instruct, `--max-num-batched-tokens 2048 --max_model_len 32768
--async-scheduling`, `sample_on_device_mode: "all"`.

The discriminating measurement is four ~2K-token prompts issued concurrently,
each carrying its own passphrase and asked to repeat it. Twenty samples:

| resume offsets floored to | recall |
|---|---|
| chunked prefill off | 20/20 |
| `block_size` (64) only | 10/20 |
| `lcm(block_size, q_chunk_size)` | 20/20 |

Solo prefills pass in every configuration, so this only shows up once several
prefills share a step. Removing the floor entirely does not degrade output, it
kills the engine: `TT_FATAL @ tensor_apis.cpp:157 host_tensor.logical_shape() ==
device_tensor.logical_shape()`.

Also on device (pre-rebase head of this branch; the four commits here are
unchanged):

- Plugin host suite: 343 passed locally. After the rebase onto current plugin
  `main`, GitHub `unit-tests` (3.10 and 3.12, `pytest tests --ignore=tests/tt`)
  passed on [vllm-tt-plugin#79](https://github.com/tenstorrent/vllm-tt-plugin/pull/79).
- Plugin `tests/tt/test_chunked_prefill.py`: three cases (solo budget-sized
  split, 3x-budget span that pins `q_chunk_size=256`, concurrent ragged
  remainder). The T3K chunked-prefill CI leg runs all three.
- `tests/tt` with the T3K leg's `-k` filter: 42 passed, 1 skipped. Unfiltered,
  11 failed with chunked prefill on against 12 with it off, same families
  (`TestSeedingAndVariety`, `*_mixed_batch`) that
  [vllm-tt-plugin#77](https://github.com/tenstorrent/vllm-tt-plugin/pull/77)
  rewrites and the T3K leg already excludes.
- A 4200-word prompt whose middle chunk is a full 2048-token resumed span: 6/6.
- `vllm bench serve`, 8 x 4096-token prompts at concurrency 4: 8/8, mean TTFT
  1393 ms, mean ITL 35 ms.

Full `vLLM Model Tests`: https://github.com/tenstorrent/tt-metal/actions/runs/33627798946

- `Llama-3.1-8B-Instruct (chunked prefill) [wh_llmbox]` passed, including the
  server-log budget gate and the three device tests.
- Gemma 4 E2B, the other Llama 8B llmbox legs, and Qwen2.5-VL-72B passed.
- `Qwen3-VL-32B-Instruct` structured-output `correct_rate` 0% is not this PR.
  Chunked prefill stayed off (`Qwen3VLForConditionalGeneration` does not declare
  the capability). The same job already failed at 71.875% on the 31 Aug nightly
  (plugin `main`). Grammar was applied; decode quality under sampling is flaky.
- `Qwen3.6-27B (multimodal B=1)` server-ready timeout is not extra warmup.
  Chunked prefill stayed off. Nightly became ready at 1560 s / 1800 s; this host
  spent ~29 min loading 64 transformer layers and hit the 30 min cap three
  seconds into the same decode warmup nightly always runs. The sibling
  `batch=8` leg (`server-timeout: 35`) passed.
- Remaining reds (GPT-OSS, Galaxy sampling, Gemma4-31B verify) match existing
  matrix noise; Galaxy generators do not declare this capability.

The unit coverage for resume-offset alignment lives in
`models/common/tests/test_resume_offset_alignment.py`, merged with #54435.

## Known limitations

**Gemma 4 above `max_num_seqs 1` is unsound for a separate reason.** Its
sliding-window layers keep a rolling K/V tail in a single per-layer attribute
that is not keyed by request and is reset whenever any request starts at offset
0, so concurrent prefills clobber each other. This PR neither causes nor fixes
that; Gemma 4 is enabled here exactly as it was under the allowlist. Filed
separately.

**Qwen and Mistral have no device validation.** They share the tt_transformers
generator and SDPA configuration with Llama, so the Llama result is strong
evidence, but neither model's weights were available on the box used here.

**Prefix-cache hits below the alignment are recomputed.** That is #54435's
behaviour, not a change here. The Llama prefix-caching CI leg passes.

## Not duplicating

`gh pr list --repo tenstorrent/tt-metal --state open --search "chunked prefill"`
hits generator-internal prefill work (Kimi, GLM, DeepSeek, Minimax). None of
those declare `supports_chunked_prefill` on the vLLM `tt_transformers` bridges
or add the T3K Llama scheduler-split CI leg.

## AI assistance

Written with AI assistance (Claude, then Cursor Grok 4.6). Every line is to be
reviewed and defended by the human submitter.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:viktorpus/chunked-prefill-tt-transformers)

### DIFF
--- a/.github/time_budget.yaml
+++ b/.github/time_budget.yaml
@@ -425,7 +425,7 @@ models:
   vllm:
     wh_galaxy_perf: 430
     bh_quietbox_2: 310
-    wh_llmbox: 297
+    wh_llmbox: 342
     bh_loudbox: 75
     wh_n150: 65
 umd:
--- a/.github/workflows/vllm-model-tests-impl.yaml
+++ b/.github/workflows/vllm-model-tests-impl.yaml
@@ -659,7 +659,7 @@ jobs:
               exit 1
             fi
             RESOLVED=$(grep "$EXPECTED" "${FILE_SERVER_LOG}" | tail -1 \
-              | grep -o "max_num_batched_tokens=[0-9]\+" | cut -d= -f2)
+              | grep -o "max_num_batched_tokens=[0-9]\+" | cut -d= -f2 || true)
             if [ "$RESOLVED" != "$BUDGET" ]; then
               echo "❌ server resolved max_num_batched_tokens=$RESOLVED, tests were told $BUDGET"
               exit 1
--- a/.github/workflows/vllm-model-tests-impl.yaml
+++ b/.github/workflows/vllm-model-tests-impl.yaml
@@ -139,6 +139,7 @@ jobs:
       FILE_BENCHMARK_STRUCTURED_OUTPUT_LOG: "./output/vllm_benchmark_structured_output.log"
       FILE_MULTIMODAL_TEST_LOG: "./output/vllm_multimodal_test.log"
       FILE_SAMPLING_TESTS_LOG: "./output/vllm_sampling_tests.log"
+      FILE_CHUNKED_PREFILL_TESTS_LOG: "./output/vllm_chunked_prefill_tests.log"
     container:
       image: ${{ inputs.docker-image }}
       env:
@@ -162,6 +163,7 @@ jobs:
         FILE_BENCHMARK_STRUCTURED_OUTPUT_LOG: "./output/vllm_benchmark_structured_output.log"
         FILE_MULTIMODAL_TEST_LOG: "./output/vllm_multimodal_test.log"
         FILE_SAMPLING_TESTS_LOG: "./output/vllm_sampling_tests.log"
+        FILE_CHUNKED_PREFILL_TESTS_LOG: "./output/vllm_chunked_prefill_tests.log"
         FILE_COHERENCE_TEST_LOG: "./output/vllm_coherence_test.log"
       volumes:
         - ${{ github.workspace }}/docker-job:/work # Subdir to workaround https://github.com/actions/runner/issues/691
@@ -638,6 +640,36 @@ jobs:
             else
               pytest "${PLUGIN_DIR}/tests/tt" -v --tt-server-url=http://localhost:8000 --tt-model-name=${{ matrix.test-group.model }}
             fi
+
+      - name: 🧪 Run chunked prefill tests
+        if: ${{ matrix.test-group.chunked-prefill-budget }}
+        timeout-minutes: ${{ matrix.test-group.chunked-prefill-test-timeout || 15 }}
+        uses: ./docker-job/.github/actions/run-with-log
+        with:
+          log-file: ${{ env.FILE_CHUNKED_PREFILL_TESTS_LOG }}
+          run: |
+            # Confirm the server really has the feature on at this budget before
+            # trusting the tests. Without this they would pass against a server that
+            # silently disabled chunked prefill, e.g. one whose model class does
+            # not declare the capability.
+            BUDGET=${{ matrix.test-group.chunked-prefill-budget }}
+            EXPECTED="Chunked prefill enabled"
+            if ! grep -q "$EXPECTED" "${FILE_SERVER_LOG}"; then
+              echo "❌ server log has no '$EXPECTED' line, so chunked prefill is not active"
+              exit 1
+            fi
+            RESOLVED=$(grep "$EXPECTED" "${FILE_SERVER_LOG}" | tail -1 \
+              | grep -o "max_num_batched_tokens=[0-9]\+" | cut -d= -f2)
+            if [ "$RESOLVED" != "$BUDGET" ]; then
+              echo "❌ server resolved max_num_batched_tokens=$RESOLVED, tests were told $BUDGET"
+              exit 1
+            fi
+            echo "✅ chunked prefill active at max_num_batched_tokens=$RESOLVED"
+
+            pytest "${PLUGIN_DIR}/tests/tt/test_chunked_prefill.py" -v \
+              --tt-server-url=http://localhost:8000 \
+              --tt-model-name=${{ matrix.test-group.model }} \
+              --tt-chunked-prefill-budget="$BUDGET"
 
       - name: 🧹 Cleanup server process
         if: always()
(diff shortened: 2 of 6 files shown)


tdowdallTT (approved):

mtairum (approved):
LGTM.

Thanks for running CI and validating!

gwangTT (approved):

PR #54306, opened 2026-08-25
[Models CI] Migrate the Galaxy demo tests to tier 1 Models e2e and retire the pipeline

mtairum:
**CI runs** — verification on `3efec5fa128`, (Tier 1) Models E2E Tests:

| Model | SKU | Run |
|---|---|---|
| `wan-2.2-t2v` | `wh_galaxy_perf` | [32970009429](https://github.com/tenstorrent/tt-metal/actions/runs/32970009429) |
| `wan-2.2-t2v` | `wh_galaxy_perf` | [32970022706](https://github.com/tenstorrent/tt-metal/actions/runs/32970022706) |
| `wan-2.2-t2v` | `wh_galaxy_perf` | [32970036171](https://github.com/tenstorrent/tt-metal/actions/runs/32970036171) |
| `wan-2.2-t2v` | `bh_quietbox_2` | [32970048968](https://github.com/tenstorrent/tt-metal/actions/runs/32970048968) |

The Galaxy leg runs three times in parallel on purpose. That fleet produced four distinct hardware faults during this PR's history — device enumeration, ETH heartbeat, fabric router sync, TLB window allocation — across five different machines, so a single red run cannot be read as a verdict on the diff. Three samples on three machines are what made the previous round's failure provably deterministic rather than infrastructural.

`deepseek-v3-tg` was verified separately in [32962863330](https://github.com/tenstorrent/tt-metal/actions/runs/32962863330): the renamed dispatch id resolved and spawned the right job, then the run died on `Query mappings failed on device 17` on `OM1-01A02-STGWH02` — the same machine and same fault as 11 hours earlier. The entry itself passed unchanged on `g03glx03` in [32860391817](https://github.com/tenstorrent/tt-metal/actions/runs/32860391817).

<details><summary>Run history and what each round caught</summary>

| Run | Target | Outcome |
|---|---|---|
| [32858130258](https://github.com/tenstorrent/tt-metal/actions/runs/32858130258) | Wan Galaxy | `OfflineModeIsEnabled` — real defect, fixed in 45195b723f0 |
| [32858148176](https://github.com/tenstorrent/tt-metal/actions/runs/32858148176) | DeepSeek Galaxy | fleet — device 17, `OM1-01A02-STGWH02` |
| [32859530553](https://github.com/tenstorrent/tt-metal/actions/runs/32859530553) | Wan Galaxy | ✅ pass, pre-merge entry |
[...]

PR #52553, opened 2026-08-07
[tt-train] Migrate tt-train tests from galaxy-perf to e2e-models-t3

kevinwuTT:
### Summary
Edited to include changes needed to work with `bh_sc1`:

Move tt-train model runs from the old galaxy-perf.yaml workflow to e2e-models-t3 workflow. A few newer models are also added. Each of these should run on a Galaxy BH node from the exabox runner. 

From old workflow:
* tinyllama_1_1b_ddp32_shakespeare
* llama_8b_tp8_ddp4_shakespeare (disabled for this PR since it seems to be hanging on exabox runners).
* llama_8b_tp4_ddp8_shakespeare (previously disabled, but keeping config).

New:
* qwen3_1_7b_ddp32_shakespeare
* tiny_deepseek_v3_ddp32_char_shakespeare
* llama_70b_tp8_fsdp4_shakespeare (disabled for this PR since it's unstable). 
  * [Passing run](https://github.com/tenstorrent/tt-metal/actions/runs/32389575751/job/96494762058)
  * [Timed out run](https://github.com/tenstorrent/tt-metal/actions/runs/32408202527/job/96554549010) (hanged after step 240)

***
`mpirun` is needed when running commands on exabox. This spawns a separate environment and requires some additional work when handling generated files and Github specific processes.
  * Metrics handling is disabled. The generated files need to be copied to NFS and then processed from the Github environment. This seems more appropriate as a separate PR.
  * Added a skip for setting `GITHUB_STEP_SUMMARY` if the underlying file doesn't exist. `GITHUB_STEP_SUMMARY` points to an actual file, but this does not exist in the `mpirun` environment. The solution should also be to use the NFS as a bridge to the parent Github environment. This also seems more appropriate as a separate PR.
 
### Notes for reviewers
To keep the run time reasonable, the maximum number steps are reduced. The target number is determined by visually inspecting the loss plots from previous runs and choosing a step where the loss begins to plateau. The following table shows the updated numbers from the most recent successful run: https://github.com/tenstorrent/tt-metal/actions/runs/32389575751/job/96494762054

| Test | Old Step Count | New Step Count | Elapsed time (MM:SS) | Budget (min) |
|------|------|------|------|------|
| tinyllama_1_1b_ddp32_shakespeare | 2048 | 1000 | 41:56 | 45 |
| llama_8b_tp8_ddp4_shakespeare × | 2048 | 750 | N/A | N/A |
[...]

PR #54324, opened 2026-08-25
[Models CI] Migrate the Galaxy perf tests to tier 1 e2e and device perf, and retire the pipeline

mtairum:
> [!IMPORTANT]
> **Stacked on #54306 — do not open for review or merge until that PR lands.**
> The base branch is `mtairum/migrate-galaxy-demo-pipeline`, so the diff shown here is only this PR's own change. GitHub retargets it to `main` automatically once #54306 merges; at that point this can come out of draft.
>
> **#53553 has landed** (`e5bd6c8c2f4`), so the DeepSeek decode perf thresholds are refreshed and the migrated device-perf entry is expected to pass. Its op-to-op expectation is now 139.41 µs with a dedicated 4% margin.

**CI run** — (Tier 1) Models E2E Tests on `28606b663aa`:

| Model | SKU | Run |
|---|---|---|
| `wan-2.2-t2v` | `bh_quietbox_2` | [32971450466](https://github.com/tenstorrent/tt-metal/actions/runs/32971450466) |

First run to execute the full three-call Wan entry on a QuietBox, which is what this PR's added call needs to be checked against. Expected per call: 480p `2x2_sp0tp1` passes, 720p `4x8sp1tp0nl4_ring_is_fsdp1` skips, and this PR's 720p `wh_4x8_sp1tp0` perf call skips — both skips on exact device match, 32 ≠ 4, so the job stays green.

The Galaxy side of that entry is verified on the base PR: [32970009429](https://github.com/tenstorrent/tt-metal/actions/runs/32970009429), [32970022706](https://github.com/tenstorrent/tt-metal/actions/runs/32970022706), [32970036171](https://github.com/tenstorrent/tt-metal/actions/runs/32970036171). The device-perf entry cannot be run until #53553 lands.

---

### Ticket

N/A — CI pipeline migration, continuation of #53911, #54184 and #54306.

### Problem description

`(Galaxy) perf tests` held two entries, both `team: models`, both on `wh_galaxy_perf`. It used to hold two tt-train jobs as well, but #52553 moved those to models e2e tier 3, so nothing else is left in the pipeline:

| Entry | Runs | Gate |
|---|---|---|
| Galaxy DiT Wan2.2 model perf tests | `test_performance_wan.py -k "wh_4x8_sp1tp0 and resolution_720p and t2v"` | host wall-clock per stage vs `t2v_metrics()` — encoder 0.2 s, denoising 370 s, vae 7 s, total 375.2 s |
| Galaxy DeepSeek V3 decode perf tests | `test_decode_demo_perf.py` | tracy device profiler: kernel duration, op-to-op latency and E2E, each ±3% |

Neither needs a standalone Galaxy pipeline. Wan2.2 moves to tier 1 e2e; DeepSeek is removed for the reason below; the pipeline is deleted.

### Wan2.2 perf joins the entry that already runs this file

`models_e2e_tests.yaml` already has a tier-1 `TT-DiT Wan2.2-T2V-A14B e2e tests` entry running `test_performance_wan.py` on `bh_quietbox_2` and `wh_galaxy_perf` (the latter added in #54306). The migrated case becomes a third call on that entry rather than a fourth Wan entry in the registry:

```
pytest --timeout 1200 models/tt_dit/tests/models/wan2_2/test_performance_wan.py \
  -k "wh_4x8_sp1tp0 and resolution_720p and t2v" || fail=1
```

Nothing new is needed around it: the entry already `unset HF_HUB_OFFLINE` and exports `TT_DIT_CACHE_DIR`, and `HF_HUB_CACHE` comes from the e2e impl's container env — the same value the old entry exported by hand. `wh_galaxy_perf` goes 45 → 70 min to cover the extra call.

**The perf data keeps flowing.** I had this wrong earlier and want it on the record: the old entry sets `save-perf-data: true`, and I assumed the tiered e2e impl ignoring that key meant the upload would be silently lost. It is not — `models-e2e-tests-impl.yaml` uploads whenever `generated/benchmark_data` exists, with no key to gate it, and the test writes there through `BenchmarkData.save_partial_run_json`. What does not travel is the separately named `perf-report-csv-*` artifact, which the galaxy impl produces on its *other* branch, for entries where `save-perf-data` is false.

### DeepSeek V3 decode perf moves to tier 1 device perf

This is a tracy device-profiler test carrying `@pytest.mark.models_device_performance_bare_metal`, so `models_device_perf_tests.yaml` is the right destination rather than e2e — that pipeline builds with `tracy: true`, and `wh_galaxy_perf` was already in its `ALL_SKUS`, needing only a budget key.

```yaml
- name: DeepSeek-V3 decode device perf tests (Galaxy)
  cmd: |
    export MESH_DEVICE=TG
    uv pip install -r models/demos/deepseek_v3/reference/deepseek/requirements.txt
    pytest --timeout=1200 models/demos/deepseek_v3/demo/test_decode_demo_perf.py
  model: deepseek-v3-tg
  skus:
    wh_galaxy_perf:
      timeout: 20
```

Transcribed as-is, including leaving `DEEPSEEK_V3_HF_MODEL` and `DEEPSEEK_V3_CACHE` unset so the conftest defaults for `model_path` and `cache_path` apply — the old galaxy perf impl set neither and the test has been reaching its assertions that way for a month.

The destination does add one thing the source lacked: `HF_HUB_OFFLINE=1`, for any SKU without a `weights-cache-mode`. That is exactly what broke the Wan2.2 entry in #54306, so I checked rather than assumed — every `from_pretrained` on this path resolves a local directory (`AutoConfig.from_pretrained(model_path)` with `model_path` defaulting to `models/demos/deepseek_v3/reference`, and `load_tokenizer` against a validated local path), and the DeepSeek unit suites migrated in #54184 already run green under offline mode. No `unset` needed here.

#### Why this needed #53553 (now merged)

The test had not passed once in 30 days before that fix. Of the 49 pipeline runs since 2026-07-26 the job ran 33 times — the other 16 were dispatches filtered to another model — and failed all 33, overwhelmingly on one stale constant:

```
2026-07-26  AssertionError: Total Op-to-Op Latency 142.70 us is outside the expected range [123.19, 130.81] us
2026-08-25  AssertionError: Total Op-to-Op Latency 139.52 us is outside the expected range [123.19, 130.81] us
```

#53553 moved that expectation from 127.00 to 139.41 µs and gave the metric its own 4% margin, widening the band to roughly [133.8, 145.0]. That change is in this branch now. Measured values across the month span 135.55–144.30 µs, so all of them land inside it — 144.30 with only 0.7 µs to spare.

**One thing #53553 does not address.** Kernel duration keeps its 10050.69 µs ± 3% expectation, and the latest run measured 10205.18 µs against a 10352.21 µs ceiling — 1.4% of headroom. E2E is similar. So this entry should go green once #53553 lands, but the next drift trips a different assertion, and the two-sided bands mean that will read as a failure whether it is a regression or not. Worth a follow-up that resets all three together.

Roughly a third of the 33 failures never reached the assertions at all: of thirteen sampled runs, two died in the tracy subprocess (`subprocess.CalledProcessError` from `python3 -m tracy`) and one on `TT_FATAL: Requested mesh shape MeshShape([4, 8]) requires 32 devices, but only 16 devices are …` — Galaxy fleet health, the same class of fault that hit #54306.

### Pipeline retired

Deleted: `galaxy-perf-tests.yaml`, `galaxy-perf-tests-impl.yaml`, `tests/pipeline_reorg/galaxy_perf_tests.yaml`.

| File | Reference removed |
|---|---|
| `pipeline-select-galaxy.yaml` | the `galaxy-perf` input, the `run-model-perf` output, its seven `echo` lines across all trigger types, the `galaxy-model-perf-tests` job, and the build's `tracy:` flag which that output drove |
| `aggregate-workflow-data.yaml` | workflow list entry |
| `scripts/compare_workflow_runs.py` | workflow list entry |
| `.github/workflows/silencer.md` | dispatchable-workflow list entry |

Two notes on those. **`tracy` is now `false`** in `pipeline-select-galaxy.yaml`'s `build-artifact` — `run-model-perf` was its only driver, and no job left in that dispatcher (sanity, unit, integration, e2e) needs a profiler build. If a future Galaxy job does, it needs its own flag rather than inheriting one from a deleted pipeline.

**`silencer.md` feeds a generated lock**, so it was recompiled with `gh aw compile` v0.86.2, the version in the committed lock's own metadata. The diff is one dispatch target: the `galaxy-perf-tests` `inputSchema` block, its occurrences in the safe-outputs config JSON, and the frontmatter hash. The dispatch list goes 34 → 33 and the in-file count comment is updated to match.

### Budget

| Team / pipeline / SKU | Was | Now | Used |
|---|---|---|---|
| `models.e2e_tier1.wh_galaxy_perf` | 175 | 205 | 195 |
| `models.perf.wh_galaxy_perf` | 55 | removed | — |
| `models.device_perf.wh_galaxy_perf` | — | 25 | 20 |

The `models.perf` stanza is removed outright. `galaxy_perf_tests.yaml` was the only registry producing a `(models, wh_galaxy_perf)` key under the `perf` workflow, and `wh_galaxy_perf` was that stanza's only SKU, so zeroing it would have left an empty mapping. `llk_perf_tests.yaml`, `fabric_perf_tests.yaml` and `runtime_perf_tests.yaml` have no `team: models` entries, and both still verify clean. A future entry needing this key fails with `Could not find a 'time_budget' for Team 'models', Workflow 'perf', SKU '…'`, which is a clearer error than exceeding a 0-minute budget.

Verified with the CI check itself — `models_e2e_tests.yaml` passes at e2e tiers 1, 2 and 3 with `wh_galaxy_perf` at 195/205, and `models_device_perf_tests.yaml` passes at 20/25. `py_compile` passes on the edited Python, and every touched YAML parses.

### Testing

Not dispatched yet, deliberately. The Wan entry's `wh_galaxy_perf` leg is verified in #54306 for its first two calls and the third is the command this pipeline has run daily; the DeepSeek entry cannot pass until #53553 lands. Both are worth running once this retargets to `main` so the result reflects what actually merges.

### Checklist

- [ ] (Tier 1) Models E2E Tests — `model: wan-2.2-t2v`, `sku: wh_galaxy_perf (WH Galaxy perf)` — after #54306 merges
- [ ] (Tier 1) Models Device Perf Tests — `model: deepseek-v3`, `sku: wh_galaxy_perf (WH Galaxy perf)` — after #53553 merges
- [ ] (Galaxy) Choose your pipeline — dispatch to confirm the removed perf wiring left it valid

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/retire-galaxy-perf-pipeline)


roseli-TT (approved):

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -1293,6 +1293,8 @@
       -k "resolution_480p and t2v and 2x2_sp0tp1" || fail=1
     pytest --timeout 1500 models/tt_dit/tests/models/wan2_2/test_pipeline_wan.py \
       -k "4x8sp1tp0nl4_ring_is_fsdp1 and resolution_720p" || fail=1
+    pytest --timeout 1200 models/tt_dit/tests/models/wan2_2/test_performance_wan.py \
```
jamesleeTT, line 1299: remove mesh filter here

Issue #54330, opened 2026-08-25
Metal 2.0 port: moreh/moreh_nll_loss/moreh_nll_loss_step2


mcw-anasuya:
Port `ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2` to Metal 2.0 / Host 2.0 APIs (Gen1 WH/BH target).

Parent: #41908

## Scope

One `descriptor`-concept device operation, `MorehNllLossStep2DeviceOperation`, one program factory with three rank builders (2d / 3d / 4d), and seven own kernels — three readers, three writers, one compute kernel shared across the rank paths — plus the `moreh_common.hpp` donor. `moreh_nll_loss_step1` is a separate device operation and is not in scope here.

## Audit verdict: GREEN

Every gate clears. The work is concentrated in CB endpoints, which carry four must-fix items — two of them compile failures rather than mechanical conversions:

* The compute kernel constructs `dfb_divisor` outside `#if defined(DIVISOR)` while every use is inside. `c_3` isn't allocated at all when divisor is absent, so the six no-divisor configs won't compile until the declaration moves inside the guard.
* `c_24` (tmp_weight) must keep an unconditional spec and be self-looped when weight is absent — the compute kernel names its index in every config, so the token has to resolve even where nothing FIFO-touches it.
* `c_7` (weight scratch) has zero endpoints in every 2d and 3d config and must be dropped there; the 4d reader genuinely uses it and keeps it.
* Four dead CB declarations get deleted rather than converted — converting them would manufacture endpoints that don't exist.

## Notes

* WH/BH only; basic port, style fixup deferred.
* Twelve reachable configurations: 3 rank paths × weight present/absent × divisor present/absent.
* `unpack_modes` is not optional here. Five intermediates become `Float32` exactly when `fp32_dest_acc_en` is on, and the compute kernel consumes all five, so each needs an explicit `UnpackToSrc` entry even though the legacy vector was uniform `Default`. The trigger is the DFB's format, not the op's tensor dtypes — this op has no `Float32` tensor anywhere.
* **Test coverage notes:** the rank-4 path is reached in `test_moreh_nll_loss.py` only via a rank-6 shape, since `impl_4d` serves `rank >= 4`. 44 of 97 collected cases skip themselves on `bfloat8_b`, which `validate_inputs` rejects anyway. No C++ gtest coverage exists.
[...]

PR #54193, opened 2026-08-24
[Metal 2.0] Port `kv_cache`

gajanan-choudhary:
### Issue
Closes #54176.

### Summary
* Ports `ttnn/cpp/ttnn/operations/kv_cache` (`UpdateKVCacheOperation`) to the Metal 2.0 host API (Gen1 WH/BH)
* Both program factories move from the `ProgramDescriptor` (`create_descriptor`) API to **`CustomProgramSpecFactoryConcept`** — a `create_program_artifacts` returning a `ProgramSpec` + `ProgramRunArgs`, plus a translated `override_runtime_arguments` returning `ProgramRunArgs`:

- **`UpdateCacheMultiCoreProgramFactory`**: reader/writer + per-core-group compute `KernelSpec`s (work-split multiplicity preserved as CTAs); `interm0`/`interm1` (`c_24`/`c_25`) modeled as **aliased DFBs** (`advanced_options.alias_with`); `unpack_modes` entries derived for the `fp32_dest_acc_en` path; sharded input via a borrowed-memory DFB.
- **`FillCacheMultiCoreProgramFactory`**: reader + the existing `eltwise/unary` `writer_unary_interleaved_start_id_metal2` donor fork (reused, not re-forked).
- Kernels: named args, `dfb::`/`tensor::` bindings, `DataflowBuffer`. Address RTAs and `TensorAccessorArgs` plumbing dropped.
- The custom `compute_program_hash` and the device-operation class are left intact.

This is the codebase's first op on `CustomProgramSpecFactoryConcept`.

### Notes for reviewers

- **Cache-hit override**: each translated `override_runtime_arguments` re-applies only the hash-excluded, attribute-derived scalars (`cache_start_id`, `Wbytes`, `tile_update_offset`, `batch_read_offset`); tensor addresses refresh through the typed binding channel (the borrowed input DFB re-resolves from the input `TensorArgument`). `compute_update_cache_dynamic_args` / `compute_fill_cache_start_ids` stay the shared source of truth for the work-split, so create and override cannot drift.
- **Aliased DFBs (`c_24`/`c_25`)**: two distinct `buffer_index`es over one L1 region, each a clean 1P+1C — not multi-binding. Confirmed against the kernel-touch census.
[...]

PR #54331, opened 2026-08-25
Port moreh_nll_loss_step2 to Metal 2.0

mcw-anasuya:
### Issue

https://github.com/tenstorrent/tt-metal/issues/54330

### Summary

Metal 2.0 port of `moreh/moreh_nll_loss/moreh_nll_loss_step2` — the single factory, all three rank builders (2d/3d/4d) and all seven kernels, converted to the base `ProgramSpecFactoryConcept`. Behaviour-preserving; the mechanical parts are as prescribed.

### Notes for Reviewers

* The compute kernel's `divisor` DFB declaration moves inside `#if defined(DIVISOR)`; `tmp_weight`'s spec stays unconditional and is self-looped when weight is absent, since the kernel names it in every config. `weight_scratch` is dropped in 2d/3d (zero endpoints there) and kept in 4d.
* Two compute `KernelSpec`s in two `WorkUnitSpec`s, with the dead per-core compute RTA **deleted** rather than adopted.
* `unpack_modes` carries five explicit `UnpackToSrc` entries when `fp32_dest_acc_en` is on, written through the generation-neutral `unpack_modes()` accessor rather than `std::get<ComputeGen1Config>`, which throws on a Gen2 target. `opt_level` is stated explicitly as `O3` inside the shared `make_compute_spec` helper, so it can't be set on one instance and missed on the other.
* Seven dead reader RTAs dropped, not named — the brief listed four and the inventory found three more of the identical kind. Two near-twins were kept because they *are* used: the 2d reader's `N` and the 3d reader's `element_size`.
* Dead includes removed (`moreh_helper_functions.hpp` in the factory, `moreh_common.hpp` in the 4d writer); the factory now includes `math.hpp` and the compute-config headers for the symbols it names directly.

**Noted, not addressed:**

* Only `num_tiles_per_core` and `start_id` genuinely vary per node; the rest are `common_runtime_arg_values` candidates (3 of 5 in the 2d reader, 4 of 6 in the 3d and 4d readers, 2 of 4 in the 3d writer). Promoting them changes dispatch semantics, so it's outside a behaviour-preserving port; the list is recorded in the report for that pass.
* DFB bindings are written as `DFBBinding` aggregates rather than through `ProducerOf` / `ConsumerOf`, which the porting recipe puts out of bounds.
* `reduction` is unused by the factory but feeds the program hash, so two calls differing only in it compile identical programs. Preserved — the cache key isn't the port's to change.
* Three `TT_FATAL(false, "Core not in specified core ranges.")` guards are gone with the dead RTA branch they were the `else` arm of; the identical condition is still checked a few lines earlier by the surviving `TT_THROW`.

**Verified:** `53 passed, 44 skipped, 106 deselected` on `test_moreh_nll_loss.py` and `test_moreh_nll_loss_unreduced.py` with `-k "not backward"`. Skips are the tests' own `bfloat8_b` guard.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:anasuya/metal2_port_moreh_nll_loss_step2)


ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/device/kernels/moreh_nll_loss_step2_kernel.cpp
```diff
@@ -6,39 +6,34 @@
...
-    constexpr uint32_t cb_divisor_recip = tt::CBIndex::c_27;
-    DataflowBuffer dfb_divisor_recip_obj(cb_divisor_recip);  // 1/divisor
-    constexpr uint32_t cb_tmp3 = tt::CBIndex::c_28;
-    DataflowBuffer dfb_tmp3_obj(cb_tmp3);
-
-    constexpr uint32_t cb_output = tt::CBIndex::c_16;
-    DataflowBuffer dfb_output_obj(cb_output);
+    constexpr auto per_core_tile_cnt = get_arg(args::per_core_tile_cnt);
+
+    DataflowBuffer dfb_tmp_weight_obj(dfb::tmp_weight);
```
Aswinmcw, line 14: `dfb_tmp_weight_obj` (and `tmp1`/`divisor_recip`/`tmp3` below it) are constructed unconditionally, and `compute_kernel_hw_startup` names `dfb::tmp_weight` as icb0 — that is what forces the host to allocate and nominally self-loop those buffers in every configuration. In the no-weight/no-divisor build 4 of the 6 compute-side DFBs are allocated and never touched (~4 tiles of L1 per core).

`tmp_weight` and `tmp_input` carry the same `fp32_dest_acc_en_data_format` in every configuration, so guarding the construction with `#if defined(WEIGHT)` and passing `dfb::tmp_input` as icb0 in the no-weight build is format-identical and lets `DFB_TMP_WEIGHT` drop out of the spec entirely — the same treatment you already applied to `divisor` in this file.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/device/moreh_nll_loss_step2_program_factory.cpp
```diff
@@ -286,127 +477,185 @@ tt::tt_metal::ProgramDescriptor moreh_nll_loss_step2_impl_3d(
     auto [math_fidelity, math_approx_mode, fp32_dest_acc_en, packer_l1_acc, dst_full_sync_en] =
         get_compute_kernel_config_args(device->arch(), compute_kernel_config);
 
-    ProgramDescriptor desc;
+    ProgramSpec spec;
+    spec.name = "moreh_nll_loss_step2_3d";
 
-    // create circular buffers
+    // create dataflow buffers
```
Aswinmcw, line 483: Lines 483-582 here are byte-identical to 227-326 in `_impl_2d` — the only difference in the whole 100-line block is `3d reader` vs `2d reader` inside a comment; `_impl_4d` diverges only by the weight-scratch buffer. Same again for the two `WorkUnitSpec` pushes and the `run_args.tensor_args` blocks. Worth hoisting the DFB list / `tensor_parameters` / defines / reader bindings into shared helpers alongside `make_dfb` and `make_compute_spec`, otherwise the three impls have to be kept in sync by hand.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/METAL2_PORT_BRIEF.md
```diff
@@ -0,0 +1,267 @@
+# Metal 2.0 Port Brief — `ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2`
```
Aswinmcw: These four `METAL2_*.md` files add 1,918 lines to the op source directory — about 70% of the diff. The step1 port (#54272) and the other recent Metal 2.0 ports (#54147, #54193) each touched only their source files, so this would be the first in-tree doc set of its kind. Suggest moving the porting notes to the PR description or a dev-docs location rather than shipping them next to the kernels.

Aswinmcw (commented):
Draft, so nothing blocking — three notes inline.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/device/kernels/moreh_nll_loss_step2_kernel.cpp
Aswinmcw: Correction on my part — withdraw this as a change request for this PR. Issue #54330 lists the unconditional `tmp_weight` spec + self-loop as a must-fix, and scopes the work as "basic port, style fixup deferred", so the PR is doing exactly what was asked. The L1 observation stands as follow-up against the parent #41908, not something to change here.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/METAL2_PORT_BRIEF.md
Aswinmcw: Related: the PR description doesn't mention these four files at all. They're 1,918 of the 2,729 added lines (~70% of the diff), so the "Summary" and "Notes for Reviewers" sections read as if this were an 8-file source-only port. Whichever way the files go, the description should account for them.

mcw-anasuya: Will be removed before merging

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/device/moreh_nll_loss_step2_program_factory.cpp
mcw-anasuya: Perhaps this can be addressed in another PR. Little out of scope for this one. What do you think @Aswinmcw ?

Aswinmcw: Agreed — deduplicating the three rank builders is a refactor, not part of a behaviour-preserving port, so it doesn't belong here. Please open a follow-up against the parent #41908 so it doesn't get lost; the three blocks will only drift once someone edits one of them.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/METAL2_PORT_BRIEF.md
Aswinmcw: The PR came out of draft with all four files still in the diff and CI green, so this is now the only thing standing between it and merge. Dropping them is the last step — once they're gone this is an approve from me.

ttnn/cpp/ttnn/operations/moreh/moreh_nll_loss/moreh_nll_loss_step2/device/kernels/reader_moreh_nll_loss_step2_2d.cpp
```diff
@@ -7,64 +7,39 @@
...
 
     uint32_t Ct = (C + TILE_HEIGHT - 1) / TILE_HEIGHT;
 
-    DataflowBuffer dfb_input_obj(cb_input);
-    DataflowBuffer dfb_target_obj(cb_target);
-    DataflowBuffer dfb_tmp_input_obj(cb_tmp_input);
+    DataflowBuffer dfb_input_obj(dfb::input);
+    DataflowBuffer dfb_target_obj(dfb::target);
+    DataflowBuffer dfb_tmp_input_obj(dfb::tmp_input);
 #if defined(WEIGHT)
```
Aswinmcw, line 40: Not this PR — the loop body is untouched — but worth a follow-up while the file is open. At lines 84-86 below, the weight read derives `noc_id = target_val / TILE_WIDTH` and `get_tilized_idx(0, target_val)` from a raw `target_val`, outside the `target_val != ignore_index && 0 <= target_val < C` guard that protects the input read a few lines earlier. For an ignored or out-of-range target the result still comes out right (`tmp_input` is zeroed, so the product is zero), but the NoC read is still issued, and a negative `target_val` divided and widened to `uint32_t` gives a wildly out-of-bounds page id. The 3d and 4d readers guard the same read. Suggest filing it against the parent #41908 with the other ops-team items from #54330.

mcw-anasuya: Out of scope for this PR.

bbradelTT (approved):

PR #5023, opened 2024-01-30
#4998: Add backward support for hypot

KalaivaniMCW:
Add backward support for hypot

PR #54335, opened 2026-08-25
Support LG AI Research's EXAONE 4.5

### LINKED ISSUE #54334: Support LG AI Research's EXAONE 4.5
## Model
[LGAI-EXAONE/EXAONE-4.5-33B](https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B) — LG AI Research's first open-weight VLM: a 33B text decoder (EXAONE 4.0 architecture) + a 1.2B Qwen2.5-VL-style vision encoder.

## Why it needs work in tt-transformers
The text decoder differs from currently supported families in three ways:
- **Post-norm residual order** — no `input_layernorm`; norms apply to attention/MLP *outputs* before the residual adds (`h = x + post_attention_layernorm(attn(x))`).
- **Inverted hybrid RoPE** — llama3-scaled RoPE (theta 1e6, factor 16) on `sliding_attention` layers only; `full_attention` layers are **NoPE** (opposite pairing to Gemma-3's global/local rope split).
- **Multimodal checkpoint, text-first serving** — composite `Exaone4_5_ForConditionalGeneration` config (not in the `AutoModelForCausalLM` mapping), `model.visual.*` + `mtp.*` keys to handle at load.

Additional platform gap found on P150x8: per-device MLP width 27392/8 = 107 tiles (prime) degenerates `find_grid_k_n` to a single core and overflows L1 in `ff_norm`.

## Scope
- [x] Text decoder, TP=8 on P150x8 (PCC-validated vs HF; demo batch-1/32)
- [x] Long-context validation past the 4096 sliding window (8K prefill PCC, needle test)
- [x] On-device vision tower (`models/demos/exaone45_vl`, GQA vision attention)
- [x] vLLM bridge class (`Exaone4_5_ForConditionalGeneration`)
- [ ] Prompts > 128K (blocked on chunked prefill × sliding-window SDPA)
- [ ] MTP speculative decoding (checkpoint's `mtp.*` head; upstream vLLM's own implementation measures only ~3.2% draft acceptance on these weights — appears unvalidated upstream)

Implementation in the linked PR.

changh95:
This PR implements [EXAONE 4.5](https://huggingface.co/LGAI-EXAONE/EXAONE-4.5-33B) model from LG AI Research.

EXAONE 4.5 is a popular multimodal LLM model used in Korea. Many customers ask for EXAONE 4.5 support, so implementation is needed.

For now, It's 8 * p150 setup with TP=8.

Closes #54334

## What

Adds full inference support for `LGAI-EXAONE/EXAONE-4.5-33B` to tt-transformers, validated on P150x8 (8x Blackhole p150b, 1x8 mesh, FABRIC_1D). Six commits:

1. **Text decoder support** (`64c1c717938`) 
2. **Host-vision hybrid demo** (`2d93ef1b5e7`)
3. **Perf/accuracy tables + refpt** (`18fa1aa38b6`)
4. **Support vision encoder** (`248bb52f2e5`) — `models/experimental/exaone45_vl` 
5. **vLLM bridge class** (`6b601646f26`) 
6. **Opt-in FF2-BFP4 decoder config** (`413be92a3b7`) — +7.4% decode (39.0 vs 36.3 t/s/u) at 90.40/98.20 token accuracy; kept out of the default (top5 sits on the 98 threshold).

## Validation (P150x8, real weights)

| Gate | Result |
|---|---|
| test_attention (b1/b32) | PCC ≥ 0.99 |
| test_decoder (post-norm wiring) | PCC 0.999998 |
| test_model, 64 layers, teacher-forced (accuracy) | PCC 0.968–0.989 (gate 0.94) |
| test_model_prefill @ 8192 (past the 4096 window) | output PCC 0.9920, per-layer KV ≈ 0.99 |
| needle @ 8K (7K-token gap across NoPE global layers) | exact retrieval |
| ci-token-matching | 91.40/99.60 (performance), 95.60/100.00 (accuracy) |
| demo batch-1 / batch-32 | 36.3 t/s/u TTFT 51 ms / 1,090 t/s aggregate |
| vision tower (attention/merger/28 blocks/wrapped) | 0.9997 / 0.9995 / ≥0.995 / 0.991–0.9993 |
| e2e image chat, tower on device | tower-vs-HF PCC 0.9984 |

## Companion change & follow-ups
- **vLLM serving requires a plugin-side registration**: the vLLM TT plugin prepends `TT` to the checkpoint architecture and has no built-in EXAONE entry, so the bridge class here is not picked up by the plugin alone. tenstorrent/tt-inference-server#5023 supplies the registration via an `EXTRA_MODELS_DIR` bundle (validated end-to-end); direct plugin users need an equivalent `ModelRegistry` entry in the engine-core process **and `--max-model-len 131072`** — the checkpoint advertises 262144 tokens but prefill above 128K would hit the unsupported sliding-window chunked path (the bridge class now rejects larger `max_seq_len` up front). Treat #5023 as the deployment companion for serving.
- **CI onboarding** (tiers + pipeline_reorg registries) is tracked in #55121, which stays open after this PR merges.

## Out of scope / known limits
- Prompts > 128K (chunked prefill × sliding-window SDPA raises NotImplementedError upstream).
- `update_weights` hot-swap not extended for post-norm.
- MTP (`mtp.*`) weights are dropped: we verified with vLLM 0.25.1's CPU backend that even vLLM's own `exaone4_5_mtp` implementation gets only a **3.2% draft acceptance rate** on these dense-33B weights (target outputs perfect) — the dense-4.5 MTP path appears unvalidated upstream; details available on request.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:changh95/exaone45)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:changh95/exaone45)


changh95:
Addressed the review in a6df483da1c. **Fixed (6):**

- **`model_config.reference_vision_model`** — missing / unmatched / partially-loaded vision weights now raise `RuntimeError` instead of silently returning a randomly initialized tower; random init is reserved for the explicit `dummy_weights` branch.
- **`vision_attention.py` chunked branch** — the copied chunked-SDPA path referenced undefined `keys_BKSD`/`values_BKSD` and could never run; replaced with an explicit `NotImplementedError` (the vision tower has no KV cache).
- **`reference/functional.py`** — removed the leftover debug `print()`s.
- **`test_model.py`** — removed the unused duplicate `VisionTransformer` construction (no more double 28-block tower allocation).
- **`test_wrapped_model.py`** — tuple-valued `pooler_output` is now normalized via `torch.cat` before `comp_pcc` (multi-image safety).
- **`test_vision_block.py`** — dropped the PCC 0.85 relaxation for the last four blocks; every block now gates at 0.99 (all 28 measured ≥ 0.995 on real weights).

Re-ran the full exaone45_vl suite on P150x8 after the changes: `test_vision_attention` 1 passed, `test_patch_merger` 1 passed, `test_vision_block` 1 passed (28/28 at the 0.99 gate), `test_model` 6 passed, `test_wrapped_model` 3 passed.

**Kept as-is (2), with rationale:**
- **PERF.md rows** — the file is marked deprecated in favor of `model_targets.yaml`, but every currently supported model still has rows in both; omitting only EXAONE would make the table inconsistent. Happy to drop the rows if the team prefers starting the migration here.
- **CI tier registration** (`model_ci_tiers.md`, pipeline_reorg registries) — tracked as a follow-up in #54334; tier assignment allocates scheduled CI capacity, which seems like a maintainer call rather than something this PR should self-assign.

yieldthought:
Codex review:

- [models/tt_transformers/tt/generator_vllm.py:1012](https://github.com/tenstorrent/tt-metal/blob/a6df483da1ced7ec2b013accf3398528663c7325/models/tt_transformers/tt/generator_vllm.py#L1012) says the TT plugin can map this class directly, but the [current plugin](https://github.com/tenstorrent/vllm-tt-plugin/blob/859be837cb7bcae998597e5a21579d3cc3801227/src/vllm_tt_plugin/platform.py#L1494) prepends `TT` to the checkpoint architecture and rejects it unless `TTExaone4_5_ForConditionalGeneration` has been registered; its built-in registry has no EXAONE entry. tenstorrent/tt-inference-server#5023 supplies that missing registration through an `EXTRA_MODELS_DIR` bundle and reports a successful server validation, but this PR does not link or require the companion change. Used by itself, the advertised vLLM bridge still fails during plugin architecture validation before this class is loaded. Suggestion: either add the core plugin registration or declare the companion PR as a merge/deployment dependency and document the bundle requirement for direct plugin users.

- [models/tt_transformers/README.md:9](https://github.com/tenstorrent/tt-metal/blob/a6df483da1ced7ec2b013accf3398528663c7325/models/tt_transformers/README.md#L9) promotes EXAONE to verified support, and [models/model_targets.yaml:563](https://github.com/tenstorrent/tt-metal/blob/a6df483da1ced7ec2b013accf3398528663c7325/models/model_targets.yaml#L563) marks its P150x8 targets active, but the model is absent from `models/model_ci_tiers.md`, `tests/pipeline_reorg/models_unit_tests.yaml`, and `tests/pipeline_reorg/models_e2e_tests.yaml`; the PR's model/runtime jobs were consequently all skipped. The author says CI onboarding is tracked as a follow-up in #54334, while this PR says `Closes #54334`, so merging it will close the only stated tracker without adding recurring coverage for the new post-norm/NoPE paths or vision tower. Suggestion: register at least the critical unit and end-to-end accuracy coverage here, or create and link a separate CI-onboarding issue that remains open after this PR merges.

changh95:
Thanks @yieldthought — both points addressed:

1. **Plugin registration**: You're right, the docstring overstated what the plugin does on its own. Fixed in cb34244a73d — it now documents that the vLLM TT plugin prepends `TT` to the checkpoint architecture, has no built-in EXAONE entry, and therefore requires a `ModelRegistry` registration in the engine-core process. The PR description now declares tenstorrent/tt-inference-server#5023 (the `EXTRA_MODELS_DIR` bundle, validated end-to-end on P150x8) as the deployment companion for serving, with a note for direct plugin users to supply an equivalent registration.

2. **CI onboarding tracker**: Good catch on `Closes #54334` collapsing the follow-up. Opened #55121 as a dedicated CI-onboarding issue (tiers + `pipeline_reorg` unit/e2e registries, with the concrete coverage checklist) that stays open after this merges; the PR description links it. I've left the final tier assignment to maintainers since it allocates scheduled CI capacity — happy to prepare the registry PR once a tier is agreed.

models/model_targets.yaml
```diff
@@ -553,6 +553,30 @@ targets:
             accuracy:
               top1: 95
               top5: 100
+  exaone-4.5-33b:
+    aliases: ["EXAONE-4.5-33B", "LGAI-EXAONE/EXAONE-4.5-33B"]
```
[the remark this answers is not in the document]
changh95: CI onboarding now has a dedicated tracker that survives this PR merging: #55121 (tier assignment + pipeline_reorg unit/e2e registry entries, with a coverage checklist). Leaving the tier choice to maintainers since it allocates scheduled CI capacity.

models/model_targets.yaml
```diff
@@ -553,6 +553,30 @@ targets:
...
+              # Measured 2026-08-25 (performance/bfp8, ci-token-matching): 91.40 / 99.60.
+              top1: 89
+              top5: 98
+          - batch_size: 32
+            seq_len: 712
+            status: active
+            perf:
+              # Measured 2026-08-25 on P150x8: 34.1 t/s/u (1090 t/s), TTFT 14.4 ms.
+              decode_t/s/u: 32
+              prefill_time_to_first_token: 25
```
yieldthought: Codex: The recorded 14.4 ms result falls below this target's +/-15% acceptance band (21.25-28.75 ms), so the validator will reject it as a stale target.

```suggestion
              prefill_time_to_first_token: 14.4
```

models/tt_transformers/tt/generator_vllm.py
```diff
@@ -1006,6 +1006,99 @@ def allocate_kv_cache(self, *args, **kwargs):
...
+
+    Named after the HF architecture string. Note that the vLLM TT plugin
+    does NOT pick this class up automatically: the plugin prepends ``TT``
+    to the checkpoint architecture and its built-in registry has no EXAONE
+    entry, so serving requires registering
+    ``TTExaone4_5_ForConditionalGeneration`` -> this class in the vLLM
+    ``ModelRegistry`` of the engine-core process. tt-inference-server does
+    this via an ``EXTRA_MODELS_DIR`` plugin bundle (see
+    tenstorrent/tt-inference-server#5023); direct plugin users must supply
+    an equivalent registration.
```
yieldthought: Codex: The direct-plugin path also needs the model-specific context cap; without it, vLLM uses the checkpoint's 262144-token default.

```suggestion
    tenstorrent/tt-inference-server#5023); direct plugin users must supply
    an equivalent registration and set ``--max-model-len 131072``; longer
    contexts require unsupported sliding-window chunked prefill.
```

models/tt_transformers/tt/generator_vllm.py
```diff
@@ -1006,6 +1006,99 @@ def allocate_kv_cache(self, *args, **kwargs):
...
+        cls,
+        hf_config,
+        mesh_device,
+        max_batch_size,
+        max_seq_len,
+        n_layers=None,
+        tt_data_parallel=1,
+        optimizations: str = "performance",
+    ):
+        tt_model, model_args = initialize_vllm_text_transformer(
```
yieldthought, line 1064: Codex: Fail before model construction when a caller supplies the checkpoint's advertised 262144-token context, which this implementation cannot prefill.

```suggestion
        if max_seq_len > 131072:
            raise ValueError(
                f"EXAONE-4.5 currently supports max_seq_len <= 131072; got {max_seq_len}. "
                "Sliding-window chunked prefill is not supported."
            )
        tt_model, model_args = initialize_vllm_text_transformer(
```

yieldthought:
Codex review:

1. [models/model_targets.yaml:577](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/model_targets.yaml#L577) centers the active batch-32 TTFT target at 25 ms even though the adjacent measurement records 14.4 ms. The validator applies the file-wide symmetric +/-15% band, so this target accepts only 21.25-28.75 ms and will reject the PR's own result as "much better than expected, update target" as soon as the model is onboarded to CI in [#55121](https://github.com/tenstorrent/tt-metal/issues/55121). Suggestion: center the target on 14.4 ms (or a newer representative measurement) before enabling the scheduled job.

2. [models/tt_transformers/tt/generator_vllm.py:1055](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/generator_vllm.py#L1055) forwards vLLM's `max_model_len` unchanged, but [models/tt_transformers/model_params/EXAONE-4.5-33B/config.json:86](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/model_params/EXAONE-4.5-33B/config.json#L86) advertises 262144 tokens while [models/tt_transformers/tt/model_config.py:2468](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/model_config.py#L2468) only allows a single prefill chunk through 131072 tokens. The [TT plugin loader](https://github.com/tenstorrent/vllm-tt-plugin/blob/8b62b2bb03d4156d7eea82585ed0eaed2d067d70/src/vllm_tt_plugin/loader.py#L38-L45) deliberately preserves the HF-derived default and passes it here; its fallback all-user KV budget is 131072, so the documented direct-plugin setup fails startup unless the operator knows to add a cap. If that budget is enlarged, a prompt above 131072 instead enters the internal chunked path at [models/tt_transformers/tt/generator.py:1444](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/generator.py#L1444) and the sliding-attention layers raise `NotImplementedError` at [models/tt_transformers/tt/attention.py:1204](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/attention.py#L1204). The companion [tt-inference-server#5023](https://github.com/tenstorrent/tt-inference-server/pull/5023) avoids this with `max_context: 131072`, but the direct-plugin instructions only mention registration. Suggestion: document `--max-model-len 131072` alongside the direct registration requirement and reject larger `max_seq_len` values here with a model-specific error until sliding-window chunked prefill is supported.

yieldthought (approved):
approving to unblock but take a look at codex review please

mtairum (changes requested):
Can you move the model from models/demos to `models/experimental/exaone45_vl` instead?

mtairum:
### Tier CI dispatched on `changh95/exaone45`

**Run:** https://github.com/tenstorrent/tt-metal/actions/runs/33778099160 (`All Model Tests`, tiers 1+2+3, unit + e2e + sweep, `skus=all`, `model=all`)

**Why `model=all`:** this PR touches six shared tt-transformers files that are on the hot path for essentially every LLM in the tier registries, so the affected set can't be expressed as a single substring filter:

- `models/tt_transformers/tt/decoder.py` — `use_post_norm` branch; `attention_norm` may now be `norm=None`
- `models/tt_transformers/tt/distributed_norm.py` — new gather-only (`norm=None`) mode
- `models/tt_transformers/tt/model.py` — `nope` kwarg on the global rope setup; `rope_scaling=` now passed to the **local** rope setup
- `models/tt_transformers/tt/model_config.py` — new defaults, `force_text_only`, `get_hf_model_cls` fallback, `load_state_dict` vision/MTP filter
- `models/tt_transformers/tt/rope.py` — `get_rot_mats(..., nope=False)`
- `models/tt_transformers/tests/generate_reference_outputs.py` — `AutoModelForImageTextToText` fallback

Tier models importing those modules:

- **T1** — llama3.1-8b (n150 / bh_p150 / galaxy-dp), qwen3-32b-galaxy, qwen3.6-27b, gpt-oss-120b, gemma-4-\*, tt-transformers-common
- **T2** — llama3.1-8b (+dp), llama3.2-90b-vision, llama3.3-70b, mistral-7b, mixtral-8x7b, qwen2.5-32b, qwen2.5-coder-32b, qwen3-32b
- **T3** — gemma-2-2b/9b, llama3.2-1b/3b/11b-vision, mistral-small-3.1-24b, phi-3-mini, qwen2.5-7b/72b, qwq-32b, qwen3-0.6b/1.7b

Sweep is included because llama3.1-8b (T1) plus llama3.3-70b and qwen3-32b (T2) have tt-transformers sweep rows. Exabox SKUs (`bh_sc1/sc4/sc16`) are excluded — `skus=all` deliberately omits them.

Static review of the shared diffs suggests no behaviour change for existing models: every new branch is gated on `use_post_norm` / `use_global_nope` / `force_text_only`, all defaulted off in `ModelArgs.__init__`, and the one changed call site (local rope `rope_scaling=`) passes `None` for every non-EXAONE model, which matches the existing parameter default in `RotarySetup`, `HfRotarySetup` and `HfRotarySetupOld`. This run is to confirm that empirically, since the tier pipelines are scheduled-on-main only and PR CI gives no tier signal.

changh95:
> Codex review:
> 
> 1. [models/model_targets.yaml:577](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/model_targets.yaml#L577) centers the active batch-32 TTFT target at 25 ms even though the adjacent measurement records 14.4 ms. The validator applies the file-wide symmetric +/-15% band, so this target accepts only 21.25-28.75 ms and will reject the PR's own result as "much better than expected, update target" as soon as the model is onboarded to CI in [#55121](https://github.com/tenstorrent/tt-metal/issues/55121). Suggestion: center the target on 14.4 ms (or a newer representative measurement) before enabling the scheduled job.
> 2. [models/tt_transformers/tt/generator_vllm.py:1055](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/generator_vllm.py#L1055) forwards vLLM's `max_model_len` unchanged, but [models/tt_transformers/model_params/EXAONE-4.5-33B/config.json:86](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/model_params/EXAONE-4.5-33B/config.json#L86) advertises 262144 tokens while [models/tt_transformers/tt/model_config.py:2468](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/model_config.py#L2468) only allows a single prefill chunk through 131072 tokens. The [TT plugin loader](https://github.com/tenstorrent/vllm-tt-plugin/blob/8b62b2bb03d4156d7eea82585ed0eaed2d067d70/src/vllm_tt_plugin/loader.py#L38-L45) deliberately preserves the HF-derived default and passes it here; its fallback all-user KV budget is 131072, so the documented direct-plugin setup fails startup unless the operator knows to add a cap. If that budget is enlarged, a prompt above 131072 instead enters the internal chunked path at [models/tt_transformers/tt/generator.py:1444](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/generator.py#L1444) and the sliding-attention layers raise `NotImplementedError` at [models/tt_transformers/tt/attention.py:1204](https://github.com/tenstorrent/tt-metal/blob/cb34244a73d2e63dbeb441b87e5d3222bdb11ab3/models/tt_transformers/tt/attention.py#L1204). The companion [tt-inference-server#5023](https://github.com/tenstorrent/tt-inference-server/pull/5023) avoids this with `max_context: 131072`, but the direct-plugin instructions only mention registration. Suggestion: document `--max-model-len 131072` alongside the direct registration requirement and reject larger `max_seq_len` values here with a model-specific error until sliding-window chunked prefill is supported.

@yieldthought 
Thanks, addressed these issues

changh95:
> Can you move the model from models/demos to `models/experimental/exaone45_vl` instead?

@mtairum 
done

mtairum:
Thank you for the fast updates!

One final set of changes before I approve (mostly cleanup of code duplication).
Let me know when you have these changes and I'll approve the PR 🙇 

### Vision tower duplicates `models/demos/qwen25_vl` — please import instead of forking

Thanks for moving the tower to `models/experimental/` — that's the right home, and it also takes the vision code out of the tier-eligible tree.

One thing left to address: most of `models/experimental/exaone45_vl/` is a copy of `models/demos/qwen25_vl/`, not new code.

| file | vs `qwen25_vl` counterpart |
|---|---|
| `tt/common.py` (315) | md5-identical |
| `tt/vision_mlp.py` (125) | md5-identical |
| `tt/vision_rmsnorm.py` (94) | md5-identical |
| `reference/functional.py` (487) | identical except two deleted `print()` calls |
| `tt/vision_block.py` (125) | 2 changed lines, both imports |
| `tt/vision_attention.py` (519) | 1 changed import + the chunked-SDPA branch |

That's ~1,665 of the 2,286 new lines, with about 7 lines of real change. The concern is drift: a fix to `qwen25_vl/tt/common.py` will not reach EXAONE.

**The useful part:** your dead-code removal found a genuine bug in the original. In `models/demos/qwen25_vl/tt/vision_attention.py:467`, the `chunk_start_idx is not None` branch passes `keys_BKSD` and `values_BKSD`, which are never defined in that file — it would raise `NameError` if taken. Dead in practice (the tower has no KV cache, so `chunk_start_idx` is always `None`), but it's a live landmine in a model that runs in Tier 2 (Qwen2.5-VL-72B) and Tier 3 (Qwen2.5-VL-32B).

Suggested order:

1. Fix that branch in `models/demos/qwen25_vl/tt/vision_attention.py` as part of this PR — the same `raise NotImplementedError` you already wrote. It's a one-line, behaviour-neutral change to an unreachable branch, and having it here is what unblocks steps 2 and 3.
2. `vision_attention.py` and `vision_block.py` then become zero-delta: delete both from `exaone45_vl` and import from `qwen25_vl`.
3. Delete `tt/common.py`, `tt/vision_mlp.py`, `tt/vision_rmsnorm.py` and `reference/functional.py`, importing all four from `qwen25_vl`.

`models/experimental/` importing from `models/demos/` is established here — 58 files under `models/experimental/` already do it — so no concern about the direction.

That leaves `tt/model.py`, `tt/model_config.py` and `tt/patch_merger.py`, which have real EXAONE deltas (143 / 49 / 21 diff lines) and should stay as they are.

While you're in there: the retained files still carry Qwen naming in docstrings and the `qwen2_5_vision_transformer_preprocess` entry point, which reads as if EXAONE runs a Qwen tower. Worth a pass once the copies are gone.

Consolidating the eight vision towers in the repo into a shared home is a separate and much larger job — not asking for that here.

changh95:
@mtairum , thank you for your review.

Just replaced exaone vision tower files with qwen2.5 vision tower imports.

mtairum (approved):
Re-reviewed. Looking good. 

Thank you for addressing the changes!

dpopovTT (approved):

PR #54338, opened 2026-08-25
Make release tickets actionable (MINFRA-1651, MINFRA-1216)

dimitri-tenstorrent:
Two things, both about the RELEASE board being trustworthy and actionable.

## 1. The sim check polled the wrong commit (the bug)

`check-rtl-sim-status` resolved the commit to poll with a hardcoded branch lookup:

```bash
SHA=$(gh api repos/tenstorrent/tt-metal/branches/stable --jq .commit.sha)
```

Correct while the job only ran on `stable`. #54205 widened it to version tags,
and on a tag run stable can already have moved past the tagged commit -- so the
job reports **a different commit's** sim result as the release's. A confident
green for a release nobody tested, which is the failure mode the comment above
the job warns about.

Now reads `github.sha` on the tag path: that commit was stable HEAD when the
tag was cut, so it carries the check. The stable path is untouched.

## 2. The tickets were not actionable

RELEASE-4 read `Package and release failed on main (4 job(s))`, unassigned,
with bare SHAs and unclickable text. Compare the human-written tickets on the
same board -- "All big models failed with error: Timeout on device".

**Titles now name what broke.** A job name is a path:

```
build-test-publish (Ubuntu 22.04) / release-demo-tests / ... / Gemma-4-31B e2e tests [bh_quietbox_2]
```

The second segment is the family, the leaf is the thing that failed. One
family gets named with its leaves; a spread across families lists the families
with counts. A fixed character budget bounds the title regardless of job names
-- some carry a parenthesised cluster description and run past 70 characters on
their own.

| | |
| --- | --- |
| before | `Package and release failed on main (4 job(s))` |
| after | `4 jobs failed during release main: demo tests (3), create docker release image (release models) (1)` |
| after (single family) | `9 demo tests failed during release v0.63.0: Demo Test with Perf Metrics (DeepSeek V3 B1 Supercluster 16 aka Superpod 4) (+8 more)` |

**Descriptions link their evidence.** `_adf()` marks up `[label](url)` and bare
URLs, so every producer that goes through it gets clickable output -- the RTL
sim ticket and the evidence report included. The watcher links the commit, the
run, and each failed job to its own log, labelled with the leaf name rather
than the 80-character path every row shares.

**Tickets get an owner.** `JIRA_ASSIGNEE_ACCOUNT_ID`, honoured by all three
producers, so failures land on someone for triage instead of unassigned. Point
the `RELEASE_JIRA_ASSIGNEE` repo variable somewhere else to change owner
without touching code; unset it and the per-test owners in `ai_ip_tests.json`
resume.

## 3. Jobs whose runner never started are not test failures

When a runner fails to come up, the job's work steps are **skipped** and only
its `always()` steps run, so it fails in ~35 seconds having tested nothing.
Those were being counted and ticketed as test failures.

Across the last five failing releases, 2 of 35 failures were this -- and both
are in the run behind RELEASE-4, which is why that ticket read as four failures
spread across two unrelated families. They are dropped now, and the count is
noted in the description. A run where *every* failure is infra files nothing.

Worth stating because it is a common assumption: GitHub **skips** a job whose
`needs` failed, so a dependency cascade never appears as a failure at all. The
jobs that looked like cascades were these runner-setup failures -- and one of
them was a demo model test, so filtering by job type would have kept the bogus
failure and dropped genuine packaging ones.

The rule is "a setup step failed", which is what distinguishes a runner that
never came up from a test that ran and failed.

| | |
| --- | --- |
| before | `Package and release failed on main (4 job(s))` |
| after | `Release stable — 2 demo tests failed: TT-DiT Wan2.2-T2V-A14B +1 more` |

The title derivation moved out of the workflow into `failure_title.py` -- it
had grown past what is reasonable to keep untested in YAML, and it now carries
real rules: which trailing parenthetical is a model name and which is a runner
qualifier, and a length cap so that one 74-character job name cannot crowd out
every other. The collect step is one line again.

## Testing

66 tests, up from 45. The new ones cover the title derivation (12), the assignee reaching the
the assignee reaching the payload, link markup, and the ADF invariant that a
text node is never empty.
Both new wirings were checked negatively -- break the env-var name or the
regex and the suite fails.

The infra classifier was checked against every failed job in the last five
failing releases (35 jobs) by reading their step lists. The collect step was
executed end to end against real job data with `gh` mocked, over four paths:
a normal failure, a run with infra failures mixed in, a run that is entirely
infra, and a non-release branch. Its output was fed through `_adf()`.

RELEASE-4 and RELEASE-2 have been updated in place as a live check: RELEASE-4
now carries the new title and a linked description, and Jira stored 6 link
nodes on the round trip.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgnidash/release-sim-check-sha)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgnidash/release-sim-check-sha)


dimitri-tenstorrent:
Two more commits incorporating David Popov's feedback on the release tickets:

1. **RELEASE-7 layout + AI Summary** — the ticket body now follows the structure of RELEASE-7 (Impact / Failed jobs / Environment with the run link / Other information), and the Other information section carries per-job root causes pulled from the failed run's own `ai_run_summary` artifact (the `ai_summary/run` action already publishes it from `release-build-test-publish`; fetched cross-run via `ci_digest.fetch_run_summary`). The analysis step is best-effort — the ticket still files if the artifact is missing.
2. **Auto-close on green** — tickets get a `package-release-ref:<ref>` label; when a later Package and release run on that ref succeeds, the watcher closes the open ticket(s) with a comment linking the green run. If the Jira workflow has no Done-category transition the ticket keeps the comment but stays open.

The run link David asked for was already in the ticket body; it now lives in the Environment section per the RELEASE-7 template.

dpopovTT (approved):

PR #54340, opened 2026-08-25
Package the models/ tree as tt-metal-models for pip, apt, and dnf

knauth:
The ttnn wheel ships the runtime but not the model implementations, so serving requires a git clone and PYTHONPATH. The Tenstorrent vLLM plugin maps every registered architecture to a dotted path rooted at `models.`, so without an importable models tree every model fails to load, and vLLM reports it only as "Model architectures [...] failed to be inspected".

Add packaging/tt-metal-models/, which builds that tree as a distribution:

- The wheel takes its version from the same setuptools-scm derivation ttnn uses and declares ttnn==<that version>. The coupling is real but invisible in the source (models/ has no version asserts), so a strict pin makes skew a resolver error instead of an AttributeError deep in model construction.
- MANIFEST.in is the source of truth for the non-Python payload. models/ is a PEP 420 namespace package, so discovery needs namespaces=true; and the data lives in directories that hold no .py and whose names are not valid Python identifiers (Llama-3.1-70B-Instruct), so package-data globs cannot reach it.
- import_matrix.py is the gate: it imports every vLLM entry point against the built wheel, each in its own subprocess from a neutral directory, because a models/ directory in the cwd otherwise shadows the installed one and the matrix would silently test the checkout instead.
- build_native_packages.py produces python3-tt-metal-models from the wheel, so file selection has one source of truth rather than a second policy that drifts.

Fix the defects the matrix found, all of which broke serving from a package:

- models/common/utility_functions.py imported pytest at module scope, making a test framework a hard runtime dependency of essentially every model. Moved into the only function that uses it.
- The bge_large_en runners imported assert_with_pcc from their own test package, which the test-directory prune necessarily removes. Moved it beside the comp_pcc it wraps; the test module re-exports it, so the tests are unchanged.
- model_config.py resolved its data root from TT_METAL_HOME before __file__, so a stale value silently loaded model params from an old checkout. It now prefers the installed package when the env-pointed root has no matching tree, and warns.

Known limitations are documented in packaging/tt-metal-models/README.md. Two need a decision rather than code: the llama reference submodule is Llama 2 Community Licensed, not Apache-2.0, so the build never vendors it implicitly; and deepseek_v3_b1 refers to its kernels by repo-relative string literals that the runtime resolves against its kernel search root, not __file__.

### DIFF
--- a/.github/workflows/build-and-test-wheels.yaml
+++ b/.github/workflows/build-and-test-wheels.yaml
@@ -24,9 +24,14 @@ jobs:
     uses: ./.github/workflows/_test-wheels-impl.yaml
     with:
       wheel-artifact-name: ${{ needs.build-artifact.outputs.wheel-artifact-name }}
+  # Independent of build-artifact: tt-metal-models is pure Python and needs no C++ build,
+  # so it runs in parallel rather than waiting on the wheel. It self-verifies via the
+  # import matrix.
+  build-models-wheel:
+    uses: ./.github/workflows/models-wheel.yaml
   publish-wheels:
     name: "Publish wheels to internal PyPI"
-    needs: [build-artifact, test-wheels]
+    needs: [build-artifact, test-wheels, build-models-wheel]
     runs-on: ubuntu-latest
     permissions:
       id-token: write # Required for AWS OIDC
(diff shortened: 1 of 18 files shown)


knauth:
This is a draft to open discussion of packaging this code- it seems to be the only thing preventing us from making models entirely pip-installable, assuming successful deployment of TTTv2.

ezietlowTT:
Can we get eyes on this?  This PR is a major blocker for some model distribution work.

knauth:
Superseded by #54478 — same branch rebased onto current main, pushed in-repo so the workflow-file changes can run CI (the runner hook rejects fork PRs that modify .github/workflows/). Cycode subprocess findings were addressed in-code where meaningful (input validation, shell=False); see the note in #54478.

PR #54351, opened 2026-08-25
Remove redundant DST_ACCUM from SFPU macros; thread dest-acc explicitly

halghTT:
## Summary

- Removes the `DST_ACCUM` argument from `SFPU_UNARY_CALL`, `SFPU_BINARY_CALL`, and `SFPU_TERNARY_CALL` on WH/BH; dest accumulation mode is passed only via kernel `TEMPLATES` when the kernel actually uses it.
- Adds `SFPU_UNARY_INIT_ACCUM` for bare inits that depend on dest-acc (`asin`/`acos`/`softcap` on BH), and requires an explicit `is_fp32_dest_acc_en` template arg on `llk_math_eltwise_unary_sfpu_init` (no default).
- Templates `reconfig_data_format`, experimental SDPA, tilize/untilize, and other non-SFPU compute APIs with `is_fp32_dest_acc_en` so kernels can match the active dest mode after `set_fp32_dest_acc`.
- Updates call sites across compute API headers, DeepSeek/ttnn kernels, and LLK tests; Quasar keeps internal `*_QSR` wrappers that inject `DST_ACCUM_MODE`.

cc @ndivnicTT 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:halgh/sfpu-dest-acc-template-cleanup)


tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/llk_math_eltwise_unary_sfpu_macros.h
```diff
@@ -38,33 +39,40 @@ inline __attribute__((always_inline)) void _sfpu_check_(
...
+ * then the init callback (the FN-like argument), then the templates that
+ * parameterise it.
  */
 
 /*
- * Bare init: no callback.
+ * Bare init: no callback. Uses kernel compile-time DST_ACCUM_MODE.
  *   SFPU_UNARY_INIT(abs);
  */
 #define SFPU_UNARY_INIT(OP) ::ckernel::llk_math_eltwise_unary_sfpu_init<::SfpuType::OP, DST_ACCUM_MODE>()
```
ndivnicTT: You can just add ACCUM in the existing one and not create a new macro. It would require ~100+ updates, but in this PR you already updated a lot of calls, so I think we can go in that direction.

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/llk_math_eltwise_binary_sfpu_macros.h
```diff
@@ -43,14 +43,19 @@ inline __attribute__((always_inline)) void _sfpu_binary_check_(
  * Macro hygiene: DST_IN0/DST_IN1/DST_OUT/VECTOR_MODE are evaluated by both
  * the check and params call. Keep call sites to identifiers/literals, not
  * side effects.
+ * Shared signature matches WH/BH (no DST_ACCUM slot). Inject DST_ACCUM_MODE
+ * for dest-tile bounds; Quasar has no runtime dest-acc switch.
  */
-#define SFPU_BINARY_CALL(DST_SYNC, DST_ACCUM, FN, TEMPLATES, DST_IN0, DST_IN1, DST_OUT, VECTOR_MODE, ...) \
-    (::ckernel::_sfpu_binary_check_<DST_SYNC, DST_ACCUM>(DST_IN0, DST_IN1, DST_OUT, VECTOR_MODE),         \
-     _llk_math_eltwise_binary_sfpu_params_(                                                               \
+#define SFPU_BINARY_CALL_QSR(DST_SYNC, DST_ACCUM, FN, TEMPLATES, DST_IN0, DST_IN1, DST_OUT, VECTOR_MODE, ...) \
```
ndivnicTT, line 49: Can we not go in this direction? In any way, now that macros don't have the same signatures you need #ifdef QUASAR ... to guard arch specific code
So there is no point in renaming

ndivnicTT (commented):
I looked at whole PR, the idea is good and most of the changes are good. I assume there will be some more polishing changes to fix failing tests.
I have a concern regarding macros, I don't think any new macro should be added, and also no _QSR suffix. Macros should only be updated to not have DST_ACCUM_MODE like it is done.

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/llk_math_eltwise_binary_sfpu_macros.h
halghTT: The goal here is to keep the signature the same, with the definition below

bbradelTT (approved):

riverwuTT (approved):
approve for programming example change as it appear to be consistent with other use site updates.

mdragulaTT (approved):
Approving for tt-train

pavlejosipovic (approved):

rtawfik01 (approved):

sjameelTT (approved):

nbabinTT (approved):

llongTT (commented):
Two non-blocking notes on the dest-acc threading, details inline. Neither affects any in-tree caller today.

tt_metal/hw/inc/api/compute/reconfig_data_format.h
```diff
@@ -180,35 +184,39 @@ ALWI void reconfig_data_format(
 /**
  * Reconfigures the srcA data format for a new operand, always re-deriving the int8/unsigned state from the new format.
  */
+template <bool is_fp32_dest_acc_en = DST_ACCUM_MODE>
```
llongTT, line 187: **Non-blocking.** Adding a defaulted `is_fp32_dest_acc_en` here *and* to the deprecated shims makes every explicit template-argument spelling ambiguous:

- `reconfig_data_format_srca<true>(cb)` matches both this primary (`<bool acc>`) and the deprecated `<bool is_tile_dim, bool acc = DST_ACCUM_MODE>` at line 494.
- `reconfig_data_format_srca<true, false>(cb)` matches both deprecated families (`<is_tile_dim, acc>` and `<to_from_int8, is_tile_dim, acc>`).
- Same for `reconfig_data_format<SrcOrder::Reverse, true>(a, b)` (line 155 vs 463) and the `_srcb` / `_skip_int8` forms.

The candidates have identical function parameter lists, so overload resolution cannot prefer one. g++ and clang both report `call of overloaded 'reconfig_data_format_srca<true>(...)' is ambiguous`. Before this PR the primary had no template parameter, so arity picked the shim.

No in-tree caller uses these spellings today, so CI is green. But the deprecated forms cannot be called as documented during their support window (the note at lines 442-448 still describes a disambiguation rule that no longer holds), and the new flag cannot be passed explicitly on these names at all; it is reachable only through the default.

Suggested fix: leave the new parameter off the deprecated shims so arity disambiguates again. The `RECONFIG_DF_DEPRECATED` block further down passed its 2026-08-20 removal date before this PR opened and has no callers, so it could simply be deleted here instead of extended.

tt_metal/hw/inc/api/compute/experimental/sdpa.h
```diff
@@ -475,7 +498,8 @@ template <
     std::uint32_t scale_fp32,
     VectorMode vector_mode = VectorMode::C,
     bool pop_ms = false,
-    bool dense = false>
+    bool dense = false,
+    bool is_fp32_dest_acc_en = DST_ACCUM_MODE>
```
llongTT, line 502: **Non-blocking.** The new `is_fp32_dest_acc_en` is forwarded to `fused_max_sub_exp_add_tile` (line 522) and `sdpa_bcast_col_reuse_preamble` (line 526), but four other dest-width-sensitive calls in these same wrappers still run on the kernel-wide `DST_ACCUM_MODE`:

- `copy_init(cb_worker_ms)` at line 505
- `tile_regs_release()` at lines 539 and 597
- `pack_untilize_dest_init<...>(cb_l_out)` at lines 685-691 (7th template argument left defaulted)
- `pack_untilize_uninit(cb_l_out)` at line 705

All four already accept `is_fp32_dest_acc_en` as a template argument.

If a Blackhole kernel does `set_fp32_dest_acc<true>()` and then `sdpa_tail<..., true>(...)`, which is the use the PR description advertises, the SFPU part computes at one DEST width while the datacopy, dest clear, and packer read use the other, and the defaulted `pack_untilize_dest_init` re-programs the packer read width back to the kernel-wide mode. No in-tree caller passes a non-default value yet, so nothing breaks today.

Suggested fix: either forward `is_fp32_dest_acc_en` to those four calls (`untilize_block` in untilize.h is a fully threaded precedent), or drop the parameter from `sdpa_tail_ms_reduce` / `sdpa_tail_l_block` / `sdpa_tail` until it can be threaded end to end, as was done for `softcap_init` in this PR.

llongTT (approved):

sidnadTT (approved):

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/llk_math_eltwise_unary_sfpu_macros.h
```diff
@@ -32,33 +32,47 @@ inline __attribute__((always_inline)) void _sfpu_check_(std::uint32_t dst_index,
 
 // Macro hygiene: DST_IDX and VECTOR_MODE are evaluated by both the check and
 // params call. Keep call sites to identifiers/literals, not side effects.
-#define SFPU_UNARY_CALL(DST_SYNC, DST_ACCUM, FN, TEMPLATES, DST_IDX, VECTOR_MODE, ...) \
-    (::ckernel::_sfpu_check_<DST_SYNC, DST_ACCUM>(DST_IDX, VECTOR_MODE),               \
-     _llk_math_eltwise_unary_sfpu_params_(                                             \
+// Shared signature matches WH/BH (no DST_ACCUM slot). Inject DST_ACCUM_MODE
+// for dest-tile bounds; Quasar has no runtime dest-acc switch.
+#define SFPU_UNARY_CALL_QSR(DST_SYNC, DST_ACCUM, FN, TEMPLATES, DST_IDX, VECTOR_MODE, ...) \
```
ndivnicTT, line 37: Can you please erase SFPU_UNARY_CALL_QSR

ndivnicTT: Or do it as a follow up for removing redundant DST_ACCUM for quasar

PR #54364, opened 2026-08-25
[Bug Fix] Eltwise Unary: Update output dtypes for preallocated output tensor

KalaivaniMCW:
### PR Category

Bug fix

### Summary

- `unary_impl` doesn't update the `output_dtype` for preallocated output tensor and `output_dtype` remains same as input dtype which is the case for most unary ops

### Notes for reviewers

- This PR updates the output dtype for preallocated output tensor
- Follow up PR will add the necessary tests with ttsim skips

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kalaivani/unary_out)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kalaivani/unary_out)

### DIFF
--- a/ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
@@ -24,6 +24,9 @@ Tensor unary_impl(
         op_chain.back().type() == unary::UnaryOpType::BITCAST) {
         output_dtype = static_cast<DataType>(*op_chain.back().get_param_if<float>(1));
     }
+    if (optional_output_tensor.has_value()) {
+        output_dtype = optional_output_tensor->dtype();
+    }
     bool preserve_fp32_precision = (input_dtype == DataType::FLOAT32);
     bool fp32_dest_acc_en = preserve_fp32_precision || output_dtype == DataType::UINT32 ||
                             output_dtype == DataType::INT32 || output_dtype == DataType::FLOAT32 ||


ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -24,6 +24,9 @@ Tensor unary_impl(
         op_chain.back().type() == unary::UnaryOpType::BITCAST) {
         output_dtype = static_cast<DataType>(*op_chain.back().get_param_if<float>(1));
     }
+    if (optional_output_tensor.has_value()) {
+        output_dtype = optional_output_tensor->dtype();
```
Aswinmcw, line 28: Nothing validates that the preallocated dtype is a legal output dtype for this chain. `bitcast` already `TT_FATAL`s on a mismatch and `ttnn::typecast` has its own `typecast_impl` that never reaches here, so the two branches above are effectively already guarded — which means this block only bites on plain unary chains, where there is no conversion op at all. `ttnn.abs(bf16_in, output_tensor=int32_out)` now silently configures an INT32 output instead of being rejected.

Which `(input dtype, preallocated output dtype)` pairs is this meant to enable? If it's a specific set, worth a `TT_FATAL` for everything outside it rather than trusting whatever the caller allocated.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -24,6 +24,9 @@ Tensor unary_impl(
...
+    }
```
Aswinmcw, line 29: No test for the behaviour being fixed. Worth noting the blast radius when you add one: `output_dtype` feeds `fp32_dest_acc_en`, `bfp8_pack_precise`, *and* `operation_attributes_t::to_hash()`. Before this change, two calls differing only in preallocated output dtype hashed identically, so the second one reused a program compiled for the first dtype — so a `test_unary_program_cache.py` case belongs here alongside the numerical one.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
KalaivaniMCW: this is specific for 16 bit floating dtypes and 32-floating dtype cases

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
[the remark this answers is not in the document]
KalaivaniMCW: this has been tested locally. As the mixed dtype test will fail ttsim tests  , it will be added in a follow-up PR with necessary ttsim skips

KalaivaniMCW: this has been tested locally and verified. 
As the mixed dtype tests will fail ttsim tests , tests will be added in a follow-up PR with necessary ttsim skips

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
Aswinmcw: Nothing in the diff restricts it to those, though — `output_dtype = optional_output_tensor->dtype()` runs unconditionally, for every chain and every dtype, so `ttnn.abs(bf16_in, output_tensor=int32_out)` takes the same path and silently gets an INT32 output config. If the supported set really is {bfloat16, float32}, a `TT_FATAL` on that set makes the intent enforceable rather than implicit, and it also bounds the follow-up test surface.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
Aswinmcw: `tests/pipeline_reorg/ttsim-skip-list.yaml` takes individual test ids (`file.py::test_name[param]`), so a one-line entry per arch would let the test land in this PR instead of trailing it — no need to wait on a follow-up for the skips. Worth doing for the program-cache case at least, since that's the part local testing is least likely to have covered.

Aswinmcw (approved):
LGTM — remaining points are suggestions, not blockers.

PR #54372, opened 2026-08-25
Add pull-request gate for changed CI tests

tdowdallTT:
### PR Category
Feature

### Summary
Adds `.github/workflows/verify-changed-tests.yaml`, a gate that proves any `tests/pipeline_reorg/*.yaml` entry a PR edits still runs green before that PR can merge, instead of relying on manual review to catch a broken or silently-weakened test definition.

All the logic lives in `.github/scripts/utils/verify_changed_tests.py`:
- Diffs the changed yamls at test-entry level (keyed on `name`/`arch`/`gtest_shard_index`, since `name` alone isn't unique) to work out which entries are new or behaviour-affecting. Metadata-only edits (`owner_id`, `team`, per-SKU `timeout`) don't trigger a run — timeouts are already enforced statically by `verify_time_budget.py`.
- Calls `prepare_test_matrix.py` over the changed yamls and narrows its output to exactly the touched legs, so the gate dispatches only the affected work.
- For legs on scarce hardware (`REVIEW_ONLY_SKUS`: galaxy/sc1/sc4/sc16 SKUs), requires an owner-approving review on the head commit instead of auto-running — enforced by CODEOWNERS via the main ruleset's `require_code_owner_review`.
- Fails closed: an unresolvable leg or a matrix shorter than expected fails the gate rather than passing with reduced coverage.
- On `merge_group`, only re-resolves scope (the legs already ran on the PR head); it doesn't re-fire hardware runs.

Adds `.github.scripts.utils.test_verify_changed_tests` covering the diffing/scoping logic.

### Notes for reviewers
- The workflow reports the `Changed tests verified` status check on every PR, but **this PR does not add it to the branch protection ruleset**. See Test Plan for why.
- Draft PRs are skipped (`scope` job is gated on `!github.event.pull_request.draft`); the verdict job short-circuits to a pass for drafts since they can't be queued anyway.

### Test Plan
This is landing as a silent rollout, not a blocking gate, on purpose:
1. Merge this workflow without adding `Changed tests verified` to the branch protection ruleset. It will run on every PR and report status, but nothing is blocked on it yet.
2. Let it run against real PR traffic for a period and confirm it scopes changed tests correctly — right legs dispatched, no false blocks on metadata-only edits, owner-review gating works as expected on scarce SKUs — without any false positives/negatives.
3. Once verified, add `Changed tests verified` as a required status check in the ruleset so failures actually block merges.

- [x] Ran CI and confirmed the gate resolves the expected legs on this branch's own test changes.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:tdowdall/reorg-pre-commit)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:tdowdall/reorg-pre-commit)


.github/workflows/verify-changed-tests.yaml
```diff
@@ -0,0 +1,432 @@
...
+      volumes:
+        - ${{ github.workspace }}/docker-job:/work # Subdir to workaround https://github.com/actions/runner/issues/691
+        - /dev/hugepages-1G:/dev/hugepages-1G
+        # Weights cache, selected by the SKU's weights-cache-mode exactly as the
+        # owning pipeline selects it.
+        - ${{ matrix.leg['weights-cache-mode'] == 'cloud-mlperf' && '/mnt/MLPerf/huggingface:/mnt/MLPerf/huggingface:ro' || (matrix.leg['weights-cache-mode'] == 'yyz4-mnt-models' && '/mnt/models/huggingface:/mnt/MLPerf/huggingface:ro') || (matrix.leg['weights-cache-mode'] == 'local-disk' || matrix.leg['weights-cache-mode'] == 'lfc') && '/localdev/blackhole_demos/huggingface_data:/mnt/MLPerf/huggingface:rw' || '/mnt/MLPerf:/mnt/MLPerf:ro' }}
+      options: >-
+         ${{ (startsWith(matrix.leg.sku, 'sim_') && '--label runner-mode=sim')
+             || (matrix.leg.sku != 'github_hosted_cpu' && '--device /dev/tenstorrent')
+             || '' }}    defaults:
```
roseli-TT, line 251: I think this is a true syntax error (thanks claude). The `defaults:` here should probably be on the next line

.github/workflows/verify-changed-tests.yaml
```diff
@@ -0,0 +1,432 @@
...
+#
+# All logic lives in .github/scripts/utils/verify_changed_tests.py. This file only
+# wires it up, plus the two policy lists below.
+#
+# merge_group runs `scope` only. The hardware legs already ran on the PR head;
+# re-firing them in the queue would spend the same scarce hardware twice. The
+# queue only needs the context to report, which `scope` does. Required checks
+# are evaluated on the merge group too, so the trigger cannot simply be dropped.
+
+on:
```
roseli-TT, line 14: is `on: merge_group` missing intentionally for this first submission?

.github/scripts/utils/verify_changed_tests.py
```diff
@@ -0,0 +1,917 @@
...
+    return selected, sim_libs
+
+
+# ---------------------------------------------------------------------------
+# reviews
+# ---------------------------------------------------------------------------
+
+
+def fetch_reviews(repo, pr, token):
+    url = f"https://api.github.com/repos/{repo}/pulls/{pr}/reviews?per_page=100"
```
roseli-TT, line 702: Please verify this review comment:
```The reviews endpoint returns oldest first, and every approve / request-changes / review-comment submission counts as a review. So on a PR with >100 reviews you get the oldest 100 — and the approval on the current head, which is by construction the newest, is precisely what gets dropped. Result: a false block on a PR that is properly approved. This is a realistic shape for the long-lived galaxy PRs this gate is meant to route to owners.```

roseli-TT:
Renaming a pipeline yaml results in a gate failure. I think the expected behaviour is not to fail in that situation

roseli-TT (approved):

.github/workflows/verify-changed-tests.yaml
tdowdallTT: Yes, removed intentionally to prevent long merge queue runs. There is a potential improvement where we allow short added tests to run on `merge_group`. That said, this tool is intended to verify that new CI has the ability to pass, verification on branch should hopefully be sufficient.

PR #54380, opened 2026-08-26
[skip ci] Fix: restore the `/test` badge column and pin the report heading

blozano-tt:
### PR Category

Bug fix (regression from #54373)

### Summary

On [PR #54351](https://github.com/tenstorrent/tt-metal/pull/54351#issuecomment-5418943518) the `/test` summary table came back with a plain `` `sanity-tests` `` code span where the badge should be — for all six dispatched pipelines — and then printed *"Badges go live within ~1 minute of dispatch"* underneath a table containing no badges.

**Dispatch itself was fine.** All six runs landed on `halgh/sfpu-dest-acc-template-cleanup` with correctly narrowed inputs, and the suite narrowing from #54373 worked exactly as intended (`run-fabric-sanity-tests: false`, `run-t3000-sanity-tests: false`, …). This is purely the report.

### Root cause — self-inflicted in #54373

That PR made the inputs column emphatic and well-argued: ten lines, bolded, with a rationale. The badge survived only as an unbolded parenthetical — *"a status badge (below)"* — at the head of the same bullet.

The badge instruction never got weaker in substance. It got weaker in **prominence**, and the column it described lost out to the one being argued for. The *Status badges* section was still being read — its caveat line came out verbatim — but read as *a caption to reproduce* rather than *a column to build*. That section said "emit one badge per dispatched pipeline" and never said **where** it goes, which is exactly how it degraded into a caption.

### Fix

Rather than just re-bolding the parenthetical (which would leave the same failure one refactor away):

- **States the table's shape directly** — a three-column skeleton with the badge first — instead of describing the columns in prose.
- **Gives the failure a name the agent can catch mid-write:** if it is about to type a bare pipeline name in column 1, it has dropped the badge.
- **The badge template section now opens by saying where its output belongs** ("column 1 of the table above — not a separate section, and not a caption").
- **Makes the "badges go live shortly" caveat conditional** on the table actually containing badges. Unconditional, it is worse than nothing: it tells the reader to wait for images that are never coming — which is precisely how this regression disguised itself as *slow* badges rather than *missing* ones.

Prompt-only; the `.lock.yml` diff is the single `body_hash` line.

### Notes for reviewers

The interesting failure here isn't that an instruction was wrong — it's that two instructions competed for the same slot and the quieter one lost. Worth remembering when adding to this prompt: emphasis is finite, and adding a strongly-argued requirement next to a mildly-worded one can silently demote the latter. That's the reason for the structural fix (a table skeleton the agent fills in) over a wording fix.

I did not touch the narrowing logic — it is working, and the live output on #54351 is a good example of it: seven `sanity-tests` suites cut to the three the diff can actually reach, with the reasoning stated per row.

One small inconsistency I noticed and deliberately left alone: on #54351 the agent set `run-ttsim-sanity-tests: false` but left `run-sim: true` at its default. Those are arguably the same axis. It is low-impact and I would rather see whether it recurs than add another rule on one observation.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:test-command-restore-badges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:test-command-restore-badges)


blozano-tt:
Added a second commit (`6e082bc535`) for two report-formatting fixes raised in review.

**Heading drift.** It rendered `### `/test` — dispatched pipelines` on #54233 and `## `/test` triage` on #54351. The cause was not something overriding the instruction — there was **no heading instruction at all**, so nothing to override. Now pinned verbatim to:

```
### `/test` — dispatched pipelines
```

**Where "triage" came from.** Almost certainly the prompt itself: *"You are a triage agent, not a test author."* Reworded to *"You select and launch existing pipelines; you do not write tests"* — same meaning, without seeding a word this repo already overloads three ways (CI triage, issue triage, and the `run_triage_tests` input). The heading rule also rules the word out explicitly; the verbatim heading is what does the real work, the prohibition just closes the loop.

`run_triage_tests` in the do-not-touch list is a real input name and is deliberately untouched.

**Also confirmed, since it came up in review:** badge images *do* render inside GFM table cells. Rendered the proposed table through GitHub's own `POST /markdown` endpoint — `` wrapped in `<a>` inside `<td>`, correct src and href, slashed branch names intact.

One verified edge case, deliberately **not** fixed here: `git check-ref-format` permits `|` in a branch name (`feat|bar` is legal), and a pipe splits the markdown row, mangling that one cell. No tt-metal branch in a 60-branch sample contains one. Given this PR exists because adding text beside the badge instruction demoted it, I would rather handle it deterministically in the prefetch step (emit a URL-safe branch variant) than spend prompt attention on a pathological case — happy to do that as a follow-up if wanted.

Issue #52141, opened 2026-08-05
[ttnn/eltwise]: bfloat16 add and subtract return 0.0 for Inf - Inf where float32 returns NaN, the operand dtype selects the compute engine


kinginu:
**Component / Area:** ttnn/eltwise/binary_ng

**Issue Type:** Bad Outputs

## Observed

`ttnn.subtract` and `ttnn.add` return `0.0` for `Inf - Inf` when the operands are `bfloat16`, and `nan` when they are `float32`. Same expression, same shapes, only the dtype differs.

Measured on Blackhole P300 at `b2e77906b9`, a clean build with nothing applied. Every tile is filled with the value, and every result is uniform across all 1024 elements:

| expression | IEEE / torch | `ttnn` bfloat16 | `ttnn` float32 |
|---|---|---|---|
| `add(+Inf, -Inf)` | NaN | **`0.0`** | `nan` |
| `subtract(+Inf, +Inf)` | NaN | **`0.0`** | `nan` |
| `multiply(+Inf, 0)` | NaN | **`0.0`** | `nan` |
| `divide(+Inf, +Inf)` | NaN | **`+Inf`** | `nan` |
| `add(+Inf, 1)` | +Inf | `+Inf` | `+Inf` |
| `subtract(+Inf, 1)` | +Inf | `+Inf` | `+Inf` |
| `multiply(+Inf, 2)` | +Inf | `+Inf` | `+Inf` |
| `add(-Inf, -Inf)` | -Inf | `-Inf` | `-Inf` |
| `reciprocal(+Inf)` | +0 | `0.0` | `0.0` |
| `neg(+Inf)` | -Inf | `-Inf` | `-Inf` |
| `clone(+Inf)` | +Inf | `+Inf` | `+Inf` |

The last four rows are the control: everything IEEE defines as finite-or-infinite comes out right in both dtypes. Only the invalid operations differ, and only in bfloat16. `to_layout` on a tile carrying `0x7F80` and `0xFF80` returns both unchanged, so the packer is not involved and this reproduces on plain `main`.

The `multiply` and `divide` rows each have a different identified cause and are not what this issue is about; see Scope.

## Expected

`nan`. IEEE 754 calls `Inf - Inf` an invalid operation whose result is NaN, and `bfloat16` represents NaN perfectly well, so nothing about the type forces this.

[...]

PR #54381, opened 2026-08-26
[Bug fix] packer: use the identity path for identity packs so bfloat16 NaN survives

### LINKED ISSUE #51557: [Bug] bfloat16 loses all 254 NaN encodings through ops that neither convert nor compute
## Describe the bug

A `bfloat16` NaN does not survive a device op, even an op that performs no arithmetic and no format conversion. `ttnn.to_layout`, `ttnn.neg`, `ttnn.abs` and `ttnn.typecast(x, ttnn.bfloat16)` all return an Infinity of the same sign. A host round-trip and `ttnn.clone` preserve the value, so it reaches the device intact and is destroyed on the way back out.

In-tree this is recorded in 15 places as an expected hardware limitation of the dtype, for example `ternary_program_factory.cpp:939`: *"bfloat16 packs NaN as inf as this is an expected behaviour and a known HW limitation of bfloat16 dtype."* That is not what I measure. I have a build in which **all 254 bfloat16 NaN encodings survive these ops bit-exact**, with **all 65024 normal encodings unchanged** and no perf change I can detect. See the note at the end.

## Relationship to the two open issues this subsumes

Both of these are the same defect seen through one op each. I am filing this rather than commenting on either, because the behaviour is not specific to the op in either title, and because the fix makes both of their repros pass.

- #31406, open since 2025-10-29. Its exact repro, all four layout combinations, both dtypes, on Blackhole P300:

  | dtype | direction | unpatched | with my fix |
  |---|---|---|---|
  | float32 | all four combinations | NaN kept | NaN kept |
  | bfloat16 | ROW_MAJOR -> ROW_MAJOR | NaN kept (no conversion happens) | NaN kept |
  | bfloat16 | ROW_MAJOR -> TILE | **`+Inf`** | **NaN kept** |
  | bfloat16 | TILE -> ROW_MAJOR | **`+Inf`** | **NaN kept** |
  | bfloat16 | TILE -> TILE | NaN kept (no conversion happens) | NaN kept |

  Two notes on that issue. Its float32 claim, NaN becoming `131008.0`, no longer reproduces on Blackhole at `66d22c1f3a9`: float32 NaN survives all four combinations, so what remains of #31406 is the bfloat16 case. And #19585, which #31406 was split out of, is closed.
- #40503, open since 2026-03-23. Measured below.

What is new here relative to both: it is not one op, it is every op in the class, including `neg`, `abs` and `typecast(x, bfloat16)`; it is exhaustive over all 65536 bfloat16 encodings rather than one NaN; and the fix exists and is measured.

## To reproduce

```python
import torch, ttnn

d = ttnn.open_device(device_id=0)

def bf16(bits):                      # a 32x32 bfloat16 tile whose first element is `bits`
    r = torch.zeros(1, 1, 32, 32, dtype=torch.int16)
    r[0, 0, 0, 0] = bits - (1 << 16) if bits >= 0x8000 else bits
    return ttnn.from_torch(r.view(torch.bfloat16), dtype=ttnn.bfloat16,
                           layout=ttnn.TILE_LAYOUT, device=d)

def bits(t):
    return int(ttnn.to_torch(t).view(torch.int16).flatten()[0].item()) & 0xFFFF

print(f"round-trip, no op  0x{bits(bf16(0x7FC0)):04X}")                                             # 7FC0
print(f"clone              0x{bits(ttnn.clone(bf16(0x7FC0), dtype=ttnn.bfloat16)):04X}")            # 7FC0
print(f"to_layout ROW_MAJOR 0x{bits(ttnn.to_layout(bf16(0x7FC0), ttnn.ROW_MAJOR_LAYOUT)):04X}")     # 7F80  <-- +Inf
print(f"neg                0x{bits(ttnn.neg(bf16(0x7FC0))):04X}")                                   # FF80  <-- -Inf
print(f"abs                0x{bits(ttnn.abs(bf16(0x7FC0))):04X}")                                   # 7F80  <-- +Inf
print(f"typecast bf16->bf16 0x{bits(ttnn.typecast(bf16(0x7FC0), ttnn.bfloat16)):04X}")               # 7F80  <-- +Inf
```

Measured on Blackhole P300. `0x7FC0` is a quiet NaN; `0x7F80` is `+Inf`. The same holds for every other NaN encoding, e.g. `0x7F81 -> 0x7F80` and `0xFFC0 -> 0xFF80`.

Build a fresh input for each measurement and read each result once: several ttnn ops deallocate their input, so reusing one device tensor across ops reads a freed buffer.

## Exhaustive measurement

All 65536 bfloat16 bit patterns in a single `[1, 1, 256, 256]` tensor, one op applied to one freshly built tensor, on Blackhole P300:

| op | words changed of 65536 | NaN encodings changed | normal encodings changed |
|---|---|---|---|
| host round-trip, no device op | 0 | 0 of 254 | 0 of 65024 |
| `clone` bf16 -> bf16 | 0 | 0 of 254 | 0 of 65024 |
| `to_layout` TILE -> ROW_MAJOR | 509 | **254 of 254**, to `0x7F80` / `0xFF80` | 0 of 65024 |
| `neg` | 65535 (a sign flip changes almost everything, as expected) | **254 of 254**, to `0xFF80` / `0x7F80` | — |

`to_layout` broken down by class, on the unpatched build and on the build with my fix:

| class | count | unpatched | with my fix |
|---|---|---|---|
| normal | 65024 | 0 changed | 0 changed |
| **NaN** | 254 | **all 254 -> ±Inf**, 2 distinct outputs | **0 changed**, 254 distinct outputs, bit-exact |
| denormal | 254 | all -> `0x0000` | all -> `0x0000`, unchanged |
| −0 | 1 | -> `+0` | -> `+0`, unchanged |
| Inf | 2 | 0 changed | 0 changed |

Every normal value is bit-exact in both builds, which is the evidence that preserving the NaN does not disturb ordinary data. `ttnn.neg` likewise returns 254 distinct NaNs with my fix (`0xFF81`, `0xFF82`, ...), i.e. sign flipped and payload preserved, which is correct IEEE negation of a NaN.

## #40503 measured

#40503 carries an AI-generated root-cause guess that points at the SFPU write. That guess is wrong: the value the SFPU writes is fine, and it is lost afterwards, which is why `clone` and the host round-trip keep it while `full_like` does not. Measured with the same helpers as above:

```
ttnn.full_like(bf16 tensor, float("nan"))   unpatched 0x7F80        with my fix 0x7FC0
```

So one change closes both #40503 and the bfloat16 half of #31406, which is all that is left of it.

## In-tree workarounds this explains

15 places in 8 files work around the behaviour, all describing it as expected and none of them correct about why. `test_exp2.py:222-225` is worth singling out: it states that *"ttnn.bfloat16 host-side packing collapses NaN to +inf before the tensor reaches the SFPU"*, and the round-trip line in the repro above shows that is not so — the host path carries all 254 encodings intact.

- `tests/ttnn/unit_tests/operations/eltwise/test_unary.py:461`, `:2546`, `:2658`, `:2662`
- `tests/ttnn/unit_tests/operations/eltwise/test_where.py:407`, `:435`, `:472`
- `tests/ttnn/unit_tests/operations/eltwise/test_div_ops.py:407`, `:445`
- `tests/ttnn/unit_tests/operations/eltwise/test_pow.py:167`
- `tests/ttnn/unit_tests/operations/eltwise/test_expm1.py:42`
- `tests/ttnn/unit_tests/operations/eltwise/test_exp2.py:222-225`
- `tests/ttnn/nightly/unit_tests/operations/eltwise/test_unary_shard.py:208`, `:218`
- `ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp:939`, which describes a
  workaround of reinterpreting the bfloat16 CBs as `UINT16` so the datapath bit-copies instead of
  converting

## Expected behavior

`torch` preserves a NaN through negation and through `float32 -> bfloat16` (`0x7FC00000 -> 0x7FC0`). Independently of torch parity, a NaN silently becoming `+Inf` is worse than an abort: NaN is what a user greps for when debugging, `Inf` compares and propagates differently, and a NaN guard placed after any bfloat16 op will never fire. `bfloat16` represents NaN perfectly well, so describing this as a limitation of the dtype is inaccurate.

## Scope of this report

- Only ops that **neither convert a format nor perform arithmetic** are in scope. A genuine narrowing conversion such as `typecast(float32, bfloat16)` still yields `+Inf` from a NaN and is **not** claimed here.
- **Denormals are not part of this report.** All 254 still become `+0`, and `tech_reports/Handling_Special_Value/special_values.md` documents denormals as flushed globally in TT hardware (`denormals | all | 0x0`), so that is intended.
- **Minus zero is not part of this report.** It still becomes `+0`. One word out of 65536, and I have not isolated where it is lost, so I make no claim about it.
- Measured on **Blackhole** only. I have no Wormhole hardware.

## Environment

- Blackhole P300, device 0
- tt-metal `66d22c1f3a9`; the LLK sources involved are byte-identical to `main` at `b6d87f377ba`
- sfpi 7.67.0

## Note on the fix

I have root-caused this and I have a fix. It is small, it is confined to the LLK, and it is measured on Blackhole P300:

- all 254 NaN encodings preserved bit-exact through the ops above, with sign and payload;
- all 65024 normal encodings bit-identical to the unpatched build;
- #40503 fixed (`full_like(nan)` returns `0x7FC0`) and #31406's repro passing in all eight dtype-by-direction combinations;
- ttnn eltwise regression subset: **1064 passed, 455 skipped, 0 failed**;
- no perf change I can detect: an LLK unary SFPU sweep restricted to `Float16_b`, before and after, moves by well under 1% in both directions, i.e. noise.

I have not opened a PR yet because I would rather confirm the diagnosis with you first and avoid duplicating work someone else may already have in flight. If you would like the change, please assign the issue to me and I will submit it.

kinginu:
## PR Category
Bug fix. Closes #51557. Closes #40503. Closes the bfloat16 case of #31406.

## Summary
A pack whose source and destination formats are the same needs no conversion, but the LLK still configures the packer's rounding path. That path flushes denormals, folds minus zero to plus zero, and, for an eight-bit exponent, maps NaN to infinity. A `bfloat16` tensor therefore cannot carry a NaN through any op, since every op ends in a pack.

The hardware provides an identity path. `tt-isa-documentation` (`WormholeB0/TensixTile/TensixCoprocessor/Packers/FormatConversion.md`, at `b7738d9`) specifies it for a BF16-to-BF16 pack:

> **Rounding (i.e. flush denormals to zero and NaN to infinity):** Set `IntermediateFormat = BF16` and `Read_raw = false`.
> **Identity (i.e. preserve denormals and NaN):** Set `Read_raw = true`, then choose either of: set `IntermediateFormat = BFP8`; set `IntermediateFormat = BF16`.

Both architectures name that bit `PCK_DEST_RD_CTRL_Read_int8`, and the LLK sets it only for int8 output. This PR extends the condition to identity packs, leaving `IntermediateFormat` alone.

Two details in the implementation:

- The condition tests for `Float16_b -> Float16_b` or `Float16 -> Float16`, not for equal formats. `get_single_pack_src_format` reports `pack_src_format == pack_dst_format == Bfp8_b` for a `bfloat8_b` output, and that pack does convert: reading Dst raw there skips the early rounding and produces the `bfp8_pack_precise` result instead of the default one.
- `reconfig_packer_data_format`'s read-modify-write mask now includes `Read_int8`. Without it the bit keeps whatever `set_packer_config` left, so a kernel that initialises with an identity pack and later reconfigures to a narrowing output packs it with `Read_raw` set. `matmul` and `softmax` are such kernels.

## Tests
`test_bfloat16_nan_identity_ops.py`, 25 cases: `to_layout` in both directions, `neg`, `abs`, `typecast(bf16 -> bf16)` and `clone` over three NaN encodings; `full_like` with a NaN fill (#40503); `to_layout` at `(17, 37)` in all four layout directions (#31406's repro); all 65536 bfloat16 encodings through `to_layout`; and a guard asserting that `matmul` with a `bfloat8_b` output does not become bit-identical to the `bfp8_pack_precise` path.

Two LLK tests exercise the pack directly through an A2D datacopy in `Float16_b`.

## Measurements

Blackhole P300 and Wormhole n300. The two architectures give the same figures.

| | base | with this PR |
|---|---|---|
| `typecast(bf16 -> bf16)`, all 65536 encodings | 509 changed | **0 changed** |
| `to_layout`, all 65536 encodings | 509 changed | 255 changed |
| NaN | 254 of 254 lost | 0 of 254 changed, 254 distinct, bit-exact |
| normal | 0 of 65024 | 0 of 65024 |
| denormal / −0 | flushed | preserved through `typecast`, still flushed through `to_layout`, see Scope |
| `full_like(nan)` (#40503) | `0x7F80` | `0x7FC0` |
| #31406's repro, 8 combinations | bf16 `RM->TILE`, `TILE->RM` lose it | 8 of 8 keep it |
| block-float digests, 22 kernels | — | 0 of 22 changed |
| `test_bfloat16_nan_identity_ops.py` | 19 failed | 25 passed |
| ttnn eltwise regression incl. integer paths | — | 1548 passed, 9 skipped, 0 failed |
| LLK suite (Blackhole) | — | 9565 passed, 1246 skipped, 0 failed |

`typecast(bf16 -> bf16)` becomes an exact identity: 0 of 65536 encodings change, where 509 changed before. `to_layout` still changes 255, the 254 denormals and `0x8000 -> 0x0000`, for a reason that is upstream of the pack and described under Scope.

`#31406`'s float32 claim, NaN becoming `131008.0`, does not reproduce on either architecture on the base commit.

## Performance

LLK unary SFPU perf suite, identity-pack cases with `dest_acc=No`, paired per test, comparing the largest profiling window per test.

| | cases | mean | worst | best |
|---|---|---|---|---|
| Blackhole P300 | 48 | +0.001% | +0.012% | −0.003% |
| Wormhole n300 | 206 | −0.003% | +0.234% | −0.365% |

Each build was measured twice on each architecture. The repeat runs of a single binary are 96% bit-identical on Blackhole and 79% on Wormhole, and the second base-versus-PR comparison reproduces the mean, worst and best above. So the per-test differences are deterministic rather than noise, and they are centred on zero with the sign split across cases. The change is computed at packer configure time; the per-tile path gains no instruction.

## Scope
Not fixed: genuine narrowing conversions such as `typecast(float32, bfloat16)`; `bfloat16` with `fp32_dest_acc_en`, which the guard excludes; `0x8000 -> 0x0000` and denormal flushing in the layout-changing paths, see below; and arithmetic ops, where `ttnn.add` loses a bfloat16 NaN for an unrelated reason (upstream #52141).

## Denormals and minus zero: a wider change than the NaN fix alone

This needs stating plainly because it is the part most likely to be objected to.

The identity path is a bitcast, so it preserves everything the rounding path was destroying, not only NaN. Measured: `typecast(bf16 -> bf16)` changes 0 of 65536 encodings with this PR and 509 without, and `0x8000` and the 254 denormals now survive `typecast`, `clone`, `reshape`, `to_memory_config` and `neg(+0)`. The ISA text says the same thing: *"Identity: Denormals preserved. NaN preserved."*

That conflicts with `tech_reports/Handling_Special_Value/special_values.md`, which states denormals are flushed globally (`denormals | all | 0x0`). After this PR that is no longer true for identity packs. **The two cannot be separated with this bit**: selecting the identity path is all-or-nothing, so there is no version of this change that keeps NaN and still flushes denormals. If the global flushing statement has to hold, this change cannot be taken as it stands, and the honest answer would be to update the document instead.

The layout-changing ops, `to_layout` in either direction and `transpose`, still flush both. That is not this pack: `MOVA2D` replaces any `SrcA` datum whose exponent field is zero before the pack sees it (`MOVA2D.md:71`), which catches denormals and `0x8000` but not NaN. Filed separately as  and unaffected by this PR.

Quasar is not touched and has not been checked.

### DIFF
--- a/tests/ttnn/unit_tests/base_functionality/test_bfloat16_nan_identity_ops.py
+++ b/tests/ttnn/unit_tests/base_functionality/test_bfloat16_nan_identity_ops.py
@@ -1,0 +1,154 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""A bfloat16 NaN must survive an op that neither converts a format nor computes anything.
+
+Before the packer selects the identity path for same-format packs, every op in this file returns an
+Infinity of the same sign instead, which is issue #31406 for `to_layout` and #40503 for `full_like`.
+Genuine narrowing conversions such as `typecast(float32, bfloat16)` are out of scope: their early
+format conversion has to round, and rounding maps NaN to infinity for an 8-bit exponent.
+
+Denormals and minus zero are deliberately not asserted here. Both are still flushed, and
+`tech_reports/Handling_Special_Value/special_values.md` documents denormals as flushed globally.
+"""
+
+import pytest
+import torch
+
+import ttnn
+from models.common.utility_functions import run_for_blackhole
+
+# Three of the 254 bfloat16 NaN encodings: the canonical quiet NaN, the smallest payload, and a
+# negative one, so that a sign-preserving bug and a payload-dropping bug are both visible.
+NAN_BITS = [0x7FC0, 0x7F81, 0xFFC0]
+
+
+def bf16_tile(bits, device, shape=(1, 1, 32, 32)):
+    """A fresh device tensor whose first element holds `bits` and whose remainder is +0."""
+    raw = torch.zeros(shape, dtype=torch.int16)
+    raw[0, 0, 0, 0] = bits - (1 << 16) if bits >= 0x8000 else bits
+    return ttnn.from_torch(raw.view(torch.bfloat16), dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+
+def first_bits(tensor):
+    return int(ttnn.to_torch(tensor).view(torch.int16).flatten()[0].item()) & 0xFFFF
+
+
+def classify(bits):
+    exponent, mantissa = bits & 0x7F80, bits & 0x007F
+    if exponent == 0x7F80:
+        return "NaN" if mantissa else "Inf"
+    if exponent == 0:
+        return "zero" if mantissa == 0 else "denormal"
+    return "normal"
+
+
+@run_for_blackhole()
+@pytest.mark.parametrize("nan_bits", NAN_BITS, ids=lambda b: f"0x{b:04X}")
+@pytest.mark.parametrize(
+    "op_name",
+    ["to_layout_row_major", "to_layout_tile", "neg", "abs", "typecast_bfloat16", "clone"],
+)
+def test_identity_op_preserves_nan(device, op_name, nan_bits):
+    """Each op is given a tensor of its own, because several ttnn ops deallocate their input."""
+    ops = {
+        "to_layout_row_major": lambda t: ttnn.to_layout(t, ttnn.ROW_MAJOR_LAYOUT),
+        "to_layout_tile": lambda t: ttnn.to_layout(ttnn.to_layout(t, ttnn.ROW_MAJOR_LAYOUT), ttnn.TILE_LAYOUT),
+        "neg": ttnn.neg,
+        "abs": ttnn.abs,
+        "typecast_bfloat16": lambda t: ttnn.typecast(t, ttnn.bfloat16),
+        "clone": lambda t: ttnn.clone(t, dtype=ttnn.bfloat16),
+    }
+    got = first_bits(ops[op_name](bf16_tile(nan_bits, device)))
+
+    if op_name == "abs":
+        # The payload has to survive, and it is the payload this is about. The sign bit of a NaN is
+        # left alone by abs on this hardware, which is a separate question from the pack path.
+        assert classify(got) == "NaN" and (got & 0x7FFF) == (
+            nan_bits & 0x7FFF
+        ), f"abs(0x{nan_bits:04X}) returned 0x{got:04X}"
+        return
+
+    expected = nan_bits ^ 0x8000 if op_name == "neg" else nan_bits
+    assert got == expected, f"{op_name}(0x{nan_bits:04X}) returned 0x{got:04X}, expected 0x{expected:04X}"
+
+
+@run_for_blackhole()
+def test_full_like_nan_fill_stays_nan(device):
+    """Issue #40503: the fill value reaches the SFPU intact and is lost on the way out."""
+    reference = ttnn.from_torch(
+        torch.zeros(1, 1, 32, 32, dtype=torch.bfloat16), dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device
+    )
+    got = first_bits(ttnn.full_like(reference, float("nan")))
+    assert classify(got) == "NaN", f"full_like(nan) returned 0x{got:04X}"
+
+
+@run_for_blackhole()
+@pytest.mark.parametrize("input_layout", [ttnn.ROW_MAJOR_LAYOUT, ttnn.TILE_LAYOUT])
+@pytest.mark.parametrize("output_layout", [ttnn.ROW_MAJOR_LAYOUT, ttnn.TILE_LAYOUT])
+def test_to_layout_nan_not_mangled(device, input_layout, output_layout):
+    """Issue #31406, with its own shape, which is deliberately not a multiple of the tile size."""
+    torch_input = torch.full((17, 37), 0.05, dtype=torch.bfloat16)
+    torch_input[0, 0] = float("nan")
+    tensor = ttnn.from_torch(torch_input, device=device, layout=input_layout)
+    result = ttnn.to_torch(ttnn.to_layout(tensor, layout=output_layout))
+    assert torch.isnan(result[0, 0]), f"NaN became {float(result[0, 0])}"
+
+
+@run_for_blackhole()
+def test_every_bfloat16_encoding_through_to_layout(device):
+    """All 65536 encodings in one tensor: no NaN may change, and no normal value may change."""
+    raw = torch.arange(65536, dtype=torch.int32) & 0xFFFF
+    raw = torch.where(raw >= 0x8000, raw - (1 << 16), raw).to(torch.int16).reshape(1, 1, 256, 256)
+    tensor = ttnn.from_torch(raw.view(torch.bfloat16), dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    out = ttnn.to_torch(ttnn.to_layout(tensor, ttnn.ROW_MAJOR_LAYOUT)).view(torch.int16).flatten()
+    source = raw.flatten()
+
+    changed = {"NaN": [], "normal": []}
+    for src, got in zip(source.tolist(), out.tolist()):
+        src, got = src & 0xFFFF, got & 0xFFFF
+        kind = classify(src)
+        if kind in changed and src != got:
+            changed[kind].append((src, got))
+
+    assert not changed["normal"], f"{len(changed['normal'])} normal encodings changed, e.g. {changed['normal'][:4]}"
+    assert not changed["NaN"], f"{len(changed['NaN'])} of 254 NaN encodings changed, e.g. " + ", ".join(
+        f"0x{s:04X}->0x{g:04X}" for s, g in changed["NaN"][:4]
+    )
+
+
+@run_for_blackhole()
+def test_narrowing_pack_is_unaffected(device):
+    """Selecting the identity pack path must not leak into a pack that does convert.
+
+    The packer's read-raw bit lives in a register that `set_packer_config` writes whole and
+    `reconfig_packer_data_format` writes through a mask, so a kernel that initialises the packer for
+    an identity pack and later reconfigures it to a block-float output can carry the bit into that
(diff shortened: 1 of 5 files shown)


vtsilytskyiTT:
@kinginu please address AI comments and take a look at red sanity tests.

kinginu:
Done, and rebased onto main.

Coverage went up along the way: three bfloat16 NaN parameters in `test_where.py` and `rdiv(0, 0)` in `test_unary.py` were xfailed for this defect and now run. `test_remainder_nan` and `test_fmod_nan` keep their xfail, since they still return an infinity, so whatever loses the NaN there is upstream of the pack.

The red clang-tidy was a 120 minute timeout at 4693 of 4747 files with zero violations.

vtsilytskyiTT:
I rerun sanity tests. Something is wrong. Nodes are going down all the time:
```
  [gw0] node down: Not properly terminated
  [gw0] FAILED 
```
Also there are test failures like this https://github.com/tenstorrent/tt-metal/actions/runs/35251260282/job/105315566979?pr=54381
```
  [ RUN      ] MeshDeviceFixture.TensixTestSingleCoreMultiTileBankedL1ReaderDataCopyL1Writer
  2026-09-17 17:49:26.377 | info     |            Test | Running test using Slow Dispatch (mesh_dispatch_fixture.hpp:177)
  [1597290] ERROR: UnimplementedFunctionality: tensix_pacr: bf16 to bf16: read_raw=1
```

Issue #14222, opened 2024-10-24
Layernorm gives garbage output for sufficiently large tensors


yieldthought:
**Describe the bug**
If the tensor is large enough that subblock_w is >8, Layernorm produces garbage output. We have an assert that it should be <=4 for FP32 accumulation but not that it should be <= 8 otherwise.

**To Reproduce**
Run a large sharded multicore layernorm, e.g. 1,1,32,8192 over 16 cores with a program config specifiying subblock_w and block_w both 16. Observe that the layernorm output has 0 PCC comapred to torch and that the values are not even in comparable ranges.

**Expected behavior**
Assert: subblock width must less than 8

PR #54388, opened 2026-08-26
[Bug fix] softmax: validate the sharded subblock_w against the Dest capacity

### LINKED ISSUE #52040: [Bounty $1,500] ttnn/normalization: sharded softmax accepts a subblock_w larger than Dest and silently returns wrong results
**Component / Area:** ttnn/normalization/softmax

**Issue Type:** Bad Outputs

## Observed

On sentence_bert's own shape and program config, with `subblock_w` raised from 6 to 12, sharded softmax returns rows summing to as much as **47.42** where a softmax must sum to 1. There is no error, no warning and no log line: the op returns a tensor of the right shape and dtype containing garbage. The only bound the op applies to `subblock_w` is `block_w % subblock_w == 0`, and `block_w` is 12 here, so 12 passes it.

The repro below is that measurement. The tables here are the same sweep run one config per process, with a torch PCC alongside the row sums, at `(1, 1, 36864, 384)` `bfloat16`: same shard shape, same 6x8 grid, same `block_h` 24 and `block_w` 12, same per-core work. Default compute config, Blackhole P300, `992f3ca634`:

| subblock_w | row sum mean | min | max | PCC vs torch | |
|---|---|---|---|---|---|
| 1 | 0.999102 | 0.9779 | 1.0192 | 0.999968 | ok |
| 2 | 0.999102 | 0.9779 | 1.0192 | 0.999968 | ok |
| 3 | 0.999102 | 0.9779 | 1.0192 | 0.999968 | ok |
| 4 | 0.999102 | 0.9779 | 1.0192 | 0.999968 | ok |
| 6 | 0.999102 | 0.9779 | 1.0192 | 0.999968 | ok |
| 12 | 6.402125 | 0.8213 | 47.6707 | **0.018846** | wrong |

The threshold moves with the compute config: `subblock_w` tiles are live in Dest at once, and Dest holds 8 tiles by default, 4 with `fp32_dest_acc_en` and 16 with `dst_full_sync_en`. So the same `subblock_w` can be correct under one compute config and garbage under another. `fp32_dest_acc_en=True`, everything else unchanged:

| subblock_w | row sum mean | min | max | PCC vs torch | |
|---|---|---|---|---|---|
| 1 | 0.999166 | 0.9968 | 1.0001 | 0.999997 | ok |
| 2 | 0.999166 | 0.9968 | 1.0001 | 0.999997 | ok |
| 3 | 0.999166 | 0.9968 | 1.0001 | 0.999997 | ok |
| 4 | 0.999166 | 0.9968 | 1.0001 | 0.999997 | ok |
| 6 | 5.043384 | **-1526.7466** | 838.0242 | -0.030736 | wrong, and correct at 16 bits |
| 12 | 7.276325 | 1.9015 | 47.5277 | 0.002739 | wrong |

`block_w = 16` puts a `subblock_w` on both sides of the 8-tile and 16-tile capacities. `dst_full_sync_en` moves the limit the other way:

| subblock_w | dst_full_sync_en=False | dst_full_sync_en=True |
|---|---|---|
| 1 | 0.997798, PCC 0.999962 | 0.997798, PCC 0.999962 |
| 2 | 0.997798, PCC 0.999962 | 0.997798, PCC 0.999962 |
| 4 | 0.997798, PCC 0.999962 | 0.997798, PCC 0.999962 |
| 8 | 0.997798, PCC 0.999962 | 0.997798, PCC 0.999962 |
| 16 | 5.168172, PCC 0.018633, wrong | **0.997798, PCC 0.999962, correct** |

Ruled out:

* Not `numeric_stable`. With `numeric_stable=False`, `subblock_w` 1 to 6 give 0.999680 and `subblock_w` 12 gives row sums from -46.4514 to 49.8968, PCC 0.004103. Both paths fail at the same `subblock_w`.
* Not bfloat16 precision. The same input at `subblock_w` 6 sums to 0.999102 with PCC 0.999968. At 12 the PCC is 0.018846, so the output is uncorrelated with the answer rather than imprecise.
* Not a malformed shard config. `block_h` and `block_w` match the shard shape in tiles, which the op has validated since #37329, and `block_w % subblock_w == 0` holds for every row above.
* Not program cache reuse. Every row above ran in its own process against a freshly opened device.

## Expected

An out-of-range `subblock_w` should raise. The sibling normalization op already does exactly this: `ttnn/cpp/ttnn/operations/normalization/layernorm/device/sharded_layernorm_factory_helpers.cpp:23`, `assert_subblock_compute_config_compatible(dst_full_sync_en, fp32_dest_acc_en, subblock_wt)`, added for #14222, "Make layernorm assert instead of silently giving garbage output when subblock_w too large". Sharded softmax has the same parameter and the same failure and never got the same check.

What should not change: every `subblock_w` at or below the capacity keeps working, including all six in-tree call sites below, two of which sit exactly on the limit. The check has to read the compute config rather than assume 8, or `subblock_w = 16` under `dst_full_sync_en` starts raising even though it is correct today.

## Steps to reproduce

```python
import torch, ttnn

d = ttnn.open_device(device_id=0, l1_small_size=79104)
x = torch.randn((1, 96, 384, 384), dtype=torch.float32).to(torch.bfloat16)
mem = ttnn.MemoryConfig(
    memory_layout=ttnn.TensorMemoryLayout.HEIGHT_SHARDED, buffer_type=ttnn.BufferType.L1,
    shard_spec=ttnn.ShardSpec(
        ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(5, 7))}),
        [24 * 32, 384], ttnn.ShardOrientation.ROW_MAJOR))

for sw in (1, 2, 3, 4, 6, 12):
    pc = ttnn.SoftmaxShardedMultiCoreProgramConfig(
        compute_with_storage_grid_size=(6, 8), subblock_w=sw, block_h=24, block_w=12)
    t = ttnn.from_torch(x, device=d, layout=ttnn.Layout.TILE, memory_config=mem)
    o = ttnn.to_torch(ttnn.from_device(
        ttnn.softmax_in_place(t, program_config=pc, numeric_stable=True))).float().sum(-1)
    print(f"subblock_w={sw:2d}  row sum mean {float(o.mean()):.4f}  min {float(o.min()):.4f}  max {float(o.max()):.4f}")
ttnn.close_device(d)
```

Blackhole P300, `992f3ca634`, nothing applied:

```
subblock_w= 1  row sum mean 0.9930  min 0.9521  max 1.0181
subblock_w= 2  row sum mean 0.9930  min 0.9521  max 1.0181
subblock_w= 3  row sum mean 0.9930  min 0.9521  max 1.0181
subblock_w= 4  row sum mean 0.9930  min 0.9521  max 1.0181
subblock_w= 6  row sum mean 0.9930  min 0.9521  max 1.0181
subblock_w=12  row sum mean 6.4043  min 0.7300  max 47.4248
```

The reading needs no reference implementation: a softmax's rows sum to 1 by definition, so the op reports its own error. The row sums differ in the last digits from the first table because the logical shape is split differently across the same shards, and the third table uses `(1, 1, 6144, 512)` so that `block_w` can be 16.

## Affected call sites

Six in tree at `992f3ca634`, two of them shipped model code and four of them tests. None passes a `subblock_w` above the capacity, so this is a footgun with no current victims rather than a live wrong answer in a shipped model. Two sit exactly on the limit, which is what forces the bound to be `<=` rather than `<`.

| call site | subblock_w | capacity | |
|---|---|---|---|
| `models/demos/blackhole/sentence_bert/ttnn/common.py:70` | 6 | 8 | ok at 16 bits, and wrong under `fp32_dest_acc_en` |
| `models/demos/falcon7b_common/tt/falcon_attention.py:409` | 8 | 8 | on the limit |

The tests are the same story: `tests/ttnn/nightly/unit_tests/operations/fused/test_softmax_sharded.py:72,164,262,344`, `tests/ttnn/unit_tests/operations/fused/test_softmax.py:215`, and `tests/tt_eager/python_api_testing/sweep_tests/pytests/tt_dnn/test_softmax_sharded.py:112`. None exceeds the capacity and none tests above it, which is how this survived. What they do show is that the cap is folklore every caller re-derives by hand, including the `fp32_dest_acc_en` halving that softmax applies by default whenever the input is `float32` (`softmax_device_operation.cpp:386`):

```python
    # calculate the max subblock size < 8
    upper_limit = min(block_w // 2, 8)
```

## Trade-offs a fix has to answer

The capacity depends on the compute config, so a hardcoded `<= 8` would reject `subblock_w = 16` under `dst_full_sync_en`, measured correct above, and would accept `subblock_w = 6` under `fp32_dest_acc_en`, measured wrong above. `ttnn::get_dest_reg_count` already derives the number from both flags and from the tile shape.

Cost on the dispatch path is zero: the values are ones the op already has, and `validate_on_program_cache_miss` runs once per program cache miss rather than per invocation.

Nothing in tree starts raising. Every configuration measured above that is correct today returns bit-identical numbers with the check in place.

## Environment information

* OS: Ubuntu 22.04
* Version of software: tt-metal `992f3ca634`, clean build, nothing applied
* Hardware: Blackhole P300, device 0

Not tested on Wormhole: no hardware available.

kinginu:
## PR Category

Bug fix.

## Summary

Closes #52040.

`SoftmaxShardedMultiCoreProgramConfig` lets the caller choose `subblock_w`, and nothing validated it against the Dest capacity. An oversized value ran to completion and returned a tensor of the right shape and dtype containing garbage, with no exception and no warning.

`softmax_program_factory_attention_optimized_sharded.cpp:197` passes `subblock_w` to the compute kernel as a compile-time argument, and the kernel indexes Dest with it directly (`kernels/attention/compute/softmax_sharded.cpp:57`):

```cpp
        for (uint32_t w = 0; w < subblock_w; w++) {
            pack_tile(w, dfb_out_id);
        }
```

so tiles `0 .. subblock_w-1` have to exist in Dest. The only bound the op applied was `block_w % subblock_w == 0` (`softmax_device_operation.cpp:240`), and that one sits inside the mask-present branch. That placement is also why `subblock_w = 0` belongs in the same check: with no mask the divisibility check never runs, so a zero reaches `num_subblocks_w = block_w / subblock_w` in the sharded factory and divides by zero — a SIGFPE that takes the process down with a core dump rather than an exception.

The bound comes from `ttnn::get_dest_reg_count`, not a constant, because the capacity depends on the compute config: 8 tiles by default, 4 with `fp32_dest_acc_en`, 16 with `dst_full_sync_en`. A hardcoded `<= 8` would reject `subblock_w = 16` under `dst_full_sync_en`, which runs correctly today. Layernorm spells the same table by hand in `sharded_layernorm_factory_helpers.cpp:23`, added for #14222 after the identical failure in that op; this calls the shared helper instead of adding a third copy.

## Accuracy

What the guard catches, measured on N300 at `141fbea49c5`, shape `[8, 4, 128, 512]`, height sharded on an 8x4 grid, bfloat16. Each row of a softmax must sum to 1:

| compute config | Dest capacity | `subblock_w` | row sum, min .. max | PCC vs torch |
|---|---|---|---|---|
| default | 8 | 8 | 0.9471 .. 1.0181 | 0.999860 |
| default | 8 | **16** | 0.8597 .. **159.7318** | **0.012314** |
| `fp32_dest_acc_en` | 4 | 4 | 0.9611 .. 1.0008 | 0.999815 |
| `fp32_dest_acc_en` | 4 | **8** | 2.6942 .. 3.9061 | **-0.063760** |
| `dst_full_sync_en` | 16 | 16 | 0.9471 .. 1.0181 | 0.999860 |

The two over-capacity rows are the defect. Nothing is raised for either. The `fp32_dest_acc_en` row is the one that shows the capacity is not a constant: `subblock_w = 8` is correct under the default config and wrong under this one, and its output is not merely inaccurate but anti-correlated with the answer.

What must not move: with the guard in place, all three in-capacity rows return **the same numbers to every printed digit**, and the two over-capacity rows raise instead:

```
TT_FATAL @ softmax_device_operation.cpp:342: program_config.subblock_w > 0 &&
program_config.subblock_w <= dest_tile_capacity
info: subblock_w (16) must be greater than 0 and at most the Dest capacity of 8 tiles for this
compute config (fp32_dest_acc_en=false, dst_full_sync_en=false).
```

## Cost

Host-side only. One `TT_FATAL` in `validate_on_program_cache_miss`, in the same `std::visit` block as the `block_h` and `block_w` checks from #37329. It runs once per program cache miss, before any program is built, so there is no per-call cost and the dispatch count is unchanged.

Reading `attributes.compute_kernel_config` there is safe: it is resolved by `softmax_init_compute_kernel_config` before it is stored (`softmax_device_operation.cpp:401`), and `get_compute_kernel_config_args`, which the program factory uses to configure the kernel, is a passthrough of the same fields `get_dest_reg_count` reads. The check and the kernel cannot disagree.

## Tests

`test_softmax_sharded_subblock_w_dest_capacity`, 6 cases: both bounds against all three capacities. On the base commit the two over-capacity cases fail; the `subblock_w = 0` case cannot be run there at all, since it aborts the interpreter rather than failing.

Nothing that works today starts raising:

| | P300 | N300 |
|---|---|---|
| `tests/ttnn/unit_tests/operations/fused/`, whole directory | 3380 passed, 540 skipped, 796 xfailed, 0 failed | not run |
| `tests/ttnn/nightly/.../test_softmax_sharded.py` | 13 passed | not run |
| `models/demos/blackhole/sentence_bert/tests/pcc/` | 9 passed | not run |
| the five configurations in the Accuracy table | — | 3 of 3 in-capacity identical, 2 of 2 over-capacity raise |

No in-tree caller exceeds the capacity, and the two that sit exactly on it — falcon7b at 8 (`falcon_attention.py:409`) and `test_softmax_sharded.py:72` at 8 — still pass.

## Not covered

- The interleaved factory's `fp32_dest_acc_en ? 4 : 8` and layernorm's four-branch `assert_subblock_compute_config_compatible` are both hand-rolled copies of what `get_dest_reg_count` computes, and neither accounts for `dst_full_sync_en` the same way. Consolidating them is a cleanup across two ops and would put a refactor in a bug-fix PR.
- The message names `fp32_dest_acc_en` and `dst_full_sync_en` rather than telling the caller which value to use, because the largest legal `subblock_w` also has to divide `block_w`, and picking it for them would need the divisor search the eager sweep does by hand.
- The regression suites above were run on P300 only. The N300 evidence is the five configurations in the Accuracy table, not a suite run.

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `141fbea49c5`
* Hardware: N300 and P300

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1218171627388572

### DIFF
--- a/tests/ttnn/unit_tests/operations/fused/test_softmax.py
+++ b/tests/ttnn/unit_tests/operations/fused/test_softmax.py
@@ -8,7 +8,7 @@ import torch
 import torch.nn.functional as F
 
 import ttnn
-from tests.ttnn.utils_for_testing import assert_numeric_metrics
+from tests.ttnn.utils_for_testing import assert_numeric_metrics, assert_with_pcc
 from models.common.utility_functions import torch_random
 
 TEST_PADDING_VALUE = -42
@@ -292,6 +292,68 @@ def test_softmax_sharded_stable_with_program_cache(
             memory_config=ttnn.L1_MEMORY_CONFIG,
         )
     assert device.cache_entries_counter.total == 1
+
+
+# The sharded compute kernel keeps subblock_w tiles live in Dest, so a subblock_w above the Dest
+# capacity used to run to completion and return wrong numbers with no diagnostic. The capacity is
+# 8 tiles by default, 4 with fp32_dest_acc_en and 16 with dst_full_sync_en, so the guard has to read
+# the compute config rather than assume the default. subblock_w = 0 is the other end of the same
+# bound: block_w % subblock_w == 0 is only checked when a mask is present, so without one a zero
+# used to reach block_w / subblock_w in the program factory and take the process down with SIGFPE.
+@pytest.mark.parametrize(
+    "fp32_acc_en, dst_full_sync_en, subblock_w, expect_raise",
+    [
+        (False, False, 0, True),
+        (False, False, 8, False),
+        (False, False, 16, True),
+        (True, False, 4, False),
+        (True, False, 8, True),
+        (False, True, 16, False),
+    ],
+)
+def test_softmax_sharded_subblock_w_dest_capacity(
+    device, expect_error, fp32_acc_en, dst_full_sync_en, subblock_w, expect_raise
+):
+    torch.manual_seed(0)
+    grid_size = (8, 4)
+    batch_size, num_heads, h, w = 8, 4, 128, 512
+
+    torch_input_tensor = torch_random((batch_size, num_heads, h, w), -10, 10, dtype=torch.bfloat16)
+    memory_config = ttnn.create_sharded_memory_config(
+        torch_input_tensor.shape,
+        core_grid=ttnn.CoreGrid(y=grid_size[1], x=grid_size[0]),
+        strategy=ttnn.ShardStrategy.HEIGHT,
+        orientation=ttnn.ShardOrientation.ROW_MAJOR,
+    )
+    program_config = ttnn.SoftmaxShardedMultiCoreProgramConfig(
+        compute_with_storage_grid_size=grid_size,
+        subblock_w=subblock_w,
+        block_h=batch_size * num_heads * h // 32 // (grid_size[0] * grid_size[1]),
+        block_w=w // 32,
+    )
+    compute_kernel_config = ttnn.init_device_compute_kernel_config(
+        device.arch(),
+        math_fidelity=ttnn.MathFidelity.HiFi4,
+        fp32_dest_acc_en=fp32_acc_en,
+        packer_l1_acc=False,
+        dst_full_sync_en=dst_full_sync_en,
+    )
+    input_tensor = ttnn.from_torch(
+        torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=memory_config
+    )
+
+    if expect_raise:
+        with expect_error(RuntimeError, "must be greater than 0 and at most the Dest capacity"):
+            ttnn.softmax_in_place(
+                input_tensor, program_config=program_config, compute_kernel_config=compute_kernel_config
+            )
+        return
+
+    output_tensor = ttnn.softmax_in_place(
+        input_tensor, program_config=program_config, compute_kernel_config=compute_kernel_config
+    )
+    torch_output_tensor = F.softmax(torch_input_tensor.float(), dim=-1)
+    assert_with_pcc(torch_output_tensor, ttnn.to_torch(output_tensor).float(), 0.999)
 
 
 @pytest.mark.parametrize("batch_size", [1, 16])
--- a/tests/ttnn/unit_tests/operations/fused/test_softmax.py
+++ b/tests/ttnn/unit_tests/operations/fused/test_softmax.py
@@ -343,7 +343,10 @@ def test_softmax_sharded_subblock_w_dest_capacity(
     )
 
     if expect_raise:
-        with expect_error(RuntimeError, "must be greater than 0 and at most the Dest capacity"):
+        # The bound is enforced by two checks now: the zero end runs ahead of the mask branch, which
+        # carries a block_w % subblock_w, and the capacity end needs the compute config.
+        message = "subblock_w must be greater than 0" if subblock_w == 0 else "must be at most the Dest capacity"
+        with expect_error(RuntimeError, message):
             ttnn.softmax_in_place(
                 input_tensor, program_config=program_config, compute_kernel_config=compute_kernel_config
             )
@@ -354,6 +357,38 @@ def test_softmax_sharded_subblock_w_dest_capacity(
     )
     torch_output_tensor = F.softmax(torch_input_tensor.float(), dim=-1)
     assert_with_pcc(torch_output_tensor, ttnn.to_torch(output_tensor).float(), 0.999)
+
+
+# subblock_w = 0 with a mask present used to reach `block_w % subblock_w` in the mask branch of
+# validate_on_program_cache_miss, which is a modulo by zero, before any check on subblock_w itself
+# ran. The zero check is hoisted ahead of both branches, so this path raises like the unmasked one.
+def test_softmax_sharded_masked_subblock_w_zero(device, expect_error):
+    torch.manual_seed(0)
+    grid_size = (8, 4)
+    batch_size, num_heads, h, w = 8, 4, 128, 512
+
+    torch_input_tensor = torch_random((batch_size, num_heads, h, w), -10, 10, dtype=torch.bfloat16)
+    attention_mask = torch.zeros(batch_size, 1, 1, w, dtype=torch.bfloat16)
+    attention_mask_t = ttnn.from_torch(attention_mask, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+
+    memory_config = ttnn.create_sharded_memory_config(
+        torch_input_tensor.shape,
+        core_grid=ttnn.CoreGrid(y=grid_size[1], x=grid_size[0]),
+        strategy=ttnn.ShardStrategy.HEIGHT,
+        orientation=ttnn.ShardOrientation.ROW_MAJOR,
+    )
+    program_config = ttnn.SoftmaxShardedMultiCoreProgramConfig(
+        compute_with_storage_grid_size=grid_size,
+        subblock_w=0,
+        block_h=batch_size * num_heads * h // 32 // (grid_size[0] * grid_size[1]),
+        block_w=w // 32,
+    )
+    input_tensor = ttnn.from_torch(
+        torch_input_tensor, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device, memory_config=memory_config
+    )
+
+    with expect_error(RuntimeError, "subblock_w must be greater than 0"):
+        ttnn.scale_mask_softmax_in_place(input_tensor, 1.0, attention_mask_t, program_config=program_config)
 
 
 @pytest.mark.parametrize("batch_size", [1, 16])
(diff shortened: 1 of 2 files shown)


jberkowitzTT:
@kinginu - One change is needed before approval: the new subblock_w > 0 guard comes after the existing block_w % subblock_w check in the mask-present path. With a masked sharded softmax and subblock_w=0, it can still hit modulo-by-zero before the guard. Please move the zero check before any divisibility check and add a masked subblock_w=0 regression case. The DEST-capacity validation and current tests otherwise look good.

kinginu:
Done. Zero check hoisted ahead of both branches, masked subblock_w=0 case added.

jberkowitzTT:
@kinginu - I'll tag in the code owners, but from a bounty perspective, this looks good! Thanks for the great work and I'll close out the bounty issue with the standard payment instructions

gajanan-choudhary (approved):
LGTM!

gajanan-choudhary:
### Code review

Found 1 issue:

1. The new Dest-capacity `TT_FATAL` will hard-fail an existing in-tree caller, contradicting the PR body's "no in-tree caller exceeds the capacity."

`models/demos/bge_large_en` builds the sharded softmax config with `subblock_w=seqL_t` and calls `attention_softmax_` without a `compute_kernel_config`, so the default capacity of 8 applies. Its tests run seq_len 384 (`seqL_t=12`) and 512 (`seqL_t=16`) — both exceed 8, so the new check aborts where the op previously returned (wrong, but non-fatal) results.

https://github.com/tenstorrent/tt-metal/blob/a0630078ff96b52a1fa2f24066143f08119c8043/models/demos/bge_large_en/ttnn/ttnn_bge_self_attention.py#L110-L123

https://github.com/tenstorrent/tt-metal/blob/a0630078ff96b52a1fa2f24066143f08119c8043/ttnn/cpp/ttnn/operations/normalization/softmax/device/softmax_device_operation.cpp#L344-L357

Turning silent corruption into a hard error is the right direction, but these BGE configs need a fix (smaller `subblock_w`, or `dst_full_sync_en=True`) landed with this PR, and the "no in-tree caller exceeds the capacity" claim in the description should be corrected. Note BGE model tests don't appear to be in the pre-merge gate, so this may only surface later (nightly/model suites).

Two out-of-scope follow-ups (not blocking): the general `block_w % subblock_w == 0` check is still only reached on the non-sharded-mask path, so an unmasked sharded call can still silently truncate `num_subblocks_w = block_w / subblock_w` in the program factory; and layernorm's sharded validate has the identical unguarded pattern.

PR #54389, opened 2026-08-26
[Bug fix] clamp: use an infinity, not FLT_MAX, for an omitted bound

### LINKED ISSUE #51498: ttnn.clamp: omitting a bound substitutes a finite sentinel, so `clamp(x, min=0)` turns `+inf` into 3.4028235e38
# ttnn.clamp: omitting a bound substitutes a finite sentinel, so `clamp(x, min=0)` turns `+inf` into 3.4028235e38

`ttnn::clamp` has no way to tell the kernel "this bound does not exist", so the host invents one. For
float it uses `±FLT_MAX`; for int32 it uses `±16775716`. The kernel then applies that sentinel as a
real bound. The tensor-bound overload in the same function does not have this problem.

**Scope.** Directly observable on `float32` tensors. In `bfloat16` the `FLT_MAX` sentinel
(`0x7F7FFFFF`) sits exactly one round-to-nearest-even step below `+inf`: its low half is `0xFFFF`, so
packing rounds it up to `0x7F80` = `+inf` and the clip becomes invisible. Whether that happens
depends on the pack rounding mode, which I could not verify without a device — cf.
[#25815](https://github.com/tenstorrent/tt-metal/issues/25815) ("Make pow function round outputs
to-nearest bfloat16 instead of truncating it"), which shows the mode is not uniform across the tree.
If the pack truncates rather than rounds, `bfloat16` is affected too.

The two overloads of the same public op disagree on the same request, and both are in the same file:

- `clamp(Tensor, ScalarVariant min, nullopt)` →
  [`unary_composite_op.cpp:213-215`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp#L213-L215)
  → `clamp_tss(a, min_val, FLT_MAX)` → `calculate_clamp` applies **both** bounds
- `clamp(Tensor, Tensor min, nullopt)` →
  [`unary_composite_op.cpp:226-232`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp#L226-L232)
  → `where(ge(a, min), a, min)` → no upper bound at all

```
x = +inf, float32

ttnn.clamp(x, min=0.0)                    ->  3.4028234663852886e+38
ttnn.clamp(x, min=torch.zeros_like(x))    ->  inf
torch.clamp(x, min=0.0)                   ->  inf
```

`torch.clamp(input, min=m)` is documented as `torch.maximum(input, m)` — there is no upper bound to
apply. This is the behaviour [#13527](https://github.com/tenstorrent/tt-metal/issues/13527) and
[#13512](https://github.com/tenstorrent/tt-metal/issues/13512) ("Update `ttnn.clamp` logic to match
PyTorch API", "`min`, `max` value should have default value as `None`") set out to deliver; the
sentinel is the part of that work that does not match.

**The claim below involves no NaN**, so nothing in it depends on how the SFPU orders NaN in
`SFPSWAP`: `|inf|` is `0x7F800000` and `|FLT_MAX|` is `0x7F7FFFFF`, so sign-magnitude order and real
order agree on every operand used here. (`clamp` does also mishandle NaN, for an unrelated reason in
the kernel rather than the host — that is [#51470](https://github.com/tenstorrent/tt-metal/issues/51470),
and the interaction is spelled out under *Known limits*.)

### Summary

[`unary_composite_op.cpp:185-216`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp#L185-L216):

```cpp
    if (has_int32_scalar) {
        // All scalars are int32_t (or null)
        std::int32_t min_val = min.has_value() ? std::get<std::int32_t>(min.value()) : -16775716;
        std::int32_t max_val =
            max.has_value() ? std::get<std::int32_t>(max.value())
                            : 16775716;  // max_val and min_val will be updated once unary infra supports int32 scalar.
        return ttnn::clamp_tss(a, min_val, max_val, output_mem_config, output_tensor);
    }  // All scalars are float (or null)
    float min_val = min.has_value() ? std::get<float>(min.value()) : std::numeric_limits<float>::lowest();
    float max_val = max.has_value() ? std::get<float>(max.value()) : std::numeric_limits<float>::max();
    return ttnn::clamp_tss(a, min_val, max_val, output_mem_config, output_tensor);
```

`clamp_tss` lowers through
[`unary_op_utils.cpp:652-668`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_op_utils.cpp#L652-L668)
to `clamp_tile` / `clamp_tile_int32`, and the kernel has no "bound absent" branch — it runs a max
body and then a min body, unconditionally:

```c
// ckernel_sfpu_clamp.h — calculate_clamp
        load_value_param_float(min_val);
        calculate_unary_max_min_float_body<Max>();
        load_value_param_float(max_val);
        calculate_unary_max_min_float_body<Min>();
```

So `max = None` is not "no upper bound", it is `max = 3.4028235e38`.

**Why this has survived.** Two things hide it at once.

1. No clamp test feeds a non-finite value.
   [`test_composite.py::test_unary_composite_clamp_ttnn`](https://github.com/tenstorrent/tt-metal/blob/main/tests/ttnn/unit_tests/operations/eltwise/test_composite.py)
   does exercise the single-bound path — its matrix contains `(-10, None)`, `(None, 10)`,
   `(-1.0, None)`, `(None, 1.0)`, `(-0.5, None)`, `(None, -0.5)` — but the input comes from
   `data_gen_with_range(input_shapes, -100, 100, device)`, which is `torch.rand(...).bfloat16()`
   scaled into `[-100, 100]`. The comparison is `compare_pcc`.
2. The matrix also contains `("tensor", None)` and `(None, "tensor")`, i.e. the *correct* overload,
   checked by the same PCC assertion on the same finite data. Both overloads pass, so nothing signals
   that they diverge.

There is also a rounding coincidence worth noting, because it means the defect would stay quiet even
if bf16 inputs did include `inf`: `FLT_MAX` is `0x7F7FFFFF`, whose discarded low half is `0xFFFF`, so
round-to-nearest-even to bfloat16 carries into `0x7F80` = `+inf`. In bfloat16 the sentinel is exactly
one rounding step below infinity. In float32 there is no such step.

### Measured

Method note: these are host-side evaluations of the exact source path above — the two sentinel
constants, then the kernel's max-then-min sequence in float32. No device is involved and none is
needed: every row is ordinary arithmetic on `±inf`, finite values, or exactly-representable integers,
with no NaN and no rounding in dispute. `torch.clamp(x, min=m) == torch.maximum(x, m)` /
`torch.clamp(x, max=M) == torch.minimum(x, M)` is the reference.

**float32, one bound omitted:**

| call | ttnn | torch | error |
|---|---:|---:|---|
| `ttnn.clamp(+inf, min=0.0)` | 3.4028234663852886e+38 | `inf` | absolute error `inf`; a finite value is returned for an infinite one |
| `ttnn.clamp(-inf, max=1.0)` | -3.4028234663852886e+38 | `-inf` | absolute error `inf` |
| `ttnn.clamp(+inf, max=1.0)` | 1.0 | 1.0 | — |
| `ttnn.clamp(3.4e38, min=0.0)` | 3.3999999521443642e+38 | 3.3999999521443642e+38 | — |
| `ttnn.clamp(7.5, min=0.0)` | 7.5 | 7.5 | — |

**Same call, tensor bound instead of scalar bound** (`unary_composite_op.cpp:226-232`):

| call | ttnn | torch |
|---|---:|---:|
| `ttnn.clamp(+inf, min=0.0)` (scalar) | **3.4028234663852886e+38** | `inf` |
| `ttnn.clamp(+inf, min=zeros_like(x))` (tensor) | `inf` | `inf` |

Same op, same mathematical request, two answers, decided by the type of the bound.

**int32, one bound omitted** — supporting evidence rather than the main point, since the
`16775716` line carries its own `TODO`:

| call | ttnn | torch | relative error |
|---|---:|---:|---:|
| `ttnn.clamp(2147483647, min=0)` | 16775716 | 2147483647 | 99.22% |
| `ttnn.clamp(2000000000, min=0)` | 16775716 | 2000000000 | 99.16% |
| `ttnn.clamp(16777216, min=0)` | 16775716 | 16777216 | 0.0089% |
| `ttnn.clamp(16775717, min=0)` | 16775716 | 16775717 | — (first clipped value) |
| `ttnn.clamp(-2147483648, max=0)` | -16775716 | -2147483648 | 99.22% |

The sentinel leaves **0.78%** of the int32 range unclipped; 2,130,707,931 positive int32 values are
silently capped. The int32 clamp test
([`test_composite.py::test_unary_composite_clamp_int_ttnn`](https://github.com/tenstorrent/tt-metal/blob/main/tests/ttnn/unit_tests/operations/eltwise/test_composite.py))
draws inputs from `torch.randint(-100, 100, ...)`, so it never reaches the sentinel.

### Root cause

`clamp_tss` takes two mandatory bounds and the kernel applies both. `std::optional` is resolved at the
host boundary by picking a value that is *supposed* to be a no-op, and neither choice is one:

- `std::numeric_limits<float>::max()` is not the identity for `min()` — `+inf` is larger.
- `16775716` is not the identity for int32 `min()` — most of the int32 range is larger. It is also not
  `INT32_MAX`, nor `2^24`; it is 1500 below `2^24`.

The int32 value is constrained by the plumbing rather than chosen: `unary_op_utils.cpp:652-660` reads
the scalar out of a `std::vector<float>` and casts it back
(`static_cast<uint32_t>(static_cast<int32_t>(params[0]))`), so any sentinel must survive a float32
round trip. That is what the `"will be updated once unary infra supports int32 scalar"` comment is
about. `INT32_MAX` does not survive it (`static_cast<float>(2147483647) == 2147483648.0f`), which is
presumably why a small value was chosen.

### Files

- `ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp` — `clamp`, lines 185-216
  (scalar overload, defect) and 218-251 (tensor overload, correct)
- `ttnn/cpp/ttnn/operations/eltwise/unary/common/unary_op_utils.cpp` — `CLAMP_TSS`, lines 652-668
- `tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_clamp.h` (WH)
- `tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_clamp.h` (BH, logic-identical)

The kernels need no change for the float fix.

### Proposed fix

**float32 — one word per line, no kernel change.** The identity element for `min` is `+inf`, not
`FLT_MAX`:

```cpp
    float min_val = min.has_value() ? std::get<float>(min.value())
                                    : -std::numeric_limits<float>::infinity();
    float max_val = max.has_value() ? std::get<float>(max.value())
                                    :  std::numeric_limits<float>::infinity();
```

`min(x, +inf) == x` and `max(x, -inf) == x` for every finite value and for both infinities, under
either real order or the SFPU's sign-magnitude order (`|inf|` is the largest non-NaN magnitude). It
restores `+inf` and `-inf` and changes nothing else.

**int32 — a strictly better sentinel that still round-trips through float32.** `2147483520`
(= `2^31 - 128` = `16777215 * 128`) is exactly representable in float32, and so is `-2147483520`:

```cpp
        std::int32_t min_val = min.has_value() ? std::get<std::int32_t>(min.value()) : -2147483520;
        std::int32_t max_val = max.has_value() ? std::get<std::int32_t>(max.value()) : 2147483520;
```

That shrinks the silently-clipped band from 2,130,707,931 values to 254 (127 at each end), with no
infra work.

I deliberately do **not** propose `std::numeric_limits<std::int32_t>::min()` for the lower sentinel,
even though `-2^31` is exactly representable in float32. It arrives at the kernel as the bit pattern
`0x80000000`, which `calculate_unary_max_min_int32_body` routes through `load_value_param_int`'s
`~value` inversion plus a `TTI_SFPNOT`/`SFPSWAP` pair precisely because `SFPSWAP` wants
sign-magnitude — the same `INT32_MIN` handling class that
[#20852](https://github.com/tenstorrent/tt-metal/issues/20852) is still open about. Using
`-2147483520` avoids that pattern entirely. If a device check shows `INT32_MIN` is safe here, the
band closes to 127. Closing the last 127 needs the int32-scalar path the `TODO` refers to, or a
`CLAMP_MIN` / `CLAMP_MAX` unary op that applies a single bound — which is the structurally right fix
for both dtypes and would also let the scalar and tensor overloads share one definition.

**Known limits of the proposed fix, stated up front:**

- The int32 half is a narrowing, not an elimination: 127 values at the top of the range remain
  clipped until an int32 scalar can be passed without a float32 round trip.
- **The float half moves NaN from one wrong answer to a different wrong answer.** Today
  `clamp(NaN, min=m)` compares `NaN` against the `FLT_MAX` sentinel and yields `FLT_MAX`; with the
  sentinel changed to `+inf` the same comparison yields `+inf`. Neither equals `torch`'s `NaN`.

  This is worth stating plainly rather than burying, but it is **not a regression introduced here**:
  `clamp` already fails to propagate NaN for reasons that have nothing to do with the sentinel. The
  cause is in the kernel — `clamp` is two bare `SFPSWAP` instructions, and a compare-and-exchange
  cannot preserve NaN — and it is reported separately as
  [#51470](https://github.com/tenstorrent/tt-metal/issues/51470), which covers `maximum`, `minimum`,
  `clamp` and `hardtanh` together. Fixing that one makes NaN correct here too, in either sentinel.

  **The two reports do not overlap**: #51470 changes the kernel and is about NaN; this one changes a
  host-side constant and is about `±inf` and the int32 range. Neither fix depends on the other, and
  applying both in either order gives the same end state.
- I have not measured this on silicon. The float change is a host-side constant substitution and the
  kernel path is unchanged, so the risk surface is a compile-time literal; but it should be confirmed
  on both WH and BH before merging.

### Success criteria

- [ ] `ttnn.clamp(x, min=m)` with `max` omitted equals `torch.maximum(x, m)` on float32, including
      `x = ±inf`
- [ ] `ttnn.clamp(x, max=M)` with `min` omitted equals `torch.minimum(x, M)` on float32, including
      `x = ±inf`
- [ ] the scalar-bound and tensor-bound overloads of `ttnn.clamp` agree on finite inputs and `±inf`:
      `ttnn.clamp(x, min=0.0) == ttnn.clamp(x, min=ttnn.zeros_like(x))`
      (NaN inputs are deliberately excluded here — the two overloads disagree on NaN both before and
      after this change, for a reason that lives in the kernel rather than in the host, and that is
      the subject of [#51470](https://github.com/tenstorrent/tt-metal/issues/51470))
- [ ] int32: `ttnn.clamp(x, min=0)` leaves every value in `[0, 2147483520]` unchanged
- [ ] the existing clamp/clip test matrices gain a float32 case with `±inf` in the input and an exact
      comparison — `compare_pcc` on bfloat16 in `[-100, 100]` cannot see any of this
- [ ] int32 clamp inputs extended past `torch.randint(-100, 100, ...)` to include `INT32_MAX`,
      `INT32_MIN` and values near `2^24`
- [ ] confirm on Blackhole

### Why this matters

`clamp(x, min=...)` with no upper bound is the ordinary spelling of a lower-bounded rectifier, and
`clamp(x, max=...)` with no lower bound is the ordinary spelling of a ceiling. Neither is asking for
the other bound, and PyTorch's contract says neither gets one. Today both get one, silently, and the
value substituted for `inf` is finite, plausible-looking, and the largest number that exists — so
`isinf` downstream flips to false, an overflow that a model was relying on to stay visible becomes an
enormous finite number, and `1 / clamp(x, min=0)` returns a subnormal instead of `0`.

The clearest statement of the problem is that the same op gives two answers depending on whether the
bound arrived as a Python float or as a tensor, and the two implementations are adjacent overloads in
one file — `clamp` at line 185 and `clamp` at line 218.

---

*Method: host-side evaluation of the source path above (numpy `float32`; kernel `max`-then-`min` body
transcribed from `ckernel_sfpu_clamp.h`, with sign-magnitude ordering per `ckernel_sfpu_unary_max_min.h`);
I do not have a device. No row uses NaN, so no assumption about the SFPU's NaN ordering enters. The
reference is `torch.clamp`'s documented equivalence to `torch.maximum` / `torch.minimum` when one bound
is `None`. The float32 round-trip claims about the int32 sentinel were recomputed directly from IEEE-754
binary32.*

I can prepare a PR with the host-side change and a pytest covering the special-value cases above, if this is accepted.

kinginu:
## PR Category

Bug fix.

## Summary

Closes #51498.

`torch.clamp(x, min=m)` is documented equivalent to `torch.maximum(x, m)` — there is no upper bound. The scalar-bound overload of `ttnn::clamp` (`unary_composite_op.cpp`) substituted `FLT_MAX` for the omitted side and passed it into `clamp_tss`, so an infinity came back as the largest finite float. The tensor-bound overload in the same function uses `where(ge(a, min), a, min)` and never had the problem, so **the two overloads of one public op disagreed on the same input**.

The substitution becomes the identity it was meant to be:

```
std::numeric_limits<float>::lowest() -> -std::numeric_limits<float>::infinity()
std::numeric_limits<float>::max()    -> +std::numeric_limits<float>::infinity()
```

`±inf` compares correctly against every finite operand under the kernel's sign-magnitude ordering — `|inf| = 0x7F800000` sits above every finite `|v|` — so the kernel path is untouched.

## Accuracy

N300 at `141fbea49c5`, one tile, against `torch.clamp`:

| | input | bound | main | this branch | torch |
|---|---|---|---|---|---|
| float32 | `+inf` | `min=0` | **3.4028234663852886e+38** | **`inf`** | `inf` |
| float32 | `-inf` | `max=0` | **-3.4028234663852886e+38** | **`-inf`** | `-inf` |
| bfloat16 | `+inf` | `min=0` | **3.3895313892515355e+38** | **`inf`** | `inf` |
| bfloat16 | `-inf` | `max=0` | **-3.3895313892515355e+38** | **`-inf`** | `-inf` |

What must not move does not. The bound that is supplied was never affected — `clamp(+inf, max=0)` returns `0.0` on both branches in both dtypes — and a finite input near the top of the range is unchanged: `clamp(3e38, min=0)` returns `3.0000000054977558e+38` on both branches in float32.

bfloat16 was already `+inf` at the sentinel in some paths, since `FLT_MAX` rounds up to `+inf` when the packer rounds to nearest; the rows above show it was not in this one.

## Cost

Host-side only: two words in the scalar-bound overload. No kernel change, no dispatch change, no per-call cost.

## Tests

`test_clamp_open_bound.py` asserts three things: that `clamp(±inf, min/max=m)` with the other bound omitted returns `±inf`, matching `torch.clamp`; that `clamp(FLT_MAX, min=0)` passes through unchanged rather than being clipped at value; and that the scalar-bound and tensor-bound overloads produce identical output on the same input, `±inf` included.

## Not covered

- **The int32 scalar path.** The `±16775716` sentinel at `:207-210` is left alone; extending it to `INT32_MIN`/`INT32_MAX` needs the integer `SFPSWAP` verified over the full range, which is a kernel-side question and not a host-side substitution.
- **NaN.** `clamp` loses NaN through the kernel's two `SFPSWAP` instructions whatever the sentinel is. That is a separate defect with its own upstream report; it is not touched here.

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `141fbea49c5`
* Hardware: N300 and P300

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/test_clamp_open_bound.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_clamp_open_bound.py
@@ -1,0 +1,71 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+"""Regression tests for ttnn.clamp with an omitted bound.
+
+`torch.clamp(x, min=m)` is documented as `torch.maximum(x, m)` — there is no upper bound.
+Before this fix the host substituted `FLT_MAX` for the omitted side, so `clamp(+inf, min=0)`
+returned `3.4028235e+38` where torch returns `+inf`. This asserts the fixed contract on
+float32 for the three inputs that reveal the sentinel: `+inf`, `-inf`, and `FLT_MAX` itself.
+"""
+
+import pytest
+import torch
+
+import ttnn
+
+
+def _tt(x, device):
+    return ttnn.from_torch(x, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+
+
+def _one(x, device):
+    padded = torch.zeros((1, 1, 32, 32), dtype=torch.float32)
+    padded.view(-1)[0] = x
+    tt = ttnn.from_torch(padded, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+    return tt
+
+
+@pytest.mark.parametrize(
+    "x, min_v, max_v, expected",
+    [
+        (float("inf"), 0.0, None, float("inf")),
+        (float("inf"), -1.0, None, float("inf")),
+        (float("-inf"), None, 0.0, float("-inf")),
+        (float("-inf"), None, 1.0, float("-inf")),
+        # FLT_MAX is the old sentinel; with the omitted upper bound now +inf,
+        # clamp(FLT_MAX, min=0) should pass FLT_MAX through unchanged.
+        (torch.finfo(torch.float32).max, 0.0, None, torch.finfo(torch.float32).max),
+    ],
+)
+def test_clamp_open_bound_float32(device, x, min_v, max_v, expected):
+    """clamp with one bound omitted must not clip ±inf or FLT_MAX."""
+    got_scalar = ttnn.to_torch(ttnn.clamp(_one(x, device), min=min_v, max=max_v)).view(-1)[0].item()
+    assert got_scalar == expected, f"clamp({x}, min={min_v}, max={max_v}) = {got_scalar}, expected {expected}"
+
+
+def test_clamp_open_bound_matches_tensor_overload(device):
+    """The scalar-bound and tensor-bound overloads of ttnn.clamp must agree on finite input and ±inf."""
+    x = torch.tensor([[1.0, -2.0, float("inf"), float("-inf"), 3.0]] + [[0.0] * 5] * 31 * 32, dtype=torch.float32).view(
+        1, 1, 32, 32
+    )[:, :, :1, :5]
+    # Pad to a full 32x32 tile of zeros with the interesting row at [0, 0, 0, :5]
+    padded = torch.zeros((1, 1, 32, 32), dtype=torch.float32)
+    padded[0, 0, 0, :5] = x.view(-1)[:5]
+
+    tx = ttnn.from_torch(padded, dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+    zero_tensor = ttnn.from_torch(torch.zeros_like(padded), dtype=ttnn.float32, layout=ttnn.TILE_LAYOUT, device=device)
+
+    got_scalar = ttnn.to_torch(ttnn.clamp(tx, min=0.0))
+    got_tensor = ttnn.to_torch(ttnn.clamp(tx, min=zero_tensor))
+    ref = torch.clamp(padded, min=0.0)
+
+    # Both overloads must equal torch, exactly, on the five special values in row 0.
+    for i, v in enumerate(padded[0, 0, 0, :5].tolist()):
+        assert (
+            got_scalar[0, 0, 0, i].item() == ref[0, 0, 0, i].item()
+        ), f"scalar overload at {v}: {got_scalar[0, 0, 0, i].item()} != torch {ref[0, 0, 0, i].item()}"
+        assert (
+            got_tensor[0, 0, 0, i].item() == ref[0, 0, 0, i].item()
+        ), f"tensor overload at {v}: {got_tensor[0, 0, 0, i].item()} != torch {ref[0, 0, 0, i].item()}"
--- a/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp
@@ -210,8 +210,15 @@ Tensor clamp(
                             : 16775716;  // max_val and min_val will be updated once unary infra supports int32 scalar.
         return ttnn::clamp_tss(a, min_val, max_val, output_mem_config, output_tensor);
     }  // All scalars are float (or null)
-    float min_val = min.has_value() ? std::get<float>(min.value()) : std::numeric_limits<float>::lowest();
-    float max_val = max.has_value() ? std::get<float>(max.value()) : std::numeric_limits<float>::max();
+    // Omitted bounds must not be substituted with a finite sentinel: on float32 that turns
+    // `clamp(+inf, min=m)` into `FLT_MAX` (3.4028235e38) where `torch.clamp` returns `+inf`.
+    // ±inf compares correctly against every finite operand under the kernel's sign-magnitude
+    // ordering (|inf| = 0x7F800000 is above every finite |v|), so passing ±inf preserves
+    // torch's contract: `clamp(x, min=m)` is `maximum(x, m)` and `clamp(x, max=M)` is
+    // `minimum(x, M)` when the other bound is None. NaN is not addressed here — the kernel's
+    // two SFPSWAP instructions lose it in either sentinel, tracked separately in #51470.
+    float min_val = min.has_value() ? std::get<float>(min.value()) : -std::numeric_limits<float>::infinity();
+    float max_val = max.has_value() ? std::get<float>(max.value()) : std::numeric_limits<float>::infinity();
     return ttnn::clamp_tss(a, min_val, max_val, output_mem_config, output_tensor);
 }
 


mateusznowakTT:
@kinginu : One of the added tests fail, please address it.

PR #54392, opened 2026-08-26
[Metal 2.0] Per-width buffer sets for the untilize/untilize_with_unpadding block factories

abdulla-TT:
### PR Category
Cleanup (Metal 2.0 prereq)

### Summary
Relates to #51305. Same fix as #54140, now applied to the other two ops in that issue.

The `untilize` and `untilize_with_unpadding` `_multi_core_block[_interleaved]` factories had the
same problem the tilize ones did: up to four `(input, output)` CB pairs, all reusing `c_0`/`c_16`,
on disjoint core ranges, at two different sizes. Named DFBs can't size a buffer per node, so this
blocks the Metal 2.0 port.

And the size isn't only a perf knob here either, just for the opposite reason. In tilize it's the
reader that walks a whole block as raw contiguous L1; here it's the writer. It takes
`get_read_ptr()` once per block and then walks the block's rows forward from there, so if a buffer
isn't an exact multiple of the block being drained, the walk runs off the end into whatever is
next instead of wrapping.

The work split only ever hands out two block widths, so each factory now declares two buffer sets
instead of four colliding pairs: `full` (`c_0`/`c_16`, sized `single_sub_block_size`) and
`cliffrow` (`c_2`/`c_17`, sized `single_block_size_cliff_row`). Each non-empty set gets its own
reader/writer pair over its own cores. Compute stays split per region, since each region has its
own block *count*, but each instance binds the set matching its block *width*. Which set a core
belongs to is read from the work split's own core assignment rather than re-derived from the
runtime-arg branching, and a couple of TT_FATALs assert that a set's buffers are sized for exactly
the width its cores are fed. L1 usage is unchanged, because the two sets sit on disjoint cores.

`untilize_wh.cpp`, `reader_unary_interleaved_wh_multicore.cpp` and
`writer_unary_stick_layout_wh_multicore.cpp` now take their buffer indices as compile-time args
instead of hardcoding `c_0`/`c_16`. All three are used only by these two factories (the quasar
ports have their own copies), and both call sites are updated.

`BlockBufferSet::staging_index` becomes optional. The staging buffer is a reader-side
DRAM-alignment scratchpad that only the tilize direction needs, so `push_buffer_set` emits it only
for a set that declares one. That's the one-line change you'll see in each tilize factory.

### Notes for reviewers
This stacks on #54140, since it uses `BlockBufferSet` / `push_buffer_set` / `union_of` and the
shared planner from it. The top two commits are mine — `#51305: untilize & untilize_with_unpadding
fix` and the follow-up that shares the planner; the three below them are #54140's, replayed onto
main and patch-identical to that branch. They disappear once it merges.

Unlike tilize, neither of these factories defines `override_runtime_arguments`, so
`resolve_bindings` walks every kernel's bindings and the second reader/writer pair repoints its
addresses on a cache hit by itself. None of the hook plumbing #54140 needed applies here.

Both untilize factories originally kept a local `BlockPlan` / `make_block_plan`. That is gone: all
four block factories now go through the shared `make_block_plan(BlockDirection, BlockCoreOrder, …)`
in `data_movement/common`, and there is exactly one definition of each left in the tree.

Two parameters carry what legitimately differs. `BlockDirection` covers staging (only tilize's
reader touches DRAM row by row, so only it gets `c_1`/`c_3` and only it reserves staging room in
the L1 budget) and the shape the split is driven from (tilize the output padded shape, untilize the
input — they differ for unpadding). `BlockCoreOrder` is the one worth a look: the two
`split_blocks_for_tilize_wh` overloads assign cores in different orders, since `corerange_to_cores`
defaults to `row_wise=false`, and plain untilize's runtime-arg loop walks
`grid_to_cores(…, row_wise=true)`. Each factory passes the order its own loop already uses, so no
core's args move; a `TT_FATAL` rejects `RowMajor` together with `sub_core_grids`, which the
`CoreCoord` overload cannot honour.

The only behavioural change from the consolidation is untilize's cliffrow input index moving
`c_1` → `c_2`, adopting the shared index table so there is one rule rather than one per direction.

`unpack_to_dest_mode` is now built inside each compute kernel from that instance's own
`set.input_index`, rather than once with both sets' indices marked. The old form set fp32 unpack on
`cliffrow_set.input_index` even when that set was empty — an operand with no CB on any core — and
handed the same vector to the full-set kernels. No behaviour change, but it matches the
`empty()`-guarded treatment every other use of a set gets.

Unrelated bug I hit while picking test shapes, now filed as #54524:
`untilize_with_unpadding` hangs when both the height and width are unpadded on a padded
`(1, 1, 64, 8192)` input. It reproduces on main, so it is not from this change; the shape is simply
kept out of the tests here.

### Testing
New `test_untilize_block_per_node_cb_size` and `test_untilize_with_unpadding_block_per_node_cb_size`,
following the ones in #54140: widths indivisible by any plausible block size so both sets are live
at once, one single-width control, and each shape run twice to cover the program-cache hit.

All four suites on WH, no failures:

- `test_tilize` — 208 passed
- `test_untilize` — 832 passed
- `test_tilize_with_val_padding` — 103 passed
- `test_untilize_with_unpadding` — 357 passed

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/untilize_2.0)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/untilize_2.0)


ttnn/cpp/ttnn/operations/data_movement/untilize/device/factories/untilize_multi_core_block_program_factory.cpp
```diff
@@ -211,43 +174,55 @@ ProgramDescriptor UntilizeMultiCoreBlockProgramFactory::create_descriptor(
     }
     std::vector<UnpackToDestMode> unpack_to_dest_mode(NUM_CIRCULAR_BUFFERS, UnpackToDestMode::Default);
     if (fp32_dest_acc_en) {
-        unpack_to_dest_mode[tt::CBIndex::c_0] = UnpackToDestMode::UnpackToDestFp32;
+        unpack_to_dest_mode[full_set.input_index] = UnpackToDestMode::UnpackToDestFp32;
+        unpack_to_dest_mode[cliffrow_set.input_index] = UnpackToDestMode::UnpackToDestFp32;
```
Aswinmcw: When `cliffrow_set` is empty its `input_index` is still `c_2`, so this marks fp32 unpack on an operand that has no CB on any core — and it does so in the vector handed to the full-set compute kernels too. Harmless today, but every other use of a set in this file is `empty()`-guarded. Building the vector inside `make_compute_kernel` from that instance's `set.input_index` would make each compute kernel declare fp32 unpack for exactly the buffer it reads.

ttnn/cpp/ttnn/operations/data_movement/untilize_with_unpadding/device/factories/untilize_with_unpadding_multi_core_block_interleaved_program_factory.cpp
```diff
@@ -216,37 +182,50 @@ tt::tt_metal::ProgramDescriptor UntilizeWithUnpaddingMultiCoreBlockInterleavedPr
     std::vector<tt::tt_metal::UnpackToDestMode> unpack_to_dest_mode(
         NUM_CIRCULAR_BUFFERS, tt::tt_metal::UnpackToDestMode::Default);
     if (fp32_dest_acc_en) {
-        unpack_to_dest_mode[tt::CBIndex::c_0] = tt::tt_metal::UnpackToDestMode::UnpackToDestFp32;
+        unpack_to_dest_mode[full_set.input_index] = tt::tt_metal::UnpackToDestMode::UnpackToDestFp32;
+        unpack_to_dest_mode[cliffrow_set.input_index] = tt::tt_metal::UnpackToDestMode::UnpackToDestFp32;
```
Aswinmcw: Same as the untilize factory: set unconditionally, so an empty `cliffrow_set` still marks `c_2`.

Aswinmcw (commented):
Description has drifted from the head commit in three places — worth a refresh before this leaves draft:

- "`cliffrow` (`c_1`/`c_17`, sized `single_block_size_cliff_row`)" — `make_block_plan` assigns `c_2`/`c_17`. `c_1`/`c_3` are the tilize *staging* indices, so `c_1` would be a real collision if the body were accurate, and it sent me looking for one.
- The "Still open" paragraph (and both of Copilot's inline asks, which were made against `9a2017038`) are already answered by `787a9304f`: neither factory keeps a local `BlockPlan`/`make_block_plan` any more, both go through the shared `make_block_plan(BlockDirection, BlockCoreOrder, …)`. Answering it anyway: folding it in here was right. I diffed both untilize paths against `3a435afac` and the split is bit-identical — same `a.padded_shape()` source, same `max_l1_size / (input_tile + output_tile)` limit once staging drops out of the budget, same four regions mapping onto the same two widths — so there's no behaviour delta hiding in the parameterisation.
- "Only the top commit is mine" — there are two now.

`787a9304f` is also titled "update comments", but it's the commit that moved the planner into `common.cpp` and added `BlockDirection`/`BlockCoreOrder`. Worth a message that says so for whoever bisects this later.

The claim I most wanted to check holds: `untilize_wh.cpp`, `reader_unary_interleaved_wh_multicore.cpp` and `writer_unary_stick_layout_wh_multicore.cpp` are consumed only by these two factories — the quasar ports source their own copies under `experimental/quasar/…/kernels/` — so shifting `TensorAccessorArgs<3>`→`<4>` and `<4>`→`<5>` breaks no other caller. L1 is unchanged too: the four old `push_cb_pair` calls and the two new sets carry the same per-region sizes on the same (disjoint) cores.

Please do file the `untilize_with_unpadding` `(1,1,64,8192)`→`(1,1,60,8000)` hang as its own issue and link it here rather than leaving it in the description — it reproduces on main, so it will outlive this PR.

Aswinmcw (approved):
All four points addressed. Re-diffed `787a9304f` → `02d3a681e`: `common.{cpp,hpp}` and all three kernels are byte-identical, and the only code change is the `unpack_to_dest_mode` move into `make_compute_kernel` in both factories — each instance now marks fp32 unpack for exactly the buffer it reads. Description now matches the diff on `c_2`/`c_17` and the commit count, the consolidation is documented instead of left as an open question, and the hang is filed as #54524.

Approving for whenever #54140 lands and this rebases onto main.

VirdhatchaniKN (approved):
Pls address the above comments

ttnn/cpp/ttnn/operations/core/to_layout/to_layout_op.cpp
```diff
@@ -182,14 +182,17 @@ Tensor to_layout_impl(
...
                 // workaround by applying padding and then tilizing
-                ttsl::SmallVector<std::array<uint32_t, 2>> padding = {
-                    {0, 0},
-                    {0, 0},
-                    {0, padded_output_shape[2] - output_shape[2]},
-                    {0, padded_output_shape[3] - output_shape[3]}};
+                // Pad only the last two dims (height/width), sized to the tensor's rank.
+                int padding_rank = padded_output_shape.rank();
+                ttsl::SmallVector<std::array<uint32_t, 2>> padding(padding_rank, {0, 0});
+                padding[padding_rank - 2] = {0, padded_output_shape[-2] - output_shape[-2]};
```
mateusznowakTT, line 192: Is it possible to hit here empty or 1d tensor?

abdulla-TT: No, a rank <2 sharded tensor can't reach TILE layout since it becomes (1, N), so the shard height of 1 fails tile alignment earlier: "Physical shard shape (1, 32) must be tile {32, 32} sized!". Scalars fail the same way.
Fair point on the line itself though — at rank 1 it'd be padding[-1], so I've added a
TT_FATAL(padding_rank >= 2) above it.

mateusznowakTT (approved):

bbradelTT (approved):

mcw-anasuya (approved):

ttnn/cpp/ttnn/operations/data_movement/common/common.cpp
```diff
@@ -789,4 +792,159 @@ uint32_t per_shard_page_size_bytes(const ttnn::Tensor& t, uint32_t row_bytes) {
...
 }
 
+tt::tt_metal::CoreRangeSet union_of(const tt::tt_metal::CoreRangeSet& a, const tt::tt_metal::CoreRangeSet& b) {
+    if (a.empty()) {
+        return b;
+    }
+    if (b.empty()) {
+        return a;
+    }
+    return a.merge(b);
```
amorrisonTT: Is this function necessary?

amorrisonTT:
It might be worth running all models pipeline to ensure there is no perf degradation. These ops are pervasive.

amorrisonTT (approved):

ttnn/cpp/ttnn/operations/data_movement/common/common.cpp
abdulla-TT: Dropped the condition since,  It was only guarding the empty cases, and merge already handles those a.merge(empty).

Issue #53613, opened 2026-08-19
Eltwise unit tests restructuring - standardised Eltwise tests


KalaivaniMCW:
- A structured `tt-metal/tests/ttnn/unit_tests/operations/eltwise/`
- Identify and remove redundant tests for ops that are now stable 
- Limit infra testing to one or two ops 
- Group test categories based on underlying infra and API structure
- Use bfloat16 bits for input and valid across valid input domain 
- Segregate tests further based on dtypes - float32, bloack format etc.
- Move long running tests to Nightly 

**Categories at Infra level:** 
- [ ] Unary : 106 ops grouped into 6 categories ; Done (2 categories)
- [ ] Binary : 66 Binary ops
- [ ] Ternary : 6 ternary ops

**Sub-categories:**
- [ ] Unary Composite 
- [ ] Binary composite 
- [ ] Ternary composite

**Expected :** 
- Add exhaustive tests for unary ops
- for binary and ternary, use a reasonable input coverage 
- remove redundant test cases in eltwise unit tests 
- follow a category-wise test file structure for forward ops in eltwise unit tests

KalaivaniMCW:
Need to look into ops: 
for supported input domain - i0, i1
Trig ops covered only for valid domain

KalaivaniMCW:
unary category 1 - PR https://github.com/tenstorrent/tt-metal/pull/53700
Group | WH baseline | WH my branch | Δ | BH baseline | BH my branch | Δ
-- | -- | -- | -- | -- | -- | --
1 | 8m 48s | 8m 50s | +2s | 7m 59s | 7m 18s | −41s
2 | 7m 56s | 7m 42s | −14s | 8m 12s | 7m 45s | −27s
3 | 12m 53s | 9m 57s | −2m 56s | 10m 03s | 9m 34s | −29s
4 | 14m 43s | 12m 09s | −2m 34s | 12m 54s | 12m 59s | +5s

KalaivaniMCW:
### PRs under this issue

**Merged**

| PR | Scope |
|---|---|
| #53700 | Unary bfloat16 coverage tests — category 1 ops |
| #53893 | Eltwise unit tests — remove redundant / outdated tests |
| #53894 | Remove all eltwise tests from `tt_dnn` sweep_tests |
| #54533 | Unary bfloat16 coverage tests — category 5 ops |
| #54701 | Fix mixed dtype cases in Eltwise Unary |
[...]

Issue #47177, opened 2026-06-16
Add Mixed dtype functionality for Elementwise infra


KalaivaniMCW:
- Add Mixed dtype functionality for binary_ng infra  and ternary infra following the typecast rules as Pytorch wherever necessary
- Decide on the dtype of the output tensor
- All the old issues are tagged here as sub-issues for tracking and testing

Explicit test cases for ops:

- [ ] where
- [ ] add and other binary arithmetics
- [ ] eq, ne and other relational ops

PR #54425, opened 2026-08-26
[Test] Mixed dtype cases in Eltwise Binary

KalaivaniMCW:
### PR Category

Fix
### Summary

#53963 #53613

### Notes for reviewers

- Added some tests covering mixed dtype case in eltwise ops
- Added restrictions to block unsafe mixed dtype cases

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kalaivani/mixed)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kalaivani/mixed)


tests/pipeline_reorg/ttsim-skip-list.yaml
```diff
@@ -30,6 +30,9 @@ wormhole_b0:
   # Likely OOM killer related failures -- probably need to run these tests serially
   - tests/ttnn/unit_tests/base_functionality/test_reshape.py::test_reshape_oob
 
+  # Mixed dtype operation failures
+  - tests/ttnn/unit_tests/operations/eltwise/test_ttsim_skipped.py::test_add_accurate_mode_for_mixed_dtype
```
Aswinmcw: `test_neg_mixed_dtype` is missing from both arch lists, but its own docstring records `ttsim [5082] ERROR: UndefinedBehavior: tensix_unpacr` — so it will still be collected on the `sim_*` SKUs and fail there. Add it under both `wormhole_b0` and `blackhole`; the list is consumed as `--deselect=<entry>` (see `Load TTSim skip list` in `ttnn-sanity-tests-impl.yaml`), and a bare `file.py::test_name` deselects every parametrization, so no per-param entries are needed.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_skipped.py
```diff
@@ -0,0 +1,72 @@
...
+    golden = golden_fn(torch_input).to(TORCH_DTYPE[out_dtype])
+    result = ttnn.to_torch(output_tensor)
+    assert output_tensor.dtype == out_dtype
+    assert torch.equal(result, golden)
+    return output_tensor
+
+
+@pytest.mark.parametrize("in_dtype", [ttnn.bfloat16, ttnn.float32])
+@pytest.mark.parametrize("out_dtype", [ttnn.bfloat16, ttnn.float32])
+def test_neg_mixed_dtype(device, in_dtype, out_dtype):
```
Aswinmcw: The `in_dtype`/`out_dtype` params are dead: the body runs the full 2×2 loop on lines 67-70 regardless of them, so all four cases do identical work plus one repeated combination at line 72. That's 4× the runtime for no extra coverage. Either drop the two `parametrize` decorators and keep the loop (it's what actually exercises the back-to-back program-cache path), or drop the loop and let the params do the work — not both.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_skipped.py
```diff
@@ -0,0 +1,72 @@
...
+    """Preallocated output dtype can differ from the input. Every in/out pair is
+    invoked back-to-back to check for program-cache collision.
+    ttsim [5082] ERROR: UndefinedBehavior: tensix_unpacr: unpack_to_dst=0 in_data_format=0 out_data_format=0"""
+    torch.manual_seed(0)
+    torch_input = torch.ones((64, 64), dtype=TORCH_DTYPE[in_dtype]) * 1.22
+    ttnn_op = ttnn.neg
+
+    for in_dt in (ttnn.bfloat16, ttnn.float32):
+        for out_dt in (ttnn.bfloat16, ttnn.float32):
+            torch_input = torch_input.to(TORCH_DTYPE[in_dt])
```
Aswinmcw: Chain-converting `torch_input` in place means the float32 cases never see a float32-precision value: `1.22` is truncated to bf16 (`1.21875`) on the first iteration and never regains precision, and `1.21875` is bf16-exact. So every in/out pair compares a value that survives a bf16 round-trip unchanged, and `torch.equal` on line 53 would pass even if the float32→float32 path silently went through bf16 — which is the bug class this test is here to catch. Rebuild the input from a float32 master inside the loop and pick a value that isn't bf16-representable.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_errors.py
```diff
@@ -0,0 +1,72 @@
...
+    torch.manual_seed(0)
+    dtype_b = ttnn.float32
+
+    torch_input_tensor_a = torch.randn((64, 64)) * 100
+    torch_input_tensor_b = torch.randn((64, 64)) * 100
+
+    input_tensor_a = ttnn.from_torch(torch_input_tensor_a, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device)
+    input_tensor_b = ttnn.from_torch(torch_input_tensor_b, dtype=dtype_b, layout=ttnn.TILE_LAYOUT, device=device)
+
+    output = ttnn.add(input_tensor_a, input_tensor_b, fast_and_approximate_mode=False)
```
Aswinmcw: Only the tensor-tensor form is covered here. The tensor-scalar path is where mixed-dtype scalar bugs usually hide (the scalar is packed per-dtype in `pack_scalar_runtime_arg`), so it's worth adding `ttnn.add(bf16_tensor, <python float>)` alongside this.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_skipped.py
```diff
@@ -0,0 +1,72 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import pytest
+import ttnn
+from tests.ttnn.utils_for_testing import assert_equal, assert_with_ulp, assert_with_pcc, flush_subnormal_values_to_zero
+
+pytestmark = pytest.mark.use_module_device
+
+# The tests in this file are skipped for ttsim because they are not supported, will error out with known reasons.
```
Aswinmcw: Grouping by "which platform skips this" rather than by op/functionality will age badly — when ttsim gains support the filename becomes wrong and the tests have to move, and mixed-dtype coverage isn't discoverable from the name. Since the skip list already keys on individual `file.py::test_name` entries, these could live in the op-appropriate file and still be deselected. This also cuts against the structure goals in #53613.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_skipped.py
```diff
@@ -0,0 +1,72 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import pytest
+import ttnn
+from tests.ttnn.utils_for_testing import assert_equal, assert_with_ulp, assert_with_pcc, flush_subnormal_values_to_zero
```
Aswinmcw: `assert_equal`, `assert_with_pcc` and `flush_subnormal_values_to_zero` are unused — only `assert_with_ulp` is called.

Aswinmcw (commented):
Description references `#56963`, which doesn't exist in this repo — did you mean #47177 (mixed dtype for eltwise infra)? Also worth using `Relates to #<n>` so the links register.

tests/pipeline_reorg/ttsim-skip-list.yaml
KalaivaniMCW: the ttsim error for neg test is different, I have triggered a CI run to check if workaround is pre-applied for this case during CI

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_errors.py
KalaivaniMCW: added TS variant

tests/pipeline_reorg/ttsim-skip-list.yaml
KalaivaniMCW: CI applies the workaround automatically for this case, hence the skip is not needed; the error is recorded in the test desc.

KalaivaniMCW: This is error will still ocuur when anyone tests locally on a ttsim environment

tests/pipeline_reorg/ttsim-skip-list.yaml
```diff
@@ -30,6 +30,9 @@ wormhole_b0:
   # Likely OOM killer related failures -- probably need to run these tests serially
   - tests/ttnn/unit_tests/base_functionality/test_reshape.py::test_reshape_oob
 
+  # Mixed dtype operation failures
+  - tests/ttnn/unit_tests/operations/eltwise/test_ttsim_errors.py::test_add_accurate_mode_for_mixed_dtype
```
mcraigheadTT, line 34: Not sure what this is about, first I'm hearing about it. in general I think we need silicon vs. sim test coverage parity going forward and that we shouldn't be adding anything new to this skip list.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -23,9 +23,16 @@ Tensor unary_impl(
...
-    if (optional_output_tensor.has_value()) {
+        if (optional_output_tensor.has_value()) {
+            TT_FATAL(
+                (output_dtype == optional_output_tensor->dtype()),
+                "Preallocated tensor's dtype is not compatible with the expected dtype");
+        }
+    } else if (optional_output_tensor.has_value()) {
         output_dtype = optional_output_tensor->dtype();
+        TT_FATAL(
+            (output_dtype == input_dtype) || (is_floating_point(output_dtype) && is_floating_point(input_dtype)),
```
Aswinmcw, line 34: `is_floating_point` also returns true for `BFLOAT8_B`, `BFLOAT4_B` and `FP8_E4M3`, so this declares e.g. `ttnn.neg(bf16_in, output_tensor=<bfloat4_b>)` safe — but `TORCH_DTYPE` in the new test maps only bfloat16 and float32, so no block-float or fp8 output is ever exercised. Given the bug underneath #53963 is a packer emptying at the wrong width, a shared-exponent output format is exactly where I'd expect the next instance of it. Either widen the test matrix to the formats this predicate admits, or narrow the predicate to the ones you've actually validated.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -23,9 +23,16 @@ Tensor unary_impl(
     if (op_chain.back().type() == unary::UnaryOpType::TYPECAST ||
         op_chain.back().type() == unary::UnaryOpType::BITCAST) {
         output_dtype = static_cast<DataType>(*op_chain.back().get_param_if<float>(1));
-    }
-    if (optional_output_tensor.has_value()) {
+        if (optional_output_tensor.has_value()) {
+            TT_FATAL(
+                (output_dtype == optional_output_tensor->dtype()),
+                "Preallocated tensor's dtype is not compatible with the expected dtype");
```
Aswinmcw, line 29: Both new `TT_FATAL`s carry byte-identical text and neither prints the dtypes, so a user can't tell which rule fired or what they passed. `validate_on_program_cache_miss` in `unary_device_operation.cpp` is where this codebase already puts unary dtype invariants (LGAMMA, SOFTCAP), it already computes `output_tensor.has_value() ? output_tensor->dtype() : args.output_dtype`, and every message there formats the dtype with `{}` — worth matching, or at minimum adding the dtypes here.

Also worth knowing for the message wording: this branch is *not* on the `ttnn.typecast` path. `ttnn::typecast` goes to `prim::typecast` and has had its own `output_dtype == optional_output_tensor->dtype()` check at `typecast.cpp:67` for a while. This only fires when a TYPECAST/BITCAST is the tail of a fused chain (`unary_chain(..., output_tensor=t)`), and someone hitting it there has no way to guess the complaint is about the trailing typecast's target dtype.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_errors.py
```diff
@@ -0,0 +1,73 @@
...
+    invoked back-to-back to check for program-cache collision.
+    ttsim [5082] ERROR: UndefinedBehavior: tensix_unpacr: unpack_to_dst=0 in_data_format=0 out_data_format=0"""
+    torch.manual_seed(0)
+    fixed_input = torch.ones((64, 64), dtype=torch.float32) * 1.22
+    ttnn_op = ttnn.neg
+
+    for in_dt in (ttnn.bfloat16, ttnn.float32):
+        for out_dt in (ttnn.bfloat16, ttnn.float32):
+            torch_input = fixed_input.to(TORCH_DTYPE[in_dt])
+            _unary_with_preallocated_output(device, torch_input, in_dt, out_dt, ttnn_op)
```
Aswinmcw: Both tests cover the combinations that are now *allowed*; nothing asserts either new `TT_FATAL` actually fires. Since blocking unsafe combinations is half of what the PR does, a `pytest.raises` pair would pin it — one for `ttnn.neg(bf16_in, output_tensor=<uint32>)` and one for a fused chain ending in a typecast whose target disagrees with the preallocated tensor. Without them, loosening the predicate later is a silent change.

tests/ttnn/unit_tests/operations/eltwise/test_ttsim_errors.py
```diff
@@ -0,0 +1,73 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+
+# SPDX-License-Identifier: Apache-2.0
+
+import torch
+import pytest
+import ttnn
+from tests.ttnn.utils_for_testing import assert_with_ulp, assert_equal
```
Aswinmcw: `assert_equal` is still unused — only `assert_with_ulp` is called.

Aswinmcw (commented):
Re-review of the new commits. The test rework addresses the earlier points — the dead `in_dtype`/`out_dtype` params are gone, `torch_input` is now rebuilt from a float32 master each iteration so the fp32→fp32 pair finally sees a value bf16 can't represent, and the TS variant is in. Three things on the restriction half, inline.

One scoping note worth putting in the description: #53963's actual cache-key fix (`output_dtype = optional_output_tensor->dtype()` for the non-typecast path) already landed on main in #54364, so what this PR adds on top is the guard rails plus coverage, not the collision fix itself. Reads today as if it's the fix.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
KalaivaniMCW: torch doesn't have block float formats, so there is no direct comparisons for these cases

mateusznowakTT (approved):

mcraigheadTT:
Didn't see a reply to my comments/questions on the skip list, also, I see a test called test_ttsim_errors.py and I'm not really sure what this is, that seems like an odd name for a test to merge into tt-metal main (especially since I still haven't heard from anyone about what is going on here).

KalaivaniMCW:
@mcraigheadTT  
The failing case here is add operation performed with mixed dtype on the SFPU  (bf16 + fp32) 
The Tensix unpacker has two configuration contexts (0 and 1) for pipelining. Every tile unpack alternates between them
With mixed dtypes (bf16 + fp32), the kernel calls `copy_tile_to_dst_init_short_with_dt` to reconfigure the unpacker format between operands. This writes the new base address to whichever context is active. Now the two contexts have different configurations (one for bf16, one for fp32). When UNPACR executes with `cfg_context_id`=1 and the context 1 registers hold genuinely different data, ttsim cannot resolve it and reports `UnsupportedFunctionality`. it's that mixed dtype forces the two unpack config contexts to diverge (different source addresses/formats), and ttsim doesn't implement reading from the context 1 register bank when it actually differs from context 0.
I'll check once if there is anything that needs updation in the program factory/kernel side.

VirdhatchaniKN (approved):

mcraigheadTT:
Per https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/UNPACR_Regular.md, cfg_context_id (bits 12:10) in the UNPACR instruction - called ContextNumber in tt-isa-documentation - is officially UnsupportedFunctionality and is not to be used, per https://github.com/tenstorrent/tt-isa-documentation/blob/main/Glossary.md#unsupportedfunctionality and https://github.com/tenstorrent/ttsim-private/blob/main/docs/unsupported_functionality.md (slightly cut down public version at https://github.com/tenstorrent/ttsim/blob/main/docs/unsupported_functionality.md).

Please change your software to avoid using this unsupported feature.

mcraigheadTT (changes requested):
Code that hits UnsupportedFunctionality needs to be rewritten to avoid using the officially unsupported hardware feature.

KalaivaniMCW:
**Summary**
**What happens:** When we have mixed dtypes (input_a = bf16, input_b = fp32), the hardware unpacker needs two configuration contexts to handle the two different data formats simultaneously:

```
cfg_context_id=0 → unpack configuration for bf16 operand
cfg_context_id=1 → unpack configuration for fp32 operand
```
The Tensix hardware supports multiple unpack configuration contexts natively, so this works fine on real silicon.

**Why ttsim fails:** The ttsim simulator only implements cfg_context_id=0. When the kernel tries to use cfg_context_id=1 for the second operand format, ttsim reports it as UnsupportedFunctionality because it hasn't implemented multi-context unpacking.

**Result:** The kernel works on real hardware. The test passes in metal device but not in ttsim environments.

PR #54429, opened 2026-08-26
[LLK]Oprimize SFPU ema

ldjurovicTT:
### PR Category

Performance

### Summary

Records the EMA SFPU math body into the replay buffer once per tile and replays it for the
remaining seven blocks. **The arithmetic is untouched, so results are bit-identical.**

| | main | this PR | Δ | |
|---|---|---|---|---|
| **MATH_ISOLATE** cycles/tile | 320.77 | **247.93** | **−72.84** | **−22.71 %** |
| L1_TO_L1 cycles/tile | 338.64 | 267.66 | −70.98 | −20.96 % |
| Math ELF `TEXT_SIZE` | 3015 | 2599 | −416 B | −13.80 % |
| Output bits, 1000 alphas | — | **identical** | — | 0 / 2,048,000 differ |

Measured on **Wormhole n300**, `--speed-of-light`, `loop_factor=16`, `tile_cnt=8`,
`Float16_b→Float16_b`, `dest_acc=No`. Both configurations run twice; every figure reproduced
to the last decimal.

### What changed

`_compute_ema_math_body_()` takes no template parameters, so all 8 blocks of a tile emit
byte-identical instructions — only the surrounding loads and stores differ, through
compile-time dst offsets. So the body is recorded on the first block and replayed on the
other seven:

```c
template <bool RECORD>
sfpi_inline void _compute_ema_math_()
{
    TTI_SFPTRANSP(0, 0, 0, 0);                       // inline on all 8 blocks

    if constexpr (RECORD) {
        lltt::record<lltt::Exec>(EMA_MATH_REPLAY_SLOT, EMA_MATH_REPLAY_LEN);  // LEN == 16
        _compute_ema_math_body_();                   // exactly 8 SFPMAD + 8 SFPNOP
    } else {
        lltt::replay(EMA_MATH_REPLAY_SLOT, EMA_MATH_REPLAY_LEN);
    }

    TTI_SFPMOV(0, LREG3, LREG4, 0);                  // inline on all 8 blocks
    TTI_SFPTRANSP(0, 0, 0, 0);                       // inline on all 8 blocks
}
```

8 × 16 recorded instructions collapse to 16 + 7; the bracketing `SFPTRANSP`/`SFPMOV`/
`SFPTRANSP` stay inline. **No instruction inside the body changed** — same two SFPMADs per
row, same 8 SFPNOPs, same registers, same order. That is why it is bit-exact by
construction, and the sweep below confirms it rather than assuming it.

The record is held to 16 rather than the body's full 19 deliberately: `cmath_common.h`
splits the per-thread replay buffer between issuers (`replay_buf_offset == 16`), so `[0,16)`
is the SFPU half and `[16,32)` is the FPU's. Recording 19 at slot 0 overwrites three FPU
slots — see **Replay buffer boundary** below.

### Why this is a cycle win, which was not expected

Replay is normally described as a code-size optimisation. `ckernel_sfpu_exp.h` states
"Per-element runtime is unchanged: the recorded instructions execute exactly as if they had
been issued inline". For this kernel that is not what happens — replay is worth −72.88
cycles/tile.

The explanation is that **the inlined kernel is bound by the rate at which the TRISC can
push TTI instructions into Tensix, not by SFPU execution latency.** A `REPLAY` expands
inside Tensix without a per-instruction push, so the 8 `SFPNOP`s per block become almost
free once recorded: they are expensive to *issue* even though they are free to *execute*.

**But the win is not linear in instruction count, and it is worth saying so** — the obvious
reading of the above is that every collapsed instruction buys ~0.58 cycles (126 collapsed,
72.9 saved). That is contradicted by direct measurement: moving 3 of the 19 instructions
back inline puts 21 pushes per tile back, which the linear model prices at ~12 cycles. It
costs **0.04** (247.89 with all 19 recorded, 247.93 with 16). What fits is that the fully
inlined kernel at 152 pushes/tile really is push-bound — that is where the 72.9 cycles come
from — but once the body is replayed the bottleneck moves to replay expansion inside Tensix,
and a few extra pushes per block hide underneath it. So the first ~126 collapsed
instructions are nearly all of the win and the last few are worth almost nothing.

Two pieces of evidence for that reading:

- The win tracks instruction count, not arithmetic. A variant of this kernel with the NOPs
  removed (13 instructions/block instead of 19) gains **less** from replay in absolute terms
  — −46.3 vs −72.9 cycles — precisely because it has fewer instructions left to collapse.
- `TEXT_SIZE` falls 500 B, matching 126 collapsed instructions × 4 B.

This matters beyond EMA: it means NOP-heavy hand-scheduled SFPU bodies on Wormhole may be
paying for the NOPs at *issue* rather than at execution, and replay addresses that directly
without touching numerics.

### Correctness

The instruction sequence is unchanged, so bit-exactness is expected — but "expected" is not
evidence, so it was measured across the whole documented alpha range.

`ttnn.ema` takes an arbitrary float alpha and derives `beta = 1 - alpha`
(`ema_program_factory.cpp:99`), with no default. The pre-existing test pins a single value,
`alpha = 0.25`, which is the least informative choice available: 0.25 is an exact power of
two, so `alpha * EMA_old` is a pure exponent shift with no rounding.

`test_sfpu_ema_alpha_sweep` (new, nightly, in `test_sfpu_ema.py`) sweeps `alpha = k/1000` for k in 0..999 — 1000
cases, 2048 outputs each — measured against an fp64 reference recurrence rather than a
same-precision golden, so the residual is the kernel's own rounding rather than a mix of
kernel and golden error. Run on both kernels:

| | |
|---|---|
| alphas compared | **1000** |
| outputs compared | **2,048,000** |
| **differing** | **0** |
| `test_sfpu_ema.py` | 3 passed |
| `test_perf_header_gate.py` | 12 passed |

Accuracy characterisation from the same sweep, for reference: the kernel sits within about
one bfloat16 step of the exact recurrence at every alpha (worst peak error 1.881 ULP, at
alpha=0.962). Absolute error falls 65× from alpha=0.1 to alpha=0.999 while the signal itself
shrinks 85×, so ULP error — flat at 1.0–1.3 — is the honest reading and absolute error is
just tracking the signal.

### Raw runs

`TILE_LOOP` row of `perf_data/perf_sfpu_ema/perf_sfpu_ema.post.csv` (cycles/tile — the
`.post.csv` divides the `TILE_LOOP` means by `loop_factor * tile_cnt`):

| run | kernel | mean(MATH_ISOLATE) | mean(L1_TO_L1) | TEXT_SIZE(MATH_ISOLATE) |
|---|---|---|---|---|
| 1 | main | 320.7734375 | 338.640625 | 3015 |
| 2 | main | 320.7734375 | 338.640625 | 3015 |
| 1 | this PR (16 recorded + 3 inline) | 247.9296875 | 267.6640625 | 2599 |
| 2 | this PR (16 recorded + 3 inline) | 247.9296875 | 267.6640625 | 2599 |
| 1 | 19 recorded (superseded, crosses the FPU boundary) | 247.890625 | 267.6328125 | 2515 |
| 2 | 19 recorded (superseded, crosses the FPU boundary) | 247.890625 | 267.6328125 | 2515 |

Zero run-to-run variance. The baseline was produced by checking out the kernel as it stands
at the merge base into the same tree, so `_compute_ema_math_` is the only difference — same
harness, flags, silicon and session.

**Re-measured after rebasing onto `a6c0acfae42`** (48 commits of upstream movement): every
figure above reproduced exactly, and a full re-capture of the baseline's raw output bits came
back byte-identical to the pre-rebase capture, so none of the upstream changes affect this
kernel's results.

### Considered and not taken: reassociating the recurrence

The 8 `SFPNOP`s come from two chained SFPMADs per row (`LREG7 = alpha*prev`, then
`row = beta*input + LREG7`), each needing a filler behind it. They cannot be filled by
reordering: the rows must occupy LREG0-3 to be transposed as a group by `SFPTRANSP`, they are
live for the whole span where the slots exist, and the stores need the post-transpose layout,
so no independent work can be scheduled in. Nor by freeing registers — moving alpha/beta into
the programmable constants LREG12/LREG13 was tried, is bit-exact, and measured **321.20
cycles/tile, i.e. very slightly slower than main**.

Pre-scaling the inputs by beta *does* remove 6 of the 8 NOPs, and combined with replay reaches
**199.91 cycles/tile (−37.7 %)**. It was not taken here because it is not bit-neutral: it gives
`beta*input` its own rounding instead of the partially-fused one (`SFPMAD` performs exactly one
rounding per instruction — WormholeB0 `SFPMAD.md`), so both forms round the same number of times
but round different quantities. Measured over the same 1000-alpha sweep:

- accuracy unchanged — mean RMS error ratio 1.000001 against the fp64 reference, identical
  worst peak error, no alpha differing by more than 2 %;
- but **923 of 1000 alphas bit-identical and 77 not** — 87 of 2,048,000 outputs differ, 84 of
  them by exactly one bfloat16 ULP. Only the 8 exactly-representable alphas (multiples of
  0.125) are structurally safe.

So it is a real further −19 % available at the cost of an ULP-level change to a public op's
results. Worth revisiting deliberately; not folded into a change that is otherwise free.

### New coverage

- `tests/sources/sfpu_ema_perf.cpp` — perf driver, `MATH_ISOLATE` + `L1_TO_L1`. The EMA kernel
  had no perf coverage at all, which is why this change could not previously be quantified.
- `tests/python_tests/perf_sfpu_ema.py` — `dest_acc=No` only (the kernel is hardcoded for
  16-bit bf16 DEST), `loop_factor=16`, `[128, 64]`.
- `tests/python_tests/test_sfpu_ema.py` — the 1000-alpha sweep above added as `test_sfpu_ema_alpha_sweep`, nightly.
  Takes `LLK_EMA_SWEEP_DUMP=<path>` to append raw output bits per alpha, which is how the
  kernel-to-kernel bit comparisons in this PR were produced.
- `helpers/perf/test_schemas.py`, `helpers/perf/wide_schema.py` — schema entries; without the
  latter the CSV→Parquet conversion fails strict mode.

Reproduce:

```bash
cd tt_metal/tt-llk/tests
CHIP_ARCH=wormhole pytest -q --speed-of-light --compile-producer -m perf ./python_tests/perf_sfpu_ema.py
CHIP_ARCH=wormhole pytest -q --speed-of-light --compile-consumer -m perf ./python_tests/perf_sfpu_ema.py
# each consumer run writes a fresh timestamped directory
LATEST=$(ls -dt ../perf_data/runs/local-* | head -1)
awk -F, 'NR==1 || /TILE_LOOP/' "$LATEST/perf_sfpu_ema/perf_sfpu_ema.post.csv"
```

Output lands under `tt_metal/tt-llk/perf_data/runs/local-<timestamp>/`, **not**
`tests/perf_data/`. cycles/tile is the `TILE_LOOP` row of the `.post.csv`, which divides the
raw means by `loop_factor * tile_cnt`; the sibling `.parquet` holds the un-normalised means
(31735 raw / 128 = 247.93).

### Notes for reviewers

- **Replay buffer boundary — the record must stay at 16.** `cmath_common.h` splits the
  per-thread buffer at `replay_buf_offset == 16`: `[0,16)` SFPU, `[16,32)` FPU. An earlier
  revision of this PR recorded 19 at slot 0 and silently overwrote three FPU slots.
  `_llk_math_transpose_dest_init_` records its MOP into `[16,32)` and the per-tile op replays
  out of it, so a math kernel that inits transpose_dest and then runs an EMA tile got wrong
  transposes. Measured on n300 by calling the real `_compute_ema_math_` between that init and
  its replays — it touches only LREGs, transpose_dest works through SrcA/SrcB/DEST, so the
  replay buffer is the only thing they share:

  | EMA interleaved | dest_acc=Yes | dest_acc=No |
  |---|---|---|
  | 19 recorded | **12288 / 16384 wrong** | **3059 / 16384 wrong** |
  | 16 recorded (this PR) | **0 / 16384** | **0 / 16384** |

  A `NoExec` probe claiming `[0,N)` in the same position isolates the boundary exactly:
  `N <= 16` clean, `N = 17/18/19` corrupts 4096/8192/12288 outputs — precisely 4096 per FPU
  slot crossed. Holding the record to 16 costs 0.04 cycles/tile and 84 B.

  Note this is not unique to EMA: `ckernel_sfpu_exp.h` records `(0, 32)` and `ckernel_sfpu_topk.h`
  up to `(0, 29)`, both of which cross the same boundary far harder. Worth a separate look.
- **`EMA_MATH_REPLAY_LEN` must equal the recorded instruction count exactly** (16 = 8 SFPMAD
  + 8 SFPNOP) or the replay buffer misaligns and the wrong sequence executes silently. Any
  edit to the body must update it. Nothing enforces this — `lltt::record` has no bounds check
  and no static assert; the guard that would catch it, `make_sequence` in
  `ckernel_code_sequence.h`, is used only by `llk_math_matmul.h`.
- **Within the SFPU half, slot 0 is still shared** with the accurate `exp` path and topk, so a
  caller interleaving EMA with either must not rely on slot 0 surviving.
- **Wormhole only.** Not applied to the Blackhole copy, which has no NOPs and would need its
  own measurement to justify.
- **`MATH_ISOLATE` includes the datacopy**, matching `eltwise_unary_sfpu_perf.cpp` — it is what
  retires the SrcA valid bits unpack sets. Fixed cost inside the marker, cancels in the delta,
  but the absolute number is not the SFPU block alone.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ldjurovic/sfpu_wh_nop_overlap)


tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_ema.h
```diff
@@ -79,45 +79,46 @@ sfpi_inline void _compute_ema_math_()
...
-    // Step 1(in3): Calculate α * EMA_old in LREG7
-    // LREG7 = LREG5 * LREG2 (α * EMA_old)
-    TTI_SFPMAD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LREG2, ckernel::p_sfpu::LCONST_0, ckernel::p_sfpu::LREG7, 0);
-    TTI_SFPNOP; // Next cycle cannot read from LREG7 (2-cycle operation)
-
-    // Step 2(in3): Calculate final EMA = β * in3 + α * EMA_old
-    // LREG3 = (LREG6 * LREG3) + LREG7
-    TTI_SFPMAD(ckernel::p_sfpu::LREG6, ckernel::p_sfpu::LREG3, ckernel::p_sfpu::LREG7, ckernel::p_sfpu::LREG3, 0);
-    TTI_SFPNOP; // Next cycle cannot read from LREG3 (2-cycle operation)
+    // EMA_new = alpha * EMA_old + beta * input, chained across the 4 rows.
```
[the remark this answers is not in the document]
ldjurovicTT: Fixed in 50d05689be9. The file had two contradictory register maps: the pre-existing one above the new explanation still described the old two-MAD schedule as current (`LREG7=EMA_new`, `LREG7 = LREG5 * LREG4 + LREG6 * LREG0`).

Replaced it with a map of what the code actually does now — LREG0-3 hold the 4 rows and are updated in place to become the outputs, LREG4 carries EMA_old in and the new carry out, LREG5 = alpha, LREG6 = beta, and LREG7 is unused because the pre-scaling removes the need for a temp.

On the line-96 reference: the remaining `LREG7` mentions are deliberate — they describe the *previous* schedule as the thing being replaced. I have made that framing explicit ("That earlier schedule (no longer what the code below does) spent two MADs per row...") so it cannot be read as a description of the current register usage.

tt_metal/tt-llk/tests/sources/sfpu_ema_perf.cpp
```diff
@@ -0,0 +1,239 @@
...
+            return;
+        }
+        else if constexpr (PERF_RUN_TYPE == PerfRunType::UNPACK_ISOLATE || PERF_RUN_TYPE == PerfRunType::L1_CONGESTION)
+        {
+            _perf_math_loop_clear_valid</* clear A */ true, /* clear B */ false>(TILE_CNT * LOOP_FACTOR);
+        }
+        else if constexpr (PERF_RUN_TYPE == PerfRunType::MATH_ISOLATE)
+        {
+            // Only the SFPU math block: no datacopy and no dest handshake, so the
+            // measurement is the EMA kernel itself.
```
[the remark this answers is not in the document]
ldjurovicTT: Correct, and fixed in 50d05689be9. That claim was a leftover from before the datacopy turned out to be necessary, and I did not go back and delete it when I added the explanation below it — so the file asserted both things three lines apart.

It now reads: "Isolates the math pipe: no dest handshake with pack, so what is left is the datacopy plus the EMA kernel." The paragraph that follows already explained why the datacopy has to stay (it is what retires the SrcA valid bits unpack sets — dropping it and using a bare `TTI_CLEARDVALID` hangs the math thread) and that it is therefore a fixed cost inside the marker.

tt_metal/tt-llk/tests/python_tests/perf_sfpu_ema.py
```diff
@@ -0,0 +1,104 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+MATH_ISOLATE perf for the EMA SFPU kernel (ckernel_sfpu_ema.h).
+
+The EMA kernel had no perf coverage at all, which is why the Wormhole scheduling
+change to _compute_ema_math_ could not be quantified from the existing suite. This
+drives sources/sfpu_ema_perf.cpp, whose TILE_LOOP marker under MATH_ISOLATE is just
+the SFPU math block -- no datacopy, no dest handshake -- so the number attributable
+to the kernel is not diluted by the surrounding pipeline.
```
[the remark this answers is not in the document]
ldjurovicTT: Agreed, fixed in 50d05689be9 — same leftover claim as in the C++ driver.

The docstring now says explicitly that `mean(MATH_ISOLATE)` is the SFPU block **plus a fixed datacopy**, not the standalone SFPU cost; that the offset is constant across a before/after comparison of the kernel so it cancels in a delta; and that the absolute number should not be read as the kernel alone. Same wording as the "Notes for reviewers" entry in the PR body, so the two no longer disagree.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_ema.h
```diff
@@ -74,50 +74,52 @@ sfpi_inline void _compute_ema_math_()
...
```
fvranicTT: Can we make a similar change for BH? Even though BH doesn't have NOPs, it does scoreboarding internally and stalls. And from what I can see, we would benefit from updating BH code too:

```cpp
sfpi_inline void _compute_ema_math_()
{
    // Transpose the input data to the correct order
    TTI_SFPTRANSP(0, 0, 0, 0);

    // EMA equation: EMA_new = α * EMA_old + β * input
    // Where: LREG0-3=input rows, LREG4=EMA_old, LREG5=α, LREG6=β=(1-α)
    // Each row is computed in place: LREG_i = LREG5 * LREG_{i-1} + LREG6 * LREG_i
    //
    // EMA_new = alpha * EMA_old + beta * input, chained across the 4 rows.
    //
    // Blackhole interlocks on LREG results, so no SFPNOP padding is needed here. The
    // stalls are still paid, they are just invisible in the instruction stream: the
    // straightforward encoding is two MADs per row -- alpha * prev, then
    // row = beta * input + that -- which makes all 8 MADs one serial dependency chain,
    // so every issue stalls on its predecessor.
    //
    // Scaling every input by beta up front instead leaves one MAD per row on the chain
    // (row_i = alpha * row_{i-1} + beta_scaled_i, a single fused multiply-add), halving
    // the chain to 4. The four scaling multiplies are mutually independent, so three of
    // them can be dealt into the chain's stall slots. Instruction count is unchanged at
    // 8 MADs, so this costs nothing in code size; it only removes stall cycles. Two
    // stall slots remain unfilled: there is no independent work left to schedule into
    // them, because all 8 LREGs are committed (4 rows, the carry, alpha, beta) and the
    // only spare sits in the LREG4-7 quad that SFPTRANSP rewrites. See the Wormhole
    // copy of this kernel for the full slot accounting.
    //
    // This reassociates the rounding -- the old form rounded alpha*prev alone and fused
    // beta*input into the add, this one rounds beta*input alone and fuses alpha*prev --
    // so it is not bit-identical in fp32. It is expected to be bit-identical at the
    // output, because DEST here is bfloat16 and an fp32 reassociation lands ~2^-24
    // relative, far under bfloat16's 2^-9. The equivalent Wormhole change was verified
    // over 24 stimulus sets (8 seeds x 3 amplitudes, 172032 outputs) with every value
    // identical to the two-MAD form; that sweep has not yet been rerun on Blackhole.
    //
    // Pre-scale in0/in1 by beta. Independent of everything below.
    TTI_SFPMAD(ckernel::p_sfpu::LREG6, ckernel::p_sfpu::LREG0, ckernel::p_sfpu::LCONST_0, ckernel::p_sfpu::LREG0, 0);
    TTI_SFPMAD(ckernel::p_sfpu::LREG6, ckernel::p_sfpu::LREG1, ckernel::p_sfpu::LCONST_0, ckernel::p_sfpu::LREG1, 0);

    // in0: LREG0 = alpha * EMA_old + beta * in0
    TTI_SFPMAD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LREG4, ckernel::p_sfpu::LREG0, ckernel::p_sfpu::LREG0, 0);
    // Pre-scale in2, covering the stall behind the MAD above.
    TTI_SFPMAD(ckernel::p_sfpu::LREG6, ckernel::p_sfpu::LREG2, ckernel::p_sfpu::LCONST_0, ckernel::p_sfpu::LREG2, 0);

    // in1: LREG1 = alpha * LREG0 + beta * in1
    TTI_SFPMAD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LREG0, ckernel::p_sfpu::LREG1, ckernel::p_sfpu::LREG1, 0);
    // Pre-scale in3, covering the stall behind the MAD above.
    TTI_SFPMAD(ckernel::p_sfpu::LREG6, ckernel::p_sfpu::LREG3, ckernel::p_sfpu::LCONST_0, ckernel::p_sfpu::LREG3, 0);

    // in2: LREG2 = alpha * LREG1 + beta * in2
    TTI_SFPMAD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LREG1, ckernel::p_sfpu::LREG2, ckernel::p_sfpu::LREG2, 0);

    // in3: LREG3 = alpha * LREG2 + beta * in3
    TTI_SFPMAD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LREG2, ckernel::p_sfpu::LREG3, ckernel::p_sfpu::LREG3, 0);

    // Update EMA_old for next iteration
    // LREG4 = LREG3 (copy new EMA to old EMA register)
    TTI_SFPMOV(0, ckernel::p_sfpu::LREG3, ckernel::p_sfpu::LREG4, 0);

    // Transpose the output data to the correct order
    TTI_SFPTRANSP(0, 0, 0, 0);
}
```
Results in less dependent-adjacent issue pairs. We went from 8 down to 2.

ldjurovicTT: I am planning on changing BH kernel too, but first I need to take a very close look into bit-exactness of results since copilot caught a potential problem

fvranicTT (approved):
Let's tackle BH version independently then.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_ema.h
ldjurovicTT: On BH side putting this body in replay buffer doesn't bring any speedup and also we can't rearange these MULs without braking dependency chain

tt_metal/tt-llk/tests/python_tests/test_sfpu_ema_alpha_sweep.py
skrsmanovicTT, line 1: nitpick: Should this be included in already existing functional ema test, since you are reusing same kernel?

skrsmanovicTT (approved):
Please wait for @nstojicTT to confirm perf infra changes are correct since he is in charge for new infra features

tt_metal/tt-llk/tests/python_tests/helpers/perf/test_schemas.py
```diff
@@ -438,6 +438,32 @@
...
+            "marker",
+            "num_faces",
+            "num_faces_A",
+            "num_faces_B",
+            "tile_cnt",
+            "unpack_to_dest",
+            "unpack_transpose_faces",
+            "unpack_transpose_within_face",
+        ],
+        "test_name_aliases": {"perf_sfpu_ema": "perf_sfpu_ema"},
```
nstojicTT, line 466: could we add "aliases": {}
since its version 1 I understand it will be empty, its just for the sake of keeping the catalog uniform as all the other tests have it even if its empty

nstojicTT: Other than that, LGTM

ldjurovicTT: Done in 2a57e561232 — added `"aliases": {}`.

You are right that it was the odd one out: `perf_sfpu_ema` was the only entry of the 24 in the catalog without the key. Empty at version 1 as you say, so it changes nothing functionally, it just keeps the catalog uniform.

tt_metal/tt-llk/tests/python_tests/test_sfpu_ema_alpha_sweep.py
ldjurovicTT: Agreed — merged into `test_sfpu_ema.py` in 2a57e561232 and the separate file is gone.

You are right that the split was not paying for itself: the sweep drove the same `sources/sfpu_ema_test.cpp` through the same `TestConfig`, so the whole stimuli/tilize/config block was duplicated verbatim between the two files. That is now one `_run_ema_on_device(alpha, beta, num_time_tiles, dest_acc)` helper shared by both tests, which is a net −53 lines.

What stays different between them is only what actually differs: the functional test checks one alpha against the fp32 golden with `passed_test`, the sweep checks 1000 alphas against an fp64 reference recurrence with a ULP bound. The sweep keeps its `@pytest.mark.nightly`, so the default suite is unaffected — collection is 3 selected / 1000 deselected under `-m "not nightly"`.

For what it is worth this also matches what the rest of the directory does — `test_sfpu_reduce.py`, `test_eltwise_unary_sfpu.py` and `test_sfpu_ternary.py` all keep nightly and default-suite tests in one file.

Verified on Wormhole n300: `test_sfpu_ema` 3 passed, `test_perf_header_gate.py` 17 passed, and a sweep slice reproduces the numbers in the PR body (alpha=0.962 → 3.763 half-ULPs, i.e. the 1.88 ULP worst case).

PR #35904, opened 2026-01-15
Prefix caching for Llama3 70B Galaxy

viktorpusTT:
### Ticket
https://github.com/tenstorrent/vllm/issues/268

### Problem description
Automatic prefix caching (https://docs.vllm.ai/en/latest/features/automatic_prefix_caching/#enabling-apc-in-vllm) allows for re-use of cached KV entries when a new prompt arrives which shares its prefix with a previous prompt which was already processed. It will help reduce time to first token (TTFT) significantly when there are multiple users with shared prefixes and when a conversation with a user is continued over many turns.

### What's changed
After previous changes:
* enabling VLLM to support APC: https://github.com/tenstorrent/vllm/pull/272
* adding APC support to tt-transformers: https://github.com/tenstorrent/tt-metal/pull/33883
* adding support for paged+chunked prefill in ring distributed sdpa: https://github.com/tenstorrent/tt-metal/pull/35742

This PR adds support for Llama 70B Galaxy optimized model.

Related VLLM PR: https://github.com/tenstorrent/vllm/pull/335

### Performance
Depends on total seq len and the ratio of cached tokens. The higher the ratio the better, of course. For longer sequences, the benefits are lower, because prefix caching cannot use ring SDPA, which is faster. The implementation includes a heuristic and ignores prefix caching when the performance benefit would be negative. Thus the runtime ratio cached vs non-cached is never over 100%:
image

### CI
* VLLM: https://github.com/tenstorrent/tt-metal/actions/runs/23286297205 
  * failing the same way as main+dev: https://github.com/tenstorrent/tt-metal/actions/runs/23273761652/job/67672749751
  * Likely because of https://github.com/tenstorrent/tt-metal/issues/37397
* Galaxy demo tests: https://github.com/tenstorrent/tt-metal/actions/runs/23297050855/job/67748571451
* Tier 1 models sweep tests: https://github.com/tenstorrent/tt-metal/actions/runs/23286658562/job/67712587506

### Checklist

- [ ] [All post-commit tests](https://github.com/tenstorrent/tt-metal/actions/workflows/all-post-commit-workflows.yaml?query=branch:viktorpus/70b-galaxy-prefix-caching)
- [ ] [Blackhole Post commit](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:viktorpus/70b-galaxy-prefix-caching)
[...]

PR #54435, opened 2026-08-26
tt_transformers: align resumed prefill offsets and reserve their traces

viktorpusTT:
Fixes a silent wrong-answer bug in traced prefill with automatic prefix caching,
and stops the traces it needs from being allocated behind the operator's back.

Independent of, and a prerequisite for,
[tenstorrent/tt-metal#54302](https://github.com/tenstorrent/tt-metal/pull/54302)
(chunked prefill), which makes the same code path far busier.

## The bug

`tenstorrent/tt-metal#43165` made resumed prefill traceable by passing
`chunk_start_idx` to `ttnn.transformer.chunked_scaled_dot_product_attention` as a
`ttnn.Tensor` that is refreshed on every replay. The captured program config
therefore cannot be derived from the offset, so
`models/tt_transformers/tt/attention.py` builds it with `chunk_start_idx=0`:

| padded length of the resumed span | `q_chunk_size` the traced program is pinned to |
|---|---|
| `< 2048` | 64 |
| `>= 2048` | 256 |

`ttnn.transformer.chunked_scaled_dot_product_attention` requires
`chunk_start_idx` to be a multiple of the `q_chunk_size` its program was built
with. It reads the wrong prefix rather than raising when it is not.

Automatic prefix caching hands `Generator.prefill_forward_text` a `start_pos`
that is a multiple of the paged KV `block_size` (64), which need not be a
multiple of 256. So a cache hit whose remaining span reaches 2048 tokens can land
on an offset the captured program cannot honour, and the request answers from the
wrong prefix.

A full-prompt repeat never hits it, because the remaining span is tiny and the
pin is 64. It needs a partial prefix overlap followed by enough new tokens.

## Measurement

T3K, `meta-llama/Llama-3.1-8B-Instruct`, tracing on, automatic prefix caching on.
Two prompts share a prefix carrying a passphrase, then diverge for ~2300 tokens so
the second request's resumed padded length is 4096. Sweeping the shared prefix
length across every residue mod 256:

| | recall |
|---|---|
| before | **13/40** |
| after | **40/40** |

Failures are the shared prefix being read back as unrelated text, never an error.

## Changes

`models/tt_transformers/tt/generator.py`:

- `_align_resume_offsets` floors each `num_cached_tokens` to
  `math.lcm(block_size, q_chunk_size)`, iterating until the offset stops
  changing. It has to iterate: flooring lengthens the suffix, a longer suffix can
  pin a larger `q_chunk_size`, and that can demand a smaller offset again. With
  `block_size` 32, `seq_len` 1120 and `start_pos` 100 a single pass settles on 64
  against a pin of 256. Flooring recomputes at most `alignment - 1` tokens whose
  K/V is rewritten identically into the same blocks, so it is semantically a
  no-op.
- `_resume_offset_alignment` takes the pin from the model's own
  `get_attn_sdpa_program_config`, so a short cache hit on a short suffix keeps its
  64-token alignment and is not rounded away. A model whose `model_args` exposes
  no such method must declare
  `model_capabilities['resumed_prefill_token_alignment']`; the generator raises
  rather than fall back to block-only alignment, which is the wrong-prefix bug
  itself. `Gemma4ModelArgs` is that case, so both Gemma 4 generator classes
  declare `SDPA_CHUNK_ALIGN`, the same 128 that `align_num_cached_tokens_to_sdpa`
  already applies locally.
- `_warmup_prefill_resumed_sweep` captures the resumed ("sp1") prefill traces at
  warmup, batch 1, one per traced sequence length, at the offset a real request
  will be floored to. The warmup prompt is the offset plus a full bucket, capped
  at `capped_warmup_seq_len`, so the suffix padded into the bucket is the bucket
  itself. Spanning only the bucket leaves no suffix at all once `block_size`
  reaches the smallest traced length of 128. Gated on the model declaring
  `supports_prefix_caching` or `supports_chunked_prefill`.

Both mirror `models/demos/llama3_70b_galaxy/tt/generator.py`, which has paired a
`SDPA_CHUNK_ALIGN` round-down with a phase-2 resumed-prefill warmup since
`tenstorrent/tt-metal#35904`. `models/tt_transformers` took the tracing half of
that design without either.

## Why the warmup matters

`Generator.warmup_model_prefill` never passed a `start_pos`, so it only ever
captured the `sp0` half of the prefill trace key
`f"{prefill_seq_len}_{model_id}_{batch_size}_{use_start_pos}"`. Every `sp1` key
was captured lazily by whichever request happened to resume first, allocating
trace region at serving time that nothing had sized for. That is how a server
gets a fatal `TT_FATAL @ mesh_trace.cpp:82 ... only 50000000B is allocated for
trace region` mid-benchmark rather than at startup.

Reserving them makes the requirement a function of configuration instead of
traffic. Measured on T3K with `Llama-3.1-8B-Instruct`: 53248000B, against the
50000000B default. The three `Llama-3.1-8B-Instruct` legs still on that default
(`(prefix caching) [wh_llmbox]`, `with sampling-tests [bh_loudbox]`,
`DP=4 [wh_galaxy_perf]`) get an explicit `trace_region_size` of 85000000, the
value the sibling `with sampling-tests [wh_llmbox]` leg already uses.

Models that declare neither `supports_prefix_caching` nor
`supports_chunked_prefill` capture nothing extra at warmup, so their trace region
is unchanged. That covers Gemma 3, GPT-OSS, Mistral 3, Mllama and the Qwen VL
bridges, including the ones on a 28467200-byte region. Gemma 4 overrides
`warmup_model_prefill` in both its generator classes and is not reached.

The gate decides what to reserve, not what can happen. A model reaches
`_align_resume_offsets` on any nonzero `start_pos`, whatever it declares, which is
why the alignment has to be derivable or declared for every model that can resume
and not only for the ones warmed here.

Gemma 4 is the case in point. The plugin surfaces a chunked-prefill continuation as
a nonzero `start_pos`
([`model_runner.py:1159`](https://github.com/tenstorrent/vllm-tt-plugin/blob/main/src/vllm_tt_plugin/model_runner.py#L1159):
`input_positions = num_computed_tokens_cpu[req_indices]`), and the Gemma 4 legs run
`enable_chunked_prefill=True` with `max_num_batched_tokens=2496`. Any prompt past
that budget resumes at an offset that is not a multiple of 128, through the
fall-through `super().prefill_forward_text(..., start_pos=start_pos)` in
`models/demos/gemma4/tt/generator.py`. Its declaration is load-bearing, not
defensive hardening.

## Validation

https://github.com/tenstorrent/tt-metal/actions/runs/33065341377

Default `block_size` 64:

- Partial-prefix-hit sweep, 40 shared-prefix lengths: 13/40 before, 40/40 after.
- Whole-prompt repeat at 2000/3000/4000/6000 words, cold and cached: 8/8 before
  and after (this shape was never broken; included to show no regression).
- Warmup reserves exactly 5 resumed traces, at `num_cached` 64 for the 128 and
  1024 buckets and 256 for 2048, 4096 and 8192, matching the pinned
  `q_chunk_size` per length.
- With the 50000000B default the server now fails at startup rather than
  mid-serving, which is the intended consequence of reserving the traces.

`--block_size 128`, where the resume offset `lcm(block_size, q_chunk_size)`
reaches the smallest traced bucket and the warmup prompt has to clear it:

| | prompt spanning the bucket | prompt of offset plus a bucket |
|---|---|---|
| 128 bucket | `AssertionError: user 0: num_cached=128 must be < seq_len=128` | warms at `num_cached` 128 |
| server | `EngineCore failed to start` | `Application startup complete` |
| resumed traces reserved | 0 | 5, at `num_cached` 128/128/256/256/256 |
| partial-prefix-hit sweep | never ran | 40/40 |

## Known gap

Prefill warmup prepares each trace variant while earlier traces are already live,
against the invariant `_prepare_trace_prefill` documents. The existing sp0 sweep
already does this, so it is pre-existing rather than introduced here, and fixing
it changes behaviour beyond this PR. Tracked in
https://github.com/tenstorrent/tt-metal/issues/54441.

No CI leg runs a non-default `block_size`, so the `--block_size 128` results above
come from manual runs. Because the resumed traces are reserved at warmup, a
regression there fails at server startup rather than silently, so it stays loud.

The 53248000B measurement is T3K only. `bh_loudbox` and `wh_galaxy_perf` take the
same 85000000 without their own measurement. Because the traces are now reserved
at warmup, an undersized region fails at startup rather than mid-serving, so a
wrong value is loud.

## AI assistance

Written with AI assistance (Claude). Every line is to be reviewed and defended by
the human submitter.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:viktorpus/traced-apc-alignment-and-warmup)

### DIFF
--- a/models/common/tests/test_resume_offset_alignment.py
+++ b/models/common/tests/test_resume_offset_alignment.py
@@ -1,0 +1,179 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+Host-only tests for the resumed-prefill offset alignment in
+``models.tt_transformers.tt.generator.Generator``.
+
+A resumed prefill (prefix caching, or a prompt split across engine steps) hands
+the traced chunked SDPA a ``chunk_start_idx``. That op reads the wrong prefix
+instead of raising when the offset is not a multiple of the ``q_chunk_size`` its
+program was captured with, so the offset is floored to
+``lcm(block_size, q_chunk_size)`` before use.
+
+These are pure functions of ``(num_cached, seq_len, block_size, q_chunk_size)``.
+The tests bind the real methods to a stub so they run without a device.
+"""
+
+import math
+from types import SimpleNamespace
+
+import pytest
+
+from models.tt_transformers.tt.common import get_padded_prefill_len
+from models.tt_transformers.tt.generator import Generator
+
+# The buckets Llama-3.1-8B traces on T3K, and the ceiling warmup applies to them.
+TRACED_BUCKETS = (128, 1024, 2048, 4096, 8192)
+CAPPED_WARMUP_SEQ_LEN = 131072
+
+
+def _q_chunk_size(seq_len):
+    """The rule ``ModelArgs.get_attn_sdpa_prefill_program_config`` applies at chunk_start_idx=0."""
+    return 256 if seq_len >= 2048 else 64
+
+
+class _StubModelArgs:
+    def __init__(self, q_chunk_size_fn=_q_chunk_size):
+        self._q_chunk_size_fn = q_chunk_size_fn
+        self.capped_warmup_seq_len = CAPPED_WARMUP_SEQ_LEN
+
+    def get_attn_sdpa_program_config(self, mode, seq_len, chunk_start_idx, _unused):
+        return SimpleNamespace(q_chunk_size=self._q_chunk_size_fn(seq_len))
+
+
+class _StubModelArgsNoProgramConfig:
+    """A model whose args cannot describe the pin, e.g. Gemma4ModelArgs."""
+
+    def __init__(self):
+        self.capped_warmup_seq_len = CAPPED_WARMUP_SEQ_LEN
+
+
+class _StubGenerator:
+    """Borrows the real methods so the tests exercise shipped code, not a copy."""
+
+    _traced_sdpa_q_chunk_size = Generator._traced_sdpa_q_chunk_size
+    _resume_offset_alignment = Generator._resume_offset_alignment
+    _assert_uniform_resume_alignment = Generator._assert_uniform_resume_alignment
+    _align_resume_offsets = Generator._align_resume_offsets
+    # Read off the class a staticmethod is a plain function, which would rebind here.
+    _resumed_warmup_prompt_len = staticmethod(Generator._resumed_warmup_prompt_len)
+
+    def __init__(self, block_size, model_args=None, model_capabilities=None):
+        args = model_args if model_args is not None else [_StubModelArgs()]
+        self.model_args = args
+        self.data_parallel = len(args)
+        self.model_capabilities = model_capabilities or {}
+        self._block_size = block_size
+
+    def _paged_prefill_block_size(self, _kv_cache):
+        return self._block_size
+
+    def align(self, num_cached_per_user, prompt_lens):
+        # kv_cache only has to be non-None: _paged_prefill_block_size is stubbed.
+        return self._align_resume_offsets(num_cached_per_user, prompt_lens, kv_cache=[object()])
+
+
+@pytest.mark.parametrize("block_size", [32, 64, 128, 256])
+@pytest.mark.parametrize("bucket", TRACED_BUCKETS)
+def test_alignment_is_lcm_of_block_size_and_pin(block_size, bucket):
+    gen = _StubGenerator(block_size)
+    alignment = gen._resume_offset_alignment(bucket, block_size)
+    expected = math.lcm(block_size, _q_chunk_size(bucket))
+    assert alignment == expected
+    assert alignment % block_size == 0, "paged ops need the page-table slice to land on a block"
+    assert alignment % _q_chunk_size(bucket) == 0, "the traced SDPA program pins this q_chunk_size"
+
+
+def test_zero_offset_short_circuits_without_consulting_the_alignment():
+    """An ordinary prefill passes a zero-filled start_pos and must not need an alignment."""
+    gen = _StubGenerator(64, model_args=[_StubModelArgsNoProgramConfig()], model_capabilities={})
+    assert gen.align([0, 0], [1024, 4096]) == [0, 0]
+
+
+def test_undeclarable_alignment_raises_on_a_real_resume(expect_error):
+    gen = _StubGenerator(64, model_args=[_StubModelArgsNoProgramConfig()], model_capabilities={})
+    with expect_error(ValueError, "resumed_prefill_token_alignment"):
+        gen.align([128], [4096])
+
+
+def test_declared_alignment_is_used_when_the_pin_cannot_be_derived():
+    """Gemma 4's case: chunked_prefill_sdpa pins 128 and the model args cannot say so."""
+    gen = _StubGenerator(
+        64,
+        model_args=[_StubModelArgsNoProgramConfig()],
+        model_capabilities={"resumed_prefill_token_alignment": 128},
+    )
+    assert gen.align([200], [4096]) == [128]
+
+
+def test_already_aligned_offset_is_unchanged():
+    gen = _StubGenerator(64)
+    # 4096 pins 256; 512 is already a multiple of lcm(64, 256) = 256.
+    assert gen.align([512], [8192]) == [512]
+
+
+def test_settle_loop_takes_the_second_hop():
+    """One pass is not enough: flooring lengthens the suffix, which can raise the pin.
+
+    block_size 32, seq_len 1120, start_pos 100. A single pass floors 100 against the
+    suffix at offset 96, whose padded length is 1024 and pins 64, giving 64. The
+    longer suffix at offset 64 pads to 2048 and pins 256, so the offset must settle
+    lower still.
+    """
+    gen = _StubGenerator(32)
+    (settled,) = gen.align([100], [1120])
+    assert settled == 0
+    # The result is a fixed point: re-running does not move it.
+    assert gen.align([settled], [1120]) == [settled]
+
+
+def test_settled_offset_is_a_fixed_point_across_the_bucket_range():
+    gen = _StubGenerator(64)
+    for seq_len in (200, 1120, 2100, 4500, 9000):
+        for start_pos in range(0, seq_len, 37):
+            (settled,) = gen.align([start_pos], [seq_len])
+            assert 0 <= settled <= start_pos, "flooring never raises the offset"
+            assert gen.align([settled], [seq_len]) == [settled], "not a fixed point"
+
+
+def test_each_user_is_aligned_independently():
+    gen = _StubGenerator(64)
+    assert gen.align([0, 100, 512], [1024, 1120, 8192]) == [0, 0, 512]
+
+
+@pytest.mark.parametrize("block_size", [64, 128, 256, 1024])
+@pytest.mark.parametrize("bucket", TRACED_BUCKETS)
+def test_resumed_warmup_prompt_reaches_the_intended_bucket(block_size, bucket):
+    """Regression for the block_size >= 128 startup failure.
+
+    Spanning the bucket alone gave a suffix of ``bucket - num_cached``, which is 0 at
+    block_size 128 and negative at 256, so warmup tripped
+    ``assert 0 <= num_cached < seq_len`` and the server never started.
+    """
+    gen = _StubGenerator(block_size)
+    num_cached = gen._resume_offset_alignment(bucket, block_size)
+    total_seq_len = gen._resumed_warmup_prompt_len(bucket, num_cached, CAPPED_WARMUP_SEQ_LEN)
+    suffix = total_seq_len - num_cached
+
+    assert suffix > 0, "a resumed warmup with no suffix has nothing to prefill"
+    assert get_padded_prefill_len(suffix) == bucket, "the captured trace would be for another bucket"
+    assert 0 <= num_cached < total_seq_len, "would trip the guard in _prefill_forward_text_impl"
+
+
+def test_resumed_warmup_prompt_respects_the_warmup_ceiling():
+    """When the ceiling equals the bucket the prompt is capped, and the suffix still fits."""
+    gen = _StubGenerator(64)
+    bucket = 8192
+    num_cached = gen._resume_offset_alignment(bucket, 64)
+    total_seq_len = gen._resumed_warmup_prompt_len(bucket, num_cached, capped_warmup_seq_len=bucket)
+    assert total_seq_len == bucket
+    assert get_padded_prefill_len(total_seq_len - num_cached) == bucket
+
+
+def test_heterogeneous_replicas_are_rejected_rather_than_silently_aligned_to_replica_zero(expect_error):
+    replica_0 = _StubModelArgs()
+    replica_1 = _StubModelArgs(q_chunk_size_fn=lambda seq_len: 512)
+    gen = _StubGenerator(64, model_args=[replica_0, replica_1])
+    with expect_error(AssertionError, "replica 1"):
+        gen.align([300], [4096])
--- a/models/demos/gemma4/tt/generator.py
+++ b/models/demos/gemma4/tt/generator.py
@@ -1495,6 +1495,11 @@ class Gemma4Generator(ChunkedPrefillPageTableGuardMixin, Generator):
     model_capabilities = {
         "supports_prefix_caching": False,
         "supports_async_decode": False,
+        # Gemma4ModelArgs exposes no get_attn_sdpa_program_config, so the resume
+        # offset alignment cannot be derived. chunked_prefill_sdpa pins
+        # q_chunk_size=128 and documents that its base_offset must be a multiple
+        # of it.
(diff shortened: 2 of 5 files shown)


yieldthought:
Codex review:

[models/tt_transformers/tt/generator.py:251](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator.py#L251)

This gate treats the inherited base capability as if the concrete bridge had declared resume support. `Generator.model_capabilities` sets `supports_prefix_caching=True` at [models/tt_transformers/tt/generator.py:177](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator.py#L177), so [models/tt_transformers/tt/generator_sglang.py:247](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator_sglang.py#L247)'s `GptOssForCausalLM` enters this sweep even though it has no capability override. Its custom model args allow the 128-token ordinary trace on T3K/TG but reject nonzero cached-token traces at [models/demos/gpt_oss/tt/model_config.py:158](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/demos/gpt_oss/tt/model_config.py#L158), and expose neither `get_attn_sdpa_program_config` nor `resumed_prefill_token_alignment`. With the default `warmup_prefill=True`, `_warmup_prefill_resumed_sweep` therefore calls `_resume_offset_alignment` and raises `ValueError` before the first traced paged prefill completes; the zero-offset short-circuit does not help because this warmup calls the alignment helper directly. Suggestion: make GPT-OSS's SGLang bridge explicitly opt out of prefix caching (as its vLLM bridge already does), and only reserve a resumed variant when `can_enable_trace(prefill_seq_len, num_cached)` accepts that nonzero offset.

viktorpusTT:
> Codex review:
> 
> [models/tt_transformers/tt/generator.py:251](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator.py#L251)
> 
> This gate treats the inherited base capability as if the concrete bridge had declared resume support. `Generator.model_capabilities` sets `supports_prefix_caching=True` at [models/tt_transformers/tt/generator.py:177](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator.py#L177), so [models/tt_transformers/tt/generator_sglang.py:247](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/tt_transformers/tt/generator_sglang.py#L247)'s `GptOssForCausalLM` enters this sweep even though it has no capability override. Its custom model args allow the 128-token ordinary trace on T3K/TG but reject nonzero cached-token traces at [models/demos/gpt_oss/tt/model_config.py:158](https://github.com/tenstorrent/tt-metal/blob/4f791ff3c5b81a1d140ad61da75602f1105f6c27/models/demos/gpt_oss/tt/model_config.py#L158), and expose neither `get_attn_sdpa_program_config` nor `resumed_prefill_token_alignment`. With the default `warmup_prefill=True`, `_warmup_prefill_resumed_sweep` therefore calls `_resume_offset_alignment` and raises `ValueError` before the first traced paged prefill completes; the zero-offset short-circuit does not help because this warmup calls the alignment helper directly. Suggestion: make GPT-OSS's SGLang bridge explicitly opt out of prefix caching (as its vLLM bridge already does), and only reserve a resumed variant when `can_enable_trace(prefill_seq_len, num_cached)` accepts that nonzero offset.

Good catch, resolved.

yieldthought (approved):

mtairum (approved):

mtairum:
## Gemma 4: the resumed path looks live under the shipping scheduler config, and no CI leg reaches it

I went looking for evidence that the Gemma 4 side of this change is exercised anywhere, and pulled the server log from your own validation run [33065341377](https://github.com/tenstorrent/tt-metal/actions/runs/33065341377), job `Gemma4-31B [wh_llmbox]` ([98496332330](https://github.com/tenstorrent/tt-metal/actions/runs/33065341377/job/98496332330)).

### The green Gemma 4 legs do not reach the code

Full prefill sequence for that job:

```
Warming up prefill for sequence length: 32 / 128 / 512  ->  Prefill seq len: 128 / 128 / 1024, trace: False
warmup_gemma4_batched_prefill_traces: batch sizes (1,) x trace ISLs [128]
Prefill seq len: 128, trace: True  ->  Done Capturing Prefill Trace
... then 32 x "Prefilling User 1 up to 100 tokens"  ->  "Prefill seq len: 128"
```

`get_padded_prefill_len(100) == 128`, so `num_cached == 0` on all 32 serving requests and `_align_resume_offsets` short-circuits at `models/tt_transformers/tt/generator.py:1555` every time. Exactly one prefill trace is captured, at warmup, `sp0` only. The leg is green because the workload cannot reach the path, not because the path is covered.

### But the same log shows the mechanism is live

Engine config from that run:

```
enable_prefix_caching=False, enable_chunked_prefill=True
max_num_batched_tokens=2496, max_num_seqs=1
```

Chunked prefill is on at the engine level and the leg runs `--max_model_len 32768`. Any Gemma 4 prompt over 2496 tokens is split, and the continuation step starts at 2496. `2496 = 128*19 + 64` — not a multiple of the 128 that `chunked_prefill_sdpa` pins (`models/demos/gemma4/tt/attention/operations.py:286`). The benchmark runs `random_input_len=100`, which is why it never fires.

If the plugin surfaces a chunked continuation as a nonzero `start_pos`, then `resumed_prefill_token_alignment: 128` is not defensive hardening — it is fixing a live wrong-prefix read for every Gemma 4 prompt above 2496 tokens, on a path no CI leg has ever run.

### Two things worth resolving before merge

**1. The description argues both sides.** It says `Gemma4Generator.prefill_forward_text` forwards `start_pos` slices to the base implementation from its own batch-chunking loop — but that loop requires `all(n == 0 for n in num_cached_per_user)` (`models/demos/gemma4/tt/generator.py:362`), so it only ever forwards zeros. Meanwhile Gemma 4 declares `supports_prefix_caching: False` (`models/demos/gemma4/tt/generator.py:35`) and nothing in-tree declares `supports_chunked_prefill`, so the gate at `models/tt_transformers/tt/generator.py:253` is False and `_warmup_prefill_resumed_sweep` never runs for it. Either Gemma 4 resumes — in which case its `sp1` traces are captured lazily mid-serving, which is exactly the failure this PR's second half exists to remove — or it does not, and the declaration is dead config. The description currently asserts both.

**2. Both `Gemma4-31B` legs carry no `trace_region_size`** (the bare 50 MB default), while `warmup_gemma4_batched_prefill_traces` reserves `sp0` for ISL `[128]` only. If (1) resolves to "resumes", they need the same explicit sizing the three Llama legs got in this PR.

### Suggested run to settle it

Dispatch **vLLM Model Tests** on this branch — `model: google/gemma-4`, `sku: wh_llmbox (T3000)`, `tier: 2` — with one temporary edit to the `Gemma4-31B` [T3K] entry pushing the prompt past the chunk budget:

```yaml
  additional-benchmark-args: "--random-input-len 8000"
```

Plus one temporary line for an unambiguous signal, in `models/tt_transformers/tt/generator.py` right after `trace_key` is built (line 800):

```python
logger.info(f"Prefill trace key: {trace_key} num_cached_tokens={num_cached_tokens}")
```

The key already encodes the variant (`f"{prefill_seq_len}_{model_id}_{batch_size}_{use_start_pos}"`). Read the **Show server log** step and `grep "Prefill trace key"`.

- **`sp1` keys appear** -> Gemma 4 resumes under the shipping config. The 128 is load-bearing, the trace is being captured lazily on the 50 MB default, and both legs need explicit sizing. The same run against `main` should produce garbled output above 2496 tokens, which turns this into a regression test.
- **No `sp1` keys even at 8000 tokens** -> the plugin does not surface continuations as `start_pos`. The declaration is defensive; worth saying so and dropping the batch-chunking-loop justification.

Without the extra log line there is still a zero-edit signal: per request, compare `Prefilling User 1 up to N tokens` against the `Prefill seq len: P` that follows — `P != get_padded_prefill_len(N)` means that call carried a nonzero `start_pos`. Less crisp, since the worker may report the chunk end rather than the full prompt as `prompt_lens`.

One caveat on the whole thing: whether the TT worker turns a chunked-prefill continuation into a nonzero `start_pos` on `prefill_forward` lives in `tenstorrent/vllm-tt-plugin`, outside this repo. That is precisely what the run above settles.

---
<sub>Review assisted by Claude; findings verified against the tree and the linked CI logs.</sub>

viktorpusTT:
Good catch on the batch-chunking loop, that justification was wrong. `can_batch_prefill` requires `all(n == 0 for n in num_cached_per_user)`, so that loop only ever forwards zeros. Fixed the description.

The plugin question is answerable without a run. `model_runner.py:1159` in vllm-tt-plugin:

```python
input_positions = input_batch.num_computed_tokens_cpu[req_indices]
# - Chunked continuation: start_pos=computed, chunk=budget => computed + budget
```

So chunked continuations do arrive as a nonzero `start_pos`. Gemma 4 resumes, via the fall-through `super().prefill_forward_text(..., start_pos=start_pos)` rather than the batched path. The declaration is load-bearing: without it `_align_resume_offsets` raises on the first prompt past 2496.

On point 2, declaring `supports_chunked_prefill` would not help. Both Gemma 4 classes override `warmup_model_prefill` and never call `super()`, so the gate is unreachable for them whatever they declare, which is why your log shows `warmup_gemma4_batched_prefill_traces` capturing `sp0` only. So Gemma 4 does capture `sp1` lazily on the 50 MB default, and you are right that it is a risk. But it is that way on main too: main already aligns Gemma 4 locally (`align_num_cached_tokens_to_sdpa`) and already allows the resume. This PR neither causes it nor fixes it, so I would rather not size those legs here and hide a main-side gap.

Running your suggested job anyway, `--random-input-len 8000` on the Gemma 4 T3K leg, to see whether the lazy `sp1` capture actually blows the default. That is the part I cannot settle by reading. Will post the result.

Also rebased onto main. Two Gemma 4 conflicts: main's env-gated `supports_async_decode`, which I kept, and the new `SDPA_CHUNK_ALIGN`, which the declaration now uses instead of a bare 128.

viktorpusTT:
Probe run: https://github.com/tenstorrent/tt-metal/actions/runs/33500246606 (Gemma4-31B [wh_llmbox], 8000-token prompts, temporary trace-key log line on a throwaway branch).

Gemma 4 does resume. Four chunks per request, clearly:

```
Prefilling User 1 up to 2048 tokens  ->  Prefill seq len: 2048, trace: False
Prefilling User 1 up to 4096 tokens  ->  Prefill seq len: 2048, trace: False
Prefilling User 1 up to 6144 tokens  ->  Prefill seq len: 2048, trace: False
Prefilling User 1 up to 8000 tokens  ->  Prefill seq len: 2048, trace: False
```

Prefill seq len is the padded suffix, so a prompt end of 4096 with a 2048 suffix means start_pos=2048. Offsets were 0 / 2048 / 4096 / 6144. That matches model_runner.py:1159 and confirms the declaration is load-bearing.

But no sp1 trace is captured. Every chunked call logs trace: False. The only two trace keys in the whole run are 128_0_1_sp0 from warmup, since Gemma 4 traces ISL 128 only on T3K. So there is nothing to reserve and no lazy capture at serving time.

On the 2496: the engine used max_num_batched_tokens: 2048 in this run, which is 16*128. Every offset it produced was already a multiple of the pin, and there were zero SDPA chunk alignment and zero Resume offset alignment lines. So the misalignment did not fire here. I do not know where 2496 came from in your log; if you can point me at it I will look, but it is not what this leg configures.

That does not make the declaration optional. A deployment whose chunk budget is not a multiple of 128 still lands on a bad offset, and _align_resume_offsets runs on all four of those calls regardless of tracing, so without the declaration it raises. The protection is real, the CI leg just happens not to need it.

Run was clean: 4/4 successful, 0 failed, coherence guard passed, no errors.

dpopovTT (approved):

PR #53405, opened 2026-08-17
feat: implement qsr sfpu reduce

fvranicTT:
### PR Category
Feature

### Summary
Implement SFPU reduce MIN/MAX/SUM/AVG

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:filip/sfpu-reduce-qsr)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:filip/sfpu-reduce-qsr)

PR #54437, opened 2026-08-26
[Feature] Add Quasar SFPU reduce kernel and tests

### LINKED ISSUE #50161: [Quasar] SFPU reduce
Need to add SFPU reduce for full INT32 support in dest, since FPU reduce cannot accurately do INt32 reduce sum scalar. Can port BH SFPU reduction to Quasar with minimal changes.

@fvranicTT @nmohammedTT fyi

jmacanovicTT:
### Summary
Closes #50161, overlaps with #53405 (improves what it's missing)

Quasar had no SFPU reduce. The kernel existed for Blackhole and Wormhole,
but nothing equivalent on Quasar, so there was no way to reduce a tile on
the SFPU there — including Int32, which the FPU path cannot do.

### What's changed
Ported the SFPU reduce to Quasar and added tests at both layers.

**Kernel** — `ckernel_sfpu_reduce.h` for Quasar:
- SUM, AVG, MAX and MIN on both axes (integer AVG is column-only: a column
  always divides by 32, which a shift does exactly; a row divides by the
  runtime column count, which only the float reciprocal-multiply gets right)
- Float32, Float16_b, Float16 and Int32, with Int32 held in a 32-bit Dest
- MAX/MIN pick their winner by operand order on a single `SFPSWAP`, and
  floats use its fp32 compare so both-negative pairs need no correction

**LLK tests** — 858 functional variants and a 69-variant perf sweep,
covering both axes, all four pools, sub-tile column extents and the Int32
extremes where the comparator could misorder `INT32_MIN`/`INT32_MAX`.

**Metal test** — `test_sfpu_reduce.cpp`, 47 configs, driving the op through
the public Compute API (`sfpu_reduce`) rather than the LLK directly. This
covers the layer above: that a normal metal program can reach the op and
get the right answer. It needed three fixes to `compute_kernel_api.h`,
where `sfpu_reduce` was arch-guarded out of Quasar entirely and its
`static_assert` named a `DataFormat::UInt32` that Quasar does not have.

### Known gaps
Both are API limitations, documented in the test:
- **Int32 gets no multi-tile block.** It must unpack to Dest, and that path
  skips the datacopy loop that advances the Dest tile index, so every tile
  of a block lands on Dest tile 0. Float formats use `UnpackToSrc` and do
  get multi-tile.
- **Row AVG is unreachable.** `sfpu_reduce`'s `static_assert` allows
  `REDUCE_ROW` for SUM/MAX/MIN only, on every format — stricter than the
  LLK, which takes a float row AVG.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jmacanovic/quasar-sfpu-reduce)


tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,518 @@
...
+        constexpr std::uint32_t SWAP_IMM12 = IS_INT ? REDUCE_SWAP_IMM12_INT32 : REDUCE_SWAP_IMM12_FP32;
+
+        if constexpr (POOL_TYPE == PoolType::MAX) {
+            // DST as lreg_c: DST <- larger.
+            TTI_SFPSWAP(SWAP_IMM12, DST, SRC, p_sfpswap::ALL_ROWS_MAX);
+        } else {
+            // DST as lreg_dest: DST <- smaller. Same mode, operands reversed.
+            TTI_SFPSWAP(SWAP_IMM12, SRC, DST, p_sfpswap::ALL_ROWS_MAX);
+        }
+        TTI_SFPNOP(0 /* srcs_wr_done */, 0 /* srcs_rd_done */, 0 /* dest_done */);
```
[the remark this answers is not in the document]
jmacanovicTT: Did not improve performance. Skipping this suggestion.

fvranicTT (approved):

abhullar-tt (approved):

tt_metal/tt-llk/tests/python_tests/quasar/perf_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,106 @@
...
+    reduce_pool=PERF_POOLS,
+    formats=PERF_FORMATS,
+    mathop=get_supported_reduce_axes,
+    dimension_combinations=perf_input_dimensions,
+    implied_math_format=[ImpliedMathFormat.Yes],
+    run_types=PERF_RUN_TYPES_QUASAR,
+    loop_factor=[PERF_LOOP_FACTOR_QUASAR],
+    is_perf=[True],
+)
+def test_perf_sfpu_reduce_quasar(
```
mvlahovicTT, line 80: this perf test has no entry in `helpers/perf/test_schemas.py` `PERF_TEST_SCHEMAS_QSR`, so `test_perf_header_gate.py` fails on this branch: `"perf_sfpu_reduce_quasar: perf test found in source but missing from the catalog. Add an entry with its columns and version=1"`. a local perf run does not notice because `_assert_matches_catalog` skips unknown names, the static gate is what catches it and it will fail in ci. 

the entry needs `block_ct_dim`, `block_rt_dim`, `dest_acc`, `dest_sync`, `dst_index`, `face_c_dim`, `face_r_dim`, `formats.input_A`, `formats.input_B`, `formats.output`, `formats.register_A`, `formats.register_B`, `formats.sfpu_dst`, `formats.sfpu_src`, `full_ct_dim`, `full_rt_dim`, `implied_math_format`, `loop_factor`, `marker`, `mathop`, `num_faces`, `num_faces_A`, `num_faces_B`, `pool_type`, `tile_cnt`, `unpack_to_dest`, `unpacker_engine_sel` plus the `formats.sfpu_math` alias and `test_name_aliases` like the neighbours. all of those names already exist in `wide_schema_quasar.py` so nothing else is needed

tests/tt_metal/tt_metal/llk/sources.cmake
```diff
@@ -21,6 +21,7 @@ set(UNIT_TESTS_LLK_SRC
     test_reduce.cpp
     test_sfpu_binary_bcast.cpp
     test_sfpu_compute.cpp
+    test_sfpu_reduce.cpp
```
mvlahovicTT, line 24: as far as i can tell the metal test is compiled everywhere and executed nowhere. `quasar_regression_tests.yaml` has no unit_tests_llk group, the only unit_tests_llk block in `quasar_sim_regresion_tests.yaml` is commented out, and the two llk merge gate jobs that filter on `LLKQuasarMeshDeviceSingleCardFixture` run on wh and bh cards where the fixture GTEST_SKIPs. on the pr itself llk-unit-tests and ttsim-quasar-merge-gate-tests are skipped and llk-build-quasar only builds tt-llk. so neither the 66 metal configs nor the jit compile of `sfpu_reduce_quasar.cpp` have ci evidence. 

please add a unit_tests_llk group with a `TensixComputeSfpuReduce` filter to `quasar_regression_tests.yaml` (and the ttsim list if ttsim models SFPSWAP and SFPSHFT2), and say in the description where the metal configs and the llk sweep were actually run (ttsim or emulator)

tt_metal/tt-llk/tests/python_tests/quasar/test_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,557 @@
...
+
+    res_tensor = torch.tensor(res_from_L1, dtype=format_dict[formats.output_format])
+    res_tensor = untilize_block(res_tensor, formats.output_format, dst_dim)
+
+    if mathop == MathOperation.ReduceColumn:
+        golden_slice, res_slice = golden_tensor[0], res_tensor[0]
+    else:
+        golden_slice, res_slice = golden_tensor[:, 0], res_tensor[:, 0]
+
+    assert passed_test(
```
mvlahovicTT, line 436: for Max and Min get_reduce_atol returns None, so passed_test falls back to the default Tolerance(atol=0.05, rtol=0.05) plus a pcc gate a single wrong lane cannot move. that is 5 percent on an operation that should be exact: the stimuli are generated in the target dtype, the golden picks from those same quantized values and the device stores the chosen word back unchanged. i checked with the actual sweep stimuli (manual_seed 0, same shapes): a kernel returning the second largest value passes 17 to 18 of 32 rows on the [32,32] ReduceRow Max variants for every float format, and a fold bug that drops sfpu column 7 (what a lost SFPSHFT2 rotation would look like) passes 32 of 32 rows on the Float32 [32,32] row Max variant. the column axis is fine, a dropped SFPLOAD fails every shape. 

please pass custom_atol=0 and custom_rtol=0 for Max and Min (passed_test honours 0), or one format ulp as atol if Float16 in dest does not round trip bit exactly

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+// in the destination format, while MAX/MIN select an input unchanged and the integer pools are exact.
+double reduce_rtol(const SfpuReduceConfig& config) {
+    if (is_int_format(config.format)) {
+        return 0.0;
+    }
+    if (config.pool == ReducePool::Max || config.pool == ReducePool::Min) {
+        return 0.0;
+    }
+    return 0.05;
+}
```
mvlahovicTT: the Float32 tolerance ends up 4 to 5 orders of magnitude looser than the device. the computed atol is around 5e-6 for a column and then floored to 0.05, with a flat rtol of 0.05. the folds happen in fp32 lregs (SFPADD in ckernel_sfpu_reduce.h:156) so the real fp32 error is about 1e-6. a Float32 result that came back rounded to bf16 or tf32 precision passes every Float32 SUM and AVG config, and that is exactly the failure class the kernel depends on the metal layer to avoid: reduce_sfpmem_mode uses DEFAULT and lets the hardware resolve the dest word format from ALU_ACC_CTRL_SFPU_Fp32 and the srcb format register, which this test is the one covering. for AVG the stimulus is zero mean uniform(-4,4) so a typical result is 0.41 and the band is 0.07, a divisor of 30 or 34 instead of 32 is invisible. the comment at 288 to 291 is also not what this kernel does, partial sums never leave the lregs, only the final SFPSTORE rounds. suggest Float32 atol around 1e-4 times sqrt(N) with rtol 1e-5, Float16_b one ulp of the result, and drop the 0.05 floor the comment argues against

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+        if (format == tt::DataFormat::Int32) {
+            values[i] = static_cast<double>(std::bit_cast<std::int32_t>(packed[i]));
+        } else {
+            values[i] = static_cast<double>(std::bit_cast<float>(packed[i]));
+        }
+    }
+    return values;
+}
+
+// The input as the kernel actually sees it. Float16_b is already rounded by encode_elements and Int32
```
mvlahovicTT: this hard codes a truncating fp32 to tf32 conversion on the srca path for the Float32 golden. the quasar unpacker takes Float32 from l1 to srca only as tf32, fp16 or bf16 depending on the host programmed unpack dst format, and i could not find the rounding mode of that conversion written down anywhere. since the Float32 MAX and MIN configs compare with atol and rtol 0, a round to nearest conversion or a bf16 unpack dst format fails all 16 of them. one more reason the test needs a real quasar run before merge

tt_metal/hw/inc/api/compute/compute_kernel_api.h
```diff
@@ -690,54 +691,21 @@ ALWI void silu_tile_init_pack() { PACK(SFPU_UNARY_INIT_FN(silu, sfpu::silu_init,
...
-        (true /* APPROXIMATE */, is_fp32_dest_acc_en, num_rows, ITERATIONS, layout, accumulate),
-        idst,
-        idst_idx,
-        0 /* DST out unused, but required for _llk_math_eltwise_binary_sfpu_params_ */,
-        VectorMode::RC,
-        chunk)));
+// Formats the SFPU reduce kernel implements on this arch. Mirrors is_supported_reduce_format() in the
+// arch's ckernel_sfpu_reduce.h, which is only included on the MATH TRISC and so cannot be used here.
+#ifdef ARCH_QUASAR
+#define SFPU_REDUCE_SUPPORTED_FORMATS "Float32, Float16_b, Float16, Int32"
```
mvlahovicTT, line 697: Float16 is now reachable through sfpu_reduce on quasar and the description lists it as supported, but format_define throws on Float16 and datum_bytes would return 4 for it, so the newly opened api path has zero metal coverage. either add Float16 to format_define, datum_bytes, encode and decode, or say in the test header that Float16 is llk only

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+    switch (pool) {
+        case ReducePool::Sum: return "PoolType::SUM";
+        case ReducePool::Avg: return "PoolType::AVG";
+        case ReducePool::Max: return "PoolType::MAX";
+        case ReducePool::Min: return "PoolType::MIN";
+    }
+    TT_THROW("unreachable pool type");
+}
+
+std::string format_define(tt::DataFormat format) {
```
mvlahovicTT, line 82: Float16 is now reachable through sfpu_reduce on quasar and the description lists it as supported, but format_define throws on Float16 and datum_bytes would return 4 for it, so the newly opened api path has zero metal coverage. either add Float16 to format_define, datum_bytes, encode and decode, or say in the test header that Float16 is llk only

tt_metal/tt-llk/tests/python_tests/quasar/test_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,557 @@
...
+    NUM_FACES,
+    TEST_FACE_DIMS,
+    TILE_COUNT,
+    UNPACKER_ENGINE_SEL,
+    generate_input_dim,
+)
+from helpers.tilize_untilize import tilize_block, untilize_block
+from helpers.utils import passed_test
+
+MAX_TILES_16_BIT = 8
```
mvlahovicTT, line 42: the description says 858 functional variants but collect-only gives 2844 (2820 plus the 24 int32 extreme cases). 858 is what the sweep produced with a 4 tile ceiling for every format, so the number predates MAX_TILES_16_BIT = 8 and the doubling under IMPLIED_MATH_FORMATS. implied_math_format only changes the unpack to dest hw_configure and never touches the sfpu instruction stream, test_reduce_quasar.py pins it to Yes for most formats rather than sweeping it, so either trim the multipliers or update the description

tt_metal/tt-llk/tests/python_tests/quasar/test_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,557 @@
...
+]
+
+# Relative precision per float format: bf16 has 7 explicit mantissa bits, fp16 10, fp32 23.
+_FLOAT_FORMAT_EPS = {
+    DataFormat.Float16_b: 2.0**-8,
+    DataFormat.Float16: 2.0**-11,
+    DataFormat.Float32: 2.0**-24,
+}
+
+IMPLIED_MATH_FORMATS = [ImpliedMathFormat.No, ImpliedMathFormat.Yes]
```
mvlahovicTT: the description says 858 functional variants but collect-only gives 2844 (2820 plus the 24 int32 extreme cases). 858 is what the sweep produced with a 4 tile ceiling for every format, so the number predates MAX_TILES_16_BIT = 8 and the doubling under IMPLIED_MATH_FORMATS. implied_math_format only changes the unpack to dest hw_configure and never touches the sfpu instruction stream, test_reduce_quasar.py pins it to Yes for most formats rather than sweeping it, so either trim the multipliers or update the description

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+inline void reduce_combine() {
+    static_assert(DST <= p_sfpu::LREG7 && SRC <= p_sfpu::LREG7, "reduce may only write LREG0-7");
+    static_assert(DST != SRC, "reduce_combine needs two distinct registers");
+
+    if constexpr (POOL_TYPE == PoolType::MAX || POOL_TYPE == PoolType::MIN) {
+        constexpr std::uint32_t SWAP_IMM12 = IS_INT ? REDUCE_SWAP_IMM12_INT32 : REDUCE_SWAP_IMM12_FP32;
+
+        if constexpr (POOL_TYPE == PoolType::MAX) {
+            // DST as lreg_c: DST <- larger.
+            TTI_SFPSWAP(SWAP_IMM12, DST, SRC, p_sfpswap::ALL_ROWS_MAX);
```
mvlahovicTT, line 128: the golden docstring says the MAX/MIN reduce is a bare SFPSWAP with no nan guard so the sfpu total order reaches the result and +nan outranks every finite value. the quasar rtl does the opposite: with imm12 bit 0 set the swap condition is fp_smaller and not srcb_nan and not srcc_nan, so a nan on either side means no swap and the value already in the DST register survives. in the column tree the DST chain is LREG0 row 0, so a column max gives nan only if the row 0 element is nan and a nan in rows 1 to 31 is silently dropped. no stimulus contains nan so nothing fails today, but the golden models something the hardware does not do. please either document nan input as unspecified and position dependent in the kernel and fix the docstring, or add a nan guard. same story for signed zero, the rtl treats -0 and +0 as equal so the DST side sign wins, the golden ranks -0 below +0

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+    } else {
+        // SFPADD is dest = a * b + c, so a = DST and b = 1.0 makes it dest = DST + SRC.
+        TTI_SFPADD(DST, p_sfpu::LCONST_1, SRC, DST, 0 /* instr_mod1: no negation */);
+    }
+}
+
+/**
+ * @brief Divide LREG0 by the 32-row column extent for an integer AVG, in place.
+ *
+ * A right shift by 5 divides by 32 - but rounds the wrong way. On a negative two's-complement
```
mvlahovicTT: "both a logical shift and Quasar's arithmetic shift round toward negative infinity" is not right for the logical shift, which shifts zeros into a negative two's complement value and yields a large positive garbage value. only the arithmetic shift rounds toward negative infinity. the code is fine (shift the magnitude, restore the sign), only the justification is off

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+ * @brief Prepare the math thread's SFPU state for a run of reduce calls.
+ *
+ * Only one thing to do: zero the RWC counters. The reduce addresses Dest with immediates measured
+ * from the base @ref _llk_math_eltwise_sfpu_start_ programs, so they must start from zero.
+ *
+ * The rest - SFPU config register, ADDR_MOD_7 - is already set up by
+ * @ref _llk_math_eltwise_sfpu_init_, and no reduce path claims a programmable constant register or
+ * a replay slot, so this cannot disturb a neighbouring op's setup.
+ *
+ * The reverse is not guaranteed. The folds read LCONST_0 / LCONST_1 (LREG9/10) at their reset
```
mvlahovicTT, line 405: "only one thing to do: zero the rwc counters ... so they must start from zero" is redundant and the reason is not the mechanism. llk_math_eltwise_sfpu_init already does reset_counters<SET_ABD_F> and the wrapper calls it right before init_reduce, and the SFPLOAD immediates are relative to the dest section base written by set_dst_write_addr (a cfg write, not the rwc). harmless, but either drop the call or fix the wording

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+    static_assert(
+        is_supported_reduce_format(FORMAT), "Unsupported reduce format: expected Float32, Float16_b, Float16 or Int32");
+    static_assert(
+        REDUCE_DIM == ReduceDim::REDUCE_COL || REDUCE_DIM == ReduceDim::REDUCE_ROW,
+        "Unsupported reduce_dim: expected REDUCE_COL or REDUCE_ROW");
+    static_assert(
+        !reduce_is_32_bit_format<FORMAT>() || IS_FP32_DEST_ACC_EN,
+        "a 32-bit reduce format requires a 32-bit Dest (IS_FP32_DEST_ACC_EN)");
+    // A column AVG always divides by 32, which an integer can do with a shift
+    // (@ref reduce_int_average_col). A row AVG divides by the runtime column count - rarely a power
```
mvlahovicTT, line 472: 1/num_cols is only exact for power of two widths. for 96, 160 or 224 columns the fp32 reciprocal is rounded and x times 1/n differs from x/n by up to one ulp, so "divides it exactly" should read "to within fp32 rounding". also Int32 SUM wraps silently mod 2^32 (the rtl takes iadd_res[31:0]) and neither the kernel doc nor the tests mention it, one line in the brief would be good since Int32 is the headline feature

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+    experimental::ProgramSpec spec{
+        .name = "sfpu_reduce",
+        .kernels = {reader_spec, writer_spec, compute_spec},
+        .dataflow_buffers = {in_dfb_spec, out_dfb_spec},
+        .work_units = {wu},
+    };
+
+    Program program = experimental::MakeProgramFromSpec(*mesh_device, spec);
+
+    const std::vector<double> input = generate_stimulus(
```
mvlahovicTT: the stimulus seed comes from system_clock and is never logged, so a MAX/MIN exact compare failure could not be replayed. please log the seed in the mismatch message

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+        ASSERT_TRUE(ok) << fmt::format(
+            "mismatch at {}: golden={} device={} (atol={} rtol={})", i, golden[i], device[i], atol, rtol);
+    }
+}
+
+}  // namespace unit_tests::compute::sfpu_reduce
+
+using namespace unit_tests::compute::sfpu_reduce;
+
+// Column reduce: every pool, including the integer AVG that divides by 32 with a shift.
```
mvlahovicTT: this comment says holding a multi tile block in dest and reducing each tile in place is a separate path that does not work from here yet, but the configs right below run exactly that (block_ct_dim 2 column reduce for both float formats). also the description says 47 configs and the head runs 66

tt_metal/tt-llk/tests/python_tests/quasar/perf_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,106 @@
...
+    one extra row-axis case: both factors of row_base = rt * block_ct_dim * REDUCE_TILE_STRIDE
+    exceed one there. The column axis reduces tile by tile, so its rows across these shapes are
+    knowingly duplicates.
+    """
+    dest_acc = (
+        DestAccumulation.Yes
+        if formats.input_format.is_32_bit()
+        else DestAccumulation.No
+    )
+    return generate_perf_input_dimensions(dest_acc, DestSync.Half) + [
```
mvlahovicTT, line 42: if blackhole parity is a goal for this perf data, add [TILE_DIM, TILE_DIM] here. the bh perf_sfpu_reduce is a single 32x32 Float32 ReduceRow Max variant, so with only the [128,32], [32,128] and [64,64] shapes there is no like for like point, and the only comparable number would be per tile MATH_ISOLATE TILE_LOOP after the loop_factor division

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+#include "tt_metal/test_utils/packing.hpp"
+
+// Metal-layer coverage for the Quasar SFPU reduce, driven through the public Compute API rather
+// than the LLK directly. The LLK sweep in tt-llk covers the kernel itself; this covers the layer
+// above -- that a normal metal program can reach the op and gets the right answer back.
+//
+// The kernel writes only the axis it collapses and leaves the rest of the tile holding fold
+// leftovers, so the comparison reads just that axis, which is what a reduce's consumers read:
+//   * REDUCE_COL folds each tile's 32 rows onto its row 0; tiles reduce independently.
+//   * REDUCE_ROW folds a tile row's columns onto column 0 of that row, spanning every tile in it.
```
mvlahovicTT: are these comments necessary, can they be reduced/removed?

tests/tt_metal/tt_metal/llk/test_sfpu_reduce.cpp
```diff
@@ -0,0 +1,627 @@
...
+}
+
+bool is_int_format(tt::DataFormat format) { return format == tt::DataFormat::Int32; }
+
+// How the operand reaches Dest, which is what decides whether a multi-tile block can be staged.
+//
+// UnpackToSrc goes through SrcA and the math-side datacopy, whose per-tile loop advances the Dest
+// index, so copy_block gives each tile its own Dest tile. UnpackToDest skips that loop
+// (llk_math_eltwise_unary_datacopy_block: "math is a sync-only forwarder") and every tile lands on
+// Dest tile 0, overwriting the last. Int32 must unpack to Dest, so Int32 cannot use a multi-tile block.
```
mvlahovicTT: are these comments necessary, can they be reduced/removed?

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+// A 32x32 tile is not a flat square in Dest but four 16x16 faces: faces 0/1 side by side over
+// rows 0-15, faces 2/3 the same over rows 16-31. One address holds one face row (16 datums), so
+// face f starts at 16f - the four faces sit at 0, 16, 32 and 48.
+//
+// One SFPLOAD reads four consecutive face rows, and address bit 1 picks the even or odd 8
+// columns of each - 4 rows x 8 columns = 32 datums, exactly the SFPU's 4 rows x 8 columns.
+//
+// Both axes load four at a time, but group them differently:
+//   column group - one face, one column parity: 8 columns, that face's 16 rows.
+//   row quad     - four whole tile rows: both parities of two side-by-side faces.
```
mvlahovicTT: are these comments necessary, can they be reduced/removed?

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+ * @brief SFPLOAD/SFPSTORE format-select code for a reduce operand format.
+ *
+ * Floats take DEFAULT, letting the hardware resolve the Dest word format at runtime from
+ * ALU_ACC_CTRL_SFPU_Fp32 and the SrcB format register - both already programmed from formats.math.
+ * One kernel body therefore covers Float32, Float16_b and Float16 in either Dest width.
+ *
+ * Int32 names its mode outright: implied formats are unreliable under unpack-to-dest (TEN-4674).
+ *
+ * @tparam FORMAT: Math-side data format of the reduce operand
+ */
```
mvlahovicTT, line 78: are these comments necessary, can they be reduced/removed?

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+ * larger in its lreg_c, so the operand order alone decides which @p DST keeps. The compare domain
+ * comes from @ref REDUCE_SWAP_IMM12_INT32 / @ref REDUCE_SWAP_IMM12_FP32.
+ *
+ * @tparam POOL_TYPE: Reduction operator, values = <SUM/AVG/MAX/MIN>
+ * @tparam IS_INT: Whether the operands are integers (selects the adder and the compare domain)
+ * @tparam DST: LREG holding the running result; overwritten with the folded value
+ * @tparam SRC: LREG holding the incoming value; clobbered
+ * @note Every MAX/MIN fold ends in an SFPNOP: SFPSWAP takes 2 cycles and its result must not be
+ *       read by the next instruction (the SFPSWAP -> SFPSTORE auto-stall bug). That trailing NOP
+ *       is what makes this safe back-to-back and directly before a store.
```
mvlahovicTT: are these comments necessary, can they be reduced/removed?

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+ * r+4, r+8, r+12 at column c. Those quarters are stacked in LREG0's rows, out of reach - which is
+ * what SFPTRANSP is for. It spreads them into row 0 of LREG0-3, where the ordinary fold finishes.
+ *
+ * @tparam POOL_TYPE: Reduction operator, values = <SUM/AVG/MAX/MIN>
+ * @tparam MODE: SFPLOAD/SFPSTORE format-select code, @ref reduce_sfpmem_mode
+ * @tparam IS_INT: Whether operands are integers
+ * @tparam TOP_FACE_ADDR: Dest address of the top face of the pair, values = <0/REDUCE_FACE_STRIDE>
+ * @tparam COLUMN_OFFSET: Column parity selector, values = <EVEN_COL/ODD_COL>
+ * @note Writes the totals to row 0 of the top face; rows 1-3 keep the transpose's leftovers, which
+ *       is fine - row 0 is all a column reduce's consumers read.
```
mvlahovicTT: are these comments necessary, can they be reduced/removed?

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
```diff
@@ -0,0 +1,523 @@
...
+// amount, so shifting right means passing a negative one.
+constexpr std::uint32_t REDUCE_AVG_SHIFT_LOG2 = 5;
+constexpr std::uint32_t REDUCE_AVG_SHIFT_IMM12 = static_cast<std::uint32_t>(-REDUCE_AVG_SHIFT_LOG2) & 0xFFF;
+constexpr std::uint32_t REDUCE_AVG_SHFT_MOD_LOGICAL_IMM = 0x1;  // amount from imm12, logical shift
+constexpr std::uint32_t REDUCE_SETCC_IMM12_INT32 = 0x0;         // read src_c as two's-complement int32
+constexpr std::uint32_t REDUCE_SETCC_MOD_NEGATIVE = 0x0;        // set CC where src_c is negative
+constexpr std::uint32_t REDUCE_IADD_MOD_NEGATE_KEEP_CC = 0x6;   // dest = lreg_c - dest, leaving CC alone
+static_assert(REDUCE_COL_EXTENT == (1U << REDUCE_AVG_SHIFT_LOG2), "the integer column AVG shift must divide by 32");
+
+// SFPSWAP imm12 bit 0 picks how the comparison reads its operands (instructions/assembly.yaml):
```
mvlahovicTT, line 62: this relies on imm12 bit 0 = 0 being a two's complement int32 compare, which is what assembly.yaml and the a0 rtl say (tt_sfpu.sv does a 33 bit signed subtract). ckernel_sfpu_binary_max_min.h:55-58 says the same mode is an unsigned compare on the lreg bits and applies a cc guarded correction swap to both negative pairs for Int32 as well as floats. on that rtl the correction would invert correctly ordered both negative int32 pairs, so its Int32 test and test_int32_reduce_extreme all_negative cannot both pass on the same model. this kernel is the one consistent with the rtl, but it is worth settling with an emulator run of the all negative MAX/MIN variants of both kernels, since the quasar llk tests are excluded from tt-metal ci

mvlahovicTT (commented):
please go over the comments in the new files once more and cut the ones that are overly verbose or that restate what the code already says. a lot of the doc blocks in ckernel_sfpu_reduce.h and test_sfpu_reduce.cpp explain intent that is obvious from the line below them, and a few are wrong now (the shift rounding note, the init_reduce reason, the multi tile block note in the metal test)

and make sure these tests are passing and the relevant ci are green on the final head before this goes in

tt_metal/tt-llk/tests/python_tests/quasar/perf_sfpu_reduce_quasar.py
```diff
@@ -0,0 +1,106 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
```
ndivnicTT, line 4: Can you please run metal/tt-metal/tt_metal/tt-llk/tests/python_tests/test_perf_*.py locally and fix eventual failures?

You will need to update metal/tt-metal/tt_metal/tt-llk/tests/python_tests/helpers/perf/wide_schema_quasar.py

ndivnicTT (approved):
LGTM!
Please address comment before merging

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_reduce.h
jmacanovicTT: confirmed on the emulator today, both the reduce's Int32 guard and sweep and the binary max/min Int32 test pass on the same model. They agree with the RTL's two's-complement compare; they differ in operand encoding. The reduce and the metal layer use two's-complement Int32, the binary test stages sign-magnitude via the harness default, which under a signed compare inverts exactly the both-negative pairs its correction swap fixes. The "unsigned compare" wording in that kernel's comment is inaccurate and worth a follow-up there.

tt_metal/tt-llk/tests/python_tests/quasar/perf_sfpu_reduce_quasar.py
jmacanovicTT: Ran them, our schema entry passes, the 4 remaining failures are pre-existing on main and come from the matmul entry and the report tool tests.

tests/tt_metal/tt_metal/llk/sources.cmake
jmacanovicTT: @fvranicTT could you check up on this one please? This is connected to quasar CI.

fvranicTT: These tests are currently being run manually, and not as a regular CI, due to lack of hardware. This is not a new issue, and there's nothing to be solved in this PR.

PR #54443, opened 2026-08-26
[Models CI] Migrate the Galaxy TTTv2 2D module tests to tier 1 Models unit CI

mtairum:
**CI run** — [(Tier 1) Models Unit Tests 33078550091](https://github.com/tenstorrent/tt-metal/actions/runs/33078550091) on `f1ef37038ad`, `model: tt-transformers-common`, `sku: all`.

One dispatch covers every entry this PR touches — all 11 `tt-transformers-common` entries:

| SKU | Entries | What it proves |
|---|---|---|
| `wh_galaxy_perf` | 3 | the migrated 2D entries pass, with the MLP filter narrowed to the two `from_model_args` failures |
| `wh_llmbox` | 8 | the eight 1D entries still run there under the new cluster-type gate |

> `TT-Transformers common unit tests [wh_llmbox]` will fail. That is **not** this PR: it has timed out at its 15-minute limit on `main` for at least four consecutive scheduled runs (32929600689, 32893573236, 32808313730, 32772260578). Flagged separately.

<details><summary>Superseded run on <code>4123f80d225</code></summary>

| SKU | Run |
|---|---|
| `wh_galaxy_perf` | [32984785776](https://github.com/tenstorrent/tt-metal/actions/runs/32984785776) |
| `wh_llmbox` | [32984794342](https://github.com/tenstorrent/tt-metal/actions/runs/32984794342) |

Predates the rebase onto main, the `-k` narrowing, and the switch to `ttnn.cluster.get_cluster_type()`.

</details>

<details><summary>Earlier round, and what it settled</summary>

The first attempt folded the 2D calls into the existing 1D entries. It failed on **both** SKUs, and the logs showed why — every pytest call carries `--cov` with `fail_under = 80` from `models/common/tests/setup.cfg`, so whichever call skips on a given SKU fails the gate despite zero test failures:

| | 1D call | 2D call | Job |
|---|---|---|---|
| [`wh_galaxy_perf`](https://github.com/tenstorrent/tt-metal/actions/runs/32980799692) | 27 / 28 skipped → cov 15.96% / 15.31% ❌ | passed → cov 87.91% / 91.82% ✅ | fail |
| [`wh_llmbox`](https://github.com/tenstorrent/tt-metal/actions/runs/32980818155) | 27 / 28 passed → cov 90.43% / 89.86% ✅ | skipped → cov 21.98% / 26.67% ❌ | fail |

The `auto-compose` entry, the one already structured separately, passed on Galaxy in that same run — and the five other untouched 1D entries passed on `wh_llmbox`. That is what settled the structure.

</details>

---

### Ticket

N/A — CI pipeline migration, continuation of #53911, #54184, #54306 and #54324.

### Problem description

`Galaxy tttv2 tests` was the only `team: models` entry in `(Galaxy) unit tests`. It runs the **2D** TTTv2 modules — the same family and the same owner as the **1D** suites that #53911 moved out of the T3K pipelines:

```
models/common/tests/modules/rmsnorm/test_rmsnorm_2d.py
models/common/tests/modules/mlp/test_mlp_2d.py
models/common/tests/test_auto_compose.py
```

Nothing in any tiered registry referenced those three files, so this is a genuine migration with an obvious destination. Unlike the demo and perf pipelines, `(Galaxy) unit tests` does **not** empty out — its other five jobs belong to `umd`, `scaleout` and `ttnn` — so the pipeline stays and only the models entry and its dispatch option leave.

### Three separate entries

| Entry | SKU | Timeout | Runs |
|---|---|---|---|
| `TT-Transformers RMSNorm 2D module unit tests` | `wh_galaxy_perf` | 5 min | `test_rmsnorm_2d.py` |
| `TT-Transformers MLP 2D module unit tests` | `wh_galaxy_perf` | 5 min | `test_mlp_2d.py`, `-k "not (test_mlp_2d_vs_reference_from_model_args and 8x4)"` |
| `TT-Transformers auto-compose 2D unit tests` | `wh_galaxy_perf` | 5 min | `test_auto_compose.py` |

All three share `model: tt-transformers-common` / `model_family: TTTv2` and the existing owner, so one dispatch filter still covers the family. `wh_galaxy_perf` was already in the tier-1 unit workflow's `ALL_SKUS`, so no workflow wiring was needed. The eight 1D entries are untouched and stay `wh_llmbox`-only.

**Why not extra calls on the existing 1D entries.** That was tried first and CI rejected it — see the collapsed section above. One `cmd` runs on every SKU an entry declares, and each call's `--cov … fail_under = 80` then fails on whichever SKU its suite skips. The two escapes were both worse: `--cov-fail-under=0` would disable enforcement on `wh_llmbox` too, where those gates currently read 90.43% and 89.86%; branching on device count inside `cmd` would reimplement SKU dispatch in shell. Splitting needs no new machinery, and matches how #53911 split this family into eight entries rather than one.

**`test_mlp_2d.py` is migrated with two cases excluded.** `test_mlp_2d_vs_reference_from_model_args[32-8x4]` and `[512-8x4]` hit `TT_FATAL @ tt_metal/fabric/fabric.cpp:174 forwarding_direction` on a Galaxy; the `4x8` orientation passes. That is a pre-existing defect in the migrated test rather than a migration artefact. The filter names those two cases specifically rather than a bare `not 8x4`, which would also have dropped the two passing `test_mlp_2d_vs_reference` cases and `test_ttnn_linear_2d_mesh_topology_bug[8x4]` — an `xfail(strict=True)` canary whose whole job is to fail when the TTNN topology bug gets fixed. The exclusion is marked inline with the reason so it is not mistaken for a tuning choice. It wants an issue against the 8×4 fabric path.

### Hardware gating, both directions

**2D outside a Galaxy: already handled, no change made.** The 2D suites request `(4,8)` and `(8,4)`. Neither is in the allowed set for a `(2,4)` system, so on a T3K they already skip with `Requested mesh (4, 8) unsupported on system (2, 4)`.

**1D outside a T3K: needed a gate.** The eight 1D suites now opt into a new fixture via a module-level `pytestmark`:

```python
# 1D module suites target the T3K; skip when the host system is a Galaxy.
pytestmark = pytest.mark.usefixtures("skip_on_galaxy_system")
```

Two deliberate choices in that fixture:

- **Opt-in per module, not a change to `_allowed_req_shapes_for_system`.** Removing 1D shapes from the `(8,4)` allow-list would have been a one-line fix, but `test_auto_compose.py` deliberately parametrises `1x1, 1x2, 1x4, 1x8, 2x4, 4x8, 8x4` and expects the 1D shapes to run on a Galaxy. A global rule would have silently killed that coverage.
- **Queried at fixture setup, not collection time.** Probing the device while collecting has deadlocked nested-pytest runs in this repo before, so the check runs at the same point as the existing `ttnn_mesh_device` fixture’s own system lookup.
- **Fingerprinted by cluster type, not device count.** `ttnn.cluster.get_cluster_type()` is checked against `{GALAXY, TG, BLACKHOLE_GALAXY}`, matching `models/tt_transformers/tt/model_config.py:545-549` and the note in `_allowed_req_shapes_for_system` that a `ClusterType` is the robust fingerprint where a mesh-shape tuple is not. A hand-wired 32-chip rig reports `ClusterType.CUSTOM` and so is not matched; an indeterminate cluster type does not skip.

The fixture returns without skipping if the system cannot be determined, leaving the decision to the mesh fixture rather than failing closed on an unrelated error.

The guard is independently worthwhile even with the split: the 1D entries are `wh_llmbox`-only today, and this stops a future SKU addition from quietly widening them onto a Galaxy. The first CI round confirmed it works — the 1D suites skipped on Galaxy with `T3K-targeted; system mesh is (8, 4)`, and on `wh_llmbox` all 27 `rmsnorm_1d` and 28 `mlp_1d` tests passed unchanged.

### `test_auto_compose.py` is not duplicated

`TT-Transformers common unit tests` already collects it on `wh_llmbox` via the directory glob, covering its `1x1` through `2x4` cases. The new Galaxy entry adds the `4x8` and `8x4` cases, which skip on an LLMBox. Same file, disjoint case sets.

### Source side

`Galaxy tttv2 tests` is removed from `tests/pipeline_reorg/galaxy_unit_tests.yaml` and `tttv2` from `(Galaxy) unit tests`' dispatch enum. The pipeline itself stays — its five remaining jobs belong to `umd`, `scaleout` and `ttnn`.

`galaxy-unit-tests` is one of the workflows the silencer can dispatch, so `silencer.lock.yml` embeds that enum and was regenerated with `gh aw compile` v0.86.2 — the version named in the committed lock's own metadata. The diff is the single enum entry.

### Budget

| Team / pipeline / SKU | Was | Now | Used |
|---|---|---|---|
| `models.unit_tier1.wh_galaxy_perf` | 45 | 60 | 55 |
| `models.unit.wh_galaxy_perf` | 10 | removed | — |

The source key is removed rather than zeroed: this job was its entire 10-minute allocation, and no models entry verifies against `models.unit` on that SKU any more. A future entry needing it fails with `Could not find a 'time_budget'`, which is a clearer error than exceeding a 0-minute grant — the same reasoning applied to `models.perf` in #54324.

Verified with the CI check itself: `models_unit_tests.yaml` passes at unit tiers 1, 2 and 3 with `wh_galaxy_perf` at 55/60, and `galaxy_unit_tests.yaml` and `t3k_unit_tests.yaml` both still pass.

### Testing

Not dispatched yet — the three entries land on `wh_galaxy_perf`, and Galaxy capacity has been the bottleneck across this migration series. Worth one run of `(Tier 1) Models Unit Tests` with `model: tt-transformers-common`, `sku: wh_galaxy_perf (WH Galaxy perf)` to confirm the three new entries, plus one on `wh_llmbox` to confirm the `pytestmark` addition did not disturb the eight 1D entries.

### Checklist

- [ ] (Tier 1) Models Unit Tests — `model: tt-transformers-common`, `sku: wh_galaxy_perf (WH Galaxy perf)` — the three 2D entries
- [ ] (Tier 1) Models Unit Tests — `model: tt-transformers-common`, `sku: wh_llmbox (T3000)` — the 1D entries unchanged
- [ ] File an issue for the `test_mlp_2d.py` 8×4 fabric fatal
- [ ] (Galaxy) unit tests — dispatch to confirm the removed `tttv2` option left the pipeline valid

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mtairum/migrate-galaxy-tttv2-2d)

### DIFF
--- a/tests/pipeline_reorg/models_unit_tests.yaml
+++ b/tests/pipeline_reorg/models_unit_tests.yaml
@@ -1751,4 +1751,3 @@
       tier: 1
   owner_id: U08JFM3CFA4 # Gongyu Wang
   team: models
-
--- a/tests/pipeline_reorg/models_unit_tests.yaml
+++ b/tests/pipeline_reorg/models_unit_tests.yaml
@@ -1728,15 +1728,16 @@
   team: models
 
 - name: TT-Transformers MLP 2D module unit tests
-  # 8x4 excluded: test_mlp_2d_vs_reference_from_model_args[32-8x4] and [512-8x4] hit
-  # TT_FATAL @ tt_metal/fabric/fabric.cpp:174 forwarding_direction. The 4x8 orientation passes.
+  # Only the two from_model_args 8x4 cases are excluded: they hit TT_FATAL @
+  # tt_metal/fabric/fabric.cpp:174 forwarding_direction. A bare "not 8x4" would also drop
+  # the two passing direct-API cases and the strict-xfail topology canary.
   cmd: |
     unset TT_METAL_WATCHER
     export HF_HOME=/mnt/MLPerf/huggingface
     HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
     pytest models/common/tests/modules/mlp/test_mlp_2d.py \
       -m "not slow" \
-      -k "not 8x4" \
+      -k "not (test_mlp_2d_vs_reference_from_model_args and 8x4)" \
       --tb=short \
       --durations-min=3.0 \
       --cov=models.common.modules.mlp.mlp_2d \
--- a/tests/pipeline_reorg/models_unit_tests.yaml
+++ b/tests/pipeline_reorg/models_unit_tests.yaml
@@ -1550,6 +1550,7 @@
   cmd: |
     unset TT_METAL_WATCHER
     export HF_HOME=/mnt/MLPerf/huggingface
+    fail=0
     HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
     TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/tttv2/mlp_1d \
     pytest models/common/tests/modules/mlp/test_mlp_1d.py \
@@ -1558,12 +1559,24 @@
       --durations=10 \
       --cov=models.common.modules.mlp.mlp_1d \
       --cov-report=term-missing \
-      --cov-config=models/common/tests/setup.cfg
+      --cov-config=models/common/tests/setup.cfg || fail=1
+    HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
+    pytest models/common/tests/modules/mlp/test_mlp_2d.py \
+      -m "not slow" \
+      --tb=short \
+      --durations-min=3.0 \
+      --cov=models.common.modules.mlp.mlp_2d \
+      --cov-report=term-missing \
+      --cov-config=models/common/tests/setup.cfg || fail=1
+    exit $fail
   model: tt-transformers-common
   model_family: TTTv2
   skus:
     wh_llmbox:
       timeout: 6
+      tier: 1
+    wh_galaxy_perf:
+      timeout: 5
       tier: 1
   owner_id: U08JFM3CFA4 # Gongyu Wang
   team: models
@@ -1572,6 +1585,7 @@
   cmd: |
     unset TT_METAL_WATCHER
     export HF_HOME=/mnt/MLPerf/huggingface
+    fail=0
     HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
     TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/tttv2/rmsnorm_1d \
     pytest models/common/tests/modules/rmsnorm/test_rmsnorm_1d.py \
@@ -1580,12 +1594,24 @@
       --durations=10 \
       --cov=models.common.modules.rmsnorm.rmsnorm_1d \
       --cov-report=term-missing \
-      --cov-config=models/common/tests/setup.cfg
+      --cov-config=models/common/tests/setup.cfg || fail=1
+    HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
+    pytest models/common/tests/modules/rmsnorm/test_rmsnorm_2d.py \
+      -m "not slow" \
+      --tb=short \
+      --durations-min=3.0 \
+      --cov=models.common.modules.rmsnorm.rmsnorm_2d \
+      --cov-report=term-missing \
+      --cov-config=models/common/tests/setup.cfg || fail=1
+    exit $fail
   model: tt-transformers-common
   model_family: TTTv2
   skus:
     wh_llmbox:
       timeout: 3
+      tier: 1
+    wh_galaxy_perf:
+      timeout: 5
       tier: 1
   owner_id: U08JFM3CFA4 # Gongyu Wang
   team: models
@@ -1702,48 +1728,6 @@
   skus:
     wh_llmbox:
       timeout: 4
-      tier: 1
-  owner_id: U08JFM3CFA4 # Gongyu Wang
-  team: models
-
-- name: TT-Transformers RMSNorm 2D module unit tests
-  cmd: |
-    unset TT_METAL_WATCHER
-    export HF_HOME=/mnt/MLPerf/huggingface
-    HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
-    pytest models/common/tests/modules/rmsnorm/test_rmsnorm_2d.py \
-      -m "not slow" \
-      --tb=short \
-      --durations-min=3.0 \
-      --cov=models.common.modules.rmsnorm.rmsnorm_2d \
-      --cov-report=term-missing \
-      --cov-config=models/common/tests/setup.cfg
-  model: tt-transformers-common
-  model_family: TTTv2
-  skus:
-    wh_galaxy_perf:
-      timeout: 5
-      tier: 1
-  owner_id: U08JFM3CFA4 # Gongyu Wang
-  team: models
-
-- name: TT-Transformers MLP 2D module unit tests
-  cmd: |
-    unset TT_METAL_WATCHER
-    export HF_HOME=/mnt/MLPerf/huggingface
-    HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
-    pytest models/common/tests/modules/mlp/test_mlp_2d.py \
-      -m "not slow" \
-      --tb=short \
-      --durations-min=3.0 \
-      --cov=models.common.modules.mlp.mlp_2d \
-      --cov-report=term-missing \
-      --cov-config=models/common/tests/setup.cfg
-  model: tt-transformers-common
-  model_family: TTTv2
-  skus:
-    wh_galaxy_perf:
-      timeout: 5
       tier: 1
   owner_id: U08JFM3CFA4 # Gongyu Wang
   team: models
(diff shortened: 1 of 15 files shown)


roseli-TT (approved):

gwangTT (approved):

tests/pipeline_reorg/models_unit_tests.yaml
```diff
@@ -1705,3 +1705,68 @@
...
+- name: TT-Transformers MLP 2D module unit tests
+  # 8x4 excluded: test_mlp_2d_vs_reference_from_model_args[32-8x4] and [512-8x4] hit
+  # TT_FATAL @ tt_metal/fabric/fabric.cpp:174 forwarding_direction. The 4x8 orientation passes.
+  cmd: |
+    unset TT_METAL_WATCHER
+    export HF_HOME=/mnt/MLPerf/huggingface
+    HF_MODEL=meta-llama/Llama-3.1-8B-Instruct \
+    pytest models/common/tests/modules/mlp/test_mlp_2d.py \
+      -m "not slow" \
+      -k "not 8x4" \
```
yieldthought: Codex: 8x4 also matches the two passing direct-API variants and the topology xfail. Narrow this to the two documented from_model_args failures.

```suggestion
      -k "not (test_mlp_2d_vs_reference_from_model_args and 8x4)" \
```

yieldthought:
Codex review:

1. [tests/pipeline_reorg/models_unit_tests.yaml:1739](https://github.com/tenstorrent/tt-metal/blob/4123f80d225fd59afc58d1505d2ffb5a12caab24/tests/pipeline_reorg/models_unit_tests.yaml#L1739) — Narrow the exclusion to the two failing `from_model_args` cases. `-k "not 8x4"` deselects every node whose parameter ID contains `8x4`, not only `test_mlp_2d_vs_reference_from_model_args[512-8x4]` and `[32-8x4]`. [Run 32980799692](https://github.com/tenstorrent/tt-metal/actions/runs/32980799692/job/98219577428) shows that the direct-API prefill and decode `test_mlp_2d_vs_reference[...,8x4]` cases both passed; this selector drops that known-good orientation coverage (and the topology xfail probe) during the migration. Suggestion: use `-k "not (test_mlp_2d_vs_reference_from_model_args and 8x4)"` so only the two documented fabric failures are excluded.

2. [models/common/tests/conftest.py:298](https://github.com/tenstorrent/tt-metal/blob/4123f80d225fd59afc58d1505d2ffb5a12caab24/models/common/tests/conftest.py#L298) — Do not silently bypass the Galaxy gate when topology detection fails. The catch-all `except Exception: return` allows the suite to continue after any transient descriptor error. A later successful `SystemMeshDescriptor()` call in `ttnn_mesh_device` will not make the same decision: `_allowed_req_shapes_for_system()` deliberately permits these 1D shapes on Galaxy, so the tests can then consume the hardware this fixture is meant to protect. Suggestion: distinguish the verified no-device/host-only case from descriptor failures on a hardware runner, and skip or fail explicitly for the latter instead of returning.

3. [tests/pipeline_reorg/models_unit_tests.yaml:1709](https://github.com/tenstorrent/tt-metal/blob/4123f80d225fd59afc58d1505d2ffb5a12caab24/tests/pipeline_reorg/models_unit_tests.yaml#L1709) — The final separated Galaxy entries have not been exercised at the current PR head. The cited [head run 32984785776](https://github.com/tenstorrent/tt-metal/actions/runs/32984785776) ended before `load-test-matrix`; its `models-unit-tests` job was skipped and no Galaxy test job was created. The earlier run used the folded commands from `f4f91cf53ec` and therefore does not verify these three entries or the new selector. Suggestion: rerun the exact `tt-transformers-common` / `wh_galaxy_perf` dispatch successfully before deleting the legacy registration.

yieldthought:
Approving to unblock but please use your best judgement regarding the codex review issues

yieldthought (approved):

Issue #18948, opened 2025-03-11
[Test Infra] Test narrow tile feature


pmilenkovicTT:
Create C++ tests for narrow tile feature on both BH and WHB0.

vmilicevicTT:
@fvranicTT , is this covered by the work you are doing on Test Infra?

fvranicTT:
Most likely yes. But we need to be more specific, which API methods need to be tested with narrow tile feature @pmilenkovicTT?

pmilenkovicTT:
I stumbled upon it when working with llk_pack_init - from my conversion with @rtawfik01 in March, it is yet to be tested, but I think it will be a part of the parameter sweep effort.

vmilicevicTT:
@fvranicTT , can you please check if this is covered in test infra, and if so, please close this issue?  Thanks!

nstamatovicTT:
The LLK Issue Solver marked this issue as outdated based on its current description, discussion, and the implementation on main. Unless told otherwise, this issue will be closed in 48 hours.

PR #54467, opened 2026-08-26
[SAN] Refactor State Store to Improve Flexibility. 

sstanisicTT:
### PR Category

Feature

### Summary

Introduces the new API for LLK Sanitizer hooks. The new API is based on KV approach. 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:llk/san/state-refactor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:llk/san/state-refactor)


tt_metal/tt-llk/tools/include/sanitizer/impl.h
```diff
@@ -0,0 +1,645 @@
...
+        {
+            snapshot_assert<Trigger::ERROR, hook>(exu.operation.snapshot.subset_of(exu.operand), exu.context.operation, exu.context.current);
+        }
+
+        exu.operation.status = OperationStatus::Executed;
+    }
+}
+
+// uninit() is execute()'s twin: the same arguments, the same checks, a different status left behind.
+// A nullary uninit<Op>() is still checked, because the operation is named rather than inferred.
```
sstanisicTT: Delete comment,  obvious

tt_metal/tt-llk/tools/include/sanitizer/operation.h
```diff
@@ -0,0 +1,487 @@
...
+    struct DataCopyType : Field<std::uint32_t>
+    {
+    };
+
+    struct BroadcastType : Field<std::uint32_t>
+    {
+    };
+
+    // Blackhole's MOP depends on the face count, so its halves restate it; Wormhole's execute is
+    // never handed it, and leaves the field unknown.
```
sstanisicTT:
```suggestion
// Note: Blackhole specific
```

tt_metal/tt-llk/tools/include/sanitizer/operation.h
```diff
@@ -0,0 +1,487 @@
...
+        BroadcastType,
+        NumFaces>;
+};
+
+// ------------------
+// OPERATION - PACK
+// ------------------
+
+// sstanisic todo: contract cannot be enforced if Pack has an uninit, without killing performance --
+// hence Hoistable::Yes, even though init does program per-op packer state.
```
sstanisicTT:
```suggestion
// sstanisic fixme: rework Pack Init/Uninit interaction
```

tt_metal/tt-llk/tools/include/sanitizer/operation.h
```diff
@@ -0,0 +1,487 @@
...
+        BlockCtDim,
+        FullCtDim,
+        NarrowRow,
+        RowNumDatums>;
+};
+
+// At namespace scope so ExuOperations can name them. Registering a list here is the only place it
+// is named: ExuState<Exu> reaches for it through ExuOperations, so State below cannot disagree with
+// these four specializations about which operations an Exu holds.
+using UnpackOperations = OperationList<OperationUnpackUnary, OperationUnpackBinary, OperationUnpackMatmul, OperationUnpackTilize>;
```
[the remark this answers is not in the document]
sstanisicTT: Expected and legitimate.

tt_metal/tt-llk/tools/include/sanitizer/operation.h
```diff
@@ -0,0 +1,487 @@
...
+        /* Fields */
+        BlockCtDim,
+        FullCtDim,
+        NarrowRow,
+        RowNumDatums>;
+};
+
+// At namespace scope so ExuOperations can name them. Registering a list here is the only place it
+// is named: ExuState<Exu> reaches for it through ExuOperations, so State below cannot disagree with
+// these four specializations about which operations an Exu holds.
```
sstanisicTT: Remove comment, redundant.

tt_metal/tt-llk/tools/include/sanitizer/output.h
```diff
@@ -0,0 +1,359 @@
...
+    }
+}
+
+/**
+ * @brief Check a value a hook named against the recorded state, and report when the two disagree.
+ *
+ * Prints the site the recorded value was written from as well as the hook's own, so a reader has both
+ * halves of the disagreement. A field the recorded state never saw prints as UNKNOWN, not as zero.
+ * The silent zone is the caller's to honour; this reports whenever the values differ.
+ *
```
sstanisicTT: Remove, Redundant.

tt_metal/tt-llk/tools/include/sanitizer/output.h
```diff
@@ -0,0 +1,359 @@
...
+
+/**
+ * @brief Check that the hook named the Operation its Exu holds a record for, and report when it did not.
+ *
+ * @tparam L The trigger level for this report.
+ * @tparam A The class of hook that triggered the check.
+ * @tparam E The Exu the hook names.
+ * @tparam R The type of record the hook looked up.
+ * @param exu The Exu's state, read for the seated status, the site it was seated at, the hook's own
+ *        site, and the silent zone, which this one honours.
```
sstanisicTT: Just the Exu's state, drop the rest

tt_metal/tt-llk/tools/include/sanitizer/output.h
```diff
@@ -0,0 +1,359 @@
...
+ * @param record The record the hook looked up, null when its Operation is not the seated one.
+ * @return Whether the Operation is seated, so the caller can stop without testing the record again.
```
sstanisicTT:
```suggestion
 * @return Whether the Operation is seated
```

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,804 @@
...
+{
+
+// A dependent false for the unreachable else of an if constexpr chain -- over a type or
+// over a value. The overloads are distinguished by template parameter kind: passing a
+// value to the first (or a type to the second) is a deduction failure, not an error.
+template <typename...>
+constexpr bool always_false()
+{
+    return false;
+}
```
sstanisicTT:
```suggestion
template <typename...>
constexpr bool always_false()
{
    return false;
}
```

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,804 @@
...
+ * Extracts the Group and Type.
+ *
+ * @tparam V Type to check.
+ */
+template <typename V>
+struct value_of : std::false_type
+{
+    // ----- invalid type escape hatch -----
+    // Mirrors the specialization's member names, for the reason extract_field_base spells out.
+    using Group = void;
```
sstanisicTT, line 477:
```suggestion
    // ----- invalid type escape hatch -----
    extract_field_base spells out.
    using Group = void;
```

sstanisicTT: Left this one open because I could not apply the suggestion verbatim — as written the middle line has lost its `// ` prefix:

```
    // ----- invalid type escape hatch -----
    extract_field_base spells out.
    using Group = void;
```

That is a bare statement and would not compile. Since the rest of this review pass removes verbose comments, I read the intent as "drop that line" and deleted it, keeping the banner and the member:

```cpp
struct value_of : std::false_type
{
    // ----- invalid type escape hatch -----
    using Group = void;
};
```

If you actually wanted shortened comment text there instead, say the word and I will change it — otherwise this thread is fine to resolve.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -34,6 +35,20 @@ inline void llk_unpack_A_init(
         tensor_shape.face_r_dim,
         tensor_shape.total_num_faces())));
 
+    // transpose_of_faces and within_face_16x16_transpose are not OperationUnpackUnary fields --
+    // transpose is modelled on the matmul operation, not the unary one.
```
sstanisicTT: This comment is wrong, it's not modeled because there is nothing to check against these. remove the comment.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -94,5 +133,9 @@ inline void llk_unpack_A_block(
 
 template <BroadcastType BType = BroadcastType::NONE>
 inline void llk_unpack_A_uninit() {
+    // No operand is in scope here; restating the broadcast variant is what ties the uninit to
+    // the operation that was initialised.
```
sstanisicTT: Remove comment.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_common_api.h
```diff
@@ -112,6 +126,14 @@ inline void llk_unpack_reconfig_data_format_srca(const std::uint32_t srca_new_op
     // Currently, there is a constraint that tile size is equal to the fifo page size
     // TODO NC: tile size should be computed in the LLK instead, as the part of #34495
     const std::uint32_t tile_size = get_local_cb_interface(srca_operand_id).fifo_page_size;
+    // face_r_dim and num_faces are only committed when dim_stride_target reprograms dim/stride, so
+    // this reconfig does not claim them.
```
sstanisicTT: Remove comment.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_common_api.h
```diff
@@ -136,6 +158,14 @@ inline void llk_unpack_reconfig_data_format_srcb(const std::uint32_t srcb_new_op
     // Currently, there is a constraint that tile size is equal to the fifo page size
     // TODO NC: tile size should be computed in the LLK instead, as the part of #34495
     const std::uint32_t tile_size = get_local_cb_interface(srcb_operand_id).fifo_page_size;
+    // face_r_dim and num_faces are only committed when dim_stride_target reprograms dim/stride, so
+    // this reconfig does not claim them.
```
sstanisicTT: Ditto.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_math_matmul_api.h
```diff
@@ -28,8 +28,19 @@ inline void llk_math_matmul_init(
 
     const bool partial_face = (in0_tile_r_dim < FACE_R_DIM);
 
-    // In0/operandA -> srcB, In1/operandB -> srcA
-    llk::san::math_operand_check(unpack_dst_format[in1_id], unpack_dst_format[in0_id]);
+    // In0/operandA -> srcB, In1/operandB -> srcA; SrcA's format is the FPU operand state.
```
sstanisicTT: Revert to the old comment.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_math_matmul_api.h
```diff
@@ -53,5 +64,16 @@ inline void llk_math_matmul(const uint dst_index, const std::uint32_t ct_dim = 1
     //     ct_dim,
     //     rt_dim);
 
+    // The FPU matmul execute takes no operand, so there is no Operand<Exu::Fpu> state to restate
+    // here; the four operation fields are the whole of what init seated, and restating them is what
+    // catches a matmul executed under a fidelity, throttle or block shape other than the one it was
+    // initialised for.
```
sstanisicTT: Remove comment.

tt_metal/tt-llk/tools/include/sanitizer/operation.h
```diff
@@ -0,0 +1,377 @@
...
+    {
+    };
+
+    struct PartialFace : Field<std::uint32_t>
+    {
+    };
+
+    struct NarrowTile : Field<std::uint32_t>
+    {
+    };
```
sstanisicTT: Should have // Note: Blackhole only, and be the last Field and the last in the struct.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_pack_tile_api.h
```diff
@@ -29,6 +29,28 @@ inline void llk_pack_init(const std::uint32_t pack_output, std::uint32_t num_til
         LLK_ASSERT_BLOCK(are_packers_configured_correctly(pack_src_format[output_id], pack_dst_format[output_id]));
     }
 
+    // pack_dst_format only reaches per-op state that a later reconfig cannot repair when the MOP bakes
+    // it in: the multi-tile output-address offset, and the partial-face BFP PACR variant. Otherwise
+    // reconfig_packer_data_format re-derives everything init took from the format (set_packer_l1_offset),
+    // so a format reconfig between init and execute is legal and must not read as drift.
```
sstanisicTT: Make more concise, the current wording is confusing.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -33,6 +34,20 @@ inline void llk_unpack_A_init(
         tensor_shape.face_r_dim,
         tensor_shape.total_num_faces())));
 
+    // transpose_of_faces and within_face_16x16_transpose are not OperationUnpackUnary fields --
+    // transpose is modelled on the matmul operation, not the unary one.
```
sstanisicTT: Remove comment

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -93,5 +132,9 @@ inline void llk_unpack_A_block(
 
 template <BroadcastType BType = BroadcastType::NONE>
 inline void llk_unpack_A_uninit() {
+    // No operand is in scope here; restating the broadcast variant is what ties the uninit to
+    // the operation that was initialised.
```
sstanisicTT: Remove comment.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_unpack_common_api.h
```diff
@@ -112,6 +126,14 @@ inline void llk_unpack_reconfig_data_format_srca(const std::uint32_t srca_new_op
     // Currently, there is a constraint that tile size is equal to the fifo page size
     // TODO NC: tile size should be computed in the LLK instead, as the part of #34495
     const std::uint32_t tile_size = get_local_cb_interface(srca_operand_id).fifo_page_size;
+    // face_r_dim and num_faces are only committed when dim_stride_target reprograms dim/stride, so
+    // this reconfig does not claim them.
```
sstanisicTT: Remove comment.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_unpack_common_api.h
```diff
@@ -136,6 +158,14 @@ inline void llk_unpack_reconfig_data_format_srcb(const std::uint32_t srcb_new_op
     // Currently, there is a constraint that tile size is equal to the fifo page size
     // TODO NC: tile size should be computed in the LLK instead, as the part of #34495
     const std::uint32_t tile_size = get_local_cb_interface(srcb_operand_id).fifo_page_size;
+    // face_r_dim and num_faces are only committed when dim_stride_target reprograms dim/stride, so
+    // this reconfig does not claim them.
```
sstanisicTT: Remove comment.

tt_metal/tt-llk/tools/include/sanitizer/operation.h
sstanisicTT: Moved `NarrowTile` last in both the field declarations and the `StateStruct`, as asked — but I annotated it `// Note: Wormhole only` rather than Blackhole, because the evidence points the other way:

| | Blackhole | Wormhole |
|---|---|---|
| `narrow_tile` in pack lib (`llk_pack*.h`) | **0** | 23 |
| `Operand<Exu::Pack>::NarrowTile` stated in pack api | **0** | 6 |
| `Operand<Exu::Pack>::TileWidth` stated in pack api | 5 | **0** |

Blackhole's `_llk_pack_init_` takes `(pack_src_format, face_r_dim, tile_c_dim, num_faces, num_tiles, skip_bh_tilize_workaround)` — no `narrow_tile`; Wormhole's takes `(pack_dst_format, face_r_dim, num_faces, partial_face, narrow_tile, num_tiles)` — no `tile_c_dim`. Blackhole also defines `get_output_narrow_tile` and never calls it, and `llk_pack_init` there carries `// TODO (#18948): Revisit for narrow_tile`.

So on the pack path `NarrowTile` is Wormhole-only and **`TileWidth` is the Blackhole-only field**. If you meant `TileWidth` should get the note and move last, say so and I will swap them — happy to annotate both, `TileWidth` as Blackhole-only and `NarrowTile` as Wormhole-only.

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+    }
+}
+
+} // namespace detail
+
+// ---------------
+// OperationStatus
+// ---------------
+
+enum class OperationStatus : std::uint32_t
```
ncvetkovicTT: Why don't we bring this buddy down next to `ApiClass`? Kinda makes it hard to read to jump up and down?

tt_metal/tt-llk/tools/include/sanitizer/output.h
```diff
@@ -0,0 +1,354 @@
...
+    SAN_ASSERT(false, "llk::san | fault   | api_class_name() was given an ApiClass no hook names");
+    return SAN_STRING("<unknown>");
+}
+
+/**
+ * @brief The seated operation's status, as a report spells it.
+ *
+ * @param status
+ */
+inline string operation_status_name(const OperationStatus status)
```
ncvetkovicTT: I guess this too can be lowered somewhere down near `ApiClass`?

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+
+#include <bitset>
+#include <cstddef>
+#include <cstdint>
+#include <tuple>
+#include <type_traits>
+#include <utility>
+#include <variant>
+
+// #include "llk_defs.h"
```
ncvetkovicTT: Why?

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+        static_assert(contains<F>(), "Field is not a member of this StateStruct.");
+        if constexpr (contains<F>())
+        {
+            constexpr std::size_t idx = index<F>();
+            return std::get<idx>(values);
+        }
+        __builtin_unreachable();
+    }
+
+    bool subset_of(const StateStructImpl& other) const
```
ncvetkovicTT: Maybe we should state how this differs from `equal`, because there we say that a field is not a member of the struct, for which, intuitively, this can also be used, right? Maybe just put a comment here and in `equal` saying what each does.

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+
+template <>
+struct OperandContext<Exu::Unpack>
+{
+    UnwindContext unpack_a;
+    UnwindContext unpack_b;
+};
+
+template <>
+struct OperandContext<Exu::Fpu>
```
ncvetkovicTT, line 720: Maybe it looks ugly, but won't you reduce the code significantly by splitting the unpack vs non-unpack case? I.e. jam fpu, sfpu and pack together.

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+ *  @brief Check if an Exu is native to a Thread.
+ */
+constexpr bool is_exu_native(Exu e, Thread t)
+{
+    switch (t)
+    {
+        case Thread::TRISC0:
+            return e == Exu::Unpack;
+        case Thread::TRISC1:
+            return e == Exu::Fpu || e == Exu::Sfpu; // the math thread drives both
```
ncvetkovicTT, line 644:
```suggestion
            return e == Exu::Fpu || e == Exu::Sfpu; // the math thread drives both, except when TRISC2 drives the Sfpu :P
```

But seriously, if we ever plan to expand the usage of this helper past `is_thread_supported`, how do we handle TRISC2->SFPU case?

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+            return false;
+    }
+
+    __builtin_unreachable();
+}
+
+/**
+ * @brief Check if a Thread is supported by sanitizer.
+ *
+ * Used to prevent hooks being called from TRISC3.
```
ncvetkovicTT, line 657: How? I mean, what's the intended usage of this.

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+// Operation<Exu, Hoistable>
+// ---------------------------------
+
+/**
+ * @brief Base class for Operation state.
+ *
+ * Operation state is specified by deriving from Operation.
+ *
+ * @tparam E Exu that the operation is native to.
+ * @tparam H Indicates whether the operation is hoistable.
```
ncvetkovicTT, line 120: Alright but we should somewhere (I honestly don't know where) state that on llk-lib level, this may be arch-dependent. For example, we can have reduce non-hoistable on WH/BH but hoistable on QSR. However, since it's non-hoistable for at least one arch, then it's non-hoistable for all. Or something along those lines. 

That will make it impossible to hoist something on QSR when writing llk-lib-only code, but oh well, I think we can live with that. 

Since this is for offline discussion, I am fine with leaving it like this.

tt_metal/tt-llk/tools/include/sanitizer/types.h
```diff
@@ -0,0 +1,800 @@
...
+    Execute,
+    Uninitialize,
+    Reconfigure
+};
+
+// -------------
+// UnwindContext
+// -------------
+
+struct UnwindContext
```
ncvetkovicTT, line 695: Slop warning, but seems based:
`UnwindContext::UNKNOWN` restates the sentinel that the member initializers already produce. pc and ra both default to `UINTPTR_MAX ` on lines 701-702, so `UnwindContext{}` and `UnwindContext::UNKNOWN` are the same value, written twice. If the default ever changes, the two silently disagree. Can we derive it instead?
```
// line 707
 -inline const UnwindContext UnwindContext::UNKNOWN {UINTPTR_MAX, UINTPTR_MAX};
 +inline const UnwindContext UnwindContext::UNKNOWN {};
```
(Or drop UNKNOWN altogether and write {} at the two use sites, `impl.h:294` and `impl.h:374`. It's also an out-of-line object, so it costs 8 bytes of .rodata per kernel binary that a value-init wouldn't.)

tt_metal/tt-llk/tools/include/sanitizer/impl.h
```diff
@@ -0,0 +1,643 @@
...
+        }
+
+        (field_check<hook>(sanitizer, values), ...);
+
+        if (!is_exu_silent(exu))
+        {
+            snapshot_assert<Trigger::ERROR, hook>(exu.operation.snapshot.subset_of(exu.operand), exu.context.operation, exu.context.current);
+        }
+
+        exu.operation.status = OperationStatus::Uninitialized;
```
ncvetkovicTT, line 637: Slop warning:
llk_unpack_tilize_init(cb, ct)   → seats OperationUnpackTilize, status = Initialized
llk_unpack_tilize(cb, 0)         → status = Executed
llk_unpack_tilize_uninit(cb)     → status = Uninitialized
llk_unpack_tilize(cb, 1)         → should report "you executed an op you tore down"
It doesn't. uninit (impl.h:637) only assigns status; operation.specific still holds the tilize alternative and snapshot still holds init's values. operation_seated_assert (output.h:233) tests record != nullptr and never looks at status — even though it prints it at :254. So the fourth call passes the seated check and validates against the torn-down record. Since tilize is the one Hoistable::No op, this is exactly the op where it matters.

ncvetkovicTT (commented):
1. Does it make sense to run some tests with ` TT_METAL_LLK_SANITIZER` defined? Or add some tests to see if this all compiles and all.
2. Slop warning: The sanitizer's reports are built to name two places: where a value was recorded, and where the mismatch was noticed. Both come from UnwindContext, which is filled in by LLK_SAN_FUNCTION(). In this PR that macro is used in two places, both overloads of compute_kernel_hw_startup in tt_metal/hw/inc/api/compute/compute_kernel_hw_startup.h, against the 47 SAN_HOOK call sites the PR adds. I also looked through the rest of the stack (55062, 55054, 55055, 55093, 55094 and 55057) and none of them add further uses outside the lit tests. For every hook that is not reached through compute_kernel_hw_startup, context_depth stays at 0, so context.current keeps its initial value of UnwindContext::UNKNOWN. Both source locations in field_assert, operation_seated_assert and snapshot_assert then print as empty. There is a second effect on the hooks that do get a real location. field_update calls operand_context_update for Reconfigure as well as Configure (impl.h:423 and impl.h:434), so every llk_unpack_reconfig_data_format_srca, llk_unpack_reconfig_data_format_srcb and llk_pack_reconfig_data_format overwrites the location that llk_unpack_hw_configure or llk_pack_hw_configure recorded, replacing it with UNKNOWN. Is adding LLK_SAN_FUNCTION() to the Compute API entry points planned for a later PR in the stack, or is the intention that reports carry no source location for now? The answer also decides whether the per-operand slots in OperandContext are worth keeping in their current shape, since operand_context_update and operand_context_get together are about eighty lines of impl.h that only start to matter once the locations are real.

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: I don't want to do that because now if you try to update Operand state of FPU from TRISC2 it will error (because sanitizer doesn't yet support that), and if we put it in the same container it won't do that.

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: Good question, don't remember, will remove.

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: I think the current model is not mature enough to support that case in any way. Theoretically all it would take is some Mutex style primitive but I'm positive it would break in integration.

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: We can always split it into  OperationWh and OperationBh which will have independat fields and hoists. I just didn't find that usecase yet so I reused with unknown fields. That would introduce Arch dependency on LLK API level though which is kind of annoying.

tt_metal/tt-llk/tools/include/sanitizer/impl.h
sstanisicTT: I want to fix this in a following revision, that will also handle the Reconfig / Uninit (Hoistability) interaction, but this diff is alraedy too large

tt_metal/tt-llk/tools/include/sanitizer/types.h
ncvetkovicTT: Ok once you remove it I will resolve the convo.

tt_metal/tt-llk/tools/include/sanitizer/types.h
ncvetkovicTT: Deal.

tt_metal/tt-llk/tools/include/sanitizer/types.h
ncvetkovicTT: Alright let's leave it for some other time.

tt_metal/tt-llk/tools/include/sanitizer/types.h
ncvetkovicTT: But LLK API si already arch-dependent. I will close this thread, no need to continue here tbh, you already kinda answered.

tt_metal/tt-llk/tools/include/sanitizer/impl.h
ncvetkovicTT: Alright I will close the thread.

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: LGTM

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT:
```    
if constexpr (!is_thread_supported(T))
{
    return OperationDefect::NotThreadSupported;
}
```
It's a bit of a roundabout. It checks that none of the Sanitizer EXUs are native to this thread. If this is the case, Thread is unsupported. This gets triggered when you call any API from TRISC3 on QSR. I added this because I think that QSR support will first be rolled out in Sanitizer without TRISC3, using the standard WH/BH UNPACK/{FPU,SFPU}/PACK split

tt_metal/tt-llk/tools/include/sanitizer/types.h
sstanisicTT: LGTM

ncvetkovicTT (approved):
Fix PR gate, run some pipelines before merging to make sure we're not breaking anything. LGTM.

nvelickovicTT (approved):

dpopovTT (approved):

kstevensTT (approved):

PR #54474, opened 2026-08-26
[Feature] Support bias in 1D depthwise conv1d

ctxbot:
### PR Category

Feature

### Summary

Depthwise conv1d (groups == C) with bias doesn't work today. The sharded depthwise factory throws `Bias is not supported for depthwise conv1d`, and the grouped-layout fallback can't TILE-slice a prepared device bias, so at large channel counts the op just runs out of L1 — with the grouped-delta-net decode shapes (C=10240, groups=10240, K=4) a prepared-weights-with-bias call fatals with `DRAM Auto slice could not find valid slice configuration`. I hit exactly that running depthwise conv1d with bias at C=10240.

This lets the 1D depthwise path take bias directly:

- `is_1d_depthwise_conv` no longer excludes `has_bias`; the factory throw is removed.
- The bias is prepared row-replicated in the depthwise tiled layout (logical `[1,1,32,C_padded]`), so the row broadcast comes from the tile shape and the kernel folds bias into DST on the last tap with a dest-reuse add — both the tap loop and the coalesced path. No pack-format change, and the bias CB is never popped.
- Convs that lower to `ttnn::linear` keep the stock bias form; the prep config now carries an `mm_conv` flag so weight prep and bias prep stay consistent.
- On the channel-chunk path, `can_chunk_channels` accepts a prepared device bias and the chunk loop TILE-slices it along the last dim the same way as the prepared weights, so prepared weights + bias chunks instead of fataling.

Based on top of #53314; the chunk-path pieces need that to land first.

### Testing

Added prepared-weights-with-bias rows to `test_conv1d_grouped_dram_channel_chunk_routes`; existing host-bias, prepared-no-bias and mm_conv rows unchanged. All on a single P150a (Blackhole), bf16, fp32 dest acc, PCC vs torch conv1d:

| weights | bias | C | slice | PCC |
|---|---|---|---|---|
| host | host | 10240 | auto | 0.999992 |
| host | host | 10240 | L1_FULL | 0.999992 |
| prepared | prepared | 10240 | auto | 0.999992 |
| prepared | prepared | 10240 | L1_FULL | 0.999992 |
| prepared | prepared | 96 | auto | 0.999997 |
| prepared | prepared | 64 | auto | 0.999996 |
| host | host | 96 | auto | 0.999997 |
| host | host | 64 | auto | 0.999996 |
| prepared | none | 96 | auto | 0.999999 |
| prepared | none | 64 | auto | 0.999999 |
| linear (mm_conv) | host | 32 | L1_FULL | 0.999999 |

The prepared+prepared rows at C=10240 are the new capability — same PCC as the host-bias chunked value at that shape, which is what you'd expect when the chunk path engages with the device bias sliced alongside the weights. The mm_conv row is the guard that a conv lowering to `ttnn::linear` still takes the stock bias form and doesn't throw.

Also ran the full node under the Wormhole ttsim build (v1.10.2) — 11/11 there as well, so no WH-specific surprises. GS is the one arch I can't check locally.

Surrounding suites on the same setup: `test_conv1d.py` 56 passed, `test_prepare_conv_weights.py` 69 passed, `test_conv2d.py` 161 passed.

### Notes for reviewers

/cc @vtsilytskyiTT


vtsilytskyiTT:
@ctxbot can you address copilot issues? Then we'll run CI and do a round of reviews.

PR #54478, opened 2026-08-26
Package the models/ tree as tt-metal-models for pip, apt, and dnf

knauth:
The ttnn wheel ships the runtime but not the model implementations, so serving requires a git clone and PYTHONPATH. The Tenstorrent vLLM plugin maps every registered architecture to a dotted path rooted at `models.`, so without an importable models tree every model fails to load, and vLLM reports it only as "Model architectures [...] failed to be inspected".

Add packaging/tt-metal-models/, which builds that tree as a distribution:

- The wheel takes its version from the same setuptools-scm derivation ttnn uses and declares ttnn==<that version>. The coupling is real but invisible in the source (models/ has no version asserts), so a strict pin makes skew a resolver error instead of an AttributeError deep in model construction.
- MANIFEST.in is the source of truth for the non-Python payload. models/ is a PEP 420 namespace package, so discovery needs namespaces=true; and the data lives in directories that hold no .py and whose names are not valid Python identifiers (Llama-3.1-70B-Instruct), so package-data globs cannot reach it.
- import_matrix.py is the gate: it imports every vLLM entry point against the built wheel, each in its own subprocess from a neutral directory, because a models/ directory in the cwd otherwise shadows the installed one and the matrix would silently test the checkout instead.
- build_native_packages.py produces python3-tt-metal-models from the wheel, so file selection has one source of truth rather than a second policy that drifts.

Fix the defects the matrix found, all of which broke serving from a package:

- models/common/utility_functions.py imported pytest at module scope, making a test framework a hard runtime dependency of essentially every model. Moved into the only function that uses it.
- The bge_large_en runners imported assert_with_pcc from their own test package, which the test-directory prune necessarily removes. Moved it beside the comp_pcc it wraps; the test module re-exports it, so the tests are unchanged.
- model_config.py resolved its data root from TT_METAL_HOME before __file__, so a stale value silently loaded model params from an old checkout. It now prefers the installed package when the env-pointed root has no matching tree, and warns.

Known limitations are documented in packaging/tt-metal-models/README.md. Two need a decision rather than code: the llama reference submodule is Llama 2 Community Licensed, not Apache-2.0, so the build never vendors it implicitly; and deepseek_v3_b1 refers to its kernels by repo-relative string literals that the runtime resolves against its kernel search root, not __file__.

### Notes for reviewers

---

**Supersedes #54340.** That PR was opened from a fork, and the CI runner hook rejects fork PRs that modify `.github/workflows/*` (the `Build Release` job failed before building anything). Same branch, rebased onto current main, now pushed in-repo. Prior discussion and Cycode history live on #54340.

**Note on Cycode findings:** the SAST rule flags every `subprocess.run` in the packaging scripts as "unsanitized user input in an OS command". All calls are list-form with `shell=False`; inputs (interpreter paths, wheel paths, module names) are validated before use — module names in particular are checked against a dotted-identifier regex because the import matrix interpolates them into probe source. The scanner does not recognize these as sanitizers and re-flags the sinks, so remaining findings on this developer/CI tooling need a platform-side dismissal rather than further code change.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:knauth/tt-metal-models-packaging)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:knauth/tt-metal-models-packaging)

### DIFF
--- a/.github/workflows/package-and-release.yaml
+++ b/.github/workflows/package-and-release.yaml
@@ -243,7 +243,7 @@ jobs:
 
       - name: Publish to PyPI (Trusted Publishing)
         if: ${{ !inputs.dry-run }}
-        uses: pypa/gh-action-pypi-publish@ed0c539abdc6b55ad89f3ea7b8e96860a5ddef80 # v1.13.0
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
 
   # A separate job, not another file in the job above's dist/, because PyPI trusted
   # publishing issues a token scoped to a single project: the `pypi` environment's
@@ -282,7 +282,7 @@ jobs:
 
       - name: Publish to PyPI (Trusted Publishing)
         if: ${{ !inputs.dry-run }}
-        uses: pypa/gh-action-pypi-publish@ed0c539abdc6b55ad89f3ea7b8e96860a5ddef80 # v1.13.0
+        uses: pypa/gh-action-pypi-publish@ed0c53931b1dc9bd32cbe73a98c7f6766f8a527e # v1.13.0
 
   create-changelog:
     needs: [create-tag, release-precheck]
--- a/.github/workflows/package-and-release.yaml
+++ b/.github/workflows/package-and-release.yaml
@@ -162,6 +162,55 @@ jobs:
       is-release-candidate: ${{ needs.release-precheck.outputs.is-release-candidate }}
       tests-yaml-gist-url: ${{ inputs.tests-yaml-gist-url }}
 
+  # Built once, not inside build-test-publish's per-platform matrix: tt-metal-models is
+  # py3-none-any, so a per-platform build would produce identical wheels under one
+  # artifact name and collide on upload. Built from the release tag so it carries the
+  # same version as the ttnn wheel and its `ttnn==<version>` pin resolves.
+  build-models-wheel:
+    needs: create-tag
+    uses: ./.github/workflows/models-wheel.yaml
+    with:
+      # On a dry run the tag is calculated but never created, so checking it out would
+      # fail; build from the triggering ref instead. The version string differs from a
+      # real release in that case, which is correct -- a dry run is not a release.
+      ref: ${{ inputs.dry-run && github.ref || format('refs/tags/{0}', needs.create-tag.outputs.version) }}
+
+  publish-models-to-internal-pypi:
+    name: "Publish tt-metal-models to internal PyPI"
+    needs: [build-models-wheel, create-tag]
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write # Required for AWS OIDC
+    if: ${{ github.ref == 'refs/heads/main' && !inputs.dry-run }}
+    steps:
+      - name: Set up Python
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: "3.10"
+
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.PYPI_ROLE }}
+          aws-region: ${{ secrets.PYPI_REGION }}
+
+      - name: Install s3pypi
+        run: pip install s3pypi
+
+      - name: Download tt-metal-models artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: ${{ needs.build-models-wheel.outputs.artifact-name }}
+          path: ./wheels
+
+      - name: Publish wheel to internal PyPI
+        run: |
+          found_wheel_files=$(find ./wheels -type f -name "*.whl" -exec realpath {} \;)
+          for wheel in $found_wheel_files; do
+            echo "Publishing wheel: $wheel"
+            s3pypi upload "$wheel" --put-root-index --bucket ${{ secrets.PYPI_BUCKET }}
+          done
+
   publish-to-pypi:
     needs: [build-test-publish, create-tag]
     runs-on: ubuntu-latest
@@ -196,7 +245,44 @@ jobs:
         if: ${{ !inputs.dry-run }}
         uses: pypa/gh-action-pypi-publish@ed0c539abdc6b55ad89f3ea7b8e96860a5ddef80 # v1.13.0
 
+  # A separate job, not another file in the job above's dist/, because PyPI trusted
+  # publishing issues a token scoped to a single project: the `pypi` environment's
+  # identity is bound to `ttnn` and would reject a `tt-metal-models` upload. This needs
+  # its own environment with its own trusted publisher registered on PyPI for the
+  # `tt-metal-models` project, pointing at this workflow and this environment name.
+  #
+  # For the same reason the artifact is named `tt-metal-models-dist-*` rather than
+  # `ttnn-dist-*`: the job above globs that pattern and would otherwise sweep this wheel
+  # into the ttnn upload.
+  publish-models-to-pypi:
+    needs: [build-models-wheel, create-tag]
+    runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/stable' || startsWith(github.ref, 'refs/tags/v') || inputs.dry-run }}
+    environment:
+      name: pypi-models
+      url: https://pypi.org/project/tt-metal-models/
+    permissions:
+      id-token: write
+      contents: read
+    steps:
+      - name: Download tt-metal-models artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          pattern: tt-metal-models-dist-*
+          path: artifacts
+          merge-multiple: true
 
+      # Only the wheel goes to PyPI. The .deb/.rpm in the same artifact reach users via
+      # the GitHub release assets instead.
+      - name: Flatten into dist/
+        run: |
+          mkdir -p dist
+          find artifacts -type f -name '*.whl' -exec cp {} dist/ \;
+          ls -l dist
+
+      - name: Publish to PyPI (Trusted Publishing)
+        if: ${{ !inputs.dry-run }}
+        uses: pypa/gh-action-pypi-publish@ed0c539abdc6b55ad89f3ea7b8e96860a5ddef80 # v1.13.0
 
   create-changelog:
     needs: [create-tag, release-precheck]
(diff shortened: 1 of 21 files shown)


yieldthought:
Codex review:

I found three additional issues beyond the existing inline review comments:

1. [INSTALLING.md:122](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/INSTALLING.md#L122) breaks the documentation build at this PR head. The `build-docs` job fails with `myst.xref_missing` because `INSTALLING.md` is copied into the Sphinx source tree, where `packaging/tt-metal-models/README.md` is not a documentation target. Suggestion: use the repository's absolute GitHub URL here, or add the packaging README to the documentation tree and link to its doc target.

2. [.github/workflows/models-wheel.yaml:122](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/.github/workflows/models-wheel.yaml#L122) runs the matrix without vLLM and without `--strict`, while [packaging/tt-metal-models/import_matrix.py:48](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/packaging/tt-metal-models/import_matrix.py#L48) deliberately treats a missing `vllm` import as `SKIP`. The current successful matrix run skipped 6 of the 11 `generator_vllm` modules, including `models.tt_transformers.tt.generator_vllm`, so a missing packaged module, data file, or dependency behind those first vLLM imports will not fail this release gate. This package exists specifically to support those serving entry points. Suggestion: add a strict matrix leg with the supported Tenstorrent vLLM/plugin environment (or an equivalent consumer integration test) and gate publication on it.

3. [packaging/tt-metal-models/pyproject.toml:26](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/packaging/tt-metal-models/pyproject.toml#L26) declares the distribution as Apache-2.0-only, but [packaging/tt-metal-models/MANIFEST.in:55](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/packaging/tt-metal-models/MANIFEST.in#L55) makes the wheel redistribute third-party code under other terms. The CI-built wheel contains [models/demos/deepseek_v3_d_p/reference/kimi_k3/attn_res/hf_attn_res.py:5](https://github.com/tenstorrent/tt-metal/blob/824666dc93e6b0c8e6918b23c3dd88a24e454d3f/models/demos/deepseek_v3_d_p/reference/kimi_k3/attn_res/hf_attn_res.py#L5), whose governing `LICENSE-Kimi-K3` adds service and attribution conditions, as well as MIT-tagged Z Lab and Motif sources for which those subtrees contain no full MIT notice. The wheel metadata nevertheless reports only `License: Apache-2.0` and `License-File: LICENSE`. Suggestion: either exclude those sources from the distribution, or declare a complete license expression (including a `LicenseRef` for Kimi K3), enumerate every applicable license file, and add the missing MIT notices before publishing.

yieldthought:
Approving to unblock but please use your best judgement regarding the codex review issues

yieldthought (approved):

blozano-tt (changes requested):
Need some questions answered before proceeding 

Does this unconditionally couple the models/ directory with ttnn (Is this the right design for the product? Would pip install torch bring in similar baggage? How wil it work when models/ directory moves to another repo?)What will the wheel size be?Can it still be published to pypi?Do package and release pipelines pass?

.github/workflows/package-and-release.yaml
```diff
@@ -162,6 +162,55 @@ jobs:
       is-release-candidate: ${{ needs.release-precheck.outputs.is-release-candidate }}
       tests-yaml-gist-url: ${{ inputs.tests-yaml-gist-url }}
 
+  # Built once, not inside build-test-publish's per-platform matrix: tt-metal-models is
+  # py3-none-any, so a per-platform build would produce identical wheels under one
+  # artifact name and collide on upload. Built from the release tag so it carries the
+  # same version as the ttnn wheel and its `ttnn==<version>` pin resolves.
+  build-models-wheel:
```
dpopovTT, line 169: Ubuntu 22 and Ubuntu 24 don't share the same python version, because of that we need different wheel for every OS (python version)

.github/workflows/package-and-release.yaml
```diff
@@ -162,6 +162,55 @@ jobs:
...
+    needs: create-tag
+    uses: ./.github/workflows/models-wheel.yaml
```
dpopovTT, line 171: Where is this tested?

.github/workflows/package-and-release.yaml
```diff
@@ -162,6 +162,55 @@ jobs:
...
+    with:
+      # On a dry run the tag is calculated but never created, so checking it out would
+      # fail; build from the triggering ref instead. The version string differs from a
+      # real release in that case, which is correct -- a dry run is not a release.
+      ref: ${{ inputs.dry-run && github.ref || format('refs/tags/{0}', needs.create-tag.outputs.version) }}
+
+  publish-models-to-internal-pypi:
```
dpopovTT, line 178: This also needs to move to build test publish

.github/workflows/package-and-release.yaml
```diff
@@ -196,7 +245,44 @@ jobs:
...
+  # A separate job, not another file in the job above's dist/, because PyPI trusted
+  # publishing issues a token scoped to a single project: the `pypi` environment's
+  # identity is bound to `ttnn` and would reject a `tt-metal-models` upload. This needs
+  # its own environment with its own trusted publisher registered on PyPI for the
+  # `tt-metal-models` project, pointing at this workflow and this environment name.
+  #
+  # For the same reason the artifact is named `tt-metal-models-dist-*` rather than
+  # `ttnn-dist-*`: the job above globs that pattern and would otherwise sweep this wheel
+  # into the ttnn upload.
+  publish-models-to-pypi:
```
dpopovTT, line 257: Can we merge this into one step that uploads multiple wheels to pypi

dpopovTT (changes requested):

dpopovTT:
I can't find any CI run for this branch, and it is editing one of our main gates for release. Please test this thoroughly

mtairum (approved):
Approving from the models side. Please address infra team concerns on the packaging.

Please address the copilot/claude comment on the `model_config.py` as well.

models/tt_transformers/tt/model_config.py
```diff
@@ -63,11 +63,37 @@
...
+            return candidate
+        if (_INSTALLED_ROOT / _PARAMS_DIRNAME).is_dir():
+            logger.warning(
+                f"{env_var}={value} does not contain {_PARAMS_DIRNAME}; "
+                f"resolving bundled model parameters against {_INSTALLED_ROOT} instead. "
+                f"Unset {env_var} if you meant to use the installed models package."
+            )
+            return _INSTALLED_ROOT
+        return candidate
+    return _INSTALLED_ROOT
```
mtairum, line 93: Following up on the Copilot comment about the resolution order here — it is correct about the control flow, though I read the impact a bit differently, and I think it points at a second branch that matters more.

**The precedence break is real.** Once the first env var is non-empty, every path returns, so `TT_METAL_HOME` is unreachable whenever `TT_METAL_RUNTIME_ROOT` is set: `continue` is the only escape to the next iteration, and it is guarded by `if not value`. That contradicts the comment this PR leaves in place directly above (L60-62), which still promises "Prefers `TT_METAL_RUNTIME_ROOT` ... falling back to the traditional `TT_METAL_HOME`". The old `or` chain fell back on *unset*; the new code introduces a validity check but only half-applies the fallback to it, and the comment was not updated to match.

**Impact is narrower than "silently loads a different parameter tree", though.** For the two paths to actually differ you need all of: both vars set, `TT_METAL_RUNTIME_ROOT` invalid, `TT_METAL_HOME` valid, *and* `_INSTALLED_ROOT != TT_METAL_HOME`. In a checkout `PYTHONPATH=$TT_METAL_HOME`, so `_INSTALLED_ROOT` and `TT_METAL_HOME` are the same directory and the fallback lands on the identical path. It only becomes observable in the split layout (wheel in site-packages alongside a separate checkout) that this PR introduces — so worth fixing, but not a regression against current behaviour.

For what it's worth, I checked every place in the repo that sets `TT_METAL_RUNTIME_ROOT`, and none of them can hit this today:

- `tests/scripts/quasar/run_quasar_regression.sh:8`, `bootstrap_pipeline_dir.sh:306`, `tt-train/tools/profiling/run_experiments.sh:143`, `tests/scripts/multihost/run_quad_galaxy_tests.sh:498`, `tools/scaleout/exabox/health_check_test_suite/diag_runner.py:1226-1227` — all set it equal to `TT_METAL_HOME`, or derive one from the other.
- `models/demos/deepseek_v3_b1/scaleout_configs/generate_blitz_decode_pipeline_configs.py:291` — sets it only for a C++ gtest invocation; no Python `model_config` in that process.

So no model-tier CI exposure either way.

**The branch I would worry about more is the last `return candidate`.** It returns a root that was just proven not to contain `_PARAMS_DIRNAME`, with no warning at all. That is the silent-wrong-paths failure mode the new comment at L67-72 says the change exists to prevent, and it is the one case where the user gets no signal.

A version that fixes both — restores "first *valid* root wins", preserves the historical behaviour when nothing on disk is valid, and stops returning a known-bad root silently:

```python
def _resolve_repo_root() -> Path:
    rejected = []
    for env_var in ("TT_METAL_RUNTIME_ROOT", "TT_METAL_HOME"):
        value = os.environ.get(env_var)
        if not value:
            continue
        candidate = Path(value)
        if (candidate / _PARAMS_DIRNAME).is_dir():
            return candidate
        rejected.append((env_var, candidate))

    if not rejected:
        return _INSTALLED_ROOT

    names = ", ".join(f"{var}={path}" for var, path in rejected)
    if (_INSTALLED_ROOT / _PARAMS_DIRNAME).is_dir():
        logger.warning(
            f"{names} does not contain {_PARAMS_DIRNAME}; resolving bundled model "
            f"parameters against {_INSTALLED_ROOT} instead. Unset it if you meant "
            f"to use the installed models package."
        )
        return _INSTALLED_ROOT

    # Nothing on disk has the tree; preserve the historical env-root behaviour
    # rather than silently substituting a root that is equally wrong.
    logger.warning(f"{_PARAMS_DIRNAME} not found under {names} or {_INSTALLED_ROOT}.")
    return rejected[0][1]
```

The comment at L60-62 should also be amended to say the fallback now triggers on validity, not only on the var being unset.

PR #54479, opened 2026-08-26
Allocator: add opt-in range lockstep for HYBRID L1 buffers

ronnie00TT:
### Problem description

In `AllocatorMode::HYBRID`, placing a **lockstep** L1 buffer gathers the per-core allocators' occupied ranges so its address avoids them — scanning **every bank on every device**, with no filtering by the cores the buffer occupies.

The default has to be that broad in general, since a multicast writes every core in its rectangle whether or not that core is a destination. But the ranges are neither deduplicated nor attributable to a core, so their union grows until placements fail with most of L1 free. One measured failure: 72 KB requested across 8 cores that held nothing, refused with 587 KB free and a 278 KB contiguous block, after 2952 ranges summing to 342 MB were subtracted from a 1.36 MB address space.

### What's changed

A third allocation mode a buffer opts into. Lockstep semantics are unchanged.

| mode | address | kept clear of per-core allocations on |
| --- | --- | --- |
| per-core | independent on each core | n/a |
| lockstep (default, unchanged) | one address across its cores | **every** core on every device |
| **range lockstep (new, opt-in)** | one address across its cores | the cores it **occupies**, on every device |

Both lockstep modes still scan every device, since a mesh buffer holds one address on all of them; the opt-in narrows *which cores*.

**The opt-in.** `set_range_lockstep_allocation(target, enable)`, the same shape as the existing `per_core_allocation` free functions, on `BufferShardingArgs` for constructing a `Buffer` directly and on `MemoryConfig` for tensors. Enabling it is rejected if the buffer already asks for per-core allocation, and on interleaved buffers, which occupy every bank and so have no smaller set of cores to scope to.

**The narrowed scan.** With the flag set, the allocator collects occupied ranges only from the buffer's own cores. Which cores those are comes from whichever sharding description the buffer carries — a distribution spec's `cores_with_data()`, or a shard spec's grid. Cores that have no allocator bank are skipped: a service core claimed by fast dispatch is a legal shard core with real L1, but the allocator gives it no bank, so it holds nothing to avoid.

**A more useful out-of-memory message.** The free-bytes and largest-free-block figures it already printed are one allocator's own view and say nothing about what was subtracted on behalf of the others, so this kind of failure reads as ordinary exhaustion. It now also prints how many ranges were subtracted and how many bytes survived them.

Two pre-existing bugs found while testing, fixed in their own commits: `MemoryConfig`'s hand-written `operator==` and JSON converters listed `per_core_allocation` but not the new flag, so configs differing only in it compared equal and serialized identically; and top-down range selection formed `r.second - size_per_bank` before checking it, so an oversized request wrapped unsigned and was taken as a valid placement past the end of L1.

### Using it

There is no core list to pass: the scan is narrowed to the cores the memory config already shards over, so the flag is one extra call on a config you were building anyway.

```python
core_grid = ttnn.CoreRangeSet([ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 0))])

mem_config = ttnn.MemoryConfig(
    ttnn.BufferType.L1,
    ttnn.ShardSpec(core_grid, [1, num_bytes], ttnn.ShardOrientation.ROW_MAJOR),
)
mem_config.experimental_set_range_lockstep_allocation(True)

tensor = ttnn.from_torch(data, ..., memory_config=mem_config, device=mesh_device)
```

Here the scan covers (0,0) through (3,0) and nothing else, where the default would have covered every core on every device. The C++ equivalent is the same call from `experimental/range_lockstep_allocation/memory_config.hpp`, or on `BufferShardingArgs` when constructing a `Buffer` directly. Requires `TT_METAL_ALLOCATOR_MODE_HYBRID=1` before device open. Disabling is always legal; the guards apply only when enabling.

**When not to use it.** Only opt in when nothing reaches the buffer on a core it was not allocated on. A multicast rectangle wider than the buffer's own grid is the case to check: `num_mcast_dests` counts destinations but does not stop the NOC writing the whole rectangle.

### Test plan

Blackhole p150b; the C++ suite needs slow dispatch (its fixture skips otherwise) plus `TT_METAL_ALLOCATOR_MODE_HYBRID=1`.

| suite | result |
| --- | --- |
| `unit_tests_per_core_allocation` | **21/21** (11 new) |
| `tests/ttnn/unit_tests/range_lockstep_allocation/` | **23/23** (new) |
| `unit_tests_ttnn_tensor --gtest_filter=TensorSerializationFlatbufferTest.*` | **5/5** (2 new) |

Scenario: hog most of one core's L1 per-core, then request the same size on a different core. Range lockstep must place it; default lockstep must not. Through a mesh deliberately, since `hybrid_device_allocators_` is only populated for mesh allocations.

The Python suite adds the direction the C++ one missed — a range lockstep buffer must still avoid occupancy on the cores it **does** hold — parametrized over the hogged core's position. Mutation-tested by breaking the scan three ways: collecting nothing fails all three positions, scanning only the grid's first core fails two, dropping its last core fails one. None were caught before.

One more C++ test covers the nd shard spec conversion, the only place where losing the flag is completely silent: `TensorSpec` rebuilds the `MemoryConfig` from named fields there, which drops the experimental flags unless they are restored, and the buffer then reverts to the full scan with nothing reporting it. Whether the conversion runs at all is shape-dependent, so the test uses a shape it succeeds for. Verified by reverting just that restore and rebuilding: it fails, and passes again with it.

Default mode is a no-op: a 146-test allocator subset of `distributed_unit_tests` under LOCKSTEP gives byte-identical per-test results with and without this change.

### Checklist

- [x] New/Existing tests provide coverage for changes
- [ ] Post-commit CI
- [ ] `(Runtime) Unit Tests` — `runtime_sd_api` covers `unit_tests_per_core_allocation` on `wh_n150_civ2` / `wh_n300_civ2` / `bh_p150b_civ2` / `bh_p300` / `bh_p100a_civ2_viommu`. Nightly + manual dispatch only.

### Notes for reviewers

**Longer term**, the toggle is expected to be transitional: it exists while ops that reach a buffer across cores are reworked and mcast gets proper enforcement. What replaces it — scoped lockstep as the default, or per-core allocation as the default with lockstep opted into on a core range — was left open; tracked in #55253. This PR is the mechanism only.

I've also run [(Runtime) Unit Tests](https://github.com/tenstorrent/tt-metal/actions/runs/33522146295) and [T3K Unit tests](https://github.com/tenstorrent/tt-metal/actions/runs/33529780425).

Note that the failing test cases in CI pipeline also failed in main branch.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ronnie/hybrid-alloc-core-filtered-ranges)


tt_metal/impl/allocator/allocator.cpp
```diff
@@ -179,12 +179,37 @@ DeviceAddr AllocatorImpl::allocate_buffer(Buffer* buffer) {
...
+                    for (const auto& core : cores) {
+                        occupied_banks.push_back(logical_core_to_bank_ids_.at(BufferType::L1).at(core).at(0));
+                    }
+                } else {
+                    uint32_t num_l1_banks = l1_manager_->num_banks();
+                    occupied_banks.reserve(num_l1_banks);
+                    for (uint32_t bank_id = 0; bank_id < num_l1_banks; bank_id++) {
+                        occupied_banks.push_back(bank_id);
+                    }
+                }
```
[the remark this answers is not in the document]
ronnie00TT: Both points are real; fixed in 9ed08f3.

**Service cores.** Confirmed against `Buffer`'s own shard-grid validation, which explicitly admits a shard core with no bank:

```cpp
TT_FATAL(allocator.has_bank(bank_type, core) || is_claimed_service_core(core), ...);
```

so the unconditional `logical_core_to_bank_ids_.at(...)` would have thrown on a case that is legal today. Cores are now added through a helper that skips any core without an L1 bank; such a core holds no per-core ranges to avoid, so skipping it is also semantically right rather than just defensive.

**Distribution specs.** Also right, and it is what `Buffer::num_cores()` keys off for that path:

```cpp
if (buffer_distribution_spec_.has_value()) { return buffer_distribution_spec_.value().num_cores_with_data(); }
```

Since `num_cores()` is what becomes `num_banks` for the allocation, `cores_with_data()` is exactly the set that can collide — using `cores()` instead would under-scan and risk a real overlap. The distribution-spec branch is now checked first, with the shard-spec branch retained after it and the all-banks scan kept for interleaved buffers.

**Regressions for both paths**, as suggested: `LockstepIgnoresPerCoreRangesOnOtherCores` and `LockstepIgnoresPerCoreRangesOnOtherCoresWithDistributionSpec`. Verified in both directions on Blackhole p150b — with the change 12/12 pass; with `allocator.cpp` reverted to `origin/main` and the tests kept, exactly those two fail and the other ten are unaffected.

They are deliberately separate `TEST_F`s rather than one parameterised case: I first wrote them as two flavours in a single test and the second silently passed without the fix, because the first flavour's failed allocation leaves the allocator perturbed enough to mask it. Worth flagging as a trap for anyone extending this suite.

No regression added for the service-core path — it needs the `ServiceCoreFdFixture` machinery in `test_service_core_manager.cpp` under fast dispatch, which does not compose with this suite's HYBRID slow-dispatch fixture. Happy to add one there as a follow-up if you would rather have it covered.

abhullar-tt (approved):
lgtm, @yugaoTT could you confirm you are okay with these changes for hybrid mode?

tt-aho (approved):

yugaoTT:
so it solves one problem: you allocated some PCA tensors first, then when allocate LSA, if cores not overlap with those PCAs then we get free space.
one concern i had, with blaze is do we need to guard the mcast op (or any future ops that does full grid mcast) to always allocate the tensor on all cores in stead of a subset, otherwise we might see data corruption. That might be hard to capture if somehow it happened

abhullar-tt:
> t solves one problem: you allocated some PCA tensors first, then when allocate LSA, if cores not overlap with those PCAs then we get

I don't think its easy to capture this right now but when most things most to the new Metal 2.0 apis we would have information on which tensors are bound to kernels. This can help us check for things like invalid mcasting. For now though I don't think it is safe

sjameelTT (approved):

tt_metal/impl/allocator/allocator.cpp
```diff
@@ -178,15 +179,69 @@ DeviceAddr AllocatorImpl::allocate_buffer(Buffer* buffer) {
...
+            // In HYBRID mode the per-core allocators hand out addresses this one cannot see, so
+            // gather their occupied ranges and keep the lockstep address clear of them.
+            //
+            // By default that means every bank, since an op may reach the buffer on a core outside
+            // its own shard grid; experimental/range_lockstep_allocation/buffer.hpp covers when a
+            // buffer may instead be scanned against just the cores it occupies.
+            //
+            // Either way the scan spans devices: a mesh buffer holds the same address on all of them.
             std::vector<std::pair<DeviceAddr, DeviceAddr>> additional_ranges;
             if (!hybrid_device_allocators_.empty()) {
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/impl/range_lockstep_allocation/buffer.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include <tt_stl/assert.hpp>
+
+namespace tt::tt_metal::experimental::range_lockstep_allocation {
+
+BufferShardingArgs& set_range_lockstep_allocation(BufferShardingArgs& args, bool enable) {
+    if (enable) {
+        TT_FATAL(
+            !per_core_allocation::is_per_core_allocation(args),
+            "range_lockstep_allocation and per_core_allocation are mutually exclusive: a buffer either takes one "
+            "address across its cores or an independent address on each");
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/impl/buffers/buffer.cpp
```diff
@@ -429,7 +429,8 @@ Buffer::Buffer(
     page_size_(page_size),
     shard_spec_(sharding_args.shard_spec()),
     buffer_distribution_spec_(sharding_args.buffer_distribution_spec()),
-    per_core_allocation_(experimental::per_core_allocation::is_per_core_allocation(sharding_args)) {
+    per_core_allocation_(experimental::per_core_allocation::is_per_core_allocation(sharding_args)),
+    range_lockstep_allocation_(experimental::range_lockstep_allocation::is_range_lockstep_allocation(sharding_args)) {
     TT_FATAL(this->device_ != nullptr, "Device needs to not be null.");
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/api/tt-metalium/tensor/spec/memory_config/memory_config.hpp
```diff
@@ -52,13 +52,15 @@ class MemoryConfig final {
...
+        "per_core_allocation",
+        "range_lockstep_allocation");
     std::tuple<
         const TensorMemoryLayout&,
         const BufferType&,
         const std::optional<ShardSpec>&,
         const std::optional<NdShardSpec>&,
         const bool&,
+        const bool&,
         const bool&>
```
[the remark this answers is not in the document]
ronnie00TT: Metalium promises API (source-level) stability, not ABI stability — see `.github/instructions/metal-api.instructions.md`, and commit `6955404` which specifically reframed this and added the note that binary-level changes (struct/class field layout, `sizeof()`, vtable order) are not release blockers, since downstream consumers recompile against the headers. This change is also entirely under `experimental/`, which carries no API-stability guarantee at all.

tt_metal/api/tt-metalium/buffer.hpp
```diff
@@ -171,13 +177,22 @@ class BufferShardingArgs {
...
 private:
     std::optional<BufferDistributionSpec> buffer_distribution_spec_;
     std::optional<ShardSpecBuffer> shard_spec_;
     TensorMemoryLayout buffer_layout_ = TensorMemoryLayout::INTERLEAVED;
     // per_core_allocation is experimental functionality
     // access is through experimental::per_core_allocation free functions
     bool per_core_allocation_ = false;
+    // range_lockstep_allocation is experimental functionality
+    // access is through experimental::range_lockstep_allocation free functions
+    bool range_lockstep_allocation_ = false;
```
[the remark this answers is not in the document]
ronnie00TT: Metalium promises API (source-level) stability, not ABI stability — see `.github/instructions/metal-api.instructions.md`, and commit `6955404` which specifically reframed this and added the note that binary-level changes (struct/class field layout, `sizeof()`, vtable order) are not release blockers, since downstream consumers recompile against the headers. This change is also entirely under `experimental/`, which carries no API-stability guarantee at all.

tt_metal/impl/allocator/bank_manager.cpp
```diff
@@ -495,28 +495,46 @@ uint64_t BankManager::allocate_buffer(
...
+            placeable_bytes += (r.second - r.first);
+            largest_placeable = std::max(largest_placeable, r.second - r.first);
+        }
         TT_FATAL(
             false,
             "Out of Memory: Not enough space after considering dependencies to allocate {} B {} across {} banks ({} B "
-            "per bank), bank size is {} B (allocated: {} B, free: {} B, largest free block: {} B)",
+            "per bank), bank size is {} B (allocated: {} B, free: {} B, largest free block: {} B). After subtracting "
+            "{} dependency and {} additional occupied range(s), {} B remained placeable across {} window(s), largest "
+            "{} B",
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/api/tt-metalium/experimental/range_lockstep_allocation/memory_config.hpp
```diff
@@ -0,0 +1,19 @@
...
+#pragma once
+
+#include <tt-metalium/tensor/spec/memory_config/memory_config.hpp>
+
+namespace tt::tt_metal::experimental::range_lockstep_allocation {
+
+// MemoryConfig free functions — access experimental range lockstep state on MemoryConfigImpl.
+//
+// See experimental/range_lockstep_allocation/buffer.hpp for what the mode means. Setting it here
+// is how a tensor carries the request down to buffer allocation.
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/api/tt-metalium/experimental/range_lockstep_allocation/buffer.hpp
```diff
@@ -0,0 +1,27 @@
...
+// keeps that address clear of per-core allocations on EVERY core, which is what an op needs when
+// it addresses the buffer on cores outside its own shard grid — a multicast, for instance, writes
+// to every core in its rectangle whether or not that core is a destination.
+//
+// Range lockstep narrows that guarantee to the cores the buffer actually occupies. Use it only
+// when nothing reaches the buffer on a core it was not allocated on; otherwise the placement may
+// land on top of a per-core allocation elsewhere on the grid.
+//
+// Experimental, and expected to become the default once ops that address buffers across cores
+// declare the region they reach.
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/impl/allocator/allocator.cpp
```diff
@@ -178,15 +179,69 @@ DeviceAddr AllocatorImpl::allocate_buffer(Buffer* buffer) {
...
+                        // so a core's bank id here is not necessarily its bank id there once the
+                        // compute/dispatch split or the harvesting differs across the mesh.
+                        //
+                        // A core with no bank contributes nothing: a service core claimed by fast
+                        // dispatch is a legal shard core with real L1, but the allocator gives it no
+                        // bank, so it holds no per-core ranges to avoid.
+                        if (!dev_alloc->has_bank(BufferType::L1, core)) {
+                            continue;
+                        }
+                        gather_from(dev_alloc, dev_alloc->logical_core_to_bank_ids_.at(BufferType::L1).at(core).at(0));
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

tt_metal/impl/allocator/allocator.cpp
```diff
@@ -178,15 +179,69 @@ DeviceAddr AllocatorImpl::allocate_buffer(Buffer* buffer) {
...
+                    auto ranges = dev_alloc->get_l1_allocated_ranges(AllocatorID{bank_id + 1});
+                    additional_ranges.insert(additional_ranges.end(), ranges.begin(), ranges.end());
+                };
+
+                std::vector<CoreCoord> cores_to_scan;
+                const bool scope_to_own_cores =
+                    experimental::range_lockstep_allocation::is_range_lockstep_allocation(*buffer);
+                // A distribution spec reserves L1 on its cores_with_data(), which is what
+                // Buffer::num_cores() reports for it; shard_spec_ is not set on that path.
+                if (const auto& distribution_spec = buffer->buffer_distribution_spec();
```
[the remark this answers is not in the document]
ronnie00TT: Fixed

bbradelTT (approved):

riverwuTT (approved):

yugaoTT (approved):

abhullar-tt (approved):

nsextonTT (approved):

tt_metal/impl/tensor/spec/tensor_spec.cpp
```diff
@@ -280,6 +281,24 @@ void TensorSpec::populate_sharding_specs() {
...
 }
 
+namespace {
+
+// create_memory_config_with_prepopulated_shard_specs() rebuilds a MemoryConfig from a handful of
+// named fields, which drops the experimental allocation flags. Every rebuild on the sharding-spec
+// conversion paths has to put them back, or an opt-in silently degrades to the default at
+// allocation time with nothing on the way there reporting it.
+MemoryConfig restore_experimental_flags(const MemoryConfig& source, MemoryConfig rebuilt) {
+    if (tt::tt_metal::experimental::per_core_allocation::is_per_core_allocation(source)) {
```
vtsilytskyiTT:
```suggestion
MemoryConfig restore_experimental_flags(const MemoryConfig& source, MemoryConfig rebuilt) {
    using namespace tt::tt_metal::experimental;
    if (per_core_allocation::is_per_core_allocation(source)) {
```
Nit1: `using namespace` might help a little.
Nit2: function definition is not nice. You take a source (const ref, that is OK) and destination as a copy, to return that copy. What is more, function is always used together with `create_memory_config_with_prepopulated_shard_specs`. I suggest to wrap `create_memory_config_with_` and add experimental flags handling. Or even extend `create_memory_config_with_` function with your 2 flags.

tests/pipeline_reorg/ttnn_sanity_tests.yaml
```diff
@@ -265,11 +265,11 @@
...
 
 - name: "core ttnn unit test group"
-  # per_core_allocation runs as its own pytest process: TT_METAL_ALLOCATOR_MODE_HYBRID is read
+  # The allocator suites run as their own pytest process: TT_METAL_ALLOCATOR_MODE_HYBRID is read
   # once when MetalContext is first constructed, so it must be set before this process opens a
   # device. Exporting it here rather than relying on the directory conftest's fixture also makes
   # the requirement visible, and satisfies the requires_hybrid_allocator marker on those tests.
-  cmd: 'pytest --timeout 300 tests/ttnn/unit_tests/base_functionality tests/ttnn/unit_tests/benchmarks tests/ttnn/unit_tests/tensor -xv -m "not disable_fast_runtime_mode" && TT_METAL_ALLOCATOR_MODE_HYBRID=1 pytest --timeout 300 tests/ttnn/unit_tests/per_core_allocation -xv'
+  cmd: 'pytest --timeout 300 tests/ttnn/unit_tests/base_functionality tests/ttnn/unit_tests/benchmarks tests/ttnn/unit_tests/tensor -xv -m "not disable_fast_runtime_mode" && TT_METAL_ALLOCATOR_MODE_HYBRID=1 pytest --timeout 300 tests/ttnn/unit_tests/per_core_allocation -xv && TT_METAL_ALLOCATOR_MODE_HYBRID=1 pytest --timeout 300 tests/ttnn/unit_tests/range_lockstep_allocation -xv'
```
vtsilytskyiTT, line 272:
```suggestion
  cmd: 'pytest --timeout 300 tests/ttnn/unit_tests/base_functionality tests/ttnn/unit_tests/benchmarks tests/ttnn/unit_tests/tensor -xv -m "not disable_fast_runtime_mode" && TT_METAL_ALLOCATOR_MODE_HYBRID=1 pytest --timeout 300 tests/ttnn/unit_tests/per_core_allocation tests/ttnn/unit_tests/range_lockstep_allocation -xv'
```
Unless you explicitly want a fresh & clean execution environment.

tests/ttnn/unit_tests/range_lockstep_allocation/test_range_lockstep_allocation.py
```diff
@@ -0,0 +1,197 @@
...
+        try:
+            holder.append(_allocate(mesh, core, num_bytes, per_core=True))
+        except RuntimeError as exc:
+            # HYBRID is latched at the first MetalContext construction, so the env var the fixture
+            # sets is too late if an earlier test in the same pytest process already opened a
+            # device. Skip rather than fail: under LOCKSTEP no per-core ranges are gathered at all,
+            # and these tests would prove nothing. Per-core allocation is what reports it.
+            if "AllocatorMode::HYBRID" not in str(exc):
+                raise
+            pytest.skip("HYBRID allocator mode is not active in this process")
```
vtsilytskyiTT: No `requires_hybrid_allocator` marker. Sibling `per_core_allocation` tests use it. Here a shared pytest process that already opened a device just skips (`AllocatorMode::HYBRID` in the hog path). The yaml job is fine; a local `pytest tests/ttnn/unit_tests` is easy to misread as a pass.

ttnn/core/tensor/flatbuffer/tensor_spec.fbs
```diff
@@ -103,6 +103,7 @@ table MemoryConfig {
     nd_shard_spec: NdShardSpec;
     created_with_nd_shard_spec: bool;
     per_core_allocation: bool = false;
+    range_lockstep_allocation: bool = false;
```
vtsilytskyiTT, line 106: Nit: no Python getter. Fine if you keep testing via JSON; a tiny `experimental_is_range_lockstep_allocation` would make e2e tests able to assert the flag on the allocated tensor as a second signal.

tests/pipeline_reorg/ttnn_sanity_tests.yaml
ronnie00TT: Tried that first — it's what this PR had, and CI aborted the whole pytest process on `bh_p150b` twice (exit -6). The other suite's `expect_error` tests leave tensors of a closed device in a reference cycle; the collector reaches them when the next fixture opens a device and the destructor calls `std::terminate`. Separate processes contain it, at ~10s.

tt_metal/impl/tensor/spec/tensor_spec.cpp
ronnie00TT: Fixed

tests/ttnn/unit_tests/range_lockstep_allocation/test_range_lockstep_allocation.py
ronnie00TT: Fixed

ttnn/core/tensor/flatbuffer/tensor_spec.fbs
ronnie00TT: Fixed

tests/tt_metal/tt_metal/api/allocator/test_per_core_allocation.cpp
```diff
@@ -138,6 +146,257 @@ TEST_F(PerCoreAllocationTest, PerCoreAndLockstepCoexist) {
...
+namespace {
+MemoryConfig sharded_l1_memory_config() {
+    return MemoryConfig(
+        TensorMemoryLayout::HEIGHT_SHARDED,
+        BufferType::L1,
+        ShardSpec(CoreRangeSet(CoreCoord(0, 0)), {32, 32}, ShardOrientation::ROW_MAJOR));
+}
+}  // namespace
+
+TEST_F(PerCoreAllocationTest, RangeLockstepMemoryConfigReachesShardingArgs) {
```
obacherikovTT: I have a request regarding tests since you're mentioning that
> Longer term, the intent is for range lockstep to become the default

Could you please move range lockstep tests that don't interact with per-core allocation into a separate test file? And it would be really helpful if they clearly showed the difference between outcomes depending on `set_range_lockstep_allocation` being called or not.

tests/tt_metal/tt_metal/api/tensor/test_tensor_sharding.cpp
```diff
@@ -232,14 +240,18 @@ TEST_F(MemoryConfigEqualityTest, MixedCreationPath_NdFieldDiffers) {
         BufferType::L1,
         shard_spec_a_,
         nd_shard_spec_a_,
-        /*created_with_nd_shard_spec=*/true);
+        /*created_with_nd_shard_spec=*/true,
+        /*per_core_allocation=*/false,
+        /*range_lockstep_allocation=*/false);
```
obacherikovTT: This is likely out of scope of this PR, but having 3 `bool` parameters next to each other is very error-prone (and you used comments to distinguish them).

It would be safer for `create_memory_config_with_prepopulated_shard_specs` to take a struct and use designated initializers to specify named parameters and have compiler to always check them.

obacherikovTT (commented):
Could you please create an issue to track this initiative?

> Longer term, the intent is for range lockstep to become the default once ops that reach a buffer across cores declare the region they reach, at which point the toggle goes away and full-grid lockstep becomes the multicast special case. This PR is the mechanism only.

ronnie00TT:
@obacherikovTT Thanks for pointing it out, I've made an issue #55253 to track this

tests/tt_metal/tt_metal/api/allocator/test_per_core_allocation.cpp
ronnie00TT: I've moved tests that don't interact with PCA to a separate file. And I've moved tests with `set_range_lockstep_allocation` on/off to one [block](https://github.com/tenstorrent/tt-metal/blob/021200e4cf5d627fc01728d2b3f755dab07742c3/tests/tt_metal/tt_metal/api/allocator/test_range_lockstep_allocation.cpp#L127) in `tests/tt_metal/tt_metal/api/allocator/test_range_lockstep_allocation.cpp`

tests/tt_metal/tt_metal/api/tensor/test_tensor_sharding.cpp
ronnie00TT: Thanks for pointing it out, I have fixed it.

tests/pipeline_reorg/ttnn_sanity_tests.yaml
vtsilytskyiTT: I see. Then a separate process.

vtsilytskyiTT (approved):
Core TTNN part approved.

tests/tt_metal/tt_metal/api/allocator/test_per_core_allocation.cpp
obacherikovTT: Thank you!
I see that you also added new test executable. It makes it harder to run all the allocation tests together, so I'm curious why decided to add it.

obacherikovTT (approved):

tests/tt_metal/tt_metal/api/allocator/test_per_core_allocation.cpp
ronnie00TT: Thanks! I've combined several executables into one.

PR #54480, opened 2026-08-26
[Feature] Metal 2.0: Implement variadic tensor parameters

riverwuTT:
### PR Category
Feature

### Summary
Add Metal 2.0 variadic tensor parameters: a compile-time pack of existing `TensorBinding`s.

Named single-tensor bindings stay the default. A group is not a new binding kind and not a run-arg; `ProgramRunArgs.tensor_args` is unchanged. Codegen emits each group into the same `tensor::` namespace as a `constexpr std::tuple` of member tokens in `members` order (empty members → `std::tuple<>`).

Constraints (checked at `MakeProgramFromSpec`):
- Every `members` entry is a `TensorBinding::accessor_name` on this kernel
- No duplicate members within a group (empty and size 1 are allowed)
- One binding may appear in more than one group
- Group `accessor_name` is a valid C++ identifier
- Group `accessor_name` is unique in this kernel's `tensor::` namespace (must not collide with a binding name or another group)

**Host**

```cpp
KernelSpec reader{
    .tensor_bindings = {
        TensorBinding{.tensor_parameter_name = IN0, .accessor_name = "in0"},
        TensorBinding{.tensor_parameter_name = IN1, .accessor_name = "in1"},
        TensorBinding{.tensor_parameter_name = OUTPUT, .accessor_name = "output"},
    },
    .advanced_options = {
         // NEW: declares a variadic tensor group.
        .tensor_groups = {
            {
                .accessor_name = "inputs",
                .members = {"in0", "in1"},
            },
        },
    },
};
```

**Device**

```cpp
namespace tensor {
constexpr TensorBindingToken<...> in0{};
constexpr TensorBindingToken<...> in1{};
constexpr TensorBindingToken<...> output{};
// NEW: tensor group — tuple order matches host `members` order
constexpr auto inputs = std::make_tuple(in0, in1);
}
```

### Notes for reviewers
- Public API lives only on `KernelAdvancedOptions::tensor_groups` in `advanced_options.hpp`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/m2-var-dfb-tensor)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/m2-var-dfb-tensor)

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -3979,14 +3979,14 @@ void kernel_main() {
 }
 
 // ============================================================================
-// TensorGroup: validation + JIT smoke (compile-only, no hardware loop)
+// TensorBindingSequence: validation + JIT smoke (compile-only, no hardware loop)
 // ============================================================================
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupSeveralMembersJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceSeveralMembersJITSmoke) {
     NodeCoord node{0, 0};
 
     ProgramSpec spec;
-    spec.name = "tensor_group_several";
+    spec.name = "tensor_binding_sequence_several";
 
     auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
     dm_kernel.source = KernelSpec::SourceCode{R"(
@@ -3999,8 +3999,8 @@ void kernel_main() {
     static_assert(std::is_same_v<std::tuple_element_t<2, inputs_t>, tensor::in2_t>);
 }
 )"};
-    dm_kernel.advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "inputs", .members = {"in0", "in1", "in2"}},
+    dm_kernel.advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "inputs", .members = {"in0", "in1", "in2"}},
     };
 
     spec.kernels = {dm_kernel};
@@ -4018,11 +4018,11 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupEmptyMembersJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceEmptyMembersJITSmoke) {
     NodeCoord node{0, 0};
 
     ProgramSpec spec;
-    spec.name = "tensor_group_empty";
+    spec.name = "tensor_binding_sequence_empty";
 
     auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
     dm_kernel.source = KernelSpec::SourceCode{R"(
@@ -4032,8 +4032,8 @@ void kernel_main() {
     static_assert(std::is_same_v<std::remove_cv_t<decltype(tensor::empty)>, std::tuple<>>);
 }
 )"};
-    dm_kernel.advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "empty", .members = {}},
+    dm_kernel.advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "empty", .members = {}},
     };
 
     spec.kernels = {dm_kernel};
@@ -4043,11 +4043,11 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupSingletonMembersJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceSingletonMembersJITSmoke) {
     NodeCoord node{0, 0};
 
     ProgramSpec spec;
-    spec.name = "tensor_group_singleton";
+    spec.name = "tensor_binding_sequence_singleton";
 
     auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
     dm_kernel.source = KernelSpec::SourceCode{R"(
@@ -4058,8 +4058,8 @@ void kernel_main() {
     static_assert(std::is_same_v<std::tuple_element_t<0, solo_t>, tensor::in0_t>);
 }
 )"};
-    dm_kernel.advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "solo", .members = {"in0"}},
+    dm_kernel.advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "solo", .members = {"in0"}},
     };
 
     spec.kernels = {dm_kernel};
@@ -4071,11 +4071,11 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupSameBindingInTwoGroupsJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceSameBindingInTwoSequencesJITSmoke) {
     NodeCoord node{0, 0};
 
     ProgramSpec spec;
-    spec.name = "tensor_group_shared_member";
+    spec.name = "tensor_binding_sequence_shared_member";
 
     auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
     dm_kernel.source = KernelSpec::SourceCode{R"(
@@ -4089,9 +4089,9 @@ void kernel_main() {
     static_assert(std::is_same_v<std::tuple_element_t<0, g1_t>, tensor::in0_t>);
 }
 )"};
-    dm_kernel.advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "g0", .members = {"in0"}},
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "g1", .members = {"in0"}},
+    dm_kernel.advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "g0", .members = {"in0"}},
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "g1", .members = {"in0"}},
     };
 
     spec.kernels = {dm_kernel};
@@ -4103,15 +4103,15 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupOnComputeKernelJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceOnComputeKernelJITSmoke) {
     ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
     ASSERT_TRUE(spec.kernels[1].is_compute_kernel());
 
     spec.tensor_parameters = {MakeMinimalTensorParameter("t0"), MakeMinimalTensorParameter("t1")};
     BindTensorParameterToKernel(spec.kernels[1], "t0", "in0");
     BindTensorParameterToKernel(spec.kernels[1], "t1", "in1");
-    spec.kernels[1].advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "inputs", .members = {"in0", "in1"}},
+    spec.kernels[1].advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "inputs", .members = {"in0", "in1"}},
     };
     spec.kernels[1].source = KernelSpec::SourceCode{R"(
 #include <type_traits>
@@ -4127,14 +4127,14 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupNameEqualsDfbAccessorJITSmoke) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceNameEqualsDfbAccessorJITSmoke) {
     ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
 
     spec.tensor_parameters = {MakeMinimalTensorParameter("t0")};
     BindTensorParameterToKernel(spec.kernels[0], "t0", "in0");
     // Minimal program already binds dfb accessor "input_dfb" on kernels[0].
-    spec.kernels[0].advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "input_dfb", .members = {"in0"}},
+    spec.kernels[0].advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "input_dfb", .members = {"in0"}},
     };
     spec.kernels[0].source = KernelSpec::SourceCode{R"(
 #include <type_traits>
@@ -4150,12 +4150,12 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
-TEST_F(ProgramSpecTestGen1, CPU_TensorGroupUnknownMemberFails) {
+TEST_F(ProgramSpecTestGen1, CPU_TensorBindingSequenceUnknownMemberFails) {
     ProgramSpec spec = MakeMinimalGen1ValidProgramSpec();
     spec.tensor_parameters = {MakeMinimalTensorParameter("t0")};
     BindTensorParameterToKernel(spec.kernels[0], "t0", "in0");
-    spec.kernels[0].advanced_options.tensor_groups = {
-        KernelAdvancedOptions::TensorGroup{.accessor_name = "inputs", .members = {"in0", "missing"}},
+    spec.kernels[0].advanced_options.tensor_binding_sequences = {
+        KernelAdvancedOptions::TensorBindingSequence{.sequence_name = "inputs", .members = {"in0", "missing"}},
     };
 
     EXPECT_THAT(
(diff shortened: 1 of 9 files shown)


tt_metal/jit_build/jit_build_settings.hpp
```diff
@@ -130,6 +130,11 @@ class JitBuildSettings {
     virtual void process_scratchpad_binding_handles(
         std::function<void(const std::string& accessor_name, uint32_t size_bytes, uint32_t addr_crta_word)>) const {}
 
+    // Tensor group callback: accessor_name + ordered member TensorBinding accessor names.
+    // Emitted as constexpr std::tuple tokens in the `tensor::` namespace (user order; no sort).
+    virtual void process_tensor_groups(
+        std::function<void(const std::string& accessor_name, const std::vector<std::string>& members)>) const {}
```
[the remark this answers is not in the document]
riverwuTT: nop, not of concern in this PR.

tt_metal/impl/kernels/kernel.cpp
```diff
@@ -591,6 +599,15 @@ uint64_t Kernel::compute_hash() const {
         hasher.update(static_cast<uint64_t>(handle.size_bytes));
         hasher.update(static_cast<uint64_t>(handle.addr_crta_word));
     }
+    // Tensor groups: user order matches genfiles emission; hash size, name, and ordered members.
+    hasher.update(static_cast<uint64_t>(this->tensor_groups_.size()));
+    for (const auto& group : this->tensor_groups_) {
+        hasher.update(group.accessor_name);
+        hasher.update(static_cast<uint64_t>(group.members.size()));
+        for (const auto& member : group.members) {
+            hasher.update(member);
```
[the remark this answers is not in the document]
riverwuTT: accepted in 6d7823dfb2e

tt_metal/impl/metal2_host_api/program_spec.cpp
```diff
@@ -533,6 +533,43 @@ CollectedSpecData CollectSpecData(const ProgramSpec& spec) {
...
+            TT_FATAL(
+                IsValidCppIdentifier(group.accessor_name),
+                "Kernel '{}' tensor group accessor_name '{}' must be a valid C++ identifier",
+                kernel.unique_id,
+                group.accessor_name);
+            TT_FATAL(
+                !accessor_names.contains(group.accessor_name),
+                "Kernel '{}' tensor group accessor_name '{}' collides with a TensorBinding accessor_name",
+                kernel.unique_id,
+                group.accessor_name);
```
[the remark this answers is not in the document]
riverwuTT: accepted in 6d7823dfb2e

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,34 @@ struct KernelAdvancedOptions {
     //          Always prefer regular, named compile-time arguments.
     [[deprecated("Compile-time varargs is a temporary feature that will be removed in the future.")]]
     std::vector<uint32_t> compile_time_varargs;
+
+    ////////////////////////////////////////////////////////////////////////////////
+    // Tensor groups
```
akerteszTT: Elsewhere in the API, we use the term "Group" very explicitly to mean a set of things where the order doesn't matter. This is the opposite; we need a different term. You're using the term "pack" here too, which is likewise not a great choice -- it's used extensively in Metal for compute APIs (pack_tile, llk_pack_init, unpack_modes, etc), so an unnecessary overload

I think we should call this a `TensorBindingSequence` - what do you think? It does what it says on the tin.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,34 @@ struct KernelAdvancedOptions {
...
+    //   - No duplicate members within a group
+    //   - Group accessor_name is a valid C++ identifier, unique in this kernel's
+    //     tensor:: namespace.
+    //
+    // CAUTION: This feature exists to address niche use cases only.
+    //          Named TensorBindings remain how a kernel binds tensors; a group
+    //          only packs some of those bindings when the kernel must consume them as a pack.
+
+    struct TensorGroup {
+        std::string accessor_name;         // device: tensor::<accessor_name>
```
akerteszTT: Can you make this `sequence_name` instead? We're starting to have some serious overloading of "accessor", and here it's not necessary.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,34 @@ struct KernelAdvancedOptions {
...
+        std::vector<std::string> members;  // TensorBinding::accessor_name; order is the device tuple order
+    };
+    Group<TensorGroup> tensor_groups;
```
akerteszTT: `Group<TensorBindingSequence> tensor_binding_sequences;`

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,34 @@ struct KernelAdvancedOptions {
     //          Always prefer regular, named compile-time arguments.
     [[deprecated("Compile-time varargs is a temporary feature that will be removed in the future.")]]
     std::vector<uint32_t> compile_time_varargs;
+
+    ////////////////////////////////////////////////////////////////////////////////
+    // Tensor groups
+    ////////////////////////////////////////////////////////////////////////////////
+
+    // A tensor group bunches existing TensorBindings on this kernel into a
```
akerteszTT: I find this comment a bit unclear. It leads with the mechanism rather than the purpose. And also, it gives no indication of how the resulting tuple is intended to be used on the kernel side.

How about this?

```
    // A tensor binding sequence gives a kernel an additional way to retrieve its tensor
    // bindings: positionally, by index, rather than by name.
    //
    // In kernel code, a tensor binding is normally retrieved by name (e.g. `tensor::in0`).
    // Declaring a TensorBindingSequence with a list of tensor binding names also emits the
    // following into the generated header's `tensor` namespace, in the order specified:
    //
    //    constexpr auto my_binding_sequence = std::make_tuple(in0, in1, /* ... */ inN);
    //
    // To make use of this, the kernel then calls:
    //   /* create a tuple of TensorAccessor from the binding token tuple */
    //   auto accessor_tuple =  make_tensor_accessors(my_binding_sequence); 
    //   /* create an array non-owning, type-erased TensorAccessor handles */
    //   auto accessor_array = make_abstract_tensor_accessor_wrappers(accessor_tuple); 
    //
    // Usage: A niche mechanism for a kernel that wishes to express a compile-time-variadic number of
    //        tensor bindings, and therefore needs to access them positionally. Prefer the default
    //        named TensorBindingToken access whenever possible.
    //
    // Notes:
    //   - The named tokens are still emitted, so a sequence adds positional access rather than
    //      replacing named access. 
    //   - A separate count argument is not required, as the sequence is available kernel-side via
    //      `std::tuple_size_v<decltype(tensor::my_binding_sequence)>`
    //
    // Constraints:
    //   - Every `members` entry is a TensorBinding::accessor_name on this kernel
    //   - No duplicate members within a sequence (one binding may appear in several sequences)
    //   - `accessor_name` is a valid C++ identifier, unique in this kernel's `tensor::` namespace
    //   - An empty members list is legal; this produces an empty std::tuple<>.
    //
```

akerteszTT (commented):
Looks great overall.

Partial review -- API surface nits only. Will follow up with a few code-side nits :)

tt_metal/jit_build/genfiles.cpp
```diff
@@ -230,20 +243,32 @@ void write_kernel_bindings_generated_header(const string& out_dir, const JitBuil
...
+            //
+            // Tensor groups are constexpr std::tuple packs of those member tokens (members order).
             content << "namespace tensor {\n";
             for (const auto& entry : ta_entries) {
                 content << "using " << entry.name << "_t = ::tensor_accessor::TensorBindingToken<" << entry.cta_offset
                         << "u, " << entry.addr_crta_offset << "u>;\n";
                 content << "constexpr " << entry.name << "_t " << entry.name << "{};\n";
             }
+            for (const auto& group : tg_entries) {
+                content << "constexpr auto " << group.name << " = std::make_tuple(";
```
ruizhangTT: nit: possible to use fmt::join here?

ruizhangTT (approved):

tt_metal/impl/kernels/kernel.cpp
```diff
@@ -346,6 +346,14 @@ void Kernel::process_scratchpad_binding_handles(
     }
 }
 
+void Kernel::process_tensor_groups(
```
akerteszTT: When you rename this, please make sure to get rid of the word "group" everywhere that it appears in the implementation... there's lots of them.

tt_metal/impl/kernels/kernel.cpp
```diff
@@ -591,6 +599,17 @@ uint64_t Kernel::compute_hash() const {
         hasher.update(static_cast<uint64_t>(handle.size_bytes));
         hasher.update(static_cast<uint64_t>(handle.addr_crta_word));
     }
+    // Tensor groups: user order matches genfiles emission; hash size, name, and ordered members.
+    // Per-member size is hashed before the bytes so {"a","bc"} and {"ab","c"} do not collide.
+    hasher.update(static_cast<uint64_t>(this->tensor_groups_.size()));
+    for (const auto& group : this->tensor_groups_) {
```
akerteszTT: This is correct! But, definitely worth a comment to emphasize that order matters for the hash. Other vectors of things that go into the kernel hash get sorted first -- i.e. true "groups" where the order is unimportant, like named CTAs.

Seems like an easy error some someone to introduce later if the reasoning isn't made clear.

All the more reason to get rid of "group" everywhere :)

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,34 @@ struct KernelAdvancedOptions {
...
+    // compile-time-sized pack.
+    //
+    // Each group is emitted into the same `tensor::` namespace as the member
+    // tokens, as a constexpr std::tuple of those tokens in `members` order:
+    //   constexpr auto inputs = std::make_tuple(in0, in1, /* ... */ inN);
+    // An empty members list is std::tuple<>.
```
akerteszTT: Do you have a use case in mind for allowing an empty list? Just curious, no objection.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -172,6 +172,15 @@ void write_kernel_bindings_generated_header(const string& out_dir, const JitBuil
             scratch_entries.push_back({name, size_bytes, addr_crta_word});
         });
 
+    // Tensor groups: user order (matches Kernel::compute_hash); no sort.
+    struct TgEntry {
```
akerteszTT: Acronym will need to change when renaming "group". 

Also, please consider just writing it out in full. I'm still regretting my obfuscated "TA" acronym that you doubtless copied here... which is particularly unfortunate since TensorAccessor isn't the same as TensorBindingToken.

tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
```diff
@@ -3978,6 +3978,294 @@ void kernel_main() {
...
+    NodeCoord node{0, 0};
+
+    ProgramSpec spec;
+    spec.name = "tensor_group_several";
+
+    auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
+    dm_kernel.source = KernelSpec::SourceCode{R"(
+#include <type_traits>
+void kernel_main() {
+    using inputs_t = std::remove_cv_t<decltype(tensor::inputs)>;
```
akerteszTT, line 3996: Humans and AIs might look to these tests to figure out how to use this construct in kernel code. This is a bit of C++ esoterica I definitely had to look up. 

Can you please add a comment everywhere that this appears to emphasize that this is a test-only thing for asserting on, not something to use in real kernel code?

akerteszTT (commented):
Left a few more minor nits, mostly concerning the "Group" rename.

One high-level comment:
It wasn't at all obvious to me how the kernel code should make use of the generated tuple of `TensorBindingTokens`. I've documented the how-to in the public header comment (my suggested rewrite). 

In doing so... I now think that for feature to be complete, we need a lightweight helper method in `tensor_accessor.h` to convert the tuple of binding tokens into the tuple of TensorAccessor that the existing `make_abstract_tensor_accessor_wrappers ` thingy wants:

```
template <typename... Tokens>
auto make_tensor_accessors(const std::tuple<Tokens...>& tokens) {
    return std::apply([](const auto&... toks) { return std::make_tuple(TensorAccessor(toks)...); }, tokens);
}
```

tt_metal/jit_build/genfiles.cpp
riverwuTT: Yes will do, this is a piece of AI slop I missed.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: nop, just "why not".

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: good idea, I'll do this.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: adopted!

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: adopted

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: adopted.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
riverwuTT: adopted.

tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
riverwuTT: done

tt_metal/jit_build/genfiles.cpp
riverwuTT: done!

tt_metal/impl/kernels/kernel.cpp
riverwuTT: done!

tt_metal/impl/kernels/kernel.cpp
riverwuTT: done!

tt_metal/jit_build/genfiles.cpp
riverwuTT: done!

ruizhangTT: Hmmm I didn't see any changes here

riverwuTT: ops, didn't push

riverwuTT: See: d0c1aec7a3b015f8f0aeaecc00582011ec25c825

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,47 @@ struct KernelAdvancedOptions {
...
+    // Notes:
+    //   - The named tokens are still emitted, so a sequence adds positional access rather than
+    //      replacing named access.
+    //   - A separate count argument is not required, as the sequence is available kernel-side via
+    //      `std::tuple_size_v<decltype(tensor::my_binding_sequence)>`
+    //
+    // Constraints:
+    //   - Every `members` entry is a TensorBinding::accessor_name on this kernel
+    //   - No duplicate members within a sequence (one binding may appear in several sequences)
+    //   - `accessor_name` is a valid C++ identifier, unique in this kernel's `tensor::` namespace
```
akerteszTT: Oops. Should be `sequence_name`.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,47 @@ struct KernelAdvancedOptions {
...
+    //
+    // In kernel code, a tensor binding is normally retrieved by name (e.g. `tensor::in0`).
+    // Declaring a TensorBindingSequence with a list of tensor binding names also emits the
+    // following into the generated header's `tensor` namespace, in the order specified:
+    //
+    //    constexpr auto my_binding_sequence = std::make_tuple(in0, in1, /* ... */ inN);
+    //
+    // To make use of this, the kernel then calls:
+    //   /* create a tuple of TensorAccessor from the binding token tuple */
+    //   auto accessor_tuple =  make_tensor_accessors(my_binding_sequence);
```
akerteszTT: My bad on this one. Should be `tensor::my_binding_sequence`.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -108,6 +108,47 @@ struct KernelAdvancedOptions {
...
+    //   /* create an array non-owning, type-erased TensorAccessor handles */
```
akerteszTT: Also my bad. Should be "create an array of non-owning". (Missing word)

akerteszTT (approved):
Looks great, thanks @riverwuTT !

I noticed some minor errors in (my) header comments; please fix before merging :)

Also -- no test currently exercises the new helpers `make_tensor_accessors` or `make_local_tensor_accessors`. Probably worth adding a quick unit test before merging.

jbaumanTT (approved):

PR #54489, opened 2026-08-26
Make cluster-health store date dirs writable by the store group

tt-mnijjar:
## Summary
- `report_cluster_health.py` creates `--store-root/<YYYY-MM-DD>/` as `03770` (setgid, sticky, owner/group write, **no world write**) with the store root's group, so the first writer of the day no longer leaves a umask-masked `0755` directory owned by their uid that locks out every later user.
- Only the date directory is chmod'd. The store root and its ancestors are left alone, so a mistyped `--store-root` cannot loosen an unrelated tree.
- Sticky keeps one writer from unlinking another's records. `chown` runs before `chmod` because a non-privileged `chown` can drop setgid.
- `chmod`/`chown` are best effort. A date directory already owned by a different user still fails on the write that follows, exactly as before, with the existing `store write failed` warning.

Record files themselves still follow the caller's umask; the README now calls out that a restrictive umask writes records the log shipper cannot read.

## Test plan
- [ ] `python3 tools/scaleout/exabox/tests/test_report_cluster_health.py`
- [ ] `python3 tools/scaleout/exabox/tests/test_report_backfill.py` (covers concurrent writers into one date dir)
- [ ] First write creates `<store>/<date>/` group-writable, sticky, not world-writable
- [ ] A second user in the store's group can publish another `*.json` into that same date directory
- [ ] An existing `0755` date dir owned by the current user is repaired on the next write
- [ ] A missing `--store-root` ancestor is created without group or other write

## Note on the Cycode findings
Two `Permissive file assignment` comments remain, and the `Cycode: SAST` check passes:
- `os.chmod(date_dir, STORE_DIR_MODE)` — intentional and now group-scoped rather than world-writable, which is what the rule's own guidance recommends ("prefer assigning file permissions to groups rather than other").
- `os.chmod(date_dir, 0o755)` in the tests — a false positive. That line sets up a *restrictive* directory inside a `TemporaryDirectory` so the test can assert the repair path.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mnijjar/cluster-health-shared-store-dir-mode)


mgajewskiTT (approved):
Please answer/resolve AI comments.

PR #54497, opened 2026-08-26
refactor: deduplicate bfloat16 ULP bit-manipulation helpers in eltwise tests

blozano-tt:
## What

Removes ~90 lines of verbatim copy-paste: an identical block of five pure bfloat16 bit-manipulation helpers was duplicated in four eltwise ULP test files. This PR moves the block into the existing shared helper module and imports it instead. Net delta: **+104 / −296**.

**Functions moved** (byte-identical relocation — bodies untouched):
- `float_to_bf16_bits`
- `bf16_bits_to_float`
- `is_bf16_denormal`
- `bf16_daz_normalize`
- `bf16_value_order_index_daz`
- `ulp_distance_bf16_daz` — 4 copies
- `bf16_quantize_rne` — 2 copies (the two `tanh` files only)

**Files deduplicated** (all in `tests/ttnn/unit_tests/operations/eltwise/`):
- `test_tanh_bw_ulp.py`
- `test_tanh_fw_ulp.py`
- `test_gelu_bw_ulp.py`
- `test_gelu_bw_main_ulp.py`

All four copies were confirmed byte-identical (SHA-256 of the extracted block matches across all four files) before extraction.

## Why `eltwise_test_utils.py`

- Same directory as all four consumers, and already the established import target (`test_unary_category1_bfloat16.py` imports from it today).
- It already hosts bfloat16 bit-pattern helpers (`generate_bfloat16_bits`, `flush_to_zero`, `generate_bfloat16_bits_in_range`) — same domain.
- Follows the existing two-tier convention: generic ttnn helpers in `tests/ttnn/utils_for_testing.py`, eltwise-specific helpers here. No pre-existing equivalent of these five exists in either tier, so this creates no third variant.

Notes:
- `import struct` was removed from all four test files — it was used only inside the moved block.
- Each file's import list is derived from what it actually calls. `is_bf16_denormal` and `bf16_value_order_index_daz` are no longer imported anywhere: both are only reached internally, via `bf16_daz_normalize` and `ulp_distance_bf16_daz` respectively.
- `import torch` stays in all four files — still used outside the moved block.

## Verification

These tests require Tenstorrent hardware, which was not available — **the tests were not executed**. Static checks performed instead:
- `python -m py_compile` passes on all five touched files.
- The moved functions were exercised in isolation (extracted via AST, since `torch`/`ttnn` are unavailable off-hardware): round-trip 1.0 ↔ 0x3F80, denormal detection, DAZ normalization of denormals and −0, and zero's order index all behave as before.
- `ulp_distance_bf16_daz` was diffed against its pre-move implementation across **every BF16 bit pattern (65,536) × 11 probe values** — including ±0, ±inf, NaN and a denormal — for **721,017 pairs, 0 mismatches**. `bf16_quantize_rne` needs `torch` at runtime so it could not be executed off-hardware; it is a byte-identical move into a module that already imports `torch`.
- `black==23.10.1` (the version pinned in `.pre-commit-config.yaml`) reports all files in the eltwise directory unchanged; no trailing whitespace; all files end with a newline.
- Import path resolves by the same mechanism as the existing `from tests.ttnn.unit_tests.operations.eltwise.eltwise_test_utils import ...` usage.

CI should be relied on to run the actual hardware tests.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dedupe-bf16-ulp-helpers)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dedupe-bf16-ulp-helpers)

### DIFF
--- a/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
@@ -107,7 +107,7 @@ def to_tt_tensor(
 
 
 def float_to_bf16_bits(f: float) -> int:
-    """Convert float to BFloat16 bit representation."""
+    """Convert float to BFloat16 bits by truncating the lower 16 FP32 mantissa bits."""
     f32_bits = struct.unpack(">I", struct.pack(">f", f))[0]
     return f32_bits >> 16
 
--- a/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
@@ -154,3 +154,30 @@ def bf16_value_order_index_daz(bits: int) -> int:
         return 0x7F7F - magnitude
     else:
         return 32640 + bits - 0x007F
+
+
+def ulp_distance_bf16_daz(a: float, b: float) -> int:
+    """Calculate ULP distance with DAZ+FTZ model."""
+    a_bits = bf16_daz_normalize(float_to_bf16_bits(a))
+    b_bits = bf16_daz_normalize(float_to_bf16_bits(b))
+
+    a_exp = (a_bits >> 7) & 0xFF
+    b_exp = (b_bits >> 7) & 0xFF
+    if (a_exp == 0xFF and (a_bits & 0x7F) != 0) or (b_exp == 0xFF and (b_bits & 0x7F) != 0):
+        return -1
+
+    idx_a = bf16_value_order_index_daz(a_bits)
+    idx_b = bf16_value_order_index_daz(b_bits)
+
+    if idx_a < 0 or idx_b < 0:
+        return -1
+
+    return abs(idx_a - idx_b)
+
+
+def bf16_quantize_rne(x: float) -> float:
+    """RNE-quantize a float to BF16 (matches torch's BFloat16 conversion).
+    Required because the bit-level helpers above truncate, but torch — and therefore
+    the device input — uses round-to-nearest-even. For test points that are not
+    exact BF16 values (e.g., 2.9, 3.01), truncation and RNE diverge."""
+    return float(torch.tensor([x], dtype=torch.bfloat16).item())
--- a/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import struct
 import torch
 import ttnn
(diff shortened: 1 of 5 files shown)


tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
```diff
@@ -103,3 +104,80 @@ def to_tt_tensor(
         layout=layout,
         memory_config=memory_config,
     )
+
+
+def float_to_bf16_bits(f: float) -> int:
+    """Convert float to BFloat16 bit representation."""
```
Aswinmcw: Now that this is a shared helper, worth stating in the docstring that this **truncates** rather than round-to-nearest-even. The divergence is currently only documented inside `bf16_quantize_rne` below, so a caller who imports just this one gets no warning that it disagrees with `torch.tensor(x, dtype=torch.bfloat16)` for non-exact BF16 inputs.

Aswinmcw (approved):
One doc-level note: the opening line still says "an identical block of **five** pure bfloat16 bit-manipulation helpers", but the list below it (and the diff) moves seven — `ulp_distance_bf16_daz` and `bf16_quantize_rne` included. Both automated reviewers read the stale count and reported that `ulp_distance_bf16_daz` "remains duplicated across the four files" and was "intentionally left out of scope", which is not what the diff does. Worth fixing the count so the record isn't misleading.

Verified independently: all seven bodies are byte-identical to their `main` counterparts across all four sources, every imported name is used in each consumer, `struct` is unused after the move in all four, and the import path matches the existing `test_unary_category1_bfloat16.py` usage.

vtsilytskyiTT (approved):
LGTM

mateusznowakTT (approved):

VirdhatchaniKN (approved):

tests/ttnn/unit_tests/operations/eltwise/eltwise_test_utils.py
blozano-tt: @copilot please address this comment.

PR #54510, opened 2026-08-27
refactor: deduplicate mesh-packing machinery in topology_mapper_utils.cpp

blozano-tt:
### What

`tt_metal/fabric/topology_mapper_utils.cpp` contained **124 lines of byte-identical duplication**: the two `build_physical_multi_mesh_adjacency_graph` overloads (single-MGD and multi-MGD) each carried their own private copy of the same three entities (verified byte-identical by `diff` before the change):

| entity | copy A (old line) | copy B (old line) | size |
|---|---|---|---|
| `struct MeshEnumState` | 993 | 1817 | 11 lines |
| lambdas `bitset_disjoint`, `mark_used`, `unmark` | 1103–1125 | 1943–1965 | 23 lines |
| `struct DisjointPackingSearch` | 1217 | 2057 | 90 lines |

This PR hoists a single copy of all three into the file-level anonymous namespace (next to `mesh_physical_layouts_from_psd_placements`, which both overloads already share) and deletes both function-local copies. Internal linkage is preserved. **Net delta: +191 / −352 = −161 lines** (the extra savings beyond 124x2 come from the duplicated doc comments moving with the code).

### The coupling and how it was resolved

The three entities form a dependency cluster: `DisjointPackingSearch` held members typed via `decltype` of the enclosing function's lambdas (`decltype(bitset_disjoint)& is_disjoint;` etc.), and those lambdas captured `asic_word_count` from function scope, so none of them could be moved to file scope unchanged. Resolution:

- The three lambdas become plain free functions with `std::size_t asic_word_count` as an explicit trailing parameter. Bodies are otherwise verbatim; a by-value capture becomes a by-value parameter, so each call computes exactly what it did before.
- `DisjointPackingSearch` replaces the three `decltype(...)&` members with a single `std::size_t asic_word_count;` member and calls the free functions directly. The value stored is the same `const std::size_t asic_word_count` local the lambdas used to capture, passed at the same aggregate-init position where the three lambda references were passed. Everything else in the struct is verbatim.
- `MeshEnumState` is hoisted verbatim except that `MappingConstraints`, `TopologyMappingEnumerationSession`, and `MappingResult` are now spelled `::tt::tt_fabric::...` — at namespace scope the function-local `using namespace ::tt::tt_fabric;` no longer applies. These are the same types the unqualified names resolved to before (the public header `topology_mapper_utils.hpp` imports only `AdjacencyGraph`/`MeshId` into `tt::tt_metal::experimental::tt_fabric`, and itself spells `::tt::tt_fabric::MappingConstraints` for the same reason).

This was chosen over templating `DisjointPackingSearch` on the callable types (which would only recover ~101 of the 124 lines and leave the lambdas duplicated) and over a factory returning lambdas to keep `decltype` alive (more indirection for no benefit). No renaming beyond the member-name cleanup the hoist forces (`is_disjoint`/`mark_used_fn`/`unmark_fn` reference members are gone; the struct calls the free functions by their real names). Both call sites execute the same logic as before.

### Verification (exactly what was and was not done)

- **Done:** `clang++ -std=c++20 -fsyntax-only` (clang 20.1.8) on the full modified TU with reconstructed include paths (repo `tt_metal/api`, `tt_metal`, `tt_stl`, `tt_metal/hw/inc`, umd submodule at its pinned commit, hal codegen headers generated via `tt_metal/llrt/hal/codegen/codegen.sh`, third-party deps from a CPM cache). **Passes with no errors.** The pristine `origin/main` version of the file passes the identical command, so the check is apples-to-apples.
- **Done:** `git-clang-format --style=file` against `origin/main` — no changes on the modified lines.
- **Done:** pre-change `diff` confirming all three duplicated regions were byte-identical between the two overloads.
- **Not done:** a full `./build_metal.sh --build-tests` build or link (impractical in this environment), and no runtime/hardware testing. CI should be relied on for the full build.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:refactor/dedupe-topology-mapper-packing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:refactor/dedupe-topology-mapper-packing)

### DIFF
--- a/tt_metal/fabric/topology_mapper_utils.cpp
+++ b/tt_metal/fabric/topology_mapper_utils.cpp
@@ -833,6 +833,183 @@ std::map<MeshId, MeshPhysicalLayout> mesh_physical_layouts_from_psd_placements(
     return layouts;
 }
 
+// -----------------------------------------------------------------------------
+// Shared round-robin mesh-packing machinery used by both
+// build_physical_multi_mesh_adjacency_graph overloads (single-MGD and
+// multi-MGD). MeshEnumState, the bitmask helpers, and DisjointPackingSearch
+// below were previously duplicated verbatim inside each overload.
+//
+// MeshEnumState bundles everything needed to drive one shape's solver and
+// remember what it has found so far:
+//   logical_graph   — the pattern we want to place: either the mesh
+//                     instances from the logical descriptor, or the full
+//                     physical coarse topology when only one shape exists
+//                     and all placements need to be assigned.
+//   physical_graph  — the hardware mesh-level connectivity for this shape.
+//   session         — the incremental solver; it keeps prior constraints in
+//                     memory so each call only adds new work.
+//   excluded        — solutions already returned; fed back to the solver so
+//                     it doesn't repeat them.
+//   solutions       — all solutions found so far, in the order returned.
+//   solution_bits   — a chip bitmask for each solution; used for fast
+//                     overlap checks in the packing search below.
+//   embedding_sizes — how many meshes each solution places; used to try
+//                     larger placements first (better pruning).
+//   exhausted       — set to true once the solver finds no more options.
+// -----------------------------------------------------------------------------
+struct MeshEnumState {
+    AdjacencyGraph<MeshId> logical_graph;
+    AdjacencyGraph<MeshId> physical_graph;
+    ::tt::tt_fabric::MappingConstraints<MeshId, MeshId> constraints;
+    ::tt::tt_fabric::TopologyMappingEnumerationSession<MeshId, MeshId> session;
+    std::vector<std::map<MeshId, MeshId>> excluded;
+    std::vector<::tt::tt_fabric::MappingResult<MeshId, MeshId>> solutions;
+    std::vector<std::vector<std::uint64_t>> solution_bits;
+    std::vector<std::size_t> embedding_sizes;
+    bool exhausted = false;
+};
+
+// -----------------------------------------------------------------------------
+// BITMASK HELPERS
+// Each cached solution carries a bitmask with one bit per chip in the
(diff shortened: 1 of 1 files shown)


Riddy21 (approved):
Looks good! Dont' forget to run CPU tests on https://github.com/tenstorrent/tt-metal/actions/workflows/tm-fabric-tests.yaml

p1-0tr (approved):

blozano-tt:
Started the dedicated fabric CPU-only unit tests for commit `78c5b481dd7410c6ab610f30fdac227fbb2b9620`.

[Workflow run](https://github.com/tenstorrent/tt-metal/actions/runs/34541278989) — currently in progress.

Configuration: Release build on Ubuntu 22.04, CPU tests enabled, hardware and performance tests disabled.

PR #54522, opened 2026-08-27
[Bug fix] Pass output_mem_config to ones_like in polar_bw

### LINKED ISSUE #54523: `ones_like` ignores `output_mem_config`, so an intermediate lands in the wrong memory space
Item 17 of #48240.

### Observed

In `_polar_bw` (`complex_unary_backward/device/complex_unary_backward_op.cpp:45-46`), `flip_tensor` is built from two creation calls. The first passes `output_mem_config`, the second does not:

```cpp
ComplexTensor flip_tensor = ComplexTensor(
    {ttnn::zeros_like(input.real(), input.real().dtype(), input.real().layout(), std::nullopt, output_mem_config),
     ttnn::ones_like(input.imag())});   // <- no output_mem_config
```

`ones_like` with no memory config inherits the **reference tensor's** config, i.e. the caller's input, not the requested output config. So whenever the requested config differs from the input's config, the two halves of one `ComplexTensor` are allocated in different memory spaces:

```
input.imag() in DRAM, output_mem_config = L1   ->  real = L1,   imag = DRAM
input.imag() in L1,   output_mem_config = DRAM ->  real = DRAM, imag = L1
```

Every other creation call in this file (lines 45, 68, 81) passes `output_mem_config`. Line 46 is the only one that doesn't.

### Impact

Low, and worth stating plainly so nobody over-prioritises it:

- **Output values are unaffected.** Running `polar_bw` with `inputs=DRAM, memory_config=L1` gives values byte-identical to `inputs=L1, memory_config=L1`.
- **Output placement is unaffected.** The final outputs still honour the requested config, because the ops consuming `flip_tensor` are all passed `output_mem_config` explicitly.
- **The real effect is the intermediate allocation.** The direction that matters is a caller requesting DRAM while passing L1 inputs: `ones_like` then allocates in **L1** even though DRAM was asked for. That is unrequested L1 consumption, which can contribute to L1 pressure in memory-tight models.

### Why existing tests do not catch it

`tests/ttnn/nightly/unit_tests/operations/eltwise/backward/complex_ops/test_backward_polar.py` uses the same `memcfg` for the inputs and for `memory_config=`, so the requested and inherited configs are always equal and can never diverge.

### Repro

```python
t_dram = ttnn.from_torch(x, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,

ttnn.ones_like(t_dram).memory_config().buffer_type                      # BufferType.DRAM
ttnn.zeros_like(t_dram, memory_config=ttnn.L1_MEMORY_CONFIG) \
    .memory_config().buffer_type                                        # BufferType.L1
```

Those two calls are lines 46 and 45 respectively, so this is the mismatch `flip_tensor` ends up with.

### Fix

One line — pass `output_mem_config`, matching the `zeros_like` directly above it:

```cpp
ttnn::ones_like(input.imag(), input.imag().dtype(), input.imag().layout(), std::nullopt, output_mem_config)
```

### Goal / expected result / acceptance criteria

**Goal:** `polar_bw` should allocate every tensor it creates in the memory config the caller asked for, so a caller requesting DRAM never gets an unrequested L1 allocation.

**Expected result:** both halves of `flip_tensor` are created in `output_mem_config`, exactly like the `zeros_like` on the line above.

**Acceptance criteria**

1. `ttnn::ones_like` at `complex_unary_backward_op.cpp:46` is passed `output_mem_config`.
2. With the requested config differing from the input's config, both halves of `flip_tensor` report the same buffer type. Using the repro above, `ones_like(..., memory_config=L1)` and `zeros_like(..., memory_config=L1)` both return `BufferType.L1` for a DRAM input.
3. No change to `polar_bw` output values. Outputs for `inputs=DRAM, memory_config=L1` stay byte-identical to `inputs=L1, memory_config=L1`.
4. No change to `polar_bw` output placement: outputs continue to honour `memory_config`.
5. `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/complex_ops/test_backward_polar.py` still passes.

VirdhatchaniKN:
### Issue

Closes #54523. Item 17 of #48240.

### What this does

Passes `output_mem_config` to the `ones_like` that builds the imaginary half of `flip_tensor` in `_polar_bw`, matching the `zeros_like` above it.

### Why

`ones_like` with no memory config inherits its input's config, not the requested one. When those differ, the two halves of one `ComplexTensor` land in different memory spaces — so a caller asking for DRAM gets an unrequested L1 allocation. Output values and placement were never affected, so this is a placement fix, not a correctness fix. Detail in #54523.

### Tests

| Test | Result |
| --- | --- |
| `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/complex_ops/test_backward_polar.py` | 12 passed |
| `tests/ttnn/docs_examples/test_backward_examples.py -k polar_bw` | 1 passed |
| `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/complex_ops/` (whole dir) | 188 passed, 16 skipped |

No existing test covers the mixed-config case, so these show no regression. The fix itself was checked with a probe: both halves now report the same buffer type, and values are unchanged.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/ones_bug)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/ones_bug)


ttnn/cpp/ttnn/operations/eltwise/complex_unary_backward/device/complex_unary_backward_op.cpp
```diff
@@ -43,7 +43,7 @@ std::vector<ComplexTensor> _polar_bw(
     sgn_result.deallocate();
     ComplexTensor flip_tensor = ComplexTensor(
         {ttnn::zeros_like(input.real(), input.real().dtype(), input.real().layout(), std::nullopt, output_mem_config),
-         ttnn::ones_like(input.imag())});
+         ttnn::ones_like(input.imag(), input.imag().dtype(), input.imag().layout(), std::nullopt, output_mem_config)});
```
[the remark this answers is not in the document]
VirdhatchaniKN: Added test and it passed
Screenshot 2026-08-27 at 12 48 24 PM
Removing the test now.

Aswinmcw, line 46: Fix is right and this file is now consistent (45/46 match, 68 and 81 already passed it). Since #54523 is item 17 of #48240, a grep for the same pattern turns up several siblings worth folding into the series — not asking for them here:

- `unary_backward.cpp:539` (`fill_bw`) — line 535 computes `auto output_memory_config = output_mem_config.value_or(input.memory_config());` and then never uses it; the `zeros_like(grad)` on 539 drops it. Dead local, so the compiler won't warn.
- `unary_backward.cpp:258` (`pow_bw`, `|exponent| < eps` path) — `input_grad = ttnn::zeros_like(input);` replaces the optional rather than filling it, so in addition to ignoring `output_mem_config` it leaves a caller-supplied preallocated `input_grad` unwritten. `fill_bw` does this correctly by threading the tensor in as the last `zeros_like` arg.
- `binary_backward.cpp:38,41` — `detail::preallocated_tensors_check` takes no memory-config parameter at all, so every binary backward op that allocates its grad outputs through it inherits the input's config. That one needs the parameter threaded through, not a call-site fix.
- `binary/device/binary_composite_op.cpp:677,793` (`polyval`, scalar-base `pow`) with verbatim copies at `experimental/quasar/binary/device/binary_composite_op.cpp:605,721` — any fix there needs applying to both.

The `empty_like` cases are the ones I'd rank above this one: the tensor created is the op's actual output, so the whole result lands in the wrong memory space rather than just an intermediate.

Aswinmcw (approved):

ttnn/cpp/ttnn/operations/eltwise/complex_unary_backward/device/complex_unary_backward_op.cpp
VirdhatchaniKN: Checked all four — they hold, with a couple of extra details:

- **`fill_bw`** (`unary_backward.cpp:535`) — confirmed dead: `output_memory_config` appears exactly once in the function, the declaration. Never read. Both `zeros_like` calls on 539 pass `std::nullopt` for memory config.
- **`pow_bw`** (`:258`) — confirmed, and it is the worse of the two failure modes: `input_grad = ttnn::zeros_like(input);` rebinds the local `std::optional` instead of writing through the caller's tensor, so a preallocated `input_grad` comes back untouched. Worth noting line 254 right above it (`input_grad.value_or(ttnn::empty_like(input))`) is itself one of the `empty_like` cases.
- **`preallocated_tensors_check`** (`binary_backward.cpp:38,41`) — confirmed, no memory-config parameter in the signature. Agreed this needs the parameter threaded through rather than a call-site fix.
- **`polyval` / scalar-base `pow`** — confirmed, and the two differ in severity: `binary_composite_op.cpp:677` is `return ttnn::full_like(...)`, i.e. the op's actual output, whereas `:793` is only an intermediate like this PR's. The quasar copies at `605`/`721` are verbatim, so any fix needs applying in both places.

Agreed on the ranking, and for the reason you give: where the created tensor is the returned output, the whole result lands in the wrong memory space, not just a temporary. `preallocated_tensors_check` and `:677` are the two that fit that description.

Leaving all of these out of this PR as you suggest — happy to file them against #48240 as follow-ups if useful.

VirdhatchaniKN: Tracked here https://github.com/tenstorrent/tt-metal/issues/54546

mateusznowakTT (approved):

PR #54525, opened 2026-08-27
[Bug fix] Take the plain RMSNorm op when the cluster axis is length 1

### LINKED ISSUE #54518: [model]: distributed RMSNorm aborts on a single-column mesh (TP=1)
### Component / Area

model — `models/demos/deepseek_v3_d_p` (distributed RMSNorm)

### Issue Type

Runtime Crash

**Found during:** Mistral Small 4 119B prefill bring-up (#53688), whose PP=4 layout is the first to use a single-column mesh. The bug is in shared DeepSeek-family prefill code — any single-column mesh hits it, not just Mistral.

## Observed

`TtDistributedRmsNorm` aborts on any mesh whose cluster axis has length 1. Neither stats gather degenerates to a copy on a singleton axis — both `TT_FATAL` — so which error you get depends only on whether the tensor's topology metadata supports the high-bandwidth path:

```
all_gather_device_operation.cpp:41
    all_gather collective will only work for num_devices > 1, got 1

high_bw_all_gather_device_operation.cpp:330
    high_bw_all_gather cluster_axis 1 selects a singleton mesh axis in mesh shape MeshShape([8, 1])
```

TP=1 is a configuration this package otherwise supports — `tt_reduce.py:131`, `tt_moe_gate_prefill.py:814` and `tt_ccl.py:366` all guard a length-1 axis already; the norm was simply inconsistent with its siblings. It is reached by any single-column mesh, including every `(8,1)` stage submesh of a PP=4 × TP=1 × SP=8 layout.

Nothing exercises TP=1 today, which is why the inconsistency survived: `test_rmsnorm_distributed`'s existing `(1,4)` row is not feasible on Blackhole galaxy and skips there.

## Expected

At a length-1 cluster axis there is nothing to distribute — every device already holds the full hidden dim — so the norm computes normally rather than aborting.

## Steps to Reproduce

```
pytest models/demos/deepseek_v3_d_p/tests/pcc/test_rmsnorm.py::test_rmsnorm_tp1
```

(the regression test added with the fix; carves an `(8,1)` submesh out of the `8x4` mesh)

**Input data:** none — random weights and input, no checkpoint.
**Frequency:** Always, on any single-column mesh.

## Software Versions

- tt-metal main `58342a6f63d` (2026-08-26) — bug present at this commit, re-confirmed by reverting the guard and re-running the regression test
- Ubuntu 22.04.5 LTS

## Hardware Details

- Product: Blackhole

## Regression

Not a regression — present since the module was written. Reached only once a single-column mesh was used.

## Priority

P1

## Impact

Blocks any single-column mesh, including all four stages of the PP=4 × TP=1 × SP=8 Mistral Small 4 prefill layout.

kmabeeTT:
### PR Category
Bug fix

### Summary

Fixes #54518.

At a length-1 cluster axis every device already holds the full hidden dim, so nothing is distributed: return `ttnn.rms_norm` before choosing a gather at all.

The cached gamma is reused as-is — plain `layer_norm` accepts the same ROW_MAJOR `[1, 1, dim/32, 32]` layout the distributed op takes — so there is no second weight, no second cache entry, and no existing line changes.

Split out of #53747 per review, and reworked there per @mbezuljTT's suggestion: the version in that PR kept the distributed op pair and fed it un-gathered local stats, which forced a restructure of the surrounding code. Taking the plain op instead makes the fix a 12-line insertion and lets steps 2–3 stay exactly as they are.

### Notes for reviewers

- Measured on an `(8,1)` submesh of an `8x4` Blackhole galaxy — the real PP=4 stage shape:
  - output is **bit-identical** to the pre/post pair (0 of 21M elements differ, max abs diff 0), so this cannot change any existing result;
  - **2.7×** faster device-side at the production row count (182.8 → 67.1 µs traced), 11× eager;
  - full PP=4 256k run unchanged (17.9k vs 17.8k tok/s steady), and the TP=2 shape re-checked to confirm the untouched distributed path still works.
- `test_rmsnorm_tp1` runs on an `(8,1)` submesh rather than a bare `(1,1)` mesh, which is not feasible on this hardware and would only ever skip. Verified it discriminates: with the fix reverted it fails with the `high_bw_all_gather` singleton-axis `TT_FATAL`.
- The underlying CCL behaviour — neither gather degenerating to a copy at one device — is worth fixing centrally; every model hand-rolls this guard today. Out of scope here.

### Testing

- `pcc/test_rmsnorm.py::test_rmsnorm_tp1` — passes on an `(8,1)` submesh of the `8x4` galaxy, and **fails with the fix reverted** (`high_bw_all_gather ... singleton mesh axis`), so it discriminates
- Output is **bit-identical** to the pre/post pair at TP=1 — 0 of 21M elements differ, max abs diff 0 — so no existing result can change
- **2.7x faster** device-side at the production row count (182.8 → 67.1 us traced), 11x eager
- Mistral Small 4 PP=4 x TP=1 x SP=8, all 36 layers, real weights, 256k chunked prefill (51 x 5120) — passed, 17841 tok/s steady (unchanged vs the pre-rework 17803). Throughput/no-crash only: the run is headless (no LM head or sampling) and asserts no PCC, so correctness rests on the two bullets above
- TP=2 shape re-checked, confirming the untouched distributed path still works
- Will launch CI after the initial round of Copilot feedback (if any) is addressed, to avoid a rerun before merge:
  ```
  gh workflow run "Blaze Models Prefill tests" --repo tenstorrent/tt-metal --ref BRANCH -f test-type=prefill_block,prefill_transformer_determinism
  ```
- `kmabee/issue_54515-54519_integration` carries all five fixes in order, for one full `test-type=all` + `(Blackhole) e2e` run linked from all five PRs.

### DIFF
--- a/models/demos/deepseek_v3_d_p/tests/pcc/test_rmsnorm.py
+++ b/models/demos/deepseek_v3_d_p/tests/pcc/test_rmsnorm.py
@@ -18,7 +18,7 @@ from loguru import logger
 from tracy import signpost
 
 import ttnn
-from models.demos.deepseek_v3_d_p.tests.fabric_profiles import torus_x_device_params
+from models.demos.deepseek_v3_d_p.tests.fabric_profiles import torus_x_device_params, torus_xy_device_params
 from models.demos.deepseek_v3_d_p.tt.tt_ccl import per_axis_topology
 from models.demos.deepseek_v3_d_p.tt.tt_distributed_rms_norm import TtDistributedRmsNorm
 from tests.ttnn.utils_for_testing import assert_with_pcc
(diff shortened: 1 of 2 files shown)


kmabeeTT:
onPR jobs aren't launching for some reason on this PR (tried open/close, force push), going to close and try re-filing.  (Edit: Ah, conflict prevents jobs from running)

PR #54526, opened 2026-08-27
[Bug fix] Reject a tuned MLA matmul config whose in0_block_w does not divide K

### LINKED ISSUE #54517: [model]: tuned MLA matmul config applied to a model with a different K
### Component / Area

model — `models/demos/deepseek_v3_d_p` (MLA tuned matmul configs)

### Issue Type

Runtime Crash

**Found during:** Mistral Small 4 119B prefill bring-up (#53688). Mistral Small 4 is the variant that hits the crash today, but the cause — a tuned table keyed without regard to the variant's dimensions — is shared DeepSeek-family prefill code.

## Observed

`MLA_MATMUL_CONFIG` is keyed on `(weight_name, seq_len_local)` and carries nothing about the variant's dimensions, so a config tuned for one model is applied to another with a different K. Where the tuned `in0_block_w` does not divide that model's `Kt`, the matmul dies:

```
TT_FATAL: MatmulMultiCoreReuseMultiCastProgramConfig: Kt (32) must be divisible by in0_block_w (14)
```

Mistral Small 4 at `seq_len_local` 3200 (25600/8, the 25k production ISL) hits exactly this: it picks up a row tuned for DeepSeek, whose `Kt` is 56 (hidden 7168 / TP 4 / 32) against Mistral's 32.

The crash is the *lucky* case. Where the block width happens to divide — e.g. `Kt` 64 with `in0_block_w` 8 — another model's tiling is applied silently and nothing reports it.

## Expected

A tuned config that cannot apply to this weight is not used; the matmul falls back to the default program config (untuned, so possibly slower) rather than aborting.

## Steps to Reproduce

1. Run any MLA variant whose `Kt` differs from the tuned variant's at a `seq_len_local` present in `MLA_MATMUL_CONFIG` — e.g. Mistral Small 4, single-rank, 25k ISL.
2. The `q_a_proj` matmul aborts with the `TT_FATAL` above.

**Input data:** any MLA variant whose hidden/TP combination gives a `Kt` not divisible by the tuned `in0_block_w`.
**Frequency:** Always, for an affected (variant, seq_len_local) pair.

## Software Versions

- tt-metal main `58342a6f63d` (2026-08-26) — bug present at this commit
- Ubuntu 22.04.5 LTS

## Hardware Details

- Product: Blackhole

## Regression

Not a regression — present since the tuned table was introduced. It only became reachable when a second variant with a different `Kt` started using those `seq_len_local` values.

## Priority

P1

## Impact

Blocks any MLA model whose K differs from the tuned one, including Mistral Small 4 at the 25k production ISL. The silent-misapplication case affects correctness-adjacent tuning on models nobody has checked.

kmabeeTT:
### PR Category
Bug fix

### Summary

Fixes #54517.

`MLA_MATMUL_CONFIG` is keyed on `(weight_name, seq_len_local)` and carries nothing about a variant's dimensions, so a config tuned for one model could be applied to another with a different K — fatal where the tuned `in0_block_w` does not divide that model's `Kt`:

    TT_FATAL: MatmulMultiCoreReuseMultiCastProgramConfig: Kt (32) must be
              divisible by in0_block_w (14)

Mistral Small 4 at `seq_len_local` 3200 (the 25k production ISL) hits exactly this, picking up a row tuned for DeepSeek's `Kt` of 56.

K is now a **gating tag in `_cfg_matches`**, beside `num_heads` / `q_lora_rank` / `chunked_only` / `dense_head_cap_non_dsa`: a candidate whose `in0_block_w` cannot divide this model's per-device `Kt` is skipped, and the next candidate in the slot is considered.

Split out of #53747 per review.

### Notes for reviewers

- Skipping a candidate rather than vetoing the slot is the point. A slot may hold a list, so two variants can each tune the same `(weight_name, seq_len_local)` and each will select its own entry — DeepSeek's `in0_block_w=14` at `Kt=56`, Mistral's `in0_block_w=8` at `Kt=32`. Rejecting the whole slot on a mismatch would make the second entry unreachable.
- `Kt` is the per-device extent: weights are `[K, N]` and `mapper_tp0` shards K across the TP axis, so `shape[-2]` is already post-shard. Confirmed on silicon — the check computed `Kt=32` for Mistral `q_a_proj` at tp=4 (4096 / 4 / 32), where the global value would have been 128.
- Existing entries are untagged and accepted for any K their tiling divides, which is what keeps DeepSeek and Kimi selecting them today. Declaring the `kt` each entry was tuned at would make matching exact, but the table does not record it and deriving 26 values by hand risks the current tunings — better done as entries are touched.
- Rebased onto `58342a6f63d`; no conflict with the GLM-5.2 indexer changes to `mla.py` from #53637.

### Testing

Local, on a 32-chip Blackhole galaxy:

- Mistral Small 4 MLA at 25k (`test_mistral4_mla -k "seq25k and check_pcc"`) — the config that reaches this check. Passed on every mesh profile this galaxy supports, with the DeepSeek-tuned entry correctly not selected at Mistral's `Kt=32`.
- Blaze on this branch, `test-type=mla,mla_chunked,kimi_mla,k3_mla,dsa_mla_glm`: [run 33403558186](https://github.com/tenstorrent/tt-metal/actions/runs/33403558186).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kmabee/issue_54517_fix-mla-matmul-cfg-guard)


models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
```diff
@@ -0,0 +1,50 @@
...
+    """Guards `perf(mla): guard tuned matmul configs against the wrong K`.
+
+    `MLA_MATMUL_CONFIG` is keyed on (weight_name, seq_len_local) and knows nothing about the
+    variant's dimensions, so a config tuned for one model is applied to another with a different K.
+    Rejecting a non-dividing block width degrades to the generic program config -- untuned, so
+    possibly slower -- instead of dying.
+    """
+    stub = SimpleNamespace(q_a_proj_weight=SimpleNamespace(shape=(kt * ttnn.TILE_SIZE, 1024)))
+    cfg = {"program_config": SimpleNamespace(in0_block_w=in0_block_w)}
+    assert ttMLA._cfg_fits_weight(stub, cfg, "q_a_proj") is fits
```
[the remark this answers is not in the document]
kmabeeTT: Fixed in d9831643f1a — both assertions now name the dimensions under test.

models/demos/deepseek_v3_d_p/tt/mla/mla.py
```diff
@@ -717,7 +717,43 @@ def _resolve_mm_cfg(self, weight_name: str, seq_len_local: int) -> dict | None:
...
+            return True  # nothing to check against; keep the tuned config
+        # Weights are [K, N]; only the K dim participates in the in0 block tiling.
+        kt = weight.shape[-2] // ttnn.TILE_SIZE
+        if kt % in0_block_w == 0:
+            return True
+        logger.warning(
+            f"[ttMLA] tuned matmul config for '{weight_name}' has in0_block_w={in0_block_w}, which does "
+            f"not divide this model's Kt={kt} — falling back to the default program config. The tuned "
+            f"table is keyed on (weight, seq_len) only, so it is another variant's tuning."
+        )
```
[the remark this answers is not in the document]
kmabeeTT: Fixed in d9831643f1a — warns once per distinct `(weight_name, in0_block_w, Kt)` instead of once per resolution.

kmabeeTT:

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
```diff
@@ -0,0 +1,56 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+"""Host-only regression tests for the MLA tuned-matmul config guard. No device, no weights.
+
+Fails on the code as it stood before `perf(mla): guard tuned matmul configs against the wrong K`.
```
ipotkonjak-tt: is there a real need for this file?

ipotkonjak-tt (commented):
is this the right solution? Seems it's more practical to add K to config selection. You'll need to do the above if you're to tune Mistral's matmuls, this way they'd collide with DeepSeek

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
```diff
@@ -0,0 +1,56 @@
...
+        # so the matmul TT_FATALs: "Kt (32) must be divisible by in0_block_w (14)".
+        (14, 32, False),
+        (8, 32, True),
+        # Kt = 64 (tp 2) divides by 8, so the config is ACCEPTED -- the guard's known gap: another
+        # model's tiling then applies silently. Pinned here so a future change to that behaviour is
+        # a deliberate decision rather than a surprise.
+        (8, 64, True),
+    ],
+)
+def test_tuned_matmul_cfg_is_rejected_when_block_width_does_not_divide_k(in0_block_w, kt, fits):
```
nbabinTT: Please rename this function to something more meaningful. Do we actually need this test if it's not part of any CI?

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
```diff
@@ -0,0 +1,56 @@
...
+    possibly slower -- instead of dying.
+    """
+    stub = SimpleNamespace(q_a_proj_weight=SimpleNamespace(shape=(kt * ttnn.TILE_SIZE, 1024)))
+    cfg = {"program_config": SimpleNamespace(in0_block_w=in0_block_w)}
+    assert (
+        ttMLA._cfg_fits_weight(stub, cfg, "q_a_proj") is fits
+    ), f"in0_block_w={in0_block_w} against Kt={kt}: expected fits={fits}"
+
+
+def test_tuned_matmul_cfg_kept_when_there_is_nothing_to_check_against():
```
nbabinTT: Please rename this function to something more meaningful. Do we actually need this test if it's not part of any CI?

nbabinTT:
Thank you for running CI. Adding links to Blaze CI for visibility: https://github.com/tenstorrent/tt-metal/actions/workflows/blaze-models-prefill-tests.yaml?query=branch%3Akmabee%2Fissue_54517_fix-mla-matmul-cfg-guard

models/demos/deepseek_v3_d_p/tt/mla/mla.py
```diff
@@ -717,7 +717,43 @@ def _resolve_mm_cfg(self, weight_name: str, seq_len_local: int) -> dict | None:
...
+        divide, another model's tuning is applied silently and nobody finds out. The real fix is to
+        key the table on the variant or on the actual (K, N); this is the guard until then.
+        """
+        pc = cfg.get("program_config")
+        in0_block_w = getattr(pc, "in0_block_w", None)
+        weight = getattr(self, f"{weight_name}_weight", None)
+        if in0_block_w is None or weight is None:
+            return True  # nothing to check against; keep the tuned config
+        # Weights are [K, N]; only the K dim participates in the in0 block tiling.
+        kt = weight.shape[-2] // ttnn.TILE_SIZE
```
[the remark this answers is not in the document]
kmabeeTT: Per-device, confirmed on silicon: Mistral at tp=4 computed Kt=32 for q_a_proj (4096 / 4 / 32) where the global value would be 128; wkv_b1 Kt=2 and wkv_b2 Kt=8 agree. mapper_tp0 shards K, so .shape[-2] is already post-shard — the extent the matmul program actually consumes.

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
kmabeeTT: Thx for raising - no...  deleted in 794e74cc811 after also considering feedback from Nenad below. It existed to pin the guard's fit decision; now that K is just another selection tag there's nothing separate to test.

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
kmabeeTT: Fair points... deleted the file at 794e74cc811 — the guard it tested was replaced by ordinary tag selection. On the CI half: you're right it wasn't gated, and nothing under `tests/torch/` is, including the five files that predate this PR. Worth fixing separately — the mechanism exists (`cpu_medium` + `fabric_cpu_only_unit_tests.yaml`), but `time_budget.yaml` only grants `cpu_medium` to `scaleout` and these are `team: models`.

models/demos/deepseek_v3_d_p/tests/torch/test_mla_matmul_configs.py
kmabeeTT: Same as above — file deleted in 794e74cc811.

nbabinTT (approved):

ipotkonjak-tt (commented):
approved assuming [Blaze Models Prefill tests](https://github.com/tenstorrent/tt-metal/actions/workflows/blaze-models-prefill-tests.yaml) passes

ipotkonjak-tt (approved):
approved assuming [Blaze Models Prefill tests](https://github.com/tenstorrent/tt-metal/actions/workflows/blaze-models-prefill-tests.yaml) passes

kmabeeTT:
> approved assuming [Blaze Models Prefill tests](https://github.com/tenstorrent/tt-metal/actions/workflows/blaze-models-prefill-tests.yaml) passes

Thanks! Yep, link in PR desc for this PR as well as it combined with the 4 other bug fixes (also passing).

PR #54529, opened 2026-08-27
[Bug fix] Compare the KVPE the device actually caches

### LINKED ISSUE #54519: [model]: per-layer KVPE comparison never runs — every row reports -1.0
### Component / Area

model — `models/demos/deepseek_v3_d_p` (prefill reference path)

### Issue Type

Bad Outputs

**Found during:** Mistral Small 4 119B prefill bring-up (#53688). The bug is in shared DeepSeek-family prefill code — it affects every variant whose reference caches expanded per-head keys, not just Mistral.

## Observed

The per-layer KVPE comparison never happens, and nothing says so. The reference caches expanded per-head keys `[b, n_heads, seq, head_dim]` while the device caches the compressed MLA latent `[b, 1, seq, kv_lora_rank + qk_rope_head_dim]`. The two do not correspond in rank **or** width, so every per-layer KVPE row reports `-1.0` — the sentinel for "not measured".

It fails quietly rather than loudly: a `[..., :kv_lora_rank]` slice of a 128-wide last dim just clamps to 128, so what reaches `comp_pcc` is a shape error rather than a number.

A related failure is in how the reference layer is called. The cache kwarg is `past_key_value` on the vendored DeepSeek/Kimi/GLM layers and `past_key_values` on transformers ≥ 5; the wrong name lands silently in `**kwargs`, so no KV is captured at all.

## Expected

Per-layer KVPE rows report a real PCC against the tensor the device actually caches, and a reference cache written with the wrong KVPE layout is rejected rather than reused.

## Steps to Reproduce

1. `pytest models/demos/deepseek_v3_d_p/tests/test_prefill_block.py` for any MLA variant whose reference is a stock transformers attention (e.g. `Mistral4Attention`).
2. Every per-layer KVPE row in the output reads `-1.0`.

**Input data:** any MLA variant and its HF checkpoint.
**Frequency:** Always, on every affected variant.

## Software Versions

- tt-metal main `58342a6f63d` (2026-08-26) — bug present at this commit
- Ubuntu 22.04.5 LTS

## Hardware Details

- Product: Blackhole

## Regression

Dates from the device switching to a compressed-latent KV cache while the reference kept expanding per-head keys.

## Priority

P2

## Impact

KVPE has never actually been validated on any variant whose reference caches expanded keys. A whole class of KV-cache bug is currently invisible to the test suite.

## Note on ordering

Depends on #54516: the derived KVPE rotates its pe part with the reference rope table, so on a bf16 rope the derived value carries exactly the position-dependent phase error that issue describes. Land the rope fix first.

kmabeeTT:
### PR Category
Bug fix

### Summary

Fixes #54519.

Derive the KVPE line the device actually stores — `kv_a_layernorm(latent)` concatenated with the rope-rotated pe part, computed from the layer's own weights — whenever the reference caches the expanded per-head form instead. Qualify the reference-cache identity by KVPE width so a cache written before this fix is rejected rather than reused.

`decoder_layer_kwargs` binds the reference layer call to that layer's **own signature**, which fixes the silently-wrong cache kwarg across transformers generations and supplies `position_embeddings` where the layer requires it.

Split out of #53747 per review.

### ⚠️ Merge order

Built on top of #54528 and should land after it. `derive_mla_kvpe` rotates the pe part with the reference rope table; on the pre-fix bf16 rope the derived KVPE would carry the position-dependent phase error that PR removes, so merging this first would measure the wrong thing. 

### Notes for reviewers

- The derivation runs in fp32: accumulating the norm + projection in bf16 costs ~2e-3 of PCC against the device's own KVPE, enough to fail a 0.999 bar with nothing actually wrong.
- `derive_mla_kvpe` must be called while the layer's weights are still loaded — the layer-by-layer reference frees them right after the forward.
- `tests/torch/test_prefill_layer_kwargs.py` is host-only and pins both signature shapes (singular/plural cache kwarg, with/without `position_embeddings`).
- Verified on silicon (32-chip Blackhole galaxy, `test_prefill_block.py -k mistral4`, 6 passed). The derivation path is demonstrably live, and both halves report a real number rather than the pre-fix `-1.0` sentinel:

  ```
  Reference caches expanded KV (last dim 128 != 320); deriving the compressed MLA KVPE line from each layer instead
  KVPE cache KV part PCC: 0.999923 (threshold: 0.999)
  KVPE cache PE part PCC: 0.999952 (threshold: 0.999)
  ```

### Testing

- `tests/torch/test_prefill_layer_kwargs.py` — host-only, 3 passed; pins both signature shapes (singular/plural cache kwarg, with/without `position_embeddings`)
- Mistral Small 4 prefill block, 8x4 — 6 passed. The derivation path is demonstrably live and both halves report a real number rather than the pre-fix `-1.0`:
  ```
  Reference caches expanded KV (last dim 128 != 320); deriving the compressed MLA KVPE line from each layer instead
  KVPE cache KV part PCC: 0.999923 (threshold: 0.999)
  KVPE cache PE part PCC: 0.999952 (threshold: 0.999)
  ```
- Blaze dispatched on this branch, `test-type=prefill_block,glm_prefill_block,kimi_prefill_block,kv_cache_table,prefill_runner_kv_pcc`: [run 33043229940](https://github.com/tenstorrent/tt-metal/actions/runs/33043229940) (passing)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kmabee/issue_54519_fix-prefill-kvpe)


models/demos/deepseek_v3_d_p/tests/test_prefill_block.py
```diff
@@ -285,18 +287,18 @@ def run_model(
...
-                )
-                torch_output = layer_out[0]
+                layer_out = hf_model.layers[layer_idx](torch_input, **layer_kwargs)
+                torch_output = layer_out[0] if isinstance(layer_out, (tuple, list)) else layer_out
             logger.info(f"Torch reference output shape: {torch_output.shape}")
             if ref_cache is not None:
-                ref_kvpe = hf_cache_layer_kv(ref_cache, layer_idx)[0]
+                ref_kvpe = reference_kvpe_for_layer(
+                    hf_model.layers[layer_idx], layer_idx, torch_input, layer_kwargs, ref_cache, config
+                )
```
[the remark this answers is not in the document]
kmabeeTT: Correct — the derivation only ran on a cache miss. Fixed in e87d72cb958: the loaded `ref_kvpe` width is checked against `mla_kvpe_width(config)` and a stale cache is recomputed instead of reused.

kmabeeTT:

models/demos/deepseek_v3_d_p/tests/torch/test_prefill_layer_kwargs.py
```diff
@@ -0,0 +1,72 @@
...
+
+
+def test_old_style_layer_kwargs_match_the_explicit_call_they_replaced():
+    """`decoder_layer_kwargs` now binds the reference call in test_prefill_block for EVERY model, so
+    on a vendored layer it must still reproduce the explicit call it replaced -- a silent drift here
+    moves the reference those models' PCC rows are judged against.
+    """
+    cache = object()
+    mask = torch.zeros(1, 1, 4, 4)
+    pos = torch.arange(4).unsqueeze(0)
```
pavlepopovic, line 40: do we run this in ci anywhere? is there a point in these tests if we don't ?

pavlepopovic (approved):

models/demos/deepseek_v3_d_p/tests/torch/test_prefill_layer_kwargs.py
kmabeeTT: Nope it is not - `tests/torch/` appears in zero pipeline or workflow files, and neither are the four files already in that directory. That's about half the test files under `models/`, there is some benefit to keeping this one though:

`decoder_layer_kwargs` fails silently, not loudly — bind the cache under the wrong name and it lands in `**kwargs`, no KV is captured, and you get a confusing PCC number in a device test instead of an error. The plural (`past_key_values`) branch has no other coverage, and which branch fires depends on the installed transformers version, which no device test pins. 72 lines, no device, milliseconds.

Gating it in CI is blocked on budget rather than pipeline: `time_budget.yaml` grants `cpu_medium` only to `scaleout`, and `models` has no CPU grant at all. The cheap route is to hang a host-only file off a leg that already boots a device — it needs neither, so it rides for milliseconds and no new budget line.

PR #53700, opened 2026-08-19
#53613: Eltwise Unary bfloat16 coverage tests - category 1 ops

KalaivaniMCW:
### PR Category

Test Only

### Summary

#53613 

### Notes for reviewers

Eltwise unit tests restructuring
- Add `test_unary_category1_bfloat16.py` with exhaustive bfloat16 bitpattern coverage for 47 category-1 unary ops (trig, hyperbolic, comparison, rounding, special math, logical, and utility)
- Add shared test utilities in `eltwise_test_utils.py` (`generate_bfloat16_bits`, `generate_bfloat16_bits_in_range`, `flush_to_zero`)
- Each op is tested against its golden function over every representable bfloat16 value in its valid domain, with appropriate tolerance (exact match, ULP=1, or PCC≥0.999 depending on kernel accuracy)
- Remove redundant tests from existing files that are now subsumed by the exhaustive coverage

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kalaivani/unary_cat1_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kalaivani/unary_cat1_bf16)

PR #54533, opened 2026-08-27
#53613: Eltwise Unary bfloat16 coverage tests - category 5 ops

### LINKED ISSUE #55338: #53613: Category 5 — Exhaustive BF16 coverage for GLU-family ops (glu, reglu, geglu, swiglu)
## Parent Issue
Closes / contributes to #53613 (Eltwise unit tests restructuring — standardised Eltwise tests)

## PR
Implemented in #54533

## Summary

Adds exhaustive bfloat16 bit-pattern coverage for all four **Category-5 gated linear unit ops** that share the `split + gate + multiply` infra pattern:

| Op | Gate function | Accuracy criterion |
|---|---|---|
| `ttnn.glu` | sigmoid(B) | ULP ≤ 2 |
| `ttnn.reglu` | relu(B) | ULP ≤ 1 |
| `ttnn.geglu` | gelu(B) — tanh approx | PCC ≥ 0.999 |
| `ttnn.swiglu` | silu(B) = B·sigmoid(B) | PCC ≥ 0.999 |

All ops split the input at `dim` and compute `output = A * gate_fn(B)`.

## What was added

- **`test_unary_category5_bfloat16.py`** — exhaustive normal bf16 sweep for all four ops.
  - `glu` / `reglu`: all 32 512 positive + 32 512 negative normal bf16 values through the gate half; ULP asserted.
  - `geglu` / `swiglu`: same coverage; PCC ≥ 0.999; value half A clamped to `[-1, 1]` to prevent `A * gate(B)` overflow (which `comp_pcc` would silently discard).
  - FTZ mask derived from `golden` only — `result` is not independently flushed.
  - All tests parametrized over `dim ∈ {-1, 3}` to cover positive-index normalization in `split_tensor_for_glu`.

- Redundant tests removed from `test_activation.py` and `test_composite.py`.

## Accuracy notes

| Op | Observed max ULP | Notes |
|---|---|---|
| glu | ≤ 2 | sigmoid SFPU ≤ 2 ULP |
| reglu | ≤ 1 | relu is exact piecewise-linear |
| geglu | ~62 571 | FAST_LUT GELU approximation — use `GeluVariant::ACCURATE` to improve |
| swiglu | ~128 | SFPU silu approximation; FTZ for B < −88 |

## Labels
op_cat: eltwise

abdulla-TT:
### PR Category
Test Only

Ticket
#53613

### Summary
- Add `test_unary_category5_bfloat16.py` with exhaustive bfloat16 bit-pattern coverage for all 4 Category-5 gated linear unit ops (`ttnn.glu`, `ttnn.reglu`, `ttnn.geglu`, `ttnn.swiglu`)
- Each op is tested against its golden function over every representable bfloat16 value in its valid domain, with appropriate tolerance:
  - `glu`    → ULP ≤ 2  (sigmoid via SFPU)
  - `reglu`  → ULP ≤ 1  (relu is exact piecewise linear)
  - `geglu`  → PCC ≥ 0.999  (GELU uses a tanh polynomial approximation)
  - `swiglu` → PCC ≥ 0.999  (SFPU flushes sigmoid subnormals to zero for large negative gate inputs)
- For `geglu` and `swiglu`, the value half A uses the transpose of B so that A and B are distinct at every off-diagonal position — operand-wiring bugs are detectable and saturated values do not dominate the PCC
- Remove redundant tests from `test_activation.py` and `test_composite.py` that are now subsumed by the exhaustive coverage

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/unary_cat5_bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/unary_cat5_bf16)


tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
```diff
@@ -0,0 +1,138 @@
...
+def test_swiglu_op(device):
+    """Exhaustive normal bfloat16 coverage for swiglu.
+
+    PCC ≥ 0.999 is used because SFPU flushes sigmoid subnormals to zero for
+    large negative gate inputs, causing small legitimate differences from the
+    CPU reference.  A = B.T so both halves cover all normal values with
+    distinct pairings, making operand-wiring bugs detectable.
+    """
+    B = generate_bfloat16_bits(include_spl_values=False)  # (256, 256) — all normal bf16 values
+    A = B.T.contiguous()  # same values, transposed ordering
```
Aswinmcw: With `A = B.T`, both halves span the full bf16 magnitude range, so `A * gate(B)` overflows: I get **4160 of the 65536 elements (6.3%) landing on ±inf** in the golden for both swiglu and geglu (another ~36% are exact zero). `comp_pcc` rewrites inf/NaN to `0.0` in *both* tensors before correlating (`models/common/utility_functions.py:527-529`), so those 4160 points are unconditionally discarded — the device can return anything there and the test still passes. The glu/reglu docstring above already calls out overflow as the reason `A` is pinned to 1.0; the same hazard is present here, just hidden.

Bounding `A` (e.g. scale into `[-1, 1]`, or restrict it to a limited exponent window) keeps every product finite, which makes the comparison actually cover all 65536 points — and once the outputs are finite, `assert_with_ulp` with a measured threshold becomes usable here too, matching glu/reglu. Same applies to `test_geglu_op` (line 128).

The ULP switch is a suggestion, not a blocker; the inf masking is the part worth fixing.

tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    B = generate_bfloat16_bits(include_spl_values=False)  # (256, 256)
+    A = torch.ones_like(B)  # A=1.0: output = gate_fn(B) directly
+    input_tensor = _build_glu_input(A, B)
+
+    tt_in = to_tt_tensor(input_tensor, device)
+    golden_function = ttnn.get_golden_function(ttnn_op)
+    golden = golden_function(input_tensor, dim=-1, device=device)
+
+    tt_result = ttnn_op(tt_in, dim=-1)
```
Aswinmcw: All four new tests hardcode `dim=-1`, but both deleted suites parametrized `dim` over `[-1, 3]`. Nothing else under `tests/ttnn` exercises the GLU family with a positive `dim` after this PR (`test_situ_glu_sfpu.py` is a BH-only low-level kernel test, and the `python_api_testing` yamls aren't run here), so the positive-index normalization path loses its only coverage. Worth a `@pytest.mark.parametrize("dim", [-1, 3])` on these.

tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
```diff
@@ -0,0 +1,138 @@
...
+    result = ttnn.to_torch(tt_result)
+
+    result = flush_to_zero(result)
```
Aswinmcw: Separate from the threshold question already raised below: this masks the **outputs**, on both sides. Any element where the device returns a near-zero for a legitimately non-zero expected value gets rewritten to `0.0` on both sides and passes silently — the mask removes exactly the failure it should catch.

If the intent is to model SFPU FTZ, derive the mask from the *expected* side only (`golden`), or better, filter the **inputs** so the sweep stays inside the range where device and CPU agree, and give the FTZ boundary its own explicit case. This is the same shape as the `atanh` and near-zero-mask findings on #53700.

Aswinmcw (approved):
Both earlier points are addressed in this revision — FTZ is now derived from the golden alone and applied to both sides, and bounding A to [-1, 1] closes the inf-masking hole in swiglu/geglu. LGTM.

VirdhatchaniKN (approved):

tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
```diff
@@ -0,0 +1,145 @@
...
+
+@pytest.mark.parametrize("dim", [-1, 3])
+def test_geglu_op(device, dim):
+    """Exhaustive normal bfloat16 coverage for geglu.
+
+    PCC ≥ 0.999 is used because GELU uses a tanh polynomial approximation
+    with ~0.1-0.5% relative error in [-3, 3].  A is clamped to [-1, 1] to
+    prevent A * gelu(B) from overflowing to inf for large-magnitude B.
+    """
+    B = generate_bfloat16_bits(include_spl_values=False)  # (256, 256) — all normal bf16 values
```
KalaivaniMCW: If you are using a specific range of inputs, pls use the other function in eltwise_test_utils that generates bfloat16 bits for a particular range

tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
```diff
@@ -0,0 +1,145 @@
...
+@pytest.mark.parametrize("dim", [-1, 3])
+def test_swiglu_op(device, dim):
+    """Exhaustive normal bfloat16 coverage for swiglu.
+
+    PCC ≥ 0.999 is used because SFPU flushes sigmoid subnormals to zero for
+    large negative gate inputs, causing small legitimate differences from the
+    CPU reference.  A is clamped to [-1, 1] to prevent A * silu(B) from
+    overflowing to inf for large-magnitude B.
+    """
+    B = generate_bfloat16_bits(include_spl_values=False)  # (256, 256) — all normal bf16 values
```
KalaivaniMCW: If you are using a specific range of inputs, pls use the other function in eltwise_test_utils that generates bfloat16 bits for a particular range

tests/ttnn/unit_tests/operations/eltwise/test_unary_category5_bfloat16.py
abdulla-TT: Updated with `generate_bfloat16_bits_in_range`.

mateusznowakTT (approved):

PR #54539, opened 2026-08-27
argmax: Fix uint32 underflow in sub_core_grids work split

### LINKED ISSUE #51228: [ttnn/reduction] argmax sub_core_grids work-split: 2 uint32 underflows on the last core
These are statically audited bugs. The issues have enough detail that you can point claude and ask it to fix them. There is a small chance of false positives.

---

Two high-severity defects in `argmax`'s `distribute_work_to_cores()`, both on the `sub_core_grids` path, both underflowing uint32 for the last core.

**Line references are against `02246470f99`, and every file below was re-checked as byte-identical on `main` @ `007e45f8bfe`, so the line numbers are valid as of filing. If lines have since moved, the quoted code in each entry is the reliable anchor.**

Each item is self-contained — location, triggering configuration, evidence, and a suggested fix — and is independently fixable. Please tick items off as they land rather than closing the issue on the first fix.

CODEOWNERS: @tenstorrent/metalium-developers-mmfusedreduce

**These are two facets of the same broken work-split** — one because work runs out before the last core, one because the core count is never capped at the number of reduction blocks. Fix them together in `distribute_work_to_cores()`; a single clamp likely addresses both.

| # | Sev | Location | Defect |
|---|---|---|---|
| 1 | H | [`reduction/argmax/device/argmax_multi_core_program_factory.cpp:327`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp#L327) | In the sub_core_grids path, distribute_work_to_cores hands every core blocks_per_core blocks even when the work runs out before the last core, so the 'last core'... |
| 2 | H | [`reduction/argmax/device/argmax_multi_core_program_factory.cpp:331`](https://github.com/tenstorrent/tt-metal/blob/007e45f8bfe/ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp#L331) | In the sub_core_grids branch, distribute_work_to_cores() never caps the number of cores at the number of reduction blocks, so when cores > blocks the 'last core'... |

---

### 1. `ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp:327` — integer-underflow

**In the sub_core_grids path, distribute_work_to_cores hands every core blocks_per_core blocks even when the work runs out before the last core, so the 'last core' trim (red_dim_units_last0/1 = red_dim_units - (ideal - red_dim_units)) can go negative and wraps to a ~4 G uint32 read size.**

**How it fails.** ttnn.argmax(dim=-1) on a ROW_MAJOR bfloat16 DRAM tensor of shape [1,1,1,80] with sub_core_grids = CoreRange({0,0},{3,0}) (4 cores). alignment=32 -> min_red_dim_units_per_core=16, total_blocks=div_up(80,16)=5, total_cores=4 -> blocks_per_core=div_up(5,4)=2 -> red_dim_units0=32 for every core. ideal_red_dim_units = 4*32 = 128 != 80, num_cores1==0 so red_dim_units_last0 = 32 - (128-80) = 32-48 = 4294967280 (uint32 wrap). The last core receives num_reduce_units=4294967280 and src_read_size_last0 = 4294967280*2 (~4 GB after truncation) as runtime args 5/6, and reader_argmax_interleaved_multicore.cpp issues noc.async_read(..., src_read_size, ...) with that size and loops i over ~4 G elements -> L1 corruption / hang. Even before the wrap, cores 1 and 2 are given src_offset 64 and 128 bytes into a 160-byte page and read past the end of the row (garbage compared, indices >= red_dim_units reported).

**Evidence.** distribute_work_to_cores (same file, lines 45-72): `const uint32_t blocks_per_core = tt::div_up(total_blocks, total_cores); red_dim_units0 = blocks_per_core * min_red_dim_units_per_core;` — total_cores is the user's sub-grid core count, never reduced to div_up(total_blocks, blocks_per_core), so the assigned work can exceed the real work by more than one core's share. Lines 322-335 then assume exactly one partial core: `const int ideal_red_dim_units = (num_cores0*red_dim_units0)+(num_cores1*red_dim_units1); ... red_dim_units_last1 = ideal_red_dim_units == red_dim_units ? red_dim_units1 : red_dim_units1 - (ideal_red_dim_units - red_dim_units);` (and the mirrored last0 at 331-333). The non-sub_core_grids branch is safe precisely because split_work_to_cores() sizes num_cores from the work, so `ideal - red_dim_units < min_red_dim_units_per_core <= red_dim_units{0,1}`; the sub-grid branch breaks that invariant. Nothing guards it: ArgMaxDeviceOperation::validate_on_program_cache_miss (argmax_device_operation.cpp) only checks `sub_core_grids->ranges().size() <= 2`. Consumer: reader_argmax_interleaved_multicore.cpp lines 230-233 (`src_read_size`, `red_dim_units_this_core`) and lines 58-72.

**Suggested fix.** Clamp the number of participating cores to div_up(total_blocks, blocks_per_core) in the sub_core_grids branch (and give the trailing cores zero work), or compute each core's units as `std::min(red_dim_units0, red_dim_units > i*red_dim_units0 ? red_dim_units - i*red_dim_units0 : 0)` per core instead of trimming only the last one; add a TT_FATAL that the computed per-core unit count never underflows.

<sub>CODEOWNERS: `@tenstorrent/metalium-developers-mmfusedreduce`</sub>

- [ ] fixed

---

### 2. `ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp:331` — integer-underflow

**In the sub_core_grids branch, distribute_work_to_cores() never caps the number of cores at the number of reduction blocks, so when cores > blocks the 'last core' correction underflows uint32 and the last core is handed a ~4 GB NoC read size and a ~4e9-iteration scan loop.**

**How it fails.** ttnn.argmax(x, dim=-1, sub_core_grids=CoreRangeSet([CoreRange((0,0),(7,0))])) on a ROW_MAJOR bfloat16 DRAM tensor of shape [1,1,1,64]. min_red_dim_units_per_core = dram_alignment/sizeof(bfloat16) = 16, total_blocks = div_up(64,16) = 4, total_cores = 8 -> blocks_per_core = div_up(4,8) = 1 -> red_dim_units0 = 16 for all 8 cores. ideal_red_dim_units = 8*16 = 128 != red_dim_units (64), so lines 331-333 compute red_dim_units_last0 = 16u - (128u - 64u) = 16u - 64u = 4294967248. src_read_size_last0 = 4294967248*2 wraps to 4294967200. Core 7 then receives runtime arg 5 = 4294967200 (bytes to noc.async_read) and runtime arg 6 = 4294967248 (red_dim_units_this_core), i.e. a multi-GB DRAM read into a 32-byte L1 CB and a ~4-billion-iteration L1 scan -> L1 corruption / watcher fault / hang. Independently, cores 4..7 get red_dim_offset 64/80/96/112 >= red_dim_units = 64 and src_offset past the end of the input row, so they scan adjacent DRAM and can report out-of-range indices as the argmax. Reproduces for any W < 16*num_sub_grid_cores (e.g. W=144 with 8 cores also underflows: last0 = 32u-224u).

**Evidence.** distribute_work_to_cores(), sub_core_grids branch:
```
const uint32_t total_blocks = tt::div_up(red_dim_units, min_red_dim_units_per_core);
...
const auto total_cores = cores0.num_cores();
const uint32_t blocks_per_core = tt::div_up(total_blocks, total_cores);
red_dim_units0 = blocks_per_core * min_red_dim_units_per_core;
```
and the last-core correction:
```
red_dim_units_last0 = ideal_red_dim_units == red_dim_units
                          ? red_dim_units0
                          : red_dim_units0 - (ideal_red_dim_units - red_dim_units);
```
(red_dim_units_last0 / red_dim_units_last1 are uint32_t; the correction assumes overshoot < per-core units, which only holds when every core gets at least one block.)

The sibling non-sub_core_grids branch of the SAME function delegates to tt::tt_metal::split_work_to_cores (tt_metal/common/work_split.cpp:348-354), which explicitly caps the core count:
```
if (units_to_divide >= max_num_cores) { target_num_cores = max_num_cores; ... }
else { target_num_cores = units_to_divide; all_cores = num_cores_to_corerangeset(target_num_cores, grid_size, row_wise); }
```
so in that path ideal_red_dim_units - red_dim_units is always < min_red_dim_units_per_core <= red_dim_units_last and the subtraction can never underflow. The sub_core_grids branch is the one that omits the cap.

The consumer confirms the arguments are used raw, with no clamp: reader_argmax_interleaved_multicore.cpp lines 230-233 read src_read_size / red_dim_units_this_core straight from runtime args, feed src_read_size to noc.async_read (lines 58-63) and loop `for (uint32_t i = red_dim_offset; i < (red_dim_offset + red_dim_units_this_core); ++i)` (line 72).

Validation does not prevent it: ArgMaxDeviceOperation::validate_on_program_cache_miss only checks `args.sub_core_grids->ranges().size() <= 2` (argmax_device_operation.cpp:175-179); nothing relates the sub-grid core count to red_dim_units.

**Suggested fix.** Cap the number of cores used in the sub_core_grids branch at total_blocks before computing blocks_per_core (e.g. select the first min(total_cores, total_blocks) cores of the sub-grid via num_cores_to_corerangeset_in_subcoregrids / select_from_corerangeset, or reuse the CoreRangeSet overload of tt::tt_metal::split_work_to_cores(sub_core_grids, total_blocks) which already caps), so that every participating core owns at least one block and ideal_red_dim_units - red_dim_units < red_dim_units_last. As a belt-and-braces guard, TT_FATAL that (ideal_red_dim_units - red_dim_units) < red_dim_units_last before the subtraction.

<sub>CODEOWNERS: `@tenstorrent/metalium-developers-mmfusedreduce`</sub>

- [ ] fixed

---

ign-sampanna:
### PR Category

Bug fix

### Summary

`ttnn.argmax(..., sub_core_grids=...)` could hang the device or corrupt L1.

The multi-core split gave every core in the user grid the same rounded-up share, then subtracted the whole overshoot from the last core. When that overshoot was larger than one core's share, the `uint32_t` subtraction wrapped and the last core was handed a large NoC read and large iteration scan.

Both items in #51228 are this one invariant: the last-core trim assumed overshoot < one core's share. That holds on the default path (`split_work_to_cores` sizes the core count from the work) and fails on the `sub_core_grids` path (the core count is the user's grid).

Each core is now given `min(its share, units left at its offset)`, floored at 0. There is no last-core subtraction, so the wrap cannot happen. Cores with 0 units skip the NoC read (a zero-length read wedges the core) and still write a default partial for the reducer.

Closes #51228

### Notes for reviewers

- **Why this approach:** The issue suggests either capping the number of cores or clamping each core's work. This change uses the per-core clamp because it fixes both reported cases, while keeping the existing core set, semaphore configuration, and multicast setup unchanged.

- **Idle cores:** Cores with no remaining work still write their default partial result and signal `done_sem`. The reducer expects a result from every participating core, so this behavior is required.

- **Validation:** Since the last-core subtraction has been removed, the requested underflow `TT_FATAL` is no longer applicable. Instead, a coverage check verifies that the per-core slices exactly cover `red_dim_units`.

- **No change to existing splits:** For work distributions that were valid before, the new calculation produces the same per-core work.

- **Test coverage:** Added tests for both issue repros, a two-range core grid, clean work splits, and `dim=None` with idle cores. The operations run on a worker thread with a timeout so a device hang is reported as a test failure instead of blocking the test session.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/armax_sub_core_grids)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/armax_sub_core_grids)

### DIFF
--- a/tests/ttnn/unit_tests/operations/reduce/test_argmax.py
+++ b/tests/ttnn/unit_tests/operations/reduce/test_argmax.py
@@ -268,3 +268,92 @@ def test_argmax_reduce_all_multicore_no_deadlock(device):
     ref = int(torch.argmax(t.reshape(-1)))
     got = int(ttnn.to_torch(ttnn.from_device(result["out"])).item())
     assert got == ref, f"argmax reduce_all mismatch: got {got}, expected {ref}"
+
+
+def _run_argmax_in_worker(fn, timeout_s=60.0):
+    """Run `fn` on a worker thread so a device hang surfaces as an assertion, not a hung session."""
+    result = {}
+
+    def _run():
+        try:
+            result["out"] = fn()
+        except Exception as exc:  # surface device/compile errors to the main thread
+            result["error"] = exc
+
+    worker = threading.Thread(target=_run, daemon=True)
+    worker.start()
+    worker.join(timeout=timeout_s)
+    return worker, result
+
+
+@pytest.mark.timeout(120, method="thread")
+@pytest.mark.parametrize(
+    "tensor_shape, core_ranges",
+    [
+        # More sub-grid cores than blocks of work
+        ([8, 32], [((0, 0), (1, 1))]),  # 4 cores, 2 blocks
+        ([1, 64], [((0, 0), (7, 0))]),  # 8 cores, 4 blocks
+        # Fewer cores than blocks
+        ([1, 80], [((0, 0), (3, 0))]),  # 4 cores, 5 blocks -> 2 blocks each = 128 units for 80
+        ([1, 144], [((0, 0), (7, 0))]),  # 8 cores, 9 blocks -> 2 blocks each = 256 units for 144
+        # Two ranges exercise the second core group (cores1 / red_dim_units1).
+        ([4, 80], [((0, 0), (1, 0)), ((0, 1), (1, 1))]),
+        # Widths that already split cleanly
+        ([1, 128], [((0, 0), (7, 0))]),
+        ([8, 1024], [((0, 0), (7, 0))]),
+    ],
+)
+def test_argmax_sub_core_grids_work_split(device, tensor_shape, core_ranges):
+    """
+    Guard the sub_core_grids work-split: rounded-up per-core shares can overshoot the
+    reduction dim by more than one core's share, and trimming only the last core then underflowed
+    uint32 into a large read size and a hang.  Slices are now clamped per core instead.
+    """
+    torch.manual_seed(0)
+    grids = ttnn.CoreRangeSet(
+        {ttnn.CoreRange(ttnn.CoreCoord(*start), ttnn.CoreCoord(*end)) for start, end in core_ranges}
+    )
+
+    torch_tensor = torch.randn(*tensor_shape, dtype=torch.bfloat16)
+    ttnn_tensor = ttnn.from_torch(torch_tensor, ttnn.bfloat16, layout=ttnn.ROW_MAJOR_LAYOUT, device=device)
+
+    def _run():
+        out = ttnn.argmax(ttnn_tensor, dim=-1, keepdim=False, sub_core_grids=grids)
+        ttnn.synchronize_device(device)
+        return out
+
+    worker, result = _run_argmax_in_worker(_run)
+
+    assert not worker.is_alive(), f"ttnn.argmax timed out for shape={tensor_shape}, " f"sub_core_grids={core_ranges}"
+    if "error" in result:
+        raise result["error"]
+
+    ttnn_result = ttnn.to_torch(ttnn.from_device(result["out"])).to(torch.int32)
+    assert_equal(torch.argmax(torch_tensor, dim=-1, keepdim=False), ttnn_result)
+
+
+@pytest.mark.timeout(120, method="thread")
+def test_argmax_reduce_all_sub_core_grids_idle_cores(device):
+    """
+    dim=None uses the kernel's separate reduce_all block (its own partial-write and semaphore
+    sequence); drive it with a sub_core_grids wide enough to leave cores with no work.
+    """
+    torch.manual_seed(0)
+    # 4 cores for a 32-wide reduction dim = 2 blocks of work, so two cores are left idle.
+    grids = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(1, 1))})
+    t = torch.randn(8, 32, dtype=torch.bfloat16)
+    ttnn_in = ttnn.from_torch(t, ttnn.bfloat16, layout=ttnn.ROW_MAJOR_LAYOUT, device=device)
+
+    def _run():
+        out = ttnn.argmax(ttnn_in, sub_core_grids=grids)  # dim=None -> reduce_all
+        ttnn.synchronize_device(device)
+        return out
+
+    worker, result = _run_argmax_in_worker(_run)
+
+    assert not worker.is_alive(), "ttnn.argmax(dim=None) timed out with sub_core_grids leaving idle cores"
+    if "error" in result:
+        raise result["error"]
+
+    got = int(ttnn.to_torch(ttnn.from_device(result["out"])).item())
+    assert got == int(torch.argmax(t.reshape(-1))), f"reduce_all mismatch: got {got}"
--- a/ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp
@@ -4,6 +4,7 @@
 #include "argmax_device_operation.hpp"
 #include "ttnn/operations/reduction/reduce_op_validation.hpp"
 
+#include <algorithm>
 #include <tt-metalium/bfloat16.hpp>
 #include <tt-metalium/experimental/metal2_host_api/program_run_args.hpp>
 #include <tt-metalium/experimental/metal2_host_api/program_spec.hpp>
@@ -25,7 +26,7 @@ using namespace tt::tt_metal::experimental;
  *
  * If a sub_core_grids is provided, it will be used to distribute the work evenly across the cores.
  * Otherwise, we distribute to maximum of two core groups, with each core group getting a minimum of
- * `min_red_dim_units_per_core` elements to process, except the last core.
+ * `min_red_dim_units_per_core` elements to process, except where the reduction dim runs out.
  * @param device Pointer to the device
  * @param red_dim_units Total units in the reduction dimension
  * @param min_red_dim_units_per_core Minimum units per core (for alignment)
@@ -34,8 +35,9 @@ using namespace tt::tt_metal::experimental;
  *         - all_cores: CoreRangeSet of all cores
  *         - cores0: First group of cores
  *         - cores1: Second group of cores (if any)
- *         - red_dim_units0: Units assigned to first group per core
- *         - red_dim_units1: Units assigned to second group per core
+ *         - red_dim_units0: Nominal per-core units for the first group: the offset stride and CB page size.
+ *           The units a core actually processes are clamped against what is left (see units_for_core).
+ *         - red_dim_units1: Same, for the second group
  */
 static inline std::tuple<CoreRangeSet, CoreRangeSet, CoreRangeSet, uint32_t, uint32_t> distribute_work_to_cores(
     const tt::tt_metal::IDevice* device,
@@ -138,7 +140,7 @@ static inline std::tuple<CoreRangeSet, CoreRangeSet, CoreRangeSet, uint32_t, uin
  *    - Each group handles a different portion of the reduction dimension
  *    - cores0 handles red_dim_units0 elements
  *    - cores1 handles red_dim_units1 elements
- *    - Each core gets a minimum of `min_red_dim_units_per_core` elements to process, except the last core
+ *    - Each core gets a minimum of `min_red_dim_units_per_core` elements, except where the reduction dim runs out
  *
  * 2. Core Layout:
  *    - Cores are arranged in a grid pattern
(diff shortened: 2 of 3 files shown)


ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_interleaved_multicore.cpp
```diff
@@ -56,13 +56,17 @@ inline void find_argmax_for_core(
...
-        noc.async_read(
-            s_src,
-            src_dfb,
-            src_read_size,
-            {.page_id = outer_idx * inner_dim_units + j, .offset_bytes = src_offset},
-            {.offset_bytes = 0});
-        noc.async_read_barrier();
+        // A zero-length NoC read wedges the core, so a core with no units skips it. It still writes its
+        // default partial below -- keep that write, the reducer reads every core's slot.
+        if (red_dim_units_this_core > 0) {
```
iwroszTT, line 60: How about moving this check to the very beginning of the reader kernel, that is somewhere in the main()?

ign-sampanna: Moving the check out and returning early won't work: the reducer reads every core's result slot. If a zero-work core returns without writing its result, the reducer may read stale L1 data and produce a wrong index.

So a zero-work core must still:
- write a default result (the dtype's sentinel, index `0`)
- signal `done_sem`

Only the NoC read needs to be skipped.

### Possible approaches

**1. Move the check to the top of `find_argmax_for_core`** *(smallest change)*  
Hoist the check out of the `j` loop. The function can return early after writing the default results, avoiding the check on every `j`.

**2. Handle it on the host** *(no kernel change)*  
Give a zero-work core a valid one-block read at offset `0`. The scan still processes zero elements, so the data is ignored. A comment would explain why the read is intentional.

**3. Don't launch idle cores**  
Limit the participating cores so every core has work. This avoids the zero-work case, but changing the core count can break the current multicast setup because it supports only two core ranges. A reduced grid could require more ranges, causing some cores to miss the `start_sem` signal and hang.

The three approaches are possible. Option 1 is the smallest change, option 2 avoids kernel changes, and option 3 would require a fallback for multi-range grids. Happy to proceed with whichever approach you think is best.

ttnn/cpp/ttnn/operations/reduction/argmax/device/argmax_multi_core_program_factory.cpp
```diff
@@ -442,41 +425,55 @@ ttnn::device_operation::ProgramArtifacts ArgMaxMultiCoreProgramFactory::create_p
...
+                 {"src_read_size", static_cast<uint32_t>(red_dim_units_this_core * input_unit_size)},
+                 {"red_dim_units_this_core", red_dim_units_this_core}});
+            assigned_red_dim_units += red_dim_units_this_core;
         }
         kernel_run_args.push_back(std::move(reader_run_args1));
     }
 
+    // The per-core slices must tile the reduction dim exactly: no gaps (dropped elements) and no
+    // overlap (double-counted elements).
+    TT_FATAL(
```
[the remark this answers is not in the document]
ign-sampanna: Confirmed — `TT_FATAL` is active in Release builds and is not guarded by `DEBUG`/`NDEBUG`. `TT_ASSERT` is the one that becomes a no-op outside DEBUG, so `TT_FATAL` is intentional here. The check runs on the host during program creation, so there is no device-side runtime cost.

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
```diff
@@ -268,3 +268,92 @@ def _run():
...
+    "tensor_shape, core_ranges",
+    [
+        # More sub-grid cores than blocks of work
+        ([8, 32], [((0, 0), (1, 1))]),  # 4 cores, 2 blocks
+        ([1, 64], [((0, 0), (7, 0))]),  # 8 cores, 4 blocks
+        # Fewer cores than blocks
+        ([1, 80], [((0, 0), (3, 0))]),  # 4 cores, 5 blocks -> 2 blocks each = 128 units for 80
+        ([1, 144], [((0, 0), (7, 0))]),  # 8 cores, 9 blocks -> 2 blocks each = 256 units for 144
+        # Two ranges exercise the second core group (cores1 / red_dim_units1).
+        ([4, 80], [((0, 0), (1, 0)), ((0, 1), (1, 1))]),
```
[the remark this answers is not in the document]
ign-sampanna: `distribute_work_to_cores` is currently file-local, so testing it directly would require extracting the split arithmetic into a device-independent helper (and adding a separate unit-test path). For this bug fix, the existing device tests plus the host-side `TT_FATAL` coverage check provide the regression coverage. I can follow up with a standalone unit test if we decide to refactor the split logic.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_interleaved_multicore.cpp
iwroszTT: Let's keep the original code. Please update the comment to explain when `red_dim_units_this_core == 0` can happen.

vtsilytskyiTT (approved):
Looks good to me.

ttnn/cpp/ttnn/operations/reduction/argmax/device/kernels/reader_argmax_interleaved_multicore.cpp
ign-sampanna: Addressed!

bbradelTT (approved):

iwroszTT (approved):

PR #54548, opened 2026-08-27
[Bug fix] Power gradient: write through preallocated output and honour memory_config

### LINKED ISSUE #54549: `pow_bw`: preallocated `input_grad` left unwritten for `exponent ≈ 0`, and `memory_config` ignored when no output is preallocated
### Observed

`unary_backward.cpp:254-259`:

```cpp
input_grad = input_grad.value_or(ttnn::empty_like(input));          // 254: no memory_config
const float ZERO_THRESHOLD = std::numeric_limits<float>::epsilon() * 10.0f;
if (std::abs(exponent) < ZERO_THRESHOLD) {
    input_grad = ttnn::zeros_like(input);                           // 258: rebinds the optional
    grad_tensor.emplace_back(input_grad);
    return grad_tensor;
}
```

Two independent defects.

**1. Correctness — `exponent ≈ 0` only.** Line 258 assigns a *new* tensor to the local `std::optional` rather than writing into the caller's tensor, so a preallocated `input_grad` is never touched. The non-zero-exponent path does not have this problem: it ends in

```cpp
where(ttnn::lez(input), std::numeric_limits<float>::infinity(), final_result, output_mem_config, input_grad);
```

which passes `input_grad` as the output tensor and so writes through correctly. `fill_bw` threads the tensor in the same way. Line 258 is the outlier.

**2. Placement — only when nothing is preallocated.** Neither `empty_like` (254) nor `zeros_like` (258) is given `output_mem_config`, so the tensor they create inherits the input's config. When the caller *does* supply `input_grad`, the result correctly goes into that tensor and `memory_config` is rightly moot — a caller-supplied output should win. So this defect is confined to the no-preallocation case, and it affects both the `exponent ≈ 0` and normal paths.

Measured on `main`, bfloat16 / TILE, inputs in L1, `memory_config=DRAM`:

```
no prealloc, exponent 0   ->  returned config L1        (DRAM requested)   wrong
no prealloc, exponent 2   ->  returned config L1        (DRAM requested)   wrong
prealloc (L1),  exponent 0 ->  caller tensor untouched, still 7.0          wrong
prealloc (L1),  exponent 2 ->  caller tensor written, 7.0 -> 0.5           correct
```

### Assumptions

- The `input_grad` out-parameter is intended to be written in place. Every other backward op with preallocated outputs follows that contract, and `pow_bw`'s own non-zero-exponent path already does.
- `memory_config` is intended to control placement only when the op allocates the output itself. A caller-supplied `input_grad` takes precedence, so returning it in its own memory space is correct behaviour, not a defect.

### Gaps

- No test passes a preallocated `input_grad` to `pow_bw` and then inspects that tensor, so defect 1 is invisible.
- No test passes a `memory_config` that differs from the input's config without also preallocating, so defect 2 is invisible.

### Fix required

- Line 258: write through the preallocated tensor instead of rebinding, mirroring `fill_bw`:
  `ttnn::zeros_like(input, std::nullopt, std::nullopt, std::nullopt, output_mem_config, input_grad)`
- Line 254: pass `output_mem_config` to `empty_like`.
- Regression test covering both `exponent = 0` and `exponent != 0`, in both shapes:
  preallocated `input_grad` (assert the caller's tensor is written), and no preallocation
  with a `memory_config` differing from the input's (assert the result lands in the
  requested config).

VirdhatchaniKN:
### Issue
https://github.com/tenstorrent/tt-metal/issues/54549

### What this does

Two changes in `pow_bw`:

- For `|exponent| < eps`, writes the zero gradient **through** the caller's `input_grad` instead of rebinding the local `std::optional`.
- Passes `output_mem_config` to the `empty_like` and `zeros_like` calls.

### Why

`input_grad = ttnn::zeros_like(input);` replaced the optional rather than filling the caller's tensor, so a preallocated `input_grad` came back holding its previous contents — a caller reading its own tensor instead of the returned vector silently got stale data. The non-zero-exponent path already writes through via `where(..., input_grad)`, and `fill_bw` does the same; this line was the outlier.

Separately, neither creation call passed `output_mem_config`, so with nothing preallocated the gradient inherited the input's config instead of the requested one.

Measured on a Wormhole card, bfloat16/TILE, inputs in L1, `memory_config=DRAM`:

| | before | after |
| --- | --- | --- |
| no prealloc, exp 0 | L1 | DRAM |
| no prealloc, exp 2 | L1 | DRAM |
| prealloc, exp 0 | 7.0 (untouched) | 7.0 → 0.0 |
| prealloc, exp 2 | 7.0 → 0.5 | 7.0 → 0.5 |

A caller-supplied tensor still wins over `memory_config`, which is why the preallocated rows stay in L1; only the written value changes.

### Tests

| Test | Result |
| --- | --- |
| `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py` | 98 passed |
| `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_rpow.py` | 12 passed |
| `tests/ttnn/docs_examples/test_backward_examples.py -k pow_bw` | 2 passed |
| `tests/ttnn/nightly/unit_tests/operations/eltwise/backward/ -k "pow or fill"` | 119 passed |

No existing test preallocates `input_grad` or requests a differing `memory_config`, so these confirm no regression; the behaviour in the table was verified with a direct probe.

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/pow_bw_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/pow_bw_check)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
+++ b/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
@@ -302,9 +302,22 @@ def test_bw_pow_preallocated_output_wins_over_memory_config(input_shapes, expone
 
     This is the invariant whose behaviour changed: the gradient is written into the
     supplied tensor, so it stays in that tensor's memory space even when a different
-    config is requested. `full_like_impl` reaches this via
-    `optional_output_tensor.value().memory_config()` in `full_impl`, so flipping that
-    precedence would break it silently.
+    config is requested.
+
+    The two parametrizations reach it by different routes, and neither goes through
+    `full_impl`:
+
+    - exponent 0: `zeros_like` -> `full_like_impl` takes the fast path for a device
+      tensor in TILE layout whose dtype already matches, short-circuiting to
+      `ttnn::fill(tensor, 0.0f, memory_config, optional_output_tensor)`. `fill` is a
+      DEFINE_UNARY_OP_SCALAR_VARIANT, so precedence is decided by `unary_impl`
+      (eltwise/unary/unary.cpp:38-40), which prefers
+      `optional_output_tensor.value().memory_config()` over the requested config.
+    - exponent 2: `full_like_impl` is not involved. The write-through is the
+      `where(lez(input), inf, final_result, output_mem_config, input_grad)` at the end
+      of `pow_bw`, via `where`'s output-tensor handling.
+
+    Changing either of those would break this silently.
     """
     grad_data, grad_tensor = data_gen_with_range(input_shapes, -10, 10, device)
     in_data, input_tensor = data_gen_with_range(input_shapes, 0.1, 10, device)
--- a/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
+++ b/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
@@ -293,3 +293,30 @@ def test_bw_pow_honours_requested_memory_config(input_shapes, exponent, device):
     assert (
         result[0].memory_config().buffer_type == ttnn.BufferType.L1
     ), f"requested L1 but gradient landed in {result[0].memory_config().buffer_type}"
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("exponent", [0.0, 2.0])
+def test_bw_pow_preallocated_output_wins_over_memory_config(input_shapes, exponent, device):
+    """A caller-supplied input_grad takes precedence over memory_config.
+
+    This is the invariant whose behaviour changed: the gradient is written into the
+    supplied tensor, so it stays in that tensor's memory space even when a different
+    config is requested. `full_like_impl` reaches this via
+    `optional_output_tensor.value().memory_config()` in `full_impl`, so flipping that
+    precedence would break it silently.
+    """
+    grad_data, grad_tensor = data_gen_with_range(input_shapes, -10, 10, device)
+    in_data, input_tensor = data_gen_with_range(input_shapes, 0.1, 10, device)
+
+    preallocated = ttnn.full(input_shapes, 7.0, ttnn.bfloat16, ttnn.TILE_LAYOUT, device, ttnn.L1_MEMORY_CONFIG)
+
+    result = ttnn.pow_bw(
+        grad_tensor, input_tensor, exponent, memory_config=ttnn.DRAM_MEMORY_CONFIG, input_grad=preallocated
+    )
+
+    assert result[0].memory_config().buffer_type == ttnn.BufferType.L1, (
+        "preallocated output should win over memory_config, but the result moved to "
+        f"{result[0].memory_config().buffer_type}"
+    )
+    assert preallocated.memory_config().buffer_type == ttnn.BufferType.L1
--- a/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
+++ b/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
@@ -26,7 +26,7 @@ def test_negative_exponent(input_shapes, exponent, device):
     in_data, input_tensor = data_gen_with_range(input_shapes, -100, 100, device, True, seed=0)
     grad_data, grad_tensor = data_gen_with_range(input_shapes, -20, 20, device, seed=1)
 
-    with pytest.raises(RuntimeError) as _e:
+    with pytest.raises(RuntimeError) as _e:  # allow-pytest.raises: pre-existing, predates this hook
         tt_output_tensor_on_device = ttnn.pow_bw(grad_tensor, input_tensor, exponent)
     assert "exponent >= 0.0" in str(_e.value)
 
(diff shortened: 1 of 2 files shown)


ttnn/cpp/ttnn/operations/eltwise/unary_backward/unary_backward.cpp
```diff
@@ -251,11 +251,17 @@ std::vector<std::optional<Tensor>> pow_bw(
...
     std::optional<Tensor> input_grad) {
     std::vector<std::optional<Tensor>> grad_tensor;
-    input_grad = input_grad.value_or(ttnn::empty_like(input));
+    // Not value_or: its argument is evaluated even when input_grad already holds a
+    // tensor, which would allocate and immediately discard a device tensor.
+    if (!input_grad.has_value()) {
+        input_grad = ttnn::empty_like(input, std::nullopt, std::nullopt, std::nullopt, output_mem_config);
+    }
     const float ZERO_THRESHOLD = std::numeric_limits<float>::epsilon() * 10.0f;
     TT_FATAL(exponent >= 0.0, "negative exponents are not supported; use recip(pow(input,abs(exponent)))");
```
Aswinmcw: Since you're already removing a throwaway allocation here, this guard is worth hoisting above the `if (!input_grad.has_value())` block — for a negative exponent we now allocate a device tensor via `empty_like` and then immediately throw. `test_negative_exponent` (line 29, whose marker you touched in this PR) walks that exact path, so it's allocating and discarding on every run. Pure move, no behaviour change otherwise.

tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
```diff
@@ -248,3 +248,48 @@ def test_bw_unary_pow_edge_case_exponents(device, input_shapes, exponent, high1,
...
+    """
+    grad_data, grad_tensor = data_gen_with_range(input_shapes, -10, 10, device)
+    in_data, input_tensor = data_gen_with_range(input_shapes, 0.1, 10, device)
+
+    # inputs land in DRAM by default, so request L1 to force a divergence
+    result = ttnn.pow_bw(grad_tensor, input_tensor, exponent, memory_config=ttnn.L1_MEMORY_CONFIG)
+
+    assert (
+        result[0].memory_config().buffer_type == ttnn.BufferType.L1
+    ), f"requested L1 but gradient landed in {result[0].memory_config().buffer_type}"
```
Aswinmcw, line 295: Optional: the two tests each vary one axis — preallocated with no `memory_config`, and `memory_config` with nothing preallocated. Row 3 of your table (prealloc L1 + `memory_config=DRAM` → tensor wins, stays L1) is the one whose *behaviour changed* and the least obvious invariant of the three, and nothing pins it. `full_like_impl` gets there via `optional_output_tensor.value().memory_config()` in `full_impl`, so it'd break silently if that precedence were ever flipped. A third param case asserting the result stays L1 while DRAM was requested would cover it.

Aswinmcw (approved):

tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_pow.py
```diff
@@ -248,3 +248,75 @@ def test_bw_unary_pow_edge_case_exponents(device, input_shapes, exponent, high1,
...
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("exponent", [0.0, 2.0])
+def test_bw_pow_preallocated_output_wins_over_memory_config(input_shapes, exponent, device):
+    """A caller-supplied input_grad takes precedence over memory_config.
+
+    This is the invariant whose behaviour changed: the gradient is written into the
+    supplied tensor, so it stays in that tensor's memory space even when a different
+    config is requested. `full_like_impl` reaches this via
+    `optional_output_tensor.value().memory_config()` in `full_impl`, so flipping that
+    precedence would break it silently.
```
Aswinmcw: My fault from the last round — I pointed you at `full_impl` and that isn't the code this test actually reaches, for either parametrization.

For `exponent == 0.0`, `zeros_like` → `full_like_impl` takes the *fast path*: a device tensor in TILE layout whose dtype already matches (bfloat16 here) short-circuits to `ttnn::fill(tensor, 0.0f, memory_config, optional_output_tensor)` before the `full_impl` call. `fill` is a `DEFINE_UNARY_OP_SCALAR_VARIANT`, so precedence is actually decided in `unary_impl` (`eltwise/unary/unary.cpp:38-40`):

```cpp
auto output_memory_config = optional_output_tensor.has_value()
                                ? optional_output_tensor.value().memory_config()
                                : memory_config.value_or(input_tensor.memory_config());
```

For `exponent == 2.0`, `full_like_impl` isn't involved at all — the write-through is the `where(lez(input), inf, final_result, output_mem_config, input_grad)` at the end of `pow_bw`.

So the invariant holds and the assertions are right; only the citation is off. Worth fixing because it cuts both ways: someone changing `full_impl` would think this test covers them, and someone changing `unary_impl:38` wouldn't know this test is what guards it. Suggest naming `unary_impl` (and `where`'s output-tensor handling) instead.

Aswinmcw (approved):
Both points addressed. Re-diffed `2b756f7f3` → `4080f05dc`: the guard hoist is a pure move (`ZERO_THRESHOLD` is only read by the `if` below it, so nothing is reordered around a dependency), and the new precedence test is the case that was missing. One correction to a mechanism I gave you in the last round — details inline.

mateusznowakTT (approved):

PR #54553, opened 2026-08-27
Add Metalium 2.0 migration compatibility infrastructure

pavlejosipovic:
## Summary

This PR contains only the Metalium/TTNN framework contracts needed to migrate legacy Program factories to `ProgramSpec`. Operation factories and operation kernels are intentionally excluded and will remain in their individual migration commits.

Every compatibility path is explicit and opt-in. None adds a CB, backing buffer, data copy, runtime feature branch, or alternate operation kernel.

### DFB accessor aliases

One kernel binding can publish multiple named constexpr accessors for the same realized DFB slot.

Halo, Pool2D, Conv2D, and Conv3D have legacy kernels where distinct semantic CB names intentionally resolve to one physical CB index. Treating those names as separate DFBs would incorrectly model separate resources. Aliases preserve the existing physical CB identity without duplicate endpoints, storage, or data movement.

### Incomplete Gen1 endpoint coverage

A Gen1 DFB placement may omit its producer or consumer on selected nodes while still allowing at most one of each endpoint per node and requiring at least one complete placement.

Conv2D multicast grids contain padding and no-op core placements whose specialized kernel is guaranteed not to access one side of the DFB. Requiring a synthetic endpoint would misrepresent the topology and add synchronization that does not exist today. This option records the actual asymmetric placement, is incompatible with multi-binding, and remains rejected on Gen2.

### Gen1 plain-CB multi-binding

An explicit compatibility option allows multiple kernel instances to bind one physical Gen1 CB while lowering it as a plain CB.

Conv2D uses this for overlapped `ACT`/`ACT_TILIZED` storage and an explicitly synchronized split-reader CB. Conv3D uses it for compute/writer reduction buffers whose active role varies by core. Converting these relationships into ordinary DFB credit-flow edges would change the existing handshake and could double-account credits. The option preserves the original CB and kernel-owned synchronization and is rejected on Gen2.

### Borrowed-memory subranges

A DFB can borrow a bounds-checked non-zero byte offset inside an existing tensor allocation.

Conv2D stores matmul partials in a legacy subrange of the output allocation. Expressing that offset directly preserves the accumulator address, L1 footprint, and output layout. Without it, the migration would require scratch storage and a copy or a different physical layout.

### Realized borrowed-DFB validation

Borrowed sharded DFB ranges are validated against their physical backing buffers after lowering.

The declarative spec does not yet know the final allocation and offsets needed to prove that the Conv2D subrange is in bounds. Post-realization validation retains this safety check at the first point where the required physical information exists, without altering allocation.

### Constexpr-discard-only bindings

Named DFB, semaphore, scratchpad, and tensor bindings may remain unresolved only when a compile-time branch proves their accessor unreachable.

Fold has optional DFBs, ConvertToHWC has optional tensors, Halo has optional scratchpad and tensor accessors, and Conv2D has optional DFBs and semaphores. Their typed kernel schemas need to name these resources even when a specialization compiles their use away. The discard-only state emits an invalid placeholder instead of manufacturing a dummy resource, rejects inconsistent declarations, and leaves runtime access invalid.

### Typed constexpr arguments

Named compile-time arguments can emit typed template values.

Pool consumers require typed reduction selections to retain compile-time specialization. The generic mechanism avoids feature macros and runtime branches. This PR adds only the argument facility; the consuming operation-kernel changes are excluded.

### Architecture-correct DFB implicit synchronization

Implicit DFB synchronization applies only to endpoints using `DataMovementGen2Config`.

Wormhole and Blackhole kernels already execute explicit reserve/push/wait/pop handshakes. Adding implicit credit accounting would create a second synchronization path and could change performance or deadlock. The restriction preserves the existing Gen1 protocol while retaining Quasar implicit-sync behavior and opt-out controls.

### Post-realization validation hooks

Program factories can inspect read-only realized DFB allocation information from a cache-miss post-realization hook.

Conv2D previously used `post_conv2d_op_memory_checks` for realized CB-size and allocator-delta diagnostics. Those properties cannot be checked before `ProgramSpec` lowering because physical placement does not yet exist. The hook restores the diagnostic point and cannot mutate allocation or dispatch.

The PR remains draft so these shared framework contracts can be reviewed independently before the operation-specific migrations are proposed.

## Related migration branch

The vision-operation migrations that consume this infrastructure are available on [`pjosipovic/metalium-2.0-convolution-migration-assessment`](https://github.com/tenstorrent/tt-metal/tree/pjosipovic/metalium-2.0-convolution-migration-assessment).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:pjosipovic/metalium-2.0-migration-infrastructure)

### DIFF
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
@@ -203,6 +203,7 @@ struct DFBAdvancedOptions {
     // It is NOT supported on Gen2 architectures: setting this flag on a Gen2
     // target is a hard error, whether or not any instance is actually multi-bound.
     bool allow_instance_multi_binding = false;
+
 };
 
 struct AdvancedKernelRunArgs {
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/dataflow_buffer_spec.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/dataflow_buffer_spec.hpp
@@ -129,6 +129,13 @@ struct DataflowBufferSpec {
     // (TODO: this should become std::variant<TensorParamName, BufferParameterName>.)
     std::optional<TensorParamName> borrowed_from = std::nullopt;
 
+    // Byte offset from the borrowed memory object's per-node base address. This
+    // preserves legacy globally allocated CB views that occupy a subrange of a
+    // sharded tensor without allocating another DFB or copying data. A nonzero offset must be aligned
+    // both to entry_size (so the view begins on a DFB entry boundary) and to the backing buffer's
+    // L1/NoC address alignment (so the absolute base + offset remains a legal NoC address).
+    uint32_t borrowed_memory_offset = 0;
+
     //////////////////////////////
     // Advanced options (see advanced_options.hpp)
     //////////////////////////////
--- a/tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
+++ b/tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
@@ -138,9 +138,19 @@ struct KernelSpec {
         enum class AccessPattern { STRIDED, ALL, BLOCKED };
 
         DFBSpecName dfb_spec_name;   // identify the DFB within the ProgramSpec
-        std::string accessor_name;   // DFB accessor name (used in the kernel source code)
+        std::string accessor_name;   // Primary DFB accessor name (used in the kernel source code)
         EndpointType endpoint_type;  // producer or consumer
         AccessPattern access_pattern = AccessPattern::STRIDED;
+        // Additional constexpr accessor names for the same DFB device slot. This is an escape hatch
+        // for migrating legacy kernels in which multiple semantic CB names intentionally shared one
+        // physical CB index. Aliases do not add endpoints, storage, or data movement.
+        std::vector<std::string> accessor_aliases = {};
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named spec is absent, emit a placeholder token and allocate no DFB storage. Using that
+        // token at runtime is invalid; normal DFB bindings must leave this false. The placeholder is
+        // a bare numeric sentinel to preserve Gen1 APIs that accept raw CB ids, so it cannot carry a
+        // TensorAccessor-style live-use trap.
+        bool allow_unbound_for_constexpr_discard = false;
     };
     Group<DFBBinding> dfb_bindings;
 
@@ -150,6 +160,10 @@ struct KernelSpec {
     struct SemaphoreBinding {
         SemaphoreSpecName semaphore_spec_name;  // identify the semaphore within the ProgramSpec
         std::string accessor_name;              // semaphore accessor name (used in the kernel source code)
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named spec is absent, emit an invalid numeric placeholder and allocate no semaphore.
+        // The numeric form preserves legacy semaphore APIs and therefore cannot carry a live-use trap.
+        bool allow_unbound_for_constexpr_discard = false;
     };
     Group<SemaphoreBinding> semaphore_bindings;
 
@@ -160,6 +174,10 @@ struct KernelSpec {
     struct ScratchpadBinding {
         ScratchpadSpecName scratchpad_spec_name;  // identify the scratchpad within the ProgramSpec
         std::string accessor_name;                // scratchpad accessor name (used in the kernel source code)
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named spec is absent, emit a zero-sized token and allocate no L1. Using that token at
+        // runtime is invalid; normal scratchpad bindings must leave this false.
+        bool allow_unbound_for_constexpr_discard = false;
     };
     Group<ScratchpadBinding> scratchpad_bindings;
 
@@ -174,6 +192,9 @@ struct KernelSpec {
     struct TensorBinding {
         TensorParamName tensor_parameter_name;  // identify the TensorParameter within the ProgramSpec
         std::string accessor_name;              // tensor accessor name (used in the kernel source code)
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named parameter is absent, emit token metadata but no CTA/CRTA payload or runtime patching.
+        bool allow_unbound_for_constexpr_discard = false;
     };
     Group<TensorBinding> tensor_bindings;
 
--- a/tests/tt_metal/tt_metal/api/dataflow_buffer/test_borrowed_memory_dataflow_buffer.cpp
+++ b/tests/tt_metal/tt_metal/api/dataflow_buffer/test_borrowed_memory_dataflow_buffer.cpp
@@ -78,6 +78,7 @@ struct BorrowedDFBTestConfig {
     DFBAccessPattern cap      = DFBAccessPattern::STRIDED;  // producer is always STRIDED
     bool tensix_consumer      = false;
     bool verify_data          = false;
+    uint32_t borrowed_memory_offset = 0;
     std::optional<uint32_t> num_entries_override = std::nullopt;
     bool expect_attach_fatal = false;
 };
@@ -180,12 +181,16 @@ void run_borrowed_memory_dfb_program(
         .num_entries = cfg.num_entries,
         .data_format_metadata = tt::DataFormat::Float16_b,
         .borrowed_from = experimental::TensorParamName{"dfb_ring_tensor"},
+        .borrowed_memory_offset = cfg.borrowed_memory_offset,
     };
 
     // --- TensorParameters ---
     const TensorSpec src_spec  = make_flat_dram_tensor_spec(cfg.entry_size, cfg.num_entries);
     const TensorSpec dst_spec  = make_flat_dram_tensor_spec(cfg.entry_size, cfg.num_entries);
-    const TensorSpec ring_spec = make_flat_l1_tensor_spec(cfg.entry_size, cfg.num_entries);
+    ASSERT_EQ(cfg.borrowed_memory_offset % cfg.entry_size, 0u)
+        << "The runtime test keeps the borrowed subrange aligned to a DFB entry";
+    const uint32_t borrowed_prefix_entries = cfg.borrowed_memory_offset / cfg.entry_size;
+    const TensorSpec ring_spec = make_flat_l1_tensor_spec(cfg.entry_size, cfg.num_entries + borrowed_prefix_entries);
 
     spec.tensor_parameters.push_back({.unique_id = experimental::TensorParamName{"src_tensor"}, .spec = src_spec});
     if (!cfg.tensix_consumer) {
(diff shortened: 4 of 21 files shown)


tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
```diff
@@ -138,9 +138,19 @@ struct KernelSpec {
...
 
         DFBSpecName dfb_spec_name;   // identify the DFB within the ProgramSpec
-        std::string accessor_name;   // DFB accessor name (used in the kernel source code)
+        std::string accessor_name;   // Primary DFB accessor name (used in the kernel source code)
         EndpointType endpoint_type;  // producer or consumer
         AccessPattern access_pattern = AccessPattern::STRIDED;
+        // Additional constexpr accessor names for the same DFB device slot. This is an escape hatch
+        // for migrating legacy kernels in which multiple semantic CB names intentionally shared one
+        // physical CB index. Aliases do not add endpoints, storage, or data movement.
+        std::vector<std::string> accessor_aliases = {};
```
akerteszTT, line 147: DFB aliasing is already fully supported via advanced options. This is superfluous, and does not solve the general problem -- aliased CBs/DFBs are permitted to different in other dimensions than just name.

tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
```diff
@@ -138,9 +138,19 @@ struct KernelSpec {
...
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named spec is absent, emit a placeholder token and allocate no DFB storage. Using that
+        // token at runtime is invalid; normal DFB bindings must leave this false. The placeholder is
+        // a bare numeric sentinel to preserve Gen1 APIs that accept raw CB ids, so it cannot carry a
+        // TensorAccessor-style live-use trap.
+        bool allow_unbound_for_constexpr_discard = false;
```
akerteszTT, line 153: This is the same feature for unbound DFBs that you previously requested. Something similar is already in PR review.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
```diff
@@ -162,6 +163,14 @@ struct DFBAdvancedOptions {
...
     // target is a hard error, whether or not any instance is actually multi-bound.
     bool allow_instance_multi_binding = false;
+
+    // Legacy no-op placement compatibility. Allows a per-node DFB instance to have only one side
+    // of its endpoint pair when the kernel on that node is guaranteed not to access the DFB.
+    // This does not create an endpoint, storage copy, or synchronization; misuse is unsafe. It is
+    // mutually exclusive with allow_instance_multi_binding: incomplete coverage still permits at
+    // most one producer and at most one consumer on each node. Gen1 compatibility only: Gen2 DFB
+    // lowering requires a real producer and consumer endpoint on every instantiated node.
+    bool allow_incomplete_endpoint_coverage = false;
```
akerteszTT: We have discussed this at length within the runtime team, and we will NOT permit DFBs that are bound at only one endpoint.

A DFB is a SW FIFO abstraction. There is no legitimate use case for a FIFO with only one endpoint. The purpose of a FIFO is to provide synchronization semantics between two endpoints.

The reason that this type of use case proliferated with CBs is that CBs gradually took on a huge assortment of semantics that are completely orthogonal to their original purpose as SW FIFOs. A "single ended" CB makes sense only when a "pretend" CB is overlain onto an allocated tensor, and used to populate that tensor's data.

One of the goals of Metal 2.0 is to create runtime abstractions that are appropriate to the use case. A CB built on borrowed memory is very seldom that.

There are two solutions to the single-ended problem already supported in Metal 2.0:
 1. (Ideal) Use a `LocalTensorAccessor`. This is abstraction for the local region of L1 memory occupied by a tensor, as bound to a kernel for access.
 2. (Cop-out) Use a fictional kernel binding for the other endpoint.

The current porting recipe uses the option 2 "cop-out", and binds both endpoints to the same kernel. However, in time we would like to attempt to migrate as many use cases as possible to the new semantics (which will be fully supported by LLK, by the way). Using a DFB to represent an arbitrary region of memory wastes scarce hardware resources for synchronization -- as especially acute problem in Quasar and in megakernel applications.

tt_metal/api/tt-metalium/experimental/metal2_host_api/dataflow_buffer_spec.hpp
```diff
@@ -129,6 +129,13 @@ struct DataflowBufferSpec {
     // (TODO: this should become std::variant<TensorParamName, BufferParameterName>.)
     std::optional<TensorParamName> borrowed_from = std::nullopt;
 
+    // Byte offset from the borrowed memory object's per-node base address. This
+    // preserves legacy globally allocated CB views that occupy a subrange of a
+    // sharded tensor without allocating another DFB or copying data. A nonzero offset must be aligned
+    // both to entry_size (so the view begins on a DFB entry boundary) and to the backing buffer's
+    // L1/NoC address alignment (so the absolute base + offset remains a legal NoC address).
+    uint32_t borrowed_memory_offset = 0;
```
akerteszTT, line 137: This is a legacy hack that will not be adopted by Metal 2.0. The same functionality will be provided via a different mechanism.

Briefly: The "pinned CB" / "borrowed memory DFB" mechanism has two entirely orthogonal use cases:
 1. A representation of a tensor's local L1 memory, maskerading as a DFB. These use cases frequently either don't need FIFO semantics, or abuse FIFO semantics (e.g. by reaching in a moving write pointers). This was forced by the fact that the legacy LLK APIs demanded a view of L1 memory in CB form.
 2. A way to self-manage the allocation of ephemeral (i.e. Program-execution scope) memory, directing it explicitly into a reserved allocation rather than relying on the runtime's own allocator. This used by Blaze and Forge.

These are very distinct semantics. The `offset` CB hack was added to support 2 for Blitz/Blaze. We have a plan to support lightweight area allocation properly, which is fully aligned with the Blaze team; it will be implemented shortly.

At last analysis, there were no uses of `offset` outside the area allocation mechanism. This should not be permitted.

tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
```diff
@@ -150,6 +160,10 @@ struct KernelSpec {
     struct SemaphoreBinding {
         SemaphoreSpecName semaphore_spec_name;  // identify the semaphore within the ProgramSpec
         std::string accessor_name;              // semaphore accessor name (used in the kernel source code)
+        // Compatibility escape hatch for a token referenced only from a compile-time-discarded branch.
+        // If the named spec is absent, emit an invalid numeric placeholder and allocate no semaphore.
+        // The numeric form preserves legacy semaphore APIs and therefore cannot carry a live-use trap.
+        bool allow_unbound_for_constexpr_discard = false;
```
akerteszTT, line 166: See above. This requested functionality will supported by a different mechanism.

akerteszTT (commented):
I've reviewed all of the proposed API changes. 

All of them are Metal 2.0 roadmap items for which we want to improve on the legacy semantics to expose a clean programming model. These are all under development.

In each case, we have actively discussed existing op use cases with the stakeholder teams (Blaze and the ops porting team) to figure out the best solution. In some cases, we've had to make compromises with the desired programming model; in others, we've agreed to evolve the ops usage towards the new, cleaner semantics.

I want to avoid at all costs unthinkingly reproducing legacy hacks with ill-conceived semantics in the new API. Doing so negates all the effort that has gone into trying to clean up the Metalium programming model.

tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
pavlejosipovic: Thanks, I audited these uses against existing DFB aliasing.

Optional-resource aliases can move to the unbound-resource mechanism under review. The remaining use case needs several kernel constexpr names to resolve to one realized hardware CB slot. `alias_with` appears to share backing storage but still creates distinct DFB/device-slot identities, which can increase scarce CB-slot usage.

Is there an existing configuration that guarantees one realized slot with multiple kernel-visible names? If not, I would like this use case considered in the planned aliasing model.

tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
pavlejosipovic: Thanks, this is great. I will migrate to the runtime-owned mechanism and remove this implementation when it lands.

Fold and Conv2D have kernel schemas that name optional DFBs, while an NTTP specialization discards every access when that DFB is absent. NTTP removes the device-code access, but currently does not make the missing named binding valid during ProgramSpec validation and lowering.

I need the replacement to allow that binding to remain absent without a dummy DFB/CB, endpoint, runtime argument, allocation, synchronization, or copy.

tt_metal/api/tt-metalium/experimental/metal2_host_api/advanced_options.hpp
pavlejosipovic: Thanks, this worked well. I replaced address-only Conv2D DFBs with `LocalTensorAccessor` and used fictional opposite-role bindings for padding/no-op placements.

I then removed `allow_incomplete_endpoint_coverage` entirely. The replacements add no CB, allocation, copy, kernel access, or synchronization.

tt_metal/api/tt-metalium/experimental/metal2_host_api/dataflow_buffer_spec.hpp
pavlejosipovic: Thanks for explaining the planned area allocator. I agree that a generic borrowed-memory offset is not the desired final API.

Conv2D needs `MATMUL_PARTIALS` at an exact non-zero subrange of the existing output tensor allocation, preserving per-shard bounds, cache-hit address patching, L1 footprint, and zero-copy behavior.

Will the planned area allocator support an exact tensor-backed subrange, rather than only a separate ephemeral allocation? If so, it should replace this extension cleanly.

tt_metal/api/tt-metalium/experimental/metal2_host_api/kernel_spec.hpp
pavlejosipovic: Thanks, I will use the runtime-owned mechanism when available.

Conv2D needs an optional semaphore to be absent when NTTP specialization discards every access, without allocating a dummy semaphore or emitting a runtime argument. The same mechanism should ideally cover DFB, semaphore, scratchpad, and tensor bindings consistently.

Issue #51249, opened 2026-07-27
Implement the shared PyArrow schema and LLK Parquet writer


ndivnicTT:
Parent: #51243

## Summary
Add Parquet output alongside current CSV reporting using one typed writer that aligns every LLK performance frame to the canonical shared nullable schema.

## Scope
- Define the explicit PyArrow schema.
- Rename aliases, validate types, add absent fields as `NULL`, and reorder columns.
- Integrate Parquet output with the current report fixture while retaining CSV compatibility.
- Include mandatory run/test metadata and separate congestion UNPACK/PACK fields.

## Deliverables
- Shared PyArrow schema and writer API.
- Representative test-family unit tests.

## Acceptance Criteria
- [ ] SFPU unary, math matmul, and pack/untilize use identical physical schemas.
- [ ] Test-specific and unexecuted run-type fields are `NULL`.
- [ ] Values preserve canonical types and units.
- [ ] Unknown fields fail before writing.
- [ ] Existing tests emit CSV and Parquet from the same report data.

## Dependencies
- Milestone 1 of #51243.

PR #54563, opened 2026-08-27
[Documentation Only] Document the LLK perf infrastructure

nstojicTT:
### PR Category

Documentation Only. One new file, `tt_metal/tt-llk/docs/perf_infra/README.md`. No code, test, or workflow changes.

Branched from `main`, so it merges independently of every perf PR in review (#53928, #53021, #51954, #53752).

### Summary

The LLK perf pipeline enforces rules that are nowhere explained. A developer who adds a sweep parameter learns about `wide_schema.py` when the Parquet writer raises at the end of a 20-minute run, and about `test_schemas.py` when the gate fails in CI. This documents the pipeline so both are findable before the run.

**Structure: problem → solution first, implementation last.**

| Section | Answers |
|---|---|
| The two rules | The two things that break a build, stated in a table |
| I want to… | A routing table from your task to the procedure |
| **Tasks** | Twelve procedures, opening with a gate-failure → fix lookup keyed on the actual assertion text |
| **Common pitfalls** | Fifteen mistakes that pass every gate, or fail far from their cause |
| **How it works** | Pipeline, schema ownership, the published table, run identity, storage, tools |

A sample of the pitfalls, since that section is the part hardest to reconstruct from the code:

- Updating `columns` without increasing `version` — passes every gate, and silently makes two reports of the same test incomparable
- `-n auto` addresses an 8×8 core grid; Wormhole ships with harvested rows, so it is *wrong*, not slow
- `<test>.csv` holds raw loop totals, not per-tile figures
- A 2–6% step on a matmul test is bistable hardware behaviour, not a regression
- Lost sweep coverage looks exactly like an improvement in any diff

### Review history

Two rounds of feedback, both applied.

**Round 1 — "reads like a novel, not a reference; the before/motivation goes stale."** Removed the six-defect problem statement, the before/now table, the merged-PR inventory, the design-principles essay, and the gate-study narrative. Restructured around tasks.

**Round 2 — "50% of this is not needed, and I do not want to read four files."** Collapsed `tasks.md`, `pitfalls.md` and `reference.md` into this one file: 1294 lines in the first draft, 638 now. What went beyond the merge itself was per-task preamble that restated the heading, a Tools section and a File map that listed the same paths twice, the gate-study table reduced to a paragraph, two pitfalls merged where they shared a mechanism, and the pipeline diagram cut from fifteen lines to four.

### Notes for reviewers

**The numbers are read from the code on `main`, not remembered** — 23 WH/BH and 18 Quasar perf tests under the gate (`derive_perf_test_schemas()`), 82 published columns of which 75 come from tests and 7 from CI (`DB_SCHEMA`). Worth checking; a reference with one wrong count stops being trusted.

**All 25 internal links verified** against the 42 headings.

**Where a tool is not on `main`, it is not presented as available** — `--perf-run-types` is left out of the task text for that reason.

**Not included:** a discovery link from `tests/README.md` or the LLK root `README.md`. Say the word and I will add it.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-infra-docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-infra-docs)


mvlahovicTT:
@nstojicTT heads up on two lines here, since they came up while fixing #54592 and you flagged the approach on my first attempt.

The docs currently say, in two places:

> **Counter and metric columns.** Deferred (issue #51249). They join the table as nullable once a counter run captures their exact names.

> Metric columns are **not** in `DB_SCHEMA` yet. They join the published table as nullable once a counter run captures their exact names (issue #51249).

That was the plan, and `wide_schema.py` carried a matching `TODO`, but it is not what we are doing. I did exactly what those lines describe in #54594, adding all 306 as nullable, and it is the wrong shape: [DATA-1652](https://tenstorrent.atlassian.net/browse/DATA-1652) is a contract with the data team covering timing, sweep config, SoL and provenance in about 82 columns, and [DATA-1669](https://tenstorrent.atlassian.net/browse/DATA-1669) asks to make it narrower. Publishing them takes it to 388, and since no workflow passes `--enable-perf-counters` they are NULL in every row the gate sees (measured: 0 of 306 on the no-counters path).

#54607 drops them through `DROPPED_COLUMNS` instead, alongside `TEXT_SIZE(...)`, and removes the stale `TODO` from `wide_schema.py`. Since your PR is still in review it is probably cleanest if you adjust those two lines here rather than me doing it in a follow-up, but happy either way, just say which you prefer.

One thing worth writing into the docs while you are in there, if you agree: the rule a developer needs is not only "a new CSV column must exist in the published table" (Rule 1) but "or be in `DROPPED_COLUMNS` if it is diagnostic output the gate does not consume". The current Rule 1 wording sends you straight to widening the table, which is the trap I fell into.

Also flagging that DATA-1669 is closed as Done with no comments and unchecked acceptance criteria, so the decision to stop widening the table is not recorded anywhere a developer would find it. That is really what caused this.

tt_metal/tt-llk/docs/perf_infra/README.md
```diff
@@ -0,0 +1,638 @@
...
+
+### A dynamically built parameter list is invisible to the gate
+
+The gate reads `templates=` and `runtimes=` with `ast`. It handles a list written inline,
+assigned to a variable, or given as a dict entry — but not one built by a comprehension
+or returned by a helper. Such a test under-reports its columns, the gate passes, and the
+run then fails at the Parquet writer. Keep the lists literal; if you cannot,
+`test_perf_report_hw_free.py` is the exact check.
+
+### `-n auto` addresses cores that may not exist
```
ndivnicTT: What is this? Can you research and check do we have this issue anywhere?

tt_metal/tt-llk/docs/perf_infra/README.md
```diff
@@ -0,0 +1,638 @@
...
+Mistakes that pass every gate, or fail far from their cause.
+
+### Changing `columns` without increasing `version`
+
+The gate compares the column **set**, so this passes, and nothing else notices either.
+`version` is the only signal that two reports of the same test measure the same thing —
+two v4 reports with different columns are silently incomparable, and it surfaces weeks
+later as a trend nobody can explain.
+
+### A dynamically built parameter list is invisible to the gate
```
ndivnicTT: You fixed this issue, didn't you?

tt_metal/tt-llk/docs/perf_infra/README.md
```diff
@@ -0,0 +1,638 @@
...
+| `helpers/perf/wide_schema.py` | `DB_SCHEMA` — the published WH/BH table |
+| `helpers/perf/wide_schema_quasar.py` | The published Quasar table |
+| `helpers/perf/test_schemas.py` | Per-test columns, versions, aliases |
+
+No header is ever a literal: `stat_column("L1_TO_L1", "mean")` → `mean(L1_TO_L1)`, and
+likewise `metric_column`, `text_size_column`, `counter_base`, `cycles_of`.
+
+## The published table
+
+`DB_SCHEMA` is 82 columns; one row is one test configuration in one run. It is both the
```
ndivnicTT: Is this accurate? Maybe reference the code instead of giving number 82 as it might change occasionally

General suggestion for concrete numbers in this section

tt_metal/tt-llk/docs/perf_infra/README.md
```diff
@@ -0,0 +1,638 @@
...
+
+Everything except `marker` and the mandatory provenance keys is **nullable**. That is
+what lets 23 tests with different sweeps share one table — each fills its own columns and
+the rest stay NULL, so the table need not be complete to be useful. Timing columns are
+formula-driven: the full `{mean, std} × _TIMING_BASES` grid, so the schema is a superset
+of any run configuration rather than of what one nightly sampled.
+
+Deferred: counter and metric columns join as nullable once a counter run captures their
+names (#51249); canonicalizing columns that name the same thing differently —
+`c_dimm`/`k_dimm`/`r_dimm` vs `in0_c_dim`/`ct_dim` — needs sign-off (#51245).
```
ndivnicTT: needs sign-off - please erase this

ndivnicTT (approved):
LGTM!
The doc needs to be updated to the fresh state, probably would be the best to merge it after perf regression PR gate.
Also, please check the comments

PR #47359, opened 2026-06-17
[Bug fix] Fix Llama-3.1-8B eval accuracy under vLLM async scheduling

ifabijanic-tt:
### Summary
Under --async-scheduling, Llama-3.1-8B produced incorrect decode output at reset / batch-condense steps, regressing the eval/accuracy tests. The main fix restores correct generation by keeping the authoritative device token across decode resets. Two supporting fixes restore seeded-sampling determinism on multi-chip Blackhole (which also unblocks the seed=0 chat-completions test).

### Main fix - keep the device "async-ahead" token across decode reset (generator.py)
Under async scheduling, the host-supplied last token lags device sampling at reset / prefill→decode mode-switch steps. Staging that stale host token (and its lagging position) re-processes a token at the wrong position - overwriting KV / regenerating positions - which produces duplicate or flipped tokens under concurrency and degrades eval accuracy.

This keeps the authoritative device token (and its position) for continuing slots, taking host tokens only for slots freshly prefilled since the last decode submit (_slots_prefilled_since_decode). slot_remap carries global slot indices (the plugin offsets each DP rank's local [0,B) map by rank*B), so it's rebased to per-rank local before gathering dev_toks/dev_pos, otherwise rank ≥1 indexes out of bounds. Position pairing is gated by TT_FIX_RESET_POS (default on).

#### Restore seed registration + RNG-counter↔position alignment (generator.py)
Main #45166 ("Clean up sampling separation in generators") dropped reset_seed_from_slots_if_needed + align_seed_counters_to_positions from sample_decode_on_device. Without them the per-request seed never reaches the device (the seed manager stays unseeded) and the RNG counter isn't tied to the absolute decode position, so seeded sampling is non-deterministic under async scheduling. Restores both calls before get_new_values, matching the known-good path. (This is what was failing the seed=0 test; the SeedManager methods themselves were untouched - they just stopped being called.)

#### Disable batched prefill on multi-chip Blackhole (model_config.py, #4004)
Prefill logits are batch-variant on multi-chip Blackhole (float-reduction order depends on batched-token count), so same-seed requests in different-sized prefill batches diverge. Disabled for Llama-3.1-8B on P300/P150x4/P150x8 (alongside main's existing Qwen3-32B/P150x4 case).

### Potential impact on other models
[...]

PR #54565, opened 2026-08-27
[Bug fix] Gate the async-ahead token keep on `supports_async_decode`

nyoshifujiTT:
### PR Category

Bug fix

### Summary

At a reset step `Generator.decode_forward` may discard the host's token and position for a slot and adopt whatever the previous decode left in the device trace-input buffers, whenever the two positions are continuous (`dev_pos == host_pos` or `host_pos + 1`). That recovers the authoritative token under vLLM async scheduling, where the host trails the device by one step, and it was introduced for that case in #47359.

Without that lag there is nothing to recover. For a model declaring `model_capabilities["supports_async_decode"] = False` the same rule can only substitute a stale token from an earlier request that happened to reuse the position, silently changing what the model conditions on. Observed on Qwen3.6-27B P150x4 with on-device sampling: 8 of 32 requests were truncated at the second token, 0 of 32 after this change, with no measurable throughput difference at B=1.

This affects every model declaring `supports_async_decode = False` alongside on-device sampling, today Qwen3.6-27B, Gemma4 and Diffusion-Gemma; it is not specific to the model it was found on. The fix consults the declared capability before applying the keep.

This is the second time the keep has had to be excluded from models it was not designed for: #48042 made it fall back to the host tokens for row-sharded models, whose device buffers it cannot read correctly. That exclusion is a shape check inside the loop, because it depends on the chunk; this one is a property of the model, so it belongs on the entry condition.

This fix is needed to run https://github.com/tenstorrent/tt-inference-server/pull/4837 , which our customers requested.

#### Relates to

* #47359
* #48042
* tenstorrent/tt-inference-server#4837

#### Where to focus

The change is in the shared `Generator`, so every `tt_transformers` model runs the new condition. It is a narrowing: the keep now fires strictly less often, and only for models declaring `supports_async_decode = False`. Models declaring `True` (Llama, Qwen3-32B, Mistral, Phi, GPT-OSS, the Qwen-VL adapters) and any caller that never declares capabilities keep today's behaviour, since the lookup defaults to `True`. The tt-metal demo path is unaffected for the reason given in #47359: it never reaches a reset step with a decode trace already captured.

#### Worth confirming

That `supports_async_decode = False` is the right signal for the three models that declare it, rather than something narrower. The claim is that async-ahead and the keep are the same property, so a model that opts out of one must opt out of the other; if a model wanted the keep without async scheduling it would now silently lose it.

#### LLM/AI Usage

This PR incorporates output generated by my AI agent, which is why the commit author is set to the agent's name.

However, I have thoroughly reviewed all changes (as seen in https://github.com/nyoshifujiTT/tt-metal/pull/9 ) and, as the committer (@nyoshifujiTT), I take full responsibility for this PR.


yieldthought:
Codex review:

**[🟡] Correctness: Treat an undeclared async capability as disabled**

[models/tt_transformers/tt/generator.py:1612](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/models/tt_transformers/tt/generator.py#L1612) defaults `supports_async_decode` to `True`, but the vLLM capability contract says missing keys default to `False` at [tech_reports/LLMs/vLLM_integration.md:43](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/tech_reports/LLMs/vLLM_integration.md#L43). This is not theoretical: `Mistral3ForConditionalGeneration` enables on-device sampling but omits `supports_async_decode` at [models/tt_transformers/tt/generator_vllm.py:413](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/models/tt_transformers/tt/generator_vllm.py#L413). vLLM therefore schedules it synchronously, while this generator still enters the async-ahead keep at reset steps and can substitute the same stale device token this PR is meant to prevent. The undeclared case in [models/tt_transformers/tests/test_async_ahead_token_keep_unit.py:134](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/models/tt_transformers/tests/test_async_ahead_token_keep_unit.py#L134) currently locks in that mismatch.

**Why it matters:** the fix remains incomplete for an in-tree synchronous/on-device-sampling model and makes the shared generator disagree with the scheduler's source of truth.

**Suggested fix:**
Suggestion: use the platform's fail-closed default and update the undeclared-capability test to expect the host token.

```python
supports_async_decode = self.model_capabilities.get("supports_async_decode", False)
```

**[🟡] CI: Keep the new tests within the tier-3 Wormhole budget**

[tests/pipeline_reorg/models_unit_tests.yaml:1547](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/tests/pipeline_reorg/models_unit_tests.yaml#L1547) adds a 10-minute `models/wh_n150` leg, but that team/SKU was already using the full 75-minute `unit_tier3` allocation in [.github/time_budget.yaml:384](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/.github/time_budget.yaml#L384). Running the same verifier used by [.github/workflows/models-unit-tests-impl.yaml:64](https://github.com/tenstorrent/tt-metal/blob/676dbc182b73ea09a23bcff1e58e1ed3faff0aef/.github/workflows/models-unit-tests-impl.yaml#L64) now returns nonzero at `requested=85, budget=75`; a synthetic merge with current `main` produces the same total.

**Why it matters:** every tier-3 unit workflow will fail in `load-test-matrix` before any model test starts.

**Suggested fix:**
Suggestion: because the PR says these two files complete in under four seconds and they use the same WH/BH SKUs, add them to the existing `On-device sampling unit tests` matrix leg instead of reserving another 10-minute leg, while preserving failure propagation between pytest commands. If the separate job is required, update the time budget with the measured-runtime justification.

```yaml
cmd: |
  pytest --timeout 300 models/common/tests/test_sampling.py -k "ttsampling or num_single_device_vocab_splits or untilize_chunk_count or seed or slot_remap or topk_route_mirror_parity" && \
  pytest --timeout 300 models/tt_transformers/tests/test_async_ahead_token_keep_unit.py \
    models/tt_transformers/tests/test_async_ahead_token_keep_decode.py
```

models/tt_transformers/tt/generator.py
```diff
@@ -1609,6 +1609,8 @@ def decode_forward(
         start_pos = torch.chunk(start_pos, self.data_parallel, 0)
         page_table = torch.chunk(page_table, self.data_parallel, 0) if page_table is not None else None
 
+        supports_async_decode = getattr(self, "model_capabilities", {}).get("supports_async_decode", True)
```
yieldthought: Codex: The vLLM platform treats a missing capability as disabled, and `Mistral3ForConditionalGeneration` omits this key while enabling on-device sampling. Match the scheduler contract and use the existing class attribute directly.

```suggestion
        supports_async_decode = self.model_capabilities.get("supports_async_decode", False)
```

models/tt_transformers/tests/test_async_ahead_token_keep_unit.py
```diff
@@ -0,0 +1,164 @@
...
+            {"supports_async_decode": True, "supports_sample_on_device": True},
+            KEPT_TOKENS,
+            KEPT_POSITIONS,
+            id="async_model_keeps_the_device_token",
+        ),
+        pytest.param(
+            None,
+            KEPT_TOKENS,
+            KEPT_POSITIONS,
+            id="undeclared_capability_keeps_the_device_token",
```
yieldthought: Codex: Keep the regression case aligned with the platform's fail-closed default.

```suggestion
            None,
            HOST_TOKENS,
            HOST_POSITIONS,
            id="undeclared_capability_decodes_the_tokens_it_was_given",
```

models/tt_transformers/tt/generator.py
nyoshifujiTT: Yes, you're right. I misunderstood. Fixed.

models/tt_transformers/tests/test_async_ahead_token_keep_unit.py
nyoshifujiTT: Also fixed to follow "default false" settings.

nyoshifujiTT:
@yieldthought Thanks for your review. I confirmed you're right and fixed. Could you check and review again?

mtairum:
Rebased onto latest `main` (`a831f59`) and force-pushed: `393c934` -> `d787c41`. The three merge-from-main commits are gone; the 13 original commits replay linearly and the resulting diff is identical to the pre-rebase one. One conflict, in `tests/pipeline_reorg/models_unit_tests.yaml`: resolved by keeping main's current `-k` filter (`topk_authoritative_route_decision`, which replaced `topk_route_mirror_parity`) and layering your two `pytest` lines on top.

Kicked off model tier CI for the affected models plus the new tests. These run off `mtairum/ci-pr54565`, a mirror of the rebased branch in this repo, because `workflow_dispatch` cannot target a fork branch:

- On-device sampling unit tests, i.e. the new `test_async_ahead_token_keep_{unit,decode}.py` files: https://github.com/tenstorrent/tt-metal/actions/runs/33758757093
- Qwen3.6-27B unit + e2e: https://github.com/tenstorrent/tt-metal/actions/runs/33758761176
- Mistral-Small-3.1-24B unit + e2e: https://github.com/tenstorrent/tt-metal/actions/runs/33758765899

Mistral-Small-3.1 is included because `Mistral3ForConditionalGeneration` omits `supports_async_decode` while enabling on-device sampling, so the fail-closed default now removes the keep for it as well.

Worth noting the new tests had not run anywhere before this: the `On-device sampling unit tests` leg is tier 3 and only reachable from the scheduled/dispatched models-unit pipeline, not from PR gate or sanity.

Once these are green I will approve.

nyoshifujiTT:
@mtairum Thank you for running the CI for me. I checked the test results, and all 3 failures were due to timeouts or dispatch hangs, which are unrelated to the 3-line change in generator.py in this PR (modifying which token to decode with). The new test regression itself passed. The two Qwen3.6 failures happen intermittently on main as well, and Mistral just barely timed out. Could you rerun it, please?

mtairum (approved):
I've re-run the CI, but with the status provided until they reach the timeouts, I'm happy with pre-approving the PR.

mtairum:
Re-kicked the two model tier runs (tiers 1, 2 and 3, unit + e2e) in parallel off `mtairum/ci-pr54565`, which is at this PR's current head `13820c6a726`:

- Qwen3.6-27B unit + e2e: https://github.com/tenstorrent/tt-metal/actions/runs/34487552171
- Mistral-Small-3.1-24B unit + e2e: https://github.com/tenstorrent/tt-metal/actions/runs/34487556920

nyoshifujiTT:
@mtairum Thanks again for the rebase, the CI runs and the approval.

I think this PR is now mergeable_state: clean with your approval on the current head (13820c6a726) and PR Gate green, so I believe everything on the review side is settled. Could you queue it when you get a chance?

For completeness, the state of the last CI round:
- Mistral-Small-3.1-24B unit + e2e: passed (this is the model whose behaviour actually changes under the fail-closed default, so it is the one that mattered most here)
- Qwen3.6-27B unit + e2e: the two bh_quietbox_2 legs timed out again, matching the intermittent failures on main rather than anything in this change
- load-test-matrix / Verify test timeouts against budget: passed, confirming the tier-3 budget fix in this PR

This one is a prerequisite for tenstorrent/tt-inference-server#5046, which a customer is waiting on, so getting it in would unblock the rest of that work.

mtairum:
Added to merge queue 👍

PR #54566, opened 2026-08-27
[Feature] Quasar ATT maps: window math, product map configurations, typed resolution, bring-up replay

vvukomanovicTT:
### PR Category
Feature

### Summary
**Stacked on #54415** (the ATT stack is one linear chain: #54285 → #54415 → this → #54576; this PR's diff is its three commits only). The Quasar ATT maps: window math, the two transcribed product map configurations, the typed resolution layer, and the bring-up register-image replay — all host-only data and `constexpr` code. **Nothing consumes any of it yet**: no kernel or firmware includes these headers until the ATT enablement PR (#54576) selects a map.

Three commits, each reviewable against its own ground truth:

1. **Window math + transcribed configs.** `noc_att::Window` mirrors one ATT mask-table entry with constexpr address math over it (`make_address`, `matches`, selector/local extraction, complete-transfer validation). `mask_bits` uses the hardware/QMK-dump convention — count of **low wildcard bits**; the programming guide's "MSB count" framing is inverted and deliberately not used. Two data-only configs, transcribed row-for-row (never derived by formula): the QSR1 boot map slots tt-metal uses (worker 4, GDDR 5, loopback-scratch 13, per-tile config 14) and the aether 2x3 bring-up map. The aether map only exists because the bring-up replay programs it, so it is ours to shape: its **local window carries QSR1 slot-14 geometry** (compare `0x18_0000_0000`, `translate=1`, BAR 0, 6-bit selector, patched self endpoint) instead of the prototype's pass-through `compare=0` window. That makes every emulator run exercise write-source/read-return/atomic-return roles through a **translating** window — the exact role/window combination QSR1's local operands use, which a pass-through window can never produce — and removes the `compare=0` window, so a leaked raw local address matches no window and faults instead of being silently absorbed. The remote window is correspondingly narrowed (6 selectors, 2^32 span) to keep the windows disjoint — an overlap the disjointness host test caught during development. Each config declares its local (self) window and the derived `LOCAL_WINDOW_BASE` (`0x18_0000_0000` on QSR1 via the boot-patched ep256-self config window; `0x0` on aether) — the constant the enablement will surface as `NOC_ATT_LOCAL_WINDOW_BASE` for the V3 API, static_asserted against the window so they cannot drift. The worker frames transcribe the coordinate frame kernels receive today — the soc descriptors' physical worker positions, passed through unchanged because Quasar NOC translation is the identity placeholder — recorded in the headers so the finalized translated scheme (UMD #2494) is revisited together with these origins.
2. **Typed resolution layer — one resolver over declarative maps.** `noc_att::Address` (local / worker / logical DRAM / dispatch / loopback-scratch + offset) with `encode<MAP>()` as the single resolution+encoding step. A map is pure data: a `MapData` aggregate the config header declares next to its windows and tables (per-role window assignment, local window class, worker frame, inverse endpoint tables, optional DRAM/dispatch bindings — absent means the kind resolves invalid, QSR1's state until descriptor-owned rows exist). One shared resolver interprets every map, so map-specific logic cannot exist and adding a map is adding data; `att_address.h` names no product configuration. Backend helpers for the enablement: `transfer_supported` on a finished operand, `is_self_address`, `extract_local_address`, `resolve_current` initiator identity via the generated inverse endpoint tables, and the worker multicast rectangle with the packed software-descriptor round trip.
3. **Bring-up programming replay + image test.** Production boot/UMD owns ATT programming; the emulator bring-up targets need a one-shot data-driven replay from firmware `noc_init` (hooked up by the enablement, gated by `ATT_PROGRAM_FOR_TEST`, which `#error`s outside firmware builds). The generated Grendel qsr.s1 tile-NIU register image (plus delivery overrides) and the compact aether program are replayed verbatim, never edited by hand.

### Notes for reviewers
- **Tests are host-only** (`unit_tests_noc`, no device). Commit 1 asserts every window field against the QMK-dump values and selector-table uniqueness/coverage/disjointness. Commit 2's goldens are all `static_assert`: exact encoded addresses for every kind on both maps (e.g. `Address::local(0x300000)` on QSR1 → `0x18_0030_0000`; worker (9,5) → `0x100_1F00_1234`), aperture/window transfer bounds, out-of-map rejection, self-detection for every self route, extract/encode inversion. Commit 3 decodes **every checked register write** in the images through the hardware control-word union and compares bit-for-bit with the commit-1 data — mask slots 4/5/13/14 field-by-field (including the GDDR Mimir BAR `0x8_0000_0000`), all 32 worker + 60 full-tile endpoint rows (override-aware), and the replay contract.
- The commit-2 goldens caught one inherited prototype bug, fixed here: the loopback-scratch self check sat below the initiator-identity early-return, so scratch operands were not recognized as self when the identity was unknown.
- This is a re-slicing of the working `vvukomanovic/quasar-att-v3` prototype (DmLoopback passes on emu-quasar-1x3 with this data), with the corrections recorded in the plan: no `embed_local_address` slot-13 relocation, transcribed selector tables, per-map code replaced by the declarative `MapData` + shared resolver.
- All new headers are in the `jit_api` install manifest (`sources.cmake`).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:vvukomanovic/att-pr3-att-maps)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:vvukomanovic/att-pr3-att-maps)


tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,379 @@
...
+inline constexpr ResolvedTile INVALID_TILE{false, WindowClass::None, 0, 0, 0};
+
+/// The declarative description of one product map. Every field is transcribed
+/// configuration data; the resolver below is the only code that interprets it.
+struct MapData {
+    std::uint64_t local_window_base;
+
+    /// Window per role, indexed by WindowClass. None slots stay nullptr; a
+    /// nullptr for a role makes every identity of that role resolve invalid.
+    const Window* windows[WINDOW_CLASS_COUNT];
```
tt-vjovanovic: Consider using `std::array<type, WINDOW_CLASS_COUNT> windows;` so that users can directly use `.windows.size()` instead of knowing it is `WINDOW_CLASS_COUNT` or using `sizeof(windows)/sizeof(windows[0])`.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+        return {Kind::LoopbackScratch, Id{}, absolute_l1_address};
+    }
+
+    constexpr Kind kind() const { return kind_; }
+    constexpr std::uint64_t offset() const { return offset_; }
+
+    // Per-kind identity accessors; resolution reads the field matching kind().
+    constexpr std::uint32_t x() const { return id_.xy.x; }
+    constexpr std::uint32_t y() const { return id_.xy.y; }
+    constexpr std::uint32_t bank() const { return id_.bank; }
```
tt-vjovanovic: If kind decides whether one should read x/y or kind, you should add ASSERT to it, so that running tests can catch wrong access.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+        constexpr Id() : xy{0, 0} {}
+        constexpr Id(std::uint16_t x, std::uint16_t y) : xy{x, y} {}
+        constexpr explicit Id(std::uint16_t logical_bank) : bank(logical_bank) {}
+    };
+
+    constexpr Address(Kind kind, Id id, std::uint64_t offset) : kind_(kind), id_(id), offset_(offset) {}
+
+    Kind kind_;
+    Id id_;
+    std::uint64_t offset_;
```
tt-vjovanovic: I'm not sure how optimal data layout is in this structure.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att.h
```diff
@@ -0,0 +1,109 @@
...
+
+/// @brief Extract the local-address field from an address in this window.
+constexpr std::uint64_t local_address(const Window& window, std::uint64_t address) {
+    return address & low_mask(local_address_bits(window));
+}
+
+/// @brief Whether @p address falls inside the window's compare range.
+constexpr bool matches(const Window& window, std::uint64_t address) {
+    return (address & ~low_mask(window.mask_bits)) == window.compare;
+}
```
tt-vjovanovic: Consider moving these functions inside the `Window` struct. Then you can use implicit `this` and they can still stay `constexpr`.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+    std::uint32_t extent_xy;
+    std::uint32_t rectangle_count;
+};
+
+/// The roles a window can play in a map. A map assigns a Window to each role it
+/// supports (several roles may share one Window, as on maps with a single
+/// remote window).
+enum class WindowClass : std::uint8_t { None = 0, LoopbackScratch, Worker, Dram, FullTile };
+
+inline constexpr std::size_t WINDOW_CLASS_COUNT = 5;
```
tt-vjovanovic: You can write something like this:
```
enum class WindowClass : std::uint8_t {
   LoopbackScratch = 0,
   Worker,
   Dram,
   FullTile,

   // Add values before this line. Count represent number of valid values.
   Count,
   Invalid = Count   // Invalid replaced None, figure out if you actually need None entry in arrays or not.
};
```
Then you don't need to maintain `WINDOW_CLASS_COUNT`, but let it be automatically maintained by items inside the `WindowClass`. You can define it as:
```
inline constexpr std::size_t WINDOW_CLASS_COUNT = static_cast<std::size_t>(WindowClass::Count);
```

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+ * the resolver once per product map.
+ */
+namespace noc_att {
+
+using NocAddress = std::uint64_t;
+
+struct AddressResult {
+    bool valid;
+    NocAddress address;
+};
```
tt-vjovanovic: This sounds like `std::optional<NocAddress>`. Maybe you should prefer that if you aren't able to say some value is not valid (like 0 or -1).

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+inline constexpr std::size_t WINDOW_CLASS_COUNT = 5;
+
+/// A resolved ATT identity: which window and selector reach a tile.
+struct ResolvedTile {
+    bool valid;
+    WindowClass window;
+    std::uint32_t selector;
+    std::uint32_t noc_x;
+    std::uint32_t noc_y;
+};
```
tt-vjovanovic: Depending on alignment requirement, this structure might be better "compressed" with a bit different order of data:
```
struct ResolvedTile {
    std::uint32_t selector; // 4 bytes
    std::uint32_t noc_x; // 4 bytes
    std::uint32_t noc_y; // 4 bytes
    bool valid;   // 1 byte
    WindowClass window; // 1 byte
};
```

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+    std::uint32_t count = 0;
+
+    constexpr Table() = default;
+    template <std::size_t N>
+    constexpr Table(const T (&array)[N]) : data(array), count(N) {}
+
+    constexpr const T& operator[](std::uint32_t index) const { return data[index]; }
+    constexpr std::uint32_t size() const { return count; }
+    constexpr bool empty() const { return count == 0; }
+};
```
tt-vjovanovic, line 89: This sounds like `std::span<>`. What standard are we using? `std::span<>` is introduced in C++20

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,408 @@
...
+    template <const MapData& Map>
+    constexpr AddressResult encode(std::uint64_t size = 1) const;
+
+private:
+    union Id {
+        struct Xy {
+            std::uint16_t x;
+            std::uint16_t y;
+        } xy;
+        std::uint16_t bank;
```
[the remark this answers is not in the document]
vvukomanovicTT: fixed - identities that do not fit 16 bits clamp to 0xFFFF, which no map can resolve (endpoint coordinates and bank counts are far smaller), so an oversized identity is rejected at resolution instead of wrapping onto a valid one. added static_asserts covering worker/dram/dispatch overflow on both maps.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/configs/grendel_qsr1_att_config.h
```diff
@@ -0,0 +1,134 @@
...
+ *
+ * Data only. Address resolution over this data lives in noc/att/att_address.h.
+ */
+namespace grendel_qsr1_att_config {
+
+// The worker frame below is the coordinate frame kernels receive today: the
+// soc descriptor's physical functional_workers positions (quasar_32_arch.yaml,
+// workers 2-2..9-5), passed through unchanged because Quasar NOC translation
+// is the identity placeholder (UMD #2494). Revisit together with #2494 if the
+// finalized translated scheme moves the worker frame.
```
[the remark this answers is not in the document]
vvukomanovicTT: fixed, the comments now reference tenstorrent/tt-umd#2494.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att.h
```diff
@@ -0,0 +1,109 @@
...
+ * The names match the ATT programming model so a generated configuration can
+ * be checked directly against its source data.
+ *
+ * mask_id is deliberately absent: it selects a hardware register slot but does
+ * not change an address. Register addresses, mask IDs, and endpoint coordinates
+ * belong to boot-owned programming data.
+ */
+struct Window {
+    /// Window base: the address bits above the wildcard region.
+    std::uint64_t compare;
```
tt-vjovanovic: Maybe better name would be `base` or `base_address`.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
```diff
@@ -0,0 +1,411 @@
...
+/// @brief The declarative description of one product map. Every field is
+/// transcribed configuration data; the resolver below is the only code that
+/// interprets it.
+///
+/// Each configuration header declares exactly one inline constexpr instance,
+/// consumed only as a `const MapData&` template argument - everything
+/// constant-folds, so the instance is never odr-used at runtime and
+/// contributes no storage to the device binary.
+struct MapData {
+    std::uint64_t local_window_base;
```
tt-vjovanovic: Does this exist because we don't have `Local` in `WindowsClass`?

tt-vjovanovic: Or is this value that should be equal to `windows[(uint32_t)local_window_class].compare`?

tests/tt_metal/tt_metal/noc/test_quasar_att_image.cpp
```diff
@@ -0,0 +1,186 @@
...
+
+#include <cstdint>
+#include <optional>
+
+#include "internal/tt-2xx/quasar/noc/att/temporary_programming/grendel_qsr1_att_data.h"
+#include "internal/tt-2xx/quasar/noc/att/temporary_programming/quasar_aether_2x3_att_data.h"
+
+namespace {
+
+using noc_att::RegisterWrite;
```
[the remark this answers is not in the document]
vvukomanovicTT: removed.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att.h
vvukomanovicTT: This is the name ATT in hardware defines, I would prefer to have that matching because we cant change it in HW repo atm.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: the union is gone since the layout rework - fields coexist and unused ones are zero, so there is no active-member hazard left to assert on. can add kind checks on the accessors if you still want them after seeing the new layout.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: reworked per the slack thread - members sorted by size, no union, and a static_assert pins the struct at 16 bytes.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att.h
vvukomanovicTT: done - the window math is member functions on Window now (still constexpr), low_mask stays free since it is not window-specific.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: adopted, including dropping the none slot - WINDOW_CLASS_COUNT derives from Count, the windows array holds only the four real roles, and Invalid = Count marks no-role in ResolvedTile.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: switched to std::optional<NocAddress> - it is c++17 so the device jit builds take it, and there is no honest sentinel since 0 is a structurally valid operand. compiles clean on the riscv toolchain and everything stays constexpr.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: reordered largest-first and pinned with static_assert(sizeof(ResolvedTile) == 16).

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: device jit builds are c++17, so no std::span - Table is the minimal c++17 span stand-in with the count deduced from the array (noted in its comment now). happy to swap it for std::span the day device moves to c++20.

tt_metal/hw/inc/internal/tt-2xx/quasar/noc/att/att_address.h
vvukomanovicTT: half and half - it is the local window's make_address(selector 0, local 0), which happens to equal that window's compare on both current maps but means 'the selector-0 base'. it also has to exist as a preprocessor macro because the v3 header #errors on it before any type exists. dropped the stored field though: local_window_base(map) is derived from the map now, so nothing can drift.

kstevensTT (approved):

arikTT (approved):

jbaumanTT (approved):

PR #54568, opened 2026-08-27
[Feature] Report AI job and run summaries for the L2 nightly pipeline

ppetrovicTT:
### PR Category
Feature

### Summary

Turns on AI job + run summaries for the L2 nightly pipeline, and stops the reports
calling passing jobs crashed.

- Every job now runs under `run-with-log`, so a `timeout-minutes` kill is
  distinguishable from a real failure instead of reporting as success.
- Per-job summaries on all 9 impls; one `AI Run Summary` job aggregates them.
  Jobs that produced nothing are reported as infra failures rather than dropped.
- Tests that deliberately provoke a device error now declare it, so log triage
  stops reading them as crashes. **33 of 160 green legs were affected per run; now 0.**
- Several jobs that can't declare it (C++ negative tests, a provoked device hang, some ops tests) opt
  out via an `ai_summary` section in their test yaml.

Full run showing the output: https://github.com/tenstorrent/tt-metal/actions/runs/33167878929

### Notes for reviewers

- `prepare_test_matrix.py` flattens the new `ai_summary` yaml section into a flat
  matrix key, so workflows read one plain value.
- Expected-error patterns must come from the message the logger emits, not the
  assert condition and not across a `{}` — those match the exception but never the
  log line, so the suppression silently does nothing.
- `matmul` uses the opt-out rather than a test fix: it logs a fatal from a program
  config it then successfully retries. That's a question for the matmul owners.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ppetrovic/l2-nightly-summary)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ppetrovic/l2-nightly-summary)

### DIFF
--- a/.github/scripts/utils/prepare_test_matrix.py
+++ b/.github/scripts/utils/prepare_test_matrix.py
@@ -285,6 +285,11 @@ def build_test_matrix(tests, enabled_skus, sku_config, event=None, allow_missing
             for key, value in sku_test_config.items():
                 if key != "timeout" and value is not None:
                     entry[key] = value
+            # ai_summary is the authoring surface in the tests yaml; flatten it to the
+            # keys the impls read so every leg carries them and no workflow has to guard
+            # on the section being absent.
+            ai_summary = entry.pop("ai_summary", None) or {}
+            entry["authoritative_job_status"] = bool(ai_summary.get("authoritative_job_status", False))
             sku_entry = sku_config[concrete_sku]
             if "weights-cache-mode" in sku_entry:
                 entry["weights-cache-mode"] = sku_entry["weights-cache-mode"]
--- a/.github/scripts/utils/prepare_test_matrix.py
+++ b/.github/scripts/utils/prepare_test_matrix.py
@@ -285,9 +285,10 @@ def build_test_matrix(tests, enabled_skus, sku_config, event=None, allow_missing
             for key, value in sku_test_config.items():
                 if key != "timeout" and value is not None:
                     entry[key] = value
-            # ai_summary is the authoring surface in the tests yaml; flatten it to the
-            # keys the impls read so every leg carries them and no workflow has to guard
-            # on the section being absent.
+            # ai_summary is the authoring surface in the tests yaml; flatten it so an impl
+            # reads a plain key instead of guarding on the section. Only the impls that
+            # forward authoritative_job_status act on it -- setting it on a test under any
+            # other impl is computed here and then dropped.
             ai_summary = entry.pop("ai_summary", None) or {}
             entry["authoritative_job_status"] = bool(ai_summary.get("authoritative_job_status", False))
             sku_entry = sku_config[concrete_sku]
--- a/.github/workflows/tt-train-tests.yaml
+++ b/.github/workflows/tt-train-tests.yaml
@@ -280,7 +280,7 @@ jobs:
               "input_dirs": ["generated/test_logs"],
               "output_dir": "generated/ai_summaries",
               "scope": "${{ inputs.summary-scope }}",
-              "authoritative_job_status": ${{ matrix.test-group.authoritative_job_status || false }}
+              "authoritative_job_status": ${{ (matrix.test-group.ai_summary && matrix.test-group.ai_summary.authoritative_job_status) || false }}
             }
           api-key: ${{ secrets.TT_CHAT_API_KEY }}
           api-url: ${{ secrets.TT_CHAT_URL }}
--- a/.github/workflows/tt-train-tests.yaml
+++ b/.github/workflows/tt-train-tests.yaml
@@ -63,6 +63,15 @@ on:
           Max allowed timeout (minutes) for any individual test SKU entry in the
           tests yaml. Verified during load-test-matrix via verify_time_budget.py.
           Empty skips the per-test ceiling check.
+      summary-scope:
+        required: false
+        type: string
+        default: ""
+        description: "Discriminator for the AI summaries when the calling workflow is invoked more than once per run"
+    outputs:
+      matrix:
+        description: "Filtered matrix of legs that ran (consumed by ai_summary/run for INFRA_FAILURE reconciliation)"
+        value: ${{ jobs.load-test-matrix.outputs.matrix }}
 
 permissions:
   contents: read
@@ -190,11 +199,14 @@ jobs:
 
       - name: ${{ matrix.test-group.name }}
         timeout-minutes: ${{ inputs.timeout-minutes-override > 0 && inputs.timeout-minutes-override || matrix.test-group.timeout }}
-        run: |
-          python3 tools/tt-power-sidecar/tt_power_sidecar.py \
-            --backend sysfs \
-            --out /work/power_report.json \
-            -- bash -c '${{ matrix.test-group.cmd }}'
+        uses: ./docker-job/.github/actions/run-with-log
+        with:
+          log-file: generated/test_logs/${{ matrix.test-group.name }}.log
+          run: |
+            python3 tools/tt-power-sidecar/tt_power_sidecar.py \
+              --backend sysfs \
+              --out /work/power_report.json \
+              -- bash -c '${{ matrix.test-group.cmd }}'
 
       - name: Upload loss plot
         id: upload-loss-plot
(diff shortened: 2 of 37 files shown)


.github/workflows/ttsim-unit-tests-impl.yaml
```diff
@@ -150,12 +159,31 @@ jobs:
...
+            ${{ matrix.test-group.cmd }}
       - uses: tenstorrent/tt-metal/.github/actions/upload-artifact-with-job-uuid@main
         timeout-minutes: 10
         if: ${{ !cancelled() }}
         with:
           prefix: "test_reports_"
+      - name: 🤖 AI job summary
+        if: ${{ !cancelled() }}
+        continue-on-error: true
+        uses: tenstorrent/tt-github-actions/.github/actions/ai_summary/job@main
```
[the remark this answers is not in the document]
ppetrovicTT: Not pinning: this action is first-party (`tenstorrent/tt-github-actions`) and every other consumer in this repo tracks `@main`. Pinning only these ten call sites would leave the pipeline on a stale tool while the rest float, and the action is still changing. The same applies to the nine sibling comments.

tests/ttnn/nightly/unit_tests/operations/transformers/test_paged_update_cache.py
```diff
@@ -1378,5 +1378,5 @@ def test_paged_fill_cache_batched_rejects_non_row_major_batch_idx_tensor(device)
     except (RuntimeError, ValueError):
         pytest.skip("Cannot construct a TILE_LAYOUT uint32 tensor on this build; layout assertion still in place.")
 
-    with pytest.raises(RuntimeError):
+    with expect_error(RuntimeError, "Batch idx tensor must have input_tensor batch dim"):
```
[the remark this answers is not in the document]
ppetrovicTT: Correct, and the pin is what makes it visible: the count check at `paged_fill_cache_device_operation.cpp:126` fires before the ROW_MAJOR check at :139, because tilizing pads the physical volume to 1024 against `input_batch` 4. So this test duplicates `test_paged_fill_cache_batched_rejects_mismatched_batch_idx_tensor` and the layout guard has no coverage.

Pre-existing gap, not introduced here, and out of scope for this PR — it needs a tensor whose padded volume is still 4 in a non-ROW_MAJOR layout, which is a change for the op owners. Filing separately rather than widening this change.

tests/ttnn/unit_tests/operations/data_movement/test_pad_subcoregrids.py
```diff
@@ -300,7 +300,7 @@ def test_pad_subcoregrids_rejects_sharded(device):
     input_tensor = ttnn.to_memory_config(input_tensor, sharded_mem)
 
     sub_core_grid = ttnn.CoreRangeSet([ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(3, 3))])
-    with pytest.raises(RuntimeError):
+    with expect_error(RuntimeError, "must not exceed number of cores"):
```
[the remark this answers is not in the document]
ppetrovicTT: Agreed and confirmed: the error comes from `tt_metal/impl/tensor/spec/tensor_spec.cpp:47` while building the padded output spec, so pad's own guard at `pad_device_operation.cpp:174` is never reached. The sibling `test_pad_subcoregrids_rejects_singlecore` does hit pad's guard, so the pair is inconsistent.

Pre-existing gap that the pinned message exposed rather than caused. Leaving it to the data-movement owners rather than reshaping the test here.

.github/scripts/utils/prepare_test_matrix.py
```diff
@@ -286,6 +286,11 @@ def build_test_matrix(tests, enabled_skus, sku_config, event=None, allow_missing
             for key, value in sku_test_config.items():
                 if key != "timeout" and value is not None:
                     entry[key] = value
+            # ai_summary is the authoring surface in the tests yaml; flatten it to the
+            # keys the impls read so every leg carries them and no workflow has to guard
+            # on the section being absent.
+            ai_summary = entry.pop("ai_summary", None) or {}
+            entry["authoritative_job_status"] = bool(ai_summary.get("authoritative_job_status", False))
```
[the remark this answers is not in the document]
ppetrovicTT: Fair — the comment overclaimed. Narrowed it to say only the impls that forward `authoritative_job_status` act on the flag, and that setting it elsewhere is computed and then dropped.

Not wiring all nine deliberately: the flag disables crash detection on a green job, so requiring a workflow edit to enable it is the point rather than friction. Five test entries use it today, all under impls that forward it.

vtsilytskyiTT:
`s/leg/log/g` in PR description, please :)

vtsilytskyiTT (approved):
ttnn-core part approved.

cglagovichTT (approved):

tests/ttnn/unit_tests/base_functionality/test_torch_conversion.py
```diff
@@ -1187,18 +1185,22 @@ def test_from_torch_sharded_tilize_dispatch_core_overlap(device):
...
+            layout=ttnn.TILE_LAYOUT,
+            device=device,
+            memory_config=sharded_mem_config,
+            mesh_mapper=ttnn.ReplicateTensorToMesh(device),
+        )
 
-    assert result.layout == ttnn.TILE_LAYOUT
-    assert result.dtype == ttnn.bfloat16
-    assert list(result.shape) == list(shape)
+    # assert result.layout == ttnn.TILE_LAYOUT
```
bbradelTT, line 1201: Please remove commented out code. Alternatively you could add a comment about why this is kept here.

bbradelTT (approved):

tests/ttnn/unit_tests/base_functionality/test_torch_conversion.py
ppetrovicTT: i thought the comment one call above was enough?

rdjogoTT (approved):

tdowdallTT (approved):

tdowdallTT:
Excited to see AI run summary extended to l2, thank you for this feature!

KalaivaniMCW (approved):
Approved for eltwise test files.

.github/workflows/tt-train-tests.yaml
```diff
@@ -63,6 +63,15 @@ on:
           Max allowed timeout (minutes) for any individual test SKU entry in the
           tests yaml. Verified during load-test-matrix via verify_time_budget.py.
           Empty skips the per-test ceiling check.
+      summary-scope:
```
Aswinmcw, line 66: This input is added here but no caller sets it, and `tt-metal-l2-nightly.yaml` calls this workflow three times in one run (`tt-train-unit-tests`, `tt-train-perf-pretraining-tests`, `tt-train-perf-finetuning-tests`) — the exact condition the description names. All three get `scope: ""`.

It's safe today only because the three matrices happen to be disjoint: unit reads `train_unit_tests.yaml`, and the two perf calls partition `train_perf_tests.yaml` by `category` (10 pretraining / 2 finetuning, no leg without a category). Add one leg with the same `name` under both categories, or one shared with the unit yaml, and two jobs upload a summary under the same `job-name` + empty scope, while `EXPECTED` in `ai-run-summary` double-counts it — `add` with no dedupe.

Passing `summary-scope: unit` / `pretraining` / `finetuning` at the three call sites removes the dependency on that invariant, and matches how `sanity-tests.yaml` scopes its impls.

Aswinmcw (approved):
Eltwise file is correct — `negative exponents are not supported` matches the `TT_FATAL` in `unary_backward.cpp:254` verbatim. One suggestion below on the tt-train call sites; not a blocker.

.github/workflows/tt-train-tests.yaml
ppetrovicTT: Great catch! Thanks!

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
```diff
@@ -21,16 +21,16 @@
...
 
 def _argmax_misc_and_rank_special():
     return [
         _case([], RM, None, True, torch.bfloat16),
         _case([32], RM, -1, False, torch.float32),
-        _case([32, 0], RM, 1, True, torch.bfloat16),
+        _case([32, 0], RM, 1, True, torch.bfloat16, "Expected reduction dim 1 to have non-zero size"),
         _case([64], RM, -1, True, torch.bfloat16),
         _case([1, 512], RM, -1, True, torch.float32),
         _case([1, 1024], RM, -1, True, torch.int32),
```
jvegaTT, line 36: Shouldn't we put None on the other cases?

jvegaTT (approved):
Small nit, approving data movement changes to unlock

tests/ttnn/unit_tests/operations/reduce/test_argmax.py
ppetrovicTT: i didn't want to pollute all the lines
it's handled by a default parameter in `_case` function on line 24

PR #51556, opened 2026-07-30
[Performance] Qwen3.6-27B: Added bucketing fix and TP=8 support

atupe-tt:
### Summary

- ~~Added async scheduling~~ - removed since it is dependent on vLLM PRs that are not yet merged
- Adds decode batch bucketing for Qwen3.6-27B. The model uses the smallest suitable batch size, so a single request stays fast even when the server supports high concurrency.
- Compiles all batch sizes before trace capture and safely keeps separate traces for each size.
- Preserves shared GDN and device state when the batch size changes.
- Fixes unseeded on-device sampling, which could repeat the same random result.
- Adds TP=8 support, including KV-head replication and TP-specific prefill tuning.
- Adds regression tests and CI coverage.
- Async scheduling is deferred until the required vLLM changes are ready.

### Changes to common files

- `models/common/sampling/generator.py`: Adds optional batch-size keys for sampling traces. Models that do not use bucketing keep the existing behavior.
- `models/tt_transformers/tt/generator.py`: Adds optional two-stage trace setup, correct seed initialization, and cleanup for all bucket traces. Buffer reuse is opt-in for Qwen.
- `models/common/warmup/warmup_utils.py`: Lets Qwen compile all batch sizes before trace capture. The default path is unchanged.
- `models/tt_transformers/tt/model_config.py`: Adds opt-in KV-head replication for TP configurations with more devices than KV heads. Other models still use the existing validation.
- `models/demos/utils/device_sku.py`: Adds the P150×8 LoudBox SKU mapping.
- `.github/workflows/vllm-model-tests-impl.yaml`: Disables thinking in the coherence check and increases its output limit.
- Common sampling tests verify that bucketed traces stay isolated without changing other models.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:atupe/qwen36-bucketing-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:atupe/qwen36-bucketing-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-sanity-tests.yaml?query=branch:atupe/qwen36-bucketing-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:atupe/qwen36-bucketing-fix)
[...]

PR #48861, opened 2026-07-02
[Feature] Optimizations for Qwen3.6-27B

atupe-tt:
### Summary
Added optimizations to Qwen3.6-27B - including a kernel similar to `chunked_gated_delta_rule`

Here are the perf numbers with the new kernel - 
image

Old numbers for reference - 
image

### CI Status
[vLLM Nightly](https://github.com/tenstorrent/tt-metal/actions/runs/29600676589)
[(Tier 1) Models Unit Tests](https://github.com/tenstorrent/tt-metal/actions/runs/29600697000)
[(Tier 1) Models E2E](https://github.com/tenstorrent/tt-metal/actions/runs/29600729369)

PR #47738, opened 2026-06-22
Feature: Qwen 3.5 & 3.6 bringup on BH

atupe-tt:
### Summary
Bringing up of Qwen3.5-9B, 27B and Qwen3.6-27B models on BH (9B on a single P150 and 27B on P150x4)

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:qwen9b-p150)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:qwen9b-p150)

PR #54572, opened 2026-08-27
Bring up Qwen3.5-9B on Wormhole N300 and Qwen3.5/3.6-27B on T3K

ign-christy:
## Summary
Enables the Qwen3.5-9B and Qwen3.5/3.6-27B hybrid Gated-DeltaNet + Gated-Full-Attention model on Wormhole N300
(`(1,2)` mesh, TP=2) and Wormhole T3K(`(1,8)` mesh, TP=8), retunes the prefill/decode/vision hot paths for that mesh, adds a full-depth HF-parity test suite, fixes a permuted-RoPE weight-cache staleness gap, and lands
tier-3 CI coverage.

Refer `models/demos/blackhole/qwen36/README-N300-9B.md` and `models/demos/blackhole/qwen36/README-T3K-27B.md` for more information.

### Enablement

- Wormhole-gated the Blackhole-only fusions and relieved L1 pressure to get Qwen3.5-9B running on
  Wormhole at all, with a model-local WH compat layer for GDN L1 OOMs.
- Moved RoPE and greedy sampling on-device for trace-safe decode; allowed `text_demo` to target
  Wormhole B0 alongside Blackhole; limited device RoPE / on-device argmax to Wormhole.
- Fixed Wormhole TP attention KV decode corruption at B=32 and N300 TP batched-decode / GDN
  fused-batch-vs-compute-grid issues; stabilized the batched TP tests.

### Performance — Gated DeltaNet (GDN)

- Prefill: avoided redundant layout relayouts, moved FIR taps + in-proj width padding to L1,
  switched to native `conv1d` with tuned prefill CCL, and a one-K-pass matmul + L1 placement pass.
- Decode: bf16 recurrence with a local delta-rule path, TILE-native GQA expand with an a/b split,
  a fused KV-cache write with a gated sigmoid-multiply, and TILE optimizations gated to 9B only.
- A single-layer prefill profiler was added alongside these to drive the tuning.

### Performance — Gated full attention

- Prefill: BFP8 SDPA with tuned QKV/`wo` matmuls.
- Decode: permuted full-width RoPE (folds the head_dim channel permutation into the weights,
  collapsing the partial-rope slice/transpose/concat chain into one call) with a post-concat gate;
  RoPE/prefill constants moved off host onto `ttnn`; the traced chunked-prefill path fixed to
  gather RoPE inside the trace.

### Performance — MLP and vision tower

- MLP: two optimization passes, scoped to `bfloat8_b` for the 9B/N300 gate/up projections.
- Vision tower enabled on Wormhole, then tuned: SDPA + redundant-op removal + row padding,
  per-family matmul program configs, tuned CCL workers on N300, host-tilized RoPE/bilinear
  weights, on-device greedy argmax for the TP decode path, and a residual-precision fix with a
  tightened PCC gate. Full writeup in `VISION_TOWER_PERF.md`.
- All of the above narrowed to `wh_9b_n300` (or otherwise scoped) so Blackhole/other-mesh paths
  are unaffected.

### Testing

- `tests/unit/test_prefill.py` / `test_decode.py`: full-depth (all 32 layers, real checkpoint)
  logits PCC vs `Qwen3_5ForCausalLM` — the primary correctness gate for a full-model change,
  sharing `tests/unit/full_depth_pcc_common.py`.
- RoPE unit tests added alongside the Wormhole device-RoPE change; test oracles scoped to
  Wormhole where the numerics differ from Blackhole.
- Permuted-RoPE weight cache (`tt/attention/tp.py::load_attention_weights_tp`) now self-invalidates:
  the `.rp` cache tag is a content hash of the pre-permutation source weight plus a version marker
  (`rope_tp.ROPE_PERM_VERSION`), so a stale cache built by an older permutation construction can no
  longer be silently reused by a full-model run — previously only `test_attention_tp.py` (which
  forces a cold cache) could catch a regression there.

### Measured — N300, Qwen3.5-9B

| Test | Gate | Measured |
| --- | --- | --- |
| Full-depth prefill logits PCC (32 layers, 128-token prompt) | 0.98 | 0.998400 |
| Full-depth decode logits PCC (32 layers, 5 steps) | 0.95 | 0.994791 (mean) |
| Vision tower, depth 27 (real checkpoint) | 0.985 | 0.98850 → 0.99929 (after sequence-padding fix) |

Text generation (all 8 traced ISLs, batch 1, greedy):

| ISL | TTFT | Decode | ms/tok |
| --- | --- | --- | --- |
| 128 | 0.24 s | 22.84 tok/s | 43.8 |
| 4k | 0.95 s | 22.58 tok/s | 44.3 |
| 32k | 8.33 s | 22.71 tok/s | 44.0 |
| 128k | 33.50 s | 20.76 tok/s | 48.2 |
| 256k | 125.43 s | 18.15 tok/s | 55.1 |

Vision-tower kernel tuning (full detail in `VISION_TOWER_PERF.md`):

| Stage | Before | After (N300) |
| --- | --- | --- |
| Depth-1 window (one block, device time) | 11,876 µs | 6,887 µs (1.72x) |
| Full tower device time | 52,477 µs | 45,343 µs (−13.6%) |
| SDPA (8 heads/device, 256/512) | 20.62 ms | 15.29 ms (1.35x) |
| CCL workers (`wpl=4` vs `wpl=2`) | 5.77 ms | 4.69 ms (−19%) |

### Measured — T3K, Qwen3.6-27B

| Test | Gate | Measured (T3K) |
| --- | --- | --- |
| `test_full_depth_prefill_logits_pcc` (64 layers, 128-token prompt) | 0.98 | 0.995684 |
| `test_full_depth_decode_logits_pcc` (64 layers, 5 steps) | 0.95 | 0.993884 (mean), 0.990054 (worst step) |
| `test_teacher_forcing_e2e` top-1 / top-5 vs HF (128 steps) | 0.87 / 0.93 | **93.02% / 97.67%** |
| `test_teacher_forcing_logits_pcc` (128 decode steps) | 0.75 | 0.967950 (mean), 0.876796 (worst step) |

Text generation:
| ISL | Gen tokens | TTFT | Decode | ms/tok |
| --- | --- | --- | --- | --- |
| 128 | 50 | 0.73 s | 17.87 tok/s | 56.0 |
| 4k | 100 | 1.31 s | 17.67 tok/s | 56.6 |
| 8k | 100 | 1.97 s | 17.69 tok/s | 56.5 |
| 16k | 100 | 4.07 s | 17.32 tok/s | 57.7 |
| 32k | 100 | 8.83 s | 16.99 tok/s | 58.9 |
| 64k | 500 | 20.64 s | 16.05 tok/s | 62.3 |
| 128k | 100 | 38.68 s | 14.86 tok/s | 67.3 |
| 256k | 100 | 156.97 s | 11.79 tok/s | 84.8 |

### CI

Eight pipeline entries in total — three for the 9B, five for the 27B. Every one is reachable by
name from `workflow_dispatch` (`model:` column), not only under `model: all`.

| Entry | Tier | SKU | Budget | `workflow_dispatch` model |
| --- | --- | --- | --- | --- |
| `Qwen3.5-9B unit tests` | 3 | `wh_n300` | 30 | `qwen3.5-9b` |
| `Qwen3.5-9B e2e tests` (`release_ready: false`) | 3 | `wh_n300` | 20 | `qwen3.5-9b` |
| `Qwen3.5-9B P150 unit tests` | 3 | `bh_p150` | 25 | `qwen3.5-9b` |
| `Qwen3.6 unit tests` (extended: `test_rope_tp.py`, `test_generate_tp.py`) | 1 | `bh_quietbox_2` | 28→36 | `qwen3.6-27b` |
| `Qwen3.6-27B e2e tests` (extended: HF cache seed step) | 1 | `bh_quietbox_2` | 45 | `qwen3.6-27b` |
| `Qwen3.6 full-depth PCC tests` | 2 | `bh_quietbox_2` | 40 | `qwen3.6-27b` |
| `Qwen3.6-27B long-context e2e tests` (`release_ready: false`) | 2 | `bh_quietbox_2` | 40 | `qwen3.6-27b` |
| `Qwen3.6-27B teacher-forced accuracy` (`release_ready: false`) | 3 | `bh_quietbox_2` | 70 | `qwen3.6-27b` |

The 9B legs cover TP contract + vision component tests and `text_demo`/`vision_demo` smoke cases.
The 27B legs at tiers 2-3 are the full-depth HF-parity gate (`tests/unit/test_prefill.py` /
`test_decode.py`), the long-context ISL ladder past the first traced prefill chunk, and the
teacher-forced accuracy regression detector — all three split off the tier-1 entry so a failure is
attributable and does not gate merges.

**`workflow_dispatch` registration.** `qwen3.5-9b` on the tier-3 pair; `qwen3.6-27b` on the tier-1
pair (already present on `main`) plus the three workflows this PR adds 27B entries to —
`models-t2-unit-tests.yaml`, `models-t2-e2e-tests.yaml`, `models-t3-e2e-tests.yaml`. Without that
last group the three new 27B entries would only fire under `model: all`, which fans out every model
on the tier at shared `bh_quietbox_2` capacity.

**`time_budget.yaml` — seven entries.** Five are `bh_quietbox_2` raises for the cherry-picked 27B
entries and are annotated `needs #tt-metal-infra sign-off`; `e2e_tier3.bh_quietbox_2` nearly triples.
The two `wh_n300` caps are measured, not placeholders (26 min / 16 min on
[unit #33774111206](https://github.com/tenstorrent/tt-metal/actions/runs/33774111206) and
[e2e #33774115934](https://github.com/tenstorrent/tt-metal/actions/runs/33774115934)).

| Key | Before | After | Driver |
| --- | --- | --- | --- |
| `models.unit_tier1.bh_quietbox_2` | 390 | 400 | `Qwen3.6 unit tests` +8 |
| `models.unit_tier2.bh_quietbox_2` | 60 | 95 | `Qwen3.6 full-depth PCC tests` (40) |
| `models.unit_tier3.wh_n300` | 33 | 53 | `Qwen3.5-9B unit tests` (30) |
| `models.unit_tier3.bh_p150` | 30 | 50 | `Qwen3.5-9B P150 unit tests` (25) |
| `models.e2e_tier2.bh_quietbox_2` | 186 | 220 | `Qwen3.6-27B long-context e2e` (40) |
| `models.e2e_tier3.wh_n300` | 60 | 70 | `Qwen3.5-9B e2e tests` (20) |
| `models.e2e_tier3.bh_quietbox_2` | 42 | 120 | `Qwen3.6-27B teacher-forced accuracy` (70) |

Existing Gemma-2-2B/9B + Qwen2.5-7B tier-3 jobs already consumed most of the prior `wh_n300` caps.
`verify_time_budget.py` passes on unit and e2e tiers 1-3.

**Known gaps.**
- **Cold HF cache needs a writable mount.** Four entries run `HF_HUB_OFFLINE=0 … snapshot_download`
  to seed the shared `HF_HOME`, which needs `mlperf-read-only=false`. Scheduled runs use the default
  `:ro` `/mnt/MLPerf`, so a cold cache there fails rather than seeds. Not yet resolved.
- **Outstanding manual step:** `test-command.lock.yml` / `silencer.lock.yml` are gh-aw–compiled
  locks that need `gh aw compile` (unavailable here) to pick up `qwen3.5-9b` for `/test`-comment
  dispatch; scheduled and `workflow_dispatch` triggers are unaffected.
- A stray untiered `models.unit:` block (`wh_llmbox: 25`, `wh_galaxy_perf: 10`) is still in the
  `time_budget.yaml` diff and is unrelated to Qwen — to be dropped.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ign/qwen359b)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ign/qwen359b)


yieldthought:
Codex review:

1. [P1] Size the fused KV-write grids from the active decode bucket.

[models/demos/blackhole/qwen36/tt/model_config.py:443](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/demos/blackhole/qwen36/tt/model_config.py#L443) derives the two fused-write shard grids once from `max_batch_size`, and [models/demos/blackhole/qwen36/tt/attention/tp.py:949](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/demos/blackhole/qwen36/tt/attention/tp.py#L949) uses those fixed grids for every N300/9B paged decode. That conflicts with the serving contract: [models/demos/blackhole/qwen36/tt/qwen36_vllm.py:312](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/demos/blackhole/qwen36/tt/qwen36_vllm.py#L312) slices tokens, positions, and the page table to an active bucket `B < max_batch_size`. The fused kernel marks every core in the supplied grid active ([ttnn/cpp/ttnn/operations/experimental/paged_cache/device/fused_update_cache/paged_tiled_fused_update_cache_program_factory.cpp:393](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/fused_update_cache/paged_tiled_fused_update_cache_program_factory.cpp#L393) and [ttnn/cpp/ttnn/operations/experimental/paged_cache/device/fused_update_cache/paged_tiled_fused_update_cache_program_factory.cpp:406](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/fused_update_cache/paged_tiled_fused_update_cache_program_factory.cpp#L406)), then each core reads `index_ptr[my_batch_idx]` ([ttnn/cpp/ttnn/operations/experimental/paged_cache/device/kernels/dataflow/reader_paged_fused_update_cache_interleaved_start_id.cpp:99](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/ttnn/cpp/ttnn/operations/experimental/paged_cache/device/kernels/dataflow/reader_paged_fused_update_cache_interleaved_start_id.cpp#L99)). For example, a max-batch-32 model serving bucket 1 still launches 32 K cores and 32 V cores against one position/page-table row; cores 1..31 read beyond those buffers and can update arbitrary cache locations or fault. The non-fused path already avoids this via `_kv_shard_cfg(B)` at [models/demos/blackhole/qwen36/tt/attention/tp.py:806](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/demos/blackhole/qwen36/tt/attention/tp.py#L806), but the fused branch bypasses it. Suggestion: cache disjoint K/V (and corresponding RoPE) shard configs per active `B`, or otherwise ensure only `B` cores have work, and add an N300/9B regression with `max_batch_size > B`. The new N300 CI entry currently excludes this suite as a “27B/P150x4 path only” at [tests/pipeline_reorg/models_unit_tests.yaml:223](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/tests/pipeline_reorg/models_unit_tests.yaml#L223).

2. [P2] Restore the copyright declaration in the four new sweep files.

[models/tt_transformers/tests/sweeps/sweep_common.py:1](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/tt_transformers/tests/sweeps/sweep_common.py#L1), [models/tt_transformers/tests/sweeps/sweep_create_heads.py:1](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/tt_transformers/tests/sweeps/sweep_create_heads.py#L1), [models/tt_transformers/tests/sweeps/sweep_llama_mm_v2.py:1](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/tt_transformers/tests/sweeps/sweep_llama_mm_v2.py#L1), and [models/tt_transformers/tests/sweeps/sweep_sdpa_decode.py:1](https://github.com/tenstorrent/tt-metal/blob/77344bc6508e2a5a8d8e02119eebfeaa37807cdf/models/tt_transformers/tests/sweeps/sweep_sdpa_decode.py#L1) start with only the SPDX license identifier. The PR's `check-spdx-licenses` job is currently failing after the hook rewrites exactly these files. Suggestion: add the standard 2026 Tenstorrent copyright line (and separating blank comment line) above the existing identifier in all four files.

models/tt_transformers/tests/sweeps/sweep_common.py
```diff
@@ -0,0 +1,98 @@
+# SPDX-License-Identifier: Apache-2.0
```
yieldthought, line 3: Codex: The SPDX job is currently failing because this new file has the license identifier but no copyright declaration.

```suggestion
# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.

# SPDX-License-Identifier: Apache-2.0
```

models/tt_transformers/tests/sweeps/sweep_create_heads.py
```diff
@@ -0,0 +1,238 @@
+# SPDX-License-Identifier: Apache-2.0
```
yieldthought, line 3: Codex: The SPDX job is currently failing because this new file has the license identifier but no copyright declaration.

```suggestion
# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.

# SPDX-License-Identifier: Apache-2.0
```

models/tt_transformers/tests/sweeps/sweep_llama_mm_v2.py
```diff
@@ -0,0 +1,522 @@
+# SPDX-License-Identifier: Apache-2.0
```
yieldthought, line 3: Codex: The SPDX job is currently failing because this new file has the license identifier but no copyright declaration.

```suggestion
# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.

# SPDX-License-Identifier: Apache-2.0
```

models/tt_transformers/tests/sweeps/sweep_sdpa_decode.py
```diff
@@ -0,0 +1,278 @@
+# SPDX-License-Identifier: Apache-2.0
```
yieldthought, line 3: Codex: The SPDX job is currently failing because this new file has the license identifier but no copyright declaration.

```suggestion
# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.

# SPDX-License-Identifier: Apache-2.0
```

mtairum:
## Review — "should fix" items

Separate from the blocking items (I'll raise those inline). These are all things I'd want resolved before merge but none of them are correctness bugs in the model code.

**1. `tt/model_config.py:158,160` — the two `gdn_ab_gap` branches are byte-identical.**
Both compute `-(-self.gdn_nv_tp // 32) * 32 - self.gdn_nv_tp`. This is the same thing the `tt-comment-hygiene` bot flagged on 2026-09-02. Related: the comment block above still reads *"Mutually exclusive by construction, asserted below"*, but the assert was removed in f2a9944c when the `if/elif` refactor landed. Either restore genuinely distinct per-model expressions or collapse to one branch, and fix the stale sentence either way.

**2. `.github/time_budget.yaml` — the PR description understates the budget impact.**
The description mentions two bumps (`unit_tier3.wh_n300` 33→56, `e2e_tier3.wh_n300` 60→75). The diff actually changes seven entries, five of which you've annotated `needs #tt-metal-infra sign-off`:

| Key | Before | After |
| --- | --- | --- |
| `e2e_tier2.bh_quietbox_2` | 186 | 220 |
| `e2e_tier3.bh_quietbox_2` | 42 | 120 |
| `unit_tier1.bh_quietbox_2` | 390 | 400 |
| `unit_tier2.bh_quietbox_2` | 60 | 95 |
| `unit_tier3.bh_p150` | 30 | 50 |

`e2e_tier3.bh_quietbox_2` nearly triples. Could you get the infra sign-off onto this thread, and update the description to list all seven?

**3. `.github/time_budget.yaml` — stray `models.unit:` key.**
A bare `unit:` block (`wh_llmbox: 25`, `wh_galaxy_perf: 10`) is added between `e2e_tier3` and `unit_tier1`. `models` is a tiered team (`query_time_budget.py:78`, `TIERED_MODEL_TESTTYPES = {"unit","e2e","sweep"}`), so an untiered `unit` key looks like dead config. It's also unrelated to Qwen and unmentioned in the description — rebase artifact? Please drop it if so.

**4. HF cache seeding assumes a writable `/mnt/MLPerf`.**
Four CI entries now run `HF_HUB_OFFLINE=0 python3 -c "... snapshot_download(...)"`, and your own comment next to it says it *"needs `mlperf-read-only=false` on this run too, since /mnt/MLPerf is otherwise mounted `:ro`"*. Scheduled tier-3 runs don't set that input, so on a cold cache these jobs fail rather than seed. Worth confirming against the scheduled path before this merges.

**5. `tt/wh_compat.py` monkeypatches a shared module at import time.**
The file ends in a bare `apply()`, which rebinds `_shared_seq.chunk_gated_delta_rule_seq` for the whole process. The blast-radius note correctly establishes that Blackhole is unaffected (per-call `is_blackhole()`), but it doesn't cover the Wormhole case: any *other* WH model importing `models/experimental/gated_attention_gated_deltanet` in the same process silently picks up the bf16 fork. Either make `apply()` opt-in from the qwen36 entry points only, or extend the docstring to state the WH-side exposure.

**6. T3K numbers are reported as measured, but the code says they're an estimate.**
`tt/chunk_seq_wh.py` routes 8-device meshes back to upstream, with the comment: *"This is a head-count/TP-division ESTIMATE, not a T3K hardware measurement."* The description carries a "Measured — T3K, Qwen3.6-27B" table (0.995684 prefill PCC, 93.02% top-1). If those runs predate the narrowing they describe a different kernel than what ships. Could you confirm which, and re-label or re-run accordingly?

---

Separately on CI: the last green `(Tier 3) Models Unit` is `a04c7a5f` (2026-08-30) and the last `(Tier 3) Models E2E` is `709d12d6` (2026-08-29). Neither is an ancestor of the current head — the branch was rebased at `39f18555`, and `tt/attention/tp.py` and `tt/gdn/recurrent_decode_wh.py` have changed since. Both tiers will need a re-run on the current head before merge.

mtairum (changes requested):
.

mtairum (changes requested):
Is there a reason to bring this one up on N300 and not just P150?

I worry a bit since this is a Tier 3 model and we're adding around 50 min extra to N300.

I've left a comment with some Claude findings, please address them.

For the CI tests, please reduce the scope a bit.
My suggestion:
-  `Qwen3.5-9B` - P150 only
   - if we really must keep on N300, drop the N300 unit tests, and leave just the e2e.
-  `Qwen3.6-27B` - QB2 ✅ 

Also is there a way to reduce the amount (time) of unit tests?

mtairum:
## Rebased on latest `main` + tier CI dispatched

Rebased onto `main` (130 commits of drift) and force-pushed — head is now `a4fe6720`. All 57 commits replayed with no conflicts; the five files that overlapped with main (`time_budget.yaml`, `model_targets.yaml`, `tt_transformers/tt/model_config.py`, and both `pipeline_reorg` yamls) auto-merged cleanly, and the PR-only diff is unchanged at 84 files / +15699/-909.

Heads up that the rewrite detached the four inline SPDX comments from their diff anchors. They were already fixed in `d5ea1a59`, so nothing is outstanding there.

### Runs dispatched on `a4fe6720`

| Run | Covers | SKU |
| --- | --- | --- |
| [(Tier 3) Models Unit #33774111206](https://github.com/tenstorrent/tt-metal/actions/runs/33774111206) | `Qwen3.5-9B unit tests`, `Qwen3.5-9B P150 unit tests` | `wh_n300`, `bh_p150` |
| [(Tier 3) Models E2E #33774115934](https://github.com/tenstorrent/tt-metal/actions/runs/33774115934) | `Qwen3.5-9B e2e tests` | `wh_n300` |
| [(Tier 1) Models Unit #33774183327](https://github.com/tenstorrent/tt-metal/actions/runs/33774183327) | `Qwen3.6-27B unit tests` (the `test_rope_tp.py` / `test_generate_tp.py` additions) | `bh_quietbox_2` |

All three dispatched with read-write `/mnt/MLPerf` (`mlperf-read-only=false`, `mlperf-write-access=true` on the tier-1 workflow), since the new `HF_HUB_OFFLINE=0 snapshot_download` seed step can't persist against the default `:ro` mount. Worth noting the scheduled runs use the default, so a cold cache there will still fail — that's the point I raised as item 4 above.

### Three added entries can't be dispatched by name

`qwen3.6-27b` is missing from the `workflow_dispatch` model list of three workflows this PR adds entries to. Only `qwen3.5-9b` was registered, and only on the two tier-3 workflows:

| Entry added | Tier / workflow | Blocker |
| --- | --- | --- |
| `Qwen3.6 full-depth PCC tests` | 2 — `models-t2-unit-tests.yaml` | `qwen3.6-27b` not in the model choice list |
| `Qwen3.6-27B long-context e2e tests` | 2 — `models-t2-e2e-tests.yaml` | same |
| `Qwen3.6-27B teacher-forced accuracy` | 3 — `models-t3-e2e-tests.yaml` | same |

They'd only run under `model: all`, which fans out every model on that tier — not something I wanted to fire at shared `bh_quietbox_2` capacity to reach three jobs. Could you add `qwen3.6-27b` to those three lists the same way `qwen3.5-9b` was added to the tier-3 pair? The teacher-forcing one in particular is the accuracy-regression detector the description calls out, so it's worth having reachable on demand rather than only on the nightly.

Same note applies to the description's CI section — it says *"Registered `qwen3.5-9b` on the tier-3 `workflow_dispatch` lists"*, which is accurate but doesn't cover the 27B entries now landing at tiers 1-3.


We also should tighten the timeouts after these run.

mtairum:
## Review items addressed + tier-2/3 CI dispatched

Two commits on top of `a4fe6720`, no force-push:

- `de3e37c3453` — register `qwen3.6-27b` for `workflow_dispatch`; tighten the 9B `wh_n300` timeouts
- `cf8c3982b33` — compile the gh-aw locks; drop the stray budget key; document `wh_compat`'s Wormhole exposure

### On re-running with a writable `/mnt/MLPerf`

Worth being precise here, because it does not clear the blockers: the failing tier-1 run
[#33774183327](https://github.com/tenstorrent/tt-metal/actions/runs/33774183327) **already ran with
`mlperf-read-only: false` and `mlperf-write-access: true`** (both confirmed in the job log). The
`test_gdn_tp_prefill[in_bf8]` failure is a numerics assert, not an I/O or cache-permission failure,
so a re-run with the same mount flags reproduces it unchanged. The three runs dispatched below were
likewise dispatched with `mlperf-read-only=false`.

The writable-mount problem is a **separate** issue and is about *scheduled* runs specifically: they
use the default `:ro` mount and cannot pass the input, so a cold cache fails there rather than
seeds. That needs a config/infra change, not a re-run.

### CI runs — status

_Last updated 2026-09-04. Six pipelines dispatched; all four `bh_quietbox_2` legs are still waiting
on QB2 capacity._

| Run | Entry | Status |
| --- | --- | --- |
| [(Tier 3) Models Unit #33868620429](https://github.com/tenstorrent/tt-metal/actions/runs/33868620429) | `Qwen3.5-9B unit tests [wh_n300]`, `Qwen3.5-9B P150 unit tests [bh_p150]` | ✅ **pass** — 24.7 min / 19.5 min |
| [(Tier 3) Models E2E #33868624096](https://github.com/tenstorrent/tt-metal/actions/runs/33868624096) | `Qwen3.5-9B e2e tests [wh_n300]` | ✅ **pass** — 16.2 min |
| [(Tier 1) Models E2E #33868616032](https://github.com/tenstorrent/tt-metal/actions/runs/33868616032) | `Qwen3.6-27B e2e tests [bh_quietbox_2]` | ⏳ build + `load-test-matrix` green, test job queued on QB2 |
| [(Tier 2) Models Unit #33865697984](https://github.com/tenstorrent/tt-metal/actions/runs/33865697984) | `Qwen3.6 full-depth PCC tests [bh_quietbox_2]` | ⏳ queued on QB2 |
| [(Tier 2) Models E2E #33865701121](https://github.com/tenstorrent/tt-metal/actions/runs/33865701121) | `Qwen3.6-27B long-context e2e tests [bh_quietbox_2]` | ⏳ queued on QB2 |
| [(Tier 3) Models E2E #33865704242](https://github.com/tenstorrent/tt-metal/actions/runs/33865704242) | `Qwen3.6-27B teacher-forced accuracy [bh_quietbox_2]` | ⏳ queued on QB2 |

The three tier-2/3 runs were dispatched on `de3e37c3453`, the other three on `cf8c3982b33`. All with
`mlperf-read-only=false` (tier 1 E2E with `mlperf-write-access=true`). The four 27B entries resolved
by name from `model: qwen3.6-27b`, which is what `de3e37c3453` fixed — previously they were only
reachable under `model: all`.

**What the two greens settle.** Second consecutive green for all three 9B tier-3 jobs, now on the
current head, and the tightened caps hold with real headroom:

| Job | Measured | Cap | Headroom |
| --- | --- | --- | --- |
| `Qwen3.5-9B unit tests [wh_n300]` | 24.7 min (prev. 26.0) | 30 | 18% |
| `Qwen3.5-9B P150 unit tests [bh_p150]` | 19.5 min | 25 | 22% |
| `Qwen3.5-9B e2e tests [wh_n300]` | 16.2 min (prev. 16.0) | 20 | 19% |

`load-test-matrix` also passes on the tier-1 E2E run, which is the in-pipeline check on the
`time_budget.yaml` edits.

**Still to run: (Tier 1) Models Unit.** Not dispatched, because it re-fails identically until
blocker 1 below is resolved — the job exits at `test_gdn_tp.py` and the remaining 11 test files
never execute. That is the gate on blockers 1 and 2.

Note that (Tier 1) Models E2E had **never** been run on this branch before now, despite this PR
modifying its `Qwen3.6-27B e2e tests` entry (the HF cache seed step) — worth watching.

---

## Blockers

1. **`test_gdn_tp_prefill[blackhole-in_bf8-device_params0-1x4]` — `PCC: 0.0`.** Prefill output is
   constant (`PCC is NaN (zero variance / constant tensor)`, allclose fallback also fails). This is
   **PR-introduced, not main drift**: the `in_dtype` parametrize is new at `test_gdn_tp.py:472`, and
   `main`'s `test_gdn_tp.py` is 13/13 green. The `in_bf16` sibling passes at 0.99996. Per the test's
   own docstring, `layer.py` narrows attention_norm's prefill gather to bf8 on GDN layers, so in-proj
   sees a bf8 `in0` in prefill and a bf16 one in decode — suspect either the harness `ttnn.typecast`
   to `bfloat8_b` on the K-sharded `_pf_in` tensor, or the fused AGMM prefill path zeroing on a bf8
   `in0`. Needs a real answer either way: if the asymmetry is benign, the row should be relaxed with
   a measured threshold, not deleted.

2. **Tier-1 coverage is 2 of 13 test files.** The job exits at `test_gdn_tp.py`, so `test_rope_tp.py`,
   `test_generate_tp.py`, `test_model_tp.py`, `test_decode_bucketing.py` and 7 others never ran on
   this branch. Separately, `main` has been red on
   `test_decode_bucketing.py::test_all_buckets_fit_trace_region` (`Timeout (>300.0s)`) every run
   since 2026-08-28 — unknown whether this PR inherits it. Tier 1 needs a clean re-run once item 1
   is resolved.

3. **Cold HF cache needs a writable mount.** Four entries seed `HF_HOME` via
   `HF_HUB_OFFLINE=0 … snapshot_download`, which requires `mlperf-read-only=false`. Scheduled runs
   use the default `:ro`, so a cold cache fails rather than seeds. Every green run so far was
   hand-dispatched with the override.

## Should fix before merge

4. **Infra sign-off on the five `bh_quietbox_2` budget raises** — `e2e_tier3.bh_quietbox_2` 42→120
   nearly triples. The PR description now tables all seven raises with drivers; the sign-off itself
   is still outstanding on this thread.
5. **T3K numbers labelled "Measured" while `tt/chunk_seq_wh.py` calls them an ESTIMATE.** If those
   runs predate the 8-device narrowing they describe a different kernel than what ships — re-label
   or re-run.
6. **`tt/model_config.py:158,160` — the two `gdn_ab_gap` branches are byte-identical**, and the
   comment above still reads *"Mutually exclusive by construction, asserted below"* after the assert
   was dropped in `f2a9944c`.

## Closed in these two commits

- ~~`qwen3.6-27b` missing from three `workflow_dispatch` model lists~~ — registered on
  `models-t2-unit-tests.yaml`, `models-t2-e2e-tests.yaml`, `models-t3-e2e-tests.yaml`.
- ~~PR description doesn't cover the 27B entries~~ — CI section rewritten: all 8 pipeline entries
  (tier / SKU / budget / dispatch name) and all 7 `time_budget.yaml` raises.
- ~~Placeholder timeouts~~ — `Qwen3.5-9B unit tests` `wh_n300` 33→30 (measured 26 min),
  `Qwen3.5-9B e2e tests` `wh_n300` 25→20 (measured 16 min), with
  `models.unit_tier3.wh_n300` 56→53 and `models.e2e_tier3.wh_n300` 75→70.
- ~~`test-command.lock.yml` / `silencer.lock.yml` need `gh aw compile`~~ — compiled with gh-aw
  v0.86.2, matching the `compiler_version` already in both lock headers; `.md` sources untouched.
  Heads up that the recompile also picks up unrelated drift that landed on `main` without a
  recompile (`kimi_k3`, `disaggregated_prefill`, `indexed_fused_update_cache`,
  `run-llk-sanity-tests`) — kept whole rather than hand-trimmed, since these files are marked
  DO NOT EDIT.
- ~~Stray untiered `models.unit:` block~~ — dropped; `TIERED_MODEL_TESTTYPES` in
  `query_time_budget.py` includes `unit`, so nothing read it. The six per-entry budget annotations
  went with it (the file carries none anywhere else).
- ~~`wh_compat.py` Wormhole blast radius undocumented~~ — docstring now states that importing any
  qwen36 GDN module runs `apply()` as an import side effect and rebinds the shared module
  process-wide, so any other Wormhole consumer silently gets DRAM chunk-seq activations and the
  bf16 relayout. Documented rather than made opt-in (qwen36 is the only Wormhole consumer today,
  and both overrides are L1-relief on a path that OOMs without them), with a note on what to change
  if a second consumer appears and why a shared pytest session is the case to watch.

mtairum:
Please address the blockers above.

One suggestion would be to split the Qwen3.5-9B from qwen3.6-27B changes into 2 PRs.
But I'm happy either way.

We just need to make sure the CI tests we are adding all pass. Until then I cannot approve.

ign-christy:
@mtairum 

Ran the 27B e2e coverage: T3K [unit](https://github.com/tenstorrent/tt-metal/actions/runs/34309139516/job/102335239495) and [e2e](https://github.com/tenstorrent/tt-metal/actions/runs/34309032526/job/102352583676) both pass, and so does the [QB2 tier-1 e2e](https://github.com/tenstorrent/tt-metal/actions/runs/34316684312/job/102355775012) entry. Only the [QB2 long-context run](https://github.com/tenstorrent/tt-metal/actions/runs/34316767237/job/102357216844) fails, and not on a test — the model is now ~3.5× faster on TTFT and ~50% faster on decode than the June targets (`9a71e73`, #47738), and `validate_perf_targets` flags "too fast" the same as "too slow", so every line reads `[much better than expected, update target]`.

The speedup comes from main's 27B work (#49946, #49565, #50089, #48861, #51556), not this PR — its perf changes are N300-only (`tp_common.py:170`) — so could whoever owns the 27B Blackhole numbers refresh them? `16384`/`32768` will fail the same way later. Happy to update them here instead if you'd prefer.

One note: for these dispatches I temporarily raised the two 27B e2e timeouts from 1200 s to 1800 s, then reverted them.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -482,6 +488,91 @@
...
+  # Same -k filters as the 27B leg: single ISL + determinism + one vision case, not the full
+  # 4k-256k ISL ladder or vision demo matrix (out of CI budget on both device targets).
+  cmd: |
+    export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=Qwen/Qwen3.5-9B TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/Qwen/Qwen3.5-9B MESH_DEVICE=N300
+    # One-time seed of the shared HF_HOME cache when this revision isn't on the mount yet.
+    # This job's env sets HF_HUB_OFFLINE=1 (and CI=true forces local_files_only=True
+    # downstream too), so a cold cache fails outright with LocalEntryNotFoundError instead of
+    # downloading. Override HF_HUB_OFFLINE for just this command; needs mlperf-read-only=false
+    # on this run too, since /mnt/MLPerf is otherwise mounted :ro and the download can't persist.
+    if [ -w "$HF_HOME" ]; then HF_HUB_OFFLINE=0 python3 -c "import os; from huggingface_hub import snapshot_download; snapshot_download(os.environ['HF_MODEL'])"; else echo "HF_HOME is read-only; relying on the pre-seeded cache"; fi
```
mtairum, line 503: Can we remove if statements from the yaml please? The HF_HOME is read only always, unless you pass the flag when running CI.

tests/pipeline_reorg/models_e2e_tests.yaml
```diff
@@ -482,6 +488,91 @@
...
+  owner_id: U08HL8X1ECD # Aniruddha Tupe
+  team: models
+
+# T3K (1x8) e2e for the 27B -- the demo path on Wormhole TP=8, which no other entry runs.
+# Single ISL only: the bh_quietbox_2 tier-1 entry stays the primary e2e gate, this covers the
+# 8-device mesh the Wormhole-specific gates (chunk_seq_wh, wh_compat) are selected by.
+- name: Qwen3.6-27B T3K e2e tests
+  cmd: |
+    export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=Qwen/Qwen3.6-27B TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/Qwen/Qwen3.6-27B MESH_DEVICE=T3K
+    if [ -w "$HF_HOME" ]; then HF_HUB_OFFLINE=0 python3 -c "import os; from huggingface_hub import snapshot_download; snapshot_download(os.environ['HF_MODEL'])"; else echo "HF_HOME is read-only; relying on the pre-seeded cache"; fi
```
mtairum, line 564: Can we remove if statements from the yaml please? The HF_HOME is read only always, unless you pass the flag when running CI.

tests/pipeline_reorg/models_unit_tests.yaml
```diff
@@ -194,6 +194,12 @@
   cmd: |
     unset TT_METAL_WATCHER  # blocked by #50886
     export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=Qwen/Qwen3.6-27B TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/Qwen/Qwen3.6-27B
+    # One-time seed of the shared HF_HOME cache when this revision isn't on the mount yet.
+    # This job's env sets HF_HUB_OFFLINE=1 (and CI=true forces local_files_only=True
+    # downstream too), so a cold cache fails outright with LocalEntryNotFoundError instead of
+    # downloading. Override HF_HUB_OFFLINE for just this command; needs mlperf-read-only=false
+    # on this run too, since /mnt/MLPerf is otherwise mounted :ro and the download can't persist.
+    if [ -w "$HF_HOME" ]; then HF_HUB_OFFLINE=0 python3 -c "import os; from huggingface_hub import snapshot_download; snapshot_download(os.environ['HF_MODEL'])"; else echo "HF_HOME is read-only; relying on the pre-seeded cache"; fi
```
mtairum, line 202: Can we remove if statements from the yaml please? The HF_HOME is read only always, unless you pass the flag when running CI.

tests/pipeline_reorg/models_unit_tests.yaml
```diff
@@ -205,15 +211,81 @@
...
+# T3K (1x8) coverage for the 27B. chunk_seq_wh.py sends 8-device meshes to upstream's fp32
+# chunk-seq relayout while N150/N300 take the bf16 fork -- a per-mesh branch no other entry
+# reaches, so a regression in it is currently invisible. Tier 3: secondary platform, and the
+# 27B's primary gate stays the bh_quietbox_2 tier-1 entry.
+- name: Qwen3.6-27B T3K unit tests
+  cmd: |
+    unset TT_METAL_WATCHER  # blocked by #50886
+    export HF_HOME=/mnt/MLPerf/huggingface HF_MODEL=Qwen/Qwen3.6-27B TT_CACHE_PATH=/mnt/MLPerf/huggingface/tt_cache/Qwen/Qwen3.6-27B
+    export MESH_DEVICE=T3K
+    if [ -w "$HF_HOME" ]; then HF_HUB_OFFLINE=0 python3 -c "import os; from huggingface_hub import snapshot_download; snapshot_download(os.environ['HF_MODEL'])"; else echo "HF_HOME is read-only; relying on the pre-seeded cache"; fi
```
mtairum, line 278: Can we remove if statements from the yaml please? The HF_HOME is read only always, unless you pass the flag when running CI.

mtairum (approved):
Approving now, but please remove the if statement from the model e2e and unit yaml.

mtairum:
And please rebase the PR on top of latest main.

models/demos/blackhole/qwen36/tt/chunk_seq_wh.py
```diff
@@ -0,0 +1,459 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""Wormhole variant of the shared GDN chunk-seq kernel wrapper.
+
+WHY THIS FILE EXISTS
+--------------------
+``chunk_gated_delta_rule_seq`` in
+``models/experimental/gated_attention_gated_deltanet/tt/ttnn_delta_rule_seq.py`` relayouts the
```
atupe-tt, line 8: Why not just add a dtype kwarg to `ttnn_delta_rule_seq` file? That way, this entire file won't be required

models/demos/blackhole/qwen36/tt/wh_compat.py
```diff
@@ -0,0 +1,129 @@
...
+
+    _check_fork_is_current()
+
+    # --- 1. chunk-seq activations: DRAM on Wormhole -------------------------------------- #
+    _orig_seq_memory_config = _shared._seq_memory_config
+
+    def _seq_memory_config(seq_len):
+        """Wormhole: always DRAM for the chunk-seq activations.
+
+        Upstream is ``L1 if seq_len <= _L1_SEQ_THRESHOLD else None``. On Wormhole those L1
```
atupe-tt, line 104: Let's make `_L1_SEQ_THRESHOLD` arch-aware. That way we don't need this new file

models/demos/blackhole/qwen36/tt/gdn/conv_fir_wh.py
```diff
@@ -0,0 +1,297 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""Wormhole variant of the shared GDN depthwise causal conv1d FIR.
+
+WHY THIS FILE EXISTS
+--------------------
+``_causal_conv1d_fir`` in
+``models/experimental/gated_attention_gated_deltanet/tt/ttnn_gated_deltanet.py`` builds its padded
```
atupe-tt, line 8: Can we add the 2 required kwargs to the `_causal_conv1d_fir` in `ttnn_gated_deltanet.py`?
Then this entire file is not needed

models/demos/blackhole/qwen36/tt/prefill_norm_tuned.py
```diff
@@ -0,0 +1,218 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+"""DistributedNorm with a TUNED prefill all-gather.
+
+WHY THIS FILE EXISTS
+--------------------
+``DistributedNorm.forward`` in ``models/tt_transformers/tt/distributed_norm.py`` reads
```
atupe-tt, line 7: Try to reuse it instead of adding a new file

models/demos/blackhole/qwen36/tt/tp_common.py
```diff
@@ -716,6 +1290,38 @@ def prepare_gdn_qkv(qkv_w, key_dim, value_dim, nk, dk, nv, dv, tp):
     return torch.cat(shards, dim=0)
 
 
+def tuned_vocab_all_gather(input_tensor, mesh_device, tt_ccl, dim, topology, num_workers_per_link, chunks_per_sync):
```
atupe-tt, line 1293: Can we delete this method and reuse from ` models/tt_transformers/tt/ccl.py::tt_all_gather`?

models/demos/blackhole/qwen36/tt/tp_common.py
```diff
@@ -211,19 +458,179 @@ def _full_grid_crs(grid):
...
+    "widest subblock" can force a narrow block and MORE passes -- N=7168 gets the ideal 1x4 subblock
+    and is still slower than N=6912's 1x3. A prime tile count (193) is worst of all: its only divisors
+    are 1/5/25, and 25 (one pass) overflows L1.
+
+    So: pad the width until the tile count factors well. Candidates are multiples of grid_size[0] (so
+    per_core_N divides exactly and every column of the grid is used), within max_waste extra tiles.
+    Scored by (passes, waste) -- fewest K passes, then least padding. Returns n unchanged when nothing
+    beats it, so callers can use it unconditionally.
+
+    NOT CURRENTLY CALLED BY THE MODEL. Its one user was the GDN in-proj width, which now reaches ONE K
```
atupe-tt, line 498: If this is not called anywhere, can we remove this?

models/demos/blackhole/qwen36/tt/vision/vision_model_config.py
```diff
@@ -20,6 +26,166 @@ def __init__(self, model_name):
...
+# cores, and at 11008 rows q=256 gives 344 units (6 rounds, 40 idle slots) against q=128's 688 (11
+# rounds, 16 idle) -- the balance q=256 lost once the row count stopped being a multiple of
+# 256x64/heads. Measured at the demo shape against the HiFi4/bf16-K baseline: 9B/N300 18.27 -> 13.43
+# ms (1.36x), 27B/T3K 6.06 -> 3.80 ms (1.59x), pcc 0.999909 both.
+#
+# `exp_approx=True` measured 0.6% (inside noise) and accumulates approximate exp over flash chunks
+# across 27 blocks -- not taken. 512/512 is rejected: the flash CBs reach 1,949,888 B against L1's
+# 1,499,136 B.
+_VISION_SDPA_TUNING = dict(fidelity="hifi2", k_bf8b=True, q_chunk=128, k_chunk=512, exp_approx=False)
+_VISION_SDPA_TUNING_BY_DEVICE = {}
```
atupe-tt, line 163: Where does this get populated? If not anywhere, then remove this

models/demos/blackhole/qwen36/tests/perf/test_profile_lm_head.py
```diff
@@ -0,0 +1,320 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
```
atupe-tt, line 1: Are all the `test_profile_*` files really required? If I understand it correctly, they were only used for Tracy reports. Please remove them if they were only used for testing

models/tt_transformers/model_params/Qwen3.5-9B/config.json
```diff
@@ -0,0 +1,103 @@
+{
```
atupe-tt, line 1: Why is this config file required? Why not pull configs from HF?

models/tt_transformers/tests/sweeps/matmul_sweep.py
atupe-tt, line 1: Are all these `/sweeps` required? If they are model specific, move them to the model directory. If not needed, please remove them

models/demos/blackhole/qwen36/README-N300-9B.md
atupe-tt, line 1: Move this content to the main README. Please do not copy paste as it is. Keep it concise

models/demos/blackhole/qwen36/README-T3K-27B.md
atupe-tt, line 1: Move this content to the main README. Please do not copy paste as it is. Keep it concise

atupe-tt (changes requested):

atupe-tt:
@ign-christy I have added my review. Please address all the comments. 
Also, cut down the size of comments/docstrings. They are too long

atupe-tt:
@ign-christy Please let me know once the comments are addressed. 
Also, update the branch. It has many merge conflicts

ign-christy:
@atupe-tt All comments are addressed and the branch is rebased onto latest main. The four Wormhole shims are gone — chunk_seq_wh, wh_compat, conv_fir_wh and prefill_norm_tuned are folded into upstream as kwargs (out_dtype, pad_layout, prefill_ag_tuning) and an arch-aware _L1_SEQ_THRESHOLD; tuned_vocab_all_gather now reuses ccl.py::tt_all_gather. The test_profile_* files, the sweeps and model_params/Qwen3.5-9B/config.json are removed (configs come from HF), both per-device READMEs are merged into the main one, and comments/docstrings are cut down throughout.

models/demos/blackhole/qwen36/tt/tp_common.py
atupe-tt: @ign-christy I believe you missed this

models/demos/blackhole/qwen36/tt/vision/vision_model_config.py
atupe-tt: @ign-christy I believe you missed this as well

atupe-tt (approved):
2 changes are not addressed. Please address them
Approved to unblock

PR #54575, opened 2026-08-27
[FIX] Reject duplicate sweep keys instead of averaging them silently

nstojicTT:
### PR Category

Test infrastructure

### Summary

`_collapse_duplicate_keys` averaged perf rows that shared a `(sweep-params, marker)`
key, then logged a warning. The warning was unreachable, so the averaging was
silent. This PR raises `PerfSchemaError` instead.

A key is a row's identity. Two rows with the same key mean the measurement is
ambiguous. That must fail the run, not produce a quiet mean.

**Depends on #<PR-1> (`nstojictt/perf-dedupe-dest-acc-sweep`).** Merged alone, the
next nightly fails on `perf_eltwise_unary_sfpu` and `perf_eltwise_binary_sfpu`.

### Why the warning was never seen

`helpers/logger.py` sends loguru to two places:

1. stdlib logging, which pytest shows only when `log_cli` is on. The perf runners
   use `-q` with `log_cli` off.
2. `test_run*.log`, which `llk-perf-impl.yaml` uploaded only under `if: failure()`.

So a green nightly wrote the warning to a file and discarded the file. Grepping all
ten perf jobs of run 33040431355 returns no loguru line at all — not the warning,
and not even the routine `Raw appending` info lines.

PR #<PR-3> (`nstojictt/perf-always-upload-test-logs`) fixes the upload separately.

### Before and after

**Before** — two rows in, one row out, a warning nobody receives:

```python
logger.warning(
    "{}: collapsing {} duplicate (sweep-params, marker) key(s) spanning "
    "{} rows into one row each (mean of metric columns); {} key(s) had "
    "differing metric values (run-to-run noise, or a distinguishing "
    "parameter not recorded as a column).", ...)

agg = {c: ("mean" if c in numeric_cols else "first") for c in value_cols}
collapsed = frame.groupby(key_cols, dropna=False, sort=False).agg(agg).reset_index()
return collapsed[list(frame.columns)]
```

**After** — the run fails and names the report and the offending keys:

```
PerfSchemaError: perf_eltwise_binary_sfpu.csv: 1 duplicate (sweep-params, marker)
key(s) spanning 2 rows; 1 key(s) disagreed on a metric value. A key must identify
one measurement. Either the sweep varies something that is not recorded as a
column, or two sweep points normalize onto the same recorded value (e.g. dest_acc
promoted from No to Yes for an outlier format combo). First duplicate key(s):
[{'formats.input_A': 'Float16_b', 'formats.output': 'Float16',
  'dest_acc': 'DestAccumulation.Yes', 'approx_mode': 'ApproximationMode.No',
  'mathop': 'MathOperation.SfpuElwadd', 'marker': 'TILE_LOOP'}]
```

The message keeps the `differing` count from the old warning. A pair that disagrees
on a metric is the stronger signal: it means the two collapsed rows were not the
same kernel, and the old mean was meaningless.

### The change

- Rename `_collapse_duplicate_keys` to `_reject_duplicate_keys`, matching what it
  now does. Three call sites in `combine_perf_reports` (raw, post, counters).
- Raise `PerfSchemaError` with the report label and the first three duplicate keys.
- Keep the fast path untouched: an empty frame, a frame with no `marker` column, or
  a frame with no duplicates is returned unchanged.

The broad handler also changes. It used to swallow any failure inside the check:

```diff
-    except Exception as e:
-        logger.warning("{}: duplicate-key collapse skipped due to error: {}", label, e)
-        return frame
+    except PerfSchemaError:
+        raise
+    except Exception as e:
+        raise PerfSchemaError(
+            f"{label}: duplicate-key check failed: {type(e).__name__}: {e}"
+        ) from e
```

Without the explicit re-raise, the broad `except` would catch the new
`PerfSchemaError` and return the un-collapsed frame. The check would then hide the
very duplicate it just found. The second clause means a bug inside the check fails
loudly instead of disabling it.

### Tests

Two tests added to `test_perf_report_hw_free.py`. Neither needs hardware.

```python
def test_duplicate_sweep_key_is_rejected_not_averaged():
    # two rows, identical key
    with pytest.raises(PerfSchemaError) as excinfo:
        _reject_duplicate_keys(frame, "perf_example.csv")
    assert "perf_example.csv" in str(excinfo.value)
    assert "duplicate" in str(excinfo.value)


def test_distinct_sweep_keys_pass_through_unchanged():
    # same rows, dest_acc differs
    pd.testing.assert_frame_equal(
        _reject_duplicate_keys(frame, "perf_example.csv"), frame
    )
```

### Documentation

`.claude/skills/perf-report/SKILL.md` and its `.cursor` copy told the reader to
subtract "any rows merged by duplicate-key collapse" when checking a row count.
Nothing is merged now, so both lines say duplicates are rejected instead.

### Verification

- The two new unit tests.
- A full LLK perf run on a branch carrying this PR plus #<PR-1> and #<PR-3>:
  run 33068927365. Every test/shard report satisfies `rows == cases x markers`,
  which is the property this PR now enforces.
- To confirm the raise path on real data, run the perf suite with this PR but
  **without** #<PR-1>. It must fail on `perf_eltwise_unary_sfpu` and name the
  duplicate keys.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-duplicate-key-error)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-duplicate-key-error)


njokovicTT (approved):

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
```diff
@@ -343,20 +343,19 @@ def get_unique_base_names(input_dir: Path):
...
-    silently.
+def _reject_duplicate_keys(frame: pd.DataFrame, label: str) -> pd.DataFrame:
+    """Fail the session if any rows share the same (sweep-params, marker) key.
+
+    The key columns are a row's identity, so two rows carrying the same key make
+    the measurement ambiguous. It happens when a sweep varies something the CSV
+    does not record, or when the harness normalizes a parameter onto a value
+    another sweep point already uses (dest_acc promoted from No to Yes for an
+    outlier format combo in TestConfig).
+
```
skrsmanovicTT: Let's not keep implementation history in comments, use commit message for that purpose. 

Claude always use this approach by default but there is really no need to document old behaviour and justification for changes as part of comment or function header

skrsmanovicTT (commented):
If there is no quick solution for previously stacked PR let's discuss best approach for merging strategy

tt_metal/tt-llk/tests/python_tests/helpers/perf/core.py
nstojicTT: agree

ndivnicTT (approved):

PR #54617, opened 2026-08-27
Qwen3-32B Blackhole Galaxy: nightly vLLM serving lane, sampling correctness fixes, FF2 prefill L1 fix

mgiermakowskiTT:
### Ticket

N/A

### CI
[qwen e2e](https://github.com/tenstorrent/tt-metal/actions/runs/33859128846)
[qwen vLLM](https://github.com/tenstorrent/tt-metal/actions/runs/33865234451)
[llama WH clx](https://github.com/tenstorrent/tt-metal/actions/runs/33872125651/job/101042877438)

### Problem description

Qwen3-32B on Blackhole Galaxy had no vLLM serving coverage in CI, and three independent defects stood between the model and a servable lane.

**1. Long prompts overran L1 in the FF2 prefill matmul.** The matmul sized its output block from a hardcoded 320-tile volume with `per_core_M = ceil(seq_len / (7 * 32))`, so its static circular buffers grew with sequence length (~763 KB at 16k, ~980 KB at 32k) until they collided with the contiguous L1 window left by the resident decode-path buffers:

```
Statically allocated circular buffers in program 4878 clash with L1 buffers on core range [0-0 - 3-6].
L1 buffer allocated at 988992 and static circular buffer region ends at 990848
```

**2. A request that changed sampling mode killed the engine.** Sampling traces are keyed on `(penalties, log_probs, force_argmax, bucket)`, but warmup pre-compiled only the combination that happened to be active then. When a later request asked for logprobs or penalties, `sample()` found an uncaptured slot and — because the decode path passes `skip_precompile=True` — compiled and ran that program for the first time *inside* a live trace capture. That trips `TT_FATAL !is_capturing_trace`, which aborts the process rather than surfacing an error to the client.

**3. Sampling returned a token from the previous decode step.** `capture_trace()` returned `slot["output"]` immediately after `end_trace_capture`. But `begin/end_trace_capture` only *record* ops — nothing executes — so the output buffer still held whatever the last replay left in it. Every time a new trace key appeared, the caller got the prior step's token. This is what made seeded sampling look non-deterministic on BH: the sequence was shifted, not random.

Two smaller device-parameter bugs surfaced while chasing (3), both reachable from ordinary vLLM requests:

- **`k` was passed to the kernel unclamped.** `ttnn.sampling` walks `k` entries of each user's two-face candidate row. vLLM encodes "no top-k filter" as `k = vocab_size`, which is far above `max_top_k`, so the top-p scan and the draw ran off the end of that row into unrelated L1.
- **`temperature = 0` inverted its own meaning.** `temp` multiplies the logits before the softmax, so the greedy encoding `temp = 0` flattened every candidate to equal probability and made the draw uniform — the opposite of the argmax the caller asked for.

#### `models/demos/llama3_70b_galaxy/tt/qwen_model_config.py` — bound the FF2 output block

Caps the output-block volume at 160 tiles on Blackhole (320 elsewhere, unchanged), and always pads `per_core_M` up to a multiple of 8 on BH. The padding matters as much as the bound: `find_out_block_h` only accepts a block height that divides `per_core_M`, so a prime-ish `per_core_M` leaves no divisor near the cap and collapses the block to a tiny, slow shape.

#### `models/common/sampling/generator.py` — `precompile(all_configs=True)`

Adds an opt-in mode that compiles every `_TraceKey` flag combination (`itertools.product((False, True), repeat=3)`) instead of only the currently active one, so no request can be the first to compile a program during trace capture. The loop sets `_force_argmax_sampling` directly rather than going through `reset_params()`, because `reset_params()` re-derives the flag from live `k`/`p`/`temp` and would overwrite the in-flight request params; only the flag selects which program gets compiled. Saved state is restored in `finally`.

Defaults to `False`, so this is inert for every existing caller. `models/demos/llama3_70b_galaxy/tt/generator.py` is the one opt-in (the call site right before decode trace capture).

#### `models/common/sampling/generator.py` — replay the trace where its result is consumed

The stale-token fix is placed in `sample()`'s lazy-capture branch, not inside `capture_trace()`:

```python
self.capture_trace(logits, tt_out_tok=tt_out_tok, skip_precompile=skip_precompile)
# begin/end_trace_capture only records the ops, so the captured output buffer
# still holds the previous step's token; replay before returning it as this
# step's sample. Callers that only capture (warmup) must not pay for this.
return self._execute_trace(key)
```

`capture_trace()` has two callers with opposite needs. `sample()` consumes the return value as this step's token. `models/tt_transformers/tt/generator.py` calls it at warmup and **discards** the return. Executing inside `capture_trace()` would therefore make tt_transformers' warmup write a token derived from dummy warmup logits into `tt_out_tok` — the decode token-feedback buffer — silently corrupting its first real decode step. Fixing at the point of consumption leaves capture-only callers untouched.

#### `models/common/sampling/tt_sampling.py` — normalize device params at both entry points

New `_normalize_device_params(k, temp)` clamps `k` to `max_top_k` and maps `temp == 0 → 1.0` (the `force_argmax` path already handles greedy correctly). It is applied in both places params reach the device — initial setup and the dynamic `update` path — so a mid-flight `update` cannot reintroduce an out-of-range `k`. The `update` path also now reuses the returned tensors instead of re-wrapping the raw arguments with `torch.tensor(...)`, including for the `k == 1` greedy tie-break mask, which otherwise would have gone out of sync with the clamped `k`.

#### CI lane

New nightly lane *"Qwen3-32B device (BH Galaxy)"* in `tests/pipeline_reorg/vllm_model_tests.yaml` (TG mesh, `arch-blackhole`, `--data_parallel_size 4 --max_num_seqs 8 --async-scheduling`), plus the `bh_galaxy` SKU wiring in `.github/workflows/vllm-model-tests.yaml` and `.github/time_budget.yaml`, and serving instructions in the demo README.

The lane pins `fabric_config: FABRIC_2D_TORUS_XY`. Blackhole Galaxy decode runs column-axis collectives that a 1D fabric cannot route across columns; `worker_l1_size` and `trace_region_size` follow the BH Qwen demo (`text_qwen_demo.py`).

### Validation

Run on Blackhole Galaxy (32 chips, mesh 8×4) against the CI-accurate serving stack — `vllm==0.25.1` plus the canonical `tenstorrent/vllm-tt-plugin` @ `859be83`, driven through tt-inference-server, matching what nightly CI assembles:

| Suite | Result |
| --- | --- |
| Chat conformance | **22 passed** in 434 s |
| Streaming spec | **3 passed** in 172 s |
| Serving benchmark sweeps | **6/6 exit 0** |

Benchmark throughput (3 sweeps each, output token throughput / mean TTFT / mean TPOT):

| Concurrency | Throughput | Mean TTFT | Mean TPOT |
| --- | --- | --- | --- |
| 1 | 32.3–32.5 tok/s | 99.3–121.9 ms | 30.2 ms |
| 32 | 917.4–918.6 tok/s | 518.1–524.9 ms | 31.0 ms |

The prefill L1 fix was verified separately by sweeping warmup sequence lengths 128 → 32768 with zero circular-buffer clashes.

Server warmup measured 700–761 s across runs on this machine.

### Checklist

- [x] Blackhole Galaxy hardware validation on the CI-accurate stack (table above)
- [x] Prefill L1 clash fix verified across sequence lengths 128 → 32768
- [x] Sampling fixes validated end-to-end through the vLLM serving path
- [ ] New/existing tests provide coverage for changes

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mgiermakowski/qwen3-32b-vllm)

### DIFF
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import copy
+import itertools
 import random
 import secrets
 from dataclasses import dataclass, fields, replace
@@ -348,6 +349,7 @@ class SamplingGenerator:
         logits: ttnn.Tensor,
         *,
         tt_out_tok: Optional[ttnn.Tensor] = None,
+        all_configs: bool = False,
     ) -> None:
         """Run the sampling pipeline once without capturing, to compile it and size its scratch.
 
@@ -357,13 +359,48 @@ class SamplingGenerator:
         left inline, this pass allocates device buffers that a live trace can corrupt on replay.
 
         ``logits`` only has to match the spec of the tensor that will later be captured, not be it.
+
+        ``all_configs`` compiles every ``_TraceKey`` flag combination rather than just the one active
+        now. Traces are keyed on (penalties, log_probs, force_argmax), but warmup only ever runs one of
+        those, so a request asking for logprobs or penalties later finds an uncaptured slot and, because
+        callers pass ``skip_precompile=True``, executes its program for the first time inside a live
+        trace capture -- TT_FATAL !is_capturing_trace, which kills the engine rather than erroring.
         """
-        self._run_sampling(
-            logits,
-            penalties_on=self._penalties_active,
-            tt_out_tok=tt_out_tok,
-            count_tokens=False,
-        )
+        if not all_configs:
+            self._run_sampling(
+                logits,
+                penalties_on=self._penalties_active,
+                tt_out_tok=tt_out_tok,
+                count_tokens=False,
+            )
+            return
+
+        log_probs = self.tt_sampling.log_probs_calculator
+        saved_penalties = self._penalties_active
+        saved_force_argmax = self.tt_sampling._force_argmax_sampling
+        saved_enabled = list(log_probs.logprobs_enabled)
+        saved_num_logprobs = list(log_probs.num_logprobs)
+        try:
+            for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=3):
+                self._penalties_active = penalties_on
+                # Set the flag directly: reset_params() would re-derive it from k/p/temp and overwrite
+                # the live request params, and only the flag selects the program being compiled.
+                self.tt_sampling._force_argmax_sampling = force_argmax
+                log_probs.set_log_probs_mode(log_probs_on, num_logprobs=0)
+                self._run_sampling(
+                    logits,
+                    penalties_on=penalties_on,
+                    tt_out_tok=tt_out_tok,
+                    count_tokens=False,
+                )
+        finally:
+            self._penalties_active = saved_penalties
+            self.tt_sampling._force_argmax_sampling = saved_force_argmax
+            log_probs.logprobs_enabled = saved_enabled
+            log_probs.num_logprobs = saved_num_logprobs
+            log_probs.enable_log_probs = any(saved_enabled)
+            log_probs.topk_logprobs_needed = log_probs.enable_log_probs
+            self._log_probs_active = log_probs.enable_log_probs
 
     def capture_trace(
         self,
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -452,6 +452,10 @@ class SamplingGenerator:
         slot["kwargs"] = {"tt_out_tok": tt_out_tok}
         _mark_trace_buffers_corruptible(self._active_trace_bucket, (logits, output))
 
+        # begin/end_trace_capture only records the ops, so the output buffer still holds the
+        # previous step's token. Run the trace once or this call replays that stale token.
+        ttnn.execute_trace(self.mesh_device, trace_id, cq_id=self.cq_id, blocking=False)
+
         return slot["output"]
 
     def _execute_trace(self, key: _TraceKey) -> ttnn.Tensor:
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -452,10 +452,6 @@ class SamplingGenerator:
         slot["kwargs"] = {"tt_out_tok": tt_out_tok}
         _mark_trace_buffers_corruptible(self._active_trace_bucket, (logits, output))
 
-        # begin/end_trace_capture only records the ops, so the output buffer still holds the
-        # previous step's token. Run the trace once or this call replays that stale token.
-        ttnn.execute_trace(self.mesh_device, trace_id, cq_id=self.cq_id, blocking=False)
-
         return slot["output"]
 
     def _execute_trace(self, key: _TraceKey) -> ttnn.Tensor:
@@ -498,11 +494,15 @@ class SamplingGenerator:
         else:
             key, slot = self._trace_slot(penalties_on, log_probs_on, force_argmax)
             if slot["id"] is None:
-                return self.capture_trace(
+                self.capture_trace(
                     logits,
                     tt_out_tok=tt_out_tok,
                     skip_precompile=skip_precompile,
                 )
+                # begin/end_trace_capture only records the ops, so the captured output buffer
+                # still holds the previous step's token; replay before returning it as this
+                # step's sample. Callers that only capture (warmup) must not pay for this.
+                return self._execute_trace(key)
 
             self._validate_trace_inputs(slot, logits, tt_out_tok)
             tt_out = self._execute_trace(key)
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -92,6 +92,12 @@ class _TraceKey:
     bucket: int | None = None
 
 
+# precompile(all_configs=True) enumerates every combination of the bool fields above. Derive the
+# count so a new flag breaks the unpacking there instead of silently leaving its programs
+# uncompiled -- which reopens TT_FATAL !is_capturing_trace on the first request needing it.
+_TRACE_KEY_FLAGS = sum(f.type in (bool, "bool") for f in fields(_TraceKey))
+
+
 class SamplingGenerator:
     """
     High-level sampling helper that owns both `TTSampling` and `TTPenalties`
@@ -381,7 +387,7 @@ class SamplingGenerator:
         saved_enabled = list(log_probs.logprobs_enabled)
         saved_num_logprobs = list(log_probs.num_logprobs)
         try:
-            for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=3):
+            for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=_TRACE_KEY_FLAGS):
                 self._penalties_active = penalties_on
                 # Set the flag directly: reset_params() would re-derive it from k/p/temp and overwrite
                 # the live request params, and only the flag selects the program being compiled.
@@ -396,10 +402,8 @@ class SamplingGenerator:
         finally:
             self._penalties_active = saved_penalties
             self.tt_sampling._force_argmax_sampling = saved_force_argmax
-            log_probs.logprobs_enabled = saved_enabled
-            log_probs.num_logprobs = saved_num_logprobs
-            log_probs.enable_log_probs = any(saved_enabled)
-            log_probs.topk_logprobs_needed = log_probs.enable_log_probs
+            # Restore through the setter that owns the derived flags rather than re-deriving them here.
+            log_probs.set_log_probs_mode(saved_enabled, num_logprobs=saved_num_logprobs)
             self._log_probs_active = log_probs.enable_log_probs
 
     def capture_trace(
@@ -476,6 +480,9 @@ class SamplingGenerator:
         """
         Convenience wrapper that either runs the sampling module directly or
         replays a captured trace.
+
+        ``count_tokens`` only applies to the untraced path: the token-count update is recorded into
+        the trace at capture time, so a replay always performs it.
         """
 
         penalties_on = self._penalties_active
@@ -484,6 +491,8 @@ class SamplingGenerator:
         # Explicit request seeds update a persistent seed tensor every token;
         # run them directly so trace replay cannot observe stale seed state.
         use_internal_trace = enable_trace and not self.seed_manager.has_active_request_seed()
+        if use_internal_trace and not count_tokens:
+            raise ValueError("count_tokens=False cannot be honoured on a traced sample(); pass enable_trace=False.")
         if not use_internal_trace:
             tt_out = self._run_sampling(
                 logits,
--- a/models/common/sampling/generator.py
+++ b/models/common/sampling/generator.py
@@ -388,6 +388,10 @@ class SamplingGenerator:
         saved_num_logprobs = list(log_probs.num_logprobs)
         try:
             for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=_TRACE_KEY_FLAGS):
+                # Models that disable force-argmax never reach that program, and it is not runnable
+                # under their sub-device config (untilize with sub_core_grids=None).
+                if force_argmax and not self.tt_sampling._allow_force_argmax_sampling:
+                    continue
                 self._penalties_active = penalties_on
                 # Set the flag directly: reset_params() would re-derive it from k/p/temp and overwrite
                 # the live request params, and only the flag selects the program being compiled.
@@ -426,12 +430,17 @@ class SamplingGenerator:
             logger.debug(
                 f"Pre-compiling sampling path before trace capture (penalties={penalties_on},log_probs_on={log_probs_on},force_argmax={force_argmax})"
             )
+            # TTPenalties.apply() rewrites its input in place, so compiling on `logits` itself would
+            # leave the capture buffer already penalized and make the first replay penalize it twice.
+            scratch = ttnn.clone(logits) if penalties_on else logits
             self._run_sampling(
-                logits,
+                scratch,
                 penalties_on=penalties_on,
                 tt_out_tok=tt_out_tok,
                 count_tokens=False,
             )
+            if scratch is not logits:
+                ttnn.deallocate(scratch)
 
         trace_id = ttnn.begin_trace_capture(self.mesh_device, cq_id=self.cq_id)
         sampled = self._run_sampling(
--- a/models/common/sampling/tt_sampling.py
+++ b/models/common/sampling/tt_sampling.py
@@ -144,6 +144,20 @@ class TTSampling(LightweightModule):
     def force_argmax_sampling(self) -> bool:
         return self._force_argmax_sampling
 
+    def _normalize_device_params(self, k, temp):
+        """Clamp k to the candidate row the kernel actually holds and replace the greedy temp=0.
+
+        ttnn.sampling walks k entries of each user's two-face candidate row, so a k above max_top_k
+        (callers pass vocab_size to mean "no top-k filter") runs the top-p scan and the draw off the
+        end of that row into unrelated L1. temp multiplies the logits before the softmax, so a caller
+        encoding greedy as temp=0 flattens every candidate to an equal probability and turns the draw
+        uniform -- the opposite of the argmax it asked for.
+        """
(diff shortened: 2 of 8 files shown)


viktorpusTT (approved):
Approve (vllm scope, did not review sampling code) if CI is green for this model.

models/common/sampling/tt_sampling.py
```diff
@@ -144,6 +144,20 @@ def _is_force_argmax_sampling(self, k, p, temp):
...
+    def _normalize_device_params(self, k, temp):
+        """Clamp k to the candidate row the kernel actually holds and replace the greedy temp=0.
+
+        ttnn.sampling walks k entries of each user's two-face candidate row, so a k above max_top_k
+        (callers pass vocab_size to mean "no top-k filter") runs the top-p scan and the draw off the
+        end of that row into unrelated L1. temp multiplies the logits before the softmax, so a caller
+        encoding greedy as temp=0 flattens every candidate to an equal probability and turns the draw
+        uniform -- the opposite of the argmax it asked for.
+        """
+        k = torch.clamp(torch.as_tensor(k), max=self.max_top_k)
```
mtairum: Clamp is upper-bound only. `k < 1` is the documented "no restriction" encoding (`generator.py:650`: `if top_k[i] < 1: top_k[i] = 32`) and passes straight through; a negative `k` becomes a huge value via `from_torch(dtype=ttnn.uint32)` — the exact out-of-bounds walk this guards.

```python
k = torch.as_tensor(k)
k = torch.where(k < 1, torch.full_like(k, self.max_top_k), k).clamp(max=self.max_top_k)
```

models/common/sampling/tt_sampling.py
```diff
@@ -144,6 +144,20 @@ def _is_force_argmax_sampling(self, k, p, temp):
...
+        temp = torch.as_tensor(temp, dtype=torch.float32)
+        temp = torch.where(temp == 0.0, torch.ones_like(temp), temp)
```
mtairum, line 171: `temp == 0 -> 1.0` without also forcing `k = 1`. `format_sampling_params` rewrites greedy as the full triple `(temp=1, k=1, p=0)`; here only `temp` changes, so a direct caller with `temp=0, top_k=50` samples from top-32 instead of the argmax it asked for. Less wrong than the old uniform draw, still wrong.

Mirror the full greedy rewrite, or state in the docstring that only `format_sampling_params` output is supported.

models/common/sampling/tt_sampling.py
```diff
@@ -600,6 +615,7 @@ def reset_params(
         empty_slots: list[int] | None = None,
     ):
         """Update sampling parameters (k, p, temperature, logprobs) dynamically."""
+        k, temp = self._normalize_device_params(k, temp)
```
mtairum, line 631: `torch.as_tensor(None)` raises, so `reset_params` no longer accepts `k=None` / `temp=None` — which `_is_force_argmax_sampling` on the very next line still tolerates via `is_default_value`. Unreachable from the one caller today, but that tolerance is now dead and misleading.

models/common/sampling/generator.py
```diff
@@ -357,13 +359,48 @@ def precompile(
...
+            )
+            return
+
+        log_probs = self.tt_sampling.log_probs_calculator
+        saved_penalties = self._penalties_active
+        saved_force_argmax = self.tt_sampling._force_argmax_sampling
+        saved_enabled = list(log_probs.logprobs_enabled)
+        saved_num_logprobs = list(log_probs.num_logprobs)
+        try:
+            for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=3):
```
mtairum: `repeat=3` hardcodes `_TraceKey`'s boolean arity. A fourth bool field silently reopens `TT_FATAL !is_capturing_trace` for the new dimension. Deriving the product from `dataclasses.fields(_TraceKey)` makes adding a field force this loop to notice.

models/common/sampling/generator.py
```diff
@@ -357,13 +359,48 @@ def precompile(
...
+                    penalties_on=penalties_on,
+                    tt_out_tok=tt_out_tok,
+                    count_tokens=False,
+                )
+        finally:
+            self._penalties_active = saved_penalties
+            self.tt_sampling._force_argmax_sampling = saved_force_argmax
+            log_probs.logprobs_enabled = saved_enabled
+            log_probs.num_logprobs = saved_num_logprobs
+            log_probs.enable_log_probs = any(saved_enabled)
```
mtairum: Duplicates the derivation `set_log_probs_mode` owns (`tt_log_probs.py:355-359`). `log_probs.set_log_probs_mode(saved_enabled, num_logprobs=saved_num_logprobs)` restores through the method that owns it, so a future third dependent flag can't go stale here.

models/common/sampling/generator.py
```diff
@@ -457,11 +494,15 @@ def sample(
...
-                return self.capture_trace(
+                self.capture_trace(
                     logits,
                     tt_out_tok=tt_out_tok,
                     skip_precompile=skip_precompile,
                 )
+                # begin/end_trace_capture only records the ops, so the captured output buffer
+                # still holds the previous step's token; replay before returning it as this
+                # step's sample. Callers that only capture (warmup) must not pay for this.
+                return self._execute_trace(key)
```
mtairum, line 523: `sample()`'s `count_tokens` argument is ignored on this branch. Harmless while capture didn't execute; now the trace runs and `update_output_tokens` is counted regardless. No caller pairs `count_tokens=False` with `enable_trace=True` today (`llama3_70b_galaxy/tt/generator.py:867` uses `enable_trace=False`), but the trap is live now.

models/demos/llama3_70b_galaxy/tt/qwen_model_config.py
```diff
@@ -675,21 +675,30 @@ def w2_prg_config(seq_len):
...
+
+                # On Blackhole, persistent decode-path L1 buffers sit at ~989 KB on core range
+                # [0-0 - 3-6], leaving this matmul only ~761 KB of contiguous L1 for its static CBs,
+                # whose size is dominated by the output block volume. Bound that volume, and always pad
+                # per_core_M to a multiple of 8 so that a block height close to the bound actually divides
+                # per_core_M (otherwise a prime-ish per_core_M collapses the block to a tiny, slow shape).
+                max_out_block_volume = 160 if self.is_blackhole else 320
                 per_core_M = (
-                    next_multiple_of_8(total_per_core_out_M) if total_per_core_out_M > 320 else total_per_core_out_M
+                    next_multiple_of_8(total_per_core_out_M)
```
mtairum, line 686: Padding to a multiple of 8 doesn't guarantee a divisor near the cap when `per_core_M / 8` is prime. seq 32768: 147 -> 152 = 8·19 -> best block is 19x5 = 95 tiles, 59% of the 160 cap, while 16k gets the full 16x10 = 160. The longest sequence ends up with the smallest block.

Not a correctness issue and `next_multiple_of_16` trades it for more padded rows at 4k/8k — flagging in case it shows up in 32k TTFT.

tests/pipeline_reorg/vllm_model_tests.yaml
```diff
@@ -375,6 +375,32 @@
...
+  # fabric (1D fabrics fail on the cross-column route); worker_l1_size follows the
+  # BH Qwen demo (text_qwen_demo.py).
+  tt-config: '{"dispatch_core_axis": "col", "sample_on_device_mode": "all", "fabric_config": "FABRIC_2D_TORUS_XY", "worker_l1_size": 1345000, "trace_region_size": 184915840}'
+  additional-server-args: "--data_parallel_size 4 --max_num_seqs 8 --async-scheduling"
+  # Off until BH sampling is fixed: seed determinism is unsupported on BH (the full
+  # ttnn.sampling pipeline does not run there; see SAMPLING_AG_CONFIG in
+  # qwen_model_config.py), and the penalty/structured-output tests currently fail on
+  # BH Galaxy on main as well. The filter below covers the seeding gap once enabled.
+  sampling-tests: false
+  sampling-test-filter: "not test_seeding_and_variety and not test_mixed_params_batch"
```
mtairum: Dead while `sampling-tests: false` — the step is gated on `sampling-tests` alone (`vllm-model-tests-impl.yaml:590`). Drop it, or add one line saying it's a ready-to-flip marker so a future reader doesn't delete it as cruft.

mtairum (commented):
Full review, focused on the three fixes. Verified the mechanisms against source rather than the PR body — summary of what held up, then the concerns.

**Verified**

- FF2: `out_CB_tiles` and `interm0_CB_tiles` are both `out_block_h*out_block_w` (`matmul_multicore_reuse_mcast_2d_program_factory.cpp:165-173`); `per_core_M*per_core_N` only applies to a sharded output. So the volume cap is the whole L1 lever and the `per_core_M` padding is L1-neutral. Simulating `w2_prg_config` reproduces your numbers: 16k -> 37x5 = 185 tiles (~758 KB across the two CBs), 32k -> 49x5 = 245 (~1003 KB). Padding is required: with a 160 cap and unpadded `per_core_M=74`, 16k collapses to 2x10 = 20 tiles. Non-BH path is bit-identical.
- `sample()` replay: both branches return the same `slot["output"]` object, only delta is the `execute_trace`. Placing it here rather than in `capture_trace()` is right — `models/tt_transformers/tt/generator.py:2146` captures at warmup and discards the return.
- `precompile(all_configs=True)` is safe on every flag it forces: `TTPenalties` allocates masks/counts in `__init__` (`tt_penalties.py:139-144`); `reset_params` copies `k/p/temp` into the persistent device tensors immediately (`tt_sampling.py:650-652`), so toggling `_force_argmax_sampling` never reads unset tensors; `set_log_probs_mode(x, num_logprobs=0)` compiles the same program as `num_logprobs>0` (`topk_logprobs_needed = enable_log_probs`, `tt_log_probs.py:359`).
- `reset_trace()` calls `ttnn.release_trace`, so no trace-region leak from the force_argmax flips that `allow_force_argmax=True` makes frequent on BH.

**Most pressing**

1. **`ensure-bh-links-online` never runs for this lane.** `.github/workflows/vllm-model-tests-impl.yaml:195-197` gates on `contains(matrix.test-group.mesh-device, 'P150')`; this lane sets `mesh-device: TG`, so a 32-chip Blackhole box gets no eth-link recovery. Every other bh_galaxy pipeline gates on the SKU (`runtime-unit-tests-impl.yaml:103`, `runtime-integration-tests-impl.yaml:94`) or `startsWith(sku, 'bh_')` (`tm-fabric-tests-impl.yaml:186`, `fabric-sanity-tests-impl.yaml:138`). Flaky links will land as unexplained fabric hangs. `sku` is set by `prepare_test_matrix.py`, so `if: ${{ startsWith(matrix.test-group.sku, 'bh_') }}` works.

2. **`k` clamp is upper-bound only** — see inline on `tt_sampling.py:156`. `k < 1` is the documented "no restriction" encoding and still reaches the kernel.

3. **The same block sizer is unfixed one file over.** `models/demos/llama3_70b_galaxy/tt/model_config.py:1062-1095` is a byte-for-byte copy with the hard 320 and no BH padding, for Llama-70B on the same Galaxy generator. Not reachable in CI today (no BH Galaxy Llama lane), but it is the identical latent clash — worth extracting the helper or mirroring the bound.

**Follow-ups, outside this diff**

- `models/experimental/llama32_1b_quasar/sampling/generator.py:409-415` has the identical uncorrected `return self.capture_trace(...)` stale-token pattern.
- `models/tt_transformers/tt/generator.py:2078` still calls `precompile()` without `all_configs=True`, and `_record_decode_trace_text` captures only the current config. Softer manifestation: some paths pass `skip_precompile=False`, so it precompiles inline and hits the #52176 allocation hazard rather than the `TT_FATAL`.

On the test-coverage points from the bots: agree, and `_normalize_device_params` is the cheap one — pure tensor function, no device needed. A host-only test over `w2_prg_config` asserting `out_block_h*out_block_w <= 160` and `per_core_M % out_block_h == 0` for 4096..32768 would also gate the FF2 regression without hardware.

viktorpusTT:
Please add CI job link in PR desc.

yieldthought:
Codex review:

**[P1] Respect the force-argmax capability during precompile** — [models/common/sampling/generator.py:390](https://github.com/tenstorrent/tt-metal/blob/02ef823fb970d29ea513e223f183303435f52f42/models/common/sampling/generator.py#L390)

The new loop sets `_force_argmax_sampling=True` even when the model disables that path. Qwen's Wormhole config explicitly sets `allow_force_argmax=False`, but the shared Galaxy caller now opts into `all_configs=True`. This breaks the existing Wormhole serving lane: [CI job 100899342800](https://github.com/tenstorrent/tt-metal/actions/runs/33831695715/job/100899342800) reaches `precompile -> TTSampling.forward -> ttnn.split` and aborts with `Kernel group cores do not match sub device cores for programmable core type TENSIX`. The relevant Python is unchanged between that run's `0fc0d52456b6` and the current head. Suggestion: skip force-argmax combinations when `_allow_force_argmax_sampling` is false; the remaining combinations still cover every supported request.

**[P2] Preserve unmodified logits across lazy precompile and replay** — [models/common/sampling/generator.py:514](https://github.com/tenstorrent/tt-metal/blob/02ef823fb970d29ea513e223f183303435f52f42/models/common/sampling/generator.py#L514)

For an unseeded request with active penalties and the default `skip_precompile=False`, `capture_trace()` first runs sampling eagerly. `TTPenalties.apply()` modifies the supplied logits in place. The newly added replay then applies the same penalties again before selecting the returned token. The generic transformer caller forwards this default, so the first sample for a new trace configuration can be wrong. For example, logits `[2.0, 1.2]` with a `0.5` presence penalty on token 0 should become `[1.5, 1.2]` and select token 0; precompile plus replay produces `[1.0, 1.2]` and selects token 1. Suggestion: use separate scratch logits for the eager precompile and add a first-capture regression with nonzero penalty history.

Validation: reviewed all eight changed files and relevant callers. Host checks passed for FF2 block sizing over 897 lengths (4K–32K), unchanged Wormhole configs, and trace control flow; the double-penalty counterexample was reproduced using the actual Python orchestration with a simulated device backend. No new hardware tests were run.

models/common/sampling/generator.py
```diff
@@ -357,13 +365,46 @@ def precompile(
...
+            return
+
+        log_probs = self.tt_sampling.log_probs_calculator
+        saved_penalties = self._penalties_active
+        saved_force_argmax = self.tt_sampling._force_argmax_sampling
+        saved_enabled = list(log_probs.logprobs_enabled)
+        saved_num_logprobs = list(log_probs.num_logprobs)
+        try:
+            for penalties_on, log_probs_on, force_argmax in itertools.product((False, True), repeat=_TRACE_KEY_FLAGS):
+                self._penalties_active = penalties_on
```
yieldthought, line 395: Codex: Skip force-argmax warmup when the model disables this path; Qwen3-32B on Wormhole currently aborts during this iteration.

```suggestion
                if force_argmax and not self.tt_sampling._allow_force_argmax_sampling:
                    continue
                self._penalties_active = penalties_on
```

mtairum (approved):

tdowdallTT (approved):

PR #54628, opened 2026-08-27
Make Quasar compute sync_threads() barrier all TRISCs using HW semaphores and add sync_all_cores().

### LINKED ISSUE #46296: [QSR]: Extend sync_threads to work with Tensix
Currently `sync_threads` only enables a barrier point for DMs but for completeness sake we should enable syncing threads across all triscs within a Neo, and all triscs across all Neos, and all triscs + DMs 

We can also try to backport this but not sure if that is necessary.

msudumbrekar-TT:
## Summary

On Quasar, a compute kernel runs on 16 TRISCs (4 NEOs × 4 TRISCs). `sync_threads()` used to wait only within one NEO, so threads on other NEOs could race ahead.

This PR makes compute `sync_threads()` wait for **all TRISCs in the kernel**.

DM `sync_threads()` is unchanged: it still waits only for the DM threads in that kernel. Folding DMs into the same call would hang if compute is not at the same wait.

For the case where **DMs and TRISCs** need to wait on each other, there is a new API: `sync_all_cores(participants)`. Both the DM kernel and the compute kernel must call it with the **same** count (`num_dm_threads + 4 * num_neos`).

Fixes #46296

## Why two APIs

| Call | Who waits | When to use |
|---|---|---|
| `sync_threads()` on DM | DM threads in this kernel | existing DM code |
| `sync_threads()` on compute | all TRISCs in this kernel (4 * get_num_threads()) | existing compute code |
| `sync_all_cores(n)` | DMs + TRISCs | only when both kernels opt in with the same `n` |

## How it works

TRISC L1 atomics are not coherent across TRISC roles, so compute barriers use tensix HW semaphores instead of L1.

- Compute `sync_threads()`: semaphores **30** (arrival) and **29** (generation)
- `sync_all_cores()`: semaphores **28** and **27**
- Watcher already owns **31**; that is unchanged
- DM0 `thread_sync_init()` zeros 27–30 at kernel start

WH/BH: these calls stay no-ops.

## Test plan

`KernelThreadSyncTest` (Quasar only for the new cases):

- [x] `BarrierSynchronizesThreads` — DM `sync_threads()` still works
- [x] `ComputeBarrierSynchronizesAllTriscs` — 16 TRISCs wait together
- [x] `AllCoresBarrierSynchronizesDmAndTriscs` — 6 DMs + 16 TRISCs via `sync_all_cores(22)`

Verified on Quasar 1x3 emu (`TT_METAL_SLOW_DISPATCH_MODE=1`).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:msudumbrekar/qsr-sync-threads-tensix)

### DIFF
--- a/tt_metal/hw/firmware/src/tt-2xx/dmk.cc
+++ b/tt_metal/hw/firmware/src/tt-2xx/dmk.cc
@@ -27,6 +27,7 @@
 // Per-processor kernel thread info for Quasar (set from kernel_config before kernel runs)
 thread_local uint32_t num_sw_threads __attribute__((used));
 thread_local uint32_t my_thread_id __attribute__((used));
+thread_local uint32_t my_barrier_id __attribute__((used));
 
 extern "C" [[gnu::section(".start")]]
 uint32_t _start() {
@@ -98,6 +99,12 @@ uint32_t _start() {
     // Setup after the go signal so the previous kernel has completed.
     num_sw_threads = launch_msg->kernel_config.num_sw_threads[hartid];
     my_thread_id = launch_msg->kernel_config.kernel_thread_id[hartid];
+    // Barrier slot for this kernel. thread_0_hartid is the lowest hart running this same kernel
+    // text, so every thread of a kernel derives the same value and two co-resident kernels derive
+    // different ones — exactly the per-kernel identity sync_threads() needs to keep its rendezvous
+    // separate. This reuses the identity the thread-0 election above already relies on, rather
+    // than spending launch-message space on it (kernel_config_msg_t has no room to spare).
+    my_barrier_id = thread_0_hartid;
 
     // Paint stack after all thread_local writes and CRT init are done.
     mark_stack_usage();
--- a/tt_metal/hw/firmware/src/tt-2xx/dmk.cc
+++ b/tt_metal/hw/firmware/src/tt-2xx/dmk.cc
@@ -100,10 +100,8 @@ uint32_t _start() {
     num_sw_threads = launch_msg->kernel_config.num_sw_threads[hartid];
     my_thread_id = launch_msg->kernel_config.kernel_thread_id[hartid];
     // Barrier slot for this kernel. thread_0_hartid is the lowest hart running this same kernel
-    // text, so every thread of a kernel derives the same value and two co-resident kernels derive
-    // different ones — exactly the per-kernel identity sync_threads() needs to keep its rendezvous
-    // separate. This reuses the identity the thread-0 election above already relies on, rather
-    // than spending launch-message space on it (kernel_config_msg_t has no room to spare).
+    // text, so every thread of a kernel derives the same slot and two co-resident kernels derive
+    // different ones, keeping their sync_threads() rendezvous separate.
     my_barrier_id = thread_0_hartid;
 
     // Paint stack after all thread_local writes and CRT init are done.
--- a/tt_metal/hw/inc/api/kernel_thread_globals.h
+++ b/tt_metal/hw/inc/api/kernel_thread_globals.h
@@ -42,11 +42,13 @@ extern volatile KernelBarrier g_kernel_barrier[NUM_KERNEL_BARRIERS];
 
 #endif  // !COMPILE_FOR_TRISC
 
-// Semaphores 29/30 reserved for compute sync_threads; 31 is the watcher ring buffer.
+// Semaphores 27–30: thread barriers; 31: watcher ring buffer.
 constexpr uintptr_t TENSIX_GLOBAL_SEM_BASE = 0x01840000;
 constexpr uint32_t TENSIX_GLOBAL_SEM_STRIDE = 0x40;
 constexpr uint32_t COMPUTE_BARRIER_ARRIVED_SEM_IDX = 30;
 constexpr uint32_t COMPUTE_BARRIER_GENERATION_SEM_IDX = 29;
+constexpr uint32_t ALL_CORES_BARRIER_ARRIVED_SEM_IDX = 28;
+constexpr uint32_t ALL_CORES_BARRIER_GENERATION_SEM_IDX = 27;
 constexpr uint32_t TENSIX_GLOBAL_SEM_VALUE_MASK = 0xFFFFu;
 
 inline volatile uint32_t* tensix_global_sem(uint32_t idx) {
@@ -108,9 +110,34 @@ inline void thread_sync_init() {
     }
     tensix_global_sem_init(COMPUTE_BARRIER_ARRIVED_SEM_IDX, 0);
     tensix_global_sem_init(COMPUTE_BARRIER_GENERATION_SEM_IDX, 0);
+    tensix_global_sem_init(ALL_CORES_BARRIER_ARRIVED_SEM_IDX, 0);
+    tensix_global_sem_init(ALL_CORES_BARRIER_GENERATION_SEM_IDX, 0);
 #endif
 }
 #endif  // !COMPILE_FOR_TRISC
+
+inline void tensix_global_sem_barrier(uint32_t arrived_idx, uint32_t generation_idx, uint32_t participants) {
+#if defined(ARCH_QUASAR)
+    if (participants <= 1) {
+        return;
+    }
+    asm volatile("fence rw, rw" ::: "memory");
+    uint32_t next_generation = (tensix_global_sem_read(generation_idx) + 1) & TENSIX_GLOBAL_SEM_VALUE_MASK;
+    uint32_t arrived = tensix_global_sem_fetch_add(arrived_idx, 1) + 1;
+    if (arrived == participants) {
+        tensix_global_sem_init(arrived_idx, 0);
+        tensix_global_sem_init(generation_idx, next_generation);
+    } else {
+        while ((tensix_global_sem_read(generation_idx) & TENSIX_GLOBAL_SEM_VALUE_MASK) != next_generation) {
+        }
+    }
+    asm volatile("fence rw, rw" ::: "memory");
+#else
+    (void)arrived_idx;
+    (void)generation_idx;
+    (void)participants;
+#endif
+}
 
 // barrier_idx selects an independent barrier so co-resident kernels with different
 // participant counts (e.g. a DFB's producer vs consumer kernel) don't share a counter.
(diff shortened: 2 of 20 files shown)


tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -95,19 +151,36 @@ inline void wait_threads(uint32_t participants, uint32_t barrier_idx = 0) {
     }
 
 #if defined(ARCH_QUASAR)
+#if defined(COMPILE_FOR_TRISC)
+    (void)barrier_idx;  // compute has one semaphore pair; barrier_idx is DM-only
```
abhullar-tt, line 155: I know this isn't introduced by your changes but `barrier_idx` was added to the DM `sync_threads` to make sure separate groups of DM kernels can sync without impacting each other. e.g. readers use barrier_idx = 0 and writers use barrier_idx = 1. 

this isn't really intuitive from a user pov. Instead could the barrier_id be encoded in the kernel with fw exposing a get_barrier_id and have sync_threads use that. This would remove the barrier_idx argument entirely.

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -40,9 +40,34 @@ struct KernelBarrier {
...
 
-#endif // ARCH_QUASAR
+inline uint32_t tensix_global_sem_read(uint32_t idx) { return *tensix_global_sem(idx); }
+
+inline uint32_t tensix_global_sem_fetch_add(uint32_t idx, uint32_t inc) {
+    return *reinterpret_cast<volatile uint32_t*>(reinterpret_cast<uintptr_t>(tensix_global_sem(idx)) + 4 * (inc + 8));
+}
+
+#if defined(COMPILE_FOR_TRISC)
+constexpr uint32_t kTriscCoresPerNeo = 4;
```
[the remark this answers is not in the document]
abhullar-tt: agreed to not duplicate this here

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -95,19 +151,36 @@ inline void wait_threads(uint32_t participants, uint32_t barrier_idx = 0) {
...
 
 inline void sync_threads(uint32_t barrier_idx = 0) {
+#if defined(ARCH_QUASAR) && defined(COMPILE_FOR_TRISC)
+    wait_threads(get_num_threads() * kTriscCoresPerNeo, barrier_idx);
+#else
     wait_threads(get_num_threads(), barrier_idx);
+#endif
+}
+
+inline void sync_all_cores(uint32_t participants) {
```
abhullar-tt, line 180: Is num_participants the sum of all dm riscs + tensixes (or does it count individual triscs). I think it should be at tensix granularity because from user pov they parallelize across tensixes. Would it be possible to structure this such that user does not have to supply num_participants? 

Also `sync_all_cores` might not be the best name because it might look like it does cross node synchornization. I think renaming it to reflect its a dm + compute barrier would be better.

abhullar-tt (commented):
could you also add a test that is multiple types of DM kernels (eg three dm kernels each 2 way parallelized) + a compute kernel parallelized 4 ways. 

the comments in kernel_thread_globals.h are also specific to DMs. could you also extend the comments to cover the compute sync and the DM <-> compute sync

tt-vjovanovic (approved):

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -7,6 +7,58 @@
...
+//      different thread counts sharing one counter would never agree on a
+//      target and would hang.
+//
+//   2. The TRISCs of one compute kernel -- sync_threads() from a compute
+//      kernel. Participants are every TRISC of every NEO the kernel occupies,
+//      i.e. get_num_threads() * NUM_TRISC_CORES, so the barrier releases only
+//      once unpack, math, pack and SFPU on all of those NEOs have arrived.
+//      The L1 barrier of flavor 1 cannot be reused: the TRISC roles do not
+//      share a coherent view of L1, so an arrival written by one role is not
+//      guaranteed to be seen by another. This rides on a pair of tensix global
```
kstevensTT: This is not true. TRISCs share a coherent view of L1 and can use atomics to L1 to sychronize. It is still fine to use global semaphores instead, if there is another reason to (e.g. lower latency), but at minimum this comment needs to be updated with the correct reasoning.

TRISCs vs. DMs do not share a coherent view of L1 over atomics (case #3), so we still need to use global semaphores there.

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -105,10 +237,61 @@ inline void wait_threads(uint32_t participants, uint32_t barrier_idx = 0) {
...
+ * From a DM kernel the participants are that kernel's DM harts. From a compute kernel they are
+ * all NUM_TRISC_CORES TRISCs of each NEO the kernel occupies, so unpack, math, pack and SFPU
+ * are released together. No-op on WH/BH, where a kernel is single-threaded.
+ *
+ * Which barrier a DM kernel rendezvouses on is decided by firmware, not the caller, so two
+ * co-resident kernels cannot collide. This does not synchronize with the other kernels on
+ * the core; use dm_compute_barrier() for a rendezvous that spans DM and compute.
+ */
+// clang-format on
+inline void sync_threads() {
```
kstevensTT, line 260: This changes the API to remove the argument, but I see places in the code where the argument is still used. Please make sure you update all uses of `sync_threads(0)` and all comments that refer to `sync_threads` with an argument before merging this change.

tt_metal/hw/inc/api/kernel_thread_globals.h
msudumbrekar-TT: Oh okay yes! Thanks fixed this!

tt_metal/hw/inc/api/kernel_thread_globals.h
msudumbrekar-TT: Good catch on this! Thanks! fixed it!

kstevensTT (approved):

tt_metal/hw/firmware/src/tt-2xx/dmk.cc
```diff
@@ -98,6 +99,12 @@ uint32_t _start() {
     // Setup after the go signal so the previous kernel has completed.
     num_sw_threads = launch_msg->kernel_config.num_sw_threads[hartid];
     my_thread_id = launch_msg->kernel_config.kernel_thread_id[hartid];
+    // Barrier slot for this kernel. thread_0_hartid is the lowest hart running this same kernel
+    // text, so every thread of a kernel derives the same value and two co-resident kernels derive
+    // different ones — exactly the per-kernel identity sync_threads() needs to keep its rendezvous
+    // separate. This reuses the identity the thread-0 election above already relies on, rather
+    // than spending launch-message space on it (kernel_config_msg_t has no room to spare).
```
abhullar-tt: could you make this comment more succinct. we don't need to mention the launch msg consideration.

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -105,10 +239,61 @@ inline void wait_threads(uint32_t participants, uint32_t barrier_idx = 0) {
...
+ * synchronize with any other core or node, despite spanning both processor kinds. No-op on
+ * WH/BH.
+ *
+ * | Argument     | Description                                      | Type     | Required |
+ * |--------------|--------------------------------------------------|----------|----------|
+ * | dm_threads   | Number of DM harts that will arrive.             | uint32_t | True     |
+ * | tensixes     | Number of tensix engines (NEOs) that will arrive.| uint32_t | True     |
+ */
+// clang-format on
+inline void dm_compute_barrier(uint32_t dm_threads, uint32_t tensixes) {
```
abhullar-tt: nit we use sync_threads on dm and compute but the cross dm + compute one is called barrier. could we make the names match up

abhullar-tt:
FYI @rtawfik01 / @fvranicTT this uses up more GlobalSemaphores

tt_metal/hw/inc/api/kernel_thread_globals.h
```diff
@@ -95,19 +147,36 @@ inline void wait_threads(uint32_t participants, uint32_t barrier_idx = 0) {
     }
 
 #if defined(ARCH_QUASAR)
+#if defined(COMPILE_FOR_TRISC)
+    (void)barrier_idx;
+    tensix_global_sem_barrier(COMPUTE_BARRIER_ARRIVED_SEM_IDX, COMPUTE_BARRIER_GENERATION_SEM_IDX, participants);
```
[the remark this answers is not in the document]
arikTT: @msudumbrekar-TT this is a good suggestion

abhullar-tt (approved):

tt_metal/hw/inc/api/kernel_thread_globals.h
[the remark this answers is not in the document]
msudumbrekar-TT: barrier_idx was removed from the API entirely per @kstevensTT review, so there's nothing left to discard or assert and sync_threads(1) no longer compiles. DM slots are now firmware-assigned.

PR #54644, opened 2026-08-27
fix ccl test failure on bh_loudbox

ubcheema:
## Summary
Three fixes for `BH ccl nightly tests` failures on the LoudBox leg.

### 1. `reduce_scatter_minimal_direct` TT_FATAL on 2D_TORUS_X + 2x4 mesh
`ttnn/cpp/ttnn/operations/ccl/ccl_common.cpp`: align `get_usable_topology` with `get_axis_topology` — when `cluster_axis` is supplied, resolve wrap using the physical wiring check (`is_axis_wrap_wired`) instead of `get_boundary_mode`, matching what the C++ CCL op consults internally.

The old `get_boundary_mode`-only heuristic classified a non-wrap-wired axis of length 4 (bh_loudbox row) as WRAP → `get_usable_topology` returned `Torus`, the Python-side skip guard let the test through, and then `reduce_scatter_direct_fabric_supported` inside the op saw `axis_topology == Linear` and hit:
> `reduce_scatter_minimal_direct needs a 1D fabric, or a 2D fabric whose ACTIVE AXIS wraps; got FABRIC_2D_TORUS_X on MeshShape([2,4]) with axis topology Linear`

The non-`cluster_axis` path still falls back to `get_boundary_mode` so the multi-host shard-span case (which physical wiring alone can't express) keeps working.

### 2. `test_all_to_all_async_generic_production_perf_and_accuracy` — inactive real-time profiler
`tests/pipeline_reorg/fabric_tests.yaml`: swap `bh_loudbox` for `bh_loudbox_civ2_viommu` under `BH ccl nightly tests`. That perf test requires the real-time profiler, which needs 64-bit PCIe / host IOMMU on Blackhole; `bh_loudbox` runners are not IOMMU-enabled and so the test hard-fails at `require_realtime_profiler`. `bh_loudbox_civ2_viommu` is the IOMMU-enabled equivalent — same 8-chip LoudBox fabric, strict superset for these tests. Also adds `bh_loudbox_civ2_viommu` to `scaleout.unit` in `.github/time_budget.yaml` and a `workflow_dispatch` checkbox for it in `.github/workflows/tm-fabric-tests.yaml`.

### 3. `loudbox_8x1 / head_to_sequence` bandwidth gate re-baseline
With #2 in place the perf test now actually runs on LoudBox, and the `head_to_sequence` reshard measured **29.598 GB/s** — below the existing 2%-margin gate against the 31.8 GB/s reference (31.164 GB/s minimum). Lower `_REFERENCE_USEFUL_BANDWIDTH_GBPS[(8, "head_to_sequence")]` from 31.8 → 29.6, matching what the IOMMU host actually delivers. `quietbox_4x1` references (both cases) and `(8, "sequence_to_head")` are untouched — the latter never executed this run and will be re-baselined separately if it turns out to need it.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ucheema/bh-ccl-test-fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ucheema/bh-ccl-test-fix)

### DIFF
--- a/.github/time_budget.yaml
+++ b/.github/time_budget.yaml
@@ -199,7 +199,7 @@ scaleout:
     wh_n300_civ2: 15
     wh_llmbox_civ2: 65
     bh_loudbox: 195
-    bh_loudbox_civ2_viommu: 195
+    bh_loudbox_civ2_viommu: 45
     bh_p300: 195
     bh_quietbox_2: 195
   stress:
--- a/.github/time_budget.yaml
+++ b/.github/time_budget.yaml
@@ -199,6 +199,7 @@ scaleout:
     wh_n300_civ2: 15
     wh_llmbox_civ2: 65
     bh_loudbox: 195
+    bh_loudbox_civ2_viommu: 195
     bh_p300: 195
     bh_quietbox_2: 195
   stress:
--- a/.github/workflows/tm-fabric-tests.yaml
+++ b/.github/workflows/tm-fabric-tests.yaml
@@ -64,6 +64,11 @@ on:
         required: false
         default: true
         type: boolean
+      run-bh_loudbox_civ2_viommu:
+        description: "Run on bh_loudbox_civ2_viommu"
+        required: false
+        default: true
+        type: boolean
       # Keep the "Available names" list below in sync with tests/pipeline_reorg/fabric_tests.yaml.
       hw-test-names:
         description: >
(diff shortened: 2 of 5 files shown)


.github/time_budget.yaml
```diff
@@ -199,6 +199,7 @@ scaleout:
     wh_n300_civ2: 15
     wh_llmbox_civ2: 65
     bh_loudbox: 195
```
roseli-TT, line 201: would you mind removing this line for the previous sku

sjameelTT (approved):

.github/time_budget.yaml
ubcheema: We have tests that run on bh_loudbox as well.

ubcheema: I have lowered the budged to 45 minutes though, as 195 was too big for these tests.

tdowdallTT (approved):

PR #54651, opened 2026-08-27
[Metal 2.0] Port `data_movement/concat`: 3 of 7 ProgramFactories

### LINKED ISSUE #55370: Metal 2.0 port: data_movement/concat
Port `ttnn/cpp/ttnn/operations/data_movement/concat` to Metal 2.0 / Host 2.0 APIs (Gen1 WH/BH target).

Parent: #41908

## Scope
Three native `ConcatDeviceOperation` program factories cleared for the port:
- `ConcatProgramFactory` — interleaved concat (row-major + tiled)
- `ConcatS2SRMProgramFactory` — sharded→sharded row-major width concat
- `ConcatS2STiledProgramFactory` — sharded→sharded tiled width concat (with groups)

Out of scope (still on the legacy API): `ConcatS2SMultiProgramFactory`, `ConcatS2IProgramFactory`, `ConcatBlockShardedProgramFactory`, `ConcatTiledUnalignedProgramFactory`.

## Target concept
Base `ProgramSpecFactoryConcept` — each ported factory exposes `create_program_artifacts` returning `ProgramArtifacts { spec, run_params }` and defines **no** `override_runtime_arguments`, so it takes the framework's tensor-binding-refresh cache-hit path (not the custom `ProgramRunArgs` override path). The `program_factory_t` variant now mixes ported (`create_program_artifacts`) and legacy (`create_descriptor`) factories; the device-op framework dispatches each alternative independently by concept and caches the selected factory per index, so op selection is unchanged.

## Tensor bindings (per factory)
- `ConcatProgramFactory` — interleaved I/O, all **Case 1** (via `TensorAccessor`). Inputs bound through a `TensorBindingSequence` (`inputs`) so a run-time-advanced index can pick the source tensor; output bound on the donor writer as `dst`.
- `ConcatS2SRMProgramFactory` / `ConcatS2STiledProgramFactory` — borrowed-memory input/output DFBs (`DataflowBufferSpec::borrowed_from`) → **clean**; no `TensorAccessor` in the kernels. The tiled factory additionally allocates four ordinary L1 scratch DFBs (per-input transposes, concat, output transpose).

## Kernels
Interleaved writers reuse existing shared donor `_metal2` forks — **no new forks**:
- `ttnn/cpp/ttnn/kernel/dataflow/writer_unary_stick_layout_interleaved_start_id_metal2.cpp`
- `ttnn/cpp/ttnn/operations/eltwise/unary/device/kernels/dataflow/writer_unary_interleaved_start_id_metal2.cpp`

Op-owned kernels converted in place: `reader_concat_interleaved_start_id.cpp`, `reader_concat_stick_layout_interleaved_start_id.cpp`, `reader_height_sharded_width_concat_two_tensors.cpp`, `reader_height_sharded_width_concat_two_tensors_tiled.cpp`, `writer_height_sharded_width_concat_two_tensors_tiled.cpp`, `compute/height_sharded_width_concat_two_tensors.cpp`.

## Status
Ported in #54651 — pending on-device validation on Wormhole and Blackhole (`tests/ttnn/unit_tests/operations/data_movement/test_concat.py`).

## Notes
- WH/BH only — no Quasar emulation.
- Basic Metal 2.0 port (no functional changes expected). Named `dfb::` / `args::` bindings, borrowed-memory DFBs, and a tensor binding sequence for the variable (up to 47) input count preserve legacy behavior.
- Legacy behavior preserved where Metal 2.0 defaults differ: compute pinned to `opt_level=O3` (Metal 2.0 defaults compute to O2); explicit `UnpackToSrc` unpack modes on the Float32 tiled path (legacy default was empty).
- **Watch:** the tiled writer binds the borrowed output DFB as both producer and consumer (self-loop) since it is the buffer's only toucher; the s2s-rm reader/writer pair bind all three DFBs role-free (one producer, one consumer per DFB) to satisfy the validator. Confirm both hold on device.
- **Watch:** the Float32 tiled unpack-mode branch is lightly covered — `test_sharded_concat_with_groups` parametrizes bf16/int32/bfloat8_b, not float32. Add/confirm a float32 tiled sharded→sharded width-concat test.

akerteszTT:
### Issue
Closes #55370.

### Changes
Ports 3 of the 7 `ttnn.concat` device program factories from the legacy `ProgramDescriptor` API to the Metal 2.0 `ProgramSpec` / `ProgramRunArgs` API:

- `ConcatProgramFactory` — interleaved concat (row-major + tiled)
- `ConcatS2SRMProgramFactory` — sharded→sharded row-major width concat
- `ConcatS2STiledProgramFactory` — sharded→sharded tiled width concat (with groups)

The remaining 4 factories (`S2SMulti`, `S2I`, `BlockSharded`, `TiledUnaligned`) stay on the legacy API. The device-op framework dispatches each variant alternative independently through C++20 concepts, so a `program_factory_t` variant mixing legacy `create_descriptor` and ported `create_program_artifacts` factories is fully supported; the selected factory is cached per index. Op selection is unchanged.

### What changed (per factory, mechanical / behavior-preserving)

- Entry point `create_descriptor` → `create_program_artifacts`, returning `ProgramArtifacts { spec, run_params }`.
- `CBDescriptor` → `DataflowBufferSpec`: tensor-backed CBs become `borrowed_from` DFBs, scratch CBs become plain DFBs.
- Positional args → **named** args (`get_arg(args::NAME)`); buffer-index args dropped (now DFB bindings); per-tensor arrays become runtime/compile-time varargs; interleaved reader uses a tensor binding sequence.
- Kernels updated to the Metal 2.0 kernel-args API (`dfb::`, `tensor::`, `get_vararg`, `get_compile_time_vararg`). Interleaved writers now bind the shared `writer_unary_*_metal2.cpp` donor kernels.

Legacy behavior is preserved where Metal 2.0 defaults differ: compute pinned to `opt_level=O3`, explicit `UnpackToSrc` on the Float32 tiled path, and the per-group compile-time / per-RISC page-range splits kept intact. No functional or numerical change intended.

### Testing

Covered by existing `tests/ttnn/unit_tests/operations/data_movement/test_concat.py` (interleaved, sharded, sharded-with-groups, fp32, program-cache). Needs to pass on Wormhole and Blackhole.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:akertesz/concat-port-test)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:akertesz/concat-port-test)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:akertesz/concat-port-test)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:akertesz/concat-port-test)


ttnn/cpp/ttnn/operations/data_movement/concat/device/kernels/dataflow/reader_concat_interleaved_start_id.cpp
```diff
@@ -8,41 +8,41 @@
...
 
-    tt_l1_ptr uint32_t* arg_ptr = (tt_l1_ptr uint32_t*)get_arg_addr(src_addr_base_idx);
+    // Two num_tensors-element runtime vararg blocks, in the order the host supplies them:
+    // num_tiles_per_block first, then tile_id_per_tensor.
+    constexpr uint32_t tile_id_per_tensor_offset = num_tensors;
     for (uint32_t i = 0; i < num_tensors; ++i) {
-        num_tiles_per_block[i] = arg_ptr[num_tiles_per_block_base_offset + i];
-        tile_id_per_tensor[i] = arg_ptr[tile_id_per_tensor_offset + i];
+        num_tiles_per_block[i] = get_vararg(i);
+        tile_id_per_tensor[i] = get_vararg(tile_id_per_tensor_offset + i);
```
gajanan-choudhary, line 38: `get_vararg` stuff here looks fine to me!

ttnn/cpp/ttnn/operations/data_movement/concat/device/kernels/dataflow/reader_concat_stick_layout_interleaved_start_id.cpp
```diff
@@ -67,7 +66,7 @@ void kernel_main() {
         }
         curr_tensor = 0;
 #else
-        auto page_size = kernel_compile_time_args[page_size_base_idx + curr_tensor];
+        auto page_size = get_compile_time_vararg(curr_tensor);
```
gajanan-choudhary, line 69: The port log contains a "complaint" that Claude had to figure out on its own that the `vararg` offset starts from `0` and that `page_size_base_idx` should be removed: https://github.com/tenstorrent/tt-metal/blob/97f7633ea1c02594dc89abd3d3b7bec2f8b840c8/ttnn/cpp/ttnn/operations/data_movement/concat/METAL2_PORT_REPORT.md?plain=1#L180-L193

 @akerteszTT, I don't know if you want to update the recipe with that instruction. (not a blocker here of course)

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_program_factory.cpp
```diff
@@ -267,34 +381,49 @@ tt::tt_metal::ProgramDescriptor ConcatProgramFactory::create_descriptor(
...
-    desc.kernels.push_back(std::move(reader_desc));
-    desc.kernels.push_back(std::move(writer_desc));
-    return desc;
+    ProgramRunArgs run_args;
+    // push_back rather than brace-init: an initializer_list would copy the per-node arg tables.
+    run_args.kernel_run_args.reserve(2);
+    run_args.kernel_run_args.push_back(std::move(reader_run_args));
+    run_args.kernel_run_args.push_back(std::move(writer_run_args));
+    for (uint32_t i = 0; i < num_input_tensors; ++i) {
+        run_args.tensor_args.emplace(INPUTS[i], inputs[i].get());
```
llongTT, line 419: **Correctness (silent wrong results): a cache entry created by `concat([x, x])` binds both inputs to argument 0, so a later `concat([a, b])` with identical specs returns `concat(a, a)`.**

How it happens:

1. The program-cache key covers only op attributes, tensor specs and the factory index, so `concat([x, x])` and `concat([a, b])` share one cached program. That part is by design.
2. On the cache miss, `SpecBindingSupport::resolve_bindings` (`ttnn/api/ttnn/mesh_device_operation_adapter.hpp`) turns each `TensorArgument` into a position in the argument list by searching for its `MeshTensor` object and taking the first match. When the same tensor handle is passed twice, both `input_0` and `input_1` resolve to position 0, and that table is frozen into the cache entry.
3. On the cache hit, `refresh_tensor_args` rebinds `input_1` from position 0, which is now `a`. Tensor `b` is never bound anywhere.
4. Nothing catches it. `ValidateTensorArgs` only compares the bound tensor's spec with the declared one (`a` and `b` have the same spec), and it is skipped by default (`validate_program_args = false`). On device, reading `a` twice is an ordinary memory access.

The legacy `Buffer*` patcher avoided this: `program_descriptor_patching.cpp` bails to empty bindings when a buffer repeats among the inputs (the `matmul(X, X)` comment), which forces a full argument rebuild on every hit of that entry. That is why `main` is correct for the same call sequence.

Reproduced on Blackhole with this branch for all three ported factories: this one, and the equivalent bindings in `concat_s2s_rm_program_factory.cpp` (lines 278-279) and `concat_s2s_tiled_program_factory.cpp` (lines 398-399). All four factory paths (interleaved tiled, interleaved row-major, height-sharded row-major, height-sharded tiled) return `concat(a, a)`. A three-input variant, `concat([x, y, x])` followed by `concat([a, b, c])`, returns `[a, b, a]`, so only the aliased slot goes stale. The reverse order (`[a, b]` first, then `[x, x]`) is correct, which means whether a process is affected depends on call order.

The gap is in the shared spec adapter rather than in this factory, but concat is the op most likely to hit it: `repeat_interleave` pushes the same handle up to 32 times per concat, and the diffusion demos do `concat([latents, latents])` for guidance. After one such call, every later same-shaped concat of two distinct tensors in the process is wrong.

Suggested fix, either of:
- A custom `compute_program_hash` (this op already has one) that mixes in an aliasing signature, for example for each input the index of the first input sharing its `MeshTensor`. Then `[x, x]` and `[a, b]` get separate entries, each correct.
- A duplicate check in `resolve_bindings` that mirrors the legacy bail-out, which fixes every spec-ported multi-input op at once.

A pytest file that pins this (and the sharded-size regression noted on `concat_s2s_rm_program_factory.cpp`) is in the review summary. It passes on `main` and fails on this branch.

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_s2s_rm_program_factory.cpp
```diff
@@ -34,172 +40,249 @@ static CoreRangeSet cores_to_corerangeset(const std::vector<CoreCoord>& cores) {
...
-            .format_descriptors = {{CBFormatDescriptor{
-                .buffer_index = static_cast<uint8_t>(input_id),
-                .data_format = cb_data_format,
-                .page_size = input_page_size,
-            }}},
-            .buffer = input_tensors[input_id].buffer(),
+        dataflow_buffers.push_back(DataflowBufferSpec{
+            .unique_id = input_dfb_names[input_id],
+            .entry_size = input_page_size,
+            .num_entries = num_input_units,
```
llongTT, line 94: **Regression (loud failure): a height-sharded tensor smaller than one shard is now rejected at spec validation.**

The borrowed DFB is sized to the full shard (`num_entries` is the shard row count here; the tiled factory does the same with tiles per shard at lines 128-135). `ProgramSpec` validation (`program_spec.cpp`, the "is larger than its borrowed TensorParameter" check) compares that against `TensorSpec::compute_packed_buffer_size_bytes()`, the tensor's total size across all cores, because no `Buffer` exists at spec time. Any tensor with two or more shards passes, since its total is at least one shard. A tensor with a single partial shard fails, even though the precise per-bank check in `AttachBorrowedDFBBuffers` would pass.

Example: two bf16 row-major tensors of shape (1,1,30,64), height-sharded with shard (32,64) on a 1-core or 2-core grid, concatenated on dim 3:

```
TT_FATAL: DFB 's2s_rm_input_0' (entry_size 128 * num_entries 32 = 4096 bytes) is larger than its borrowed TensorParameter 'input_0' (3840 bytes).
```

Row-major sharded layouts are aligned in width only, so 30 rows pack to 3840 bytes while the shard allocation is 4096. The tiled factory fails the same way for (1,1,32,64) with shard (64,64): 8192 vs 4096 bytes. The legacy factory compared the CB size against `Buffer::aligned_size_per_bank()` and ran these cases correctly; verified on `main` for both layouts on 1 and 2 cores.

This is narrow (the tensor must be smaller than its own shard) and it fails loudly rather than producing wrong data, but it is a valid `TensorSpec` that used to work. Suggested fix: clamp `num_entries` to the tensor's actual rows/tiles when the grid holds a single shard (a run-time DFB override does not help, because the check runs on the spec values), or have the framework's spec-time check use the per-shard size for sharded specs.

llongTT (commented):
Two correctness findings from a review of this PR, both reproduced on Blackhole (this branch vs `main`), with a pytest file that pins them.

**1. Input aliasing poisons the program-cache entry (silent wrong results).** `concat([x, x])` followed by `concat([a, b])` of the same spec returns `concat(a, a)` on all three ported factories. Details and suggested fixes in the inline comment on `concat_program_factory.cpp`.

**2. A height-sharded tensor smaller than one shard is now rejected (loud failure).** The borrowed DFBs are sized to the shard but validated against the tensor's total packed size. Details in the inline comment on `concat_s2s_rm_program_factory.cpp`.

Results of the test file below:

| Test | `main` | this branch |
|---|---|---|
| `test_alias_then_distinct_inputs` (4 factory paths) | pass | fail, output equals `concat(a, a)` |
| `test_distinct_then_alias_inputs` (control, reverse order) | pass | pass |
| `test_partial_alias_three_inputs` | pass | fail, output equals `concat(a, b, a)` |
| `test_single_partial_shard` (row-major and tiled, 1 and 2 cores) | pass | fail, borrowed-DFB size check aborts |

Proposed location: `tests/ttnn/unit_tests/operations/data_movement/test_concat_cache_aliasing.py`.

<details>
<summary>test_concat_cache_aliasing.py</summary>

```python
#

"""
Regression tests for two program-cache hazards in the Metal 2.0 concat program factories.

1. Input aliasing (silent wrong results).
   The program-cache key holds only tensor specs, so concat([x, x]) and concat([a, b]) share
   one compiled program. On a cache hit the framework re-points that program at the new
   tensors using a mapping recorded when the entry was created. That mapping must identify
   each input by its position in the argument list. If the entry created by concat([x, x])
   records both inputs as "argument 0", a later concat([a, b]) reads `a` twice and never
   touches `b`, with no error: the only hit-time check compares tensor specs, which match.

2. Single partial shard (loud failure).
   A height-sharded tensor whose total row count is smaller than one shard is a valid
   TensorSpec. The sharded two-tensor factories size their borrowed L1 buffers from the
   shard, so a spec-time size check against the tensor's total packed size must not reject
   the call.
"""

import pytest
import torch

import ttnn


@pytest.fixture
def isolate_program_cache(device):
    device.disable_and_clear_program_cache()
    device.enable_program_cache()
    yield
    device.disable_and_clear_program_cache()


def _grid(num_cores):
    return ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(num_cores - 1, 0))})


def _height_sharded(num_cores, shard_shape):
    return ttnn.MemoryConfig(
        ttnn.BufferType.L1,
        ttnn.ShardSpec(_grid(num_cores), shard_shape, ttnn.ShardOrientation.ROW_MAJOR),
    )


# One row per concat program factory reachable with two inputs of identical spec.
ALIAS_CASES = {
    "interleaved_tile_dim0": dict(
        shape=(1, 1, 64, 64),
        layout=ttnn.TILE_LAYOUT,
        dim=0,
        in_mem=ttnn.DRAM_MEMORY_CONFIG,
        out_mem=ttnn.DRAM_MEMORY_CONFIG,
    ),
    "interleaved_rm_dim2": dict(
        shape=(1, 1, 32, 64),
        layout=ttnn.ROW_MAJOR_LAYOUT,
        dim=2,
        in_mem=ttnn.DRAM_MEMORY_CONFIG,
        out_mem=ttnn.DRAM_MEMORY_CONFIG,
    ),
    "height_sharded_rm_dim3": dict(
        shape=(1, 1, 64, 64),
        layout=ttnn.ROW_MAJOR_LAYOUT,
        dim=3,
        in_mem=_height_sharded(2, (32, 64)),
        out_mem=_height_sharded(2, (32, 128)),
    ),
    "height_sharded_tile_dim3": dict(
        shape=(1, 1, 64, 64),
        layout=ttnn.TILE_LAYOUT,
        dim=3,
        in_mem=_height_sharded(2, (32, 64)),
        out_mem=_height_sharded(2, (32, 128)),
    ),
}


def _to_device(device, torch_tensor, case):
    return ttnn.from_torch(torch_tensor, layout=case["layout"], device=device, memory_config=case["in_mem"])


def _check_concat(device, torch_inputs, tt_inputs, case, context):
    """Run concat on tt_inputs and compare against torch.cat(torch_inputs)."""
    out = ttnn.concat(tt_inputs, dim=case["dim"], memory_config=case["out_mem"])
    got = ttnn.to_torch(out)
    expected = torch.cat(torch_inputs, dim=case["dim"])
    if not torch.equal(got, expected):
        # Diagnose the aliasing signature: every slot filled from the first input.
        first_only = torch.cat([torch_inputs[0]] * len(torch_inputs), dim=case["dim"])
        pytest.fail(
            f"{context}: wrong result; output equals concat of the first input repeated: "
            f"{torch.equal(got, first_only)}"
        )


@pytest.mark.parametrize("case_name", ALIAS_CASES.keys())
def test_alias_then_distinct_inputs(device, isolate_program_cache, case_name):
    """concat([x, x]) creates the cache entry; concat([a, b]) must then read both a and b."""
    case = ALIAS_CASES[case_name]
    torch.manual_seed(0)

    x_t = torch.randn(case["shape"]).bfloat16()
    x = _to_device(device, x_t, case)
    _check_concat(device, [x_t, x_t], [x, x], case, "priming call concat([x, x])")
    entries_after_prime = device.num_program_cache_entries()

    a_t = torch.randn(case["shape"]).bfloat16()
    b_t = torch.randn(case["shape"]).bfloat16()
    a = _to_device(device, a_t, case)
    b = _to_device(device, b_t, case)
    _check_concat(device, [a_t, b_t], [a, b], case, "concat([a, b]) after concat([x, x])")
    assert device.num_program_cache_entries() == entries_after_prime, "second call was expected to be a cache hit"


@pytest.mark.parametrize("case_name", ALIAS_CASES.keys())
def test_distinct_then_alias_inputs(device, isolate_program_cache, case_name):
    """Control for the test above: the reverse order is correct by construction."""
    case = ALIAS_CASES[case_name]
    torch.manual_seed(0)

    a_t = torch.randn(case["shape"]).bfloat16()
    b_t = torch.randn(case["shape"]).bfloat16()
    _check_concat(device, [a_t, b_t], [_to_device(device, a_t, case), _to_device(device, b_t, case)], case, "priming")
    entries_after_prime = device.num_program_cache_entries()

    x_t = torch.randn(case["shape"]).bfloat16()
    x = _to_device(device, x_t, case)
    _check_concat(device, [x_t, x_t], [x, x], case, "concat([x, x]) after concat([a, b])")
    assert device.num_program_cache_entries() == entries_after_prime, "second call was expected to be a cache hit"


def test_partial_alias_three_inputs(device, isolate_program_cache):
    """concat([x, y, x]) then concat([a, b, c]): a stale position table yields [a, b, a]."""
    case = ALIAS_CASES["interleaved_tile_dim0"]
    torch.manual_seed(0)

    x_t = torch.randn(case["shape"]).bfloat16()
    y_t = torch.randn(case["shape"]).bfloat16()
    x = _to_device(device, x_t, case)
    y = _to_device(device, y_t, case)
    _check_concat(device, [x_t, y_t, x_t], [x, y, x], case, "priming call concat([x, y, x])")
    entries_after_prime = device.num_program_cache_entries()

    a_t, b_t, c_t = (torch.randn(case["shape"]).bfloat16() for _ in range(3))
    a, b, c = (_to_device(device, t, case) for t in (a_t, b_t, c_t))
    out = ttnn.concat([a, b, c], dim=case["dim"], memory_config=case["out_mem"])
    got = ttnn.to_torch(out)
    assert device.num_program_cache_entries() == entries_after_prime, "second call was expected to be a cache hit"
    if not torch.equal(got, torch.cat([a_t, b_t, c_t], dim=case["dim"])):
        stale = torch.equal(got, torch.cat([a_t, b_t, a_t], dim=case["dim"]))
        pytest.fail(f"concat([a, b, c]) after concat([x, y, x]) returned wrong data; equals concat(a, b, a): {stale}")


@pytest.mark.parametrize("num_cores", [1, 2])
@pytest.mark.parametrize(
    "layout, shape, in_shard, out_shard",
    [
        (ttnn.ROW_MAJOR_LAYOUT, (1, 1, 30, 64), (32, 64), (32, 128)),
        (ttnn.TILE_LAYOUT, (1, 1, 32, 64), (64, 64), (64, 128)),
    ],
    ids=["rm_30_rows_shard_32", "tile_32_rows_shard_64"],
)
def test_single_partial_shard(device, layout, shape, in_shard, out_shard, num_cores):
    """A height-sharded tensor smaller than one shard must be accepted by the sharded factories."""
    torch.manual_seed(0)
    a_t = torch.randn(shape).bfloat16()
    b_t = torch.randn(shape).bfloat16()
    in_mem = _height_sharded(num_cores, in_shard)
    a = ttnn.from_torch(a_t, layout=layout, device=device, memory_config=in_mem)
    b = ttnn.from_torch(b_t, layout=layout, device=device, memory_config=in_mem)

    out = ttnn.concat([a, b], dim=3, memory_config=_height_sharded(num_cores, out_shard))
    assert torch.equal(ttnn.to_torch(out), torch.cat([a_t, b_t], dim=3))
```

</details>

_Review prepared with Claude Code._

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_program_factory.cpp
gajanan-choudhary: Fixed in 251ff951218. compute_program_hash now folds an input-aliasing signature (for each input, the index of the first input backed by the same MeshTensor), so aliased and distinct call patterns get separate program-cache entries and each rebinds correctly — concat([a,b]) after concat([x,x]) is now a miss instead of a wrong-data hit. Verified on Wormhole: the repro no longer returns concat(a,a) for any of the three ported factories.

This is an op-local fix. The root cause is the missing aliasing bail-out in the shared spec-path resolve_bindings, which still exposes other base-concept multi-input ports; that general fix is tracked separately.

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_s2s_rm_program_factory.cpp
gajanan-choudhary: Resolved on the current base by the framework fix #56780 ("Validate borrowed DFBs against the per-bank allocation, not the packed tensor size"). After rebasing this branch onto latest main, the single-partial-shard cases pass with no op-local change — both the row-major factory here and the tiled factory (verified on Wormhole, 1- and 2-core grids). No factory-side clamp is needed.

llongTT (approved):

PR #54669, opened 2026-08-27
BGE-M3: Data Parallel Optimization on N300

gtobarTT:
### Ticket

N/A

### What's changed

BGE-M3 had no data-parallel serving path for one N300. The long-context shape
(batch 12, sequence length 8192) ran on one chip and took about 3.2 s. This
change adds that path and brings the same shape to 989 ms.

Each chip runs an independent replica of batch 6 over the full 8192 sequence.
The path uses no inter-chip collectives.

The change adds these optimizations:

- **QKV scatter.** One program runs the QKV projection, the head split, and
  the BF4 conversion. This removes 24 head-split programs per pass.
- **Query head-fold.** The path folds 2 query chunks into the head dimension,
  so SDPA reads 4096 queries per head instead of 8192. The GQA head-broadcast
  keeps every query head over the full sequence, so the result stays exact.
- **Encoder SDPA.** A model-local descriptor runs a q256/k2048 chunk plan with
  BF4 K and V. The writer emits the concat-head order, so the separate
  concat-heads program is no longer needed.
- **Compact valid-length masking.** The kernels read the valid length of each
  request and build only the boundary mask tiles. They skip a key chunk that
  holds only padding, so a short request completes sooner.
- **Folded attention scale.** The build folds the scale into the Q weight.
  SDPA then runs with scale 1.0 and drops the multiply in every layer.
- **LoFi LayerNorm and BF4 MLP** for the serving shape.

`ModelArgs` sets `use_jit` for the B12/S8192 data-parallel shape and maps the
serving options, so a caller passes `data_parallel` alone. `encoder.py` reads
that flag and selects `BgeM3AttentionJit` and `BgeM3MLPJit`. The base classes
keep the S512 behavior.

**Entry points**

| Command | Purpose |
| --- | --- |
| `pytest tests/perf/perf.py::test_perf -s` | wall-clock latency |
| `pytest tests/perf/tracy_perf.py -k n300_dp_tracy` | per-op device profile |
| `pytest tests/pcc/model_dp.py::test_model_dp_2 -s` | accuracy against HF |
| `python demo/demo_traced.py --batch 12 --seq-len 8192 --data-parallel` | demo |
| `python demo/mteb_eval_minimal.py --mode both` | MTEB scores |

`test_perf` reads the card while pytest collects the tests. An N300 runs the
B12/S8192 data-parallel shape on a (2, 1) mesh with fabric. Every other card
runs the S512 batch sweep on one device.

### Checklist

Measured on one N300 (Wormhole, 2 chips), `bfloat8_b`:

| Metric | Value |
| --- | --- |
| Latency, 8192 tokens | 988.7 ms (12.1 embeddings/s) |
| Latency, 128 to 4096 tokens padded to 8192 | 535 to 675 ms |
| PCC against HF, unmasked | 0.8644 |
| PCC against HF, masked CLS | 0.9352 |
| MTEB STSBenchmark | 0.8132 against 0.8485 on HF |
| MTEB SICK-R | 0.7821 against 0.7971 on HF |

### Notes for reviewers

1. The `encoder_sdpa` kernels started as a copy of
   `ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/compute/`. That
   copy drifted, so the second commit re-syncs four SFPU call signatures and
   deletes a 2739-line copy of the streaming pipeline that the serving path
   never ran. The copy will drift again. I would like input on whether these
   kernels should include the ttnn headers instead of copying them.

2. `optimizations.py` still carries configuration that only the S512 and
   Blackhole shapes use. Removing it needs those shapes re-tested, so it is
   left for a later change.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:gtobarTT/bge_m3_n300_dp2)


dvartaniansTT (approved):
LGTM

PR #54675, opened 2026-08-27
conv2d: exact fp32 depthwise conv1d via SFPU accumulation

### LINKED ISSUE #54877: Pushes fp32 precision into the depthwise conv1d kernel
Pushes fp32 precision into the depthwise conv1d kernel (SFPU accumulation + UnpackToDestFp32), making it bit-equal to the exact MAC  form at 2.1–4.1x MAC's speed — so the prefer_mac model-code workaround is deleted. Worth ~548 ms/decode in accurate mode.

rouzbehTT:
## Summary
`ttnn.conv1d`'s depthwise path (used by MiniMax-H3's audio decoder) truncates fp32 operands through SrcA/SrcB (~9/13 significand bits regardless of MathFidelity), measuring 1.6e-03 rel RMSE against a float64 golden. The workaround was an exact but slower elementwise MAC fallback, opted into via `prefer_mac`.

Two commits:
1. **Kernel fix**: route the multiply/accumulate through the SFPU when both operands are fp32, and deliver ACT/ACT_TILIZED/WEIGHTS/scratch as `UnpackToDestFp32` so the unpacker doesn't round-trip through TF32 first. Dtype-gated at compile time — no flag, no change for bf16/bf8 callers (LTX included).
2. **Retire `prefer_mac`**: with conv1d now exact, the lever selects nothing. Removed from `depthwise_tap_filter` and the constructor chain; MAC stays only as the fallback for shapes conv1d cannot configure.

## Measured (hardware, float64 golden, four production depthwise shapes)

| shape | conv1d before | conv1d after | MAC | conv1d / MAC time |
|---|---|---|---|---|
| C=512,K=12,s=2 | 1.588e-03 | 7.946e-08 | 7.946e-08 | 0.69 / 2.81 ms (4.1x) |
| C=16,K=7,s=1 | 1.675e-03 | 6.256e-08 | 6.256e-08 | 0.76 / 1.88 ms (2.5x) |
| C=8,K=7,s=1 | 1.442e-03 | 6.190e-08 | 6.190e-08 | 0.52 / 1.09 ms (2.1x) |
| C=8,K=12,s=2 | 1.614e-03 | 7.585e-08 | 7.585e-08 | 0.67 / 2.41 ms (3.6x) |

conv1d now matches the MAC fallback bit-for-bit at every shape, at 2.1–4.1x its speed.

## Tests
`test_decode` (real weights, accurate-mode PSNR/rmse/PCC bars) and `test_depthwise_chunked_conv1d_matches_torch` pass through the conv1d path.

## Not in scope
`split_mode`/`tap_matmul` work around the same truncation in the conv3d path; retiring those needs an equivalent conv3d kernel fix.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rouzbeh/improve-audio-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rouzbeh/improve-audio-kernels)


cglagovichTT (approved):

rouzbehTT:
Addressed the review comments in aeb4cdf5241:

- **Stale DST comment** (x2): fixed — the block-SFPU form uses 2 DST tiles, not 3.
- **fp32_operands not gating on the scratch CB**: the gate now also checks the scratch format. Two findings from verifying the claim: (1) the divergence scenario is structurally impossible today — ACT_TILIZED and the dest-reuse scratch both take the output data format in `conv2d_op_program_factory_common.cpp`; (2) even under host/kernel gate divergence (fp32 inputs, non-fp32 output), `UnpackToDestFp32` on a non-Float32 CB is a no-op by construction (`get_unpack_dst_formats` only honors the mode when the CB format is Float32), and the kernel independently falls back to the FPU path. The check is added as future-proofing, with a comment saying so.
- **Test coverage**: added three regressions to `test_conv1d.py` with true fp32 inputs (not bf16-widened), bounding rel RMSE at 5e-6 vs a float64 golden — the SFPU path measures ~7e-8, the old FPU path ~1.5e-3, so an unpack-mode regression fails by ~300x. Covers the coalesced layout (C=64 K=12), the non-coalesced scratch layout (C=512 K=12 s=2, exceeds every arch's NoC burst), and bf16+`fp32_dest_acc_en` staying on the FPU path. All three pass on Blackhole.

Drive-by finding while writing these: depthwise conv1d with dilation>1 returns wrong data on this route (rel RMSE ~1.4 vs golden), identically on the old FPU path with bf16 — pre-existing, unrelated to this PR (output length honors dilation; the data layout doesn't). No in-tree caller hits it; can file separately.

ttnn/cpp/ttnn/operations/conv/conv2d/device/kernels/compute_depthwise_conv1d.cpp
```diff
@@ -147,6 +271,17 @@ void kernel_main() {
     // height block (in-place) or at a dedicated scratch CB for multiple blocks.
     constexpr uint32_t partials_cb_id = get_compile_time_arg_val(11);
 
+    // SFPU tap accumulation only when both operands are genuinely fp32 -- derived purely from the
+    // per-CB unpack format, no new compile-time arg or host flag. Gating on format (not just
+    // DST_ACCUM_MODE) keeps every bf16/bf8 caller, LTX included, bit-exact and on the faster FPU path.
+    // The scratch check is structurally redundant today (ACT_TILIZED and the dest-reuse scratch both
+    // take the output data format in the CB table), but keeps this gate correct if that ever diverges:
+    // the SFPU path reads the scratch through `copy_tile` and needs it delivered UnpackToDestFp32.
```
pavlejosipovic: Could we pass `use_fp32_sfpu_depthwise` as a compile-time kernel arg from the factory and use it both for the unpack overrides and kernel dispatch? The current CB-format inference duplicates the host eligibility condition and can misclassify mixed-dtype calls because `ACT_TILIZED` follows output dtype. An explicit arg makes the selected arithmetic contract unambiguous.

rouzbehTT: Done, the factory now computes use_fp32_sfpu_depthwise once and it drives both the unpack-to-dest overrides and the kernel dispatch (compile arg 12). Also, the kernel-side format inference is removed.

Good call on the mixed-dtype case: the condition now also requires the output format to be fp32. ACT_TILIZED/scratch follow output dtype, so the SFPU contract needs it. Also, any mixed call falls back to the stock FPU path unchanged.

Also, validated on BH: full conv1d suite green, fp32-exact gates and the bf16-unchanged regression included.

pavlejosipovic (approved):

pavlejosipovic (approved):

rouzbehTT:
Ran both of the CIs:
Nightly passes without issue
https://github.com/tenstorrent/tt-metal/actions/runs/33544779561

A few of the sanity ones fail 
https://github.com/tenstorrent/tt-metal/actions/runs/33544775590
but they are inherited from failure on the main branch
https://github.com/tenstorrent/tt-metal/actions/runs/33546202729

Issue #53394, opened 2026-08-17
TT-NOP injector tool race findings - LLK


nmohammedTT:
Parent tracker for timing races found by the TT-NOP injector (tt_metal/tt-llk/tests/python_tests/ttnop).

Unperturbed tests pass. These only fail when a delay is inserted at a sync site.

**Pinned tree** (findings were taken from this commit; later pushes to the branch do not change it):
- Branch: [`nmohammed/timing-perturbations-tt-llk-ci`](https://github.com/tenstorrent/tt-metal/tree/nmohammed/timing-perturbations-tt-llk-ci)
- Commit: [`1cfdc6227eb`](https://github.com/tenstorrent/tt-metal/commit/1cfdc6227ebf76390354891a3141cb136a636dfe) (`updated NOP count`)

Struck-through rows have a merged fix. Open PRs are left as-is.

## Blackhole

- SFPI: 7.68.0
- Reservation: yyzo-bh-02-special-nmohammed-for-reservation-183021

| Case | NOP | Symptom |
| --- | --- | --- |
| ~~test_generalized_moe_gate.py::test_generalized_moe_gate_dest_sections[what:(3, 2)]~~ | ~~tti_nop~~ | ~~hang~~ (#54483) |
| test_mul_reduce_scalar.py::test_mul_reduce_scalar[formats:Float16_b->Float16_b-math_fidelity:HiFi2-num_tiles:1-tile_dimensions:[16, 16]] | unpacr0 | mismatch |
| test_mul_reduce_scalar.py::test_mul_reduce_scalar[formats:Float16_b->Float16_b-math_fidelity:HiFi2-num_tiles:1-tile_dimensions:[16, 16]] | unpacr0, unpacr1 | hang (math thread; injector artifact, #54689) |
| test_sfpu_binary.py::test_sfpu_binary_int_uniform[mathop:SfpuLcm-dest_acc:Yes] | tti_nop | hang |
| test_transpose_dest.py::test_transpose_dest_float[fmt_dest_acc_math_transp_unpack_to_dest:(InputOutputFormat[A:Float32,B:Float32,out:Bfp8_b], dest_acc=Yes, transpose=Yes, unpack_to_dest=True)] | tti_nop | hang |
| test_fused.py::test_fuser[tiny_tiles_multi_op] | unpacr1 | hang |
[...]

PR #52256, opened 2026-08-05
[Bug fix] Fix Blackhole destination-reuse synchronization

jasondavies:
### PR Category

Bug fix

### Summary

- Closes https://github.com/tenstorrent/tt-metal/issues/52252
- Closes https://github.com/tenstorrent/tt-metal/issues/46523

Blackhole destination reuse uses dummy unpack operations to publish source banks consumed by `MOVD2A`/`MOVD2B`. The math-side wait previously checked only source `DVALID`. Because a preceding math instruction or `SETRWC` can release and advance the source bank one or two cycles after the wait is processed, the wait can observe stale state and allow a dummy source clear to collide with a destination-to-source move.

This PR adds the `MATH` condition to the existing `SRCA_VLD`/`SRCB_VLD` waits. Draining preceding FPU-pipeline instructions first makes their bank release visible before the validity of the bank selected for `MOVD2A`/`MOVD2B` is evaluated.

It also sets `Stall_Clr_Cntrl=1` on destination-reuse dummy unpacks. This is the preferred Blackhole operating mode: it gates the dummy operation using the unpacker-selected bank and allows unpack to prepare the next bank while math consumes the current one. It is complementary to, rather than a replacement for, the explicit math-pipeline drain.

This PR:

- Drains preceding math instructions before waiting for the source bank used by `MOVD2A`/`MOVD2B`.
- Sets `Stall_Clr_Cntrl` on destination-reuse dummy unpack operations so unpack can operate on the next bank.
- Fixes regular `DEST_TO_SRCA` and `DEST_TO_SRCB` reuse.
- Fixes the supported column-broadcast `DEST_TO_SRCA` path, including FP32 destination accumulation.
- Fixes row-broadcast `DEST_TO_SRCA`, including supported tile shapes with fewer than four faces.
- Centralizes the race-sensitive dummy-unpack encoding so all supported paths use the same synchronization.
- Re-enables the Blackhole FP32 Welford layernorm cases previously skipped for #46523.

### Notes for reviewers

`move_d2a_fixed_face()` and `move_d2b_fixed_face()` now wait on `MATH | SRCA_VLD` and `MATH | SRCB_VLD`, respectively. `STALL_MATH` prevents new FPU instructions from entering while the `MATH` condition drains older ones; the source-valid condition is then evaluated after any delayed bank release or pointer advance has taken effect.

[...]

PR #54683, opened 2026-08-28
[LLK] Gate mul_reduce_scalar's Dest->Src moves on source bank-valid

### LINKED ISSUE #53412: [TT-NOP Injector][BH] mismatch race on mul_reduce_scalar
Sub-issue of #53394 (TT-NOP injector tool race findings - LLK).

## Summary

Blackhole `mul_reduce_scalar` mismatches when unpack sync is delayed. Same site across the HiFi / tile-count / tile-shape sweep.

## Recreate

Branch: [`nmohammed/timing-perturbations-tt-llk`](https://github.com/tenstorrent/tt-metal/tree/nmohammed/timing-perturbations-tt-llk)
Commit: [`1ec846ac199`](https://github.com/tenstorrent/tt-metal/commit/1ec846ac19961585cecd2fee44e9bed5321e27bd) (`Add the ttnop focus tool for timing-perturbation triage`)
```bash
cd tt_metal/tt-llk/tests/python_tests/ttnop
CHIP_ARCH=blackhole ./focus.sh --site sync --sites unpack:3 --nop tti_nop \
  --delays 5-100 --repeats 5 \
  'test_mul_reduce_scalar.py::test_mul_reduce_scalar[formats:Float16_b->Float16_b-math_fidelity:HiFi2-num_tiles:1-tile_dimensions:[16, 16]]'
```

amahmudTT:
Fixes #53412 (sub-issue of #53394, TT-NOP injector race findings).

## The race

`MOVD2A` / `MOVD2B` write SrcA/SrcB **from Dest**. The Src auto-wait covers instructions that *read* Src, so these moves fall outside it and need an explicit bank-valid gate. The reduce-phase moves had **no gate at all** — not a wrong mask, absent:

- `_llk_math_mul_reduce_scalar_move_dest_to_src_` — 16x `MOVD2A` (`DEST_TO_SRCA`) and 16x `MOVD2B` (`DEST_TO_SRCB`)
- `_llk_math_mul_reduce_scalar_` — the row-0 `MOVD2B` into the SrcB scratch rows

They race `_llk_unpack_mul_reduce_scalar_switch_to_reduce_`'s `UNPACR_NOP(SrcX, SET_DVALID, UNP_ZEROSRC)`. When MATH wins, it fills the bank from Dest and the unpacker's **`UNP_ZEROSRC` then zeroes the rows just written**, so `GAPOOL` reduces zeros and the scalar comes out `0.0`.

## The fix

Adopt the idiom already three functions away in the same tree — `cmath_common.h`'s `move_d2a_fixed_face` / `move_d2b_fixed_face`, whose own comment states it outright ("MOVD2A does not auto-wait for `SrcA[MatrixUnit.SrcABank].AllowedClient == MatrixUnit`, so gate on SRCA_VLD before the row moves"):

```c
TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);   // before the MOVD2A rows
TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);   // before the MOVD2B rows
```

The `MATH` condition additionally drains the multiply-phase Dest writes that the moves read back.

**Cost is bounded to one wait per reduce phase.** The unpacker posts the dummy DVALID once in `switch_to_reduce_`, and nothing clears it until the closing `CLEARDVALID` — so the gate blocks only the first move of a phase; the `move_dest_to_src<DEST_TO_SRCA>(i)` calls in the per-tile loop find the bank already valid and fall straight through. Checked every caller (`mul_reduce_scalar.h`, `sum_reduce_scalar.h`, `rmsnorm.h`, and the four tt-llk test sources) — all pair `switch_to_reduce` with the moves before any `clear_dvalid`, so no call site can deadlock on the new wait.

## Also in this diff: the unpack-side bank wait (throughput, not part of the fix)

`llk_unpack_mul_reduce_scalar.h` (both arches) also switches its two dummy publications to the pipelined bank wait — `Stall_Clr_Cntrl=1` on Blackhole, `UNP_ZEROSRC_STALL_RESET_WR_RDY` on Wormhole. Note the shapes differ: BH carries `Set_Dvalid` and `Unpack_Pop=UNP_ZEROSRC` in one instruction, whereas WH needs two, and it is the `ZEROSRC` of the pair that takes the bit — the following bare `SET_DVALID` performs no wait of its own and inherits one by sequencing. **This is a throughput/parity change and is not what the validation below exercises.** Calling it out because the diff contains it and the rest of this description is about the math-side gate.

The bit selects only *which bank the publication waits on*: set, it gates on `Unpackers[i].SrcBank` — the bank it actually clears — so unpack can prepare the next bank while math consumes the current one; clear (the default), it gates on `MatrixUnit.Src?Bank`, which under the bank-pointer lockstep invariant is back with the unpackers only when *no* bank is outstanding. The default is therefore the **stronger, serializing** wait, and it implies the own-bank condition — so the old form over-waits, it does not under-wait. The bank cleared is `UnpackBank` either way.

This matches #52256, whose commit message calls `Stall_Clr_Cntrl=1` *"the preferred Blackhole operating mode ... complementary to, rather than a replacement for, the explicit math-pipeline drain."* The gate added above is that drain; this bit is the mode.

Both publications clear a single bank (`Bank_Clr_Ctrl=0`). That matters: with a both-banks clear the own-bank wait would cover only the bank being prepared and could overwrite one the Matrix Unit still owns, so the bit is valid only for single-bank publications.

The **36/36** result cited below is a no-regression result for this half, not fail-without/pass-with evidence — a pipelining change is expected to be behaviour-neutral. The Wormhole side of it is unexercised for the same reason the gate is (see the reviewer note at the end).

## Validation (Blackhole p100a, sfpi 7.72.0)

Fail-without / pass-with, using the issue's own recreate command:

```bash
CHIP_ARCH=blackhole ./focus.sh --site sync --sites unpack:3 --nop tti_nop \
  --delays 5-100 --repeats 5 \
  'test_mul_reduce_scalar.py::test_mul_reduce_scalar[formats:Float16_b->Float16_b-math_fidelity:HiFi2-num_tiles:1-tile_dimensions:[16, 16]]'
```

| tree | result |
| --- | --- |
| `origin/main` (pristine) | **96/96 variants mismatch**, 5/5 runs each, `device=0.0 golden=130.703125` |
| this PR | **0 findings**, every variant passes |
| control: this PR **minus only the SrcA gate** | failure returns, same site and signature |

The third row is there because "no findings" on its own does not prove the injector was still
firing. Dropping just the `SRCA_VLD` gate from the patched tree brings the mismatch straight back
at the same `SEMGET` site, so injection is live on the patched tree and the SrcA gate is the one
this failure needs.

`device=0.0` is the mechanism above showing through directly: the reduction ran on a zeroed bank.

Unperturbed suite: `test_mul_reduce_scalar.py` **54/54 pass** on Blackhole with the fix.

The delay band is 5-100 rather than the issue's original counts because NOP counts are harness-build-specific (the report was taken at an older sfpi pin); the site and the mechanism are the same.

## Wormhole

The two `llk_math_mul_reduce_scalar.h` files are byte-identical apart from the `ZEROACC` signature, so Wormhole carries the same missing gate, and WH's own `cmath_common.h` has the same `move_d2*_fixed_face` idiom. Fixed there too, deliberately, to avoid the recurring pattern of a Blackhole dest-reuse fix never reaching Wormhole.

Verified the WH edit is actually emitted rather than compiled away: the WH math kernel builds clean, and its `.text` grows from 3280 to 3332 bytes against pristine (`INSTRUCTION_WORD` embeds the Tensix word as a RISC-V immediate, so the instructions do not appear as standalone `.text` words).

**Reviewer note:** the Wormhole change is reasoned, compile-verified and binary-verified, but **not HW-validated** — there is no WH silicon on this box, and `test_mul_reduce_scalar.py` skips itself on non-Blackhole (`"mul_reduce_scalar is a Blackhole-only experimental LLK"`). Worth a WH run before merge if anyone has a board. Note the compute APIs `mul_reduce_scalar.h` / `sum_reduce_scalar.h` are *not* `ARCH_BLACKHOLE`-gated the way `rmsnorm.h` is, so the WH path is reachable live code, not dead.

## Not in this PR

The audit behind #53412 found two further defects on this path, left out to keep the fix reviewable. Neither is required for the mismatch above:

1. `_llk_math_mul_reduce_scalar_init_` rewrites `ADDR_MOD_0/1/2` with no drain while multiply-phase `ELWMUL`s are still in the FPU (`ApplyAddrMod` runs as the FPU instruction's last step and reads those words).
2. No SFPU->FPU Dest drain between `_calculate_fill_`'s `SFPSTORE`s and the following `MOVD2B`/`GAPOOL`; `DISABLE_IMPLIED_SRCA_FMT_Base` is also never set around the Dest->Src moves, though the ISA calls it strongly recommended on Blackhole.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:amahmud/bh-mul-reduce-scalar-sync)

### DIFF
--- a/tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
@@ -139,6 +139,19 @@ inline void incr_counters(const std::uint32_t incr_a, const std::uint32_t incr_b
     TT_INCRWC(incr_cr, incr_d, incr_b, incr_a);
 }
 
+// MOVD2A/MOVD2B write SrcA/SrcB from Dest, so they fall outside the Src auto-wait, which covers
+// only instructions that read Src. Gate the row moves on the target bank's DVALID, and drain
+// in-flight math so the Dest values those moves read back have settled.
+inline void srca_bank_wait()
+{
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
+}
+
+inline void srcb_bank_wait()
+{
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+}
+
 inline void move_d2a_fixed_face(const std::uint8_t addrmod)
 {
     // Drain preceding math instructions so their source-bank release is visible before testing
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_mul_reduce_scalar.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_mul_reduce_scalar.h
@@ -89,6 +89,10 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
             TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
         }
 
+        // MOVD2A writes SrcA from Dest and so is outside the Src auto-wait. Gate on the unpacker's
+        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
+
         switch (idst)
         {
             case 0:
@@ -121,6 +125,10 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
     }
     else if constexpr (binary_reuse_dest == EltwiseBinaryReuseDestType::DEST_TO_SRCB)
     {
+        // MOVD2B writes SrcB from Dest and so is outside the Src auto-wait. Gate on the unpacker's
+        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+
         TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_BD);
 
         TTI_MOVD2B(0, p_movd2b::SRC_ZERO_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_4_ROWS, 0);
@@ -247,6 +255,9 @@ inline void _llk_math_mul_reduce_column_(const std::uint32_t dst_index, const ck
 template <MathFidelity math_fidelity>
 inline void _llk_math_mul_reduce_scalar_()
 {
+    // Same gate as the DEST_TO_SRCB move: MOVD2B is outside the Src auto-wait.
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+
     // Copy row 0 from dest to srcB (rows 16-31 as scratch) and transpose
     TTI_MOVD2B(0, p_movd2b::SRC_ROW16_OFFSET, ADDR_MOD_0, p_movd2b::MOV_1_ROW, 0);
     TTI_GATESRCRST(0b1, 0b1);
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_mul_reduce_scalar.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_mul_reduce_scalar.h
@@ -89,9 +89,7 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
             TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
         }
 
-        // MOVD2A writes SrcA from Dest and so is outside the Src auto-wait. Gate on the unpacker's
-        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
-        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
+        math::srca_bank_wait();
 
         switch (idst)
         {
@@ -125,9 +123,7 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
     }
     else if constexpr (binary_reuse_dest == EltwiseBinaryReuseDestType::DEST_TO_SRCB)
     {
-        // MOVD2B writes SrcB from Dest and so is outside the Src auto-wait. Gate on the unpacker's
-        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
-        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+        math::srcb_bank_wait();
 
         TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_BD);
 
@@ -255,8 +251,7 @@ inline void _llk_math_mul_reduce_column_(const std::uint32_t dst_index, const ck
 template <MathFidelity math_fidelity>
 inline void _llk_math_mul_reduce_scalar_()
 {
-    // Same gate as the DEST_TO_SRCB move: MOVD2B is outside the Src auto-wait.
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+    math::srcb_bank_wait();
 
     // Copy row 0 from dest to srcB (rows 16-31 as scratch) and transpose
     TTI_MOVD2B(0, p_movd2b::SRC_ROW16_OFFSET, ADDR_MOD_0, p_movd2b::MOV_1_ROW, 0);
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_unpack_mul_reduce_scalar.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_unpack_mul_reduce_scalar.h
@@ -27,7 +27,9 @@ inline void _llk_unpack_mul_reduce_scalar_switch_to_reduce_()
 {
     TTI_SETADCZW(0b011, 0, 0, 0, 0, 0b1111);
     semaphore_post(semaphore::UNPACK_SYNC);
-    TTI_UNPACR_NOP(SrcA, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
-    TTI_UNPACR_NOP(SrcB, 0, 0, p_unpacr_nop::SET_DVALID, 0, 0, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
+    // Stall_Clr_Cntrl=1: wait on Unpackers[i].SrcBank -- the bank actually cleared and published --
+    // not MatrixUnit.Src?Bank, which is a different bank once double-buffering reaches steady state.
+    TTI_UNPACR_NOP(SrcA, 0, 0, p_unpacr_nop::SET_DVALID, 0, 1, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
+    TTI_UNPACR_NOP(SrcB, 0, 0, p_unpacr_nop::SET_DVALID, 0, 1, 0, 0, p_unpacr_nop::UNP_ZEROSRC);
     t6_semaphore_get(semaphore::UNPACK_SYNC);
 }
--- a/tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/cmath_common.h
+++ b/tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/cmath_common.h
@@ -138,6 +138,19 @@ inline void incr_counters(const std::uint32_t incr_a, const std::uint32_t incr_b
     TT_INCRWC(incr_cr, incr_d, incr_b, incr_a);
 }
 
+// MOVD2A/MOVD2B write SrcA/SrcB from Dest, so they fall outside the Src auto-wait, which covers
+// only instructions that read Src. Gate the row moves on the target bank's DVALID, and drain
+// in-flight math so the Dest values those moves read back have settled.
+inline void srca_bank_wait()
+{
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
+}
+
+inline void srcb_bank_wait()
+{
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
+}
+
 inline void move_d2a_fixed_face(const std::uint8_t addrmod)
 {
     TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCA_VLD); // MOVD2A for a whole face assumes unpacker will set a dummy data_valid, so we want to wait on that
--- a/tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/experimental/llk_math_mul_reduce_scalar.h
+++ b/tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/experimental/llk_math_mul_reduce_scalar.h
@@ -89,6 +89,10 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
             TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
         }
 
+        // MOVD2A writes SrcA from Dest and so is outside the Src auto-wait. Gate on the unpacker's
+        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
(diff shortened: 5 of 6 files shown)


amahmudTT:
## Comprehensive perturbation sweep, both trees

Widened from the single site in the issue to **every sync site on all three threads, all five filler types, delays 0-100, 10 repeats** — 165 `(filler, site)` cells per tree, 166,650 test bodies per tree, run on p100a / sfpi 7.72.0.

`auto` was not used: it only offers `unpacr*` on unpack-thread sites and `sfpnop` on SFPU sites, so each of the five fillers was forced explicitly at every site to cover the cross-unit combinations too.

Coverage was ledgered per cell rather than inferred from silence — **165/165 cells ran on both trees, 0 unverified.**

### Result

| site | filler | pristine | this PR |
| --- | --- | --- | --- |
| `unpack:3` `SEMGET@0x06898` | `tti_nop` | 100 delays fail, **1000/1000** | **clean** |
| `unpack:3` | `risc_nop` | 100 delays fail, **1000/1000** | **clean** |
| `unpack:3` | `sfpnop` | 100 delays fail, **1000/1000** | **clean** |
| `unpack:3` | `unpacr0` | 100 delays fail, **1000/1000** | **clean** |
| `unpack:3` | `unpacr1` | 100 delays fail, **1000/1000** | **clean** |
| `math:2` `SEMWAIT@0x0b3c8` | `unpacr0` | hang, delays 79-86, 8/80 | hang, delays 79-86, 8/80 (**unchanged**) |
| `math:2` | `unpacr1` | hang, delays 80-87, 8/80 | hang, delays 80-87, 8/80 (**unchanged**) |
| other 31 sites x 5 fillers | | clean | clean |

Two things worth noting beyond the fix itself:

1. **The race is deterministic, not flaky, and not filler-specific.** On pristine, `unpack:3` fails at *every* delay 1-100 under *all five* fillers, 1000/1000 — broader than the `tti_nop` / 5-100 characterisation in the issue. Delay 0 is absent from every record, i.e. the delay-0 control passed in all cases, so the cave detour and jump are not the cause.
2. **The fix introduces no new perturbation sensitivity.** The only surviving findings are the two `math:2` hangs, and they are identical on both trees — same delay bands, same 8/80 rate, same call chain (`_llk_math_wait_for_dest_available_` -> `math_dest_wait`, `cmath_common.h:268`). Pre-existing on `main`, untouched by this PR.

### On the `math:2` hang (not this PR's business - now tracked in #54689)

It is in the MATH_PACK dest-sync handshake, unrelated to the SrcA/SrcB bank-valid gate here. The filler is a pure no-op with respect to source-register state — per `tt_llk_blackhole/instructions/assembly.yaml`, our word has `Unpack_Pop=2` (`UNP_NOP`, "inject cycle delay between back to back unpack") with `Clr_to1_fmt_Ctrl=0` and `Set_Dvalid=0`, so neither of the two dvalid-setting encodings is active (`Clr_to1_fmt_Ctrl==3` is the one that "sets dvalid of Src Regs, but NOP in all other respects"), and all the clear-related fields are documented as valid only under `UNP_CLR_SRC`. The WH ISA page for this exact form gives its functional model as `// Causes no effects`.

But `UNPACR_NOP` carries `ex_resource: UNPACK`, so injecting ~80 of them into the **math** thread's stream makes math occupy unpacker 0 while the unpack thread is using it. No production LLK path issues an UNPACR from the math thread, so this stimulus looks unreachable in real code.

This has since been written up and filed as **#54689**, a sub-issue of #53394, rather than left as a loose note here. That issue records the full signature, the evidence that it is an injector artifact rather than a hardware race, and a suggested tool change so a future exhaustive sweep does not re-discover and re-triage it.

It also pins down why the unpacker fillers reach this site when 100 `tti_nop`s at the same site do nothing: `NOP` is dispatched to the Miscellaneous Unit and "executes in a single cycle", so n of them is about n cycles regardless of the other threads, whereas `UNPACR_NOP` is dispatched to an unpacker and only leaves the math thread's stream once that unpacker accepts it - a cost set by the unpack thread's workload, not by the NOP count. Past ~32 queued instructions the per-thread instruction FIFO fills and the RISCV core itself stalls (`BabyRISCV/PushTensixInstruction.md`), and that FIFO is in-order. So `unpacr*` counts are not comparable to `tti_nop` counts, and on a non-unpack thread they are not really a delay knob at all.

Raw ledgers and per-cell `failures.jsonl` archived at
`/proj_sw/user_dev/amahmud/race-audit/issue-53394-ttnop-llk/53412-comprehensive/`.

tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_mul_reduce_scalar.h
```diff
@@ -89,6 +89,10 @@ inline void _llk_math_mul_reduce_scalar_move_dest_to_src_([[maybe_unused]] std::
             TTI_SETRWC(p_setrwc::CLR_NONE, 0, 0, 0, 0, p_setrwc::SET_D);
         }
 
+        // MOVD2A writes SrcA from Dest and so is outside the Src auto-wait. Gate on the unpacker's
+        // dummy DVALID -- its UNP_ZEROSRC zeroes the bank these rows fill -- and drain in-flight math.
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
+
```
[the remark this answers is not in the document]
amahmudTT: Done in e80e48805c4 — extracted `srca_bank_wait()` / `srcb_bank_wait()` into each arch's `cmath_common.h` and routed all six sites through them. Codegen is unchanged: the math kernel `.text` is byte-identical before and after on both arches (BH 3200 bytes, WH 3332 bytes, matching md5s), and `test_mul_reduce_scalar.py` stays 54/54 on Blackhole.

I did not fold `move_d2a_fixed_face` / `move_d2b_fixed_face` into the same helper, though. They are the same idiom but not the same condition — Blackhole gates on `MATH | SRC*_VLD`, Wormhole gates on `SRC*_VLD` with no `MATH` drain:

```
BH cmath_common.h:126   TTI_STALLWAIT(STALL_MATH, MATH | SRCA_VLD)
WH cmath_common.h:123   TTI_STALLWAIT(STALL_MATH, SRCA_VLD)
```

Sharing one helper across both would silently add a stall bit to the Wormhole fixed-face path, which is existing HW-validated code on an architecture not available on this box. The WH/BH divergence looks like a real gap, but it wants its own change and its own validation rather than riding along here.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
```diff
@@ -119,6 +119,19 @@ inline void incr_counters(const std::uint32_t incr_a, const std::uint32_t incr_b
     TT_INCRWC(incr_cr, incr_d, incr_b, incr_a);
 }
 
+// MOVD2A/MOVD2B write SrcA/SrcB from Dest, so they fall outside the Src auto-wait, which covers
+// only instructions that read Src. Gate the row moves on the target bank's DVALID, and drain
+// in-flight math so the Dest values those moves read back have settled.
+inline void srca_bank_wait()
```
ncvetkovicTT, line 145: I don't really see the need for these two functions. We have numerous places in code where we don't wrap STALLWAIT in a function, and the mere number of possibilities for stalling and waiting resources deem this impractical, if you ask me.

ncvetkovicTT (commented):
This is experimental LLK. We must make sure that all occurrences of this call is thoroughly tested - from LLK tests up to ops tests and then models. It might just be that some model was falsely relying on this race, and that we need to fix something else in pair with this.

tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
amahmudTT: well in this file we are using the same pattern multiple times, and since this is a complex issue it is better to comment in every instance which is leading to redundancy , having the comment along with the function makes it easier not to make the mistake of not including p_stall::MATH when using p_stall::STALL_MATH  (rather contuerintuitive) which is plagueing multiple places. A better thing to do is to have these functions in a common place so that whenever we use MODVD2A/B we call these canonical ones.

nmohammedTT:
I was looking at this on my side as well. The all-zeros result comes from unpack’s UNPACR_NOPs (UNP_ZEROSRC) landing after math has already staged the tile into SrcA and the scaler into SrcB. Those NOPs zero the banks, SrcB is only written once, and every GAPOOL then multiplies by 0.
I had a working fix that stalled math on SRCA_VLD | SRCB_VLD at the end of _llk_math_mul_reduce_scalar_init_(), which is enough to make the unpack arm finish before the first MOVD2A/MOVD2B. I think this PR is the better version of that, because it waits on the correct src register. I reran the test case with your changes with --all mode and it passed.

nmohammedTT (approved):

tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
ncvetkovicTT: Ok if you want to go with the wrapper, I approve. Please still address my main comment about making sure we tested this nicely in relevant models/ops tests.

ncvetkovicTT (approved):

tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
nmohammedTT: I had @halghTT manually add 56 TTI_NOPs and run a blaze model, and it did hang. So this issue does exist in blaze but nothing has caught it yet.

amahmudTT: We have gathering on, we need to turn that off. Also I had missed the unpacker side changes.

amahmudTT:
CI does not exercise this path, and the All Model Tests failures are unrelated — host-side untilize shape errors, infra timeouts and an SDXL PCC assert, spread across galaxy/quietbox/llmbox/n150/n300. The only file here with broad reach, `cmath_common.h`, is purely additive (two new `inline` helpers), so it cannot change codegen for any other model.

The tests that are relevant pass, run locally on a Blackhole **p100a** at `81e5b87`:

| test | result | signal |
| --- | --- | --- |
| `test_sum_reduce_scalar` | 40 passed, 14 skipped, 0 failed | ✅ real |
| `test_mul_reduce_scalar` | 54 passed, 0 failed | ✅ real |
| `test_mul_reduce_scalar_chunked` | 36 xfailed | ⚠️ none — suite is `xfail` by design |

`sum_reduce_scalar` is the meaningful one: it is the test failing in #54806's CI, and the exact failing variant there (`Float16_b->Float16_b, HiFi2, num_tiles=2, [16, 32]`) passes here. All 14 skips are the known, unrelated `[16, 16]` single-face TODO in that file.

This is a regression check on silicon, not a repro — the race does not reproduce naturally, so the fail-without/pass-with evidence remains the TT-NOP injector run in the description. The hang is a separate issue: gathering hangs test cases, and #54806 (disable gathering) depends on this fix, since without it the race reproduces.

amahmudTT:
## Wormhole check

Ran the WH half on an **n150** (branch head `bd5db87f648` vs its merge-base `e8784f4d23a`, whose diff is exactly the 6 LLK files).

**Result: no WH regression.**

Two things worth flagging, both **pre-existing and unrelated to this PR** — verified at the merge-base:

- `test_mul_reduce_scalar` and `test_sum_reduce_scalar` call the pack/datacopy LLKs with the **Blackhole** argument list unconditionally. `_llk_pack_hw_configure_` has no `tile_c_dim` on WH, so `total_col_dim()` misbinds to `num_faces` and trips `LLK_ASSERT(num_faces == 1||2||4)` on TRISC2 — all args are `uint32_t`/`bool`, so it compiles clean and only fails at runtime. `_llk_pack_init_`/`_llk_pack_dest_init_` diverge the same way.
- `_llk_math_eltwise_unary_datacopy_` takes a 4th `num_faces` arg on BH but not WH — a hard compile error, so `test_sum_reduce_scalar` has **never compiled for Wormhole**.

So simply relaxing the `ChipArchitecture.BLACKHOLE` skip turns both suites red regardless of this PR.

With those calls routed through the existing `_wrapper_` helpers in `tests/helpers/include/llk_lib_{pack,math}_wrappers.h` (local, ~10 lines, not part of this PR), both arms are green, 3/3 runs each:

| test | merge-base | this branch |
|---|---|---|
| `test_mul_reduce_scalar` | 54 passed, 0 skipped, 0 failed | 54 passed, 0 skipped, 0 failed |
| `test_sum_reduce_scalar` | 40 passed, 14 skipped, 0 failed | 40 passed, 14 skipped, 0 failed |

The 14 skips are the test's own 16x16 multi-tile guard, not the arch skip — matching the Blackhole 40/14. Greens were sanity-checked with a golden-perturbation positive control (the suites do go red, on real device values), and `/tmp/tt-llk-build` was wiped between arms since the variant cache keys on include paths, not header contents — confirmed the two arms built different ELFs.

**Caveat for the reviewer note:** the WH half is HW-validated only *together with* that harness fix. I'll put it up separately — it is independent of this PR and is what actually gives WH CI coverage of these experimental LLKs.

Issue #54692, opened 2026-08-28
JIT kernel and firmware compilation require -flto=auto; removing it fails -Werror=array-bounds on fixed-address mailbox accesses


blozano-tt:
JIT kernel and firmware compilation both require `-flto=auto`. Removing it fails with `-Werror=array-bounds` errors on fixed-address mailbox accesses.

## Affected builds

`-flto=auto` is set in `common_flags` (`tt_metal/jit_build/build.cpp:172`) and assigned to both `cflags_` (`:188`) and `lflags_` (`:358`). It applies to every `JitBuildState`:

| Build | Sources | Compiled | Sites failing without LTO |
|---|---|---|---|
| `-DFW_BUILD` | `brisc.cc`, `trisc.cc`, `ncrisc.cc`, `idle_erisc.cc` | during host build | 69 |
| `-DKERNEL_BUILD` | `brisck.cc`, `ncrisck.cc`, `trisck.cc` | at runtime JIT | 3 |

Reproduced on Wormhole and Blackhole, at `-O2` and `-O3`, on Ubuntu 22.04 and 24.04.

## Steps to reproduce

Delete the token from `common_flags`:

```diff
--- a/tt_metal/jit_build/build.cpp
+++ b/tt_metal/jit_build/build.cpp
@@ -170,8 +170,6 @@
         "-ftt-no-dyninit "
-        // Rely on Link Time Optimization (removes globally unreachable code)
-        "-flto=auto "
         // Fast math allows non-IEEE compliant optimizations ...
```

Build. The host build fails while compiling firmware.

CI run demonstrating this: https://github.com/tenstorrent/tt-metal/actions/runs/33140947675

### Reproducing a single kernel compile

```sh
export TT_METAL_LOG_KERNELS_COMPILE_COMMANDS=1     # build.cpp:659-661
# run any workload that JIT-compiles a kernel, capturing output to log.txt
grep -m1 'g++ compile cmd:' log.txt | sed 's/.*g++ compile cmd: //; s/ (build\.cpp:[0-9]*)$//' > c.sh
sed 's/ -flto=auto//' c.sh > c_nolto.sh
cd "$(dirname "$(grep -oE '\-o [^ ]+\.o' c.sh | head -1 | cut -d' ' -f2)")"
bash c.sh        # exit 0
bash c_nolto.sh  # exit 1
```

[...]

Issue #21844, opened 2025-05-08
LTO support


ThisIsFineTM:
### Is your feature request related to a problem? Please describe

Right now there is no link-time optimization (LTO) support in our cmake config. Additionally, there is an explicit `-fno-lto` flag in the top level CMakeLists.txt. I've done some local testing, and LTO seems to work just fine. 

Also, these files have `-flto`:
METALIUM_GUIDE.md
tt_metal/hw/CMakeLists.txt
tt_metal/jit_build/build.cpp
tt_metal/third_party/tt_llk/tt_llk_blackhole/common/inc/ckernel_sfpi.h

Using `-flto` by itself will enable LTO, but only single threaded. `-flto=auto` will enable multithreaded LTO. So those references should also be updated.

### Describe the solution you'd like

Add opt-in build option for enabling LTO and update instances of `-flto` to `-flto=auto`. I'll put in a PR.

PR #54693, opened 2026-08-28
[EXPERIMENT — DO NOT MERGE] Drop -flto=auto from JIT kernel common_flags to see what CI reports

blozano-tt:
> **This PR is expected to fail and must not be merged.** It exists to get CI's verdict on a one-token change, as an independent check against a local reproduction. Draft, and I'll close it once CI has reported.

## What this changes

One token removed from `common_flags` in `tt_metal/jit_build/build.cpp:172`:

```diff
-        // Rely on Link Time Optimization (removes globally unreachable code)
-        "-flto=auto "
```

`common_flags` is assigned to **both** `cflags_` (`:188`) and `lflags_` (`:358`), so this affects JIT kernel compile *and* link.

## What I expect CI to report

Locally, replaying a real logged compile command (captured with `TT_METAL_LOG_KERNELS_COMPILE_COMMANDS=1`) with exactly this token deleted fails with three `-Werror=array-bounds` errors:

| Location | Expression |
|---|---|
| `tt_metal/hw/inc/internal/tt-1xx/blackhole/eth_fw_api.h:257` | `GET_MAILBOX_ADDRESS_DEV(link_status_check_timestamp)` |
| `tt_metal/hw/inc/internal/firmware_common.h:201` | `mailboxes->go_message_index` |
| `tt_metal/hw/inc/internal/firmware_common.h:203` | `mailboxes->go_messages[go_message_index].signal` |

each with `cc1plus: note: source object is likely at address zero`, then `cc1plus: all warnings being treated as errors`.

**Diagnosis:** these are casts of literal device addresses to pointers. With LTO enabled GCC cannot see across translation units to bounds-check them; with it disabled it can, concludes the object is at address zero, and `-Werror` makes it fatal. The diagnostic is correct in the abstract and inapplicable in fact — the "array at address 0" is a memory-mapped mailbox.

**If CI reports something different**, that is the more interesting outcome and worth knowing — my repro was on Blackhole `-DCOMPILE_FOR_BRISC` only, replaying one object's command rather than a full build, so CI covers ground I could not.

## Why it's worth knowing

Measured on a CPU-only MockDevice rig (tt-metal `d103fdd`, sfpi 7.69.0, per-kernel timings from `TT_METAL_LOG_KERNEL_COMPILE`):

| Phase | Time |
|---|---|
| Compile (direct measurement, 3 runs, ±3 ms) | 313 ms |
| Per-binary total (compile + link) | 688 ms |
| Link / LTO (inferred) | **~375 ms (~55%)** |

LTO link is the majority of per-binary JIT compile cost. Whether it can be removed or replaced is therefore worth measuring — but today that experiment cannot be run at all, because the build does not complete without it.

**To be clear: I am not proposing removing LTO.** The ask in #54692 is only that building without it should be *possible*, so the trade can be measured. Any real change would also need to replace what LTO provides (`build.cpp:171`: *"removes globally unreachable code"*) and re-verify L1 text budgets.

## Related

- #54692 — tracking issue, full commands and error output
- #41370 — `-mno-tt-fix-whbhebreak` silently defeated by `-flto=auto` in the same JIT path
- #21844 — LTO for the host CMake build (notes `jit_build/build.cpp` already carried `-flto`)
- tenstorrent/tt-blaze#3680 — downstream report with the generated kernel source

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:test/no-lto-jit-kernels)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:test/no-lto-jit-kernels)

### DIFF
--- a/tt_metal/jit_build/build.cpp
+++ b/tt_metal/jit_build/build.cpp
@@ -168,8 +168,9 @@ void JitBuildEnv::init(
         "-std=c++17 -ftt-nttp -ftt-constinit -ftt-consteval "
         // Ban dynamic initializations, via a check we've added
         "-ftt-no-dyninit "
-        // Rely on Link Time Optimization (removes globally unreachable code)
-        "-flto=auto "
+        // EXPERIMENT (do not merge): "-flto=auto " removed here to see what CI reports.
+        // See tenstorrent/tt-metal#54692 -- locally this fails with three
+        // -Werror=array-bounds errors on fixed-address mailbox accesses.
         // Fast math allows non-IEEE compliant optimizations ...
         "-ffast-math "
         // ... but we require these IEEE behaviors


blozano-tt:
## Result: confirmed, and broader than predicted

Sanity tests dispatched against this branch: https://github.com/tenstorrent/tt-metal/actions/runs/33140947675 — **failed**, as expected, with the same `-Werror=array-bounds=` error class.

Two ways my prediction was wrong, both worth recording:

1. **It breaks the host build, not runtime.** I expected `build.cpp` to compile fine and the failure to surface on first kernel JIT. In fact tt-metal compiles device firmware (`-DFW_BUILD`) through the same `jit_build` path at build time, so the build itself fails.
2. **72 error sites, not 3.** Across `brisc.cc` (33), `trisc.cc` (20), `ncrisc.cc` (12), `firmware_common.h` (8), `idle_erisc.cc` (1) — every RISC, on **Wormhole** as well as the Blackhole I tested locally, at `-O3` as well as `-O2`, on both Ubuntu 22.04 and 24.04.

Full breakdown on the issue: #54692 (comment).

The practical upshot: making LTO optional needs a shared accessor/macro for memory-mapped structures, not a handful of local suppressions. Closing — this PR existed only to get that answer.

PR #52081, opened 2026-08-04
[Bug fix] Avoid use of hard-coded bfloat16 intermediate CBs in Layernorm

ign-sudha:
### PR Category
Bug fix

### Summary
Fixes #43946.

Non-Welford `layer_norm_pre_all_gather` / `rms_norm_pre_all_gather` used hard-coded bf16 intermediate CBs (`x^2`, fused pre-add). With Float32 input and `fp32_dest_acc_en=true`, that still truncated values and left reduce on the FPU/tf32 path, so accuracy stayed poor even after only changing CB formats.

This PR:
1. Stores intermediate CBs (and the reduce scaler) in Float32 when `fp32_dest_acc_en` is set. Input/residual keep their own dtypes.
2. Adds an SFPU full-fp32 path for Float32 + `fp32_dest_acc_en` + `fast_and_approximate_mode=false`: `UnpackToDestFp32`, SFPU square/pre-add, and `ReduceFp32Mode::Accurate`. Otherwise keeps the existing FPU/tf32 path (Quasar always uses FPU).
3. Rejects Float32 input when `fp32_dest_acc_en` is false, instead of silently losing precision.
4. Adds `fast_and_approximate_mode` to both public APIs (default `false` = accurate).
5. Adds Float32 precision tests for layernorm (SFPU vs FPU, with/without residual) and a check for the `fp32_dest_acc_en` rejection.

### Notes for reviewers
- Main host changes: `layernorm_pre_all_gather_program_factory.cpp` (1D + 2D) — CB formats and `unpack_fp32_active`.
- Main device changes: layernorm/rmsnorm pre_allgather compute kernels and `pre_add.h` — SFPU vs FPU branches.
- Accurate reduce uses `PoolType::SUM` because SFPU Accurate only supports SUM. With the reader scaler of `1.0`, that matches the previous AVG+scaler behavior.
- `fast_and_approximate_mode` only matters for Float32 with `fp32_dest_acc_en=true`.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ign/sudha-hardcoded-bf16)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ign/sudha-hardcoded-bf16)
[...]

PR #54698, opened 2026-08-28
Size the pre-all-gather x^2 buffer at Wt, not Wt*2

### LINKED ISSUE #54697: [Bug] rms_norm_pre_all_gather: width-4096 row with fp32_dest_acc_en cannot fit L1 (regression from #52081)
### Summary

Since #52081, `ttnn.rms_norm_pre_all_gather` cannot allocate a width-4096 row when
`fp32_dest_acc_en=True`. It throws before the op runs. The same call succeeds on the
commit immediately before #52081, and succeeds today with `fp32_dest_acc_en=False`.

`fp32_dest_acc_en=True` is the value `models/common/rmsnorm.py` (`RMSNorm`, line 57)
passes **by default**, so this is a configuration the models use rather than a corner case.

### Minimal repro

Single device, no mesh, no model, no test framework:

```python
import torch, ttnn

cfg = lambda fp32: ttnn.WormholeComputeKernelConfig(
    math_fidelity=ttnn.MathFidelity.HiFi2, math_approx_mode=False,
    fp32_dest_acc_en=fp32, packer_l1_acc=True)

for W in (2048, 4096):
    for fp32 in (False, True):
        x  = torch.randn((1, 1, 32, W), dtype=torch.bfloat16)
        tt = ttnn.from_torch(x, dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT,
        ttnn.rms_norm_pre_all_gather(tt, compute_kernel_config=cfg(fp32), dtype=ttnn.bfloat16)
        print(f"W={W} fp32_dest_acc_en={fp32} -> OK")
ttnn.close_device(dev)
```

**Observed** — only the last combination fails:

```
W=2048 fp32_dest_acc_en=False -> OK
W=2048 fp32_dest_acc_en=True  -> OK
W=4096 fp32_dest_acc_en=False -> OK
W=4096 fp32_dest_acc_en=True  -> TT_THROW dataflow_buffer.cpp
    Statically allocated dataflow buffers on core range [0-0 - 0-0] grow to
    1684672 B which is beyond max L1 size of 1499136 bytes
```

Reproduces on **wormhole** (1684672 B vs 1499136 B) and **blackhole p300a**
(1690624 B vs 1572864 B), so it is not board specific.

### Bisect

Same script, same machine:

| build | `W=4096, fp32_dest_acc_en=True` |
|---|---|
| `4dfe546c002` — parent of #52081 | **OK** |
| current main | **FAIL** |

### Root cause

`layernorm_pre_all_gather_program_factory.cpp` sizes the `x2` (x squared) buffer at the
full row, double buffered:

```cpp
const uint32_t intermed0_tiles = Wt * double_buffer_constant;   // x^2
```

#52081 changed the intermediate CB format from hard-coded bf16 to fp32 when
`fp32_dest_acc_en` is set:

```cpp
- tt::DataFormat cb_data_format = tt::DataFormat::Float16_b;
+ tt::DataFormat cb_data_format = fp32_dest_acc_en ? Float32 : Float16_b;
```

At Wt=128 that takes the `x2` buffer from 256x2048 = 524,288 B to 256x4096 = 1,048,576 B.
The **+524,288 B is exactly the overshoot**:

| | x2 buffer | program total | fits blackhole (1,572,864)? |
|---|---|---|---|
| before #52081 | 524,288 | ~1,166,000 | yes |
| after #52081 | 1,048,576 | 1,690,624 | **no** |

To be clear, **#52081 is a correct fix** — using bf16 intermediates while the caller asked
for fp32 accumulation loses precision. The problem is that the op has no headroom for it.

### Suggested fix

The `x^2` buffer is sized at `Wt * double_buffer_constant`, i.e. a full row, double buffered.
Sizing it **adaptively** — keep the double buffer whenever the double-buffered program fits
L1, fall back to a single buffer only when it would not fit at all — recovers the bytes the
format change added, while keeping #52081's precision improvement and leaving every shape
that already fits byte-for-byte unchanged.

The double buffer is worth keeping where it fits: the packer and the unpacker are different
RISCs, so it buys real overlap (measured 1-3% at widths that fit). An unconditional single
buffer would pay that cost everywhere, which is why the fit check is the right shape for this.

`in0` keeps its double buffer in all cases — there the reader genuinely prefetches ahead of
compute.

Measured on wormhole with the adaptive form:

- the five previously failing heights at width 4096 all run, PCC 0.998 vs an fp32 golden
- widths that already worked (640 / 1024 / 2048) return **bit-identical** PCCs
- the layernorm (non-RMS) path still produces correct `sum(x)` and `sum(x^2)`
- no perf change on any shape main can already run (identical program emitted)

### Impact

Any distributed norm reaching this op with a wide per-chip row and the default
`fp32_dest_acc_en=True`. It is how Llama-3.3-70B lands on a 2-device mesh
(hidden 8192 / 2 = 4096 per chip); those traced configs have thousands of recorded
executions from before #52081.

Aswinmcw:
Fixes #54697

### Problem

Since #52081 the intermediate CB follows `fp32_dest_acc_en` instead of being hard-coded bf16. That is a **correct precision fix**, but it doubles the `x^2` buffer, and a width-4096 row no longer fits L1:

```
W=4096, fp32_dest_acc_en=True -> 1684672 B vs 1499136 B (wormhole)
                                 1690624 B vs 1572864 B (blackhole p300a)
```

`fp32_dest_acc_en=True` is what `models/common/rmsnorm.py` `RMSNorm` passes by default (line 57), so this is a configuration the models use — it is how Llama-3.3-70B lands on a 2-device mesh (hidden 8192 / 2 = 4096 per chip).

Bisected with the single-device repro in #54697: **passes** on `4dfe546c002` (parent of #52081), **fails** on main, **passes** again with this change.

### Fix

Size the `x^2` buffer **adaptively**: keep the double buffer whenever the double-buffered program fits L1, and fall back to a single buffer only when it would not fit at all.

```cpp
const uint32_t intermed0_tiles =
    (static_bytes_double_buffered <= device->l1_size_per_core()) ? x2_tiles_double_buffered : Wt;
```

An earlier revision of this PR dropped the double buffer unconditionally, on the reasoning that the same thread fills and drains `x^2` so the second half could never be live. **That reasoning was wrong** — the compute kernel is three RISCs, and the *packer* produces into `x^2` while the *unpacker* consumes it, so there is real cross-RISC overlap. Measuring it (below) showed a 1-3% cost, hence the adaptive form: only shapes that cannot fit the double buffer give it up.

`in0` keeps its double buffer in all cases — there the reader genuinely prefetches ahead of compute. #52081's precision improvement is preserved rather than reverted. The `intermed0_tiles % block_size == 0` assert still holds: `block_size` is 1 in this factory.

### Performance vs main (wormhole N300, H=1024, 200 iterations, warm cache)

| width | main | this PR | note |
|---|---|---|---|
| 1024 | 82.1 us | 83.0 - 84.6 us | identical program, delta within noise |
| 2048 | 103.5 us | 103.9 - 106.0 us | identical program, delta within noise |
| 3072 | 130.0 us | 130.9 - 132.2 us | identical program, delta within noise |
| 4096 | **fails to allocate** | 149.5 - 155.2 us | previously impossible |

For every shape main can run, the fit check passes and `intermed0_tiles` takes exactly main's `Wt * double_buffer_constant`, so the emitted program is **identical** — perf equality is structural, not statistical. The PR ranges above are three repeats of the same binary; that run-to-run spread (1.3-2.1 us at these widths) is as large as the apparent gap to main, so the gap is not measurable signal.

For reference, the single-vs-double comparison that motivated the adaptive form, measured same-binary via a temporary env toggle:

| shape | single buffer | double buffer | double faster by |
|---|---|---|---|
| H=1024 W=1024 | 90.9 us | 88.3 us | 2.9% |
| H=2048 W=2048 | 114.7 us | 112.6 us | 1.9% |
| H=1024 W=2048 | 104.7 us | 103.8 us | 0.9% |

At width 4096 the double buffer is not an option at all — forcing it still fails to allocate — so no perf is being left on the table there.

### Validation (wormhole N300)

| | result |
|---|---|
| five previously failing heights at width 4096 (32/128/1024/2048/4096) | now run, **PCC 0.998** vs an fp32 golden |
| widths that already worked (640 / 1024 / 2048) | **bit-identical** PCCs before and after |
| layernorm (non-RMS) path | still correct `sum(x)` and `sum(x^2)` |
| `test_pre_all_gather_wide_row_fp32_dest_acc_fits_l1` | added, covers RMSNorm and LayerNorm on a single device |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:Aswinmcw/fix-layernorm-preallgather-l1)

### DIFF
--- a/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
@@ -147,13 +147,19 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
     const uint32_t res_tiles = Wt * double_buffer_constant;    // residual b
     const uint32_t fused_tiles = Wt;                           // a + b
 
-    // x^2 is filled to Wt tiles and then drained by a single BulkWaitBulkPop reduce in the SAME
-    // compute kernel, so the second half of a double buffer can never be live: nothing produces
-    // into it while the reduce consumes the first. Sizing it at Wt keeps the access pattern
-    // identical -- tiles are packed at absolute index wt+wtr over one row -- and gives back a full
-    // row's worth of L1, which a wide fp32_dest_acc_en row needs to fit at all (see #54697).
-    // in0 keeps its double buffer: there the reader genuinely prefetches ahead of compute.
-    const uint32_t intermed0_tiles = Wt;  // x^2
+    // The x^2 buffer is sized per ROW, so a wide fp32_dest_acc_en row is what pushes this program
+    // past L1 (#54697). Keep the double buffer wherever it fits -- the packer and the unpacker are
+    // different RISCs, so it buys real overlap (measured ~1-3% on shapes that already fit) -- and
+    // fall back to a single buffer only when the double-buffered program would not fit at all,
+    // which today throws at allocation instead of running.
+    const uint32_t x2_tiles_double_buffered = Wt * double_buffer_constant;
+    const uint32_t out0_tiles_estimate = is_rmsnorm ? 1 : 2;
+    const uint32_t static_bytes_double_buffered =
+        in0_tiles * in_single_tile_size + in1_tiles * scaler_tile_size +
+        (fuse_pre_add ? (res_tiles * inb_single_tile_size + fused_tiles * single_tile_size) : 0) +
+        x2_tiles_double_buffered * single_tile_size + out0_tiles_estimate * out_single_tile_size;
+    const uint32_t intermed0_tiles =
+        (static_bytes_double_buffered <= device->l1_size_per_core()) ? x2_tiles_double_buffered : Wt;  // x^2
     uint32_t out0_tiles = 1;
     if (!is_rmsnorm) {
         out0_tiles = 2;
--- a/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
@@ -150,10 +150,9 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
     // x^2 is filled to Wt tiles and then drained by a single BulkWaitBulkPop reduce in the SAME
     // compute kernel, so the second half of a double buffer can never be live: nothing produces
     // into it while the reduce consumes the first. Sizing it at Wt keeps the access pattern
-    // identical (tiles are packed at absolute index wt+wtr over one row) and saves Wt * fp32 tile
-    // bytes, which is exactly what pushes a wide fp32_dest_acc_en row past L1: at width 4096 the
-    // program asks 1684672 B against 1499136 B (wormhole) / 1572864 B (blackhole). in0 keeps its
-    // double buffer -- there the reader genuinely prefetches ahead of compute.
+    // identical -- tiles are packed at absolute index wt+wtr over one row -- and gives back a full
+    // row's worth of L1, which a wide fp32_dest_acc_en row needs to fit at all (see #54697).
+    // in0 keeps its double buffer: there the reader genuinely prefetches ahead of compute.
     const uint32_t intermed0_tiles = Wt;  // x^2
     uint32_t out0_tiles = 1;
     if (!is_rmsnorm) {
--- a/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
@@ -147,7 +147,14 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
     const uint32_t res_tiles = Wt * double_buffer_constant;    // residual b
     const uint32_t fused_tiles = Wt;                           // a + b
 
-    const uint32_t intermed0_tiles = Wt * double_buffer_constant;  // x^2
+    // x^2 is filled to Wt tiles and then drained by a single BulkWaitBulkPop reduce in the SAME
+    // compute kernel, so the second half of a double buffer can never be live: nothing produces
+    // into it while the reduce consumes the first. Sizing it at Wt keeps the access pattern
+    // identical (tiles are packed at absolute index wt+wtr over one row) and saves Wt * fp32 tile
+    // bytes, which is exactly what pushes a wide fp32_dest_acc_en row past L1: at width 4096 the
+    // program asks 1684672 B against 1499136 B (wormhole) / 1572864 B (blackhole). in0 keeps its
+    // double buffer -- there the reader genuinely prefetches ahead of compute.
+    const uint32_t intermed0_tiles = Wt;  // x^2
     uint32_t out0_tiles = 1;
     if (!is_rmsnorm) {
         out0_tiles = 2;
--- a/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
+++ b/ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
@@ -158,8 +158,16 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
         in0_tiles * in_single_tile_size + in1_tiles * scaler_tile_size +
         (fuse_pre_add ? (res_tiles * inb_single_tile_size + fused_tiles * single_tile_size) : 0) +
         x2_tiles_double_buffered * single_tile_size + out0_tiles_estimate * out_single_tile_size;
+    // Budget is L1 above the reserved base, matching what validate_dataflow_buffer_region compares
+    // against (it checks an absolute region end, so the reserved base must come off the top here).
+    // Deliberately STATIC: shape, dtypes and device config only. Sizing a buffer from dynamic L1
+    // occupancy (lowest_occupied_compute_l1_address) would make the program depend on state the
+    // program-cache key does not cover, so a program built under one occupancy could be replayed
+    // under another -- unsound with the program cache and with tracing.
+    const uint32_t l1_budget = static_cast<uint32_t>(
+        device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1));
     const uint32_t intermed0_tiles =
-        (static_bytes_double_buffered <= device->l1_size_per_core()) ? x2_tiles_double_buffered : Wt;  // x^2
+        (static_bytes_double_buffered <= l1_budget) ? x2_tiles_double_buffered : Wt;  // x^2
     uint32_t out0_tiles = 1;
     if (!is_rmsnorm) {
         out0_tiles = 2;
--- a/tests/ttnn/unit_tests/operations/fused/test_distributed_layernorm.py
+++ b/tests/ttnn/unit_tests/operations/fused/test_distributed_layernorm.py
@@ -11,7 +11,7 @@ import ttnn
 from models.common.utility_functions import tt2torch_tensor
 
 from loguru import logger
-from tests.ttnn.utils_for_testing import assert_numeric_metrics
+from tests.ttnn.utils_for_testing import assert_numeric_metrics, check_with_pcc
 from tests.tests_common.skip_reasons import LEGACY_CCL_SKIP
 from ttnn import ShardTensorToMesh, ConcatMeshToTensor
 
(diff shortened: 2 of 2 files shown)


bbradelTT (commented):
Would there be any perf improvement if the double buffer was used?

fplavecTT:
While in theory double-buffer doesn't matter for this, we should run at least some perf tests to prove that this is also true in practice (particularly if the theoretical conclusion comes from AI).

fplavecTT:
PR description says "LayerNormPreAllGather2DProgramFactory has the same pattern and the same over-allocation (1725632 B at this shape). I have not validated that path, so it is left alone here and noted in https://github.com/tenstorrent/tt-metal/issues/54697"

But the PR description also includes "Fixes #54697", so merging this PR will close the issue #54697 although not everything the issue reports is fixed by this PR

ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
```diff
@@ -147,7 +147,14 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
...
 
-    const uint32_t intermed0_tiles = Wt * double_buffer_constant;  // x^2
+    // x^2 is filled to Wt tiles and then drained by a single BulkWaitBulkPop reduce in the SAME
+    // compute kernel, so the second half of a double buffer can never be live: nothing produces
+    // into it while the reduce consumes the first. Sizing it at Wt keeps the access pattern
+    // identical (tiles are packed at absolute index wt+wtr over one row) and saves Wt * fp32 tile
+    // bytes, which is exactly what pushes a wide fp32_dest_acc_en row past L1: at width 4096 the
+    // program asks 1684672 B against 1499136 B (wormhole) / 1572864 B (blackhole). in0 keeps its
+    // double buffer -- there the reader genuinely prefetches ahead of compute.
+    const uint32_t intermed0_tiles = Wt;  // x^2
```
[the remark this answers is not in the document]
Aswinmcw: Agreed — fixed in 21b8d86.

The totals are a property of one configuration (dtypes, RMSNorm vs LayerNorm output size, residual fusion, arch, and any future neighbouring buffer), so they would go stale and could misdescribe other paths through this factory. The comment now keeps only the lifetime argument, which holds across all of them, and points at #54697 where the measured numbers live permanently.

Aswinmcw:
Both review points addressed in 21b8d86.

**1. Point-in-time allocation totals removed from the comment** — replied inline. They describe one configuration and would go stale; the lifetime argument stands on its own and #54697 retains the measured numbers.

**2. Regression test added** — `test_pre_all_gather_wide_row_fp32_dest_acc_fits_l1`, covering both RMSNorm and LayerNorm.

Worth being explicit about why the existing tests could not have caught this, since it is the whole reason the regression reached main unnoticed: every parametrization in this module shards `8192` across **4 or 8 devices**, so each chip only ever sees a **1024-wide** row. The buffer that overflows is sized per *per-chip* row, so no amount of multi-device coverage exercises it.

The new test instead runs the widest per-chip row the models actually use — Llama-3.3-70B on a 2-device mesh, `8192 / 2 = 4096` — on a **single device** with `fp32_dest_acc_en=True`. It needs no T3K, so it runs on any single-chip lane, and it fails at allocation time rather than on numerics if the sizing regresses.

One caveat on scope: I verified the **RMSNorm** variant fails pre-fix (that is the bisect in the PR description). The **LayerNorm** variant is included because it shares the same buffer sizing and is the strictly larger case (`out0_tiles = 2`), but I have not run it against a pre-#52081 binary.

Aswinmcw:
> perf change?

Good question, and checking it found a real problem with my reasoning — **fixed in 8ef367f**.

**I had not measured it.** I argued the `x^2` double buffer could never be live because "the same thread fills and drains it". That is wrong: the compute kernel is three RISCs, and the **packer** produces into `x^2` while the **unpacker** consumes it, so there is genuine cross-RISC overlap. Dropping it unconditionally does cost throughput. Same binary, env-toggled sizing, 200 iterations each, warm cache:

| shape | Wt (unconditional) | Wt*2 (old) | delta |
|---|---|---|---|
| H=1024 W=1024 | 90.9 us | 88.3 us | **+2.9%** |
| H=2048 W=2048 | 114.7 us | 112.6 us | **+1.9%** |
| H=1024 W=2048 | 104.7 us | 103.8 us | **+0.9%** |

So the change now sizes the buffer **adaptively**: keep `Wt*2` whenever the double-buffered program fits L1, and fall back to `Wt` only when it would not fit at all — the case that today throws at allocation rather than merely running slower.

Re-measured with that:

| shape | adaptive (this PR) | Wt*2 (old) |
|---|---|---|
| H=1024 W=1024 | 87.5 us | 88.3 us |
| H=2048 W=2048 | 113.2 us | 112.6 us |
| H=1024 W=2048 | 103.9 us | 103.8 us |

Shapes that already worked keep the double buffer and their timings (within noise of the old sizing), and width 4096 with `fp32_dest_acc_en` now runs at PCC 0.998. Numerics are unchanged throughout — the already-working widths still return bit-identical PCCs.

The code comment now records the cross-RISC overlap and the measured cost, so the next person does not repeat my mistake.

Aswinmcw:
> PR description says "LayerNormPreAllGather2DProgramFactory has the same pattern and the same over-allocation (1725632 B at this shape). I have not validated that path, so it is left alone here and noted in #54697"
> 
> But the PR description also includes "Fixes #54697", so merging this PR will close the issue #54697 although not everything the issue reports is fixed by this PR

@fplavecTT i have updated the PR and issue description to reflect the proper change

ttnn/cpp/ttnn/operations/normalization/layernorm_distributed/device/layernorm_pre_all_gather_program_factory.cpp
```diff
@@ -147,7 +147,19 @@ ttnn::device_operation::ProgramArtifacts LayerNormPreAllGatherProgramFactory::cr
...
+    // fall back to a single buffer only when the double-buffered program would not fit at all,
+    // which today throws at allocation instead of running.
+    const uint32_t x2_tiles_double_buffered = Wt * double_buffer_constant;
+    const uint32_t out0_tiles_estimate = is_rmsnorm ? 1 : 2;
+    const uint32_t static_bytes_double_buffered =
+        in0_tiles * in_single_tile_size + in1_tiles * scaler_tile_size +
+        (fuse_pre_add ? (res_tiles * inb_single_tile_size + fused_tiles * single_tile_size) : 0) +
+        x2_tiles_double_buffered * single_tile_size + out0_tiles_estimate * out_single_tile_size;
+    const uint32_t intermed0_tiles =
+        (static_bytes_double_buffered <= device->l1_size_per_core()) ? x2_tiles_double_buffered : Wt;  // x^2
```
bbradelTT: I think this should take into account more than just the total L1 size. See https://github.com/tenstorrent/tt-metal/pull/54991/changes for an example.
The size comparison should be for all CBs, not just one.

Also, if you do this, then this will not work properly with tracing. You need to provide a program config parameter to allow the user to specify what needs to be done.

bbradelTT: The program config support may be ok to be delayed to a future PR.

bbradelTT (approved):

fplavecTT:
> > PR description says "LayerNormPreAllGather2DProgramFactory has the same pattern and the same over-allocation (1725632 B at this shape). I have not validated that path, so it is left alone here and noted in #54697"
> > But the PR description also includes "Fixes #54697", so merging this PR will close the issue #54697 although not everything the issue reports is fixed by this PR
> 
> @fplavecTT i have updated the PR and issue description to reflect the proper change

Can you elaborate what exactly was done? I see that mention of LayerNormPreAllGather2DProgramFactory is gone from the PR. Is there another issue opened to track that work?

Aswinmcw:
> > > PR description says "LayerNormPreAllGather2DProgramFactory has the same pattern and the same over-allocation (1725632 B at this shape). I have not validated that path, so it is left alone here and noted in #54697"
> > > But the PR description also includes "Fixes #54697", so merging this PR will close the issue #54697 although not everything the issue reports is fixed by this PR
> > 
> > 
> > @fplavecTT i have updated the PR and issue description to reflect the proper change
> 
> Can you elaborate what exactly was done? I see that mention of LayerNormPreAllGather2DProgramFactory is gone from the PR. Is there another issue opened to track that work?

The same bug is there is 2D as well but same fix results in hang so i limited the issue and fix for 1D alone for now

fplavecTT:
> The same bug is there is 2D as well but same fix results in hang so i limited the issue and fix for 1D alone for now

Please open an issue for addressing 2D so it's not forgotten.

PR #54701, opened 2026-08-28
[Fix] Mixed dtype cases in Eltwise Unary #54425

KalaivaniMCW:
### PR Category

Fix
### Summary

https://github.com/tenstorrent/tt-metal/issues/53963 https://github.com/tenstorrent/tt-metal/issues/53613

### Notes for reviewers

Added some tests covering mixed dtype case in eltwise ops
Added restrictions to block unsafe mixed dtype cases

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kalaivani/mixed_ttsim)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kalaivani/mixed_ttsim)


ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -23,9 +23,20 @@ Tensor unary_impl(
...
+            TT_FATAL(
+                (output_dtype == optional_output_tensor->dtype()),
+                "Preallocated tensor's dtype {} is not compatible with the expected dtype {}.",
+                static_cast<int>(optional_output_tensor->dtype()),
+                static_cast<int>(output_dtype));
+        }
+    } else if (optional_output_tensor.has_value()) {
         output_dtype = optional_output_tensor->dtype();
+        TT_FATAL(
+            (output_dtype == input_dtype) || (is_floating_point(output_dtype) && is_floating_point(input_dtype)),
```
Aswinmcw: `is_floating_point` also returns true for `BFLOAT4_B` and `FP8_E4M3` (tensor_types.cpp:72-81), so this gate admits e.g. `float32` input with an `FP8_E4M3` preallocated output for every unary op on every arch — and the enum itself warns that FP8_E4M3 is "Blackhole only, ROW-MAJOR only for now ... Check op support before opting in" (tensor_types.hpp:35-37). Since the point of the change is to block unsafe mixes, consider gating on the set actually validated here (bf16 / fp32 / bf8) rather than the whole floating-point family.

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
```diff
@@ -23,9 +23,20 @@ Tensor unary_impl(
     if (op_chain.back().type() == unary::UnaryOpType::TYPECAST ||
         op_chain.back().type() == unary::UnaryOpType::BITCAST) {
         output_dtype = static_cast<DataType>(*op_chain.back().get_param_if<float>(1));
-    }
-    if (optional_output_tensor.has_value()) {
+        if (optional_output_tensor.has_value()) {
+            TT_FATAL(
+                (output_dtype == optional_output_tensor->dtype()),
+                "Preallocated tensor's dtype {} is not compatible with the expected dtype {}.",
```
Aswinmcw: `fmt::formatter<DataType>` is specialized (tensor_types.hpp:98), so `"{}"` with the dtype prints `BFLOAT8_B` instead of `3` and the `static_cast<int>` can go. Also, both new TT_FATALs use this identical message string, so the typecast-mismatch and the int/float-mix failures are indistinguishable in a log — worth naming each condition.

tests/ttnn/unit_tests/operations/eltwise/test_eltwise_mixed_dtype.py
```diff
@@ -0,0 +1,63 @@
...
+    # Pass the ttnn tensor into assert_with_ulp so bfloat8_b is compared at bfloat16
+    # resolution; ttnn.to_torch() would upcast it to float32 and make ULP too strict.
+    golden = golden_fn(ttnn.to_torch(input_tensor)).to(TORCH_DTYPE[out_dtype])
+    assert output_tensor.dtype == out_dtype
+    assert_with_ulp(golden, output_tensor, ulp_threshold)
+
+
+def test_neg_mixed_dtype(device):
+    """Preallocated output dtype can differ from the input. Every in/out pair is
+    invoked back-to-back to check for program-cache collision."""
```
Aswinmcw: The docstring says the back-to-back invocation checks for program-cache collision, but nothing asserts it — the coverage rests entirely on `pytest.mark.use_module_device` keeping the cache warm across the loop. If that marker is dropped or the fixture scope changes, the test keeps passing while no longer testing #53963. An assertion that `device.num_program_cache_entries()` grows once per distinct (in, out) pair would pin the intent.

tests/ttnn/unit_tests/operations/eltwise/test_eltwise_mixed_dtype.py
```diff
@@ -0,0 +1,63 @@
...
+    torch.manual_seed(0)
+    fixed_input = torch.ones((64, 64), dtype=torch.float32) * 1.22
+    ttnn_op = ttnn.neg
+
+    for in_dt in FLOAT_DTYPES:
+        for out_dt in FLOAT_DTYPES:
+            torch_input = fixed_input.to(TORCH_DTYPE[in_dt])
+            _unary_with_preallocated_output(
+                device, torch_input, in_dt, out_dt, ttnn_op, ulp_threshold=ULP_THRESHOLD[(in_dt, out_dt)]
+            )
```
Aswinmcw: Neither new TT_FATAL is exercised. Two cases would cover them: an int input with a float preallocated output, and `ttnn.typecast(x, ttnn.bfloat16, output_tensor=<float32 tensor>)` — both expected to raise.

Aswinmcw (commented):
The TYPECAST branch is a user-visible behaviour change that the description does not mention: `ttnn.typecast(x, dtype, output_tensor=y)` previously ignored `dtype` and silently used `y.dtype()`, and now hard-errors. Worth a line in the summary so it is not found by a downstream break.

tests/ttnn/unit_tests/operations/eltwise/test_eltwise_mixed_dtype.py
KalaivaniMCW: The unsafe mixed dtype cases were verified locally. 
image

tests/ttnn/unit_tests/operations/eltwise/test_eltwise_mixed_dtype.py
KalaivaniMCW: added assert to check the 9 pgm cache entries

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
KalaivaniMCW: removed static cast

ttnn/cpp/ttnn/operations/eltwise/unary/unary.cpp
KalaivaniMCW: allowing only bf16 / fp32 / bf8 , as other two dtypes will see a precision loss.

Aswinmcw (approved):
All four earlier points are addressed. One doc item left: the TYPECAST/BITCAST branch is a user-visible behaviour change — `ttnn.typecast(x, dtype, output_tensor=y)` previously ignored `dtype` and used `y.dtype()`, and now raises when they disagree. Worth a line in the description so it isn't a surprise to callers.

PR #54714, opened 2026-08-28
[Bug fix] Honour the requested memory_config when allocating gradient outputs

### LINKED ISSUE #54715: preallocated_tensors_check`: binary backward grad outputs ignore `memory_config`
### Observed

`binary_backward.cpp:29-43`:

```cpp
void preallocated_tensors_check(
    std::optional<Tensor>& input_grad,
    std::optional<Tensor>& other_grad,
    const Tensor& input,
    const Tensor& other,
    const std::array<bool, 2>& required_outputs) {          // no memory config parameter
    ...
    if (required_outputs[0] && !input_grad.has_value()) {
        input_grad = ttnn::empty_like(input);               // 38
    }
    if (required_outputs[1] && !other_grad.has_value()) {
        other_grad = ttnn::empty_like(other);               // 41
    }
}
```

The helper cannot honour a memory config because it is never given one. Every gradient it allocates inherits the corresponding input's config.

Measured on `main`, inputs in L1, `memory_config=DRAM`:

```
add_bw -> grad configs = [L1, L1]        (DRAM requested)
```

These tensors are the op's **actual returned outputs**, not intermediates, so the whole result lands in the wrong memory space.

### Assumptions

- `memory_config` is intended to control where the returned gradients are allocated, consistent with the rest of the eltwise surface.
- Callers that supply their own preallocated grads keep full control; only the helper-allocated path is in question.

### Gaps

- 8 call sites route through this helper, so the behaviour is uniform across the binary backward ops and no single op looks anomalous.
- Existing tests pass the same memory config for inputs and output, so inherited and requested never diverge.

### Fix required

- Add a memory-config parameter to `preallocated_tensors_check` and forward it to both `empty_like` calls.
- Thread the caller's `output_mem_config` through all 8 call sites (`binary_backward.cpp` lines 133, 157, 194, 500, 525, 574, 688, 821).
- Regression test on at least one representative op, asserting the returned grads honour a `memory_config` that differs from the inputs'.

VirdhatchaniKN:
### Issue

Closes #54715. Sub-issue B of the memory-config series; related to item 17 of #48240 (#54523).

### What this does

Gradient tensors that these ops allocate themselves now land in the caller's requested `memory_config` instead of inheriting the input's.

- `detail::preallocated_tensors_check` gains a `memory_config` parameter, forwarded to both `empty_like` calls and threaded through the 8 tensor-tensor call sites.
- The four tensor-scalar overloads (`add_bw`, `sub_bw`, `div_bw`, `mul_bw`) allocate in the requested config.
- `output_mem_config` is restored in the signatures of `assign_bw`, `concat_bw`, and the `add_bw`/`sub_bw` scalar overloads, where it had been commented out.

### Why

These helpers allocate the gradient when the caller supplies none, but had no memory config to honour, so every tensor they created inherited the input's config. These are the ops' returned outputs, not intermediates.

`assign`, `concat`, and the `add_bw`/`sub_bw` scalar overloads ignored the config entirely rather than only in the helper, so they change from never honouring it to honouring it — the largest behaviour delta here.

Measured on Wormhole, bfloat16/TILE, inputs in **L1**, `memory_config=DRAM`:

| Path | before | after |
| --- | --- | --- |
| tensor-tensor: `add, mul, div, rsub, assign, addalpha, subalpha, concat` | L1 | DRAM |
| tensor-scalar: `add, sub, div, mul` | L1 | DRAM |

The input placement matters: with DRAM inputs the inherited and requested configs coincide and the defect is invisible, so any check has to use differing configs.

### No accuracy impact

Placement only. Gradient values are identical across `memory_config` unset / L1 / DRAM, `max|diff| = 0.0`. With `memory_config` unset — the default — behaviour is unchanged, since `nullopt` still falls through to the input's config.

### One change deliberately left out

`concat_bw`'s two `ttnn::slice` calls keep `std::nullopt` for the memory config. `slice.cpp:123` resolves it as `optional_output_tensor.has_value() ? its config : memory_config_arg`, and `input_grad` is always set by the helper immediately above, so the argument could never take effect. `concat_bw` still returns DRAM for L1 inputs without it. A comment records this so it is not re-added.

### Tests

Regressions in `test_backward_add.py`, all with inputs in **L1**, DRAM requested, no preallocated outputs, comparing the whole `memory_config` rather than `buffer_type` alone. The tensor-tensor case is table driven (`_TT_GRAD_OPS`), so a new op added to the helper needs one line.

| Path | Ops |
| --- | --- |
| tensor-tensor (× 3 `are_required_outputs`) | `add, addalpha, subalpha, rsub, div, mul, assign` |
| tensor-tensor, separate shape (× 3) | `concat` |
| tensor-scalar | `add, sub, div, mul` |

Every call site this diff touches has a test that fails without its fix, and the two halves fail independently:

| Control | Result |
| --- | --- |
| revert the helper's two `empty_like` args | 24 failed |
| revert the four scalar sites | 4 failed |
| with both fixes | 28 passed |

| Suite | Result |
| --- | --- |
| `eltwise/backward`, affected ops | 819 passed, 20 skipped |
| `docs_examples` for those ops | 1 passed |

Note for anyone reproducing: input placement matters. With DRAM inputs the inherited and requested configs coincide and the defect is invisible, so a check must use differing configs.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/bb_grad_check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/bb_grad_check)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_add.py
+++ b/tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_add.py
@@ -122,3 +122,107 @@ def test_bw_add_scalar(input_shapes, scalar, device):
 
     status = compare_pcc(tt_output_tensor_on_device, golden_tensor)
     assert status
+
+
+# Tensor-tensor ops whose gradients are allocated by preallocated_tensors_check.
+# concat_bw is covered separately: it needs a differently shaped grad tensor.
+_TT_GRAD_OPS = {
+    "add": lambda g, a, b, req, mc: ttnn.add_bw(g, a, b, are_required_outputs=req, memory_config=mc),
+    "addalpha": lambda g, a, b, req, mc: ttnn.addalpha_bw(g, a, b, 1.0, are_required_outputs=req, memory_config=mc),
+    "subalpha": lambda g, a, b, req, mc: ttnn.subalpha_bw(g, a, b, 1.0, are_required_outputs=req, memory_config=mc),
+    "rsub": lambda g, a, b, req, mc: ttnn.rsub_bw(g, a, b, are_required_outputs=req, memory_config=mc),
+    "div": lambda g, a, b, req, mc: ttnn.div_bw(g, a, b, are_required_outputs=req, memory_config=mc),
+    "mul": lambda g, a, b, req, mc: ttnn.mul_bw(g, a, b, are_required_outputs=req, memory_config=mc),
+    "assign": lambda g, a, b, req, mc: ttnn.assign_bw(g, a, b, are_required_outputs=req, memory_config=mc),
+}
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("are_required_outputs", [[True, True], [True, False], [False, True]])
+@pytest.mark.parametrize("op", sorted(_TT_GRAD_OPS))
+def test_bw_helper_allocated_grads_honour_memory_config(input_shapes, device, are_required_outputs, op):
+    """Gradients the op allocates itself must land in the requested memory config.
+
+    Regression for `preallocated_tensors_check`, which allocated via `empty_like`
+    without a memory config, so the gradients inherited the input's config instead.
+    Existing coverage either omits `memory_config` or supplies preallocated outputs,
+    so the requested and inherited configs never diverge there.
+
+    Inputs are placed in L1 and DRAM is requested: with DRAM inputs the inherited and
+    requested configs coincide and the defect is invisible.
+    """
+    _, input_tensor_a = data_gen_with_range(input_shapes, 1, 100, device, True)
+    _, input_tensor_b = data_gen_with_range(input_shapes, 1, 100, device, True)
+    _, grad_tensor = data_gen_with_range(input_shapes, -50, 50, device)
+
+    input_tensor_a = ttnn.to_memory_config(input_tensor_a, ttnn.L1_MEMORY_CONFIG)
+    input_tensor_b = ttnn.to_memory_config(input_tensor_b, ttnn.L1_MEMORY_CONFIG)
+    grad_tensor = ttnn.to_memory_config(grad_tensor, ttnn.L1_MEMORY_CONFIG)
+
+    # no preallocated outputs, so the op allocates them through the helper
+    result = _TT_GRAD_OPS[op](
+        grad_tensor, input_tensor_a, input_tensor_b, are_required_outputs, ttnn.DRAM_MEMORY_CONFIG
+    )
+
+    for i, required in enumerate(are_required_outputs):
+        if not required:
+            continue
+        assert result[i] is not None
+        # compare the whole config, not just buffer_type: a wrong memory_layout or shard
+        # spec is the same class of bug, and empty_like is handed the full config
+        assert (
+            result[i].memory_config() == ttnn.DRAM_MEMORY_CONFIG
+        ), f"{op}_bw grad {i} was requested in {ttnn.DRAM_MEMORY_CONFIG} but landed in {result[i].memory_config()}"
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("are_required_outputs", [[True, True], [True, False], [False, True]])
+def test_bw_concat_helper_allocated_grads_honour_memory_config(input_shapes, device, are_required_outputs):
+    """concat_bw reaches its grads through ttnn.slice, a different write path from assign."""
+    _, input_tensor_a = data_gen_with_range(input_shapes, -100, 100, device, True)
+    _, input_tensor_b = data_gen_with_range(input_shapes, -100, 100, device, True)
+    _, grad_tensor = data_gen_with_range(torch.Size([1, 1, 64, 32]), -50, 50, device)
+
+    input_tensor_a = ttnn.to_memory_config(input_tensor_a, ttnn.L1_MEMORY_CONFIG)
+    input_tensor_b = ttnn.to_memory_config(input_tensor_b, ttnn.L1_MEMORY_CONFIG)
+    grad_tensor = ttnn.to_memory_config(grad_tensor, ttnn.L1_MEMORY_CONFIG)
+
+    result = ttnn.concat_bw(
+        grad_tensor,
+        input_tensor_a,
+        input_tensor_b,
+        0,
+        are_required_outputs=are_required_outputs,
+        memory_config=ttnn.DRAM_MEMORY_CONFIG,
+    )
+
+    for i, required in enumerate(are_required_outputs):
+        if required:
+            assert result[i].memory_config() == ttnn.DRAM_MEMORY_CONFIG
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("op", ["add", "sub", "div", "mul"])
+def test_bw_scalar_overload_grads_honour_memory_config(input_shapes, device, op):
+    """The tensor-scalar overloads allocate their own gradient and must honour memory_config.
+
+    add_bw and sub_bw scalar previously ignored output_mem_config outright (it was
+    commented out in the signature); div_bw and mul_bw scalar took it for the compute but
+    allocated with a bare empty_like. Existing scalar coverage passes no memory_config, so
+    inherited and requested never diverged.
+
+    Inputs must be in L1 with DRAM requested: with DRAM inputs the two coincide and the
+    defect is unobservable.
+    """
+    _, input_tensor = data_gen_with_range(input_shapes, 1, 100, device, True)
+    _, grad_tensor = data_gen_with_range(input_shapes, -50, 50, device)
+
+    input_tensor = ttnn.to_memory_config(input_tensor, ttnn.L1_MEMORY_CONFIG)
+    grad_tensor = ttnn.to_memory_config(grad_tensor, ttnn.L1_MEMORY_CONFIG)
+
+    fn = {"add": ttnn.add_bw, "sub": ttnn.sub_bw, "div": ttnn.div_bw, "mul": ttnn.mul_bw}[op]
+    result = fn(grad_tensor, input_tensor, 2.0, memory_config=ttnn.DRAM_MEMORY_CONFIG)
+
+    assert (
+        result[0].memory_config() == ttnn.DRAM_MEMORY_CONFIG
+    ), f"{op}_bw scalar overload was requested in DRAM but landed in {result[0].memory_config()}"
--- a/ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
+++ b/ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
@@ -31,14 +31,15 @@ void preallocated_tensors_check(
     std::optional<Tensor>& other_grad,
     const Tensor& input,
     const Tensor& other,
-    const std::array<bool, 2>& required_outputs) {
+    const std::array<bool, 2>& required_outputs,
+    const std::optional<MemoryConfig>& memory_config) {
     TT_FATAL(required_outputs[0] || required_outputs[1], "At least one gradient is expected to be calculated.");
 
     if (required_outputs[0] && !input_grad.has_value()) {
-        input_grad = ttnn::empty_like(input);
+        input_grad = ttnn::empty_like(input, std::nullopt, std::nullopt, std::nullopt, memory_config);
     }
     if (required_outputs[1] && !other_grad.has_value()) {
-        other_grad = ttnn::empty_like(other);
+        other_grad = ttnn::empty_like(other, std::nullopt, std::nullopt, std::nullopt, memory_config);
     }
 }
 
@@ -131,7 +132,7 @@ std::vector<std::optional<Tensor>> addalpha_bw(
     std::vector<std::optional<Tensor>> result = {std::nullopt, std::nullopt};
 
     operations::binary_backward::detail::preallocated_tensors_check(
-        input_grad, other_grad, input_a, other, {are_required_outputs[0], are_required_outputs[1]});
+        input_grad, other_grad, input_a, other, {are_required_outputs[0], are_required_outputs[1]}, output_mem_config);
 
     if (are_required_outputs[0]) {
         ttnn::assign(grad_tensor, input_grad.value());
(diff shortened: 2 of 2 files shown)


ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
```diff
@@ -31,14 +31,15 @@ void preallocated_tensors_check(
...
     const Tensor& input,
     const Tensor& other,
-    const std::array<bool, 2>& required_outputs) {
+    const std::array<bool, 2>& required_outputs,
+    const std::optional<MemoryConfig>& memory_config) {
     TT_FATAL(required_outputs[0] || required_outputs[1], "At least one gradient is expected to be calculated.");
 
     if (required_outputs[0] && !input_grad.has_value()) {
-        input_grad = ttnn::empty_like(input);
+        input_grad = ttnn::empty_like(input, std::nullopt, std::nullopt, std::nullopt, memory_config);
```
Aswinmcw, line 39: The helper is fixed, but the same file still has four gradient-output allocations with the identical defect — all in the **scalar overloads**: `add_bw` (:180), `sub_bw` (:235), `div_bw` (:668) and `mul_bw` (:821). The first two still have `output_mem_config` commented out in the signature; the latter two take it live and use it for the compute but allocate with a bare `empty_like`. After this PR `add_bw(grad, a, b, memory_config=DRAM)` honours the request while `add_bw(grad, a, 2.0, memory_config=DRAM)` does not, which is a harder split to reason about than the current uniform behaviour. Worth pulling into this PR, or filing so it does not sit half-fixed.

ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
```diff
@@ -492,13 +493,18 @@ std::vector<std::optional<Tensor>> assign_bw(
     const Tensor& input_tensor,
     const Tensor& other_tensor,
     const std::vector<bool>& are_required_outputs,
-    const std::optional<MemoryConfig>& /*output_mem_config*/,
+    const std::optional<MemoryConfig>& output_mem_config,
```
Aswinmcw, line 501: `assign_bw` and `concat_bw` are the two ops with the largest behaviour delta here — never honouring the config to always honouring it — and the new test only covers `add_bw`, whose parameter was already live. `concat_bw` in particular reaches its grads through `ttnn::slice` rather than `ttnn::assign`, so it is a different write path, not just a different caller. One case on each would cover what actually changed.

ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
```diff
@@ -517,13 +523,18 @@ std::vector<std::optional<Tensor>> concat_bw(
...
     std::vector<std::optional<Tensor>> grad_tensor = {std::nullopt, std::nullopt};
 
     operations::binary_backward::detail::preallocated_tensors_check(
-        input_grad, other_grad, input_tensor_a_arg, other, {are_required_outputs[0], are_required_outputs[1]});
+        input_grad,
+        other_grad,
+        input_tensor_a_arg,
+        other,
+        {are_required_outputs[0], are_required_outputs[1]},
+        memory_config);
```
Aswinmcw, line 542: Now that `memory_config` is a live parameter, the two `ttnn::slice` calls below (:547, :568) still pass `std::nullopt` for it. Harmless while the preallocated output wins, but it leaves the function half-threaded and is the same shape as the bug being fixed.

tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_add.py
```diff
@@ -122,3 +122,41 @@ def test_bw_add_scalar(input_shapes, scalar, device):
...
+        are_required_outputs=are_required_outputs,
+        memory_config=ttnn.DRAM_MEMORY_CONFIG,
+    )
+
+    for i, required in enumerate(are_required_outputs):
+        if not required:
+            continue
+        assert result[i] is not None
+        assert (
+            result[i].memory_config().buffer_type == ttnn.BufferType.DRAM
```
Aswinmcw: Asserting only `buffer_type` lets a wrong `memory_layout` or shard spec through. `assert result[i].memory_config() == ttnn.DRAM_MEMORY_CONFIG` costs nothing and pins the whole config, which is what `empty_like` is being handed.

ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
```diff
@@ -533,7 +549,7 @@ std::vector<std::optional<Tensor>> concat_bw(
             input_tensor_a_arg.logical_shape()[2],
             input_tensor_a_arg.logical_shape()[3]};
         ttsl::SmallVector<uint32_t> step = {1, 1, 1, 1};
-        ttnn::slice(grad_tensor_arg, start_index, end_index, step, std::nullopt, input_grad);
+        ttnn::slice(grad_tensor_arg, start_index, end_index, step, memory_config, input_grad);
```
Aswinmcw: Correcting my earlier suggestion here: threading `memory_config` into these two `slice` calls is inert. `slice.cpp:123` resolves the config as `optional_output_tensor.has_value() ? optional_output_tensor->memory_config() : memory_config_arg.value_or(...)`, so the preallocated tensor always wins — and inside `are_required_outputs[0]`, `input_grad` is guaranteed non-null because the helper just allocated it. The concat fix works entirely through the helper's `empty_like`; these arguments are dead and read as if they are doing the work. Either drop them or add a one-line comment saying the output tensor governs.

tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_add.py
```diff
@@ -122,3 +122,94 @@ def test_bw_add_scalar(input_shapes, scalar, device):
...
+    )
+
+    for i, required in enumerate(are_required_outputs):
+        if required:
+            assert result[i].memory_config() == ttnn.DRAM_MEMORY_CONFIG
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("are_required_outputs", [[True, True], [True, False], [False, True]])
+def test_bw_concat_helper_allocated_grads_honour_memory_config(input_shapes, device, are_required_outputs):
```
Aswinmcw, line 180: The three new tests all cover the TT (tensor-tensor) path, but the revision also changed four tensor-scalar overloads — `add_bw` :181, `sub_bw` :240, `div_bw` :674, `mul_bw` :828 — and none of them has coverage. `add_bw`/`sub_bw` scalar went from ignoring `output_mem_config` outright to honouring it, which is the same magnitude of behaviour change as `assign`/`concat`. `test_bw_add_scalar` passes no `memory_config`, so nothing pins this. One TS case (inputs L1, `memory_config=DRAM`) would close it — scalar paths are where this class of bug usually survives.

Aswinmcw (commented):
Description is now stale against the diff: it still says "threads the caller's config through all 8 call sites", but the revision also changes the four **scalar** overloads (`add_bw` :181, `sub_bw` :240, `div_bw` :674, `mul_bw` :828), un-comments `output_mem_config` in `add_bw`/`sub_bw`, and threads `memory_config` into `concat_bw`'s two `ttnn::slice` calls — none of which are mentioned. The Tests section also names one new test while three were added, and `sub` is missing from the verified-ops list even though its scalar overload now changes behaviour. Please refresh it before undrafting.

ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
Aswinmcw, line 39: This line is the fix and it's correct. My concern is the four **scalar** overloads that `28b721a6ac0` reverted back to a bare `empty_like` — commenting here since those lines are no longer in the diff.

For `add_bw` scalar (:180) the description's justification — "placement is decided downstream of `empty_like`" — can't apply, because there is no downstream step:

```cpp
input_grad = input_grad.value_or(ttnn::empty_like(input_tensor));  // inherits input's config
ttnn::assign(grad_tensor, input_grad.value());                     // writes into that buffer
result.emplace_back(input_grad);                                   // returns that same tensor
```

`assign` writes into the preallocated buffer rather than relocating it, and `output_mem_config` is commented out, so nothing between allocation and return can move the tensor. With `input_tensor` in L1 and `memory_config=DRAM`, this returns L1. `sub_bw` scalar (:235) has the same shape.

The reported capture (`total=2, L1=0, DRAM=2`) looks consistent with the inputs having been in **DRAM**, where inherit and honour coincide and the change is unobservable by construction. The condition that separates them is the one your three new TT tests already use: inputs in L1, DRAM requested.

`mul_bw` (:824) and `div_bw` (:671) scalar are a weaker case — `output_mem_config` does reach the compute call, so there is at least a candidate downstream step. But whether it can relocate an already-allocated output is the open question, and `slice.cpp:123` resolves exactly that conflict in favour of the preallocated tensor.

Could you re-run the scalar case with inputs in L1 and DRAM requested? If it returns DRAM the revert stands and I'll drop this. If it returns L1, please restore the four scalar changes and add one TS case — that was the intent of my previous comment, and it pins the behaviour either way.

Aswinmcw (commented):
The `slice` revert plus the explanatory comment is exactly right — that one was genuinely inert and is now settled.

The scalar revert I'd push back on. My earlier comment asked for a TS **test**, not for the fix to be removed; reverting it reinstates the tensor/scalar asymmetry I raised in the first round. Detail inline — happy to be proved wrong, but I don't think the current measurement is the discriminating one.

ttnn/cpp/ttnn/operations/eltwise/binary_backward/binary_backward.cpp
VirdhatchaniKN: You were right, and I have restored all four scalar changes in `28b26b52147`.

Re-ran exactly as you asked, `input_tensor` in L1 with `memory_config=DRAM`:

```
add_bw scalar   before -> L1     after -> DRAM
sub_bw scalar   before -> L1     after -> DRAM
div_bw scalar   before -> L1     after -> DRAM
mul_bw scalar   before -> L1     after -> DRAM
```

It returns L1, so the revert does not stand. Your read of `add_bw` was correct: `assign` writes into the buffer and there is no step that can relocate it.

Your diagnosis of my capture was also right — it was taken with DRAM inputs, where inherit and honour coincide. The reason it looked reproducible is worse than a one-off: reverting only the `empty_like` argument **does not compile**, because `output_mem_config` then goes unused in `add_bw`/`sub_bw` scalar, which is why it was commented out to begin with. My script ran `ninja ttnn >/dev/null 2>&1 && ninja install`, so the compile error was swallowed, `install` never ran, and the stale *fixed* library served both halves of the comparison. A valid control has to revert to main's actual form with the parameters re-commented so that it builds. With that:

```
revert the 4 scalar sites  ->  4 failed
with the fixes             ->  4 passed
```

The TS test is restored too — it was never vacuous, the control was.

The `concat_bw` `slice` arguments stay reverted, since that one rests on reading `slice.cpp:123` rather than on a build: `optional_output_tensor` wins and `input_grad` is always set by the helper above, and `concat_bw` still returns DRAM for L1 inputs without them. A comment records it.

The description is updated: the claim that there is no user-visible split was wrong and is gone.

Aswinmcw (approved):
LGTM. Scalar fixes restored with a TS case over all four ops, and the test discriminates for each — `add`/`sub`/`div` allocate from `input_tensor` and `mul` from `grad_tensor_arg`, and both are placed in L1. The `slice` omission is correctly reasoned and now documented in the code and the description.

Optional nit, not blocking: `empty_like(x, std::nullopt, std::nullopt, std::nullopt, mc)` appears six times now; a file-local wrapper taking just `(tensor, memory_config)` would make the positional chain harder to get wrong later.

mateusznowakTT (approved):

tests/ttnn/nightly/unit_tests/operations/eltwise/backward/test_backward_add.py
```diff
@@ -122,3 +122,121 @@ def test_bw_add_scalar(input_shapes, scalar, device):
...
+    )
+
+    for i, required in enumerate(are_required_outputs):
+        if required:
+            assert result[i].memory_config() == ttnn.DRAM_MEMORY_CONFIG
+
+
+@pytest.mark.parametrize("input_shapes", ((torch.Size([1, 1, 32, 32])),))
+@pytest.mark.parametrize("op", ["add", "sub", "div", "mul"])
+def test_bw_scalar_overload_grads_honour_memory_config(input_shapes, device, op):
```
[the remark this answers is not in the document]
VirdhatchaniKN: Valid — those five sites had no regression pinning them. Closed in `172b4beb2bc`.

Rather than copy the `add_bw` test five times, the tensor-tensor case is now table driven: `_TT_GRAD_OPS` maps a name to a call and the test is parameterised over it, so a new op added to the helper needs one line. That replaces the separate `add_bw`/`assign_bw` tests. `concat_bw` stays its own test since it needs a differently shaped grad.

Covered now, all with inputs in L1 and DRAM requested, no preallocated outputs, comparing the whole `memory_config`:

```
tensor-tensor   add, addalpha, subalpha, rsub, div, mul, assign   x 3 are_required_outputs
tensor-tensor   concat                                            x 3
tensor-scalar   add, sub, div, mul
```

Both halves verified to fail without their fix, and independently of each other:

```
revert the helper's two empty_like args   24 failed
revert the four scalar sites               4 failed
with both fixes                           28 passed
```

Full affected-op suite: 819 passed, 20 skipped.

One note for anyone reproducing this: the input placement matters. With DRAM inputs the inherited and requested configs coincide and the defect is invisible — the tests place inputs in L1 and request DRAM for that reason, and the docstring says so.

PR #54741, opened 2026-08-28
[Performance] Optimize fused Ring Indexer communication and scheduling

pavlejosipovic:
### PR Category

Performance

### Summary

This PR improves the fused Ring Indexer pipeline in three trace-stable layers.

#### 1. Coalesce Ring Attention all-gather DRAM and Fabric traffic

- Give workers disjoint destination-DRAM-bank ownership instead of walking pages in logical tensor order.
- Read pages in each bank's physically contiguous order and pack each Fabric transfer up to the available payload.
- Rotate packet issue across banks so independent DRAM banks remain active.
- Prefetch batches of up to four packets and use double-buffered reader CB storage to overlap the next DRAM batch with Fabric drain.
- Split the diametric shard across both ring directions without splitting a bank stream.
- Select the bank-owned path from static capabilities: Blackhole, sequence-axis gather, interleaved DRAM layout, and compatible page/payload geometry.
- Derive pages per packet from the configured payload, so both the default limit and the production 14 KiB Fabric payload use the same implementation. With BFP8 tiles, 14 KiB carries 13 complete 1,088-byte pages instead of the old two-page scatter cap.

#### 2. Start compute from local and partially arrived shards

- Replace the old global logical-K walk with shard-major KC work units, so units fully contained in the initially local shard can run before the first all-gather round completes.
- Keep units inside physical shard boundaries while preserving their logical score destinations.
- Use a static `KC=10` unit (320 keys), aligned to the all-gather row geometry and independent of runtime `kv_len`.
- A post-firmware Ring-8 sweep of every fitting KC from 1 through 15 confirmed KC=10 as the unique 55K optimum. KC=15 improved 512K by only 0.6% while regressing 55K by 40%; KC=20 exceeded L1.
- Add a row-aligned midpoint readiness marker in addition to the existing shard-completion marker.
- Synchronize the midpoint across all links before publishing it, so a consumer never observes a partially completed ready region.
- Permit a whole KC unit contained in the first half to use the midpoint marker; units crossing the midpoint or in the second half still require completion.
- Interleave equal offsets from the two shards arriving in a paired forward/backward wave (`A0, B0, A1, B1, ...`), allowing compute to move to the other ready half instead of waiting for one complete shard.
- Update valid page counts, readiness, and work bounds on device from the runtime prefix while reusing the same cached program and trace.

#### 3. Balance Ring Indexer work across compute lanes

- Derive a cyclic source-column shift from ring-wave and compute-grid geometry instead of using Ring-4/Ring-8 lookup cases.
- Apply one shift to both shards in a paired arrival wave, preserving their interleaved readiness order.
- Spread repeated valid-prefix remainders across block-column lanes rather than repeatedly starting the extra work at lane zero.
- Preserve row blocks and adjacent-block DRAM pairing; only the source-column assignment is permuted.
- Keep offsets within every lane monotonic and assign every KC unit exactly once.
- Leave singleton local and diametric waves unshifted, and leave Linear/Ring-2 scheduling unchanged.
- Derive the schedule from static topology, capacity, phase, and grid geometry—not runtime `kv_len`—so it remains program-cache/trace stable and extends naturally to larger rings such as Ring-32.

There is no runtime `kv_len` or tensor-capacity heuristic. Runtime valid prefixes change device-side bounds and readiness only; they do not select a different program.

### Performance guide

[Open the rendered performance guide](https://htmlpreview.github.io/?https://gist.githubusercontent.com/pavlejosipovic/be696451c612f61c6c0b0f08f7d401c3/raw/97998c0c47ddf7629b6aac156955e0d338a1a652/ring-indexer-performance-guide.html)
([gist source](https://gist.github.com/pavlejosipovic/be696451c612f61c6c0b0f08f7d401c3)).

The scrolling HTML guide provides the detailed visual walkthrough behind the summary above:

- the fused DRAM → all-gather → Fabric → score-reader → compute pipeline;
- Blackhole's eight-bank page interleave, disjoint worker bank ownership, diametric-shard split, bank-local packet construction, packet-level bank rotation, four-packet prefetch, and double buffering;
- why the 14 KiB payload matters and how 13 BFP8 tile pages fit in one packet;
- why origin/main could not execute even initially local data: its logical KC units crossed source-shard boundaries;
- the new shard-major local-first KC layout, midpoint/completion readiness protocol, and paired-direction arrival order;
- the 10×10 and 11×10 measured compute grids, their 20/22 block-column scheduling lanes, the repeated-remainder imbalance, and the geometry-derived paired-wave permutation;
- the static KC=1–15 sweep and its compute-granularity/L1 trade-off;
- per-commit invariants, isolated scheduling ablations, and the end-to-end origin/main-versus-latest measurements.

### End-to-end performance

The comparable production-configuration A/B is LoudBox, Ring-8/SP=8, explicit 14 KiB Fabric payload, and warm traced fused-program duration. The latest values are the median of three independent test invocations, each using the median of three profiler replays.

| KV prefix | `origin/main` | Latest | Latency reduction | End-to-end speedup | FPU model |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 55K | 0.338 ms | 0.204 ms | 39.6% (0.134 ms) | 1.66× | 35.13% → 58.10% |
| 512K | 2.870 ms | 1.867 ms | 34.9% (1.003 ms) | 1.54× | 39.28% → 60.39% |

These absolute-duration comparisons use 14 KiB on both sides. The old `origin/main` targets were recorded with the default Fabric payload, so target-to-target movement is not a valid speedup measurement. FPU utilization here is ideal LoFi work divided by measured end-to-end duration, not a hardware-counter reading.

The paired-wave scheduler was also isolated on the same QuietBox runner with seven warm trace replays per mode. At the pathological 56,320-token remainder it reduced latency from 0.285490 ms to 0.253710 ms (11.13%); at 524,288 tokens it changed 1.976846 ms to 1.972230 ms (0.23%). The worst point across the 21-prefix sweep was a 1.34% regression, inside the symmetric 2% guard. The relevant performance job passed in [run 33313112257](https://github.com/tenstorrent/tt-metal/actions/runs/33313112257/job/99262729001); the workflow's overall red result came from a separate SDPA-nightly job.

The checked-in warm traced-median CI contract uses a symmetric ±2% band and the 14 KiB payload:

| Blackhole box | 55K target | 512K target |
| --- | ---: | ---: |
| QuietBox, Ring-4 | 52.22% | 62.99% |
| LoudBox, Ring-8 | 58.10% | 60.39% |

### Validation

- [T3K E2E CCL](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch%3Apjosipovic/ring-indexer-dram-coalescing-multiworker+event%3Aworkflow_dispatch)

- [Blaze Models Prefill — GLM 5.2](https://github.com/tenstorrent/tt-metal/actions/runs/33768863554)

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:pjosipovic/ring-indexer-dram-coalescing-multiworker)

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
+++ b/tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
@@ -854,9 +854,8 @@ def test_indexer_score_full_mesh_rejects_invalid_contracts(expect_error):
 def test_indexer_score_ring8_partial_readiness_reference_cache_hit(mesh_device):
     """Reference-check the production Ring two-marker protocol, including its cache-hit runtime patch.
 
-    The smaller correctness shapes above intentionally use the legacy packet schedule. This test keeps query
-    work small but makes the BF16 K capacity large enough that forwarded payload exceeds the 20 MiB bank-owned
-    threshold. Two runtime prefixes then exercise different midpoint/completion locations on one cached program.
+    This test keeps query work small while using a large BF16 K capacity to exercise the bank-owned schedule's
+    midpoint/completion protocol. Two runtime prefixes exercise different marker locations on one cached program.
     Sampled first/last rows on every rank cover both directions without constructing a capacity-sized CPU score.
     """
     sp = mesh_device.shape[0]
@@ -866,10 +865,6 @@ def test_indexer_score_ring8_partial_readiness_reference_cache_hit(mesh_device):
     k_capacity = 128 * 1024
     kv_lens = (56320, 112640)
     dim = 128
-    bf16_bytes = 2
-    forwarded_capacity_bytes = k_capacity // sp * dim * bf16_bytes * (sp - 1)
-    assert forwarded_capacity_bytes > 20 * 1024 * 1024, "test capacity must activate bank-owned partial readiness"
-
     grid = mesh_device.compute_with_storage_grid_size()
     worker_cores = ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(grid.x - 1, grid.y - 1))})
     subdevice_id = ttnn.SubDeviceId(0)
--- a/tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
+++ b/tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
@@ -60,6 +60,21 @@ pytestmark = [
 ]
 
 
+def _ring8_partial_router_config():
+    config = ttnn.FabricRouterConfig()
+    config.max_packet_payload_size_bytes = 14 * 1024
+    return config
+
+
+_RING8_PARTIAL_DEVICE_PARAMS = {
+    "fabric_config": ttnn.FabricConfig.FABRIC_2D_TORUS_XY,
+    "reliability_mode": ttnn.FabricReliabilityMode.STRICT_INIT,
+    "fabric_tensix_config": ttnn.FabricTensixConfig.DISABLED,
+    "fabric_router_config": _ring8_partial_router_config(),
+    "require_exact_physical_num_devices": True,
+}
+
+
 def _fused_dev_inputs(submesh, q_g, w_g, k_host, *, k_dtype=ttnn.bfloat16):
     """Fused op inputs: SP-shard q/w (bf16) on dim 2, SP-shard k_local (the AG input), and a zero-seeded
     gathered buffer (AG fills remote bands; zeros prove the reader dual-sources the local band). k_dtype sets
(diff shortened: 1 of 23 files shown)


ipotkonjak-tt (approved):
approved assuming [Blaze Models Prefill tests](https://github.com/tenstorrent/tt-metal/actions/workflows/blaze-models-prefill-tests.yaml) passes

if the perf is improved maybe glm perf test targets need adjusting

ttnn/cpp/ttnn/operations/experimental/indexer_score/device/ring_indexer_score_schedule.hpp
```diff
@@ -0,0 +1,102 @@
...
+    // KC-unit tails more broadly than a unit stride while keeping neighboring waves close on the physical grid.
+    const uint32_t column_stride = std::max(1u, column_count / (wave_count + 1));
+    return ((wave + wave / 2) * column_stride) % column_count;
+}
+
+// Build each (row-block, column) lane's shard-major list of physical K-tile starts. Rotation changes only which
+// column residue a lane owns in paired arrival waves. It is a cyclic column permutation within each row block, so
+// every KC unit remains in exactly one lane, adjacent row blocks retain their paired DRAM access pattern, and
+// paired shards retain matching, monotonically increasing offsets.
+inline WorkList make_work_list(
```
skrsticTT, line 66: does it make sense to add host unit test that verifies rotation and that we scheduled everything right? guess that currently we test perf and accuracy on some specific cases but we don't prove this is done right

tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
```diff
@@ -833,6 +848,119 @@ def _call(**overrides):
...
+    sp = mesh_device.shape[0]
+    heads = 4
+    q_per_rank = 32
+    chunk_global = sp * q_per_rank
+    k_capacity = 128 * 1024
+    kv_lens = (56320, 112640)
+    dim = 128
+    bf16_bytes = 2
+    forwarded_capacity_bytes = k_capacity // sp * dim * bf16_bytes * (sp - 1)
+    assert forwarded_capacity_bytes > 20 * 1024 * 1024, "test capacity must activate bank-owned partial readiness"
```
skrsticTT: not sure from where this 20mib comes from?

ttnn/cpp/ttnn/operations/experimental/indexer_score/device/ring_indexer_score_schedule.hpp
```diff
@@ -0,0 +1,102 @@
...
+    ArrivalWaves waves(ring_size / 2 + 1);
+    RingIdSequencer seq(ring_index, ring_size, writes.backward_writes_expected, writes.forward_writes_expected);
+    for (uint32_t iteration = 0; iteration < ring_size; ++iteration) {
+        const uint32_t shard = seq.get_next_ring_id([](uint32_t, uint32_t) {});
+        waves[(iteration + 1) / 2].push_back(shard);
+    }
+    return waves;
+}
+
+inline ArrivalWaves arrival_waves(uint32_t ring_size, uint32_t ring_index, ttnn::ccl::Topology topology) {
```
skrsticTT: this one func is not used?

ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_writer.cpp
```diff
@@ -199,76 +287,122 @@ void kernel_main() {
...
+            } else {
+                tile_id_start = my_tensor_rank * input_pages_per_batch_head;
+            }
+
+            for (uint32_t bh_idx = 0; bh_idx < input_batch_head_count[input_idx]; bh_idx++) {
+                while (tiles_read < tiles_to_read) {
+                    uint32_t num_pages_to_read = std::min(tiles_to_read - tiles_read, packet_size_in_pages);
+                    cb_output.wait_front(packet_size_in_pages);
+                    const size_t l1_read_addr_base = cb_output.get_read_ptr();
+                    size_t l1_read_addr = l1_read_addr_base;
```
skrsticTT: this can be one line assign

ttnn/cpp/ttnn/operations/experimental/indexer_score/device/kernels/compute_indexer_score.cpp
```diff
@@ -613,7 +665,8 @@ void kernel_main() {
 
             k.pop_front(k_chunk_tiles);
         }
-        // group's bands done: release its resident q/w (one block per group).
+        // Group's units are done: release its resident q/w block. When every unit is kv-empty, no compute path
+        // dereferences q/w; pop_front intentionally returns the matching CB credit without an extra wait.
         CircularBuffer(cb_w).pop_front(w_group_tiles);  // gates waited in the mul/scale phase
```
skrsticTT, line 670: can we have case where we do pop without wait?

skrsticTT (approved):

tests/ttnn/nightly/unit_tests/operations/experimental/indexer_score/test_ring_indexer_score_dsa.py
pavlejosipovic: stale stuff, I used to have a heuristc based on this amount of data but I've removed it

ttnn/cpp/ttnn/operations/experimental/ccl/ring_attention_all_gather_async/device/kernels/ring_attention_all_gather_writer.cpp
pavlejosipovic: yep

ttnn/cpp/ttnn/operations/experimental/indexer_score/device/ring_indexer_score_schedule.hpp
pavlejosipovic: yep

ttnn/cpp/ttnn/operations/experimental/indexer_score/device/kernels/compute_indexer_score.cpp
pavlejosipovic: Yes, when every unit is KV-empty. The reader still pushes the resident w group, but no compute path dereferences it, so the pop only returns that matching CB credit. I clarified this invariant beside the pop.

PR #54752, opened 2026-08-28
🚧 Add topology solver tests for a positional GlobalNode

p1-0tr:
The topology solver is generic over its GlobalNode parameter: node identity is erased to dense size_t indices in GraphIndexData before any search runs, so topology_solver.tpp references ASICs only in two comments and topology_solver_sat.cpp not at all. These tests establish that mechanically by instantiating the whole solver on a (hostname, tray_id, asic_location) node instead of AsicID, with no change to any library source.

Motivation is the open question of whether the topology mapper should switch its physical node identity from AsicID (an opaque uint64 from UMD chip_unique_ids, or synthesized in physical_descriptor_builder.cpp) to the position tuple. Two tests pin down the payoff:

  - position traits are read off the node, so the AsicPositionMap side table in TopologyMappingConfig::asic_positions and the tray/location trait maps built in physical_grouping_descriptor_matching.cpp have nothing to translate.
  - host partitions are read off the node, so TopologyMappingConfig::hostname_to_asics is derivable rather than an input the caller must supply.

Two tests record costs rather than wins: node index order (which drives SAT variable numbering and DFS candidate ordering, so it selects among equally valid mappings) becomes positional, meaning existing tests asserting a specific mapping would need re-baselining; and with a string host key that order is lexicographic rather than natural, so host10 precedes host2 and the order is not the rack order a locality heuristic would assume.

Test-only; no library files touched, and already covered by tests/tt_metal/tt_fabric/sources.cmake. Not yet compiled - authored on macOS without a tt-metal build environment, so it needs a CI or local build pass before being relied on.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:p1-0tr/test-asic-location-tuple-in-solver)

### DIFF
--- a/tests/tt_metal/tt_fabric/fabric_router/test_topology_solver.cpp
+++ b/tests/tt_metal/tt_fabric/fabric_router/test_topology_solver.cpp
@@ -9,7 +9,9 @@
 #include <map>
 #include <memory>
 #include <set>
+#include <string>
 #include <tuple>
+#include <unordered_map>
 #include <vector>
 #include <gtest/gtest.h>
 #include <tt-logger/tt-logger.hpp>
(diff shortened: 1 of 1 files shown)


p1-0tr:
We'll go with a different type to identify ASICs, closing

Issue #54691, opened 2026-08-28
untilize writer rebuilds the row_stick_ids array once per tile when it changes once per row


AJ0070:
**Component / Area:** kernels

**Issue Type:** Performance

## Observed

`writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp` rebuilds the same
`row_stick_ids` array once per tile, when its contents only change once per row of tiles.

[`writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp:56-66`](https://github.com/tenstorrent/tt-metal/blob/main/ttnn/cpp/ttnn/operations/data_movement/untilize/device/kernels/dataflow/writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp#L56-L66):

```cpp
for (uint32_t i = 0; i < num_sticks / tile_height; i++) {
    for (uint32_t tile_id = 0; tile_id < num_tiles_per_core; tile_id++) {
        for (uint32_t j = 0; j < tile_height; j++) {
            row_stick_ids[j] = stick_id + j;      // depends only on stick_id and j
        }
        curr_stick_offset = curr_offset;
        write_tiles(1, tile_width_size);
        curr_offset += tile_width_size;
    }

    stick_id += tile_height;                       // advances only in the outer loop
}
```

`stick_id` is assigned once before the loop nest and advances only at the end of each `i`
iteration. It is invariant across the entire `tile_id` loop, so the inner `j` loop writes the
identical 32 values on every one of the `num_tiles_per_core` passes.

Nothing between the passes invalidates the array. `write_tiles` only reads it
(`row_stick_ids[k]` at line 46) and mutates a local read pointer; `row_stick_ids` is declared at
line 34 and written at line 59 and nowhere else in the file. What genuinely varies per `tile_id`
is `curr_stick_offset` and `curr_offset`, neither of which feeds the array.

With `tile_height = 32` (constexpr, line 15), the redundant work is

```
(num_sticks / tile_height) * (num_tiles_per_core - 1) * 32   scalar stores
```

all of them recomputing values already present in the array.

## Expected

Hoist the `j` loop above the `tile_id` loop so the array is built once per row of tiles:

```cpp
for (uint32_t i = 0; i < num_sticks / tile_height; i++) {
    for (uint32_t j = 0; j < tile_height; j++) {
        row_stick_ids[j] = stick_id + j;
    }
[...]

PR #54753, opened 2026-08-28
Hoist redundant `row_stick_ids` build out of the tile loop in untilize

mcw-anasuya:
### Ticket
https://github.com/tenstorrent/tt-metal/issues/54691

### Summary
The untilize parallel-columns writer rebuilt `row_stick_ids` once per tile, but the table depends only on `stick_id`, which advances once per row of tiles. Hoisting it out of the tile loop removes `tile_height * (num_tiles_per_core - 1)` redundant stores per row.

### Measured Performance
Tested across 5 runs, best-case (min) BRISC times:
| **shape** | **tiles/core** | **before (ns)** | **after (ns)** | **Δ (ns)** | **Δ (%)** |
|---|---|---|---|---|---|
| [1,1,32,512] | 4 | 8573 | 8087 | −486 | −5.7% |
| [1,1,32,1024] | 8 | 16579 | 15423 | −1156 | −7.0% |
| [1,1,32,2048] | 16 | 32707 | 30114 | −2593 | −7.9% |
| [1,1,32,4096] | 32 | 64989 | 59632 | −5357 | −8.2% |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:anasuya/untilize_stick_id)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:anasuya/untilize_stick_id)

### DIFF
--- a/ttnn/cpp/ttnn/operations/data_movement/untilize/device/kernels/dataflow/writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp
+++ b/ttnn/cpp/ttnn/operations/data_movement/untilize/device/kernels/dataflow/writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp
@@ -54,10 +54,10 @@ void kernel_main() {
 
     uint32_t curr_offset = offset_within_stick;
     for (uint32_t i = 0; i < num_sticks / tile_height; i++) {
+        for (uint32_t j = 0; j < tile_height; j++) {
+            row_stick_ids[j] = stick_id + j;
+        }
         for (uint32_t tile_id = 0; tile_id < num_tiles_per_core; tile_id++) {
-            for (uint32_t j = 0; j < tile_height; j++) {
-                row_stick_ids[j] = stick_id + j;
-            }
             curr_stick_offset = curr_offset;
             write_tiles(1, tile_width_size);
             curr_offset += tile_width_size;


ttnn/cpp/ttnn/operations/data_movement/untilize/device/kernels/dataflow/writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp
```diff
@@ -54,10 +54,10 @@ void kernel_main() {
 
     uint32_t curr_offset = offset_within_stick;
     for (uint32_t i = 0; i < num_sticks / tile_height; i++) {
+        for (uint32_t j = 0; j < tile_height; j++) {
+            row_stick_ids[j] = stick_id + j;
```
Aswinmcw, line 58: `experimental/quasar/untilize/.../writer_unary_stick_layout_split_rows_interleaved_parallel_columns_metal2.cpp` is a copy of this kernel with the identical inner `j` loop and would take the same hoist verbatim. Worth folding in here so the two do not diverge, or a follow-up if you would rather keep this PR to one file.

Aswinmcw (approved):
Correctness verified: `stick_id` only advances at the end of the outer loop and `write_tiles` never writes `row_stick_ids`, so the hoist is semantically identical. One thing on the numbers below — the saving per tile is a fixed 32 stores against a fixed per-tile cost (32 NOC writes + barrier), so the percentage should be roughly shape-independent; 8.3% / 1.6% / 2.1% instead spans 5x. Worth a couple of repeat runs to see whether the 8.3% row is real or run-to-run variance.

ttnn/cpp/ttnn/operations/data_movement/untilize/device/kernels/dataflow/writer_unary_stick_layout_split_rows_interleaved_parallel_columns.cpp
mcw-anasuya: Issue does not mention `experimental/quasar/`. Not touching it yet.

amorrisonTT (approved):

PR #54776, opened 2026-08-28
conv3d: fp32-exact output path; retire tap_matmul

### LINKED ISSUE #54878: Fixes three TF32-rounding defects
The same idea applied to the dense conv3d kernel: fixes three TF32-rounding defects (partial reduction,  bias add, untilize) behind a dtype-gated flag, making conv3d+split match the tap-matmul reformulation — so tap_matmul is deleted (~99 ms/decode + restores T-parallel scaling). Ends  with the question to the LLK team that decides whether the last lever, split_mode, can die too.

rouzbehTT:
## Summary
Stacked on `rouzbeh/improve-audio-kernels` (#54675, the depthwise conv1d fix). Same idea applied to the dense convs in MiniMax-H3's audio decoder: push fp32 precision into the conv3d kernel, then delete the model-code workaround it obsoletes.

Two commits:
1. **Kernel**: three TF32-rounding defects in conv3d's fp32 tail, fixed behind a dtype-gated `use_fp32_exact` (fp32 dtype + fp32 dest, no flags): the C_in-block partial reduction re-rounded the running partial through SrcA/SrcB on every add (now SFPU with a dedicated fp32 accumulator CB); the bcast-rows bias add rounded the finished output the same way (now `copy_tile` + `unary_bcast<ROW>` + SFPU add); untilize re-read the output through SrcA (interm CB now `UnpackToDestFp32`). bf16-dtype callers keep the original FPU path byte-identical.
2. **Retire `tap_matmul`**: it reformulated stride-1 convs as K per-tap matmuls purely to dodge those roundings. conv3d+split now matches it, so it's removed from `Conv1dViaConv3d`, the constructor chain, and `weights_variant` (new cache suffix, no stale-cache collision).

## Measured (rel RMSE vs float64 golden, Blackhole)
| shape | conv3d+split before | after | tap+split (removed) |
|---|---|---|---|
| conv_pre 2048→1024 k7 | 9.98e-4 | 2.85e-4 | 2.83e-4 |
| AMP 512 k3 d1/3/5 | 6.99e-4 | 2.90e-4 | 2.74e-4 |
| conv_pre @ C_in_block=32 | 1.66e-3 | 2.85e-4 | == 128 (blocking decoupled from accuracy) |

Plain fp32 conv3d also improved 1.86e-3 → 1.14e-3, now equal to plain `ttnn.matmul`.

## Performance (Blackhole, d03u02; layer-level via Conv1dViaConv3d, fp32, bias, 50-iter mean)
| shape | before (this base) | after (exact tail, 9ba0ccd) | tap+split (removed) |
|---|---|---|---|
| conv_pre 2048->1024 k7, split off | 3.81 ms | 4.42 ms (+16%) | -- |
| conv_pre, split full | 11.6 ms | 13.2 ms (+14%) | 4.7 ms |
| AMP 512 k3 (d1/d3/d5), split off | ~0.98 ms | ~1.21 ms (+24%) | -- |
| AMP, split full | ~3.0 ms | ~3.5 ms (+17%) | ~1.9 ms |
| amp_narrow 32->32 k3 | 89 us | 98 us | -- |
| streaming-output probe (raw conv3d, 128B writes, M_t=2) | 136 us | 131 us (-3%) | -- |

Sub-150 us rows fluctuate +/-10% run to run (dispatch/clock noise); the remap-toggle removal itself measured perf-neutral (big shapes 0.4-1% faster, within variance).

- The +14-25% on fp32 convs is the SFPU tail itself (per-tile copy/add + fp32 unpack); bf16-dtype convs are byte-identical. The Blackhole `llk_math_reconfig_remap` toggles that originally guarded the SFPU sections were measured at parity (the streaming probe was their worst case, one pair per tile-row) and then removed outright in 9ba0ccd -- current BH LLKs use logical DST tile indices consistently across copy/bcast/SFPU/pack, verified bit-identical on BH silicon.
- Per-op, retired tap+split was 2-2.8x faster than exact conv3d+split at the two large shapes. It is still removed because: it only covered stride-1, non-channel-sharded convs (the exact tail covers every conv3d caller, which is where the e2e fidelity gain comes from), it carried K per-tap weight copies, and e2e the branch is net faster anyway -- H3 audio decode accurate mode 7.6 s @ 44.7 dB (tap+split + MAC depthwise) -> 7.3 s @ 67.6 dB (this branch default, re-certified at 9ba0ccd; the bias-CB UnpackToDestFp32 fix added ~2 dB on BH over the earlier 65.5 dB measurement).
- Wormhole: correctness is CI-verified (nightly fp32-exact gates green on wh_n150/wh_n300); WH latency is unmeasured -- no WH hardware on our side. tap_matmul never targeted WH.

## Verified
Real-weight e2e on the H3 audio VAE: `test_decode` (accurate-mode PSNR/rmse/PCC bars, previously reachable only via tap_matmul), `test_encode`, `test_roundtrip`, `test_depthwise_chunked_conv1d_matches_torch`, `test_conv_operand_split_improves_precision` — all pass.

## Question for the conv/LLK team
With both operands exactly bf16-representable (nothing left to truncate at unpack), a matmul product still rounds at ~2^-11.6 relative — flat in reduction depth, present at single-K-tile shapes (hi@hi carries 3.097e-4 while hi@lo / lo@hi are ~1e-6). Is that per-product/phase rounding configurable? If yes, the remaining `split_mode` lever can be deleted outright; if not, splitting (any flavor) is capped at ~2.8e-4 and `split_mode` stays as the one remaining model-code lever.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rouzbeh/audio-exact-fp32-dense)


rouzbehTT:
Addressed the review comments in b07f76528b0:

- **Stale local-CB comment (:155)**: correct — that text described the interim design where the interm CB stayed on the default unpack path. The final design flags interm/reduction/acc `UnpackToDestFp32`; the comment now says so. The same interim-design leftover existed in the factory's accumulator-CB comment ("the only CBs delivered UnpackToDestFp32") — fixed there too, and the surrounding blocks trimmed.
- **Direct test coverage for the exact tail**: added `test_conv3d_fp32_exact_tail` — true fp32 end to end, HiFi4, bias on, scored against a float64 golden, at `C_in_block=full` (single block: bias/untilize branch) and `=32` (4 blocks: reduction branch). Two gates: rel RMSE ≤ 1.35e-3 per blocking, and ≤5% error spread between blockings. Verified both directions on Blackhole: with the fix 1.155e-3 at both blockings, 0.0% spread; with the fix reverted 1.507e-3 / 1.638e-3 and 8.7% spread — both gates fail. (The equivalent conv1d gates landed on the base PR #54675.)

rouzbehTT:
Sanity CI caught a real regression in 598f6b4c270's predecessor: the earlier rewiring repurposed kernel arg 27 from `use_fp32_partials` to `use_fp32_exact`, which silently dropped the format reconfigs the **bf16-dtype + fp32-partials** path needs (fp32-format interm CB under a bf16 output) — the pre-existing conv3d sweeps failed with garbage PCC on both WH and BH. The all-fp32 audio testing could not have seen it.

Fixed by carrying both flags: arg 27 is `use_fp32_partials` again (reconfigs key on either flag), new arg 30 is `use_fp32_exact` (SFPU tail + flagged-CB reads only), so the bf16 FPU path matches main. Full conv3d unit suite green locally (36 passed, incl. the failing sweeps and the fp32-exact gates). CI re-dispatched: sanity https://github.com/tenstorrent/tt-metal/actions/runs/33300911679, L2-nightly conv https://github.com/tenstorrent/tt-metal/actions/runs/33300913090 (the earlier red runs were on the broken commit).

ttnn/cpp/ttnn/operations/experimental/conv3d/device/conv3d_program_factory.cpp
```diff
@@ -266,6 +271,22 @@ tt::tt_metal::ProgramDescriptor Conv3dProgramFactory::create_descriptor(
         });
     }
 
+    // fp32 running-partial accumulator for the exact reduction (see reduce_block_fp32_sfpu in
+    // the compute kernel); avoids round-tripping the running partial through the interm CB.
+    uint32_t cb_reduction_acc_tiled_id = 32;
+    if (use_fp32_partials && use_fp32_exact) {
```
pavlejosipovic, line 277: Please include this allocation in the L1 prefetch budget: `cb_reduction_acc_tiled` adds `matmul_M_t * matmul_N_t * partial_tile_size` bytes on the exact multi-C_in-block path, but `other_cbs_bytes` only charges the existing reduction CB and worker-ack CB. Without the additional charge when `use_fp32_partials && use_fp32_exact`, `l1_prefetch_max_bytes` can admit a shard allocation that overcommits L1.

ttnn/cpp/ttnn/operations/experimental/conv3d/device/kernels/compute.cpp
```diff
@@ -144,11 +145,104 @@ void add_block_inplace_math(uint32_t inout_cb, uint32_t add_cb) {
...
+template <uint32_t rows, uint32_t cols>
+void add_bias_inplace_sfpu(uint32_t inout_cb, uint32_t bias_cb) {
+    CircularBuffer inout_cb_obj(inout_cb);
+    CircularBuffer bias_cb_obj(bias_cb);
+    constexpr uint32_t DST_ACC = 0;
+    constexpr uint32_t DST_OPERAND = 1;
+    inout_cb_obj.wait_front(rows * cols);
+    bias_cb_obj.wait_front(cols);
+#ifdef ARCH_BLACKHOLE
+    MATH((llk_math_reconfig_remap(false)));
```
pavlejosipovic: Could we quantify the Blackhole cost here and consider hoisting remap transitions to a phase boundary? `llk_math_reconfig_remap` performs a Tensix sync and waits for pending packs before changing the stride-of-16/untilize mapping. In the streaming-output path this helper runs per subblock, so the `false`/`true` pair may add a synchronization point for every small output block. Disabling remap is semantically needed while SFPU/datacopy uses linear DST slots, but it would be good to demonstrate that repeated toggling does not regress the relevant BH paths (or restructure it so it happens once around a batch of SFPU work).

pavlejosipovic:
Please document the performance impact in the PR description (or commit message) before retiring `tap_matmul`: latency for old normal Conv3D, new exact Conv3D, and the old `tap_matmul` workaround on the representative H3 shapes, ideally on both Blackhole and Wormhole. The PR currently establishes the numerical improvement, but the new path introduces per-tile SFPU work and, on Blackhole, remap transitions that synchronize Tensix, so the replacement’s throughput/latency trade-off should be explicit.

tests/ttnn/unit_tests/operations/conv/test_conv3d.py
```diff
@@ -799,3 +799,92 @@ def test_conv3d_halo_buffer_rejects_unsupported(device, expect_error, padding_mo
             config=config,
             halo_buffer=halo_buffer,
         )
+
```
pavlejosipovic: I do not think we need to run this precision gate in both suites. Please move the full test implementation to the nightly Conv3D file and remove the unit-test copy/wrapper import. This is a hardware/numerics regression rather than a fast unit gate, and keeping one canonical test avoids the unusual pattern of importing and invoking a `test_*` function from another suite.

tests/ttnn/unit_tests/operations/conv/test_conv3d.py
```diff
@@ -799,3 +799,92 @@ def test_conv3d_halo_buffer_rejects_unsupported(device, expect_error, padding_mo
...
+    H_out = _out_size(H, padding[1], 1, kernel_size[1], 1)
+    W_out = _out_size(W, padding[2], 1, kernel_size[2], 1)
+
+    kernel_config = ttnn.init_device_compute_kernel_config(
+        device.arch(),
+        math_fidelity=ttnn.MathFidelity.HiFi4,
+        math_approx_mode=False,
+        fp32_dest_acc_en=True,
+        packer_l1_acc=False,
+    )
```
pavlejosipovic: Please add an FP32 case that exercises the streaming-output branch too. This test uses the default spatial output blocks (`T_out_block/H_out_block/W_out_block = 1`), so `matmul_M_t` is one and `enable_streaming_output` remains off. A multi-tile spatial output block with a single C_in block would cover the newly added streaming SFPU bias/untilize path.

ttnn/cpp/ttnn/operations/experimental/conv3d/device/conv3d_program_factory.cpp
rouzbehTT: Good catch, fixed in commit 52367f0a361

ttnn/cpp/ttnn/operations/experimental/conv3d/device/kernels/compute.cpp
rouzbehTT: Quantified on BH (table in the PR description).

tests/ttnn/unit_tests/operations/conv/test_conv3d.py
rouzbehTT: Done in 31ef88ad63e.

tests/ttnn/unit_tests/operations/conv/test_conv3d.py
rouzbehTT: Added in commit 27b78308c0b. Config forces enable_streaming_output (single C_in block, 128B writes, matmul_M_t=2)

1.500e-3 without the fix, 1.150e-3 with it, bound 1.35e-3.

pavlejosipovic:
Pushed 9ba0ccd6e27 to remove the Blackhole DEST-remap disable/restore pairs around the fp32-exact reduction and bias helpers.

The current BH LLKs use logical DST tile indices consistently across `copy_tile`, `unary_bcast<ROW>`, `add_binary_tile`, matmul, and pack. Leaving remap enabled produced identical results in the full-block reduction, bias, streaming, and temporary no-bias reduction paths:

- `test_conv3d_fp32_exact_tail`: passed at C_in_block 128 and 32, both 1.155e-03 relative RMSE
- `test_conv3d_fp32_exact_tail_streaming`: passed at 1.172e-03 relative RMSE
- temporary no-bias reduction isolation: passed at 1.155e-03 for both blockings

The same reasoning likely applies to `rgb_to_yuv::typecast_and_pack`, where `llk_math_reconfig_remap(false)` was added as a precaution after untilize. I A/B tested the current RGB-to-YUV Y and UV correctness cases on BH with and without that toggle; both passed unchanged. I suggest removing that transition from RGB-to-YUV as well, after running its broader shape sweep.

pavlejosipovic (approved):

cglagovichTT (approved):

rouzbehTT:
all the tests passed:
- https://github.com/tenstorrent/tt-metal/actions/runs/33530869501
- https://github.com/tenstorrent/tt-metal/actions/runs/33530390215

Issue #51083, opened 2026-07-24
Operator-precedence bug in repeat_program_factory_last_dim.cpp


fplavecTT:
This latent bug was caught by Claude while porting ops to Metal 2.0, but I've also inspected it myself and do believe this is a real bug.

**Prevention note**: I was always taught to parenthesize anything other than a single variable in a ternary condition, which would have likely prevented this bug. We should consider following that guideline.

In `repeat_program_factory_last_dim.cpp:53-57`, `cb_size_bytes` is written as:
```
const uint32_t cb_size_bytes = READ_ALIGNMENT * 2 + (source_page_size_bytes & 0xF) == 0 ? source_page_size_bytes
                               : (source_page_size_bytes & 0x7) == 0                    ? source_page_size_bytes * 2
                               : ...
```
C++ binds `*` and `+` tighter than `==`, so the first condition parses as `((READ_ALIGNMENT * 2 + (source_page_size_bytes & 0xF)) == 0)`, i.e. `((128 + (spb & 0xF)) == 0)`, which is never true. Two consequences:
1. the intended "multiple of 16 → source_page_size_bytes" branch is unreachable, so a multiple-of-16 page falls through to the "multiple of 8" branch and gets `source_page_size_bytes * 2`; and 
2. the `READ_ALIGNMENT * 2` (128-byte) alignment headroom that the kernel comment in `repeat_last_dim_rm_interleaved.cpp:29-34` documents for each CB page is dropped from every branch. Computed sizes fall below the documented intent for most page sizes (for example spb = 16 yields 32 bytes rather than the intended 144, and spb = 64 yields 128 rather than 192). The analogous higher-dim factory computes the size correctly as `(READ_ALIGNMENT * 2) + page_size_bytes` (`repeat_program_factory_higher_dim.cpp:66`).

Link to source line:
https://github.com/tenstorrent/tt-metal/blob/cebd1d45a19f3f5e051406cee3d2abb70cf67bcf/ttnn/cpp/ttnn/operations/data_movement/repeat/device/repeat_program_factory_last_dim.cpp#L53

PR #43582, opened 2026-05-04
#39225: use ADDR_MOD_7 in welford's sfpu llk

bbradelTT:
### Summary

Closes #39225

On Blackhole, a specific model with specific weights was failing when using a welford-mode ttnn.group_norm op. The root cause was eventually found as using ADDR_MOD_3 instead of ADDR_MOD_7 in the BH Welford's sfpu llk. This PR updates that file accordingly.

Pytest is AI generated.

Tested locally that tests pass.                                  

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/blackhole-post-commit.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:bbradel-39225_welford_init)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:bbradel-39225_welford_init)

Issue #39225, opened 2026-03-05
[group_norm]: BH P150 PCC drop with Welford alg enabled


nmilicevicTT:
### Component / Area

ops

### Issue Type (optional)

Bad Outputs

### Observed

DRAM group_norm on blackhole P150 with `use_welford=True` causes PCC drop. Noticed it when running in comparison mode.

### Expected

Tests should pass PCC check. 

### 1. Steps (exact commands)

```
git fetch origin
git checkout nmilicevic/sdxl-bh-group-norm-welford
TT_METAL_CORE_GRID_OVERRIDE_TODEPRECATE="7,7" pytest models/experimental/stable_diffusion_xl_base/vae/tests/pcc/test_welford_bug_reproduction.py
```

If wondering why TT_METAL_CORE_GRID_OVERRIDE_TODEPRECATE="7,7"  is used, this is the first step for porting SDXL to blackhole from wormhole. The issue persists even without it since the core grid is configured as 8x8.

### 2. Input data / link or description

/ 

### 3. Frequency

Same test suite will produce same PCC. It looks like the PCC depends on what tests were run together. The same test that failed when all tests were run can even pass when run standalone. 

### 1. Software Versions

bh-30 p150b
fw 19.4.0.0

### 2. Hardware Details

P150

### Is this a regression?

Unknown

### Logs & Diagnostics

```
.
.
.

PASSED models/experimental/stable_diffusion_xl_base/vae/tests/pcc/test_welford_bug_reproduction.py::test_resnetblock2d_welford_disabled_EXPECT_PASS[device_params0-down_blocks-2-1]
PASSED models/experimental/stable_diffusion_xl_base/vae/tests/pcc/test_welford_bug_reproduction.py::test_resnetblock2d_welford_disabled_EXPECT_PASS[device_params0-down_blocks-3-0]
FAILED models/experimental/stable_diffusion_xl_base/vae/tests/pcc/test_welford_bug_reproduction.py::test_resnetblock2d_welford_enabled_EXPECT_FAIL[device_params0-up_blocks-1-0] - AssertionError: 0.9336986788295536
FAILED models/experimental/stable_diffusion_xl_base/vae/tests/pcc/test_welford_bug_reproduction.py::test_resnetblock2d_welford_enabled_EXPECT_FAIL[device_params0-up_blocks-3-0-conv_shortcut] - AssertionError: 0.9988206810945045
[...]

PR #54786, opened 2026-08-28
[Performance] Optimise Welford paths with stable two-pass statistics

### LINKED ISSUE #54016: [Bounty $35000] Welford Two-Pass Statistics Optimisation

jasondavies:
### PR Category

Performance

### Summary

Closes https://github.com/tenstorrent/tt-metal/issues/54016

Replace online Welford statistics with stable shifted two-pass statistics in `var`, `std`, GroupNorm and LayerNorm on Wormhole and Blackhole. FP32 accumulation and centred statistics preserve accuracy for large-offset inputs while removing the per-sample Welford recurrence.

The implementation adds allocator-aware L1 replay, compact SFPU statistics combining, compensated LayerNorm finalisation and targeted residual/affine data reuse. The `use_welford` option keeps its name; calibrated selectors retain tile/FPU reductions where appropriate. Distributed LayerNorm remains online Welford; Quasar is excluded.

The shared L1-capacity helper also excludes reserved `L1_SMALL` space and prevents unsigned underflow in existing matmul and data-movement callers.

### API Change

Remove `ttnn.group_norm(..., reciprocals=...)` and `ttnn.create_group_norm_reciprocals`. Callers should remove GroupNorm reciprocal-table creation and argument passing. LayerNorm's reciprocal API remains for paths that still consume the LUT.

### Performance

Representative operation benchmarks use synchronised trace latency, including dispatch overhead. Speed-up is baseline latency divided by PR latency.

The family results below are the **18 September snapshot**, comparing merge base `38afb36a2b6` with `b7ccb4a97ad`. They have not been refreshed after the latest fixes and merge.

| Family | Cases per architecture | Blackhole geometric-mean speed-up | Wormhole geometric-mean speed-up |
|---|---:|---:|---:|
| Generic `var`/`std` | 14 | 12.45x | 17.60x |
| GroupNorm | 4 | 3.07x | 3.50x |
| Interleaved LayerNorm | 8 | 1.10x | 1.09x |

That snapshot includes FP32 interleaved LayerNorm regressions: plain `(32, 4096)` is 26.9% slower on Blackhole and 23.3% on Wormhole; affine `(512, 256)` is 7.7% and 2.9% slower respectively. These paths also correct large-offset accuracy failures in the baseline. Wormhole residual-plus-affine `(1024, 2880)` is 12.4% slower, an outstanding performance gap without an established accuracy justification.

#### Sharded LayerNorm

The **22 September comparison** uses unmodified main `b8915544692` and PR candidate `7eabffb4107` on Blackhole. All cases are `(1024, 256)`, block-sharded over a 4x4 grid; affine means gamma and beta without residual addition.

| Workload | Main | This PR | Speed-up |
|---|---:|---:|---:|
| FP32 affine | 42.36 us | 57.70 us | 0.73x |
| FP32 plain | 41.13 us | 54.38 us | 0.76x |
| BF16 affine | 39.92 us | 21.22 us | 1.88x |
| BFP8 affine | 39.93 us | 21.30 us | 1.87x |

The FP32 slowdown fixes a pre-existing precision problem: main passes only **8/28** large-offset probes, versus **28/28** for this PR. Applying the same precision fixes to online Welford gives 60.73 us affine and 57.33 us plain, making two-pass **5.0–5.2% faster at comparable tested accuracy**. This supplementary control does not remove the user-visible slowdown against unmodified main. Updated Wormhole sharded timings remain pending.

### Validation

- **Blackhole, 22 September:** 627 existing reduction tests plus 24 new small-population cases passed; 670 LayerNorm/GroupNorm cases passed with 24 skips; three host selector tests passed.
- **Wormhole at `7eabffb4107`:** the complete sharded LayerNorm suite passed, **284 passed / 25 skipped**, including all 28 new FP32 large-offset cases. The same 28 cases pass on Blackhole.
- **After merging main at `381207dcf37`:** the host build, 274 Blackhole reduction/cache/padding tests and the row-major residual LayerNorm regression passed. Wormhole has not been rerun after this merge.
- Coverage includes large offsets, low variance, partial tiles, correction modes, batch/HW merging, output padding, scalar cache reuse and occupied-L1 replay boundaries. Counts from overlapping runs should not be added.

The benchmark tables identify their tested revisions; they are not a complete current-head refresh. Current-head CI status is reported separately by GitHub.


bbradelTT:
@jasondavies are you able to run the various CI pipelines to measure perf changes for models or does someone from Tenstorrent need to do that?

fvranicTT:
> @jasondavies are you able to run the various CI pipelines to measure perf changes for models or does someone from Tenstorrent need to do that?

We will have to do that. That's how we did it so far.

jasondavies:
I can run some representative model performance benchmarks myself (I have a QB2 and N300) but I'll need someone to run the real CI at some point.  Let me know if you want any specific numbers e.g. last I checked, this achieved 1.1x end-to-end speedup for a representative SDXL 512x512 50-step on WH N300.

fvranicTT:
Running
- Single-card device perf tests: https://github.com/tenstorrent/tt-metal/actions/runs/33393095655
- L2 nightly: https://github.com/tenstorrent/tt-metal/actions/runs/33393228444
- Sanity tests: https://github.com/tenstorrent/tt-metal/actions/runs/33393331098

FYI @jasondavies @bbradelTT

ttnn/cpp/ttnn/operations/core/program_cache_l1.hpp
```diff
@@ -0,0 +1,23 @@
...
+#include <cstdint>
+
+#include <tt-metalium/device.hpp>
+#include <tt-metalium/hal_types.hpp>
+
+namespace ttnn::operations::core {
+
+// Keep a conservative margin for non-CB L1 users. Callers should hash the
+// selected program plan, not this allocator-dependent byte count.
+inline std::uint64_t usable_program_l1_capacity(const tt::tt_metal::IDevice* device) {
```
vtsilytskyiTT, line 25: We have similar functions `get_max_l1_space` in `ttnn/cpp/ttnn/operations/data_movement/common/common.hpp` and `ttnn/cpp/ttnn/operations/matmul/device/utilities/matmul_utilities.hpp`. Feel free to reuse.

jasondavies: Addressed in bdb95751b9c21618a6ddc986e58206b7c6b42dc8.

I consolidated the allocator-aware L1 calculation into a shared core helper and updated the existing data-movement and matmul get_max_l1_space() implementations to delegate to it.

Replay selection applies its additional 5% safety margin on top of the shared raw capacity calculation, preserving the existing behaviour without duplicating the allocator logic.

jasondavies:
Resolved the Wormhole SDXL GroupNorm performance failure in 790d31f46d3 by packing the final affine result directly to the output buffer and combining each two-pass statistics state switch into a single SFPU dispatch. Latency improved from 1.457 ms to 1.346 ms, a 7.63% reduction, and passed the existing target across five repeated Wormhole runs. Blackhole improved from 826 µs to 792 µs.

I believe the other failures are all flaky tests/CI issues/timeout issues.  The one exception is that I'm expecting a TTSim failure due to https://github.com/tenstorrent/ttsim/issues/14

jasondavies:
Wormhole now uses the DEST-to-SrcA + ELWADD workaround instead of temporarily overriding SrcB to TF32. Both approaches preserve the required precision and showed similar performance, but the workaround avoids relying on tenstorrent/ttsim#14.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_stats_selector.hpp
```diff
@@ -0,0 +1,100 @@
...
+    const BlackholeStatsSelectorParams& blackhole_params) {
+    if (!requested_use_welford) {
+        return StatisticsBackend::TILE_REDUCTION;
+    }
+    if (arch == tt::ARCH::QUASAR) {
+        return StatisticsBackend::TILE_REDUCTION;
+    }
+    // RMSNorm does not execute the Welford calculation, but its existing
+    // Welford-configured route has distinct kernel and CB requirements.
+    if (arch != tt::ARCH::BLACKHOLE || rms_norm || input_is_row_major) {
```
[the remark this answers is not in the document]
jasondavies: Addressed in 1f735f4bb73.

The `fp32_dest_acc_en` requirement now precedes the architecture, RMSNorm, and row-major selector branches, and the FP32 SFPU finaliser predicate independently carries the same requirement. I also added selector coverage for Wormhole, row-major input, and RMSNorm with FP32 DEST accumulation disabled. All `LayerNormStatsSelector.*` tests pass.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_op_multi_core.cpp
```diff
@@ -164,6 +176,222 @@ std::optional<tt::DataFormat> dfb_data_format(const m2::ProgramSpec& spec, const
...
+            plan.large_tensor = true;
+            plan.width_block_tiles = std::min(plan.width_block_tiles, with_weights_max);
+        } else if (!selected_fits) {
+            plan.large_tensor = true;
+            plan.width_block_tiles = std::min(plan.width_block_tiles, without_weights_max);
+        }
+    }
+    plan.compact_fp32_finalizer = fp32_sfpu_finalizer && !residual.has_value() && !plan.large_tensor;
+    if (fp32_sfpu_finalizer && !plan.compact_fp32_finalizer) {
+        plan.large_tensor = true;
```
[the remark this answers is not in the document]
jasondavies: Addressed in a51a1285913.

The large-tensor plan is now gated for tiled input with row-major affine tensors before either the reader or compute kernel is selected. Program construction also asserts the pairing invariant, and the compute selection no longer repeats a divergent condition. A focused FP32 residual-plus-affine regression test exercises the previously mismatched configuration and passes on Blackhole.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_op_multi_core.cpp
```diff
@@ -164,6 +176,222 @@ std::optional<tt::DataFormat> dfb_data_format(const m2::ProgramSpec& spec, const
...
+        const std::uint32_t full_row_tiles = tt::round_up(width_tiles, block_size);
+        auto replay_footprint = footprint(true);
+        replay_footprint.residual_values = static_cast<std::uint64_t>(full_row_tiles) * intermediate_tile_size;
+        if (replay_footprint.fits(usable_l1)) {
+            plan.residual_value_tiles = full_row_tiles;
+            plan.fused_pre_add_replay = true;
+        }
+    }
+    const bool affine_mcast_core_set_safe =
+        requested_cores == default_core_range(device) || all_cores.bounding_box().size() == num_cores;
```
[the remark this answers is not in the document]
jasondavies: Addressed in 7635bd0bf1f.

Affine multicast is now enabled only when the participating cores completely fill their bounding box. The default-grid exception was removed, and program construction asserts the same invariant before constructing the multicast rectangle. The Blackhole residual-plus-affine replay/program-cache test passes.

tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_eltwise_binary_custom.h
```diff
@@ -188,3 +188,54 @@ inline void _llk_math_sub_bcast_cols_reuse_custom_(
...
+    LLK_ASSERT(tensor_shape.total_num_faces() == 4, "compensated column broadcast requires 32x32 tiles");
+
+    for (std::uint32_t tile = 0; tile < ct_dim; ++tile)
+    {
+        math::set_dst_write_addr<DstTileShape::Tile32x32, UnpackDestination::SrcRegs>(dst_index + tile);
+        math::clear_dst_reg_addr();
+
+        // Seed DEST with anchor - mean. The split statistic tile stores the
+        // upper and lower row halves in faces 1 and 3 respectively.
+        _compensated_move_broadcast_col_to_dest_(16, 0);
```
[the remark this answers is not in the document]
jasondavies: Addressed in bb37c1d8441.

The Blackhole compensated column-broadcast primitive now selects the preserve-zero state before seeding DEST with `MOVB2D`, then restores the default state afterwards. This prevents data-dependent zeroing of the `anchor - mean` correction. All eight FP32 large-offset finaliser variants pass on Blackhole.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_binary_bcast.h
```diff
@@ -136,10 +136,12 @@ constexpr std::uint32_t LREG_MASK  = p_sfpu::LREG6;
...
 // so that the 4 binops can be issued back-to-back without inter-op NOPs.
-constexpr std::uint32_t LREG_DATA0 = p_sfpu::LREG1;
-constexpr std::uint32_t LREG_DATA1 = p_sfpu::LREG3;
-constexpr std::uint32_t LREG_DATA2 = p_sfpu::LREG4;
-constexpr std::uint32_t LREG_DATA3 = p_sfpu::LREG5;
+constexpr std::uint32_t LREG_DATA0          = p_sfpu::LREG1;
+constexpr std::uint32_t LREG_DATA1          = p_sfpu::LREG3;
+constexpr std::uint32_t LREG_DATA2          = p_sfpu::LREG4;
+constexpr std::uint32_t LREG_DATA3          = p_sfpu::LREG5;
+constexpr std::uint32_t LREG_SCALAR_MEAN    = p_sfpu::LREG6;
```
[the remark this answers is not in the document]
jasondavies: Addressed in 16737a42b16.

The Wormhole and Blackhole LReg assignment blocks now explicitly document that scalar normalisation borrows LREG6 for the mean and therefore invalidates the persistent column-broadcast mask. The scalar helper comment now states the same lifetime rule. A later column broadcast must reinitialise its mask, matching the existing documented rule for row broadcasts.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/kernels/compute/layernorm_large_tensor_welford.cpp
```diff
@@ -451,21 +549,39 @@ void kernel_main() {
...
+            tile_width,
+            W,
+            blk>(*p_reciprocals);
 #endif
         // We should expect that either of the two would have have populated dst regs with mean and
         // variance in mean_dst and var_dst respectively.
 
         dfb_ex_obj.reserve_back(onetile);
         dfb_ex2_obj.reserve_back(onetile);
+        if constexpr (fp32_sfpu_finalizer) {
```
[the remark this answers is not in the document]
jasondavies: Addressed in c2cdefab227.

All mean/M2 alias FIFO operations, including the transpose input selection, are now controlled by `welford_state_fp32_alias`. When the aliases resolve to the primary CBs, the kernel no longer performs duplicate reserve/push/wait/pop operations on the same FIFO. The Blackhole FP32 residual-plus-affine large-offset case passes.

ttnn/cpp/ttnn/operations/experimental/ccl/dit_fused_distributed_groupnorm/device/dit_fused_distributed_groupnorm_program_factory.cpp
```diff
@@ -631,12 +638,19 @@ DitFusedDistributedGroupnormMeshWorkloadFactory::create_at(
...
-        {"reciprocal_size", 0u},  // welford native
         {"TILE_WIDTH", tile_w},
         {"cb_in0_welford", cb_in0_welford_index},
         {"welford_fp32_alias", static_cast<uint32_t>(welford_fp32_alias)},
         {"welford_unpack_fp32_active", static_cast<uint32_t>(welford_unpack_fp32_active)},
         {"enable_fp32_reconfig", static_cast<uint32_t>(enable_fp32_reconfig)},
+        {"fp32_sfpu_normalizer", static_cast<uint32_t>(fp32_sfpu_normalizer)},
+        {"cb_normalize_in_fp32", cb_in0_welford_index},
+        {"cb_ex_global_fp32", cb_ex_global_index},
+        {"cb_ex2pe_fp32", cb_ex2pe_index},
```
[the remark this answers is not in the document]
jasondavies: Addressed in a9785156e04.

The v1 distributed GroupNorm factory now fixes `fp32_sfpu_normalizer` to `false`. That implementation has no dedicated FP32 normalisation alias CBs and currently accepts only BF16 input, so a future FP32 input extension can no longer accidentally turn the primary producer/consumer CBs into self-aliasing FIFOs. The host target builds successfully.

models/demos/stable_diffusion_xl_base/vae/tt/tt_resnetblock2d.py
```diff
@@ -143,19 +117,10 @@ def forward(self, input_tensor, input_shape):
...
-                core_grid=self.groupnorm_config_1["core_grid"],
-                strategy=ttnn.ShardStrategy.HEIGHT,
-                orientation=ttnn.ShardOrientation.ROW_MAJOR,
-            )
-            reciprocals_tensor = ttnn.to_memory_config(self.reciprocals_tensor_1, sharded_mem_config)
 
         hidden_states = ttnn.to_memory_config(hidden_states, mem_cfg)
-        # NOTE: On Blackhole, using welford causes PCC drop in unit tests
-        use_welford = reciprocals_tensor is not None
+        # Request two-pass statistics for interleaved DRAM; sharded inputs retain tile reductions.
```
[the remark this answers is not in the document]
jasondavies: Addressed in 9cdedfea962.

`#39225` was subsequently fixed on main by `#43582`, and its dedicated Blackhole state-leak regression passes with this branch. I removed the corresponding VAE ResNet/up-block skips, stale comments, and Blackhole-only relaxed thresholds. Both formerly skipped ResNet cases, all four 1024-resolution up-block cases, the three restored-threshold ResNet cases, and the dedicated state-leak regression pass on Blackhole.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
```diff
@@ -277,118 +261,155 @@ void kernel_main() {
...
         }
 
-        for (uint32_t out_block_index = 0; out_block_index < num_out_blocks_padded; out_block_index++) {
-            uint32_t out_block_h_actual = out_block_h_normal;
-            if (extra_out_block && (out_block_index == (num_out_blocks_padded - 1))) {
-                out_block_h_actual = out_block_h_last;
+        // First statistics traversal: accumulate sums relative to each group's first value.
+        for (uint32_t mt = 0; mt < block_h; ++mt) {
+            if (mt > 0) {
+                two_pass_stats_switch_group(mean_dst, active_group, 0);
```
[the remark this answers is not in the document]
jasondavies: Addressed in 317245b9f01.

Both GroupNorm compute implementations now compile out the start-of-row group save/restore dispatches when `num_groups == 1`. I extended the FP32 large-offset DRAM coverage to one- and two-group configurations; all four plain/affine cases pass on Blackhole.

ttnn/cpp/ttnn/operations/core/program_cache_l1.hpp
```diff
@@ -0,0 +1,26 @@
...
+inline std::uint64_t available_program_l1_capacity(const tt::tt_metal::IDevice* device) {
+    const auto lowest_occupied_l1 = device->lowest_occupied_compute_l1_address().value_or(device->l1_size_per_core());
+    const auto cb_l1_base = device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+    return lowest_occupied_l1 > cb_l1_base ? lowest_occupied_l1 - cb_l1_base : 0;
+}
+
+// Keep a conservative margin for non-CB L1 users. Callers should hash the
+// selected program plan, not this allocator-dependent byte count.
+inline std::uint64_t usable_program_l1_capacity(const tt::tt_metal::IDevice* device) {
+    return available_program_l1_capacity(device) * 95 / 100;
```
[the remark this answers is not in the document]
jasondavies: Addressed in e66a1df5e45.

The shared program-cache L1 helper now names its 95% usable-capacity policy as `kProgramL1UsagePercent`, matching the existing GroupNorm convention and retaining the documented 5% margin. The host target builds successfully.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_op_multi_core.cpp
```diff
@@ -140,18 +137,33 @@ void bind_tensor(m2::KernelSpec& kernel, const m2::TensorParamName& tensor, std:
...
+    kernel.semaphore_bindings.push_back(m2::SemaphoreBinding{
+        .semaphore_spec_name = semaphore,
+        .accessor_name = std::move(accessor_name),
+    });
+}
+
 // Make the two members of a legacy two-index circular buffer into an alias clique. Both members
 // must name each other, which is what makes the group legal.
-void alias_pair(m2::ProgramSpec& spec, const m2::DFBSpecName& first, const m2::DFBSpecName& second) {
+void alias_group(m2::ProgramSpec& spec, const std::vector<m2::DFBSpecName>& group) {
```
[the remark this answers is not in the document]
jasondavies: Addressed in b52c0c099df.

Updated the stale two-member comment to document the actual N-member alias-group contract: each member must name every other member in the clique.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/groupnorm_mcast_program_factory.cpp
```diff
@@ -595,6 +604,10 @@ tt::tt_metal::ProgramDescriptor GroupNormDeviceOperation::GroupNormMcastProgramF
         {"padded_hw", pad.padded_hw},
         {"pad_scaler_bits", pad.scaler_bits(num_rows_per_batch_per_core_group_1 * num_channels_per_group)},
         {"has_row_mask", static_cast<uint32_t>(pad.active)},
+        {"sfpu_two_pass_reciprocal",
+         std::bit_cast<uint32_t>(
+             1.0f / static_cast<float>(
+                        std::max(1U, num_channels_per_group * num_rows_per_batch_per_core_group_1 / tile_width)))},
```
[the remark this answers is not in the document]
jasondavies: Addressed in 82a786600bc.

Removed `sfpu_two_pass_reciprocal` from the multicast and no-multicast writer named compile-time argument maps. The reciprocal remains only in the compute maps that consume it, eliminating three duplicated expressions and unnecessary writer variants. The host target builds successfully.

tt_metal/tt-llk/common/ckernel_sfpu_welfords_common.h
```diff
@@ -0,0 +1,1211 @@
...
+    TTI_SFPADD(ckernel::p_sfpu::LREG4, ckernel::p_sfpu::LCONST_1, ckernel::p_sfpu::LREG2, ckernel::p_sfpu::LREG4, 0);
+    TTI_SFPADD(ckernel::p_sfpu::LREG5, ckernel::p_sfpu::LCONST_1, ckernel::p_sfpu::LREG3, ckernel::p_sfpu::LREG5, 0);
+    TTI_SFPNOP;
+}
+
+template <std::uint32_t input_lreg, std::uint32_t accumulator_lreg>
+sfpi_inline void _two_pass_accumulate_shifted_sum_row_()
+{
+    TTI_SFPMAD(ckernel::p_sfpu::LREG11 /* -1 */, ckernel::p_sfpu::LREG4, input_lreg, input_lreg, 0);
+    TTI_SFPNOP;
```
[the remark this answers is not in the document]
jasondavies: Addressed in 860f79a28aa.

The 11 register RAW hazard sites now use `WELFORD_SFPU_ONLINE_HAZARD_NOP()`, while the separate transpose-drain NOPs remain unconditional. Blackhole assembly for a concrete FP32 GroupNorm instantiation contains 36 fewer instantiated `sfpnop` instructions, with the affected SFPU arithmetic unchanged; Wormhole still expands every hazard macro to `TTI_SFPNOP`. The corresponding numerical test passes on Blackhole.

tt_metal/hw/inc/api/compute/welford.h
```diff
@@ -70,6 +71,87 @@ ALWI void welford_reinit(uint32_t cbid, uint32_t call_line = __builtin_LINE()) {
...
+ * @brief Initializes the SFPU state used by the two-pass statistics helpers.
+ *
+ * Configures the Welford address mode used by their SFPU loads without
+ * programming Welford's unused replay buffer. The first shifted update with
+ * `initialize_anchor=true` initializes the accumulators.
+ */
+ALWI void two_pass_stats_init_shifted() { MATH((llk_math_two_pass_sfpu_init())); }
+
+template <bool accumulate_m2, bool dual_m2 = true>
+ALWI void two_pass_stats_update_rows(std::uint32_t input_dst_idx, std::uint32_t start_row, std::uint32_t num_rows) {
```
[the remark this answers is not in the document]
jasondavies: Addressed in 923632b5203.

Added Doxygen `@brief`, template-parameter, and function-parameter documentation for all 15 previously undocumented `two_pass_stats_*` APIs, including state layout and reciprocal-bit semantics.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_stats_selector.hpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+    if (!params.has_gamma && !params.has_beta && params.input_format == tt::DataFormat::Float16_b) {
+        return true;
+    }
+    if (!params.compact_two_pass_fits_in_l1) {
+        return false;
+    }
+    if (params.fuse_pre_add && has_full_affine) {
+        if (params.input_format == tt::DataFormat::Float16_b) {
+            return params.padded_width <= 256 || params.padded_width >= 2880;
```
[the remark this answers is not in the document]
jasondavies: Addressed in 81d812fbac7.

Replaced the four bare crossover literals with named constants. BF16 and BFP8 retain separate wide fused-residual constants even though both are currently 2880, so future per-format calibration cannot accidentally couple them. All `LayerNormStatsSelector.*` tests pass.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
```diff
@@ -466,52 +516,84 @@ void kernel_main() {
...
+                            reconfig_data_format_srca(dfb_ex_global_fp32_id, dfb_ex2pe_fp32_id);
+                            copy_init(dfb_ex2pe_fp32_id);
+                            copy_tile(dfb_ex2pe_fp32_id, g, inv_std_dst);
+                            sfpu_normalize_bcast_scalar(data_dst, mean_dst, inv_std_dst);
+                            tile_regs_commit();
+                            tile_regs_wait();
+                            pack_tile(data_dst, dfb_xmm_id);
+                            tile_regs_release();
+                            dfb_xmm.push_back(1);
+                            reconfig_data_format_srca(dfb_ex2pe_fp32_id, dfb_normalize_in_fp32_id);
```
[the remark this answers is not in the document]
jasondavies: Rechecked this against the updated ISA documentation and Blackhole silicon. These aliases use UnpackToDestFp32, so copy_tile selects the direct-to-DEST path from the alias metadata. For direct-to-DEST unpacking, the ISA explicitly defines Float32 -> Tf32 as preserving the full FP32 value; TF32 mantissa truncation applies only when writing through SrcA/SrcB.

I also tested the proposed unconditional reconfiguration in both GroupNorm implementations. Placing it before tile_regs_acquire() or after copy_init() corrupted the Blackhole output. The restored implementation passes all six sharded and DRAM FP32 large-offset regressions.

The conditional reconfigurations between the three (Float32, Float32) aliases are therefore intentionally no-ops. The LLK_ASSERT_BLOCK failure is a separate debug-state validation issue: it does not account for the documented equivalence of Tf32 and Float32 on the direct-to-DEST path. No GroupNorm change is required here.

jasondavies:
@Aswinmcw Can we run CI on this?  All failures should now be addressed and it's ready for further review.

jasondavies:
Issue tracking LLK rerun failure here https://github.com/tenstorrent/tt-metal/issues/55703.  Edit: fix is merged.

jasondavies:
@Aswinmcw Would be great to re-run CI now and also ping codeowners, thanks

jasondavies:
@Aswinmcw Fixed merge conflicts - ready for another CI pass.  Can we also ping codeowners and get some code review?

jasondavies:
@jberkowitzTT @cmaryanTT @Aswinmcw Can we ping codeowners and get this reviewed and merged?  Thanks.

bbradelTT (approved):

tech_reports/op_kernel_dev/accuracy_tips/accuracy_tips.md
```diff
@@ -138,10 +138,10 @@ With changes:
 
 <img src="images/error_after.png" style="width:600px;"/>
 
-# Welford's Algorithm For Mean and Variance
-As stated above, the LayerNorm and GroupNorm algorithms require computing the mean and variance per layer (LayerNorm) or per group (GroupNorm). By default, the mean and variance are computed with the two-pass method detailed above, namely compute $\mu$ in one pass then use it to compute $\sigma^2$ in the second pass.
+# Welford-Compatible Mean and Variance Paths
+LayerNorm and GroupNorm require the mean and variance per layer or group. Their `use_welford` option is retained as a compatibility name for requesting the numerically stable statistics backends; it no longer selects one algorithm unconditionally. Depending on the architecture, tensor layout and calibrated performance policy, ordinary LayerNorm and GroupNorm select either shifted two-pass SFPU statistics with FP32 accumulation or a tile/FPU reduction. Distributed LayerNorm continues to use an online Welford merge protocol.
```
fplavecTT, line 142: "or a tile/FPU reduction"?
This makes it sound like use_welford may use FPU with reduced precision (TF32)? Is that true or just awkward wording?

tech_reports/op_kernel_dev/accuracy_tips/accuracy_tips.md
```diff
@@ -182,10 +182,10 @@ program_config=ttnn.LayerNormShardedMultiCoreProgramConfig(<other configs>, use_
 GroupNorm (specified as boolean input to `ttnn.group_norm`):
 
 ```python
-output_tensor = ttnn.group_norm(<other inputs>, use_welford=use_welford)
+output_tensor = ttnn.group_norm(<other inputs>, use_welford=True)
 ```
 
-While the Welford method invoked as defined above works, it is advisable to create and pass in a tensor of reciprocals to substantially accelerate the calculation. This can be done by the following:
+GroupNorm uses compile-time reciprocal constants in its shifted two-pass path and does not require a reciprocals tensor. LayerNorm's selected SFPU statistics path uses a reciprocal lookup table; the tile-reduction path does not. The lookup table can be created as follows:
```
fplavecTT, line 188: Curious: why does one require reciprocals and the other one does not?

tests/ttnn/unit_tests/operations/fused/test_group_norm.py
```diff
@@ -173,8 +182,153 @@ def manual_group_norm(input_tensor, num_groups, eps=1e-2):
     return input_tensor
 
 
+@run_for_blackhole("low-variance stable-statistics regression is calibrated on Blackhole")
+@pytest.mark.parametrize("base, amplitude", [(1.0, 0.01), (10.0, 0.05)])
+@pytest.mark.parametrize("device_params", DEVICE_PARAMS_L1_SMALL_SIZE, indirect=True)
+def test_group_norm_stable_stats_translation_stability(device, base, amplitude):
+    torch.manual_seed(7)
```
fplavecTT, line 189: Nit: Add a brief docstring?

tests/ttnn/unit_tests/operations/fused/test_group_norm.py
```diff
@@ -173,8 +182,153 @@ def manual_group_norm(input_tensor, num_groups, eps=1e-2):
...
+    )
+    actual = ttnn.to_torch(ttnn.from_device(ttnn.to_memory_config(output, ttnn.DRAM_MEMORY_CONFIG))).float()
+
+    error = actual - reference
+    assert torch.isfinite(actual).all()
+    if constant is not None:
+        # Exact equality also detects small means flushed by premature scaling.
+        assert torch.equal(actual, reference)
+    assert error.abs().max() < 0.02
+    assert error.abs().mean() < 0.004
```
fplavecTT, line 327: We tend to use `assert_numeric_metrics` if possible, which covers multiple metrics (of course that's unnecessary for the case where `torch.equal` is asserted).

tests/ttnn/unit_tests/operations/fused/test_layer_norm.py
```diff
@@ -102,6 +158,235 @@ def test_layer_norm(device, h, w, use_welford, dtype):
...
+        program_config=ttnn.LayerNormDefaultProgramConfig(use_welford=True),
+        recip_tensor=create_recip_tensor(device, width, use_welford=True),
+        compute_kernel_config=compute_kernel_config,
+    )
+    actual = ttnn.to_torch(output).to(torch.float64)
+
+    error = actual - reference
+    assert torch.isfinite(actual).all()
+    assert error.abs().max() < 0.025
+    assert actual.mean(dim=-1).abs().max() < 0.004
```
fplavecTT, line 198: Similar in this and other test files: `assert_numeric_metrics` (or if appropriate `assert_output_accuracy`, which calls `assert_numeric_metrics`)

tests/ttnn/unit_tests/operations/fused/test_layer_norm.py
```diff
@@ -465,6 +817,272 @@ def test_l1_interleaved(device, use_welford, dtype):
...
+    config = ttnn.LayerNormDefaultProgramConfig(use_welford=True)
+
+    # No reciprocal tensor: falling back to the large streaming kernel is an error.
+    warm_output = ttnn.layer_norm(device_inputs[0], program_config=config)
+    assert_output_accuracy(references[0], ttnn.to_torch(warm_output), use_welford=True)
+    warm_output.deallocate(force=True)
+
+    # The compact footprint fits only if the unused full-row XMM buffer is absent.
+    grid = device.compute_with_storage_grid_size()
+    pressure_tiles = (650 * 1024 * grid.x * grid.y + 2047) // 2048
```
fplavecTT, line 836: Could we have some comments (or meaningful constant names, whichever one is more appropriate) for some of these less obvious magic numbers.
Same for few other places in this file.

tests/ttnn/unit_tests/operations/fused/test_layer_norm_sharded.py
```diff
@@ -162,6 +163,39 @@ def test_layer_norm_sharded_with_weight_and_bias_row_major(device, use_welford,
...
+        block_wt=block_wt,
+        provide_reciprocal=False,
+    )
+
+    assert_numeric_metrics(
+        reference,
+        output,
+        pcc_threshold=0.9999,
+        rtol=0.065,
+        atol=0.065,
```
fplavecTT, line 194: 0.065 + 6.5% seems like a very high error bar for float32.
Keep in mind that `assert_numeric_metrics` calls https://docs.pytorch.org/docs/2.14/generated/torch.allclose.html which checks for `∣ input - other ​∣ ≤ atol + rtol * ∣ other ​∣`

Frobenius also seems high.

tests/ttnn/unit_tests/operations/reduce/test_reduction.py
```diff
@@ -6,14 +6,107 @@
...
+    values = torch.randn(shape) + (torch.arange(shape[-1]) % 32).float() + 1024
+    tt_input = ttnn.from_torch(values, dtype=dtype, layout=ttnn.TILE_LAYOUT, device=device)
+    quantised_input = ttnn.to_torch(tt_input).to(torch.float64)
+    rtol = 2e-4 if dtype == ttnn.float32 else 0.01 if dtype == ttnn.bfloat16 else 0.03
+    for torch_op, ttnn_op in ((torch.var, ttnn.var), (torch.std, ttnn.std)):
+        expected = torch_op(quantised_input, dim=dim, keepdim=True, correction=int(correction))
+        for _ in range(3):
+            actual = ttnn.to_torch(ttnn_op(tt_input, dim=dim, keepdim=True, correction=correction)).to(torch.float64)
+            assert torch.isfinite(actual).all()
+            torch.testing.assert_close(actual, expected, rtol=rtol, atol=1e-7)
```
fplavecTT, line 90: Same comment as elsewhere. Use `assert_numeric_metrics` whenever possible.

tests/ttnn/unit_tests/operations/reduce/test_reduction.py
```diff
@@ -145,11 +323,144 @@ def test_var_fp32_translation_invariance(device, dim, offset, scalar):
...
+        warm_output = ttnn_op(warm_input, dim=-1, keepdim=True, correction=True)
+        ttnn.synchronize_device(device)
+        warm_output.deallocate(force=True)
+    warm_input.deallocate(force=True)
+
+    # Reuse the same operation keys after allocator state changes. Without the
+    # occupied-L1 cache discriminator, this resurrects the replay programs.
+    # Occupying 1100 KiB/core leaves room for streaming, but not 512 KiB row replay.
+    grid = device.compute_with_storage_grid_size()
+    pressure_tiles = (1100 * 1024 * grid.x * grid.y + 2047) // 2048
```
fplavecTT, line 389: I'm wondering if we could avoid hard coding 1100 (and other similar numbers) and at the same time make the test architecture-independent. I see there is `ttnn.get_memory_view(device, ttnn.BufferType.L1).largest_contiguous_bytes_free_per_bank`. Maybe we could express these values as fractions of that?

ttnn/cpp/ttnn/operations/core/program_cache_l1.hpp
```diff
@@ -0,0 +1,43 @@
+// SPDX-FileCopyrightText: © 2026 Tenstorrent AI ULC
```
fplavecTT, line 1: Tenstorrent USA, Inc.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
```diff
@@ -277,118 +261,165 @@ void kernel_main() {
...
-        welford_init();
-
-        uint32_t block_xy_coord = 0;
+        two_pass_stats_init_shifted();
+        two_pass_stats_clear();
+        uint32_t active_group = 0;
 
-        for (uint32_t g = 0; g < num_groups; ++g) {
+        for (uint32_t g = 1; g < num_groups; ++g) {
             welford_save_state(mean_dst, g);
```
fplavecTT, line 269: Is `welford_save_state` still needed now that two-pass approach is used?
I assume it is, but a brief comment why would be useful here.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
```diff
@@ -466,52 +516,84 @@ void kernel_main() {
...
+                            tile_regs_release();
+                            dfb_xmm.push_back(1);
+                            reconfig_data_format_srca(dfb_ex2pe_fp32_id, dfb_normalize_in_fp32_id);
+                            copy_init(dfb_normalize_in_fp32_id);
                         } else {
-                            reconfig_data_format_srcb(dfb_eps_id, dfb_ex_global_id);
+                            // // a. x-u
+                            // fp32: SrcA needs dfb_in0 (fp32 input), SrcB needs dfb_ex_global (fp32 mean); the prior
+                            // group's mul_tiles(dfb_xmm) left SrcA on dfb_xmm. Use the unconditional 1-arg form: the
+                            // old 2-arg srcb(dfb_eps -> dfb_ex_global) never reset SrcA at all.
```
fplavecTT, line 550: Nit: narrating past state of code is usually not useful to future readers.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
```diff
@@ -466,52 +516,84 @@ void kernel_main() {
...
+                            if constexpr (enable_fp32_reconfig) {
+                                reconfig_data_format_srca(dfb_in0_id);
+                                reconfig_data_format_srcb(dfb_ex_global_id);
+                            } else {
+                                reconfig_data_format_srcb(dfb_eps_id, dfb_ex_global_id);
+                            }
+                            sub_bcast_scalar_init(dfb_in0_id, dfb_ex_global_id);
+
+                            tile_regs_acquire();
+                            sub_tiles_bcast_scalar(dfb_in0_id, dfb_ex_global_id, 0, 0 + (g << 1), dst0);
```
fplavecTT, line 560: "0 + (g << 1)" looks odd. Could use a comment or a meaningfull named const var.
Also, is the intention to actually do bit manipulation with `<< 1` rather than `* 2`? If the intention is arithmetic, `* 2` is preferred because the intent is clearer and compiler will pick the best optimized implementation anyways.

Whatever the decision, please apply it to other locations/files where this pattern occurs

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/groupnorm_device_operation.cpp
```diff
@@ -44,14 +48,158 @@ GroupNormDeviceOperation::program_factory_t GroupNormDeviceOperation::select_pro
...
+    if (height_tiles < num_virtual_rows || height_tiles % num_virtual_rows != 0) {
+        return {};
+    }
+    std::uint32_t per_core_height_tiles = height_tiles / num_virtual_rows;
+    std::uint32_t per_core_height = per_core_height_tiles * tile_height;
+    const std::uint32_t per_core_width = width / num_virtual_cols;
+    const std::uint32_t per_core_width_tiles = (per_core_width + tile_width - 1) / tile_width;
+    const std::uint32_t channels_per_group = width / num_groups;
+    const std::uint32_t num_row_shards = height / per_core_height;
+    const std::uint32_t batches_group_1 = num_batches > num_row_shards ? num_batches / num_row_shards : 1;
```
fplavecTT, line 91: May be prudent to assert that `num_row_shards` and `num_col_shards` are not 0 before dividing by them.
(My guess is they can't be 0 by construction right now, but a future refactoring could inadvertently change that).

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
```diff
@@ -16,6 +17,101 @@
...
+    plan.is_std = operation_attributes.math_op == ReduceOpMath::STD;
+    const float post_mul_scaler =
+        plan.is_std ? std::abs(operation_attributes.scalar) : operation_attributes.scalar * operation_attributes.scalar;
+    plan.use_post_mul = post_mul_scaler != 1.0f;
+    plan.post_mul_scaler_bits = std::bit_cast<std::uint32_t>(post_mul_scaler);
+    plan.narrow_scratch_to_bf16 = !plan.is_std && !plan.use_post_mul && plan.output_format == DataFormat::Float16_b;
+    plan.combined_format = plan.narrow_scratch_to_bf16 ? DataFormat::Float16_b : DataFormat::Float32;
+    const auto arch = tensor_arg.device()->arch();
+    plan.use_sfpu_leaf_combine = plan.reduce_hw && (arch == tt::ARCH::BLACKHOLE || arch == tt::ARCH::WORMHOLE_B0) &&
+                                 plan.fp32_dest_acc_en && plan.W % plan.tile_width == 0 &&
```
fplavecTT, line 63: Why the restriction on `plan.W % plan.tile_width`? Can't we do masking in the kernel/LLK to ignore columns in the partial tile, or address it some other way?
The way it is now, my quick experiment showed about 17X difference in performance between 
shape `(1, 1, 32, 10528)` and `(1, 1, 32, 10529)`.
At minimum this should be documented, but ideally we would have a way to mask unused columns.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
```diff
@@ -16,6 +17,101 @@
...
+                                 static_cast<std::uint64_t>(plan.W) * plan.reduce_batch_size >= 128;
```
fplavecTT, line 64: Similar here. Where does 128 come from?
I asked Claude to run some experiments and changing this 128 to 1 helped performance in some cases and never hurt it. Of course, my testing wasn't exhaustive.
If this stays, rationale should be documented.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
```diff
@@ -506,7 +507,10 @@ WelfordReduceDeviceOperation::WelfordReduceProgramFactory::create_program_artifa
     std::visit(
         [&](auto& compute_cfg) {
             compute_cfg.sfpu_precision_mode = Precision::Precise;  // legacy math_approx_mode = false
-            compute_cfg.double_buffer_dest = true;                 // legacy dst_full_sync_en = false
+            // Compact HW statistics need single-buffered DST: overlapping these
+            // sparse writes with PACK intermittently loses leaf statistics on
```
fplavecTT, line 511: Intermittently?
Is this a known limitation or something that should be investigated (possibly outside of this PR)?

ttnn/cpp/ttnn/operations/normalization/layernorm/device/kernels/compute/layernorm_welford.cpp
```diff
@@ -128,136 +164,145 @@ void kernel_main() {
...
+            if constexpr (welford_fp32_alias) {
+                dfb_x_welford_obj.wait_front(static_cast<uint16_t>(wt + 1));
+            } else {
+                dfb_x_obj.wait_front(static_cast<uint16_t>(wt + 1));
+            }
+            const uint32_t stats_input_dst =
+                wt < num_front_retained_limit ? (wt == 0 ? retained_input_dst : wt) : input_dst;
+            transpose_tile(dfb_x_welford, wt, stats_input_dst);
+            if (wt == 0) {
+                two_pass_stats_update_shifted_rows<false /* accumulate_m2 */, true /* initialize_anchor */>(
```
fplavecTT, line 255: If I understand correctly, anchor is just the first element along W of each row? What if that element is not representative of the rest of the row? Do any tests cover this case (for Layernorm, or ttnn.std/var)? How does the error compare to Welford?
Mean is likely to be more affected than variance. Depending on what the tests show, there may be some cheap way to compensate (if needed at all).

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/compute/welford_reduce_h.cpp
```diff
@@ -38,115 +40,111 @@ void kernel_main() {
...
-                // Apply the user scalar to the reduced output: var(s*x)=s^2 var(x),
-                // std(s*x)=|s| std(x). mul_unary_tile is an SFPU op on DEST at full fp32
-                // precision .
-                binop_with_scalar_tile_init();
-                mul_unary_tile(var_dst, post_mul_scaler_bits);
 #endif
-                tile_regs_commit();
+            if (ht == 0) {
+                two_pass_stats_update_shifted_rows<false, true>(
+                    stats_input_dst, 0, ht == Ht - 1 ? last_tile_rows : tile_height);
```
fplavecTT: Not a blocker, but I would recommend parenthesizing non-trivial conditions in ternary op to avoid things like #51083
Current condition is less complex than that, but these things tend to grow :)

fplavecTT:
@dgomezTT Can you take a look at this PR to see if this will have any impact on Metal 2.0 porting? I didn't notice anything myself, but another pair of eyes wouldn't hurt.

tech_reports/op_kernel_dev/accuracy_tips/accuracy_tips.md
jasondavies: Added a clarification about the FP32 issue.  In short: yes, `use_welford=True` will route via TF32 SrcA/SrcB.  The shifted backed preserves full FP32.

tech_reports/op_kernel_dev/accuracy_tips/accuracy_tips.md
jasondavies: It's complicated: essentially I would like to remove the reciprocals tensor everywhere entirely (perhaps in a follow-up issue) but this requires additional work.  For example, distributed LayerNorm still uses online Welford updates.  Blockwise LayerNorm looks up reciprocals for individual blocks.

tests/ttnn/unit_tests/operations/fused/test_group_norm.py
jasondavies: Done.

tests/ttnn/unit_tests/operations/fused/test_layer_norm.py
jasondavies: Done.

tests/ttnn/unit_tests/operations/fused/test_layer_norm.py
jasondavies: Done.

tests/ttnn/unit_tests/operations/fused/test_layer_norm_sharded.py
jasondavies: Tightened this case to `rtol=0`, `atol=0.015`, relative Frobenius `0.005` and PCC `0.99999`. It passes on Blackhole. In particular, the absolute-error allowance no longer grows with the expected output magnitude.

tests/ttnn/unit_tests/operations/reduce/test_reduction.py
jasondavies: Replaced 1100 KiB with an allocation derived from `get_memory_view(...).largest_contiguous_bytes_free_per_bank` and `num_banks`. It leaves half the replay row's storage free, verifies that replay cannot fit, and checks for new cached programs after the allocation. Removed this test's Blackhole-only skip.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
jasondavies: The save/restore primitive is still needed: two-pass reuses its register layout for each group's anchor and partial sums while traversing tiles shared by several groups. Added that explanation to both compute kernels; it does not invoke online Welford updates.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/kernels/compute/welford_groupnorm.cpp
jasondavies: Named the two-tile group stride and replaced all corresponding shifts/add-zero expressions in both GroupNorm compute kernels with arithmetic indexing. The comment identifies the mean followed by variance layout.

ttnn/cpp/ttnn/operations/normalization/groupnorm/device/groupnorm_device_operation.cpp
jasondavies: Geometry derivation is now shared with the factories; I added an explicit nonzero row/column-shard guard there, before either is used as a divisor. Invalid geometry returns `valid=false`, matching the selector's existing contract.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
jasondavies: Documented.  Maybe better addressed as follow-on work.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
jasondavies: Removed the gate on Blackhole as the compact case seems faster in every case.  Need to measure on Wormhole.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
jasondavies: OK, I implemented a scalar-based fallback for the tail.  So it won't be a drastic performance cliff.  Might be possible to optimise further using SFPU later.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
jasondavies: Looks like the root cause was due to sparse writes to uninitialised Dst.  This contract is not currently enforced in ttsim so I've opened an issue: https://github.com/tenstorrent/ttsim/issues/19 in case this helps anyone debugging this in future.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/kernels/compute/layernorm_welford.cpp
jasondavies: Yes: each independent accumulation stream uses its first valid sample, which need not be representative. The shift is algebraically valid for any anchor, but an outlier can leave large shifted sums and increase mean rounding error. In exact arithmetic, an error `e` in the estimated mean adds `e^2` to the second-pass variance; normalisation is also affected directly by subtracting that mean. Neither this formulation nor online Welford is uniformly more accurate for every input.

Added a LayerNorm regression with one outlier and identical remaining values, moving the outlier between the first and last columns at widths 128, 4096 and 16384. It checks FP64-reference agreement and explicitly bounds the output row mean, which PCC alone would miss. All six cases passed on Blackhole with matched explicit epsilon. Also extended the existing low-variance/outlier reduction regression from HW to W and H; all 12 dtype/correction/direction cases passed within the 140-case reduction run. The accuracy guide now describes a first-sample shift rather than calling it representative.

jasondavies:
@fplavecTT Thanks for the review!  Everything should now be addressed, but I still need to validate on WH (and update/remove the use_sfpu_leaf_combine gate).

jasondavies:
Ping @fvranicTT @Aswinmcw for CI re-run.  Thanks!

ipotkonjak-tt (commented):
Can you post passing pipelines? Since SDXL model is affected please make sure to include [(Tier 1) Models Device Perf Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t1-device-perf-tests.yaml) and [(Tier 2) Models Unit Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t2-unit-tests.yaml)

jasondavies:
@ipotkonjak-tt

> Can you post passing pipelines? Since SDXL model is affected please make sure to include [(Tier 1) Models Device Perf Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t1-device-perf-tests.yaml) and [(Tier 2) Models Unit Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t2-unit-tests.yaml)

Edit: sorry, got confused with my other PR (drop legacy rsqrt) which affects SDXL!

jasondavies:
> Can you post passing pipelines? Since SDXL model is affected please make sure to include [(Tier 1) Models Device Perf Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t1-device-perf-tests.yaml) and [(Tier 2) Models Unit Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t2-unit-tests.yaml)

@fvranicTT @Aswinmcw Please run these too.  I think the current set of failing CI tests are unrelated to this PR BTW.

Aswinmcw:
Triggered both on `fork/jasondavies/two-pass` (mirror of this PR's head `c86e4519f99` in the main repo, so `workflow_dispatch` can target it):

| Pipeline | Inputs | Run |
|---|---|---|
| [(Tier 1) Models Device Perf Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t1-device-perf-tests.yaml) | `model: all` | https://github.com/tenstorrent/tt-metal/actions/runs/35080598815 |
| [(Tier 2) Models Unit Tests](https://github.com/tenstorrent/tt-metal/actions/workflows/models-t2-unit-tests.yaml) | `model: all` | https://github.com/tenstorrent/tt-metal/actions/runs/35080602051 |

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/compute/welford_reduce_hw.cpp
```diff
@@ -36,120 +35,127 @@
...
+                    const std::uint32_t stats_input_dst =
+                        ht < num_front_retained_limit ? (ht == 0 ? retained_input_dst : mean_dst) : input_dst;
+                    copy_tile(dfb::in, 0, stats_input_dst);
                     dfb_in.pop_front(onetile);
-
-                    if (ht < (Ht - 1)) {
-                        welford_update<0>(input_dst, start_N, {});
+#endif
+                    if (ht == 0) {
+                        two_pass_stats_update_shifted_rows<false /* accumulate_m2 */, true /* initialize_anchor */>(
```
fplavecTT: Templated function calls like this are easier to read if the parameters use enums rather than Booleans.
E.g. `two_pass_stats_update_shifted_rows<AccumM2, InitAnchor>(...)`

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/compute/welford_reduce_w.cpp
```diff
@@ -37,124 +38,97 @@ void kernel_main() {
...
-            if constexpr (welford_fp32_input) {
-                welford_init<WelfordInitMode::PreserveStats>();
-            }
-
-            if (wt < (Wt - 1)) {
-                welford_update<0>(input_dst, start_N, {});
+#endif
+            if (wt == 0) {
+                two_pass_stats_update_shifted_rows<false, true>(
+                    stats_input_dst, 0, wt == Wt - 1 ? last_tile_rows : tile_width);
```
fplavecTT: Similar comment in this file about parenthesizing ternary condition.

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/reader_unary_reduce_universal_twopass_start_id.cpp
```diff
@@ -0,0 +1,81 @@
...
+                    tile_bytes,
+                    {.page_id = row_start_id + wt_base + wt},
+                    {.offset_bytes = wt * tile_bytes});
+            }
+            noc.async_read_barrier();
+            dfb_in0.push_back(read_batch);
+        }
+#else
+        constexpr std::uint32_t num_front_retained = 2;
+        constexpr std::uint32_t num_passes = Wt <= num_front_retained + 1 ? 1 : 2;
```
fplavecTT: Parenthesize condition here as well.

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/dataflow/writer_welford_hw.cpp
```diff
@@ -153,6 +151,24 @@ void kernel_main() {
...
+            auto* combined_ptr = reinterpret_cast<volatile std::uint16_t*>(dfb_combined.get_write_ptr());
+            for (std::uint32_t i = 0; i < combined_tile_size_bytes / sizeof(std::uint16_t); ++i) {
+                combined_ptr[i] = 0;
+            }
+        } else {
+            auto* combined_ptr = reinterpret_cast<volatile float*>(dfb_combined.get_write_ptr());
+            for (std::uint32_t i = 0; i < combined_tile_size_bytes / sizeof(float); ++i) {
+                combined_ptr[i] = 0.0f;
+            }
+        }
```
fplavecTT: I think this whole block could be replaced with `noc.async_write_zeros()` followed by `noc.write_zeros_l1_barrier()`

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_stats_selector.hpp
```diff
@@ -0,0 +1,113 @@
...
+constexpr StatisticsBackend select_sharded_statistics_backend(
+    bool requested_use_welford, tt::ARCH arch, bool fp32_dest_acc_en) {
+    if (!requested_use_welford || !fp32_dest_acc_en) {
+        return StatisticsBackend::TILE_REDUCTION;
+    }
+    if (arch == tt::ARCH::QUASAR) {
+        return StatisticsBackend::TILE_REDUCTION;
+    }
+    // Blackhole's centred tile reducer is 1.86-2.23x faster than the former online-Welford route across the
+    // representative sharded matrix and showed no numerical disadvantage. Wormhole retains the two-pass backend.
```
fplavecTT: "Showed no numerical disadvantage" is a red flag. That means that Welford was also losing precision somewhere through the TF32 path, which it shouldn't do. And if this persists in the new implementation, then the new implementation has the same bug. Unfortunately, none of the existing tests cover this case. I asked Claude to create the same test with large offset that already exists for interleaved path, but apply it to the sharded path, and it confirmed that there is a bug (both with the old and the new implementation).
The Claude-generated test can be found in https://github.com/tenstorrent/tt-metal/tree/fplavec/sharded-layernorm-large-offset-test 

Rather than bypassing the two-pass SFPU implementation, its sharded implementation should be fixed and `fp32_dest_acc_en` should be honored.

ttnn/cpp/ttnn/operations/reduction/generic/device/kernels/compute/welford_reduce_h.cpp
```diff
@@ -38,115 +40,111 @@ void kernel_main() {
...
-        //   rows so only valid elements participate in the statistics.
-        // - welford_finalize_to_row(mean_dst, scale_idx, ...) converts M2 into variance and
-        //   writes final mean/variance tiles into DST.
-        // - If is_std, sqrt_tile() turns variance into standard deviation in place.
-        // - start_N advances by one tile height each iteration so Welford sees the correct
-        //   element count / divisor progression across the whole H reduction.
         copy_init(dfb::in);
         tile_regs_acquire();
+        if constexpr (materialize_padding) {
+            // Tile-wide SFPU operations read physical DST, not its lazy zero flags.
```
fplavecTT: Nit: clarify this comment. Not sure what this is referring to.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
```diff
@@ -16,6 +17,109 @@
...
+    constexpr std::uint32_t hw_multi_column_replay_min_tiles = 16;
+    const bool hw_multi_column = (arch == tt::ARCH::WORMHOLE_B0 || arch == tt::ARCH::BLACKHOLE) && plan.reduce_hw &&
+                                 (plan.reduce_batch_size > 1 || plan.Wt > 1);
+    const std::uint32_t replay_min_tiles =
+        plan.work_group_1 > 1 || plan.work_group_2 > 1
+            ? multi_output_replay_min_tiles
+            : (hw_multi_column ? hw_multi_column_replay_min_tiles : single_output_replay_min_tiles);
+    std::uint64_t footprint =
+        static_cast<std::uint64_t>(replay_tiles) * plan.input_tile_size + 2 * plan.output_tile_size;
+    footprint += plan.reduce_hw ? 4 * tile_size(DataFormat::Float32) + tile_size(plan.combined_format) : 0;
```
fplavecTT: A brief comment where `4` (and `2` on the previous line) come from, or a meaningful constant name.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
```diff
@@ -291,16 +312,31 @@ WelfordReduceDeviceOperation::WelfordReduceProgramFactory::create_program_artifa
...
+        reduce_defines["WELFORD_SFPU_LEAF_COMBINE"] = "1";
+    }
+
+    const std::uint32_t two_pass_reduce_size = reduce_w ? W : H;
+    const std::uint32_t two_pass_variance_divisor =
+        reduce_hw ? H : (operation_attributes.correction ? two_pass_reduce_size - 1 : two_pass_reduce_size);
+    const std::uint32_t two_pass_mean_reciprocal =
+        std::bit_cast<std::uint32_t>(1.0f / static_cast<float>(two_pass_reduce_size));
+    const std::uint32_t two_pass_variance_reciprocal =
+        std::bit_cast<std::uint32_t>(1.0f / static_cast<float>(two_pass_variance_divisor));
```
fplavecTT, line 337: `two_pass_variance_divisor` could be `two_pass_reduce_size - 1`, so at least in theory it could be 0.
I'm sure there is some far away validation somewhere that prevents that, but it may be prudent to assert here before division in case that validation goes away in the future.

ttnn/cpp/ttnn/operations/reduction/generic/device/welford_reduce_program_factory.cpp
jasondavies: Removed the gate on Wormhole too.

ipotkonjak-tt (commented):
I see there are both perf and pcc regression on the SDXL modules. @jasondavies to comment; are these expected? if not do we know the root cause?

@jmitrovicTT if pcc drop is expected we should validate the change on our model accuracy tests

jasondavies:
@ipotkonjak-tt So it looks like we have three Wormhole VAE PCC failures, but interestingly switching GroupNorm's legacy rsqrt to non-legacy makes all three pass their unchanged thresholds.  This supports the idea of merging https://github.com/tenstorrent/tt-metal/pull/56292 (remove legacy rsqrt) before this one.

Will re-run performance checks.  The only regression I'm aware of is Blackhole VAE encode, 1024×1024: 143.374 ms versus 141.175 ms expected: 1.56% slower.

jasondavies:
@ipotkonjak-tt Blackhole VAE encode perf regression is resolved now.

I do think it's best to merge the legacy removal PR https://github.com/tenstorrent/tt-metal/pull/56292 before this one so that everything is using the non-legacy/high-precision rsqrt.

ttnn/cpp/ttnn/operations/normalization/layernorm/device/layernorm_stats_selector.hpp
jasondavies: Fixed the sharded FP32 path to preserve precision through residual addition, statistics merging and normalisation. All 28 new large-offset cases pass on Blackhole and Wormhole, and both complete sharded suites pass (285/284 passes respectively, with 24/25 skips). Unmodified main passed only 8/28 in the Blackhole comparison.

The two measured Blackhole FP32 shapes cost 32–36% latency versus main, but two-pass is 5.0–5.2% faster than online Welford with the same precision fixes.

fplavecTT: > Fixed the sharded FP32 path to preserve precision through residual addition, statistics merging and normalisation. All 28 new large-offset cases pass on Blackhole and Wormhole, and both complete sharded suites pass (285/284 passes respectively, with 24/25 skips). Unmodified main passed only 8/28 in the Blackhole comparison.
> 
> The two measured Blackhole FP32 shapes cost 32–36% latency versus main, but two-pass is 5.0–5.2% faster than online Welford with the same precision fixes.

Perfect. Thank you.

Can you help me understand this bit in the same function:
```
    // Retain Blackhole's calibrated tile path for lower-precision input. FP32
    // requires two-pass statistics and SFPU subtraction: TF32 intake loses
    // small variations around a large shared offset, even with FP32 DEST.
    if (arch == tt::ARCH::BLACKHOLE && input_format != tt::DataFormat::Float32) {
        return StatisticsBackend::TILE_REDUCTION;
    }
```

I understand that bypassing TF32 may not be beneficial (or benefits may be limited only to computations in the loop) when input is not FP32. But why would that apply only to Blackhole?
Why wouldn't this condition be architecture-agnostic?

jasondavies: Yes that does seem strange.  Let me run the comparison on WH and get back to you.

jasondavies: Validated on Wormhole and removed the Blackhole-only restriction in 429b46836b2. Tile reductions were faster across all 48 BF16/BFP8 cases tested, with geometric-mean speed-ups of 1.43× and 1.45× respectively. FP32 retains the SFPU path for accuracy. The full sharded LayerNorm suite passes on both architectures.

fplavecTT (approved):

tests/ttnn/unit_tests/operations/fused/test_layer_norm_sharded.py
```diff
@@ -162,6 +163,42 @@ def test_layer_norm_sharded_with_weight_and_bias_row_major(device, use_welford,
     )
 
 
+@run_for_blackhole("Blackhole retains tile reductions for lower-precision sharded LayerNorm")
+def test_layer_norm_sharded_tile_backend_does_not_require_reciprocal(device):
```
fplavecTT, line 167: One remaining nit: This behaviour (no reciprocals needed) shouldn't be architecture specific either, so it feels odd that it runs only for BH. Should the guard be removed?

mbezuljTT:
please show green CI runs for SDXL; especially on perf side.

jasondavies:
> please show green CI runs for SDXL; especially on perf side.

Someone else needs to run these for me.  Locally, perf doesn't regress; DRAM GroupNorm fails only because it is 6.5% faster than the expected range.

As I mentioned earlier my preference is to merge the legacy removal PR https://github.com/tenstorrent/tt-metal/pull/56292 before this one, I believe this will resolve three Wormhole SDXL VAE accuracy checks: the 512×512 ResNet block, encoder, and full encode.

Issue #54798, opened 2026-08-29
MatmulMultiCoreReuse silently drops output blocks when num_output_blocks > num_cores


kgrujcic-tt:
### Summary

`MatmulMultiCoreReuseProgramConfig` writes exactly one output block per core. When the program config asks for more output blocks than the compute grid has cores, the surplus blocks are never computed and the tail of the output tensor is left at whatever the buffer already held. There is no error and no warning: the op returns a tensor that is partly correct and partly uninitialized.

Block count is `batch * (M_t / per_core_M) * (N_t / per_core_N)`.

### Repro

Blackhole, 8 devices, grid 11x10 = 110 cores. Per chip: batch 32, `[20, 16] x [16, 4]` tiles (this is the MLA `wkv_b2` shape at the 640-token production chunk), bf16 in / bf8 weights, interleaved L1 activation, interleaved DRAM output.

```python
pc = ttnn.MatmulMultiCoreReuseProgramConfig(
    compute_with_storage_grid_size=(11, 10),
    in0_block_w=2, out_subblock_h=1, out_subblock_w=4,
    per_core_M=5, per_core_N=4,
)   # blocks = 32 * (20/5) * (4/4) = 128 > 110
out = ttnn.linear(act, weight, program_config=pc, memory_config=ttnn.DRAM_MEMORY_CONFIG, dtype=ttnn.bfloat8_b)
```

PCC vs `torch.matmul` is `-0.000009`, reproducible on 3 consecutive runs. Dropping to `per_core_M=10` (64 blocks, under the core count) gives PCC `0.999917` on the same shape, also reproducible.

### Why it is easy to miss

The failure only shows when the destination buffer does not already contain the right answer. Running any other matmul over the same output buffer first leaves the correct values there, so the partial write reads back as a pass. That is what made this look non-deterministic at first: in a sweep, the same over-subscribed config returned PCC ~0 when it ran first and PCC 0.9999 when it ran after a config that had filled the buffer.

### Suggested fix

`validate_matmul_reuse_sharded_output_block_divisibility` in `ttnn/cpp/ttnn/operations/matmul/device/matmul_device_operation.cpp` already computes `num_output_blocks` and compares it against a core count, but it returns early unless one of the inputs is L1-sharded:

```cpp
if (sharded == nullptr) {
    return;
}
```

[...]

PR #54547, opened 2026-08-27
Refactor prefill module/model tests for chunk size 5k

mbezuljTT:
### PR Category
Test Only

### Summary
All tests under `models/demos/deepseek_v3_d_p/tests/` now run the production prefill chunk: 5120
global, **640 tokens per chip**, replacing 3200 / 4096 / 1600 / 6400 per chip and a 1024 smoke tier.
Per-chip is the invariant — a smaller mesh shortens the global length rather than loading each chip
more. `utils/chunk_config.py` holds the numbers once; 5120 was re-declared in thirteen places.

Perf baselines were re-measured, not rescaled: ten gates moved 1.8x–4.4x. Validated on hardware —
117 tests on a 32-chip galaxy, 118 on a LoudBox.

### Notes for reviewers
- Rebased onto #53530, which already retired most CI selectors this PR was updating. `test_kimi_mla`
  and the `test_ttnn_moe` rows it deleted stay deleted.
- `perf-random-5k` restored — it is 640 tokens/chip, and #53530's determinism job selects it.
- Constants named per review: `PREFILL_SP_FACTOR` / `PREFILL_CHUNK_TOKENS` /
  `PREFILL_CHUNK_TOKENS_PER_CHIP`; the rename reaches two files not otherwise touched.
- `perf-host-64` kept — sole selector for the 8x1 perf proxy. Galaxy perf gates on `is_high_power`.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mbezulj/refactor_prefill_tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mbezulj/refactor_prefill_tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mbezulj/refactor_prefill_tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mbezulj/refactor_prefill_tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mbezulj/refactor_prefill_tests)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mbezulj/refactor_prefill_tests)
[...]

PR #54801, opened 2026-08-29
[Test Only] MLA matmul unit tests: retune at the 640-token production shape

### LINKED ISSUE #54761: MLA matmul unit tests: add tuned 640-token configs, drop the one-shot 100K/128K shapes
`test_mla_matmuls.py::test_mla_mm` has 12 parametrize rows and none is a production shape:

- 6 at `SEQ_LEN_25K = 6400` — *"dividing 100 * 1024 over 4 galaxies"*
- 6 at `SEQ_LEN_32K = 8192` — *"dividing 128 * 1024 over 4 galaxies"*

Both model one-shot 100K/128K prefill. Production is chunked at 5120 tokens / 640 per chip, so these
shapes and their 12 hand-tuned `prog_config_*` blocks are obsolete.

Wanted, in one change:

1. **Add real 640-row program configs.** Existing ones are sized for 6400 rows (200 tiles); at 20
   tiles their `per_core_M` / `out_subblock_h` are invalid. Each matmul (mm0–mm5) needs a config
   tuned at the production shape, plus rows to exercise it.
2. **Delete the obsolete rows and configs**, and `SEQ_LEN_25K` / `SEQ_LEN_32K`.
3. **Remove the stopgap skip** #54547 added for the 6 `SEQ_LEN_25K` rows — it leaves `SEQ_LEN_32K`
   running, equally non-production, so the file gates nothing at 640 today.

Four `blackhole_e2e_tests.yaml` jobs run `op_unit_tests/` as a directory, so collected counts change.

kgrujcic-tt:
### PR Category
Test Only

### Summary
Closes #54761.

`test_mla_mm` had 12 parametrize rows, none of them a production shape: 6 at `SEQ_LEN_25K = 6400`
and 6 at `SEQ_LEN_32K = 8192`, both modelling one-shot 100K/128K prefill. Production is chunked at
5120 tokens / 640 per chip, and #54547 had already skipped the `SEQ_LEN_25K` half, so the file gated
nothing at the shape the model actually runs.

All 12 rows are replaced with 6 at `SEQ_LEN = PREFILL_CHUNK_TOKENS` (5120), which SP-shards over the
8 mesh rows to 640 rows == 20 M tiles per chip. The obsolete rows, their 12 hand-tuned
`prog_config_*` blocks, `SEQ_LEN_25K` / `SEQ_LEN_32K` and the #54547 skip are gone.

### Notes for reviewers
- K and N are sequence-independent and carry over unchanged; only the M split is retuned.
  `per_core_M = 2` covers 20 tiles over the 10 grid rows for the 2D mcast matmuls.
- `q_a_proj` and `kv_a_proj_with_mqa` are head-count independent, so their per-chip shape is
  identical to `MLA_MATMUL_CONFIG[...][640]` and they now mirror those entries exactly. The other
  four are wider at 128 heads than any shipped 640 entry and are tuned here.
- **The two batched matmuls change config type.** `MLA_MATMUL_CONFIG` picks
  `MatmulMultiCoreReuse` for `wkv_b1` / `wkv_b2` at 640 (`MultiCast1D` only appears in the
  4096/3200 slots), so the old rows were gating a factory production never selects at this length.
  Switching also measured **3.8x / 2.7x faster** (247 -> 65 us and 197 -> 72 us per chip).
- That factory writes one output block per core, and asking for more blocks than the 110-core grid
  holds silently leaves the tail of the output unwritten -- filed separately as #54798. `per_core_M
  = 10` holds both at 64 blocks. A module-level assert now pins `PREFILL_CHUNK_TOKENS_PER_CHIP ==
  640` so a chunk-size change fails at collection with a readable message instead of inside the
  matmul validator.
- Four `blackhole_e2e_tests.yaml` jobs run `op_unit_tests/` as a directory, so their collected count
  drops by 6. None pins a count, and all run on meshes where `test_mla_mm` is skipped anyway.

### Validation
The 8x4 anchor needs a galaxy, so this was validated on a Blackhole LoudBox at 2x4 with SP=2 and a
1280-token global length -- which reproduces the identical per-chip shapes (640 rows, TP=4) that the
committed 8x4 rows produce. 6/6 pass, three consecutive runs, bit-identical PCC:

| row | matmul | per-chip [M,K]x[K,N] | PCC |
| --- | --- | --- | --- |
| mm0 | `q_a_proj` | [20,56]x[56,48] | 0.999901 |
| mm1 | `q_b_proj` | [20,48]x[48,192] | 0.999903 |
| mm2 | `wkv_b1` | b32 [20,4]x[4,16] | 0.999954 |
| mm3 | `kv_a_proj_with_mqa` | [20,56]x[56,18] | 0.999850 |
| mm4 | `wkv_b2` | b32 [20,16]x[16,4] | 0.999917 |
| mm5 | `o_proj` | [20,128]x[128,224] | 0.999898 |

Program-config choices were swept on the same proxy (min of 10 interleaved rounds x 20 iterations).

### CI runs
Triggered for this branch:

- (Blackhole) e2e, `model: disaggregated_prefill` -- the four jobs that run `op_unit_tests/` as a directory and therefore see the collected-count change: https://github.com/tenstorrent/tt-metal/actions/runs/33238397320

`test_mla_mm` itself pins an (8,4) mesh, so it is skipped on every Blackhole e2e SKU; this run gates that the file still collects cleanly there.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kgrujcic/54761-mla-matmul-640)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kgrujcic/54761-mla-matmul-640)


mbezuljTT:
@ipotkonjak-tt please take a look

mbezuljTT:
@kgrujcic-tt could you please update PR description with compute utilization %

ipotkonjak-tt (commented):
Does the mla config for 640 tokens agree with config used here? If not do we need to update the model config file as well?

Note that this file was used while optimizing matmuls and is not in anyway linked to the actual configuration used in the model itself. Maybe we should actually make that link if we are to keep the file up to date.

Also, i'm unsure whether this file is actually tested somewhere in the CI? Not sure if there is a need? Unit now it was useful as a tool rather that a classic test

nostojicTT:
Moved to utils after offline discussion, to keep local testing of the MLA matmuls; no CI collects it.
 Program configs are taken from each weight's 4096 entry in MLA_MATMUL_CONFIG, with only per_core_M adjusted for the 640-row chunk.

ipotkonjak-tt (approved):

PR #54806, opened 2026-08-29
[Bug fix][LLK tests] Disable Blackhole instruction gathering, as tt-metal firmware does

### LINKED ISSUE #52997: [LLK][BH] SfpuLcm dest_acc=Yes hangs: extra dest SFPLOAD after SFPU is codegen-sensitive
## Summary

`test_sfpu_binary.py::test_sfpu_binary_int_uniform[mathop:SfpuLcm-dest_acc:Yes]` hangs on Blackhole with:

```
TENSIX TIMED OUT ... Timeout reached: waited 2 seconds for Math, Packer, Unpacker
```

The unmodified test passes on `main`. Inserting extra RISC `nop`s in the always-inlined SFPU binary caller is enough to hang it.

GCD dest_acc=Yes on the same harness still passes.

## Failing test

```
tt_metal/tt-llk/tests/python_tests/test_sfpu_binary.py::test_sfpu_binary_int_uniform[mathop:SfpuLcm-dest_acc:Yes]
```

## Hang site

BH LCM (`tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/ckernel_sfpu_lcm.h`) loads `a`/`b`, runs `gcd_body<15>`, then a Newton reciprocal in LRegs, then **reloads `a`/`b` from dest**:

```c++
TTI_SFPSETEXP(0, p_sfpu::LREG0, p_sfpu::LREG3, 0);

// Load a and multiply by 1/gcd(a, b)
TT_SFPLOAD(p_sfpu::LREG0, InstrModLoadStore::INT32, 3, dst_index_in0 * dst_tile_size);
```

That extra dest `SFPLOAD` after the reciprocal is the hang. GCD never reloads dest after compute, so it does not hit this.

## Repro

Repro branch: halgh/lcmrace

Branch this bug was first found on: https://github.com/tenstorrent/tt-metal/pull/52887

Confirmed on `main` at `7566a67e9c96193297e16b35f6e58751e925f214`. The only local change is eight RISC nops at the end of `_sfpu_binary_check_` in

`tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_sfpu/llk_math_eltwise_binary_sfpu_macros.h`

```c++
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
    asm volatile("nop");
```

That function is always-inlined immediately before LCM. Then on Blackhole, from `tt_metal/tt-llk`:

```bash
tt-smi -r   # if a prior run timed out
bash .cursor/scripts/run_test.sh run \
  --worktree "$(pwd)" \
  --arch blackhole \
  --test test_sfpu_binary.py \
  --k "test_sfpu_binary_int_uniform and SfpuLcm" \
  --timeout 120
```

Expected without the nops: pass (~1s).
Expected with the nops: Tensix timeout (~2s wait for Math/Packer/Unpacker).

## Notes

- Int32 LCM only runs with `dest_acc=Yes`.
- A `STALLWAIT(STALL_SFPU, WAIT_SFPU)` before the extra dest reload unblocks the hang. The LCM sequence is valid on `main` without it; the hang is sensitive to RISC codegen in the inlined caller.

### LINKED ISSUE #53415: [TT-NOP Injector][BH] hang race on transpose_dest
Sub-issue of #53394 (TT-NOP injector tool race findings - LLK).

## Summary

Blackhole `transpose_dest` hangs when math sync is delayed. Seen on dest-acc float and int params.

## Recreate

Branch: [`nmohammed/timing-perturbations-tt-llk`](https://github.com/tenstorrent/tt-metal/tree/nmohammed/timing-perturbations-tt-llk)
Commit: [`1ec846ac199`](https://github.com/tenstorrent/tt-metal/commit/1ec846ac19961585cecd2fee44e9bed5321e27bd) (`Add the ttnop focus tool for timing-perturbation triage`)

```bash
cd tt_metal/tt-llk/tests/python_tests/ttnop
CHIP_ARCH=blackhole TTNOP_SITE_MODE=sync TTNOP_THREADS=math TTNOP_SITES=math:9 TTNOP_FILLER=tti_nop TTNOP_DELAYS=6-7,27-29 ./focus.sh 'test_transpose_dest.py::test_transpose_dest_float[fmt_dest_acc_math_transp_unpack_to_dest:(InputOutputFormat[A:Float32,B:Float32,out:Bfp8_b], <DestAccumulation.Yes: True>, <Transpose.Yes: True>, True)]'
```

amahmudTT:
### PR Category
Bug fix

### Summary

Fixes #52997
Fixes #53415

Blackhole boots with instruction gathering enabled. tt-metal firmware disables it on
every RISC around Tensix in `configure_gathering()`
(`tt_metal/hw/inc/internal/firmware_common.h`, the #16439 workaround). The tt-llk test
harness never did, so the tests ran in a configuration that never ships — and that is
the root cause behind both hangs above.

`ckernel.h`'s `load_replay_buf()` is the complement: it brackets the replay-record window
with disable/enable, but only `#if defined(ENABLE_GATHERING)`. Those are two halves of
one contract — *either* gathering is off globally, *or* it is on and the record window is
bracketed:

| | `ENABLE_GATHERING` defined | not defined |
|---|---|---|
| tt-metal `configure_gathering()` | leaves gathering on | disables globally |
| tt-llk `load_replay_buf()` | brackets the record window | bracket compiles out |

The harness implemented **neither** half: it never defines `ENABLE_GATHERING`, so no
bracketing, and never wrote cfg0, so no global disable. It fell between the two.

This calls `configure_gathering()` from `do_crt0()`, which both `_start()`s
(`brisc.cpp`, `trisc.cpp`) run — so it covers BRISC and all three TRISCs, before any
global constructor and before any Tensix instruction. The guard is copied from tt-metal
verbatim, so defining `ENABLE_GATHERING` both leaves gathering on *and* switches on the
`load_replay_buf` bracketing.

Blackhole only. Wormhole has no gathering (see the note in
`tt_llk_wormhole_b0/common/inc/ckernel_code_sequence.h`); Quasar manages its own policy
in `tt_llk_quasar/common/inc/ckernel.h`.

### Notes for reviewers

**Scope of the fix.** This removes a *test-only* exposure: production firmware already
disables gathering, so there was no product-side gap. It does not change the hardware
behaviour behind #16439 — if anyone sets `ENABLE_GATHERING`, the hazard is back (and the
replay-record bracketing turns on with it, which is the intended pairing).

**Why the regression tests assert on the binary, not on behaviour.** The gathering hazard
is a codegen lottery: a kernel that happens to be laid out safely passes with gathering
on, so a green functional test proves nothing about the configuration. CSR `0x7c0` is
write-only throughout this codebase (every use is `csrrs zero` / `csrrc zero`), so it
cannot be read back and checked at runtime either. `test_gathering_config.py` therefore:

1. asserts every RISC's built ELF carries the gathering-disable sequence, decoded as
   RISC-V instruction words in executable sections;
2. asserts the harness copy has not drifted from tt-metal's `configure_gathering()`.

Both fail without the `boot.h` change (verified with the build cache wiped, so a stale
ELF cannot produce a false pass) and pass with it. They run under `--compile-producer`,
so they need no device.

The ELF check uses a lower bound rather than an exact count, so an unrelated future cfg0
write (tt-metal's `configure_l1_data_cache()` uses the same encoding) cannot turn it into
a spurious failure; the exact sequence is pinned by the drift test instead.

**Verification.** 933 Blackhole and 654 Wormhole test compiles pass. In simulation, the
`transpose_dest` configuration that reproduces #53415 completes cleanly with this change,
where it previously did not.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:amahmud/llk-tests-disable-gathering)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:amahmud/llk-tests-disable-gathering)

### DIFF
--- a/tt_metal/hw/inc/internal/firmware_common.h
+++ b/tt_metal/hw/inc/internal/firmware_common.h
@@ -271,6 +271,9 @@ bool is_message_go() {
 #define EARLY_RETURN_FOR_DEBUG_EXIT
 #endif
 
+// NOTE: the tt-llk test harness keeps a synchronized copy of this function in
+// tt_metal/tt-llk/tests/helpers/include/boot.h, so its firmware boots in the same gathering
+// configuration this one does. test_gathering_config.py fails if the two drift apart; edit both.
 inline __attribute__((always_inline)) void configure_gathering() {
 #if defined(ARCH_BLACKHOLE) && !defined(ENABLE_GATHERING)
     // Workaround for tt-metal#16439, making sure gathering multiple instructions issued to Tensix is disabled
--- a/tt_metal/tt-llk/tests/helpers/include/boot.h
+++ b/tt_metal/tt-llk/tests/helpers/include/boot.h
@@ -26,6 +26,30 @@ extern void (*__init_array_end[])(void);
 // even though -fno-asynchronous-unwind-tables -fno-exceptions flags are set
 void* __gxx_personality_v0;
 
+// Mirror of tt-metal firmware's configure_gathering() (tt_metal/hw/inc/internal/firmware_common.h).
+// Blackhole boots with instruction gathering on; tt-metal firmware disables it on every RISC.
+// The guard matches tt-metal's, so defining ENABLE_GATHERING both leaves gathering on and
+// switches on the load_replay_buf record-window bracketing in ckernel.h.
+TT_ALWAYS_INLINE void configure_gathering()
+{
+#if defined(ARCH_BLACKHOLE) && !defined(ENABLE_GATHERING)
+    asm(R"ASM(
+        .option push
+        li   t1, 0x2
+        csrrs zero, 0x7c0, t1
+        li   t1, 0x1
+        slli t1, t1, 18
+        fence
+        csrrs zero, 0x7c0, t1
+        li   t1, 0x2
+        csrrc zero, 0x7c0, t1
+        fence
+        .option pop
+         )ASM" ::
+            : "t1");
+#endif
+}
+
 __attribute__((no_profile_instrument_function)) TT_ALWAYS_INLINE void do_crt0()
 {
     asm volatile(
@@ -37,6 +61,9 @@ __attribute__((no_profile_instrument_function)) TT_ALWAYS_INLINE void do_crt0()
 
     // Set stack pointer
     asm volatile("la sp, %0" : : "i"(__stack_top) : "memory");
+
+    // Before any global constructor or Tensix instruction can run.
+    configure_gathering();
 
     // Initialize .bss
     for (volatile std::uint32_t* p = (volatile std::uint32_t*)__ldm_bss_start; p < (volatile std::uint32_t*)__ldm_bss_end; p++)
--- a/tt_metal/tt-llk/tests/python_tests/test_gathering_config.py
+++ b/tt_metal/tt-llk/tests/python_tests/test_gathering_config.py
@@ -40,11 +40,18 @@ CSRRS_ZERO_CFG0_T1 = 0x7C032073  # csrrs zero, 0x7c0, t1   (funct3=010)
 CSRRC_ZERO_CFG0_T1 = 0x7C033073  # csrrc zero, 0x7c0, t1   (funct3=011)
 
 # configure_gathering() issues csrrs twice (bit 1, then bit 18) and csrrc once (bit 1).
-# Asserted as a lower bound: other cfg0 writes (tt-metal's configure_l1_data_cache()
-# uses the same encoding) must not turn this into a spurious failure. The exact
-# sequence is pinned by test_harness_gathering_sequence_matches_tt_metal below.
+# Counted as a lower bound so an unrelated cfg0 write (tt-metal's
+# configure_l1_data_cache() uses the same encoding) cannot cause a spurious failure.
+# That tolerance alone would admit a false pass -- another writer could satisfy the
+# bound while configure_gathering() itself had been dropped -- so the ordered triple
+# must also appear inside one short window, which is what proves this sequence, and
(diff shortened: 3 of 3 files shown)


fvranicTT (approved):

amahmudTT:
Disabling gathering is exposing an issue, that is why pr gate is failing. There is a separate issue to track that problem once that one is merged, this one will pass too. Will merge it then.

amahmudTT:
This needs to be merged before this can be merged, as this (gathering) pr exposes the issue fixed by this pr https://github.com/tenstorrent/tt-metal/pull/54683

abhullar-tt (approved):

PR #54829, opened 2026-08-29
Add ttnn.rmsnorm_distributed_bw

abogdanovicTT:
### Ticket

N/A

### Problem description

Forward already has ttnn.rms_norm_pre_all_gather / ttnn.rms_norm_post_all_gather. There was no matching backward, so tensor-parallel RMSNorm could not be trained with the same pre / all-gather / post split.

### What's changed

Adds ttnn.rms_norm_pre_all_gather_bw and ttnn.rms_norm_post_all_gather_bw, matching the forward split. Only E[x^2] (forward stats) and E[x * g] (bw stats) are all-gathered; weight_grad stays on the shard.

- ttnn.rms_norm_pre_all_gather_bw(input, output_grad, stats, …) - local sum(x · g) padded to one tile column, then all-gather on the last dim.
- ttnn.rms_norm_post_all_gather_bw(input, output_grad, stats, bw_stats, …) - returns [input_grad, weight_grad]; weight_grad is None when weight is omitted.

Call sequence matches forward: pre → all_gather on the last dim → post.

The post **apply** step is now one reader / compute / writer program (`generic_op`), not a chain of full-size eltwise ops:
- `dx = gamma * dy / rms - x * E[x·g] / rms^2` stays in L1 (SFPU dest-chain, unpack-to-dest fp32).
- When weight is set, `dgamma = sum(dy * x / rms)` is pack-accumulated per core and reduced on-chip (two-stage Tensix tree) instead of `ttnn.sum` over `{0,1,2}`.
- Tiny column stats (`reciprocal` / `square` / `multiply` of `rms` and `E[x·g]`) stay as host composites.

### Note 
ttml can now work around a fused distributed RMSNorm backward by composing these ops with all-gather, rmsnorm and scatter. After adding ttnn.rmsnorm_distributed_bw, a ttml wrapper for it could be added.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abogdanovic/rmsnorm_distributed_bw)


ttnn/cpp/ttnn/operations/normalization/rmsnorm_distributed/device/kernels/compute/rmsnorm_bw_apply.cpp
```diff
@@ -0,0 +1,152 @@
...
+        }
+    };
+
+    if constexpr (with_dgamma) {
+        cb_reserve_back(cb_acc, Wt);
+        zero_fill_acc();
+    }
+
+    mul_binary_tile_init();
+    addcmul_tile_init();
```
mstaletovicTT, line 51: you generally do not want to have multiple inits on the same resource "live" at the same time

Your main resources are unpacker, math, packer, and SFPU.

Unpacker and math are mostly inited together, with calls such as `mul_tiles_init` or `matmul_block_init`.

Packer is most often inited only once, in `compute_kernel_hw_startup`

SFPU is also standalone, with calls such as `mul_binary_tile_init()`

You are supposed to only have one init "live" per resource, so stacking two inits here back to back is, in theory, incorrect. I think that these particular ones do not do anything super important, so it works, but in general this is not good practice and could lead to issues

ttnn/cpp/ttnn/operations/normalization/rmsnorm_distributed/device/kernels/compute/rmsnorm_bw_apply.cpp
```diff
@@ -0,0 +1,152 @@
...
+            tile_regs_wait();
+            pack_reconfig_data_format(cb_out);
+            pack_tile(0, cb_out);
+            tile_regs_release();
+
+            if constexpr (with_dgamma) {
+                tile_regs_acquire();
+                copy_tile_init(cb_dy);
+                copy_tile(cb_dy, c, 0);
+                copy_tile_init(cb_x);
```
mstaletovicTT, line 87: if cb_dy and cb_x are the same datatype, this init is unnecessary. I would guess that they are, since if they were not, a reconfig would also be needed

ttnn/cpp/ttnn/operations/normalization/rmsnorm_distributed/device/kernels/dataflow/writer_rmsnorm_bw_apply.cpp
```diff
@@ -0,0 +1,123 @@
...
+            if (role == 2) {
+                open_and_signal(ready2, n_leaders, /*tab_off=*/row_cols, /*along_x=*/false);
+            }
+            ready2.wait_min(1);
+            scatter_row(get_read_ptr(cb_part), gather_l1, root_vx, root_vy, my_row, Wt, page);
+            noc_async_write_barrier();
+            arrive2.up(noc, root_vx, root_vy, 1);
+            cb_pop_front(cb_part, Wt);
+        }
+        if (role == 2) {
```
mstaletovicTT, line 110: probably document what these roles mean

PR #54850, opened 2026-08-29
Functional Kimi-K3 disaggregated prefill

### LINKED ISSUE #53508: Functional Kimi-K3 Prefill
Wire Kimi-K3 components into the model and make a test run against the golden trace.

Branches: **`feature/kimi-k3-modules`** (unmerged ttKDA #52799 + AttnRes #53116/#53126/#53318 rebased on `main`) and **`feature/kimi-k3-prefill`** (the model, branched from it). Progress comments below.

kgrujcic-tt:
Closes #53508.

Functional Kimi-K3 (2.8T, 93 layers, MLA + KDA hybrid with AttnRes) disaggregated prefill, on the
shared prefill runner and in CI. Rebased onto `main` @ `54a826b1d93`, 354 commits of drift.

## Hardware
93 layers do not fit one Galaxy. Four ranks, 24/24/24/21: three is lopsided and puts one rank on the
MLA circular-buffer ceiling (#54876).

## Accuracy, 4xGLX, 93 layers, vs the 1M vLLM golden

| measurement | result |
|---|---|
| KV, head `[0,56320)` | per-rank min **0.9938 / 0.9915 / 0.9379 / 0.8999**, per-layer min 0.8999 at L87, 24/24 layers |
| KV, tail `[993280,1044480)` | 24/24 layers; `nope` mean 0.8405, `pe` mean 0.8072, min 0.5503 at L51 |
| hidden state | 93 layers x 21 chunks = 1953 comparisons; head mean 0.9715, tail mean 0.8355 |
| next token | **21/21 identical** to the golden, head and tail |

Single Galaxy up to 24 layers: depth ladder, chunked prefill, traced path and KV slots all pass.
Per-layer tables, plots and the PCC methodology are in
[#53508](https://github.com/tenstorrent/tt-metal/issues/53508#issuecomment-5747526350).

Intermediate PCC understates output correctness at depth: at position 1,040,383 the layer-92 hidden
state scores 0.31 and the logits 0.78, yet the predicted token matches and so do all five top
candidates.

## Throughput
Captured-trace path, one user, 93 layers on 4xGLX, KV allocated for the full 1,049,600-token
window. Derived from a single 204-chunk run: prefill is sequential, so the first N chunks are what
an N-chunk run computes.

| tokens | steady tok/s, traced | steady tok/s, eager |
|---|---|---|
| 56,320 | 14,501 | 6,732 |
| 261,120 | 10,232 | 6,210 |
| 522,240 | 6,794 | 5,797 |
| 1,044,480 | 3,988 | 3,658 |

Cost per chunk is `ms = 282 + 1334 * depth/1M`, against `374 + 1426` on the pre-rebase traced run,
so both the fixed cost and the depth slope improved. The traced gain shrinks with depth because
tracing removes a fixed dispatch cost, a falling share as MLA attention grows with position.

Wall-clock throughput is feed-limited rather than device-limited: at ~325 ms/chunk traced against
~950 ms/push the single-process producer is the bottleneck, and the device idles 22% of the run.
Startup dropped from ~8 s to under 1 s per rank by warming the KV-ack programs during compile.

Multi-user traced is not measurable on this branch -- a capture resolves the KDA slot at capture
time, so `set_trace_controller` refuses more than one allocated slot (#57135).

## CI
12 Kimi-K3 legs in `blaze_models_prefill_tests.yaml`: host-only contracts, depth ladder, chunked and
traced prefill, KDA and MLA accuracy and perf, MoE, and the disaggregated runner on SC1 and SC4.
10 green on [run 35504488109](https://github.com/tenstorrent/tt-metal/actions/runs/35504488109);
the SC4 leg hit a runner chip-init fault that skipped the test step (#57119) and `kimi_moe_perf`
fails on `main` with the identical measurement (#57103).

## Fixed here
- Layer-ack service configured in layer space, so a K3 rank's 6 acks per chunk were numbered as 24.
  Now derived from `kv_slot_layer_ids`, on both ack transports; a no-op for dense models.
- Producer device-map name did not match the writer's at one rank; reader uses
  `rank_scoped_device_map_path`.
- A traced capture resolved the KDA slot at capture time; multi-slot traced now refuses rather than
  corrupting silently.
- A missing KV golden warned instead of failing.
- Multi-user KDA reset wiped every user's recurrent state instead of the one starting.
- Two post-rebase breakages from other models' changes (`tp_shard_kv`, `rope_setup` on a NoPE
  model), now gated by a host-only contract test.
- Runner KV read-back for a stack where only some layers own a KV slab, and a table-wait timeout
  sized for a 93-layer bringup.
- CI: `k3_contracts` was the only bare `cmd` in the matrix and the container hook could not launch
  it; `k3_transformer_traced` was capped from a warm-cache measurement while the weight load
  dominates.

## Not included
Multi-turn KDA resume. Pre-existing KDA op issues against `main`: #56239, #56240.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:feature/kimi-k3-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:feature/kimi-k3-prefill)


models/demos/common/prefill/runners/ci/run_multirank_pcc.sh
kgrujcic-tt, line 1: Note to myself: This file has be unslopped; looks like the default behavior is changed.

dpopovTT (approved):

.github/workflows/blaze-models-prefill-tests.yaml
```diff
@@ -12,7 +12,7 @@ on:
...
+          Valid stages: dsa_mla_glm, glm_chunked, glm_chunked_accuracy, glm_moe, glm_prefill_block, hca_chunked, hca_perf, high_bw_all_gather_perf, k3_kda, k3_kda_perf, k3_mla, k3_mla_perf, k3_mla_trace, k3_prefill_runner_24l, k3_prefill_runner_93l, kimi_chunked, kimi_chunked_padded, kimi_dflash, kimi_k3_moe, kimi_moe, kimi_moe_perf, kv_cache_table, minimax_prefill_kv_pcc, minimax_prefill_perf, mistral4_chunked_padded, mistral4_mla, mistral4_mla_chunked, mistral4_moe, mistral4_prefill_block, mistral4_prefill_transformer, mla_chunked, moe_gate, prefill_block_determinism, prefill_runner, prefill_runner_kv_pcc, prefill_transformer_determinism, ring_joint_sdpa_perf, swa_mixed, swa_perf.
```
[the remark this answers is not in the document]
mbezuljTT: looks fair? @kgrujcic-tt

models/demos/common/prefill/runners/topology_configuration/ci/kimi_k3_sc1_mgd.textproto
```diff
@@ -0,0 +1,15 @@
+mesh_descriptors {
```
mbezuljTT, line 1: I dont understand why we have mdg per model; I acknowledge that is how it is being done - please follow up with Jaksa when he is back

models/demos/common/prefill/runners/prefill_runner.py
```diff
@@ -94,7 +94,9 @@ def sd(key, val):
 )
 SYNC_PER_CHUNK = os.environ.get("PREFILL_SYNC_PER_CHUNK", "0") == "1"
 TIMING_DIR = os.environ.get("PREFILL_TIMING_DIR", "")
-_L1_SMALL_SIZE = ADAPTER.l1_small_size
+# Env-overridable: on Kimi-K3 AttnRes wants L1_SMALL large and MLA's chunked attention wants it
```
mbezuljTT, line 97: this should not stay ambiguous; there needs to be clear and only way to define L1_SMALL

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
```diff
@@ -0,0 +1,255 @@
...
+    cache_user_id: int = 0
+    actual_start: Optional[int] = None
+    # Absolute KV position past this chunk's last real token. MLA clamps its cache write to it
+    # (#54744); the traced path takes the same value off `metadata[2]` on device instead.
+    actual_end: Optional[int] = None
+    # The traced path's `(slot_id, actual_start, actual_end)` triple of 1-element uint32 tensors.
+    metadata: Optional[tuple] = None
+
+
+class K3Attention(Protocol):
```
mbezuljTT, line 53: "house rule" was to inherit from LightweightModule which was historically used to replace pytorch nn module that had bunch of (hist) overhead we didn't care about. 

it is literally this 

```
class LightweightModule:
    """Torch modules add a surprising amount of host overhead for attribute
    access and method calls. This class is a lightweight alternative that
    just wraps a forward function for now."""

    def __call__(self, *args, **kwargs):
        return self.forward(*args, **kwargs)
```

I dont know if this is a real problem with protocol. no actions required at this time.

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
```diff
@@ -0,0 +1,255 @@
...
+        hidden = ttnn.squeeze(gathered, dim=0)
+        ttnn.deallocate(gathered)
+
+        if self._states is None:
+            raise ValueError(
+                f"KDA layer {self.layer_idx} has no state cache; call bind_state_cache() before the first forward"
+            )
+        output, new_state = self.kda.forward(hidden, self._states.read(self.layer_idx, ctx.cache_user_id))
+        ttnn.deallocate(hidden)
+        self._states.commit(self.layer_idx, new_state, ctx.cache_user_id)
```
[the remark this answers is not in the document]
mbezuljTT: is this a real issue? @kgrujcic-tt

models/demos/deepseek_v3_d_p/tt/kimi_k3/transformer.py
```diff
@@ -0,0 +1,498 @@
...
+        # as they are created.
+        self.kda_states: KdaStateCache | None = None
+        kda_layers: dict[int, object] = {}
+
+        self.layers = []
+        for local_idx in range(num_layers):
+            layer_idx = first_layer_idx + local_idx
+            is_last = local_idx == num_layers - 1
+            layer_state = state_dict["layers"][local_idx] if state_dict.get("layers") else {}
+            attention = build_attention(
```
mbezuljTT, line 249: what is the motivation to build attention outside of block and then pass it into the block? 

is this on purpose or ai slop ? 
connecting this to previous question of attention using protocol and lightweight module used here. 

pick one and stick to it - please.

models/demos/deepseek_v3_d_p/tt/kimi_k3/weights.py
```diff
@@ -0,0 +1,290 @@
...
+    while everything `quantization_config.ignore` covers is bf16 in both. A caller with both on the
+    box can take the attention from whichever is closer to hand.
+    """
+    checkpoint_dir = Path(checkpoint_dir)
+    prefix = layer_prefix(checkpoint_dir, layer_idx)
+    is_mla = layer_idx in model_cfg.mla_layer_ids()
+    is_moe = layer_idx >= model_cfg.NUM_DENSE_LAYERS
+
+    wanted = {alias: prefix + key for alias, key in LAYER_NORM_KEYS.items()}
+    if is_mla:
```
mbezuljTT, line 136: kda layers not covered here - is it on purpose?

mtairum (approved):

models/demos/common/prefill/runners/prefill_producer.py
```diff
@@ -80,14 +80,19 @@ def sd_bool(key, val):
...
-    global SP_AXIS, TP_AXIS, GLOBAL_MESH_SHAPE, CHUNK_SIZE, MAX_SEQ_LEN, NUM_LAYERS, ADAPTER
+    global SP_AXIS, TP_AXIS, GLOBAL_MESH_SHAPE, CHUNK_SIZE, MAX_SEQ_LEN, NUM_LAYERS, ADAPTER, NUM_ACK_LAYERS
     SP_AXIS = int(os.environ.get("PREFILL_SP", 8))
     TP_AXIS = int(os.environ.get("PREFILL_TP", 4))
     GLOBAL_MESH_SHAPE = (SP_AXIS, TP_AXIS)
     CHUNK_SIZE = int(os.environ.get("PREFILL_CHUNK_SIZE", 5 * 1024))
     MAX_SEQ_LEN = int(os.environ.get("PREFILL_MAX_SEQ_LEN", CHUNK_SIZE * 11))
     NUM_LAYERS = int(os.environ.get("PREFILL_NUM_LAYERS", 61))
     ADAPTER = get_adapter(os.environ.get("PREFILL_MODEL", DEFAULT_MODEL))
+    # One ack fires per layer that WRITES KV, not per layer. Dense stacks are the same number, but a
```
mbezuljTT, line 91: for my education, why are we not acking on KDA layers?

mbezuljTT (approved):
Thank you @kgrujcic-tt for driving this bringup. 

My comments are mostly cosmetics / code organizations and are not blocking. 

Please address them (by dismissing, addressing, or creating a follow up issues) 


Please have someone approve from runner perspective.

.github/time_budget.yaml
```diff
@@ -192,8 +192,12 @@ shield:
     wh_n150: 100
     bh_p150: 60
   demo:
-    bh_sc4: 50
-    bh_sc1: 75
+    # Sum of the three SC4 disaggregated-prefill legs: Kimi-K2.7 20, GLM-5.2 30, Kimi-K3 100.
+    # Kimi-K3 is 93 layers against the others' depths and its bringup is dominated by streaming a
+    # 1.5 TB TTNN cache off the models share on four ranks at once -- measured 42.6 and 65.9 min to
+    # all ranks serving on runs 34648535999 and 34659612357.
```
ipotkonjak-tt: [nit] lets minimize adding claude comments to the code

models/demos/common/prefill/runners/prefill_producer.py
```diff
@@ -80,14 +80,19 @@ def sd_bool(key, val):
...
+    # hybrid one is not: Kimi-K3 at 93 layers acks only its 24 full-attention layers, so draining
+    # NUM_LAYERS * pushes waits for acks that are never sent and burns the full 600 s timeout twice,
+    # once after warmup and once after the measured request.
+    NUM_ACK_LAYERS = getattr(ADAPTER, "num_kv_cache_layers", lambda n: n)(NUM_LAYERS)
```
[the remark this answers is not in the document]
ipotkonjak-tt: high lvl makes sense, should be verified @kgrujcic-tt 

but i think decode needs us to migrate  last KDA state per layer in addition to MLA cache?

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
```diff
@@ -0,0 +1,255 @@
...
+        )
+        # `[1, 1, T, d]` -> `[1, T, d]`: ttKDA takes three dimensions, and the leading one is batch.
+        hidden = ttnn.squeeze(gathered, dim=0)
+        ttnn.deallocate(gathered)
+
+        if self._states is None:
+            raise ValueError(
+                f"KDA layer {self.layer_idx} has no state cache; call bind_state_cache() before the first forward"
+            )
+        output, new_state = self.kda.forward(hidden, self._states.read(self.layer_idx, ctx.cache_user_id))
```
ipotkonjak-tt: is kda metadata aware, seems it's not passed to the module? if not we should assert that traced execution of the model is not supported atm

models/demos/deepseek_v3_d_p/tt/kimi_k3/block.py
```diff
@@ -0,0 +1,280 @@
...
+        layer_idx: int,
+        local_idx: int,
+        attention,
+        seq_len: int,
+        *,
+        num_links: int = 1,
+        topology=ttnn.Topology.Linear,
+        sp_axis: int = 0,
+        tp_axis: int = 1,
+        is_balanced: bool = False,
```
ipotkonjak-tt: do we plan to support balanced execution? if no lets remove to avoid confusion

models/demos/common/prefill/runners/prefill_producer.py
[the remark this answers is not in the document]
kgrujcic-tt: Confirmed and fixed in 27db30eb.

The service was configured in layer space. Deriving the counts from `kv_slot_layer_ids` instead gives 6/6/6/6 over a total of 24, so the seq stream is dense from chunk 0 and matches the producer's `NUM_ACK_LAYERS` exactly. Dense models have one ack per layer, so acks == layers and it is a no-op for them; sc1 was already accidentally correct because its single rank collapses seq to the record index.

Simulating the old derivation over 21 chunks on the 24/24/24/21 split: 504 records pushed, 471 injected, 33 stranded. The lag is what your "final ACKs remain buffered" describes; there are no permanent mid-stream gaps, because the per-rank blocks tile the seq space.

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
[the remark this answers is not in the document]
kgrujcic-tt: Confirmed and fixed in ca3d0b4c by refusing the combination rather than silently corrupting it.

There is no small change that makes multi-slot traced KDA correct: replay runs no Python, so the carry would have to be one `[num_slots, ...]` tensor indexed on-device from the metadata, or one capture per slot, which a 256 MB trace region cannot afford. The guard is at `set_trace_controller`, the earliest K3-specific trace hook.

No CI leg hits it today, since every K3 leg is either untraced or single-slot.

models/demos/common/prefill/runners/prefill_producer.py
```diff
@@ -165,19 +170,41 @@ def _parse(raw: str) -> dict:
...
 
-    def _matches():
+    # A KV read is UMD-local: read_dram_umd only reaches chips visible to THIS process. Under
+    # pipeline parallelism every rank exports its own map, and merging them all makes the caller
+    # believe it can resolve another galaxy's chips -- such a layer then clears the visibility guard
+    # and dies inside the read with "no visible chip with ASIC unique_id".
+    def _own():
+        # This rank's own map: the only chips this process can actually reach.
+        own = f"{stem}_r{rank}{ext}"
+        return [own] if os.path.exists(own) else []
```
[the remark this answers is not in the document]
kgrujcic-tt: Confirmed and fixed in 1abb3bd3.

`_own()` now resolves the name with `rank_scoped_device_map_path()`, the same helper the runner writes it with, so the single-rank case picks the base path. Threading `num_ranks` through keeps the property the rank scoping exists for -- never adopt a peer's map -- rather than falling back when a sidecar is merely late.

The sc1 leg had not reached this code: it was being killed on the job timeout during bringup. Its timeout is now 60 min, so it would have surfaced here.

models/demos/common/prefill/runners/prefill_producer.py
```diff
@@ -712,22 +761,44 @@ def _read_slot_kv_and_check_pcc_mla(table, device_map: dict, slot_id: int, real_
...
+        + (f"; {len(unreferenced)} layers carry no KV golden (hybrid stack): {unreferenced}" if unreferenced else "")
     )
+    if missing_golden:
+        # These layers DO own a slab, so the golden should have covered them. Naming them is the
+        # difference between a partial trace and a gate that quietly narrowed itself.
+        logger.warning(
+            f"[producer] slot {slot_id}: {len(missing_golden)} slab-owning layer(s) have no KV golden "
+            f"under {trace_dir} and were not scored: {missing_golden}. The trace is partial or "
+            f"PREFILL_TRACE_DIR is wrong; the PCC below covers only the rest."
+        )
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in d508b103. `missing_golden` now raises.

Checked it cannot false-fail: for every configuration we run -- sc4 at 93 layers, sc1 at 24, and the head/tail trace -- all expected slabs have a golden, so the set is empty.

models/demos/common/prefill/runners/topology_configuration/ci/kimi_k3_sc1_mgd.textproto
kgrujcic-tt: Agreed, and it is worse than it looks: all six per-model descriptors under `topology_configuration/ci/` are byte-identical once comments are stripped. I diffed them -- `kimi_k3_sc1` vs `glm52_sc1` vs `kimi27_sc1`, and `kimi_k3_sc4` vs `kimi27_sc4`, all zero differing lines.

The only thing keeping them apart is `run_multirank_pcc.sh` deriving the path as `${MGD_DIR}/${MODEL}_${CONFIG}_mgd.textproto`. Collapsing them to `sc1`/`sc4` is a one-line change there plus four deletions, but it touches three models' legs, so I would rather not fold it into the rebase. Filing a follow-up and will raise it with Jaksa.

models/demos/common/prefill/runners/prefill_runner.py
kgrujcic-tt: Fixed in c33efcba -- there is now one definition, `KimiK3Config.L1_SMALL_SIZE`.

There were two 4096s: the adapter's `l1_small_size`, which the runner reads via `PREFILL_L1_SMALL_SIZE`-overridable `ADAPTER.l1_small_size`, and a literal in the chunked test's `device_params`, which exists because the mesh fixture is built before any adapter is resolved. Both now read the config constant, which sits next to `FABRIC_PAYLOAD_SIZE` that config already owns for the same reason.

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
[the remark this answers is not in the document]
kgrujcic-tt: Yes, it is real. Verified against the rebased tree and fixed in ca3d0b4c.

The KDA arm selects its carry with the host scalar `ctx.cache_user_id`, so the slot is resolved when the trace is captured. `_forward_traced` captures with `cache_user_id=0` while the eager path passes the real slot, so every replay reads and overwrites slot 0's carry. It is silent: the chunk's own output still looks right, only the next chunk inherits the wrong state. MLA is unaffected because it takes its slot from device-resident metadata.

No CI leg hits it -- every K3 leg is untraced or single-slot -- which is why it went unnoticed. A correct fix needs the carry indexed on-device from that metadata, so for now the combination is refused at `set_trace_controller` rather than silently corrupted.

models/demos/deepseek_v3_d_p/tt/kimi_k3/weights.py
kgrujcic-tt: They are covered, just a few lines further down. This block builds the `wanted` key map for tensors read out of the MLA checkpoint; the KDA tensors are loaded separately at the `else` branch of `if is_mla`:

```python
state_dict["kda_weights"] = load_kda_layer_state_dict(
    Path(kda_checkpoint_dir or checkpoint_dir), layer_idx, kimi_k3_kda_config()
)
```

Separate because the two checkpoints are not interchangeable per-tensor -- the routed experts are MXFP4 in the published checkpoint and bf16 only in the dequantized export -- so `kda_checkpoint_dir` can point somewhere else. Every layer gets exactly one attention weight group either way.

models/demos/common/prefill/runners/prefill_producer.py
kgrujcic-tt: Because a KDA layer owns no KV slab, so it has nothing to ack. The ack signals that a layer's KV is resident and readable by the decode side; `block.py` gates it on `attention.writes_kv`, so only the 24 MLA layers of 93 fire one.

That asymmetry was in fact mis-wired on the service side, which your question is close to: `LayerAckService` was still configured with the rank's LAYER count while the rank emits only ack-count records. Fixed in 27db30eb -- details in the Copilot thread above.

models/demos/deepseek_v3_d_p/tt/kimi_k3/attention.py
kgrujcic-tt: Done anyway, in 4f4e0f1 -- `TtK3MlaAttention` and `TtK3KdaAttention` now inherit `LightweightModule`. It costs nothing and matches what everything else here does.

Keeping the `K3Attention` Protocol alongside it: the two serve different jobs. `LightweightModule` supplies `__call__ -> forward`; the Protocol is the structural contract the block programs against (`layer_idx`, `writes_kv`, `forward`) and is what lets `block.py` stay branchless on MLA vs KDA.

kgrujcic-tt: Correction: that commit is `2fe66b3a`, not `4f4e0f1`.

models/demos/deepseek_v3_d_p/tt/kimi_k3/transformer.py
kgrujcic-tt: On purpose, and it is load-bearing in two ways.

`build_attention` is the single place the MLA/KDA branch happens. `TtKimiK3Block` then never asks what kind of layer it is -- it asks `attention.writes_kv` and nothing else -- so the block has one code path for a stack whose layers alternate 3 KDA : 1 MLA. If the block constructed its own attention, every block would need the schedule and the branch, and the hybrid decision would be smeared across the layer loop instead of living in one function.

The second reason is ordering: `KdaStateCache` allocates its carries *from* the `ttKDA` instances, so the cache cannot exist until the layers do, and the layers must not each allocate their own -- one address per carry for the life of the run is the whole point. Construction is therefore two-phase, and `bind_state_cache` is that seam. Building attention inside the block would invert the dependency.

The comment at `attention.py:53` states the contract; happy to move that rationale next to `build_attention` if that is where you would look for it.

models/demos/deepseek_v3_d_p/tests/kimi_k3/test_layer_schedule.py
```diff
@@ -0,0 +1,102 @@
...
+the global model layer index that names weights, and the rank-local KV slot that indexes the cache —
+and the failure mode when they are confused is silent: `ttMLA._cache_batch_idx` computes
+`cache_user_id * layer_num + cache_layer_idx` against a cache sized to this rank's MLA count, so a
+global slot on a rank with `first_layer_idx > 0` is a plausible integer pointing at another user's
+rows, or past the end.
+
+`KimiK3Config.mla_kv_slot()` is the model-wide map and returns the global slot. The
+`NotImplementedError` in `adapters/kimi_k3.py` names it as the thing to use when wiring the cache
+up, which is right for a single-rank run and wrong for every other one. The test below pins that
+difference rather than leaving it to be rediscovered.
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in `9fc604ad`. `adapters/kimi_k3.py` raises nothing of the sort any more, so the note now states the model-wide versus rank-local slot rule directly instead of sending readers to a diagnostic that is not there.

models/demos/deepseek_v3_d_p/tt/kimi_k3/layer_schedule.py
```diff
@@ -0,0 +1,104 @@
...
+* a **global** model layer index, 0..92, which names the weights in the checkpoint and decides
+  whether a layer is MLA or KDA;
+* a **rank-local** KV slot, which indexes the KV cache this rank allocated.
+
+They are not the same and the difference is not cosmetic. `ttMLA._cache_batch_idx` computes
+`cache_user_id * self.layer_num + cache_layer_idx` against a cache sized to *this rank's* layer
+count, so a rank starting at layer 24 owns MLA layers 27, 31, ... whose slots must be 0, 1, ... —
+not the 6, 7, ... that `KimiK3Config.mla_kv_slot()` returns. That classmethod is the model-wide
+map; `adapters/kimi_k3.py`'s `NotImplementedError` points at it, and taking that hint literally
+gives a pipeline-parallel rank a plausible, wrong, out-of-range slot. Use `kv_slot_of_local` here.
```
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in `9fc604ad`, same as the sibling note in the test. The rationale now states the rule directly and points at `KimiK3Adapter.allocate_kv_cache`, whose docstring already carries it.

models/demos/common/prefill/runners/ci/run_multirank_pcc.sh
kgrujcic-tt: Two defaults changed for models that already used this script; everything else is additive or holds its old value.

- `TABLE_WAIT_SECS` (new, default 3600) replaces the hard-coded `seq 1 360` at 5 s, so the bound on mesh bringup plus weight load plus warmup compile went from 30 to 60 minutes. It bounds bringup, not the chunk loop, and the loop breaks as soon as the table appears, so the shallower models pay nothing for it. 30 min was already too short: K3 at 93 layers over 4 ranks measured 42.6 min from step start to `_serve_request` (run 34648535999), and the script exited 1 while the ranks were still compiling.
- The producer `mpirun` now gets `-x PATH -x LD_LIBRARY_PATH`, matching what the runner already had. Without it the ranks come up and the producer dies with `No module named 'numpy'`, which surfaces only as `PCC GATE FAIL: 0/4 producer verdict file(s) present`.

Unchanged: `SC1_MAX_SEQ_LEN` moved to the top of the file but is still 256000; `SC1_NUM_LAYERS` defaults empty, so `NUM_LAYERS_ENV` is unset for every model that does not ask for it; the `kimi_k3` case is new.

.github/workflows/blaze-models-prefill-tests.yaml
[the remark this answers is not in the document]
kgrujcic-tt: Fixed in this PR. The help text now lists all 12 K3 stages, checked against the matrix rather than by eye: `k3_chunked_prefill`, `k3_contracts`, `k3_kda`, `k3_kda_perf`, `k3_mla`, `k3_mla_perf`, `k3_mla_trace`, `k3_prefill_runner_24l`, `k3_prefill_runner_93l`, `k3_transformer_depth`, `k3_transformer_traced`, `kimi_k3_moe`. Diffing the stage tokens in `tests/pipeline_reorg/blaze_models_prefill_tests.yaml` against the tokens in the description leaves an empty set both ways.

Issue #32620, opened 2025-11-18
Correct Documentation for Ops Under `Tensor Creation` category


ctr-avalevkoTT:
### Issue

Extraction of code examples from docstrings to CI revealed formatting or structural inconsistencies in the operations listed below.

### Category

Tensor creation

### Findings

OP | Binding| Issues
-- | -- | --
ttnn.arrange | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.empty_like | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.ones_like | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.ones | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.full | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.full_like | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table
ttnn.rand | ttnn/cpp/ttnn/operations/rand/rand_pybind.cpp | No supported dtypes and layouts table & `shape (list[int])` without description
ttnn.from_buffer | ttnn/cpp/ttnn-pybind/operations/creation.cpp | No supported dtypes and layouts table

### Deliverables

- Apply fixes to the operations listed in the table above.
- Ensure consistent formatting of Args, Keyword Args, and Examples.
- Validate that examples in the CI (L2 nightly) work as expected.

fahmedchanTT:
@ntarafdar Does your team own this list ?

PR #54898, opened 2026-08-31
#32620: Add supported dtypes and layouts tables to tensor-creation op docstrings

mouliraj-mcw:
### Ticket
Link to Github Issue #32620

### Problem
`ttnn.arange`, `empty_like`, `ones_like`, `ones`, `full`, `full_like`, `rand`, and `from_buffer` in `creation_nanobind.cpp`/`rand_nanobind.cpp` render without a "Supported dtypes and layouts" table, and `ttnn.rand`'s `shape` param has no description — breaking the docstring shape the L2-nightly example extraction relies on.

### Fix
Rewrite the 11 tensor-creation docstrings to the peer single-row `.. list-table::` + prose-caveat shape (matches `binary_nanobind.cpp:267-277`, `ternary_nanobind.cpp:42-53`, `sort_nanobind.cpp`) using each op's actual `full_impl`/`from_buffer`/`arange`/`tensor_spec.cpp:validate_dtype_and_layout` coverage. Drop the stale `Ranks 2, 3, 4` column (no rank guard in `full_impl`) and the stale `TILE-default-when-dtype=BFLOAT4/8_B` clause from `from_buffer` (`creation.cpp:328` fixes `ROW_MAJOR`, `from_buffer_impl` throws for block-float before the default fires — clause was unreachable). L2-nightly examples remain driven by `tests/ttnn/docs_examples/test_tensor_creation_examples.py` + `examples_mapping.py`, with `docs/source/ttnn/_ext/doc_modifier.py` auto-injecting the rendered `Example:` block per page (16/16 pass locally).

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mouliraj/tensor_creation_docs)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mouliraj/tensor_creation_docs)

### DIFF
--- a/ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
+++ b/ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
@@ -71,19 +71,30 @@ void bind_full(nb::module_& mod) {
 
         Args:
             shape (ttnn.Shape): The shape of the tensor.
-            fill_value (float): The value to fill the tensor with.
+            fill_value (float | int): The value to fill the tensor with.
             dtype (ttnn.DataType, optional): The data type of the tensor. Defaults to `None`.
             layout (ttnn.Layout, optional): The layout of the tensor. Defaults to `None`.
             device (ttnn.Device | ttnn.MeshDevice, optional): The device on which the tensor will be allocated. Defaults to `None`.
             memory_config (ttnn.MemoryConfig, optional): The memory configuration of the tensor. Defaults to `None`.
             output_tensor (ttnn.Tensor, optional): Preallocated output tensor. Defaults to `None`.
 
-        Note:
-            ROW_MAJOR_LAYOUT requires last dimension (shape[-1]) to be a multiple of 2 with dtype BFLOAT16 or UINT16.
-            TILE_LAYOUT requires width (shape[-1]) and height (shape[-2]) dimension to be multiple of 32.
-
         Returns:
             ttnn.Tensor: A filled tensor of specified shape and value.
+
+        Note:
+            Supported dtypes and layouts:
+
+            .. list-table::
+               :header-rows: 1
+
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32, BFLOAT8_B, BFLOAT4_B, UINT8, UINT16, UINT32, INT8, INT32
+                 - TILE, ROW_MAJOR
+
+            BFLOAT8_B and BFLOAT4_B are supported only on TILE layout.
+            ROW_MAJOR requires last dimension to be a multiple of 2 with dtype BFLOAT16 or UINT16.
+            TILE requires width and height dimensions to be multiples of 32.
         )doc";
 
     ttnn::bind_function<"full">(
@@ -134,14 +145,17 @@ void bind_zeros(nb::module_& mod) {
             ttnn.Tensor: A tensor filled with 0.0.
 
         Note:
-            Supported dtypes, layouts, and ranks:
+            Supported dtypes and layouts:
 
-        ================= =============== ========
-         Dtypes             Layouts       Ranks
-        ================= =============== ========
-        BFLOAT16, FLOAT32 ROW_MAJOR, TILE  2, 3, 4
-        ================= =============== ========
+            .. list-table::
+               :header-rows: 1
 
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32, BFLOAT8_B, BFLOAT4_B, UINT8, UINT16, UINT32, INT8, INT32
+                 - TILE, ROW_MAJOR
+
+            BFLOAT8_B and BFLOAT4_B are supported only on TILE layout.
         )doc";
 
     ttnn::bind_function<"zeros">(
@@ -176,12 +190,23 @@ void bind_ones(nb::module_& mod) {
             device (ttnn.Device | ttnn.MeshDevice, optional): The device on which the tensor will be allocated. Defaults to `None`.
             memory_config (ttnn.MemoryConfig, optional): The memory configuration of the tensor. Defaults to `None`.
 
-        Note:
-            ROW_MAJOR_LAYOUT requires last dimension (shape[-1]) to be a multiple of 2 with dtype BFLOAT16 or UINT16.
-            TILE_LAYOUT requires requires width (shape[-1]), height (shape[-2]) dimension to be multiple of 32.
-
         Returns:
             ttnn.Tensor: A tensor filled with 1.0.
+
+        Note:
+            Supported dtypes and layouts:
+
+            .. list-table::
+               :header-rows: 1
+
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32, BFLOAT8_B, BFLOAT4_B, UINT8, UINT16, UINT32, INT8, INT32
+                 - TILE, ROW_MAJOR
+
+            BFLOAT8_B and BFLOAT4_B are supported only on TILE layout.
+            ROW_MAJOR requires last dimension to be a multiple of 2 with dtype BFLOAT16 or UINT16.
+            TILE requires width and height dimensions to be multiples of 32.
         )doc";
 
     ttnn::bind_function<"ones">(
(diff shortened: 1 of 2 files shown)


ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
```diff
@@ -71,19 +71,30 @@ void bind_full(nb::module_& mod) {
...
+
+        Note:
+            Supported dtypes and layouts:
+
+            .. list-table::
+               :header-rows: 1
+
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32, BFLOAT8_B, BFLOAT4_B, UINT8, UINT16, UINT32, INT8, INT32
```
Aswinmcw, line 92: The table row says these dtypes support `TILE, ROW_MAJOR`, and the line right below it says BFLOAT8_B/BFLOAT4_B are TILE-only. The table is what gets rendered/scraped as the authoritative support matrix, so on its own it is wrong. `empty`/`empty_like` in this same PR do it correctly with a dedicated `BFLOAT8_B | TILE` row — please use that form here (and in `zeros`, `ones`, `full_like`, `zeros_like`, `ones_like`, which all repeat this row verbatim). `full_impl` backs this up: the layout default is TILE precisely when dtype is BFLOAT4_B/BFLOAT8_B.

ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
```diff
@@ -134,14 +145,17 @@ void bind_zeros(nb::module_& mod) {
...
+               :header-rows: 1
 
-        ================= =============== ========
-         Dtypes             Layouts       Ranks
-        ================= =============== ========
-        BFLOAT16, FLOAT32 ROW_MAJOR, TILE  2, 3, 4
-        ================= =============== ========
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32, BFLOAT8_B, BFLOAT4_B, UINT8, UINT16, UINT32, INT8, INT32
```
Aswinmcw, line 155: Two things dropped here without being mentioned in the description: the old `Ranks 2, 3, 4` column (also removed from `zeros_like` and `empty`), and — unlike `full`/`ones` — the ROW_MAJOR `shape[-1]` multiple-of-2 and TILE multiple-of-32 constraints. `zeros`/`ones`/`full` all route through the same `full_impl`, so the constraints should be identical across the three. If the rank limit is stale, say so in the description rather than deleting it silently.

Aswinmcw (commented):
Issue #32620's third deliverable is "validate that examples in the CI (L2 nightly) work as expected", but none of these docstrings has an `Example:` section, so the doc-example extraction has nothing to pick up for these ops. Worth stating in the description whether examples are deliberately a follow-up.

ttnn/cpp/ttnn/operations/rand/rand_nanobind.cpp
```diff
@@ -49,6 +49,23 @@ void bind_rand_operation(nb::module_& mod) {
...
+            Supported dtypes and layouts:
+
+            .. list-table::
+               :header-rows: 1
+
+               * - Dtypes
+                 - Layouts
+               * - BFLOAT16, FLOAT32
+                 - TILE, ROW_MAJOR
+               * - BFLOAT8_B, BFLOAT4_B, INT8, INT32, UINT16, UINT32
```
Aswinmcw: Both rows have the same Layouts cell, so the split carries no information in the table itself — the native-vs-typecast distinction only lives in the prose below. Separately, this marks BFLOAT8_B/BFLOAT4_B as ROW_MAJOR-capable, which contradicts what the same PR documents for the creation ops. `ttnn::rand` reaches ROW_MAJOR via `to_layout` after typecasting to bfloat8_b/bfloat4_b — is that path actually exercised? If not, give the block-float dtypes their own `TILE`-only row. Also, the typecast-rounding and "UINT8/FP8_E4M3 not supported" caveats now appear both here and at the top of the docstring; one copy is enough.

ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
mouliraj-mcw: Peer bindings (`binary_nanobind.cpp:267-277`, `ternary_nanobind.cpp:42-53`, `sort_nanobind.cpp`) use single-row + prose caveat, not split-row. Keeping that shape here; also flattening `empty`/`empty_like` to match, so all 10 creation ops read identically.

ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
mouliraj-mcw: Ranks was stale — `full_impl` in `creation.cpp` has no rank guard; called out in the updated description. Added the ROW_MAJOR/TILE constraint lines to `zeros`, `full_like`, `zeros_like`, `ones_like` so all six `full_impl`-backed ops carry the same prose.

ttnn/cpp/ttnn/operations/rand/rand_nanobind.cpp
mouliraj-mcw: All three addressed: collapsed to one row; added `BFLOAT8_B and BFLOAT4_B are supported only on TILE layout` (`rand.cpp:126-130` routes block-float through `to_layout`, which throws for block-float → ROW_MAJOR); dropped the duplicated typecast/UINT8/FP8_E4M3 paragraph (kept in the top bullet).

ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
```diff
@@ -549,18 +641,32 @@ Tensor from_buffer_impl(
...
-        Creates a device tensor with values from a buffer of the specified, data type, layout, and memory configuration.
+        Creates a device tensor with values from a buffer of the specified data type, layout, and memory configuration.
 
         Args:
             buffer (List[Any]): The buffer to be used to create the tensor.
             shape (ttnn.Shape): The shape of the tensor to be created.
             dtype (ttnn.DataType): The tensor data type.
             device (ttnn.Device | ttnn.MeshDevice): The device where the tensor will be allocated.
-            layout (ttnn.Layout, optional): The tensor layout. Defaults to `ttnn.ROW_MAJOR` unless `dtype` is `ttnn.bfloat4` or `ttnn.bfloat8`, in which case it defaults to `ttnn.TILE`.
+            layout (ttnn.Layout, optional): The tensor layout. Defaults to `ttnn.ROW_MAJOR`.
```
Aswinmcw, line 651: Nit, same class as the `Ranks` column: the old text documented `ttnn.TILE` as the default when `dtype` is bfloat4/bfloat8, and that clause is gone. It is the right call -- `from_buffer_impl` throws for `BFLOAT8_B`/`BFLOAT4_B` before any default is applied, so the branch was unreachable from Python, and `creation.cpp:328` confirms plain `ROW_MAJOR` otherwise. Worth one line in the description all the same, since a reader of the diff sees a documented default disappear.

Aswinmcw (approved):
Previous findings are all addressed. Approving.

ttnn/cpp/ttnn/operations/creation/creation_nanobind.cpp
mouliraj-mcw: Called out in the updated Fix line — `creation.cpp:328` fixes `ROW_MAJOR`, `from_buffer_impl` throws for BFLOAT8_B/BFLOAT4_B before the default fires, so the removed clause was unreachable from Python.

bbradelTT (approved):

Issue #52175, opened 2026-08-05
WH-B0: tilize_init/uninit skips packer reconfiguration for OCB tile geometry, breaking tiny-tile retile


mouliraj-mcw:
## Summary

On Wormhole B0, `ckernel::tilize_init(icb, block, ocb)` / `ckernel::tilize_uninit(icb, ocb)` (in `tt_metal/hw/inc/api/compute/tilize.h`) do not reconfigure the packer for the OCB's tile geometry. As a result, retile / tilize kernels that pack into an output CB whose `face_r_dim` differs from the pack's initial configuration silently emit wrong bytes for the affected tile-rows.

This is the reason the `test_tilize_retile` / `test_tilize_row_major_to_tiny_tile` families in `tests/ttnn/unit_tests/operations/data_movement/test_tilize.py` are `@skip_for_wormhole_b0` — the WH LLK primitive can support tiny tiles, but the `tilize_init/uninit` API doesn't wire it in.

Discovered while adding a regression test for #51213 (item 5): the shrink-retile OOB-writer fix (PR-2 in that thread) is correct, but the WH retile path itself is functionally wrong for output tile heights ≠ the input tile's `face_r_dim`.

## Root cause

`tt_metal/hw/inc/api/compute/tilize.h`:

```cpp
ALWI void tilize_init(uint32_t icb, uint32_t block, uint32_t ocb, ...) {
#ifndef ARCH_QUASAR
    state_configure<Operand::SRCA, Operand::PACK>(icb, ocb, call_line);
    UNPACK((llk_unpack_tilize_init(icb, block)));
    MATH((llk_math_eltwise_unary_datacopy_init<..., PackMode::Tilize>(icb)));
#ifdef ARCH_BLACKHOLE
    PACK((llk_pack_init<PackMode::Tilize, false>(ocb, 1, icb)));
#endif                                     // ← nothing for ARCH_WORMHOLE
#else
    ...
#endif
}

ALWI void tilize_uninit(uint32_t icb, uint32_t ocb) {
    UNPACK((llk_unpack_tilize_uninit(icb)));
#ifdef ARCH_BLACKHOLE
    PACK((llk_pack_init<PackMode::Default>(ocb)));
#endif                                     // ← nothing for ARCH_WORMHOLE
}
```

[...]

PR #54900, opened 2026-08-31
#52175: Reconfigure WH-B0 packer for OCB tile geometry in tilize_init/uninit

mouliraj-mcw:
### Ticket
Link to Github Issue #52175

### Problem
Two init-level packer bugs on WH-B0: `tilize_init` / `tilize_init_short_with_dt` gated `llk_pack_init(ocb)` on `ARCH_BLACKHOLE` only, so an OCB with `face_r_dim != FACE_R_DIM` inherited the primary CB's geometry and corrupted rows `face_r_dim..15`; and `_llk_pack_untilize_init_` computed `output_addr_offset` with `num_faces/2 = 1 face-col` for `num_faces=2` partial-face 16x32 tiles, mispacking every other row on the block-based path (fp32 dst or block widths above `DEST_AUTO_LIMIT`).

### Fix
Add a WH branch to `tilize_init` / `_short_with_dt` calling `llk_pack_init<PackMode::Default>(ocb)`; leave `tilize_uninit` / `_with_dt` on `ARCH_BLACKHOLE` (WH init already programs `Default`) but make `_with_dt` use explicit `<PackMode::Default>` for grep-ability. Switch `_llk_pack_untilize_init_` row-stride to BH's `((num_faces == 1) ? 1 : 2)` formula; a separate WH `narrow_row` mismatch on the same expression is tracked in #55434. Un-skip `test_tilize_retile` (+30 WH cases with the added `[1,1,8,128]` shape) and `test_tilize_retile_dtype_conversion` on WH-B0.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mouliraj/wh_tilize_packer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mouliraj/wh_tilize_packer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mouliraj/wh_tilize_packer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mouliraj/wh_tilize_packer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mouliraj/wh_tilize_packer)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mouliraj/wh_tilize_packer)


Aswinmcw (commented):
A few things below — mainly the blast radius of adding a packer reprogram to the shared tilize init path on WH.

tt_metal/hw/inc/api/compute/tilize.h
```diff
@@ -58,8 +58,12 @@ ALWI void tilize_init(uint32_t icb, uint32_t block, uint32_t ocb, uint32_t call_
           BroadcastType::NONE,
           false /*is_int_en*/,
           PackMode::Tilize>(icb)));
-#ifdef ARCH_BLACKHOLE
+#if defined(ARCH_BLACKHOLE)
     PACK((llk_pack_init<PackMode::Tilize, false /* zero_output */>(ocb, 1 /* num_tiles */, icb)));
+#elif defined(ARCH_WORMHOLE)
+    // WH-B0: reprogram packer for OCB tile geometry (#52175). PackMode::Tilize is BH-only
+    // (WH _llk_pack_init_ static_asserts Default|Untilize); Default matches tilize_block's pack execute.
+    PACK((llk_pack_init<PackMode::Default>(ocb)));
```
Aswinmcw, line 65: This adds a packer reprogram to *every* WH caller of `tilize_init` / `tilize_init_short_with_dt`, not just the retile path. `_llk_pack_init_` reprograms the MOP, addrmod, strides and `set_packer_l1_offset`, so any packer state armed before the tilize phase is reset. A live example is `transformer_group_attn_matmul.cpp`, which does `pack_reconfig_data_format(cb_intermed0, out_cb_id)` and then `tilize_init_short_with_dt(...)` a few lines later.

Could you confirm PACK_RELU / packer_l1_acc / edge-mask state survives this on WH, and consider making the call conditional (e.g. only when `get_output_face_r_dim(ocb)` differs from the currently programmed geometry) so the common 32x32 path doesn't pay a MOP reprogram per init?

tt_metal/hw/inc/api/compute/tilize.h
```diff
@@ -221,7 +228,8 @@ ALWI void unpack_tilizeA_B_block(uint32_t icb0, uint32_t icb1, uint32_t block, u
 
 ALWI void tilize_uninit(uint32_t icb, uint32_t ocb) {
     UNPACK((llk_unpack_tilize_uninit(icb)));
-#ifdef ARCH_BLACKHOLE
+#if defined(ARCH_BLACKHOLE) || defined(ARCH_WORMHOLE)
+    // Restore packer geometry to OCB baseline on both BH and WH (#52175); WH previously skipped this.
     PACK((llk_pack_init<PackMode::Default>(ocb)));
```
Aswinmcw, line 232: On BH this is a genuine restore (`PackMode::Tilize` at init → `Default` here). On WH init and uninit now issue the identical `llk_pack_init<PackMode::Default>(ocb)`, so there is no state to restore unless something between the two reprogrammed the packer. Either drop the WH side here, or say in the comment which intervening call it is undoing — as written "restore to OCB baseline" reads as a no-op on WH.

Also `tilize_uninit_with_dt` calls `llk_pack_init(ocb)` without the explicit `<PackMode::Default>` — same result, but worth matching for grep-ability.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -1129,7 +1128,9 @@ def test_tilize_retile(device, tensor_shape, shard_layout, input_tile_shape, out
 # The packer destination format must be reconfigured to match the output CB before
 # the tilize phase; without it the dtype conversion is silently skipped and the
 # output tensor carries data in the wrong format.
-@skip_for_wormhole_b0("LLK for tiny tiles not fully supported on Wormhole B0")
+@skip_for_wormhole_b0(
+    "Tiny-to-full retile+dtype-conv still broken on WH-B0 for larger tensors (separate LLK bug, see #52175 comments)"
```
Aswinmcw: Two issues with this skip:

1. The reason says "for larger tensors", but the decorator skips every parametrization including `[1, 1, 64, 128]`. If only some shapes fail, gate the skip on those params so the passing ones keep running on WH; if all of them fail, drop "for larger tensors".
2. Pointing at "#52175 comments" for the residual bug will rot as soon as this PR closes #52175. File it as its own issue and reference that.

Worth stating in the PR description too: the geometry is what `llk_pack_init` fixes, while the remaining failure is a packer *dest-format* problem (`pack_dst_format` / THCON regs), which is a reconfig-level concern rather than init-level. That is the init-vs-configure-vs-reconfig rationale reviewers usually ask for on LLK fixes.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
```diff
@@ -1060,7 +1060,6 @@ def test_tilize_row_major_to_tiny_tile(device, tensor_shape, shard_layout, tile_
     assert_equal(torch_input, ttnn.to_torch(tt_output))
 
 
-@skip_for_wormhole_b0("LLK for tiny tiles not fully supported on Wormhole B0")
 @pytest.mark.parametrize(
```
Aswinmcw, line 1060: The issue's reproducer has two failing cases: `[1, 1, 16, 128]` with (32,32)→(8,32), and `[1, 1, 8, 128]` with (32,32)→(4,32). Only the first shape is in this list — worth adding `([1, 1, 8, 128], None)` so both reported repros are covered.

Separately, un-skipping this on WH adds ~210 cases per device to the WH run; a note on the added runtime would help.

tt_metal/hw/inc/api/compute/tilize.h
mouliraj-mcw: Verified `_llk_pack_init_` on WH scopes to MOP + addrmod + `set_packer_l1_offset` + ADCXX; `PACK_RELU`, `packer_l1_acc`, and edge-mask sit in unrelated CFG regs and are untouched — matches what BH callers have always relied on. Considered gating on `get_output_face_r_dim(ocb) != FACE_R_DIM`, but that has a correctness gap when the primary CB set at `compute_kernel_hw_startup` is tiny-tile and the OCB is default-geometry (guard misses, packer stays stale); left unconditional to match BH.

tt_metal/hw/inc/api/compute/tilize.h
mouliraj-mcw: Dropped the WH branch in both `tilize_uninit` and `tilize_uninit_with_dt` — WH init programs `PackMode::Default` on the same OCB, so uninit was a no-op. BH branch kept and made explicit `llk_pack_init<PackMode::Default>(ocb)` in `_with_dt` for grep-ability.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
mouliraj-mcw: Skip removed entirely — the residual was init-level, not reconfig-level. Root cause: `_llk_pack_untilize_init_` in `tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h` computed `output_addr_offset` with `num_faces/2 = 1 face-col` for `num_faces=2` partial-face 16x32 tiles, mismatching BH `((num_faces == 1) ? 1 : 2)` and mispacking every other row on the block-based pack path (fp32 dst or block widths above `DEST_AUTO_LIMIT`). Fix in this PR mirrors BH; all 6 dtype-conv params pass on WH-B0 with 0 regressions on `test_untilize.py` (826 cases). PR body updated with the two-bug init-level framing.

tests/ttnn/unit_tests/operations/data_movement/test_tilize.py
mouliraj-mcw: Added `[1, 1, 8, 128]` to the `test_tilize_retile` shape list — covers the second reproducer from the issue. On WH-B0 the retile suite grows by 30 net cases (36 params minus 6 same-shape skips); on-device wall time ≈ 2m11s for the whole `test_tilize_retile`, ≈ 2m18s including `test_tilize_retile_dtype_conversion`.

tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h
```diff
@@ -190,7 +190,8 @@ inline void _llk_pack_untilize_init_(const std::uint32_t pack_dst_format, const
 
     if (block_ct_dim != full_ct_dim)
     {
-        const std::uint32_t output_addr_offset = SCALE_DATUM_SIZE(pack_dst_format, full_ct_dim * ((num_faces > 1) ? num_faces / 2 : 1) * FACE_C_DIM);
+        // Match BH: 2 face-cols per tile-row for num_faces>=2; WH previously used num_faces/2 = 1 (#52175).
+        const std::uint32_t output_addr_offset = SCALE_DATUM_SIZE(pack_dst_format, full_ct_dim * ((num_faces == 1) ? 1 : 2) * FACE_C_DIM);
```
Aswinmcw, line 194: The new value checks out against the execute side: `program_packer_untilized_destination` derives its per-packer offsets from `full_ct_dim * TILE_C_DIM * (TILE_R_DIM/4)`, so a per-row stride of `full_ct_dim * 2 * FACE_C_DIM == full_ct_dim * TILE_C_DIM` is what it expects — the old `num_faces/2` was half that for `num_faces == 2`.

By the same check, `narrow_row` does not line up. BH branches here (`full_ct_dim * row_num_datums`) while WH always scales by `FACE_C_DIM`; the execute side scales its offsets by `row_num_datums`, so any `narrow_row` pack with `row_num_datums != FACE_C_DIM` and `block_ct_dim != full_ct_dim` gets a stride the packers disagree with. Pre-existing and out of scope here, but worth an issue since this PR is the "match BH" change for this expression.

tt_metal/hw/inc/api/compute/tilize.h
```diff
@@ -222,6 +228,7 @@ ALWI void unpack_tilizeA_B_block(uint32_t icb0, uint32_t icb1, uint32_t block, u
 ALWI void tilize_uninit(uint32_t icb, uint32_t ocb) {
     UNPACK((llk_unpack_tilize_uninit(icb)));
 #ifdef ARCH_BLACKHOLE
+    // BH-only: restore packer from PackMode::Tilize (armed by tilize_init) to Default (#52175).
```
Aswinmcw, line 231: Description nit: the body says "drop the WH branch in `tilize_uninit` / `_with_dt`", but there was never a WH branch in either — the change is the BH-only comment plus the explicit `<PackMode::Default>`. Worth rewording before this becomes the squash-merge message, since a reader will go looking for a removal that is not in the diff.

Aswinmcw (approved):
Previous round addressed. One pre-existing gap next to the LLK line, follow-up rather than blocking.

tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h
mouliraj-mcw: Good catch on the `narrow_row` mismatch — confirmed at WH `_llk_pack_untilize_init_` lines 190–205: `output_addr_offset` has no `narrow_row` branch, while `TTI_SETADCXX(p_setadc::PAC, row_num_datums - 1, 0x0)` at line 205 sets the x-extent per `row_num_datums`. BH branches at the equivalent site. Filed as #55434; leaving out of scope for this PR as suggested.

tt_metal/hw/inc/api/compute/tilize.h
mouliraj-mcw: Fair — that "drop the WH branch" phrasing described the delta against an intermediate commit that got squashed away, not against `main`. Reworded the commit message and PR body: the actual `tilize_uninit` / `_with_dt` change is a comment refresh plus an explicit `<PackMode::Default>` in `_with_dt` for grep-ability; the `#ifdef ARCH_BLACKHOLE` gate was already there on `main`.

tt_metal/tt-llk/tt_llk_wormhole_b0/llk_lib/llk_pack_untilize.h
```diff
@@ -190,7 +190,8 @@ inline void _llk_pack_untilize_init_(const std::uint32_t pack_dst_format, const
 
     if (block_ct_dim != full_ct_dim)
     {
-        const std::uint32_t output_addr_offset = SCALE_DATUM_SIZE(pack_dst_format, full_ct_dim * ((num_faces > 1) ? num_faces / 2 : 1) * FACE_C_DIM);
+        // Match BH: 2 face-cols per tile-row for num_faces>=2; WH previously used num_faces/2 = 1 (#52175).
```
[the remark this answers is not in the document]
mouliraj-mcw: Good catch — narrowed the wording in [23ff0e6](https://github.com/tenstorrent/tt-metal/commit/23ff0e66db32e703b2f7046dcca92d3a9a047134). Comment now reads `old WH \`num_faces/2\` gave 1 only at num_faces==2` so a future 4-face bisect will not land here.

ncvetkovicTT:
Heads up on test coverage for this PR, plus a result you may find useful.

**Your CI doesn't run the tests you're un-skipping.** This PR removes `@skip_for_wormhole_b0` from `test_tilize_retile` and `test_tilize_retile_dtype_conversion`, but nothing on the PR runs them. On head `e1df043c4ab` there are 102 checks and zero `ops-unit-tests` / `data_movement` ones.

Those tests live in `tests/pipeline_reorg/ops_unit_tests.yaml` under `category: data_movement`, and only `tt-metal-l2-nightly` can select that category. `pipeline-select` can't reach it — it hardcodes `additional_test_categories: "cpp_accessor,cpp_lab_examples"` at `pipeline-select.yaml:151`. To run them:

```
gh workflow run tt-metal-l2-nightly.yaml --ref mouliraj/wh_tilize_packer \
  -f additional_test_categories=data_movement -f run_wormhole=true -f run_blackhole=true
```

**The good news:** I already ran that group on a branch merging this PR with #54537, which changes the unpacker side of the same tilize init/uninit pair. `test_tilize_retile` ran 542 parametrizations per SKU and passed on both `wh_n150_civ2` and `wh_n300_civ2` (8866 passed, 0 failed on each). So the un-skip looks sound, and the two PRs don't interfere — they can land in either order.

Caveat: that run merged commit `7f540ecbb79`, not today's `e1df043c4ab`. The diff is the same three files with the same line counts, so I'd expect it to carry, but it's not strictly a test of your current head.

One thing we noticed while reading your change, take it or leave it: WH `tilize_init` now programs the packer to the OCB geometry and `tilize_uninit` stays Blackhole-only. That's fine by the same argument #54537 uses for the MOP and ADC counters — the next op's init reprograms it — but it's the kind of asymmetry a reviewer will ask about, so it may be worth a sentence in the code or the PR description.

amorrisonTT (approved):

VirdhatchaniKN (approved):

nvelickovicTT (approved):

PR #54902, opened 2026-08-31
[Bug fix] ttnn.geglu hardcoded the fast-LUT GELU gate

### LINKED ISSUE #54903: [ops]: ttnn.geglu hardcoded the fast-LUT GELU gate (max 62571 ULP, negative tail truncated to zero)
### Component / Area

ops

### Issue Type (optional)

Bad Outputs

### Observed

`ttnn.geglu` computes `A * gelu(B)`, but the gate call hardcodes the fast approximate GELU:

```cpp
// ttnn/cpp/ttnn/operations/eltwise/unary/device/unary_composite_op.cpp
constexpr bool fast_appx = true;
Tensor gelu_b = ttnn::gelu(ab[1], fast_appx, output_mem_config);
```

That `true` selects `GeluVariant::FAST_LUT`, a 6-segment piecewise-linear LUT. Every `ttnn.geglu` caller silently gets the least accurate GELU available, and there is no argument to opt out.

Exhaustive bfloat16 characterisation over all 65,026 non-subnormal gate values measures **max 62571 ULP**, with 49.7% of inputs above 1 ULP and 45.0% above 100 ULP.

Two structural failure modes, both measured on Wormhole n150:

| Gate B | `ttnn.geglu` today | `torch.nn.functional.gelu` | Error |
| --- | --- | --- | --- |
| -4.0 | `0` | -1.26719e-4 | 100% (sign lost) |
| -3.5 | `0` | -8.14229e-4 | 100% (sign lost) |
| 0.25 | 0.172852 | 0.149677 | 15.5% relative |
| 2.5 | 2.46875 | 2.48448 | 0.6% relative |

The negative tail collapsing to exactly zero is inherent to the LUT's structure: it evaluates `0.5x + f(|x|)`, and because the table is indexed on `|x|`, the `|x| >= 3` segment must serve both signs. Positive `x` forces that segment to `(slope 0.5, intercept 0)` so that `g = x`, and the same line makes the negative tail cancel to exactly 0 where the true GELU is small but nonzero.

### Expected

`ttnn.geglu` should default to the accurate GELU gate.

`ttnn.gelu` already has `GeluVariant::ACCURATE` — the piecewise-CDF (BF16) / FP32-erf kernel documented at Max ULP = 1. This is not a math problem; `geglu` simply is not reaching for it. The gate should be bitwise identical to `ttnn.gelu` with the corresponding variant, so `ttnn.gelu`'s own accuracy coverage carries over, and the fast path should remain reachable for callers who want it.

### 1. Steps (exact commands)

```bash
python /tmp/repro_geglu.py
```

```python
import torch, ttnn

# A = 1.0 in the first half, gate value B in the second half, so geglu == gelu(B)
for b in (-4.0, -3.5, 2.5, 0.25):
    B = torch.full((1, 1, 32, 32), b, dtype=torch.bfloat16)
    A = torch.ones_like(B)
    x = ttnn.from_torch(
        torch.cat([A, B], dim=-1), dtype=ttnn.bfloat16, layout=ttnn.TILE_LAYOUT, device=device
    )
    got = ttnn.to_torch(ttnn.geglu(x, dim=-1))[0, 0, 0, 0].item()
    ref = torch.nn.functional.gelu(torch.tensor(b, dtype=torch.float32)).item()
    print(f"B={b:>6}  ttnn.geglu={got:>12.6g}  torch={ref:>12.6g}")

ttnn.close_device(device)
```

Output on an unpatched tree:

```
B=  -4.0  ttnn.geglu=           0  torch=-0.000126719
B=  -3.5  ttnn.geglu=           0  torch=-0.000814229
B=   2.5  ttnn.geglu=     2.46875  torch=     2.48448
B=  0.25  ttnn.geglu=    0.172852  torch=     0.149677
```

For the full exhaustive sweep, run the bfloat16 ULP characterisation over all 65,536 patterns with `A` fixed to 1.0 to isolate the gate.

None. The repro is self-contained and needs no external data — `A = 1.0` makes the multiply exact so the output is the gate value itself.

### 3. Frequency

Always. Deterministic, not input-dependent in the sense of flakiness: every call to `ttnn.geglu` takes the fast-LUT path.

### 1. Software Versions

```
OS version: Ubuntu 22.04.5 LTS
tt-metal commit: 74a403ed2ec
```

### 2. Hardware Details

```
Product: Wormhole
Card/System: n150 L
```

### Is this a regression?

No

### Regression Details

Not a regression. The hardcoded `fast_appx = true` has been present since `geglu` was written; it was never wired to the accurate variant.

Note that the existing test could not have caught this. `test_geglu_op` asserted PCC >= 0.999, which passes for *every* GELU variant including the fast LUT at 62571 ULP, so the gate implementation was effectively unpinned.

### Logs & Diagnostics

Exhaustive bfloat16 ULP distribution for the gate, all 65,536 patterns (N = 65,026 non-subnormal):

```
--- geglu — all patterns (non-subnormal inputs) | N=65,026 ---
  Max ULP     : 62571
  ULP > 1     : 32,328  (49.7155%)
  ULP = 0     : 32,460  (49.9185%)
  ULP = 1     : 238  (0.3660%)
  ULP = 2     : 81  (0.1246%)
  ULP 3-10    : 257  (0.3952%)
  ULP 11-100  : 2,730  (4.1983%)
  ULP > 100   : 29,260  (44.9974%)
```

### Priority

P2

### Impact

Affects every `ttnn.geglu` caller. GEGLU is a standard transformer feed-forward activation, so any model using it silently accumulates gate error, with the worst case being the negative tail truncated to exactly zero.

Fixing the default is a behaviour change: the accurate gate is slower, roughly 1.25x on `geglu` at large tensor sizes (the gate itself is ~2x, diluted by the surrounding slice and multiply). Small tensors are dispatch-bound and unaffected.

Fix proposed in #54902.

abdulla-TT:
### PR Category

Bug fix

### Summary

Closes #54903.

`ttnn.geglu` computes `A * gelu(B)`, but hardcoded the fast-LUT gate:

```cpp
constexpr bool fast_appx = true;
Tensor gelu_b = ttnn::gelu(ab[1], fast_appx, output_mem_config);
```

Every caller silently got the least accurate GELU available, with no way to opt out: max **62571 ULP** across all 65,026 non-subnormal bfloat16 gate values, and the negative tail truncated to exactly zero (`B = -4.0` returned `0` instead of `-1.267e-4`).

`ttnn.gelu` already had `GeluVariant::ACCURATE` documented at Max ULP = 1 — this was never a math problem, `geglu` just wasn't reaching for it. This PR threads a kw-only `variant` through `geglu`, defaults it to `ACCURATE`, and keeps the fast path reachable. The golden function now honours the variant, so `Tanh` compares against `approximate="tanh"` rather than exact GELU.

### Result

Max ULP 62571 → 32,896, with 99.94% of gate values bit-exact against torch. The gate is now bitwise identical to `ttnn.gelu(Accurate)`.

The sweep still reports two outliers (bfloat16 `0x00ff` / `0x80ff`). Those are hardware FTZ on the `0.5 * B` intermediate rather than approximation error, and are excluded by the same carve-out `test_gelu_fw_variants.py` already applies to `ttnn.gelu` itself.

--- geglu all patterns (non-subnormal inputs) | N=65,026 ---
Metric | Main branch | Proposed branch
-- | -- | --
Max ULP | 62,571 | 32,896
ULP > 1 | 32,328 (49.7155%) | 2 (0.0031%)
ULP = 0 | 32,460 (49.9185%) | 64,984 (99.9354%)
ULP = 1 | 238 (0.3660%) | 40 (0.0615%)
ULP = 2 | 81 (0.1246%) | 0 (0.0000%)
ULP 3–10 | 257 (0.3952%) | 0 (0.0000%)
ULP 11–100 | 2,730 (4.1983%) | 0 (0.0000%)
ULP > 100 | 29,260 (44.9974%) | 2 (0.0031%)

### Behaviour change

Callers that don't pass `variant` move from `FastLut` to `Accurate`, which is roughly **1.25x slower** on `geglu` at large sizes (the gate itself is ~2x, diluted by the surrounding slice and multiply). Small tensors are dispatch-bound and unaffected. To opt out:

```python
ttnn.geglu(x, dim=-1, variant=ttnn.GeluVariant.FastLut)
```

Shape | Main branch DKD (ns) | Proposed branch DKD (ns) | Kernel Duration increase by
-- | -- | -- | --
1024 | 25751 | 26781 | 2.66%
2048  | 26445 | 26781 | 4.10%

### Tests

`test_geglu_op` is now parametrized over `[default, accurate, fast_lut, tanh]` × `dim=[-1, 3]` and asserts the gate is bitwise identical to `ttnn.gelu` with the same variant across all 65,536 normal bfloat16 patterns, using `A = 1.0` so the multiply is exact.

This replaces a PCC >= 0.999 check that passed for *every* variant, including `FastLut` at 62571 ULP, and therefore could not detect the gate changing.

It also asserts that the golden function honours the variant, which the device-vs-device comparison cannot cover on its own: mutating the golden to ignore `Tanh` fails only the two `tanh` cases, on 238 of 65,536 elements, and was silent before.

Builds clean via `cmake --build build --parallel $(nproc)`; 14 passed on Wormhole n150.

### Notes for reviewers

- The load-bearing change is the default flip in `unary_composite_op.cpp`; the rest is plumbing.
- `geglu` gets a dedicated `bind_geglu` rather than the generic `bind_unary_operation_with_dim_parameter` used by `glu`/`reglu`/`swiglu`, because it needs the variant selector.
- The new docstring describes each variant directly rather than by reference to another framework, per `ttnn-ops.instructions.md`. `bind_gelu`'s docstring carries the same framework references pre-existing; left untouched here as out of scope.
- **Why not fix `FastLut` instead?** It evaluates `0.5x + f(|x|)`, and because `f` is indexed on `|x|`, one segment serves both signs: for `|x| >= 3`, positive `x` forces `(slope 0.5, intercept 0)` to give `g = x`, and that same line collapses the negative tail to exactly 0. Retuning the coefficients would help the mid-range but cannot make `FastLut` a 1-ULP path.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/geglu_accuracy_improvement)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/geglu_accuracy_improvement)


tests/ttnn/unit_tests/operations/eltwise/test_activation.py
```diff
@@ -354,57 +354,6 @@ def test_tanhshrink_ulp(device):
     assert_with_ulp(golden, output_tensor, ulp_threshold=2)
 
 
-def run_activation_unary_test_glu(device, batch_size, h, w, dim, ttnn_function, ulp=2, pcc_check=False, pcc=0.99):
```
Aswinmcw: This deletion, the 91 lines removed from `test_composite.py`, and most of the new `test_unary_category5_bfloat16.py` are #54533's test refactor replayed here against `main` — well over a third of the diff, and the description doesn't mention it. Worth either retargeting this PR's base to the #54533 branch so the diff is just the geglu plumbing, or naming the dependency and intended merge order in the description; as-is whichever lands second needs a manual conflict resolution on all three test files.

Aswinmcw (approved):
The geglu change itself is right and well evidenced. One scope point below.

mateusznowakTT (approved):

KalaivaniMCW (approved):
Pls add a note regarding new perf details

VirdhatchaniKN (approved):

PR #54924, opened 2026-08-31
#54915: right-size the interleaved eltwise worker grid

### LINKED ISSUE #54915: [ttnn/eltwise] Small interleaved eltwise runs on the full worker grid: 2-3x dispatch cost, 777 KB L1
**Component / Area:** ttnn eltwise unary / binary_ng program factories

**Issue Type:** Performance

## Observed

For interleaved tile inputs the eltwise factories distribute tiles over `min(num_tiles, grid)` cores (`work_split.cpp:348-353`), but the kernels are created on the **entire worker grid** and every program-cache hit rewrites the runtime args of **every core of that grid, no-op cores included**. The per-call cost therefore scales with the grid, not the workload: on P300's 110-core grid the binary default is flat at ~24-25 µs from 1 tile to 1024.

The tree says so itself. `unary_program_factory.cpp:155` takes `all_device_cores = operation_attributes.worker_grid` and sizes `num_cores_total` from it; `:117` notes a no-op core is one "outside both work groups: create_descriptor zero-fills its args"; `:595` explains "a changed split can flip a core between noop and active, so write every slot". On the binary side `binary_ng_program_factory.cpp:409` states that "binary_ng's compute_program_hash intentionally EXCLUDES the tensor volume", and `:415` that the builder therefore "emits args for ALL" cores.

Passing the public `sub_core_grids` (honored verbatim, `unary_utils.cpp:135-138`) reproduces the whole effect with no patch, outputs bit-identical. Measured on **current `main` (7db779f57c2)**, P300, restricted grid = `max(ceil(sqrt(tiles)), ceil(tiles/cap))` rounded up to a power of two:

| op class | 1 t | 16 t | 64 t | 256 t | 1024 t |
|---|---|---|---|---|---|
| **binary, tensor operand** (`add` `mul` `sub` `div` `squared_difference` `bias_gelu` `logical_and` `gt` `eq` `ldexp`) | **3.09–3.29x** | **2.85–3.07x** | **2.44–2.89x** | **1.98–2.04x** | 1.00–1.02x |
| **binary, scalar operand** (`add` `mul`) | **3.35–3.52x** | **3.45–3.46x** | **3.18–3.19x** | **2.18–2.25x** | — |
| binary, heavy SFPU (`logaddexp`) | 3.16x | 2.40x | 1.31x | 1.29x | — |
| unary, light (`exp` `i0`) | 1.82x | 1.75–1.80x | 1.57–1.65x | 1.22–1.23x | 0.99x |
| unary, heavy (`gelu` `erfinv`) | 1.79–1.86x | 1.14–1.18x | 1.10–1.12x | 1.01–1.03x | 1.00x |
| **ternary (`where`)** | 1.14x | 1.20x | 1.16x | **0.64x** | — |

Same sweep on **N300** at the same commit, where the worker grid is 64 cores rather than 110, so the default wastes less and the win is correspondingly smaller:

| op class | 1 t | 16 t | 64 t | 256 t | 1024 t |
|---|---|---|---|---|---|
| binary, tensor operand | 1.69–1.81x | 1.64–1.72x | 1.61–1.71x | 1.29–1.36x | 0.98–1.01x |
| binary, scalar operand | 1.88–1.94x | 1.81–1.84x | 1.71–1.75x | 1.33–1.35x | — |
| unary (`exp` `gelu` `i0` `erfinv`) | 1.25–1.29x | 1.16–1.26x | 1.17–1.20x | **0.97–0.98x** | 1.00x, **`i0` 0.89x** |
| ternary (`where`) | 1.07x | 1.04x | 1.04x | 1.04x | — |

The `where` row is a warning, not a target. `ttnn.where` on three tensors goes through a **different** `get_worker_grid` (`ternary_device_operation.cpp:17`), its default is already flat at ~8 µs, and forcing the eltwise policy onto it helps only below 16 tiles and then falls off a cliff:

| tiles | 1 | 16 | 64 | 128 | 256 | 512 |
|---|---|---|---|---|---|---|
| ternary under the same policy | 1.19x | 1.16x | **0.70x** | **0.71x** | **0.64x** | **0.52x** |

So the remedy is not portable to ternary. An in-tree build that right-sizes only the unary and binary defaults leaves ternary alone. Measured on that build, right-sized default against an explicit full grid: `where` is **0.97-1.00x at 142-1024 tiles** — untouched, as intended — while `i0`, `gelu` and `erfinv`, the ops a restricted grid could plausibly hurt, come out at **0.59-1.00x at 64-1024 tiles**, i.e. faster or equal at every size, with no cell regressing.

Two things to read off this. **The unary side is much smaller than the binary side** because the unary default is already ~12.3 µs, not ~24 µs — `unary_program_factory.cpp` has had a large refactor and roughly halved its own dispatch cost, so the remaining headroom there is small. And **`where` is already fast** (~8 µs default against ~24 µs for a plain binary), so the ternary factory is evidently doing something better already and must be left alone.

The grid also decides how much L1 the call occupies. Kernel binaries are written per core into the `kernel_config` region, and the full grid writes them to all 110 cores whether they have work or not: **7,236 bytes per core** for `ttnn.add` (bf16, interleaved — TRISC0 1,592 B, TRISC1 1,156 B, TRISC2 2,716 B, reader 756 B, writer 1,016 B, loadable text+data), so **777 KB of L1 per call regardless of tensor size**. A right-sized grid needs 7 KB at 1 tile and 113 KB at 256 tiles. Heavier ops have larger compute kernels, so `add` is the small end.

## Where it lands in models

Classified every eltwise call site under `models/` by resolving the memory config through the helper that produces it: **72 sites are interleaved, small and on a per-token or per-layer path; 86 hot sites are ruled out; 22 could not be resolved statically.**

**Out of reach, and it is the flagship:** `models/tt_transformers` and the llama / deepseek family are sharded end to end in decode. `get_residual_mem_config(Mode.DECODE)` returns a `create_sharded_memory_config` on every branch except galaxy (`model_config.py:1230-1260`), so the residual adds at `decoder.py:277,299,329` never take this path. Prefill is interleaved but large, where the win is gone.

**In reach:** the strongest is **Mamba** (`models/demos/wormhole/mamba`) — 9 qualifying sites, all explicit `ttnn.L1_MEMORY_CONFIG`, 80 and 160 tiles, and they run per layer per decode token over 64 layers: 4 multiplies, 5 adds and a silu per layer, so **576 binary dispatches per token**. Then the newer **TTTv2 ports** (`models/common/models/**`, `gpt_oss` 7 sites, `gemma4` 6) which deliberately keep the residual stream interleaved rather than sharded — so the reach grows with the newer stack rather than shrinking. Also `falcon7b_common` (`falcon_decoder.py:174,184`, 142 tiles, per layer per token), `ttnn.sampling`, Whisper's autoregressive decode, and diffusion step/conditioning glue. The CNN and detection demos are essentially all ruled out.

At those exact tile counts, on current `main`:

| shape | P300 add | P300 mul | P300 silu | N300 add | N300 mul | N300 silu |
|---|---|---|---|---|---|---|
| mamba residual `[1,1,32,2560]` (80 t) | **2.55x** | **2.55x** | 1.50x | 1.57x | 1.56x | 1.08x |
| falcon7b hidden `[1,1,32,4544]` (142 t) | **2.02x** | **2.01x** | 1.27x | 1.25x | 1.26x | 0.98x |
| mamba d_inner `[1,1,32,5120]` (160 t) | 1.99x | 1.98x | 1.25x | 1.38x | 1.27x | 0.97x |

## Not established

- Sharded paths and non-eltwise families are not measured.
- The measured optimum uses contiguous rectangles from (0,0); NoC placement effects are not isolated.
- The 22 undetermined sites, whose memory config depends on runtime configuration.

Related: #49996 targets kernel count for the same small-tensor overhead, and #53793 is the merged precedent for model-based split selection (topk).

Hardware: N300 and P300, both built and measured at tt-metal `7db779f57c2` (current `main`). The P300 figures are an in-tree A/B of the fix as the default; the N300 figures drive the same grids through `sub_core_grids` on a stock build of the same commit. Every source citation re-checked there.

kinginu:
## PR Category

Performance.

## Summary

Fixes #54915.

For interleaved tensors, the unary and binary_ng factories create kernels on the **entire worker grid** and every program-cache hit rewrites runtime args for every core of that grid, so the per-call cost scales with the grid, not the workload — flat ~24 µs on P300's 110-core grid and ~48 µs (binary) on N300's 64-core grid from 1 tile to 1024. #54915 measured that a right-sized `sub_core_grids` is faster below ~256 tiles, bit-identically — 2.0-3.5x for binary and 1.0-1.9x for unary on current `main`.

This makes that the default. `get_worker_grid` (unary and binary_ng) now sizes the interleaved default grid as `max(ceil(sqrt(tiles)), ceil(tiles/cap))` cores — the sqrt term is the measured optimum for light ops; the per-core cap is the measured bound below which no swept op loses to the full grid on either architecture (**4 tiles/core** for SFPU-chain work: unary, and binary with activations; **8** for plain FPU binary) — rounded up to a power of two, and snapped to the full grid once the target reaches it (large tensors keep today's grid and cache entry). The grid is a contiguous rectangle from the worker-grid origin; non-rectangular sub-device worker sets take the first N cores row-major.

The program cache forks per bucket instead of per volume. No hashing change is needed: both factories already key on the worker grid — unary in its `to_hash`, and binary_ng in its attribute tuple, which gained `worker_grid` upstream while this branch was in flight (an earlier revision of this branch added it, and that part of the diff is now gone). Distinct buckets are logarithmic in the grid (≤8 per op config on P300), and volumes within a bucket share one entry exactly as before — `override_runtime_arguments` re-splits on every hit unchanged. Explicit `sub_core_grids` and every sharded path are untouched.

## Cost

Batched sync-bounded protocol (5 warm-ups, 5 runs × 50 calls, one `synchronize_device` per run, median; re-measure once if run spread >10%), shape `[1,1,32,32·T]` TILE DRAM interleaved. Base is the same tree at `7db779f57c2` with the three source files reverted.

### In-tree A/B on current `main` (7db779f57c2), P300

Rebased onto `main` and measured as an in-tree A/B of the default call path: base is the same tree with the three source files reverted. The branch's original tables were taken at `b2e77906b92`, and `unary_program_factory.cpp` has since had a large refactor that roughly **halved the unary default** (~24 µs -> ~12.2 µs), so the unary win is much smaller than it was; the binary default is unchanged at ~24-25 µs. Median µs/call, 5x50 calls:

| tiles | add | multiply | exp | silu |
|---|---|---|---|---|
| 1 | 24.12 -> 7.35 (**3.28x**) | 23.88 -> 7.33 (**3.26x**) | 11.79 -> 6.14 (1.92x) | 11.78 -> 6.07 (1.94x) |
| 16 | 23.70 -> 7.55 (**3.14x**) | 23.31 -> 7.78 (**3.00x**) | 12.01 -> 6.26 (1.92x) | 12.05 -> 6.30 (1.91x) |
| 64 | 23.94 -> 8.23 (**2.91x**) | 24.07 -> 8.15 (**2.95x**) | 11.85 -> 6.90 (1.72x) | 11.85 -> 7.04 (1.68x) |
| 80 | 23.96 -> 9.38 (**2.55x**) | 23.69 -> 9.30 (**2.55x**) | 11.82 -> 7.92 (1.49x) | 11.90 -> 7.91 (1.50x) |
| 142 | 24.37 -> 12.06 (**2.02x**) | 24.68 -> 12.27 (**2.01x**) | 12.32 -> 9.74 (1.26x) | 12.50 -> 9.85 (1.27x) |
| 160 | 24.73 -> 12.41 (1.99x) | 24.70 -> 12.49 (1.98x) | 12.19 -> 9.87 (1.24x) | 12.33 -> 9.87 (1.25x) |
| 256 | 24.85 -> 12.51 (1.99x) | 24.52 -> 12.55 (1.95x) | 12.21 -> 9.78 (1.25x) | 12.31 -> 9.92 (1.24x) |
| 1024 | 25.11 -> 24.70 (1.02x) | 24.97 -> 25.01 (1.00x) | 12.25 -> 12.14 (1.01x) | 12.47 -> 12.29 (1.01x) |

80, 142 and 160 tiles are the shapes real decode paths use: mamba's residual stream and `d_inner`, and falcon7b's hidden state. At 1024 tiles the policy is the full grid, so the program is identical to base and the +-2% is run-to-run spread.

### N300, same commit

The worker grid is 64 cores rather than 110, so the default wastes less and the win is smaller. Stock build of the same commit driven through `sub_core_grids`, which builds the identical program:

| op class | 1 t | 16 t | 64 t | 256 t | 1024 t |
|---|---|---|---|---|---|
| binary, tensor operand (10 ops) | 1.69-1.81x | 1.64-1.72x | 1.61-1.71x | 1.29-1.36x | 0.98-1.01x |
| binary, scalar operand | 1.88-1.94x | 1.81-1.84x | 1.71-1.75x | 1.33-1.35x | — |
| unary (`exp` `gelu` `i0` `erfinv`) | 1.25-1.29x | 1.16-1.26x | 1.17-1.20x | 0.97-0.98x | 1.00x |

At the decode shapes: mamba 80 t 1.57x / 1.56x / 1.08x (add / mul / silu), falcon7b 142 t 1.25x / 1.26x / 0.98x, mamba 160 t 1.38x / 1.27x / 0.97x.

### Ternary must not take this treatment

`ttnn.where` on three tensors goes through `ternary_device_operation.cpp:17`, a different `get_worker_grid` that this change does not touch. That is deliberate: forcing the same policy onto it through `sub_core_grids` helps below 16 tiles and then collapses.

| tiles | 1 | 16 | 64 | 128 | 256 | 512 |
|---|---|---|---|---|---|---|
| ternary under this policy | 1.19x | 1.16x | **0.70x** | **0.71x** | **0.64x** | **0.52x** |

Its default is already flat at ~8 µs against ~24 µs for a plain binary, so the ternary factory is doing something better already. An earlier revision of this PR proposed extending the remedy there as a follow-up; that was wrong.

### L1 occupancy

The kernel binaries are written per core into the `kernel_config` region, so the grid decides the L1 footprint of a call. Loadable text+data for `ttnn.add` (bf16, interleaved), read from the JIT build output with `riscv-tt-elf-size`:

| | bytes |
|---|---|
| compute TRISC0 / TRISC1 / TRISC2 | 1,592 / 1,156 / 2,716 |
| dataflow reader / writer | 756 / 1,016 |
| **per core** | **7,236** |

| shape | full grid (110 cores) | right-sized | saved |
|---|---|---|---|
| 1 tile | 777.3 KB | 7.1 KB (1 core) | **770.2 KB (99%)** |
| 16 tiles | 777.3 KB | 28.3 KB (4 cores) | **749.0 KB (96%)** |
| 64 tiles | 777.3 KB | 56.5 KB (8 cores) | **720.8 KB (93%)** |
| 256 tiles | 777.3 KB | 113.1 KB (16 cores) | **664.2 KB (85%)** |

`add` is the small end: heavier SFPU ops have larger compute kernels. Program-cache entries hold no L1 of their own — `kernel_config` is a fixed per-core-type ring buffer (`dispatch.cpp:2617-2634`) and `reserve_space_in_kernel_config_buffer` takes from it at dispatch time — so bucketing does not multiply the footprint; only the number of cores written to changes.

### What the bucketing costs

Measured on a stock build by varying `sub_core_grids`, which forks the cache exactly as the `worker_grid` bucketing does. 8 distinct grids produce **8 cache entries and no more**; first call per bucket is **0.17–0.41 ms**; a warm cycle over all 8 runs at 12.44 µs/call with the entry count unchanged.

Cycling between buckets costs nothing measurable:

| | µs/call | spread |
|---|---|---|
| 16t on 4 cores, repeated | 7.23 | 4.8% |
| 1024t on 110 cores, repeated | 24.61 | 1.7% |
| **alternating the two** | **16.16** | 0.6% |
| expected if switching were free | 15.92 | |
| **-> overhead** | **+1.5%** | |
| control: 4-core grid alternating with *itself* | 7.15 | 2.1% |
| alternating 4-core and 8-core grids | 7.50 | 6.0% |
| **-> overhead, tight pair** | **+1.1%** | |

The control fixes the noise floor at about 1%, so +1.1–1.5% is at or just above it — against a 2.0-2.55x steady-state win at real decode shapes.

## Correctness

**Bit-identity on current `main`, with the fix as the default**: right-sized default vs an explicit full grid, seeded inputs, `add` / `multiply` / `subtract` / `exp` / `silu` / `gelu` at 1, 5, 16, 64, 80, 100, 142, 160, 256 and 512 tiles — **60 of 60 cells bit-identical**.

**No op regresses.** Right-sized default vs an explicit full grid on the same build, for the ops a restricted grid could plausibly hurt: `i0` / `gelu` / `erfinv` are 0.59-1.00x from 64 to 1024 tiles (faster or equal everywhere, no cell above 1.05x), and `where` — which this change does not touch — is 0.97-1.00x from 142 to 1024 tiles.

- **Exhaustive bf16**: all 65,536 bit patterns through `exp`, `gelu`, `relu`, `add`, `mul` at 64-tile and 32-tile shapes (multi-core policy grids), bf16 and fp32 inputs — 30 cells, every output **bit-identical** to the base build's full-grid results.
- The policy sweep bit-compared every measured cell against the default grid on both machines (508 restricted cells, 0 mismatches).
- **Within-bucket cache hits**: one grid shared across volumes 65–128 (uneven splits, active↔noop flips on an RM case, in-place variants) — correct on every hit, one cache entry per grid (probe on base via `sub_core_grids` = the identical path, re-run on this build).

## Tests

- Suites re-run on this build: the three files this PR migrates **9025 passed, 1 xfailed**; `test_binary_bcast.py` **3848 passed** (broadcast paths through the resized grids); `test_unary_subcoregrids.py` and `test_backward_gelu.py` **46 passed** (the latter's cache asserts are unchanged — fixed shape).
- `test_unary_program_cache.py` and `test_binary_ng_program_cache.py` updated where they pinned the old one-entry-per-volume contract: the migration guards keep their intent with same-bucket volume pairs (5→16 tiles on one 4-core grid, uneven group-1/group-2 splits), and new bucket-boundary tests pin the new contract — the binary one also pins the worker_grid hash as a regression guard (without it, differently-sized grids would collide on one entry); both new tests fail on base as intended, confirmed by running the same files against a build of this tree with the change reverted, where the bucket-boundary assert fails `1 == 2`. `test_quantization.py`'s per-tensor cache test asserted entry-count equality across differently-sized iterations; it now asserts the surviving invariant — a second identical pass is all cache hits (seeded per pass).
- Cache accounting: an 11-size sweep (1–1024 tiles) creates 8 entries per op (was 1); each new bucket pays one program creation on first call (<0.5 ms warm-process), JIT binaries are shared across buckets (same kernels, same defines).

## Not covered

- N300 in-tree A/B not run (no writable build there); the N300 column is the stock build driven through `sub_core_grids`, which constructs byte-identical programs to this change's default (same `worker_grid` attribute, same factory path). The measured policy cells and bit-identity above cover it.
- Sharded paths and non-eltwise families: untouched, unmeasured (issue scope).
- **Ternary is out of scope and must stay so.** See the table above: the same policy costs 0.52-0.71x on `where` from 64 tiles up. `where_ttt` has its own `get_worker_grid` and this change does not touch it.
- **The per-core cap is wrong for heavy SFPU binaries.** `logaddexp` falls to 1.29x at 64-256 tiles under the binary cap of 8; it belongs on the SFPU cap of 4 or its own.
- Workloads asserting exact eltwise cache-entry counts outside these two test files would see counts change (bounded by the bucket count).

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `7db779f57c2` (current `main`) on both machines — P300 as an in-tree A/B of this branch against the same tree with the change reverted, N300 as a stock build of the same commit driven through `sub_core_grids`, which constructs the identical program. The branch was rebased onto `main` for this: `unary_program_factory.cpp` had a large refactor that halved the unary default, and `binary_ng_device_operation.cpp` drifted +119/-6.
* Hardware: N300 and P300

### DIFF
--- a/tests/ttnn/nightly/unit_tests/operations/eltwise/test_quantization.py
+++ b/tests/ttnn/nightly/unit_tests/operations/eltwise/test_quantization.py
@@ -435,31 +435,33 @@ def test_quantization_per_channel_4d(device, x0, x1, x2, x3, input_dtype):
 # kernels used
 @pytest.mark.parametrize("input_dtype", [ttnn.float32, ttnn.bfloat16])
 def test_quantization_per_tensor_program_cache(device, input_dtype):
-    torch.manual_seed(0)
+    def sweep():
+        torch.manual_seed(0)  # identical data both passes; only the cache state differs
+        for dim in [1, 2, 3, 4]:
+            for i in range(5):
+                # Each iteration gets completely different input tensors, quant ranges, etc.
+                input_tr = torch.rand([30 + i] * dim, dtype=torch.float32)
 
-    num_program_cache_entries_list = []
+                scale, zero_point = calculate_scale_zero_point_per_tensor(input_tr, -120 + i, 121 - i)
+                scale_r, zero_point_r = calculate_scale_zero_point_per_tensor(input_tr, -50 - i, 42 + i)
 
-    for dim in [1, 2, 3, 4]:
-        for i in range(5):
-            # Each iteration gets completely different input tensors, quant ranges, etc.
-            input_tr = torch.rand([30 + i] * dim, dtype=torch.float32)
+                input_tt = ttnn.from_torch(input_tr, dtype=input_dtype, layout=ttnn.TILE_LAYOUT, device=device)
+                quantized_tt = ttnn.quantize(input_tt, scale, zero_point)
+                requantized_tt = ttnn.requantize(quantized_tt, scale, zero_point, scale_r, zero_point_r)
+                derequantized_tt = ttnn.dequantize(requantized_tt, scale_r, zero_point_r, dtype=input_dtype)
+                result_tr = ttnn.to_torch(derequantized_tt)
 
-            scale, zero_point = calculate_scale_zero_point_per_tensor(input_tr, -120 + i, 121 - i)
-            scale_r, zero_point_r = calculate_scale_zero_point_per_tensor(input_tr, -50 - i, 42 + i)
+                check_pcc(input_tr, result_tr, False)
+                check_match_ratio(input_tr, result_tr, input_dtype)
 
-            input_tt = ttnn.from_torch(input_tr, dtype=input_dtype, layout=ttnn.TILE_LAYOUT, device=device)
-            quantized_tt = ttnn.quantize(input_tt, scale, zero_point)
-            requantized_tt = ttnn.requantize(quantized_tt, scale, zero_point, scale_r, zero_point_r)
-            derequantized_tt = ttnn.dequantize(requantized_tt, scale_r, zero_point_r, dtype=input_dtype)
-            result_tr = ttnn.to_torch(derequantized_tt)
-
-            check_pcc(input_tr, result_tr, False)
-            check_match_ratio(input_tr, result_tr, input_dtype)
-
-            num_program_cache_entries_list.append(device.num_program_cache_entries())
-
-    assert num_program_cache_entries_list[0] > 0
-    assert max(num_program_cache_entries_list) == min(num_program_cache_entries_list)
+    # First pass may create one entry per (op, worker-grid bucket): the interleaved default
+    # grid is right-sized to the tensor volume, so volumes in different buckets get separate
+    # (logarithmically many) entries. A second identical pass must be all cache hits.
+    sweep()
+    entries_after_first = device.num_program_cache_entries()
+    assert entries_after_first > 0
+    sweep()
+    assert device.num_program_cache_entries() == entries_after_first
 
 
 @pytest.mark.parametrize("x0", [32])
--- a/tests/ttnn/unit_tests/operations/eltwise/test_binary_ng_program_cache.py
+++ b/tests/ttnn/unit_tests/operations/eltwise/test_binary_ng_program_cache.py
@@ -28,6 +28,11 @@ Fields correctly excluded from hash (handled by override_runtime_arguments):
   kernel defines), and on the accessor path the TensorAccessor's page geometry is
   baked into the program. So sharded operands and sharded outputs each contribute
   their tensor shape in pages to the key (issue #54138)
+  Within the interleaved case this now means: shapes whose right-sized worker grid is the
+  same share an entry (the worker grid IS hashed and is sized from the tensor volume:
+  max(ceil(sqrt(tiles)), ceil(tiles/cap)) cores, cap 8 tiles/core for plain binary, 4 with
+  an SFPU activation chain, power-of-two rounded); volumes in different buckets get
+  separate entries
 - scalar.has_value(): not in attribute_values(), but compute_program_hash branches
   on input_tensor_b presence
 - input_dtype: not in attribute_values(), but compute_program_hash includes input
@@ -110,17 +115,18 @@ def test_ng_cache_reuse_same_config(device, isolate_program_cache):
 @pytest.mark.parametrize(
     "shape_first, shape_second",
     [
-        ([1, 1, 32, 64], [1, 1, 128, 256]),  # grow volume
-        ([1, 1, 128, 256], [1, 1, 32, 64]),  # shrink volume
+        ([1, 1, 32, 160], [1, 1, 128, 128]),  # grow volume: 5 tiles -> 16 tiles, same 4-core bucket
+        ([1, 1, 128, 128], [1, 1, 32, 160]),  # shrink volume
     ],
 )
 def test_ng_inplace_cache_reuse_different_shapes(device, isolate_program_cache, shape_first, shape_second):
     """binary_ng re-applies all per-dispatch state on a cache hit via override_runtime_arguments
-    (#48928). An in-place add (output_tensor aliases input) with different logical shapes shares one
-    cache entry (volume is excluded from the hash), so the second call is a cache HIT that reuses the
-    first program WITHOUT rebuild. binary_ng's override_runtime_arguments must re-derive every per-core
-    arg for the current shape or the reused program corrupts the result. Regression guard for the
-    in-place cache-hit path (the SDXL silu / moreh class of bug)."""
+    (#48928). An in-place add (output_tensor aliases input) with different logical shapes in the same
+    worker-grid bucket shares one cache entry (volume is excluded from the hash within a bucket), so
+    the second call is a cache HIT that reuses the first program WITHOUT rebuild. binary_ng's
+    override_runtime_arguments must re-derive every per-core arg for the current shape or the reused
+    program corrupts the result. Regression guard for the in-place cache-hit path (the SDXL silu /
+    moreh class of bug)."""
 
     def inplace_add(shape, seed):
         torch.manual_seed(seed)
@@ -441,10 +447,11 @@ def test_ng_different_input_dtypes_same_output_dtype(device, isolate_program_cac
 
 
 def test_ng_cache_reuse_different_logical_shapes(device, isolate_program_cache):
-    """Different logical shapes share 1 cache entry, different outputs (by design).
-    logical_shape is correctly excluded from compute_program_hash();
-    override_runtime_arguments handles shape differences at runtime."""
-    torch_ref1, tt_out1 = run_binary_ng_op(device, ttnn.add, [1, 1, 32, 32], [1, 1, 32, 32], dtype=ttnn.float32)
+    """Logical shapes in the same worker-grid bucket share 1 cache entry, different outputs
+    (by design). logical_shape is correctly excluded from compute_program_hash();
+    override_runtime_arguments handles shape differences at runtime. A volume in a different
+    bucket gets its own right-sized worker grid and therefore its own entry."""
+    torch_ref1, tt_out1 = run_binary_ng_op(device, ttnn.add, [1, 1, 32, 64], [1, 1, 32, 64], dtype=ttnn.float32)
     assert_with_pcc(torch_ref1, tt_out1, 0.9999)
 
     torch_ref2, tt_out2 = run_binary_ng_op(device, ttnn.add, [1, 1, 64, 64], [1, 1, 64, 64], dtype=ttnn.float32)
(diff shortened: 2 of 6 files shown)


mateusznowakTT:
This PR is aiming to fine tune generic codes for specific case.
The same can be achieved by passing properly configured sub_core_grids parameter to op invocation.

Closing.

Issue #15632, opened 2024-12-03
[Bug Report] UMD headers are leaking from tt-metal


rfurko-tt:
**Describe the bug**
UMD headers are leaking from tt-metal. 

**To Reproduce**
Include of `host_api.hpp` results in include of `program.hpp` -> `kernel_types.hpp` -> `tt_cluster.hpp`

And `tt-cluster.hpp` includes bunch of `UMD` headers:

```
#include "umd/device/device_api_metal.h"
#include "umd/device/tt_cluster_descriptor.h"
#include "umd/device/tt_xy_pair.h"
```

**Expected behavior**
`tt-metal` should not leak headers of internal libraries. It makes library import more complicated.

**Please complete the following environment information:**
- OS: Ubuntu 20.04
- Version of software 070a226c183d8bc33f57e7edd34763c753ad58be

rfurko-tt:
@dmakoviichuk-tt @ayerofieiev-tt @davorchap FYI

pgkeller:
this will be covered by the metal API refactor, assigning Eyon for tracking

prajaramanTT:
@eyonland @rfurko-tt is this still an open issue ? If not, can you please close this ? Thanks

blozano-tt:
> [@eyonland](https://github.com/eyonland) [@rfurko-tt](https://github.com/rfurko-tt) is this still an open issue ? If not, can you please close this ? Thanks

The problem is still there:
https://github.com/tenstorrent/tt-metal/blob/8c9694509d79cecf8501d5c7f33b4264eb4ae94e/tt_metal/llrt/tt_cluster.hpp#L14-L16

We should probably reassign this issue to someone who routinely develops tt_metal.

blozano-tt:
It is difficult to fix this within tt_cluster.hpp, because the return type of this function is defined in UMD. So it can't be forward declared:
https://github.com/tenstorrent/tt-metal/blob/8c9694509d79cecf8501d5c7f33b4264eb4ae94e/tt_metal/llrt/tt_cluster.hpp#L112C5-L116C6

blozano-tt:
I broke this chain: Include of host_api.hpp results in include of program.hpp -> kernel_types.hpp -> tt_cluster.hpp

In #16691 

However, I'm not sure it will fully take tt_cluster.hpp out of the public interface.

There could be other files that leak the same header.

eyonland:
@pgkeller , who would be the right person to look into this?

[...]

PR #54934, opened 2026-08-31
[Cleanup] Rename CoreCoord to xy_pair in tt_metal/llrt

aleksamarkovicTT:
### PR Category
Cleanup

### Summary
Related to: https://github.com/tenstorrent/tt-umd/issues/1216

`CoreCoord` and `xy_pair` are both aliases of `tt_xy_pair`, so this is a mechanical, source- and ABI-compatible rename, scoped to `tt_metal/llrt` only.

- Adds `using xy_pair = tt_xy_pair;` next to the existing `CoreCoord` alias in `tt_metal/api/tt-metalium/core_coord.hpp`.
- Replaces `CoreCoord` with `xy_pair` throughout `tt_metal/llrt` (234 occurrences across 13 files), including comment mentions of the type.

### Notes for reviewers
We're in a really tricky situation today, in UMD there's two coordinate types with separate purposes:
1. xy_pair, which is just two scalars, called **CoreCoord** in metal. An incomplete type that doesn't convey the coord. system and cannot be translated to other coord. system.
2. CoreCoord, which contains core type and coordinate system and completely covers xy_pair.

The UMD stance is that metal should adopt our complete CoreCoord and move away from its differently-named coordinate systems. There are many benefits to this, one of which is that coordinate correctness can be easily verified.

The way we will achieve this is:
1. Gradually remove the confusing `CoreCoord` alias and use `xy_pair`, which is the actual name of the type.
2. Move Metal's CoreCoord methods that are missing for umd::CoreCoord to UMD.
3. Gradually adapt metal to use umd::CoreCoord and its coordinate systems.

This is the first PR out of many for step 1, the type will be changed to xy_pair and then the alias to CoreCoord will be removed.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:amarkovic/llrt-corecoord-to-xy-pair)

### DIFF
--- a/tt_metal/api/tt-metalium/core_coord.hpp
+++ b/tt_metal/api/tt-metalium/core_coord.hpp
@@ -29,6 +29,7 @@ struct to_json_t;
 namespace tt::tt_metal {
 
 using CoreCoord = tt_xy_pair;
+using xy_pair = tt_xy_pair;
 
 class CoreRangeSet;
 
--- a/tt_metal/llrt/core_descriptor.cpp
+++ b/tt_metal/llrt/core_descriptor.cpp
@@ -236,7 +236,7 @@ const tt::core_descriptor_t& MetalEnvImpl::get_core_descriptor_config(
     size_t start_x = compute_with_storage_start[0].as<size_t>();
     size_t start_y = compute_with_storage_start[1].as<size_t>();
 
-    tt::tt_metal::CoreCoord compute_grid_size;
+    tt::tt_metal::xy_pair compute_grid_size;
     // When slow dispatch is on, use full logical grid (no dispatch cores to reserve).
     // Quasar dispatch-engine FD uses dedicated dispatch tiles from the soc `dispatch:` list, so
     // Tensix workers listed under dispatch_cores in the fast-dispatch core descriptor YAML are not
@@ -257,7 +257,7 @@ const tt::core_descriptor_t& MetalEnvImpl::get_core_descriptor_config(
                 compute_grid_size.y);
         }
     } else {
-        compute_grid_size = tt::tt_metal::CoreCoord((end_x - start_x) + 1, (end_y - start_y) + 1);
+        compute_grid_size = tt::tt_metal::xy_pair((end_x - start_x) + 1, (end_y - start_y) + 1);
     }
 
     std::vector<RelativeCoreCoord> compute_cores;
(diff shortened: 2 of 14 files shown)


tt_metal/api/tt-metalium/core_coord.hpp
```diff
@@ -29,6 +29,7 @@ struct to_json_t;
 namespace tt::tt_metal {
 
 using CoreCoord = tt_xy_pair;
+using xy_pair = tt_xy_pair;
```
riverwuTT, line 32: `tt_xy_pair` is a umd symbol leaked to the tt-metal. There's an ongoing effort to remove this coupling: https://github.com/tenstorrent/tt-metal/issues/15632

If we introduce this to the `api/tt-metalium` surface, we will be more hopelessly coupled to umd headers.

Please put this alias somewhere outside of the api folder.

riverwuTT (changes requested):
Reject putting the `xy_pair` type alias in the api folder as this introduces more coupling between runtime api user's include surface with umd apis. See more in: https://github.com/tenstorrent/tt-metal/issues/15632

Can you provide more context on why this change is made?
Is `xy_pair` and `CoreCoord` referring to different coordinate systems?
If so, you have an excellent, once-in-a-life-time opportunity to give it a better name.

PR #53904, opened 2026-08-21
[Feature] Make sliding RingJointSDPA trace-replay safe

svuckovicTT:
### PR Category

Feature

### Summary

Makes sliding-window RingJointSDPA trace-replay safe for Gemma4 chunked prefill.

- Extends the metadata-tensor contract introduced by #48907 from `ring_mla` to `ring_joint_scaled_dot_product_attention`.
- Treats the full physical KV-cache allocation as the gather ceiling while deriving the active prefix on-device from `kv_actual_isl`.
- Derives the sliding neighbor-halo source range and flattened cache batch on-device, allowing one captured program to replay across growing prefixes and cache slots.
- Isolates concurrent halo metadata reads and gives the halo its own semaphore, preventing intermittent cache-slot corruption and mixed sliding/dense replay deadlocks.
- Adds focused 32-device coverage for scalar equivalence, mixed sliding/dense capture, changing-prefix replay, and readback-free stress replay.

### Notes for reviewers

The metadata read protocol and API naming intentionally follow #48907. The main new logic is specific to the sliding halo, which that PR did not cover.

Please focus review on:

- the metadata-path gather ceiling in `compute_gather_valid_Ht`;
- host/device parity for `compute_halo_tail_start_Ht`;
- on-device cache-batch selection in the halo reader;
- separate reader/writer metadata CBs and the dedicated third halo semaphore.

Validation:

- `./build_metal.sh -ce`
- focused mixed sliding/dense trace-replay test: five consecutive passes plus a final post-rewrite pass
- first-group and cache-growth sliding accuracy/determinism regression
- indexed KV-cache and cache-hit sliding regression
- formatting, pre-commit hooks, Python compilation, collection, and `git diff --check`

The deadlock root-cause analysis and reproducer are included in `docs/superpowers/specs/2026-08-06-ring-trace-replay-deadlock.md`.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:svuckovic/gemma4-prefill-pr-sdpa)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:svuckovic/gemma4-prefill-pr-sdpa)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:svuckovic/gemma4-prefill-pr-sdpa)
[...]

PR #53153, opened 2026-08-14
Fix GPT-OSS first-group chunked sliding RingJointSDPA

pavlejosipovic:
## Summary

- unify the first-group RingJointSDPA prefill path with the regular chunked sliding flow
- update GPT-OSS disaggregated-prefill attention/runtime handling and regression coverage
- refine sliding-window work-plan and halo-layout behavior

## Validation

- Sanity Tests: already run
- All Model Tests (Tier 2 unit, gpt-oss-120b on bh_p150): running

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:pjosipovic/ring-joint-sdpa-first-group-unified-prefill)

PR #54935, opened 2026-08-31
[gpt_oss_d_p] Bounded sliding-window KV cache (PR2: RingJointSDPA bounded ring read)

mdragulaTT:
## Problem
PR1 (#53773) bounds the sliding-layer KV cache to a circular window, but the ring cache-read still assumes the full-length cache layout: on-device slab addressing walks `[0, kv_actual)` linearly, so chunked prefill beyond the window capacity reads garbage on sliding layers.

## Solution
RingJointSDPA learns the circular layout behind one hashed bool, `circular_kv_cache` (the slab count is derived on-device from the cache/Q geometry — whole Q-sized slabs, >= 2, validated). The shared work-plan header wraps slab indices modulo that count, matching PR1's host-side modulo write; full-attention layers and unbounded callers are untouched (flag off keeps the legacy path bit-for-bit).

Stacked on #53773 (PR1: allocation + circular host write); this PR adds the on-device ring read, so chunked runs cover every chunk.

Variable-chunk integration (the stack is rebased over #53672): the circular capacity is sized off the **largest** served chunk size (`2 × max_chunk_size`), `gather_layer`/`kv_cache_pcc_check` take the per-call chunk size, and every served size must divide the capacity (asserted at allocation — keeps the slab math sound for whichever uniform size a sequence uses).

## Testing
- CPU unit tests for the bounded ring plan math (slab wrap, boundary chunks, capacity edge cases) — pass; plus PR1's bounded KV math tests.
- Galaxy, 36L, real weights, 5k golden @1024 chunks (PR1+PR2 stack):
  - regression (bounded OFF): min KV PCC across 36 layers K=0.982 / V=0.920 — matches the pre-change baseline.
  - bounded ON: sliding layers hold only **1928/5000 positions resident** (2×1024 circular window, wraps multiple times) and land the **same min PCC** K=0.980 / V=0.920 (the min V lives on a full-attention layer in both runs).
- Post-rebase revalidation on the #53672-integrated stack (fresh build, CPU unit tests, both device runs): **identical PCCs** — regression K=0.982 / V=0.920, bounded K=0.980 / V=0.920. The variable-chunk integration preserves behavior exactly.
- **Rebased over #53904** (trace-safe sliding RingJointSDPA): compile-time-arg slots renumbered around the new rank-mapping descriptor (reader 44, compute 58); full revalidation on the rebased stack — 32 CPU unit tests, the 1x1 bounded device-write test, and galaxy regression / bounded / 2-user runs all land **min KV PCC 0.92036**, identical to the unbounded baseline.
- **Review round** (Copilot + tdd + op-owner feedback): API narrowed to `circular_kv_cache: bool` (count derived on-device); `dense_sp` rejects circular + direct-write (foot-gun); the program factory carries compile-time `static_assert`s pinning the plan's wrap behavior to CPU-mirror-generated values, so header drift fails the host build. Full revalidation after the refactor: same 0.92036 everywhere (regression / bounded / 2-user + the 1x1 device write test).
- **Perf-band regression found and fixed (review round 3)**: BH sanity's QB2 sdpa leg failed only `test_ring_joint_attention_minimax3_gqa_chunked_perf_check` (46.26% vs band [46.53, 47.47]) on two boxes / two days, while PR1's branch and `main` measured 46.9% the same day; every other ring-joint perf check moved < 0.2 pts. Root cause: three PR changes touched code shared with every ring kernel — an immediately-invoked `[&]` lambda in the kv-pad rotation mask loop, a fourth field on `SlidingKChunkRef` (returned by `k_chunk_at`, which dense kernels also instantiate), and plan-origin lookups referenced inside the existing capture-all lambdas. GCC fixes `[&]` captures at template definition, so even sliding-only references inside discarded `if constexpr` branches force those objects into memory and reshuffle the inlined kernel's register allocation. Fix: every deviation is gated on the compile-time bounded slab count, unbounded kernels keep the original statements, the bounded path takes absolute K origins from the work plan outside the lambdas, and `SlidingKChunkRef` is back to three fields. Verified with the JIT's logged compile commands + `objdump` diff: the dense minimax3 kernel (the unbounded sliding kernel differs only by the plan builder storing each range's absolute origin: +24–36 bytes in `kernel_main`, outside the K-chunk loop; no CI perf band covers that configuration) compile instruction-for-instruction identical to the pre-change binaries on all three triscs. Galaxy re-check of the sliding tests: production 0.99967, circular (shared runtime, both parities) 0.99967 / 0.99967, metadata rejection fires. QB2 sanity re-run: green — https://github.com/tenstorrent/tt-metal/actions/runs/34360226101 (QB2 p03t03: minimax3 1.086 ms / 46.92% vs main 1.087 / 46.89% and PR1 1.087 / 46.91% the same day; 140 passed, every band in range).
- **Known interplay**: `circular_kv_cache` rejects the trace-safe metadata path (validated in the device op) — bounded SWA and trace-replay are mutually exclusive until the wrap can be derived from the metadata tensors; tracked in #56115.

## Impact
With PR1+PR2, the 18 sliding layers keep a small circular cache instead of full-length: at 128k with the 8k default chunk this saves ~44% of KV DRAM per chip (153 → 86 MiB per user) — headroom for multi-user prefill. Opt-in via `bounded_sliding_kv_cache`; default behavior unchanged.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mdragula/gpt-oss-bounded-sliding-pr2)


ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_device_operation.cpp
```diff
@@ -612,6 +633,24 @@ void RingJointSDPADeviceOperation::validate_on_program_cache_miss(
...
+        // slab-major addressing and would read garbage from a wrapped cache.
+        TT_FATAL(
+            args.has_sliding_window() && is_chunked,
+            "bounded_kv_slab_count requires chunked sliding-window attention");
+        TT_FATAL(
+            has_kv_pad_rotation,
+            "bounded_kv_slab_count requires kv_actual_isl (KV-pad rotation): the wrap position is "
+            "derived from the true absolute logical_n/kv_actual_isl");
+        TT_FATAL(
+            !tensor_args.has_metadata(),
```
pavlejosipovic: How should bounded KV compose with the #53904 trace-metadata path used by Gemma4? This guard is necessary today: metadata clamps absolute kv_actual_isl/logical_nt to physical capacity, and halo derivation lacks the circular slab modulo. Removing it would misaddress data after the first wrap. If support is follow-up work, please track it explicitly and add a rejection test.

ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
```diff
@@ -649,6 +651,11 @@ void bind_sdpa(nb::module_& mod) {
                 compute kernels prune K chunks outside the window. Ring attention currently supports the
                 GPT-OSS specialization: a 128-token window, local 8Q:1K:1V heads with D64, BF16 Q,
                 BFP8_B K/V, SP4 production or SP8 test topology, and chunked prefill without joint tokens.
+            bounded_kv_slab_count (int, optional): Bounded circular sliding KV cache. The per-device
```
pavlejosipovic: Could the API expose only a circular/bounded mode and derive the slab count internally? The only valid value is N_local_kv / N_local_q; exposing it leaks chunk geometry, duplicates Python logic, and varies with served chunk size. Prefer circular_kv_cache=True plus internal divisibility and n_slabs >= 2 validation. The gather-buffer shape cannot provide this signal because bounded and unbounded sliding calls use the same halo.

models/demos/gpt_oss_d_p/tt/attention/dense_sp.py
```diff
@@ -72,8 +73,16 @@ def dense_sp_attention(
...
+    bounded_kv_slab_count  bounded circular sliding cache: the cache holds only this many chunk
+                      slabs and the ring read wraps its local slab addressing (chunk group g in slab
+                      g mod n_slabs). None on full layers / unbounded caches. kv_actual and logical_n
+                      stay TRUE ABSOLUTE lengths either way — the op derives the wrap on-device.
     -> out            [1, n_q_local, chunk_local, head_dim]  block-cyclic over the chunk
     """
+    assert bounded_kv_slab_count is None or sliding_window_size is not None, (
+        "bounded_kv_slab_count is only meaningful for sliding-window layers "
+        f"(got bounded_kv_slab_count={bounded_kv_slab_count}, sliding_window_size=None)"
+    )
```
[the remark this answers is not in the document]
pavlejosipovic: Agreed. Please centralize circular offset derivation instead of duplicating kv_actual % cache_capacity in Python. Writer and reader need one shared layout contract; otherwise a call path can write linearly while the reader interprets the cache circularly. Ideally the cache abstraction/write op derives the physical offset from the cache shape.

pavlejosipovic:
Before merge, please run the Sanity Tests workflow with Blackhole enabled and require the BH legs to pass. This PR changes Blackhole RingJointSDPA kernels and compile-time argument layouts, while the current hardware sanity jobs are skipped.

pavlejosipovic:
Please update test_ring_joint_attention_gpt_oss_chunked_sliding_production_accuracy in tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py to cover this feature. It currently uses the full 55K prefix and never wraps. Run it with a two-chunk physical cache, circularly place the final chunk and predecessor, keep logical_n/kv_actual_isl absolute, enable bounded mode, and compare against the existing sliding-window+sinks reference. This provides production-kernel hardware coverage beyond the CPU plan test.

models/demos/gpt_oss_d_p/tt/attention/dense_sp.py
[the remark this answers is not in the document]
mdragulaTT: Real foot-gun — fixed in 575a81b8a3: dense_sp now asserts that circular caches are written via write_kv_chunk (the host-side modulo path); the direct write_chunk=True path rejects circular_kv_cache.

ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa.hpp
```diff
@@ -110,6 +110,7 @@ std::tuple<ttnn::Tensor, ttnn::Tensor, ttnn::Tensor> ring_joint_scaled_dot_produ
     std::optional<uint32_t> kv_actual_isl = std::nullopt,
     const std::optional<ttnn::Tensor>& attention_sink = std::nullopt,
     std::optional<uint32_t> sliding_window_size = std::nullopt,
+    std::optional<uint32_t> bounded_kv_slab_count = std::nullopt,
```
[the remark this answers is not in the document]
mdragulaTT: Addressed two ways in 575a81b8a3: (1) the program factory now carries compile-time static_asserts pinning build_sliding_q_work_plan's wrap behavior (wrapped slab bases, absolute K origins, unbounded identity) to values generated by the CPU mirror — a header change that shifts behavior now fails the HOST build instead of drifting; (2) the galaxy PCC runs vs the independent HF golden remain the end-to-end backstop for the reader/compute plumbing.

models/demos/gpt_oss_d_p/tt/attention/prefill.py
```diff
@@ -209,17 +209,18 @@ def attention_forward(
...
-                raise NotImplementedError(
-                    f"bounded_sliding_kv_cache: on-device cache-read of a bounded sliding layer "
-                    f"(layer {layer_idx}, cached_len={cached_len}) is not supported — this build "
-                    f"has allocation + the circular write only; validate via host readback "
-                    f"(kv_cache_pcc_check)."
+                assert cache_capacity % chunk_global == 0, (
+                    f"bounded sliding cache capacity ({cache_capacity}) must be a whole number of "
+                    f"chunk slabs (chunk_global={chunk_global})"
                 )
+                bounded_kv_slab_count = cache_capacity // chunk_global
```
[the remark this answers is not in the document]
mdragulaTT: Resolved by removing the Python-side computation entirely (also per pavlejosipovic's API suggestion): the op now takes circular_kv_cache (bool) and derives the slab count on-device from the cache/Q geometry, TT_FATALing on non-whole slabs or fewer than 2.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/sliding_window_work_plan.hpp
```diff
@@ -172,12 +200,17 @@ constexpr SlidingQWorkPlan build_sliding_q_work_plan(
         if (plan.source_range_count == SlidingQWorkPlan::max_source_ranges) {
             return SlidingQWorkPlan{};
         }
+        // Absolute global chunk of first_k_chunk: undo the (possibly wrapped) slab base. q_local %
+        // k_chunk == 0 (checked above), so both bases are whole k-chunk multiples.
+        const uint32_t first_global_k_chunk =
```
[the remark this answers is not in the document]
mdragulaTT: Extracted in 575a81b8a3: slab_global_k_chunk_base / source_local_k_chunk_base, with the same extraction mirrored in the CPU test.

jmalone-tt:
Wanted to make sure these changes won't break shared prefix (not supported today, but we want it eventually). Is it possible to have 128-token sliding window KV cache read but allocate the full max sequence length for SWA so we can use shared prefixes between requests?

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
```diff
@@ -5732,6 +5774,122 @@ def test_ring_joint_attention_gpt_oss_chunked_sliding_production_accuracy():
...
+            mesh_config,
+            GPT_OSS_CHUNKED_MODEL,
+            chunk_size=GPT_OSS_CHUNKED_CHUNK_SIZE,
+            total_seq=GPT_OSS_CHUNKED_TOTAL_SEQ,
+            qk_configs=[(GPT_OSS_Q_CHUNK_SIZE, GPT_OSS_K_CHUNK_SIZE)],
+            persistent_buffer_mode="exact_per_chunk",
+            sliding_window_size=GPT_OSS_RING_SINK_CONFIG.sliding_window_size,
+            use_attention_sink=True,
+            circular_kv_cache=True,
+        )
```
pavlejosipovic: This covers both slab parities, but each parametrized case creates and closes its own runtime and makes only one call, so both exercise program creation rather than a cache hit. Please run the two chunks through one shared runtime so the second call verifies `apply_ring_joint_scalar_runtime_args` relocates the circular halo correctly when `logical_n` and slab parity change.

pavlejosipovic:
The Blackhole sanity run still shows possible default-path performance regressions. Minimax3 measured 46.27% and 46.26% in two consecutive runs, below the 46.53% lower bound, and the latest MLA run measured 62.55% vs a 62.57% lower bound. Since this PR changes the shared RingJointSDPA kernel and says the flag-off path is unchanged, do these failures also reproduce on the stacked base? Could you investigate with a same-QB2 base-vs-PR comparison and rerun the BH sanity legs before merge?

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_device_operation.cpp
mdragulaTT: Agreed, the guard stays; it is now the first check and covered by `test_ring_joint_circular_kv_cache_rejects_metadata_path`. I will file a follow-up issue for the trace composition.

ttnn/cpp/ttnn/operations/transformer/sdpa/sdpa_nanobind.cpp
mdragulaTT: Done: `circular_kv_cache: bool`, slab count derived and validated on-device.

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
mdragulaTT: Done: both parities go through one runtime, and the test asserts the program-cache entry count is unchanged after the second call.

mdragulaTT:
@jmalone-tt @pavlejosipovic 

> Before merge, please run the Sanity Tests workflow with Blackhole enabled and require the BH legs to pass.

Done: BH sanity on the current head, both QB2 legs green — https://github.com/tenstorrent/tt-metal/actions/runs/34360226101.

> Please update test_ring_joint_attention_gpt_oss_chunked_sliding_production_accuracy ... to cover this feature.

Done as a sibling test so the unbounded baseline stays intact: `test_ring_joint_attention_gpt_oss_chunked_sliding_circular_cache_accuracy` runs the production sliding+sink config from a two-slab circular cache (final chunk + predecessor placed circularly, logical_n/kv_actual_isl absolute), both slab parities through one runtime, same reference; PCC 0.99967 in the sdpa nightly.

> The Blackhole sanity run still shows possible default-path performance regressions ... do these failures also reproduce on the stacked base?

No, they were ours. Same-day QB2 comparison: main and PR1 both at 1.087 ms / 46.9% on minimax3, this PR at 1.102 ms / 46.3%. Cause: a [&] lambda plus two struct/lookup changes in code shared with the dense kernel changed its register allocation (GCC fixes captures at template definition, even inside discarded if constexpr branches). Fixed so unbounded kernels compile instruction-identical to pre-change (verified by diffing the JIT binaries); the rerun above has minimax3 at 1.086 ms / 46.92%. mla_100k sits on its band edge on main and PR1 too (62.65–62.69 vs 62.57), so that one is pre-existing.

> Is it possible to have 128-token sliding window KV cache read but allocate the full max sequence length for SWA so we can use shared prefixes between requests?

Yes: bounded is opt-in, and with the flag off SWA layers keep the full max_seq_len allocation with the 128-token read, byte-identical to today. We will keep the flag as a deployment choice rather than make bounded the only layout. With the flag on, prefix reuse for a sliding layer only needs the prefix tail (window + halo), which the circular cache always retains; sharing circular slots across requests is the follow-up migration/prefix-sharing work.

pavlejosipovic (approved):

ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/sliding_window_work_plan.hpp
```diff
@@ -24,6 +28,14 @@ struct SlidingKChunkRef {
     uint32_t compact_k_chunk = 0;
 };
 
+// Bounded circular sliding KV cache (bounded_kv_slab_count = n_slabs): each device keeps only
+// n_slabs Q-sized slabs and chunk group g lives in local slab (g % n_slabs). 0 or 1 means the
+// cache is unbounded (identity). This is the ONLY place local slab placement is derived; keep
```
philei-tt, line 33: This is not the only place: `compute_halo_tail_start_Ht` in `ring_attention_all_gather_metadata.hpp` (marked KEEP IN SYNC with `chunked_sliding_halo_source_start_tile`) still returns `source_group * q_local_tile_rows + ...` without the wrap. Unreachable today only because circular rejects the metadata path, but whoever lifts that TT_FATAL will find one twin unwrapped. Either plumb the slab count into the halo config and have the twin call this helper now, or name that fourth consumer and the guard here.

models/demos/gpt_oss_d_p/tests/unit/test_bounded_ring_plan_math.py
```diff
@@ -0,0 +1,378 @@
+# SPDX-FileCopyrightText: © 2026 Tenstorrent USA, Inc.
+# SPDX-License-Identifier: Apache-2.0
+
+"""
+CPU-only unit test for the bounded sliding-window ring-read plan math. No device, no ttnn ops.
+
+``_plan`` below is a LINE-FOR-LINE python mirror of the C++ single source of truth
```
philei-tt, line 7: A line-for-line Python mirror tests the mirror, not what the kernels run; drift only shows up on a galaxy PCC run. The header is already host-compiled and only includes `<array>`/`<cstdint>`, and `tests/ttnn/unit_tests/gtests/ccl/test_ccl_helpers.cpp` is an existing device-less gtest pinning a kernel constexpr helper. A `test_sliding_window_work_plan.cpp` gtest (or binding the two plan functions in `sdpa_nanobind.cpp`, which this PR already touches, so these pytest properties run against the real function) would remove the lockstep obligation.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_program_factory.cpp
```diff
@@ -39,6 +40,35 @@ namespace {
...
+               : 0;
+}
+
+// Compile-time lockstep checks for sliding_window_work_plan.hpp. The header is constexpr and
+// shared by this factory, the reader, and the compute kernel; a behavioral change that breaks
+// these asserts fails the HOST build instead of silently drifting from the CPU mirror in
+// models/demos/gpt_oss_d_p/tests/unit/test_bounded_ring_plan_math.py (which generated these
+// expected values). Geometry: sp=4 ring, 8-tile Q slabs, 4-tile K chunks, 128-token window,
+// 5 chunks written (post-wrap), Q chunk at local tile 0 on device 0.
+constexpr auto kCircularPlanDev0 = ring_joint::build_sliding_q_work_plan(0, 4, 0, 8, 4, 128, 32, 16, 4, 160, 2);
```
philei-tt, line 59: Valuable checks, but twelve output-value static_asserts on one hand-picked geometry in the production factory TU means a legitimate header change breaks the library build for everyone, with only the raw expression as the message. They belong in the gtest suggested on the unit test file, where they can sweep sp/slabs/chunks and carry a diagnostic.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/compute/compute_streaming.hpp
```diff
@@ -2557,6 +2572,21 @@ void sdpa_ring_v2(
         const uint32_t q_k_loop_count = has_sliding_window ? per_q_valid_kv : num_kv_chunks;
         for (uint32_t k_chunk = 0; k_chunk < q_k_loop_count; ++k_chunk) {
             const auto sliding_k_chunk = sliding_q_plan.k_chunk_at(k_chunk);
+            // Bounded circular cache: absolute K-chunk index of this work item from the plan's ranges.
+            // Inverting local cache rows to global positions is ambiguous once several chunk groups
+            // alias one local slab; unbounded caches keep the inversion.
+            uint32_t bounded_global_k_chunk = 0;
+            if constexpr (bounded_kv_cache) {
```
philei-tt, line 2579: This re-walks `source_ranges` with the same loop `k_chunk_at` runs four lines above, and the result is then selected by two separate `bounded_kv_cache ? plan : inversion` ternaries (k_global_start and step_k_start_tile). I get the constraint (a 4th field on `SlidingKChunkRef` perturbs dense codegen), but a header-owned `SlidingQWorkPlan::global_k_chunk_at(work_index)` called only under `if constexpr (bounded_kv_cache)` is never instantiated in unbounded kernels, so it keeps binary identity while giving the walk one owner. Then compute one `k_origin_tile` and reuse it at both sites.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_program_factory.cpp
```diff
@@ -39,6 +40,35 @@ namespace {
 namespace ag_rt = ttnn::ring_attention_all_gather_async_detail;
 namespace ring_joint = ttnn::operations::transformer::sdpa::ring_joint;
 
+// Circular sliding KV slab count derived from the cache/Q geometry (0 = unbounded). Validation
+// already required whole Q-sized slabs and >= 2 of them; both lengths are per-device, so the
+// ratio is unit-free.
+uint32_t derived_kv_slab_count(
```
philei-tt, line 46: The slab count now has two owners: this (local_kv / local_q) and `validate_runtime_patched_scalars` (cache_capacity / chunk_capacity with a cancelling ring factor). A `kv_slab_count()` on `RingJointSDPAInputs` next to `local_kv_seq_len()`/`is_chunked()` serves both. The new `cache_capacity % chunk_capacity` TT_FATAL also duplicates the pre-existing `N_local_kv % N_local_q` check that always runs in this mode; keep only the `n_slabs >= 2` one.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_device_operation_types.hpp
```diff
@@ -93,6 +101,8 @@ struct RingJointSDPAParams {
 
     bool has_sliding_window() const { return sliding_window_size.value_or(0) > 0; }
 
+    bool has_bounded_kv() const { return circular_kv_cache; }
```
philei-tt, line 104: Zero-logic alias for a public bool, and the diff uses both spellings (`has_bounded_kv()` at three sites, raw `circular_kv_cache` at two). The feature also carries two stems end to end: API/Python/tests say *circular*, kernels/header/layout say *bounded*. Drop the accessor and pick one stem so a single grep finds the feature.

models/demos/gpt_oss_d_p/tt/attention/dense_sp.py
```diff
@@ -72,8 +73,19 @@ def dense_sp_attention(
...
+    circular_kv_cache the sliding cache is a circular buffer of whole chunk slabs; the ring read
+                      wraps its local slab addressing (chunk group g in slab g mod n_slabs, slab count
+                      derived on-device from the cache/Q geometry). False on full layers / unbounded
+                      caches. kv_actual and logical_n stay TRUE ABSOLUTE lengths either way.
     -> out            [1, n_q_local, chunk_local, head_dim]  block-cyclic over the chunk
     """
+    assert (
+        not circular_kv_cache or sliding_window_size is not None
+    ), "circular_kv_cache is only meaningful for sliding-window layers (got sliding_window_size=None)"
+    assert not (circular_kv_cache and write_chunk), (
```
philei-tt, line 85: This assert exists because there are now two KV writers with different wrap laws: the inline `write_chunk` branch below uses the absolute `kv_actual`, `write_kv_chunk` applies the modulo. The only production caller passes `write_chunk=False`, so the guarded branch is dead. Deleting the `write_chunk` parameter and branch removes the second writer and the need for this assert. (The sibling `sliding_window_size is not None` assert duplicates the op's TT_FATAL.)

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
```diff
@@ -806,6 +806,28 @@ def to_balanced_growing_cache_layout(src_full, sp_size, chunk_size, last_uploade
     return perm
 
 
+def to_circular_two_slab_cache_layout(src_full, sp_size, chunk_size, chunk_idx):
```
philei-tt, line 809: `src_full[:, :, bounded_blockcyclic_positions(sp_size, chunk_size, 2*chunk_size, (chunk_idx+1)*chunk_size), :]` is bit-identical to this for chunk_idx >= 1 (the file already imports from deepseek_v3_d_p). Hard-coding `n_slabs = 2` here while the model derives it from `min(2*max_chunk, max_seq_len)` means a capacity change in the model leaves this test passing on a layout the model no longer produces. Also the K/V selection at the two call sites is the same if/else written twice; pick the layout fn once above the loop.

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
```diff
@@ -5732,6 +5774,137 @@ def test_ring_joint_attention_gpt_oss_chunked_sliding_production_accuracy():
...
+            else:
+                assert runtime.mesh_device.num_program_cache_entries() == cache_entries, (
+                    "second circular-cache chunk (other slab parity) must hit the cached program, not "
+                    "create a new one"
+                )
+    finally:
+        close_ring_joint_sdpa_runtime(runtime)
+
+
+def test_ring_joint_circular_kv_cache_rejects_metadata_path(expect_error):
```
philei-tt, line 5815: ~90 lines to assert one host-side TT_FATAL that fires before any value is inspected; the scaffolding is copied from the existing metadata test (~L4575) and the trailing positive call asserts nothing the circular accuracy test doesn't already cover with a PCC. Zeros of the right shapes plus a shared fixture with that test brings this to ~25 lines.

models/demos/gpt_oss_d_p/tests/unit/test_bounded_ring_plan_math.py
```diff
@@ -0,0 +1,378 @@
...
+``build_sliding_q_work_plan`` with the ``bounded_kv_slab_count`` wrap and the
+``first_global_k_chunk`` range field). That header is compiled into the HOST program factory, the
+ring READER kernel, and the COMPUTE kernel — any change to it MUST be mirrored here in lockstep
+(and vice versa), exactly like ring_joint_kv_pad_derivation.hpp's host/device copies.
+
+What is verified, cross-checked against the production un-rotation helper
+``bounded_blockcyclic_positions`` (kv_cache.py — the same helper the host-readback PCC instrument
+uses, itself validated against a write simulation in test_bounded_kv_math.py):
+
+  1. POSITION IDENTITY: for every chunk 0..N-1 (chunk 0 included — #53153 unified the first group
```
philei-tt, line 19: nit: the `#53153` tags here and at L83/145/198/293 record which PR introduced each constraint; the mirrored C++ header states the same constraints tag-free. Keep the constraint text, drop the tags. Same for `dense_sp.py:158` (restates the parameter) and the "kept beside the capacity check they replace" wording in `ring_joint_sdpa_device_operation.cpp`, which is diff-relative.

philei-tt (approved):
Approving. The mechanism is sound: one constexpr helper wraps the local slab, reader/compute/host halo layout all go through it, the plan carries absolute origins so the mask no longer inverts local rows, alignment is enforced by the existing ring-group checks, the hash covers the flag, and the slot renumbering is consistent (only one factory feeds these kernels; the Quasar copy of the compute header stays unbounded via the default).

Comments inline. The first one (a fourth consumer of the slab math on the metadata path that did not get the wrap) is the one I'd fix before the trace follow-up. The rest are quality cleanups; the header gtest suggestion would replace both the Python mirror and the factory static_asserts.

ayerofieiev-tt (approved):

ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/sliding_window_work_plan.hpp
mdragulaTT: Named it at the guard and on both helpers: `compute_halo_tail_start_Ht` stays unwrapped and circular caches are rejected on the metadata path; plumbing the slab count into the halo config is the trace follow-up (#56115).

models/demos/gpt_oss_d_p/tests/unit/test_bounded_ring_plan_math.py
mdragulaTT: Replaced by `tests/ttnn/unit_tests/gtests/sdpa/test_sliding_window_work_plan.cpp` against the real header: slab wrap, circular == unbounded up to the slab wrap, absolute origins, remote ranges inside the halo, work-item walk, plus the pinned dev0 geometry. Mirror deleted.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_program_factory.cpp
mdragulaTT: Moved into that gtest (the pinned geometry is one test there); the factory no longer carries them.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/kernels/compute/compute_streaming.hpp
mdragulaTT: Done as suggested: `SlidingQWorkPlan::global_k_chunk_at` in the header, called once under `if constexpr (circular_kv_cache)` into one `circular_k_origin_tile` that both sites use. The dense kernel stays instruction-identical to main on all three triscs.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_program_factory.cpp
mdragulaTT: Done: `RingJointSDPAInputs::kv_slab_count()` next to `local_kv_seq_len()`, used by the factory and the validation; the duplicate whole-slab TT_FATAL is gone, only `n_slabs >= 2` remains.

ttnn/cpp/ttnn/operations/transformer/sdpa/device/ring_joint_sdpa_device_operation_types.hpp
mdragulaTT: Accessor dropped; kernels, header and halo layout renamed to the `circular` stem (`circular_kv_slab_count`, `circular_kv_local_slab`, `circular_kv_cache`). The model-side allocation flag keeps its merged name.

models/demos/gpt_oss_d_p/tt/attention/dense_sp.py
mdragulaTT: Done: `write_chunk` parameter and branch deleted, `dense_sp_attention` no longer takes the chunk K/V, both asserts gone.

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
mdragulaTT: Done: the layout now comes from `sliding_capacity_tokens` + `bounded_blockcyclic_positions`, picked once above the loop.

tests/nightly/blackhole/sdpa/test_ring_joint_sdpa.py
mdragulaTT: Done: zeros of the right shapes, ~35 lines, no positive call.

models/demos/gpt_oss_d_p/tests/unit/test_bounded_ring_plan_math.py
mdragulaTT: Done: the mirror is gone; the parameter-restating comment in dense_sp.py and the diff-relative wording in the device op are removed.

PR #54938, opened 2026-08-31
[LLK] Reduce instruction count in some SFPU kernels

ldjurovicTT:
### PR Category
Performance

### Summary
Accuracy-neutral SFPU instruction-count reductions on **Wormhole and Blackhole**, plus a doc
recording the cost model, the per-architecture differences and the traps.

Three ideas, all bit-exact: park hot fp32 poly coefficients in the program CREGs (§1), evaluate two
independent chains in lockstep (§2), and replace `v_if` predicate blocks with branch-free idioms
(§3). Everything below is `MATH_ISOLATE` cycles from the tt-llk perf suite (Float16_b,
`tile_cnt=8`, `loop_factor=16`, `iterations=32`), each op A/B'd against the same tree with only
that change reverted and the build root wiped between variants.

#### Wormhole n300

| kernel | change | KERNEL before → after | delta |
|---|---|---|---|
| `relu_max` (LLK) | two `v_if` predicate blocks → branch-free `max(min(x,t),0)` | 62850 → **38205** | **−39.2 %** |
| `Hardsigmoid` (Metal) | §3, free rider via `_relu_max_body_` | 61883 → **37052** | **−40.1 %** |
| `i0` (Metal) | poly coefficients → CREGs, + 2 elements per iteration hand-interleaved | 181428 → **138391** | **−23.7 %** |
| `softplus` (Metal) | 3 fp32 poly coefficients → `vConstFloatPrgm0/1/2` | 198231 → **185561** | **−6.4 %** |
| `reciprocal_compat` (LLK) | conditional negate → `copysgn` | 139707 → **135610** | **−2.9 %** |
| `rsqrt_compat` (LLK) | same change, shared call site | 160181 → **156085** | **−2.6 %** |
| `i1` (Metal) | interleave co-resident `numer`/`denom` Horner chains | 548206 → **538323** | **−1.8 %** |

Refreshed after the review round: every Wormhole row above was re-measured in one session against
`origin/main` with all of this PR's changes reverted at once, rather than one change at a time.
`Square` was carried as a control and read 25018 → 25018 exactly. `reciprocal_compat` and
`rsqrt_compat` are quoted at `approx_mode:Yes`; their `approx_mode:No` paths improve too, by
−2.4 % and −1.2 %. `i0` and `i1` now read ~0.3 % higher on **both** sides of the A/B than in the
first round — the baseline moved with them, so the deltas stand; every SFPU header lands in the
same math ELF via the `SfpuType` dispatch switch, so unrelated edits shift code layout for the
largest bodies. Prefer the deltas to the absolutes for those two.

#### Blackhole p100a

| kernel | change | KERNEL before → after | delta | INIT |
|---|---|---|---|---|
| `relu_max` (LLK) | §3 branch-free clamp | 53167 → **32815** | **−38.3 %** | 170 → 171 |
| `Hardsigmoid` (Metal) | §3, free rider via `_relu_max_body_` | 61615 → **36782** | **−40.3 %** | 185 → 185 |
| `i0` (Metal) | §1 CREGs + §2 interleave | 167855 → **130987** | **−22.0 %** | 171 → 195 |
| `softplus` (Metal) | §1 CREGs + §2 interleave | 192177 → **155310** | **−19.2 %** | 170 → 188 |
| `softplus` dest_acc (Metal) | same | 188082 → **149164** | **−20.7 %** | 170 → 195 |

`TILE_LOOP` tracks `KERNEL` within 0.1 % on both architectures. `Square` was carried as a control
in every Blackhole run and read 20398 → 20398 exactly (25018 → 25018 on Wormhole). On Wormhole, `relu_max`'s `L1_TO_L1` also
fell 63720 → 38935.

Static, per architecture: on Wormhole `relu_max` 14 → 8 SFPU instrs, `i0` 38.0 → 31.5
slots/element, `rsqrt_compat` `SFPSETCC` 17 → 9. On Blackhole `relu_max` 11 → 6 slots/element, and
`i0` §2 holds the count flat at ~28.5 slots/element while halving its dependency stalls.

`Hardsigmoid` is a free rider — it calls `_relu_max_body_`, so §3 reaches it with no separate
change. `softplus` §2 is applied on Blackhole first and is still open on Wormhole; the rewrite is
arch-independent, so it should transfer.

### Correctness and bit-exactness

All changes are bit-exact. CREG parking keeps the exact fp32 coefficient, interleaving only
reorders independent chains, and the SFPU compare shares one sign-magnitude total order with
`SFPSWAP` and `SFPSETSGN` — so the clamp and `copysgn` rewrites preserve NaN and ±0 too.

- **Wormhole:** 527 passed / 9 skipped for `{I0, Softplus, ReluMax, Hardsigmoid, Silu}`, 32 for
  `{RsqrtCompat, ReciprocalCompat, SqrtCompat}`, 8 for `I1` — including the fp32 `_edges` variants
  that feed NaN, ±inf and −0.0. Skips are the pre-existing ReluMin gate (tt-llk#1120).
- **Blackhole:** 318 passed / 234 skipped for
  `{I0, Softplus, ReluMax, ReluMin, Hardsigmoid, Lrelu}`, and 65/65 in `test_sfpu_sdpa.py` — which
  calls `calculate_softplus_body` directly and so exercises §1's CREG contract through a second
  init path. Outcomes were compared **per test ID** against the unmodified tree and are identical,
  so the 234 skips are pre-existing.
- **Bit-exactness measured, not just argued (Blackhole):** all **308** result buffers, across every
  format pair including the Float32→Float32 `_edges` cases carrying NaN, ±inf and ±0.0, are
  **byte-identical** to the unmodified tree. Method: wrap `TestConfig.run`, sha256
  `outcome.result`, key by node id, diff the hash files. The golden comparison is tolerance-based
  and cannot distinguish bit-exact from merely-within-tolerance — worth doing routinely, and it is
  what actually validates the `softplus` predication hoist below.
- **Full-sweep regression check (Blackhole):** the whole 1154-variant unary SFPU perf sweep on both
  trees, 1152 passed / 2 skipped each. Of 3456 matched rows **3400 are within ±0.5 %**, and **no
  `KERNEL` or `TILE_LOOP` row outside the intended ops moved at all** — the rest are `INIT` rows
  shifting ≤6 cycles out of ~170–230.

### Notes for reviewers

1. **Clamp bound order matters.** The original applied the high bound first, so a negative
   threshold must yield 0. `max(min(x,t),0)` reproduces that; `min(max(x,0),t)` does not.
2. **`relu_max`'s `vInt` path is deliberately left predicated.** `sfpi::min`/`max` cover
   `vFloat`/`vSMag` only on WH/BH — the `vInt` overloads are gated on `__riscv_xtttensixqsr`.
3. **The interleaves are written statement by statement on purpose.** The sfpi builtins are
   ordered side effects, so GCC will not interleave two independent chains — calling `POLYVAL10`
   twice back to back measured no per-element change at all.
4. **§2 normally cannot cross a `v_if`, and `softplus` goes through it anyway.** The
   condition-code state is one shared resource, so two elements taking different branches
   serialise their predicated blocks. `softplus` puts its whole polynomial inside
   `v_if (t <= threshold)`. The way through is not to interleave inside the guard but to lift the
   chain out and predicate only the store — SFPU predication masks the destination *write*, not the
   execution, so evaluating on all lanes and discarding on the `t > threshold` lanes leaves them
   holding the `dst_reg` value they would have kept anyway. Dependent-adjacent pairs fall 12 → 6.5
   per element with loop `cc` unchanged. This is the one step in the PR that is an argument rather
   than a mechanical transform, which is why the byte-comparison above was run.
5. **`softplus` §2 is bf16-path only.** The fp32 path's exp tail is a second, deeper chain that
   does not fit alongside a second element's live values, and the perf sweep drives Float16_b, so
   an fp32 interleave could not be measured in this harness. `calculate_softplus_body` is left
   untouched, so the SDPA consumer of it is unaffected.
6. **`rsqrt_compat`'s rewrite** relies on `_reciprocal_compat_` always returning a non-negative
   magnitude (`setsgn` forces `val` negative, the Newton iteration stays positive, `setexp`
   preserves the sign, and the pole cases are `+0`/`+inf`), which makes the conditional negate
   exactly `copysgn`.

Also adds `numer` to `tt_metal/tt-llk/.codespellignore`: it is established vocabulary here, but the
codespell hook in `tt_metal/tt-llk/.pre-commit-config.yaml` has no `files:` filter and was
rewriting `numer` → `number` across the kernel and its pre-existing comments.

### Blackhole differs from Wormhole in mechanism, not outcome

Worth knowing before picking up the next kernel, because the doc's §2 candidate table does not
transfer as written:

- **There is essentially no `SFPNOP` tax on Blackhole to remove.** Scanning the largest math ELF
  for each of 103 unary SFPU ops (5760 ELFs, full build) finds **145** `SFPNOP` in total, all in
  six hand-written TTI kernels — `exponential` 33, `exp_with_base` 33, `topk_local_sort` 32,
  `topk_rebuild` 12, `reciprocal` 9, `erf` 1. **97 of 103 ops carry zero**, and every
  compiler-scheduled polynomial kernel is at 0. Ranking by `SFPNOP` share ranks the whole Blackhole
  candidate list at zero.
- **The replacement signal is dependent-adjacency count** — pairs where instruction *i+1* reads a
  register instruction *i* wrote. It also separates the two levers cleanly: `relu_max` §3 is 14 → 9
  instructions with no change in dependent pairs (a slot win), while `i0` §2 holds the instruction
  count flat at ~28.5 slots/element and halves the pairs 15 → 8 (a stall win, −15.8 %).
- **`polygamma` ranks *ideal* on WH (14 stalls, 0 `cc`) and *poor* on BH (124 pairs, 56 `cc`).** The
  two architectures compile it differently enough that the feasibility verdict flips — read the
  target architecture's own scan rather than carrying a verdict across. `power` tops the Blackhole
  list (616 pairs) and was absent from the Wormhole survey entirely.
- **"Issue slots are free on Blackhole" is corrected as too broad.** It came from a load/store-bound
  replay experiment where removing 69 instructions per tile bought zero cycles, and does not
  generalise to compute-dense polynomial bodies — §1 removed exactly 6 slots/element from `i0` and
  was worth −7.3 %.

The doc also records the dead ends: CREG parking saves ~1 slot per coefficient on WH rather than 2
(the loads were filling `SFPMAD` latency), interleaving needs stalls *and* low predication *and*
spare registers (`sigmoid` and `gelu_tanh` regressed on the third), `SFPMOV` counts are predication
artifacts rather than register spills, 3-way interleave spills on both architectures, and the two
scan methodology traps — a partial build, and taking the first rather than the largest ELF per op —
the second of which caught the Blackhole scan too.

### Not done yet
- `rsqrt_compat` / `reciprocal_compat` §3 is Wormhole-only; the Blackhole port is the obvious next
  step, as is `i1`'s chain interleave.
- `softplus` §2 has not been applied on Wormhole.
- The two remaining two-sided clamp sites in the Metal `ckernel_sfpu_relu.h` are still open on both
  architectures; one of them takes `min(max(..))` because it applies its low bound first.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ldjurovic/sfpu_instr_count_reduction)

### DIFF
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
@@ -13,16 +13,6 @@ using namespace sfpi;
 namespace ckernel {
 namespace sfpu {
 
-#define POLYVAL10(coef10, coef9, coef8, coef7, coef6, coef5, coef4, coef3, coef2, coef1, coef0, t4)               \
-    ((coef0 +                                                                                                     \
-      (coef1 +                                                                                                    \
-       (coef2 +                                                                                                   \
-        (coef3 +                                                                                                  \
-         (coef4 + (coef5 + (coef6 + (coef7 + (coef8 + (coef9 + coef10 * t4) * t4) * t4) * t4) * t4) * t4) * t4) * \
-            t4) *                                                                                                 \
-           t4) *                                                                                                  \
-          t4) *                                                                                                   \
-     t4)
 // Park the three costliest series coefficients in the SFPU's programmable constant
 // registers: SFPMAD names a CREG directly in an operand field, so each costs zero
 // instructions per element instead of the two SFPLOADI a full-fp32 literal needs, and the
@@ -39,9 +29,10 @@ inline void i0_init() {
 
 // Horner step of the I0 series, applied to two independent accumulators in lockstep.
 // The steps must alternate at STATEMENT level: GCC will not interleave two independent
-// expression trees on its own (measured -- writing the two POLYVAL10 calls back to back
-// leaves all 9 SFPNOP per element in place). Coefficients ascend exactly as POLYVAL10
-// consumed them, so each element's chain keeps its operations, operands and order: bit-exact.
+// expression trees on its own (measured -- writing the two series out as two back-to-back
+// Horner expressions leaves all 9 SFPNOP per element in place). Coefficients ascend in the
+// order the series is evaluated, so each element's chain keeps its operations, operands and
+// order: bit-exact.
 #define I0_STEP2(c)         \
     do {                    \
         r0 = r0 * t0 + (c); \
@@ -77,7 +68,7 @@ inline void calculate_i0() {
         I0_STEP2(sfpi::vConstFloatPrgm0);  // 0.0004340277778f,   parked by i0_init
         I0_STEP2(0.015625f);
         I0_STEP2(0.25f);
-        r0 = r0 * t0;  // POLYVAL10's trailing multiply by t4
+        r0 = r0 * t0;  // the series has no constant term: trailing multiply by t
         r1 = r1 * t1;
 
         dst_reg[0] = 1.0f + r0;
@@ -86,7 +77,11 @@ inline void calculate_i0() {
         dst_reg += 2;
     }
 
-    // Odd ITERATIONS: finish the trailing element on its own.
+    // Odd ITERATIONS: finish the trailing element on its own. Written out rather than handed
+    // to PolynomialEvaluator::eval: eval builds `c + t * inner`, which compiles to the same
+    // 37 (WH) / 28 (BH) instructions but with the SFPMAD multiplicands commuted (`t * r`
+    // instead of `r * t`) on every step. Commuting a multiply is IEEE-safe, but this kernel's
+    // guarantee is byte-identical output, so the chain stays as the interleaved path writes it.
     if constexpr (ITERATIONS % 2 != 0) {
         vFloat in = dst_reg[0];
         vFloat t = in * in;
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
@@ -23,7 +23,19 @@ namespace sfpu {
            t4) *                                                                                                  \
           t4) *                                                                                                   \
      t4)
-inline void i0_init() { math::reset_counters(p_setrwc::SET_ABD_F); }
+// Park the three costliest series coefficients in the SFPU's programmable constant
+// registers: SFPMAD names a CREG directly in an operand field, so each costs zero
+// instructions per element instead of the two SFPLOADI a full-fp32 literal needs, and the
+// value keeps its exact fp32 bit pattern (bit-exact, not an accuracy trade). The other
+// coefficients stay literals -- 0.25f and 0.015625f are already bf16-exact and free, and
+// on Wormhole a dependent Horner chain hides some load latency, so draining every load
+// out of the chain trades loads for SFPNOPs rather than removing them outright.
+inline void i0_init() {
+    math::reset_counters(p_setrwc::SET_ABD_F);
+    sfpi::vConstFloatPrgm0 = 0.0004340277778f;
+    sfpi::vConstFloatPrgm1 = 0.000006781684028f;
+    sfpi::vConstFloatPrgm2 = 6.78E-08f;
+}
 
 template <bool APPROXIMATION_MODE, int ITERATIONS = 8>
 inline void calculate_i0() {
@@ -41,9 +53,9 @@ inline void calculate_i0() {
                             9.39E-15f,
                             2.40E-12f,
                             4.71E-10f,
-                            6.78E-08f,
-                            0.000006781684028f,
-                            0.0004340277778f,
+                            sfpi::vConstFloatPrgm2,  // 6.78E-08f,          parked by i0_init
+                            sfpi::vConstFloatPrgm1,  // 0.000006781684028f, parked by i0_init
+                            sfpi::vConstFloatPrgm0,  // 0.0004340277778f,   parked by i0_init
                             0.015625f,
                             0.25f,
                             x);
--- a/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
+++ b/tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
@@ -37,30 +37,71 @@ inline void i0_init() {
     sfpi::vConstFloatPrgm2 = 6.78E-08f;
 }
 
+// Horner step of the I0 series, applied to two independent accumulators in lockstep.
+// The steps must alternate at STATEMENT level: GCC will not interleave two independent
+// expression trees on its own (measured -- writing the two POLYVAL10 calls back to back
+// leaves all 9 SFPNOP per element in place). Coefficients ascend exactly as POLYVAL10
+// consumed them, so each element's chain keeps its operations, operands and order: bit-exact.
+#define I0_STEP2(c)         \
+    do {                    \
+        r0 = r0 * t0 + (c); \
+        r1 = r1 * t1 + (c); \
+    } while (0)
+
 template <bool APPROXIMATION_MODE, int ITERATIONS = 8>
 inline void calculate_i0() {
+    // Two elements per iteration, hand-interleaved. The series is a fully dependent Horner
+    // chain, so on Wormhole every step whose coefficient is a CREG or an immediate stalls one
+    // cycle on the previous SFPMAD -- 9 SFPNOP per element, 24% of the loop body. Two
+    // independent chains fill those slots, and one TTINCRWC covers both elements.
+    //
+    // Register pressure: 2 x (t, r) plus the two loads = 6 of the 8 general LRegs, with the
+    // coefficients in CREGs. A third interleaved element spills.
 #pragma GCC unroll 0
 
-    for (int d = 0; d < ITERATIONS; d++) {
-        vFloat result = 0.0f;
-        vFloat input = dst_reg[0];
-        vFloat x = input * input;
+    for (int d = 0; d < ITERATIONS / 2; d++) {
+        vFloat in0 = dst_reg[0];
+        vFloat in1 = dst_reg[1];
+        vFloat t0 = in0 * in0;
+        vFloat t1 = in1 * in1;
 
-        result = 1.0f + POLYVAL10(
-                            1.50E-22f,
-                            7.24E-20f,
-                            2.90E-17f,
-                            9.39E-15f,
-                            2.40E-12f,
-                            4.71E-10f,
-                            sfpi::vConstFloatPrgm2,  // 6.78E-08f,          parked by i0_init
-                            sfpi::vConstFloatPrgm1,  // 0.000006781684028f, parked by i0_init
-                            sfpi::vConstFloatPrgm0,  // 0.0004340277778f,   parked by i0_init
-                            0.015625f,
-                            0.25f,
-                            x);
+        vFloat r0 = 1.50E-22f;
+        vFloat r1 = 1.50E-22f;
+        I0_STEP2(7.24E-20f);
+        I0_STEP2(2.90E-17f);
+        I0_STEP2(9.39E-15f);
+        I0_STEP2(2.40E-12f);
+        I0_STEP2(4.71E-10f);
+        I0_STEP2(sfpi::vConstFloatPrgm2);  // 6.78E-08f,          parked by i0_init
+        I0_STEP2(sfpi::vConstFloatPrgm1);  // 0.000006781684028f, parked by i0_init
+        I0_STEP2(sfpi::vConstFloatPrgm0);  // 0.0004340277778f,   parked by i0_init
+        I0_STEP2(0.015625f);
+        I0_STEP2(0.25f);
+        r0 = r0 * t0;  // POLYVAL10's trailing multiply by t4
+        r1 = r1 * t1;
 
-        dst_reg[0] = result;
+        dst_reg[0] = 1.0f + r0;
+        dst_reg[1] = 1.0f + r1;
+
+        dst_reg += 2;
+    }
+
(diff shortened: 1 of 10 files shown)


tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
```diff
@@ -23,32 +23,85 @@ namespace sfpu {
...
-        dst_reg[0] = result;
+    for (int d = 0; d < ITERATIONS / 2; d++) {
+        vFloat in0 = dst_reg[0];
+        vFloat in1 = dst_reg[1];
+        vFloat t0 = in0 * in0;
+        vFloat t1 = in1 * in1;
+
+        vFloat r0 = 1.50E-22f;
+        vFloat r1 = 1.50E-22f;
+        I0_STEP2(7.24E-20f);
```
fvranicTT, line 61: Can we use PolynomialEvaluator instead?

ldjurovicTT: Not for the interleaved chain — that's actually the reason the macro exists.

`PolynomialEvaluator::eval` returns a single expression tree (`coeff0 + x * eval(x, rest...)`). Two
calls give two *independent* trees, and GCC will not interleave those. I measured it: writing the
two chains back to back left all 9 `SFPNOP` per element in place — 37.5 vs 38.0 slots/element, i.e.
only the `TTINCRWC` amortising. The win appears solely when the two chains alternate at **statement**
level, which is what `I0_STEP2` forces and what an expression-tree evaluator can't express.

Two things your comment did surface, though, and I'll fold both in:

1. `POLYVAL10` is now dead code — the interleave replaced its only call site, and nothing else in
   the repo uses it. It should go.
2. The odd-`ITERATIONS` scalar tail *is* a plain Horner chain, so it can use
   `PolynomialEvaluator::eval` bit-exactly (same ascending coefficient order):
   `1.0f + t * eval(t, 0.25f, 0.015625f, Prgm0, Prgm1, Prgm2, 4.71E-10f, ..., 1.50E-22f)`.

If you'd rather the two-chain form were library code than a per-kernel macro, the right home is an
`eval2(x0, x1, coeffs...)` in `ckernel_sfpu_polyval.h` — a variadic template emitting alternating
statements (statement order survives even though expression trees don't), which would replace both
`I0_STEP2` here and `SOFTPLUS_STEP2` in this PR's Blackhole softplus. I'd want to re-measure it
against the macro before trusting it, since "did the compiler actually keep the interleave" is the
entire risk in this class of change. Happy either way — in this PR, or as a follow-up?

fvranicTT: Check PolynomialEvaluator in Quasar. I changed its implementation to have two chains. I just forgot to port it to BH/WH.

ldjurovicTT: Correcting point 2 above — I tried it and backed it out, which turns out to be the more useful
answer to your question.

`PolynomialEvaluator::eval` builds `coeff0 + x * inner`, i.e. the *variable* on the left of the
multiply. The hand-written chain is `r = r * t + c`, accumulator on the left. Compiling both forms
of the scalar tail side by side (sfpi 7.71, `-O3`, `-mcpu=tt-wh-tensix` and `tt-bh-tensix`):

- identical instruction count — 37 on WH, 28 on BH — and identical opcodes,
- but every step comes out `SFPMAD L0, L1, L0, L2` instead of `SFPMAD L0, L0, L1, L2`: the two
  multiplicands are swapped.

Commuting a multiply is IEEE-safe and I'd normally not care. Here I do: this PR's correctness claim
is byte-identical result buffers against the unmodified tree (all 308 of them, including the fp32
`_edges` cases that feed NaN), and NaN-payload selection between two NaN operands is exactly the
case where `a*b` and `b*a` are allowed to differ. Zero instructions saved, non-zero risk to the one
guarantee the PR is making — so the tail stays hand-written, with a comment recording why.

Point 1 stands and is done: `POLYVAL10` is deleted from both the Wormhole and Blackhole `i0`.

This also sharpens the `eval2(x0, x1, coeffs...)` idea: to be a drop-in for `I0_STEP2` /
`SOFTPLUS_STEP2` it would need to emit `r = r * x + c` (accumulator-left), not eval's operand
order. Worth doing, still wants measuring, and still better as a follow-up than as more scope here.

fvranicTT: Can we replace the original eval with this new approach? That would bring benefits to all kernels that are using PoynomialEvaluator.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_softplus.h
```diff
@@ -93,7 +93,31 @@ sfpi_inline sfpi::vFloat softplus_exp_negative(sfpi::vFloat x) {
...
+// production (SFPU_UNARY_INIT) and the tt-llk harness run before the calculate step.
+#ifdef INP_FLOAT32
+#define SOFTPLUS_CREG_C0 SOFTPLUS_POLY_C0
+#define SOFTPLUS_CREG_C1 SOFTPLUS_POLY_C1
+#define SOFTPLUS_CREG_C2 SOFTPLUS_POLY_C2
+#else
+#define SOFTPLUS_CREG_C0 SOFTPLUS_BF16_POLY_C0
+#define SOFTPLUS_CREG_C1 SOFTPLUS_BF16_POLY_C1
+#define SOFTPLUS_CREG_C2 SOFTPLUS_BF16_POLY_C2
+#endif
```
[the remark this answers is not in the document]
ldjurovicTT: Applied in f0e025b — the macros are gone and `softplus_init` now carries the
`#ifdef INP_FLOAT32` inline, matching what `calculate_softplus_body` already does a few lines
down. Done on the Blackhole copy of the file too, which had picked up the same pattern.

tt_metal/tt-llk/tt_llk_wormhole_b0/common/inc/sfpu/ckernel_sfpu_relu.h
```diff
@@ -58,16 +53,28 @@ inline void _relu_max_impl_(const int iterations, VecType threshold)
...
+            // See _relu_max_body_: branch-free clamp, high bound applied first.
+            result = sfpi::max(sfpi::min(result, threshold), 0.0f);
         }
-        v_endif;
-        v_if (result < 0)
+        else
         {
-            result = 0;
+            // sfpi::min/max cover vFloat and vSMag only on Wormhole/Blackhole (the vInt
+            // overloads are gated on __riscv_xtttensixqsr), and SFPSWAP orders operands as
```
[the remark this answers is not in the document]
ldjurovicTT: Grounded in f0e025b. The claim checks out, and you were right that it wasn't verifiable from the
repo — by construction.

`min(vInt, int)`, `max(vInt, int)` and the `is_base_of<vInt, Type>` arm of `min_max`'s `enable_if`
all sit inside `#if __riscv_xtttensixqsr` in sfpi's `include/sfpi_lib.h`. Verified against both
7.71.0 (the copy the tt-llk tests build against, at `tt_metal/tt-llk/tests/sfpi/`) and 7.72.0 (the
release `tt_metal/sfpi-version` pins).

The reason your grep came back empty is that the header ships with the toolchain and is gitignored
(`tt_metal/tt-llk/tests/.gitignore:2`), so it is on every developer's disk after a test build but
never in a diff. The comment now says all of that, so the next reader doesn't repeat the search.

Also added, since it is the obvious next question: `sfpi::clamp()` exists but is defined as
`min(max(val, lower), upper)` — the bound order this kernel specifically must not use.

tt_metal/hw/ckernels/wormhole_b0/metal/llk_api/llk_sfpu/ckernel_sfpu_i0.h
ldjurovicTT: You're right about the mechanism — split() emits the two chains as statements, so it does interleave, and my earlier objection was wrong. But measuring it: with fp32 literal coefficients (what almost every WH/BH call site uses today) the split is 2-3 instructions worse on both arches, because each literal's two SFPLOADI already fill the SFPMAD latency — it only wins once the coefficients are free, where WH goes 18 → 17 instructions and BH halves dependent pairs 6 → 3. It's also not bit-exact and ~1.8× less accurate (Estrin accumulates in x²), and at the default threshold it flips 21 of 41 WH call sites — so it can't land in this PR, whose claim is byte-identical buffers.


We can try this ins some follow up pr in detail

Issue #42193, opened 2026-04-15
Migrate operations to descriptors


dgomezTT:
Migrate all current operations to work with descriptors

PR #54939, opened 2026-08-31
[Cleanup] matmul: remove smuggled buffer-address seeds from descriptor runtime args

dgomezTT:
### Problem

Every buffer address the matmul `ProgramDescriptor` factories put in a runtime arg is already declared as a tensor binding via `KernelDescriptor::emplace_runtime_args`, so the framework patches it on a program-cache hit. Ten of those slots were nevertheless seeded with a raw `.address()` value that the binding then overwrote.

The seed is a dead value, but it makes a lost or mis-indexed binding **silent**: the cache-miss call is correct (it uses the raw seed) and only cache hits go stale. The bindings are installed by hand-counted index — e.g. `in1_sender_variant[18] = *bias_tensor` — which is exactly what drifts when the arg layout changes.

Only two of the ten were visible to `scripts/detect_smuggled_rta.py`. The detector matches `push_back(...address())` but not an address inside a braced init list, so the eight in the 1D/2D factories went unreported — matmul looked like 2 hits when it was really 10.

| factory | slots |
| --- | --- |
| `create_program_dram_sharded_descriptor` | in1, bias *(the 2 the detector flagged)* |
| `create_program_mcast_in0_descriptor` (1D) | in0, in1, out, bias |
| `create_program_mcast_in0_in1_descriptor` (2D) | in0, in1, out, out (receiver) |

### Change

Every descriptor-path address slot now holds `0u`. The address comes only from the binding, so a lost binding traps on address 0 instead of decaying into a stale-address bug. No arg ordering, count, or byte changes — pure construction-mechanism cleanup.

Also drops three `// smuggled-rta-ok` markers:
- one sat on a descriptor slot that **is** bound, mislabelling correct code as parked;
- the other two are on legacy `SetRuntimeArgs` paths that the detector no longer flags at all, so they suppressed nothing.

Legacy paths re-patched by `override_*_program_parameters` are left untouched (descriptors-only scope).

### Test

New `test_matmul_cache_hit_patches_addresses` in `tests/ttnn/unit_tests/operations/matmul/test_matmul_program_cache.py`, covering the default, 1D `mcast_in0`, 1D `mcast_in1`, 2D mcast and DRAM-sharded routings.

It keeps the first operand set alive so the cache-hit call's tensors are forced to land at different addresses, asserts they actually moved, then asserts PCC and a single cache entry. A `0u` placeholder that ever survived to the device fails loudly instead of silently — which turns "I read every index and they are all rebound" into a machine-checked invariant.

### Checks

- `scripts/detect_smuggled_rta.py` clean across the whole matmul tree.
- Full `pre-commit run` clean on all four files.

### DIFF
--- a/tests/ttnn/unit_tests/operations/matmul/test_matmul_program_cache.py
+++ b/tests/ttnn/unit_tests/operations/matmul/test_matmul_program_cache.py
@@ -103,3 +103,201 @@ def test_matmul_cache_miss_different_dtype(device, isolate_program_cache):
     assert (
         count == 2
     ), f"different dtype (bfloat16 vs float32) must produce 2 distinct cache entries, got {count} (dtype not keyed)"
+
+
+# ---------------------------------------------------------------------------
+# Cache-hit address patching (smuggled-pointer regression net)
+#
+# Every buffer address a matmul descriptor factory puts in a runtime arg is
+# declared as a tensor binding (KernelDescriptor::emplace_runtime_args); the arg
+# vectors hold a 0 placeholder in that slot. The framework patches bindings on a
+# program-cache hit, so a slot that lost its binding would keep the placeholder
+# (address 0) or a stale address from the cache-miss call. These tests force the
+# second call's tensors to land at DIFFERENT addresses than the first, so a
+# missing binding shows up as wrong data instead of passing by luck.
+# ---------------------------------------------------------------------------
+
+L1_WIDTH_SHARDED_3_CORES = ttnn.MemoryConfig(
+    memory_layout=ttnn.TensorMemoryLayout.WIDTH_SHARDED,
+    buffer_type=ttnn.BufferType.L1,
+    shard_spec=ttnn.ShardSpec(
+        ttnn.CoreRangeSet({ttnn.CoreRange(ttnn.CoreCoord(0, 0), ttnn.CoreCoord(2, 0))}),
+        (32, 32),
+        ttnn.ShardOrientation.ROW_MAJOR,
+    ),
+)
+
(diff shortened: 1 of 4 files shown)


dgomezTT:
Closing — the premise was wrong.

The matmul descriptor factories were never smuggling: every address slot there already has a registered `BufferBinding`, so the framework knows it is a pointer. The raw `.address()` seed this PR replaced is a dead value the binding overwrites. The change was cosmetic churn on non-problems.

The genuine smuggles in matmul are all on legacy `SetRuntimeArgs` paths, where no `BufferBinding` API exists at all:

- `process_gather_in0_program_...` — in1, reachable from `ttnn.matmul` via `gather_in0`
- legacy 1D mcast_in0/mcast_in1 (9 slots) and legacy 2D (5 slots) — reachable only from `all_gather_matmul_async`, `llama_reduce_scatter_matmul`, `matmul_reduce_scatter_async`, which hand matmul a `Program&` plus a `MatmulFusedOpSignaler`
- sparse matmul factory (~6 slots)

All are functionally correct today (re-patched by `override_*_program_parameters`). Because a binding only exists on the descriptor side, "de-smuggle" and "migrate the path off legacy" are the same action — so these disappear with the legacy-path migration (#42193), not with a separate smuggle fix. Tracking them there rather than duplicating ~785 lines of the ring-matmul path to remove 1 of 15 slots.

Separately noted for the guard: `scripts/detect_smuggled_rta.py` matches `push_back(...address())` but not an address inside a braced init list, so it reported 2 sites in matmul where the pattern actually appears 10 times. Not a correctness bug (all 10 are bound), but the detector undercounts.

PR #54943, opened 2026-08-31
ttnn: add the measured matmul config registry

nkapreTT:
## What this does

Adds one opt-in TTNN matmul registry backed by checked-in tables generated from banked matmul sweeps.

- `ttnn.CONFIG.matmul_registry_mode`: `off` (default), `shadow`, or `on`
- exact matching on operation semantics, shape, dtype, layout, device/grid, and compute-kernel knobs
- dense matmul: 60,221 measured entries across 1,302 M/K/N shapes
- AGMM: 48 Blackhole 32-chip entries across the existing 8x4 sweep cohort and exact Wan 4x8 model contracts
- compact generated C++; no Python or generation step in the tt-metal build

`shadow` performs lookup without changing behavior. `on` applies an exact matched recipe. For AGMM, an exact match supersedes the model's explicit program and compute-kernel configs so model code can use the unified registry; a normal miss preserves the caller's existing config. Strict fallback mode turns a miss into an error for coverage validation.

## Model A/B flow

All Models now exposes `matmul-registry-mode=off|shadow|on` and passes it to single-host E2E jobs through `TTNN_CONFIG_OVERRIDES`. The same unmodified model test can therefore run registry off and on; there is no registry-specific model fork or handcrafted fallback path.

The production Wan 2.2 14B 4x8 pipeline was captured at 480p over 40 inference steps:

- 16,800 AGMM calls
- four unique exact contracts at the real `M=4096` sequence length
- call counts: 3,360 / 3,360 / 6,720 / 3,360
- inference completed in 52.23s on the capture runtime

The capture pytest later failed only its cold-load encoder timing threshold (`0.154s` observed vs `0.1s` expected); inference and registry capture completed first. The full-pipeline contracts are banked in [tt-matmul-codegen commit 5943260](https://github.com/tenstorrent/tt-matmul-codegen/commit/5943260d3e).

Paired All Models Wan runs on this exact PR head are in progress:

- [registry off](https://github.com/tenstorrent/tt-metal/actions/runs/35147890336)
- [registry on](https://github.com/tenstorrent/tt-metal/actions/runs/35147894268)

## Generated-table flow

The generator remains in the separate `tt-matmul-codegen` repository because it owns the sweep results and table source. Its deterministic install commands are:

```bash
scripts/pipeline.sh registry ~/tt-metal
scripts/pipeline.sh multichip-registry ~/tt-metal
```

The flow validates, sorts, and emits the checked-in C++ tables. There are no runtime digests, provenance sidecars, locks, counters, circuit breakers, or build-time Python hooks.

## Selection policy

Each dense exact key ships the fastest legal, verified measured recipe after the noise-band tie-break. If a matching `generic_ttnn_auto` baseline exists, the recipe must beat it; otherwise no speedup is claimed for that entry.

For the 1,806 dense entries with a measured baseline: median 1.43x, geomean 1.54x, and no selected entry is slower than the measured baseline. The other 58,415 entries cover explicit knob vectors for which no auto baseline exists.

## Host lookup characterization

Standalone lookup time over the generated tables:

- dense 60,221-entry table: 120-162 ns per lookup across first/middle/last hits; 142 ns miss
- AGMM 4-entry cohort: 113-122 ns
- AGMM 40-entry cohort: 132-163 ns

Five AGMM table lookups are under 1 microsecond total. These measurements isolate table search and exclude request-key construction and recipe materialization; the paired model jobs measure the complete host/device E2E effect.

## Verification

- `tt-matmul-codegen`: `PYTHONPATH=. pytest -q tests/test_multichip_registry.py` — 5 passed
- generated 48-entry AGMM header compiled with `clang++ -std=c++20 -fsyntax-only`
- `git diff --check` — clean in both repositories
- previous C++ head full Release Ubuntu 22.04 build — [passed](https://github.com/tenstorrent/tt-metal/actions/runs/35128127186)
- final C++ head full Release Ubuntu 22.04 build — [passed](https://github.com/tenstorrent/tt-metal/actions/runs/35147864557)

The local `copilot-build.sh --build-ttnn-tests` wrapper cannot run because Docker is unavailable on this host. Current `Scope changed tests` / `Changed tests verified` failures are caused by the unrelated `Gemma3-27B [wh_llmbox]` entry in `tests/pipeline_reorg/vllm_model_tests.yaml` missing `cmd`; the PR Gate itself passes.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nkapre/matmul-registry-productized)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nkapre/matmul-registry-productized)


ttnn/cpp/ttnn/operations/matmul/device/config/registry/codegen/emit_cpp.py
```diff
@@ -0,0 +1,1017 @@
...
+from typing import Any
+
+ARTIFACT_KIND = "ttnn_matmul_registry_lock"
+GENERATOR_VERSION = "ttnn-matmul-lock-emitter-v1"
+POLICY_VERSION = "matmul-direct-bank-v2"
+DIRECT_BANK_EVIDENCE_POLICY_VERSION = "deterministic-matmul-bank-v2"
+LOCK_SCHEMA_VERSION = 2
+KEY_SCHEMA_VERSION = 1
+CODEGEN_RECIPE_ABI = 2
+MAX_ENTRIES = 4096
```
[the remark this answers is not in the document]
nkapreTT: Fixed. The stale in-tree emitter and lock are deleted; generation now lives in tt-matmul-codegen commit 4664d34 and emits the schema-v2 60,221-entry table through the documented one-command flow.

tests/ttnn/unit_tests/gtests/test_matmul_registry.cpp
```diff
@@ -0,0 +1,578 @@
...
+    bad_metadata.lock_schema_version = 99;
+    const auto entry = exact_entry(req);
+    EXPECT_EQ(
+        resolve_with_compact_table_for_testing(req, eligibility(), bad_metadata, {&entry, 1}).reason,
+        ResolutionReason::UnsupportedArtifact);
+    bad_metadata = metadata();
+    bad_metadata.semantic_source_sha256 = {};
+    EXPECT_EQ(
+        resolve_with_compact_table_for_testing(req, eligibility(), bad_metadata, {&entry, 1}).reason,
+        ResolutionReason::UnsupportedArtifact);
```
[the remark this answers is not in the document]
nkapreTT: Fixed by deleting the obsolete metadata contract and its contradictory tests. The checked table is now plain generated C++ with schema/key validation only.

ttnn/cpp/ttnn/operations/matmul/CMakeLists.txt
```diff
@@ -3,6 +3,36 @@ include(sources.cmake)
 add_library(ttnn_op_matmul ${LIB_TYPE})
 add_library(TTNN::Ops::Matmul ALIAS ttnn_op_matmul)
 
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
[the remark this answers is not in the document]
nkapreTT: Fixed. The unconditional Python discovery is removed; tt-metal only compiles the checked-in C++ table.

ttnn/cpp/ttnn/operations/matmul/device/config/matmul_config_registry.hpp
```diff
@@ -0,0 +1,336 @@
...
+    bool mode_is_frozen = false;
+    Mode frozen_mode = Mode::Off;
+    std::size_t exact_entry_count = 0;
+    std::array<DomainStatsSnapshot, kOperationDomainCount> domains{};
+};
+
+StatsSnapshot stats_snapshot() noexcept;
+void reset_stats_for_testing() noexcept;
+
+class SelectedExecutionGuard {
```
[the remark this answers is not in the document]
nkapreTT: Fixed by removing the unused execution guard, telemetry, and circuit-breaker machinery entirely. Dispatch now simply applies a valid hit or leaves the caller parameters untouched.

ttnn/cpp/ttnn-nanobind/core.cpp
```diff
@@ -62,6 +63,8 @@ void py_module_types(nb::module_& mod) { nb::class_<ttnn::Config>(mod, "Config")
 void py_module(nb::module_& mod) {
     namespace lightmetal = tt::tt_metal::experimental::lightmetal;
 
+    export_enum<ttnn::MatmulRegistryMode>(mod, "MatmulRegistryMode", nb::is_arithmetic());
```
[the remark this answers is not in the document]
nkapreTT: Fixed. Added Python coverage for set/get of Off, Shadow, and On plus load_config_from_dictionary; the mode also participates in config save/load and fixture restoration.

tests/ttnn/unit_tests/gtests/sources.cmake
```diff
@@ -30,6 +30,7 @@ set(UNIT_TESTS_TTNN_BASIC_SOURCES
     test_graph_query_op_runtime.cpp
     test_launch_operation.cpp
     test_matmul.cpp
+    test_matmul_registry.cpp
```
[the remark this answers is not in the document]
nkapreTT: Fixed. The PR body now says the focused gtest is wired into unit_tests_ttnn_basic and clearly states that the full target was not built on this host because Docker is unavailable.

tests/ttnn/unit_tests/gtests/test_agmm_registry.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+TEST(AgmmRegistry, CohortsDoNotCrossDeviceCounts) {
+    EXPECT_TRUE(registry::entries_for_device_count(8).empty());
+    EXPECT_EQ(registry::entries_for_device_count(32).size(), 40U);
+    EXPECT_TRUE(registry::entries_for_device_count(1).empty());
+    EXPECT_TRUE(registry::entries_for_device_count(16).empty());
+}
+
+TEST(AgmmRegistry, EveryBh32EntryRoundTripsAndMaterializes) {
+    const auto entries = registry::generated::bh32_entries();
```
bbradelTT: What is bh32?

bh is usually Blackhole, but I've never heard of Blackhole32.

You might need to fully expand what the acronym should be.

bbradelTT (commented):
Please add documentation about how to update the registry. Also, about how to enable or disable this.

What is the plan for bringing this functionality up, and updating models / perf tests?

mtairum (changes requested):
You've added the  `matmul-registry-mode` + TTNN override to models Tier 1 e2e.
Why not to Tier 2 and Tier 3 as well, for completion? Also tier 1/2/3 models unit and tier 1/2 sweep?

I understand this was tested on Wan, and that's a tier 1 model, but if we're making this change, it should be propagated to the other pipelines.

nkapreTT:
@mtairum Addressed in e0eaf4b9c4.

The `matmul-registry-mode` choice (`off`/`shadow`/`on`, default `off`) is now propagated through:

- Tier 2 and Tier 3 E2E
- Tier 1, Tier 2, and Tier 3 unit tests, including the multihost runner
- Tier 1 and Tier 2 sweep tests
- `all-model-tests.yaml` for every corresponding tier job

Each shared runner exports the same `TTNN_CONFIG_OVERRIDES` value to test processes. Tier 3 sweeps are intentionally absent because Tier 3 has no sweep pipeline.

I reran the applicable pre-commit suite on all 11 workflow files; check-yaml and yamllint pass.

bbradelTT:
@nkapreTT ran some tests and found that the overhead of the registry is minimal:
<1% variation in runtimed

  - Registry ON: https://github.com/tenstorrent/tt-metal/actions/runs/35282691384
  - Registry OFF: https://github.com/tenstorrent/tt-metal/actions/runs/35282694925

ttnn/cpp/ttnn/operations/matmul/device/config/matmul_registry_dispatch.cpp
```diff
@@ -0,0 +1,255 @@
...
+        return inspection;
+    }
+
+    const auto* device_a = input_tensor_a.device();
+    const auto* device_b = input_tensor_b.device();
+    // This registry contains local dense-matmul recipes only. Distributed
+    // matmul families (for example AGMM) own separate exact tables keyed by
+    // their collective topology; a mesh-wide tensor must never consume this
+    // single-device evidence merely because each local launch is legal.
+    if (input_tensor_a.logical_shape().rank() != 2 || input_tensor_b.logical_shape().rank() != 2 ||
```
bbradelTT: Please replace rank() != 2 with “rank ≥ 2 and every dim except the last two is 1” for both tensors and add at least one test to verify.

bbradelTT (approved):
From an ops perspective this looks good. I just asked for a minor change.

In regards to my previous comment, documentation is in tech_reports/ttnn/Certified_Matmul_Registry.md

ttnn/cpp/ttnn/operations/matmul/device/config/matmul_registry_dispatch.cpp
nkapreTT: Fixed in e3036525c46. Registry eligibility now accepts rank >= 2 when every dimension before the final two is 1 for both tensors. Added focused coverage for rank-2, [1,1,M,K], and rejection of a non-unit front dimension.

tests/ttnn/unit_tests/gtests/test_agmm_registry.cpp
nkapreTT: bh32 meant a 32-device Blackhole configuration, not a product name. I removed the abbreviation from the checked-in C++ identifiers in e3036525c46 (blackhole_32_device_cohorts) and updated the generator on tt-matmul-codegen main in 40f924c165f.

mtairum (approved):
Approving from the model tier yaml side.

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_config_registry.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+#include "agmm_registry_data.hpp"
+#include "ttnn/operations/compute_throttle_utils.hpp"
+#include "ttnn/operations/matmul/device/config/matmul_config_registry.hpp"
+
+namespace ttnn::experimental::all_gather_minimal_matmul_registry {
+namespace {
+
+bool is_default_tile(const compact::TensorDescriptor& tensor) noexcept {
+    return tensor.layout == 1 && tensor.tile_height == 32 && tensor.tile_width == 32 &&
```
vtsilytskyiTT:
```suggestion
    return tensor.layout == 1 && tensor.tile_height == tt::constants::TILE_HEIGHT && tensor.tile_width == tt::constants::TILE_WIDTH &&
```
Include `tt_metal/api/tt-metalium/constants.hpp` if required.

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_config_registry.cpp
```diff
@@ -0,0 +1,275 @@
...
+    return tensor.present ? tensor_is_complete(tensor.tensor) && is_default_tile(tensor.tensor)
+                          : tensor.tensor == compact::TensorDescriptor{};
+}
+
+std::optional<std::uint64_t> checked_product(std::uint64_t lhs, std::uint64_t rhs) noexcept {
+    if (lhs != 0 && rhs > std::numeric_limits<std::uint64_t>::max() / lhs) {
+        return std::nullopt;
+    }
+    return lhs * rhs;
+}
```
vtsilytskyiTT: Nit1: this is a good generic function, may be useful elsewhere. We have similar functions in `tt_metal/api/tt-metalium/math.hpp`. What do you think about placing `checked_product` there + some test coverage?
Nit2: would be nice to follow this function definition: https://en.cppreference.com/cpp/numeric/ckd_mul . Future migration will be smooth.

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_config_registry.cpp
```diff
@@ -0,0 +1,275 @@
...
+            fidelity = tt::tt_metal::MathFidelity::HiFi2;
+            break;
+        case static_cast<std::uint32_t>(tt::tt_metal::MathFidelity::HiFi3):
+            fidelity = tt::tt_metal::MathFidelity::HiFi3;
+            break;
+        case static_cast<std::uint32_t>(tt::tt_metal::MathFidelity::HiFi4):
+            fidelity = tt::tt_metal::MathFidelity::HiFi4;
+            break;
+        default: return std::nullopt;
+    }
```
vtsilytskyiTT: This should be common reusable function.

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_config_registry.cpp
```diff
@@ -0,0 +1,275 @@
...
+                .subblock_h = config.subblock_h,
+                .subblock_w = config.subblock_w,
+                .compute_with_storage_grid_size = {config.compute_grid_x, config.compute_grid_y}},
+        .compute_kernel_config = compute_kernel_config};
+}
+
+std::optional<Recipe> select_recipe(const Mode mode, const RegistryRequestFacts& facts) {
+    if (mode != Mode::On) {
+        return std::nullopt;
+    }
```
vtsilytskyiTT: Shouldn't we select a recipe for `Shadow` mode?

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_config_registry.cpp
```diff
@@ -0,0 +1,275 @@
...
+    const auto key = build_registry_key(facts);
+    const auto* entry = key ? lookup(*key) : nullptr;
+    std::optional<Recipe> recipe;
+    if (entry != nullptr) {
+        recipe = materialize_recipe(*entry);
+    }
+    if (!recipe && ttnn::operations::matmul::registry::fallback_is_error(mode)) {
+        TT_THROW("AGMM registry required an exact recipe, but no exact match was found");
+    }
```
vtsilytskyiTT: TT_FATAL = condition + TT_THROW.

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_registry_descriptor.hpp
```diff
@@ -0,0 +1,180 @@
...
+    return true;
+}
+inline constexpr const EntryDescriptor* lookup_exact(
+    const KeyDescriptor& key, std::span<const EntryDescriptor> entries) noexcept {
+    const auto candidate = std::lower_bound(
+        entries.begin(), entries.end(), key, [](const EntryDescriptor& entry, const KeyDescriptor& requested) {
+            return entry.key < requested;
+        });
+    return candidate != entries.end() && candidate->key == key ? &*candidate : nullptr;
+}
```
vtsilytskyiTT: No need, just use [std::find](https://en.cppreference.com/cpp/algorithm/find)

ttnn/cpp/ttnn/operations/experimental/ccl/all_gather_minimal_matmul_async/registry/agmm_registry_descriptor.hpp
```diff
@@ -0,0 +1,180 @@
...
+    std::uint32_t output_dtype{};
+    bool output_memory_config_present{};
+    std::uint32_t output_memory_layout{};
+    std::uint32_t output_buffer_type{};
+    std::uint32_t output_layout{};
+    std::uint16_t output_tile_height{};
+    std::uint16_t output_tile_width{};
+    bool output_tile_transpose_of_faces{};
+    bool output_tile_transpose_within_face{};
+    auto operator<=>(const OperationDescriptor&) const = default;
```
vtsilytskyiTT: Do you really need a spaceship operator for any of those structures? Equality should be enough for most of them. < or > makes no sense to me. What is the rule to order descriptors?

ttnn/cpp/ttnn/operations/matmul/device/config/registry/matmul_registry_exact.hpp
```diff
@@ -0,0 +1,196 @@
...
+    bool fuse_batch{};
+    bool mcast_in0{};
+    bool transpose_mcast{};
+    bool fused_activation_present{};
+    bool gather_in0{};
+    bool hop_cores_present{};
+    bool untilize_out{};
+    bool stream_in1{};
+
+    auto operator<=>(const ProgramConfigDescriptor&) const = default;
```
vtsilytskyiTT: I'm quite skeptical about spaceship operator here. Do you need < and > for sorting those structures?

ttnn/cpp/ttnn/operations/matmul/device/config/registry/matmul_registry_exact.hpp
```diff
@@ -0,0 +1,196 @@
...
+    const KeyDescriptor& key, const std::span<const ProgramConfigExactEntry> entries) noexcept {
+    const auto candidate = std::lower_bound(
+        entries.begin(),
+        entries.end(),
+        key,
+        [](const ProgramConfigExactEntry& entry, const KeyDescriptor& requested_key) {
+            return entry.key < requested_key;
+        });
+    return candidate != entries.end() && candidate->key == key ? &*candidate : nullptr;
+}
```
vtsilytskyiTT: std::find

ttnn/cpp/ttnn/operations/matmul/device/config/registry/matmul_registry_exact.hpp
```diff
@@ -0,0 +1,196 @@
...
+// entry whose value disagreed with its key would answer a caller with numerics
+// they never asked for.
+constexpr bool entries_bind_key_compute_kernel(const std::span<const ProgramConfigExactEntry> entries) noexcept {
+    for (const auto& entry : entries) {
+        if (entry.key.compute_kernel != entry.compute_kernel_config) {
+            return false;
+        }
+    }
+    return true;
+}
```
vtsilytskyiTT: https://cppreference.com/cpp/algorithm/all_any_none_of (though they are constexpr only since C++20, so up to you).

ttnn/cpp/ttnn/operations/matmul/device/config/registry/matmul_registry_exact.hpp
```diff
@@ -0,0 +1,196 @@
...
+    const std::span<const ProgramConfigExactEntry> entries) noexcept {
+    for (const auto& entry : entries) {
+        if (entry.key.has_activation || entry.program_config.fused_activation_present ||
+            entry.key.compute_kernel.math_approx_mode != kMathApproxModeIsInertAt ||
+            entry.compute_kernel_config.math_approx_mode != kMathApproxModeIsInertAt) {
+            return false;
+        }
+    }
+    return true;
+}
```
vtsilytskyiTT: Same, std has all_of, any_of, none_of.

ttnn/cpp/ttnn/operations/matmul/device/config/registry/README.md
vtsilytskyiTT: I saw same file in tech_reports

ttnn/cpp/ttnn/operations/matmul/device/config/matmul_registry_dispatch.cpp
```diff
@@ -0,0 +1,269 @@
...
+            has_bias,
+            call_semantics,
+            parameters,
+            optional_output_tensor,
+            /*trace_capture_active=*/false);
+    } catch (...) {
+        // Request construction may inspect tensors before legacy validators
+        // do. Preserve their exception timing by falling back untouched.
+        if (fallback_is_error(mode)) {
+            throw;
```
vtsilytskyiTT: Throw in device code?

ttnn/cpp/ttnn/operations/matmul/device/config/matmul_registry_dispatch.cpp
```diff
@@ -0,0 +1,269 @@
...
+        }
+    } catch (...) {
+        if (fallback_is_error(mode)) {
+            throw;
+        }
+        // An unavailable trace state is indistinguishable from active capture.
+        inspection.eligibility.trace_capture_active = true;
+        static_cast<void>(resolve_for_dispatch(mode, std::nullopt, inspection.eligibility, parameters));
+        return;
+    }
```
vtsilytskyiTT: Do we throw in device code?

ttnn/core/config.cpp
```diff
@@ -22,6 +24,16 @@
 
 namespace ttnn::core {
 
+namespace {
```
vtsilytskyiTT: Why is matmul_registry_mode_ the only atomic field? Is set_matmul_registry_mode() expected to run concurrently with matmul dispatch? If mode is fixed before execution starts, this could be a regular field like the others. If concurrent switching is supported, making only the mode atomic protects the enum but not the registry state governed by it. Could we document the concurrency contract and explain how registry population/lookup and mode transitions are synchronized? Also, if the mode publishes initialized registry contents, the getter/setter should use acquire/release explicitly; otherwise relaxed ordering should be sufficient.
```c++
RegistryMode registry_mode() const noexcept {
    return registry_mode_.load(std::memory_order_acquire);
}

void set_registry_mode(RegistryMode mode) noexcept {
    registry_mode_.store(mode, std::memory_order_release);
}
```

vtsilytskyiTT (changes requested):

vtsilytskyiTT:
Please add a "motivation" chapter in PR description. I see what PR does, but don't see what is the benefit.

Issue #54329, opened 2026-08-25
[Quasar] Resnet LLK workarounds master issue


rtawfik01:
Workarounds that were introduced during resnet bringup that are merged in main but should be fixed properly:

- [x] TEN-4746: dummy copy_tile (real unpack TDMA) forced between wait_front → pop_front

Bare wait_front(); pop_front(); traps the Quasar unpacker (POP_TILES races WAIT_TILES → TILE_COUNTERS 0x10000). Workaround interposes a throwaway copy_tile of tile 0 with a full tile_regs_acquire/commit/wait/release dance, #ifdef ARCH_QUASAR. TTI_NOP/DMANOP are documented as insufficient — must be a real TDMA op.

Need to investigate if can use UNPACR_NOP instead
```
matmul/.../compute/bmm_large_block_zm_fused_bias_activation_metal2.cpp:468, :500
matmul/.../compute/bmm_large_block_zm_fused_bias_activation.cpp (same pattern)

```
Jira: TEN-4746 / #48552. Removal candidate once UNPACR_NOP works
- [x] Raw llk_* calls inside compute .cpp

Compute kernels must use the compute API, never the LLK directly.

```
PACK((llk_pack_relu_config(...))) — ~35 files (binary/binary_ng/matmul). e.g. binary/.../compute/eltwise_binary_kernel.cpp:43, bmm_*.cpp:280/294/474/561
PACK((llk_pack_init(out_cb_id))) — pool_generic/.../compute/compute_pool_2d.cpp:315
MATH((llk_math_reconfig_remap(true))) — compute_pool_2d.cpp:337
```

Fix = route through / add the missing compute-API wrapper.

- [x] Inline ISA register pokes (TTI_STALLWAIT / TT_SETC16)
Hand-written asm barriers in compute kernels. Arch-gated because arity differs (Quasar 4 args, WH 2).

```
PACK(TTI_STALLWAIT(STALL_MATH,...,PACK)) — compute_pool_2d.cpp:429 — removal candidate: author says "remove once the sim models PUSH_TILES.packer_wr_done_wait_mask"; no-op on WH/BH HW.

PACK(TTI_STALLWAIT(STALL_PACK, WAIT_SFPU)) — bmm_*_metal2.cpp:608, bmm_*.cpp:542, bmm_fused_activation.hpp:117

[...]

PR #54948, opened 2026-08-31
[Quasar] Replace TEN-4746 copy_tile drain with dummy_unpack() helper

rtawfik01:
The Quasar matmul kernel bmm_large_block_zm_fused_bias_activation_metal2.cpp drains the mm_partials CB between wait_front/pop_front purely for flow control (PACKER_L1_ACC path). A bare wait->pop trips the Quasar unpacker (TEN-4746: POP_TILES resolves before WAIT_TILES because the tile-counter compare is only re-sampled by a real same-engine TDMA on the dfb). The existing workaround interposed a full copy_tile plus a reconfig + matmul_block_init bracket to restore matmul state -- ~20 lines and an in-kernel ARCH_QUASAR block per drain.

Replace it with a single clear-SrcA unpacker op wrapped in a new compute-API helper, ckernel::dummy_unpack() (tile_move_copy.h):

    UNPACK(TTI_UNPACR_NOP(p_unpacr::UNP_A, 0, 0, 0,
                          p_unpacr::UNP_CLRSRC_ZERO, p_unpacr::UNP_CLRSRC));

A clear-source op is a real unpacker transaction, so it satisfies the POP-after-WAIT ordering that a bare NOP/DMANOP/TTI_NOP cannot; but unlike copy_tile it does not read the CB, so it leaves the PACKER_L1_ACC offset bookkeeping untouched (no reconfig / matmul_block_init needed). Clearing SrcA is harmless -- the next matmul re-unpacks SrcA. On WH/BH there is no such HW ordering bug, so dummy_unpack() compiles to nothing.

Both TEN-4746 drain loops now read wait_front(); dummy_unpack(); pop_front(); and carry no ARCH_QUASAR ifdef. The two genuine mm_partials consumes (copy_block reload, add_tiles bias) are unchanged.

Validated on the 2x3 Zebu emulator: resnet50 quasar fc linear (test_resnet50_fc_linear) passes, 20/20 deterministic.

### PR Category
#54329 

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:rtawfik/quasar-dummy-unpack)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:rtawfik/quasar-dummy-unpack)

### DIFF
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
@@ -120,7 +120,7 @@ inline void llk_unpack_A_block(
  * in-flight op; it clears SrcA only (the next op re-unpacks it) and reads nothing from L1. Kernels do not
  * need this in normal operation -- use it only for debug when an explicit SrcA flush is wanted.
  */
-inline void llk_unpack_dummy() {
+inline void llk_unpack_dummy([[maybe_unused]] const std::uint32_t dfb_id) {
     TTI_STALLWAIT(ckernel::p_stall::STALL_UNPACK, ckernel::p_stall::SRCA_CLR);
     TTI_UNPACR_NOP(ckernel::SrcA, 0, 0, 0, 0, 0, 0, ckernel::p_unpacr_nop::CLR_SRC_0, ckernel::p_unpacr_nop::CLR_SRC);
 }
--- a/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
+++ b/tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
@@ -6,6 +6,10 @@
 #include <cstdint>
 #include "llk_unpack_A.h"
 #include "llk_unpack_common_api.h"
+// llk_unpack_dummy(): debug-only SrcA flush on WH/BH (no WAIT/POP ordering role), defined in the debug
+// folder. Pulled in only so the shared compute-API dummy_unpack() resolves; on Quasar the same call is a
+// required TEN-4746 drain primitive defined inline in that arch's llk_unpack_A_api.h.
+#include "debug/llk_unpack_dummy.h"
 
 /*************************************************************************
  * LLK UNPACK A
@@ -111,23 +115,6 @@ inline void llk_unpack_A_block(
         address += offset_address;
         WAYPOINT("UPAD");
     }
-}
-
-/**
- * @brief Flush the SrcA bank: STALLWAIT on SrcA-clear then a clear-SrcA UNPACR_NOP (no CB read, no DEST
- *        write). Reads nothing from L1.
- *
- * The STALLWAIT ensures SrcA is free before the clear, so this cannot clobber a SrcA bank still owned by an
- * in-flight op; SrcA is cleared only (the next op re-unpacks it). WH/BH have no WAIT/POP ordering
- * requirement, so this is a debug-only flush -- use it only when an explicit SrcA flush is wanted. (On
- * Quasar the same call is a required drain primitive; see that arch's llk_unpack_dummy.)
- *
- * @param dfb_id  Unused on WH/BH; accepted only to match the Quasar signature, where it names the drained
- *                dataflow buffer.
- */
-inline void llk_unpack_dummy([[maybe_unused]] const std::uint32_t dfb_id) {
-    TTI_STALLWAIT(ckernel::p_stall::STALL_UNPACK, ckernel::p_stall::SRCA_CLR);
-    TTI_UNPACR_NOP(ckernel::SrcA, 0, 0, 0, 0, 0, 0, ckernel::p_unpacr_nop::CLR_SRC_0, ckernel::p_unpacr_nop::CLR_SRC);
 }
 
 template <BroadcastType BType = BroadcastType::NONE>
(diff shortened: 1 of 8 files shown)


rtawfik01:
@bbradelTT @amokanTT please review

bbradelTT (approved):

tt_metal/hw/inc/api/compute/tile_move_copy.h
```diff
@@ -71,6 +71,23 @@ ALWI void copy_init(
...
+ * consume. On Quasar this satisfies the unpacker's requirement that a POP_TILES be ordered after its
+ * WAIT_TILES by a real unpacker op (TEN-4746 / #48552) -- a bare TTI_NOP/DMANOP does NOT (they issue no
+ * unpacker transaction), and a real read (copy_tile / UNPACR_TILE) is heavier and, if partial, corrupts
+ * PACKER_L1_ACC offsets. This clears SrcA only (harmless: the next op re-unpacks SrcA) and never reads the
+ * CB, so it is correct and side-effect free. On WH/BH there is no such HW ordering bug, so this is a no-op.
+ * Return value: None
+ */
+// clang-format on
+ALWI void dummy_unpack() {
+#ifdef ARCH_QUASAR
```
fvranicTT: Could we have an #else #error for non-quasar archs? We had this issue in the past where AI agent would ignore empty-function warnings emitted by the compiler.

fvranicTT: Looking at the rest of the code, I now think you made this intentionally like this.

rtawfik01: i fixed this entirely

fvranicTT (approved):

tt_metal/hw/ckernels/quasar/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -193,5 +193,20 @@ inline void llk_unpack_A_block(
     }
 }
 
+/**
+ * @brief Drain / flow-control NOP for the unpacker (no CB read, no DEST write).
+ *
+ * Issues a STALLWAIT on SrcA-clear followed by a clear-SrcA UNPACR_NOP. Used between cb_wait_front and
```
ncvetkovicTT: I don't mind referencing the issue, but let's just make sure the comments are similar between archs. I  prefer the wording in Wormhole version, given that we reword the bit that I flagged down there. Ideally, I'd like to read from a comment how/why/if this QSR version of `llk_unpack_dummy` differs from WH/BH.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
```diff
@@ -92,6 +92,18 @@ inline void llk_unpack_A_block(
     }
 }
 
+/**
+ * @brief Flush the SrcA bank: STALLWAIT on SrcA-clear (stalling the unpacker) then a clear-SrcA UNPACR_NOP.
+ *
+ * The STALLWAIT ensures SrcA is free before the clear, so this cannot clobber a SrcA bank still owned by an
+ * in-flight op; it clears SrcA only (the next op re-unpacks it) and reads nothing from L1. Kernels do not
+ * need this in normal operation -- use it only for debug when an explicit SrcA flush is wanted.
```
ncvetkovicTT: But we are not using this only for debug, right? I understand the intent - it's unsafe, unwise and undesirable to use this function regularly. If we want to make this a debug-only function, let's put it in a separate file inside `tt-metal/tt_metal/tt-llk/*/llk_lib/debug`. Notice that I am suggesting that we move this lower to llk-lib level.

ncvetkovicTT (commented):
If we're going to go with the current fix, I suggest:
* removing `dummy_unpack` compute API
* using LLK API in the kernel
*  moving the functionality with Tensix instructions a layer lower (llk-lib)
I am on the fence of making this "debug-only" vs. a regular API. I don't mind it being regular API, but if it's debug, we have to move it to corresponding folder.

tt_metal/hw/ckernels/blackhole/metal/llk_api/llk_unpack_A_api.h
rtawfik01: its only debug for WH/BH, but is necessary for Quasar, but I get your point, will fix for WH/BH

rtawfik01: ok fixed, @ncvetkovicTT please review

ncvetkovicTT (approved):
lgtm

PR #54958, opened 2026-08-31
Remove the four fidelity counters that are tied off in hardware

### LINKED ISSUE #54950: Three fidelity metrics are constant because their counters are tied off in hardware
## What happens

Three metrics in `ops_perf_results.csv` are constant regardless of the workload:

| metric | always reads |
|---|---|
| Fidelity Stall Rate | 0% |
| HiFi Fraction | 0% |
| Avg HF Cycles Per Instrn | exactly 1.00 |

So every op is reported as pure LoFi with no fidelity stalls, whatever math fidelity it actually ran at. That is worse than a missing column, because the numbers look plausible.

## Why

Four counters feed them, and all four are tied off in hardware on both Wormhole and Blackhole. Confirmed by reading the RTL for both architectures:

- The signal that gates the fidelity stall counter is hardwired to zero, so `MATH_FIDELITY_STALL` never increments.
- The high fidelity cycle counter only advances while that same signal is set, so it never leaves zero. That makes `MATH_INSTRN_HF_1_CYCLE` a duplicate of `MATH_INSTRN_STARTED`, and `MATH_INSTRN_HF_2_CYCLE` and `MATH_INSTRN_HF_4_CYCLE` constant zero.

The counters are declared in `hw_counters.h` for both architectures and consumed in `tools/tracy/perf_counter_analysis.py`. The formulas are also documented in `tech_reports/PerfCounters/perf-counters.md` and in the ttnn profiling docs.

## Scope

- Drop the four counters from the per-architecture arrays in `hw_counters.h` so nothing captures them.
- Drop the three metrics that read them, along with their CSV columns.
- Remove the formulas from the two documents that describe them.

## Enum ordinals

The profiler stores and decodes the compiled ordinal as the wire value, so removing enumerators from the middle of `PerfCounterType` shifts every ordinal above them and a capture taken before the change would decode incorrectly after it. Keep the four enumerators in place as reserved entries and only remove them from the arrays. There is no pressure on the 256 limit, so the placeholders cost nothing.

## Note

This removes four columns from `ops_perf_results.csv`. Anything parsing that file by column name will stop finding them, which is the intended outcome given the values were never real.

mvlahovicTT:
Closes #54950

### What was wrong
Three metrics in `ops_perf_results.csv` were constant on every op: Fidelity Stall Rate 0%, HiFi Fraction 0%, Avg HF Cycles Per Instrn exactly 1.00. Every op read as pure LoFi with no fidelity stalls, whatever fidelity it actually ran at, and the values looked plausible enough to reason from.

### Why
The four counters behind them are tied off in the RTL on both Wormhole and Blackhole. The signal that gates `MATH_FIDELITY_STALL` is hardwired to zero, and the high-fidelity cycle counters only advance while it is set, so `MATH_INSTRN_HF_1_CYCLE` duplicates `MATH_INSTRN_STARTED` and `HF_2` / `HF_4` never leave zero.

### What changed
- The four entries leave the per-arch `hw_counters.h` arrays; `unpack_counters` drops from 22 to 18 on both arches, and `NUM_INSTRN_COUNTERS` names the instrn array size instead of repeating it.
- The three metrics and their twelve CSV columns leave the Tracy consumer (both metric paths), `process_ops_logs.py`, the tech report and the ttnn profiling doc. The tech report no longer calls these workload-dependent zeros.
- `PerfCounterType` keeps the four enumerators as reserved: the profiler stores the ordinal on the wire, so removing them would shift every later ordinal.

### Verification
- `ops_perf_results.csv` goes from 336 to 324 columns, the twelve removed.
- LLK harness built and run on Blackhole with counters enabled; output byte-identical with and without this change on the same variants.
- Rebased on main. Blackhole op sweep with `TT_METAL_PROFILE_PERF_COUNTERS=47` and `process_ops_logs.py --device-only`: 8/8 ops pass, 8/8 reports generated. Quasar: ttsim compile and boot and a Zebu emulator smoke (compute kernel and LLK matmul, profiler on) pass. CI: profiler pipeline https://github.com/tenstorrent/tt-metal/actions/runs/33988518026; sanity tests https://github.com/tenstorrent/tt-metal/actions/runs/33993119165.

|---|---|---|---|---|---|
| matmul | 1 | PASS | OK | 60.96% | 50.00% |
| binary | 22 | PASS | OK | 48.56% | 35.09% |
| tilize | 3 | PASS | OK | 61.01% | 56.91% |
| relu | 22 | PASS | OK | 77.80% | 25.29% |
| sqrt | 1 | PASS | OK | n/a | 10.93% |
| silu | 102 | PASS | OK | 47.74% | 47.34% |
| concat | 661 | PASS | OK | 65.31% | 29.34% |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mvlahovic/perf-counter-drop-fidelity)


tt_metal/hw/inc/internal/tt-1xx/blackhole/hw_counters.h
```diff
@@ -149,67 +149,38 @@ constexpr std::array<std::pair<PerfCounterType, uint16_t>, 16> l1_4_counters = {
...
-     {PerfCounterType::WAITING_FOR_MATH_IDLE_2, 52},
-     {PerfCounterType::WAITING_FOR_NONZERO_SEM_2, 53},
-     {PerfCounterType::WAITING_FOR_NONFULL_SEM_2, 54},
-     {PerfCounterType::WAITING_FOR_MOVE_IDLE_2, 55},
-     {PerfCounterType::WAITING_FOR_MMIO_IDLE_2, 56},
-     {PerfCounterType::WAITING_FOR_SFPU_IDLE_2, 57},
-     {PerfCounterType::THREAD_INSTRUCTIONS_0, 256},
-     {PerfCounterType::THREAD_INSTRUCTIONS_1, 264},
-     {PerfCounterType::THREAD_INSTRUCTIONS_2, 272},
+constexpr std::array<std::pair<PerfCounterType, std::uint16_t>, 59> instrn_counters = {
```
vvukomanovicTT: nit: magic number, can you use NUM_INSTRN_COUNTERS?

mvlahovicTT: resolved

kstevensTT (approved):

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34246404785

jbaumanTT (approved):

mo-tenstorrent (approved):
Please make sure profiler CI is green.

LGTM, however @sohaibnadeemTT should also approve before merging.

aczajkowskiTT (approved):

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378017414
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378013150

TT-BrianLiu (approved):

sohaibnadeemTT (approved):
lgtm

PR #54988, opened 2026-08-31
Let a spec factory build a different program per mesh coordinate

dgomezTT:
Right now a spec factory gets asked for its `ProgramArtifacts` exactly once, and the adapter maps that one `ProgramSpec` onto every range of the mesh. Fine for SPMD, but it makes two things impossible: an op whose program differs per device, and an op that wants the same program everywhere with different run args per device. Both keep coming up as "needs TTNN infra support" when people look at porting ops to the Metal 2.0 spec path.

The metal side isn't the problem. `MakeMeshWorkloadFromSpecs` already takes a range-to-spec map and runs distinct specs on distinct devices. The adapter just never had a way for a factory to say "my programs vary".

So this adds a factory concept for exactly that:

```cpp
static MeshWorkloadArtifacts create_mesh_workload_artifacts(
    const operation_attributes_t&, const tensor_args_t&, tensor_return_value_t&,
    const ttnn::MeshCoordinateRangeSet& tensor_coords);
```

One call returns the workload-scoped resources plus a `ProgramSpec` and `ProgramRunArgs` per coordinate range. A program uniform over part of the mesh is one entry covering that range, so you don't have to emit one per device. `MeshWorkloadSpecFactoryConcept` is mutually exclusive with the other four, so the "exactly one concept per `program_factory_t` alternative" rule still holds and existing factories classify the same as before.

Optionally declare a cache-hit refresh:

```cpp
static ProgramRunArgs override_runtime_arguments(
    const operation_attributes_t&, const tensor_args_t&, tensor_return_value_t&,
    const ttnn::MeshCoordinateRange& range);
```

It takes the range rather than a coordinate because one Program backs the whole range, so what it returns applies to every device the range covers. Handing it a coordinate invited factories to derive per-device args and get device 0's values replicated with nothing complaining.

Guardrails, since these failure modes are all quiet ones:

* A near-miss `override_runtime_arguments` signature is a compile error instead of a silent fallback to the tensor-only refresh.
* Two entries with the same range, or a range outside `tensor_coords`, are fatal. Metal catches overlapping ranges but never sees an exact duplicate, and it only checks the mesh extent.
* A factory declaring `create_mesh_workload_artifacts` as an overload set is rejected. Address-of an overload set is ill-formed, so the concept goes false and the factory would quietly fall through to the SPMD path and get one spec replicated across the mesh.

Not in here: GlobalSemaphore. The spec API only has program-scoped `SemaphoreSpec` today, so CCL-style cross-device sync is still blocked on the metal side. Separate problem, separate owner.

### Testing

Runtime coverage on the 2x4 fixture. One test builds a workload from a factory that leaves odd columns idle and asserts per-coordinate presence and absence, plus that every range shares one parked resource vector. A second drives both cache-hit branches, and for the override branch asserts it fires once per range rather than once per workload, which is the part that would regress silently.

Compile coverage pins the classification in both directions so a refactor can't flip everything onto one path.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:dgomez/spec-per-coordinate-artifacts)

### DIFF
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -159,7 +159,7 @@ struct MeshWorkloadSpecFactory {
     }
 };
 
-// Same, plus a per-coordinate cache-hit run-args refresh.
+// Same, plus a per-range cache-hit run-args refresh.
 struct MeshWorkloadSpecFactoryWithOverride {
     static ttnn::device_operation::MeshWorkloadArtifacts create_mesh_workload_artifacts(
         const OperationAttributes& /*attrs*/,
@@ -172,7 +172,7 @@ struct MeshWorkloadSpecFactoryWithOverride {
         const OperationAttributes& /*attrs*/,
         const Tensor& /*tensor_args*/,
         Tensor& /*tensor_return_value*/,
-        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
+        const ttnn::MeshCoordinateRange& /*range*/) {
         return {};
     }
 };
@@ -485,7 +485,7 @@ struct PerCoordProdAllFactoryWithOverride : PerCoordProdAllFactory {
         const Op::operation_attributes_t& attrs,
         const Op::tensor_args_t& tensor_args,
         Op::tensor_return_value_t& tensor_return_value,
-        const std::optional<ttnn::MeshCoordinate>& mesh_dispatch_coordinate) {
+        const ttnn::MeshCoordinateRange& range) {
         ++override_calls;
         return Op::ProdAllProgramFactory::create_program_artifacts(attrs, tensor_args, tensor_return_value).run_params;
     }
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -2,6 +2,7 @@
 //
 // SPDX-License-Identifier: Apache-2.0
 
+#include <atomic>
 #include <gmock/gmock.h>
 #include <type_traits>
 
@@ -473,6 +474,57 @@ TEST_F(LaunchOperation2x4Test, MeshWorkloadSpecAdapterMapsProgramsPerCoordinate)
         }
         EXPECT_EQ(static_cast<const void*>(sv.op_owned_tensors.get()), first_resources) << "range " << range;
     }
+}
+
+// Same per-coordinate mapping, but with an override so the cache-hit path takes the
+// UpdateProgramRunArgs branch instead of the tensor-only refresh.
+struct PerCoordProdAllFactoryWithOverride : PerCoordProdAllFactory {
+    static std::atomic<int> override_calls;
+
+    static tt::tt_metal::experimental::ProgramRunArgs override_runtime_arguments(
+        const Op::operation_attributes_t& attrs,
+        const Op::tensor_args_t& tensor_args,
+        Op::tensor_return_value_t& tensor_return_value,
+        const std::optional<ttnn::MeshCoordinate>& mesh_dispatch_coordinate) {
+        ++override_calls;
+        return Op::ProdAllProgramFactory::create_program_artifacts(attrs, tensor_args, tensor_return_value).run_params;
+    }
+};
+std::atomic<int> PerCoordProdAllFactoryWithOverride::override_calls{0};
+
+TEST_F(LaunchOperation2x4Test, MeshWorkloadSpecAdapterAppliesDescriptorOnCacheHit) {
+    using Op = ttnn::prim::ProdAllDeviceOperation;
+    using Adapter =
+        device_operation::MeshDeviceOperationAdapter<Op>::MeshWorkloadSpecFactoryAdapter<PerCoordProdAllFactory>;
+    using OverrideAdapter = device_operation::MeshDeviceOperationAdapter<Op>::MeshWorkloadSpecFactoryAdapter<
+        PerCoordProdAllFactoryWithOverride>;
+    static_assert(!Adapter::has_override_runtime_arguments());
+    static_assert(OverrideAdapter::has_override_runtime_arguments());
+
+    auto input = make_tensor_with_num_shards(8, mesh_device_.get());
+    Op::operation_attributes_t attrs{.output_mem_config = MemoryConfig{}};
+    Op::tensor_args_t tensor_args{.input = input};
+    auto output = Op::create_output_tensors(attrs, tensor_args);
+
+    ttnn::MeshCoordinateRangeSet tensor_coords;
+    tensor_coords.merge(ttnn::MeshCoordinateRange(mesh_device_->shape()));
+
+    // Tensor-refresh branch: a second apply must not disturb the per-coordinate mapping.
+    auto cached = Adapter::create_mesh_workload(attrs, tensor_coords, tensor_args, output);
+    const auto ranges_before = cached.workload.get_programs().size();
+    Adapter::apply_descriptor(cached, attrs, tensor_args, output);
+    EXPECT_EQ(cached.workload.get_programs().size(), ranges_before);
+    for (const auto& [range, program] : cached.workload.get_programs()) {
+        EXPECT_TRUE(cached.shared_variables.contains(range)) << "range " << range;
+    }
+
+    // Override branch: called once per range, not once for the whole workload.
+    PerCoordProdAllFactoryWithOverride::override_calls = 0;
+    auto override_cached = OverrideAdapter::create_mesh_workload(attrs, tensor_coords, tensor_args, output);
+    const auto override_ranges = override_cached.workload.get_programs().size();
+    EXPECT_GT(override_ranges, 1u);
+    OverrideAdapter::apply_descriptor(override_cached, attrs, tensor_args, output);
+    EXPECT_EQ(PerCoordProdAllFactoryWithOverride::override_calls.load(), static_cast<int>(override_ranges));
 }
 
 TEST_F(LaunchOperation2x4Test, ProgramSpecAdapterProgramsOnUniformTensorCoords) {
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -485,7 +485,7 @@ struct PerCoordProdAllFactoryWithOverride : PerCoordProdAllFactory {
         const Op::operation_attributes_t& attrs,
         const Op::tensor_args_t& tensor_args,
         Op::tensor_return_value_t& tensor_return_value,
-        const ttnn::MeshCoordinateRange& range) {
+        const ttnn::MeshCoordinateRange& /*range*/) {
         ++override_calls;
         return Op::ProdAllProgramFactory::create_program_artifacts(attrs, tensor_args, tensor_return_value).run_params;
     }
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -465,15 +465,6 @@ TEST_F(LaunchOperation2x4Test, MeshWorkloadSpecAdapterMapsProgramsPerCoordinate)
     EXPECT_GT(expected, 0u);
     EXPECT_EQ(cached.workload.get_programs().size(), expected);
     EXPECT_EQ(cached.shared_variables.size(), expected);
-
-    // Workload-scoped, not per-device: every range shares one parked resource vector.
-    const auto* first_resources = static_cast<const void*>(nullptr);
-    for (const auto& [range, sv] : cached.shared_variables) {
-        if (first_resources == nullptr) {
-            first_resources = static_cast<const void*>(sv.op_owned_tensors.get());
-        }
-        EXPECT_EQ(static_cast<const void*>(sv.op_owned_tensors.get()), first_resources) << "range " << range;
-    }
 }
 
 // Same per-coordinate mapping, but with an override so the cache-hit path takes the
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -146,6 +146,35 @@ struct CustomProgramSpecFactory {
     }
 };
 
+// Per-coordinate variants of both flavors: create_program_artifacts takes the MeshCoordinate, so the
+// adapter calls it once per coordinate and each call may return its own spec and run args.
+struct PerCoordProgramSpecFactory {
+    static ttnn::device_operation::ProgramArtifacts create_program_artifacts(
+        const OperationAttributes& /*attrs*/,
+        const Tensor& /*tensor_args*/,
+        Tensor& /*tensor_return_value*/,
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
+        return ttnn::device_operation::ProgramArtifacts{};
+    }
+};
+
+struct PerCoordCustomProgramSpecFactory {
+    static ttnn::device_operation::ProgramArtifacts create_program_artifacts(
+        const OperationAttributes& /*attrs*/,
+        const Tensor& /*tensor_args*/,
+        Tensor& /*tensor_return_value*/,
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
+        return ttnn::device_operation::ProgramArtifacts{};
+    }
+    static tt::tt_metal::experimental::ProgramRunArgs override_runtime_arguments(
+        const OperationAttributes& /*attrs*/,
+        const Tensor& /*tensor_args*/,
+        Tensor& /*tensor_return_value*/,
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/ = std::nullopt) {
+        return {};
+    }
+};
+
 // Minimal device operation supplying just the typedefs the adapter inherits.
 struct ProgramSpecMinimalOp {
     using operation_attributes_t = OperationAttributes;
@@ -167,6 +196,22 @@ static_assert(!ttnn::device_operation::ProgramFactoryConcept<CustomProgramSpecFa
 static_assert(!ttnn::device_operation::MeshWorkloadFactoryConcept<CustomProgramSpecFactory>);
 static_assert(!ttnn::device_operation::ProgramDescriptorFactoryConcept<CustomProgramSpecFactory>);
 
+// Taking the coordinate doesn't change which flavor a factory is.
+static_assert(ttnn::device_operation::ProgramSpecFactoryConcept<PerCoordProgramSpecFactory>);
+static_assert(!ttnn::device_operation::CustomProgramSpecFactoryConcept<PerCoordProgramSpecFactory>);
+static_assert(ttnn::device_operation::CustomProgramSpecFactoryConcept<PerCoordCustomProgramSpecFactory>);
+static_assert(!ttnn::device_operation::ProgramSpecFactoryConcept<PerCoordCustomProgramSpecFactory>);
+
+template <typename Factory>
+using SpecAdapter =
+    device_operation::MeshDeviceOperationAdapter<ProgramSpecMinimalOp>::ProgramSpecMeshWorkloadFactoryAdapter<Factory>;
+
+// The mesh-wide vs per-coordinate build path is picked by this predicate; pin both directions.
+static_assert(SpecAdapter<PerCoordProgramSpecFactory>::create_program_artifacts_uses_mesh_dispatch_coordinate());
+static_assert(SpecAdapter<PerCoordCustomProgramSpecFactory>::create_program_artifacts_uses_mesh_dispatch_coordinate());
+static_assert(!SpecAdapter<ProgramSpecFactory>::create_program_artifacts_uses_mesh_dispatch_coordinate());
+static_assert(!SpecAdapter<CustomProgramSpecFactory>::create_program_artifacts_uses_mesh_dispatch_coordinate());
+
 // Compile-coverage: taking the adapter methods' addresses ODR-uses them, forcing
 // the (otherwise un-instantiated) bodies to compile. Never dispatched.
 TEST(LaunchOperationTest, ProgramSpecAdapterCompiles) {
@@ -180,6 +225,13 @@ TEST(LaunchOperationTest, ProgramSpecAdapterCompiles) {
         ProgramSpecMinimalOp>::CustomProgramSpecMeshWorkloadFactoryAdapter<CustomProgramSpecFactory>;
     [[maybe_unused]] auto ccreate = &CustomAdapter::create_mesh_workload;
     [[maybe_unused]] auto capply = &CustomAdapter::apply_descriptor;
+
+    // The per-coordinate build path is a separate if-constexpr branch; instantiate it too.
+    [[maybe_unused]] auto pcreate = &SpecAdapter<PerCoordProgramSpecFactory>::create_mesh_workload;
+    using PerCoordCustomAdapter = device_operation::MeshDeviceOperationAdapter<
+        ProgramSpecMinimalOp>::CustomProgramSpecMeshWorkloadFactoryAdapter<PerCoordCustomProgramSpecFactory>;
+    [[maybe_unused]] auto pccreate = &PerCoordCustomAdapter::create_mesh_workload;
+    [[maybe_unused]] auto pcapply = &PerCoordCustomAdapter::apply_descriptor;
     SUCCEED();
 }
 
--- a/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
+++ b/tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
@@ -146,31 +146,32 @@ struct CustomProgramSpecFactory {
     }
 };
 
-// Per-coordinate variants of both flavors: create_program_artifacts takes the MeshCoordinate, so the
-// adapter calls it once per coordinate and each call may return its own spec and run args.
-struct PerCoordProgramSpecFactory {
-    static ttnn::device_operation::ProgramArtifacts create_program_artifacts(
+// Programs vary across the mesh: one call returns workload-scoped resources plus a program per
+// coordinate range.
+struct MeshWorkloadSpecFactory {
+    static ttnn::device_operation::MeshWorkloadArtifacts create_mesh_workload_artifacts(
         const OperationAttributes& /*attrs*/,
         const Tensor& /*tensor_args*/,
         Tensor& /*tensor_return_value*/,
-        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
-        return ttnn::device_operation::ProgramArtifacts{};
+        const ttnn::MeshCoordinateRangeSet& /*tensor_coords*/) {
+        return ttnn::device_operation::MeshWorkloadArtifacts{};
     }
 };
 
-struct PerCoordCustomProgramSpecFactory {
-    static ttnn::device_operation::ProgramArtifacts create_program_artifacts(
+// Same, plus a per-coordinate cache-hit run-args refresh.
+struct MeshWorkloadSpecFactoryWithOverride {
+    static ttnn::device_operation::MeshWorkloadArtifacts create_mesh_workload_artifacts(
         const OperationAttributes& /*attrs*/,
         const Tensor& /*tensor_args*/,
         Tensor& /*tensor_return_value*/,
-        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
-        return ttnn::device_operation::ProgramArtifacts{};
+        const ttnn::MeshCoordinateRangeSet& /*tensor_coords*/) {
+        return ttnn::device_operation::MeshWorkloadArtifacts{};
     }
     static tt::tt_metal::experimental::ProgramRunArgs override_runtime_arguments(
         const OperationAttributes& /*attrs*/,
         const Tensor& /*tensor_args*/,
         Tensor& /*tensor_return_value*/,
-        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/ = std::nullopt) {
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
         return {};
     }
 };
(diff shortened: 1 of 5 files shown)


ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -891,49 +920,47 @@ struct MeshDeviceOperationAdapter {
...
 
-            // Apply the factory's initial ProgramRunArgs to each MeshWorkload program.
-            for (auto& [range, program] : mesh_workload.get_programs()) {
-                tt::tt_metal::experimental::SetProgramRunArgs(program, artifacts.run_params);
+            if constexpr (create_program_artifacts_uses_mesh_dispatch_coordinate()) {
+                for (const auto& coord : tensor_coords.coords()) {
+                    auto artifacts = SpecFactory::create_program_artifacts(
+                        attrs, tensor_args, tensor_return_value, std::optional<ttnn::MeshCoordinate>(coord));
+                    // No kernels means no work at this coordinate, as in the ProgramDescriptor path.
+                    if (artifacts.spec.kernels.empty()) {
```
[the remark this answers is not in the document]
dgomezTT: Fixed in b2c37870e. The skip now requires every resource group to be empty, matching the descriptor sibling at line 588 rather than only claiming to:

```cpp
if (spec.kernels.empty() && spec.dataflow_buffers.empty() &&
    spec.cross_node_dataflow_buffers.empty() && spec.semaphores.empty() &&
    spec.scratchpads.empty() && spec.work_units.empty()) { continue; }
```

A partially populated spec now falls through to MakeProgramFromSpec, which validates and rejects it, instead of the coordinate being silently dropped. The comment claiming parity with the ProgramDescriptor path is now true.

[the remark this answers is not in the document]
dgomezTT: Fixed in b2c37870e. The skip now requires every resource group to be empty, matching the descriptor sibling at line 588 rather than only claiming to:

```cpp
if (spec.kernels.empty() && spec.dataflow_buffers.empty() &&
    spec.cross_node_dataflow_buffers.empty() && spec.semaphores.empty() &&
    spec.scratchpads.empty() && spec.work_units.empty()) { continue; }
```

A partially populated spec now falls through to MakeProgramFromSpec, which validates and rejects it, instead of the coordinate being silently dropped. The comment claiming parity with the ProgramDescriptor path is now true.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -891,49 +920,47 @@ struct MeshDeviceOperationAdapter {
...
+                        continue;
```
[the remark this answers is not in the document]
dgomezTT: Added in b2c37870e: `LaunchOperation2x4Test.ProgramSpecAdapterPerCoordinateSkipsIdleCoords`, a runtime test on the existing 2x4 fixture.

It wraps the real ProdAll factory in a coordinate-aware one that returns an empty spec on odd columns and a real spec elsewhere, so a single `create_mesh_workload` call exercises coordinate enumeration, per-coordinate spec/run-arg mapping, and the empty-spec skip together. It asserts presence/absence per coordinate rather than just a count, so a regression that dispatched to the wrong range would fail rather than pass on an accidental total, and `static_assert(Adapter::create_program_artifacts_uses_mesh_dispatch_coordinate())` pins that the per-coord branch is the one under test.

One honest limit: the 2x4 fixture needs 8 devices, so it SKIPs on a 2-device host along with the five pre-existing `LaunchOperation2x4Test` cases. It compiles and registers locally; it executes wherever that fixture runs.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -891,49 +920,47 @@ struct MeshDeviceOperationAdapter {
...
+                    }
+                    const ttnn::MeshCoordinateRange range(coord);
+                    shared_variables.emplace(range, make_shared_variables(artifacts, io_mesh_tensors));
```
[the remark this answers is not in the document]
dgomezTT: Added in b2c37870e: `LaunchOperation2x4Test.ProgramSpecAdapterPerCoordinateSkipsIdleCoords`, a runtime test on the existing 2x4 fixture.

It wraps the real ProdAll factory in a coordinate-aware one that returns an empty spec on odd columns and a real spec elsewhere, so a single `create_mesh_workload` call exercises coordinate enumeration, per-coordinate spec/run-arg mapping, and the empty-spec skip together. It asserts presence/absence per coordinate rather than just a count, so a regression that dispatched to the wrong range would fail rather than pass on an accidental total, and `static_assert(Adapter::create_program_artifacts_uses_mesh_dispatch_coordinate())` pins that the per-coord branch is the one under test.

One honest limit: the 2x4 fixture needs 8 devices, so it SKIPs on a 2-device host along with the five pre-existing `LaunchOperation2x4Test` cases. It compiles and registers locally; it executes wherever that fixture runs.

tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
```diff
@@ -146,6 +146,35 @@ struct CustomProgramSpecFactory {
...
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/) {
+        return ttnn::device_operation::ProgramArtifacts{};
+    }
+    static tt::tt_metal::experimental::ProgramRunArgs override_runtime_arguments(
+        const OperationAttributes& /*attrs*/,
+        const Tensor& /*tensor_args*/,
+        Tensor& /*tensor_return_value*/,
+        const std::optional<ttnn::MeshCoordinate>& /*mesh_dispatch_coordinate*/ = std::nullopt) {
+        return {};
+    }
```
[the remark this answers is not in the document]
dgomezTT: Added in b2c37870e: `LaunchOperation2x4Test.ProgramSpecAdapterPerCoordinateSkipsIdleCoords`, a runtime test on the existing 2x4 fixture.

It wraps the real ProdAll factory in a coordinate-aware one that returns an empty spec on odd columns and a real spec elsewhere, so a single `create_mesh_workload` call exercises coordinate enumeration, per-coordinate spec/run-arg mapping, and the empty-spec skip together. It asserts presence/absence per coordinate rather than just a count, so a regression that dispatched to the wrong range would fail rather than pass on an accidental total, and `static_assert(Adapter::create_program_artifacts_uses_mesh_dispatch_coordinate())` pins that the per-coord branch is the one under test.

One honest limit: the 2x4 fixture needs 8 devices, so it SKIPs on a 2-device host along with the five pre-existing `LaunchOperation2x4Test` cases. It compiles and registers locally; it executes wherever that fixture runs.

dgomezTT:
Pushed b2c37870e addressing the review, plus one issue nobody raised.

**Empty-spec check (SHOULD-FIX x2 + design bot).** The skip tested only `kernels.empty()` while claiming parity with the ProgramDescriptor path, which checks all three of its collections. It now requires every `ProgramSpec` resource group to be empty — `kernels`, `dataflow_buffers`, `cross_node_dataflow_buffers`, `semaphores`, `scratchpads`, `work_units`. A partially populated spec falls through to `MakeProgramFromSpec`, which rejects it, rather than the coordinate being silently dropped with its resources.

**Runtime coverage (Copilot, /tdd, tt-test-coverage-review).** Added `LaunchOperation2x4Test.ProgramSpecAdapterPerCoordinateSkipsIdleCoords`. It wraps the real ProdAll factory so odd columns return an empty spec, so one `create_mesh_workload` call covers coordinate enumeration, per-coordinate mapping, and the skip path. Assertions are per-coordinate presence/absence, not a total, so dispatching to the wrong range fails rather than passing on a coincidental count.

**Overload-set hazard (not raised in review).** Both the concept comment and the adapter probe said "declare exactly one signature" but nothing enforced it: an overload set makes `&T::create_program_artifacts` ambiguous, which `ProgramSpecFactoryConcept` keys on, so a factory declaring both would have silently stopped matching. Now a `static_assert` catches it, keyed on the signature where the two probes disagree — the call expression succeeds while address-of fails.

Verification: `./build_metal.sh --disable-profiler` clean (1163 compiles) and `--build-ttnn-tests` clean (363), which also confirms the new `static_assert` does not misfire on any of the five factory shapes the test file instantiates. The new test SKIPs on this 2-device host, as do the five pre-existing `LaunchOperation2x4Test` cases; it needs the 8-device fixture to execute.

Unrelated to the diff: the ASan job on the previous head failed with its `Compile` step never started and no log blob — a lost runner, not a build error.

bbradelTT (approved):

dgomezTT:
/codeowners ping Sanity, Nightly L2 and Runtime Sanity are all green at b2c37870e, all review threads resolved. Ready for codeowner review.

akerteszTT (commented):
I'm not convinced this quite works as intended.

The `ProgramArtifacts` object as written is designed for an SPMD op. The thing that you want to allow one-per-MeshCoordinate of must not contain an op-owned tensors vector: 
 - We don't want to allow op-owned tensors on the `MeshWorkload` concept
 - If we did, it would need to be MeshWorkload-wide, not per-device

Also, the `MeshWorkload` factory concept will need to support `GlobalSemaphore` (which is still as-yet unsupported in Device 2.0 / Host 2.0). You need to create a dedicated `MeshWorkloadArtifacts` object for this to work. Since we know it's coming, I recommend stubbing it in now.

Lastly, how does this factory concept handle the case where there is one Program that wants to have device-specific `ProgramRunArgs`? The current implementation appears to force a separate `ProgramSpec` per `MeshCoordinate`. If that's at parity with the existing `MeshWorkloadFactoryConcept` and `WorkloadFactoryConcept`, that's fine -- but worth checking.

Previous TTNN iterations have used a dedicated MeshWorkload concept... might be totally unnecessary, but it's a deviation that's worth explaining the design decision for.

dgomezTT:
Re-dispatched Sanity, Nightly L2 and Runtime Sanity at `632072ffd` — the green runs were against `b2c37870e`, so they no longer applied.

Two notes on the rework.

**The earlier resolved threads describe code that is no longer in the diff.** The five threads resolved on `b2c37870e` were answered by widening the per-coordinate empty-spec check to all six `ProgramSpec` resource groups. `create_mesh_workload_artifacts` dissolves that concern in a better way — the factory returns only the coordinate ranges it wants, so there is no emptiness predicate left to get wrong, and `TT_FATAL(!artifacts.programs.empty(), ...)` catches the degenerate case. Flagging it only so a reviewer re-reading those threads is not hunting for a check that is intentionally gone.

**The overload-set hazard came back in a new form, and it is now silent.** `MeshWorkloadSpecFactoryConcept` keys on `requires { &T::create_mesh_workload_artifacts; }`, and address-of an overload set is ambiguous, so that requirement is `false` for a factory that declares two signatures. On its own that is caught: no concept matches, the sum in `all_factories_valid` is 0, and the `static_assert` fires.

The bad case is a factory declaring a single `create_program_artifacts` **and** an overloaded `create_mesh_workload_artifacts`. Then `MeshWorkloadSpecFactoryConcept` is false, which makes the newly added `!MeshWorkloadSpecFactoryConcept<T>` clause in `ProgramSpecFactoryConcept` true, so `ProgramSpecFactoryConcept` matches, the sum is exactly 1, and it compiles. The factory silently takes the SPMD path: one `ProgramSpec` replicated across the mesh instead of a program per coordinate — which is the specific bug this change set exists to make expressible.

That is not hypothetical shape-wise: the natural way to want both call shapes is an overload with and without the coordinate argument, which is what the previous revision's `mesh_dispatch_coordinate` parameter did. A `static_assert` keyed on "declares `create_mesh_workload_artifacts` in a form whose address cannot be taken" would close it, the same way the one on the previous revision did. Leaving the call to you since the factory split is your design.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -996,6 +1002,117 @@ struct MeshDeviceOperationAdapter {
...
+        using ResolvedTensorBinding = typename SpecBindingSupport::ResolvedTensorBinding;
+
+        struct shared_variables_t {
+            std::vector<ResolvedTensorBinding> bindings;
+            // Workload-scoped, shared by every range: see MeshWorkloadArtifacts.
+            std::shared_ptr<std::vector<tt::tt_metal::MeshTensor>> op_owned_tensors;
+        };
+        using cached_mesh_workload_t = AdaptedCachedMeshWorkload<shared_variables_t>;
+
+        static consteval bool has_override_runtime_arguments() {
```
ayerofieiev-tt, line 1052: **Bug (silent): a near-miss `override_runtime_arguments` is ignored instead of rejected.**

`has_override_runtime_arguments()` only says "yes" when the factory has an override that takes exactly `(attrs, tensor_args, tensor_return_value, const std::optional<MeshCoordinate>&)`. If a factory author writes the 3-argument form (no coordinate), or takes `const MeshCoordinate&` instead of the optional, this returns `false`, the code compiles, and the hit path quietly falls back to the tensor-only refresh. Run args then go stale on every cache hit and nothing tells you why.

The sibling Custom adapter does not have this problem: `detail::HasSpecRuntimeArgsOverride` matches on the return type, and the 4-argument call in `CustomProgramSpecMeshWorkloadFactoryAdapter::apply_descriptor` fails to compile on a mismatch, so the mistake is loud there.

Suggested fix: add a static check so "has an override" and "the adapter can call it" cannot disagree, e.g.

```cpp
static_assert(
    !detail::HasSpecRuntimeArgsOverride<WorkloadSpecFactory> || has_override_runtime_arguments(),
    "override_runtime_arguments must take (attrs, tensor_args, tensor_return_value, const std::optional<MeshCoordinate>&)");
```

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -996,6 +1002,117 @@ struct MeshDeviceOperationAdapter {
...
+                std::make_shared<std::vector<tt::tt_metal::MeshTensor>>(std::move(artifacts.op_owned_tensors));
+            auto mesh_tensors = SpecBindingSupport::collect_mesh_tensors(tensor_args, tensor_return_value);
+            for (const auto& op_owned_tensor : *op_owned_tensors) {
+                mesh_tensors.push_back(std::cref(op_owned_tensor));
+            }
+
+            std::unordered_map<ttnn::MeshCoordinateRange, tt::tt_metal::experimental::ProgramSpec> program_specs;
+            std::unordered_map<ttnn::MeshCoordinateRange, tt::tt_metal::experimental::ProgramRunArgs> run_params;
+            std::unordered_map<ttnn::MeshCoordinateRange, shared_variables_t> shared_variables;
+            for (auto& per_coord : artifacts.programs) {
```
ayerofieiev-tt, line 1085: **Bug (silent): the adapter does not check the ranges the factory hands back.**

Two cases slip through here:

1. Two `PerCoordProgram` entries with the *same* range. `unordered_map::emplace` keeps the first and silently drops the second for all three maps. Metal never sees the duplicate, so its overlap check (`MeshWorkloadImpl::add_program`) cannot catch it. One program just disappears.
2. A range that is inside the mesh but *outside* `tensor_coords`. `MakeMeshWorkloadFromSpecs` only checks the mesh extent, so a program gets built for a device that holds no shard of the tensors it binds.

Suggested fix: fatal on both, since the factory got `tensor_coords` and is expected to stay inside it:

```cpp
const auto [_, inserted] = program_specs.emplace(per_coord.range, std::move(per_coord.spec));
TT_FATAL(inserted, "create_mesh_workload_artifacts returned range {} twice", per_coord.range);
TT_FATAL(tensor_coords.contains(per_coord.range) /* or a per-coordinate check */,
         "range {} is not covered by tensor_coords {}", per_coord.range, tensor_coords);
```
(Overlapping-but-not-equal ranges are already caught by Metal, so only equality and containment need checking here.)

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -996,6 +1002,117 @@ struct MeshDeviceOperationAdapter {
...
+            const tensor_args_t& tensor_args,
+            tensor_return_value_t& tensor_return_value) {
+            const bool skip_validation = !ttnn::CONFIG.get<"validate_program_args">();
+            if constexpr (has_override_runtime_arguments()) {
+                for (auto& [coordinate_range, program] : cached_workload.workload.get_programs()) {
+                    auto run_args = WorkloadSpecFactory::override_runtime_arguments(
+                        attrs,
+                        tensor_args,
+                        tensor_return_value,
+                        std::optional<ttnn::MeshCoordinate>(coordinate_range.start_coord()));
```
ayerofieiev-tt: **Design: the override only sees the first coordinate of a range.**

A `PerCoordProgram` can cover many devices (the header says "a program uniform over part of the mesh is one entry covering that range"). But on a cache hit the factory's `override_runtime_arguments` is given `start_coord()` only. If the factory derives device-specific run args from the coordinate, every other device in that range gets the args of the first one.

The Custom adapter has the same shape, but there the range always comes from `tensor_coords`, not from the factory. Here the factory chose the range itself, so handing it a single coordinate loses information it owns.

Options: pass the `MeshCoordinateRange` (or both) to the override, or document clearly at the concept that per-device run args require single-coordinate ranges. Either way, the comment above ("calls override_runtime_arguments(attrs, args, ret, coord)") should say it is the range's start coordinate.

ttnn/api/ttnn/operation_concepts.hpp
```diff
@@ -115,11 +115,23 @@ concept HasSpecRuntimeArgsOverride = requires {
...
 }  // namespace detail
 
-// Base spec factory: cache hit refreshes only tensor bindings.
+// A factory whose programs vary across the mesh: one call returns the workload-scoped resources
+// plus a ProgramSpec and ProgramRunArgs per coordinate range.
+//   static MeshWorkloadArtifacts create_mesh_workload_artifacts(
+//       const operation_attributes_t&, const tensor_args_t&, tensor_return_value_t&,
+//       const ttnn::MeshCoordinateRangeSet& tensor_coords);
+template <typename T>
+concept MeshWorkloadSpecFactoryConcept = requires {
```
ayerofieiev-tt, line 130: **Agree with your PR comment on the overload-set hole; suggest closing it in this PR.**

Restating for the thread: `&T::create_mesh_workload_artifacts` is ill-formed when that name is an overload set, so this concept is `false` for such a factory. If the same factory also has a single `create_program_artifacts`, `ProgramSpecFactoryConcept` becomes `true` (because of the new `!MeshWorkloadSpecFactoryConcept<T>` clause), `AllFactoriesValid` sums to exactly 1, and the factory silently takes the SPMD path. That is the exact bug this PR exists to make expressible, so it should not be able to happen quietly.

A cheap guard: detect "the name exists but its address cannot be taken" and `static_assert` on it, the same way the previous revision did for `create_program_artifacts`. Since you own the factory split, I would rather land it here than in a follow-up.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -974,6 +978,8 @@ struct MeshDeviceOperationAdapter {
     // Like ProgramSpecMeshWorkloadFactoryAdapter (cache-miss build inherited unchanged), but the
     // cache-hit path calls the factory's override_runtime_arguments and applies the returned
     // ProgramRunArgs via UpdateProgramRunArgs instead of the base's tensor-only UpdateTensorArgs.
+    // The miss path applies create_program_artifacts' run args, not the override, so per-coordinate
+    // run args must come from a per-coordinate create_program_artifacts.
```
ayerofieiev-tt: **Stale comment.** There is no "per-coordinate `create_program_artifacts`" anywhere at this revision (that was the previous design). A factory author reading this will look for a signature that does not exist. It should say: "per-coordinate run args must come from a `MeshWorkloadSpecFactoryConcept` factory (`create_mesh_workload_artifacts`)."

Same family, smaller: the `TT_FATAL` messages in `SpecBindingSupport::source_mesh_device` (line ~824) still say "ProgramSpec factory adapter", but the helper now serves the mesh-workload adapter too.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
```diff
@@ -996,6 +1002,117 @@ struct MeshDeviceOperationAdapter {
...
+                for (auto& [coordinate_range, program] : cached_workload.workload.get_programs()) {
+                    auto run_args = WorkloadSpecFactory::override_runtime_arguments(
+                        attrs,
+                        tensor_args,
+                        tensor_return_value,
+                        std::optional<ttnn::MeshCoordinate>(coordinate_range.start_coord()));
+                    tt::tt_metal::experimental::UpdateProgramRunArgs(program, run_args, skip_validation);
+                }
+            } else {
+                const auto io_mesh_tensors = SpecBindingSupport::collect_mesh_tensors(tensor_args, tensor_return_value);
```
ayerofieiev-tt: **Cleanup: this `else` branch is a verbatim copy of `ProgramSpecMeshWorkloadFactoryAdapter::apply_descriptor` (lines ~957-975), and `shared_variables_t` (line 1023) is identical to the one at line 883.**

Since `SpecBindingSupport` was introduced exactly to hold the code both adapters share, both belong there:

- move `shared_variables_t` into `SpecBindingSupport` (both adapters then `using SpecBindingSupport::shared_variables_t;`),
- move the refresh loop into a `SpecBindingSupport::refresh_tensor_bindings(cached_workload, tensor_args, tensor_return_value)` helper and call it from both `apply_descriptor`s.

While there: `if (sv.op_owned_tensors)` is a dead check in the new adapter (the pointer is always set by `create_mesh_workload`), and the three `using TensorParamName/TensorArgument/ResolvedTensorBinding` lines are already inherited.

Not a bug, but the next fix to the hit path will otherwise have to be made twice.

tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
```diff
@@ -349,6 +407,74 @@ TEST_F(LaunchOperation2x4Test, CachingHeterogeneousDispatch) {
...
+                    {.range = ttnn::MeshCoordinateRange(coord),
+                     .spec = std::move(program.spec),
+                     .run_params = std::move(program.run_params)});
+            }
+        }
+        return artifacts;
+    }
+};
+
+TEST_F(LaunchOperation2x4Test, MeshWorkloadSpecAdapterMapsProgramsPerCoordinate) {
```
ayerofieiev-tt, line 440: **Test gap: the new adapter's cache-hit path never runs.**

This test covers `create_mesh_workload` (the miss path) on hardware, which is good. But `MeshWorkloadSpecFactoryAdapter::apply_descriptor` -- both the `override_runtime_arguments` branch and the tensor-refresh branch -- is only instantiated by the compile-coverage test, never executed. That is the path every dispatch after the first goes through, and it is where a silent override fallback would show up.

A cheap extension: after `create_mesh_workload`, build a second input tensor and call `Adapter::apply_descriptor(cached, attrs, new_args, output)`, then check the programs' bound tensors moved. Doing it once with `PerCoordProdAllFactory` and once with a variant that declares `override_runtime_arguments` would pin both branches.

ttnn/api/ttnn/metal_v2_artifacts.hpp
```diff
@@ -33,4 +35,22 @@ struct ProgramArtifacts {
...
+// factory allocates, plus the per-coordinate programs that reference them. Splits the same way
+// tt::tt_metal::WorkloadDescriptor does, for the same reasons.
+struct MeshWorkloadArtifacts {
+    // Parked for the cache entry's lifetime, so the allocation outlives the miss at a stable address.
+    std::vector<tt::tt_metal::MeshTensor> op_owned_tensors;
+
+    // Semaphores for cross-device sync belong here once Metal 2.0 has its own semaphore object.
+
+    // Range-keyed: a program uniform over part of the mesh is one entry covering that range.
+    struct PerCoordProgram {
```
ayerofieiev-tt, line 46: **Naming nit:** `PerCoordProgram` holds a `MeshCoordinateRange`, and the comment right above says a program can cover a whole range. The name says the opposite. `RangeProgram` / `ProgramOnRange` would match what it is, and would stop people from assuming one entry per device.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
dgomezTT: Fixed in b3842250b1f, using your suggested form verbatim — `!detail::HasSpecRuntimeArgsOverride<WorkloadSpecFactory> || has_override_runtime_arguments()`. It works because `HasSpecRuntimeArgsOverride` keys on the return type of `&T::override_runtime_arguments`, so it is satisfied by the 3-argument shape and by the `const MeshCoordinate&` shape, which is exactly the set that used to slip through.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
dgomezTT: Both fatal now, in b3842250b1f.

The duplicate check is your suggestion — `emplace`'s `inserted` flag on `program_specs`, fataling with the range in the message.

For containment I went per-coordinate rather than `tensor_coords.contains(per_coord.range)`: `contains` is a member of `MeshCoordinateRange`, not `MeshCoordinateRangeSet` (mesh_coord.hpp:156-159), and the obvious substitute — `any_of` over `tensor_coords.ranges()` asking each for `contains(range)` — would reject a legitimate range that spans two adjacent entries of the set, since no single entry contains it. Iterating the coordinates of `per_coord.range` and requiring each to be covered by some range in the set gives the check you asked for without that false positive. The message names the offending range, the specific coordinate, and `tensor_coords`.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
dgomezTT: Leaving this one for Diego — it is a design question about the factory contract rather than a defect in the plumbing, and he owns the factory split.

Recording the shape of it so the decision is cheap: a `PerCoordProgram` covering N devices gets exactly one `override_runtime_arguments` call, handed `start_coord()`, and whatever it returns is applied to that one program via `UpdateProgramRunArgs`. That is coherent only while the run args are uniform over the range. A factory that derives anything device-specific from the coordinate silently gets the first device's values replicated across the range. The options are to pass the range instead of a coordinate, or to keep the coordinate and state in the concept's contract that a range must be uniform in its run args — the second costs nothing here but needs saying out loud, since the natural reading of a `std::optional<MeshCoordinate>` parameter is "this is called per device".

ttnn/api/ttnn/operation_concepts.hpp
dgomezTT: Closed in b3842250b1f, in `ProgramSpecMeshWorkloadFactoryAdapter` — the adapter that would otherwise quietly accept such a factory.

The assert is keyed on the disagreement itself: it fires when the factory *is* callable as `create_mesh_workload_artifacts(attrs, tensor_args, tensor_return_value, tensor_coords)` yet still landed in the SPMD adapter, which can only happen when the concept's `&T::create_mesh_workload_artifacts` was ill-formed. A factory that declares the name only as an overload set and nothing else is already caught by `all_factories_valid` (no concept matches, the sum is 0); this covers the case that sums to exactly 1 and compiles.

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
dgomezTT: Fixed in b3842250b1f — it now reads "per-coordinate run args must come from a `MeshWorkloadSpecFactoryConcept` factory's `create_mesh_workload_artifacts`".

ttnn/api/ttnn/mesh_device_operation_adapter.hpp
dgomezTT: Done in b3842250b1f, both pieces into `SpecBindingSupport` as you suggested: `SpecSharedVariables` (both adapters now `using shared_variables_t = typename SpecBindingSupport::SpecSharedVariables`) and `refresh_tensor_args`, which both hit paths call — the SPMD `apply_descriptor` body and the `else` branch here are each one line now.

One incidental fix it forced: `skip_validation` was declared before the `if constexpr`, and with the `else` branch reduced to a delegation it became unused whenever `has_override_runtime_arguments()` is false, so it moved inside the branch that uses it.

tests/ttnn/unit_tests/gtests/test_launch_operation.cpp
dgomezTT: Added in b3842250b1f: `MeshWorkloadSpecAdapterAppliesDescriptorOnCacheHit` drives both branches of `MeshWorkloadSpecFactoryAdapter::apply_descriptor` at runtime.

The tensor-refresh branch re-applies over an already-built workload and asserts the per-range mapping survives. The override branch uses a `PerCoordProdAllFactory` subclass with an `override_runtime_arguments` that counts its calls, and asserts the count equals the number of ranges — so it pins "once per range", not merely "was called", which is the part that would regress silently if the loop were hoisted out.

Verified by local build (1459 compiles, no errors or warnings) and confirmed present in the binary via `--gtest_list_tests`. The body itself skips on this 2-device host, as every `LaunchOperation2x4Test` does; it needs the 8-device fixture to execute, so CI is where it actually runs.

ttnn/api/ttnn/metal_v2_artifacts.hpp
dgomezTT: Your reading of the name is right, but I would rather Diego make this call than rename unilaterally, because the name is not arbitrary: it mirrors `tt::tt_metal::WorkloadDescriptor::PerCoordProgram` (workload_descriptor.hpp:64), which has the same `range` member and whose own comment says "Each entry covers a contiguous MeshCoordinateRange".

So the misnomer originates on the Metal side and the ttnn struct is deliberate parity with it. Renaming only the ttnn one buys accuracy here and costs a vocabulary split across the boundary for the same concept. Worth doing if the Metal-side name is also going to move; otherwise it is a coin flip that Diego should call.

dgomezTT: Keeping this one. The name mirrors tt_metal::WorkloadDescriptor::PerCoordProgram, which has the same range member and the same wording in its comment. Renaming only the ttnn side splits the vocabulary across the boundary for one concept.

dgomezTT:
https://github.com/tenstorrent/tt-metal/actions/runs/33788878420

t3k with t3000-unit

ayerofieiev-tt (approved):

akerteszTT (approved):
This is ultimately a TTNN decision, but I would much prefer it if you dropped op-owned tensors from the MeshWorkload factory concept. It is unnecessary.

My reasoning:
 - This is a genuinely horrible feature, which we would eliminate were it not for the ops that grandfather it in
 - All existing ops that use it are SPMD
 - The Metal 2.0 SPMD factory concept supports it (unlike the PD SPMD concept)

Not a huge deal either way, but why encourage potential new adopters by including it in the new factory concept?

PR #54991, opened 2026-08-31
Fix to_memory_config CB/L1 clash by checking free L1 instead of total L1

### LINKED ISSUE #54621: to_memory_config uses max L1 size instead of free L1, causing CB/L1 clash when a DRAM shard fits total L1 but not free L1
Tentative description, will update with more details and smaller unit test:

Running `test_attention_update_construction.py::test_attention_update_round_trip` on a P100 produced this error: `Statically allocated circular buffers in program 155 clash with L1 buffers on core range [0-0 - 0-0]. L1 buffer allocated at 1378048 and static circular buffer region ends at 1430656`. This runs fine on a P150. What seems to be happening is that `copy_to_buffer` calls `ttnn.to_memory_config` which compares the max L1 size instead of free L1 size. 

https://github.com/tenstorrent/tt-metal/blob/4a5bfad59c63ad061dc050a9374619300789f575/models/common/utility_functions.py#L1266
https://github.com/tenstorrent/tt-metal/blob/4a5bfad59c63ad061dc050a9374619300789f575/ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp#L141-L144
On a P100, the max L1 size is 1.46 MB. In this test, the wqkv's shard, (2048, 448) = 1.84 MB, which exceeds max L1 so it reroutes to generic copy. However, `wo`'s shard, (2048, 320) = 1.31 MB just happens to be below max L1, so it will call `interleaved_to_sharded` which tries to use an L1 CB sized to 1.31 MB. This collides with other buffers.

Does this only affect DRAM-sharded destinations vs L1-sharded?

shard | CB bytes | vs max_l1_size = 1,461,248 | routed to
-- | -- | -- | --
wqkv | 2048×448 | 1,835,008 | over | generic copy (works)
wo | 2048×320 | 1,310,720 | under | interleaved_to_sharded (throws)

On P100, `max_l1_size` is 1,572,864 − 111,616 = 1,461,248.

***

Re-enable the follow tests once this is fixed:
- [ ] test_attention_update_construction.py::test_attention_update_round_trip https://github.com/tenstorrent/tt-metal/pull/54012

ichovpanTT:
Fixes #54621

## Summary

- `can_use_interleaved_to_sharded`, `can_use_sharded_to_interleaved`, and `can_use_reshard` in `to_memory_config_op.cpp` compared CB size against total L1 (`l1_size_per_core - base_addr`), ignoring already-allocated L1 buffers. This caused the fast path to be selected when CBs didn't actually fit, resulting in a CB/L1 clash at dispatch.
- Replace the static computation with `get_max_l1_space()`, which queries `lowest_occupied_compute_l1_address` for actual free L1. This is the same utility already used by matmul, untilize, concat, and other ops.

## Reproducer

See [standalone script on the issue](https://github.com/tenstorrent/tt-metal/issues/54621#issuecomment-5483033500). Before the fix it crashes on P100; after the fix it falls back to `ttnn::prim::copy`.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ichovpan/fix-to-memory-config-l1-check)

### DIFF
--- a/tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
+++ b/tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
@@ -2732,3 +2732,96 @@ def test_to_memory_config_tile_interleaved_l1_dram(device, src_buffer, dst_buffe
     assert output_tensor.memory_config().buffer_type == dst_buffer
     output_torch = ttnn.to_torch(output_tensor)
     assert_equal(torch_input, output_torch)
+
+
+def test_to_memory_config_interleaved_to_sharded_with_resident_l1(device):
+    """Regression #54621: interleaved_to_sharded fast-path selection must budget CB size
+    against LIVE free L1, not against the whole of L1 (``l1_size_per_core - base_addr``).
+
+    Statically allocated CBs grow upward from the allocator's base L1 address while L1
+    tensors are allocated downward from the top of L1, so the space actually available to
+    a program's CBs is ``lowest_occupied_compute_l1_address() - base``. Before this fix,
+    ``can_use_interleaved_to_sharded`` in ``to_memory_config_op.cpp`` compared against the
+    whole-of-L1 budget, so the fast path was still selected when a resident L1 buffer left
+    insufficient space, and the program's static CB region clashed with the resident buffer
+    at dispatch:
+
+        Statically allocated circular buffers in program N clash with L1 buffers on core range ...
+
+    The fix routes the eligibility check through ``get_max_l1_space()``, which subtracts
+    ``lowest_occupied_compute_l1_address()``. This test reserves L1 so that the required CB
+    for the DRAM-sharded destination would fit under the old total-L1 budget but not under
+    free L1, and asserts that ``to_memory_config`` still succeeds (by falling back to
+    ``ttnn.prim.copy``) rather than clashing.
+
+    Asserts only on the result, never on which implementation served it: a future refactor
+    that goes back to whole-of-L1 budgeting would compile fine and pass the other tests in
+    this file, but would crash here at dispatch.
+    """
+    torch.manual_seed(0)
+    TILE_BYTES = 2048  # bfloat16 tile
+
+    info = ttnn._ttnn.reports.get_device_info(device)
+
+    # DRAM-sharded destinations go through the ``dst_is_dram`` branch of
+    # ``can_use_interleaved_to_sharded``, where the CB is sized to the full shard
+    # (num_units_per_shard * output_page_size). Size the shard so its CB is 70% of the
+    # whole-of-L1 budget the buggy check used; the resident reservation below then pushes
+    # free L1 below that CB size, so a whole-of-L1 check accepts the fast path while a
+    # free-L1 check must reject it.
+    cb_target_bytes = int(info.cb_limit * 0.70)
+    shard_tiles = cb_target_bytes // TILE_BYTES
+    if shard_tiles < 4:
+        pytest.skip(f"device cb_limit {info.cb_limit} B is too small for this regression test")
+
+    # Width-shard across every DRAM bank: shard height spans the whole tensor height and
+    # shard width is one tile column per bank, so num_units_per_shard = shard_tiles.
+    num_dram_banks = device.dram_grid_size().x
+    shape = [1, 1, 32 * shard_tiles, 32 * num_dram_banks]
+    torch_input = torch.randn(shape, dtype=torch.bfloat16)
+    # Built before the L1 reservation so any device-side tilize in from_torch runs with the
+    # usual amount of L1 available; only the to_memory_config under test should see the pressure.
+    input_tensor = ttnn.from_torch(
+        torch_input,
+        dtype=ttnn.bfloat16,
+        layout=ttnn.TILE_LAYOUT,
+        device=device,
+        memory_config=ttnn.DRAM_MEMORY_CONFIG,
+    )
+    shard_grid = make_full_dram_core_range_set(device)
+    shard_shape = (32 * shard_tiles, 32)
+    shard_spec = ttnn.ShardSpec(shard_grid, shard_shape, ttnn.ShardOrientation.ROW_MAJOR)
+    output_mem_config = ttnn.MemoryConfig(
+        ttnn.TensorMemoryLayout.WIDTH_SHARDED,
+        ttnn.BufferType.DRAM,
+        shard_spec,
+    )
+
+    # Reserve L1 so the fast-path CB is oversized against free L1 but still under the
+    # whole-of-L1 budget. Interleaved L1 spreads pages round-robin across every bank, so
+    # tiles_per_bank tiles per bank pushes the lowest occupied L1 address down uniformly.
+    headroom_target = cb_target_bytes // 2  # target free L1 ~ half the fast-path CB
+    tiles_per_bank = (info.cb_limit - headroom_target) // TILE_BYTES
+    if tiles_per_bank <= 0:
+        pytest.skip("device L1 too small to leave a meaningful headroom window")
+
+    resident = ttnn.allocate_tensor_on_device(
+        ttnn.Shape([1, 1, 32 * tiles_per_bank, 32 * info.l1_num_banks]),
+        ttnn.bfloat16,
+        ttnn.TILE_LAYOUT,
+        device,
+        ttnn.MemoryConfig(ttnn.TensorMemoryLayout.INTERLEAVED, ttnn.BufferType.L1),
+    )
+    try:
+        # Guard against a vacuous pass: if the reservation did not land where intended
+        # there is no L1 pressure to regress against.
+        actual_free_l1 = resident.buffer_address() - info.address_at_first_l1_cb_buffer
+        assert actual_free_l1 < cb_target_bytes, (
+            f"resident L1 buffer left {actual_free_l1} B free above the CB base; the "
+            f"regression scenario needs free L1 < fast-path CB size {cb_target_bytes} B"
+        )
+
+        output_tensor = ttnn.to_memory_config(input_tensor, memory_config=output_mem_config)
+        assert_equal(torch_input, ttnn.to_torch(output_tensor))
+    finally:
+        ttnn.deallocate(resident)
--- a/ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
+++ b/ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
@@ -11,6 +11,7 @@
 #include "ttnn/operations/data_movement/sharded/sharded_to_interleaved/device/sharded_to_interleaved_device_operation.hpp"
 #include "ttnn/operations/data_movement/sharded/reshard/device/reshard_device_operation.hpp"
 #include "ttnn/operations/data_movement/copy/device/copy_device_operation.hpp"
+#include "ttnn/operations/data_movement/common/common.hpp"
 #include "ttnn/types.hpp"
 #include <tt-metalium/tt_align.hpp>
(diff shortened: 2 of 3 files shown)


bbradelTT (approved):

bbradelTT (approved):

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
```diff
@@ -138,8 +138,7 @@ bool can_use_interleaved_to_sharded(
         total_cb_size += num_units_per_shard * output_page_size;
     }
 
-    uint32_t max_l1_size =
-        device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+    uint32_t max_l1_size = operations::data_movement::get_max_l1_space(input_tensor);
```
ayerofieiev-tt, line 141: This check runs before the op allocates its own output. When the destination is L1, `create_output_tensors` later allocates one shard per core in L1, which lowers the free-L1 frontier by `num_units_per_shard * output_page_size` bytes. The static CBs counted here (the 4-page scratch CB always, plus the input CB when `convert_df` is true) must still fit under that new frontier at enqueue time, but this check compares them against the frontier measured now.

So the fast path can still be picked when it does not fit, and dispatch throws the same "clash with L1 buffers" error this PR is fixing. The PR's reproducer uses a DRAM destination, which is why it passes.

Suggested fix: when `!dst_is_dram`, subtract the L1 output shard size from `max_l1_size` before comparing. Note `get_pending_l1_output_reservation` returns 0 for sharded outputs, so it cannot be reused as-is here. The same applies to the quasar copy of this function.

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
```diff
@@ -59,8 +60,7 @@ bool can_use_sharded_to_interleaved(
         // Input CB aliases the shard buffer (no extra L1). Output CB is additional.
         uint32_t total_cb_size = num_units_per_shard * aligned_output_page_size;
 
-        uint32_t max_l1_size =
-            device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+        uint32_t max_l1_size = operations::data_movement::get_max_l1_space(input_tensor);
```
ayerofieiev-tt, line 63: Same ordering problem as the interleaved_to_sharded check. When `output_mem_config` is L1 interleaved and `convert_df` is true, the op allocates the interleaved L1 output buffer after this check, and the separate `OUT_DFB` (sized `num_units_per_shard * aligned_output_page_size`) must fit under the lowered frontier at enqueue. Comparing against free L1 measured before that allocation can still choose the fast path when it does not fit.

For an interleaved L1 output the codebase already has a helper for this: `get_pending_l1_output_reservation` in `data_movement/common/common.cpp`. `tilize.cpp` (line 72) and `untilize.cpp` (line 72) subtract it from `get_max_l1_space` for exactly this reason. Suggest doing the same here.

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
```diff
@@ -212,8 +211,7 @@ bool can_use_reshard(
         bool unaligned = (remote_unit_size_padded != unit_size) || (local_unit_size_padded != unit_size);
         if (unaligned) {
             uint32_t scratch_cb_size = remote_units_per_shard * remote_unit_size_padded;
-            uint32_t max_l1_size =
-                device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+            uint32_t max_l1_size = operations::data_movement::get_max_l1_space(input_tensor);
```
ayerofieiev-tt, line 214: In `reshard_program_factory_same_width.cpp` the scratch DFB is only created when `use_scratch = unaligned && local_is_output`, i.e. only when the output is L1. In that same case the output shard is allocated in L1 after this check runs, so the frontier drops by the output shard size before the scratch DFB has to fit. Suggest subtracting the L1 output shard bytes here too, or at least gating this check on `local_is_output` and accounting for the shard.

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
```diff
@@ -138,8 +138,7 @@ bool can_use_interleaved_to_sharded(
         total_cb_size += num_units_per_shard * output_page_size;
     }
 
-    uint32_t max_l1_size =
-        device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+    uint32_t max_l1_size = operations::data_movement::get_max_l1_space(input_tensor);
 
     return total_cb_size < max_l1_size;
```
ayerofieiev-tt, line 143: Not on this line, but related to this check: in the reshard workaround at line 317 (`can_use_interleaved_to_sharded(temp, ...)`), the stricter free-L1 check will now fail more often. When it does, `temp` (a full sharded_to_interleaved result already written to DRAM) is dropped. Control then falls to the two checks below it, which cannot pass in that path: interleaved_to_sharded validation rejects a sharded input, and sharded_to_interleaved validation rejects a sharded output. The code then runs `copy(tensor)` from scratch, so one whole device op was wasted.

Cheaper: in the else case of that `if`, `return ttnn::prim::copy(temp, memory_config, ...)` directly. `temp` is already interleaved DRAM, which `copy` handles.

ttnn/cpp/ttnn/operations/experimental/quasar/to_memory_config/to_memory_config_op.cpp
```diff
@@ -138,8 +138,7 @@ bool can_use_interleaved_to_sharded(
         total_cb_size += num_units_per_shard * output_page_size;
     }
 
-    uint32_t max_l1_size =
-        device->l1_size_per_core() - device->allocator()->get_base_allocator_addr(tt::tt_metal::HalMemType::L1);
+    uint32_t max_l1_size = operations::data_movement::get_max_l1_space(input_tensor);
```
ayerofieiev-tt, line 141: The three comments on the non-quasar `to_memory_config_op.cpp` (L1-destination output shard not accounted for at lines 63, 141, and 214) apply to this file as well, since the code is identical.

tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
```diff
@@ -2732,3 +2732,96 @@ def test_to_memory_config_tile_interleaved_l1_dram(device, src_buffer, dst_buffe
...
+
+    # Reserve L1 so the fast-path CB is oversized against free L1 but still under the
+    # whole-of-L1 budget. Interleaved L1 spreads pages round-robin across every bank, so
+    # tiles_per_bank tiles per bank pushes the lowest occupied L1 address down uniformly.
+    headroom_target = cb_target_bytes // 2  # target free L1 ~ half the fast-path CB
+    tiles_per_bank = (info.cb_limit - headroom_target) // TILE_BYTES
+    if tiles_per_bank <= 0:
+        pytest.skip("device L1 too small to leave a meaningful headroom window")
+
+    resident = ttnn.allocate_tensor_on_device(
```
ayerofieiev-tt, line 2808: Minor: this file uses `pytestmark = pytest.mark.use_module_device`, so the device is shared across every test in the module. This allocation reserves roughly 65% of each L1 bank. If any earlier test leaves an L1 tensor resident (a held Python reference, or a failure before its deallocate), this call raises Out of Memory before the `try`, and the test errors instead of exercising the regression.

A cheap guard: compute free L1 before allocating (from `device.lowest_occupied_compute_l1_address` or the allocator stats) and `pytest.skip` if the window is not there, so a leak upstream shows up as a skip rather than an unrelated OOM.

ayerofieiev-tt (commented):
Switching these checks to free L1 is the right direction and fixes the DRAM-destination case from #54621. Inline comments cover the L1-destination cases where the check still runs too early, plus one wasted-work path and a test fragility note.

One follow-up thought, not blocking: the fallback this PR routes to, `ttnn::prim::copy`, still sizes its own CBs against total L1 in `copy_device_operation.cpp` (line 58), and ten other files under `ttnn/cpp` keep the same `l1_size_per_core() - get_base_allocator_addr` idiom (concat, untilize codegen, several moreh ops, layernorm_post_all_gather). Those are out of scope here but worth a tracking issue so the fallback is not the next thing to clash.

vtsilytskyiTT:
@ichovpanTT this change is useful. Any updates?

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
ichovpanTT: Addressed in b1d6361d7a2f07c5a1022dd49a9b3e5724254400

tests/ttnn/unit_tests/base_functionality/test_to_memory_config.py
ichovpanTT: Good catch, Rewrote the test, now the size of the L1 blocker and tensor to call to_memory_config on are based of the available L1 space, not total L1 space.

Fixed in 09d7042b5306bad5eaf7a6fab289082d7c7bd7a6

ttnn/cpp/ttnn/operations/experimental/quasar/to_memory_config/to_memory_config_op.cpp
ichovpanTT: See response from 3 comments above

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
ichovpanTT: The patches for this comment and others below are getting complicated to read, they introduce the bugs themselves. Essentially patching this duplicates the complex program logic. I think these checks should be added in another PR, or generalized within ttnn core

ichovpanTT: @riverwuTT

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
ichovpanTT: See response above

ttnn/cpp/ttnn/operations/core/to_memory_config/to_memory_config_op.cpp
ichovpanTT: Tbh I don't really understand what's going on there. Let's patch this in another PR if we decide to

PR #54992, opened 2026-08-31
[Performance] Reduce MeshWorkload host dispatch overhead

### LINKED ISSUE #52756: Host-side EnqueueMeshWorkload cost scales linearly with mesh size (~19 us/device), dominating op2op gap on 32-chip meshes

jbaumanTT:
### PR Category
Performance

### Summary

Closes #52756 

Repeated multi-program `MeshWorkload` enqueues were redoing work that cannot change between launches. This PR removes four such costs from the host submission path.

The first commit set caches derived workload/device metadata, runtime-argument sizes, and direct circular-buffer patch destinations, so the serial enqueue path only refreshes launch-time values.

The follow-up commits come from instrumenting a GLM chunked-prefill CI run per enqueue ([run 33439571293](https://github.com/tenstorrent/tt-metal/actions/runs/33439571293), 8x4 Blackhole Galaxy, 55k@5k, 11 chunks x 10 iters). That log shows 5,518 enqueues per 1.61 s chunk carrying 64,015 programs and 176,576 program-to-device command writes, with 273 ms per chunk spent inside `enqueue_mesh_workload` alone. Three costs in there do no dispatch work:

- `set_program_binary_status` writes an Inspector YAML record on every enqueue. The Inspector is on by default and enqueue re-commits `Committed` each launch, so a mutex and a file append were paid 5,518 times per chunk for a status that changes once. Now only real transitions are reported.
- `RecordProgramSubDevice` and `RecordProgramMetadata` are read back only by the profiler, but ran unconditionally: once per program for the metadata and once per program per device for the sub-device mapping. ttnn stamps a fresh runtime id on every dispatch, so the `std::map` behind the sub-device mapping grows to a node per device per 16-bit id and each insert becomes a tree walk through cold memory. Measured at 311 ns per call and still climbing, against 32 calls per enqueue. Both are now gated on a lock-free predicate, and the enqueue path folds it into `record_sub_device` so the per-device loop is skipped outright.
- `compute_with_storage_grid_size()` and `l1_size_per_core()` answer from a cross-device agreement check that rebuilds the device list and walks the whole mesh. Circular-buffer validation asks for both once per program on every enqueue, so on 32 devices they cost about 3.3 us per program. This alone made `MeshWorkloadImpl::compile()`, which is meant to be a no-op re-validation after the first launch, cost 121 us on a workload carrying one program per device. Both values are fixed once the devices are open, so they are now established on first use and invalidated on reshape.

### Measured
A/B on one 8x4 (32-chip) Wormhole Galaxy, GLM-shaped workloads (32 devices, ~120 worker cores, 12 CBs, ~6 KB of RTAs per device, 20 distinct workloads cycled, fresh runtime id per dispatch):

| workload shape | before | after |
| --- | --- | --- |
| 1 program spanning the mesh | 51.8 us/enqueue | 32.3 us/enqueue (-38%) |
| 32 programs, one per device | 199.4 us/enqueue | 67.9 us/enqueue (-66%) |

The before column also drifts upward across chunks (50.6 -> 52.3 us) as the profiler map fills; after is flat to within 0.03 us.

Applying the measured per-unit costs to GLM's own per-chunk counts gives roughly 295 ms of host submission removed per 1.61 s chunk (27 ms Inspector, 57 ms profiler metadata, 211 ms mesh property lookups).

#### On models

Blaze prefill on an 8x4 Blackhole Galaxy ([run 33534897935](https://github.com/tenstorrent/tt-metal/actions/runs/33534897935)), against the 2026-09-01 nightly on `main`. How much of the removed host time becomes wall time depends on how host-bound the leg is, and the suite contains both kinds:

| leg | metric | before | after |
| --- | --- | --- | --- |
| GLM-5.2-744B chunked prefill 55k@5k | wall, per chunk | 1.696 s | 1.267 s (-25.3%) |
| Kimi-K2.6-1T chunked prefill 55k@5k, untraced | wall, per chunk | 1.043 s | 0.805 s (-22.8%) |
| Kimi-K2.6-1T chunked prefill 55k@5k, traced | wall, per chunk | 0.519-0.855 s | within 0.8% |
| Kimi-K3-2.8T MLA chunked 55k@5k | device, realtime profiler | 11.956 ms | 11.830 ms (-1.1%) |
| Kimi-K2.6-1T MoE 5k | device, realtime profiler | 6.486 ms | 6.283 ms (-3.1%) |

GLM's `tt_forward` drops 354.1 s -> 306.0 s. Per-chunk stddev is 0.007-0.029 s and the four most recent nightlies span 1.680-1.729 s, so the 429 ms shift is an order of magnitude outside the noise; the branch point is two commits behind that nightly, neither touching this path.

`KIMI_UNTRACED_BASELINE_CHUNK_TIMES_S` is re-centered here because all eleven chunks fell out the bottom of a band centered on the old level. The Kimi-K2.6 MoE gate also trips low, by 0.46%, and is left alone: that metric excludes op-to-op dispatch gaps by construction, its shift is inside main's own 3.7% spread, and re-centering it on this branch would fail two of the four most recent main values.

### Notes for reviewers
- `MeshWorkloadImpl` caches immutable post-finalization classifications and config sizes; sub-device IDs remain keyed by mesh and active manager.
- `MeshDeviceImpl` owns the device-range, compute-grid and L1-size caches. The two scalar properties are established once the devices are open and re-established after `reshape()`, so their accessors stay pure reads -- ttnn calls them from many threads without the api lock, and a lazily-filled `mutable` would have raced there.
- Circular-buffer command patch destinations are cached when commands are assembled; the values themselves still come from the `CircularBuffer`, whose `page_size()` / `num_pages()` carry the divisibility and 16-bit page-count checks. Caching only the destinations is what avoids a second owner for the page size.
- Full-mesh workloads avoid building/scanning the unused-device set; partial workloads retain the existing GO-signal path.
- Gating the profiler metadata is safe because both readers already tolerate a miss: `add_program_sub_device_meta_data` returns without annotating and `get_available_worker_core_count_for_program` falls back to the full compute grid. A real-time profiler that activates mid-run starts annotating from the dispatches that follow.
- The gate does not make the recording itself cheaper. With a profiler enabled, `runtime_id_to_sub_device` is still an unpruned `std::map` that grows to roughly 2 M nodes on a 32-device mesh. Worth a follow-up, either an `unordered_map` or the lock-free fixed array already used for `runtime_id_to_kernel_sources_` beside it.
- Not included: fanning the per-device command writes out across the dispatch thread pool. That is where the remaining time goes (1.43 GB per chunk of non-temporal stores at the single-core store ceiling), and a chunked fan-out measured 36.0 -> 27.0 us/enqueue, or 22.1 us if pool workers spin longer before parking. It only helps whole-mesh workloads and needs a global change to thread-pool spin policy, so it belongs in its own PR.

Validation:
- `~/bin-metal/build_metal_cpp.sh`
- `pre-commit run --from-ref origin/main --to-ref HEAD`
- On a 32-chip Galaxy: `MeshWorkloadTestSuite.MeshWorkloadCBUpdate`, all four `MeshSubDeviceTestSuite` tests, `MeshTraceTestSuite`, and both reshape round-trips (`From1x8To2x4ThenBackTo1x8`, `From1x4To2x2Valid`).
- `MeshDeviceFixture.TensixTestUpdateCircularBufferPageSize` (slow dispatch)
- Blaze models prefill, 26 of 30 legs green. Two failures are perf gates tripping their *lower* bound (above); the other two are infra -- a UMD device-enumeration failure at collection and a `Setup multihost job` failure, both on `bh_sc1_high_power`.
- `runtime-sanity-tests` green. `sanity-tests` has one leg red, a pytest-xdist worker crash in `test_lockstep_input_unaffected` on `sim_wh_n150` that passes on the job's own serial re-run.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jbauman/glm-dispatch-host-optimization)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jbauman/glm-dispatch-host-optimization)


tt_metal/impl/program/program.cpp
```diff
@@ -1452,6 +1452,28 @@ std::shared_ptr<CircularBufferImpl> detail::ProgramImpl::get_circular_buffer(CBH
...
+            }
+            for (auto& update : command_sequence.remote_cb_config_updates) {
+                if (update.circular_buffer == circular_buffer.get() && update.buffer_index == buffer_index) {
+                    update.page_size = page_size;
+                }
+            }
+        }
+    };
+    update_cached_page_sizes(cached_program_command_sequences_);
+    update_cached_page_sizes(trace_cached_program_command_sequences_);
```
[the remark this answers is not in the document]
jbaumanTT: Fixed in f05ce95703f. MeshWorkloadCBUpdate now populates both normal and trace command caches before UpdateCircularBufferPageSize, checks the refreshed cached page-size metadata, verifies the normal re-enqueue, and replays a trace whose device kernel reports fifo_page_size/fifo_num_pages. The full C++ build and the focused fast-dispatch test pass.

tt_metal/impl/program/dispatch.cpp
```diff
@@ -3462,7 +3503,8 @@ uint32_t program_base_addr_on_core(
     distributed::MeshWorkloadImpl& mesh_workload,
     distributed::MeshDevice* mesh_device,
     HalProgrammableCoreType programmable_core_type) {
-    const auto& sub_device_ids = mesh_workload.determine_sub_device_ids(mesh_device);
+    const auto& sub_device_ids =
+        mesh_workload.determine_sub_device_ids(mesh_device, *mesh_device->get_active_sub_device_manager_id());
```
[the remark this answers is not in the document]
jbaumanTT: Fixed in f05ce95703f. Address queries now return the HAL kernel-config base before consulting finalized metadata when no fast-dispatch CQ owns a config-ring slot; program-config validation also permits classified service workloads that intentionally bypass finalization. Slow-dispatch CB and semaphore query tests pass.

abhullar-tt (approved):

sidnadTT (approved):

mo-tenstorrent (approved):

pavlepopovic (approved):
Thank you <3

cfjchu (approved):
Approving to unblock

models/demos/deepseek_v3_d_p/tests/test_prefill_transformer_chunked.py
```diff
@@ -228,7 +227,7 @@ def _pad_overrun_summary(seq_len_cache, overruns):
     #
     # If this goes flaky, re-center on the median over several runs before widening. Widening to 10% is
     # the fallback after that -- a band that needs more than 10% is a regression, not noise.
-    (61, 11, 10): [1.043, 1.036, 1.037, 1.047, 1.044, 1.034, 1.034, 1.034, 1.046, 1.041, 1.044],
+    (61, 11, 10): [0.805, 0.808, 0.811, 0.816, 0.808, 0.806, 0.806, 0.805, 0.802, 0.823, 0.854],
```
ayerofieiev-tt, line 230: **Bug risk: this gate can fail a run that is actually faster.**

This baseline comes from one run only. Chunks 0-8 are all about 0.806 s, but chunk 9 is 0.823 and chunk 10 is 0.854 (+6%).

The gate is two-sided, per chunk, with a 5% band (`low <= median <= high`, line ~1324, `UNTRACED_PERF_MARGIN = 0.05`). For chunk 10 the band is `[0.811, 0.897]`. So if the next run is flat at 0.806 for chunk 10, the test fails and reports a "regression" even though it is faster.

The header of this file says to cut the baseline from the median of several runs, not from one run (lines 186-187 and 228). The old comment that warned "a single run is thin evidence" was removed in this PR.

Also two comments are now out of date: line 192 still says the untraced run is "flat ~0.81 s/chunk" (it is not flat any more), and line 222 says "0.33 s (~30%)" - 0.33 s is 41% of 0.81 s, so one of those two numbers is stale.

Suggestion: re-cut chunks 9-10 from 2-3 runs, or confirm the tail ramp is real and fix the "flat" comment.

tt_metal/distributed/mesh_device.cpp
```diff
@@ -949,6 +985,8 @@ void MeshDeviceImpl::reshape(const MeshShape& new_shape) {
     }
     auto new_view = std::make_unique<MeshDeviceView>(new_shape, new_device_order, new_fabric_node_ids);
     view_ = std::move(new_view);
+    local_devices_by_range_.clear();
```
ayerofieiev-tt, line 988: **Race: `reshape()` clears and rewrites the new caches without the api lock.**

`reshape()` does not take `lock_api()`. But:
- `enqueue_mesh_workload` holds a `const std::vector<IDevice*>&` from `get_local_devices()` (fd_mesh_command_queue.cpp:571) while it writes program commands. `local_devices_by_range_.clear()` here frees that vector under the reader -> use-after-free.
- `establish_device_property_caches()` writes `compute_with_storage_grid_size_` / `l1_size_per_core_`. The header comment (mesh_device_impl.hpp:145-148) says these are "a value nobody writes" so ttnn may read them lock-free. `reshape()` does write them.

The old code already had a similar hole (swapping `view_`), so this is not new, but the PR adds a new dangling-reference path and a comment that says it is safe.

Simple fix: take `auto lock = lock_api();` at the top of `MeshDeviceImpl::reshape`, or document that `reshape()` must only run when no other thread is using the device. Also, the comment at mesh_device_impl.hpp:143 says the map is "only ever read" - it is written here and in `get_local_devices` (`try_emplace`). Please fix the wording.

(A previous review raised the lock question in a non-inline note; this adds the concrete dangling-reference case.)

tt_metal/distributed/fd_mesh_command_queue.cpp
```diff
@@ -429,7 +448,8 @@ void FDMeshCommandQueue::enqueue_mesh_workload(MeshWorkload& mesh_workload, bool
         TT_FATAL(!blocking, "Blocking is not supported when recording a trace.");
         trace_nodes_.push_back(MeshTraceNode{});
         auto& trace_node = trace_nodes_.back();
-        bool use_prefetcher_cache = mesh_workload.impl().max_program_kernels_sizeB_ <= this->prefetcher_cache_sizeB_;
+        bool use_prefetcher_cache =
+            mesh_workload.impl().get_finalized_metadata().max_program_kernels_sizeB <= this->prefetcher_cache_sizeB_;
```
ayerofieiev-tt, line 452: **Trace and normal enqueue disagree when a workload has no kernel binaries (pre-existing, but this PR touched both lines).**

Here (trace capture) the check is `max_program_kernels_sizeB <= prefetcher_cache_sizeB_`. In `MeshWorkloadImpl::generate_dispatch_commands` (mesh_workload.cpp:274) the check is `max_program_kernels_sizeB and max_program_kernels_sizeB <= prefetcher_cache_sizeB`.

For a workload with no kernels (`sizeB == 0`, which `load_binaries` says is a valid test case), the normal path picks `false` but the trace path picks `true`. With `true`, `assemble_device_commands` runs `program.get_kernels_buffer(mesh_device)->address()` and the kernels buffer is null for such a program -> crash during trace capture. The normal path never hits this.

Simple fix: use the value already computed by `generate_dispatch_commands`:
```suggestion
        bool use_prefetcher_cache = mesh_workload.impl().use_prefetcher_cache_;
```
This also removes the second spelling of the same value (line 517 already reads `use_prefetcher_cache_`).

tt_metal/distributed/mesh_workload.cpp
```diff
@@ -425,15 +447,15 @@ void MeshWorkloadImpl::finalize_offsets(MeshDevice* mesh_device) {
...
+    const uint32_t max_program_kernels_sizeB = tt::tt_metal::detail::ProgramImpl::finalize_program_offsets(
         extract_context_id(mesh_device),
         mesh_device,
         kernels_getter,
         kernel_groups_getter,
         semaphores_getter,
         programs);
 
-    set_finalized();
+    set_finalized(max_program_kernels_sizeB);
```
ayerofieiev-tt, line 458: **The old safety check is gone, and the programs are never marked finalized.**

Before this PR, `get_program_config_sizes()` re-read every program's sizes on each enqueue and `TT_FATAL`ed if they did not match. Now the sizes are frozen once, here.

But this joint finalize never sets `ProgramImpl::finalized_` (the static `finalize_program_offsets` does not call `set_finalized()`). So if any code later calls `program.impl().finalize_offsets(device)` or `LaunchProgram` on a program taken from `get_programs()` (both are public), that one program is re-laid out standalone, its live sizes change, and nothing notices any more. The dispatch keeps using the frozen snapshot (so the device stays consistent), but `get_sem_base_addr` / `get_cb_base_addr` return the new standalone offsets that no longer match what is on the device.

No in-tree code does this today, so this is a lost guard, not a live bug.

Suggestion: after the joint pass, mark each program finalized (`for (auto* p : program_impls) p->set_finalized();`). That also makes `LaunchProgram` skip the re-layout. Or keep a debug `TT_ASSERT` in `enqueue_mesh_workload` that compares the snapshot with each program's live `get_program_config_sizes()`.

tt_metal/distributed/fd_mesh_command_queue.cpp
```diff
@@ -1594,7 +1626,13 @@ void FDMeshCommandQueue::record_end() {
                 std::pair<bool, int>(mesh_node.unicast_go_signals, num_virtual_eth_cores),
                 static_cast<uint8_t>(this->id()));
 
-            record_program_sub_device_for_range(mesh_device_, range, node.program_runtime_id, sub_device_id);
+            const SubDeviceRecorder trace_sub_device_recorder(mesh_device_, sub_device_id);
```
ayerofieiev-tt, line 1629: **Trace capture is the only time a traced program gets recorded - if the gate is off then, replay is never annotated.**

`record_end` (here) and `create_trace_node` (dispatch.cpp ~3612, through `RecordProgramMetadata`) now skip recording when `IsProgramMetadataRecordingEnabled()` is false. Trace replay never records again. So a trace captured while the gate is off will always show empty kernel sources / no sub-device info in the profiler.

Today this cannot happen: the real-time profiler is only started inside `MeshDevice::create`, before any enqueue, and the env flag is read before `DataCollector` is built. But the gate has no re-sync, so any future "attach later" path (or a runtime flip of `profiler_enabled`) would silently break traced programs.

Suggestion: keep the gate for the per-launch path (that is the hot one), but record unconditionally at trace-capture time. Capture runs once per trace, so the cost is small.

Also, small cleanup: this block builds a new `SubDeviceRecorder` for every trace node and hand-rolls the loop. It could use the same shape as `enqueue_mesh_workload`: `recorder.record(mesh_device_->impl().get_local_devices(range), node.program_runtime_id, sub_device_id);` (same device set, same filter). Then `enabled()` and the single-device `record` overload are not needed. Note: do not hoist one recorder above the node loop, because `sub_device_id` (and so `num_worker_cores`) is per node.

tt_metal/distributed/mesh_device.cpp
```diff
@@ -825,6 +830,14 @@ std::vector<IDevice*> MeshDeviceImpl::get_devices() const {
     return devices;
 }
 
+const std::vector<IDevice*>& MeshDeviceImpl::get_local_devices(const MeshCoordinateRange& range) const {
+    auto [entry, inserted] = local_devices_by_range_.try_emplace(range);
```
ayerofieiev-tt, line 834: **Cache entry is inserted before it is filled - an exception leaves an empty entry behind.**

`try_emplace(range)` inserts an empty vector first, then `view_->get_devices(range)` fills it. If that call throws (it does `TT_FATAL` for a coordinate outside the mesh), the empty vector stays in the map. The next `get_local_devices(range)` for the same range returns the empty vector: no program commands are written for that range, and no error is raised.

Today this needs two kernel-less workloads on a bad range (workloads with kernels fail earlier in `load_binaries`), so it is low risk. But the fix is two lines - compute first, then insert:
```cpp
if (auto it = local_devices_by_range_.find(range); it != local_devices_by_range_.end()) {
    return it->second;
}
auto devices = view_->get_devices(range);
return local_devices_by_range_.emplace(range, std::move(devices)).first->second;
```
The same insert-then-fill pattern is in `MeshWorkloadImpl::determine_sub_device_ids` (mesh_workload.cpp:343); a throw from a later program leaves a half-filled set cached there too.

tt_metal/impl/program/dispatch.cpp
```diff
@@ -2703,6 +2706,37 @@ void assemble_device_commands(
...
+    const uint32_t tensix_index = hal.get_programmable_core_type_index(HalProgrammableCoreType::TENSIX);
+    const uint32_t remote_offset_index = program.get_program_config(tensix_index).local_cb_size / sizeof(uint32_t);
+    for (size_t range_index = 0; range_index < program_command_sequence.circular_buffers_on_core_ranges.size();
+         ++range_index) {
+        uint32_t* const payload = program_command_sequence.cb_configs_payloads[range_index];
+        for (const auto& circular_buffer : program_command_sequence.circular_buffers_on_core_ranges[range_index]) {
+            for (const uint32_t buffer_index : circular_buffer->local_buffer_indices()) {
+                local_cb_updates.push_back(
+                    {circular_buffer.get(),
+                     payload + UINT32_WORDS_PER_LOCAL_CIRCULAR_BUFFER_CONFIG * buffer_index,
```
ayerofieiev-tt, line 2727: **The CB slot math now lives in three places.**

`UINT32_WORDS_PER_LOCAL_CIRCULAR_BUFFER_CONFIG * buffer_index` and `remote_offset_index + (max_cbs - 1 - buffer_index) * UINT32_WORDS_PER_REMOTE_CIRCULAR_BUFFER_CONFIG` appear here (2727 / 2734), in the assemble-time fill (1576 / 1586), and in `create_trace_node` (3556 / 3564). If one copy changes, eager and trace replay write different slots.

Suggestion: two tiny helpers, e.g. `local_cb_config_slot(buffer_index)` and `remote_cb_config_slot(remote_offset_index, max_cbs, buffer_index)`, used at all three sites. (`create_trace_node` works on the trace-cached sequence, whose `local/remote_cb_config_updates` are already filled, so it could also reuse the descriptors directly.)

tests/tt_metal/distributed/test_mesh_workload.cpp
```diff
@@ -928,7 +961,21 @@ TEST_F(MeshWorkloadTestSuite, MeshWorkloadCBUpdate) {
...
+        *program,
+        "tests/tt_metal/tt_metal/test_kernels/dataflow/report_cb_config.cpp",
+        cr_set,
+        DataMovementConfig{
+            .processor = DataMovementProcessor::RISCV_1,
+            .noc = NOC::RISCV_1_default,
+            .compile_args = {updated_cb_index}});
+    uint32_t max_cb_region_size = 0;
+    for (const auto& cb_config : cb_config_vector) {
+        max_cb_region_size += 2 * cb_config.num_pages * cb_config.page_size;
```
ayerofieiev-tt, line 974: **The report address has zero slack and the `2 *` is a hidden link to a later step.**

With the four CBs here, `2 * sum` = 45056 bytes. After the `num_pages *= 2` loop below (line ~992) the CB region is exactly 45056 bytes. So `report_addr` sits exactly at the end of the CB region, with no gap. The `2` only works because that later loop doubles every CB once.

If someone adds a CB, changes the growth factor, or something reserves L1 on these cores before the CBs, the report will land inside a live CB and the test will pass or fail for the wrong reason.

Suggestion: compute `report_addr` from the biggest *updated* footprint plus some slack, or at least add a comment that says where the `2` comes from.

tt_metal/distributed/fd_mesh_command_queue.cpp
```diff
@@ -511,11 +533,23 @@ void FDMeshCommandQueue::enqueue_mesh_workload(MeshWorkload& mesh_workload, bool
...
     }
+
+    const uint32_t mcast_launch_msg_wptr =
+        cq_shared_state_->worker_launch_message_buffer_state[*sub_device_id].get_mcast_wptr();
+    const uint32_t unicast_launch_msg_wptr =
+        cq_shared_state_->worker_launch_message_buffer_state[*sub_device_id].get_unicast_wptr();
+    const CoreCoord dispatch_core = this->virtual_program_dispatch_core();
+    const SubDeviceRecorder sub_device_recorder(mesh_device_, sub_device_id);
+#if defined(TRACY_ENABLE)
+    const bool tag_tracy_zones = !tt::tt_metal::getDeviceProfilerState();
```
ayerofieiev-tt, line 544: **This reads the profiler state from the default context, but the rest of this function uses the mesh's own context.**

`getDeviceProfilerState()` takes a `ContextId` (default = `DEFAULT_CONTEXT_ID`). The recorder in this function uses `extract_context_id(mesh_device_)`, and `finish_nolock` was changed in this PR to `IsProgramRealtimeProfilerActive(extract_context_id(mesh_device_))`. This line was also changed by the PR but still uses the default context.

Small fix:
```suggestion
    const bool tag_tracy_zones = !tt::tt_metal::getDeviceProfilerState(extract_context_id(mesh_device_));
```
Low impact (only the Tracy tag), but it keeps one rule in one function.

tt_metal/impl/program/dispatch.cpp
```diff
@@ -1360,6 +1358,11 @@ BatchedTransfers assemble_runtime_args_commands(
         "incorrect.",
         command_count,
         program_command_sequence.runtime_args_command_sequences.size());
+    program_command_sequence.runtime_args_sizeB = std::accumulate(
```
ayerofieiev-tt, line 1361: **Small: the trace log later in `assemble_device_commands` still re-adds the RTA sizes by hand.**

Now that `runtime_args_sizeB` is stored here, the "Command Sequence Summary" log (~line 2832-2836) can read `program_command_sequence.runtime_args_sizeB` instead of looping over `runtime_args_command_sequences` again. Nothing changes the list between the two points. One definition of "RTA total bytes" is better than two.

tt_metal/distributed/mesh_workload.cpp
```diff
@@ -86,6 +86,41 @@ MeshWorkloadImpl::MeshWorkloadImpl() : id(get_next_counter()) {
...
+        metadata.num_program_devices += device_range.shape().mesh_size();
+        metadata.runs_on_noc_multicast_only_cores |= program_impl.runs_on_noc_multicast_only_cores();
+        metadata.runs_on_noc_unicast_only_cores |= program_impl.runs_on_noc_unicast_only_cores();
+
+        const auto& program_config_sizes = program_impl.get_program_config_sizes();
+        if (metadata.program_config_sizes.empty()) {
+            metadata.program_config_sizes = program_config_sizes;
+        } else {
+            TT_FATAL(
+                metadata.program_config_sizes.size() == program_config_sizes.size(),
```
ayerofieiev-tt, line 108: **Small: this is `std::vector::operator==` written by hand.**

The size check plus the element loop, with the same message twice, is the same as:
```cpp
TT_FATAL(
    metadata.program_config_sizes == program_config_sizes,
    "Expected config sizes to be identical across all programs in a MeshWorkload.");
```
Same behavior (size first, then elements), less code, one message.

tt_metal/impl/program/program_command_sequence.hpp
```diff
@@ -10,7 +10,6 @@
 #include "llrt/hal.hpp"
 #include "tt_metal/impl/dispatch/device_command.hpp"
 #include "tt_metal/impl/dataflow_buffer/dataflow_buffer_impl.hpp"
```
ayerofieiev-tt, line 12: **Small: one more file lost its `<numeric>` when this include was removed.**

`tests/tt_metal/tt_metal/perf_microbenchmark/dispatch/test_prefetcher.cpp` uses `std::accumulate` (line ~1060), includes `program_impl.hpp` (which pulls in this header), and has no `#include <numeric>` of its own. This PR added `<numeric>` to two sibling files for the same reason but missed this one.

It still builds today because `<random>` (via `common.h`) and `nlohmann/json.hpp` happen to include `<numeric>` on both libstdc++ and libc++, but that is luck, not a guarantee. Please add `#include <numeric>` to `test_prefetcher.cpp`.

tt_metal/distributed/mesh_device.cpp
```diff
@@ -875,10 +888,33 @@ DeviceIds MeshDeviceImpl::get_device_ids() const {
...
+    // single-threaded. Answer without caching so this accessor never writes.
     return validate_and_get_reference_value(
         this->get_devices(), [](const auto* device) { return device->compute_with_storage_grid_size(); });
 }
 
+// The cross-device agreement check behind these two properties rebuilds the device list and walks
+// the whole mesh, and circular buffer validation asks for both once per program on every enqueue.
+// They are fixed once the devices are open, so resolve them here: the accessors then answer from a
+// value nobody writes, which is what makes them safe to call without holding the api lock.
+void MeshDeviceImpl::establish_device_property_caches() {
```
ayerofieiev-tt, line 904: **Design note (non-blocking): this fixes two accessors, but ~20 more use the same slow pattern.**

`compute_with_storage_grid_size` and `l1_size_per_core` are cached here because they showed up in the profile. About 20 other `MeshDeviceImpl` accessors still call `validate_and_get_reference_value(this->get_devices(), ...)` on every call (a fresh vector plus a walk over all devices). One of them, `worker_core_from_logical_core` (line ~1377), is re-run per core on program-cache hits in a few ttnn ops (e.g. `nlp_create_qkv_heads` sharded `override_runtime_arguments`, `roll`).

Each new hot accessor will need another `std::optional`, another line here, and another reset in `reshape`. A more general fix would be: check that all devices agree once (at init and after reshape), then have every accessor read the reference device directly. That is bigger than this PR, so this is just a note for a follow-up.

PR #55001, opened 2026-08-31
Quasar Pool: Use All 4 NEOs Per Cluster (num_threads=4)

wransom-TT:
## Problem description

The Quasar max_pool2d/avg_pool2d port ran on one NEO per cluster: two split readers on two DM cores feeding a single compute kernel at num_threads=1, leaving 3 of 4 Tensix engines idle. No ttnn op used Metal 2.0 num_threads yet, so this is also the pathfinder for SPMD multi-threading through DFBs.

Two latent correctness bugs surfaced on the way: the max-pool input-CB pre-clear wrote 0 instead of -inf (only valid for post-ReLU inputs; all-negative input returned 0.0), and the compute kernel did not compile for WH/BH (a header only reachable through the Quasar include chain).

Two further Quasar bugs were found but are pre-existing (identical on main at the merge base) and are not addressed here: windows of 8 or fewer rows return wrong values, and TILE output layout returns wrong values.

## What's changed

- One reader KernelSpec at num_threads=4 (SPMD over 4 DM cores) replaces the split readers, paired with num_threads=4 compute. With equal producer and consumer thread counts the STRIDED bindings become four private (DM, NEO) lanes, each owning whole output sticks, so wide-C c-blocks and large-kernel chunks stay in-lane and run unchanged.
- The lane count is always 4 on Quasar (MPWI and TILE output stay at 1). The reader deals sticks round-robin (stick i to lane i mod 4) and each compute lane derives its own share (quotient, plus one for the first sticks-mod-4 lanes), so any per-core stick count works: a remainder leaves the tail lanes one stick short, and fewer sticks than lanes leaves them idle (the batch-1 global pool is 1 stick per core). No divisibility requirement.
- Every DFB bound to a 4-thread kernel is lane-proof: the borrowed input/output shard views are carved into 4·k entries (k keeps the TRISC-side stride inside its uint16 limit for shards over ~1 MB, e.g. the resnet stem), the row-major out_cb and the avg scalar-config table use the same lane-agnostic view, and the per-stick scalar lookup indexes by global stick instead of a per-lane counter.
- Input-CB init writes the pool identity (-inf for max, 0 for avg) over the whole ring via a CPU fill plus cache flush; each thread recovers the ring base and fills it all, since per-thread contiguous clears miscover a strided ring. Fixes the negative-input max bug. The large-kernel partial-chunk clear is restored from the non-Quasar reader (avg only).
- Shared pool_utils: num_threads_per_cluster, single_reader_stream caller flag, CB sizing scaled by thread count.
- Tests: a Quasar-gated harness under tests/ttnn/unit_tests/operations/pool/quasar (kernel ladder 2x2 to 9x9 incl. chunked large kernels, batch/tall/wide geometry, wide reduction at C=280, avg incl. per-stick scalars, all-negative inputs, block/width sharding, odd stick counts and idle lanes), plus perf-ab and span-ab threads A/B modes. Verified exact on WH silicon (one lane) and on the ZeBu emulator at four lanes; the resnet50 Quasar manifest passes 13/13 and resnet50 e2e matches main's PCC (0.999874) while running faster. craq-sim hangs on uneven lane splits are simulator artifacts (exact on RTL) and are tagged with that evidence.

## Perf (craq-sim clocks, threads=1 → threads=4; relative comparison only — real numbers need the emulator)

| case | threads=1 | threads=4 | delta% | speedup |
|------|----------:|----------:|-------:|--------:|
| avg_k7x7_s1_large | 223,493 | 61,203 | -72.6% | 3.65x |
| k9x9_s2_3chunks | 577,209 | 176,789 | -69.4% | 3.26x |
| k8x8_s2_large | 449,398 | 144,400 | -67.9% | 3.11x |
| avg_k3x3_s1 | 410,842 | 134,565 | -67.2% | 3.05x |
| k7x7_s2_large | 357,467 | 121,375 | -66.0% | 2.95x |
| k3x3_s1 | 164,407 | 71,869 | -56.3% | 2.29x |
| wide_c280_3blocks | 183,520 | 89,078 | -51.5% | 2.06x |
| block_2x2 | 49,015 | 29,923 | -39.0% | 1.64x |
| k5x5_s2 | 63,934 | 46,695 | -27.0% | 1.37x |
| batch2 / tall_32x4 / wide_4x32 / width_1x2_c128 | ~51,000 | ~43,200 | -15.2% | 1.18x |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:wransom/quasar_pool_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:wransom/quasar_pool_opt)


tests/ttnn/nightly/unit_tests/operations/experimental/quasar/pool_quasar_test_utils.py
```diff
@@ -0,0 +1,55 @@
...
+def _dump_mismatches(got, golden, out_h, out_w, channels, n_dump):
+    """got/golden: (sticks, C) float tensors. Prints the n_dump worst sticks."""
+    diff = (got - golden).abs()
+    per_stick = diff.max(dim=1).values
+    bad = int((per_stick > 0).sum().item())
+    n = min(n_dump, bad)
+    if n == 0:
+        return
+    worst = torch.topk(per_stick, n).indices
+    print(f"\nQPOOL: {bad}/{got.shape[0]} sticks mismatch; worst {n}:")
```
bbradelTT: Is there a reason to prefer print over logger? If not, please use logger.

bbradelTT (commented):

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/dataflow/reader_pool_2d.cpp
```diff
@@ -298,90 +303,69 @@ void kernel_main() {
...
     if constexpr (is_avg_pool || need_to_initialize_in_cb || force_max_clear) {
-        if constexpr (reader_id == 0) {
-            fill_with_val(clear_value_cb.get_write_ptr(), TILE_HEIGHT * TILE_WIDTH, bf16_init_value);
-            clear_value_cb.push_back(1);
-        }
-        if constexpr (reader_id == 1) {
-            clear_value_cb.wait_front(1);
-        }
-        // for average pool clear out tiles runs in loop, no need to initialize here
+        fill_with_val(clear_value_cb.get_write_ptr(), TILE_HEIGHT * TILE_WIDTH, bf16_init_value);
```
dstoiljkovicTT, line 309: Are we missing some cb push after this fill of the clear cb value?

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/dataflow/reader_pool_2d.cpp
```diff
@@ -298,90 +303,69 @@ void kernel_main() {
...
+            // (the Metal 2.0 TensorBinding tensor::reader_indices replaces a CTA-baked DRAM address).
             Noc cfg_noc;
             const auto reader_indices_accessor = TensorAccessor(tensor::reader_indices);
             cfg_noc.async_read(
                 reader_indices_accessor, reader_indices_cb, reader_page_size, {.page_id = core_nhw_index}, {});
             cfg_noc.async_read_barrier();
-            reader_indices_cb.push_back(1);
-        } else {
-            reader_indices_cb.wait_front(1);
+            // No push: census-only (nothing waits); read back at this thread's write cursor.
```
dstoiljkovicTT, line 360: Not sure I understand how the two threads are synchronized if we are lacking wait and push for the reader_indices

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
```diff
@@ -1232,106 +1290,160 @@ ttnn::device_operation::ProgramArtifacts pool2d_create_program_artifacts(
...
-            .endpoint_type = DFBEndpointType::CONSUMER});
+            .endpoint_type = DFBEndpointType::CONSUMER,
+            // Should be ALL — same workaround as in_scalar_cb_0 above (issue #54505).
+            .access_pattern = DFBAccessPattern::STRIDED});
     }
     if (!return_indices) {
+        // Census partners for the reader's raw-view DFBs: Gen2 forbids DM self-loops, so the
+        // single multi-threaded reader takes only the producer face and compute carries these
+        // never-constructed consumer bindings (no credit traffic ever flows on them).
+        if (!params.split_reader) {
```
dstoiljkovicTT, line 1324: So we excluded the if constexprs regarding different readers but are still supporting spli_reader feature, not sure how this correlates with the rest of the change

tests/ttnn/nightly/unit_tests/operations/experimental/quasar/pool_quasar_test_utils.py
wransom-TT: No, I will use the logger!

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
wransom-TT: Split reader isn't being supported on Quasar at least for now. We only have 6 user controllable DM cores for 4 compute NEOs, so the split reader as it existed (2 readers per compute) won't work. Currently I have it so each NEO has an assigned reader. 2 readers are therefore still idle, but getting them involved with strided DFBs isn't so easy so I'm going. todo that as a follow on. The improvements in this PR are already a 2-4x improvement over main!

bbradelTT: Runtime said that 2 DMs per input is sufficient to saturate the NoC. There shouldn't be any need to do more.

bbradelTT (approved):

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/dataflow/reader_pool_2d.cpp
wransom-TT: No, this is intentional, nothing ever waits on this DFB, the compute-side binding exists only so the program spec has a consumer and is never constructed in the kernel. Each reader fills its own compute/NEO lane's clear tile page.

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/kernels/dataflow/reader_pool_2d.cpp
wransom-TT: The individual readers maintain their own copies now, this costs some L1 but keeps the 4 reader + NEO pairs entirely separate with no cross-thread synchronization. If we want to recover the L1 in the future, we can use the ALL pattern in the DFB instead of STRIDED.

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
wransom-TT: Ok currently I'm using 4 so there is a 1:1 relationship with the NEOs which makes dealing with the STRIDED access pattern easier. Is there a downside to using the extra cores  to simplify the kernel if they are available?

dstoiljkovicTT: > Split reader isn't being supported on Quasar at least for now. We only have 6 user controllable DM cores for 4 compute NEOs, so the split reader as it existed (2 readers per compute) won't work. Currently I have it so each NEO has an assigned reader. 2 readers are therefore still idle, but getting them involved with strided DFBs isn't so easy so I'm going. todo that as a follow on. The improvements in this PR are already a 2-4x improvement over main!

Yes, my point was more why do we check the split reader, shouldn't that always be set to false or asserted if set to true in this path since we are not supporting it on quasar implementation of the pools

dstoiljkovicTT (approved):

ttnn/cpp/ttnn/operations/experimental/quasar/pool_generic/device/pool_multi_core_program_factory.cpp
wransom-TT: Ohhh I see yes this should be removed since it's a quasar specific factory! Will do.

PR #55012, opened 2026-09-01
tt-train: one-step async GRPO trainer + phase-split refactor + MeshSocketWeightBridge

ichovpanTT:
### Summary

Stacked on #54390 (gsm8k example). Three related additions on the `tt-train` async-RL path:

* **`OneStepAsyncGRPOTrainer`** — verl-shape one-step off-policy GRPO trainer that overlaps generation of batch `t+1` on the remote rollout worker with the optimizer step on batch `t`. Staleness is exactly 1: rollouts `R_t` were generated with `theta_{t-1}` at the moment we consume them against actor `theta_t` to produce `theta_{t+1}`. Live example wiring at `tt-train/sources/examples/grpo_remote_rollout/gsm8k_onestep/`.
* **`GRPOTrainer` phase-split refactor** — `train()` now walks one method per phase (`_iter_prompt_batches` → `_rollout` / `_await_rollout` → `_expand_prompts_and_columns` → `_compute_rewards` → `_compute_advantages` → `_optimize` → `_apply_gradients` → `_publish_step_metrics` → `_maybe_checkpoint` → `_reset_step_metrics`). Every phase helper takes 1-3 raw args and writes metrics directly into `self.metrics`, replacing the previous 5-8 argument signatures. Killed side-channels: `self._reward_component_means`, `self._callback_times`, `warmup_factor` plumbing. `OneStepAsyncGRPOTrainer` now overrides only `train()` for the prime/await/resubmit loop shape; every phase helper is inherited.
* **`MeshSocketWeightBridge`** — fabric-based alternative to `HostWeightBridge` for the ttml -> ttt weight-sync path. On Qwen3-0.6B: ~0.85 s per push vs ~3.9 s over MPI (~4.6x speedup); see [`WEIGHT_SYNC_OPTIMIZATIONS.md`](tt-train/sources/examples/grpo_remote_rollout/gsm8k/WEIGHT_SYNC_OPTIMIZATIONS.md) for the design + further optimization stack. Selectable via `weight_bridge: {host | mesh_socket}` in the gsm8k example config.

### Notes for reviewers

* Base is `ichovpan/gsm8k-example-async-rl` (open as #54390). Once that lands this can rebase to main.
* Test coverage: device-less loop-shape tests for `OneStepAsyncGRPOTrainer` (prime/await/resubmit interleaving, `num_iterations == 1` guard, completer-contract preflight, `_publish_step_metrics` splat regression) at `tt-train/tests/python/test_one_step_async_grpo_trainer.py`. `MeshSocketWeightBridge` 1->1 transport test at `tt-train/tests/python/grpo_remote_rollout/weight_transfer/`.
* One `fabric` change carried along: drop `/2` on the `num_assigned` intermesh check to fix the 2-mesh single-cable case that the MeshSocketWeightBridge topology triggers.

Made with Cursor

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:ichovpan/tt-train-1-step-async-rl)

### DIFF
--- a/tt-train/sources/examples/grpo_remote_rollout/gsm8k/WEIGHT_SYNC_OPTIMIZATIONS.md
+++ b/tt-train/sources/examples/grpo_remote_rollout/gsm8k/WEIGHT_SYNC_OPTIMIZATIONS.md
@@ -1,0 +1,260 @@
+# Cutting `MeshSocketWeightBridge` sync time toward 0.5 s
+
+Context: on Qwen3-0.6B, `WeightSyncCallback_time_s` currently averages ~0.85 s
+per push with `MeshSocketWeightBridge` (down from ~3.9 s with `HostWeightBridge`
+over MPI). Target: **≤ 0.5 s per push**.
+
+Getting from 0.85 s → ≤ 0.5 s means cutting ~350 ms. That's realistically 3-4
+stacked optimizations, not one silver bullet. Ranked list with rough savings
+estimates and concrete change shape for each.
+
+## Where the ~0.85 s is going (estimate)
+
+Rough per-push breakdown, both ranks:
+
+- `qwen3_weights_ref_hf_dict` on sender: ~150-250 ms (56 `ttnn.slice` device
+  ops + ~200 Python param lookups).
+- `_send_manifest` (JSON over MPI): ~5-20 ms.
+- Loop of ~200 `send_async` calls: ~100-150 ms of Python + pybind dispatch.
+- Sender `synchronize_device` (wire drain): ~150-300 ms.
+- Receiver `_recv_manifest`: ~5-20 ms.
+- Receiver 200 × `allocate_tensor_on_device`: ~50-100 ms.
+- Receiver 200 × `recv_async` dispatch: ~100-150 ms.
+- Receiver `synchronize_device` (mostly overlaps with sender's drain).
+- Receiver `worker.update_weights` (the `copy_to_buffer` convert cascade):
+  ~200-400 ms.
+- `bridge.barrier()`: small (< 5 ms).
+
+Total ~ 800-1400 ms brackets the observed 850 ms. The two big fat bars are
+`worker.update_weights` and the sender's dict-build + slice ops.
+
+## The high-leverage stack (do these, in order)
+
+### 1. Skip `worker.update_weights` — write directly into the model's weight buffers
+
+**Expected saving: 200-400 ms.** The single biggest win, and probably the
+difference between 0.85 s and ~0.5 s all on its own.
+
+Currently the flow is:
+
+- `recv_async(scratch_tensor, socket)` → scratch tensor allocated per key
+  per push.
+- `worker.update_weights` walks `hf_dict` and calls
+  `copy_to_buffer(scratch, self.wqkv, ...)` per param, which fires up to 4
+  device ops per tensor (`to_layout` / `typecast` / `reshape` /
+  `to_memory_config` — the last one is the `interleaved_to_sharded` that
+  triggered the L1 clash we saw during the FIFO bump; see
+  [`models/common/utility_functions.py`](../../../../../models/common/utility_functions.py)
+  around line 1257).
+
+For Qwen3-0.6B that's ~200 tensors × several ms of program dispatch each.
+Reproducibly hundreds of ms per push.
+
+Shape of the fix:
+
+- At `connect()` time (or first `send_weights`), build a **destination map**
+  `hf_key → live_ttnn.Tensor` from `worker.models[0]` on the receiver. That
+  map is the *actual* model weight handles: `self.wqkv`, `self.wo`, per-layer
+  norms, etc.
+- Change `receive_weights` at [`mesh_socket_bridge.py:210-241`](../utils/mesh_socket_bridge.py)
+  to `recv_async(dest_map[key], self._socket)` — the socket writes straight
+  into the model's buffer.
+- On the sender, `qwen3_weights_ref_hf_dict` has to produce tensors in
+  **exactly** the layout/dtype/shard that `dest_map[key]` expects (e.g.
+  sharded L1 for `wqkv`, DRAM interleaved for norms). One-time conversion on
+  the sender caches the shape; subsequent pushes reuse the shaped output
+  buffers.
+- `on_weights_received` shrinks to a no-op (or bookkeeping only). No
+  `copy_to_buffer` cascade.
+
+Caveats:
+
+- Sharding metadata has to survive the socket. Today the receiver
+  `allocate_tensor_on_device(spec, target)` allocates a plain DRAM interleaved
+  tensor; you'd need the destination buffer's memory config, not a fresh one.
+  The socket API writes bytes into whatever buffer you hand it, so this should
+  just work as long as the sender's byte layout matches the destination's byte
+  layout.
+- Sender has to be reshuffled to produce tensors in destination layout. For
+  most params that's TILE bf16 DRAM (matches today). For `wqkv`, the
+  destination is sharded L1 — the sender has to shard-out before sending.
+  That shard-out is a one-time program compile + cached; per-push cost is a
+  single `to_memory_config(cached_wqkv_send_buffer)` call.
+
+This is the single change with the largest expected impact. Worth doing
+before anything else.
+
+### 2. Cache manifest + receive-side allocations across pushes
+
+**Expected saving: 80-150 ms.** Trivial change; do it in the same commit as
+#1.
+
+Current per-push cost on the receiver at
+[`mesh_socket_bridge.py:210-231`](../utils/mesh_socket_bridge.py):
+
+- `_recv_manifest` — one MPI round-trip for a JSON payload. Skip after first.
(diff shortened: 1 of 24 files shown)


tt-train/sources/examples/grpo_remote_rollout/gsm8k/gsm8k_training_example.py
```diff
@@ -226,7 +253,8 @@ def _ttml_main() -> None:
     completer: Any = None
     client: Any = None
     try:
-        bridge = HostWeightBridge.init_sender(mesh=mesh_device, peer_rank=TTT_RANK)
+        weight_bridge_kind = _resolve_weight_bridge_kind(raw)
```
jmalone-tt, line 256: can simplify - we don't need double validation here or for the receiver bridge creation

tt-train/sources/examples/grpo_remote_rollout/gsm8k/plot_rewards.py
```diff
@@ -5,11 +5,10 @@
```
jmalone-tt, line 5: can you add a comment to the PR description that shows how plots are changed (old vs new)

tt-train/sources/examples/grpo_remote_rollout/gsm8k/WEIGHT_SYNC_OPTIMIZATIONS.md
```diff
@@ -0,0 +1,260 @@
+# Cutting `MeshSocketWeightBridge` sync time toward 0.5 s
```
jmalone-tt, line 1: don't need to commit this file

tt-train/sources/examples/grpo_remote_rollout/gsm8k_onestep/gsm8k_onestep_training_example.py
```diff
@@ -0,0 +1,384 @@
...
+        str(REPO_ROOT),
+        "generated/tt-train/grpo_gsm8k_onestep_run",
+        datetime.now(timezone.utc).strftime("%Y%m%d_%H%M%S"),
+    )
+
+
+_VALID_WEIGHT_BRIDGES = ("host", "mesh_socket")
+
+
+def _resolve_weight_bridge_kind(raw: dict) -> str:
```
jmalone-tt, line 192: This doesn't seem like it belongs in the example (should be an implementation detail the user doesn't need to worry about other than selecting which to use)

tt-train/sources/examples/grpo_remote_rollout/gsm8k_onestep/gsm8k_onestep_training_example.py
```diff
@@ -0,0 +1,384 @@
...
+            inference_client=client,
+            enable_ddp=device_config.enable_ddp,
+        )
+
+        # Async trainer owns all weight pushes -- no eager boot push, no
+        # WeightSyncCallback. The prime call inside train() pushes theta_0
+        # before submitting gen_0.
+        client.connect()
+
+        trainer = OneStepAsyncGRPOTrainer(
```
jmalone-tt, line 283: Maybe this is the right place to pass the desired weight bridge and GRPOTrainer can set everything up, with a default value that will be used if unspecified.

jmalone-tt: I think having separate classes here also isn't the best design. Why have separate OneStepAsyncGRPOTrainer and GRPOTrainer instead of a unified GRPOTrainer that can handle both?

tt-train/tests/python/grpo_remote_rollout/weight_bridge/configurations/1_1/mgd.textproto
```diff
@@ -0,0 +1,44 @@
+# Mesh Graph Descriptor for the MeshSocketWeightBridge 1->1 transport test.
```
jmalone-tt, line 1: Fine to keep, but the weight bridge changes and test probably could have been a separate PR so reviewing would have been easier

tt-train/tests/python/grpo_remote_rollout/weight_transfer/test_weight_transfer.py
```diff
@@ -61,8 +61,12 @@
 NUM_SUBMESHES = 4
 
 # Post-push RPC batch; TTT_MAX_BATCH_SIZE must be >= this.
-POST_PUSH_BATCH = 8
-TTT_MAX_BATCH_SIZE = 8
+# Bumped from 8 -> 16: TttGenerationWorker on Blackhole exhibits a determinism
+# regression at max_batch_size=8 (some slots diverge from the greedy output);
+# running with max_batch_size=16 works around it until the underlying worker
+# bug is fixed.
```
jmalone-tt, line 67: do we have a tracking ticket for this issue?

tt-train/sources/ttml/ttml/trainers/grpo_trainer.py
```diff
@@ -1105,194 +1086,353 @@ def train(self) -> None:
...
+        self,
+        prompts_x: List[List[int]],
+        completions: List[List[int]],
+        advantages_np: np.ndarray,
+    ) -> None:
+        """Reference log-probs (once) + per-micro-batch forward-with-grad + GRPO
+        loss + backward. Accumulates gradients across the whole generation
+        batch; does NOT call ``optimizer.step()`` — that is ``_apply_gradients``.
+        """
+        ref_logprobs = self._compute_ref_logprobs(prompts_x, completions)
```
jmalone-tt, line 1216: I think _optimize is called once per mini-epoch, so the ref_logprobs will change with each mini-epoch, which doesn't match the policy they were computed with. This will be an issue when we use a config with num_iterations > 1. Takeaway here is that ref_logprobs should be calculated once per generation and then passed into _optimize

tt-train/sources/ttml/ttml/trainers/grpo_trainer.py
```diff
@@ -1105,194 +1086,353 @@ def train(self) -> None:
...
+            model_source=self.model_source,
+        )
+        ckpt_dir = os.path.join(cfg.output_dir, "checkpoints", f"grpo_step_{step}")
+        monitor_cb = next((cb for cb in self.callbacks if isinstance(cb, GRPOMonitor)), None)
+        for cb in self.callbacks:
+            if cb is monitor_cb:
+                continue
+            self._time_callback(cb, "on_save", self, step, ckpt_dir)
+
+    def _reset_step_metrics(self) -> None:
```
jmalone-tt, line 1397: Looks like this is called inside mini-epoch loop which will wipe batch-level reward metrics when num_iterations > 1

tt_metal/fabric/control_plane.cpp
```diff
@@ -2769,9 +2769,14 @@ void ControlPlane::generate_intermesh_connectivity() {
...
-    // Divide by 2 here, since the intermesh_connections data structure stores connections
-    // bidirectionally.
-    auto num_assigned_intermesh_connections = intermesh_connections.size() / 2;
+
+    // intermesh_connections stores each cable bidirectionally (one directed slot per direction).
+    // We compare directed-slot counts on both sides of the check below: requested_intermesh_connections
+    // is likewise populated per direction by the MGD parser, so the raw .size()s are in the same unit.
+    // Do NOT divide by 2 here -- that historical /2 (introduced in PR #34397) put the two sides in
+    // mismatched units and made any 2-mesh, single-cable topology unsatisfiable even though one
+    // physical cable is topologically sufficient for a mesh pair.
```
jmalone-tt, line 2778: I don't think this change is right - can you double check if things resolve without it?
From claude:
requested_intermesh_connections_ is one-directional, and the getter returns the outer map size - source meshes, not directed channels. Both producers of intermesh_connections emit two entries per cable, so the /2 was correct.

tt-train/tests/python/grpo_remote_rollout/weight_transfer/test_weight_transfer.py
ichovpanTT: Just submitted, https://github.com/tenstorrent/tt-metal/issues/55425

ichovpanTT:
Having fully async trainer is the priority for now, closing the PR

PR #55031, opened 2026-09-01
Fix sharded concat treating rank-3 width as height

### LINKED ISSUE #55032: concat: sharded s2s uses absolute dim 2/3 instead of rank-relative H/W, silent wrong data on rank ≠ 4
Item 4 from master ticket

## Problem

`ConcatS2SMultiProgramFactory` decides height vs width concat with the **absolute** index `2 == dim`. Height is supposed to be `dim == rank - 2`.

`operation_attributes.dim` is already **normalized**. For a rank-3 tensor, width concat (`dim=-1`) is **dim 2**, which is also `rank - 1`. The factory treats that as height concat: each input shard is stacked **below** the previous one instead of **beside** it.

`select_program_factory` has the same literals (`3 == args.dim` for the 2-tensor width factories, `2 == args.dim || 3 == args.dim` for s2s multi). Rank-3 width concat never hits the RM/tiled width factories and is forced into the multi factory with the wrong layout. Rank-5 sharded width (`dim == 4`) misses those paths entirely.

This is silent wrong data (in-bounds). With larger shard heights the bad write offsets can also go **OOB in L1**.

The sibling `ConcatBlockShardedProgramFactory` and the sharded validator already use `rank - 1` / `rank - 2`. `build_unsqueeze_concat()` (would lift rank &lt; 4 to 4D) is defined in `concat.cpp` but **not** used in the massaged concat chain.

## Example

Rank-3 ROW_MAJOR HEIGHT_SHARDED, one core, `ttnn.concat([a, b], dim=-1)`:

| | Shape / result |
|--|----------------|
| Inputs | `(1, 4, 16)` + `(1, 4, 16)`, shard `(4, 16)` |
| Output shard | `(4, 32)` |
| PyTorch | columns 0–15 from `a`, 16–31 from `b` |
| Current ttnn | `a` in rows 0–1, `b` in rows 2–3, columns 16–31 **zero** |

Same case at rank 4, shape `(1, 1, 4, 16)`, `dim=3`: **correct**.

Larger height (e.g. 64 sticks, two equal inputs): second input starts at half the output shard; last stick lands ~1.5× the shard size → OOB L1 write.

## Location

- `ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_s2s_multi_program_factory.cpp` — `is_height_concat = 2 == operation_attributes.dim`
- `ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_device_operation.cpp` — `select_program_factory` (`3 == args.dim`, `2 == args.dim \|\| 3 == args.dim`)

## Suggested fix

Mirror the block-sharded factory:

```cpp
const uint32_t rank = input_tensors[0].logical_shape().rank();
const bool is_height_concat = rank >= 2 && operation_attributes.dim == rank - 2;
```

In `select_program_factory`, replace the dim literals with `args.dim == rank - 1` (width) and `args.dim == rank - 2` (height).

## Test plan

Add cases in `tests/ttnn/unit_tests/operations/data_movement/test_concat.py`:

- Rank-3 HEIGHT_SHARDED width concat (`dim=-1`), 2 and 3 tensors (3 tensors hits the multi factory; 2-tensor RM width uses the RM factory after dispatch is fixed).
- Rank-3 WIDTH_SHARDED height concat (`dim=-2`).
- Rank-2 HEIGHT_SHARDED width concat (used to miss dim 2/3 dispatch).
- Compare shape and values against `torch.concat`.

VirdhatchaniKN:
Fixes #55032 — item 4 of #51214.

**Problem** — `dim` is already normalized when it reaches the device op, but the sharded dispatch and `ConcatS2SMultiProgramFactory` picked height vs width with the literals `2` and `3`. Those only name the right axis at rank 4, so other ranks silently returned wrong data:

- rank 3, `dim=-1` (== 2) — width concat laid out as height: shards stacked below instead of beside, and with taller shards the writes ran off the end of the output shard in L1
- rank 5, `dim=-2` (== 3) — height concat routed into the width-only RM/tiled factories
- rank 5, `dim=-1` (== 4) and rank 2, `dim=-1`/`-2` — matched no branch, fell through to the interleaved factory

**Fix**

- Use rank-relative tests (`dim == rank-1` width, `dim == rank-2` height) in `select_program_factory` and `ConcatS2SMultiProgramFactory`, matching `ConcatBlockShardedProgramFactory` and the validator
- `ConcatS2SRMProgramFactory` (which rank-3 width concat now reaches) took its row count from `padded_shape()[-2]`; a height-sharded tensor tiles `(prod(dims[0..-2]), dims[-1])`, so any leading dim > 1 under-counted and the ragged last core copied too few rows. Use the flattened row count. Was equally broken at rank 4

**Verification**

- Built before and after on Wormhole and diffed 18 configs: every case above fails before, passes after, nothing regresses
- New tests cover rank 2/3/5, width and height, 2 and 3 inputs, RM and tile, bfloat16 and a 4-byte dtype, the issue's multi-core out-of-bounds case, and ragged sharding
- 976 passed / 0 failed across `test_concat.py`, `test_concat_program_cache.py`, `test_concat_iterative.py`, `base_functionality/test_concat_issue.py`, both nightly concat suites, and `tt_eager` `misc/test_concat.py`

**Not fixed here** — sharded *height* concat appends whole shards, so it is wrong whenever the dims before rank-2 don't collapse to 1 (repro: `(2, 1, 32, 32)`, `dim=-2`, width-sharded, shard `(64, 16)`, two cores). Reproduces identically at rank 4 before and after this branch, and in `ConcatBlockShardedProgramFactory`. Tracked separately as #55342.

---

### Pipeline status

- [x] Sanity
- [x] (Runtime) Sanity Test
- [x] Nightly for DM Category - Failed same as in main branch
- [ ] All model test
- [x] Single card
- [x] T3K
- [x] Galaxy

---

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:virdhatchani/concat_51214)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:virdhatchani/concat_51214)


tests/ttnn/unit_tests/operations/data_movement/test_concat.py
```diff
@@ -628,3 +628,138 @@ def test_concat_fp32_last_dim_mantissa_not_truncated(device, width, layout):
...
+
+    ttnn_out = ttnn.concat(ttnn_inputs, dim=dim, memory_config=output_mem)
+    # Spell out the config: the two callers below hardcode shard_shape/output_shard/num_cores,
+    # so those do not show up in the pytest node id.
+    assert tuple(ttnn_out.shape) == tuple(torch_out.shape), (
+        f"wrong output shape: got {tuple(ttnn_out.shape)}, expected {tuple(torch_out.shape)} "
+        f"for concat({num_inputs}x{tuple(shape)}, dim={dim}) with {strategy}, {layout}, {dtype}, "
+        f"shard {tuple(shard_shape)} -> {tuple(output_shard)} on {num_cores} core(s)"
+    )
+    assert_equal(torch_out, ttnn.to_torch(ttnn_out))
```
Aswinmcw, line 676: These tests check shape and values but never check that the output actually came back with the requested sharded config, so they can't distinguish "the right factory ran" from "a fallback ran and happened to produce the right data" — which is exactly the failure mode this PR is fixing (rank-5 `dim=-1` silently falling through to the interleaved factory). Worth one line here: `assert ttnn_out.memory_config() == output_mem` before the value comparison.

ttnn/cpp/ttnn/operations/data_movement/concat/device/concat_device_operation.cpp
```diff
@@ -43,6 +43,9 @@ ConcatDeviceOperation::program_factory_t ConcatDeviceOperation::select_program_f
     const bool output_nd_sharded = (TensorMemoryLayout::ND_SHARDED == args.output_mem_config.memory_layout());
     if (!input_nd_sharded && !output_nd_sharded) {
         const auto memory_layout = input_tensors[0].memory_config().memory_layout();
+        const uint32_t rank = input_tensors[0].logical_shape().rank();
+        const bool is_width_concat = rank >= 2 && args.dim == rank - 1;
```
Aswinmcw: Nit: this is now the third independent copy of the same predicate (here, `ConcatS2SMultiProgramFactory`, and `ConcatBlockShardedProgramFactory`), and they have already drifted — the block-sharded one is `dim == rank - 1` with no `rank >= 2` guard, so a rank-1 input there computes `rank - 2` as an unsigned wrap. A small shared helper (`is_width_concat(rank, dim)` / `is_height_concat(rank, dim)` in the concat common header) would keep the dispatch and the factories from disagreeing again the next time one of them is touched.

Aswinmcw (approved):
LGTM. Two nits inline. One ask on the write-up: the "Not fixed here" height-concat bug is a real known-wrong path with a repro in the body — please file it and link it from here, otherwise it only survives in this PR's description.

Aswinmcw (approved):
Both nits addressed, and the predicate is now single-sourced across the validator and `concat_impl` too, not just the three sites I pointed at. Checked the new helpers: `dim + N == rank` keeps rank-1 behaviour bit-identical to the old `dim == rank - 1` in the block-sharded factory while removing the wrap, and the `rank >= 2` guard is still applied at the s2s dispatch site where the factories index `shape[-2]`. Nothing further from me.

Still open from the last round (write-up only, not blocking): the "Not fixed here" height-concat bug has a concrete repro in the description but no issue filed — please open it and link it here so it doesn't only exist in this PR's body.

mateusznowakTT (approved):

sjameelTT (approved):

VirdhatchaniKN:
https://github.com/tenstorrent/tt-metal/actions/runs/34796978737 - passed

PR #55034, opened 2026-09-01
common/prefill: unify the layer-ack path and wire it unconditionally

jjovicicTT:
## What

One commit: unifies the prefill runner's layer-completion path and makes it unconditional.

Before this, there were two ack wirings selected by rank count plus a dead
`set_layer_ack_channel` registrar. Now there is one: every rank owns a
`LayerCompletionRouter` over a per-rank host-local ring; subordinates MPI-forward to
`master_rank`, which reorders and `inject()`s into the scheduler counter channel. At
`num_ranks == 1` the router never touches MPI, so the single-rank path costs nothing extra
structurally.

`PREFILL_ENABLE_LAYER_ACK` is gone. `PREFILL_LAYER_ACK_D2H` still selects the *source* — D2H
device records vs the host `on_layer_complete` callback.

## Why remove the knob

It defaulted off, so the router, the per-rank ring and the master reorder buffer were
unexercised by every shipped config. The machinery only ever ran when someone opted in by hand.

## Stacking

- **Base is `jjovicic/prefill_runner_unslop` (#55049)**, not `main` — that PR strips comments and
  docstrings from the same runner files, so this rebases on top of it. GitHub retargets to `main`
  when #55049 merges and its branch is deleted; if the branch survives, retarget by hand.
- **#55067** (`gpt_oss/prefill: add the per-layer completion sink`) is the additive half of the
  GPT-OSS change, split out to merge against `main` on its own. It is also contained here, so this
  branch stands alone against its own base; the overlap collapses on the rebase after #55067 lands.
- An unrelated `PREFILL_MANIFEST` existence check was dropped from this branch to keep the scope to
  layer acks alone.

## Consequences — please read before approving

1. **A runtime without `set_layer_completion_sink` can no longer serve at all** (unless it runs
   `PREFILL_LAYER_ACK_D2H=1`). All three current runtimes implement it — GPT-OSS only as of this
   branch / #55067. A new model must.
2. **The stale-shm unlink now applies to every run.** That unlink *is* the post-crash recovery
   path, so it stays blind — but two concurrent runs sharing one `PREFILL_H2D_SERVICE_ID` on one
   host will evict each other's ack channel. Documented at `_unlink_stale_shm`, not fixed.
3. **Every run now takes a mandatory 30s `LayerCompletionQueue.connect` startup dependency.**
4. **No perf number for the host-callback source.** The only measurement we have is D2H
   (~7% untraced, ~1% traced). I cannot tell you what the host-callback path costs.

I argued for keeping the opt-in and enabling acks in the CI bindings instead; the always-on
call was made deliberately over that objection.

## Validation

Single-rank on `bh-glx-b06u08` (Kimi-2.7, 61 layers, 8x4), three legs, expecting 366 acks
(2 requests x 3 chunks x 61 layers) each — results in the follow-up comment.

- **H1** — a binding that still sets `PREFILL_ENABLE_LAYER_ACK: "0"` (now inert): 366/366. The
  decisive regression leg — an acks-off config must bring the router up and still serve.
- **H2** — shipped `1rank_host_ack` + mock migration, verifying producer: ack path 61/61 per slot.
  Producer exits 1 on `min_pcc=0.890920` against its 0.93 gate; that matches the recorded
  `0.890774` baseline to 1.5e-4 and is the pre-existing Kimi-2.7 `nope` depth droop, **not a pass**.
- **H3** — shipped `1rank_d2h_ack`, D2H source: 366/366.

2-rank was validated on `bh-glx-120-c10u[14,20]` before the squash — host-callback 366/366 and D2H
366/366, rank1 owning layers [31,61), so a full count proves global seqs survive the MPI forward.

**These legs predate the rebase onto #55049.** That rebase pulled in `main`'s traced-D2H path
(`set_d2h_ack_service` / `warmup_ack_count`), which the branch had never carried, so H3 in
particular needs a re-run before this leaves draft.

Two pre-existing bugs found while reading this code are deliberately left alone (duplicate
mock-migration publish under a late `not single_rank` guard; a wrong error string claiming a
merged mock table isn't implemented) — out of scope, not regressions from this branch.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:jjovicic/unslop_runner)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:jjovicic/unslop_runner)


jjovicicTT:
## Validation results — single-rank, `bh-glx-b06u08`, Kimi-2.7 (61 layers, 8x4 mesh)

Commit under test: `c6bb51460ab`. Expected 366 acks per leg (2 requests x 3 chunks x 61 layers).

| Leg | Config | Ack source | Acks | Shutdown | Verdict |
|---|---|---|---|---|---|
| H1 | binding with `PREFILL_ENABLE_LAYER_ACK: "0"` (now inert) | host callback | **366/366** in 31.3s | clean on sentinel after 6 chunks | **PASS** |
| H2 | shipped `1rank_host_ack` + `PREFILL_MOCK_MIGRATION=1` | host callback | 61/61 layers read on both slots | clean on sentinel after 6 chunks | ack path PASS, **PCC gate RED** |
| H3 | shipped `1rank_d2h_ack` | D2H device records | **366/366** in 39.8s | clean on sentinel after 6 chunks | **PASS** |

### H1 — the leg that matters

A binding explicitly asking for acks off now brings the router up anyway and serves:

```
binding_1rank_acks_off.yaml:17   PREFILL_ENABLE_LAYER_ACK: "0"
routing up: rank=0/1 master=0 ring=/tt_prefill_layer_completion_ring_0
            source=host on_layer_complete callback
            (owns scheduler channel /tt_prefill_layer_acks_ds_prefill)
[acks] FINAL 366/366 in 31.3s   VERDICT PASS
[pp rank 0] SHUTDOWN sentinel received after 6 chunks; exiting request loop
```

That was the actual regression risk of deleting the knob: a stale config that used to skip the ack
wiring entirely. It wires, routes all 366 completions in seq order into the scheduler channel, and
tears down cleanly.

### H2 — PCC is red, and it is not this change

```
slot 0 KV PCC over [0,15360) across 61/61 local layers -> 0.890920
slot 1 KV PCC over [0,15360) across 61/61 local layers -> 0.890920
ERROR  KV cache PCC below 0.93 ... min_pcc=0.890920
```

The producer's own 0.93 threshold fails, so it exits 1. This is the pre-existing Kimi-2.7 `nope`
depth droop: the unification validation on `bh-glx-120-c10u[14,20]` recorded `min_pcc 0.890774`, so
this run is a 1.5e-4 delta on different hardware. Per-layer `pe` holds ~0.988-0.996 throughout while
`nope` sags with depth (0.942 at layer 60, 0.891 at layer 59, 0.900 at layer 49). All 61/61 layers
were read on both slots and the runner shut down cleanly — the ack machinery did its job.

Calling this "unchanged from baseline", not "passing". The gate is genuinely red because the model
does not reach 0.93 on this config.

### Not covered

- **2-rank under this commit.** The unification commits were validated 2-rank on `c10u[14,20]`
  (host-callback 366/366, D2H 366/366, rank1 owning layers [31,61)); the knob-removal commit only
  deletes the gate and dedents, but it has not been re-run 2-rank.
- **Cost of the host-callback source.** Still unmeasured. The only number on record is D2H
  (~7% untraced, ~1% traced).

PR #55036, opened 2026-09-01
[LLK] Verify the perf schema catalog from all three sides

nstojicTT:
### PR Category

Test infrastructure

### Summary

`helpers/perf/test_schemas.py` is the **catalog**: it records which columns each perf
test writes, plus a `version` so two reports of the same test are comparable. The
dashboard and the future PR gate depend on it being true.

It was checked from one side only, and that side could not see itself being wrong.
`test_perf_header_gate.py` compares the catalog against columns re-derived from the
test **source** by a static AST reader — and the catalog was generated by that same
reader. Both ends share any blind spot, so they agree with each other no matter what
the CSV contains.

The reader had two blind spots. **22 of 45 entries under-report the real CSV by four
to eight columns**, and the gate has been green the whole time.

This PR fixes the reader, corrects the catalog, and adds the two checks that make the
mistake impossible to repeat.

**No data was ever at risk.** Those columns are declared in `wide_schema.py`, so every
CSV and every Parquet has always carried them correctly. Only the record was wrong.

---

### 1. The reader cannot see a helper that returns a parameter object

`helpers/test_variant_parameters.py:1155`

```python
def generate_input_dim(srcA, srcB, block_ct_dim=None, block_rt_dim=None, tile_dimensions=(32, 32)):
    ...
    return INPUT_DIMENSIONS(full_rt_dim, full_ct_dim, block_ct_dim, block_rt_dim)
```

A test writes `templates=[generate_input_dim(dims, dims)]`. The reader looks
`generate_input_dim` up in its map of parameter classes. It is a plain function, so
the lookup misses and the reader emits nothing.

**Invisible:** `full_rt_dim`, `full_ct_dim`, `block_rt_dim`, `block_ct_dim`.

### 2. The reader cannot see fields filled by `__post_init__`

`helpers/test_variant_parameters.py:1480`

```python
class NUM_BLOCKS(RuntimeParameter):
    num_blocks: int = 1
    input_num_blocks: int = None
    output_num_blocks: int = None

    def __post_init__(self):
        if self.input_num_blocks is None:
            self.input_num_blocks = self.num_blocks
        if self.output_num_blocks is None:
            self.output_num_blocks = self.num_blocks
```

The reader's rule is *"a None-default field the call does not set is dropped"*. That is
right for genuinely optional slots like `MATH_OP.pool_type`, and wrong here, because
`__post_init__` always fills these. `NUM_BLOCKS(4)` looks like one column and is really
three. `NUM_TILES_IN_BLOCK` behaves the same way.

### 3. The gate could not detect either one

```
reader says 15  ->  catalog written with 15  ->  gate compares 15 vs 15  ->  PASS
                                                              real CSV:  20
```

This is the root cause, and it is the part worth fixing. Patching the two blind spots
alone would leave the next one to hide in exactly the same way.

### 4. The catalog is wrong today, and drifting further

11 of the 25 tests that ran in the last nightly disagree with their catalog entry, and
11 Quasar entries have the same two causes. `perf_eltwise_unary_datacopy` and
`perf_pack` are both **version 1**, merged 2026-08-29: born wrong, waved through by the
gate. Every new test using `generate_input_dim` or `NUM_BLOCKS` adds another.

### 5. Nothing ever compared the catalog to the CSV

The authoritative answer to *"what columns does this test write"* is the CSV the run
wrote. It was never consulted.

### 6. Nothing checked for a dead published column

A column could sit in `wide_schema.py` that no test produces — left behind by a deleted
or renamed test — and nothing would complain.

---

## What this PR does

Four changes. They only work together: correcting the catalog without fixing the reader
breaks the static gate, and fixing the reader without the new checks leaves the same
one-sided design that hid all of this.

### `perf_schema_derive.py` — teach the reader both constructs

| Added | Job |
|---|---|
| `helper_returned_param_calls(specs)` | Maps a function name to the parameter constructors it returns, so `generate_input_dim(...)` resolves to `INPUT_DIMENSIONS` |
| `class_post_init_copies()` | Parses `if self.dest is None: self.dest = self.src` out of `__post_init__`, and emits `dest` whenever `src` is emitted |

`emitted_fields` takes the two maps as optional arguments. Callers passing neither keep
the old behaviour, so nothing outside the gate changes.

### `helpers/perf/test_schemas.py` — correct the record

22 entries gain their missing columns and get a version bump. Columns are inserted in
alphabetical position within each existing list, so the diff is pure additions. **No
column is removed anywhere, and no test source changes.**

| Test | Version | Columns added |
|---|---|---|
| `perf_eltwise_unary_datacopy` | 1 → 2 | 8 |
| `perf_pack` | 1 → 2 | 8 |
| `perf_pack_dest_bank` | 3 → 4 | 4 block |
| `perf_fast_tilize` | 3 → 4 | 4 dim |
| `perf_fast_tilize_full` | 3 → 4 | 4 dim |
| `perf_fast_untilize` | 3 → 4 | 4 dim |
| `perf_fast_untilize_baseline_compare` | 3 → 4 | 4 dim |
| `perf_pack_untilize` | 3 → 4 | 4 dim |
| `perf_sfpu_reduce` | 4 → 5 | 4 dim |
| `perf_sfpu_reduce_sdpa` | 3 → 4 | 4 dim |
| `perf_unpack_tilize` | 3 → 4 | 4 dim |

Plus 11 Quasar entries with the same two causes. The Quasar half is verified through the
reader only — its perf leg is compile-only and emits no CSV.

### `helpers/perf/core.py` — check the catalog against the CSV a run wrote

`_assert_matches_catalog`, called from `combine_perf_reports` after the duplicate-key
check. The columns the run produced must equal the catalog entry, both directions:

```
PerfSchemaError: perf_fast_tilize.csv: the CSV this run produced does not match
catalog entry 'perf_fast_tilize' (version 3). In the CSV but not the catalog:
['block_ct_dim', 'block_rt_dim', 'full_ct_dim', 'full_rt_dim']. Update
PERF_TEST_SCHEMAS in helpers/perf/test_schemas.py and raise its version. If
test_perf_header_gate.py still passes, its source reader cannot see the construct
that produced the column — fix the reader too.
```

That last sentence is the safety net for problem 3. When only the runtime side trips,
the reader has a **new** blind spot — the condition nobody could detect before.

A base name with no catalog entry is skipped rather than failed: the static gate already
reports *"found in source but missing from the catalog"*, and `combine_perf_reports`
globs whatever CSVs are on disk, so a partial local run must not blow up.

### `test_perf_header_gate.py` — check the published table against the catalog

`test_no_published_column_without_a_perf_test`: every column a wide schema says a test
fills must appear in some catalog entry. Provenance (`origin="ci"`), generated timing
columns and `speed_of_light` are excluded. Passes today, and verified to fail when a
dead column is injected.

### The three sides, after this PR

| Check | Direction | Runs in |
|---|---|---|
| catalog ↔ test source | both ways | PR gate, static |
| **catalog ↔ the CSV a run wrote** | **both ways** | **perf run** — new |
| **published table → catalog** | **one way** | **PR gate, static** — new |
| CSV → `DB_SCHEMA` | one way | perf run, existing Parquet strict mode |

A column can no longer reach the Parquet without being in both the catalog and
`DB_SCHEMA`, and a `DB_SCHEMA` column can no longer sit there unclaimed.

---

## Repro: the reader

No hardware, no test run.

```bash
python3 - <<'PY'
import ast, sys; sys.path.insert(0, ".")
import perf_schema_derive as D
specs = D.class_field_specs()
for src in ('generate_input_dim((32, 32), (32, 32))', 'NUM_BLOCKS(4)', 'NUM_TILES_IN_BLOCK(8)'):
    print(f"  {src:<42} -> {D.emitted_fields(ast.parse(src).body[0].value, specs)}")
PY
```

**Before**, on `main`:

```
  generate_input_dim((32, 32), (32, 32))     -> []
  NUM_BLOCKS(4)                              -> ['num_blocks']
  NUM_TILES_IN_BLOCK(8)                      -> ['num_tiles_in_block']
```

**After**, passing the two new maps in (`helpers = D.helper_returned_param_calls(specs)`,
`post = D.class_post_init_copies()`):

```
  generate_input_dim((32, 32), (32, 32))     -> ['block_ct_dim', 'block_rt_dim', 'full_ct_dim', 'full_rt_dim']
  NUM_BLOCKS(4)                              -> ['input_num_blocks', 'num_blocks', 'output_num_blocks']
  NUM_TILES_IN_BLOCK(8)                      -> ['input_num_tiles_in_block', 'num_tiles_in_block', 'output_num_tiles_in_block']
```

## Repro: the catalog against real data

The reader agreeing with itself proves nothing. Take any `perf-data-*` artifact from
[nightly 33465181016](https://github.com/tenstorrent/tt-metal/actions/runs/33465181016)
and compare. Key columns are everything up to and including `marker`:

```python
import pandas as pd, importlib.util
spec = importlib.util.spec_from_file_location("cat", "helpers/perf/test_schemas.py")
cat = importlib.util.module_from_spec(spec); spec.loader.exec_module(cat)

cols = list(pd.read_csv("<artifact>/perf_fast_tilize/perf_fast_tilize.csv", nrows=1).columns)
produced = set(cols[: cols.index("marker") + 1]) - {"speed_of_light"}
recorded = set(cat.PERF_TEST_SCHEMAS["perf_fast_tilize"]["columns"])
print("in CSV, not in catalog:", sorted(produced - recorded))
print("in catalog, not in CSV:", sorted(recorded - produced))
```

**Before** — `perf_fast_tilize` at version 3 records 15 columns; the CSV has 20:

```
in CSV, not in catalog: ['block_ct_dim', 'block_rt_dim', 'full_ct_dim', 'full_rt_dim']
in catalog, not in CSV: []
```

**After** — empty both ways.

---

## Verification

Checked **per arch**, not as a union, so a Wormhole/Blackhole difference cannot hide:

```
wormhole : 22 tests produced data | 0 mismatches | none missing from the catalog
blackhole: 22 tests produced data | 0 mismatches | none missing from the catalog
no shard disagreed with another shard on any test's columns

BEFORE (main)    : 11 of 25 tests disagree with the real nightly CSV
AFTER  (this PR) :  0 of 25
```

`perf_fused` is the one catalog entry that produced no data. It carries
`@skip_for_blackhole`, `@skip_for_wormhole` and `@skip_for_quasar`, so it never runs.

The derivation change was verified to produce **byte-identical** derived schemas to the
version it came from, for all 26 WH/BH and 19 Quasar tests. A full LLK perf run on this
branch is in flight.

## Tests

Six added, none needing hardware.

`test_perf_header_gate.py` — one test pinning both constructs and the rule they must not
break:

```python
assert {"full_rt_dim", "full_ct_dim", "block_ct_dim", "block_rt_dim"} <= fields("generate_input_dim((32, 32), (32, 32))")
assert {"num_blocks", "input_num_blocks", "output_num_blocks"} <= fields("NUM_BLOCKS(4)")
assert not {"unary_extra", "pool_type"} & fields("MATH_OP(MathOperation.Elwadd)")
```

The third line matters: it proves the fix did not turn every optional slot into a
column, which would swing the error the other way. Plus the dead-column test.

`test_perf_report_hw_free.py` — four for the runtime check: match passes and
`speed_of_light` is ignored; an unrecorded column raises; a recorded column the CSV lost
raises; an unknown base name is skipped. Each registers a throwaway entry in **both**
catalogs, so the WH/BH-versus-Quasar branch cannot skew the result.

## Notes

**Why the catalog-column-absent direction is safe under pytest-split.** A shard runs
part of a sweep, so this looks like it must false-positive. It cannot:
`_assert_combined_schema` already refuses a test whose cases emit different columns, so
every case of one test has an identical column set and any shard's CSV carries the full
set.

**`speed_of_light`** is excluded from both new checks. It is a property of how the run
was built, not of what the test sweeps, and the catalog has never included it.

**The 22 version bumps are a record correction, not a data change.** No test's columns
changed. Anything keyed on `version` should not read them as a discontinuity.

## Credit

The diagnosis and the derivation fix come from @ndivnicTT's work on
`ndivnic/all_perf_changes_squashed` (commit `adbd45e6aa3`), which is where this was
found. The implementation here is a shorter equivalent with the same behaviour.

## Not covered

A column the harness **overwrites** rather than adds. `dest_acc` is promoted from `No`
to `Yes` for outlier format combinations (#54574), and the CSV records the promoted
value. The column set is unchanged, so none of these checks sees anything wrong unless
the override happens to collide with another sweep point. Catching that means comparing
the value a test *requested* against the value `TestConfig` *used*. Separate issue.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nstojictt/perf-catalog-two-way-verification)


ndivnicTT:
> test review body
> 
> > 🧠 _Reviewed using Matt Pocock's skills by [Matt Pocock Skills Reviewer](https://github.com/tenstorrent/tt-metal/actions/runs/33518323219)_ · sonnet50 · 67.3 AIC · ⌖ 6.8 AIC · ⊞ 11.1K
> 
> ### Comments that could not be inline-anchored
> tt_metal/tt-llk/tests/python_tests/perf_schema_derive.py:633
> **[/codebase-design]** `helpers.setdefault(fn.name, got)` silently keeps only the first-found return set if two different files (or two functions) share the same helper name — a second, differently-shaped `generate_input_dim`-style helper would have its columns dropped with no error, which is exactly the kind of blind spot this PR set out to eliminate.
> 
> <details> <summary>💡 Suggestion</summary>
> 
> Consider raising (or at least asserting/logging) when a helper name is already registered with a di…
> 
> tt_metal/tt-llk/tests/python_tests/perf_schema_derive.py:701
> **[/tdd]** Good coverage of the two new constructs and a boundary check that optional slots (`pool_type`, `unary_extra`) stay dropped — that third assertion is exactly the kind of regression-preventing edge case `/tdd` calls for.
> 
> <details> <summary>💡 One gap worth a follow-up test</summary>
> 
> No test exercises a helper whose return value is itself another helper call one level deeper, or a class with **no** entry in `helpers`/`post_init` (the `helpers and name in helpers` / `.get(name, {})` gu…

@nstojicTT Please check this before merging

ndivnicTT (approved):
LGTM! Thanks for this change!

nstojicTT:
> > test review body
> > > 🧠 _Reviewed using Matt Pocock's skills by [Matt Pocock Skills Reviewer](https://github.com/tenstorrent/tt-metal/actions/runs/33518323219)_ · sonnet50 · 67.3 AIC · ⌖ 6.8 AIC · ⊞ 11.1K
> > 
> > 
> > ### Comments that could not be inline-anchored
> > tt_metal/tt-llk/tests/python_tests/perf_schema_derive.py:633
> > **[/codebase-design]** `helpers.setdefault(fn.name, got)` silently keeps only the first-found return set if two different files (or two functions) share the same helper name — a second, differently-shaped `generate_input_dim`-style helper would have its columns dropped with no error, which is exactly the kind of blind spot this PR set out to eliminate.
> >  💡 Suggestion
> > Consider raising (or at least asserting/logging) when a helper name is already registered with a di…
> > tt_metal/tt-llk/tests/python_tests/perf_schema_derive.py:701
> > **[/tdd]** Good coverage of the two new constructs and a boundary check that optional slots (`pool_type`, `unary_extra`) stay dropped — that third assertion is exactly the kind of regression-preventing edge case `/tdd` calls for.
> >  💡 One gap worth a follow-up test
> > No test exercises a helper whose return value is itself another helper call one level deeper, or a class with **no** entry in `helpers`/`post_init` (the `helpers and name in helpers` / `.get(name, {})` gu…
> 
> @nstojicTT Please check this before merging

This was fixed, tnx

PR #55068, opened 2026-09-01
[DeepSeek] Reduce MoE gate compute modes to device_fp32 + host_all

nostojicTT:
### Problem

`GateComputeMode` had grown to nine members. Only two matter for the DeepSeek-family gate:

- **`DEVICE_FP32`** — the production path; what CI should validate.
- **`HOST_ALL`** — an exact-torch reference, worth keeping so a device-gate failure can be bisected locally, but it should not trigger in CI by default.

`DEVICE` (bf16) is legacy, `HOST_MATMUL` / `HOST_GROUPED_GATE` are half-way debug hybrids nothing selected, and `GPT_HOST` / `GPT_DEVICE` are dropped along with GPT-OSS gate coverage. The V4 hash gate is a separate routing algorithm and stays, split the same way.

Result: `GateComputeMode` is `DEVICE_FP32 | HOST_ALL | HASH_HOST | HASH_DEVICE`.

### What `DEVICE` actually was

Worth stating, because "bf16 vs fp32" undersells it. `DEVICE` was a strict capability **subset** of `DEVICE_FP32`:

| | `DEVICE` | `DEVICE_FP32` |
|---|---|---|
| Phase-1 matmul | `_device_matmul` | `_device_matmul` — identical |
| top-k op, multi-group | `deepseek_grouped_gate` (bf16) | `moe_grouped_topk` (upcasts to fp32 in-kernel) |
| top-k op, single-group | `moe_grouped_topk` — bf16 op has no single-group support | `moe_grouped_topk` |
| padding / chunked prefill | **none** — no `actual_isl` / `padding_config` / `actual_start` forwarded | full |

Without the padding args, padded rows never get sentinel expert indices, so downstream `masked_bincount`/dispatch/combine treat them as real tokens. `tt_moe.py:661` already gated the padding-aware path on `DEVICE_FP32` explicitly. For single-group models (Kimi, GLM) the two modes already called the same op, so those conversions are near no-ops at the op level.

### Changes

**Implementation** (`tt/moe/tt_moe_gate_prefill.py`) — dropped the five enum members, their `forward` branches, and the four helpers that became unreachable (`_device_grouped_gate`, `_compose_logits_to_host`, `_device_gpt_gate`, `_host_gpt_gate`). Constructor default moves `DEVICE` → `DEVICE_FP32`; all eight call sites pass `fallback_mode` explicitly, so this is inert today.

**Tests** — every `DEVICE` case converted to `DEVICE_FP32` across the block/transformer/cache suite, with one exception: `test_ds_prefill_transformer` listed both `(256, DEVICE)` and `(256, DEVICE_FP32)`, so converting would have produced a duplicate — the `e256_device` row is **deleted** instead. See "Coverage removed" below. **Pytest ids are held byte-identical** (`moe_gate_device`, `e384_device`, `e256_device`, `gate_device`, `perf-host-64`), so every CI `-k` selector keeps matching. **No CI YAML changes.**

**Two latent bugs fixed on the way:**
- `test_hash_gate_forward_pass` had no `uncollect_if`, so `hash_host` was running in CI. Now `HASH_DEVICE` only.
- `test_gate_cache.py` and `test_moe_cache.py` were both gating CI on `GateComputeMode.DEVICE` — i.e. validating the *legacy bf16* gate. They now validate the production one, at the same item count.

**GPT-OSS** — gate cases deleted. Production GPT-OSS routing does not use this gate: `gpt_oss_d_p/tt/moe/router.py` states it "cannot reuse DeepSeek's `TtMoEGatePrefill`" and implements `TtGptOssRouter`, and `GptOssPrefillAdapter` subclasses `PrefillModelAdapter` directly rather than the `MLAPrefillAdapter` that holds the enum lookup. This removes duplicate coverage, not the real path. `GptOss120BConfig` stays — four op unit tests use it.

### The conversion is not precision-only — it turns padding-awareness on

`tt_moe.py:661` is **not touched by this PR**:

```python
if actual_isl is not None and self.gate.fallback_mode == GateComputeMode.DEVICE_FP32:
    padding_config = ...   # padded rows get sentinel expert indices
```

`DEVICE` never satisfied that guard, so every test converted `DEVICE` → `DEVICE_FP32` that passes `actual_isl` now builds a padding config where it previously built none: padded token rows are sentinel-marked and dropped from MoE instead of being dispatched as real tokens. That is the correct behaviour and is precisely the capability gap `DEVICE` had — but it changes what those tests exercise, and their PCC baselines were established without it.

Affected: `test_kimi_prefill_transformer` / `test_glm_prefill_transformer` (`actual_isl=number_of_non_padded_tokens`) and the DS chunked transformer tests. `test_glm_prefill_block` passes `actual_isl=seq_len` (full length), so its config is a no-op.

The hard-failure path is **unreachable**: `build_padding_config_device` raises on `is_balanced=True`, but `mla.py:904` already asserts `is_balanced=False` for chunked prefill and every DS chunked test hard-codes it, so the traced+balanced combination cannot occur.

No blaze leg should fail from this — the one affected CI leg (`prefill_transformer_determinism`, blaze:176) runs `smoke` + `with_determinism`, i.e. `pcc_validation=False`, and determinism compares iterations against each other, so a consistent change does not break it.

### One regression caught and fixed in review

`run_model`'s threshold branch tested `== GateComputeMode.DEVICE`, but the **Kimi** block case was already parametrized `DEVICE_FP32`. It never matched, fell to the `else` arm, and took `KIMI_THRESHOLDS.moe_gate_host = 0.950` — which is why that override exists at all. Retargeting the branch to `DEVICE_FP32` made it match, silently moving Kimi to the generic `moe_gate_device = 0.992`. Third commit moves the 0.950 onto `moe_gate_device` so the effective bar is unchanged. DSV3 is unaffected (its case took the device arm before and after).

Worth a reviewer's eye: this is the one place in the diff where a converted comparison changed which branch a *pre-existing* `DEVICE_FP32` case takes. Every other converted comparison moved together with its own parametrize list.

### Coverage removed

Two things this PR gives up, both deliberate — calling them out so reviewers can object:

1. **The model-level bf16-vs-fp32 baseline.** `test_ds_prefill_transformer` ran `e256_device` (bf16 gate) and `e256_device_fp32` side by side: identical config, same golden trace, thresholds 0.88 vs 0.95. That gap *was* the measured end-to-end cost of the bf16 gate. Deleting the bf16 row removes it, and `TRACE_PCC_THRESHOLD_DEVICE_BF16` goes with it. Keeping it would mean keeping `GateComputeMode.DEVICE`, which is the point of the PR. The op-level equivalent survives and is untouched: `tests/ttnn/nightly/unit_tests/operations/experimental/deepseek_prefill/test_moe_grouped_topk.py` parametrizes `INPUT_DTYPES = [float32, bfloat16]` with `WEIGHTS_PCC_THRESHOLD_FP32 = 0.96` vs `..._BF16 = 0.85`. Note that test is currently excluded from the Wormhole op-unit leg (`ops_unit_tests.yaml:82`, `-k "not test_moe_grouped_topk"`, hangs on WH), so the coverage exists but does not execute there.

2. **The GPT-OSS gate cases**, as described above — duplicate of a path production does not take.

### Verification

Verified statically (no ttnn/pytest in the authoring environment — **the runs below still need doing**):

- No reference to a deleted member survives in any `.py` / `.yaml` / `.json` / `.md` / `.sh`.
- Guard rewrites are behaviour-preserving — set algebra checked per guard for all four surviving modes.
- AST pass: no orphaned locals in `forward()` (the `NameError` class `py_compile` misses), no unused imports, and all 32 `uncollect_if` predicates in the module resolve against a declared param.
- Every `GateComputeMode[name]` input — 7 adapter defaults, 6 manifests, 6 topology YAMLs, 2 docs — is already `"DEVICE_FP32"`.
- `blaze:70` still collects 9 items on CI (7 gate + 2 hash), so no exit-code-5 risk.

Suggested before merge:

```bash
CI=true pytest models/demos/deepseek_v3_d_p/tests/pcc/test_moe_gate_prefill2d.py --collect-only -q
pytest models/demos/deepseek_v3_d_p/tests/pcc/test_moe_gate_prefill2d.py -k "fabric2d-mesh-2x2" -vv
pytest models/demos/deepseek_v3_d_p/tests/cache/ -vv
```

### Reviewer attention

1. **`test_prefill_block.py:611`** — the `strict=True` xfail for the FABRIC_2D 2x4 device-gate PCC regression was calibrated on the bf16 gate. Under fp32 it may XPASS, which `strict` turns into a failure. Needs one run on that combo.
2. **Kimi and GLM local PCC cases** — deleting `TRACE_PCC_THRESHOLD_DEVICE_BF16` moves both `test_kimi_prefill_transformer` (`e384_device`) and `test_glm_prefill_transformer` (`e256_device`) from the bf16 trace bar 0.88 to the fp32 bar 0.95, since both legs moved from `DEVICE` to `DEVICE_FP32`. fp32 should clear it, but 0.95 was never measured for either model. The blaze CI leg is unaffected (it selects `smoke` + `with_determinism` ⇒ `pcc_validation=False`, which never reaches that branch), so exposure is local-only, but 0.95 was never calibrated for Kimi.
3. **Open question** — the two `HOST_ALL`-only Kimi tests in `test_prefill_block_chunked.py` got the `uncollect_if` for consistency, which means they collect *zero* items under `CI=true`. That file is in no pipeline so it changes nothing today, and elsewhere the predicate means "pick the device mode over the host twin" whereas here it means "drop the test". Happy to revert those two decorators if reviewers prefer.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:nostojic/gate_clearing)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:nostojic/gate_clearing)


ddjekicTT:
- [x] https://github.com/tenstorrent/tt-metal/actions/runs/33618407425

PR #55091, opened 2026-09-01
[Bug fix] reduce_scatter: fix the zero-page worker deadlock on the mux path

llongTT:
### PR Category
Bug fix

### Summary
`reduce_scatter_minimal_async` on the ring topology deadlocked whenever the page split left a worker with nothing to do: no `TT_FATAL`, no watcher trip, the host spinning at ~110% CPU, boards needing a reset. Reproduced deterministically on a 1x8 Blackhole ring, bfloat16, scatter dim 1, 2 links:

```
per-device [8,8,32,32]   1 page  over 4 workers  ->  hang
per-device [8,8,32,64]   2 pages over 4 workers  ->  hang
per-device [8,8,64,64]   4 pages over 4 workers  ->  passes
```

**Mechanism, from tt-triage on the live hang.** The pages of one channel are divided across `num_links * num_workers_per_direction` workers, so the first two splits leave workers with no pages at all. The ring writer opens its FabricMuxV2 client in eager-staging mode: a packet only reaches the mux after a `flush()` completes the connection. Every send path in the writer flushes after its first packet except the end-of-batch increment on `batch_ready_sem`. A worker that owns no pages sends nothing else, so that increment sits in the staging slot forever and its peer never leaves the semaphore wait. The dump shows exactly that on every device: the zero-page writers parked in `noc_semaphore_wait_min` at the batch-ready wait (`ring_reduce_scatter_minimal_async_writer.cpp:822`), every one-page worker DONE, the mux forwarder and manager idling while they wait for those clients to close. It needs `B > 1` (the increment is only sent between batches), no barrier semaphore (that path would have completed the connection) and the multi-worker path (the direct fabric connection is opened blocking).

**Change.** One flush after the batch-ready send in `ring_reduce_scatter_minimal_async_writer.cpp`, like every other send in the kernel.

**Tests.** Three splits on the 4-device ring and line tests: a worker with no pages, one page per worker, two pages per worker. The line writer opens its mux with a blocking connect and has no batch-ready send, so it never had this deadlock; its cases pin that a worker with no pages completes there too.

### Notes for reviewers
- An earlier revision of this PR capped the worker count so the heuristic never built a worker without pages. It is not needed and has been dropped: with the flush alone, the full functional sweep over all 58 shape/dim/algorithm cases of the 1x8 ring passes, and the two hanging shapes complete in ~10 s under the default heuristic (2 workers per direction, so 4 workers sharing 1-2 pages, some with none).
- The fix was validated with the hang configuration forced (`num_workers_per_link=2`): both shapes pass in ~8 s.
- Local runs of `test_minimal_reduce_scatter_async_bh.py` on this branch, on its base commit and on main's tip all show 5-6 `Timeout (>300.0s)` failures in the traced `perf` line-topology cases with the batch_1/2/4/8 shapes, in a subset that shifts between runs. Nothing else fails in any run. Those timeouts are pre-existing and timing-dependent and are unrelated to this change, which only touches the ring writer; tracked separately in #55394.
- The ring regression cases are gated to 4-device boxes and skip on 8-device machines; the line cases run on both.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:llong/rs_worker_cap_fix)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:llong/rs_worker_cap_fix)


amorrisonTT:
Would it be worth running tt-triage to find the actual mechanism behind the hang? Right now this is entirely empirical. A worker with 0 pages, ok intuitively one might expect that to cause a problem but a worker with 1 page feels like it should be ok

llongTT:
> Would it be worth running tt-triage to find the actual mechanism behind the hang? Right now this is entirely empirical. A worker with 0 pages, ok intuitively one might expect that to cause a problem but a worker with 1 page feels like it should be ok

Yes, currently it's empirical, I can extend the PR a bit to dig out the real issue

llongTT:
> Would it be worth running tt-triage to find the actual mechanism behind the hang? Right now this is entirely empirical. A worker with 0 pages, ok intuitively one might expect that to cause a problem but a worker with 1 page feels like it should be ok

Done, and your intuition was right: a worker with one page is fine. The hang needs a worker with **zero** pages. The "one page per worker" reading came from counting workers per link; the pages are divided across `num_links * workers_per_direction`, and this box runs 2 links, so "2 pages over 2 workers" was really 2 pages over 4 workers with two of them owning nothing.

**Dump.** `[8,8,32,64]`, dim 1, 8-device ring, 2 links, `num_workers_per_link=2` forced. tt-triage on the live hang, per device:

| core | kernel | state |
|---|---|---|
| zero-page workers | `ring_reduce_scatter_minimal_async_writer` | `noc_semaphore_wait_min` at writer.cpp:822, the `batch_ready_sem` wait |
| same cores, reader / compute | done | |
| one-page workers | all RISCs | DONE |
| mux cores | forwarder in `service_channels`, manager in the steady-state loop | waiting for clients to close |

**Mechanism.** The ring writer opens its `FabricMuxV2Sender` with `EAGER_STAGING`: a packet written into the mux slot is only signalled to the forwarder once `flush()` completes the connection. Every send path in the writer calls the flush after its first packet (data chunks, standalone increments, the barrier) except the end-of-batch increment on `batch_ready_sem`. A worker with no pages sends nothing else, so that increment sits in its staging slot forever and its peer never leaves the wait. Nothing faults, hence no `TT_FATAL` and no watcher trip. It needs `B > 1`, no barrier semaphore, and the multi-worker path (the direct fabric connection is opened blocking).

**Fix.** `mf.flush()` after the batch-ready send. The PR is now that one-line kernel change plus regression tests; the worker cap from the first revision is dropped, since with the flush alone the full 58-case functional sweep of the 1x8 ring passes and the two hanging shapes complete in ~10 s with the default heuristic (4 workers on 1-2 pages, some with none).

**Line topology.** Its writer opens the older mux client with a blocking connect and has no batch-ready send, so it never had this deadlock. The same three boundary splits are added to the line test anyway.

llongTT:
Deadlock root cause analysis:https://claude.ai/code/artifact/a235cd2e-3770-43e1-af68-cb1ed6152f96

tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_minimal_reduce_scatter_async_bh.py
```diff
@@ -269,6 +269,20 @@ def run_op(i):
...
+        # there: its only fabric transaction was the end-of-batch increment, whose send path did not
+        # complete the staged connection, so the increment never left the staging slot and the peer
+        # waited on batch_ready_sem forever -- silently, no TT_FATAL, no watcher trip, board reset
+        # needed. Confirmed with tt-triage on an 8-device, 2-link ring as [8,8,32,32] and [8,8,32,64].
+        # The ring writer now flushes that send like every other send.
+        # First case: 1 page, 2 workers -> one worker owns nothing (the deadlock). Second: 2 pages,
+        # 2 workers -> one page each. Third: 4 pages, two each; always worked.
+        (4, [4, 4, 32, 32], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16),
+        (4, [4, 4, 32, 64], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16),
+        (4, [4, 4, 64, 64], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16),
```
scardozaTT: I feel like only the first test case here is relevant to this PR and a corner case, the others are generic. Consider keeping just the first test case and removing the other two.

tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_minimal_reduce_scatter_async_bh.py
```diff
@@ -385,6 +402,12 @@ def test_reduce_scatter_async_4dev_ring(
         # composite_reduce_scatter when the input-side alignment check is insufficient and only the
         # dispatch-side (per-device output) check fires.
         (4, [1, 1, 32, 64], 3, ttnn.TILE_LAYOUT, ttnn.bfloat8_b, True),
+        # Same three splits on the line topology. The line writer opens its mux with a blocking
+        # connect and has no batch-ready send, so it never had the ring writer's staging deadlock;
+        # these pin that a worker with no pages completes there too.
+        (4, [4, 4, 32, 32], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16, False),
+        (4, [4, 4, 32, 64], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16, False),
+        (4, [4, 4, 64, 64], 1, ttnn.TILE_LAYOUT, ttnn.bfloat16, False),
```
scardozaTT: Same here, I'd suggest keeping just the first test case.

ttnn/cpp/ttnn/operations/experimental/ccl/reduce_scatter_minimal_async/device/kernels/ring_reduce_scatter_minimal_async_writer.cpp
```diff
@@ -812,6 +812,13 @@ void kernel_main() {
                 fabric_direction_connection,
                 pkt_hdr_seminc,
                 tt::tt_fabric::NocUnicastAtomicIncCommandHeader{opposite_batch_ready_sem_noc_addr, 0});
+            // The mux client is opened in eager-staging mode: a packet only reaches the mux once a flush
+            // has completed the connection. Every other send in this kernel flushes right after its first
+            // packet; this one must as well. A worker that owns no pages sends nothing before reaching
+            // here, so without the flush this increment sits in the staging slot forever and the peer
+            // below never leaves its batch_ready_sem wait (observed with tt-triage: the zero-page
+            // writers parked on the wait below, everything else DONE, the mux waiting for them to close).
+            mf.flush();
```
scardozaTT, line 816: nit: the comment essay here is so extra, can be trimmed to a one-liner :)

scardozaTT (approved):

tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_minimal_reduce_scatter_async_bh.py
llongTT: Done in 3104cf6: kept only the zero-page case; the one-page and two-page splits are removed.

tests/ttnn/unit_tests/operations/ccl/blackhole_CI/box/nightly/test_minimal_reduce_scatter_async_bh.py
llongTT: Done in 3104cf6: only the zero-page case remains on the line test as well.

ttnn/cpp/ttnn/operations/experimental/ccl/reduce_scatter_minimal_async/device/kernels/ring_reduce_scatter_minimal_async_writer.cpp
llongTT: Trimmed to one line in 3104cf6.

PR #55100, opened 2026-09-01
[Feature] Add resource binding lookup for Metal 2.0

### LINKED ISSUE #55114: Metal 2.0 support for conditional resource binding tokens
## Background
In Metal 2.0, all `Program` resources (DataflowBuffer, Tensor parameters, semaphores, etc) are declared at the `ProgramSpec` level; individual kernels bind to these resources. The kernel's device code accesses the bound resource through a named binding token, emitted by the host into an auto-generated header. e.g.

```
DataflowBuffer dfb0(dfb::my_dfb0); // "my_dfb0" is the named kernel binding a the Program-scope resource
```

## Problem

Production kernels in TTNN ops are often designed to support a wide range of variations. From the host code, it is easy to create a range of different `Program`s with different resources. Consider the case where "my_dfb0" is used in some compile-time variants of a `ProgramFactory`, but not others. The following kernel code fails to compile if `cond` is false. The binding token is not emitted if the resource is not present, but the compiler _still evaluates_ the conditional.

```
if constexpr (cond) {
   DataflowBuffer dfb0(dfb::my_dfb0); 
}
```

The existing solution to this problem is to just use a preprocessor `ifdef` rather than a `constexpr` CTA:

```
#ifdef cond 
   DataflowBuffer dfb0(dfb::my_dfb0); 
#endif
```

This has the additional advantage of reducing the kernel binary size. However, for more complex conditions, the preprocessor-based solution can compromise kernel readability.

## Ask

The ops team would Metal 2.0 to support the legacy capability of using `if constexpr` with conditionally bound resources.

riverwuTT:
## PR Category

Feature

## Summary
For background of the "why" of this feature (and alternative approaches considered), see https://github.com/tenstorrent/tt-metal/issues/55114.

**Goal**: Allow Metal 2.0 kernels to _safely_ interact with conditionally-bound `Program` resources (`DataflowBuffer`s, `TensorAccessor`s, etc), and to do so using `if constexpr` expressions rather than preprocessor `ifdef`s. 

In the legacy Metal API, the following syntax was legal:

```
CircularBuffer my_cb(cb_index::cb0); // whether or not cb0 is declared in the host code

// Idea: The op correct sets my_cond_CTA to "false" if the cb0 resource was never allocated:
if constexpr my_cond_CTA {
   my_cb.push_back(my_data);
}

// But... if cb0 is unallocated, the following results in undefined behavior. 
// No checks (static or runtime) will catch it:
my_cb.push_back(my_other_data); //oops
``` 

This PR adds a device-side programatic lookup API for Metal 2.0 resource bindings. This enables the desired kernel syntax, while retaining safety checks.

**Solution**: The PR adds the following helper functions. Each one returns:
 - a pointer to the named resource token, if one exists, OR
 - nullptr

```cpp
namespace dfb {
  template<char* binding_name>
  constexpr const DFBBindingToken* get_token_if_present();
};

namespace tensor {
  template<char* binding_name>
  constexpr const auto * get_token_if_present()  /* -> TensorBindingToken<...> */;
};

namespace scratchpad {
  template<char* binding_name>
  constexpr const ScratchpadBindingToken* get_token_if_present();
};
```
------

## Usage example

Previous to this PR, the only way to handle conditionally bound resources in a Metal 2.0 kernel was via `#ifdef`s:
```cpp
// Normal is always present as a resource
DataflowBuffer dfb_normal(dfb::normal);

// Bias is an optional resource and is currently guarded by pre-processor directive
#ifdef BIAS_EXIST
DataflowBuffer dfb_bias(dfb::bias);
#endif

dfb_normal./* ... */;

#ifdef BIAS_EXIST
dfb_bias./* ... */;
#endif
```

After this PR, the kernel code can get the resource programmatically:
```cpp
// The binding's presence is the condition (matching CTA not required)
const DFBBindingToken* bias_token = dfb::get_token_if_present<"bias">();

std::optional<DataflowBuffer> dfb_bias;
if (bias_token != nullptr) {
    dfb_bias.emplace(*bias_token);
}

// ... later ...
if(dfb_bias) {
    dfb_bias->push_back(/* ... */);
}
```

### Notes for reviewers
Note that the original Metal 2.0 `#ifdef` solution reduces the kernel binary size versus programming retrieval. It should be preferredin any situation where the kernel binary size impacts performance.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:riverwu/optional_token_getter)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:riverwu/optional_token_getter)

### DIFF
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -4177,6 +4177,64 @@ void kernel_main() {
     EXPECT_NO_THROW(program.impl().compile(mesh_device_.get()));
 }
 
+TEST_F(ProgramSpecTestGen1, CPU_GetTokenIfPresentPrUsageExampleJITSmoke) {
+    // Kernel shape from the PR description: always-present `normal`, optional `bias` via
+    // get_token_if_present + std::optional. Compiles the same source with and without `bias` bound.
+    constexpr const char* kSource = R"(
+#include <optional>
+void kernel_main() {
+    DataflowBuffer dfb_normal(dfb::normal);
+
+    constexpr const DFBBindingToken* bias_token = dfb::get_token_if_present<"bias">();
+    std::optional<DataflowBuffer> dfb_bias;
+#pragma GCC diagnostic push
+// This is needed to avoid compiler warning the address will not/ always be null.
+#pragma GCC diagnostic ignored "-Waddress"
+    if constexpr (bias_token != nullptr) {
+        dfb_bias.emplace(*bias_token);
+    }
+
+    dfb_normal.push_back(1);
+    if constexpr (bias_token != nullptr) {
+        dfb_bias->push_back(1);
+    }
+#pragma GCC diagnostic pop
+}
+)";
+
+    auto compile_variant = [&](bool with_bias) {
+        NodeCoord node{0, 0};
+
+        auto dm_kernel = MakeMinimalGen1DMKernel("dm_kernel");
+        dm_kernel.source = KernelSpec::SourceCode{kSource};
+        auto compute_kernel = MakeMinimalGen1ComputeKernel("compute_kernel");
+
+        auto normal = MakeMinimalDFB("normal");
+        normal.data_format_metadata = tt::DataFormat::Float16_b;
+        dm_kernel.dfb_bindings.push_back(ProducerOf(DFBSpecName{"normal"}, "normal"));
+        compute_kernel.dfb_bindings.push_back(ConsumerOf(DFBSpecName{"normal"}, "normal"));
+
+        ProgramSpec spec;
+        spec.name = with_bias ? "pr_usage_bias_present" : "pr_usage_bias_absent";
+        spec.dataflow_buffers = {normal};
+        if (with_bias) {
+            auto bias = MakeMinimalDFB("bias");
+            bias.data_format_metadata = tt::DataFormat::Float16_b;
+            spec.dataflow_buffers.push_back(bias);
+            dm_kernel.dfb_bindings.push_back(ProducerOf(DFBSpecName{"bias"}, "bias"));
+            compute_kernel.dfb_bindings.push_back(ConsumerOf(DFBSpecName{"bias"}, "bias"));
+        }
+        spec.kernels = {dm_kernel, compute_kernel};
+        spec.work_units = {MakeMinimalWorkUnit("work_unit", node, {"dm_kernel", "compute_kernel"})};
+
+        Program program = MakeProgramFromSpec(*mesh_device_, spec);
+        EXPECT_NO_THROW(program.impl().compile(mesh_device_.get())) << "with_bias=" << with_bias;
+    };
+
+    compile_variant(/*with_bias=*/false);
+    compile_variant(/*with_bias=*/true);
+}
+
 TEST_F(ProgramSpecTestGen1, CPU_GetTokenIfPresentConstructsScratchpadJITSmoke) {
     NodeCoord node{0, 0};
 
--- a/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
+++ b/tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
@@ -395,22 +395,15 @@ TEST_F(ProgramSpecTestQuasar, CPU_InvalidLocalAccessorNameFails) {
                 ::testing::HasSubstr("DFB accessor_name '" + bad_name + "' must be a valid C++ identifier")))
             << "Expected rejection for name: '" << bad_name << "'";
     }
-}
 
-TEST_F(ProgramSpecTestQuasar, CPU_TooLongLocalAccessorNameFails) {
-    // An accessor_name longer than MAX_ACCESSOR_NAME_LENGTH cannot be passed to the kernel-side
-    // by-name binding lookup, so a valid-but-too-long identifier must fail at Program construction
-    // rather than as a kernel JIT static_assert.
+    // A valid-but-too-long identifier cannot be passed to the kernel-side by-name binding lookup,
+    // so it must fail at Program construction rather than as a kernel JIT static_assert.
     const std::string too_long(MAX_ACCESSOR_NAME_LENGTH + 1, 'a');
-    NodeCoord node{0, 0};
-
     ProgramSpec spec;
     spec.name = "test_program";
-
     auto kernel = MakeMinimalGen2DMKernel("kernel");
     auto dfb = MakeMinimalDFB("dfb");
     kernel.dfb_bindings.push_back(ProducerOf(DFBSpecName{"dfb"}, too_long));
-
     spec.kernels = {kernel};
     spec.dataflow_buffers = {dfb};
     spec.work_units = std::vector<WorkUnitSpec>{MakeMinimalWorkUnit("work_unit", node, {"kernel"})};
@@ -1521,8 +1514,9 @@ TEST_F(ProgramSpecTestQuasar, CPU_DuplicateScratchpadAccessorNameFails) {
 
 TEST_F(ProgramSpecTestQuasar, CPU_InvalidScratchpadAccessorNameFails) {
     // The accessor_name becomes a C++ identifier in the generated kernel_bindings header, so it must
-    // be a valid C++ identifier. (Mirrors InvalidLocalAccessorNameFails / the semaphore-accessor
-    // equivalent; here we just spot-check a couple of clearly-invalid names.)
+    // be a valid C++ identifier and fit in MAX_ACCESSOR_NAME_LENGTH. (Mirrors
+    // InvalidLocalAccessorNameFails / the semaphore-accessor equivalent; here we just spot-check a
+    // couple of clearly-invalid names plus the too-long case.)
     const std::vector<std::string> invalid_names = {
         "1bad",       // leading digit
         "has space",  // whitespace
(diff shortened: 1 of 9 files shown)


tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
```diff
@@ -3966,6 +3966,9 @@ void kernel_main() {
     auto noc_addr = accessor.get_noc_addr(0);
     (void)noc_addr;
 }
```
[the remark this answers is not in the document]
riverwuTT: I will add this

tt_metal/jit_build/genfiles.cpp
```diff
@@ -245,7 +260,7 @@ void write_kernel_bindings_generated_header(const string& out_dir, const JitBuil
             content << "}  // namespace sem\n";
         }
 
-        if (!ta_entries.empty() || !tensor_binding_sequence_entries.empty()) {
+        if (!ta_entries.empty()) {
```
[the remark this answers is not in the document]
riverwuTT: this is a drive by fix, but the logic in this PR is correct.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -234,6 +235,20 @@ void write_kernel_bindings_generated_header(const string& out_dir, const JitBuil
             for (const auto& [name, id] : dfb_entries) {
                 content << "constexpr DFBBindingToken " << name << "{" << id << "};\n";
             }
+
+            content << "template <::internal::TemplateString name>\n"
```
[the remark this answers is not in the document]
riverwuTT: yeah I could do this.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -234,6 +235,20 @@ void write_kernel_bindings_generated_header(const string& out_dir, const JitBuil
...
+                    << "constexpr const DFBBindingToken* get_binding_if_present() {\n";
```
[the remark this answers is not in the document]
riverwuTT: I guess I can see this.

tt_metal/jit_build/genfiles.cpp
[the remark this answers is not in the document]
riverwuTT: done

tt_metal/jit_build/genfiles.cpp
riverwuTT, line 1: Large diff in this file as the `get_binding_if_present()` need to be emitted unconditionally.

Which means the tensor/ dfb/ semaphore/ ... namespace is emitted unconditionally as well.

There was a fast path to emit the file as `// No bindings for this kernel.` when no binding is present, which was removed, causing a scope reduce, thus the giant diff.

tt_metal/jit_build/genfiles.cpp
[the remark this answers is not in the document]
riverwuTT: implemented, now this function is unconditionally emitted.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -123,6 +124,47 @@ bool write_named_ct_arg_map_header(const string& out_dir, const JitBuildSettings
     return true;
 }
 
+/**
+ * Emit get_binding_if_present() helper for a given binding type.
```
akerteszTT: The PR description says `get_token_if_present`; the code says `get_binding_if_present`. These need to match! 

I lean slightly towards `get_token_if_present`, as what's being gotten is the binding token, not the binding. Or the more verbose `get_binding_token_if_present` works.

tt_metal/jit_build/genfiles.cpp
```diff
@@ -8,6 +8,7 @@
 #include "data_format.hpp"
 #include <algorithm>
 #include <cstdint>
+#include <span>
```
akerteszTT: My Claude though this include was superfluous.

akerteszTT (commented):
Nice! This is a _very_ good solution to the problem of https://github.com/tenstorrent/tt-metal/issues/55114. I added some notes in the issue to further laud its benefits :).

That said, two things need fixing before this PR is ready to merge.

 1. None of the tests you've added is actually testing the kernel-side idiom. You need to make sure to have compiling kernel tests in the expected shape for each resource type. (There's an open PR that adds the missing `SemaphoreBindingToken` type -- if you merge ahead of it, it had better cause a noisy unit test failure when it tries to merge.)
 
 2. I don't think that the tensor version of the getter actually works (another reason why the explicit kernel code tests are so crucial.) Tragically, the tensor binding token is a templated type, so the type doesn't actually exist unless the token is emitted. According to my Claude, when the binding is absent, your getter will be forced to return `std::nullptr_t`, and dereferencing that is an ill-forced expression. So, we're back to square one with an angry JIT compiler upset about the `TensorAccessor` constructor call, and unwilling to be placated by the `if constexpr` guard. That will require some head scratching to solve. (You could ship the PR with the tensor part omitted, but conditional tensor bindings an important use case.) (See why I hate templates?)

Also, here's a nitty corner case (courtesy of Claude). Worth a look, but way less important the the two above.

3. I think there's a wonky corner case that breaks the helper functions. If the DFB helper happens to be used in a kernel where _no_ `DFBBindingToken` was emitted at all, nothing pulls in `dataflow_buffer.h`, so the `DFBBindingToken` type doesn't exist. Maybe this is unimportant as the kernel ought to include `dataflow_buffer.h` itself... but worth thinking about. A lot of stuff is auto-included via `dataflow_api.h`

tt_metal/jit_build/genfiles.cpp
riverwuTT: nice catch! should have been `get_token_if_present`

tt_metal/jit_build/genfiles.cpp
riverwuTT: fixed!

tt_metal/jit_build/genfiles.cpp
riverwuTT: fixed

riverwuTT:
Thanks for the careful review,
1. I have added device side usage example as tests! They caught bugs, thanks for pointing this out!
2. I have fixed that and added tests. Tensor binding needs specific null type to differentiate it from the normal binding which uses template parameter to pass meaningful offset information. 
3. The tensor header is unconditionally included as there's no way to express the null type of `get_token_if_present` without having the full definition of the binding token. The dfb/ scratchpad headers are conditionally included (preserving prior semantics). If they are not included, a forward declaration is omitted.

tests/tt_metal/tt_metal/api/metal2_host_api/test_program_spec.cpp
```diff
@@ -3977,6 +4016,438 @@ void kernel_main() {
...
+        KernelSpec::ScratchpadBinding{
+            .scratchpad_spec_name = ScratchpadSpecName{"scratch_0"}, .accessor_name = "scratch_a"},
+        KernelSpec::ScratchpadBinding{
+            .scratchpad_spec_name = ScratchpadSpecName{"scratch_1"}, .accessor_name = "scratch_b"},
+    };
+
+    spec.kernels[0].source = KernelSpec::SourceCode{R"(
+#include "api/tensor/local_tensor_accessor.h"
+
+void kernel_main() {
```
riverwuTT, line 4429: oof this test is wrong, meant to check if the function can be used in constant expression usage

riverwuTT: resolved.

akerteszTT (approved):
Nice! Thanks River. This is a great solution, should solve a lot of op porting challenges.

ruizhangTT (approved):

kstevensTT (approved):

jbaumanTT (approved):
ChatGPT suggests adding something like
```
template <typename T, bool Present>
class static_optional;

template <typename T>
class static_optional<T, true> {
public:
    template <typename... Args>
    constexpr explicit static_optional(Args&&... args)
        : value_(std::forward<Args>(args)...) {}

    static constexpr bool has_value = true;

    constexpr T& operator*() { return value_; }
    constexpr T* operator->() { return &value_; }

private:
    T value_;
};

template <typename T>
class static_optional<T, false> {
public:
    constexpr static_optional() = default;

    static constexpr bool has_value = false;
};
```

that people can use instead of std::optional to ensure that access will be optimized out at compile-time. It could then be used like

```
constexpr auto bias_token = dfb::get_token_if_present<"bias">();
constexpr bool has_bias = bias_token != nullptr;

static_optional<DataflowBuffer, has_bias> dfb_bias = [&] {
    if constexpr (has_bias) {
        return static_optional<DataflowBuffer, true>{*bias_token};
    } else {
        return static_optional<DataflowBuffer, false>{};
    }
}();

// later
if constexpr (decltype(dfb_bias)::has_value) {
    dfb_bias->push_back(/* ... */);
}
```
or you could potentially add a cleaner way of initializing it.

Though in practice we do some pretty heavy optimization, so I'd expect the compiler could constant-propagate the existence or like thereof through most things (though not non-inlined function calls).

jbaumanTT:
Or we could suggest using std::conditional_t<..., std::monostate>

riverwuTT:
Thanks for the suggestion @jbaumanTT , there were multiple options for retrieving optional argument, and one of them would be to express a null binding as a dedicated type which your proposal would be part of. That was ultimately not chosen due to readability.

Issue #53655, opened 2026-08-19
[Blackhole] Fix stale source-valid waits in custom MoE destination-reuse LLKs


jasondavies:
## Summary

This is a follow-up to the merged [PR #52256, **Fix Blackhole destination-reuse synchronization**](https://github.com/tenstorrent/tt-metal/pull/52256). That PR fixed the generic Blackhole destination-reuse paths after reproducing the race on hardware and confirming the corrected ordering in simulation.

The Blackhole DeepSeek and generalized MoE LLKs contain custom destination-reuse replay sequences that hand-code a `STALLWAIT` followed by `MOVD2A`/`MOVD2B`. These sequences wait for SFPU completion and source `DVALID`, but do not drain the preceding math pipeline:

```cpp
TTI_STALLWAIT(
    p_stall::STALL_MATH,
    p_stall::WAIT_SFPU | p_stall::SRCA_VLD);
```

This bypasses the corrected ordering introduced by [PR #52256](https://github.com/tenstorrent/tt-metal/pull/52256) in `move_d2a_fixed_face()` and `move_d2b_fixed_face()`, which wait on `MATH | SRCA_VLD` or `MATH | SRCB_VLD`.

On Blackhole, a preceding FPU instruction such as `SETRWC` can release and advance a source bank one or two cycles after a source-valid-only wait is evaluated. The wait can therefore observe stale state and allow the following `MOVD2A`/`MOVD2B` to proceed against the wrong bank state. This is the same ordering hazard addressed by [PR #52256](https://github.com/tenstorrent/tt-metal/pull/52256), but these custom MoE paths do not call the generic helpers.

No DeepSeek output failure has yet been isolated from this code. This report tracks a latent race identified by inspection after the generic Blackhole failure was reproduced and its corrected `MATH | DVALID` ordering was confirmed in simulation.

## Affected paths

The stale wait is present in four Blackhole headers:

1. Canonical DeepSeek MoE LLK:
   `tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_deepseek_moe_gate_eltwise_binary.h`
2. Canonical generalized MoE LLK:
   `tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_generalized_moe_gate_eltwise_binary.h`
3. Private copy compiled by the TTNN DeepSeek MoE operation:
[...]

PR #55107, opened 2026-09-01
[LLK][Bug fix] Drain the math pipeline in the Dest->Src bank waits (WH + BH); pipeline the dummy publications

### LINKED ISSUE #55106: [LLK] Dest->Src reuse handshake: MOVD2A/MOVD2B waits missing the FPU-pipeline drain at 33 sites (WH + BH)
## Summary

[PR #52256](https://github.com/tenstorrent/tt-metal/pull/52256) fixed the generic Blackhole destination-reuse handshake by adding a math-pipeline drain to the wait that gates `MOVD2A` / `MOVD2B`:

```cpp
// before
TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCA_VLD);
// after
TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
```

That fix was applied to exactly two functions in one file. The same bare wait is still present at **33 other sites across 16 files**, on both Wormhole and Blackhole, including the canonical `transpose_dest` LLK and several TTNN-private LLK copies compiled by production kernels.

This issue tracks the remaining sites. It deliberately excludes every site already covered by an open or merged PR (see *Out of scope* below), so it does not overlap [#53655](https://github.com/tenstorrent/tt-metal/issues/53655) / [PR #53690](https://github.com/tenstorrent/tt-metal/pull/53690) or [PR #54731](https://github.com/tenstorrent/tt-metal/pull/54731).

## Mechanism

`p_stall::SRCA_VLD` / `SRCB_VLD` are [`STALLWAIT`](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/STALLWAIT.md) conditions **C10** / **C11**:

> **C10** — keep waiting if `SrcA[MatrixUnit.SrcABank].AllowedClient != SrcClient::MatrixUnit`
> **C11** — keep waiting if `SrcB[MatrixUnit.SrcBBank].AllowedClient != SrcClient::MatrixUnit`

Both index `MatrixUnit.SrcABank` / `SrcBBank` live, at the Wait Gate. Those bank pointers are advanced in the *epilogue* of the preceding Matrix Unit instruction — the `if (FlipSrcA)` / `if (FlipSrcB)` block at the very end of [`ELWMUL`'s functional model](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/ELWMUL.md#functional-model), and equivalently for `SETRWC` with a `CLR_*` operand.

So with a bank-flipping Matrix Unit instruction still in flight, C10/C11 test the *pre-flip* bank — which the Matrix Unit still owns — are satisfied vacuously, and the wait is forgotten. By the time the following `MOVD2A` / `MOVD2B` executes, the flip has landed, and it writes the *post-flip* bank: one the unpacker still owns and may be filling.

`MOVD2A` and `MOVD2B` do not auto-wait for this. Both ISA pages say so explicitly and direct software to use `STALLWAIT` with exactly these condition codes ([`MOVD2A`](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/MOVD2A.md#instruction-scheduling), [`MOVD2B`](https://github.com/tenstorrent/tt-isa-documentation/blob/main/WormholeB0/TensixTile/TensixCoprocessor/MOVD2B.md#instruction-scheduling)). The ISA's own note on C10/C11 — *"rarely needed, as Matrix Unit (FPU) instructions which read from SrcA or SrcB automatically wait for this condition"* — is precisely why these Dest→Src **writes** are the exception that needs the explicit wait, and why the symptom is a silent value mismatch rather than a hang.

Adding `p_stall::MATH` selects condition **C7**:

> **C7** — keep waiting if the current thread has an instruction in any stage of the Matrix Unit (FPU) pipeline

which drains the pipe so C10/C11 observe the post-flip pointer. C7's documented requirement is that the block mask include **B6** (block new Matrix Unit instructions); every site here already passes `p_stall::STALL_MATH`, which *is* B6, so that precondition is already satisfied.

`STALLWAIT` stays latched while *any* selected condition says keep-waiting and is forgotten only once *all* are simultaneously met, so a single combined mask is correct and no extra instruction is needed. That matters for the sites recorded into the replay buffer (`generalized_moe_gate_eltwise_binary`, the `deepseek_moe_gate` variants), where the recorded instruction count must not change — it does not.

**Cost.** C7 is a real drain, and the ISA notes it "might wait longer than strictly necessary" when the Matrix Unit is shared across threads. This trades a small amount of math-pipeline overlap for correctness, which is the same trade #52256 already made on the two generic Blackhole paths.

`STALLWAIT` holds until all selected conditions are simultaneously true, so a single combined mask is correct and sufficient; no extra instruction is added. This matters for the sites recorded into the replay buffer (`generalized_moe_gate_eltwise_binary`, the `deepseek_moe_gate` variants) where the recorded instruction count must not change — it does not.

## Affected sites

`p_stall::SRC?_VLD` → `p_stall::MATH | p_stall::SRC?_VLD`, preserving any existing `WAIT_SFPU`.

**Canonical LLK — Wormhole**
| File | Sites |
|---|---|
| `tt_llk_wormhole_b0/common/inc/cmath_common.h` | `move_d2a_row_broadcast_fixed_face`, `wait_bank_valid<SrcA>`, `wait_bank_valid<SrcB>` |
| `tt_llk_wormhole_b0/llk_lib/llk_math_transpose_dest.h` | `_llk_math_transpose_dest_` |
| `tt_llk_wormhole_b0/llk_lib/experimental/llk_math_generalized_moe_gate_eltwise_binary.h` | RELOAD replay sequence |
| `tt_llk_wormhole_b0/llk_lib/experimental/llk_math_generalized_moe_gate_transpose_dest_single_face.h` | `copy4rows`, `step1_hi`, `step0`, `step1`, `step2` |

**Canonical LLK — Blackhole**
| File | Sites |
|---|---|
| `tt_llk_blackhole/common/inc/cmath_common.h` | `move_d2a_row_broadcast_fixed_face`, `wait_bank_valid<SrcA>`, `wait_bank_valid<SrcB>` |
| `tt_llk_blackhole/llk_lib/llk_math_transpose_dest.h` | `_llk_math_transpose_dest_` |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_generalized_moe_gate_transpose_dest_single_face.h` | `copy4rows`, `step1_hi`, `step0`, `step1`, `step2` |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_face_compressed_mm.h` | split-partial merge |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_rmsnorm_bcast_scalar_dest_reuse.h` | `rmsnorm_bcast_scalar_reuse_dest_as_src` |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_sdpa_bcast_col_srcb_reuse.h` | preamble |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_sdpa_bcast_col_srca_srcb_reuse.h` | preamble |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_sdpa_custom_mm_reuse_dest_srcb.h` | main loop |
| `tt_llk_blackhole/llk_lib/experimental/llk_math_custom_mm_reuse_dest_srcb.h` | main loop |

**TTNN-private LLK copies** under `ttnn/cpp/ttnn/operations/experimental/deepseek/moe/deepseek_moe_gate/device/kernel_includes/tt_llk/`
| File | Sites |
|---|---|
| `tt_llk_wormhole_b0/llk_lib/llk_math_deepseek_moe_gate_eltwise_binary.h` | `DEST_TO_SRCA`, `DEST_TO_SRCB` |
| `tt_llk_wormhole_b0/llk_lib/llk_math_deepseek_moe_gate_transpose_dest_single_face.h` | `step0`, `step1`, `step2` |
| `tt_llk_blackhole/llk_lib/llk_math_deepseek_moe_gate_transpose_dest_single_face.h` | `step0`, `step1`, `step2` |

These private copies are compiled by the TTNN DeepSeek MoE gate operation, so fixing only the canonical headers would leave the production path defective — the same trap called out in #53655. PR #53690 covers the Blackhole `deepseek_moe_gate_eltwise_binary` copy; the Wormhole copy and both `transpose_dest_single_face` copies are not covered by any PR.

Two of the sites (`move_d2a_row_broadcast_fixed_face`, `wait_bank_valid<>`) currently have no callers in the repository. They are included because they are pre-broken templates: the defect ships the moment something calls them.

## Out of scope

Excluded because another PR already owns them:

- `tt_llk_wormhole_b0/common/inc/cmath_common.h` `move_d2a_fixed_face` / `move_d2b_fixed_face` — open PR #54731
- `tt_llk_blackhole/.../llk_math_deepseek_moe_gate_eltwise_binary.h`, `tt_llk_blackhole/.../llk_math_generalized_moe_gate_eltwise_binary.h`, and the TTNN-private Blackhole `deepseek_moe_gate_eltwise_binary` copy — open PR #53690 (issue #53655)
- `tt_llk_blackhole/common/inc/cmath_common.h` `move_d2a_fixed_face` / `move_d2b_fixed_face` — already fixed by merged PR #52256

Excluded because they are a different hazard, not this one:

- `llk_math_compressed_custom_mm.h` — the wait gates a `SETRWC` and a matmul replay, not a Dest→Src move
- `llk_math_sdpa_custom_mm.h` — the wait gates `MOVB2D`, which *reads* SrcB rather than writing it
- `ckernel_sfpu_topk_xl.h` — `STALL_CFG` waits around SFPU configuration
- All of Quasar — `move_d2a_fixed_face` / `move_d2b_fixed_face` / `llk_math_transpose_dest.h` / `llk_math_reduce.h` / `llk_math_unary_broadcast.h` carry the same bare waits, but Quasar's `STALLWAIT` has a different operand encoding and its bank handshake has not been assessed here. This needs a Quasar owner and a separate issue.

## Verification

The wait is a mask constant, so the change is behaviour-preserving except for the added drain, and it cannot alter recorded replay lengths.

The defect class is latent by nature — it produces a wrong value only when a bank flip is in flight at the wait, which no current test forces — so this is an inspection-and-ISA-grounded fix in the same mould as #53655/#53690, not a fail-without/pass-with reproduction. Blackhole regression coverage is available on this box and will be run. **The Wormhole sites cannot be executed here and need a Wormhole board before merge.**

## Follow-up

The TTNN-private LLK copies mean a fix to a canonical header silently misses production callers. @jasondavies raised the same point on #53655. Consolidating those copies onto the canonical implementation would prevent the next instance of this and is worth a separate issue.

---

## Scope widened: the same handshake has an unpack side

The issue as written covered the **math side** — the `MOVD2A`/`MOVD2B` wait missing its FPU-pipeline drain. That remains the defect, and it is what the fix PR is for.

The same handshake has an **unpack side** worth tightening, but on investigation it is **not a second defect**, and the PR now says so:

A dummy publication (`UNPACR_NOP` doing `ZEROSRC`/`SET_DVALID`) always clears `Unpackers[i].SrcBank`, but with `Stall_Clr_Cntrl=0` it *waits* on `MatrixUnit.Src?Bank` — so the wait establishes nothing about the bank it modifies. **It is safe today anyway:** with two banks and a one-for-one fill/consume cycle the Matrix Unit pointer only advances when it hands a bank back, landing it on the bank the unpacker holds. Reaching the harmful arrangement would need the unpacker two hand-overs ahead, which two banks prevent. The old form therefore **over-waits, not under-waits** — it costs the overlap the second bank exists to give, and does not corrupt.

Setting `Stall_Clr_Cntrl=1` makes the instruction wait on exactly the bank it clears, matching `dest_reuse_dummy_unpack()` from #52256, whose own comment gives the throughput rationale. Sites (all Blackhole, 7 publications / 3 files): `llk_unpack_common.h` `_llk_unpack_set_srcb_dummy_valid_` (2 — its Wormhole twin already waits correctly), `llk_unpack_A_sdpa.h` (3), `llk_unpack_A_rmsnorm.h` (2). `llk_unpack_mul_reduce_scalar.h` (WH+BH) is #54683's.

**Open question that would change this verdict:** if any path can hand a bank over **twice without an intervening read**, the pointers *can* reach the harmful arrangement and the unpack side becomes a real correctness bug. I have not found such a path, but I have not exhaustively ruled one out either. It is also directly measurable — the over-wait should show as throughput on `transpose_dest`.

**Investigated, not a defect:** `llk_math_hadamard.h`'s `MOVD2B` has a `MATH`-only stall and no `SRCB_VLD`, but nothing flips SrcB before it (the `CLR_AB` releases come after, at the next matmul block) and its `MATH` drain guards a different, documented hazard.

amahmudTT:
### PR Category

Bug fix

### Summary

The `Dest`→`Src` reuse handshake has **two sides**, and fixing one does not fix the other. This PR fixes **both**, at every site not already owned by another PR.

Fixes #55106

### (a) Math side — the FPU-pipeline drain (33 sites, 16 files)

`MOVD2A`/`MOVD2B` *write* a Src bank, so the Wait Gate's automatic `AllowedClient` wait — which covers only Matrix Unit instructions that *read* Src — does not apply. `MOVD2A.md`/`MOVD2B.md` say so and direct software to `STALLWAIT`.

But a `STALLWAIT` selecting only the bank-valid condition is still wrong: it indexes `MatrixUnit.Src?Bank` **live**, and that pointer advances in the *epilogue* of the preceding Matrix Unit instruction (`ELWMUL.md`'s `if (FlipSrcA)` block; likewise `SETRWC` with `CLR_*`). With a flipping op in flight it tests the **pre-flip** bank, passes vacuously, and the move then writes the **post-flip** bank the unpacker owns. Selecting `p_stall::MATH` drains the pipe first.

Symptom is a **silent wrong value, never a hang** — every Src *reader* auto-waits — so "no hangs observed" is not evidence of safety.

### (b) Unpack side — the pipelined bank wait (7 publications, 3 Blackhole files)

**This half is a throughput change, not a correctness fix.** Please read it as such — and note the title says "pipeline", not "tighten": the bit *relaxes* the wait.

The publications changed here are Blackhole's combined form — one `UNPACR_NOP` carrying both `Set_Dvalid` and `Unpack_Pop=UNP_ZEROSRC` — so each does clear `Unpackers[i].SrcBank` (that is the `ZEROSRC` half; a *bare* `SET_DVALID` clears nothing and performs no wait of its own). With `Stall_Clr_Cntrl=0` the instruction *waits* on `MatrixUnit.Src?Bank`. That wait names a bank the instruction does not modify, and it is **stronger than necessary**: under the bank-pointer lockstep invariant `MatrixUnit.Src?Bank` is back with the unpackers only when *no* bank is outstanding, so the default form implies the own-bank condition and then some.

**It is nonetheless safe today.** With two banks and a one-for-one fill/consume cycle the pointers cannot reach the harmful arrangement: the Matrix Unit pointer only advances at the instant it hands a bank back, which lands it on the bank the unpacker is holding. Getting the unpacker onto a busy bank while the Matrix pointer sits on a free one would need the unpacker to run two hand-overs ahead, which two banks prevent. So the old form **over-waits rather than under-waits** — it loses the overlap the second bank exists to provide, but does not corrupt.

Setting `Stall_Clr_Cntrl=1` makes the instruction wait on exactly the bank it clears, matching Blackhole's `dest_reuse_dummy_unpack()` from #52256 — whose own comment gives the same rationale: *"so unpack can prepare the next bank while math consumes the current bank."* The bit changes **only the wait**; the bank cleared is `UnpackBank` either way.

#52256's commit message is explicit that this is a mode, not a fix, and it is worth quoting in full because it settles how (a) and (b) relate:

> It also sets `Stall_Clr_Cntrl=1` on destination-reuse dummy unpacks. This is the preferred Blackhole operating mode: it gates the dummy operation using the unpacker-selected bank and allows unpack to prepare the next bank while math consumes the current one. **It is complementary to, rather than a replacement for, the explicit math-pipeline drain.**

**One constraint this PR does not hit, but that any future site must.** The bit is only valid when the publication clears a *single* bank. With a both-banks clear (`Bank_Clr_Ctrl=1`) the own-bank wait covers only the bank being prepared, so clearing both can overwrite one the Matrix Unit still owns — a both-banks clear is correct **only** with the default drained wait. All 7 publications changed here have `Bank_Clr_Ctrl=0`, and the three in-tree both-banks sites (`llk_unpack_AB_compressed_custom_mm.h`, `llk_unpack_AB_custom_mm.h`, `llk_unpack_AB_face_compressed_mm.h`) are untouched and correctly keep the default wait. #55123 adds a checker flag for the unsafe pairing so this cannot regress silently.

| File | Publications |
|---|---|
| `llk_unpack_common.h` `_llk_unpack_set_srcb_dummy_valid_` | 2 — its Wormhole twin already waits correctly, via a preceding `STALLWAIT` on `SRCA_CLR｜SRCB_CLR` |
| `llk_unpack_A_sdpa.h` | 3, across 2 functions |
| `llk_unpack_A_rmsnorm.h` | 2 `static constexpr` MOP constants |

**Reviewer risk to weigh.** The change removes an incidental coupling: the old wait forced the unpacker to wait for the FPU, so unpack could not run ahead. Letting it run ahead is the point, but anything that silently relied on that coupling would now be exposed — and the failure would surface downstream, not at the changed instruction. **`llk_unpack_A_sdpa.h` and `llk_unpack_A_rmsnorm.h` have no test coverage in `unit_tests_llk`;** only `llk_unpack_common.h` is exercised, by the three transpose tests. If you would rather this half landed separately, say so and I will split it.

### (c) The join

Fixing (a) lengthens the math-thread wait and shifts inter-thread timing; (b) lets the unpack side run ahead again. They touch the same handshake from opposite ends, which is why they are reviewed together — but (a) is the bug fix and (b) is the throughput mode. (b) is never a substitute for (a).

### Scope — everything owned elsewhere is excluded

| Site | Owner |
|---|---|
| WH `cmath_common.h` `move_d2a/d2b_fixed_face` | open #54731 |
| BH MoE eltwise-binary + the ttnn BH deepseek copy | open #53690 |
| BH `cmath_common.h` `move_d2a/d2b_fixed_face` | merged #52256 |
| WH+BH `llk_unpack_mul_reduce_scalar.h` | open #54683 |

Also excluded as a different hazard: `compressed_custom_mm` (the wait gates a `SETRWC`/matmul replay), `sdpa_custom_mm` (`MOVB2D` *reads* SrcB), `topk_xl` (`STALL_CFG` around SFPU config), and all of Quasar, whose `STALLWAIT` encoding differs and whose bank model needs a Quasar owner.

**Investigated and deliberately not changed:** `llk_math_hadamard.h`'s `MOVD2B` has a `MATH`-only stall and no `SRCB_VLD`, but nothing flips SrcB before it (the `CLR_AB` releases come *after*, at the next matmul block) and its `MATH` drain guards a different, documented hazard — `MVMUL` writes dst on the FPU pipe while `MOVD2B` reads it on the MOV pipe.

Two of the 33 math-side sites (`move_d2a_row_broadcast_fixed_face`, `wait_bank_valid<>`) have no callers today; they are pre-broken templates, fixed so the defect does not ship the moment something calls them.

### Testing

Blackhole p100a, `unit_tests_llk` over transpose / MoE-gate / rmsnorm / sdpa / custom-mm / binary / reduce:

| | Result |
|---|---|
| branch without the producer commit | 100 ran, 86 passed, 14 skipped, **0 failed** |
| with it | identical |

The math-side change is a mask constant and cannot alter recorded replay lengths. The unpack-side change **is** behavioural — it alters which bank the instruction waits on — which is why the A/B baseline was run.

**The Wormhole sites were not executed** — no Wormhole silicon on this box. They need a WH run before merge.

Like #53655/#53690 this is an inspection- and ISA-grounded fix for a latent class: the defect only produces a wrong value when a bank flip is in flight at the wait, which no current test forces, so there is no fail-without/pass-with reproducer to add.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:amahmud/llk-dest-reuse-math-drain-sweep)

### DIFF
--- a/tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/common/inc/cmath_common.h
@@ -117,8 +117,7 @@ inline void _configure_preserve_zero_flag_state_()
 inline void _configure_copy_zero_flag_state_(const std::uint32_t src_dst_format)
 {
     const std::uint32_t fmt = src_dst_format & 0x1F;
-    const bool flush_fp8 =
-        (fmt == static_cast<std::uint32_t>(DataFormat::Fp8_e4m3)) || (fmt == static_cast<std::uint32_t>(DataFormat::Lf8));
+    const bool flush_fp8    = (fmt == static_cast<std::uint32_t>(DataFormat::Fp8_e4m3)) || (fmt == static_cast<std::uint32_t>(DataFormat::Lf8));
     _configure_src_zero_flag_(!flush_fp8);
 }
 
@@ -178,7 +177,7 @@ inline void move_d2a_row_broadcast_fixed_face(const std::uint8_t addrmod)
 {
     // MOVD2A does not auto-wait for SrcA[MatrixUnit.SrcABank].AllowedClient == MatrixUnit, so gate on SRCA_VLD
     // before the row moves (mirrors move_d2b_fixed_face's SRCB_VLD wait). See llk_math_transpose_dest.h. tt-llk#1664.
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCA_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
     // // Seems to make things 200 clocks slower. Really shouldn't though.
     TTI_MOVD2A(0, p_mova2d::MATH_HALO_ROWS + 0, addrmod, p_movd2a::MOV_1_ROW, 0);
     TTI_MOVD2A(0, p_mova2d::MATH_HALO_ROWS + 1, addrmod, p_movd2a::MOV_1_ROW, 0);
@@ -209,11 +208,11 @@ inline void wait_bank_valid()
 {
     if constexpr (SrcReg == Srcs::SrcA)
     {
-        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCA_VLD);
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD);
     }
     else
     {
-        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCB_VLD);
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCB_VLD);
     }
 }
 
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_custom_mm_reuse_dest_srcb.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_custom_mm_reuse_dest_srcb.h
@@ -139,7 +139,7 @@ inline void _llk_math_custom_mm_reuse_dest_srcb_(
         "custom_mm_reuse_dest_srcb: in0 tile height must be 1, 2, 4 or 8");
 
     const std::uint32_t dest_buffer_base = get_dest_buffer_base();
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCB_VLD | p_stall::WAIT_SFPU);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
 
     for (std::uint32_t i = 0; i < kt_dim; i++)
     {
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_face_compressed_mm.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_face_compressed_mm.h
@@ -460,7 +460,7 @@ inline void _llk_math_face_compressed_mm_(
         // Merge each face's split partial into its accumulation.
         constexpr std::int16_t partial = _llk_math_face_compressed_mm_split_acc_partial_rows_;
 
-        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::SRCA_VLD | p_stall::SRCB_VLD); // wait for both operands
+        TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::MATH | p_stall::SRCA_VLD | p_stall::SRCB_VLD); // wait for both operands
         // Move both partials into SrcB, face 0 at row 0 and face 1 at row 8.
         // ADDR_MOD_0 advances dest by one face between the two, and ADDR_MOD_2 resets it afterwards.
         TTI_MOVD2B(p_mov::DEST_NORM, 0, ADDR_MOD_0, p_movd2b::MOV_4_ROWS, partial);
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_generalized_moe_gate_transpose_dest_single_face.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_generalized_moe_gate_transpose_dest_single_face.h
@@ -218,7 +218,7 @@ inline void _llk_math_generalized_moe_gate_copy4rows_()
 {
     static_assert(!(is_32bit || is_fp32_dest_acc_en), "32-bit / fp32 dest accum not supported");
     math::reset_counters(p_setrwc::SET_ABD_F);
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
     ckernel_template::run();
 }
 
@@ -234,7 +234,7 @@ inline void _llk_math_generalized_moe_gate_transpose_dest_single_face_step1_hi_(
 {
     static_assert(!(is_32bit || is_fp32_dest_acc_en), "32-bit and fp32 dest accum enable are not supported for single face transpose");
     math::reset_counters(p_setrwc::SET_ABD_F);
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
     ckernel_template::run();
 }
 
@@ -245,7 +245,7 @@ inline void _llk_math_generalized_moe_gate_transpose_dest_single_face_step0_()
     math::reset_counters(p_setrwc::SET_ABD_F);
 
     // Wait for SFPU and SrcB to be available
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
 
     // Run the 16-bit single-face transpose MOP
     ckernel_template::run();
@@ -262,7 +262,7 @@ inline void _llk_math_generalized_moe_gate_transpose_dest_single_face_step1_()
     math::reset_counters(p_setrwc::SET_ABD_F);
 
     // Wait for SFPU and SrcB to be available
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
 
     // Run the 16-bit single-face transpose MOP
     ckernel_template::run();
@@ -279,7 +279,7 @@ inline void _llk_math_generalized_moe_gate_transpose_dest_single_face_step2_()
     math::reset_counters(p_setrwc::SET_ABD_F);
 
     // Wait for SFPU and SrcB to be available
-    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::SRCB_VLD);
+    TTI_STALLWAIT(p_stall::STALL_MATH, p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD);
 
     ckernel_template::run();
 
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_rmsnorm_bcast_scalar_dest_reuse.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_rmsnorm_bcast_scalar_dest_reuse.h
@@ -73,8 +73,8 @@ inline void rmsnorm_bcast_scalar_reuse_dest_as_src()
 {
     TTI_STALLWAIT(
         p_stall::STALL_MATH,
-        p_stall::WAIT_SFPU | p_stall::SRCB_VLD); // MOVD2B for a whole face assumes unpacker will set a dummy
-                                                 // data_valid, so we want to wait on that
+        p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCB_VLD); // MOVD2B for a whole face assumes unpacker will set a dummy
+                                                                 // data_valid, so we want to wait on that
     TTI_MOVD2B(0, p_movd2b::SRC_ZERO_OFFSET + 0, ADDR_MOD_1, p_movd2b::MOV_1_ROW, 0);
 }
 
--- a/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_sdpa_bcast_col_srca_srcb_reuse.h
+++ b/tt_metal/tt-llk/tt_llk_blackhole/llk_lib/experimental/llk_math_sdpa_bcast_col_srca_srcb_reuse.h
@@ -69,8 +69,8 @@ inline void _llk_math_sdpa_bcast_col_srca_srcb_reuse_preamble_(std::uint32_t isr
 {
     TTI_STALLWAIT(
         p_stall::STALL_MATH,
-        p_stall::WAIT_SFPU | p_stall::SRCA_VLD | p_stall::SRCB_VLD); // MOVD2B for a whole face assumes unpacker will set a dummy
-                                                                     // data_valid, so we want to wait on that
+        p_stall::WAIT_SFPU | p_stall::MATH | p_stall::SRCA_VLD | p_stall::SRCB_VLD); // MOVD2B for a whole face assumes unpacker will set a dummy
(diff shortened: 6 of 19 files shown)


bbradelTT (approved):

amahmudTT:
## Wormhole check

Ran on an **n150** (`2e90a708fa7` vs its base `df83d62ade0`). **No WH regression** — every WH test reaching the changed code passes identically on both arms, 0 failures:

| layer | control | branch |
|---|---|---|
| tt-llk (transpose_dest, generalized MoE gate, deepseek MoE gate) | 144 variants pass, 3/3 runs | same |
| `unit_tests_llk` (transpose + cumsum — the only 2 of 486 reaching a changed WH header) | 5 pass, 3/3 runs | same |
| ttnn (layernorm, layernorm/rmsnorm sharded, group_norm, transpose, MoE gate) | 1954 passed, 579 skipped | same |

A/B verified by disassembly, not by test result: the compiled kernels show the expected +128 (`p_stall::MATH`) delta — `transpose_wh_dest`, `layernorm_welford`, `layernorm_sharded_welford` all `64,19456` → `64,19584`; MoE gate `64,17408`/`64,18432` → `64,17536`/`64,18560`.

Three candidate tests turned out to carry **no** WH signal and are excluded from the counts above: `test_top32_rm` (5 skipped, own format guards), `test_topk_large_indices` (124 skipped, Blackhole-only), `test_matmul` (no matmul kernel variant contains a `64,*` stallwait, so `transpose_tile` never takes the transpose-dest path there).

**This is safety evidence only, not proof the fix works** — per the PR description the defect needs a bank flip in flight at the wait, which no current test forces, so a passing A/B cannot separate a correct fix from an inert one. It closes the "Wormhole sites were not executed" item, no more.

nvelickovicTT (approved):

bbradelTT (commented):
Does this work properly on Quasar?

If not, is there an issue to fix it in the specific way that Quasar needs it to be fixed, which may be slightly different from how it's being fixed here?

amahmudTT:
> Does this work properly on Quasar?
> 
> If not, is there an issue to fix it in the specific way that Quasar needs it to be fixed, which may be slightly different from how it's being fixed here?

One thing I must say here the stallwaits do not fix an existing problem but a potential problem, without the modified stallwait if the functions were called after a bank flipping instruction then there would be a race. Currently they are not called after a bank flipping instruction. I have seen cases where an unused funciton with such bugs were left alone and then later used and issues appeared. This might have at most 1.9% slowdown (one ci run showed) FYI. **The perf regression went away with rerun though.** 

Regarding quasar, @rtawfik01  does this type of problem exist there ?

arginugaTT (approved):

PR #55112, opened 2026-09-01
Clamp sharded output grids to the shards they hold

bmalesevicTT:
### PR Category

Bug fix

### Summary

Relates to [tt-mlir#9189](https://github.com/tenstorrent/tt-mlir/issues/9189).

`TensorSpec::height_sharded` / `width_sharded` / `block_sharded` divide the tensor over
the given grid and round the shard shape up to tile alignment, but keep the grid as-is —
so when rounding shrinks the shard count, the grid keeps cores that hold no shard.
Validation only rejects too *few* cores, and reductions inherit the input's grid whenever
the output config has a sharded layout but no shard spec, so e.g. a `1x512x2048` input
block-sharded over 88 cores reduces to a `1x512x1` output with 8 shards still claiming all
88 cores (the vocab-loss sum: 1 shard on 109 cores).

Whether an over-provisioned grid is benign depends on how the consumer reads the tensor.
Ops that address by the **page map** (e.g. the universal reduce reader) never touch a
phantom core, so for them it is only wasted L1. But ops that derive work and output
placement from the **shard-spec grid** treat grid membership as data — and
`sharded_to_interleaved` is one of them: given a `BLOCK_SHARDED` input whose grid is wider
than the data, its writer runs on the phantom cores and **overruns the interleaved output,
silently corrupting adjacent DRAM**. This reproduces in pure ttnn, no tt-mlir involved: a
`(512x32)` tensor block-sharded on an 8x11 grid, relayouted to interleaved DRAM, comes back
14336/16384 elements wrong, while the correctly-provisioned 8x1 grid is exact. In
Llama 3.2 1B forward at `optimization-level=2` on Blackhole this is exactly what happened:
an RMSNorm stat's over-provisioned block-sharded grid made its `to_memory_config` spill
overrun and corrupt the adjacent layer-14 gate weight (loss 27.44 vs 3.496). Loss is
restored by this change alone; 104 of the plan's 819 ops change layout, all matching this
pattern.

This PR removes the trigger by never over-provisioning the grid. The op-level gap it
exploits — `sharded_to_interleaved` lacks the `BLOCK_SHARDED` "grid > data" clamp that its
HEIGHT/WIDTH paths already have — is a separate, complementary tt-metal fix.

The builders now compute the aligned shard shape first and trim the grid to the shards it
yields: first N cores in assignment order for height/width (via `select_from_corerangeset`,
preserving orientation and range order), the leading rows×cols rectangle for block
(`GRID_2D`). Shard shapes are bit-identical to before; only the grid changes. The reshape
over-pad warning now measures against the trimmed grid. The docstrings already promised
"just 1 shard per core"; this makes that true.

### Notes for reviewers

- Behaviour change: reducing to a single tile-row with `keepdim=False` now returns an
  interleaved output for height sharding (previously 88 cores for one shard); block
  sharding trims to 1 core instead.
- Reshape (`recompute_shard_spec_for_output`, generic + quasar) previously got its seeded
  grid back unchanged; outputs needing fewer shards than input cores now land on a trimmed
  grid. The config is only consumed as a data-movement target, but a consumer assuming
  output grid == input grid may now reshard — a look from someone who knows the
  sharded-reshape consumers would be welcome.
- Out of scope (not addressed here): the ND path (`sharded_across_dims*`, direct
  `sharded()`, reduce's `ND_SHARDED` branch) can still keep idle cores; validation
  rejecting hand-built over-claimed specs; intra-shard padding of a ragged last shard.

### Testing

- New pytest asserts `cores <= shards` on reduce output specs (2 geometries × 3 layouts ×
  keepdim): 10 passed / 2 skipped with the fix (interleaved fallback cells), 10 failed / 2
  passed without.
- New `ShardedBuilderGridTrimTests` (`test_tensor_nd_sharding.cpp`) pin the builder
  contract device-free: shard shapes, trim placement per orientation, no-trim early
  returns, legacy/ND grid agreement, cores == shards.
- `test_reshape.py` + `test_reduction.py`: 745 passed, 0 failed.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:bmalesevic/clamp-sharded-grid-to-shards)

### DIFF
--- a/tt_metal/impl/tensor/spec/tensor_spec.cpp
+++ b/tt_metal/impl/tensor/spec/tensor_spec.cpp
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <tt-metalium/tensor/spec/tensor_spec.hpp>
+#include <tt-metalium/core_coord.hpp>
 #include <tt-metalium/tensor/tensor_types.hpp>
 #include <tt-metalium/experimental/per_core_allocation/memory_config.hpp>
 #include <tt-metalium/tensor/spec/memory_config/memory_config.hpp>
@@ -144,6 +145,35 @@ void validate_dtype_and_layout(DataType dtype, Layout layout) {
     supported_layout();
 }
 
+uint32_t aligned_extent(const Alignment& alignment, int dim_from_end, uint32_t extent) {
+    return static_cast<int>(alignment.size()) >= dim_from_end ? round_up(extent, alignment[-dim_from_end]) : extent;
+}
+
+uint32_t shards_needed(uint32_t total_extent, uint32_t shard_extent) {
+    return shard_extent == 0 ? 0 : div_up(total_extent, shard_extent);
+}
+
+CoreRangeSet first_n_cores(CoreRangeSet grid, uint32_t count, ShardOrientation orientation) {
+    if (count == 0 || count >= grid.num_cores()) {
+        return grid;
+    }
+    return CoreRangeSet(corerange_to_cores(grid, count, orientation == ShardOrientation::ROW_MAJOR))
+        .merge_ranges();
+}
+
+CoreRange leading_block(CoreRange grid, uint32_t rows, uint32_t cols) {
+    const auto size = grid.grid_size();
+    if (rows == 0 || cols == 0) {
+        return grid;
+    }
+    rows = rows < size.y ? rows : size.y;
+    cols = cols < size.x ? cols : size.x;
+    if (rows == size.y && cols == size.x) {
+        return grid;
+    }
+    return CoreRange(grid.start_coord, CoreCoord(grid.start_coord.x + cols - 1, grid.start_coord.y + rows - 1));
+}
+
 }  // namespace CMAKE_UNIQUE_NAMESPACE
 }  // namespace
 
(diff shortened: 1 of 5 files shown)


tt_metal/impl/tensor/spec/tensor_spec.cpp
```diff
@@ -144,6 +145,34 @@ void validate_dtype_and_layout(DataType dtype, Layout layout) {
...
+
+uint32_t shards_needed(uint32_t total_extent, uint32_t shard_extent) {
+    return shard_extent == 0 ? 0 : div_up(total_extent, shard_extent);
+}
+
+CoreRangeSet first_n_cores(CoreRangeSet grid, uint32_t count, ShardOrientation orientation) {
+    if (count == 0 || count >= grid.num_cores()) {
+        return grid;
+    }
+    return select_from_corerangeset(grid, 0, count - 1, orientation == ShardOrientation::ROW_MAJOR);
```
bbradelTT, line 160: What are the trade offs for select_from_corerangeset vs CoreRangeSet(span...)? And would there be benefit in dealing with full rectangles differently?

bbradelTT (commented):
Is the following going to be addressed?

Stale num_cores in reshape's over-pad warning (both reshape.cpp files, flagged inline): the diagnostic still divides by the input's pre-trim core count even though output_mem_config's grid is now trimmed by this PR's own fix — the warning's math didn't get updated for the new grid-shrinking behaviour it's supposed to be observing.

tt_metal/impl/tensor/spec/tensor_spec.cpp
bmalesevicTT: In practice I don't think there's a real difference - the grids we build are single rectangles or generated row-wise, and for those merging keeps the same enumeration order anyway, so same cores, same shard placement, just packed differently. It only matters if a multi-range grid with unusual range order reaches this code - Copilot suggested that can happen and I couldn't rule it out, so I went with the version that can't reorder shards. The only cost is N single-core ranges in the spec instead of a couple of rectangles. Could also merge when the grid is a single rectangle (order is preserved there) and use select only for multi-range grids - no strong opinion.

bmalesevicTT:
> Is the following going to be addressed?
> 
> Stale num_cores in reshape's over-pad warning (both reshape.cpp files, flagged inline): the diagnostic still divides by the input's pre-trim core count even though output_mem_config's grid is now trimmed by this PR's own fix — the warning's math didn't get updated for the new grid-shrinking behaviour it's supposed to be observing.

Yes, right - fixed in 688b9d9e0c3, both reshape copies. The warning now divides by the trimmed output grid's core count instead of the input's pre-trim count. That changes what it flags, on purpose: the old ratio measured the over-allocation this PR removes, the new one only fires when the data genuinely under-fills its tile slot - and since the only remaining remedy there is INTERLEAVED, the message now says just that. Hopefully, it makes sense for the intended use-case.

~Btw on the over-allocation itself: inside tt-metal it was intentional and harmless - data placement comes from the page map, not from grid membership, so the extra cores are at worst wasted L1 and compute that nothing ever reads. **It only bites consumers like tt-mlir that plan from the grid**. Anyhow, if some case genuinely benefits from keeping the output on the input grid (core locality), I'd rather reshape implement that explicitly than silently carry a grid that overstates where the data is, as was the case here.~
Btw I have to correct myself on the over-allocation - it's not harmless inside tt-metal either. Some ops place data from the page map (there the extra cores are just wasted L1), but others walk the shard-spec grid directly. `sharded_to_interleaved` is one of those - on an over-provisioned block-sharded grid it runs writers on the phantom cores and their output page offsets run past the buffer, clobbering whatever DRAM sits after it. Repros in pure ttnn (block-sharded L1 -> interleaved DRAM via to_memory_config), no tt-mlir. So this PR isn't just tidying planning for grid-based consumers, it removes a real silent-corruption trigger. The factory should clamp BLOCK_SHARDED too (it already does height/width) - separate follow-up.

Rest stands: if a case genuinely benefits from keeping the output on the input grid, I'd rather reshape do that explicitly than carry a grid that overstates where the data is.

bbradelTT (approved):

Issue #49781, opened 2026-07-13
Micro-benchmark fast dispatch on quasar


jbaumanTT:
We'll need to micro-benchmark fast dispatch on quasar to see where the inefficiencies are.

sidnadTT:
Here are some initial benchmarks, after iDMA vs NoC on Quasar investigation summarized summarized here: https://github.com/tenstorrent/tt-metal/issues/49293.

We started with the Quasar FD path using iDMA for local L1 --> L1 prefetcher to dispatcher copies, with L1 hashing disabled in the emulator. In that configuration, Quasar FD looked much worse than Blackhole end-to-end, roughly ~2x worse.

That kicked off two parallel investigations:
1. iDMA vs NoC: iDMA has decent bandwidth (128B/cycle theoratical), but in this FD path it forces the prefetcher to wait for completion before releasing credit to the dispatcher. NoC is different: the payload write and credit publish can be ordered through the same VC, so the prefetcher can pipeline the copy more effectively instead of synchronously waiting on iDMA completion.
2. L1 hashing: the first NoC measurements were also misleading because L1 hashing was disabled. After enabling L1 hashing, NoC bandwidth improved substantially.

So the current comparison is Quasar emulator with L1 hashing enabled and the NoC copy path selected, compared against a fresh Blackhole p300 run with the same benchmark.

Current setup:
- Quasar: emulator, L1 hashing enabled, NoC copy path selected for prefetcher --> dispatcher local L1 transfers.
- Blackhole: fresh p300 run with the same current benchmark instrumentation.
- Benchmark: `CQ_PREFETCH_CMD_RELAY_LINEAR`, 100 iterations per payload.
- Units below are device cycles per command


Benchmarking sections:
- `period`: steady state FD end-to-end command period. Includes command fetch / CQ-facing overhead; on Quasar emulator this may include emulator related overhead. 
- `relay_cmd`: time spent by prefetcher handling `CQ_PREFETCH_CMD_RELAY_LINEAR` payload i.e. `process_relay_linear_cmd`
- `publish`: within `process_relay_linear_cmd`, overall time spent copying prefetcher L1 to dispatcher L1: including acquiring/releasing credits
- `dram_read`: exposed wait for the payload read into scratch buffer.
- `relay_internal` = `relay_cmd` - `publish` - `dram_read` : time spent calculating addresses, command parsing, loop/chunk etc.
[...]

PR #55122, opened 2026-09-02
[Performance]: Use cached L1 on FD command path on Quasar

sidnadTT:
### PR Category
Performance

### Summary
Relates to https://github.com/tenstorrent/tt-metal/issues/49781
Quasar  prefetcher/dispatcher uses uncached accesses to read command data. Uncached loads are expensive as compared to cache, and consequently stall the CPU pipeline. This PR moves the command read path to cached view and adds targeted `invalidate_l2_cache_range` calls where the bytes arrived over the NoC. 

Deliberately not converted: 
1. any host-visible pointers (`prefetch_q_rd_ptr`, completion-queue pointers)
2. any type of semaphores updates.
3. `aligned_go_signal_storage_uncached` in dispatch_s, dispatch.
4. `worker_completion_sem_addr` which workers increment over NoC atomic.

### Cache invalidation policy

In Prefetcher: once per fetch + two exec_buf sites. The extents are disjoint, so each line is invalidated exactly once within a single fetch cycle. All nine prefetcher `careful_copy` callers therefore pass invalidate_source = false.

In Dispatcher and dispatch_s: the header is invalidated at each command entry. Payload past that invalidates when applicable - in `careful_copy ... ` invalidate_source for sub-cmds, or sized from a count in the header (`set_go_signal_noc_data` in both kernels, `set_sub_device_worker_counts`). The extent isn't known until the header invalidate has already happened, which is also why the header invalidate is sized to `CQDispatchCmdLarge` by default.

The prefetcher can set snoop bit and relay packets so the receiving NIU packets the dispatcher's cache in place, removing the current dispatcher invalidate calls. But this was reverted for perf, not correctness: FD is prefetcher-bound, so this pays cycles on the bottleneck.

### Notes for reviewers
- Focus on the policy decision above. 
- The `noc_async_atomic_barrier()` was removed in `CBWriter::acquire_pages` for all archs -- wasn't really needed/useful

Improvements:

On Quasar:

Configuration | Transfer Size (Bytes) | Uncached + Barrier (Cycles) | Cache Opt + No Barrier (Cycles) | Total Cumulative Imp. (%)
-- | -- | -- | -- | --
256B × 1 | 256 | 1,358.90 | 1,299.30 | 4.39%
256B × 5 | 1,280 | 1,735.20 | 1,681.40 | 3.10%
256B × 35 | 8,960 | 4,673.60 | 3,883.60 | 16.90%
2,048B × 1 | 2,048 | 1,387.30 | 1,335.90 | 3.71%
2,048B × 5 | 10,240 | 1,857.10 | 1,790.00 | 3.61%
2,048B × 35 | 71,680 | 5,846.70 | 5,043.80 | 13.73%
12,288B × 1 | 12,288 | 1,669.90 | 1,617.40 | 3.14%
12,288B × 5 | 61,440 | 2,784.90 | 2,728.70 | 2.02%
12,288B × 35 | 430,080 | 17,403.50 | 15,604.00 | 10.34%

On BH (only barrier removal impacts BH/WH):

Configuration | Transfer Size (Bytes) | Baseline: With Barrier (Cycles) | Optimized: Barrier Removed (Cycles) | Barrier Removal Imp. (%)
-- | -- | -- | -- | --
256B × 1 | 256 | 1,164.40 | 1,161.90 | None (within noise margin)
256B × 5 | 1,280 | 1,495.00 | 1,495.80 | None (within noise margin)
256B × 35 | 8,960 | 3,803.00 | 3,790.10 | 0.34%
2,048B × 1 | 2,048 | 1,242.90 | 1,244.20 | None (within noise margin)
2,048B × 5 | 10,240 | 1,577.20 | 1,560.60 | 1.05%
2,048B × 35 | 71,680 | 5,279.10 | 4,991.30 | 5.45%
12,288B × 1 | 12,288 | 1,506.80 | 1,495.80 | None (within noise margin)
12,288B × 5 | 61,440 | 3,615.30 | 3,547.40 | 1.88%
12,288B × 35 | 430,080 | 19,972.20 | 19,552.20 | 2.10%

Quasar emulator prefetcher/dispatcher tests pass:
image

image

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:snadkarni/fd_cache_opt)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:snadkarni/fd_cache_opt)

### DIFF
--- a/tt_metal/impl/dispatch/kernels/cq_common.hpp
+++ b/tt_metal/impl/dispatch/kernels/cq_common.hpp
@@ -64,16 +64,6 @@ uint32_t wrap_gt(uint32_t a, uint32_t b) {
     return diff > 0;
 }
 
-// EXPERIMENT: tag the relay packets the dispatcher reads with a load -- command headers and the
-// sub-command arrays behind them -- so the receiving NIU snoops them into its cache, replacing the
-// dispatcher-side invalidations. Bulk payload is left untagged: the dispatcher only forwards it, and
-// that read is served from TL1 by the NIU rather than through the cache.
-// Merging is gated on whether a non-power-of-2 transfer is safe in inline-write mode, since TileLink
-// rounds the size up and may touch bytes outside the transfer.
-#ifndef FD_QSR_RELAY_SNOOP
-#define FD_QSR_RELAY_SNOOP 1
-#endif
-
 // Choosing the view for a sync word on Quasar: match the transport that publishes it, and use the same
 // view on both sides -- a cached store can sit dirty while an uncached poll reads TL1 and never sees it.
 //   Local CPU store (same-tile DM to DM): cached. DM caches are coherent with each other.
@@ -174,29 +164,15 @@ FORCE_INLINE void cq_noc_async_wwrite_with_state(
 // config (dst_noc, VC..) have been specified.
 // flush_last_transfer sets the flush packet tag on the final transfer so that a credit atomic issued after
 // this call -- typically from CBWriter::release_pages -- cannot commit to L1 ahead of the payload.
-// snoop_transfers tags every packet so the receiving NIU updates the destination's cache in place; set it
-// only for bytes the destination reads with a load, since it costs the receiver L2 write traffic.
 // No-op on tt-1xx, which has no packet tags.
 template <
     bool write_last_packet = true,
     bool update_counters = false,
     enum CQNocWait wait_first = CQ_NOC_WAIT,
     uint32_t cmd_buf = NCRISC_WR_CMD_BUF,
-    bool flush_last_transfer = false,
-    bool snoop_transfers = false>
+    bool flush_last_transfer = false>
 inline uint32_t cq_noc_async_write_with_state_any_len(
     uint32_t src_addr, uint64_t dst_addr, uint32_t size = 0, uint32_t ndests = 1, uint8_t noc = noc_index) {
-    static_assert(
-        !snoop_transfers || write_last_packet,
-        "snoop_transfers requires write_last_packet: the tag is persistent cmd-buf state and is cleared after the "
-        "final transfer, so a call that does not issue it would leak the tag onto unrelated writes.");
-#if defined(ARCH_QUASAR)
-    // Snoop is per-packet, so it must be set before the first of the burst-sized transfers below. Flush
-    // instead covers every packet that precedes it, so it is only needed on the last.
-    if constexpr (snoop_transfers) {
-        noc_set_packet_tags<cmd_buf>(/*snoop=*/true, /*flush=*/false);
-    }
-#endif
     if (size > NOC_MAX_BURST_SIZE) {
         cq_noc_async_write_with_state<CQ_NOC_SnDL, wait_first, CQ_NOC_SEND, cmd_buf, update_counters>(
             src_addr, dst_addr, NOC_MAX_BURST_SIZE, ndests);
@@ -214,13 +190,13 @@ inline uint32_t cq_noc_async_write_with_state_any_len(
     if constexpr (write_last_packet) {
 #if defined(ARCH_QUASAR)
         if constexpr (flush_last_transfer) {
-            noc_set_packet_tags<cmd_buf>(/*snoop=*/snoop_transfers, /*flush=*/true);
+            noc_set_packet_tags<cmd_buf>(/*snoop=*/false, /*flush=*/true);
         }
 #endif
         cq_noc_async_write_with_state<CQ_NOC_SnDL, CQ_NOC_WAIT, CQ_NOC_SEND, cmd_buf, update_counters>(
             src_addr, dst_addr, size, ndests, noc);
 #if defined(ARCH_QUASAR)
-        if constexpr (flush_last_transfer || snoop_transfers) {
+        if constexpr (flush_last_transfer) {
             noc_set_packet_tags<cmd_buf>(/*snoop=*/false, /*flush=*/false);
         }
 #endif
@@ -748,7 +724,7 @@ template <uint32_t l1_to_local_cache_copy_chunk, uint32_t l1_cache_elements_roun
 FORCE_INLINE void careful_copy_from_l1_to_local_cache(
     volatile uint32_t tt_l1_ptr* l1_ptr, uint32_t count, uint32_t* l1_cache) {
 #if defined(ARCH_QUASAR) && defined(COMPILE_FOR_DM)
-    if constexpr (invalidate_source && !FD_QSR_RELAY_SNOOP) {
+    if constexpr (invalidate_source) {
         // Upstream relayed the source by NoC write, which does not snoop, so a cached copy left over from an
         // earlier ring wrap is stale. Range covers the up-to-chunk-1 elements this function reads past count.
         // Prefetcher callers leave this off: both of its fetch paths already invalidate the extent they read.
--- a/tt_metal/impl/dispatch/kernels/cq_common.hpp
+++ b/tt_metal/impl/dispatch/kernels/cq_common.hpp
@@ -64,8 +64,22 @@ uint32_t wrap_gt(uint32_t a, uint32_t b) {
     return diff > 0;
 }
 
-// On Quasar, accesses to L1 semaphores must bypass the L1 D$ via the uncached alias so that
-// updates from other RISCs/cores are visible. No-op on BH/WH.
+// EXPERIMENT: tag the relay packets the dispatcher reads with a load -- command headers and the
+// sub-command arrays behind them -- so the receiving NIU snoops them into its cache, replacing the
+// dispatcher-side invalidations. Bulk payload is left untagged: the dispatcher only forwards it, and
+// that read is served from TL1 by the NIU rather than through the cache.
+// Merging is gated on whether a non-power-of-2 transfer is safe in inline-write mode, since TileLink
+// rounds the size up and may touch bytes outside the transfer.
+#ifndef FD_QSR_RELAY_SNOOP
+#define FD_QSR_RELAY_SNOOP 1
+#endif
+
+// Choosing the view for a sync word on Quasar: match the transport that publishes it, and use the same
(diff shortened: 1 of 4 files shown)


tt_metal/impl/dispatch/kernels/cq_common.hpp
```diff
@@ -64,8 +64,12 @@ uint32_t wrap_gt(uint32_t a, uint32_t b) {
     return diff > 0;
 }
 
-// On Quasar, accesses to L1 semaphores must bypass the L1 D$ via the uncached alias so that
-// updates from other RISCs/cores are visible. No-op on BH/WH.
+// Choosing the view on Quasar; both sides of a word must agree. No-op on BH/WH.
+//   Local CPU store, same-tile DM to DM: cached. DM caches are coherent with each other.
+//   NoC atomic: uncached unless the increment sets the snoop bit, which FD leaves off because empherical
+//   experiments snooping was some cycles more expensive than uncached poll.
```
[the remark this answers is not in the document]
sidnadTT: corrected

tt_metal/impl/dispatch/kernels/cq_dispatch_subordinate.cpp
```diff
@@ -624,8 +632,12 @@ void kernel_main() {
         device_print_dispatcher.execute();
 #endif
         cb_acquire_pages_dispatch_s<my_noc_xy, my_dispatch_cb_sem_id>(1);
-
-        volatile CQDispatchCmd tt_l1_ptr* cmd = uncached_l1_ptr<CQDispatchCmd>(cmd_ptr);
+#if defined(ARCH_QUASAR) && defined(COMPILE_FOR_DM)
+        // Upstream relays this command by NoC write, which does not snoop, so the header must be dropped before
+        // it is read cached. Arrays past the header stay uncached.
+        invalidate_l2_cache_range(cmd_ptr, sizeof(CQDispatchCmd));
```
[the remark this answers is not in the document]
sidnadTT: Fixed

tt_metal/impl/dispatch/kernels/cq_prefetch.cpp
```diff
@@ -795,7 +798,12 @@ void fetch_q_get_cmds(uintptr_t& fence, uintptr_t& cmd_ptr, uint32_t& pcie_read_
                     fence,
                     cmd_ptr);
 #endif
-
+#if defined(ARCH_QUASAR) && defined(COMPILE_FOR_DM)
+                // NoC reads land in TL1 without snooping the DM cache, so lines this ring region left
+                // cached from an earlier wrap are stale. Every command the prefetcher reads comes through
+                // here.
+                invalidate_l2_cache_range(inflight[idx].read_start, inflight[idx].reserved_size);
```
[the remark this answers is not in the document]
sidnadTT: Added

tt_metal/impl/dispatch/kernels/cq_dispatch.cpp
```diff
@@ -193,7 +193,12 @@ struct NocReleasePolicy {
     template <uint8_t noc_idx, uint32_t noc_xy, uint32_t sem_id>
     static FORCE_INLINE void release(uint32_t pages) {
 #ifdef ARCH_QUASAR
-        Semaphore<programmable_core_type>(sem_id).up(pages);
+        // get_semaphore() returns an offset into this core's L1, so this local store
+        // works only because prefetcher and dispatcher are on same dispatch engines sharing one L1.
+        // A split _h/_d build must go over the NoC instead. See the checklist at the top of cq_prefetch.cpp.
+        // Non-atomic RMW, so exactly one agent may ever increment this semaphore.
+        auto* sem_addr = reinterpret_cast<volatile tt_l1_ptr uint32_t*>(get_semaphore<programmable_core_type>(sem_id));
+        *sem_addr += pages;
```
sidnadTT, line 201: not atomic - we're leaning on there being exactly one incrementer rather than paying for an risc-v atomic.

tt_metal/impl/dispatch/kernels/cq_dispatch.cpp
```diff
@@ -1281,22 +1276,31 @@ void process_notify_dispatch_s_go_signal_cmd() {
...
 static inline bool process_cmd_d(uintptr_t& cmd_ptr, uint32_t* l1_cache) {
     bool done = false;
 re_run_command:
-    volatile CQDispatchCmd tt_l1_ptr* cmd =
-        reinterpret_cast<volatile CQDispatchCmd tt_l1_ptr*>(l1_uncached_addr(cmd_ptr));
+#if defined(ARCH_QUASAR) && defined(COMPILE_FOR_DM)
+    // Upstream relays this command by NoC write, which does not snoop, so the header must be dropped before
+    // it is read cached. Sized to CQDispatchCmdLarge because the variant is unknown until cmd_id is read.
+    // CPU reads past this window carry their own invalidate; payload handed to the NIU needs none.
+    invalidate_l2_cache_range(cmd_ptr, sizeof(CQDispatchCmdLarge));
```
sidnadTT, line 1301: @jbaumanTT  we can also set the snoop bit for downstream writes in the prefetcher and I tested this, but it costs some extra cycles (10-50cycles/period) so instead invalidating here in the dispatcher since we're prefetcher bound and this is pretty much free in the dispatcher

tt_metal/impl/dispatch/kernels/cq_prefetch.cpp
```diff
@@ -795,7 +798,12 @@ void fetch_q_get_cmds(uintptr_t& fence, uintptr_t& cmd_ptr, uint32_t& pcie_read_
                     fence,
                     cmd_ptr);
 #endif
-
+#if defined(ARCH_QUASAR) && defined(COMPILE_FOR_DM)
+                // NoC reads land in TL1 without snooping the DM cache, so lines this ring region left
+                // cached from an earlier wrap are stale. Every command the prefetcher reads comes through
+                // here.
```
sidnadTT, line 804: For now, added invalidation here i.e. as soon as we fetch into cmddat_q, invalidate the full fetched extent. This is much better for paged packed sub cmds perf (measured), but likely more expensive for inline writes. Once we characterize pgm, this might change.

tt_metal/impl/dispatch/kernels/cq_dispatch.cpp
jbaumanTT: Do we have to be careful that there's no other noc writes into this 64-byte region? Otherwise we could run into that issue where NoC writes to a dirty cache line can cause lost updates.

tt_metal/impl/dispatch/kernels/cq_prefetch.cpp
jbaumanTT: Did you measure this with write_linear or write_paged commands (presumably with a relay inline as a source)? Also, we could consider adding a flag to the fetch queue to say whether it should invalidate or not (depending on the commands it contains), but that's definitely something for a followup patch, if at all.

tt_metal/impl/dispatch/kernels/cq_dispatch.cpp
sidnadTT: Yes that's possible. I reverted this back to uncached accesses and will create a new PR which will guarantee 64B alignment based on host select semaphore config (which was introduced in this PR: https://github.com/tenstorrent/tt-metal/pull/53687)

tt_metal/impl/dispatch/kernels/cq_prefetch.cpp
sidnadTT: Ok for inline writes this approach of invalidating all full fetched payload (includes inline payload) in `fetch_q_get_cmds` gets super expensive with payload size, so I changed it to selective invalidation outside of `fetch_q_get_cmds` in `careful_copy_from_l1_to_local_cache` or `process_cmd`:

image


for paged sub cmds,  selective invalidation in `careful_copy_from_l1_to_local_cache` or `process_cmd` is slightly more expensive:

paged_packed_plot


I think your idea of invalidating by setting a bit in fetch queue entry makes sense and might be  most optimal, although we'll have to think about the exec_buf case there. I'll create a separate PR for it

tt_metal/impl/dispatch/kernels/cq_common.hpp
```diff
@@ -719,9 +717,31 @@ constexpr uint32_t l1_to_local_cache_copy_chunk = 6;
...
+        if constexpr (first_line_invalidated) {
+            // Shrink by what is skipped, not just advance: keeping the size would push the range one line
+            // past the array and hand back the line just saved. A short array can skip past its own end.
+            const uint32_t skipped = round_up_pow2(start, L2_CACHE_LINE_SIZE) - start;
+            start += skipped;
+            size = skipped < size ? size - skipped : 0;
+        }
+        invalidate_l2_cache_range(start, size);
+    }
+#endif
```
sidnadTT, line 744: Added a small optimization here to prevent re-invalidation of same cache lines, when there's overlap with cache line already invalidates in `process_cmd`. This part can be removed once we have the host side set bit in the fetchq entries to do an adaptive invalidate based on the command.

jbaumanTT (approved):

PR #55127, opened 2026-09-02
[Bug fix]: preserve ND-created output config in indexed_fill fallback

### LINKED ISSUE #46957: [ops]: ttnn.indexed_fill rebuilds ND-created output configs through legacy shard specs
Component / Area

ops

Issue Type (optional)

Bad Outputs

Observed

ttnn.indexed_fill(...) defaults output_memory_config to input_tensor_a.memory_config() when the caller does not override it.

Both the wrapper fallback and the device op then route that config through resolve_output_memory_config(...). That helper rebuilds sharded outputs from a legacy shard-spec path when the config is sharded and does not already carry a legacy shard_spec.

That is correct for legacy sharded configs. It is wrong for ND-created configs. An ND-created MemoryConfig already carries the authoritative shard distribution in nd_shard_spec plus created_with_nd_shard_spec, so rebuilding it through the legacy path rewrites the output contract even though indexed_fill preserves output shape.

Expected

When indexed_fill preserves output shape and the caller does not override the output config, the operator should keep an ND-created output MemoryConfig authoritative instead of rebuilding it through the legacy shard-spec path.

Steps to Reproduce the Issue

1. Steps (exact commands)

Source proof only so far.

2. Input data / link or description

The path is:

- ND-sharded TILE input tensor
- no explicit output memory_config
- graph capture around indexed_fill
- inspect the captured output attributes

3. Frequency

Always on the ND-created output path

System Details

1. Software Versions

OS version:

- current main
Software / firmware release version:

- current upstream main at readback time

2. Hardware Details

Product:

- tt-metal ops
Card/System (e.g., N150, P150, Loudbox, Quietbox, Galaxy, etc.):

- unknown

Regression Info (Optional)

Is this a regression?

- Unknown

Logs & Diagnostics (Optional)

Source evidence:

- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/ttnn/cpp/ttnn/operations/data_movement/indexed_fill/indexed_fill.cpp#L68-L79
- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/ttnn/cpp/ttnn/operations/data_movement/indexed_fill/indexed_fill.cpp#L112-L120
- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_device_operation.cpp#L181-L193
- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_utils.cpp#L291-L309
- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/tt_metal/impl/tensor/spec/memory_config/memory_config.cpp#L59-L83
- https://github.com/tenstorrent/tt-metal/blob/10358b932eda035c3704d5c06183a429da4c5302/tt_metal/impl/tensor/spec/tensor_spec.cpp#L246-L253

abdulla-TT:
### PR Category
Bug fix

### Summary
Fixes #46957.

When you don't pass an output memory config, `ttnn.indexed_fill` reuses input_a's. If that config came from an `NdShardSpec`, the op was throwing the ND shard distribution away and rebuilding a legacy 2D shard spec in its place. Since the op doesn't change the output shape, there was nothing to rebuild — and where no equivalent legacy spec exists, the rewrite quietly changes where the output data actually lives. A `[8, 1, 64, 64]` input sharded `[1, 1, 32, 32]` over a 3x4 grid came back as `[64, 32]` over a 2x8 grid.

It needed fixing in two places, because `indexed_fill` resolves its output config down two different paths:

- **`indexed_fill_utils.cpp`** — `resolve_output_memory_config(...)` now leaves ND-created configs alone. Legacy sharded configs without a shard spec still get one derived as before.
- **`permute_device_operation.cpp`** — for sub-page dims (`dim >= rank-2` on TILE), `indexed_fill` falls back to a permute whose intermediates are INTERLEAVED by design. Permute was only keeping ND provenance when *both* its input and output config were ND-created, so that interleaved intermediate sent the config straight back through legacy synthesis and undid the first fix. It now honors an ND output config on its own.

### Notes for reviewers
The permute change sounds broader than it is. "ND output config with a non-ND input" can only happen when a caller explicitly asks for an ND output, since permute otherwise derives its config from the input. An explicit request is already in the output frame, so it's used as-is — same as `untilize`. The one behavior change: pass an ND spec that doesn't suit the output shape and you now get your spec (and any error that follows) instead of a silently substituted legacy one.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:abdulla/indexed_fill_nd_shard)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:abdulla/indexed_fill_nd_shard)


ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_utils.cpp
```diff
@@ -261,6 +261,13 @@ tt::tt_metal::MemoryConfig resolve_output_memory_config(
     const Tensor& input_tensor_a,
     const ttnn::Shape& padded_out_shape,
     const tt::tt_metal::MemoryConfig& output_mem_config) {
+    // indexed_fill preserves the output shape, so an ND-sharded output config already
+    // carries the authoritative shard distribution. Re-deriving a legacy shard_spec here
+    // is unnecessary and would rewrite an ND-origin config onto the legacy shard-spec path.
+    if (output_mem_config.created_with_nd_shard_spec()) {
```
Aswinmcw: Nit: this guard checks only `created_with_nd_shard_spec()`, while the permute change in this same PR checks `created_with_nd_shard_spec() && nd_shard_spec().has_value()`. The public `MemoryConfig(buffer_type, nd_shard_spec)` ctor couples the two, but the 5-arg ctor takes the flag independently and `to_json` defensively nests both checks, so the flag-set/optional-empty state is representable. If one ever reaches here, this returns a sharded config with neither a legacy `shard_spec` nor an `nd_shard_spec` and the final permute in the fallback path falls back into legacy synthesis anyway. Matching the permute guard here would keep the two consistent.

tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
```diff
@@ -271,6 +271,63 @@ def test_indexed_fill_program_cache(device, variant):
...
+    assert output_mem_config.nd_shard_spec == input_mem_config.nd_shard_spec, (
+        f"dim={dim}: output ND shard spec was rebuilt.\n"
+        f"  expected (input_a): {input_mem_config.nd_shard_spec}\n"
+        f"  actual   (output) : {output_mem_config.nd_shard_spec}\n"
+        f"  full output config: {output_mem_config}"
+    )
+    logger.info(f"Indexed Fill (ND_SHARDED, dim={dim}) Output Memory Config: {output_mem_config}")
+
+    golden = golden_indexed_fill(torch_a, torch_b, batch_id, dim=dim)
+    assert_with_pcc(golden, ttnn.to_torch(output_tensor), 0.9999)
```
Aswinmcw, line 328: Nit: `indexed_fill` is pure data movement, so the output should be bit-exact against the golden — `torch.equal` (or `assert_equal`) would catch a shard-distribution bug that a 0.9999 PCC could still average away. It matches the convention in the rest of this file, so fine to leave, but exact is strictly stronger here.

Aswinmcw (approved):
LGTM. Two nits inline, neither blocking.

tests/ttnn/unit_tests/operations/data_movement/test_indexed_fill.py
abdulla-TT: For now will keep as the already existing file format with pcc.

VirdhatchaniKN (approved):
LGTM

mateusznowakTT (approved):

sjameelTT (approved):

ttnn/cpp/ttnn/operations/data_movement/indexed_fill/device/indexed_fill_utils.cpp
```diff
@@ -261,6 +261,13 @@ tt::tt_metal::MemoryConfig resolve_output_memory_config(
     const Tensor& input_tensor_a,
     const ttnn::Shape& padded_out_shape,
     const tt::tt_metal::MemoryConfig& output_mem_config) {
+    // indexed_fill preserves the output shape, so an ND-sharded output config already
```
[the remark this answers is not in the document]
abdulla-TT: Updated.

PR #55132, opened 2026-09-02
Propagate the softplus boundary fix to the Quasar kernel

AJ0070:
Completes #53261 on the third architecture.

#54208 corrected the softplus boundary on `wormhole_b0` and `blackhole`. While
re-checking that fix against `main` I found the same comparison in a third kernel
copy and said so on #53261 at the time; #54208's description records that the quasar
copy was left untouched there. This is that follow-up.

`tt_metal/hw/ckernels/quasar/metal/llk_api/llk_sfpu/ckernel_sfpu_softplus.h:58` still
reads `v_if (t < threshold)`. torch documents the linear fallback as
`input * beta > threshold`, so at exactly `t == threshold` the result should be
`log1p(exp(t)) / beta`, not `x`. The strict comparison sends the boundary lane down
the linear branch instead.

The change is the same one-character correction #54208 applied to the other two
architectures, `<` to `<=`, plus a comment recording why the boundary belongs to the
nonlinear side. `<=` is also the comparison shape #53058 prefers on Quasar.

No behaviour changes for any lane other than exactly `t == threshold`.


AJ0070:
Closing as a duplicate of #54858, which makes the same change.

PR #55133, opened 2026-09-02
[ttnn/msda] do the bilinear geometry in fixed point

sott0n:
### PR Category

Performance

### Summary

The reader in `ttnn.experimental.multi_scale_deformable_attn` computed its bilinear sampling geometry in float. The dataflow cores have no FPU, so each of those operations is a ~100-cycle software routine — that one chain measured **27 ms of the op's 57.5 ms**.

The coordinates are bounded (grid in `[-1, 1]`, pixel coords by the feature-map size) and only ever feed a bf16 result, so they fit Q16.16 integers comfortably. In fixed point the integer/fraction split becomes reading two halves of one integer, so the `floor` call and the int-to-float converts disappear.
Constants and helpers come from the existing `pool/device/kernels/fixed_point_arithmetic.hpp`, which upsample and rotate already use for the same job. 

Measured on BH P150 at N=8, Q=89023, P=4, D=32, feature map 200×334:

| | before | after |
|---|---:|---:|
| op | 57.5 ms | **31.1 ms** (1.85×) |
| with value reads disabled, i.e. this chain alone | 31.9 ms | **10.2 ms** (3.2×) |

Three follow-up commits fix review findings, each with a test that fails without it:

- the weight multiply's overflow branch dropped the sign
- the coordinate clamp was too small for `align_corners` on a 2-wide map
- feature-map dimensions past 16384 are rejected instead of overflowing int32
- the coordinate keeps its low bit: scale first, then halve via a 64-bit intermediate

### Notes for reviewers
N/A

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kyamaguchi/msda-op)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kyamaguchi/msda-op)


ttnn/cpp/ttnn/operations/experimental/multi_scale_deformable_attn/device/kernels/dataflow/reader_msda.cpp
```diff
@@ -35,36 +35,131 @@
...
+    } else {
+        magnitude = 0;
+    }
+    if (magnitude > clamp_q16) {
+        magnitude = clamp_q16;
+    }
+    return negative ? -magnitude : magnitude;
+}
+
+// Precondition: v != 0. __builtin_clz(0) is undefined, and the fallback loop
```
bbradelTT: Would you be able to add in an assert that would at least be caught if watcher or asserts are enabled?

bbradelTT (approved):

ttnn/cpp/ttnn/operations/experimental/multi_scale_deformable_attn/device/kernels/dataflow/reader_msda.cpp
sott0n: addressed

PR #55139, opened 2026-09-02
[Performance] binary_ng/ternary: replicate the broadcast row with local NoC copies

### LINKED ISSUE #55138: [Bounty $2,500] binary_ng/ternary: replicate the broadcast row with local NoC copies instead of ~2000 scalar stores per tile
**Component / Area:** ttnn eltwise dataflow kernels (`fill_tile_utils.hpp`, binary_ng / ternary / quasar readers)

**Issue Type:** Performance

## Observed

`fill_tile_with_first_row` replicates one row across a tile with a scalar loop on a data-movement core, re-reading its source through a `volatile` pointer on every iteration of the innermost loop:

```cpp
for (uint32_t k = 0, face_offset = 0; k < 2; ++k, face_offset += 512) {
    for (uint32_t row = 0; row < num_rows; ++row) {
        for (uint32_t col = 0; col < 16; ++col) {
            ptr[dst_offset + col]       = ptr[col];        // volatile load, every iteration
            ptr[dst_offset + col + 256] = ptr[col + 256];
        }
        row_offset += 16;
    }
}
```

That is ~992 volatile loads and ~992 stores per tile, and the row readers call it **once per output tile** — about two million of each for a 1024-tile call. The column reader next to it hoists its fill out of the inner loop and hoists its own load, and measures at 0.78-1.03x; only the row path pays.

## Expected

Within a face the 16 rows are contiguous — 16 words, 64 B per row, 1024 B per face — so doubling the valid region fills a face in four transfers. Faces 0 and 1 are the top half of the tile and hold the source row; faces 2 and 3 are the bottom half and have no row of their own, which is why the scalar version fills 15 rows for the first face pair and 16 for the second. Double within the top faces, then copy each finished face into the one below it.

```cpp
FORCE_INLINE void fill_tile_with_first_row(uint32_t l1_write_ptr) {
    constexpr uint32_t kFaceBytes = 1024;
    constexpr uint32_t kRowBytes = 64;
    const uint32_t f0 = l1_write_ptr;
    const uint32_t f1 = f0 + kFaceBytes;
    for (uint32_t n = kRowBytes; n < kFaceBytes; n <<= 1) {
        noc_async_read(get_noc_addr(f0), f0 + n, n);
        noc_async_read(get_noc_addr(f1), f1 + n, n);
        noc_async_read_barrier();
    }
    noc_async_read(get_noc_addr(f0), f0 + 2 * kFaceBytes, kFaceBytes);
    noc_async_read(get_noc_addr(f1), f1 + 2 * kFaceBytes, kFaceBytes);
    noc_async_read_barrier();
}
```

Ten transfers and five barriers per tile against ~1,984 volatile memory operations. #52909 makes the same move for `generate_index_tile`, and `writer_deepseek_grouped_gate.cpp:30` already uses the construction in-tree.

## Measured

Both machines at `047ec3ccf96`, per-op wall clock in microseconds, median of 7 runs of 30 back-to-back calls with a `synchronize_device` per run. Implemented and measured, not proposed. An ablation that deletes the replication entirely — wrong results by construction, run only to bound the saving — lands within noise of the after column at every large shape, so the fill is essentially the whole gap. Tables are per machine because the two behave differently and for opposite reasons.

### Blackhole P300

| op | shape | dtype | broadcast operands | before | after | |
|---|---|---|---|---|---|---|
| `ttnn.where` | `[1,1,1024,1024]` | f32 | 2 rows | 226.24 | **43.15** | -80.9% |
| `ttnn.where` | `[1,1,32,32768]` | f32 | 2 rows | 226.99 | **44.27** | -80.5% |
| `ttnn.where` | `[1,1,32,8192]` | f32 | 2 rows | 74.32 | **16.50** | -77.8% |
| `ttnn.multiply` | `[1,1,1024,1024]` | f32 | 1 row | 118.06 | **31.93** | -73.0% |
| `ttnn.add` | `[1,1,32,32768]` | f32 | 1 row | 117.91 | **31.96** | -72.9% |
| `ttnn.multiply` | `[1,1,32,32768]` | f32 | 1 row | 117.94 | **31.98** | -72.9% |
| `ttnn.gt` | `[1,1,32,32768]` | f32 | 1 row | 118.05 | **32.02** | -72.9% |
| `ttnn.add` | `[1,1,1024,1024]` | f32 | 1 row | 118.13 | **32.06** | -72.9% |
| `ttnn.gt` | `[1,1,1024,1024]` | f32 | 1 row | 118.30 | **32.30** | -72.7% |
| `ttnn.where` | `[1,1,32,2048]` | f32 | 2 rows | 26.07 | **8.77** | -66.4% |
| `ttnn.where` | `[1,1,1024,1024]` | f32 | 1 row | 120.59 | **42.34** | -64.9% |
| `ttnn.where` | `[1,1,32,32768]` | f32 | 1 row | 121.40 | **43.44** | -64.2% |
| `ttnn.where` | `[1,1,32,8192]` | f32 | 1 row | 42.70 | **15.36** | -64.0% |
| `ttnn.where` | `[1,1,32,2048]` | f32 | 1 row | 15.53 | **8.84** | -43.1% |
| `ttnn.gt` | `[1,1,32,8192]` | f32 | 1 row | 40.36 | **25.00** | -38.1% |
| `ttnn.multiply` | `[1,1,32,8192]` | f32 | 1 row | 40.17 | **25.11** | -37.5% |
| `ttnn.add` | `[1,1,32,8192]` | f32 | 1 row | 40.17 | **25.34** | -36.9% |
| `ttnn.where` | `[1,1,32,2048]` | bf16 | 2 rows | 8.83 | **8.49** | -3.9% |
| `ttnn.gt` | `[1,1,32,2048]` | f32 | 1 row | 25.35 | **24.78** | -2.2% |
| `ttnn.gt` | `[1,1,1024,1024]` | bf16 | 1 row | 28.65 | **28.03** | -2.2% |
| `ttnn.gt` | `[1,1,32,8192]` | bf16 | 1 row | 25.14 | **24.62** | -2.1% |

The remaining 19 of 40 measured rows do not move beyond the 2% noise floor on this machine.

### Wormhole N300

| op | shape | dtype | broadcast operands | before | after | |
|---|---|---|---|---|---|---|
| `ttnn.where` | `[1,1,32,32768]` | f32 | 2 rows | 547.81 | **122.30** | -77.7% |
| `ttnn.where` | `[1,1,1024,1024]` | f32 | 2 rows | 551.91 | **134.14** | -75.7% |
| `ttnn.where` | `[1,1,32,8192]` | f32 | 2 rows | 166.42 | **62.17** | -62.6% |
| `ttnn.where` | `[1,1,32,32768]` | f32 | 1 row | 304.83 | **117.42** | -61.5% |
| `ttnn.where` | `[1,1,1024,1024]` | f32 | 1 row | 307.21 | **124.50** | -59.5% |
| `ttnn.where` | `[1,1,32,2048]` | f32 | 2 rows | 49.08 | **26.24** | -46.5% |
| `ttnn.where` | `[1,1,32,8192]` | f32 | 1 row | 103.72 | **62.41** | -39.8% |
| `ttnn.where` | `[1,1,32,8192]` | bf16 | 1 row | 38.57 | **24.88** | -35.5% |
| `ttnn.where` | `[1,1,32,8192]` | bf16 | 2 rows | 32.28 | **24.89** | -22.9% |

The remaining 31 of 40 measured rows do not move beyond the 10% noise floor on this machine.

Baselines reproduce to under 2% on P300. **On N300 they do not**: the small shapes there are unstable, with run-to-run spreads from 6% to 118% on repeated sampling, so that machine's cut is drawn at 10% and every row inside it is reported as unmoved rather than as a small change. One row initially read as a +58% regression and was a sampling artifact — re-measured with 15 runs of 50 calls it is 24.37 against 24.38.

Which rows move is decided by whether that combination already had a hardware broadcast:

- **bfloat16 does not move, and does not need to.** `unary_bcast<BroadcastType::ROW>` lowers to `MOVD2B` / `MOVB2D` with `MOV_8_ROW_BRCST`, run from a MOP: the replication happens inside the Dst register file, eight instructions for a tile, and the reader's fill is compiled out entirely (`#if SRC_BCAST && !BCAST_LLK`). There is nothing here to improve.
- **float32 has no such path in the ternary readers.** `ternary_program_factory.cpp:218` admits the hardware broadcast only when all three operands are bfloat16. Forcing float32 through it anyway returns wrong values for every element, on **both** architectures — not the Blackhole-specific tenstorrent/tt-llk#449, but because the 32-bit two-pass sequence exists only in the `A2D` unpack-to-dest branch and not in the `B2D` + MOP branch that `unary_bcast` uses. So the bfloat16-only condition is correct, and software replication is the only option there.
- **float32 in binary_ng moves on Blackhole only**, because `is_llk_bcast` already routes the 32-bit dtypes through the hardware broadcast on Wormhole. That exclusion is a separate report.

## Correctness

Outputs bit-identical to the scalar fill at every shape and dtype measured, on both machines. `ttnn.where` with row-broadcast operands checked against `torch.where` at 3 widths, 0 of 65,536 elements wrong.

## Scope

| site | |
|---|---|
| `eltwise/binary_ng/device/kernels_ng/dataflow/reader_interleaved_row_bcast.cpp` | in scope |
| `eltwise/binary_ng/.../reader_interleaved_row_col_mixed_bcast.cpp`, `_scalar_bcast.cpp` | in scope |
| `eltwise/ternary/.../ternary_reader_rowbcast_ttt.cpp` | in scope, fills **three** operands per output tile |
| `eltwise/ternary/.../tts_tst_reader_row_col_bcast.cpp` | in scope |
| `experimental/quasar/binary_ng/...` | a separate duplicate of the same function, **not** changed here |
| `fill_tile_with_first_column` | already hoisted, measures 0.78-1.03x, **not** in scope |
| bfp8 / bfp4 fills | already packed, not per-element, not in scope |

Transposing a column broadcast into a row broadcast is not worth it: the column path measures 0.78-0.90x, i.e. **faster** than a same-shape operand, because a column operand is one tile per tile row and the read saving outweighs the fill.

## Not established

- A mechanism that would remove the replication altogether rather than make it cheaper: the packer can source datums from L1, and a zero row stride would pack every output row from the same source row. Unmeasured, and an API-level change, so it is not proposed here.
- Only float32 and bfloat16 were timed, through `multiply` / `add` / `gt` / `where`.
- The quasar readers were read, not measured.
- `noc_async_write` in place of the loop-back read measures identically (43.67 vs 43.72 us), so the direction of the local transfer does not matter; the read form is kept because it matches the existing in-tree precedent.
- The barrier per doubling level waits on all outstanding reads on that NoC. The current readers already barrier per tile so nothing is lost, but a reader that prefetched the next tile would lose that overlap.

kinginu:
## PR Category

Performance.

## Summary

Fixes #55138, which carries the measurements and the scope.

`fill_tile_with_first_row` replicated one row across a tile with a scalar loop on a data-movement core, re-reading its source through a `volatile` pointer on every iteration of the innermost loop — about 992 loads and 992 stores per tile, once per output tile.

Within a face the 16 rows are contiguous, 16 words and 64 B per row, 1024 B per face, so doubling the valid region fills a face in four transfers. Faces 0 and 1 are the top half of the tile and hold the source row; faces 2 and 3 are the bottom half and have none of their own, which is why the loop being replaced filled 15 rows for the first face pair and 16 for the second. Double within the top faces, then copy each finished face into the one below: ten transfers and five barriers per tile.

#52909 makes the same move for `generate_index_tile`, and `writer_deepseek_grouped_gate.cpp:30` already uses the construction in-tree.

One function, shared by the binary_ng and ternary row readers. `experimental/quasar/binary_ng` carries its own duplicate of the same helper; it is left alone here because it was read but never measured. `bfloat16` never reaches this path — the LLK broadcast compiles the call out — so this is float32 only.

Headline: `ttnn.where` at `[1,1,32,32768]` with two row-broadcast operands, 226.99 to 44.27 us on P300 and 547.81 to 122.30 us on N300. Full tables per machine in #55138.

## Correctness

Outputs bit-identical to the loop being replaced, at every shape and dtype measured, on both machines. Checked two ways because a float reference is only valid for one of them:

- `ttnn.where` with row-broadcast operands against `torch.where`. Pure selection, so an exact reference holds for both dtypes.
- `ttnn.multiply` with a row-broadcast operand against the same kernel handed a materialised full-size operand, which keeps the dtype rounding identical on both sides.

0 of 2,752,512 elements differ, across float32 and bfloat16 at `[1,1,32,2048]`, `[1,1,32,8192]` and `[1,1,1024,1024]`.

The loop-back read needs its source visible in L1 first. It is: the row is written by the reader's own `noc_async_read` from DRAM, already followed by `noc_async_read_barrier()` before the fill, and each doubling level is ordered against the previous by its own barrier — NoC to NoC throughout, never the RISCV-to-NoC handoff that needed #50374's drain guard.

## Cost

None. The change removes work and adds no instructions to any path that did not already call this function.

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `047ec3ccf96` (current `main`)
* Hardware: P300 (Blackhole) and N300 (Wormhole)


mateusznowakTT (approved):

ttnn/cpp/ttnn/operations/eltwise/binary_ng/device/kernels/dataflow/fill_tile_utils.hpp
```diff
@@ -369,25 +369,23 @@ FORCE_INLINE void fill_tile_with_first_row_bfloat16(uint32_t l1_write_ptr) {
...
+    const uint32_t f0 = l1_write_ptr;
+    const uint32_t f1 = f0 + kFaceBytes;
+    for (uint32_t n = kRowBytes; n < kFaceBytes; n <<= 1) {
+        noc_async_read(get_noc_addr(f0), f0 + n, n);
+        noc_async_read(get_noc_addr(f1), f1 + n, n);
+        noc_async_read_barrier();
     }
+    noc_async_read(get_noc_addr(f0), f0 + 2 * kFaceBytes, kFaceBytes);
+    noc_async_read(get_noc_addr(f1), f1 + 2 * kFaceBytes, kFaceBytes);
+    noc_async_read_barrier();
```
Aswinmcw: Faces 2 and 3 are contiguous with each other exactly as 0 and 1 are, so these two reads can be one:

```cpp
noc_async_read(get_noc_addr(f0), f0 + 2 * kFaceBytes, 2 * kFaceBytes);
```

Source `[f0, f0 + 2048)` is faces 0-1, both fully valid once the loop exits, and destination `[f0 + 2048, f0 + 4096)` is faces 2-3. Nine transfers per tile instead of ten, one fewer packet setup.

The doubling levels can't be merged the same way — `[f0, f0 + n)` and `[f1, f1 + n)` only become adjacent at `n == kFaceBytes` — so this last step is the one place it works.

ttnn/cpp/ttnn/operations/eltwise/binary_ng/device/kernels/dataflow/fill_tile_utils.hpp
```diff
@@ -369,25 +369,23 @@ FORCE_INLINE void fill_tile_with_first_row_bfloat16(uint32_t l1_write_ptr) {
 // Reads the very first row of the CB and fills the entire tile with the same row.
 // Tile is assumed to have 32-bit elements (float32/int32).
 FORCE_INLINE void fill_tile_with_first_row(uint32_t l1_write_ptr) {
```
Aswinmcw, line 371: `experimental/quasar/binary_ng/device/kernels/dataflow/fill_tile_utils.hpp` is a byte-identical copy of this header today; after this lands the two differ only in this function. The scope table in #55138 records the decision to leave it, but nothing in the tree will. Could you open a tracking issue for the quasar side so the divergence is visible to whoever touches that copy next?

Aswinmcw (approved):
LGTM. Two notes inline, neither blocking.

KalaivaniMCW:
Note: Wormhole widely uses the hardware LLK broadcast not `fill_tile_with_first_row` , in WH only exp-family ops use this s/w bcast. 
In Blackhole we use `fill_tile_with_first_row` for 32-bit dtype

kinginu:
> Note: Wormhole widely uses the hardware LLK broadcast not `fill_tile_with_first_row` , in WH only exp-family ops use this s/w bcast. In Blackhole we use `fill_tile_with_first_row` for 32-bit dtype

Right for binary_ng. ttnn.where goes through ternary_program_factory, which gates on all three operands being bfloat16, so float32 where takes the software fill on Wormhole too — that's the -77.7% row in the table.

PR #55145, opened 2026-09-02
[Bug fix] matmul: drain the last-K-tile padding stores before the in0 multicast reads them

### LINKED ISSUE #55141: matmul in0 mcast senders: RISC-written K padding is multicast without a drain
**Component / Area:** ttnn matmul dataflow kernels (`reader_bmm_tile_layout_in0_sender_dram_sharded.cpp`, `reader_bmm_tile_layout_in0_sender_padding.cpp`, `reader_bmm_tile_layout_in0_sender_receiver_padding_block_sharded.cpp`, `kernel_helper_functions/pad_tile.hpp`)

**Issue Type:** Bug (store-visibility, same class as #50374)

## Observed

When K (or the transposed dimension) is not a multiple of 32, the in0 multicast senders zero the padding of the last K tile with baby-RISCV stores (`pad_last_ktile` / `pad_last_transposed_ktile`, through a plain `T*`, `pad_tile.hpp:137`) and then multicast the block with the NoC. In the DRAM-sharded sender — the decode matmul — nothing at all sits between the last pad store and the NoC command:

```cpp
// reader_bmm_tile_layout_in0_sender_dram_sharded.cpp:91-115
if constexpr (in0_last_ktile_w > 0) {
    if (is_last_ktile_padded && (i == num_blocks_per_shard - 1)) {
        for (uint32_t h = 0; h < in0_block_h; ++h) {
            ...
            pad_last_ktile<in0_data_format, in0_last_ktile_w>(in0_last_ktile_ptr);   // RISC stores
        }
    }
}
...
#ifndef SKIP_MCAST
MulticastEndpoint mcast_dst;
noc.async_write_multicast(CoreLocalMem<uint32_t>(local_read_addr), mcast_dst, in0_block_size_bytes, ...);  // NoC reads the block
```

The 1D sender (`:282-292` interleaved, `:336-368` sharded) and the 2D block-sharded sender (`:168-215`) have a `sender_sem.wait()` / `.set(0)` in between — a load of a different address and a store. That orders the *emission* of the pad stores, which the following store already does, but not their *processing*, and `tt-isa-documentation`'s `TensixTile/BabyRISCV/MemoryOrdering.md` §"Enforcing stronger ordering" is explicit that the two are different:

> There is no _general_ mechanism to ensure that a store's write-request has been processed before a subsequent memory operation's request is emitted.

The one construction it does prescribe is the readback: "the store can be followed by a load from the same address, at which point the store's write-request will be processed before the load's read-request". The page is present for both `WormholeB0/` and `BlackholeA0/` with the same rule.

#50374 (merged) established the hazard for `deepseek_grouped_gate`: a baby-RISCV store can retire before its write-request reaches L1, the RISCV core and the NoC are different L1 clients with no program-order guarantee between them, and the fix is a blocking load of the last written word before the NoC is pointed at the buffer. The tree now carries that guard in nine kernels; these three senders, the most-used multicast in the tree, do not.

A second, independent path to the same symptom: the pad stores go through a non-`volatile` pointer, so the compiler is also free to schedule them after the `volatile` NoC register writes. The readback's memory clobber closes that too.

## Fix (implemented and tested)

`pad_last_ktile` and `pad_last_transposed_ktile` now end with `drain_pad_stores<format>()`: one `ckernel::load_blocking` of the tile's last word, which every padding pattern writes (right padding covers column 31 of every row, bottom padding covers row 31, the block-float variants zero the mantissa of face 3 row 15 column 15 in both cases). One header, every caller covered, including the quasar copies that include the same header. Cost: one blocking load per padded tile per block, only when `K % 32 ≠ 0`.

`pad_tile.hpp` needs `#include "ckernel.h"` for this — without it the DRAM-sharded sender (NCRISC) fails to build.

## Verified

P300, `tests/ttnn/unit_tests/operations/matmul/test_matmul.py -k test_matmul_padding`: **12 passed** (the `dram_sharded`, `multicast1d_sharded`, `multicast1d` and `reuse` configs × no / 1-face / 2-face padding).

## Scope

| | |
|---|---|
| `reader_bmm_tile_layout_in0_sender_dram_sharded.cpp:91-115`, `:160-178` | in scope, tightest window |
| `reader_bmm_tile_layout_in0_sender_padding.cpp:282-292`, `:336-368` | in scope |
| `reader_bmm_tile_layout_in0_sender_receiver_padding_block_sharded.cpp:168-215` | in scope |
| `reader_bmm_tile_layout_in0.cpp`, `reader_bmm_8bank_output_tiles_partitioned.cpp` | pad then CB push to compute, not NoC — covered by the header change anyway |
| in1 senders | do not pad in-kernel |

Reached by any `ttnn.linear` / `ttnn.matmul` with `K % 32 ≠ 0` on a multicast config. The LLM decode shapes have `K % 32 == 0`, so nothing in `tt_transformers` reaches the window today.

## Not established

- Not demonstrable by a failing test, for the reason #50374 gives: the window is a few L1-port cycles and there is no lever to delay a baby-RISCV store. Filed as the class, on the same footing as #50374.
- Wormhole: the ordering model is the same on both arch pages of `MemoryOrdering.md`, and the guard is arch-neutral, but tests were run on Blackhole only.

Hardware: P300 (Blackhole). tt-metal `83bf34eaeb4`.

kinginu:
## PR Category

Bug fix (store visibility).

## Summary

Fixes #55141, which carries the analysis and the scope.

When K is not a multiple of 32, the in0 multicast senders zero the padding of the last K tile with baby-RISCV stores and then point the NoC at that block. A store can retire before its write-request reaches L1, and the RISCV core and the NoC are different L1 clients with no program-order guarantee between them, so the multicast can publish a partially written tile.

`tt-isa-documentation`'s `TensixTile/BabyRISCV/MemoryOrdering.md` §"Enforcing stronger ordering" is explicit that emission order and processing order are different things:

> There is no _general_ mechanism to ensure that a store's write-request has been processed before a subsequent memory operation's request is emitted.

and the one construction it prescribes is a readback from the same address. `pad_last_ktile` and `pad_last_transposed_ktile` now end with `drain_pad_stores<data_format>()`, a blocking load of the tile's last word. Every padding pattern writes that word: right padding covers column 31 of every row, bottom padding covers row 31, and the block-float variants zero the mantissa of face 3 row 15 column 15 in both cases.

This is the guard #50374 established for the same hazard in `deepseek_grouped_gate`, which the tree now carries in nine kernels; these three senders, the most-used multicast in the tree, did not have it.

One header, so every caller is covered.

## Cost

One blocking load per padded tile per block, and only when `K % 32 != 0`. No path that did not already call these functions gains an instruction.

## Verified

P300 (Blackhole), this branch rebased onto `afea9f7bea1`:

`tests/ttnn/unit_tests/operations/matmul/test_matmul.py -k test_matmul_padding` — **12 passed** (the `dram_sharded`, `multicast1d_sharded`, `multicast1d` and `reuse` configs x no / 1-face / 2-face padding).

## Not verified

- **No failing test demonstrates the bug**, for the reason #50374 gives: the window is a few L1-port cycles and there is no lever to delay a baby-RISCV store. This is filed and fixed as the class, on the same footing as #50374.
- **The Quasar branch is untested.** `ckernel::load_blocking` is WH/BH only, so `drain_pad_stores` has an `#if defined(ARCH_QUASAR)` arm using a plain volatile load, the same split `data_movement/common/kernels/common.hpp` already uses. On Blackhole only the `#else` arm compiles, so the Quasar arm has not been built or run here.
- Wormhole: the ordering rule is the same on both arch pages of `MemoryOrdering.md` and the guard is arch-neutral, but tests were run on Blackhole only.

## Environment

* OS: Ubuntu 22.04
* Version of software: tt-metal `afea9f7bea1`
* Hardware: P300 (Blackhole)

### DIFF
--- a/ttnn/cpp/ttnn/operations/kernel_helper_functions/pad_tile.hpp
+++ b/ttnn/cpp/ttnn/operations/kernel_helper_functions/pad_tile.hpp
@@ -5,6 +5,13 @@
 #pragma once
 
 #include <tt-metalium/constants.hpp>
+#if !defined(ARCH_QUASAR)
+// ckernel::load_blocking (the store drain in drain_pad_stores) is WH/BH only: Quasar's ckernel.h has no
+// load_blocking, and from a data-movement build it #errors unless COMPILE_FOR_TRISC is defined. The Quasar
+// branch below uses a plain volatile load instead, so the header is not needed there. Same guard as
+// data_movement/common/kernels/common.hpp.
+#include "ckernel.h"
+#endif
 
 /**
  * @brief Pads a face within a tile.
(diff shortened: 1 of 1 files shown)


bbradelTT (approved):

bbradelTT:
@kinginu is anyone from Tenstorrent working to have these PRs approved for workflow runs and merged?

kinginu:
No, nobody. I'm an external contributor without a Tenstorrent sponsor.

bbradelTT:
Ok. Please rebase when you get the chance and I'll run workflows.

kinginu:
Thank you! rebased both #55145 and #55146 .

bbradelTT:
Sorry, please rebase the two again. Yesterday was not a good day for sanity.

kinginu:
Done.

kinginu:
@bbradelTT rebased both https://github.com/tenstorrent/tt-metal/pull/55145 and https://github.com/tenstorrent/tt-metal/pull/55146 .

bbradelTT:
Looks like rate limiting. I'll need to retry later.

PR #55162, opened 2026-09-02
Name the Blackhole L1 counters after the clients they actually count

### LINKED ISSUE #54952: Blackhole L1 counter names point at the wrong clients, and five count nothing at all
## What happens

Around thirty Blackhole L1 client names in `hw_counters.h` do not describe the interface they actually count, and five of them count nothing at all. Anyone reading an L1 counter column on Blackhole can be looking at a different client than the name suggests.

## The two problems

**Names that point at the wrong client.** Attributing grants per interface across the isolate run types (an interface busy only in `UNPACK_ISOLATE` belongs to unpack, only in `PACK_ISOLATE` to pack) disagrees with several declared names. Examples:

| declared | what the traffic says |
|---|---|
| `L1_1_RISC_CORE` | pack driven, not RISC |
| `L1_2_NOC_RING2_PORT_0..3` | unpack driven |
| `L1_3_NOC_RING3_PORT_4..6` | pack driven |
| `L1_0_UNIFIED_PACKER` | matches the unpacker port 1 counter on the Wormhole side |

The ring numbering is also wrong in principle. Blackhole has two NOCs, so names implying rings 2 and 3 cannot be describing what they claim.

**Five entries count nothing.** Confirmed against the Blackhole RTL. The L1 counter mux selects a group of eight interfaces, group four covers interfaces 32 to 39, and interfaces 35 and above are tied to zero in the two NOC configuration that Blackhole uses. That makes the last five entries of group four dead, and they currently carry names as though they were real.

Earlier captures showed non zero grants on those five, but that was the mux defect that made every group read group zero's data, not real traffic.

## Scope

- Correct the Blackhole L1 client names in `hw_counters.h`.
- Remove the five dead entries from the group four array.
- Update the two consumer sites in `tools/tracy/perf_counter_analysis.py` that refer to the old names.
- Note in the docs that these identities are measured rather than taken from a specification.

## Not in scope

Wormhole. Its sixteen interfaces are all real, there are no tied off ports, and its two groups map cleanly, so nothing needs renaming there.

## Care needed

One metric column keeps its name while changing which interface it measures. That has to be called out in the PR, since a consumer tracking it over time would see a step change with no name change to explain it. Consider renaming the column instead.

mvlahovicTT:
Fixes #54952

### What was wrong
The Blackhole L1 client names were written for a four-NOC layout the shipped part does not have. Half of them pointed at the wrong hardware and five pointed at ports that are tied off, so an L1 counter column on Blackhole could describe a different client than its name said.

### How the mapping was established
Empirically, one mux group at a time, with two workloads chosen so every slot's traffic attributes itself: a binary eltwise (two unpack operands) and a unary datacopy (one). A slot that counts only under the two-operand workload is the second unpacker; a slot that never counts is tied off.

### The renames

| ports | was | is |
|---|---|---|
| 1 | `L1_0_UNIFIED_PACKER` | `L1_0_UNPACKER_1_ECC` (unpacker 1 and the ECC scrubber; no packer on Blackhole) |
| 8 | `L1_1_RISC_CORE` | `L1_1_PACKER_IF_0` (the packer's L1 interface 0) |
| 9-11 | `L1_1_EXT_UNPACKER_1..3` | `L1_1_UNPACKER1_EXT_IF_1..3` (unpacker 1's extended read interfaces, also the packer L1-to-L1 read) |
| 16-19 | `L1_2_NOC_RING2_PORT_0..3` | `L1_2_UNPACKER1_EXT_IF_4..7` |
| 20-23 | `L1_2_NOC_RING2_PORT_4..7` | `L1_2_NOC_RING0_OUTGOING_2/3, INCOMING_2/3` |
| 24-27 | `L1_3_NOC_RING3_PORT_0..3` | `L1_3_NOC_RING1_OUTGOING_2/3, INCOMING_2/3` |
| 28-31 | `L1_3_NOC_RING3_PORT_4..7` | `L1_3_EXT_PACKER_2..5` |
| 32-34 | `L1_4_MISC_PORT_0..2` | `L1_4_EXT_PACKER_6..7, L1_4_TAG_SEARCH_PACKER_1` |
| 35-39 | `L1_4_MISC_PORT_3..7` | `L1_4_UNPACKER0_EXT_IF_1..5` (unpacker 0's extended read interfaces) |
| 40-41 | not captured | `L1_5_UNPACKER0_EXT_IF_6..7`, new group L1_5 (mux position 5, only slots 0 and 1 are wired) |

The TDMA request counters are named for what they are: `SRCA/SRCB_WRITE_REQ` and `PACKER0_DEST_READ_REQ` were `*_AVAILABLE`, and the `SRC*_WRITE_THREAD0/1` counters, which test bit 0 of the thread id, are `SRC*_WRITE_TID_EVEN/ODD`.

Blackhole-only enumerators are renamed in place; the ports Blackhole shares with Wormhole (1, 8, 9-11) get new enumerators appended at the end, so no ordinal moves and Wormhole keeps its names. The Tracy decode table is regenerated from the enum. The LLK device header can now emit L1 mux group 5 and records each bank's own reference count.

### Consumers follow the identities
- Port 1 is a second unpacker on Blackhole: it joins the unpacker port utilization and the read side of the read/write ratio there, and the packer-port metrics stay Wormhole-only (where port 1 carries pack1).
- "RISC Core L1 Util" is gone (nothing emits `L1_1_RISC_CORE` any more); "L1 TDMA Packer Port Util" reads port 8 on both arches.

### Verified on Blackhole
- Group 2 reports the full unpack stream on the renamed extended unpacker columns (1024 for a 16-tile Bfp8 operand) and zero on the NOC columns.
- Group 4 emits six columns instead of sixteen, with pack traffic on the tag-search interface and nothing fabricated.
- The LLK harness's live header parser reads the reformatted tables (16/16/16/16/6 entries).
- Rebased on main. Blackhole op sweep on the top of the stack (#55162 + #55168 + #55174) with `TT_METAL_PROFILE_PERF_COUNTERS=47` and `process_ops_logs.py --device-only`: 8/8 ops pass, 8/8 reports generated. Quasar: ttsim compile and boot and a Zebu emulator smoke (compute kernel and LLK matmul, profiler on) pass on the same build. CI: profiler pipeline https://github.com/tenstorrent/tt-metal/actions/runs/33987559586 (green on the second attempt; the first had one Wormhole N150 job time out in `test_dispatch_cores_extended_worker`, a profiler regression test unrelated to this change); sanity tests https://github.com/tenstorrent/tt-metal/actions/runs/33993143598.

|---|---|---|---|---|---|
| matmul | 1 | PASS | OK | 60.98% | 49.96% |
| binary | 28 | PASS | OK | 48.86% | 35.33% |
| tilize | 3 | PASS | OK | 61.00% | 56.92% |
| relu | 28 | PASS | OK | 65.31% | 24.56% |
| sqrt | 1 | PASS | OK | n/a | 10.70% |
| silu | 102 | PASS | OK | 47.80% | 47.34% |
| concat | 691 | PASS | OK | 65.31% | 29.21% |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mvlahovic/perf-l1-counter-names)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mvlahovic/perf-l1-counter-names)


mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34246414509

mo-tenstorrent (approved):
LGTM

Please make sure @sohaibnadeemTT has also approved this.

abhullar-tt (approved):
@mo-tenstorrent are there any profiler pipelines that should be run

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378025940
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378021570

afuller-TT (approved):

aczajkowskiTT (approved):

TT-BrianLiu (approved):

tt_metal/llrt/rtoptions.cpp
```diff
@@ -990,7 +990,7 @@ void RunTimeOptions::HandleEnvVar(EnvVarID id, const char* value) {
         case EnvVarID::TT_METAL_PROFILE_PERF_COUNTERS:
             sscanf(value, "%u", &this->profiler_perf_counter_mode);
             if (this->profiler_perf_counter_mode != 0) {
-                constexpr uint32_t L1_BITS = (1 << 3) | (1 << 4) | (1 << 6) | (1 << 7) | (1 << 8);
+                constexpr uint32_t L1_BITS = (1 << 3) | (1 << 4) | (1 << 6) | (1 << 7) | (1 << 8) | (1 << 9);
```
nvelickovicTT: Can you add a comment here to clarify what this represents?
IIUC these are possible L1 banks that can be selected depending on profiler mode, but it could use a comment to clarify.
And if L1_BITS are actually L1 banks, maybe just name them like that?

tt_metal/tools/profiler/perf_counters.hpp
```diff
@@ -285,14 +305,15 @@ constexpr uint32_t cntl_reg_for_group[9] = {
     RISCV_DEBUG_REG_PERF_CNT_L1_0,            // L1_2
     RISCV_DEBUG_REG_PERF_CNT_L1_0,            // L1_3
     RISCV_DEBUG_REG_PERF_CNT_L1_0,            // L1_4
+    RISCV_DEBUG_REG_PERF_CNT_L1_0,            // L1_5
```
nvelickovicTT: This name is a bit confusing. Comment is saying L1_5, but the name says L1_0.
If all the banks are in the same register, maybe just rename it so it doesn't cause confusion, e.g. `RISCV_DEBUG_REG_PERF_CNT_L1_x`

nvelickovicTT (approved):

sohaibnadeemTT (approved):
lgtm, but is there a reason ur adding std:: everywhere?

tt_metal/llrt/rtoptions.cpp
mvlahovicTT: renamed to L1_GROUP_BITS and added a comment

tt_metal/tools/profiler/perf_counters.hpp
mvlahovicTT: the register macro comes from the hardware header so I did not rename it, but the table now uses a named alias PERF_CNT_L1_ANY_GROUP with a comment saying that every L1 group counts through the single L1 counter block, the group only selects the mux

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34544108775
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34544106814

PR #55166, opened 2026-09-02
tracy: capture perf counter groups over multiple passes

### LINKED ISSUE #54951: tracy --profiler-capture-perf-counters=all does not fit in BRISC firmware on Blackhole
## What happens

`python -m tracy --profiler-capture-perf-counters=all` selects five counter groups at once. On Blackhole that does not fit: the BRISC firmware grows past its `.text` limit and device init fails, so the run never gets as far as producing counters.

`tech_reports/PerfCounters/perf-counters.md` also recommends the equivalent mask directly, so anyone following the document hits the same wall.

## Why

The counter config for each group is compiled into the BRISC firmware, and BRISC has a fixed `.text` budget. Measured previously on Blackhole, four groups fit and five do not, by a small margin. There is no way to fit all five in one pass.

## What it should do instead

Selecting more groups than fit should fail early with a message that tells the user how to split the capture across passes, rather than either failing at device init or silently truncating.

The multipass replay itself (running the passes and merging the logs) is a larger change and is deliberately out of scope here. An actionable error is most of the value for a fraction of the code.

## Scope

- Work out how many groups fit and reject a request for more, naming the groups and suggesting a split.
- Correct the recommended mask in `tech_reports/PerfCounters/perf-counters.md`.

## Before this lands

Re-measure the BRISC `.text` headroom on current main. The earlier measurement was taken on a branch, and if the firmware has since gained or lost room the threshold moves. The fix is the same shape either way, but the number in the error message has to be right.

mvlahovicTT:
### Ticket
Fixes #54951

### Problem
`python -m tracy --profiler-capture-perf-counters=all` cannot capture everything in one run. The BRISC firmware has room for the readout code of 3 counter groups (a 4th overflows `.text` on Blackhole), and the L1 bank counters share one count-time mux, so only one L1 bank counts per run. Blackhole has 5 L1 banks, Wormhole 2. Today `all` silently drops groups and the user schedules extra runs and merges CSVs by hand.

### What changed
- `schedule_perf_counter_passes()` splits the requested groups into passes of at most 3 groups and at most 1 L1 bank. Requests that fit stay a single run, byte-identical to before.
- `all` expands to the arch's real full set (9 groups on Blackhole, 6 on Wormhole). It needs a known architecture; if detection fails the run stops instead of guessing.
- A request that needs several passes stops and prints the pass plan. With `--perf-counter-multipass` it replays the workload once per pass, each with its own `TT_METAL_PROFILE_PERF_COUNTERS`, and merges the per-pass device logs. A pass that produces no device log fails the run rather than yielding a partial capture. Multi-pass stays opt-in: replaying the workload N times is a real cost.
- The capture guides show `all` with `--perf-counter-multipass` and the three-group limit instead of a five-group mask.
- The wrappers that already request `all` (`process_model_log`, `verify_process_ops_logs.sh`) pass the opt-in, so `run_device_profiler(..., capture_perf_counters_groups=["all"])` keeps working. Direct `--no-capture-tool` runs keep their own single-pass setup.

### Verification
- `tests/ttnn/tracy/test_perf_counter_multipass.py` covers the scheduler invariants, the bitfield mapping against `PROFILE_PERF_COUNTERS_*`, the log merge and the arch to L1-bank mapping.
- Rebased on main. Blackhole op sweep with `TT_METAL_PROFILE_PERF_COUNTERS=47` and `process_ops_logs.py --device-only`, all 8 ops pass (the binary op's report failed once inside the sweep and passed on a standalone rerun of the same tree and in the next sweep). Quasar: ttsim compile and boot with the profiler on passes. CI: profiler pipeline https://github.com/tenstorrent/tt-metal/actions/runs/33993054497; sanity tests https://github.com/tenstorrent/tt-metal/actions/runs/33993167699

|---|---|---|---|---|---|
| matmul | 1 | PASS | OK | 60.94% | 50.01% |
| binary | 28 | PASS | OK (standalone rerun) | 48.95% | 35.17% |
| tilize | 3 | PASS | OK | 60.99% | 56.92% |
| relu | 28 | PASS | OK | 62.75% | 24.63% |
| sqrt | 1 | PASS | OK | n/a | 10.65% |
| silu | 102 | PASS | OK | 47.75% | 47.34% |
| concat | 691 | PASS | OK | 65.31% | 29.23% |

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mvlahovic/perf-multipass-capture)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mvlahovic/perf-multipass-capture)


tools/tracy/__main__.py
```diff
@@ -2,13 +2,82 @@
...
 from pathlib import Path
 from shutil import copyfile
 
-
 from tracy import *
+from tracy.common import PROFILER_DEVICE_SIDE_LOG
 from tracy.serve_wasm import launch_server_subprocess, point_embed_at_trace
 
+# Bit positions match PROFILE_PERF_COUNTERS_* in tt_metal/tools/profiler/perf_counters.hpp.
+# l1_2/3/4 are Blackhole-only (its 2-NOC L1 has more client ports).
```
bbradelTT: What about l1_5? Should this be l1_2/3/4/5?

The rst and md files seem to refer up to only 4 as well.

bbradelTT (approved):

tools/tracy/__main__.py
```diff
@@ -2,13 +2,82 @@
 
 # SPDX-License-Identifier: Apache-2.0
 
+import math
 from pathlib import Path
 from shutil import copyfile
 
-
 from tracy import *
+from tracy.common import PROFILER_DEVICE_SIDE_LOG
```
mo-tenstorrent: Can you place all the post proc and aggregation of data across multiple runs into a separate script. This file should be changed for adding the cli options and executing the functions from the new script

tools/tracy/__main__.py
mvlahovicTT: changed

tools/tracy/__main__.py
mvlahovicTT: done

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34246409661

mo-tenstorrent (approved):
LGTM,

Please make sure you run a profiler CI with galaxy tests enabled

aczajkowskiTT (approved):

mvlahovicTT:
Profiler CI (with Galaxy tests enabled): https://github.com/tenstorrent/tt-metal/actions/runs/34381284308
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378030150

Issue #54235, opened 2026-08-24
[Bug] ternary: Analysis required to unblock Metal 2.0 port (program-cache key keeps only the shape volume, but the frozen per-core strides depend on the individual dims)


akerteszTT:
## Context: Metal 2.0 porting readiness

`ternary` is one of the only five ops that uses dynamic `TensorAccessorArgs`. I have been analyzing these ops to identify what `TensorParameterRelaxation` is required for the Metal 2.0 port (see https://github.com/tenstorrent/tt-metal/issues/54232). In this case, the `TensorAccessorArgs` says "dynamic shape" (which _can_ make sense for an eltwise op)... but the actual answer is indeterminate because the op seems kinda broken. 

My policy for ops bugs: 
 - they are non-blocking to a Metal 2.0 port if the port preserves the buggy behavior exactly, and 
 - blocking if the port forces a semantic change to the buggy behavior. 
  
 This one _looks_ blocking... but my analysis was relatively superficial.

**Can the ops team please vet this?** 

Claude's write-up is below. A caution: unlike the detailed (and painful) analysis of `binary_ng` (https://github.com/tenstorrent/tt-metal/issues/54138), this analysis is just a very speedy drive-by. It will most definitely need vetting.

## Porting to Metal 2.0

1. **What `TensorSpecRelaxation` does this op need?** — either one that already exists, or one I have to add
   to Metal 2.0.
2. **Does the port carry any semantic impact as a result of existing ops bugs?** My policy is that ops bugs
   are non-blocking to a port if the port preserves them intact, and blocking if the port forces a semantic
   change.

## The claim, in four verified facts

1. **The cache key coarsens each input tensor to its `padded_shape().volume()`**, plus the logical trailing
   two dims per tensor. `ternary_device_operation.cpp:556-645` — the TTT hash at `:597-614` passes
   `a_shape.volume(), b_shape.volume(), c_shape.volume(), a_h, a_w, b_h, b_w, c_h, c_w, shard_volumes`.
   The comment above the function states the intent outright: *"coarsens the input to its VOLUME (ternary is
   elementwise — program depends on tile count, not shape)."*

2. **The per-core reader strides depend on the leading dims individually**, not on their product.
   `ternary_program_factory.cpp:458-459`:
   ```cpp
   strides.n_stride = dims.Ht * dims.Wt * dims.C * (dims.N > 1);
   strides.c_stride = dims.Ht * dims.Wt * (dims.C > 1);
   ```
[...]

PR #50759, opened 2026-07-22
[ttnn] ternary: in-place override_runtime_arguments (fix get_dynamic-migration cache-hit regression)

dgomezTT:
Fixes the cache-hit dispatch regression the `get_dynamic -> override` migration introduced for ternary. **No framework or metal-API change — 2 files, ternary only.**

## Root cause
The migration gave ternary a *re-run* override: `create_descriptor()` + `apply_descriptor_runtime_args()`. That rebuilds the entire `ProgramDescriptor` (work-split + CBs + kernels + all per-core args) on **every cache hit** — ~2× the dispatch cost of the `get_dynamic` it replaced, for an op with a real work-split.

## Fix (legacy-style in-place override)
Re-apply **only the per-dispatch args**, straight into the cached program's per-core args via `GetRuntimeArgs(program, kernel)` — no rebuild, no work-split re-derivation (the cores come from the cached program itself):
- reader src addrs (slots 0/1/2, variant-aware), writer dst addr (slot 0), compute packed scalar (arg 3) — mirroring `create_descriptor`.
- Everything static (strides, tile counts, offsets) is a function of the cache-keyed shape → already correct → left untouched.

This is what legacy `shared_variables_t` overrides did; the framework already routes override ops to "override owns the re-application" (no `resolve_bindings`), so **nothing in the adapter or `ProgramDescriptor` changes.**

## Result (ternary cache-hit dispatch, exclusive box, MIN n=1000)
| re-run override (before) | legacy get_dynamic | **this (in-place)** |
|--:|--:|--:|
| 49.5 | 31.8 | **22.2** |

**~2.2× faster than the re-run override; at/below legacy.** ternary **169/169** pass; 4 different-scalar calls → **1** program-cache entry (proves cache hit, not rebuild).

## Note
This is the template for the other `get_dynamic`-migrated ops that regressed (they got the same re-run override): replace the re-run with an in-place override of the op's dynamic args. Per-op, no framework mechanism.

### CI Status
_Auto-generated on every push. Badges update live. Click a badge to filter runs by this branch._

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:dgomez/descriptor-core-partition)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:dgomez/descriptor-core-partition)
[...]

PR #55167, opened 2026-09-02
#54235: Fix ternary runtime arguments on program-cache hits

KalaivaniMCW:
### PR Category

Bug fix
### Summary

#54235
### Notes for reviewers

- Addressing Ternary cache issue and gaps
- `compute_program_hash` coarsens each input tensor to its padded VOLUME, so one cached program is shared across calls whose shapes differ in their factorisation -- e.g. a predicate of [4,1,32,32] and one of [1,4,32,32] hash identically.  Those two shapes need different reader strides (n_stride / c_stride are functions of the individual dims, not of their product), and on a cache hit the descriptor is NOT rebuilt leading to frozen args.
- This PR change rebuilds all shape/work-split-dependent runtime arguments and repoints sharded circular buffers to make volume-key cache collisions safe.
- Re-apply ALL per-dispatch state to the cached program on a program-cache hit.  No descriptor rebuild and no kernel recompile -- but every per-core arg IS re-derived, because compute_program_hash coarsens each input to its padded VOLUME and therefore shares one cached program across calls whose shapes differ in their factorization.  The reader strides are a function of the individual dims (n_stride = Ht x Wt x C x (N>1), c_stride = Ht x Wt x (C>1)), not of the product the key retains, so for ex: predicates of [4,1,32,32] and [1,4,32,32] hash the same but need different strides; re-applying only the buffer addresses left those frozen at the first-miss shape and silently corrupted the second call - We re-derive them from the SAME shared builder create_descriptor() uses, so the two paths stay byte-identical by construction.
**What fails without the fix:** 
Pre-fix these tests would fail due to stale cache (test results on current main):

image

image

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:kalaivani/ternary_port)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:kalaivani/ternary_port)


tests/ttnn/unit_tests/operations/eltwise/test_where.py
```diff
@@ -1180,3 +1180,71 @@ def test_where_program_cache_different_broadcast_shapes(device, variant):
...
+
+@pytest.mark.parametrize(
+    "pred_shape_1, pred_shape_2",
+    [
+        ((4, 1, 32, 32), (1, 4, 32, 32)),  # N/C swap, equal volume
+        ((2, 2, 1, 32, 32), (2, 1, 2, 32, 32)),  # 5D: D/N/C permutation
+        ((2, 2, 1, 32, 32), (1, 2, 2, 32, 32)),
+    ],
+)
+def test_where_program_cache_same_volume_different_leading_dims(device, pred_shape_1, pred_shape_2):
```
Aswinmcw: TTT only. `test_where_program_cache_different_broadcast_shapes` directly above parametrizes over `["TTT", "TTS", "TST"]` — please do the same here. TTS/TST are the variants this rework changes most: their reader arg layout differs (buffers at slots 0/1, slot 2 a plain scalar), and the packed scalar from `pack_compute_scalar_arg` now flows through the shared builder and is rewritten on every hit rather than patched at one fixed index. Neither new test exercises that path.

tests/ttnn/unit_tests/operations/eltwise/test_where.py
```diff
@@ -1180,3 +1180,71 @@ def test_where_program_cache_different_broadcast_shapes(device, variant):
...
+        expected = torch.where(pred.bool(), true_t, false_t)
+
+        pred_tt = ttnn.from_torch(pred, device=device, layout=ttnn.TILE_LAYOUT, dtype=ttnn.bfloat16)
+        true_tt = ttnn.from_torch(true_t, device=device, layout=ttnn.TILE_LAYOUT, dtype=ttnn.bfloat16)
+        false_tt = ttnn.from_torch(false_t, device=device, layout=ttnn.TILE_LAYOUT, dtype=ttnn.bfloat16)
+        result = ttnn.to_torch(ttnn.where(pred_tt, true_tt, false_tt))
+        results.append((result, expected))
+
+    for result, expected in results:
+        assert_equal(expected, result)
```
Aswinmcw: Neither test asserts the two calls actually shared a cache entry. Capturing `device.num_program_cache_entries()` before/after and asserting it stayed at 1 would make these self-verifying: if the hash is ever widened so the shapes no longer collide, both dispatches become independent, both pass, and the regression test silently stops covering the regression.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -1403,55 +1433,73 @@ void TernaryDeviceOperation::TernaryProgramFactory::override_runtime_arguments(
...
-    const uint32_t pred_addr = predicate_tensor.buffer()->address();
-    // 2nd src operand: the tensor operand (TST uses value_false; TTS/TTT use value_true).
-    const auto& src1_tensor = (variant == TernaryVariant::TST) ? value_false_tensor : value_true_tensor;
-    const uint32_t src1_addr = src1_tensor.value().buffer()->address();
-    const bool has_src2 = (variant == TernaryVariant::TTT);  // only TTT has a 3rd tensor operand
-    const uint32_t src2_addr = has_src2 ? value_false_tensor.value().buffer()->address() : 0u;
-    const uint32_t out_addr = output.buffer()->address();
-    const uint32_t scalar_arg = CMAKE_UNIQUE_NAMESPACE::pack_compute_scalar_arg(operation_attributes, output.dtype());
+    auto per_core = CMAKE_UNIQUE_NAMESPACE::build_per_core_runtime_args(
+        operation_attributes, tensor_args, output, operation_attributes.broadcast_type);
```
Aswinmcw, line 1558: This runs the whole `compute_core_partition` + `extract_tensor_dimensions` pass and heap-allocates three `std::vector<std::variant<uint32_t, Buffer*>>` per core on every cache hit — ~192 allocations on a 64-core grid, against the handful of scalar writes this replaces. #50759 moved that work off the dispatch path deliberately; re-deriving the args is right, but paying for a variant on all 27 reader slots when only 3 need a `Buffer*` is not. Returning `std::vector<uint32_t>` plus a small (slot index → `Buffer*`) fixup list per kernel gets the same correctness with a flat copy. Some before/after dispatch numbers would help either way.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -1403,55 +1433,73 @@ void TernaryDeviceOperation::TernaryProgramFactory::override_runtime_arguments(
...
+    auto apply = [&](uint32_t kernel_idx,
+                     const CoreCoord& core,
+                     const std::vector<std::variant<uint32_t, tt::tt_metal::Buffer*>>& args) {
+        auto& data = tt::tt_metal::GetRuntimeArgs(program, kernel_idx, core);
+        for (uint32_t arg_idx = 0; arg_idx < static_cast<uint32_t>(args.size()); ++arg_idx) {
+            const auto& slot = args[arg_idx];
+            data[arg_idx] = std::holds_alternative<tt::tt_metal::Buffer*>(slot)
+                                ? static_cast<uint32_t>(std::get<tt::tt_metal::Buffer*>(slot)->address())
+                                : std::get<uint32_t>(slot);
         }
```
Aswinmcw: Unchecked write into `data`. The code this replaces guarded every store (`if (a.size() > 1)`, `if (a.size() > kScalarArgIdx)`). Sizes do match today because `num_reader_args`/`num_writer_args`/`num_kernel_args` are `constexpr` and both paths fill all of them, but that's now an invariant spanning two functions. A `TT_FATAL(data.size() == args.size(), ...)` before the loop turns any future divergence into a loud failure instead of an out-of-bounds write.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -1403,55 +1433,73 @@ void TernaryDeviceOperation::TernaryProgramFactory::override_runtime_arguments(
...
+    // function of the individual dims (n_stride = Ht*Wt*C*(N>1), c_stride = Ht*Wt*(C>1)), not of the
+    // product the key retains, so predicates of [4,1,32,32] and [1,4,32,32] hash the same but need
+    // different strides; re-applying only the buffer addresses left those frozen at the first-miss
+    // shape and silently corrupted the second call (issue #54235).  We re-derive them from the SAME
+    // shared builder create_descriptor() uses, so the two paths stay byte-identical by construction.
+    //
+    // Kernel push order in create_descriptor(): reader(0), writer(1), compute(2).  The work-core
+    // partition shifts with the output volume (a core can flip between work and noop), so the builder
+    // emits args for ALL cores; we re-apply every one, buffer-address slots included, so a core
+    // promoted to a work core on this hit is never left with a stale base address.
```
Aswinmcw: The issue lists "whether the accessor *common* runtime args are refreshed on the sharded path" as unresolved. `reader_desc.common_runtime_args` / `writer_desc.common_runtime_args` are filled from `TensorAccessorArgs` at descriptor-build time and this function never touches `GetCommonRuntimeArgs`, so they stay at the first-miss tensors. That may well be safe (input memory configs are hashed), but the reasoning isn't recorded anywhere — worth one line here stating that the accessor common args are a pure function of hashed state, or flagging it as a remaining gap.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -1403,55 +1433,73 @@ void TernaryDeviceOperation::TernaryProgramFactory::override_runtime_arguments(
...
+        const auto& indices = cb->buffer_indices();
+        if (indices.contains(static_cast<uint8_t>(tt::CBIndex::c_0)) && pred_buffer != nullptr) {
+            tt::tt_metal::UpdateDynamicCircularBufferAddress(program, cb->id(), *pred_buffer);
+        } else if (indices.contains(static_cast<uint8_t>(tt::CBIndex::c_1)) && src1_buffer != nullptr) {
+            tt::tt_metal::UpdateDynamicCircularBufferAddress(program, cb->id(), *src1_buffer);
+        } else if (indices.contains(static_cast<uint8_t>(tt::CBIndex::c_2)) && src2_buffer != nullptr) {
+            tt::tt_metal::UpdateDynamicCircularBufferAddress(program, cb->id(), *src2_buffer);
+        } else if (indices.contains(static_cast<uint8_t>(tt::CBIndex::c_3)) && out_buffer != nullptr) {
+            tt::tt_metal::UpdateDynamicCircularBufferAddress(program, cb->id(), *out_buffer);
         }
```
Aswinmcw, line 1629: This block is new behaviour, not a refactor of anything that was here, and it only does something when a tensor is sharded (`CBDescriptor::buffer` is `nullptr` otherwise). Both new tests are interleaved/DRAM, so nothing covers it — a sharded `where` that hits the cache with a different buffer allocation is the case it exists for. Also, the `else if` chain stops at the first match: a CB carrying more than one of c_0..c_3 in `buffer_indices()` gets only its lowest index re-pointed. If that combination can't occur, a `TT_FATAL` or a comment saying so is better than the chain implying it can.

Aswinmcw (commented):
The description ("Bug fix", "#54235", "Addressing Ternary cache issue and gaps") doesn't say which of the issue's open points this closes. Two things worth writing down there: (1) the mechanism in one sentence — the reader strides are a function of the individual leading dims while the key only retains their product, so a cache hit replayed the first-miss strides; (2) the issue asks ops for an explicit decision on whether the volume-coarsening in `compute_program_hash` is intended. This PR answers it in code (keep the key coarse, re-derive everything on hit) but not in prose, and that answer is what unblocks the Metal 2.0 `TensorSpecRelaxation` call. Notes inline.

tests/ttnn/unit_tests/operations/eltwise/test_where.py
KalaivaniMCW: addressed

tests/ttnn/unit_tests/operations/eltwise/test_where.py
KalaivaniMCW: I will move cache-related tests to nightly

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
KalaivaniMCW: Replaced variant with flat copy. Small-shape regression cut from +34.6 µs to +9.7 µs. no significant changes on bigger multi-tile shapes.
image

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
KalaivaniMCW: `apply` lambda function is added to validate this.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
KalaivaniMCW: added a detailed comment reg this

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
KalaivaniMCW: the sharded test already added nightly ternary_sharding were tested and verified.

KalaivaniMCW: will check if it needs additional testing

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -636,17 +636,84 @@ uint32_t pack_compute_scalar_arg(
...
+    const uint32_t* compute_row(size_t i) const { return compute.data() + i * kNumComputeArgs; }
+};
+
+// SINGLE SOURCE OF TRUTH for ternary per-core runtime args.  Run by BOTH create_descriptor()
+// (cache miss) and TernaryProgramFactory::override_runtime_arguments() (cache hit).
+//
+// compute_program_hash coarsens each input tensor to its padded VOLUME, so one cached program is
+// shared across calls whose shapes differ in their factorization -- e.g. a predicate of [4,1,32,32]
+// and one of [1,4,32,32] hash identically.  Those two need different reader strides (n_stride /
+// c_stride are functions of the individual dims, not of their product), and on a cache hit the
```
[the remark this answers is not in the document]
KalaivaniMCW: updated

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -636,17 +636,84 @@ uint32_t pack_compute_scalar_arg(
...
+// descriptor is NOT rebuilt.  So every shape-/work-split-dependent per-core arg (strides, D/N/C/Ht/Wt,
+// per-core tile counts, c_start_id, freq/counter, packed scalar, ...) would otherwise stay frozen at
+// the first-miss shape and silently corrupt the result (issue #54235).  override_runtime_arguments()
+// re-runs THIS builder for the current tensors and re-applies every arg.
```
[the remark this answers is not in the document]
KalaivaniMCW: Earlier was the initial analysis phase, now it is updated

tests/ttnn/nightly/unit_tests/operations/eltwise/test_ternary_bcast.py
```diff
@@ -192,3 +192,107 @@ def test_ttnn_lerp_ttt_row_col_mixed_bcast(input_shape, end_shape, weight_shape,
...
+
+        result, expected = _where_variant(device, variant, pred, true_t, false_t)
+        results.append((result, expected))
+        if entries_after_first is None:
+            entries_after_first = device.num_program_cache_entries()
+
+    # The two dispatches must actually SHARE a cache entry, or this test is not exercising the
+    # regression at all: if the hash is ever widened so these shapes stop colliding, both calls
+    # become independent, both pass, and the coverage silently disappears.
+    assert device.num_program_cache_entries() == entries_after_first, (
```
Aswinmcw: This guards "the second call did not add an entry", but it also passes when *nothing* was cached — if the program cache is ever off for this run, `entries_after_first` is 0, the count stays 0, and the test goes green without having exercised a cache hit at all. Asserting `entries_after_first == 1` (or at least `> 0`) before this makes it non-vacuous. The `device.cache_entries_counter` helper from `tests/tests_common/cache_entries_counter.py` gives both halves directly: `measure()` around call 1 expecting +1, around call 2 expecting +0.

tests/ttnn/nightly/unit_tests/operations/eltwise/test_ternary_bcast.py
```diff
@@ -192,3 +192,107 @@ def test_ttnn_lerp_ttt_row_col_mixed_bcast(input_shape, end_shape, weight_shape,
...
+@pytest.mark.parametrize("variant", ["TTT", "TTS", "TST"])
+@pytest.mark.parametrize(
+    "pred_shape_1, pred_shape_2",
+    [
+        ((4, 1, 32, 32), (1, 4, 32, 32)),  # N/C swap, equal volume
+        ((2, 2, 1, 32, 32), (2, 1, 2, 32, 32)),  # 5D: D/N/C permutation
+        ((2, 2, 1, 32, 32), (1, 2, 2, 32, 32)),
+    ],
+)
+def test_where_program_cache_same_volume_different_leading_dims(device, variant, pred_shape_1, pred_shape_2):
```
Aswinmcw: Worth confirming in the PR body that these fail with the fix reverted — the mechanism is order- and stride-dependent, so a shape choice that happens to produce the same strides would pass on pre-fix main and the test would look like coverage it is not.

ttnn/cpp/ttnn/operations/eltwise/ternary/device/ternary_program_factory.cpp
```diff
@@ -1403,55 +1528,94 @@ void TernaryDeviceOperation::TernaryProgramFactory::override_runtime_arguments(
...
+    // shared builder create_descriptor() uses, so the two paths stay byte-identical by construction.
+    //
+    // Kernel push order in create_descriptor(): reader(0), writer(1), compute(2).
+    //
+    // NOT refreshed here: the reader/writer COMMON runtime args, which carry the TensorAccessorArgs
+    // words.  On interleaved tensors that is vacuous -- TensorAccessorArgs::update_args_config()
+    // resets the config to None for any buffer without a BufferDistributionSpec and re-sets only
+    // IsDram, so the requested RuntimeTensorShape is discarded and no shape words are emitted at all.
+    // On the SHARDED path the words ARE emitted and nothing re-applies them, so they stay at the
+    // first-miss tensors. The shard specs feed compute_program_hash via get_shard_volumes()
```
Aswinmcw: This sentence is cut off mid-thought — "The shard specs feed compute_program_hash via get_shard_volumes()" and then nothing. The conclusion it was heading for is the whole point of the paragraph (whether the stale sharded accessor words are safe because the shard specs are hashed, or a remaining gap), so it is worth finishing.

Aswinmcw (approved):
LGTM — the shared builder is the right shape for this, and both of my earlier test points are covered. Three nits below, none blocking. Copilot's point about the stale cache-hit contract on `ternary_device_operation.hpp:54-56` looks still open.

tests/ttnn/nightly/unit_tests/operations/eltwise/test_ternary_bcast.py
KalaivaniMCW: Pre-fix these tests would fail
image

PR #55168, opened 2026-09-02
perf counters: one metric engine for the Tracy tool and the LLK harness

### LINKED ISSUE #54953: Perf counter metric formulas are duplicated three times and have started to drift
## The problem

The perf counter metric formulas exist in three hand written copies:

1. `compute_perf_counter_metrics` in `tools/tracy/perf_counter_analysis.py`
2. `compute_device_only_metrics` in the same file, over a differently shaped frame
3. `_compute_single` in `tt_metal/tt-llk/tests/python_tests/helpers/metrics.py`

The same file also carries a hand maintained counter name table and a hand maintained CSV column list, both of which have to track the enum by hand.

This is not a tidiness complaint. The duplication has already produced a bug: the scoreboard stall metric is guarded in one of the two copies inside `perf_counter_analysis.py` and unguarded in the other, which is issue #54949. A formula that lives in three places will keep drifting in three directions.

## What the change does

Put the formulas in one module that both consumers call through a small adapter, so each consumer only has to say how to fetch a counter and a cycle count for its own data shape. Derive the counter name table and the CSV column list from the enum instead of hand maintaining them.

The draft of this work is in #50210. Landing it deletes roughly two thousand lines of duplicated logic against about seven hundred added, so it is a net reduction.

## What has to be fixed before it can land

The draft does not apply as written. Each of these is small on its own:

- The LLK report schema is not grown to match. The engine emits many more keys than `METRIC_BASES` knows about, and the strict CSV to Parquet converter rejects an unknown column, so any counter run fails at the end. `METRIC_BASES` and the dropped column set have to change in the same commit.
- The draft reintroduces `math_src_stall_pct`, which was removed deliberately because its numerator and denominator count different populations and it read a constant on one of the two markers. It has to stay out.
- The module is placed in a directory that holds no other Python and is reached by walking parent directories at import time. It belongs in the existing profiler Python package, imported normally.
- The draft also carries a runtime L1 mux selector that conflicts with the compile time model now on main, and an unbounded register poll that the current code bounds. Both should be dropped rather than ported.

One idea in the draft is worth keeping beyond the consolidation itself: separating bounded percentages from unbounded ratios, which fixes a stall overlap metric that can currently read well over 100%.

## What has to be measured before review

Two things, both about the effect on consumers rather than the design:

- **Column set.** The published CSV changes from 336 columns to 424, with about 52 renamed or removed and the order of the shared columns not preserved. Anything parsing that file positionally or asserting on a schema will break. This needs a column by column value diff between main and the change, captured on the same workload, published in the PR.
- **Runtime.** The draft replaces vectorised pandas work with a Python loop per operation per core. On a large trace that could be much slower. This needs timing on a realistic capture before anyone can judge the tradeoff.

Also worth deciding: one metric column keeps its name while measuring a different hardware interface. Either rename the column or call the change out prominently.

## Suggested order

Land the smaller fixes first, since several of them are prerequisites and all of them are independently useful:

- #54949 scoreboard stall guard
- #54950 remove the tied off fidelity counters
- #54951 reject a counter group selection that does not fit
- #54952 correct the Blackhole L1 client names

Then this one, with the schema grown in the same change.

mvlahovicTT:
### Ticket
Fixes #54953. Stacked on #55162 (its base) because the shared module uses the corrected Blackhole L1 names.

### Problem
The metric formulas existed in three hand-written copies: the per-op and device-only paths in `tools/tracy/perf_counter_analysis.py` and `_compute_single` in the LLK harness. They had drifted: different numerators for the same metric, different denominators, different clipping.

### What changed
- New `tools/tracy/perf_metrics_common.py` holds every formula, the L1 client-port groupings, the labels and the percentage/ratio classification. Consumers adapt their data to a three-method `CounterView` (count, cycles, has) and call `compute_metrics()`.
- `perf_counter_analysis.py` shrinks by about 1900 lines; both Tracy paths go through the engine and the counter name table is parsed from the `PerfCounterType` enum. LLK `helpers/metrics.py` becomes a thin adapter and `perf/schema.py` derives `METRIC_BASES` from the module.
- Metric keys split into `*_pct` (bounded) and `*_ratio` (unbounded, shown with a `(ratio)` unit). `math_src_stall_pct` stays out: its numerator and denominator count different populations.
- Arch-divergent metrics gate on `CounterView.has()` and read N/A instead of a fake value; the scoreboard-stall guard from #54954 is covered by the same gating.
- "L1 Packer Port Util" / "Backpressure" become "L1 Port 1 Util" / "Backpressure": port 1 is a different interface per arch (pack1 on Wormhole, unpacker 1 on Blackhole), so it is reported on its own; the read/write split places it by arch through `CounterView.is_blackhole()`.
- Partial captures read N/A, not 0: cross-bank formulas (FPU per issued instruction, compute-to-unpack, L1-vs-compute, the L1_0 composites) gate on the counters they need. Packer load imbalance counts idle engines as zero instead of dropping them, and the dest-write-port stall stays N/A while its counter never ticks (the known signal-mismatch case).
- Metrics with no value get no CSV column, so the two-pass L1 merge can still find the second pass by column absence. A shipped `perf_counter_type_names.json` (guarded by a test against the header) keeps decoding working in installed wheels.

### Measurements
Per-op CSV: four columns per metric of the engine's `METRIC_LABELS`. Removed: HiFi Fraction, Avg HF Cycles Per Instrn (tied-off counters, #54958), Math Src Data Ready Rate. Unit changes: Math-to-Pack Handoff, Unpacker L1 Efficiency and Compute-to-Unpack Ratio are now `(ratio)` columns (old value / 100); T0/T1/T2 Instrn Issue Rate are percentages (old value x 100). Fixes: Stall Overlap denominator, SrcB Write Actual and Unpacker Write Efficiency numerator, no clipping of Data Hazard and Write Blocked rates, Unpacker-to-Math Data Flow as the mean of the two per-unpacker ratios. 28 (WH) / 32 (BH) columns added.

Runtime at 300 ops x 121 cores (5.6M counter rows): 13.8 s against 76.8 s before. On a real Blackhole LLK capture all 16 kept LLK metrics are bitwise identical between the old and new code.

### Verification
- `tests/ttnn/tracy/test_perf_metrics_common.py`: key/label parity, None on absent counters, percentage bounds, presence gating.
- Docs are updated in #55174, stacked on this PR.
- Rebased on the updated #55162. Blackhole op sweep on the top of the stack (#55162 + #55168 + #55174) with `TT_METAL_PROFILE_PERF_COUNTERS=47` and `process_ops_logs.py --device-only`: 8/8 ops pass, 8/8 reports generated. Quasar: ttsim compile and boot and a Zebu emulator smoke (compute kernel and LLK matmul, profiler on) pass on the same build. CI: profiler pipeline https://github.com/tenstorrent/tt-metal/actions/runs/33993094294; sanity tests https://github.com/tenstorrent/tt-metal/actions/runs/33993192845; LLK perf https://github.com/tenstorrent/tt-metal/actions/runs/33993218168

|---|---|---|---|---|---|
| matmul | 1 | PASS | OK | 60.98% | 49.96% |
| binary | 28 | PASS | OK | 48.86% | 35.33% |
| tilize | 3 | PASS | OK | 61.00% | 56.92% |
| relu | 28 | PASS | OK | 65.31% | 24.56% |
| sqrt | 1 | PASS | OK | n/a | 10.70% |
| silu | 102 | PASS | OK | 47.80% | 47.34% |
| concat | 691 | PASS | OK | 65.31% | 29.21% |

### Follow-ups from the counter audit
- Every INSTRN wait rate, the stall overlap ratios and the per-thread shares read N/A when their counter was not captured (Quasar has none of the `WAITING_FOR_*` counters) instead of a fake 0%.
- `math_scoreboard_stall_pct` (TDMA_PACK over TDMA_UNPACK) is clamped at 0; `math_dest_wr_port_stall_pct` never differed from its denominator on real data and is dropped.
- The extended unpacker L1 group is split into unpacker 0 (banks 4-5) and unpacker 1 (banks 1-2); port 1 and port 8 take their per-arch names.

### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:mvlahovic/perf-metrics-engine)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:mvlahovic/perf-metrics-engine)


bbradelTT (approved):

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378043919
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34378038931

aczajkowskiTT (approved):

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34398410272
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34398405833

TT-BrianLiu (approved):

docs/source/ttnn/ttnn/profiling_ttnn_operations.rst
```diff
@@ -197,12 +197,8 @@ The following metrics are automatically computed from raw counters. Each metric
 *TDMA Stall Metrics*
 
 - **Data Hazard Stall Rate (%)**: Fraction of math-valid cycles stalled by dest-to-src data hazards (MOVD2A/MOVD2B). Computed as ``(MATH_INSTRN_AVAILABLE - MATH_NOT_D2S_STALLED) / MATH_INSTRN_AVAILABLE``.
-- **SrcA Write Port Blocked Rate (%)**: Fraction of srcA DMA write attempts blocked by port unavailability (DMA mux contention).
-- **SrcB Write Port Blocked Rate (%)**: Same for srcB.
+- **SrcB Write Port Blocked Rate (%)**: Fraction of srcB DMA write attempts blocked by port unavailability (DMA mux contention).
```
nvelickovicTT, line 200: Is this intentional to remove SrcA Write Port Blocked rate? Sicne it seems like an outlier to have something for SrcB but not SrcA, it would be good to have a comment explaining it.

nvelickovicTT (approved):

tt_metal/tt-llk/tests/python_tests/helpers/perf/schema.py
```diff
@@ -110,26 +110,35 @@ def assert_unique_columns(columns, context: str = "") -> None:
...
+            spec = importlib.util.spec_from_file_location(
+                "perf_metrics_common", candidate
+            )
+            module = importlib.util.module_from_spec(spec)
+            spec.loader.exec_module(module)
+            return module
+    raise RuntimeError(
+        "Could not locate tools/tracy/perf_metrics_common.py above this file"
+    )
+
```
sstanisicTT, line 138: isn't there a nasty failure mode here:
- Somebody changes the Path in metal
- The LLK CI doesn't trigger because the change is not in tt-llk
- PR in metal is merged
- Following PRs fail in CI because the python lib can't be imported.

tools/tracy/process_ops_logs.py
```diff
@@ -8,44 +8,44 @@
 #!/usr/bin/env -S python3 -m pdb
 
 import ast
-import os
```
mo-tenstorrent, line 11: Can we revert the import re order here.

mo-tenstorrent (approved):
Please make sure profiler CI is green.

Also please wait until @sohaibnadeemTT has also approved.

docs/source/ttnn/ttnn/profiling_ttnn_operations.rst
mvlahovicTT: it is intentional because the engine drops complement duplicates: srcA port figure is reported as SrcA Write Actual Efficiency and srcB overwrite figure as SrcB Write Actual Efficiency. added a line saying so

tt_metal/tt-llk/tests/python_tests/helpers/perf/schema.py
mvlahovicTT: yea :/ added a case to find-changed-files.sh so a change to tools/tracy/perf_metrics_common.py sets the LLK tests and Quasar flags, so the LLK suites run on such a PR and the fallback error also names the file

tools/tracy/process_ops_logs.py
mvlahovicTT: reverted

mvlahovicTT:
Profiler CI: https://github.com/tenstorrent/tt-metal/actions/runs/34544112945
Sanity CI: https://github.com/tenstorrent/tt-metal/actions/runs/34544110957

mvlahovicTT:
closing this one. the metric engine moved to #56320, which keeps the same single engine but puts the counter definitions and the engine itself in tt-llk, with the metal profiler consuming them through the tt_llk_perf package. that removes the header dependency that pointed the wrong way. #54953 is linked on the new pr.

PR #55170, opened 2026-09-02
P1 0tr/downed links pt1

p1-0tr:
### CI Status

- [](https://github.com/tenstorrent/tt-metal/actions/workflows/sanity-tests.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/runtime-sanity-tests.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/tt-metal-l2-nightly.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/all-model-tests.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-t3k.yaml?query=branch:p1-0tr/downed-links-pt1)
- [](https://github.com/tenstorrent/tt-metal/actions/workflows/pipeline-select-galaxy.yaml?query=branch:p1-0tr/downed-links-pt1)


tt_metal/fabric/physical_node_id.cpp
```diff
@@ -0,0 +1,97 @@
...
+std::string_view strip_rank_suffix(std::string_view host_id) {
+    const std::size_t underscore = host_id.find_last_of('_');
+    if (underscore == std::string_view::npos || underscore + 1 == host_id.size()) {
+        return host_id;
+    }
+    const std::string_view suffix = host_id.substr(underscore + 1);
+    const bool all_digits = std::all_of(
+        suffix.begin(), suffix.end(), [](const char c) { return std::isdigit(static_cast<unsigned char>(c)) != 0; });
+    return all_digits ? host_id.substr(0, underscore) : host_id;
+}
```
Riddy21: Something makes me feel like we should do this for all PSD, that even if we're running on different ranks we should have some mechanism to differentiate bewteen differetn ranks on the same host, but still keep their hostname the same. Not sure if its worth the effort?

tt_metal/fabric/physical_system_discovery.cpp
```diff
@@ -56,8 +56,18 @@ PortType to_metal_port_type(tt::scaleout_tools::PortType pt) {
...
+std::string get_local_discovery_hostname(const tt::umd::ClusterDescriptor& cluster_desc) {
+    // host_id names the group of accelerators behind a common host / controller / root complex. UMD
+    // stamps it from $TT_HOST_ID or gethostname() during live discovery, and a captured descriptor
+    // carries the id of the machine it came from. Preferring it is what lets a mock run key on the
+    // same string as the factory system descriptor instead of on an asset filename, which is the
+    // only reason the two agree on a chip's address. The field is optional, so the fallbacks below
+    // stay for descriptors that predate it.
+    if (const auto& host_id = cluster_desc.get_host_id(); host_id.has_value()) {
+        return *host_id;
+    }
```
Riddy21, line 70: Need to make sure we can differentiate between different ranks on the same host here as well. not sure if you handle this in UMD

tt_metal/fabric/physical_node_id.cpp
p1-0tr: Yep, stripping the suffix does not make much sense. Removed.

tt_metal/fabric/physical_system_discovery.cpp
p1-0tr: I removed the suffix dropping logic. I don't think we should make UMD rank aware. The way I think this should be handled is by passing `TT_HOST_ID` (though I'm more and more convinced this should be named `TT_CLUSTER_ID`) along with `TT_VISIBLE_DEVICES`, containing whatever our launch SW deems sensible.

I'll also extended the host_id field to hold 128 characters, like that we can hold `hostname + rank_id` with no issues. The reason to go straight to 128 chars is that this would accommodate any commonly used hash represented as hex string.

Riddy21:
Superseded by #54018.

The whole stack (#54018 → #55170 → #55459) has been squashed into a single commit on `ridvan/downed-links-contract` and rebased onto current `main` (513 commits). Everything from both parts is in there — `PhysicalNodeId`, the `cluster_id` keying, `diff_physical_system_descriptors`, `LinkHealth`, the FSD host filter, the ControlPlane wiring, and the test-driver fixes.

Submodules are now on their upstream `main` tips rather than feature branches: umd `de81ff1d` (tenstorrent/tt-umd#3349 is merged) and tt-cluster-descriptors `e3c31f2` (#18 and #20 are merged).

Closing this one; review continues on #54018.

